PR #472, opened 2018-01-31
Introduction of a detach handler at links in Hono Client.

pellmann:
The client application can react on a detach(close=false) and reopen the link later (retry).
In future this should be handled inside the Hono client as well as the connection retry.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/HonoClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/HonoClient.java
@@ -117,8 +117,24 @@ public interface HonoClient {
      * @param tenantId The tenant to consume data for.
      * @param telemetryConsumer The handler to invoke with every message received.
      * @param creationHandler The handler to invoke with the outcome of the operation.
+     * @param detachHandler The handler invoked on detached (not closed) link.
      * @throws NullPointerException if any of the parameters is {@code null}.
      */
+    void createTelemetryConsumer(
+            String tenantId,
+            Consumer<Message> telemetryConsumer,
+            Handler<AsyncResult<MessageConsumer>> creationHandler,
+            Handler<AsyncResult<Void>> detachHandler);
+
+    /**
+     * Creates a new consumer of telemetry data for a tenant.
+     *
+     * @param tenantId The tenant to consume data for.
+     * @param telemetryConsumer The handler to invoke with every message received.
+     * @param creationHandler The handler to invoke with the outcome of the operation.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     */
+    @Deprecated
     void createTelemetryConsumer(
             String tenantId,
             Consumer<Message> telemetryConsumer,
@@ -134,8 +150,28 @@ public interface HonoClient {
      * @param tenantId The tenant to consume events for.
      * @param eventConsumer The handler to invoke with every event received.
      * @param creationHandler The handler to invoke with the outcome of the operation.
+     * @param detachHandler The handler invoked on detached (not closed) link.
      * @throws NullPointerException if any of the parameters is {@code null}.
      */
+    void createEventConsumer(
+            String tenantId,
+            Consumer<Message> eventConsumer,
+            Handler<AsyncResult<MessageConsumer>> creationHandler,
+            Handler<AsyncResult<Void>> detachHandler);
+
+    /**
+     * Creates a new consumer of events for a tenant.
+     * <p>
+     * The events passed in to the registered eventConsumer will be settled
+     * automatically if the consumer does not throw an exception and does not
+     * manually handle the message disposition using the passed in delivery.
+     *
+     * @param tenantId The tenant to consume events for.
+     * @param eventConsumer The handler to invoke with every event received.
+     * @param creationHandler The handler to invoke with the outcome of the operation.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     */
+    @Deprecated
     void createEventConsumer(
             String tenantId,
             Consumer<Message> eventConsumer,
@@ -151,8 +187,28 @@ public interface HonoClient {
      * @param tenantId The tenant to consume events for.
      * @param eventConsumer The handler to invoke with every event received.
      * @param creationHandler The handler to invoke with the outcome of the operation.
+     * @param detachHandler The handler invoked on detached (not closed) link.
      * @throws NullPointerException if any of the parameters is {@code null}.
      */
+    void createEventConsumer(
+            String tenantId,
+            BiConsumer<ProtonDelivery, Message> eventConsumer,
+            Handler<AsyncResult<MessageConsumer>> creationHandler,
+            Handler<AsyncResult<Void>> detachHandler);
+
+    /**
+     * Creates a new consumer of events for a tenant.
+     * <p>
+     * The events passed in to the registered eventConsumer will be settled
+     * automatically if the consumer does not throw an exception and does not
+     * manually handle the message disposition using the passed in delivery.
+     *
+     * @param tenantId The tenant to consume events for.
+     * @param eventConsumer The handler to invoke with every event received.
+     * @param creationHandler The handler to invoke with the outcome of the operation.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     */
+    @Deprecated
     void createEventConsumer(
             String tenantId,
             BiConsumer<ProtonDelivery, Message> eventConsumer,
--- a/client/src/main/java/org/eclipse/hono/client/HonoClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/HonoClient.java
@@ -117,14 +117,14 @@ public interface HonoClient {
      * @param tenantId The tenant to consume data for.
      * @param telemetryConsumer The handler to invoke with every message received.
      * @param creationHandler The handler to invoke with the outcome of the operation.
-     * @param detachHandler The handler invoked on detached (not closed) link.
+     * @param closeHandler The handler invoked on detached link (detach with close=true and close=false).
      * @throws NullPointerException if any of the parameters is {@code null}.
      */
     void createTelemetryConsumer(
             String tenantId,
             Consumer<Message> telemetryConsumer,
             Handler<AsyncResult<MessageConsumer>> creationHandler,
-            Handler<AsyncResult<Void>> detachHandler);
+            Handler<AsyncResult<Void>> closeHandler);
 
     /**
      * Creates a new consumer of telemetry data for a tenant.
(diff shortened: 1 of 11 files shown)


client/src/main/java/org/eclipse/hono/client/HonoClient.java
```diff
@@ -117,13 +117,48 @@ void getOrCreateEventSender(
...
+
+    /**
+     * Creates a new consumer of telemetry data for a tenant.
+     *
+     * @param tenantId The tenant to consume data for.
+     * @param telemetryConsumer The handler to invoke with every message received.
+     * @param creationHandler The handler to invoke with the outcome of the operation.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     */
+    @Deprecated
```
sophokles73: I think we do not need to deprecate this variant. Why not leave it up to clients to choose which variant they want to use? Being forced to pass in `null` to a method invocation never looks good in code FMPOV ...

client/src/main/java/org/eclipse/hono/client/HonoClient.java
```diff
@@ -136,11 +171,31 @@ void createTelemetryConsumer(
      * @param creationHandler The handler to invoke with the outcome of the operation.
      * @throws NullPointerException if any of the parameters is {@code null}.
      */
+    @Deprecated
```
sophokles73: same here

client/src/main/java/org/eclipse/hono/client/HonoClient.java
```diff
@@ -153,6 +208,7 @@ void createEventConsumer(
...
```
sophokles73: same here

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -259,19 +267,24 @@ protected static final void setApplicationProperties(final Message msg, final Ma
...
             receiver.openHandler(openAttach -> {
-                if(openAttach.failed()) {
+                if (openAttach.failed()) {
                     LOG.debug("receiver open attach failed [{}] by peer [{}]: {}", receiver.getRemoteSource(), con.getRemoteContainer(), openAttach.cause().getMessage());
                     result.fail(openAttach.cause());
-                }
-                else {
+                } else {
+                    if(LOG.isTraceEnabled()) {
+                        LOG.debug("receiver open attach succeeded [{}] by peer [{}]", receiver.getRemoteSource(), con.getRemoteContainer());
```
sophokles73: do you want to trace or debug?

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -236,7 +243,8 @@ protected static final void setApplicationProperties(final Message msg, final Ma
             final String sourceAddress,
             final ProtonQoS qos,
             final ProtonMessageHandler messageHandler,
-            final Handler<String> closeHook) {
+            final Handler<String> closeHook,
+            final Handler<AsyncResult<Void>> detachHandler) {
```
sophokles73: IMHO we should add this as a variant of the existing method with the existing method delegating to the new variant, passing in a `null` detach handler.

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -259,19 +267,24 @@ protected static final void setApplicationProperties(final Message msg, final Ma
...
+                        LOG.debug("receiver open attach succeeded [{}] by peer [{}]", receiver.getRemoteSource(), con.getRemoteContainer());
+                    }
                     result.complete(openAttach.result());
                 }
             });
             receiver.detachHandler(remoteDetached -> {
+                receiver.detach();
                 if (remoteDetached.succeeded()) {
                     LOG.debug("receiver [{}] detached (with closed=false) by peer [{}]", receiver.getRemoteSource(), con.getRemoteContainer());
+                    detachHandler.handle(Future.succeededFuture());
```
sophokles73: check if detach handler is `null`

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -213,7 +213,7 @@ public final void setRequestTimeout(final long timoutMillis) {
 
     private Future<ProtonReceiver> createReceiver(final ProtonConnection con, final String sourceAddress, final Handler<String> closeHook) {
 
-        return AbstractHonoClient.createReceiver(context, config, con, sourceAddress, ProtonQoS.AT_LEAST_ONCE, this::handleResponse, closeHook);
+        return AbstractHonoClient.createReceiver(context, config, con, sourceAddress, ProtonQoS.AT_LEAST_ONCE, this::handleResponse, closeHook, detach -> {});
```
sophokles73: do not require existing code to pass in *dummy* handlers, instead continue to offer the existing method.

client/src/main/java/org/eclipse/hono/client/impl/EventConsumerImpl.java
```diff
@@ -58,9 +59,10 @@ public static void create(
             final ProtonConnection con,
             final String tenantId,
             final BiConsumer<ProtonDelivery, Message> eventConsumer,
-            final Handler<AsyncResult<MessageConsumer>> creationHandler) {
+            final Handler<AsyncResult<MessageConsumer>> creationHandler,
+            final Handler<AsyncResult<Void>> detachHandler) {
```
sophokles73: IMHO we should add this as a variant of the existing method ...

client/src/main/java/org/eclipse/hono/client/impl/EventConsumerImpl.java
```diff
@@ -82,7 +85,8 @@ public static void create(
             final String tenantId,
             final String pathSeparator,
             final BiConsumer<ProtonDelivery, Message> eventConsumer,
-            final Handler<AsyncResult<MessageConsumer>> creationHandler) {
+            final Handler<AsyncResult<MessageConsumer>> creationHandler,
+            final Handler<AsyncResult<Void>> detachHandler) {
```
sophokles73: IMHO we should add this as a variant of the existing method ...

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -12,20 +12,14 @@
...
-import java.util.function.Supplier;
-
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.Vertx;
+import io.vertx.proton.ProtonClientOptions;
+import io.vertx.proton.ProtonConnection;
+import io.vertx.proton.ProtonDelivery;
```
sophokles73: please revert this change and adapt your IDE settings to the import order specified in `$HONO/eclipse/hono.importorder`

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -41,14 +35,19 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
```
sophokles73: see above

example/src/main/resources/logback.xml
```diff
@@ -15,6 +15,7 @@
     <appender-ref ref="STDOUT" />
   </root>
 
-  <logger name="org.eclipse.hono.client" level="DEBUG"/>
+  <logger name="org.eclipse.hono.client" level="TRACE"/>
+  <logger name="org.eclipse.hono.client.impl" level="TRACE"/>
```
sophokles73: do not commit TRACE log level

services/messaging/src/main/java/org/eclipse/hono/messaging/SenderFactoryImpl.java
```diff
@@ -122,6 +122,14 @@ public void setConfiguration(final HonoMessagingConfigProperties config) {
                 LOG.debug("sender [{}] for container [{}] closed", address, connection.getRemoteContainer());
             }
         });
+        sender.detachHandler(remoteDetached -> {
+            if (remoteDetached.succeeded()) {
+                LOG.debug("sender [{}] detached (with closed=false) by peer [{}]", sender.getRemoteSource(), connection.getRemoteContainer());
+            } else {
+                LOG.debug("sender [{}] detached (with closed=false) by peer [{}]: {}", sender.getRemoteSource(), connection.getRemoteContainer(), remoteDetached.cause().getMessage());
+            }
+            sender.close();
```
sophokles73: I do not think that this is enough, we also need to propagate the closing of the downstream link to the upstream protocol adapter. For that, we will probably need to be able to pass in a detach handler...

client/src/main/java/org/eclipse/hono/client/HonoClient.java
pellmann: I think the question is if there is a good use case to use the API without this detach handler. My idea was to eliminate the deprecated methods, because without this handlers and a proper reaction there could always be a problem, which could not be healed without this handler. I think we should not provide an API, if we know, the usage will end in problems.

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
pellmann: Okay - found the option in the Eclipse Code Formatter plugin for IDEA..

pellmann:
@sophokles73 I merged detach and close hook and also close the upstream receiver link. Could you please take a look.

services/messaging/src/main/java/org/eclipse/hono/messaging/ForwardingDownstreamAdapter.java
```diff
@@ -327,6 +327,15 @@ public final void onClientAttach(final UpstreamReceiver client, final Handler<As
...
                 addSender(client, createdSender);
                 tracker.complete();
+                createdSender.detachHandler(detachHandler -> {
+                    if (detachHandler.succeeded()) {
+                        logger.info("downstream sender was closed detached [con: {}, link: {}]", client.getConnectionId(), client.getLinkId());
+                    } else {
+                        logger.warn("downstream sender was closed detached [con: {}, link: {}]: {}", client.getConnectionId(), client.getLinkId(), detachHandler.cause().getMessage());
+                    }
+                    createdSender.close();
+                    resultHandler.handle(Future.failedFuture(detachHandler.cause()));
```
sophokles73: this seems to be wrong. the resultHandler has already been completed by the *tracker* Future. 
when the peer (the router) detaches this sender link, then we need to do some house keeping:
1. close the link and remove it from the list of downstream senders
2. close the upstream receiver link

sophokles73: I also would prefer if we could pass in a close handler to the `SenderFactory.createSender()` method so that the set up of the link is done in one place only. WDYT?

pellmann:
Will open an other one - conflicts with master.

Issue #467, opened 2018-01-29
Propagate detach(close=false) back und allow reopen


pellmann:
As a receiver (link) in the Hono client a detach(close=false) needs to be answered the same way and later a open on this detached link need to be handled.

pellmann:
It is needed to reopen (=poll) from the receiver side.

http://qpid.2158936.n2.nabble.com/Link-handling-after-detach-close-false-td7672272.html

sophokles73:
@pellmann is this fixed with your latest commit?

pellmann:
https://github.com/eclipse/hono/commit/d0bb9aa4310ccf68b8079481b51225c157037987

PR #479, opened 2018-02-06
Handle detached and closed receiver links (#467)

pellmann:
- added closeHook, which is called on detach and close
- close receiver on sender close in hono messaging


client/src/main/java/org/eclipse/hono/client/impl/EventConsumerImpl.java
```diff
@@ -50,6 +50,7 @@ private EventConsumerImpl(final Context context, final ClientConfigProperties co
      * @param tenantId The tenant to consumer events for.
      * @param eventConsumer The consumer to invoke with each event received.
      * @param creationHandler The handler to invoke with the outcome of the creation attempt.
+     * @param closeHook The handler to invoke when the link is closed by the peer (may be {@code null}).
```
sophokles73: this contradicts the declaration of the `NullPointerException`

client/src/main/java/org/eclipse/hono/client/impl/EventConsumerImpl.java
```diff
@@ -73,6 +75,7 @@ public static void create(
      * @param pathSeparator The address path separator character used by the server.
      * @param eventConsumer The consumer to invoke with each event received.
      * @param creationHandler The handler to invoke with the outcome of the creation attempt.
+     * @param closeHook The handler to invoke when the link is closed by the peer (may be {@code null}).
```
sophokles73: this contradicts the declaration of the `NullPointerException`

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -392,51 +392,58 @@ private void reconnect(final Handler<AsyncResult<HonoClient>> connectionHandler,
...
                 tenantId,
-                () -> newTelemetryConsumer(tenantId, messageConsumer));
+                () -> newTelemetryConsumer(tenantId, messageConsumer, closeHandler));
     }
 
     private Future<MessageConsumer> newTelemetryConsumer(
             final String tenantId,
-            final Consumer<Message> messageConsumer) {
+            final Consumer<Message> messageConsumer,
+            final Handler<AsyncResult<Void>> closeHandler) {
```
sophokles73: it doesn't seem to make much sense to use a `Handler<AsyncResult<Void>>` here instead of just a `Handler<Void>` because the close hook in the consumer cannot fail. IMHO we should therefore use a plain `Handler<Void>` here as well.

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -392,51 +392,58 @@ private void reconnect(final Handler<AsyncResult<HonoClient>> connectionHandler,
...
                 tenantId,
-                () -> newEventConsumer(tenantId, messageConsumer));
+                () -> newEventConsumer(tenantId, messageConsumer, closeHandler));
     }
 
     private Future<MessageConsumer> newEventConsumer(
             final String tenantId,
-            final BiConsumer<ProtonDelivery, Message> messageConsumer) {
+            final BiConsumer<ProtonDelivery, Message> messageConsumer,
+            final Handler<AsyncResult<Void>> closeHandler) {
```
sophokles73: see above

services/messaging/src/main/java/org/eclipse/hono/messaging/ForwardingDownstreamAdapter.java
```diff
@@ -278,6 +275,13 @@ private void onDisconnectFromDownstreamContainer(final ProtonConnection con) {
         }
     }
 
+    private void closeReceiverOnDownstreamClose(final UpstreamReceiver receiver) {
+        receiver.close(ErrorConditions.ERROR_NO_DOWNSTREAM_CONSUMER);
+        metrics.decrementUpstreamLinks(receiver.getTargetAddress());
+        metrics.decrementDownstreamSenders(receiver.getTargetAddress());
+        metrics.submitDownstreamLinkCredits(receiver.getTargetAddress(),0);
```
sophokles73: formatting

services/messaging/src/main/java/org/eclipse/hono/messaging/ForwardingDownstreamAdapter.java
```diff
@@ -278,6 +275,13 @@ private void onDisconnectFromDownstreamContainer(final ProtonConnection con) {
         }
     }
 
+    private void closeReceiverOnDownstreamClose(final UpstreamReceiver receiver) {
```
sophokles73: maybe better *closeReceiverOnDownstreamDetach*?

services/messaging/src/main/java/org/eclipse/hono/messaging/SenderFactory.java
```diff
@@ -34,12 +34,14 @@
      * @param address The target address to use for the sender.
      * @param qos The quality of service the sender should use.
      * @param sendQueueDrainHandler The handler to notify about credits the sender is getting replenished with.
+     * @param closeHook The handler to invoke when the link is closed by the peer (may be {@code null}).
```
sophokles73: if we keep the `Handler<String>` type then the JavaDoc should explain the meaning of the string that the handler will be invoked with, otherwise I propose to simply use `Handler<Void>`.

pellmann: Yes, I think the string does not help.

services/messaging/src/main/java/org/eclipse/hono/messaging/ForwardingDownstreamAdapter.java
pellmann: Yes, the generic detach is more accurate.

pellmann:
@sophokles73 - integrated your comments and added a test.

services/messaging/src/test/java/org/eclipse/hono/messaging/ForwardingDownstreamAdapterTest.java
```diff
@@ -265,6 +266,47 @@ public void testDownstreamDisconnectClosesUpstreamReceivers() {
...
+        ArgumentCaptor<Handler<AsyncResult<ProtonSender>>> closeHandlerCaptor = ArgumentCaptor.forClass((Class) Handler.class);
+
+        when(downstreamSender.close()).then(invocation -> {
+            closeHandlerCaptor.getValue().handle(Future.succeededFuture(downstreamSender));
+            return downstreamSender;
+        });
+        when(downstreamSender.closeHandler(closeHandlerCaptor.capture())).then(invocation -> {
+            return downstreamSender;
+        });
+
```
sophokles73: IMHO the test should be based on something similar to `HandlerCapturingConnectionFactory` but for a sender, e.g. a `HandlerCapturingProtonSender` which offers a method to get at the registered close or detach handlers.

The whole test case would then be much simpler, I guess. You would just `adapter.addSender(client, downstreamSender)` and then trigger the detach handler of the sender and verify that the upstream client has been closed, right?

services/messaging/src/test/java/org/eclipse/hono/messaging/ForwardingDownstreamAdapterTest.java
```diff
@@ -276,35 +276,23 @@ public void testDownstreamLinkClosesUpstreamReceivers() {
...
-            closeHandlerCaptor.getValue().handle(Future.succeededFuture(downstreamSender));
-            return downstreamSender;
-        });
-        when(downstreamSender.closeHandler(closeHandlerCaptor.capture())).then(invocation -> {
-            return downstreamSender;
-        });
-
         HandlerCapturingConnectionFactory factory = new HandlerCapturingConnectionFactory(con);
         when(downstreamSender.isOpen()).thenReturn(Boolean.FALSE);
+
```
sophokles73: the line above should be removed, it is the default behavior anyway

services/messaging/src/test/java/org/eclipse/hono/messaging/ForwardingDownstreamAdapterTest.java
```diff
@@ -276,35 +276,23 @@ public void testDownstreamLinkClosesUpstreamReceivers() {
...
         adapter.start(Future.future());
-        adapter.addSender(client, downstreamSender);
 
         // WHEN the downstream sender is closed
-        adapter.onClientAttach(client, s -> {
-        });
-        downstreamSender.close();
+        adapter.onClientAttach(client, s -> {});
+        ArgumentCaptor<Handler> detachHandlerCaptor = ArgumentCaptor.forClass(Handler.class);
+        verify(downstreamSender).detachHandler(detachHandlerCaptor.capture());
```
sophokles73: the three lines above should also go into the `GIVEN` section since they are part of the fixture

services/messaging/src/test/java/org/eclipse/hono/messaging/ForwardingDownstreamAdapterTest.java
sophokles73: which is fine for this test as well IMHO

client/src/test/java/org/eclipse/hono/client/impl/EventConsumerImplTest.java
```diff
@@ -107,4 +107,42 @@ public void testCreateRegistersBiConsumerAsMessageHandler(final TestContext ctx)
...
+        // WHEN the receiver link is closed
+        EventConsumerImpl.create(vertx.getOrCreateContext(), new ClientConfigProperties(), con, "tenant", eventConsumer, ctx.asyncAssertSuccess(s -> {
+            s.close(closeHandler -> {
+                consumerCreation.complete();
+            });
+        }), closeHook );
+        consumerCreation.await(500);
+
+        // THEN the close hook is called
+        verify(closeHook).handle(any());
```
sophokles73: analogously to the test case for `ForwardingDownstreamAdapter` I would propose something like this:

```
...
when(receiver.isOpen()).thenReturn(Boolean.TRUE);
Handler<String> closeHook = mock(Handler.class);
ArgumentCaptor<Handler> detachCaptor = ArgumentCaptor.forClass(Handler.class);
Async consumerCreation = ctx.async();
EventConsumerImpl.create(vertx.getOrCreateContext(), new ClientConfigProperties(), con, "tenant", eventConsumer, 
       ctx.asyncAssertSuccess(s -> consumerCreation.complete()), closeHook );
consumerCreation.await(500);
verify(receiver).detachHandler(detachCaptor.getCapture());

// WHEN the receiver link is closed
detachCaptor.getValue().handle(..);

// THEN the close hook is called
verify(closeHook).handle(any());
// and the receiver link is closed
verify(receiver).close();
```

Issue #416, opened 2017-11-21
Support for gateways reporting data on behalf of other devices


zubairhamed:
For example, devices which are proxied by an HTTP server (or the typical Gateway pattern for non-ip devices)

Currently it looks like a 1:1 mapping between auth-id and device-id

I understand that POST omits the tenant and device-id from the path.

How can multiple devices use the same auth-id to authenticate? Combine Basic Auth with the PUT request instead?

sophokles73:
A device within Hono is defined by its (unique) (*tenant ID*, *device ID*) tuple. That said, a particular device can have arbitrary *authentication ID*s associated with it (using the Credentials API). This accounts for use cases where a device is capable of using multiple transport protocols (and authentication mechanisms) or rotating keys.
However, when a device connects to a protocol adapter and gets authenticated, its *device ID* is resolved based on the credentials (*auth ID* + secret) the device provides. Thus, if two devices use the same credentials during authentication, they are resolved to the same *device ID*, i.e. Hono treats them as the same device.
With gateways there are two scenarios possible:

1. The gateway *acts on behalf of* the sensors attached to it, i.e. each sensor is represented by its own *device ID* in Hono and the gateway effectively acts as a proxy only and is not represented by a *device ID* in Hono.
1. The sensors attached to the gateway are managed as *resources* or *properties* of the gateway itself but only the gateway is represented by a *device ID* in Hono. The Eclipse Kura gateway for example follows this approach.

zubairhamed:
Hi Kai,

I’m referring to more like the pattern of actual devices which are non ip enabled and needs the use of a gateway to proxy them.

Think of Example Zigbee where nodes should be identified by devices but the gateway could be using a single username or password(or certificate auth)

Z

sophokles73:
Still, this might be accomplished using any of the approaches outlined above. Which one are you referring to? Do the ZigBee devices/sensors need to be represented by their own *device ID* or are they mapped to *properties* of the gateway, so that when such a ZigBee device has some data to publish, it would appear as if it came from the gateway?

zubairhamed:
Yes exactly. Each Zigbee (or any other typical similar protocol)device would be its own device and would need its own Device ID.

sophokles73:
> Each Zigbee (or any other typical similar protocol)device would be its own device and would need its own Device ID.

Is that a hard requirement or is this just the first approach that came to mind? Would it also work to map the ZigBee devices to properties of the gateway?

[...]

PR #480, opened 2018-02-06
[#416] Support gateway ID in RegistrationClient.

sophokles73:
This is the 2nd PR for adding support for gateway devices.
It adds an assertRegistration method variant to RegistrationClient accepting a gateway ID in order
to get an assertion for a device that is connected via a gateway.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
@@ -12,12 +12,10 @@
  */
 package org.eclipse.hono.service;
 
-import java.net.HttpURLConnection;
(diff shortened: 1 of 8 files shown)


service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -389,25 +387,17 @@ private ProtonClientOptions createClientOptions() {
      * 
      * @param tenantId The tenant that the device belongs to.
      * @param deviceId The device to get the assertion for.
+     * @param gatewayId The gateway that wants to act on behalf of the device.
```
sysexcontrol: I would add : `May be null, if the device connects directly to the adapter.`.

PR #481, opened 2018-02-06
HTTP interface for tenant API.

sysexcontrol:
Contained is the implementation of the http interface for the tenant API, including tests.

Not contained is the documentation - this should be the discussion base for the implementation.

**Please do not review the first commit** - it is just the squashed version of the AMQP implementation PR that was discussed already. The two commits on top are the ones for the HTTP and will be the only ones that will make it to master later.


service-base/src/main/java/org/eclipse/hono/service/tenant/TenantHttpEndpoint.java
```diff
@@ -0,0 +1,232 @@
...
+
+        doTenantAction(ctx, requestMsg, (status, tenantResult) -> {
+            response.setStatusCode(status);
+            switch (status) {
+                case HttpURLConnection.HTTP_OK:
+                    final String msg = tenantResult.getJsonObject(TenantConstants.FIELD_PAYLOAD).encodePrettily();
+                    response
+                            .putHeader(HttpHeaders.CONTENT_TYPE, HttpUtils.CONTENT_TYPE_JSON_UFT8)
+                            .putHeader(HttpHeaders.CONTENT_LENGTH, String.valueOf(msg.length()))
+                            .write(msg);
```
sophokles73: isn't this what *setResponseBody()* does?

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantHttpEndpoint.java
```diff
@@ -0,0 +1,232 @@
...
+        try {
+            JsonObject payload = null;
+            if (ctx.getBody().length() > 0) {
+                payload = ctx.getBodyAsJson();
+            }
+            updateTenant(payload, ctx);
+        } catch (final DecodeException e) {
+            HttpUtils.badRequest(ctx, "body does not contain a valid JSON object");
+        }
+    }
```
sophokles73: this is the same code as *doAddTenantJson*. Maybe we can refactor the body extraction code into a separate method?

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantHttpEndpoint.java
```diff
@@ -0,0 +1,232 @@
...
+        if (payload != null) {
+            payload.remove(FIELD_TENANT_ID);
+        }
+        final HttpServerResponse response = ctx.response();
+        final JsonObject requestMsg = TenantConstants.getServiceRequestAsJson(Action.ACTION_UPDATE.toString(),
+                tenantId, payload);
+
+        doTenantAction(ctx, requestMsg, (status, tenantResult) -> {
+                response.setStatusCode(status);
+                if (status >=400) {
```
sophokles73: formatting

core/src/main/java/org/eclipse/hono/util/ProtocolAdapterConstants.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * SPDX-License-Identifier: EPL-1.0
+ */
+package org.eclipse.hono.util;
+
+/**
+ * Constants used by protocol adapters of Hono.
+ *
+ */
+public final class ProtocolAdapterConstants {
```
sophokles73: can't we simply add these to `Constants`?

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/DeviceRegistryTestUtils.java
```diff
@@ -83,6 +85,30 @@ public static JsonObject buildCredentialsPayloadPresharedKey(final String device
...
+     * <p>
+     * The tenant object created contains configurations for the http and the mqtt adapter.
+     *
+     * @param tenantId The tenant identifier.
+     * @return The tenant object.
+     */
+    public static JsonObject buildTenantPayload(final String tenantId) {
+        final JsonObject adapterDetailsHttp = new JsonObject().
+                put(TenantConstants.FIELD_ADAPTERS_TYPE, ProtocolAdapterConstants.TYPE_HTTP).
+                put(TenantConstants.FIELD_ADAPTERS_DEVICE_AUTHENTICATION_REQUIRED, "true").
```
sophokles73: use a *boolean*, not a string literal

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/DeviceRegistryTestUtils.java
```diff
@@ -83,6 +85,30 @@ public static JsonObject buildCredentialsPayloadPresharedKey(final String device
...
+                put(TenantConstants.FIELD_ENABLED, "true");
```
sophokles73: use a *boolean*, not a string literal

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/DeviceRegistryTestUtils.java
```diff
@@ -83,6 +85,30 @@ public static JsonObject buildCredentialsPayloadPresharedKey(final String device
...
+        final JsonObject adapterDetailsMqtt = new JsonObject().
+                put(TenantConstants.FIELD_ADAPTERS_TYPE, ProtocolAdapterConstants.TYPE_MQTT).
+                put(TenantConstants.FIELD_ADAPTERS_DEVICE_AUTHENTICATION_REQUIRED, "true").
```
sophokles73: use a *boolean*, not a string literal

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/DeviceRegistryTestUtils.java
```diff
@@ -83,6 +85,30 @@ public static JsonObject buildCredentialsPayloadPresharedKey(final String device
...
+                put(TenantConstants.FIELD_ENABLED, "true");
```
sophokles73: use a *boolean*, not a string literal

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/DeviceRegistryTestUtils.java
```diff
@@ -83,6 +85,30 @@ public static JsonObject buildCredentialsPayloadPresharedKey(final String device
...
+        final JsonObject tenantPayload = new JsonObject().
+                put(TenantConstants.FIELD_TENANT_ID, tenantId).
+                put(TenantConstants.FIELD_ENABLED, "true").
```
sophokles73: use a *boolean*, not a string literal

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/DeviceRegistryTestUtils.java
```diff
@@ -139,6 +165,18 @@ public static boolean testJsonObjectToBeContained(final JsonObject jsonObject, f
                             containResult.set(false);
                         }
                     }
+                } else if (entry.getValue() instanceof JsonArray) {
+                    if (!(jsonObject.getValue(entry.getKey()) instanceof JsonArray)) {
```
sophokles73: update the JavaDoc accordingly

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/DeviceRegistryTestUtils.java
```diff
@@ -148,4 +186,49 @@ public static boolean testJsonObjectToBeContained(final JsonObject jsonObject, f
...
+                return false;
+            }
+
+            boolean containingElemFound = false;
+            for (Object elemOfBiggerArray: containingArray) {
+                if (!(elemOfBiggerArray instanceof JsonObject)) {
+                    return false;
+                }
+
+                if (testJsonObjectToBeContained((JsonObject)elemOfBiggerArray, (JsonObject)containedElem)) {
```
sophokles73: formatting

core/src/main/java/org/eclipse/hono/util/ProtocolAdapterConstants.java
sysexcontrol: Yes, you are right.Will move them.

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/DeviceRegistryTestUtils.java
sysexcontrol: :+1: thanks for spotting.

sysexcontrol:
@sophokles73 : I pushed two new commits. The first one is a half rewrite of the TenantHttpEndpoint, which eliminates most code duplications existing before. Can you please have a look and tell me what you think about this way of coding it? I like it much better (although it still can be improved) - WDYT?

Not addressed the Test code duplications so far, would prefer to do this in another PR (after the release and after the merge of the Tenant API implementation).

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantHttpEndpoint.java
```diff
@@ -68,155 +69,104 @@ public String getName() {
     public void addRoutes(final Router router) {
 
         final String path = String.format("/%s", TenantConstants.TENANT_ENDPOINT);
+
+        final BodyHandler bodyHandler = BodyHandler.create();
+        bodyHandler.setBodyLimit(2048); // limit body size to 2kb
```
sophokles73: this should probably come from a config property

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantHttpEndpoint.java
```diff
@@ -68,155 +69,104 @@ public String getName() {
...
+        if (msg != null) {
+            final String body = msg.encodePrettily();
+            response.putHeader(HttpHeaders.CONTENT_TYPE, HttpUtils.CONTENT_TYPE_JSON_UFT8)
+                .putHeader(HttpHeaders.CONTENT_LENGTH, String.valueOf(body.length()))
+                .write(body);
+        }
     }
 
-    private void doTenantAction(final RoutingContext ctx, final JsonObject requestMsg, final BiConsumer<Integer, JsonObject> responseHandler) {
+    private void sendTenantAction(final RoutingContext ctx, final JsonObject requestMsg, final BiConsumer<Integer, JsonObject> responseHandler) {
```
sophokles73: we should be able to pull this up into `AbstractHttpEndpoint`, don't you think? There doesn't seem to be anything left that is tenant specific, is there?

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantHttpEndpoint.java
```diff
@@ -68,155 +69,104 @@ public String getName() {
...
 
-        doTenantAction(ctx, requestMsg, (status, tenantResult) -> {
-                response.setStatusCode(status);
-                if (status >=400) {
+        sendTenantAction(ctx, requestMsg, (status, tenantResult) -> {
+            response.setStatusCode(status);
+            if (status >=400) {
+                setResponseBody(tenantResult, response);
+            } else if (sendResponseForStatus != null) {
+                if (sendResponseForStatus.test(status)) {
```
sophokles73: why not let another handler process the response? we can provide (and register) default handlers for the CRUD responses which can be overridden by sub-classes if appropriate

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantHttpEndpoint.java
```diff
@@ -68,155 +69,104 @@ public String getName() {
...
-
-        final String tenantId = getTenantParam(ctx);
-        final HttpServerResponse response = ctx.response();
-        final JsonObject requestMsg = TenantConstants.getServiceRequestAsJson(Action.ACTION_GET.toString(), tenantId);
+    /**
+     * Check the Content-Type of the request to be 'application/json' and extract the payload if this check was
+     * successful.
+     * <p>
+     * The payload is parsed to ensure it is valid JSON and is put to the RoutingContext ctx with the
+     * key {@link #KEY_REQUEST_BODY}.
```
sophokles73: this doesn't seem to be correct

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/DeviceRegistryTestUtils.java
```diff
@@ -95,16 +95,16 @@ public static JsonObject buildCredentialsPayloadPresharedKey(final String device
      */
     public static JsonObject buildTenantPayload(final String tenantId) {
         final JsonObject adapterDetailsHttp = new JsonObject().
-                put(TenantConstants.FIELD_ADAPTERS_TYPE, ProtocolAdapterConstants.TYPE_HTTP).
-                put(TenantConstants.FIELD_ADAPTERS_DEVICE_AUTHENTICATION_REQUIRED, "true").
-                put(TenantConstants.FIELD_ENABLED, "true");
+                put(TenantConstants.FIELD_ADAPTERS_TYPE, Constants.TYPE_HTTP).
```
sophokles73: we usually put the *period* on the new line, e.g.

```
new JsonObject()
  .put("key", value);
```

sysexcontrol:
@sophokles73 : I pushed 2 new commits, one is a larger effort to further reduce the repeating code of the http endpoints (as already started before).
The registration and credentials endpoint are intentionally not fully refactored in here, but the base to do that afterwards is contained (and fully used for the tenant http endpoint).

Due to a necessary rebase, I needed to make a forced push (the last 2 commits are the new ones).

core/src/main/java/org/eclipse/hono/util/RequestResponseApiConstants.java
```diff
@@ -43,6 +43,37 @@
     public static final String FIELD_PAYLOAD   = "payload";
     public static final String FIELD_TENANT_ID = "tenant-id";
 
+    /* request actions */
+    public enum Action {
+        ACTION_GET, ACTION_ADD, ACTION_UPDATE, ACTION_REMOVE, ACTION_UNKNOWN;
```
sophokles73: I do not believe that these are the same for all request/response APIs, are they?

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -92,11 +114,107 @@ protected void extractRequiredJsonPayload(final RoutingContext ctx) {
...
+     * @param locationFormatter A Function that constructs the String being set as the HTTP location header {@link HttpHeaders#LOCATION}.
+     *                          Must not be null if the HTTP request was to add an object.
+     * @return BiConsumer<Integer, JsonObject> A consumer for the status and the JSON object that implements the default behaviour for responding to the HTTP request.
+     * @throws NullPointerException If ctx is null, if the locationFormatter is null although the request was for adding an object.
+     */
+    protected final BiConsumer<Integer, JsonObject> getDefaultResponseHandler(
+            final RoutingContext ctx,
+            final RequestResponseApiConstants.Action action,
+            final Predicate<Integer> setResponseBodyForStatus,
+            final Function<Void,String> locationFormatter
```
sophokles73: I would instead use a `Handler<HttpServerResponse>` and let the handler modify the response in any way it sees fit, i.e. not limiting it to provide a location header value.

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -92,11 +114,107 @@ protected void extractRequiredJsonPayload(final RoutingContext ctx) {
...
+    }
+
+    /**
+     * Sending a request message to a consumer on the event bus for further processing.
+     *
+     * @param ctx The routing context of the request.
+     * @param requestMsg The JSON object to send via the event bus.
+     * @param responseHandler The handler to be invoked for the response received as answer from the event bus.
+     */
+    protected void sendAction(final RoutingContext ctx, final JsonObject requestMsg, final BiConsumer<Integer, JsonObject> responseHandler) {
```
sophokles73: is this method supposed to be overridden by subclasses?

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -92,11 +114,107 @@ protected void extractRequiredJsonPayload(final RoutingContext ctx) {
...
+                });
+    }
+
+    /**
+     * Get the tenantId from the standard parameter name {@link #PARAM_TENANT_ID}.
+     *
+     * @param ctx The routing context of the request.
+     * @return The tenantId retrieved from the request.
+     */
+    protected String getTenantParam(final RoutingContext ctx) {
```
sophokles73: final?

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -92,11 +114,107 @@ protected void extractRequiredJsonPayload(final RoutingContext ctx) {
...
+        return ctx.request().getParam(PARAM_TENANT_ID);
+    }
+
+    /**
+     * Get the deviceId from the standard parameter name {@link #PARAM_DEVICE_ID}.
+     *
+     * @param ctx The routing context of the request.
+     * @return The deviceId retrieved from the request.
+     */
+    protected String getDeviceIdParam(final RoutingContext ctx) {
```
sophokles73: final?
document `NullPointerException`?

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -92,11 +114,107 @@ protected void extractRequiredJsonPayload(final RoutingContext ctx) {
...
+                    }
+                });
+    }
+
+    /**
+     * Get the tenantId from the standard parameter name {@link #PARAM_TENANT_ID}.
+     *
+     * @param ctx The routing context of the request.
+     * @return The tenantId retrieved from the request.
+     */
```
sophokles73: document `NullPointerException`?

core/src/main/java/org/eclipse/hono/util/RequestResponseApiConstants.java
sysexcontrol: Unfortunately: you are right. I haven't looked into the registration API, which defines more actions than this. With the Credentials API it would have been fitting (but only for now).

I will come up with a better solution, this has been too much generalized obviously...

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
sysexcontrol: No, will make it final.

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
sysexcontrol: Yes.

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
sysexcontrol: Yes.

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
sysexcontrol: :+1: much better indeed.
And this is solving another issue as well: the annoying action parameter, which was rather artificial in this signature.

sysexcontrol:
@sophokles73 : thanks for the last comments, I (forced) pushed 2 new commits.
Since they are not in the right order in the github display, here are the relevant ones (all other are coming from the again necessary rebase and were not changed at all):
9cf619a
35abb9f

The enum action and the generic methods are applied to the credentials implementations as well, where it fits well. The Registration API is not touched (yet).
Can you have another look?

core/src/main/java/org/eclipse/hono/util/RequestResponseApiConstants.java
```diff
@@ -43,20 +41,23 @@
     public static final String FIELD_PAYLOAD   = "payload";
     public static final String FIELD_TENANT_ID = "tenant-id";
 
-    /* request actions */
-    public enum Action {
+    /**
+     * Request standard actions to support the standard lifecycle.
+     * If more or other actions shall be used, an own enum type should be defined.
+    */
+    public enum StandardAction implements RequestResponseAction {
```
sophokles73: what is wrong with simply defining the actions that are defined for an API in the corresponding `*Constants` class?

sysexcontrol: Basically, not too much anymore, except for: 
it would be completely identical for tenant and credentials API, and it would double about 30 code lines.
Since it often is probably the case that the HTTP APIs have get,add,update,delete, I have put it into the StandardAction.

Previously, there were generic methods operating on this type. Due to the introduced handlers this does not exist anymore, so the left reason to keep them here is purely the code duplication.

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantHttpEndpoint.java
```diff
@@ -75,21 +79,26 @@ public void addRoutes(final Router router) {
         router.post(path).handler(this::extractRequiredJsonPayload);
         router.post(path).handler(this::checkPayloadForTenantId);
         router.post(path).handler(ctx -> doTenantHttpRequest(ctx, StandardAction.ACTION_ADD,
-                status -> status == HttpURLConnection.HTTP_CREATED));
+                status -> status == HttpURLConnection.HTTP_CREATED,
+                tenantId -> response -> response.putHeader(HttpHeaders.LOCATION,
```
sophokles73: IMHO this does make the code very hard to read and, worse, very hard to adapt later. FMPOV it would be much cleaner to have one method per operation/action and then implement them accordingly.

sysexcontrol: I understand your point and will come up with a less condensed version of this code.

sysexcontrol:
@sophokles73 : I refactored the TenantHttpEndpoint to a less condensed version, which IMHO improves readability and changeability a lot.

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantHttpEndpoint.java
```diff
@@ -13,17 +13,17 @@
 
 package org.eclipse.hono.service.tenant;
 
+import static java.net.HttpURLConnection.HTTP_CREATED;
```
sophokles73: please do not import the constants statically

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -225,4 +235,15 @@ protected final String getDeviceIdParam(final RoutingContext ctx) {
         return ctx.request().getParam(PARAM_DEVICE_ID);
     }
 
+    /**
+     * Get a predicate that checks if a given HTTP status is set to the passed value.
+     *
+     * @param statusToCheckFor The HTTP status, e.g. {@link HttpURLConnection#HTTP_OK}.
+     * @return The predicate.
+     */
+    protected final Predicate<Integer> getHttpStatusPredicate(final int statusToCheckFor) {
```
sophokles73: this obviously works but FMPOV it would improve readability a lot if we simply would put the lamba in the place where it is used instead of encapsulating this within a method. In the end, that's one of the reasons why lambdas have been introduced.
Compare
```
doTenantHttpRequest(ctx, tenantId, StandardAction.ACTION_ADD, 
   getHttpStatusPredicate(HTTP_CREATED),
   getLocationHeaderHandler(location));
```
with
```
doTenantHttpRequest(ctx, tenantId, StandardAction.ACTION_ADD, 
   status -> status == HTTP_CREATED,
   getLocationHeaderHandler(location));
```
How many bytes have we saved? Which version is easier to understand?
The same holds true for the *getLocationHeaderHandler()* method FMPOV.

sysexcontrol: Well, we fully agree on that - this is just like I wrote it before. Sometimes PR comments are not so clear...

The only difference with explicit getters for lambdas (or Handlers) is that they get a Javadoc and a (hopefully) good name, which sometimes can be helpful IMHO. In this case it is simple enough that this is not necessary.

Anyway, I will introduce the lambdas again and hope we both are satisfied then...

sysexcontrol:
@sophokles73 : I reintroduced the lambdas, hope we agree on this version :-) Thanks!

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantHttpEndpoint.java
```diff
@@ -128,32 +125,35 @@ private void addTenant(final RoutingContext ctx) {
...
+        doTenantHttpRequest(ctx, tenantId, RequestResponseApiConstants.StandardAction.ACTION_GET,
+                status -> status == HttpURLConnection.HTTP_OK, null);
     }
 
     private void updateTenant(final RoutingContext ctx) {
 
         final String tenantId = getTenantIdFromContext(ctx);
 
-        doTenantHttpRequest(ctx, tenantId, StandardAction.ACTION_UPDATE,null, null);
+        doTenantHttpRequest(ctx, tenantId, RequestResponseApiConstants.StandardAction.ACTION_UPDATE,null, null);
```
sophokles73: formatting

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantHttpEndpoint.java
```diff
@@ -128,32 +125,35 @@ private void addTenant(final RoutingContext ctx) {
...
     }
 
     private void removeTenant(final RoutingContext ctx) {
 
         final String tenantId = getTenantIdFromContext(ctx);
 
-        doTenantHttpRequest(ctx, tenantId, StandardAction.ACTION_REMOVE,null, null);
+        doTenantHttpRequest(ctx, tenantId, RequestResponseApiConstants.StandardAction.ACTION_REMOVE,null, null);
```
sophokles73: formatting

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantHttpEndpoint.java
sophokles73: I think we should refer to the actions via the `TenantConstants` class (this is where they will end up anyway FMPOV)

sysexcontrol: 👍

sysexcontrol:
@sophokles73 : comments adressed, thx.

sysexcontrol:
Commits combined and provided in PR [#501] for merge. This PR is closed instead.

PR #482, opened 2018-02-07
[#416] Support devices acting as gateways in HTTP adapter.

sophokles73:
This is the 3rd part of adding support for gateways.

It adds support for HTTP based gateways.

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -14,6 +14,7 @@ package org.eclipse.hono.adapter.http;
 
 import java.net.HttpURLConnection;
 import java.util.Objects;
+import java.util.Optional;
 
 import org.apache.qpid.proton.amqp.Binary;
 import org.apache.qpid.proton.amqp.messaging.Data;
@@ -21,6 +22,7 @@ import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.client.ClientErrorException;
 import org.eclipse.hono.client.MessageSender;
 import org.eclipse.hono.service.AbstractProtocolAdapterBase;
+import org.eclipse.hono.service.auth.device.Device;
 import org.eclipse.hono.service.http.HttpUtils;
 import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.EventConstants;
@@ -466,7 +468,14 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
             HttpUtils.badRequest(ctx, "missing body");
         } else {
 
-            final Future<JsonObject> tokenTracker = getRegistrationAssertion(tenant, deviceId, null);
+            final Device authenticatedDevice = Optional.ofNullable(ctx.user()).map(user -> {
+                if (Device.class.isInstance(user)) {
+                    return (Device) user;
+                } else {
+                    return null;
+                }
+            }).orElse(null);
+            final Future<JsonObject> tokenTracker = getRegistrationAssertion(tenant, deviceId, authenticatedDevice);
 
             CompositeFuture.all(tokenTracker, senderTracker).compose(ok -> {
 
--- a/site/content/user-guide/http-adapter.md
+++ b/site/content/user-guide/http-adapter.md
@@ -227,22 +227,21 @@ The same holds true analogously for the *hono-service-device-registry.hono* addr
 
 ## Using the API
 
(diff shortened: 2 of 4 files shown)


site/content/user-guide/http-adapter.md
```diff
@@ -302,15 +293,38 @@ Response:
     HTTP/1.1 202 Accepted
     Content-Length: 0
 
-An authenticated device may also use this URI. In this case the device needs to know both the *auth-id* as well as the corresponding *device identifier*:
+### Publish Telemetry Data (authenticated gateway)
```
sysexcontrol: Some lines in the introduction paragraph must be changed as well to reflect the now possible authentication of the gateway (instead of the device).
They only contain the device as authentication possibility, like e.g.
`The adapter verifies the credentials provided by the device against the credentials...`

Also I would now prefer if we change
`Publish Telemetry Data` to
`Publish Telemetry Data (authenticated devices)` to immediately giving the reader the information that there are different possibilities (now 3 in summary).
Same for event.

site/content/user-guide/http-adapter.md
```diff
@@ -242,7 +242,6 @@ There is a subtle difference between the *device identifier* (*device-id*) and t
 * Request Headers:
```
sysexcontrol: As with the device registry (where it is now fixed), I raise the question here again:
I find it strange that the usage of the HTTP adapter is documented in `Setup Http Adapter`.
And would propose to separate it into two pages instead.

As first time reader I would not expect the very valuable usage documentation of the http adapter in a `setup` page.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -48,6 +50,11 @@
  */
 public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtocolAdapterProperties> extends AbstractProtocolAdapterBase<T> {
 
+    /**
+     *
```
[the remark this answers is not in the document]
sophokles73: in fact, the whole constant is not needed anymore.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -48,6 +50,11 @@
  */
 public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtocolAdapterProperties> extends AbstractProtocolAdapterBase<T> {
 
+    /**
+     * 
+     */
+    protected static final String KEY_GATEWAY_DEVICE = "GATEWAY_DEVICE";
+
```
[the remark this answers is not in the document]
sophokles73: yes, I will remove it

site/content/user-guide/http-adapter.md
sophokles73: I tend to agree. I also do not like the `Setup Http Adapter` name so I propose to split it up into `Http Adapter Configuration` and `Http Adapter Usage`. WDYT?

site/content/user-guide/http-adapter.md
sophokles73: > Some lines in the introduction paragraph must be changed as well to reflect the now possible authentication of the gateway (instead of the device).

Can you point me to the line that needs to be changed from your point of view? I can't find anything that needs to be changed AFAIC

site/content/user-guide/http-adapter.md
sysexcontrol: :+1:

site/content/user-guide/http-adapter.md
sysexcontrol: Yes, e.g.

The adapter verifies the credentials provided by the device against the credentials that the configured Credentials service has on record for the device.

Some rephrasing into "by the device or gateway", OR: a subsequent sentence that instead of the device a gateway acting on behalf of the device would be helpful IMHO.
As short as possible, but complete, this is what I meant.
WDYT?

sophokles73: The authentication process is generic and does not really care whether it is a *device* or a *gateway* that gets authenticated. I can imagine using the abstract term *client* instead of *device*, though. Would that help?

sysexcontrol: Yes, that would improve it.
IMHO it would be best to state at the end of the paragraph that a *client* could be a *device* or *gateway*, but I leave that up to you.
*Client* fixes the main point for me.

sophokles73:
@sysexcontrol I have pushed some changes

site/content/user-guide/http-adapter.md
sophokles73: I would like to make this split for all our documents in one go after this PR based on our discussion on IRC

sysexcontrol:
@sophokles73 : I assume the split into two pages will be done separately from this PR, right (which is fine IMHO)?
Assuming that, I now approve the changes.

PR #488, opened 2018-02-12
Add sequence diagrams to Event API description.

sysexcontrol:
This PR contains the minimum of sequence diagrams that were missing in the Event API so far.

I would propose to harmonize it with the other APIs and reflect the Precondition with a sequence diagram as well.
This could be in another commit inside this PR. WDYT?


site/content/api/Event-API.md
```diff
@@ -31,7 +31,7 @@ Hono supports *AT LEAST ONCE* delivery of *Event* messages only. A client theref
 
 **Message Flow**
 
-TBD
+![Upload event data flow](../uploadEvent_Success.png)
```
sophokles73: can you also add the textual description (analogously to Telemetry API)?

when doing so, please explicitly point out that, in contrast to the Telemetry case, the protocol adapter waits for the `accepted` disposition to arrive before signaling the device that it has accepted the event.

site/content/api/Event-API.md
```diff
@@ -59,7 +59,8 @@ Hono supports *AT LEAST ONCE* delivery of *Event* messages only. A client theref
 
 **Message Flow**
 
-TBD
+
+![Receive event data flow](../consumeEvent_Success.png)
```
sophokles73: can you also add the textual description (analogously to Telemetry API)?

site/content/api/Event-API.md
```diff
@@ -18,7 +18,7 @@ The following operations can be used by *Devices* and/or *Protocol Adapters* (to
 
 Both *Devices* as well as *Protocol Adapters* will be referred to as *clients* in the remainder of this section.
 
-## Send Event
+## Upload Event Data
```
ppatierno: Why changing from send to upload ?

sysexcontrol:
@sophokles73 : I pushed a new commit containing the requested changes.

site/content/api/Event-API.md
```diff
@@ -31,8 +31,15 @@ Hono supports *AT LEAST ONCE* delivery of *Event* messages only. A client theref
...
 **Message Flow**
 
+The following sequence diagram illustrates the flow of messages involved in the *MQTT Adapter* uploading an event data message to Hono Messaging's Event endpoint.
+
 ![Upload event data flow](../uploadEvent_Success.png)
 
+1. *MQTT Adapter* sends event data for device `4711`.
+   1. *Hono Messaging* successfully verifies that device `4711` of `TENANT` exists and is enabled by means of validating the *registration assertion* included in the message (see [Device Registration]({{< relref "api/Device-Registration-API.md#assert-device-registration" >}})) and forwards data to *AMQP 1.0 Messaging Network*.
+1. *AMQP 1.0 Messaging Network* acknowledges reception of the message.
+   1. *Hono Messaging* acknowledges reception of the message. In contrast to telemetry data, the protocol adapter waits explicitly for the accepted disposition before it signals the successful delivery to the device.
```
sophokles73: we need to be careful here with the term *delivery* because people might erorneously believe that it is about delivering the event to a consuming application (which it is not). So we should probably better write something like:

In contrast to telemetry data, the protocol adapter awaits the disposition from the *AMQP 1.0 Messaging Network* 
 before signaling to the device that it has accepted the message for processing.

site/content/api/Event-API.md
sysexcontrol: I just used the same phrase as already used for the telemetry page. Do you think "sending events" and "uploading telemetry data" would be better (as it was before)? I was not quite sure myself and then decided for that... WDYT?

site/content/api/Event-API.md
sysexcontrol: :+1: that is better, I take exactly this into the description.

sysexcontrol:
@sophokles73 : fixed the `await disposition ` description.

PR #492, opened 2018-02-14
Fix m2eclipse error "Artifact has not been packaged yet..."

ctron:
This change fixes the annoying error:

> Artifact has not been packaged yet. When used on reactor artifact,
unpack should be executed after packaging: see MDEP-98.

It does work around this issue by simply disabling the plugin execution
in the m2eclipse plugin. This means that the IDE builder will no longer
execute the step of packaging the legal files. However running maven
itself, even from within the IDE, will still perform this step.

Signed-off-by: Jens Reimann <jreimann@redhat.com>


sophokles73:
Thanks :-)

PR #495, opened 2018-02-16
Forgotten rename of HTTP adapter

pho-enix:


pho-enix:
Thx for prompt reply. Changed accordingly ...

PR #498, opened 2018-02-20
[#497] Don't call the closeHook if opening the sender/receiver has failed

### LINKED ISSUE #497: AbstractHonoClient#createReceiver: closeHook called after resultFuture has already failed
We tried to create a consumer via the HonoClient and got the following log output:

>  15:38:01.368 [vert.x-eventloop-thread-3] DEBUG o.e.h.client.impl.AbstractHonoClient - receiver open attach failed [Source{address='null', durable=NONE, expiryPolicy=SESSION_END, timeout=0, dynamic=false, dynamicNodeProperties=null, distributionMode=null, filter=null, defaultOutcome=null, outcomes=null, capabilities=null}] by peer [Router.qdrouterd-b7c5bd887-4khf2]: Error{condition=amqp:unauthorized-access, description='link disallowed by local policy', info=null}
>  
>  15:38:01.369 [vert.x-eventloop-thread-3] DEBUG o.e.h.client.impl.AbstractHonoClient - receiver [event/HEALTH_CHECK_TENANT] closed by peer [Router.qdrouterd-b7c5bd887-4khf2]: Error{condition=amqp:unauthorized-access, description='link disallowed by local policy', info=null}

That means the result Future of the `AbstractHonoClient#createReceiver` invocation failed. 
The unexpected part was that afterwards, the closeHandler got called as well.

The same way as a failed `honoClient.connect` attempt only fails the returned Future - and does not trigger the disconnect Handler, IMHO a failed `createReceiver` invocation should only fail the returned Future and should not trigger the closeHandler as well (even if a "close" was received).

calohmn:
This fixes #497.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
@@ -205,7 +205,7 @@ public abstract class AbstractHonoClient {
                             con.getRemoteContainer(), remoteDetached.cause().getMessage());
                 }
                 sender.close();
-                if (closeHook != null) {
+                if (closeHook != null && !result.failed()) {
                     closeHook.handle(targetAddress);
                 }
             });
@@ -216,7 +216,7 @@ public abstract class AbstractHonoClient {
                     LOG.debug("sender [{}] closed by peer: {}", targetAddress, senderClosed.cause().getMessage());
                 }
                 sender.close();
-                if (closeHook != null) {
+                if (closeHook != null && !result.failed()) {
                     closeHook.handle(targetAddress);
                 }
             });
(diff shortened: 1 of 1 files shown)


pellmann:
Thanks for spotting this! Could you also add this for the sender link?

calohmn:
Sure, I've amended the commit accordingly.

calohmn:
The failed test above
````
testConnectFailsAfterMaxConnectionAttempts(org.eclipse.hono.client.impl.HonoClientImplTest)  Time elapsed: 5.006 sec  <<< ERROR!
org.junit.runners.model.TestTimedOutException: test timed out after 5 seconds
````
doesn't look like it is caused by the changes here - there is no receiver/sender created there.

PR #509, opened 2018-02-22
remove all devices belonging to a tenant

balsmn:
The PR adds a method in Hono Registration API to delete all devices belonging to a tenant.


balsmn:
Oh, I was expecting only one commit in the PR and not > 250 commits.

PR #510, opened 2018-02-22
Added removeAllDevices method to RegistrationService API

balsmn:
The PR adds a method in Hono Registration API to remove all devices belonging to a tenant.


balsmn:
Oops, my bad. Sorry. Of course, will fix it.

sophokles73:
@balsmn 
I have to admit that I do not really understand the reason why this should be defined as part of the Device Registration API. Can you explain why you think it is needed?

balsmn:
The reason is simple. If a tenant gets removed, then it would be better to clean up all the related data, like Devices, Credentials etc.

sophokles73:
Nothing prevents you from doing so when a tenant gets removed, does it? My point is that I cannot really see why your implementation choice of doing a *cascading delete* inside of your Device Registry would need to have an impact on the Device Registration API which is mainly intended to reflect requirements from a Hono perspective. There also is no operation defined for doing custom searches on the devices' properties while I can still imagine that such an operation would be helpful to *device managers* working with the Device Registry's browser based user interface. We explicitly distinguished between operations that are mandatory vs optional to implement in the APIs. However, FMPOV that does not mean that we should add more and more (optional) operations to the API. I also think that most implementations of the Device Registration, Credentials and Tenant APIs will be in the same component so that a cascading delete would simply be an implementation detail. And even if Tenant and Device Registration were  implemented in different components, you can simply implement a service invocation (e.g. via HTTP) between the two in order to cascade the delete. In any case, there doesn't seem to be a need to put this into the Device Registration API FMPOV.

balsmn:
@sophokles73 I see your point. The facts that we are using AMQP 1.0 to communicate between Tenant and Device Registration and that the current Device Registration and Credentials Service definitions are prohibited from further extension lead us to this PR.

Then I hope you wouldn't mind the PR #512

If you are fine with #512 I will abandon this PR.

balsmn:
@sophokles73 To your statement `Nothing is preventing you`, I don't agree. Irrespective of whether the services are consumed via AMQP or HTTP, we need either the new functions in the service API or make service API extensible (which is currently final and not allowed).

registration-service-hono-dependency

Here Hono specific code are highlighted in light-brown and the concrete custom implementation are highlighted in blue color. As `BaseRegistrationService` being the primary consumer of messages from the Vertx EventBus (via the final method `processRegistrationMessage()`), we need either the new function mapped in `processRegistrationMessage()` or allow them to be extensible (as suggested in PR #512).

As you said, these new functions don't necessarily make sense for Hono and if the event consumer can't be extended, then unfortunately the only way forward is to replicate `BaseRegistrationService`, `ConfigurationSupportingVerticle` and use our own event address, which I find it inefficient.

sophokles73:
> As BaseRegistrationService being the primary consumer of messages from the Vertx EventBus (via the final method processRegistrationMessage()), we need either the new function mapped in processRegistrationMessage() or allow them to be extensible (as suggested in PR #512).

I agree, we need to make the base classes extensible so that implementations can add their custom operations. Does this mean that if we do something like what you propose in #512 then this PR would be obsolete?

balsmn:
Yes if we do #512  then there is no need for this PR and can be abandoned.

balsmn:
Closing this PR in favor of PR #512

PR #512, opened 2018-02-26
Making the Registration and Credentials Service API extensible

balsmn:
Making the CredentialsService, RegistrationService extensible and also allowing the corresponding client implementation to be extensible. Concrete implementations will be able to extend them as per their need.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientImpl.java
@@ -39,29 +39,29 @@ import io.vertx.proton.ProtonConnection;
  * A Vertx-Proton based client for Hono's Credentials API.
  *
  */
-public final class CredentialsClientImpl extends AbstractRequestResponseClient<CredentialsResult<CredentialsObject>> implements CredentialsClient {
+public class CredentialsClientImpl extends AbstractRequestResponseClient<CredentialsResult<CredentialsObject>> implements CredentialsClient {
 
     private static Logger LOG = LoggerFactory.getLogger(CredentialsClientImpl.class);
     private static final ObjectMapper objectMapper = new ObjectMapper();
 
-    private CredentialsClientImpl(final Context context, final ClientConfigProperties config, final String tenantId) {
+    protected CredentialsClientImpl(final Context context, final ClientConfigProperties config, final String tenantId) {
         super(context, config, tenantId);
     }
 
     @Override
-    protected String getName() {
+    protected final String getName() {
 
         return CredentialsConstants.CREDENTIALS_ENDPOINT;
     }
 
     @Override
-    protected String createMessageId() {
+    protected final String createMessageId() {
 
         return String.format("cred-client-%s", UUID.randomUUID());
     }
 
     @Override
-    protected CredentialsResult<CredentialsObject> getResult(final int status, final String payload) {
+    protected final CredentialsResult<CredentialsObject> getResult(final int status, final String payload) {
         try {
             if (status == HttpURLConnection.HTTP_OK) {
                 return CredentialsResult.from(status, objectMapper.readValue(payload, CredentialsObject.class));
@@ -80,7 +80,7 @@ public final class CredentialsClientImpl extends AbstractRequestResponseClient<C
      * @return The target address.
      * @throws NullPointerException if tenant is {@code null}.
      */
-    public static String getTargetAddress(final String tenantId) {
+    public static final String getTargetAddress(final String tenantId) {
         return String.format("%s/%s", CredentialsConstants.CREDENTIALS_ENDPOINT, Objects.requireNonNull(tenantId));
     }
 
@@ -96,7 +96,7 @@ public final class CredentialsClientImpl extends AbstractRequestResponseClient<C
      * @param creationHandler The handler to invoke with the outcome of the creation attempt.
      * @throws NullPointerException if any of the parameters is {@code null}.
      */
-    public static void create(
+    public static final void create(
             final Context context,
             final ClientConfigProperties clientConfig,
             final ProtonConnection con,
@@ -124,7 +124,7 @@ public final class CredentialsClientImpl extends AbstractRequestResponseClient<C
      * on the service represented by the <em>sender</em> and <em>receiver</em> links.
      */
     @Override
-    public Future<CredentialsObject> get(final String type, final String authId) {
+    public final Future<CredentialsObject> get(final String type, final String authId) {
 
         Objects.requireNonNull(type);
         Objects.requireNonNull(authId);
--- a/client/src/main/java/org/eclipse/hono/client/impl/RegistrationClientImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/RegistrationClientImpl.java
@@ -46,16 +46,16 @@ import io.vertx.proton.ProtonSender;
  * A Vertx-Proton based client for Hono's Registration API.
  *
  */
-public final class RegistrationClientImpl extends AbstractRequestResponseClient<RegistrationResult> implements RegistrationClient {
+public class RegistrationClientImpl extends AbstractRequestResponseClient<RegistrationResult> implements RegistrationClient {
 
     private static final Logger LOG = LoggerFactory.getLogger(RegistrationClientImpl.class);
 
-    RegistrationClientImpl(final Context context, final ClientConfigProperties config, final String tenantId) {
+    protected RegistrationClientImpl(final Context context, final ClientConfigProperties config, final String tenantId) {
 
         super(context, config, tenantId);
     }
 
-    RegistrationClientImpl(final Context context, final ClientConfigProperties config, final String tenantId,
+    protected RegistrationClientImpl(final Context context, final ClientConfigProperties config, final String tenantId,
             final ProtonSender sender, final ProtonReceiver receiver) {
 
         super(context, config, tenantId, sender, receiver);
@@ -68,24 +68,24 @@ public final class RegistrationClientImpl extends AbstractRequestResponseClient<
      * @return The target address.
      * @throws NullPointerException if tenant is {@code null}.
      */
-    public static String getTargetAddress(final String tenantId) {
+    public static final String getTargetAddress(final String tenantId) {
         return String.format("%s/%s", RegistrationConstants.REGISTRATION_ENDPOINT, Objects.requireNonNull(tenantId));
     }
 
     @Override
-    protected String getName() {
+    protected final String getName() {
 
         return RegistrationConstants.REGISTRATION_ENDPOINT;
     }
 
     @Override
-    protected String createMessageId() {
+    protected final String createMessageId() {
 
         return String.format("reg-client-%s", UUID.randomUUID());
     }
 
     @Override
-    protected RegistrationResult getResult(final int status, final String payload) {
+    protected final RegistrationResult getResult(final int status, final String payload) {
 
         return RegistrationResult.from(status, payload);
     }
@@ -104,7 +104,7 @@ public final class RegistrationClientImpl extends AbstractRequestResponseClient<
      * @param creationHandler The handler to invoke with the outcome of the creation attempt.
      * @throws NullPointerException if any of the parameters other than cache manager is {@code null}.
      */
-    public static void create(
+    public static final void create(
             final Context context,
             final ClientConfigProperties clientConfig,
             final CacheManager cacheManager,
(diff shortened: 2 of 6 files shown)


client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientImpl.java
```diff
@@ -39,7 +39,7 @@
  * A Vertx-Proton based client for Hono's Credentials API.
  *
  */
-public final class CredentialsClientImpl extends AbstractRequestResponseClient<CredentialsResult<CredentialsObject>> implements CredentialsClient {
+public class CredentialsClientImpl extends AbstractRequestResponseClient<CredentialsResult<CredentialsObject>> implements CredentialsClient {
```
sophokles73: My understanding is that you want to change this in order to be able to extend the class and add your custom operations. If that is the case, then we should make the existing methods of `CredentialsClientImpl` final in order to prevent subclasses from overriding them.

client/src/main/java/org/eclipse/hono/client/impl/RegistrationClientImpl.java
```diff
@@ -46,7 +46,7 @@
  * A Vertx-Proton based client for Hono's Registration API.
  *
  */
-public final class RegistrationClientImpl extends AbstractRequestResponseClient<RegistrationResult> implements RegistrationClient {
+public class RegistrationClientImpl extends AbstractRequestResponseClient<RegistrationResult> implements RegistrationClient {
```
sophokles73: see above

service-base/src/main/java/org/eclipse/hono/service/credentials/BaseCredentialsService.java
```diff
@@ -232,6 +232,18 @@ private void processCredentialsMessageRemoveOperation(final Message<JsonObject>
         }
     }
 
+    /**
+     * Override the following method to extend RegistrationService with implementation
```
sophokles73: we are talking about the Credentials service, aren't we?

service-base/src/main/java/org/eclipse/hono/service/credentials/BaseCredentialsService.java
```diff
@@ -232,6 +232,18 @@ private void processCredentialsMessageRemoveOperation(final Message<JsonObject>
...
+     * specific operations.
+     *
+     * @param credMsg target credential message to be processed.
+     * @param action implementation specific action to be executed.
+     */
+    public void processOtherCredentialMessage(final Message<JsonObject> credMsg, String action) {
```
sophokles73: Can we rename this to e.g. *processCustomCredentialsMessage*? FMPOV it should be declared *protected* instead of *public* since it is supposed to be invoked by *processCredentialsMessage* only, isn't it?

service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
```diff
@@ -182,15 +182,27 @@ public final void processRegistrationMessage(final Message<JsonObject> regMsg) {
...
     }
 
+    /**
+     * Override the following method to extend RegistrationService with implementation
+     * specific operations.
+     *
+     * @param regMsg registration message to be processed
+     * @param operation implementation specific operation to be executed.
+     */
+    public void processOtherRegistrationMessage(final Message<JsonObject> regMsg, String operation) {
```
sophokles73: *processCustomRegistrationMessage* and declare *protected*

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
```diff
@@ -162,15 +162,27 @@ public final void processTenantMessage(final Message<JsonObject> tenantMsg) {
...
     }
 
+    /**
+     * Override the following method to extend RegistrationService with implementation
+     * specific operations.
+     *
+     * @param tenantMsg target tenant message to be processed.
+     * @param action implementation specific action to be executed.
+     */
+    public void processOtherTenantMessage(final Message<JsonObject> tenantMsg, String action) {
```
sophokles73: `protected void processCustomTenantMessage`

client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientImpl.java
balsmn: I agree. Will do.

client/src/main/java/org/eclipse/hono/client/impl/RegistrationClientImpl.java
balsmn: Will do

service-base/src/main/java/org/eclipse/hono/service/credentials/BaseCredentialsService.java
balsmn: yes, I will correct it.

service-base/src/main/java/org/eclipse/hono/service/credentials/BaseCredentialsService.java
balsmn: Yes, I can rename to `processCustomCredentialsMessage`. But if custom implementation has to override it, it has to be public, right?

service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
balsmn: will change it to `processCustomRegistrationMessage`

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
balsmn: changing the method name here to `processCustomTenantMessage`

service-base/src/main/java/org/eclipse/hono/service/credentials/BaseCredentialsService.java
balsmn: I even thought of making `processCustomCredentialsMessage` abstract, but it will break existing implementations. So went for a public method with default implementation.

sophokles73: I don't think so. See https://stackoverflow.com/a/11626042

balsmn: Oh..yes we are just invoking from within and of course the concrete implementation can override it.

client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientImpl.java
balsmn: Done

client/src/main/java/org/eclipse/hono/client/impl/RegistrationClientImpl.java
balsmn: Done

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
```diff
@@ -162,15 +162,27 @@ public final void processTenantMessage(final Message<JsonObject> tenantMsg) {
...
 
+    /**
+     * Override the following method to extend TenantService with implementation
+     * specific operations.
+     *
+     * @param tenantMsg target tenant message to be processed.
+     * @param action implementation specific action to be executed.
+     */
+    protected void processCustomTenantMessage(final Message<JsonObject> tenantMsg, String action) {
+        log.debug("invalid action in request message [{}]", action);
```
sophokles73: there are three variants of this log statement. Can we please use the same one for all three services?

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
```diff
@@ -162,15 +162,27 @@ public final void processTenantMessage(final Message<JsonObject> tenantMsg) {
...
+                break;
             }
         } catch (final ClassCastException e) {
             log.debug("malformed request message [{}]", e.getMessage());
             reply(tenantMsg, TenantResult.from(HTTP_BAD_REQUEST));
         }
     }
 
+    /**
+     * Override the following method to extend TenantService with implementation
```
sophokles73: IMHO this comment should explain what the purpose of this method is and when and in what context it will be invoked so that implementers can better understand what they need to do. In particular, the comment should include instructions on how to actually convey a response back to the client.

Same applies to the other two services.

balsmn:
@sophokles73 I just now realized. Consequently to make `CredentialsClientImpl` and `RegistrationClientImpl` extendable, I would have to make their constructors public so that the concrete implementations can delegate using `super` keyword. Do you agree?

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
balsmn: will do it.

sophokles73:
> I would have to make their constructors public so that the concrete implementations can delegate using super keyword. Do you agree?

No, it should be sufficient to make them *protected* for that.

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
balsmn: Done

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
```diff
@@ -162,16 +162,43 @@ public final void processTenantMessage(final Message<JsonObject> tenantMsg) {
...
+     *
+     * @param tenantMsg target tenant message to be processed.
+     * @param action implementation specific action to be executed.
+     */
+    protected void processCustomTenantMessage(final Message<JsonObject> tenantMsg, String action) {
+        log.debug("invalid action in request message [{}]", action);
+        reply(tenantMsg, TenantResult.from(HTTP_BAD_REQUEST));
+    }
+
+    protected void reply(final Message<JsonObject> request, final AsyncResult<TenantResult<JsonObject>> asyncActionResult) {
```
sophokles73: should be declared *final*.
IMHO you need to make the *reply* methods of `BaseRegistrationService` and `BaseCredentialsService` *protected* and *final* as well, so that sub-classes can invoke them, as recommended by your JavaDoc example code.

client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientImpl.java
```diff
@@ -80,7 +80,7 @@ protected String createMessageId() {
      * @return The target address.
      * @throws NullPointerException if tenant is {@code null}.
      */
-    public static String getTargetAddress(final String tenantId) {
+    public final static String getTargetAddress(final String tenantId) {
```
sophokles73: please change the order of the modifiers to `public static final`

client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientImpl.java
```diff
@@ -96,7 +96,7 @@ public static String getTargetAddress(final String tenantId) {
      * @param creationHandler The handler to invoke with the outcome of the creation attempt.
      * @throws NullPointerException if any of the parameters is {@code null}.
      */
-    public static void create(
+    public final static void create(
```
sophokles73: please change the order of the modifiers to `public static final`

client/src/main/java/org/eclipse/hono/client/impl/RegistrationClientImpl.java
```diff
@@ -68,24 +68,24 @@
      * @return The target address.
      * @throws NullPointerException if tenant is {@code null}.
      */
-    public static String getTargetAddress(final String tenantId) {
+    public final static String getTargetAddress(final String tenantId) {
```
sophokles73: please change the order of the modifiers to `public static final`

client/src/main/java/org/eclipse/hono/client/impl/RegistrationClientImpl.java
```diff
@@ -104,7 +104,7 @@ protected RegistrationResult getResult(final int status, final String payload) {
      * @param creationHandler The handler to invoke with the outcome of the creation attempt.
      * @throws NullPointerException if any of the parameters other than cache manager is {@code null}.
      */
-    public static void create(
+    public final static void create(
```
sophokles73: please change the order of the modifiers to `public static final`

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
balsmn: Sorry my bad. I should have seen it coming. I will add it.

client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientImpl.java
balsmn: ok

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
balsmn: Sorry, I had it in a different branch and missed actually for this PR.

PR #515, opened 2018-02-27
Hono Tenant API client implementation.

sysexcontrol:
Contained is the definition and implementation of the TenantClient, that is retrievable via the HonoClient interface (as with previous API specific clients).

Extra commits contain:
- some Unit tests for the correct triggering of the Proton client
- integration tests that test the tenant service (based on the file impl) without mocks


client/src/main/java/org/eclipse/hono/client/impl/TenantClientImpl.java
```diff
@@ -0,0 +1,155 @@
...
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.proton.ProtonConnection;
+
+/**
+ * A Vertx-Proton based client for Hono's Tenant API.
+ *
+ */
+public final class TenantClientImpl extends AbstractRequestResponseClient<TenantResult<TenantObject>>
```
balsmn: Can you not make it final in order to extend it. Instead, you can make existing methods final and allow concrete implementations to provide further functionalities. See PR #512 for similar extensions.

sysexcontrol: Good point, was written in parallel :-)
Will remove the class final :+1:

client/src/main/java/org/eclipse/hono/client/TenantClient.java
```diff
@@ -0,0 +1,43 @@
...
+     *         The future will succeed if a response with status 200 has been received from the
+     *         tenant service. The JSON object will then contain values as defined in
+     *         <a href="https://www.eclipse.org/hono/api/tenant-api/#get-tenant">
+     *         Get Tenant</a>.
+     *         <p>
+     *         Otherwise, the future will fail with a {@link ServiceInvocationException} containing
+     *         the (error) status code returned by the service.
+     * @see RequestResponseClient#setRequestTimeout(long)
+     */
+    Future<TenantObject> get();
```
sophokles73: It feels *very, very* awkward that a `TenantClient` instance is *per-se* scoped to a single tenant.

I understand that we do not really have a good concept for restricting a client's access to a subset of the existing tenant information. But at least we could include a *tenantId* parameter here and then simply authorize the client's request by comparing the identifier of the tenant that the client has been authenticated in  with the tenant ID it provides in the method call. If we then later find a better way of expressing a client's authorities, we at least have the right method signature in place.

sysexcontrol: This is exactly what I had already implemented. And then decided against it again, since this would introduce the "one tenant could wish to access the information of another tenant" - for which we do not have a concept. So I decided to eliminate this param again (check for identity is trivial, but I just ignored this param and still get the tenant from the AMQP app props).
The param felt a little like "premature optimization" to me.

I understand the awkwardness (and find the discussion of exactly this point very helpful!), but this reflects exactly the truth we have now. And forces users of the tenant API to think about why there is no param at all. And if we have a concept, there might be a second operation.

Anyway, I am partly fine with reintroducing the `Future<TenantObject> get(String tenantId)` signature again - but it is not as clear anymore and does not match a pure multi-tenancy concept IMHO. Would then vote for currently ignoring the param, so clients are not forced to provide the tenantId again, which was already provided during client generation like:

`
client.connect(new ProtonClientOptions())
            .compose(c -> c.getOrCreateTenantClient(TEST_HTTP_TENANT))
            .setHandler(ctx.asyncAssertSuccess(r -> {
                tenantClient = r;
            }));
`
So the client could just access the tenant information at other places in the code by 

`
tenantClient.get(null);
`

or (if they have the tenant avail)

`
tenantClient(TEST_HTTP_TENANT);
`

, but they do not **have** to provide their own tenantId again.

WDYT? Reintroduce the tenant param and ignore it? Or check it for identity and force clients to always provide the same tenant again (maybe use null as option to use the already provided tenantId)?

sophokles73: FMPOV the client should not need to specify a tenant ID when it creates the `TenantClient`, i.e.
```
client.connect(new ProtonClientOptions())
  .compose(c -> c.getOrCreateTenantClient())
  .setHandler(ctx.asyncAssertSuccess());
```
It can then do `client.get(tenantId)` for whichever tenant it needs information about. Whether the tenant service actually authorizes the request is another topic.

sophokles73: BTW the [Tenant API as it is specified currently](https://www.eclipse.org/hono/api/tenant-api/#preconditions) also does not scope the client's links to a particular tenant ID.

client/src/main/java/org/eclipse/hono/client/impl/TenantClientImpl.java
```diff
@@ -0,0 +1,155 @@
...
+        implements TenantClient {
+
+    private static final Logger LOG = LoggerFactory.getLogger(TenantClientImpl.class);
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    private TenantClientImpl(final Context context, final ClientConfigProperties config, final String tenantId) {
+        super(context, config, tenantId);
+    }
```
balsmn: Also, the Constructor needs to be protected to be extensible

client/src/main/java/org/eclipse/hono/client/impl/TenantClientImpl.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @Override
+    protected String createMessageId() {
+
+        return String.format("%s-%s", TenantConstants.MESSAGE_ID_PREFIX, UUID.randomUUID());
+    }
+
+    @Override
+    protected TenantResult<TenantObject> getResult(final int status, final String payload) {
```
balsmn: Can you make all the existing methods to be final, so that any extensions of the `TenantClientImpl` doesn't modify them.

sysexcontrol: Of course, when I removed the final from the class declaration, I introduced final for the methods instead. Did not push it yet ;-)

sysexcontrol:
@sophokles73 : pushed new commits. The tenant client is now not scoped to a tenant anymore, but instead the authorization checks are scoped to the tenant on a per message base - this enables the correct handling of multi-tenancy, while not opening a sender/receiver link for all tenants.

After we agreed on this PR, I would write a short documentation about how a multi-tenancy setup (at least for protocol adapters) has to be done, covering the aspects that are now possible:
- shared for all tenants
- scoped to exactly one tenant
- scoped to a specific number of tenants

The same strategy should be considered for the other Request-Response-APIs (Registration and Credentials API), but this is a different story (and would be different PRs with lower prio).

tests/src/test/java/org/eclipse/hono/tests/registry/AbstractDeviceRegistryAmqpSupportIT.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Abstract class to support integration tests for the device registry.
+ *
+ */
+public abstract class AbstractDeviceRegistryAmqpSupportIT {
+
+    /**
+     * Build a HonoClient to access the device registry service.
+     * @param vertx
+     * @return
```
sophokles73: please provide a proper JavaDoc comment

tests/src/test/java/org/eclipse/hono/tests/registry/AbstractDeviceRegistryAmqpSupportIT.java
```diff
@@ -0,0 +1,62 @@
...
+        clientProps.setName("test");
+        clientProps.setHost(IntegrationTestSupport.HONO_DEVICEREGISTRY_HOST);
+        clientProps.setPort(IntegrationTestSupport.HONO_DEVICEREGISTRY_AMQP_PORT);
+        clientProps.setUsername(IntegrationTestSupport.HONO_USER);
+        clientProps.setPassword(IntegrationTestSupport.HONO_PWD);
+        return new HonoClientImpl(vertx, clientProps);
+
+    }
+
+    protected static void shutdownDeviceRegistryClient(final TestContext ctx, final Vertx vertx, final HonoClient client) {
```
sophokles73: missing JavaDoc

tests/src/test/java/org/eclipse/hono/tests/registry/AbstractDeviceRegistryAmqpSupportIT.java
```diff
@@ -0,0 +1,62 @@
...
+import org.eclipse.hono.client.HonoClient;
+import org.eclipse.hono.client.impl.HonoClientImpl;
+import org.eclipse.hono.config.ClientConfigProperties;
+import org.eclipse.hono.tests.IntegrationTestSupport;
+
+/**
+ * Abstract class to support integration tests for the device registry.
+ *
+ */
+public abstract class AbstractDeviceRegistryAmqpSupportIT {
```
sophokles73: there doesn't seem to be any need to extend this helper class, is there?
So I guess the test classes shouldn't. Instead this class should be renamed to `DeviceRegistryAmqpTestSupport` and get a private default constructor.

client/src/main/java/org/eclipse/hono/client/StatusCodeMapper.java
```diff
@@ -92,6 +92,8 @@ public static final ServiceInvocationException from(final ErrorCondition error)
 
         if (AmqpError.RESOURCE_LIMIT_EXCEEDED.equals(error.getCondition())) {
             return new ClientErrorException(HttpURLConnection.HTTP_FORBIDDEN, error.getDescription());
+        } else if (AmqpError.UNAUTHORIZED_ACCESS.equals(error.getCondition())) {
+            return new ClientErrorException(HttpURLConnection.HTTP_UNAUTHORIZED, error.getDescription());
```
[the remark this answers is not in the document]
sysexcontrol: Agree -  403 is matching better. Will change it.

client/src/main/java/org/eclipse/hono/client/HonoClient.java
```diff
@@ -262,6 +262,17 @@
...
     Future<CredentialsClient> getOrCreateCredentialsClient(String tenantId);
 
+    /**
+     * Gets a client for interacting with Hono's <em>Tenant</em> API.
+     *
+     * @return A future that will complete with the tenant client (if successful) or
+     *         fail if the client cannot be created, e.g. because the underlying connection
+     *         is not established or if a concurrent request to create a client for the same
+     *         tenant is already being executed.
+     * @throws NullPointerException if the tenant is {@code null}.
```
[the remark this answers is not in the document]
sysexcontrol: yes, will delete it.

client/src/main/java/org/eclipse/hono/client/TenantClient.java
```diff
@@ -0,0 +1,45 @@
...
+     * @return A future indicating the result of the operation.
+     *         <p>
+     *         The future will succeed if a response with status 200 has been received from the
+     *         tenant service. The JSON object will then contain values as defined in
+     *         <a href="https://www.eclipse.org/hono/api/tenant-api/#get-tenant">
+     *         Get Tenant</a>.
+     *         <p>
+     *         Otherwise, the future will fail with a {@link ServiceInvocationException} containing
+     *         the (error) status code returned by the service.
+     * @see RequestResponseClient#setRequestTimeout(long)
```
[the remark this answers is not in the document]
sysexcontrol: Good question - with the strategy "the client is not scoped to a tenant anymore", a null value does not make sense to me. So I will change it to throwing a NullPointerException in this case.

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -142,15 +142,16 @@ protected static final void setApplicationProperties(final Message msg, final Ma
...
             for (final Map.Entry<String, ?> entry: properties.entrySet()) {
-                if (entry.getValue() instanceof List) {
-                    throw new IllegalArgumentException(String.format("Application property %s can't be a List", entry.getKey()));
-                } else if (entry.getValue() instanceof Map) {
-                    throw new IllegalArgumentException(String.format("Application property %s can't be a Map", entry.getKey()));
-                } else if (entry.getValue().getClass().isArray()) {
-                    throw new IllegalArgumentException(String.format("Application property %s can't be an Array", entry.getKey()));
-                } else {
-                    propsToAdd.put(entry.getKey(), entry.getValue());
+                if (entry.getValue() != null) {
```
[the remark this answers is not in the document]
sysexcontrol: 2 remarks:
- `entry.getValue().getClass().isArray()` was the code line I aimed at regarding being NP safe
- I wanted to (still) support null values in the message properties (since it was originally implemented to allow a null value for tenantId).
Do you have concerns with supporting null values in message properties? Thinking about it now it seems to have no different semantics than just not setting the key-value-pair at all, but it could be left to the client to decide about that IMHO.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public void addAdapterConfiguration(final Map<String, String> adapterConfiguration) {
+        if (adapterConfigurations == null) {
+            adapterConfigurations= new LinkedList<>();
+        }
+        adapterConfigurations.add(adapterConfiguration);
+    }
+
+    public static TenantObject from(final String tenantId, final Boolean enabled) {
```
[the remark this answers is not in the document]
sysexcontrol: :+1:

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
```diff
@@ -225,11 +227,12 @@ protected final void handleMessage(final ProtonConnection con, final ProtonRecei
      * Subclasses may override this method in order to do more sophisticated checks.
      * 
      * @param clientPrincipal The client.
+     * @param tenantId The tenantId (if not part of the endpoint of the resource already).
      * @param resource The resource the operation belongs to.
      * @param operation The operation.
      * @return The outcome of the check.
      */
-    protected Future<Boolean> isAuthorized(final HonoUser clientPrincipal, final ResourceIdentifier resource, final String operation) {
+    protected Future<Boolean> isAuthorized(final HonoUser clientPrincipal, final ResourceIdentifier resource, final String tenantId, final String operation) {
```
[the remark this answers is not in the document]
sysexcontrol: That makes sense, will change it.

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -81,25 +83,26 @@
      * 
      * @param context The vert.x context to run message exchanges with the peer on.
      * @param config The configuration properties to use.
-     * @param tenantId The identifier of the tenant that the client is scoped to.
-     * @throws NullPointerException if any of the parameters is {@code null}.
+     * @param tenantId The identifier of the tenant that the client can be scoped to. May be {@code null}.
```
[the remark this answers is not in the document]
sysexcontrol: :+1: will correct it.

client/src/main/java/org/eclipse/hono/client/impl/TenantClientImpl.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /**
+     * Gets the AMQP <em>target</em> address to use for sending requests to Hono's Tenant API endpoint.
+     *
+     * @return The target address.
+     * @throws NullPointerException if tenant is {@code null}.
+     */
+    public final static String getTargetAddress() {
+        return String.format("%s", TenantConstants.TENANT_ENDPOINT);
```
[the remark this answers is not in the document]
sysexcontrol: oh oh, yes :+1:

client/src/main/java/org/eclipse/hono/client/impl/TenantClientImpl.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Creates a new tenant API client for a tenant.
+     *
+     * @param context The vert.x context to run all interactions with the server on.
+     * @param clientConfig The configuration properties to use.
+     * @param con The AMQP connection to the server.
+     * @param senderCloseHook A handler to invoke if the peer closes the sender link unexpectedly.
+     * @param receiverCloseHook A handler to invoke if the peer closes the receiver link unexpectedly.
+     * @param creationHandler The handler to invoke with the outcome of the creation attempt.
+     * @throws NullPointerException if any of the parameters is {@code null}.
```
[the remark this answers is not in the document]
sysexcontrol: will correct it.

tests/src/test/java/org/eclipse/hono/tests/registry/AbstractDeviceRegistryAmqpSupportIT.java
sysexcontrol: Good proposal, will change it.

client/src/test/java/org/eclipse/hono/client/impl/AbstractRequestResponseClientTest.java
```diff
@@ -54,7 +54,7 @@
  *
  */
 @RunWith(VertxUnitRunner.class)
-public class AbstractRequestResponseClientTest {
+public class AbstractRequestResponseClientTest extends AbstractClientUnitTestSupport {
```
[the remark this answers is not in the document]
sysexcontrol: Yes, that is common sense regarding inheritance and unit tests.
Encapsulation in helper class increases some duplications again, but I agree that the tests are clearer (especially if they experience evolution in the future).
I will come up with a helper class and eliminate the inheritance.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+
+/**
+ * Encapsulates the tenant information that was found by the get operation of the
+ * <a href="https://www.eclipse.org/hono/api/Tenant-API/">Tenant API</a>.
+ * <p>
+ * Is mapped internally from json representation by jackson-databind.
```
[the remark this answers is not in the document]
sysexcontrol: Will delete that part.

sysexcontrol:
@sophokles73 : thanks for the review, I pushed 2 new commits that reflect the discussion.

Will solve the conflict later before merge (IMHO this makes the review process more handable than rebasing it now just due to one file conflict).

core/src/main/java/org/eclipse/hono/util/TenantConstants.java
```diff
@@ -24,7 +24,10 @@
  */
 
 public final class TenantConstants extends RequestResponseApiConstants {
-    /* message id prefix */
+
+    /**
+     *  All messages for the Tenant API will start with this prefix.
```
sophokles73: Let's hope that it's only the message's IDs that start with this prefix ...

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -73,6 +71,14 @@ public void addAdapterConfiguration(final Map<String, String> adapterConfigurati
         adapterConfigurations.add(adapterConfiguration);
     }
 
+    /**
+     * Creates a {@link TenantObject} for a tenantId and the enabled property.
```
sophokles73: There really is no need to link to the return type. The JavaDoc tool will add links to the return types of all methods by default.

client/src/test/java/org/eclipse/hono/client/impl/HonoClientUnitTestHelper.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Helper class that provides mocks for objects that are needed by unit tests using a Hono client.
+ *
+ */
+public final class HonoClientUnitTestHelper {
+
+    private HonoClientUnitTestHelper() {}
+
+    /**
+     * Create a mock for the Vert.x context that is used for the unit test.
```
sophokles73: I like that you added JavaDocs :+1: 
They would be even more helpful, if they would describe the returned mock's behavior, e.g. in this case something like
> Creates a mocked vert.x *Context* which immediately invokes any handler that is passed to its *runOnContext* method.

Otherwise, any user of these methods will still need to look into the source code in order to be able to use any of the methods properly.

client/src/test/java/org/eclipse/hono/client/impl/HonoClientUnitTestHelper.java
```diff
@@ -0,0 +1,79 @@
...
+        doAnswer(invocation -> {
+            Handler<Void> handler = invocation.getArgument(0);
+            handler.handle(null);
+            return null;
+        }).when(context).runOnContext(any(Handler.class));
+        return context;
+    }
+
+    /**
+     * Create a mock for a Proton sender, usable for a Hono client in the unit test.
```
sophokles73: see above

client/src/test/java/org/eclipse/hono/client/impl/HonoClientUnitTestHelper.java
```diff
@@ -0,0 +1,79 @@
...
+    public static final ProtonSender mockProtonSender() {
+
+        final ProtonSender sender = mock(ProtonSender.class);
+        when(sender.isOpen()).thenReturn(Boolean.TRUE);
+
+        return sender;
+    }
+
+    /**
+     * Create a mock for a Proton receiver, usable for a Hono client in the unit test.
```
sophokles73: see above

core/src/main/java/org/eclipse/hono/util/TenantConstants.java
sysexcontrol: Like the slight sarcasm :+1: 
Will change it to 
`Messages that are sent by the Hono client for the Tenant API use this as a prefix for the messageId.`
IMHO this improves the description...

client/src/test/java/org/eclipse/hono/client/impl/HonoClientUnitTestHelper.java
sysexcontrol: Fully agree - will improve the description!

PR #522, opened 2018-03-05
Making the HonoClient implementation extensible

balsmn:
Allowing HonoClient implementation to be extensible so that specific custom implementations can be instantiated without replicating the whole HonoClient implementation. 

The following methods will enable custom client instances to be instantiated
```
HonoClientImpl#newRegistrationClient()
HonoClientImpl#newCredentialsClient()
```

and enables cleanup by using the following methods
```
HonoClientImpl#removeRegistrationClient() 
HonoClientImpl#removeCredentialsClient()
```

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
@@ -56,7 +56,7 @@ import io.vertx.proton.ProtonDelivery;
 /**
  * A helper class for creating Vert.x based clients for Hono's arbitrary APIs.
  */
-public final class HonoClientImpl implements HonoClient {
+public class HonoClientImpl implements HonoClient {
 
     private static final Logger LOG = LoggerFactory.getLogger(HonoClientImpl.class);
 
@@ -125,7 +125,7 @@ public final class HonoClientImpl implements HonoClient {
      * @param manager The cache manager.
      * @throws NullPointerException if manager is {@code null}.
      */
-    public void setCacheManager(final CacheManager manager) {
+    public final void setCacheManager(final CacheManager manager) {
         this.cacheManager = Objects.requireNonNull(manager);
     }
 
@@ -133,7 +133,7 @@ public final class HonoClientImpl implements HonoClient {
      * {@inheritDoc}
      */
     @Override
-    public Future<Boolean> isConnected() {
+    public final Future<Boolean> isConnected() {
 
         final Future<Boolean> result = Future.future();
         context.runOnContext(check -> {
@@ -142,7 +142,7 @@ public final class HonoClientImpl implements HonoClient {
         return result;
     }
 
-    private Future<Void> checkConnected() {
+    protected final Future<Void> checkConnected() {
 
         final Future<Void> result = Future.future();
         if (isConnectedInternal()) {
@@ -168,11 +168,17 @@ public final class HonoClientImpl implements HonoClient {
         }
     }
 
+    protected final ProtonConnection getConnection() {
+        synchronized (connectionLock) {
+            return this.connection;
+        }
+    }
+
     /**
      * {@inheritDoc}
      */
     @Override
-    public Future<HonoClient> connect() {
+    public final Future<HonoClient> connect() {
         return connect(null, null);
     }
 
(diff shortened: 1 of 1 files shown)


balsmn:
I had a discussion about this PR with @sysexcontrol. Unfortunately I could add him as one of the reviewers to this PR.

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -361,23 +367,23 @@ private void reconnect(
...
+    public final Future<MessageSender> getOrCreateTelemetrySender(final String tenantId, final String deviceId) {
 
         Objects.requireNonNull(tenantId);
         return getOrCreateSender(
                 TelemetrySenderImpl.getTargetAddress(tenantId, deviceId),
                 () -> createTelemetrySender(tenantId, deviceId));
     }
 
-    private Future<MessageSender> createTelemetrySender(
+    private final Future<MessageSender> createTelemetrySender(
```
sophokles73: private methods cannot be overridden, no need to declare them *final*

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -592,7 +598,17 @@ private void reconnect(
         return result;
     }
 
-    private Future<RequestResponseClient> newCredentialsClient(final String tenantId) {
+    /**
+     * Custom implementation of {@link CredentialsClient} can be instantiated by overriding this
```
sophokles73: Please re-phrase this JavaDoc.
The first sentence should describe shortly what this method's purpose is. Subsequent paragraphs can then go into specifics and then you may add recommendations for properly overriding the method at the end.
See http://www.oracle.com/technetwork/articles/java/index-137868.html

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -642,7 +658,17 @@ private void removeCredentialsClient(final String tenantId) {
         return result;
     }
 
-    private Future<RequestResponseClient> newRegistrationClient(final String tenantId) {
+    /**
+     * Custom implementation of {@link RegistrationClient} can be instantiated by overriding this
```
sophokles73: re-phrase

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -642,7 +658,17 @@ private void removeCredentialsClient(final String tenantId) {
...
+     * method. Any such instance should be scoped to the given tenantId.
+     *
```
sysexcontrol: We always put a `<p>` after the introduction to separate it from the detailed explanation.

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -592,7 +598,17 @@ private void reconnect(
         return result;
     }
 
-    private Future<RequestResponseClient> newCredentialsClient(final String tenantId) {
+    /**
+     * Custom implementation of {@link CredentialsClient} can be instantiated by overriding this
+     * method. Any such instance should be scoped to the given tenantId.
+     *
```
sysexcontrol: We always put a `<p>` after the introduction to separate it from the detailed explanation.

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
balsmn: oops. thats was not supposed to be there. will correct it.

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -592,7 +598,18 @@ private void reconnect(
         return result;
     }
 
-    private Future<RequestResponseClient> newCredentialsClient(final String tenantId) {
+    /**
+     * Creates a new instance of {@link CredentialsClient} scoped for the given tenant identifier.
+     * <p>
+     * Custom implementation of {@link CredentialsClient} can be instantiated by overriding this method. Any such
+     * instance should be scoped to the given tenantId.Custom extension of {@link HonoClientImpl} must invoke
```
sysexcontrol: Missing space at beginning of sentence.

balsmn: Done

PR #525, opened 2018-03-06
Document http access to the Tenant API of the device registry.

sysexcontrol:

### DIFF
--- a/site/content/user-guide/device-registry.md
+++ b/site/content/user-guide/device-registry.md
@@ -6,8 +6,12 @@ weight = 205
 In addition to the AMQP 1.0 based API endpoints, the Device Registry also exposes HTTP resources for managing the contents of the registry.
 
 <!--more-->
-Please note that the HTTP resources exposed by the Device Registry are **not** part of the *official* definition of the Device Registration and Credentials APIs but have
-been implemented to provide convenient access to the registry using command line tools like *curl* or *HTTPie*. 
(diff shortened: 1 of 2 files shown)


site/content/user-guide/device-registry.md
```diff
@@ -6,9 +6,14 @@ weight = 205
 In addition to the AMQP 1.0 based API endpoints, the Device Registry also exposes HTTP resources for managing the contents of the registry.
 
 <!--more-->
-Please note that the HTTP resources exposed by the Device Registry are **not** part of the *official* definition of the Device Registration and Credentials APIs but have
+Please note that the HTTP resources exposed by the Device Registry are **not** part of the *official* definition of the Device Registration, the Credentials and (since Hono version 0.6) the Tenant APIs but have
 been implemented to provide convenient access to the registry using command line tools like *curl* or *HTTPie*. 
 
+{{% warning %}}
+The HTTP resources offered here are not secured. In production scenarios this probably needs to be addressed by securing all resources
```
sophokles73: It is fair to point out that, among many other shortcomings, the Device Registry's http endpoint also does not authorize users. AFAIC we should not hint at how to make the Device Registry work in production. Instead, we should encourage people to simply **not** use the Device Registry for production purposes as we do in many other locations already.

sysexcontrol:
@sophokles73 : update warning text.

PR #527, opened 2018-03-07
making Registration, Credentials and Tenant AMQP endpoint extensible …

balsmn:
making Registration, Credentials and Tenant AMQP endpoint extensible so that custom request verification can be used. 

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsAmqpEndpoint.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsAmqpEndpoint.java
@@ -32,7 +32,7 @@ import io.vertx.core.json.JsonObject;
  * It receives AMQP 1.0 messages representing requests and sends them to an address on the vertx
  * event bus for processing. The outcome is then returned to the peer in a response message.
  */
-public final class CredentialsAmqpEndpoint extends RequestResponseEndpoint<ServiceConfigProperties> {
+public class CredentialsAmqpEndpoint extends RequestResponseEndpoint<ServiceConfigProperties> {
 
     /**
      * Creates a new credentials endpoint for a vertx instance.
@@ -46,13 +46,13 @@ public final class CredentialsAmqpEndpoint extends RequestResponseEndpoint<Servi
 
 
     @Override
-    public String getName() {
+    public final String getName() {
         return CredentialsConstants.CREDENTIALS_ENDPOINT;
     }
 
 
     @Override
-    public void processRequest(final Message msg, final ResourceIdentifier targetAddress, final HonoUser clientPrincipal) {
+    public final void processRequest(final Message msg, final ResourceIdentifier targetAddress, final HonoUser clientPrincipal) {
 
         final JsonObject credentialsMsg = CredentialsConstants.getCredentialsMsg(msg, targetAddress);
 
@@ -82,7 +82,7 @@ public final class CredentialsAmqpEndpoint extends RequestResponseEndpoint<Servi
     }
 
     @Override
-    protected Message getAmqpReply(final io.vertx.core.eventbus.Message<JsonObject> message) {
+    protected final Message getAmqpReply(final io.vertx.core.eventbus.Message<JsonObject> message) {
         return CredentialsConstants.getAmqpReply(CredentialsConstants.CREDENTIALS_ENDPOINT, message.body());
     }
 }
--- a/service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationAmqpEndpoint.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationAmqpEndpoint.java
@@ -33,7 +33,7 @@ import io.vertx.core.json.JsonObject;
  * It receives AMQP 1.0 messages representing requests and sends them to an address on the vertx
  * event bus for processing. The outcome is then returned to the peer in a response message.
  */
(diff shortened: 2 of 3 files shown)


service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsAmqpEndpoint.java
```diff
@@ -32,7 +32,7 @@
  * It receives AMQP 1.0 messages representing requests and sends them to an address on the vertx
  * event bus for processing. The outcome is then returned to the peer in a response message.
  */
-public final class CredentialsAmqpEndpoint extends RequestResponseEndpoint<ServiceConfigProperties> {
+public class CredentialsAmqpEndpoint extends RequestResponseEndpoint<ServiceConfigProperties> {
```
sysexcontrol: Please make the `getName()` method final, too.

service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationAmqpEndpoint.java
```diff
@@ -33,7 +33,7 @@
  * It receives AMQP 1.0 messages representing requests and sends them to an address on the vertx
  * event bus for processing. The outcome is then returned to the peer in a response message.
  */
-public final class RegistrationAmqpEndpoint extends RequestResponseEndpoint<ServiceConfigProperties> {
+public class RegistrationAmqpEndpoint extends RequestResponseEndpoint<ServiceConfigProperties> {
```
sysexcontrol: Please make the `getName()` method final, too.

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantAmqpEndpoint.java
```diff
@@ -37,7 +37,7 @@
  * messages representing requests and sends them to an address on the vertx event bus for processing. The outcome is
  * then returned to the peer in a response message.
  */
-public final class TenantAmqpEndpoint extends RequestResponseEndpoint<ServiceConfigProperties> {
+public class TenantAmqpEndpoint extends RequestResponseEndpoint<ServiceConfigProperties> {
```
sysexcontrol: Please make the `getName()` method final, too.

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsAmqpEndpoint.java
balsmn: No. It was intentional so that overriding classes can give a custom name.

service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationAmqpEndpoint.java
balsmn: No. It was intentional so that overriding classes can give a custom name.

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantAmqpEndpoint.java
balsmn: Same as above.

balsmn:
@sysexcontrol Shortened the commit message

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsAmqpEndpoint.java
sysexcontrol: That is about obvious - can you explain a little why that makes sense?
IMHO if you customize the, let's say, credentials endpoint - why do you need another name for that? Would you have more than one credentials endpoint (the original one AND the customized one?).
This is currently not really clear to me ;-)

balsmn: We could name it differently, say with a version in it and we know when we instantiate, which we instantiated.

balsmn: I kept it quite similar to https://github.com/eclipse/hono/blob/master/service-base/src/main/java/org/eclipse/hono/service/amqp/AmqpServiceBase.java#L74

sysexcontrol: The `getName()` is currently mainly used for logging output of the endpoint, BUT:
not only. Which might be a piece to refactor, but since it is currently not prepared for that, I do not see the importance to let this be customized.
With the current code, there might be problems if you customize the name, but do not customize the corresponding targetAddresses the same way - so I currently stick to keeping the `getName()` method `final`.

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantAmqpEndpoint.java
balsmn: Done

service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationAmqpEndpoint.java
balsmn: Done

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsAmqpEndpoint.java
balsmn: Done

PR #534, opened 2018-03-16
Check that constructors of utility-only classes are hidden

ctron:
This change makes use of the checkstyle module "HideUtilityClassConstructor" in order to find utility-only classes which have an accessible constructor. If the class is a utility-only class creating a new instance from it is an error.

With a few exceptions Hono already follows this approach. So aside from adding the check only a few classes had to be fixed.


ctron:
I created this as a PR as this is also about code style, and everyone has an opinion there :grin:

Issue #535, opened 2018-03-16
Define device connection notifications


ctron:
This is a sub-task of #506

In order to close #506 we need to define an API (services, addresses, messages, …) on how the device connection notifications should look like.

ctron:
We are currently running a discussion around that on the mailing list: https://dev.eclipse.org/mhonarc/lists/hono-dev/msg00753.html

sysexcontrol:
@ctron : I made a PR for discussion only #537 with a seq diag containing a possible message flow (including your ideas mentioned on the mailing list). Can you have a look and comment? Thanks!

ctron:
Continuing from the mailing list:

>    > Properties:
>    >     content-type = "hono/connectivity"
>
> "hono/notification"

My idea behind the content type was to use "notification" as address, but be more specific with the message (`hono/connectivity`). If we have others, the consumer could filter based on that.

>    >    subject = "connected" | "disconnected"
>
> "connect" | "disconnect"

I used "connect" and "disconnect" initially. I am fine with both. But having the past tense seemed more appropriate as the event did already happen. "connect" and "disconnect" could be misinterpreted as a request to connect/disconnect. But I am fine with both.

>    > Application Properties
>    >     device_id = <device id>
>    >     tenant_id = <tenant_id>
>
>
> We have the tenant already in the address and with other APIs defined on a tenant base, we require  > only the device_id to be avail
> as application property.
> So: I would skip the tenant_id here.

I agree we could drop the tenant ID. But I followed the "standard response properties" of the device registry [1], which includes it.

[1] https://www.eclipse.org/hono/api/device-registration-api/#standard-response-properties

sysexcontrol:
> My idea behind the content type was to use "notification" as address, but be more specific with the message (`hono/connectivity`). If we have others, the consumer could filter based on that.

I see - so  `hono/connectivity` is fine for me.

>I used "connect" and "disconnect" initially. I am fine with both. But having the past tense seemed more appropriate as the event did already happen. "connect" and "disconnect" could be misinterpreted as a request to connect/disconnect. But I am fine with both.

Here I also see your point and agree to that - doesn't make a huge difference, but sounds reasonable.
So: "connected | disconnected" should be the name of the notification :+1: 


> I agree we could drop the tenant ID. But I followed the "standard response properties" of the device registry [1], which includes it.

[...]

PR #537, opened 2018-03-19
[#535] Discussion PR for device connection notifications.

sysexcontrol:
Just as a base for discussion:
I made a sequence diagram for the connect notification.

This PR is not intended to be ever merged, the results from the discussion should be contained in a different PR.


ctron:
Maybe I miss something, but wouldn't it be sufficient of the protocol adapter sends out the message?

sysexcontrol:
You mean: why is there still Hono Messaging in between?
If that is your question: good point, but currently we have not eliminated it. So I have put it in here still.

sysexcontrol:
Regarding the disposition frame: we send even telemetry data now unsettled, since we had no proper flow control otherwise. But the protocol adapters just throw away the acks in this case.
So, if that was your question, this fact is not reflected in the seq diag yet (should update it).

ctron:
Yes that was my question:
> You mean: why is there still Hono Messaging in between?
If that is your question: good point, but currently we have not eliminated it. So I have put it in here still.

I see your point. My initial assumption was that the protocol adapter will be in charge of sending out the notification. On the one side I don't think piping this through the messaging component adds much value. On the other side I also think that we shouldn't make this decision in the context of this issue.

pellmann:
The sequence should include the consumer application, I think - the AMQP network itself does not has a consumer with a receiver link out of the box.

ctron:
Unless we would broker this, e.g. in a queue:
> The sequence should include the consumer application, I think - the AMQP network itself does not has a consumer with a receiver link out of the box.

But then we should call it "queue".

sysexcontrol:
Ok, I will add a consumer - but we do not show that so far for telemetry either.
Without consumer there will be no credits to send notifications, but of course the AMQP network is always in between the PA, Hono Messaging (will be eliminated), and the consumer, which will result in an even longer sequence diagram.
Anyway, will update this and we will see how this looks :-)

sysexcontrol:
The notifications we are thinking of currently would never be brokered IMHO.
And brokered notifications - wouldn't they be better sent via the Event API (e.g. devices crying for help ;-))?

sysexcontrol:
@pellmann @ctron : update the seq diag to contain the consumer application expliclity

sophokles73:
> The sequence should include the consumer application, I think - the AMQP network itself does not has a consumer with a receiver link out of the box.

Why should it? It is Hono Messaging (or the protocol adapters) which maintain a send link to the AMQP Messaging Network for sending notifications. Whether there are any applications connected to the network or if the notifications are brokered, is of no real concern to Hono FMPOV. Hono should try to deliver the notifications and that's it. When we use QoS 1 we could even implement a re-try mechanism for improving the chance of delivery to applications. However, if the network uses a broker to buffer the notifications is a configuration/implementation detail of the network. I therefore also do not think that the consuming application is relevant for the sequence diagram.

sophokles73:
> The notifications we are thinking of currently would never be brokered IMHO.
And brokered notifications - wouldn't they be better sent via the Event API (e.g. devices crying for help ;-))?

They should be sent using the Event API if they are events. Maybe this is the first discussion we need to have. Are we talking about *notifications* or *events*?

FMPOV we agreed to have a separate API (Notification) because we wanted to have a separate channel for *life cycle* events. IIRC the Kapua people reasoned that it would be easier to *subscribe* to a separate channel in order to manage the device life cycle than to filter the life cycle events from the *event stream*. 

The Event API (so far) has been intended to be used for conveying events that happen as part of the device's *operational purpose*, i.e. a smoke sensor detects smoke etc. The reason why we did not want to send them as telemetry was mainly because we didn't want to require them to *stand in line* with the low priority telemetry messages.

However, thinking about it now,  the distinction between *event* and *notification* seems a little fuzzy and it is hard to see why the *life cycle events* (as the name already indicates) should not be published via the Event API. Using a specific content type for notifications would still make them easy to detect/filter when consuming events using the Event API, in this case.

sysexcontrol:
> However, thinking about it now, the distinction between event and notification seems a little fuzzy and it is hard to see why the life cycle events (as the name already indicates) should not be published via the Event API. Using a specific content type for notifications would still make them easy to detect/filter when consuming events using the Event API, in this case.

To me the lifecycle notifications should be separated from the Event API. 
Reasons for this:
- Maybe I miss something, but I cannot see why they ever should be brokered (while events probably often will be ?). So they would have to be separated from events by inspecting the message, which increases the complexity of the setup IMHO (while of course possible).
- the amount of lifecycle notifications will IMHO be much higher than events, so the events would have to "stand in line" with lifecycle events. Which is not as bad as telemetry data, but still could affect their delivery.


> They should be sent using the Event API if they are events. Maybe this is the first discussion we need to have. Are we talking about notifications or events?

FMPOV we should only name them `life-cycle notifications` (to separate them from events also in the words that are used).

WDYT?

ctron:
I think that the main difference between "events" and "notifications" are, that events are functional messages where notifications are technical events. The latter being generated by the system (hono) itself rather than by devices.

If lifecycle events are buffered or not might depend on the system setup. For Kapua I would assume buffering/queuing is desired. Others might even completely ignore those messages. QoS 1, with re-try would definitely add help in the buffering scenario.

I also think that processing from two distinct addresses would make consuming those messages (both notifications and events) much easier.

sysexcontrol:
> events are functional messages where notifications are technical events. The latter being generated by the system (hono) itself rather than by devices.

I like this phrasing :+1: 

I am currently working on a very first draft version of the API definition page and will put this at the very beginning for a clarification.
As soon as the first version of the page is ready I will push it here for discussion.

sysexcontrol:
Since there were several concerns about defining a completely new API for notifications, I stopped the definition description and rather documented what we have discussed inside the Event API.

The idea behind is that we can reuse all existing code (clients, endpoint, etc.) and define standardized events based on the `content-type` by defining the JSON payload for this `content-type`.

In addition to our ideas we had before, the usage of the Event API offer devices the possibility to send such notifications themselves (instead of "only" the protocol adapters).
This could be helpful IMHO to support devices that know themselves when they are ready to receive a command e.g and report this via such a lifecycle event.

Please have a detailed look into what I pushed as commit as extension to the Event API description and give feedback!
Thanks in advance...

sophokles73:
> I think that the main difference between "events" and "notifications" are, that events are functional messages where notifications are technical events. The latter being generated by the system (hono) itself rather than by devices. 

I find the distinction between events and notifications simply based on the source (i.e. who produced the event) rather arbitrary. Does this mean that a consumer of notifications will only (can only) ever be interested in situations that have been detected and reported by Hono components? FMPOV a consumer of events should be able to quickly determine the *nature/semantics* of an event and then decide whether it wants to/needs to process the event or not. Using the content-type we have means to enable consumers to do exactly this. In some cases the AMQP Messaging Network will also allow the consumer to specify filter criteria when opening a receiver link for the events. In any case, I would rather not start to invent new APIs without good reasons to do so. In particular not, when we already have a mechanism in place for conveying events related to devices. I propose to start off with using the existing Event API for publishing connection state events for devices and use a specific content-type to mark them as such, so that consumers can easily identify them (and then process or ignore them).

sysexcontrol:
The way @sophokles73 outlined is the content of the commit 936b9e9 pushed for discussion just before - usage of the Event API for notification purposes.

sysexcontrol:
@ctron: could you have a look into the definitions I put to the Event page (see commit 936b9d9)?
If we could work on that until tomorrow it would be perfect - thanks!

DanielMaier-BSI:
I was just thinking about what should happen with devices connected through gateway mode. Should hono send notification for the specific device or the gateway (or both?) in this case? Perhaps application wants to decide on itself if it is feasible to try to send commands to all devices connected to the gateway as application knows capabilities of the gateway (e.g. if gateway itself buffers commands for devices).

sophokles73:
> I was just thinking about what should happen with devices connected through gateway mode. Should hono send notification for the specific device or the gateway (or both?)

IMHO Hono should publish two events in this case, one for the gateway and one for the device.

sophokles73:
> IMHO Hono should publish two events in this case, one for the gateway and one for the device.

This is actually not straight forward because a gateway may (re-)use its connection to publish data for arbitrary devices, i.e. the moment the gateway connects, Hono can publish a *connect* event for the gateway, but it cannot anticipate (yet) for which device the gateway will publish data. The trigger for publishing an event for the device(s) that the gateway acts on behalf of will therefore need to be an actual message being published (be it telemetry or event) for the particular device. However, this immediately raises the question, if we publish a *connect* event every time a device publishes a message ...

site/content/api/Event-API.md
```diff
@@ -9,8 +9,13 @@ The *Event* API is used by *Devices* to send event messages downstream.
 
 The Event API is defined by means of AMQP 1.0 message exchanges, i.e. a client needs to connect to Hono using AMQP 1.0 in order to invoke operations of the API as described in the following sections. Throughout the remainder of this page we will simply use *AMQP* when referring to AMQP 1.0.
 
-The *Event* API is identical to the [*Telemetry* API]({{< relref "Telemetry-API.md" >}}) regarding the provided operations, the message format and the message flow. 
-However it provides a different quality of service for messages sent to the *Event* endpoint by routing them via a persistent message broker to guarantee *AT LEAST ONCE* delivery.
+The *Event* API is identical to the [*Telemetry* API]({{< relref "Telemetry-API.md" >}}) regarding the provided operations and the message flow.
+
+The message format for arbitrary events is identical to the [*Telemetry* API]({{< relref "Telemetry-API.md" >}}) as well, but there is a special message format for *lifecycle notifications* which is defined [here]({{< relref "#lifecycle-notifications" >}}).
```
sophokles73: IMHO there is no difference in the general message format between Telemetry and Event. However, we should add a section at the bottom where we list *well-known* message types (identified by their specific *content-type* value) and describe their semantics and intended use.

site/content/api/Event-API.md
```diff
@@ -9,8 +9,13 @@ The *Event* API is used by *Devices* to send event messages downstream.
...
+
+Events provide a different quality of service for messages sent to the *Event* endpoint by routing them via a persistent message broker
```
sophokles73: IMHO we should not talk about a *message broker* here simply because Hono does not include such a component. However, we should mention that all events have their message header *durable* set to `true`.

site/content/api/Event-API.md
```diff
@@ -9,8 +9,13 @@ The *Event* API is used by *Devices* to send event messages downstream.
...
 
-The *Event* API is identical to the [*Telemetry* API]({{< relref "Telemetry-API.md" >}}) regarding the provided operations, the message format and the message flow. 
-However it provides a different quality of service for messages sent to the *Event* endpoint by routing them via a persistent message broker to guarantee *AT LEAST ONCE* delivery.
+The *Event* API is identical to the [*Telemetry* API]({{< relref "Telemetry-API.md" >}}) regarding the provided operations and the message flow.
+
+The message format for arbitrary events is identical to the [*Telemetry* API]({{< relref "Telemetry-API.md" >}}) as well, but there is a special message format for *lifecycle notifications* which is defined [here]({{< relref "#lifecycle-notifications" >}}).
+
+Events provide a different quality of service for messages sent to the *Event* endpoint by routing them via a persistent message broker 
+to guarantee *AT LEAST ONCE* delivery. 
+*Lifecycle notifications* may be decided to not being persisted, but this is left to the setup of the *AMQP network* and is not further specified in the following.
```
sophokles73: This seems to be an issue worth discussing based on my comment above. We currently mark *all* events as *durable* btu I agree that we might want to relax this requirement and allow the issuer of the event to mark it as such or not.

site/content/api/Event-API.md
```diff
@@ -76,4 +81,56 @@ The following sequence diagram illustrates the flow of messages involved in a *B
 
 **Message Format**
 
+**Arbitrary events from the device**
+
 See [*Telemetry API*]({{< relref "Telemetry-API.md" >}}) for definition of message format. 
+
+### Lifecycle Notifications
+
+To enable *Business Applications* to react to the online status of devices belonging to a specific tenant, a special type of event - named *lifecycle notification* - is predefined.
```
sophokles73: I would rather not use the term *online* in this context, simply because there is no definition of what it means to be *online*. AFAIC we should refrain to talking about a device's connection status, i.e. we can detect when a device connects to one of the protocol adapters and we *may* be able to detect when it disconnects. I do not want to open up a wide field of speculation about what should and what shouldn't be able on a device that is supposedly *online*.

site/content/api/Event-API.md
```diff
@@ -76,4 +81,56 @@ The following sequence diagram illustrates the flow of messages involved in a *B
...
+| :------------- | :-------: | :----------------------- | :-------- | :---------- |
+| *content-type* | yes       | *properties*             | *symbol*  | Must be set to *application/vnd.eclipse-hono-notification+json* |
+
+The body of the event request MUST consist of a single *AMQP Value* section containing a UTF-8 encoded string representation of a single JSON object having the following members:
+
+| Name               | Mandatory | Type       | Description |
+| :----------------- | :-------: | :--------- | :---------- |
+| *tenant-id*        | *yes*     | *string*   | The tenant identifier to send the notification for. |
+| *device-id*        | *yes*     | *string*   | The device identifier to send the notification for. |
+| *cause*            | *yes*     | *string*   | Value must be either "connected" or "disconnected" . |
```
sophokles73: use `connected` and `disconnected` for explicit values

sysexcontrol:
>  The trigger for publishing an event for the device(s) that the gateway acts on behalf of will therefore need to be an actual message being published (be it telemetry or event) for the particular device. However, this immediately raises the question, if we publish a connect event every time a device publishes a message ...

On the one hand I cannot see any other way than publishing a "connected" event in the gateway case every time a device is publishing data.
But on the other hand this means: we publish as many events as we are receiving telemetry data (which we estimate being the absolute majority of data being sent), so **what is the extra value of such an event** then?

I like to extend the question to all "not connection oriented" protocols like HTTP (without any gateway involved): we plan so far to send a "connected" event as soon as data is published, but again : with which extra value?
- it does not mean a command can be sent down to the device after receiving such an event (could still fail)
- it is always published when upstream data was published, so as many events as upstream data (!)
- the application could just as well send its command as soon as it receives published data from the device, and not on receiving such an event?

Am I missing something, or is the whole "connected" event publishing possibly of no extra value?

The only sense I currently see in such events is if we find a better way than "always publish this event when a device is publishing upstream data". Or if some devices could publish such an event themselves, but this is purely speculative.

sophokles73:
> Am I missing something, or is the whole "connected" event publishing possibly of no extra value?
The only sense I currently see in such events is if we find a better way than "always publish this event when a device is publishing upstream data". Or if some devices could publish such an event themselves, but this is purely speculative.

This is indeed a very good question and seems to be an argument for allowing devices to indicate (e.g. by publishing a *connected* event or by other means) to indicate that they are now prepared to receive a command. For example, a device using HTTP could include a specific header in its request, a device connected via MQTT could publish a corresponding event (maybe even indicating the time period for which it will stay connected?) ...

sysexcontrol:
Regarding the frequency of notification events:
maybe it would be sufficient to define a time-interval, so that a protocol adapter *only* sends a `connected` event if it hasn't done so already inside this time-interval?

E.g. define the time-interval to 15 minutes. If the device publishes telemetry data every 5 seconds, the `connected` notification would be published only once per 180 telemetry messages. All narrowed to the instance of the protocol adapter.

sysexcontrol:
@sophokles73 : pushed new commit that reflects your comments to the Event page.

thjaeckle:
If the device initiates the `connected` event it would be very useful to also have an optional field `data` (e.g.) in the event's payload.
From an Eclipse Ditto perspective the device could insert its last known "revision" and/or the data it would like to get pushed when it changed meanwhile in the twin.

Ditto could then check if the twin knows of a newer revision and send back the updated desired state to the connected device.

But as I said, that probably only makes sense when the device sends the `connected`.

pellmann:
So let’s follow the idea, that typically the device is responsible to send a _connected_ event (and maybe the adapter could do such configurable in device registry for very constraint/limited devices). 

As discussed above for protocols like HTTP, this might need a _connected_ event together with each telemetry or event, the device sends to the adapter - a lot of messages will be sent.

So it could be a good idea to allow such a flag at the telemetry/event message itself. In the case of HTTP this could be a specific HTTP header. It also could include the time (or timeframe), the device stays connected, to be addressable by a cloud-to-device command. 

Then the next question would be, if the adapter should also include this information as an application property with the telemetry/event AMQP message or if it should send a separate event (again this could lead to always two messages for each one telemetry message from the device).

So for the AMQP side we could have just another application property “connected” with milliseconds (planned) as its data (Sigfox devices for example stay online for 30 seconds after a device initiated bidirectional communication).

If a device just wants to indicate the bidirectional communication without sending telemetry/event data, there can be a specific content-type for such events.

ctron:
As this is a task for the integration, I think it would be important to get some feedback from @stefanomorson and @mcarrer on the discussion.

ctron:
So for MQTT, if it is the responsibility of the device, then we could IMHO implement the disconnected event the same way. MQTT's way for this would be the last-will-and-testament (LWT) message which is set on the MQTT connect request. The protocol adapter could simply buffer this and release the message when the connection dies. As this functionality is a "best effort" from Hono we simply don't cover the situation of a killed pod. But when doing a graceful shutdown we could still support this.

sysexcontrol:
I would like to reanimate the discussion regarding the notifications here.

Based on what we seem to agree on, I like to propose the following (leaving intentionally out details of the different protocol adapters for now, I will come to that very soon as well):

**The standard responsibility to signal the applications that a device is ready to receive a command is at the device itself.**

Assuming that this is settled, IMHO we need two different mechanisms to support the variety of requirements to achieve that:

1.) "**piggy-back**" telemetry or event data:

this is the case when a device is ready for commands for a while after it has sent data downstream. 

For this purpose IMHO we should define specific AMQP application properties, as @pellmann  already suggested.

For now, let's not discuss 
- if this is triggered by specific protocol headers that the device specifies
- if in some situations a protocol adapter feeds these properties, so the device does not have to do it (comes later)
- which property or properties this would be (comes later)

 Important is that the **application** receives downstream data with the specific AMQP application properties and **can be sure that just then a receiver link for commands to the device is open** to send a command upstream to the protocol adapter.

2.) "**Specific event**":

Apart from piggy-backing downstream data, a device needs to be able to signal by itself that it is ready to receive commands.

For this purpose we have a well-known event, that is defined with a specific header (already specified in the last commit contained in this PR).

This can be sent without publishing other downstream data at any points in time, and will probably be sent only very rarely by a device.

Again, for now let's not discuss 
- if this is always triggered by the device directly
- if in some situations a protocol adapter sends this event, so the device does not have to do it (comes later)
----------------------------------------

If we agree on these two points, I would define the details and some sequence diagrams. And then come to the details what variants we have in the different protocol adapters (intentionally left out here for now).

@ctron  @pellmann  @sophokles73 : WDYT?

sysexcontrol:
Renamed this PR since we do not define an own API (as originally planned) anymore.

ctron:
I was just re-reading through this PR. Would a better title be something like "device ready to receive".

sysexcontrol:
@ctron: just renamed the more detailed specification PR #562 to `device notifications` - so no `lifecycle` anymore, which was not matching well.
My thoughts so far about the naming:
- can be sent directly from device -> matching with the name
- can be sent from protocol adapters for pure connection/disconnection event, when configured -> still matching, since it relates to a device

I am for now satisfied with this naming, but maybe we find a better one? :-)
If you agree, I would rename this PR here as well...

sysexcontrol:
> Sorry to jump late to the party. 

@ppatierno : no matter when, highly appreciated :+1: !

First of all: the more elaborated specification is now in #562, where the discussion about the details went further.
But you ask a good question: 
- the `source` property is made mandatory in the other PR, but
- is it really important?

I do not have a 100% elaborated answer - but we agreed on having two different possibilities which part of the system is responsible to signal `command readiness` for a device: the `device`, or a `protocol adapter` (taking over this task for a device by sending such an event when the device connects).

With that background, I feel it makes sense to know which part issued the event, but I am not 100% sure.

One additional thing is the timestamp - in current prototype the protocol adapter sets it in the message, but (if possible) I would quite like to later enable the device itself to set it in the payload. This would eliminate any latency between the device and the protocol adapter - if for some reasons the event sent by the device takes quite some time to reach the protocol adapter, the `ttl` would be not very valuable anymore (since the timestamp to add it to is not the one when the device sent the command). Demanding that the device sets an `validUntil` timestamp might be too demanding for constrained devices, since they might just know their time interval for how long they are ready for commands now (?).

Want to say that the payload may differ in completeness, depending on the source of the event, so a `source` would be helpful when evaluating the event.

Does that make sense? Comments welcome! Best in PR #562.
We can discuss this tomorrow in our meeting as well.

sysexcontrol:
Since we now have several more specific PR's for the two events we agreed on, I would like to close this PR here (to consolidate the number of PR's we have currently open).
If no one has concerns until tonight, I will tomorrow morning close this PR - ok?

sysexcontrol:
Closing this PR - we continue with the more detailed PRs now:
PR #576
PR #584
PR #569

PR #538, opened 2018-03-19
http-adapter: Implement OPTIONS method and add CORS handler

pho-enix:
This makes the API consumable by web applications (swagger-ui in my case).
The OPTIONS method is used by browsers to do a pre-flight validation,
and must be accessible without authentication. The web application will
then check for the precense and correctness of CORS related headers.

### DIFF
--- a/adapters/http-vertx/src/main/java/org/eclipse/hono/adapter/http/vertx/VertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx/src/main/java/org/eclipse/hono/adapter/http/vertx/VertxBasedHttpProtocolAdapter.java
@@ -14,7 +14,7 @@ package org.eclipse.hono.adapter.http.vertx;
 
 import java.net.HttpURLConnection;
 
-import io.netty.handler.codec.http.HttpResponseStatus;
+import io.vertx.core.http.HttpHeaders;
 import io.vertx.ext.web.handler.CorsHandler;
 import org.eclipse.hono.adapter.http.AbstractVertxBasedHttpProtocolAdapter;
 import org.eclipse.hono.adapter.http.HonoAuthHandlerImpl;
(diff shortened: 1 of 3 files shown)


adapters/http-vertx/src/main/java/org/eclipse/hono/adapter/http/vertx/VertxBasedHttpProtocolAdapter.java
```diff
@@ -57,10 +59,20 @@ protected final void addRoutes(final Router router) {
...
         addTelemetryApiRoutes(router);
         addEventApiRoutes(router);
+        setupCorsHandler(router);
+
+        router.options().handler(request -> request.response().setStatusCode(HttpResponseStatus.OK.code()).end());
+    }
+
+    private void setupCorsHandler(final Router router) {
+        router.route()
+                .handler(CorsHandler.create("*").allowedMethod(HttpMethod.PUT)
```
ctron: I think there should be an option to limit the wildcard by using some configuration option. Not everyone might want the wildcard to be hardcoded.

pho-enix: Good point. I will re-cook the patch

pho-enix: I have put the property under "hono.http.corsAllowedOrigin"

adapters/http-vertx/src/main/java/org/eclipse/hono/adapter/http/vertx/VertxBasedHttpProtocolAdapter.java
```diff
@@ -49,16 +51,26 @@ protected String getTypeName() {
...
+
         addTelemetryApiRoutes(router);
         addEventApiRoutes(router);
     }
 
+    private void setupCorsHandler(final Router router) {
+        router.route()
+                .handler(CorsHandler.create(getConfig().getCorsAllowedOrigin()).allowedMethod(HttpMethod.PUT)
+                        .allowedMethod(HttpMethod.POST).allowedMethod(HttpMethod.OPTIONS)
+                        .allowedHeader("Authorization").allowedHeader("Content-Type"));
```
sophokles73: Please use the corresponding constants for specifying the allowed headers:
`io.vertx.core.http.HttpHeaders.AUTHORIZATION` and `io.vertx.core.http.HttpHeaders.CONTENT_TYPE`.
IMHO we do not need to explicitly allow the `Options` method since this is already implicitly allowed (and implemented) by the CorsHandler, isn't it?

Issue #478, opened 2018-02-06
Remove Hono Client dependencies on Spring Framework


avgustinmm:
I order to minimize Hono client (hono-core & hono-client) dependencies on other libraries (in order to be easier for users to integrate it into their frameworks) it will be good to remove dependency on Spring Framework and stick to that in the future (if there is no real reason to add such dependency). Motivation:
1. It is always good to have as little dependencies as possible
2. Spring framework is not really needed. The only uses are:
  -  SpringUtils.isEmpty() which could be easily replaced with something like:
`public boolean isEmpty(String str) {`
` return str == null || str.length() == 0; // or str == null || str.isEmpty()`
` }`
 Note that SpringUtils.isEmpty() checks objects (not strings only) but Hono client uses it only for strings (as far as I understand).
  - For Cache manager API which is used to hold named caches. This could be easily replaced (if I'm not missing something) with a lightweight abstraction like:
 ` interface  CacheManager {`
 `   Cache getCache(String name);`
 ` }`
 ` interface Cache<K, V> {`
 `    void put(K key, V value);`
 `    V get(K key);`
 `    void evict(K key); // or invalidate `
 ` }`
   This could be instantiated wrapping Spring cache manger or other cache implementation, e.g. Guava cache. If you do that you will support easier integration of other caches (not Spring only) and it will be a benefit for the users.

sophokles73:
I agree that we can define our own cache abstraction and get rid of the (compile time) dependency on Spring. Would you like to prepare a PR for the changes you suggest?

sophokles73:
Maybe we could simply move the `SpringBasedExpiringValueCache` to the `hono-service-base` module and replace the remaining `StringUtils.isEmpty()` method invocation.

The Spring based cache is mainly used in the Protocol Adapters anyway ...

ctron:
Giving this a first look, the cache is also present in `HonoClientImpl`, `RegistrationClientImpl` and `TenantClientImpl`. So I guess simply moving stuff away to `hono-service-base` isn't good enough.

My proposal would be to create a cache abstraction interface in `hono-core` and do a thin spring implementation layer in `hono-cache-spring`. Defaulting the client to not use caching.

I will prepare this in a feature branch and create a PR once it is complete.

sophokles73:
fixed in https://github.com/eclipse/hono/commit/b8583ba7bcce1d15391efc8a45171af03f5184ba

PR #539, opened 2018-03-19
[#478]: Remove Spring dependency from core and client

ctron:
This change removes the spring dependency from 'hono-core' and
'hono-client' by introducing a CacheProvider interface. This can be
implemented as consumers of Hono would like it. The default
implementation of using Spring Cache and Guava is still provided by the
default configuration. But it is now located in 'hono-service-base'.


core/src/main/java/org/eclipse/hono/cache/CacheProvider.java
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * SPDX-License-Identifier: EPL-1.0
+ */
+
+package org.eclipse.hono.cache;
+
+public interface CacheProvider {
+
+    public <K, V> ExpiringValueCache<K, V> getCache(String cacheName);
+}
```
sophokles73: Can you please add proper JavaDoc? I also think that adding the `FunctionalInterface` annotation would fit nicely ...

ctron: :+1: for the Javadoc. About the functional interface, I had that in the beginning, but removed it again because, I am not sure if this will stay the only method over time. We might additional methods over time. Or might not. But in the end there is currently only one implementor of this interface. So the benefit of being able to use lambdas would effect only one location.

sophokles73: Ok, I see.

ctron:
I did add the missing javadoc and improved the handling of `null`.

core/src/main/java/org/eclipse/hono/cache/CacheProvider.java
```diff
@@ -0,0 +1,32 @@
...
+
+package org.eclipse.hono.cache;
+
+/**
+ * A provider for cache instances.
+ */
+public interface CacheProvider {
+
+    /**
+     * Get a new instance of a cache by name. <br>
```
sophokles73: please use a `<p>` to separate the first sentence from the remaining description
we also use a verb in *active* form at the start of the first sentence:
*Gets a new instance of a cache by name.*

core/src/main/java/org/eclipse/hono/cache/CacheProvider.java
```diff
@@ -0,0 +1,32 @@
...
+     * It is up to the provider how the cache instance is created, or it creates a new or returns an existing one.
+     * 
+     * @param cacheName the name of the cache to get. Must not be {@code null}.
```
sophokles73: We indicate the requirement to provide a non-null value by means of adding a
`@throws NullPointerException if the cache name is {@code null}`

core/src/main/java/org/eclipse/hono/cache/package-info.java
```diff
@@ -0,0 +1,17 @@
+/**
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
```
sophokles73: we do not use package-info files, please remove it

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
```diff
@@ -158,14 +159,27 @@ protected void customizeRegistrationServiceClientConfig(final RequestResponseCli
...
+    /**
+     * Exposes the provider for caches as a Spring bean.
+     * 
+     * @return The provider instance.
+     */
+    @Bean
+    @Qualifier(RegistrationConstants.REGISTRATION_ENDPOINT)
+    @Scope("prototype")
+    public CacheProvider registrationCacheProvider() {
+        return newGuavaCache(tenantServiceClientConfig());
```
sophokles73: IMHO this needs to be `registrationServiceClientConfig()`

ctron:
Should be all fixed with the latest push.

PR #201, opened 2017-06-05
[#194] MQTT Protocol Adapter send PUBACK only when remote has accepted the message

francois-travais:
Work in progress...

PR #541, opened 2018-03-19
Update the notice file, 2018, more precise

ctron:
As this touches legal, I think an additional pair of eyes is required.


kartben:
I like the idea of listing all the copyright holders but I am not 100% sure the list is correct/complete?
How about François Travais, for example? https://github.com/eclipse/hono/pull/201/files

Issue #503, opened 2018-02-22
Extend MQTT protocol adapter to support shortened topic prefixes


dejanb:
As discussed in Eclipse Integration WG we want to shorten topic prefixes used by MQTT protocol adapter, in order to decrease message sizes.

`telemetry/` -> `t/`
`event/` -> `e/`

ppatierno:
Interesting but ... are we really saving space decreasing by 5 to 10 bytes ? Considering that most of the communication should be encrypted (so on TLS) between devices and adapter I don't see a really advantage on doing that. The TLS overhead increase the message size itself.

The great advantage comes with MQTT v5 where after publishing on a topic, the next PUBLISH can have a zero length topic and the broker will publish on the same topic as before. Btw we don't have MQTT v5 ... yet

sophokles73:
> The great advantage comes with MQTT v5 where after publishing on a topic, the next PUBLISH can have a zero length topic and the broker will publish on the same topic as before

Agreed. In the meantime, the Kapua folks considered it worthwhile to shorten the prefixes. FMPOV the MQTT adapter should simply support both the original prefixes as well as the shortened ones so the device can use whatever it feels comfortable with.

ppatierno:
Agree. If it comes from a real needed for Kapua + Kura (maybe ?) then it's fine.

PR #543, opened 2018-03-20
[#503] Support short topic names in MQTT protocol adapter

dejanb:


adapters/mqtt-vertx-base/src/test/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapterTest.java
```diff
@@ -490,6 +490,44 @@ public void testOnUnauthenticatedMessageDoesNotSendPubAckOnFailure(final TestCon
         verify(endpoint, never()).publishAcknowledge(anyInt());
     }
 
+    @Test
+    public void testTopicNames(final TestContext ctx) {
```
sophokles73: what is this test supposed to verify? can you please add a JavaDoc comment?

adapters/mqtt-vertx/src/test/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapterTest.java
```diff
@@ -204,6 +204,36 @@ public void testOnPublishedMessageUsesDeviceIdentityForTopicWithoutTenant(final
         assertThat(downstreamAddress.getTenantId(), is("my-tenant"));
     }
 
+    /**
+     * Verifies that the adapter supports all required topic names
+     *
+     * @param ctx The helper to use for running tests on vert.x.
+     */
+    @Test
+    public void testTopicNames(final TestContext ctx) {
```
sophokles73: this seems to be a test case for verifying behavior of the *mapTopic* method.
IMHO you should add assertions as to whether the method has actually mapped the message topic to the correct address.

So far we try to reflect the method under test in the test method's name. Maybe something like `testMapTopicSupportsShortAndLongTopicNames`?

core/src/main/java/org/eclipse/hono/util/EndpointType.java
```diff
@@ -0,0 +1,31 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Red Hat Inc and others.
```
sophokles73: I don't know about you, but I already live in 2018 ;-)

dejanb:
@sophokles73 These should be fixed now. Thanks for the feedback

PR #545, opened 2018-03-21
Document the multi-tenancy concept of Hono.

sysexcontrol:
We have so far an empty concept page regarding multi tenancy in Hono.
This PR fills the gap and documents a lot of aspects of the multi tenancy concept.

Please have a detailed look if I forgot something, if something is unclear or even if there are any wrong statements inside.


site/content/concepts/tenancy.md
```diff
@@ -3,10 +3,134 @@ title = "Multi Tenancy"
 weight = 405
 +++
 
-This page describes how multi tenancy is handled in the Eclipse Hono project.
+Hono has a multi tenancy concept based on the handling of tenants as own *entities*. All functionality of Hono is
```
sophokles73: It would be great if you would first explain what a *tenant* actually is supposed to be and what it is used for.

sysexcontrol:
Added an explanation of what a tenant is to the beginning (not perfectly simple I find, but this should clarify IMHO).

sysexcontrol:
@sophokles73 : did you have another look?
Should we merge it and thus publish it to our webpage?

site/content/concepts/tenancy.md
```diff
@@ -1,12 +1,25 @@
...
-Hono has a multi tenancy concept based on the handling of tenants as own *entities*. All functionality of Hono is 
+Hono is designed to structure the set of all internally managed data and data streams into strictly isolated subsets. 
+This includes the registration data and credentials of devices, internal users that are used for authentication, 
+and the *Business Applications* that are part of such subsets as well.
+
+This way of *strict* isolation is generally known as multitenancy, where a **tenant** is the term for such a subset.
+Such an isolation is essential for enabling a scalable distributed architecture to handle independent subsets as if each subset had its
+own installation (which would be much harder to maintain and would not benefit from runtime cost sharing).
+
+This is why Hono implements a multitenancy concept.
```
sophokles73: this isolated sentence sounds a little strange when reading through the section. I would propose to simply omit it. You already gave the reasons why we do this in the previous paragraph ...

site/content/concepts/tenancy.md
```diff
@@ -1,12 +1,149 @@
...
+Please refer to the [Tenant API]({{< relref "Tenant-API.md" >}}) to find out which protocol adapter properties are
+configurable on a tenant level. The number of such properties may grow during further development of Hono.
+
+## AMQP 1.0 endpoints
+
+The AMQP 1.0 endpoints for all APIs of Hono are scoped to a tenant, by using the scheme `<api>/TENANT/...`.
+
+Examples are:
+
+- `telemetry/TENANT/device`
```
sophokles73: the APIs are usually not scoped to a particular device, IMHO we should therefore remove the trailing `/device` segments ...

site/content/concepts/tenancy.md
```diff
@@ -1,12 +1,149 @@
...
+The only exception to this is the [Tenant API]({{< relref "Tenant-API.md" >}}), which does not follow this scheme since it
+is addressing the tenants themselves.   
+
+## Devices and tenants
+
+All devices of Hono belong to exactly one tenant, so any data is sent in the context of the
+tenant a device belongs to.
+
+It is important here to focus on Hono's view of a device: it is on a logical level, meaning a device is created in the 
+context of a tenant purely by using Hono's Credentials and Registration APIs.
```
sophokles73: `Device Registration API`

site/content/concepts/tenancy.md
```diff
@@ -1,12 +1,149 @@
...
+It is important here to focus on Hono's view of a device: it is on a logical level, meaning a device is created in the 
+context of a tenant purely by using Hono's Credentials and Registration APIs. 
+
+Often a physical device will be created in Hono as exactly one logical device, and thus belongs to exactly one tenant.
+
+But if necessary, this can be extended:  
+
+**Advanced setup for multiple tenants**
+
+Due to the logical view Hono has to devices, a physical device can have several *logical identities* in Hono: if using
```
sophokles73: Is this really a supported use-case? It sounds as if it would be possible but I haven't heard about anybody doing this, so far. Maybe we should keep this section out for the moment? My concern is that it (currently) would confuse people even more while they are still trying to understand the basic tenant concept.

site/content/concepts/tenancy.md
```diff
@@ -1,12 +1,149 @@
...
+different credentials when connecting to a protocol adapter, the physical device can *decide* which identity to use
+for this connection, and thus can even connect in the context of different tenants.
+
+
+## Upstream data flow for tenants
+
+An important detail in Hono's architecture is that data sent upstream is transported via the tenant
```
sophokles73: we actually call this direction *downstream*: from the spring (the device) to the ocean (the applications)

site/content/concepts/tenancy.md
```diff
@@ -1,12 +1,149 @@
...
+scoped AMQP 1.0 links from the protocol adapters to the AMQP 1.0 network.
+
+Each tenant has its own pair of AMQP 1.0 links and is treated 
+independently from other tenants regarding the back pressure mechanism that AMQP 1.0 offers (using different credits).
```
sophokles73: I would omit the `(using different credits)`. This seems to be  a technical detail that is of not much importance here, is it?

site/content/concepts/tenancy.md
```diff
@@ -1,12 +1,149 @@
...
+
+This addresses the noisy neighbour problem that needs to be solved in a multitenant setup.
```
sophokles73: if the neighbor is really noisy (i.e. the consumers issue a lot of credit) then this doesn't really help unless each tenant uses its own AMQP session, does it?

site/content/concepts/tenancy.md
```diff
@@ -1,12 +1,149 @@
...
+etc.
+
+This separates the AMQP endpoints from each other on a tenant level.
+
+The only exception to this is the [Tenant API]({{< relref "Tenant-API.md" >}}), which does not follow this scheme since it
+is addressing the tenants themselves.   
+
+## Devices and tenants
+
+All devices of Hono belong to exactly one tenant, so any data is sent in the context of the
```
sophokles73: what about adding a little class diagram illustrating the relationships between *Tenant*, *Device* and *Credentials*?

site/content/concepts/tenancy.md
```diff
@@ -1,12 +1,149 @@
...
+- the credentials of devices are strictly scoped to a tenant
+- protocol adapters can be enabled/disabled for a tenant 
+- the upstream data flow is isolated for every tenant
+- *Business applications* need to authenticate to the AMQP 1.0 network and are by that mechanism scoped to their tenant
+ 
+## Notes on a production ready setup
+
+To be flexible for the different needs of production setups, Hono tries to make as few assumptions about the combination
+of the different APIs as possible.
+This means e.g. that there is no connection between the example implementations of the APIs:
```
sophokles73: Maybe we can replace *connection* with *referential integrity being enforced*?

I also think that we should more explicitly point out that the *Device Registry* component does not do this, but that a production system would (should) certainly enforce this. The term *example implementations* is a little abstract, when in fact we talk about Hono's *Device Registry* component, don't you think?

site/content/concepts/tenancy.md
sysexcontrol: Damn - you are right of course :-)! Will change it.

site/content/concepts/tenancy.md
sysexcontrol: IMHO you raise a good point here: 
1.) I agree, the credit mechanism via the links does not solve the noisy neighbour problem, it just adds some control. If the business application issues a massive amount of credits, other tenants may run into availability problems :-(
2.) own AMQP sessions improve that probably, but then we would need to introduce AMQP session based credits ("throttling"?), which we currently do not support, right?

So, I will rephrase this currently wrong sentence.

sophokles73: ad 2) an AMQP session can be limited in resource usage by means of the *session window* size which is negotiated during session establishment. using the session window you can limit the memory used for *in-flight* frames, i.e. frames that are part of unsettled message exchanges.

site/content/concepts/tenancy.md
sophokles73: makes sense, but still I think that we should postpone this *tip* until later (and maybe write a blog post about how you can cleverly partition your device set.

sysexcontrol:
@sophokles73 : updated the concept page regarding your comments.

site/content/concepts/Tenants_Device_Credentials.svg
```diff
@@ -0,0 +1 @@
...482q0 -2.359375 0.484375 -3.796875q0.484375 -1.453125 1.4375 -2.234375q0.96875 -0.78125 2.421875 -0.78125q1.078125 0 1.890625 0.4375q0.8125 0.421875 1.328125 1.25q0.53125 0.8125 0.828125 1.984375q0.3125 1.15625 0.3125 3.140625q0 2.359375 -0.484375 3.8125q-0.484375 1.4375 -1.453125 2.234375q-0.953125 0.78125 -2.421875 0.78125q-1.921875 0 -3.03125 -1.390625q-1.3125 -1.671875 -1.3125 -5.4375zm1.671875 0q0 3.296875 0.765625 4.390625q0.78125 1.078125 1.90625 1.078125q1.140625 0 1.90625 -1.09375q0.765625 -1.09375 0.765625 -4.375q0 -3.296875 -0.765625 -4.375q-0.765625 -1.078125 -1.921875 -1.078125q-1.125 0 -1.796875 0.953125q-0.859375 1.21875 -0.859375 4.5zm14.793396 6.59375l0 -1.875l1.875 0l0 1.875l-1.875 0zm10.366577 0l0 -1.875l1.875 0l0 1.875l-1.875 0zm9.913513 0l0 -9.671875l1.46875 0l0 1.375q1.0625 -1.59375 3.078125 -1.59375q0.875 0 1.609375 0.3125q0.734375 0.3125 1.09375 0.828125q0.375 0.5 0.515625 1.203125q0.09375 0.453125 0.09375 1.59375l0 5.953125l-1.640625 0l0 -5.890625q0 -1.0 -0.203125 -1.484375q-0.1875 -0.5 -0.671875 -0.796875q-0.484375 -0.296875 -1.140625 -0.296875q-1.046875 0 -1.8125 0.671875q-0.75 0.65625 -0.75 2.515625l0 5.28125l-1.640625 0z" fill-rule="nonzero"/></g></svg>
```
sophokles73: IMHO we should not use *generated* SVGs but instead use something like LibreOffice Draw to create the diagram and include it as a PNG in the document. This way, we also have the *source code* available and can adapt the diagram later.

site/content/concepts/tenancy.md
```diff
@@ -75,28 +76,25 @@ All devices of Hono belong to exactly one tenant, so any data is sent in the con
...
-An important detail in Hono's architecture is that data sent upstream is transported via the tenant
+An important detail in Hono's architecture is that data sent downstream is transported via the tenant
 scoped AMQP 1.0 links from the protocol adapters to the AMQP 1.0 network.
 
 Each tenant has its own pair of AMQP 1.0 links and is treated 
-independently from other tenants regarding the back pressure mechanism that AMQP 1.0 offers (using different credits).
+independently from other tenants regarding the back pressure mechanism that AMQP 1.0 offers.
 
-This addresses the noisy neighbour problem that needs to be solved in a multitenant setup.  
+This enables a *Business application* to control the number of AMQP 1.0 messages it can buffer for the tenant it belongs
```
sophokles73: I'd rather write something like
> This enables a *Business application* to limit the rate at which it consumes AMQP 1.0 messages *per tenant*.

site/content/concepts/tenancy.md
```diff
@@ -132,18 +128,18 @@ Here is a summary of how this is implemented:
...
-This means e.g. that there is no connection between the example implementations of the APIs:
+This means e.g. that the Device Registry does not enforce referential integrity of the APIs:
 
 - devices can be created for a tenant that is not existing (yet)
 - credentials can be created for a tenant and/or a device that is not existing (yet)
 - tenants can be deleted and leave their scoped devices and credentials still in the configuration (which may not be usable
   anymore, since the tenant is missing)
 
-These are points that production setups may want to implement differently.
+These points production setups may want to implement differently.
```
sophokles73: You seem to have garbled up the words of this sentence ;-)

site/content/concepts/Tenants_Device_Credentials.svg
sysexcontrol: Something I did not like too much either (not having the source code of the pictures), so I make a start and redo the simple diagram with LibreOffice Draw :+1:

site/content/concepts/tenancy.md
sysexcontrol: If that did not make the sentence more understandable, I just roll it back ;-)

PR #546, opened 2018-03-21
extending hono example receiver with proxy options

balsmn:
Added the following arguments that can be used to connect the example receiver from behind a proxy

```
hono.client.proxyHost
hono.client.proxyPort
hono.client.proxyUsername
hono.client.proxyPassword
```

### DIFF
--- a/core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
+++ b/core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
@@ -16,6 +16,9 @@ import java.util.Objects;
 
 import org.eclipse.hono.util.Constants;
 
+import io.vertx.core.net.ProxyOptions;
+import io.vertx.core.net.ProxyType;
+
 /**
  * Common configuration properties required for accessing an AMQP 1.0 container.
  */
@@ -34,8 +37,21 @@ public class ClientConfigProperties extends AbstractConfig {
      */
     public static final long DEFAULT_REQUEST_TIMEOUT = 200L; // ms
 
+    /**
+     * The default proxy server type
+     */
+    public static final ProxyType DEFAULT_PROXY_TYPE = ProxyType.SOCKS5;
+
+    private static final int MIN_PORT_NO = 1024;
+    private static final int MAX_PORT_NO = 65535;
+
     private String name;
     private String host = "localhost";
+    private String proxyHost;
+    private int proxyPort;
+    private String proxyUsername;
+    private String proxyPassword;
+    private ProxyType proxyType = DEFAULT_PROXY_TYPE;
     private int port = Constants.PORT_AMQPS;
     private String username;
     private char[] password;
@@ -84,7 +100,7 @@ public class ClientConfigProperties extends AbstractConfig {
      * The default value of this property is {@link Constants#PORT_AMQPS}.
      * 
      * @param port The port number.
-     * @throws IllegalArgumentException if port &lt; 1000 or port &gt; 65535.
+     * @throws IllegalArgumentException if port &lt; 0 or port &gt; 65535.
      */
     public final void setPort(final int port) {
         if (isValidPort(port)) {
@@ -95,8 +111,122 @@ public class ClientConfigProperties extends AbstractConfig {
     }
 
     /**
-     * Gets the user name that is used when authenticating to the Hono server.
+     * Gets the name or literal IP address of the proxy host that the client is configured with.
+     *
+     * @return The host name.
+     */
+    public final String getProxyHost() {
+        return proxyHost;
+    }
+
+    /**
+     * Sets the name or literal IP address of the proxy host that the client should use to connect to the an AMQP host
+     *
+     * @param proxyHost The host name or IP address of the proxy
+     * @throws NullPointerException if proxyHost is {@code null}.
+     */
+    public final void setProxyHost(final String proxyHost) {
+        this.proxyHost = Objects.requireNonNull(proxyHost);
+    }
+
+    /**
+     * Gets the TCP port of the proxy server that this client is configured to use.
+     *
+     * @return The port number.
+     */
+    public final int getProxyPort() {
+        return proxyPort;
+    }
+
+    /**
+     * Sets the TCP port of the proxy server that this client should use.
+     *
+     * @param proxyPort The port number of the proxy
+     * @throws IllegalArgumentException if port &lt; 1024 or port &gt; 65535.
+     */
+    public final void setProxyPort(final int proxyPort) {
+        if (isInValidPortRange(proxyPort)) {
+            this.proxyPort = proxyPort;
+        } else {
+            throw new IllegalArgumentException("invalid proxy port number");
+        }
+    }
+
+    /**
+     * Gets the username of the proxy server the client is configured with.
+     *
+     * @return the username
+     */
+    public String getProxyUsername() {
+        return proxyUsername;
+    }
+
+    /**
+     * Sets the username to be used for authentication with the proxy server.
+     *
+     * @param proxyUsername username to be used for authentication with the proxy server
+     * @throws NullPointerException if proxyUsername is {@code null}.
+     */
+    public void setProxyUsername(String proxyUsername) {
+        this.proxyUsername = Objects.requireNonNull(proxyUsername);
+    }
+
+    /**
+     * Gets the password of the proxy user to be used in authentication.
+     *
+     * @return password in clear text.
+     */
+    public String getProxyPassword() {
+        return proxyPassword;
+    }
+
+    /**
+     * Sets the password of the proxy user to be used in authentication.
+     *
+     * @param proxyPassword password in clear text
+     * @throws NullPointerException if proxyPassword is {@code null}.
+     */
+    public void setProxyPassword(String proxyPassword) {
+        this.proxyPassword = Objects.requireNonNull(proxyPassword);
+    }
+
+    /**
+     * Gets the proxy type to be used.
      * 
+     * <p>
+     * Default is {@link ProxyType#SOCKS5}.
+     * </p>
+     *
+     * @return the proxy type if set by the user or the default value.
+     */
+    public ProxyType getProxyType() {
+        return proxyType;
+    }
+
+    /**
+     * Sets the proxy type to be used.
+     *
+     * <p>
+     * Valid values are
+     * <ul>
+     *  <li>SOCKS4</li>
+     *  <li>SOCKS5</li>
+     *  <li>HTTP</li>
+     * </ul>
+     * Default value is set to SOCKS5
+     * </p>
+     * 
+     * @param proxyType the type of proxy to be set.
+     * @throws IllegalArgumentException if the given proxyType is not a valid type.
+     * @throws NullPointerException if proxyType is {@code null}.
+     */
+    public void setProxyType(String proxyType) {
+        this.proxyType = ProxyType.valueOf(Objects.requireNonNull(proxyType));
+    }
+
+    /**
+     * Gets the user name that is used when authenticating to the Hono server.
+     *
      * @return The user name or {@code null} if not set.
      */
     public final String getUsername() {
@@ -107,7 +237,7 @@ public class ClientConfigProperties extends AbstractConfig {
      * Sets the user name to use when authenticating to the Hono server.
      * <p>
      * If not set then this client will not try to authenticate to the server.
-     * 
+     *
      * @param username The user name.
      */
     public final void setUsername(final String username) {
@@ -334,5 +464,40 @@ public class ClientConfigProperties extends AbstractConfig {
      */
     public final void setTlsEnabled(boolean enabled) {
         this.tlsEnabled = enabled;
+    }
+
+    /**
+     * Returns an instance of ProxyOptions configured with the proxyHost, proxyPort, proxyUsername and
+     * proxyPassword.
+     * <p>
+     * If proxyHost and proxyPort are not set, then {@code null} is returned
+     * </p>
+     * 
+     * @return an instance ofProxyOptions if proxy details are provided, {@code null} otherwise.
+     */
+    public ProxyOptions getProxyOptions() {
+        if (!isEmpty(proxyHost) &&
+                isInValidPortRange(proxyPort)) {
+            ProxyOptions proxyOptions = new ProxyOptions().setHost(proxyHost)
+                    .setPort(proxyPort)
+                    .setType(getProxyType());
+
+
+            if (!isEmpty(proxyUsername) &&
+                    !isEmpty(proxyPassword)) {
+                proxyOptions.setUsername(proxyUsername);
+                proxyOptions.setPassword(proxyPassword);
+            }
+            return proxyOptions;
+        }
+        return null;
+    }
+
+    private boolean isInValidPortRange(final int proxyPort) {
+        return proxyPort >= MIN_PORT_NO && proxyPort <= MAX_PORT_NO;
+    }
+
+    private boolean isEmpty(String str) {
+        return str == null || "".equals(str);
     }
 }
--- a/example/src/main/java/org/eclipse/hono/example/AbstractExampleClient.java
+++ b/example/src/main/java/org/eclipse/hono/example/AbstractExampleClient.java
@@ -17,6 +17,7 @@ import java.util.List;
 import java.util.Objects;
 
 import org.eclipse.hono.client.HonoClient;
+import org.eclipse.hono.config.ClientConfigProperties;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -41,6 +42,7 @@ abstract class AbstractExampleClient {
     protected String tenantId;
     protected Vertx vertx;
     protected HonoClient client;
+    protected ClientConfigProperties clientConfigProperties;
     protected List<String> activeProfiles;
 
     /**
@@ -57,7 +59,7 @@ abstract class AbstractExampleClient {
     @Autowired
     public final void setVertx(final Vertx vertx) {
         this.vertx = vertx;
-        this.ctx = vertx.getOrCreateContext();
+        ctx = vertx.getOrCreateContext();
     }
 
     @Autowired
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -94,6 +98,86 @@ public final void setPort(final int port) {
...
+        this.proxyPassword = proxyPassword;
+    }
+
+    /**
+     * Sets the TCP port of the proxy server that this client should use.
+     *
+     * @param proxyPort The port number of the proxy
+     * @throws IllegalArgumentException if port is 0.
+     */
+    public final void setProxyPort(final int proxyPort) {
```
sophokles73: can you please move this up to the corresponding getter?

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -94,6 +98,86 @@ public final void setPort(final int port) {
...
+     */
+    public String getProxyPassword() {
+        return proxyPassword;
+    }
+
+    /**
+     * Sets the password of the proxy user to be used in authentication.
+     *
+     * @param proxyPassword password in clear text
+     * @throws NullPointerException if proxyUsername is {@code null}.
```
sophokles73: wrong param name, missing `null` check

example/src/main/java/org/eclipse/hono/example/AbstractExampleClient.java
```diff
@@ -57,15 +61,34 @@ public final void setActiveProfiles(final Environment env) {
...
+    @Autowired
+    public final void setClientConfigProperties(final ClientConfigProperties config) {
+        this.clientConfigProperties = Objects.requireNonNull(config);
+    }
+
     protected final ProtonClientOptions getClientOptions() {
-        return new ProtonClientOptions().setConnectTimeout(DEFAULT_CONNECT_TIMEOUT_MILLIS).setReconnectAttempts(2);
+        ProtonClientOptions protonClientOptions = new ProtonClientOptions()
+                .setConnectTimeout(DEFAULT_CONNECT_TIMEOUT_MILLIS).setReconnectAttempts(2);
+        if (!StringUtils.isEmpty(clientConfigProperties.getProxyHost()) &&
```
sophokles73: all of the following code should be moved to `ClientConfigProperties`, e.g. to a `getProxyOptions` method, so that it can be used by other components as well

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -94,6 +98,86 @@ public final void setPort(final int port) {
         }
     }
 
+    /**
+     * Gets the name or literal IP address of the host that the client is configured to connect to.
```
sophokles73: this is not correct

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -94,6 +98,86 @@ public final void setPort(final int port) {
...
+     * Gets the username of the proxy server the client is configured with.
+     *
+     * @return the username
+     */
+    public String getProxyUsername() {
+        return proxyUsername;
+    }
+
+    /**
+     * sets the username to be used for authentication with the proxy server.
```
sophokles73: Upper case ...

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
balsmn: done.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -94,6 +101,86 @@ public final void setPort(final int port) {
...
+     * Sets the TCP port of the proxy server that this client should use.
+     *
+     * @param proxyPort The port number of the proxy
+     * @throws IllegalArgumentException if port is 0.
+     */
+    public final void setProxyPort(final int proxyPort) {
+        if (proxyPort != 0) {
+            this.proxyPort = proxyPort;
+        } else {
+            throw new IllegalArgumentException("invalid proxy port number");
```
sysexcontrol: "invalid proxy port number - must be greater than zero" (or similar...)

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -94,6 +101,86 @@ public final void setPort(final int port) {
...
+    }
+
+    /**
+     * Sets the TCP port of the proxy server that this client should use.
+     *
+     * @param proxyPort The port number of the proxy
+     * @throws IllegalArgumentException if port is 0.
+     */
+    public final void setProxyPort(final int proxyPort) {
+        if (proxyPort != 0) {
```
sysexcontrol: wouldn't `proxyPort > 0` be an even better check?

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -335,4 +422,29 @@ public final boolean isTlsEnabled() {
...
+     * proxyPassword.
+     * <p>
+     * If proxyHost and proxyPort are not set, then {@code null} is returned
+     * </p>
+     * 
+     * @return either an instance of {@link ProxyOptions} if proxy details are provided, {@code null} otherwise.
+     */
+    public ProxyOptions getProxyOptions() {
+        if (!StringUtils.isEmpty(proxyHost) &&
+                proxyPort != 0) {
```
sysexcontrol: `proxyPort > 0` ?

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
sysexcontrol: Should check for max value probably as well (but vert.x might do this anyway). Have you spend any thoughts on that?

balsmn:
Can we merge this PR or is there any open issues?

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -94,6 +101,86 @@ public final void setPort(final int port) {
...
+    }
+
+    /**
+     * Sets the TCP port of the proxy server that this client should use.
+     *
+     * @param proxyPort The port number of the proxy
+     * @throws IllegalArgumentException if port &lt; 1000 or port &gt; 65535.
+     */
+    public final void setProxyPort(final int proxyPort) {
+        if (isValidPort(proxyPort)) {
```
sysexcontrol: :+1:  for using the helper class here.
But did you see that this results in "proxyPort == 0" being a valid value? Different what you had before.
And: not matching with the javadoc above (<= 1000 or > 65535).
Please adopt, thanks!

balsmn: I see now that in the helper class PortConfigurationHelper.MIN_PORT_NO is set to 0. This means, that setPort() method was also implemented wrongly and the javadoc is not correct. I would prefer to set  PortConfigurationHelper.MIN_PORT_NO to 1000 instead of 0, WDYT?

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -335,4 +422,29 @@ public final boolean isTlsEnabled() {
...
+     * <p>
+     * If proxyHost and proxyPort are not set, then {@code null} is returned
+     * </p>
+     * 
+     * @return either an instance of {@link ProxyOptions} if proxy details are provided, {@code null} otherwise.
+     */
+    public ProxyOptions getProxyOptions() {
+        if (!StringUtils.isEmpty(proxyHost) &&
+                isValidPort(proxyPort)) {
+            ProxyOptions proxyOptions = new ProxyOptions().setHost(proxyHost)
```
sophokles73: You seem to assume that the client wants to connect via an HTTP proxy, right?
This is probably not, what you usually do when connecting to a TCP socket other than an HTTP port.
FMPOV the configuration needs to take the proxy type (`ProxyOptions.setType()`) into account, using a default of `ProxyType.SOCKS5`).

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
sysexcontrol: 1.) why should a configuration like "HTTP adapter set to port 80" be always wrong?
2.) Why `1000` as a limit? I would rather set a limit to `1024`, since on most systems you need root access to open a port less than `1024`.

I would not adopt the helper class, and only add a check in your code to disallow the `0`, as you already had before.

balsmn: Ok, I will then add the same check also to setPort so that it is consistent.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
balsmn: Good point. I will add it.

balsmn:
As per javadoc of setPort() a port is valid it > 1000 or < 65535 but the implementation checked it for > 0 and < 65535. So I have made changes to setPort() to be consistent with the javadoc.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -94,9 +111,109 @@ public final void setPort(final int port) {
...
+     *
+     * @return the proxy type if set by the user or the default {@link ProxyType#SOCKS5}.
+     */
+    public ProxyType getProxyType() {
+        return proxyType != null? proxyType : DEFAULT_PROXY_TYPE;
+    }
+
+    /**
+     * Sets the proxy type to be used.
+     *
```
sophokles73: Please add an enumeration of valid values and mention the default type if not explicitly set.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -94,9 +111,109 @@ public final void setPort(final int port) {
...
+     * @param proxyType
```
sophokles73: incomplete tag

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -14,7 +14,11 @@
 
 import java.util.Objects;
 
+import io.vertx.core.net.ProxyType;
 import org.eclipse.hono.util.Constants;
+import org.springframework.util.StringUtils;
```
sophokles73: we are currently trying to get rid of the dependency on Spring altogether in `hono-core` (see #539).
please replace references to `StringUtils` with custom checks

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -335,4 +452,35 @@ public final boolean isTlsEnabled() {
...
     }
+
+    /**
+     * Returns an instance of {@link ProxyOptions} configured with the proxyHost, proxyPort, proxyUsername and
+     * proxyPassword.
+     * <p>
+     * If proxyHost and proxyPort are not set, then {@code null} is returned
+     * </p>
+     * 
+     * @return either an instance of {@link ProxyOptions} if proxy details are provided, {@code null} otherwise.
```
sophokles73: no need to mention nor link to the returned type. JavaDoc handles this automatically.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -335,4 +452,35 @@ public final boolean isTlsEnabled() {
     public final void setTlsEnabled(boolean enabled) {
         this.tlsEnabled = enabled;
     }
+
+    /**
+     * Returns an instance of {@link ProxyOptions} configured with the proxyHost, proxyPort, proxyUsername and
```
sophokles73: no need to mention nor link to the returned type. JavaDoc handles this automatically.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -94,9 +111,109 @@ public final void setPort(final int port) {
...
+     *
+     * @param proxyPassword password in clear text
+     * @throws NullPointerException if proxyPassword is {@code null}.
+     */
+    public void setProxyPassword(String proxyPassword) {
+        this.proxyPassword = Objects.requireNonNull(proxyPassword);
+    }
+
+    /**
+     * Gets the proxy type to be used. Default is {@link ProxyType#SOCKS5}.
```
sophokles73: please move the second sentence to a new HTML paragraph and remove the (duplicate) mentioning of the default value from the `@return` tag.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -94,9 +111,109 @@ public final void setPort(final int port) {
...
+     *
+     * @return the proxy type if set by the user or the default {@link ProxyType#SOCKS5}.
+     */
+    public ProxyType getProxyType() {
+        return proxyType != null? proxyType : DEFAULT_PROXY_TYPE;
```
sophokles73: we usually set the default value on property declaration, i.e. at the top of the class
`ProxyType proxyType = DEFAULT_PROXY_TYPE;`

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -94,9 +111,109 @@ public final void setPort(final int port) {
...
+    }
+
+    /**
+     * Sets the proxy type to be used.
+     *
+     * @param proxyType
+     * @throws IllegalArgumentException if the given proxyType is not a valid {@link ProxyType}
```
sophokles73: refer to the list of supported values instead of linking to `ProxyType` which is not a parameter to this method.

site/content/getting-started.md
```diff
@@ -96,6 +96,14 @@ mvn spring-boot:run -Drun.arguments=--hono.client.host=localhost,--hono.client.u
 Replace *localhost* with the name or IP address of the host that Docker is running on.
 {{% /warning %}}
 
+If the command line client is behind a proxy, you can use the proxy options to connect. The default proxy type is SOCKS5. Run the client from the `example` folder as follows:
+                                                                                                                
+~~~sh
+mvn spring-boot:run -Drun.arguments=--hono.client.host=localhost,--hono.client.username=consumer@HONO,--hono.client.password=verysecret,--hono.client.proxyHost=proxy.company.com,--hono.client.proxyPort=8080,--hono.client.proxyUsername=my-company-id,--hono.client.proxyPassword=verysecret,--hono.client.proxyType=SOCKS4
```
sophokles73: sure you want to promote `SOCKS4` instead of `SOCKS5`? Also, the port (8080) is quite unusual for a socks proxy and might confuse people. On the other hand you may want to give an example of how to (ab)use an HTTPS proxy for an AMQPS connection ;-)

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
balsmn: ok

site/content/getting-started.md
balsmn: As default was SOCKS5, I wanted to show how one can configure otherwise. On omitting this parameter, SOCKS5 is set by default. Anyways, it doesn't hurt setting default again. :-)

Yes port 8080 was misleading. I will change it.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
balsmn: Done.

sophokles73:
I have read through the [JVM's supported standard networking properties](https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html). Based on that it seems kind of awkward to define another set of (Hono specific) properties for doing exactly the same thing.

At the very least I would rather use the standard properties in order to configure the `ProxyOptions` but I also wonder if vert.x doesn't by default use the standard properties as well, i.e. you do not need to explicitly set `ProxyOptions` but instead the networking stack picks up the standard properties when establishing a connection.

@balsmn can you try to run the Hono command line consumer with the corresponding JVM standard networking properties for configuring your proxy and see if that works?

balsmn:
@sophokles73 Oh, yes. If Vertx is able to pick it up, it would be much more comfortable and less code. I will try it out and let you know.

balsmn:
@sophokles73 With this PR code, my example client worked setting `proxyType` to HTTP. So I tried the same with the following standard JVM options

```
-Dhttp.proxyHost
-Dhttp.proxyPort 
-Dhttp.proxyUser
-Dhttp.proxyPassword
```
and also 

```
-Dhttps.proxyHost
-Dhttps.proxyPort 
-Dhttps.proxyUser
-Dhttps.proxyPassword
```
Both options didn't work. Then I tried the following, 

```
-DsocksProxyHost
-DsocksProxyPort
-Djava.net.socks.username
-Djava.net.socks.password
```
and finally tried this too.

`-Djava.net.useSystemProxies=true`

But nothing worked except explicitly setting through the vertx `ProxyOptions`.

sophokles73:
@balsmn I guess then we need to set the `ProxyOptions` based on the values of
```
http.proxyHost
http.proxyPort 
http.proxyUser
http.proxyPassword

https.proxyHost
https.proxyPort 
https.proxyUser
https.proxyPassword

socksProxyHost
socksProxyPort
java.net.socks.username
java.net.socks.password
```
in order to not introduce new Hono specific command line variables/properties for the same thing.

Can you update the PR accordingly?

balsmn:
@sophokles73 yes doing it right now....

sophokles73:
@balsmn are you still working on this PR?

balsmn:
@sophokles73 Sorry for the delay. I will finish the required changes and will update the PR by this week.

sophokles73:
no activity anymore

Issue #542, opened 2018-03-20
Add checkstyle rule to find missing javadoc


ctron:
Coming out of the review on PR #539, I think it would make sense to add a checkstyle rule to find missing javadoc.

From what I see Hono has javadoc on all public elements. So that sounds be a reasonable rule.

sophokles73:
@ctron can this be closed?

ctron:
Indeed.

PR #550, opened 2018-03-26
Enable a few Javadoc codestyle checks

ctron:
This PR aims fixing #542 by adding a few codestyle checks and fixing the remaining violations detected. 

The checks enabled are:
* Require javadoc for public types and methods
* Require valid HTML, end of first sentence punctuation
* Check for empty javadocs and at-clauses

I did not enable the check to validate public (global) variables as too many javadocs would have been missing, so I think this might need further discussion. Although I would recommend to document publicly accessible fields. You can check the outcome by adding the following snipped to `default.xml` checkstyle file:

~~~xml
<module name="JavadocVariable">
  <property name="scope" value="public"/>
</module>
~~~


sophokles73:
FMPOV we should start with this one and then (later) also require public fields to have proper JavaDoc (in particular the constants!)

Issue #349, opened 2017-09-22
Support authentication of devices using X.509 certificates in MQTT adapter


sophokles73:
The MQTT adapter should support devices using an X.509 client certificate for authenticating.

sophokles73:
Vert.x 3.4.2 does not support accessing the MQTT client's certificate (yet).
@ppatierno any plans for adding this kind of functionality in `MqttEndpoint`?

ppatierno:
@sophokles73 there is this PR for that https://github.com/vert-x3/vertx-mqtt/pull/42. It should be in the 3.5.0 release.

sophokles73:
Ok, now that we have upgraded to vertx 3.5.0 I can start working on this.
The first question I stumbled upon is this: how do we determine the *tenant* of a device presenting a client certificate during authentication? An obvious option is to extract that information from the certificate, of course. We could e.g. parse the DN from the subject identifier and e.g. expect the tenant to be contained in the OU or DC. However, this would require organisations to create certificates for their devices which are either somehow specific to Hono (including a Hono tenant ID in the cert) or using standard information included in the certificates as the tenant ID (e.g. the concatenated DC values, effectively forming a *domain name*).

Any thoughts? @pellmann @dejanb @ppatierno @sysexcontrol

sophokles73:
Based on some discussions we had about the options for determining the device's tenant based on the certificate it presents during authentication, here's what I think:

Parsing the tenant from the DN of the certificate would require us to either mandate the usage of a certain component of the DN for holding the tenant ID (e.g. `O` or `OU`) or we would need to allow for the user/admin to configure e.g. a *pattern* indicating how to extract the tenant ID from the DN.

Both approaches do not seem overly appealing to me. The former would potentially render existing certificates useless for usage with Hono because the particular component of the DN is either not set at all or is already used for other purpose. The latter would require us to implement error prone pattern matching code while still requiring users to encode the tenant ID within the DN.

I am currently more in favor of using a different approach for determining the device's tenant based on the X.509 certificate presented during authentication:

In order to successfully authenticate a device based on its X.509 cert, we need to configure the protocol adapter(s) with the certificate of the CA that has signed the device's certificate. This is usually done by adding the CA certificate to the protocol adapter's *trust store*.

When a device is (successfully) authenticated, the protocol adapter establishes a chain of trust from the device's certificate up to a certificate contained in the adapter's trust store. This *certificate chain* is made available to the protocol adapter code during connection establishment.

[...]

PR #560, opened 2018-04-06
[#349] Extend Tenant API with operation to get tenant by CA.

sophokles73:
This is the first step for supporting authentication of devices using client certificates.

The PR defines a new operation in the Tenant API (*Get Tenant Information by Trusted Certificate Authority*) which is used to determine the tenant that the device belongs to when it presents its client cert during the TLS handshake.

We should discuss
* whether we want to have a separate operation for this purpose (as implemented in this PR) or whether we would rather modify the existing *get* operation which currently only supports retrieving the tenant by ID. I can also imagine to require the *get* operation to have a request payload specifying either a tenant ID or a subject DN.
* whether the *trusted-ca* configuration property serves its purpose as defined by this PR or needs modification.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/tenant/TenantAmqpEndpoint.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/tenant/TenantAmqpEndpoint.java
@@ -13,10 +13,12 @@
 
 package org.eclipse.hono.service.tenant;
 
+import java.net.HttpURLConnection;
 import java.util.Objects;
 
 import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.auth.HonoUser;
+import org.eclipse.hono.client.ClientErrorException;
 import org.eclipse.hono.config.ServiceConfigProperties;
 import org.eclipse.hono.service.amqp.RequestResponseEndpoint;
 import org.eclipse.hono.util.EventBusMessage;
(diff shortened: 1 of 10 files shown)


service-base/src/main/java/org/eclipse/hono/service/tenant/TenantAmqpEndpoint.java
```diff
@@ -53,28 +55,76 @@ public final String getName() {
...
+            return getAuthorizationService().isAuthorized(clientPrincipal, specificTenantAddress, request.getSubject());
+        }
+    }
+
+    /**
+     * Verifies that a response only contains tenant information that the
+     * client is authorized to retrieve.
+     * <p>
+     * If the response does not contain a tenant ID nor a payload, then the
+     * response is returned as-is.
```
sysexcontrol: I would prefer `unchanged` over `as-is`.

site/content/api/Tenant-API.md
```diff
@@ -277,32 +332,62 @@ Devices belonging to the tenant can connect to Hono via the rest-adapter only an
...
+      "deployment": {
+        "maxInstances": 4
+      }
     }
   ]
 }
 ~~~
 
-In the following example the tenant is allowed to use **all** adapters, as the `adapters` property is omitted in the tenant configuration:
+In the following example the tenant is allowed to use **all** adapters, as the `adapters` property is omitted in the tenant configuration (note that the payload also contains a custom property *plan* which might be used to indicate the customer's service level):
```
sysexcontrol: `...tenant configuration.
Note that ... :`

That is how it is written below  at same place, and I think to begin with a new sentence makes it easier to read.

site/content/api/Tenant-API.md
```diff
@@ -315,24 +400,28 @@ The table below contains the properties which are used to configure a *Hono prot
 
 Protocol Adapters SHOULD use the configuration properties set for a tenant when interacting with devices of that tenant, e.g. in order to make authorization decisions or to limit message rates per tenant etc.
 
-Additionally to the specified properties the JSON object MAY contain an arbitrary number of members with arbitrary names which can be of a scalar or a complex type. 
+The JSON object MAY contain an arbitrary number of additional members with arbitrary names of either scalar or complex type.
 
 ### Examples
 
-Below is an example for a payload of the response to a *get* request for tenant `TEST_TENANT`.
+Below is an example for a payload of the response to a *get* request for tenant `TEST_TENANT`. Note that the payload contains some *custom* properties at both the tenant as well as the adapter configuration level.
```
sysexcontrol: Better name the *custom* properties explicitly?
Would make it easier for readers that are completely new to the tenant API to identify them (and is not much effort).

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantAmqpEndpoint.java
sophokles73: I think that *as-is* is the right term here because we never change the payload. We fail the future if the client is not authorize to retrieve the data, but otherwise we return the payload *as-is*. Using *unchanged* instead could suggest that the payload's content is *filtered*, i.e. we would remove some of the information from the payload, which is not the case. So IMHO it should stay like it is now.

sophokles73:
@ppatierno I can think of the following arguments...

pro two separate methods:
* no change of existing method implementation or clients needed
* very easy to understand
* makes clear that *get*ting a tenant by ID is not supposed to fail because we are getting by the *primary key*, whereas getting by subject DN is actually more of a *find* operation, i.e. the result set may be empty even if the tenant to which the device belongs does actually exist.

con two separate methods(i.e. pro one single get method):
* might be seen as a first step in establishing a scheme where we add a new method for each *get*/*find* variant instead of just allowing additional search criteria in the payload. Thus, this ight *bloat* the interface over time
* requires to keep support for the existing *get by ID* method, i.e. still supporting the tenant ID in the application properties  for backward compatibility reasons.

Personally, I also do like the *one get method* approach better. I think it provides for more flexibility regarding further extensions of the search criteria in the future.

sysexcontrol:
I see the pro points for having only one get method, I can live with that as well.

Although the *bloat* danger is kind of true, it only moves the complexity to the implementation (compared to just not implementing one of the get variants). So if we have different payloads with search criteria, IMHO we should provide at least a helper class to make the implementation for the different variants as simple as possible (and use this in our own file based implementation to show the usage).

sophokles73:
fixed in bafd9c9e2e8648e0a5d357ab3dd682815d241837

Issue #504, opened 2018-02-22
Extend Credentials API to support client context


dejanb:
In order to be able to use client provided information during authentication process, we need to extend the existing Credentials API *get* operation with optional bag of properties. Credentials API implementation may then use this information in order to determine device identity. 
For example, MQTT clients always provide a MQTT *client-id* in the *CONNECT* packet.

ppatierno:
> MQTT clients always provide a MQTT client-id in the CONNECT packet.

It's not always true, not mandatory by spec. The client-id can be zero bytes length and in this case the broker can assign a random client-id to the client (which in any case doesn't know what it is because it's not provided in the CONNACK). The `MqttServer` does that (it's configurable) but how it can be related with the Credentials API feature we are discussing ?

sophokles73:
> It's not always true, not mandatory by spec. The client-id can be zero bytes length

That is true, however, the way I read the spec
> A Server MAY allow a Client to supply a ClientId that has a length of zero bytes, however if it does so the
Server MUST treat this as a special case and assign a unique ClientId to that Client. It MUST then
process the CONNECT packet as if the Client had provided that unique ClientId [MQTT-3.1.3-6].

the result will be the same, i.e. if the connection is established at all, the MQTT adapter will have a client-id at hand for the device, right?

ppatierno:
I was referring to the the phrase : 

"MQTT clients always provide a MQTT client-id in the CONNECT packet"

It's not true. As the spec says and all the brokers implement, a client can avoid to provide a client-id. In this case two things can happen : 

- if clean session is true, the server accept the CONNECT and assigns a random client-id to the client for its internal handling but the client will never know this assigned client-id
- if clean session is false, the server refuse the connection because if the client wants to store the session on the broker, it has to provide the client-id needed by the server for handling such a session internally. (of course the randomly generated id is not useful, because the client doesn't know about that and can't provide on re-connection for recovering a previous session).

So "clients always provide a client-id" is not true, it can be empty.

> the result will be the same, i.e. if the connection is established at all, the MQTT adapter will have a client-id at hand for the device, right?

Btw you are right here, because the MQTT adapter knows the randomly generated client-id (the MQTT server does it by default but it can be changed) and its related MQTT connection.

sophokles73:
@dejanb can we close this issue?

PR #561, opened 2018-04-06
[#504] extend credentials api to support client context

dejanb:


dejanb:
@sophokles73 First stab at extending an API. Will add docs if this is a way to go. Let me know if you see anything missing.

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsService.java
```diff
@@ -65,11 +66,11 @@
      * @see <a href="https://www.eclipse.org/hono/api/credentials-api/#get-credentials">
      *      Credentials API - Get Credentials</a>
      */
-    void get(String tenantId, String type, String authId, Handler<AsyncResult<CredentialsResult<JsonObject>>> resultHandler);
+    void get(String tenantId, String type, String authId, JsonObject clientContext, Handler<AsyncResult<CredentialsResult<JsonObject>>> resultHandler);
```
sophokles73: I would rather have an overloaded version of the *get* method instead of requiring client code to always provide `null` as a parameter when no client context is available. You can already see the (ugly) result in the test cases that you had to adapt accordingly...

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
```diff
@@ -322,6 +324,17 @@ private JsonObject getSingleCredentials(final String tenantId, final String auth
                     final JsonObject authIdCredential = (JsonObject) authIdCredentialEntry;
                     // return the first matching type entry for this authId
                     if (type.equals(authIdCredential.getString(CredentialsConstants.FIELD_TYPE))) {
+                        if (clientContext != null) {
```
sophokles73: FMPOV we do not need to process the client context information at all In our example Device Registry. We mainly introduced it so that other implementations could use the information for whatever purpose they see fit.

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsService.java
dejanb: +1 I missed to add it. It's now there and I removed all ugly calls from the tests

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
dejanb: I think it's valuable to have it tested at least in the most simple form, so that things don't get broken over time. It doesn't compromise anything else, so I'm inclined to leave it in.

sophokles73: My point is that if we do not specify any expected processing of the extra information in the spec (which we don't/shouldn't), then there actually is nothing to test, right? Or would you like to specify the behavior that you are implementing here as the *expected behavior* of at least our example Device Registry (if not of the operation in general)?

dejanb: My concern is if it's just ignored and not tested anywhere, someone will be inclined to just deleted after a while. Also, this gives an example of "this feature" to the people that wants to implement/adapt device registries to Hono API. It doesn't change anything for the folks that don't care about it, so I don't see a problem there.

sophokles73: Ok, then please don't forget to describe this behavior accordingly in the Device Registry User Guide.

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsServiceTest.java
```diff
@@ -339,6 +340,42 @@ public void testGetCredentialsSucceedsForExistingCredentials(final TestContext c
         get.await(2000);
     }
 
+    @Test
+    public void testGetCredentialsSucceedsForProperClientContext(final TestContext ctx) {
```
sophokles73: missing JavaDoc

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsServiceTest.java
```diff
@@ -339,6 +340,42 @@ public void testGetCredentialsSucceedsForExistingCredentials(final TestContext c
...
+            assertThat(s.getStatus(), is(HttpURLConnection.HTTP_OK));
+            assertThat(s.getPayload().getString(CredentialsConstants.FIELD_AUTH_ID), is("myId"));
+            assertThat(s.getPayload().getString(CredentialsConstants.FIELD_TYPE), is("myType"));
+            get.complete();
+        }));
+        get.await(2000);
+    }
+
+    @Test
+    public void testGetCredentialsFailsForWrongClientContext(final TestContext ctx) {
```
sophokles73: missing JavaDoc

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsServiceTest.java
```diff
@@ -454,6 +491,7 @@ private static void register(
             final String deviceId,
             final String authId,
             final String type,
+            final JsonObject clientContext,
```
sophokles73: can you also please add a variant of this method without the clientContext param and remove all the ugly `null` passing invocations from the test cases again?

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
dejanb: This change is not propagated to the default device registry implementation. It's just in the core API at the moment. It might get useful to add it as part of the integration work at some point, but I'm inclined to leave it as is for now. WDYT?

sophokles73: I don't understand your point. What do you mean by
> This change is not propagated to the default device registry implementation. It's just in the core API at the moment.

The code you are working on is part of the Device Registry component we ship with Hono.

dejanb: I meant that the HTTP API of Device Registry don't handle these extra parameters and there's no integration test for this functionality (over HTTP). I can add those for completeness.

sophokles73: Now I am completely confused. What does the HTTP API have to do with this? Do you mean that we cannot register this kind of additional information using the HTTP API?

dejanb:
@sophokles73 IT test and docs added

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
```diff
@@ -227,6 +228,16 @@ protected void doStop(final Future<Void> stopFuture) {
         }
     }
 
+    /**
+     * {@inheritDoc}
+     * <p>
+     * The result object will include a <em>no-cache</em> directive.
+     */
+    @Override
+    public void get(String tenantId, String type, String authId, Handler<AsyncResult<CredentialsResult<JsonObject>>> resultHandler) {
```
sysexcontrol: finals missing here

site/content/user-guide/device-registry.md
```diff
@@ -3,17 +3,38 @@ title = "Device Registry"
...
+
+* [Tenant API]({{< relref "api/Tenant-API.md" >}})
+* [Device Registration API]({{< relref "api/Device-Registration-API.md" >}})
+* [Credentials API]({{< relref "api/Credentials-API.md" >}})
+
+It implements AMQP 1.0 API endpoints and provides JSON based file persistence for data. 
+
 In addition to the AMQP 1.0 based API endpoints, the Device Registry also exposes HTTP resources for managing the contents of the registry.
 
-<!--more-->
```
sophokles73: please do not remove this separator, it makes sure that the overview page only contains the text above

site/content/user-guide/device-registry.md
```diff
@@ -3,17 +3,38 @@ title = "Device Registry"
...
+
+### Get Credentials
+
+The [Get Credentials]({{< relref "api/Credentials-API.md#get-credentials" >}}) operation of Credentials API defines that arbitrary array of properties can be passed to help identify the device. 
+It's not specified if and how these properties can be used by the implementation. The device registry will try to match the
+properties against the properties defined in the credentials record. It will return the credentials only if all the properties in the request are successfully matched.
+
+## HTTP Resources
+
+<!--more-->
```
sophokles73: pleas move this up to its original position

site/content/user-guide/device-registry.md
```diff
@@ -3,17 +3,38 @@ title = "Device Registry"
...
 The resources have been designed to provide convenient access to the registry's content using command line tools like *curl* or *HTTPie*.
 {{% /warning %}}
 
+## Implementation details
+
+This section describes implementation specifics of certain parts of the Device Registry.
+
+### Get Credentials
+
+The [Get Credentials]({{< relref "api/Credentials-API.md#get-credentials" >}}) operation of Credentials API defines that arbitrary array of properties can be passed to help identify the device.
```
sophokles73: FMPOV this whole section should go down to the description of the *Get Credentials* resource. It feels very much out of context in its current place.

dejanb:
@sophokles73 should be fixed now

PR #562, opened 2018-04-06
Hono device connection notifications specification

sysexcontrol:
Documented and specified the different events as part of Hono.

Not all is finished yet, but it is a full specification of how notifications are handled for the prototype of command and control.


site/content/api/Event-API.md
```diff
@@ -76,4 +81,82 @@ The following sequence diagram illustrates the flow of messages involved in a *B
...
+
+## Well-known event message types
+
+Hono defines *well-known* events that are of a specific *content-type*.
+
+NB: currently there is only one such *well-known* event.
+
+### Lifecycle Notifications
+
+A specific event can signal that a device is `connected` (e.g. to signal that it is ready to receive a command for a specific time interval)
```
sophokles73: FMPOV it is not the event that signals anything. However, the event can be used by other components to signal that a device has connected to a protocol adapter. My understanding was that we agreed to distinguish between the more technical events for signalling that a device has connected/disconnected and the *flag* that devices use to signal that they are now ready to receive a command. IMHO we should not start to mix them up again by indicating that the *connected* event is used to signal that a device is ready to receive a command. FMPOV this is not the case. This event should simply indicate that a device has connected, nothing more, nothing less.

site/content/api/Event-API.md
```diff
@@ -76,4 +81,82 @@ The following sequence diagram illustrates the flow of messages involved in a *B
...
+Hono defines *well-known* events that are of a specific *content-type*.
+
+NB: currently there is only one such *well-known* event.
+
+### Lifecycle Notifications
+
+A specific event can signal that a device is `connected` (e.g. to signal that it is ready to receive a command for a specific time interval) 
+or that a device is `disconnected` (e.g to signal that it is not ready to receive a command anymore).
+ 
+*Business Applications* may want to react to such an event by sending a command upstream to the device.
```
sophokles73: FMPOV this is **not** the purpose of this event. We cannot prevent applications from trying to send a command on reception of this event, but the real trigger for sending a command should be another event, specifically designed for that purpose.

site/content/api/Event-API.md
```diff
@@ -76,4 +81,82 @@ The following sequence diagram illustrates the flow of messages involved in a *B
...
+For that purpose, the protocol adapter instance that received such an event opens a receiver link for the device (scoped to its tenant) that
+is used to send the command to the appropriate protocol adapter instance.
+
+Such an event may be issued by the device itself or by a protocol adapter detecting that a device is connected or disconnected.
+ For further details on which part of a setup issues such an event see (... concept page).
+
+
+*Lifecycle notifications* may be decided to not being durable, but this is left to the setup of the *AMQP network* and is not further specified in the following.
```
sophokles73: Are you saying that events are no longer *durable* by definition?

site/content/api/Event-API.md
```diff
@@ -76,4 +81,82 @@ The following sequence diagram illustrates the flow of messages involved in a *B
...
+
 See [*Telemetry API*]({{< relref "Telemetry-API.md" >}}) for definition of message format. 
+
+## Well-known event message types
+
+Hono defines *well-known* events that are of a specific *content-type*.
+
+NB: currently there is only one such *well-known* event.
+
+### Lifecycle Notifications
```
sophokles73: FMPOV this has not much to do with the lifecycle of a device (which is governed by registering the device, disabling it, etc).

site/content/api/Event-API.md
```diff
@@ -76,4 +81,82 @@ The following sequence diagram illustrates the flow of messages involved in a *B
...
+| :-----------------  | :-------: | :--------- | :----------   | :---------- |
+| *tenant-id*         | *yes*     | *string*   |               | The tenant identifier to send the notification for. |
+| *device-id*         | *yes*     | *string*   |               | The device identifier to send the notification for. |
+| *cause*             | *yes*     | *string*   |               | Value must be either `connected` or `disconnected` . |
+| *source*            | *yes*      | *string*   |               | Value can be either the name of a protocol adapter (`hono-mqtt`, `hono-http`, `hono-kura`), if the device connected/disconnected to the appropriate protocol adapter,  or `device`, if the device sent the notfication itself.
+| *connection-timestamp*  | *no*      | *string*   |   null        | The point in time when the event was created (set by the issuer). If not null, the value MUST be an ISO 8601 compliant combined date and time representation in extended format. **NB:** due to network latencies this timestamp will typically be some time in the past already when the event is received by the application. |
```
sophokles73: as I stated above, I strongly believe that we should define a separate, specific event for signaling that a device is ready to receive a command

site/content/api/Telemetry-API.md
```diff
@@ -65,6 +65,9 @@ The following table provides an overview of the properties a client needs to set
...
 | *device_id*    | yes       | *application-properties* | *string*  | MUST contain the ID of the device the data in the payload has been reported by. |
 | *reg_assertion*| yes       | *application-properties* | *string*  | A [JSON Web Token](https://jwt.io/introduction/) issued by the [Device Registration service]({{< relref "api/Device-Registration-API.md#assert-device-registration" >}}) asserting the device's registration status. |
+| *connection_timestamp* | no       | *application-properties* | *string*  | The point in time that defines the beginning of a time interval for which a device may receive upstream data (e.g. a command). The value MUST be an ISO 8601 compliant combined date and time representation in extended format. **NB:** due to network latencies this timestamp will typically be some time in the past already when the event is received by the application. |
```
sophokles73: why can't we simply use the standard *creation-time* property?

BTW these properties also need to be supported by the Event API.

site/content/concepts/device-lifecycle-notifications.md
```diff
@@ -0,0 +1,93 @@
...
+Please join our mailing list or consult the IRC channel in case you have any questions or want to participate.
+{{% /warning %}}
+
+<!--more-->
+
+Business applications connected to Hono may want to invoke arbitrary functionality when a device connects or disconnects to Hono.
+This is especially necessary to implement Command and Control, but is not limited to this Use Case. Integrating digital twins, like
+[Eclipse Ditto] (https://projects.eclipse.org/proposals/eclipse-ditto), also benefit from knowing when a device just connected to Hono.
+
+For these scenarios Hono defines a lifecycle notification concept that covers different patterns that try to cover several classes
```
sophokles73: We need to distinguish between connected/disconnected events published by protocol adapters and how a device signals that it can receive a command. These are two different things and should be explained here. FMPOV they are not both part of what is called *lifecycle notifications* here. Additionally, IMHO that term does not match the purpose of either of the two concepts and should be replaced by two other terms.

site/content/concepts/device-lifecycle-notifications.md
```diff
@@ -0,0 +1,93 @@
...
+
+Besides that devices that connect to Hono may often not be directly ready for receiving upstream data, but need to separate
+their availability from the connection itself.
+
+Considering these different patterns, the main principle for lifecycle notifications is the following:
+
+# Main principle
+
+**The standard responsibility to signal that a device is ready to receive upstream data is at the device itself.**
+
```
sophokles73: This sounds a little demanding, don't you think?
Maybe we can describe this more like an offering, e.g.
>  A device MAY signal its ability to receive a command from an application.

And then elaborate on the different options and means how to do that.

site/content/api/Telemetry-API.md
```diff
@@ -65,6 +65,9 @@ The following table provides an overview of the properties a client needs to set
...
+| *stay_connected* | no       | *application-properties* | *long*  | In context with the `connection-timestamp`  it defines the time interval in milliseconds in which a device should be available for receiving upstream data. | 
+| *reg_assertion*| yes       | *application-properties* | *string*  | A [JSON Web Token](https://jwt.io/introduction/) issued by the [Device Registration service]({{< relref "api/Device-Registration-API.md#assert-device-registration" >}}) asserting the device's registration status. |
```
sophokles73: this seems to be a duplication

site/content/api/Event-API.md
sysexcontrol: After having read your other comments I think I see your concern here:

I think we agreed on ONE event, but two different events make some sense, I agree (see my comments below).
Will rephrase the wording that it is not the *event* that is signaling something itself (how should that be?), but it can be used by others to signal connection/disconnection. And remove the example in brackets here, so there is no reference to command readiness anymore.

site/content/api/Event-API.md
sysexcontrol: No, we cannot prevent that, but IMHO we had *not* agreed on two events yet.

What I did instead was to define optional parts into this event: the device can state how long it tries to stay connected and specify by that it can make sense to send a command.

We can strictly separate it:
- connected/disconnected *only* from protocol adapters (often this might not make sense, see piggy-backing telemetry data)
- command readiness only from device itself

Not convinced much about the first event though - we wanted to prevent the doubling of messages in case of telemetry messages for devices connecting via HTTP e.g.
Can imagine that protocol adapters can be configured for a tenant to always issue the first event type (for certain device setups).

Will include both events in the next commit.

site/content/api/Event-API.md
sysexcontrol: No, where are you reading that?
I left it open if the connected/disconnected events are durable. Not by definition, just left open, see your comment to the discussion PR:

`This seems to be an issue worth discussing based on my comment above. We currently mark all events as durable btu I agree that we might want to relax this requirement and allow the issuer of the event to mark it as such or not.`

Would rephrase it that the sender can decide if the event should be durable.
Often IMHO the connected event may not make too much sense to be durable, though.

site/content/api/Event-API.md
sysexcontrol: Full ack, will rename the notification - lifecycle is not a good naming.

site/content/api/Telemetry-API.md
sysexcontrol: Good point, we can indeed use the creation-time property. Will change that.

The Event API should support the `stay-connected` property as well, agree on that.

site/content/concepts/device-lifecycle-notifications.md
sysexcontrol: No, as agreed above already. Will replace it by different wording.

site/content/concepts/device-lifecycle-notifications.md
sysexcontrol: Sounds demanding, right - and it did its purpose :-) 

We can use the *offering* phrase, but we should clearly state that usually not the protocol adapters will take over the job to signal command readiness - it has to be done by the device itself. I will search for a phrasing *in between* these variants...

site/content/api/Telemetry-API.md
sysexcontrol: Yes, thanks for spotting!

site/content/api/Event-API.md
sophokles73: > connected/disconnected only from protocol adapters (often this might not make sense, see piggy-backing telemetry data)

This is exactly the use-case that the Kapua folks seem to have in mind. For Kapua, it is helpful to have an indicator of the connection status of a gateway. In the Kapua case we are only talking MQTT based gateways (Kura) which connect on startup and stay connected all the time. So this event is basically only used for feeding a *gateway connected* LED in the UI.

As such, I think it makes a lot of sense to

* have separate events for the two distinct purposes and
* make the emitting of the connected/disconnected events configurable per tenant and adapter.

sysexcontrol: Ok, this makes sense to me too: we use the tenant based configuration of protocol adapters to let them emit the connected/disconnected event (if configured).

So I will introduce both events in next commits (but will for now do some minor prototyping first).

site/content/api/Event-API.md
sysexcontrol: No, I thought of changing that (in the spec and in the implementation).

But I would propose to do that later, and for now let the connection events be durable as all other events (better get something running before dealing with the optimizations).
So I will just delete the words about durability, then the default behaviour of events for now is valid here as well.

site/content/concepts/device-lifecycle-notifications.md
sysexcontrol: I believe the same :+1: 

So far I have only thought about the events, and not even about the downstream messages.
But for the event signaling command readiness the same probably is necessary.

For now, I would only support
- devices can send the command readiness event
- devices can piggy-back telemetry/event data
Later both of that can be configured per tenant and protocol adapter to do that instead of the device (to support specific device setups that do not support it).

site/content/api/Event-API.md
```diff
@@ -76,4 +81,82 @@ The following sequence diagram illustrates the flow of messages involved in a *B
...
+now is able to receive a command for one minute:
+
+~~~json
+{
+  "tenant-id": "my-tenant",
+  "device-id": "sensor1",
+  "cause": "connect",
+  "source": "device",
+  "creation-timestamp": "2018-04-06T10:00:00+0100",
+  "ttl": "60000",
```
sysexcontrol: `60000` instead of "60000".

sysexcontrol:
@ctron @sophokles73 
I defined two separate events now in the last commit for the different notifications (connected/disconnected and "device command readiness").

This can be the base for implementing both parts - the device command readiness being already implemented in #576 (minor modifications necessary after the specification), the other part being contained in #569 .

The concept page and the piggy-backing use cases (telemetry AND event API) is for now not contained in the commit, will come soon in a separate commit.

site/content/api/Event-API.md
```diff
@@ -89,42 +89,28 @@ See [*Telemetry API*]({{< relref "Telemetry-API.md" >}}) for definition of messa
...
 
+***Device connection notification payload***
 
-*Lifecycle notifications* may be decided to not being durable, but this is left to the setup of the *AMQP network* and is not further specified in the following.
-
-
-***Lifecycle Notification Payload***
-
-The following table provides an overview of the properties a client needs to set for a *lifecycle notification* event.
+The following table provides an overview of the properties a client needs to set for a *device connection notification* event.
```
ctron: If later on the source explicitly is a "protocol adapter", then I think here "client" should also become "protocol adapter".

sysexcontrol: Currently this would be true, but what if we have other "clients" later on, that are not protocol adapters, sending this event?

How about changing it to `event sender`, so resulting in 
`... overview of the properties an event sender needs to set ...`?
Then we would be generic here, and that currently the protocol adapters are the only `event senders` would be made clear in the concept page (which I provide in another commit, needs some rework still).
WDYT?

sophokles73:
Any progress on this PR?

sysexcontrol:
@sophokles73 : to ease highly appreciated review by splitting PRs :+1: 

But I haven't understood the current problem:
The last commit reworked the Event API description by defining two separate notifications that have their own content-type, so

> if you want to establish a generic notification event that can be used for multiple situations and which need to be distinguished by means of information contained in the event payload, 

No, that wouldn't make sense, of course!

> if you want to define specific events for indicating that a device has connected/disconnected from a protocol adapter

Yes, this is one of the two contained events.

> if you want to device a specific event that a device can use to indicate that it is now ready to receive a command.

Yes, this is the other contained event.

To ease the review, I would open a new PR then containing just the event definition (which I find especially important to quickly get finished). The piggy-back definition and the concept page (currently still contained here) could go to a different PR then, afterwards closing this one.

Does this solve the problem?

sophokles73:
> To ease the review, I would open a new PR then containing just the event definition (which I find especially important to quickly get finished). The piggy-back definition and the concept page (currently still contained here) could go to a different PR then, afterwards closing this one.

So, then this PR is *only* about how a device can indicate that it is ready to receive a command, right?
If so, can we then name it accordingly, e.g. *Allow devices to indicate readiness-to-receive* or something like that? I would then actually rather have this PR address the whole issue from top to bottom, i.e. the concept and the two ways that the device can indicate its readiness so that we can discuss all aspects consistently without jumping hence and forth between PRs.

sysexcontrol:
Ok, I can split it like that (instead of defining the two events in one PR). 
Probably good for focussing the discussion to the two different notifications...

So the new PR will contain ALL aspects of a device sending a notification that it is ready to receive a command. And NOTHING of protocol adapters signalling the connection/disconnection state of a device.
But I will describe that in a second (also new) PR then and close this one completely - that reflects the separation better IMHO (and does not contain the history of the originally mixed PR anymore).

sysexcontrol:
Created the PR #584 that contains `only` the mechanisms that allows device to indicate their readiness to receive messages.

sysexcontrol:
@sophokles73 : yes, that is the plan.
Since the second new PR (for the connected/disconnected event) is not there yet, so I thought to keep it open until then, however.

But I am open to close it just now alternatively as well, just drop me a note here...

sophokles73:
@sysexcontrol do we still need this PR?

sysexcontrol:
@sophokles73 : FMPOV I can close this PR, if @ctron agrees.

@ctron : since we decided to strictly separate the PRs for specifying the "device command readiness" and the "connected/disconnected notification", I so far kept this PR open.
I think I can close this PR now, and add the specification for the "connected/disconnected" notification in a different small PR again - do you agree?

sysexcontrol:
Closing this PR now, in favour of specialized PR #584 .
Second PR for specification of "connection/disconnection" notification will be opened.

PR #567, opened 2018-04-10
Fix MQTT message count metrics

ctron:
This change does fix two issues with the MQTT connection count metrics.
The first issue is a simple mistake of calling increment instead of
decrement. The second issue ignores unauthenticated MQTT connections in
the metrics.


pellmann:
LGTM - we could also add the "unauthenticatedConnections" to the default Grafana board. Maybe create an issue if you do not want to add it?

Issue #565, opened 2018-04-10
Protocol connection events


ctron:
In addition to the device notification functionality, we discussed in Berlin that we do need some kind event which notifies of the fact that "someone" (not necessarily a device) connected.

This would be a more "debug" like event and might not make sense for all protocol adapters. It may however make sense for MQTT, where a connection can be successfully established and be kept open for a longer period of time. This might be useful when the device/gateway connections successfully, but doesn't transmit any data over the MQTT link.

The same might be useful for the (hopefully) upcoming AMQP protocol adapter.

This functionality should not be seen as a replacement for the "device notification" feature currently being discussed in #562, #535 and #537.

---

My proposal would be to start with an interface which being called to handle those events. Providing a "logging only" implementation.

---

I tried to re-cap in this issue the discussion we had in the last Hono call, splitting off the "device notification" we need for command & control, vs the MQTT connection notification event we need for better debugging.

sophokles73:
@ctron can we remove the term *alerting* from this issue? It seems to interfere with #502 which is about something different but might be mistaken for this one here.

ctron:
> can we remove the term alerting from this issue? It seems to interfere with #502 which is about something different but might be mistaken for this one here.

We could also send the notifications over this new "alerting" functionality. But we could change the naming of this functionality to not imply the *alerting* functionality has to be used.

ctron:
Maybe "protocol connection events" work better:

*  Actually we are not talking about devices/gateways here, but about connections on the side of the protocol adapter.
*  It is all about the technical connection which got opened with the protocol adapter

sophokles73:
> But we could change the naming of this functionality to not imply the alerting functionality has to be used.

That was my concern.

ctron:
Is the changed title an improvement?

ctron:
Closed by PR #569

PR #569, opened 2018-04-11
Connection events implementation

ctron:
These two commits are a base for discussion towards connection events as described in #565.

The PR is not intended to be merged right away, but to check out the direction.

Please check the second commit to see the actual changes as the first commit only contains code formatter changes.

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
@@ -40,11 +40,13 @@ import org.springframework.beans.factory.annotation.Autowired;
 
 import io.netty.handler.codec.mqtt.MqttConnectReturnCode;
 import io.netty.handler.codec.mqtt.MqttQoS;
+import io.vertx.core.AsyncResult;
 import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.json.JsonObject;
 import io.vertx.mqtt.MqttAuth;
+import io.vertx.mqtt.MqttConnectionException;
 import io.vertx.mqtt.MqttEndpoint;
 import io.vertx.mqtt.MqttServer;
 import io.vertx.mqtt.MqttServerOptions;
@@ -94,12 +96,12 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends ProtocolAd
 
     @Override
     protected final int getActualPort() {
-        return (server != null ? server.actualPort() : Constants.PORT_UNCONFIGURED);
+        return server != null ? server.actualPort() : Constants.PORT_UNCONFIGURED;
     }
 
     @Override
     protected final int getActualInsecurePort() {
-        return (insecureServer != null ? insecureServer.actualPort() : Constants.PORT_UNCONFIGURED);
+        return insecureServer != null ? insecureServer.actualPort() : Constants.PORT_UNCONFIGURED;
     }
 
     /**
@@ -187,7 +189,7 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends ProtocolAd
     private Future<MqttServer> bindMqttServer(final MqttServerOptions options, final MqttServer mqttServer) {
 
         final Future<MqttServer> result = Future.future();
-        final MqttServer createdMqttServer = (mqttServer == null ? MqttServer.create(this.vertx, options) : mqttServer);
+        final MqttServer createdMqttServer = mqttServer == null ? MqttServer.create(this.vertx, options) : mqttServer;
 
         createdMqttServer
                 .endpointHandler(this::handleEndpointConnection)
@@ -246,6 +248,36 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends ProtocolAd
     }
 
     /**
+     * Create a future indicating a rejected connection.
+     * 
+     * @param returnCode The error code to return.
+     * @return A future indicating a rejected connection.
+     */
+    protected static <T> Future<T> rejected(final MqttConnectReturnCode returnCode) {
+        return Future.failedFuture(new MqttConnectionException(
+                returnCode != null ? returnCode : MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE));
+    }
+
+    /**
+     * Create a future indicating an accepted connection with an authenticated device.
+     * 
+     * @param authenticatedDevice The device that was authenticated, may be {@code null}
+     * @return A future indicating an accepted connection.
+     */
+    protected static Future<Device> accepted(final Device authenticatedDevice) {
+        return Future.succeededFuture(authenticatedDevice);
+    }
+
+    /**
+     * Create a future indicating an accepted connection without an authenticated device.
+     * 
+     * @return A future indicating an accepted connection.
+     */
+    protected static Future<Device> accepted() {
+        return Future.succeededFuture(null);
+    }
+
+    /**
      * Invoked when a client sends its <em>CONNECT</em> packet.
      * <p>
      * Authenticates the client (if required) and registers handlers for processing messages published by the client.
@@ -256,19 +288,55 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends ProtocolAd
 
         LOG.debug("connection request from client [clientId: {}]", endpoint.clientIdentifier());
 
-        isConnected().map(ok -> {
-            if (getConfig().isAuthenticationRequired()) {
-                handleEndpointConnectionWithAuthentication(endpoint);
+        isConnected()
+                .compose(v -> handleConnectionRequest(endpoint))
+                .setHandler(result -> handleConnectionRequestResult(endpoint, result));
+
+    }
+
+    private Future<Device> handleConnectionRequest(final MqttEndpoint endpoint) {
+        if (getConfig().isAuthenticationRequired()) {
+            return handleEndpointConnectionWithAuthentication(endpoint);
+        } else {
+            return handleEndpointConnectionWithoutAuthentication(endpoint);
+        }
+    }
+
+    private void handleConnectionRequestResult(final MqttEndpoint endpoint,
+            final AsyncResult<Device> authenticationAttempt) {
+
+        if (authenticationAttempt.succeeded()) {
+
+            sendConnectedEvent(endpoint.clientIdentifier(), authenticationAttempt.result())
+                    .setHandler(sendAttempt -> {
+                        if (sendAttempt.succeeded()) {
+                            endpoint.accept(false);
+                        } else {
+                            LOG.warn(
+                                    "connection request from client [clientId: {}] rejected due to connection event "
+                                            + "failure: {}",
+                                    endpoint.clientIdentifier(),
+                                    MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE,
+                                    sendAttempt.cause());
+                            endpoint.reject(MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE);
+                        }
+                    });
+
+        } else {
+
+            final Throwable t = authenticationAttempt.cause();
+            if (t instanceof MqttConnectionException) {
+                final MqttConnectReturnCode code = ((MqttConnectionException) t).code();
+                LOG.debug("connection request from client [clientId: {}] rejected with code: {}",
+                        endpoint.clientIdentifier(), code);
+                endpoint.reject(code);
             } else {
-                handleEndpointConnectionWithoutAuthentication(endpoint);
+                LOG.debug("connection request from client [clientId: {}] rejected: {}",
+                        endpoint.clientIdentifier(), MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE);
+                endpoint.reject(MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE);
             }
-            return null;
-        }).otherwise(t -> {
-            LOG.debug("connection request from client [clientId: {}] rejected: {}",
-                    endpoint.clientIdentifier(), MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE);
-            endpoint.reject(MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE);
-            return null;
-        });
+
+        }
     }
 
     /**
@@ -276,45 +344,48 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends ProtocolAd
      * the {@linkplain ProtocolAdapterProperties#isAuthenticationRequired() authentication required} configuration
      * property to {@code false}.
      * <p>
-     * Registers a close handler on the endpoint which invokes {@link #close(MqttEndpoint)}. Registers a publish handler
-     * on the endpoint which invokes {@link #onPublishedMessage(MqttContext)} for each message being published by the
-     * client. Accepts the connection request.
+     * Registers a close handler on the endpoint which invokes {@link #close(MqttEndpoint, Device)}. Registers a publish
+     * handler on the endpoint which invokes {@link #onPublishedMessage(MqttContext)} for each message being published
+     * by the client. Accepts the connection request.
      * 
      * @param endpoint The MQTT endpoint representing the client.
      */
-    private void handleEndpointConnectionWithoutAuthentication(final MqttEndpoint endpoint) {
+    private Future<Device> handleEndpointConnectionWithoutAuthentication(final MqttEndpoint endpoint) {
 
         endpoint.closeHandler(v -> {
-            close(endpoint);
+            close(endpoint, null);
             LOG.debug("connection to unauthenticated device [clientId: {}] closed", endpoint.clientIdentifier());
             metrics.decrementUnauthenticatedMqttConnections();
         });
         endpoint.publishHandler(message -> onPublishedMessage(new MqttContext(message, endpoint)));
 
         LOG.debug("unauthenticated device [clientId: {}] connected", endpoint.clientIdentifier());
-        endpoint.accept(false);
         metrics.incrementUnauthenticatedMqttConnections();
+        return accepted();
     }
 
-    private void handleEndpointConnectionWithAuthentication(final MqttEndpoint endpoint) {
+    private Future<Device> handleEndpointConnectionWithAuthentication(final MqttEndpoint endpoint) {
 
         if (endpoint.auth() == null) {
+
             LOG.debug("connection request from device [clientId: {}] rejected: {}",
                     endpoint.clientIdentifier(), "device did not provide credentials in CONNECT packet");
-            endpoint.reject(MqttConnectReturnCode.CONNECTION_REFUSED_BAD_USER_NAME_OR_PASSWORD);
+
+            return rejected(MqttConnectReturnCode.CONNECTION_REFUSED_BAD_USER_NAME_OR_PASSWORD);
 
         } else {
 
             final DeviceCredentials credentials = getCredentials(endpoint.auth());
 
             if (credentials == null) {
+
                 LOG.debug("connection request from device [clientId: {}] rejected: {}",
                         endpoint.clientIdentifier(), "device provided malformed credentials in CONNECT packet");
-                endpoint.reject(MqttConnectReturnCode.CONNECTION_REFUSED_BAD_USER_NAME_OR_PASSWORD);
+                return rejected(MqttConnectReturnCode.CONNECTION_REFUSED_BAD_USER_NAME_OR_PASSWORD);
 
             } else {
 
-                getTenantConfiguration(credentials.getTenantId()).compose(tenantConfig -> {
+                return getTenantConfiguration(credentials.getTenantId()).compose(tenantConfig -> {
                     if (tenantConfig.isAdapterEnabled(getTypeName())) {
                         LOG.debug("protocol adapter [{}] is enabled for tenant [{}]",
                                 getTypeName(), credentials.getTenantId());
@@ -329,24 +400,24 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends ProtocolAd
                     final Future<Device> result = Future.future();
                     usernamePasswordAuthProvider.authenticate(credentials, result.completer());
                     return result;
-                }).map(authenticatedDevice -> {
+                }).compose(authenticatedDevice -> {
                     LOG.debug("successfully authenticated device [tenant-id: {}, auth-id: {}, device-id: {}]",
                             authenticatedDevice.getTenantId(), credentials.getAuthId(),
                             authenticatedDevice.getDeviceId());
                     onAuthenticationSuccess(endpoint, authenticatedDevice);
-                    return null;
-                }).otherwise(t -> {
+                    return accepted(authenticatedDevice);
+                }).recover(t -> {
                     LOG.debug("cannot authenticate device [tenant-id: {}, auth-id: {}]",
                             credentials.getTenantId(), credentials.getAuthId(), t);
-                    if (ServerErrorException.class.isInstance(t)) {
+                    if (t instanceof ServerErrorException) {
                         // one of the services we depend on might not be available (yet)
-                        endpoint.reject(MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE);
+                        return rejected(MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE);
                     } else {
                         // validation of credentials has failed
-                        endpoint.reject(MqttConnectReturnCode.CONNECTION_REFUSED_NOT_AUTHORIZED);
+                        return rejected(MqttConnectReturnCode.CONNECTION_REFUSED_NOT_AUTHORIZED);
                     }
-                    return null;
                 });
+
             }
         }
     }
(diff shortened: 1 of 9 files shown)


adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -13,6 +13,10 @@
 
 package org.eclipse.hono.adapter.mqtt;
 
+import static io.netty.handler.codec.mqtt.MqttConnectReturnCode.CONNECTION_REFUSED_BAD_USER_NAME_OR_PASSWORD;
```
sophokles73: we do not statically import constants, because we find it important to see where the constant comes from in the code.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -329,39 +390,39 @@ private void handleEndpointConnectionWithAuthentication(final MqttEndpoint endpo
...
 
     private void onAuthenticationSuccess(final MqttEndpoint endpoint, final Device authenticatedDevice) {
 
         endpoint.closeHandler(v -> {
-            close(endpoint);
+            close(endpoint, authenticatedDevice);
             LOG.debug("connection to device [tenant-id: {}, device-id: {}] closed",
                     authenticatedDevice.getTenantId(), authenticatedDevice.getDeviceId());
             metrics.decrementMqttConnections(authenticatedDevice.getTenantId());
+            sendDisconnectedEvent(endpoint.clientIdentifier(), authenticatedDevice);
```
sophokles73: looks like you are sending the disconnected twice, once as part of *close()* and once explicitly.

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
```diff
@@ -0,0 +1,49 @@
...
+ * <p>
+ * The protocol adapters may call {@link #connected(String, Device)} and {@link #disconnected(String, Device)} as they
+ * see fit. The whole process is a "best effort" process and intended to be used for monitoring/debugging.
+ * <p>
+ * When a protocol adapter calls into this producer it must provide some kind of connection ID, which might have a
+ * different meaning for different protocol adapters. This should be as identifiable as possible but there is no
+ * requirement for this to be unique. However for each connection the protocol adapter must call
+ * {@link #disconnected(String, Device)} with the same information as it called {@link #connected(String, Device)}.
+ */
+public interface ConnectionEventProducer {
```
sophokles73: I don't see why we should introduce this new concept when we already have an event propagation mechanism. FMPOV we should simply send an event using the Event API downstream. It would be sufficient to simply define the event(s) and use what we already have.

sysexcontrol: I agree to @sophokles73 - sending a special purpose event downstream is what we discussed in the PR #562.
And it is what is now implemented in the PR #576 regarding the explicity sent *device notification*s - just using the Event API, but use it with a well defined event type, supported by a specific class.

I propose we do something similar here - define a class that is JSON mappable by jackson and send the payload via the Event API. 

After having implemented this for the devices, I would next go back to the specification and include these *protocol adpater connection events* as well. So then we have a stable specification that you could follow to integrate it for the MQTT protocol adapter.
Does that make sense?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
ctron: Fixed.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
ctron: Indeed. Fixed!

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
ctron: Well the idea was to provide some kind of pluggability here. Instead of directly hooking in the Event API allow different approaches here.

As I tried to explain in #565, capture the locations triggering the event, and then providing a way to implement the handling. The default handling could be what @sysexcontrol described, using the Event API to send a specialized event downstream.

Adding an interface and allowing an alternative implementation would allow consumers of Hono to route such events to a different address (e.g. Alerts) or disable them completely. If we directly invoke the Event API that won't be possible.

sysexcontrol: Regarding *pluggability* :+1: 

But since we implemented our microservices with the Spring framework, I would prefer to use what Spring offers for pluggability, i.e. declaring a `@Bean` and autowire it to the necessary classes.
This is as flexible as it can be: we autowire a default `@Bean` (using the Event API), may provide some more (like your logging only implementation) and can easily grow in the future.
WDYT?

ctron: I completely agree. But I think that is already present in the PR. I extended the `Config` class of the protocol adapter and used the `@Autowired` annotation.

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Produce an event for a new connection.
+     * 
+     * @param connectionIdentifier The ID of the connection which connected.
```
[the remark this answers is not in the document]
ctron: I changed this to "remoteId" and "The ID of the remote endpoint which connected (e.g. a remote address, port, client id, ...)."

Does that work better?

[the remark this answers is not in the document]
sysexcontrol: IMHO this is much better :+1: 
No more concerns :-)

sysexcontrol:
@ctron: Just as a note: in the last commit d73440b in the PR #576 I provided the constants that map to the specification PR #562.
So if you continue with the device connection notifications you can use them, and we have a unified implementation for both types of event.
Review comments of course very welcome!

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Produce an event for a closed connection.
+     * 
+     * @param remoteId The ID of the remote endpoint which disconnected. The same ID used in the call to
+     *            {@link #connected(String, Device)}
+     * @param authenticatedDevice The optional authenticated device associated with the connection. May be {@code null}.
+     */
+    public void disconnected(String remoteId, Device authenticatedDevice);
+
+}
```
sophokles73: missing EOL

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+public interface ConnectionEventProducer {
+
+    /**
+     * Produce an event for a new connection.
+     * 
+     * @param remoteId The ID of the remote endpoint which connected (e.g. a remote address, port, client id, ...).
+     * @param authenticatedDevice The optional authenticated device associated with the connection. May be {@code null}.
+     */
+    public void connected(String remoteId, Device authenticatedDevice);
```
sophokles73: can we also send a *connected*/*disconnected* event for devices that have not been authenticated?

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -639,4 +653,16 @@ protected final Message newMessage(
         addProperties(msg, registrationInfo);
         return msg;
     }
+
+    protected void sendConnectedEvent(final String connectionIdentifier, final Device authenticatedDevice) {
```
sophokles73: connectionIdentifier = remoteId ?

missing JavaDoc

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -639,4 +653,16 @@ protected final Message newMessage(
...
+        if (this.connectionEventProducer != null) {
+            this.connectionEventProducer.connected(connectionIdentifier, authenticatedDevice);
+        }
+    }
+
+    protected void sendDisconnectedEvent(final String connectionIdentifier, final Device authenticatedDevice) {
```
sophokles73: connectionIdentifier = remoteId ?

missing JavaDoc

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -171,6 +174,19 @@ public final HonoClient getCredentialsServiceClient() {
         return credentialsServiceClient;
     }
 
+    public void setConnectionEventProducer(final ConnectionEventProducer connectionEventProducer) {
```
sophokles73: missing JavaDoc

service-base/src/main/java/org/eclipse/hono/service/monitoring/LoggingConnectionEventProducer.java
```diff
@@ -0,0 +1,87 @@
...
+    public LoggingConnectionEventProducer() {
+        this(null);
+    }
+
+    /**
+     * Create a new instance with the provided log level.
+     * 
+     * @param level The provided log level to use, may be {@code null} in which case {@link Level#INFO} will be used.
+     */
+    public LoggingConnectionEventProducer(Level level) {
```
sophokles73: I thought we configured log levels by means of log config files.

ctron: True. We could drop this and let people configure this using the log system configuration. I would suggest to switch to "debug" in this case in order to not spam the log if people don't want this.

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
sophokles73: and how do we convey the device ID in that case?

ctron:
I did update the logger implementation to only log to "info".  I think that is ok as, in the case someone doesn't want to have connection events at all, unconfiguring this would probably be the right choice.

Is there any blocker for moving forward? As a next step I would like to implement the actual sending as @sysexcontrol suggested. But I would like to know if everything up to now is ok.

service-base/src/main/java/org/eclipse/hono/service/monitoring/LoggingConnectionEventProducer.java
```diff
@@ -12,76 +12,27 @@
...
-
 import org.eclipse.hono.service.auth.device.Device;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.slf4j.event.Level;
 
 /**
  * A logging only implementation.
+ * <p>
+ * This implementation only create log messages as "events". It will log everything on "info" level.
```
sophokles73: In most (if not all) other JavaDoc we use `<em>` to highlight or emphasize terms and use `@code` to mark literals instead of using quotes. Can you please adapt accordingly?

So, FMPOV thsi should read:
> This implementation only create log messages as <em>events</em>. It will log everything at {@code info} level.

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
```diff
@@ -0,0 +1,51 @@
...
+public interface ConnectionEventProducer {
+
+    /**
+     * Produce an event for a new connection.
+     * 
+     * @param remoteId The ID of the remote endpoint which connected (e.g. a remote address, port, client id, ...).
+     * @param authenticatedDevice The optional authenticated device associated with the connection. May be {@code null}
+     *            if the connection is from an unauthenticated device.
+     */
+    public void connected(String remoteId, Device authenticatedDevice);
```
sophokles73: no need to declare public visibility, this is an interface

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Produce an event for a closed connection.
+     * 
+     * @param remoteId The ID of the remote endpoint which disconnected. The same ID used in the call to
+     *            {@link #connected(String, Device)}
+     * @param authenticatedDevice The optional authenticated device associated with the connection. May be {@code null}
+     *            if the connection is from an unauthenticated device.
+     */
+    public void disconnected(String remoteId, Device authenticatedDevice);
```
sophokles73: no need to declare public visibility, this is an interface

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
sophokles73: what is the value of *remoteId* and *authenticatedDevice* in case of an device that has not been authenticated?

service-base/src/main/java/org/eclipse/hono/service/monitoring/LoggingConnectionEventProducer.java
ctron: Done

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
ctron: Done.

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
sophokles73: What value does this have then? You do not know which device this is because within Hono the device is identified, well, by the device-id (which is **not** the remoteId, is it?).

ctron: Well with MQTT CONNECT you would still present some ID. Exactly the remote id is not the device id. AMQP would still maybe present come container/link ID. Or we could use the remote endpoint address (IP+Port) or IP+Port+ClientID for MQTT.

The idea is to have some kind of protocol adapter specific ID, which is available even without having an authenticated Hono device. As this is for monitoring/debugging purposes, something which helps you find out "who this is".

sophokles73:
> As this is for monitoring/debugging purposes, something which helps you find out "who this is".

But it doesn't, does it? The original use case we were discussing in the context of this PR was: Kapua wants to see a green LED when a gateway is connected and wants to see a red LED if it is disconnected.

This use case will only work for authenticated devices/gateways anyway so this is probably an academic question. BTW we are already logging the fact that a device connects/disconnects in the MQTT adapter so tome this (currently) looks like a duplication of what we already do. I do see some value of firing an event in case of an authenticated device though.

ctron:
Well Kapua also wants to see if "something" has connected. Which is a functionality that is available right now. I would assume they don't want to loose too much functionality by switching the messaging layer. Right now you can see any inbound connection in Kapua.

I also agree that the current implementation is not and actual addition. That is why this PR is "first steps" and as I tried to lay out in this PR and #565, I would like to start with a logging implementation and then add the event implementation. Based on an interface, making this swappable by the use of the "Config" class, using Spring Boot bean wiring.

But I would like to understand if so far, you have any objections. I would hate to work on the next step of implementing the event based implementation just to later find out out that the start was already heading into the wrong direction.

sophokles73:
@ctron go ahead, I have stated my opinion but I do not have any strong concerns.

ctron:
I just pushed an implementation using the Hono Event API. I stil have to test it.

But there is the event generation using the proposed event format from @sysexcontrol. I only added the `remote-id` field in order to pass long that information.

The Hono Event implementation only processes authenticated devices. For others it isn't possible as no address would exist. Still an alternative implementation could make use of this information (e.g. using a special purpose, global address).

I also chained the actions into the connection process. So the MQTT connection is only accepted when the event was sent out. For the disconnect event this isn't possible, because the connection is already lost, so waiting for the processing of the disconnect event doesn't seem useful.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -301,17 +302,27 @@ final void handleEndpointConnection(final MqttEndpoint endpoint) {
         }
     }
 
-    private void handleConnectionRequestResult ( final MqttEndpoint endpoint, final AsyncResult<Device> done) {
+    private void handleConnectionRequestResult(final MqttEndpoint endpoint, final AsyncResult<Device> done) {
```
sophokles73: can we find another name for parameter *done*? the type of the async result suggests that this is the result of the attempt to authenticate the device?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -301,17 +302,27 @@ final void handleEndpointConnection(final MqttEndpoint endpoint) {
...
 
-        if ( done.succeeded() ) {
+        if (done.succeeded()) {
 
-            endpoint.accept(false);
-            sendConnectedEvent(endpoint.clientIdentifier(), done.result());
+            sendConnectedEvent(endpoint.clientIdentifier(), done.result())
+                    .setHandler(result -> {
+                        if (done.succeeded()) {
```
sophokles73: how can *done* not be succeeded at this place in the code?

service-base/src/main/java/org/eclipse/hono/service/monitoring/AbstractMessageSenderConnectionEventProducer.java
```diff
@@ -0,0 +1,112 @@
...
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Future;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * A connection event producer based on a {@link MessageSender}.
+ */
+public abstract class AbstractMessageSenderConnectionEventProducer implements ConnectionEventProducer {
+
+    private static final String MESSAGE_CONTENT_TYPE = "application/vnd.eclipse-hono-dc-notification+json";
```
sophokles73: IMHO we should define this constant in `EventConstants` if this is supposed to be a *well-known* event

service-base/src/main/java/org/eclipse/hono/service/monitoring/AbstractMessageSenderConnectionEventProducer.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private final HonoClient client;
+    private final BiFunction<HonoClient, String, Future<MessageSender>> messageSenderSource;
+
+    protected AbstractMessageSenderConnectionEventProducer(
```
sophokles73: missing JavaDoc

the HonoClient is most likely the client for publishing the event, right? so it should probably be named something like *eventServiceClient*?

service-base/src/main/java/org/eclipse/hono/service/monitoring/AbstractMessageSenderConnectionEventProducer.java
```diff
@@ -0,0 +1,112 @@
...
+
+        if (authenticatedDevice == null) {
+            // we only handle authenticated devices
+            return Future.succeededFuture();
+        }
+
+        // get an assertion
+
+        final Future<String> assertionFuture = this.client
+                .getOrCreateRegistrationClient(authenticatedDevice.getTenantId())
```
sophokles73: this will not work because the Device Registration service will in general not be co-hosted with the Event API implementation. in the default setup the Device Registration service is implemented by the Device Registry whereas the Event API is implemented by Hono Messaging or the Dispatch Router resp.

ctron:
I think all comments should be taken care of.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -301,17 +302,30 @@ final void handleEndpointConnection(final MqttEndpoint endpoint) {
...
+                    .setHandler(result -> {
+                        if (result.succeeded()) {
+                            endpoint.accept(false);
+                        } else {
+                            LOG.info(
+                                    "connection request from client [clientId: {}] rejected due to connection event "
+                                            + "failure: {}",
+                                    endpoint.clientIdentifier(),
+                                    MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE,
+                                    authenticationResult.cause());
```
sophokles73: IMHO this needs to be `result.cause()` to make sense.

In order to improve readability, I also propose to rename `authenticationResult` to `authenticationAttempt` and `result` to `sendAttempt`.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -301,17 +302,30 @@ final void handleEndpointConnection(final MqttEndpoint endpoint) {
...
+        if (authenticationResult.succeeded()) {
 
-            endpoint.accept(false);
-            sendConnectedEvent(endpoint.clientIdentifier(), done.result());
+            sendConnectedEvent(endpoint.clientIdentifier(), authenticationResult.result())
+                    .setHandler(result -> {
+                        if (result.succeeded()) {
+                            endpoint.accept(false);
+                        } else {
+                            LOG.info(
```
sophokles73: is this worth to be logged at INFO level? If this indicates a problem that needs to be taken care of then it should be WARNING, otherwise it should be DEBUG FMPOV.

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
```diff
@@ -14,38 +14,49 @@
...
 
     /**
      * Produce an event for a new connection.
      * 
      * @param remoteId The ID of the remote endpoint which connected (e.g. a remote address, port, client id, ...).
+     * @param protocolAdapter The name of the protocol adapter sending this event. Must not be {@code null}.
      * @param authenticatedDevice The optional authenticated device associated with the connection. May be {@code null}
      *            if the connection is from an unauthenticated device.
+     * @param data Additional, protocol adapter specific data
+     * @return A future which indicates the result of the event production
```
sophokles73: There seem to be some required params for which we usually add a `@throws NullPointerException` tag to make clear what to expect ...

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
```diff
@@ -14,38 +14,49 @@
...
      * Produce an event for a closed connection.
      * 
      * @param remoteId The ID of the remote endpoint which disconnected. The same ID used in the call to
-     *            {@link #connected(String, Device)}
+     *            {@link #connected(String, String, Device, JsonObject)}
+     * @param protocolAdapter The name of the protocol adapter sending this event. Must not be {@code null}.
      * @param authenticatedDevice The optional authenticated device associated with the connection. May be {@code null}
      *            if the connection is from an unauthenticated device.
+     * @param data Additional, protocol adapter specific data
+     * @return A future which indicates the result of the event production
```
sophokles73: There seem to be some required params for which we usually add a `@throws NullPointerException` tag to make clear what to expect ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -301,17 +302,30 @@ final void handleEndpointConnection(final MqttEndpoint endpoint) {
...
+                            endpoint.reject(MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE);
+                        }
+                    });
 
         } else {
 
-            final Throwable t = done.cause();
-            if ( t instanceof MqttConnectionException ) {
+            final Throwable t = authenticationAttempt.cause();
+            if (t instanceof MqttConnectionException) {
```
sophokles73: what if t is not an instance of `MqttConnectionException`?

ctron: Then it will go into the `else` branch, which will call:

~~~java
endpoint.reject(MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE);
~~~

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
ctron: Added.

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
ctron: Added.

ctron:
I looks like there are some javadoc issues now discovered by Travis. I will those tomorrow and merge the PR then.

Thanks for your work on this!

sophokles73:
yes, I have finally been able to activate the doclint checks for OpenJDK's javadoc tool ...

ctron:
I squashed the commits, fixed the javadoc issue and added a section in the release notes.

sophokles73:
It looks like none of the two event producers are wired up by default or am I missing something?
Is this by intention?

ctron:
For HTTP you are right. But the HTTP adapter doesn't send out connection events. And I don't think it would add much value for HTTP.

For MQTT it should be configured to "logging" by default: https://github.com/eclipse/hono/pull/569/files#diff-cc90988d9f6c0108b609f084cca69b9dR112

Unless I got something from with Spring, which might very well be the case.

sophokles73:
Oh, yes, you are right. I oversaw that. All is fine :+1:  @sysexcontrol has this week off so I think you should proceed with merging this PR without waiting for his review.

PR #574, opened 2018-04-12
Command and Control

pellmann:
This is a first implementation of the [Command and Control API](https://www.eclipse.org/hono/api/command-and-control-api/) from the client (Application) to the adapters (device to adapter is not included). It is not a complete solution and I would like to discuss my decisions. It allows the following case:

- The adapter opens a receiver link for a device (that connects) e.g. control/TENANT_1/device_A
- This link is between the adapter and a Dispatch Router
- The Application can now (typically triggered by an device-connected-event) open a sender link (over Dispatch Router) to the device and send a command. (If there is no receiver link open for the device, the application could not send the command, because there are no credits - so it is always a cheap operation to ‘test’ for connected)
- The adapter gets the command (typically forwards it to the device and waits for the result). It then opens a sender link to the replyTo address, which is then used to send the response back, which is received by the application.

Notes:

- I decided to open also a link for the device response for each device (vs. one link for the tenant). One reason is, that with a tenant scoped link it would be a problem to route the response to the sender if there will be more than one application instance. In this case either another instance could get the response (balanced) or all instances get the response (multicast). The latter with the known problems of back-pressure control (not possible) and scaling (all instances need to get the response). 
- Also I think it would be a better solution to have beside this real request-response command/control pattern a second, which only sends a command (without direct response) and additionally define an event type as a (possible) response, which could be handled like all events but with a special content-type (as we also discussed with the connected state events).
- The response (sender) link of the adapter (device) stays open as long as the receiver link is open. It could be closed after the response was sended. (Also the receiver link of the application could be closed after the response was sended.) This depends on the costs of link opening vs. cost of number of open links.  

Questions regarding the defined [Command and Control API](https://www.eclipse.org/hono/api/command-and-control-api/):

- I would remove the ‘durable’ flag in the API.
- Instead of a specific ‘command’ application property I use the subject.
- Should we really set a default time for ‘ttl’? I opt for no default (so the user need to think about his use case) but a max. value might be useful.


sophokles73:
> Also I think it would be a better solution to have beside this real request-response command/control pattern a second, which only sends a command (without direct response) and additionally define an event type as a (possible) response, which could be handled like all events but with a special content-type (as we also discussed with the connected state events).

I think I don't get this. Are you talking about a *void* command/operation that doesn't expect any data to be returned as the result? FMPOV this would simply be a special case of the generic mechanism in that e.g. the protocol adapter returns the (empty) response message to the application via the reply-to address. I do not understand what the *response event* is supposed to be. Can you explain what you intend it to be used for?

sophokles73:
> I would remove the ‘durable’ flag in the API.

+1

> Instead of a specific ‘command’ application property I use the subject.

+1

> Should we really set a default time for ‘ttl’? I opt for no default (so the user need to think about his use case) but a max. value might be useful.

We would then make the TTL property mandatory, right? However, given that we do not *buffer* command request messages, why would we need this property in the first place? If the adapter cannot send the command to the device, it simply returns a response, indicating the failure, doesn't it?

pellmann:
> I think I don't get this. Are you talking about a void command/operation that doesn't expect any data to be returned as the result? FMPOV this would simply be a special case of the generic mechanism in that e.g. the protocol adapter returns the (empty) response message to the application via the reply-to address. I do not understand what the response event is supposed to be. Can you explain what you intend it to be used for?

I thought of two operations. One, which expects a direct response (with the replyTo address) and a second, which does not (Same semantic as our events in the other direction). Within this type of command one might expect a response sometimes in the future. e.g. the next time, the device goes online, which might be a day in the future. This response is now send with a normal event, but with a field, that marks this event as the result and allow a correlation to the command. So for a lot of command use-cases this might be a better solution and we do not have reply links for each device. 

I think something like light=on would be handled with the first type and something like set=newFirmWareWithReboot would be the second.

> We would then make the TTL property mandatory, right? However, given that we do not buffer command request messages, why would we need this property in the first place? If the adapter cannot send the command to the device, it simply returns a response, indicating the failure, doesn't it?

We do not think of buffering right now, but a user could configure his 'AMQP Network' to do such and it might be a good idea for some use cases. But I think to keep it simple it might be better to also remove the AMQP ttl in the beginning.

sophokles73:
> We do not think of buffering right now, but a user could configure his 'AMQP Network' to do such and it might be a good idea for some use cases. But I think to keep it simple it might be better to also remove the AMQP ttl in the beginning.

I agree, and a user can still set a ttl (which is a standard AMQP 1.0 property) if his AMQP Network supports it, even if Hono itself doesn't care about it.

sophokles73:
> This response is now send with a normal event, but with a field, that marks this event as the result and allow a correlation to the command. So for a lot of command use-cases this might be a better solution and we do not have reply links for each device.

Wouldn't we then face the same problem as with using a tenant scoped link for all responses? How do we make sure that the event with the response is consumed by the application (instance) that has issued the command?

client/src/main/java/org/eclipse/hono/client/CommandClient.java
```diff
@@ -0,0 +1,53 @@
...
+     * @return A future indicating the result of the operation.
+     *         <p>
+     *         The future will succeed if a response with status 201 has been received from the device/adapter.
+     *         <p>
+     *         Otherwise, the future will fail with a {@link ServiceInvocationException} containing
+     *         the (error) status code returned by the service.
+     * @throws NullPointerException if device ID is {@code null}.
+     * @see RequestResponseClient#setRequestTimeout(long)
+     */
+    Future<byte[]> commandWithResponse(String command, byte[] data);
```
ctron: Could we think of using `ByteBuffer`? That could prevent a few copies, and we could still add `byte[]` using default methods for simplicity.

sophokles73: or vert.x `Buffer`?

pellmann:
> Wouldn't we then face the same problem as with using a tenant scoped link for all responses? How do we make sure that the event with the response is consumed by the application (instance) that has issued the command?

We would face this problem but I see this in a clearly different use case, now. In my above example, were a firmware update will be rolled out, I would imagine, that the application needs a persistence to remember which system (device) has the new firmware. So in such scenarios it does not matter which instance gets the response. The application may assume, that in a timeframe of 2 days there need to be a response and if it is not there it could resend the update (depending on the state in the persistence).

sysexcontrol:
> Are you talking about a void command/operation that doesn't expect any data to be returned as the result? FMPOV this would simply be a special case of the generic mechanism in that e.g. the protocol adapter returns the (empty) response message to the application via the reply-to address.

AFAIK it is quite common for some protocols (LoRa-WAN e.g.) that a command will *never* be replied by the device, not in a short time frame nor in a longer time frame.
Since the applications should somehow know with what devices they are dealing, a pattern that never gets *any* response makes sense to me. For these devices we would not even have a reply-to link, since there won't be any valuable data anyway.

client/src/main/java/org/eclipse/hono/client/CommandClient.java
pellmann: Since internally we use the vert.x Buffer and the Buffer implementation uses netty ByteBuf internally (and not ByteBuffer), the use of Buffer seems better. But on the other hand the user need to use the Vert.x objects (or at least the netty ByteBuf) to prevent copies.

sophokles73:
@pellmann is this PR still relevant?

PR #575, opened 2018-04-12
Drop special PV handling for minishift

ctron:
This drops the special handling for the PV/PVC and uses the (newly added) Minishift pre-created PVs.


sophokles73:
You should also remove the part from the OpenShift deployment guide that refers to manually creating the filesystem path for the PV ...

sophokles73:
This also seems to require a certain minimum version of OpenShift supported, right?

Issue #536, opened 2018-03-16
Implement device connection notifications


ctron:
This is a sub-task of #506

After defining the API for this in #535 it needs to be implemented.

sophokles73:
@ctron @sysexcontrol  is this still relevant?

ctron:
Yes, as long as PR #569 is still open.

PR #576, opened 2018-04-12
Hono device notifications impl

sysexcontrol:
This PR implements device notifications for signalling the readiness to receive a message, as intended to be sent by devices themselves. It thus implements part of [#536].

The notification is sent via the Event API with a specific content-type, and has an own object that is mapped by jackson.
The northbound applications are supported by a multiplexer class that invokes the provided specialized callbacks in the application, so that the application itself is freed from parsing event messages itself to try to find this type of event.
The sending of such a notification is shown in an example device class that sends it via the HTTP adapter, the application example is extended to show the usage of the multiplexer.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/EventDemultiplexer.java
+++ b/core/src/main/java/org/eclipse/hono/util/EventDemultiplexer.java
@@ -1,0 +1,61 @@
+/**
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 1.0 which is available at
+ * https://www.eclipse.org/legal/epl-v10.html
+ *
+ * SPDX-License-Identifier: EPL-1.0
+ */
+
+package org.eclipse.hono.util;
+
+import org.apache.qpid.proton.amqp.messaging.Data;
+import org.apache.qpid.proton.amqp.messaging.Section;
+import org.apache.qpid.proton.message.Message;
+
+import java.util.Optional;
+import java.util.function.Consumer;
+
+/**
+ * A helper for handling different types of events with explicit callbacks.
+ * Currently it demultiplexes callbacks for generic events and notifications (being well defined events).
+ */
+public final class EventDemultiplexer {
+    /**
+     * Prevent construction.
+     */
+    private EventDemultiplexer() {
+    }
+
+
+    /**
+     * Construct a event consumer that delegates different kinds of events to specific handlers.
+     * @param eventConsumer Consumer that handles generic events.
+     * @param notificationConsumer Consumer that handles events that were detected to be a notification. Maybe {@code null}.
+     * @return The event consumer that invokes the specific handlers depending on the event type.
+     */
+    public static final Consumer<Message> createEventHandler(final Consumer<Message> eventConsumer,
+                                                             final Consumer<Message> notificationConsumer
+                                                             ) {
+        return (msg -> {
+            final Section body = msg.getBody();
+            if (!(body instanceof Data)) {
+                return;
+            }
+
+            Optional.ofNullable(notificationConsumer).map(consumer -> {
+                if (NotificationConstants.isNotification(msg)) {
+                    notificationConsumer.accept(msg);
+                } else {
+                    eventConsumer.accept(msg);
+                }
+                return consumer;
+            }).orElse(consumer -> eventConsumer.accept(msg));
+        });
+    }
+
+}
--- a/core/src/main/java/org/eclipse/hono/util/NotificationConstants.java
+++ b/core/src/main/java/org/eclipse/hono/util/NotificationConstants.java
@@ -1,0 +1,89 @@
+/**
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 1.0 which is available at
+ * https://www.eclipse.org/legal/epl-v10.html
+ *
+ * SPDX-License-Identifier: EPL-1.0
+ */
+
+package org.eclipse.hono.util;
+
+import org.apache.qpid.proton.message.Message;
+
+import java.util.Arrays;
+import java.util.Objects;
+
+/**
+ * Constants used for notification events.
+ */
+public abstract class NotificationConstants {
+
+    /**
+     * The content type that is defined for device command readiness notification events.
+     */
+    public static final String CONTENT_TYPE_DEVICE_COMMAND_READINESS_NOTIFICATION = "application/vnd.eclipse-hono-dcr-notification+json";
+
+    /**
+     * The content type that is defined for device connection notification events.
+     */
+    public static final String CONTENT_TYPE_DEVICE_CONNECTION_NOTIFICATION = "application/vnd.eclipse-hono-dc-notification+json";
+
+    /**
+     * Enum that defines all valid notifications by their content-type.
+     */
+    public enum NotificationContentType {
+        DEVICE_COMMAND_READINESS_NOTIFICATION(CONTENT_TYPE_DEVICE_COMMAND_READINESS_NOTIFICATION),
+        DEVICE_CONNECTION_NOTIFICATION(CONTENT_TYPE_DEVICE_CONNECTION_NOTIFICATION);
+
+        private final String contentType;
+
+        NotificationContentType(final String contentType) {
+            this.contentType = contentType;
+        }
+
+        /**
+         * Helper method to check if a subject is a valid content type.
+         *
+         * @param subject The subject to validate.
+         * @return boolean {@code true} if the subject denotes a valid action, {@code false} otherwise.
+         */
+        public static boolean isValid(final String subject) {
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/eclipse/hono/util/HonoDeviceNotificationConstants.java
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * Prevent construction.
+     */
+    private HonoDeviceNotificationConstants() {
+    }
+
+    /**
+     * The content type that is defined for device notification events.
+     */
+    public static final String CONTENT_TYPE_DEVICE_NOTIFICATION = "application/vnd.eclipse-hono-notification+json";
```
sophokles73: the whole idea of using a content type to identify the type of event was based on the assumption, that it is not necessary to introspect (or worse, try to deserialize a payload using Jackson) an event in order to determine if a consumer needs to process it.

Using a generic *notification* type undermines this idea FMPOV. The event for indicating that a device is now ready to receive a command should therefore have its own content-type FMPOV because it is of such importance, will probably be used very frequently and is relevant to almost all consuming applications.

core/src/main/java/org/eclipse/hono/util/HonoDeviceNotifications.java
```diff
@@ -0,0 +1,55 @@
...
+
+package org.eclipse.hono.util;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * A helper for creating device notification events.
+ */
+public final class HonoDeviceNotifications {
```
sophokles73: I do not understand the purpose of this class.
Given that you already created a class representing a notification, the factory method for creating it should probably be part of that class, shouldn't it?

core/src/main/java/org/eclipse/hono/util/HonoEventMultiplexer.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.function.Consumer;
+
+/**
+ * A helper for handling different types of events with explicit callbacks.
+ *
+ * It multiplexes callbacks for generic events, notifications (being well defined events), and specific notifications
+ * that signal the command readiness of a device. See Hono's
+ * <a href="https://www.eclipse.org/hono/api/event-api/">Event API</a> for details.
+ */
+public final class HonoEventMultiplexer {
```
sophokles73: if at all, this seems to be a *Demultiplexer*.

core/src/main/java/org/eclipse/hono/util/HonoEventMultiplexer.java
```diff
@@ -0,0 +1,143 @@
...
+                });
+
+                Optional.ofNullable(deviceReadyForCommandConsumer).map(consumer -> {
+                    // investigate if we need to call device ready for command consumer
+
+                    if (HonoDeviceNotificationConstants.CONTENT_TYPE_DEVICE_NOTIFICATION.equals(msg.getContentType())) {
+                        Optional.ofNullable(((Data)msg.getBody()).getValue()).
+                                map(data -> {
+                                    try {
+                                        return OBJECT_MAPPER.readValue(data.toString(), NotificationDeviceCommandReadyObject.class);
```
sophokles73: in order to find out if this is an event indicating that a device is ready to receive a command, we need to try to deserialize it?

core/src/main/java/org/eclipse/hono/util/HonoEventMultiplexer.java
```diff
@@ -0,0 +1,143 @@
...
+     * device for which it was received should be ready to receive a command.
+     * This is evaluated at the point in time this method is invoked.
+     *
+     * @param deviceNotification JsonObject that is evaluated.
+     * @param creationTime Time in milliseconds when the notification was created.
+     * @return Boolean {@code true} if the event signals that the device now should be ready to receive a command, {@code false}
+     * otherwise.
+     * @throws NullPointerException If deviceNotification is {@code null}.
+     */
+    private static final Boolean isDeviceCurrentlyReadyForCommands(final NotificationDeviceCommandReadyObject deviceNotification,
```
sophokles73: This belongs into `NotificationDeviceCommandReadyObject`

core/src/main/java/org/eclipse/hono/util/NotificationDeviceCommandReadyObject.java
```diff
@@ -0,0 +1,197 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+
+/**
+ * Encapsulates the device command readiness notification that is defined in Hono's
+ * <a href="https://www.eclipse.org/hono/api/event-api/">Event API</a>.
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public final class NotificationDeviceCommandReadyObject {
```
sophokles73: this looks more like the representation of a generic *Notification*.

The only data relevant for indicating that a device is ready to receive a command seems to be *timeToBeReady* (if I understand its purpose correctly in that it indicates the number of seconds the device will be *online*).

core/src/main/java/org/eclipse/hono/util/HonoDeviceNotificationConstants.java
```diff
@@ -0,0 +1,97 @@
...
+         */
+        public static boolean isValid(final String subject) {
+            return Arrays.stream(NotificationContentType.values()).filter(type -> type.contentType.equals(subject)).findFirst().isPresent();
+        }
+    }
+
+    /**
+     * JSON field name for specifying the source of the event.
+     */
+    public static final String FIELD_SOURCE = "src";
```
sophokles73: these constants seem to be only relevant to `NotificationDeviceCommandReadyObject` so we should probably move them there.

core/src/main/java/org/eclipse/hono/util/HonoDeviceNotificationConstants.java
sysexcontrol: > the whole idea of using a content type to identify the type of event was based on the assumption, that it is not necessary to introspect (or worse, try to deserialize a payload using Jackson)

Entirely true - but that is what has been implemented here (?) Only looking at the content-type, and if
- the content-type is a notification-type, and 
- if the helper class (HonoDemultiplexer) is used,

then (and only then)

- the notification object is built by using jackson.

If no helper class is used, the event is (as always) available in the eventHandler callback (no jackson involved).

core/src/main/java/org/eclipse/hono/util/HonoEventMultiplexer.java
sysexcontrol: Yes and No:
- if the event is matching at all is done by checking the content-type.
- if a matching event is valid regarding the contained data is done by looking into the JSON payload, which is done here by deserializing it. Alternative: parse it manually.

To make discussion easier, I will split this PR into
- one that contains only the definitions necessary for these notifications
- one that offers the object based functionality, so not every application needs to implement the parsing of a notification event itself.

I guess this will make the functionality contained here clearer, and we can have a quicker agreement regarding the definitions of the notification definitions.

core/src/main/java/org/eclipse/hono/util/EventDemultiplexer.java
```diff
@@ -0,0 +1,61 @@
...
+
+
+    /**
+     * Construct a event consumer that delegates different kinds of events to specific handlers.
+     * @param eventConsumer Consumer that handles generic events.
+     * @param notificationConsumer Consumer that handles events that were detected to be a notification. Maybe {@code null}.
+     * @return The event consumer that invokes the specific handlers depending on the event type.
+     */
+    public static final Consumer<Message> createEventHandler(final Consumer<Message> eventConsumer,
+                                                             final Consumer<Message> notificationConsumer
```
ctron: Wouldn't it be an idea to put in a Map here and map from Subject to consumer?

sysexcontrol: @ctron: thanks for the idea!
But I am afraid I do not understand what you mean by it - currently this class just demultiplexes based on the message, and the consumers must be provided by the user of this class (as lambdas, as method references, etc.).

Where could a `Map` improve things?

ctron: So currently the demuxing is done based on the result of "getContentType", which is a String IIRC.

So instead of using a single `Consumer<Message>` as the second argument, we could use a `Map<String,Consumer<Message>>` and use the result of `getContentType` as key. Something like:

~~~
String type = message.getContentType();
Consumer<Message> specialConsumer = consumerMap.get(type);
if ( specialConsumer != null ) {
  specialConsumer.accept ( message );
} else {
  defaultConsumer.accept ( message );
~~~

This would allow to use the demuxer with additional content types without the need to add new arguments in the "createEventHandler" all the time.

sophokles73:
@sysexcontrol  what is the status of this PR?

core/src/main/java/org/eclipse/hono/util/EventDemultiplexer.java
sysexcontrol: @ctron With the current specification, there is no "device command readiness notification" based on a content-type anymore.
So the good idea currently does not make so much sense anymore.
See PR #596 discussions for details, and thanks again for the comment!

sysexcontrol:
Since this PR is now obsolete, it is closed.
More specific implementation PRs exist:
#596 and #569

Issue #577, opened 2018-04-13
ConnectionFactoryImpl does not handle corner case when underlying connection gets closed before client received open frame


DanielMaier-BSI:
In https://github.com/eclipse/hono/blob/8320a46d3c93e76f192f53f92c2d26d3657e8c93/core/src/main/java/org/eclipse/hono/connection/ConnectionFactoryImpl.java#L138-L181
it can happen that underlying connection gets closed, e.g. AMQP server closes connection, before we receive open frame of server. As no disconnectHandler is set before we got open frame, connectionResultHandler gets never called.

I will also provide a PR for this.

DanielMaier-BSI:
Fixed with #579

PR #579, opened 2018-04-16
Fix for #577 - ConnectionFactoryImpl does not handle corner case when…

DanielMaier-BSI:
… underlying connection gets closed before client received open frame

Added disconnect handler to downstreamConnection to be aware when underlying AMQP connection gets closed after SASL handshake and sending AMQP open frame. If connection gets closed call connectionResultHandler.

If you do not like the test case, e.g. because it starts a real AMQP server and can last some time (500 msec on my machine), I can replace it with a test that mocks ProtonClient. But this test would actually only test if disconnect handler was set on connection.

### DIFF
--- a/core/src/test/java/org/eclipse/hono/connection/ConnectionFactoryImplTest.java
+++ b/core/src/test/java/org/eclipse/hono/connection/ConnectionFactoryImplTest.java
@@ -14,6 +14,9 @@ package org.eclipse.hono.connection;
 import static org.junit.Assert.*;
 import static org.mockito.Mockito.*;
 
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
+import io.vertx.proton.ProtonConnection;
 import org.eclipse.hono.config.ClientConfigProperties;
 import org.junit.Before;
 import org.junit.Test;
@@ -27,6 +30,7 @@ import io.vertx.ext.unit.TestContext;
 import io.vertx.ext.unit.junit.VertxUnitRunner;
 import io.vertx.proton.ProtonClient;
 import io.vertx.proton.ProtonClientOptions;
+import org.mockito.Mockito;
 
 /**
  * Verifies behavior of {@code ConnectionFactoryImpl}.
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/eclipse/hono/connection/ConnectionFactoryImplTest.java
```diff
@@ -73,6 +74,43 @@ public void testConnectInvokesHandlerOnfailureToConnect(final TestContext ctx) {
...
     }
 
+    /**
+     * Verifies that the given result handler is invoked if a connection gets closed after SASL auth was successful and
+     * AMQP open frame was sent by client, but no AMQP open frame was send by server yet.
+     *
+     * @param ctx The vert.x test context.
+     */
+    @Test
+    public void testConnectInvokesHandlerOnDisconnectAfterSendingOpenFrame(final TestContext ctx) {
```
sophokles73: Indeed, I would rather have a test case using a mocked connection...

DanielMaier-BSI: Ok, done.

core/src/test/java/org/eclipse/hono/connection/ConnectionFactoryImplTest.java
```diff
@@ -73,6 +77,37 @@ public void testConnectInvokesHandlerOnfailureToConnect(final TestContext ctx) {
...
+    /**
+     * Verifies that the given result handler is invoked if a connection gets closed after SASL auth was successful and
+     * AMQP open frame was sent by client, but no AMQP open frame from server was received yet.
+     */
+    @SuppressWarnings("unchecked")
+    @Test
+    public void testConnectInvokesHandlerOnDisconnectAfterSendingOpenFrame(final TestContext ctx) {
+        // GIVEN a factory configured to connect to a server (with mocked connection)
+        ConnectionFactoryImpl factory = new ConnectionFactoryImpl(vertx, props);
+        ProtonClient protonClientMock = mock(ProtonClient.class);
```
sophokles73: can you please make the declared variables final

DanielMaier-BSI: done :)

Issue #526, opened 2018-03-06
Add support for publishing data with QoS 1 via HTTP adapter.


sophokles73:
HTTP does not per se support the concept of QoS but the HTTP adapter could support a request header indicating the QoS that the device wants to use. The adapter would then return the response once it has received the disposition for the message from the downstream container and set the response's status code accordingly.

PR #581, opened 2018-04-16
[Http Adapter] Add support for publishing data with QoS 1

Alfusainey:
Pull request for issue #526

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -56,7 +56,6 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
 
     private static final Logger LOG = LoggerFactory.getLogger(AbstractVertxBasedHttpProtocolAdapter.class);
 
-    private static final String HEADER_QOS_LEVEL = "QoS-Level";
     private static final int AT_LEAST_ONCE = 1;
     private static final int HEADER_QOS_INVALID = -1;
 
@@ -482,12 +481,12 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
     private void doUploadMessage(final RoutingContext ctx, final String tenant, final String deviceId,
             final Buffer payload, final String contentType, final Future<MessageSender> senderTracker, final String endpointName) {
 
-        final Integer qosHeader = getQoSLevel(ctx.request().getHeader(HEADER_QOS_LEVEL));
+        final Integer qosHeader = getQoSLevel(ctx.request().getHeader(Constants.HEADER_QOS_LEVEL));
         if (contentType == null) {
             HttpUtils.badRequest(ctx, String.format("%s header is missing", HttpHeaders.CONTENT_TYPE));
         } else if (payload == null || payload.length() == 0) {
             HttpUtils.badRequest(ctx, "missing body");
-        } else if (qosHeader != null && (qosHeader == HEADER_QOS_INVALID || qosHeader != AT_LEAST_ONCE)) {
+        } else if (qosHeader != null && qosHeader == HEADER_QOS_INVALID) {
             HttpUtils.badRequest(ctx, "Bad QoS Header Value");
         } else {
 
@@ -541,10 +540,11 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
 
     private static Integer getQoSLevel(final String qosValue) {
         try {
-            if (qosValue == null)
+            if (qosValue == null) {
                 return null;
-            else
-                return Integer.parseInt(qosValue);
+            } else {
+                return Integer.parseInt(qosValue) != AT_LEAST_ONCE ? HEADER_QOS_INVALID : AT_LEAST_ONCE;
+            }
         } catch (NumberFormatException e) {
             return HEADER_QOS_INVALID;
         }
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -56,6 +56,10 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
 
     private static final Logger LOG = LoggerFactory.getLogger(AbstractVertxBasedHttpProtocolAdapter.class);
 
+    private static final String HEADER_QOS_LEVEL = "QoS-Level";
+    private static final int AT_LEAST_ONCE = 1;
+    private static final int HEADER_QOS_INVALID = -1;
+
     private HttpServer         server;
     private HttpServer         insecureServer;
     private HttpAdapterMetrics metrics;
@@ -478,10 +482,13 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
     private void doUploadMessage(final RoutingContext ctx, final String tenant, final String deviceId,
             final Buffer payload, final String contentType, final Future<MessageSender> senderTracker, final String endpointName) {
 
+        final Integer qosHeader = getQoSLevel(ctx.request().getHeader(HEADER_QOS_LEVEL));
         if (contentType == null) {
             HttpUtils.badRequest(ctx, String.format("%s header is missing", HttpHeaders.CONTENT_TYPE));
         } else if (payload == null || payload.length() == 0) {
             HttpUtils.badRequest(ctx, "missing body");
+        } else if (qosHeader != null && (qosHeader == HEADER_QOS_INVALID || qosHeader != AT_LEAST_ONCE)) {
+            HttpUtils.badRequest(ctx, "Bad QoS Header Value");
         } else {
 
             final Device authenticatedDevice = getAuthenticatedDevice(ctx);
@@ -491,6 +498,7 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
             CompositeFuture.all(tokenTracker, tenantConfigTracker, senderTracker).compose(ok -> {
 
                 if (tenantConfigTracker.result().isAdapterEnabled(getTypeName())) {
+                    final MessageSender sender = senderTracker.result();
                     final Message downstreamMessage = newMessage(
                             String.format("%s/%s", endpointName, tenant),
                             deviceId,
@@ -499,7 +507,11 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
                             payload,
                             tokenTracker.result());
                     customizeDownstreamMessage(downstreamMessage, ctx);
-                    return senderTracker.result().send(downstreamMessage);
+                    if (qosHeader == null) {
+                        return sender.send(downstreamMessage);
+                    } else {
+                        return sender.sendAndWaitForOutcome(downstreamMessage);
+                    }
                 } else {
                     // this adapter is not enabled for the tenant
                     return Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_FORBIDDEN));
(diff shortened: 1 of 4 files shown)


sophokles73:
Your PR has pointed me to a problem we had with the MQTT adapter supporting QoS 1 telemetry messages (#583). Can you please rebase your PR and adapt your code according to the changes I made in the MQTT adapter?

Alfusainey:
yes!!! i was really wondering how that was suppose to work before your changes. thanks!
i made the requested modifications.

sophokles73:
you seem to have forgotten to *sign-off* your commit. Please see https://github.com/eclipse/hono/blob/master/CONTRIBUTING.md for details on how to do that.

Alfusainey:
thanks. i just did

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -478,10 +481,13 @@ public final void uploadEventMessage(final RoutingContext ctx, final String tena
     private void doUploadMessage(final RoutingContext ctx, final String tenant, final String deviceId,
             final Buffer payload, final String contentType, final Future<MessageSender> senderTracker, final String endpointName) {
 
+        final Integer qosHeader = ctx.request().getHeader(HEADER_QOS_LEVEL) == null ? null : Integer.parseInt(ctx.request().getHeader(HEADER_QOS_LEVEL));
         if (contentType == null) {
             HttpUtils.badRequest(ctx, String.format("%s header is missing", HttpHeaders.CONTENT_TYPE));
         } else if (payload == null || payload.length() == 0) {
             HttpUtils.badRequest(ctx, "missing body");
+        } else if (qosHeader != null && qosHeader != AT_LEAST_ONCE) {
+            HttpUtils.notImplemented(ctx, "only QoS 1 is currently supported");
```
sophokles73: this should result in a bad request since it is not a server error.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -478,10 +481,13 @@ public final void uploadEventMessage(final RoutingContext ctx, final String tena
     private void doUploadMessage(final RoutingContext ctx, final String tenant, final String deviceId,
             final Buffer payload, final String contentType, final Future<MessageSender> senderTracker, final String endpointName) {
 
+        final Integer qosHeader = ctx.request().getHeader(HEADER_QOS_LEVEL) == null ? null : Integer.parseInt(ctx.request().getHeader(HEADER_QOS_LEVEL));
```
sophokles73: please use our code formatter rules to be found under `/eclipse` ...
this line seems to be too long

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
Alfusainey: ok, i was thinking there could be support for QoS 2 for telemetry in future. will fix

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
Alfusainey: thanks

sysexcontrol: `Integer.parseInt(...)`
And what happens if the header is not containing an integer value that can be parsed?

Can you please add some handling for this? Otherwise this would generate lots of ugly exceptions inside the adapter when a client misbehaves...

Alfusainey: sure, thanks

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -539,4 +538,15 @@ private void doUploadMessage(final RoutingContext ctx, final String tenant, fina
             });
         }
     }
+
+    private static Integer getQoSLevel(String qosValue) {
+        try {
+            if (qosValue == null)
```
sysexcontrol: check for null not necessary here, it will be transfered to a `NumberFormatException` by `parseInt`.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -539,4 +538,15 @@ private void doUploadMessage(final RoutingContext ctx, final String tenant, fina
             });
         }
     }
+
+    private static Integer getQoSLevel(String qosValue) {
```
sysexcontrol: Please make the param `final`.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
Alfusainey: Sure, but this handles the case when the qos header value is not set by the client. That should be differentiated from an invalid qos header value

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -526,4 +538,15 @@ private void doUploadMessage(final RoutingContext ctx, final String tenant, fina
             });
         }
     }
+
+    private static Integer getQoSLevel(final String qosValue) {
+        try {
+            if (qosValue == null)
```
sophokles73: please use curly braces for all statements
again, I advise to use the code formatter rules provided in the `/eclipse` folder

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -526,4 +538,15 @@ private void doUploadMessage(final RoutingContext ctx, final String tenant, fina
...
+                return null;
+            else
+                return Integer.parseInt(qosValue);
+        } catch (NumberFormatException e) {
+            return HEADER_QOS_INVALID;
```
sophokles73: introducing the `INVALID` is a good idea, you should also use it for all values other than the allowed one (`AT_LEAST_ONCE`). You can then simplify the check above ...

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
sysexcontrol: Ok, I see :+1:

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
Alfusainey: sure, i am using the code formatter since the last feedback. for some reason it feels to spot that

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
Alfusainey: @sophokles73: please note that if the qos header is not set (like in all the existing test cases before this patch), then `ctx.request().getHeader(QOS_HEADER)` will be `null`. imo this is not an invalid header and is why am not using `INVALID` when the header is not set. however, when the header is set but cannot be parsed, then its `INVALID`.

what do you think?

sophokles73: I understand, my point is that you can use `INVALID` for both a value that is not a number at all as well as for any number != 1.

Alfusainey: ok, i see. thanks

Alfusainey: how about this approach (before i update the PR)?:
```java
try {
       if (qosValue == null) {
          return null;
       }
       else {
          return Integer.parseInt(qosValue) != AT_LEAST_ONCE ? HEADER_QOS_INVALID : AT_LEAST_ONCE;
       }
} catch (NumberFormatException e) {
      return HEADER_QOS_INVALID;
}
```

sophokles73: that should do the trick. but please format like this
```
try {
       if (qosValue == null) {
          return null;
       } else {
          return Integer.parseInt(qosValue) != AT_LEAST_ONCE ? HEADER_QOS_INVALID : AT_LEAST_ONCE;
       }
} catch (NumberFormatException e) {
      return HEADER_QOS_INVALID;
}
```

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -56,6 +56,10 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(AbstractVertxBasedHttpProtocolAdapter.class);
 
+    private static final String HEADER_QOS_LEVEL = "QoS-Level";
```
sophokles73: I think we should move this to `org.eclipse.hono.util.Constants` so that client code can easily refer to it as well.

Don't forget to update your test cases accordingly ...

Alfusainey: sure, thanks! looking at the copyright section of `notice.md`, i did not see individual names added there. how does this work with students?

sophokles73: just add your name if you are not doing the work in the context of an organization

core/src/main/java/org/eclipse/hono/util/Constants.java
```diff
@@ -51,6 +51,11 @@
      */
     public static final String PROTOCOL_ADAPTER_TYPE_KURA= "hono-kura-mqtt";
 
+    /**
+     * The "QoS-Level" request header.
```
sophokles73: It seems obvious that this is the `QoS-Level` header. The name of the constant already implies it. The JavaDoc is supposed to contain a (short) description of what it is good/used for.

legal/legal/NOTICE.md
```diff
@@ -15,6 +15,7 @@ of the Eclipse Public License 1.0 which is available at https://www.eclipse.org/
 
 * Copyright 2016-2018 Bosch Software Innovations GmbH
 * Copyright 2016-2018 Red Hat, Inc.
+* [Alfusainey Jallow] (https://github.com/Alfusainey/)
```
sophokles73: please use the same format as the entries above, i.e. add the copyright notice.  you may still add the link to your GitHub profile, though.

Alfusainey: thanks. then how about `Copyright 2018 [Alfusainey Jallow] (https://github.com/Alfusainey/)`? or should it include an end date?

Alfusainey:
sure, thanks! just did that

sophokles73:
@Alfusainey  thanks for contributing :+1: 

now it would be nice if you could describe this new feature in the HTTP adapter's user guide, analogously to how it has been done in the MQTT adapter's user guide.

and don't forget to add a line to the release notes as well :-)

Alfusainey:
sure! i will create another PR for the user guide (and update the release notes)
thanks a lot!

PR #584, opened 2018-04-17
Allow devices to indicate readiness-to-receive-messages (spec).

sysexcontrol:
Contained is the full specification and concept description of the mechanisms that allow
devices to indicate their readiness to receive messages.
No implementation.


sysexcontrol:
Hint: while reworking the piggy-back variant (using a telemetry message e.g. to indicate the readiness to receive messages) it became clear that this is not possible by the device itself in all cases.
MQTT e.g. does not allow us to set any headers, so there is no way to attach the `stay_connected` value to a telemetry message.
In the specification it is now written that the protocol adapters may attach this property to the AMQP 1.0 message - but do you think this is a good idea? WDYT?

I am not certain about that, so please especially have a look at this part, thanks!

pellmann:
> MQTT e.g. does not allow us to set any headers

Some time ago we talked about property bags for MQTT, like Azure do it. I think this might be the first case where we really need it. In MQTT 5 properties are in the protocol.

pellmann:
Would it not be better to have the stay_connected (or keep_alive?) property always as an AMQP application property? So it does not matter, that there is the special event or just some property as piggy-back. This could be also true for the other (needed) fields. 

So the event data would be always free for the application event, if piggy-bagged or whatever the user wants to put in a 'special' event. Maybe this would be also better binary data for him. 

In contrast to this advantage I do not see the advantage of a JSON payload at the moment?

sysexcontrol:
> Would it not be better to have the stay_connected (or keep_alive?) property always as an AMQP application property? 

Actually I had specified it already like this. And then created the sequence diagrams where it became obvious that for MQTT (as it is now, without property bags) there is no possibility to put the `stay_connected` property into the message. So I have put it into the payload - the only thing (besides the content-type) that we have for current MQTT IMHO. Or do I miss something here?

For protocols like HTTP we could define specific headers to control this, however. To get a simple specification ready I have not gone down this road, however. But it may be considerable to do that.

> So it does not matter, that there is the special event or just some property as piggy-back. This could be also true for the other (needed) fields.

Regarding the other fields: in the case of the special event specified here, this would only be the `source` field. And this indeed could be filled by the protocol adapters as well - if they receive the event, it would be `device`, if they generate it themselves, it would be the name of the protocol adapter (like `hono-http`).

Remains the `stay-connected` JSON field for MQTT messages: any idea to get rid of this?
Maybe we should just skip this functionality?

> So the event data would be always free for the application event,

Yes, if we find a solution for `stay-connected` in MQTT messages.

> if piggy-bagged or whatever the user wants to put in a 'special' event. 

Again, the `stay-connected` field is the problem: how do we get this into a piggy-backed telemetry/event message?

> Maybe this would be also better binary data for him. In contrast to this advantage I do not see the advantage of a JSON payload at the moment?

If we solve the `stay-connected` "problem", there is no advantage anylonger, I agree to this.

pellmann:
> Yes, if we find a solution for stay-connected in MQTT messages.

Since in MQTT the device typically stays connected to receive the command I think it is not such a big issue. Beside of this for me property bags (URL encoded properties at the end of the MQTT address) are a good solution.

I think we should not restrict our AMQP APIs because some adapter-to-device protocols are restricted e.g. could not handle headers. It would also always be possible to define special workarounds just for the specific adapter-to-device protocol without reflect the same in the AMQP API.

WDYT?

site/content/api/Event-API.md
```diff
@@ -76,4 +81,63 @@ The following sequence diagram illustrates the flow of messages involved in a *B
...
+
+In the following these events (currently one) are specified in detail.
+
+
+### Device command readiness notification
+
+By sending an event of this type the receiver is notified that a device is ready to receive a command.
+ 
+*Business Applications* may want to react to such an event by sending a command upstream to the device.
+For that purpose, the protocol adapter instance that received such an event opens a receiver link for the device (scoped to its tenant) that
```
sophokles73: IMHO we should remove this sentence since it assumes a particular way of implementation which may or may not be realized this way. In any case, though, the implementation is not relevant for describing the purpose of this event.

site/content/api/Telemetry-API.md
```diff
@@ -65,6 +65,7 @@ The following table provides an overview of the properties a client needs to set
...
 | *device_id*    | yes       | *application-properties* | *string*  | MUST contain the ID of the device the data in the payload has been reported by. |
 | *reg_assertion*| yes       | *application-properties* | *string*  | A [JSON Web Token](https://jwt.io/introduction/) issued by the [Device Registration service]({{< relref "api/Device-Registration-API.md#assert-device-registration" >}}) asserting the device's registration status. |
+| *stay_connected* | no      | *application-properties* | *long*  | In context with the `creation-time` of the AMQP 1.0 message it defines the time interval in milliseconds in which a device should be available for receiving upstream data. If this property is missing, it shall be considered as not being known if the device is ready for receiving a message. If it has the value 0, the device shall be considered as being always ready to receive a message. |
```
sophokles73: again, I propose to rename to `ttd`

Instead of 0 I propose to  use -1 as the indicator for *always online* because it is completely out-of-reasonable-range.

site/content/api/Event-API.md
```diff
@@ -76,4 +81,63 @@ The following sequence diagram illustrates the flow of messages involved in a *B
...
+
+| Name           | Mandatory | Location                 | Type      | Description |
+| :------------- | :-------: | :----------------------- | :-------- | :---------- |
+| *content-type* | yes       | *properties*             | *symbol*  | Must be set to *application/vnd.eclipse-hono-dcr-notification+json* |
+
+The body of the event request MUST consist of a single *AMQP Value* section containing a UTF-8 encoded string representation of a single JSON object having the following members:
+
+| Name                | Mandatory | Type       | Default Value | Description |
+| :-----------------  | :-------: | :--------- | :----------   | :---------- |
+| *source*            | *yes*     | *string*   |               | Value can be either the name of a protocol adapter (`hono-mqtt`, `hono-http`, `hono-kura`) or `device`.|
```
sophokles73: I have a hard time seeing any good use for this information.

site/content/api/Event-API.md
```diff
@@ -76,4 +81,63 @@ The following sequence diagram illustrates the flow of messages involved in a *B
...
+| *stay-connected*    | no        | *long*     |               | In context with the `creation-time` (set for the AMQP 1.0 message by the protocol adapters) it defines the time interval in milliseconds in which a device should be available for receiving upstream data. If this property is missing or if it has the value 0, the device shall be considered as being always ready to receive a message. |
```
sophokles73: I propose to rename this property to ttd (time-til-disconnect) in analogy to ttl.

Also, FMPOV the scale should be seconds instead of milliseconds. I do not see the purpose of specifying 350 ms or 1756ms ...

site/content/api/Event-API.md
```diff
@@ -76,4 +81,63 @@ The following sequence diagram illustrates the flow of messages involved in a *B
...
+event on behalf of devices (to support setups with devices that are not capable of sending such an event).
+
+
+***Device command readiness notification Payload***
+
+The following table provides an overview of the properties an event sender needs to set for a *device command readiness notification* event.
+
+| Name           | Mandatory | Location                 | Type      | Description |
+| :------------- | :-------: | :----------------------- | :-------- | :---------- |
+| *content-type* | yes       | *properties*             | *symbol*  | Must be set to *application/vnd.eclipse-hono-dcr-notification+json* |
```
sophokles73: what about `application/vnd.eclipse-hono-ready-to-receive+json`

sophokles73:
> Since in MQTT the device typically stays connected to receive the command I think it is not such a big issue. Beside of this for me property bags (URL encoded properties at the end of the MQTT address) are a good solution.

If an MQTT based device is implemented to e.g. open a connection, send its data, then wait for a certain amount of time and then disconnect, we can simply register a default *ttd* for the device in the registry which the MQTT adapter can then include in the AMQP messages it sends downstream for the MQTT messages published by the device ...

site/content/api/Event-API.md
sysexcontrol: Agreed - is contained also in the concept page anyway.

site/content/api/Event-API.md
sysexcontrol: I understand the concern, and @ppatierno asked the same question.

So I can eliminate it, but IMHO the sender of the event is of some importance. E.g. if there are mixed setups where a device itself controls fine grained how long it can receive messages, in contrast to a protocol adapter that does this on a more coarse grained level.

But (as I stated in another conversation already), this is not a 100% sharpened use case, so I am willing to delete it now.

sysexcontrol:
> If an MQTT based device is implemented to e.g. open a connection, send its data, then wait for a certain amount of time and then disconnect, we can simply register a default ttd for the device in the registry which the MQTT adapter can then include in the AMQP messages it sends downstream for the MQTT messages published by the device ...

To bring in the device registry is an additional (new) use case, and is a good idea IMHO.
I find it hard to judge though if this matches real devices, but anyway - I will include it as another option.

However, it does not the use case of devices that want to control themselves when they are ready to receive messages.
We either exclude this use case, or need to use something like property bags for MQTT, don't we?

site/content/api/Event-API.md
sysexcontrol: Since it is used in a lot of places already, I would rather prefer
`application/vnd.eclipse-hono-ready-for-upstream+json`.

I would be uncertain as a reader about `who` is ready to receive otherwise...

WDYT?

pellmann:
> We either exclude this use case, or need to use something like property bags for MQTT, don't we?

I would propose to leave this out at the beginning (since we have solutions if really needed).

site/content/api/Event-API.md
pellmann: It defines a semantic (about what should happen) with this content-types and I am not sure if this is not to much. What is indicated is a stay-connected-timeframe. Maybe we should only indicate that and do not care, what the receiver will do with this information.

site/content/api/Event-API.md
sophokles73: What difference does it make to the consumer of the event? Will the application apply a *corrective* factor to the ttd value if the event has been sent by the protocol adapter instead of the device?

site/content/api/Event-API.md
sophokles73: I don't like that name either. If we are able to get rid of the payload, we can also use `plain/empty` ;-)

sysexcontrol:
Ok, then we would have the following behaviour now:
- MQTT devices cannot indicate their readiness themselves
- Instead the protocol adapter does this by retrieving the value for `ttd` from the device registry and puts it into the AMQP message
- In the future this could be also done on a tenant level (`ttd` tenant defaults)

When having discussed integration with Eclipse Ditto, the requirement was to let the device add some data to the event that indicates command readiness. We completely miss that then, since no default value can cover this.

To me it doesn't feel right to completely eliminate the use case that a device itself can send such an event.
I propose to let the device still send such an event, but the `ttd` comes from the registry for now (and no property bags or whatever introduced).
That would be a minimal solution for me - WDYT?

pellmann:
> That would be a minimal solution for me - WDYT?

You are right, we will not have a "full" solution for MQTT with this first implementation but we have the options to fulfill this in the future. If we first concentrate on HTTP adapter and go more straight forward  we can collect more experiences and sharpen the real requirements. 

So my concern is to not build something for what we do not know the real need well at the moment while we can first do the more simple and obvious parts (since we discussed the problems in this thread and found possible solutions).

sophokles73:
I would like to come back to the question whether we actually need any payload at all in the event. The use case described by @sysexcontrol, i.e. reporting a *last known status/revision* of the device's data to Eclipse Ditto, made me think that we probably should define an event for that particular use case and then simply include the standard (piggy backing) AMQP message properties needed to inform the recipient about the fact that the device is (also) ready to receive a command. However, the original purpose of the event would be to report its *latest revision* or *status*, if you will. The content-type could then better reflect this purpose instead of us trying to come up with some bogus type that we do not seem to be able to find. This way we also get rid of the (at least perceived) *asymmetry* between using the piggy-backing approach vs. sending an explicit event.

sysexcontrol:
> You are right, we will not have a "full" solution for MQTT with this first implementation but we have the options to fulfill this in the future. If we first concentrate on HTTP adapter and go more straight forward we can collect more experiences and sharpen the real requirements.

> So my concern is to not build something for what we do not know the real need well at the moment while we can first do the more simple and obvious parts (since we discussed the problems in this thread and found possible solutions).

Actually, I do not understand the point. The most simple (and already realized) solution was to let the device send the event with all parameters.

More complex seems to me what we have now (that we discussed for some good reasons, don't get me wrong):
- let the MQTT PA send the event for the device (when it connects)
- define HTTP headers to let the device send the event (without payload), which are mapped to the AMQP 1.0 properties like `ttd`
- let the PAs look up the value for `ttd` in either the Device Registry or via the Tenant API

But anyway, I will make a proposal in the comment to what @sophokles73 wrote below, which could be an agreed on "minimal solution" for me.

sysexcontrol:
> I would like to come back to the question whether we actually need any payload at all in the event. The use case described by @sysexcontrol, i.e. reporting a last known status/revision of the device's data to Eclipse Ditto, made me think that we probably should define an event for that particular use case and then simply include the standard (piggy backing) AMQP message properties needed to inform the recipient about the fact that the device is (also) ready to receive a command. 

That's a very good idea IMHO - and also improves clarity for me about the purpose of the different events!

I propose that we define the following "minimal approach" then:
- no payload for the "command readiness notification" anymore
- the new event type (outlined in your comment) will be defined later (and will have an arbitrary payload, using the AMQP 1.0 properties of the piggy-backed messages)
- HTTP PA: in this first step the "command readiness notification" **can** be sent by the device, but we do not define HTTP headers for it. 
  - the device sets only the content-type, does not include any payload, and uses the Event API to send it downstream.
  - instead the `ttd` value for this event is taken from either the Device Registry or from the Tenant API. If no such value can be found, the notification is NOT propagated.
  - an upstream message may be contained in the response to the HTTP request. This will use the same response headers as when piggy-backing other downstream messages.
  - the response header(s) need to be defined (unified for both cases)
- MQTT PA: the "command readiness notification" **can** be sent by the device also and works exactly the same as for the HTTP PA. 
  - the only contrast to the HTTP PA is : there will be no response containing an upstream message to the "command readiness notification". 

Piggy-backing messages:
- no difference to the "command readiness notification" anymore
- all points described above apply exactly the same here

@pellmann @sophokles73 : Does that reflect our common view on the "command readiness notification"?

site/content/api/Event-API.md
sysexcontrol: >  What is indicated is a stay-connected-timeframe

I thought differently: indicated is "ready to receive a message".
"Stay connected" is not the same for me as "willing to receive a message".

Since we have no payload anymore, we need a new type now, probably something like
`application/vnd.eclipse-hono-...+empty`

Any ideas?

site/content/api/Event-API.md
sysexcontrol: No, and I understand the concerns about the usability of this field (and remove it).
But I imagine that *a device explicitly indicating it's readiness to receive a message* could be treated with maximum importance (the device would not have sent it if not available just now).

Anyway, let's introduce it later (again) if there should be a use for it.

sophokles73:
> Does that reflect our common view on the "command readiness notification"?

No, FMPOV the only way of indicating that a device is ready to receive a command would be by means of including the `ttd` property in a downstream message. It is irrelevant whether the property's value is provided by the device itself (as part of a message it publishes) or if the protocol adapter determines the value based on e.g. a default configured for the device (or tenant).

A device may of course use the (to be defined) event for reporting its latest revision/status (or any other event for that matter) in order to indicate it readiness to receive a command. We may even define a special (empty) *dummy* event for that purpose.

> if we would use the default value for ttd here as well, every telemetry and event message via MQTT would implicitly indicate that the device is ready to receive a message (does not make much sense IMHO)

Why not? The default value either reflects the correct value because the device indeed disconnects after some time, or it is implicitly correct because the device never disconnects.

sysexcontrol:
> FMPOV the only way of indicating that a device is ready to receive a command would be by means of including the ttd property in a downstream message.

Yes. That is my view as well.

> It is irrelevant whether the property's value is provided by the device itself (as part of a message it publishes) or if the protocol adapter determines the value based on e.g. a default configured for the device (or tenant).

Yes, with the remark that the **device** can provide this property's value currently only via HTTP, and not via MQTT (only a default configured for the device or tenant can be used here).

> A device may of course use the (to be defined) event for reporting its latest revision/status (or any other event for that matter) in order to indicate it readiness to receive a command. We may even define a special (empty) dummy event for that purpose.

Yes, the special empty event is what is defined here (as "command readiness notification").
It can be sent by the device itself without publishing any other telemetry or event message, just for indicating the command readiness, and therefore we have the special content-type.

> Why not? The default value either reflects the correct value because the device indeed disconnects after some time, or it is implicitly correct because the device never disconnects.

But this comes for the price that there is no separation between a device *being connected* to MQTT and a device *having some reocurring window for receiving a command, while staying connected always*. 
With using the default values to piggy-back MQTT messages, the device would always be assumed to be ready for commands after sending any message.

For simplicity I will specify it like that for now, but I believe we need a better solution for MQTT devices later on.

So, I guess we are done now and I will continue with the prototyping and specification on this base.

Thanks for all review comments...

site/content/api/Event-API.md
sophokles73: `application/vnd.eclipse-hono-empty-notification` ?

sysexcontrol: Taken :+1:

site/content/api/Event-API.md
```diff
@@ -76,4 +81,63 @@ The following sequence diagram illustrates the flow of messages involved in a *B
 
 **Message Format**
 
+**Arbitrary events from the device**
```
sophokles73: IMHO we should omit this line. The message format is still defined as described by the Telemetry API regardless of whether a specific *well-known* event is sent. The *well-known* event types simply define some particular values that the properties need to be set to.

sophokles73:
@sysexcontrol are you going to incorporate all the feedback you received so that we can finish up this PR?

sysexcontrol:
@sophokles73 : yes, that is now top of the list after I finished with adding query parameters for `ttd` header to the HTTP adapter in PR #596 .

sysexcontrol:
@sophokles73 @pellmann : I updated the specification and considered all review comments.

site/content/api/Event-API.md
```diff
@@ -77,3 +82,31 @@ The following sequence diagram illustrates the flow of messages involved in a *B
...
+
+| Name           | Mandatory | Location                 | Type      | Description |
+| :------------- | :-------: | :----------------------- | :-------- | :---------- |
+| *content-type* | yes       | *properties*             | *symbol*  | Must be set to *application/vnd.eclipse-hono-empty-notification* |
+| *ttd*          | yes       | *application-properties* | *int*     | 'time til disconnect' : see [*Telemetry API*]({{< relref "Telemetry-API.md" >}}). | 
+
+Note that for this event type the property `ttd` must be set, as otherwise a receiver could not determine if the event is still valid when it receives it (different from other messages sent downstream).
+
+
+NB: Such an event is typically issued by the device itself to indicate that it is ready to receive a command. In future releases the protocol adapters may be able to send this
```
sophokles73: this should be rephrased to give an example of what an *empty notification* could be used for.

site/content/api/Telemetry-API.md
```diff
@@ -60,11 +60,13 @@ The following sequence diagram illustrates the flow of messages involved in the
...
-| *reg_assertion*| yes       | *application-properties* | *string*  | A [JSON Web Token](https://jwt.io/introduction/) issued by the [Device Registration service]({{< relref "api/Device-Registration-API.md#assert-device-registration" >}}) asserting the device's registration status. |
+| Name            | Mandatory | Location                 | Type      | Description |
+| :-------------- | :-------: | :----------------------- | :-------- | :---------- |
+| *content-type*  | yes       | *properties*             | *symbol*  | SHOULD be set to *application/octet-stream* if the message payload is to be considered *opaque* binary data. In most cases though, the client should be able to set a more specific content type indicating the type and characteristics of the data contained in the payload, e.g. `text/plain; charset="utf-8"` for a text message or `application/json` etc. |
+| *creation-time* | yes       | *properties*             | *timestamp* | MUST be set to the current number of milliseconds since the unix epoch (1970-01-01T00:00:00Z). |
```
sophokles73: since this is the standard property defined by AMQP 1.0 why not simply use the definition from the spec (which also notes that the point in time it refers to is supposed to be UTC).

site/content/concepts/device-notifications.md
```diff
@@ -0,0 +1,116 @@
...
+The content here summarizes the current discussion and is being prototyped.
+Please join our mailing list or consult the IRC channel in case you have any questions or want to participate.
+{{% /warning %}}
+
+<!--more-->
+
+For implementing specific behaviour *Business applications* need to know if
+
+- a device is currently ready to receive a command
+- a device is currently connected to a protocol adapter
```
sophokles73: I was under the impression that we wanted to separate these two concerns. So I propose to remove the *connected to protocol adapter* stuff.

site/content/concepts/device-notifications.md
```diff
@@ -0,0 +1,116 @@
...
+ 
+The first can serve e.g. to implement functionality like sending a command upstream to a device.
+
+In the following it is described how a *Business application* can be notified that a device is ready to receive a command. 
+This is referred to as `device command readiness`.
```
sophokles73: I believe we can find a more elegant term than `device command readiness`

site/content/concepts/device-notifications.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Device command readiness notification
+
+To notify an application about the command readiness of a device specific properties are set in a downstream message.
+
+This can be done
+
+- implicitly: when sending a downstream message for any reason (e.g. telemetry or event message)
+- explicitly: by sending a specific empty event (e.g. if a device does not have any other downstream message to send)
```
sophokles73: no, this can only be done by adding a ttd property to a telemetry or event message. using an empty event is just a special case, nothing more.

site/content/concepts/device-notifications.md
```diff
@@ -0,0 +1,116 @@
...
+
+
+The following AMQP 1.0 properties of the message are relevant for the notification:
+
+- the AMQP 1.0 property `creation-time` (which is automatically set by protocol adapters)
```
sophokles73: to which value?

site/content/concepts/device-notifications.md
```diff
@@ -0,0 +1,116 @@
...
+- an additional application-property `ttd` (containing the amount of seconds the device stays connected) in the AMQP 1.0 message
```
sophokles73: the number of seconds the device will stay connected to the protocol adapter

site/content/concepts/device-notifications.md
```diff
@@ -0,0 +1,116 @@
...
+ 
+Please refer to the [Telemetry API]({{< relref "api/Telemetry-API.md" >}}) and the [Event API]({{< relref "api/Event-API.md" >}}) for further details.
+
+### Time based validation of notification 
+
+An application receiving a downstream message containing the notification relevant properties can verify at any point in
+time if a device is ready to receive a command by:
+
+- adding the values of the `creation-time` and the `ttd` properties to build a *valid until timestamp*
```
sophokles73: we should also be able to find a better term than *valid until timestamp*

site/content/concepts/device-notifications.md
```diff
@@ -0,0 +1,116 @@
...
+- comparing the current time with the *valid until timestamp*
+
+If the current time is *before* the *valid until timestamp*, the notification shall be regarded as being currently valid.
```
sophokles73: I do not think that we are talking about the *validity* of a notification. We are talking about the expiration of the ttd value.

site/content/concepts/device-notifications.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Source of the `ttd` value
+
+While it seems to be natural that a device itself indicates when it is ready to receive a command, it may not always be
+possible or desirable to do so.
+A device could e.g. be not capable to adopt it's protocol, or all devices of a particular setup would always use the same value
```
sophokles73: what is

> A device could e.g. be not capable to adopt it's protocol

supposed to mean?

site/content/concepts/device-notifications.md
```diff
@@ -0,0 +1,116 @@
...
+
+Additionally different protocols may or may not let a sender set specific values for a message, so a device using a 
+specific protocol may not be able to
+provide a value for the `ttd` property at all.
+
+For these reasons there are (resp. will be) additional ways of setting the value of `ttd`:
+
+- Hono's Device Registry supports default values for any application-properties in the AMQP 1.0 message. By this means
+  a device can be configured to always have a specific value for `ttd`.
+- In a future extension there will be a configuration value per tenant and protocol adapter that sets the value of `ttd`
```
sophokles73: This should better be written subconjunctively because we do not know yet, if we will ever implement it

site/content/concepts/device-notifications.md
```diff
@@ -0,0 +1,116 @@
...
+  if it was not provided by other means already (like provided to the protocol adapter or by using a default value of the 
+  Device Registry).
+  
+### Hono's HTTP protocol adapter
```
sophokles73: all of the information below should be added to the HTTP adapter's user guide

site/content/api/Event-API.md
```diff
@@ -77,3 +82,31 @@ The following sequence diagram illustrates the flow of messages involved in a *B
...
 See [*Telemetry API*]({{< relref "Telemetry-API.md" >}}) for definition of message format. 
+
+## Well-known event message types
+
+Hono defines *well-known* events that are of a specific *content-type*.
+
+In the following these events (currently one) are specified in detail.
+
+
+### Device command readiness notification
```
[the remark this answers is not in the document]
sysexcontrol: Yes, is should be renamed to *empty notification*.

site/content/concepts/device-notifications.md
sysexcontrol: What I mean here is: 
there is no way to add the `ttd` value by the device, since it e.g. does not support headers or query parameters (in the case of HTTP) .
Will rephrase it.

sysexcontrol:
@sophokles73 : made all changes from your review in the last commit.

site/content/api/Telemetry-API.md
```diff
@@ -63,7 +63,7 @@ The following table provides an overview of the properties a client needs to set
 | Name            | Mandatory | Location                 | Type      | Description |
 | :-------------- | :-------: | :----------------------- | :-------- | :---------- |
 | *content-type*  | yes       | *properties*             | *symbol*  | SHOULD be set to *application/octet-stream* if the message payload is to be considered *opaque* binary data. In most cases though, the client should be able to set a more specific content type indicating the type and characteristics of the data contained in the payload, e.g. `text/plain; charset="utf-8"` for a text message or `application/json` etc. |
-| *creation-time* | yes       | *properties*             | *timestamp* | MUST be set to the current number of milliseconds since the unix epoch (1970-01-01T00:00:00Z). |
+| *creation-time* | yes       | *properties*             | *timestamp* | the time when this message was created (see the [AMQP 1.0 specification](http://docs.oasis-open.org/amqp/core/v1.0/amqp-core-messaging-v1.0.html) for details).
```
sophokles73: FMPOV this is mandatory only if ttd is set, isn't it?

site/content/concepts/device-notifications.md
```diff
@@ -17,86 +17,65 @@ Please join our mailing list or consult the IRC channel in case you have any que
...
+This notification can be triggered in an application by any downstream message as long as the following AMQP 1.0 properties are set:
 
-This can be done
+- the property `creation-time` (automatically set by protocol adapters to the current time when creating the message)
+- an additional application-property `ttd` (the number of seconds the device will stay connected to the protocol adapter)
 
-- implicitly: when sending a downstream message for any reason (e.g. telemetry or event message)
-- explicitly: by sending a specific empty event (e.g. if a device does not have any other downstream message to send)
+A device that wants to only trigger this notification (without sending other downstream data) may send an 
+*empty notification* event.
```
sophokles73: what about a link to the specification of the *empty notification* in the Event API?

site/content/concepts/device-notifications.md
```diff
@@ -17,86 +17,65 @@ Please join our mailing list or consult the IRC channel in case you have any que
...
-time if a device is ready to receive a command by:
+time if a device is ready to receive an upstream message by:
 
-- adding the values of the `creation-time` and the `ttd` properties to build a *valid until timestamp*
-- comparing the current time with the *valid until timestamp*
+- adding the values of the `creation-time` and the `ttd` properties to build an *expiration timestamp*
+- comparing the current time with the *expiration timestamp*
 
-If the current time is *before* the *valid until timestamp*, the notification shall be regarded as being currently valid.
+If the current time is *before* the *expiration timestamp*, the notification shall be regarded as being expired.
```
sophokles73: My understanding is that the notification is expired if the current time is **after** the expiration time.

site/content/user-guide/http-adapter.md
```diff
@@ -21,6 +21,7 @@ There is a subtle difference between the *device identifier* (*device-id*) and t
 * Request Headers:
   * (required) `Authorization`: The device's *auth-id* and plain text password encoded according to the [Basic HTTP authentication scheme](https://tools.ietf.org/html/rfc7617).
   * (required) `Content-Type`: The type of payload contained in the body.
+  * (optional) `hono-ttd`: The number of seconds the device stays connected.
```
sophokles73: the device will wait for the response

site/content/user-guide/http-adapter.md
```diff
@@ -48,6 +49,7 @@ Publish some JSON data for device `4711`:
 * Method: `PUT`
 * Request Headers:
   * (required) `Content-Type`: The type of payload contained in the body.
+  * (optional) `hono-ttd`: The number of seconds the device stays connected.
```
sophokles73: the device will wait for the response

site/content/user-guide/http-adapter.md
```diff
@@ -75,6 +77,7 @@ Publish some JSON data for device `4711`:
 * Request Headers:
   * (required) `Authorization`: The gateway's *auth-id* and plain text password encoded according to the [Basic HTTP authentication scheme](https://tools.ietf.org/html/rfc7617).
   * (required) `Content-Type`: The type of payload contained in the body.
+  * (optional) `hono-ttd`: The number of seconds the device stays connected.
```
sophokles73: the device will wait for the response

site/content/user-guide/http-adapter.md
```diff
@@ -106,6 +109,7 @@ Publish some JSON data for device `4712` via gateway `gw-1`:
 * Request Headers:
   * (required) `Authorization`: The device's *auth-id* and plain text password encoded according to the [Basic HTTP authentication scheme](https://tools.ietf.org/html/rfc7617).
   * (required) `Content-Type`: The type of payload contained in the body.
+  * (optional) `hono-ttd`: The number of seconds the device stays connected.
```
sophokles73: the device will wait for the response

site/content/user-guide/http-adapter.md
```diff
@@ -133,6 +137,7 @@ Publish some JSON data for device `4711`:
 * Method: `PUT`
 * Request Headers:
   * (required) `Content-Type`: The type of payload contained in the body.
+  * (optional) `hono-ttd`: The number of seconds the device stays connected.
```
sophokles73: the device will wait for the response

site/content/user-guide/http-adapter.md
```diff
@@ -160,6 +165,7 @@ Publish some JSON data for device `4711`:
 * Request Headers:
   * (required) `Authorization`: The gateway's *auth-id* and plain text password encoded according to the [Basic HTTP authentication scheme](https://tools.ietf.org/html/rfc7617).
   * (required) `Content-Type`: The type of payload contained in the body.
+  * (optional) `hono-ttd`: The number of seconds the device stays connected.
```
sophokles73: the device will wait for the response

site/content/user-guide/http-adapter.md
```diff
@@ -197,6 +203,36 @@ The adapter also considers [*defaults* registered for the device]({{< relref "ap
 
 Note that of the standard AMQP 1.0 message properties only the *content-type* can be set this way to a registered default value.
 
+## Specifying the time a device stays connected
```
sophokles73: a device will wait for a response

site/content/user-guide/http-adapter.md
```diff
@@ -197,6 +203,36 @@ The adapter also considers [*defaults* registered for the device]({{< relref "ap
...
+
+The adapter lets devices specify the number of seconds they stay connected to it by setting a header or a query parameter.
```
sophokles73: they will wait for a response

site/content/api/Telemetry-API.md
sysexcontrol: That is true, although the protocol adapters now always set it.
But since this is the specification, I completely agree - will adopt the text.

sysexcontrol:
@sophokles73 : thanks for the comments, new commit pushed.

PR #585, opened 2018-04-17
Merge the two OpenShift deployments into one

ctron:
The new OpenShift guide focuses on using EnMasse on a single node/single master minishift installation.


ctron:
@dejanb It would be great if you could test it on Mac OS

PR #588, opened 2018-04-19
Client implementation for Command and Control

pellmann:
Business Application side of Command and Control:

- changed String/JSON to Buffer in the base classes to support binary content.
- new CommandClient to send commands, which expects a result.
- in this implementation a pair of links is created for each device

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/CommandClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/CommandClient.java
@@ -1,0 +1,33 @@
+package org.eclipse.hono.client;
+
+import io.vertx.core.Future;
+import io.vertx.core.buffer.Buffer;
+
+/**
+ * A client for accessing Hono's Command and Control API.
+ * <p>
+ * An instance of this interface is always scoped to a specific tenant and device.
+ * </p>
+ */
+public interface CommandClient extends RequestResponseClient {
+
+    /**
+     * Sends a command to a device and expects a response.
+     * <p>
+     * A device needs to be (successfully) registered before a client can upload
+     * any data for it. The device also needs to be connected for a successful delivery.
+     *
+     * @param command The command name.
+     * @param data The command data to send to the device.
+     * @return A future indicating the result of the operation.
+     *         <p>
+     *         The future will succeed if a response with status 2xx has been received from the device. If the response has no payload, the future will complete with {@code null}.
+     *         <p>
+     *         Otherwise, the future will fail with a {@link ServiceInvocationException} containing
+     *         the (error) status code. Status codes are defined at <a href="https://www.eclipse.org/hono/api/command-and-control-api"/>
+     * @throws NullPointerException if command or data is {@code null}.
+     * @see RequestResponseClient#setRequestTimeout(long)
+     */
+    Future<Buffer> sendCommand(String command, Buffer data);
+
+}
--- a/client/src/main/java/org/eclipse/hono/client/HonoClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/HonoClient.java
@@ -255,6 +255,19 @@ public interface HonoClient {
     Future<TenantClient> getOrCreateTenantClient();
 
     /**
+     * Gets a client for sending command with Hono's <em>Command and Control</em> API.
+     *
+     * @param tenantId The tenant to the all command receiving devices belong.
+     * @param deviceId The device id.
+     * @return A future that will complete with the command and control client (if successful) or
+     *         fail if the client cannot be created, e.g. because the underlying connection
+     *         is not established or if a concurrent request to create a client for the same
+     *         tenant and device is already being executed.
+     * @throws NullPointerException if the tenant is {@code null}.
+     */
+    Future<CommandClient> getOrCreateCommandClient(String tenantId, String deviceId);
+
+    /**
      * Closes this client's connection to the Hono server.
      * <p>
      * This method waits for at most 5 seconds for the connection to be closed properly. Any subsequent attempts to
--- a/client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
@@ -21,6 +21,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
 
+import io.vertx.core.buffer.Buffer;
 import org.apache.qpid.proton.amqp.messaging.Accepted;
 import org.apache.qpid.proton.amqp.messaging.AmqpValue;
 import org.apache.qpid.proton.amqp.messaging.Rejected;
@@ -44,7 +45,6 @@ import io.vertx.core.AsyncResult;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
-import io.vertx.core.json.JsonObject;
 import io.vertx.proton.ProtonConnection;
 import io.vertx.proton.ProtonDelivery;
 import io.vertx.proton.ProtonHelper;
@@ -107,6 +107,26 @@ public abstract class AbstractRequestResponseClient<R extends RequestResponseRes
             this.targetAddress = String.format("%s/%s", getName(), tenantId);
             this.replyToAddress = String.format("%s/%s/%s", getName(), tenantId, UUID.randomUUID());
         }
+        this.tenantId = tenantId;
+    }
+
+    /**
+     * Creates a request-response client.
+     * <p>
+     * The client will be ready to use after invoking {@link #createLinks(ProtonConnection)} or
+     * {@link #createLinks(ProtonConnection, Handler, Handler)} only.
+     *
+     * @param context The vert.x context to run message exchanges with the peer on.
+     * @param config The configuration properties to use.
+     * @param tenantId The identifier of the tenant that the client is scoped to. May be {@code null}.
+     * @param deviceId The device id to which a specific client should be scoped.
+     * @throws NullPointerException if any of context or configuration are {@code null}.
+     */
+    AbstractRequestResponseClient(final Context context, final ClientConfigProperties config, final String tenantId, final String deviceId) {
+        super(context, config);
+        this.requestTimeoutMillis = config.getRequestTimeout();
+        this.targetAddress = String.format("%s/%s/%s", getName(), tenantId, deviceId);
+        this.replyToAddress = String.format("%s/%s/%s/%s", getName(), tenantId, deviceId, UUID.randomUUID());
         this.tenantId = tenantId;
     }
 
@@ -202,11 +222,11 @@ public abstract class AbstractRequestResponseClient<R extends RequestResponseRes
      * Creates a result object from the status and payload of a response received from the endpoint.
      *
      * @param status The status of the response.
-     * @param payload The String representation of the response's JSON payload (may be {@code null}).
+     * @param payload The representation of the payload (may be {@code null}).
      * @param cacheDirective Restrictions regarding the caching of the payload (may be {@code null}).
      * @return The result object.
      */
-    protected abstract R getResult(int status, String payload, CacheDirective cacheDirective);
+    protected abstract R getResult(int status, Buffer payload, CacheDirective cacheDirective);
 
     /**
      * Creates the sender and receiver links to the peer for sending requests
@@ -232,7 +252,8 @@ public abstract class AbstractRequestResponseClient<R extends RequestResponseRes
      *         have been created.
      * @throws NullPointerException if connection is {@code null}.
      */
-    protected final Future<Void> createLinks(final ProtonConnection con, final Handler<String> senderCloseHook, final Handler<String> receiverCloseHook) {
+    protected final Future<Void> createLinks(final ProtonConnection con, final Handler<String> senderCloseHook,
+            final Handler<String> receiverCloseHook) {
 
         Objects.requireNonNull(con);
 
@@ -335,10 +356,8 @@ public abstract class AbstractRequestResponseClient<R extends RequestResponseRes
         if (status == null) {
             return null;
         } else {
-            final String payload = MessageHelper.getPayload(message);
             final CacheDirective cacheDirective = CacheDirective.from(MessageHelper.getCacheDirective(message));
-
-            return getResult(status, payload, cacheDirective);
+            return getResult(status, MessageHelper.getPayload(message), cacheDirective);
         }
     }
 
@@ -371,7 +390,7 @@ public abstract class AbstractRequestResponseClient<R extends RequestResponseRes
     /**
      * Creates a request message for a payload and sends it to the peer.
      * <p>
-     * This method simply invokes {@link #createAndSendRequest(String, Map, JsonObject, Handler)}
+     * This method simply invokes {@link #createAndSendRequest(String, Map, Buffer, Handler)}
      * with {@code null} for the properties parameter.
      * 
      * @param action The operation that the request is supposed to trigger/invoke.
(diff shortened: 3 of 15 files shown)


client/src/main/java/org/eclipse/hono/client/impl/RegistrationClientImpl.java
```diff
@@ -101,7 +102,7 @@ protected final String createMessageId() {
     }
 
     @Override
-    protected final RegistrationResult getResult(final int status, final String payload, final CacheDirective cacheDirective) {
+    protected final RegistrationResult getResult(final int status, Buffer payload, final CacheDirective cacheDirective) {
```
sysexcontrol: final

core/src/main/java/org/eclipse/hono/util/CommandConstants.java
```diff
@@ -0,0 +1,23 @@
+package org.eclipse.hono.util;
+
+/**
+ * Commands &amp; utility methods used throughout the Command & Control API.
```
sysexcontrol: Javadoc error: &amp; instead of `& `

core/src/main/java/org/eclipse/hono/util/CommandConstants.java
```diff
@@ -0,0 +1,23 @@
...
+public class CommandConstants {
+
+    /**
+     * Empty default constructor.
+     */
+    protected CommandConstants () {
+    }
+
+    /**
+     * The name of the Command & Control API. endpoint.
```
sysexcontrol: Javadoc error: & instead of &

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -293,20 +294,19 @@ public static boolean getXOptAppCorrelationId(final Message msg) {
      */
     public static JsonObject getJsonPayload(final Message msg) {
 
-        final String payload = getPayload(msg);
-        return (payload != null ? new JsonObject(payload) : null);
+        final Buffer buffer = getPayload(msg);
+        return buffer!=null ? new JsonObject(buffer.toString(StandardCharsets.UTF_8)) : null;
```
sysexcontrol: whitespaces

pellmann:
Thank you @sysexcontrol  - I made the changes.

client/src/main/java/org/eclipse/hono/client/CommandClient.java
```diff
@@ -0,0 +1,54 @@
...
+     * @return A future indicating the result of the operation.
+     *         <p>
+     *         The future will succeed if a response with status 201 has been received from the device/adapter.
+     *         <p>
+     *         Otherwise, the future will fail with a {@link ServiceInvocationException} containing
+     *         the (error) status code returned by the service.
+     * @throws NullPointerException if device ID is {@code null}.
+     * @see RequestResponseClient#setRequestTimeout(long)
+     */
+    Future<Void> command(String command, Buffer data);
```
sophokles73: can we rename this to `sendCommand` so that the methods at least contain a verb indicating their purpose?

the same applies to the other method, of course.

client/src/main/java/org/eclipse/hono/client/CommandClient.java
```diff
@@ -0,0 +1,54 @@
...
+     * any data for it. The device also needs to be connected for a successful delivery.
+     *
+     * @param command The command name.
+     * @param data The command data to send to the device.
+     * @return A future indicating the result of the operation.
+     *         <p>
+     *         The future will succeed if a response with status 201 has been received from the device/adapter.
+     *         <p>
+     *         Otherwise, the future will fail with a {@link ServiceInvocationException} containing
+     *         the (error) status code returned by the service.
```
sophokles73: returned by the *service*? that should be *returned by the device*, right?

client/src/main/java/org/eclipse/hono/client/CommandClient.java
```diff
@@ -0,0 +1,54 @@
...
+     * @throws NullPointerException if device ID is {@code null}.
```
sophokles73: there is no device ID parameter, which is very strange because it raises the question: how do we address a device?

I am a little confused now, is this client supposed to be used by an application that wants to send a command to a device?

client/src/main/java/org/eclipse/hono/client/CommandClient.java
pellmann: 'command' is a verb, isn't it? But I am fine with sendCommand!

client/src/main/java/org/eclipse/hono/client/CommandClient.java
pellmann: Yes, but the link is created exclusively for this device - so there is no need to set it again at sending.

client/src/main/java/org/eclipse/hono/client/CommandClient.java
pellmann: Yes, that is more precise!

client/src/main/java/org/eclipse/hono/client/CommandClient.java
sophokles73: I see, then you should at least update the interface's class level JavaDoc accordingly. It states that the client is scoped to a tenant.

pellmann:
@sophokles73 - I made your recommended changed and rebased. Could you take a look?

client/src/main/java/org/eclipse/hono/client/CommandClient.java
```diff
@@ -0,0 +1,34 @@
+package org.eclipse.hono.client;
+
+import io.vertx.core.Future;
+import io.vertx.core.buffer.Buffer;
+
+/**
+ * A client for accessing Hono's Command and Control API.
+ * <p>
+ * An instance of this interface is always scoped to a specific tenant.
```
sophokles73: and device

client/src/main/java/org/eclipse/hono/client/CommandClient.java
```diff
@@ -0,0 +1,34 @@
...
+     * Sends a command to a device and expects a response.
+     * <p>
+     * A device needs to be (successfully) registered before a client can upload
+     * any data for it. The device also needs to be connected for a successful delivery.
+     *
+     * @param command The command name.
+     * @param data The command data to send to the device.
+     * @return A future indicating the result of the operation.
+     *         <p>
+     *         The future will succeed if a response with status 201 has been received from the device.
```
sophokles73: either list the relevant response codes from the spec or simply refer to the spec.

client/src/main/java/org/eclipse/hono/client/HonoClient.java
```diff
@@ -254,6 +254,19 @@
...
 
+    /**
+     * Gets a client for sending command with Hono's <em>Command and Control</em> API.
+     *
+     * @param tenantId The tenant to the all command receiving devices belong.
+     * @param deviceId The device id.
+     * @return A future that will complete with the command and control client (if successful) or
+     *         fail if the client cannot be created, e.g. because the underlying connection
+     *         is not established or if a concurrent request to create a client for the same
+     *         tenant is already being executed.
```
sophokles73: the same tenant or the same tenant *and* device?

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -202,11 +222,11 @@ public final void setRequestTimeout(final long timoutMillis) {
      * Creates a result object from the status and payload of a response received from the endpoint.
      *
      * @param status The status of the response.
-     * @param payload The String representation of the response's JSON payload (may be {@code null}).
+     * @param payload The Buffer representation of the payload (may be {@code null}).
```
sophokles73: no need to mention the type here, JavaDoc automatically adds the type

core/src/main/java/org/eclipse/hono/util/CommandConstants.java
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * Empty default constructor.
+     */
+    protected CommandConstants () {
+    }
+
+    /**
+     * The name of the Command and Control API. endpoint.
+     */
+    public static final String COMMAND_ENDPOINT = "control";
```
sophokles73: it feels strange that the *command endpoint* name is `control`

core/src/main/java/org/eclipse/hono/util/CommandResult.java
```diff
@@ -0,0 +1,47 @@
...
+
+package org.eclipse.hono.util;
+
+import io.vertx.core.buffer.Buffer;
+
+/**
+ * A container for the result returned by Hono's registration API.
+ *
+ */
+public final class CommandResult extends RequestResponseResult<Buffer> {
```
sophokles73: This doesn't seem to be specific to Command and Control at all. In fact, it couldn't be more generic, so why don't we call it `GenericResult`?

core/src/main/java/org/eclipse/hono/util/CommandResult.java
```diff
@@ -0,0 +1,47 @@
...
+ * Contributors:
+ *    Bosch Software Innovations GmbH - initial creation
+ */
+
+package org.eclipse.hono.util;
+
+import io.vertx.core.buffer.Buffer;
+
+/**
+ * A container for the result returned by Hono's registration API.
```
sophokles73: it is not related to the registration API

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -293,20 +294,19 @@ public static boolean getXOptAppCorrelationId(final Message msg) {
...
+     * Gets a message's body as Buffer object.
      *
      * @param msg The AMQP 1.0 message to parse the body of.
-     * @return The message body parsed into a JSON object or {@code null} if the message does not have a <em>Data</em>
+     * @return The message body as a Buffer or {@code null} if the message does not have a <em>Data</em>
      *         nor an <em>AmqpValue</em> section.
      * @throws NullPointerException if the message is {@code null}.
      */
-    public static String getPayload(final Message msg) {
+    public static Buffer getPayload(final Message msg) {
```
sophokles73: this is a breaking change of the public API and should therefore be at least documented in the release notes.

core/src/main/java/org/eclipse/hono/util/CommandConstants.java
pellmann: I took this from the old API definition. First I also thought it is strange but to think that the commands  are send to a control endpoint sounds not so bad.

core/src/main/java/org/eclipse/hono/util/CommandResult.java
pellmann: Or BufferResult?

client/src/main/java/org/eclipse/hono/client/HonoClient.java
pellmann: With Device.

pellmann:
I integrated the requested changes.

client/src/main/java/org/eclipse/hono/client/CommandClient.java
```diff
@@ -24,11 +24,10 @@
      *         The future will succeed if a response with status 201 has been received from the device.
      *         <p>
      *         Otherwise, the future will fail with a {@link ServiceInvocationException} containing
-     *         the (error) status code.
+     *         the (error) status code. Status codes are defined at <a href="https://www.eclipse.org/hono/api/command-and-control-api"/>
```
sophokles73: I still do not believe that the future will succeed if the device responded with a 201 status code.

pellmann: It should be 200.

sophokles73: Shouldn't this depend on the *nature* of the command, e.g. if the response contains payload (200) or not (204) or whether the command was used to create a resource (201) or not?

pellmann: What about (from API definition)?

| *2xx* | Indicating success; the codes are dependent on the adapter and device, which handles the command. |
| *4xx* | Indicating failure; the codes are dependent on the adapter and device, which handles the command. |
| *503* | The system responsible of handling the command is not available. |

sophokles73: I would replace the last description with
> The command cannot be sent to the target device.

pellmann: Okay

client/src/main/java/org/eclipse/hono/client/impl/CommandClientImpl.java
```diff
@@ -0,0 +1,107 @@
...
+
+        Objects.requireNonNull(command);
+        Objects.requireNonNull(data);
+
+        final Future<BufferResult> responseTracker = Future.future();
+        createAndSendRequest(command, data, responseTracker.completer());
+
+        return responseTracker.map(response -> {
+            switch (response.getStatus()) {
+            case HttpURLConnection.HTTP_OK:
```
sophokles73: this does not match the interface which says that the future succeeds for all 2xx codes.

client/src/main/java/org/eclipse/hono/client/CommandClient.java
```diff
@@ -0,0 +1,33 @@
...
+     * <p>
+     * A device needs to be (successfully) registered before a client can upload
+     * any data for it. The device also needs to be connected for a successful delivery.
+     *
+     * @param command The command name.
+     * @param data The command data to send to the device.
+     * @return A future indicating the result of the operation.
+     *         <p>
+     *         The future will succeed if a response with status 2xx has been received from the device.
+     *         <p>
```
sophokles73: will the future be completed with `null` or with an empty buffer if the response has no payload?

client/src/main/java/org/eclipse/hono/client/impl/RegistrationClientImpl.java
```diff
@@ -175,7 +176,7 @@ public static final void create(
         createAndSendRequest(
                 RegistrationConstants.ACTION_REGISTER,
                 createDeviceIdProperties(deviceId),
-                data,
+                data.toBuffer(),
```
sophokles73: according to the interface, data is allowed to be `null` ...

client/src/main/java/org/eclipse/hono/client/impl/RegistrationClientImpl.java
```diff
@@ -201,7 +202,7 @@ public static final void create(
         createAndSendRequest(
                 RegistrationConstants.ACTION_UPDATE,
                 createDeviceIdProperties(deviceId),
-                data,
+                data.toBuffer(),
```
sophokles73: according to the interface, data is allowed to be `null` ...

core/src/main/java/org/eclipse/hono/util/BufferResult.java
```diff
@@ -0,0 +1,47 @@
+/**
+ * Copyright (c) 2016, 2018 Bosch Software Innovations GmbH.
```
sophokles73: this is a new class, isn't it? It should be (c) 2018 only then ...

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -293,20 +294,19 @@ public static boolean getXOptAppCorrelationId(final Message msg) {
      */
     public static JsonObject getJsonPayload(final Message msg) {
 
-        final String payload = getPayload(msg);
-        return (payload != null ? new JsonObject(payload) : null);
+        final Buffer buffer = getPayload(msg);
+        return buffer != null ? new JsonObject(buffer.toString(StandardCharsets.UTF_8)) : null;
```
sophokles73: return buffer != null ? buffer.toJsonObject() : null;

pellmann: Thank you - made changes.

PR #589, opened 2018-04-19
Changed Command and Control API to newest discussion results.

pellmann:


site/content/api/Command-And-Control-API.md
```diff
@@ -12,43 +12,39 @@ THIS IS A FIRST DRAFT VERSION OF THE API SPEC AND THUS STILL SUBJECT TO DISCUSSI
...
-order to invoke operations of the API as described in the following sections.
+# Operations
 
-# Southbound Operations
+The following API defines operations that can be used by *Business Applications* to send control messages to *Protocol Adapters* (and by *Protocol Adapters* to receive them) as well as the response from the *Protocol Adapters* back to the *Business Applications*.
 
-The following operations can be used by *Devices* and/or *Protocol Adapters* to receive command & control messages sent by *Business Applications*.
-
-## Receive Control Message
+## Send and Receive a Command
```
sophokles73: If this is written from the perspective of the actor (the application) then it should probably just be *Send a Command*.

site/content/api/Command-And-Control-API.md
```diff
@@ -12,43 +12,39 @@ THIS IS A FIRST DRAFT VERSION OF THE API SPEC AND THUS STILL SUBJECT TO DISCUSSI
...
-|*ttl* | no | *properties* | *int* | Indicates the timeout value (in milliseconds) indicating when server can assume that command has not been executed because of the lack of the response from the client. Default timeout is *60000* (1 minute). |
-|*durable* | no | *properties* | *boolean* | Indicates if the message should be persisted (*store and forward*) by the underlying message transport middleware. By default command&control messages are durable.  |
+| *message-id* | yes | *properties* | *string* | MUST contain the message ID used to match incoming control message with the response message containing result of execution. |
+| *subject* | yes | *properties* | *string* | MUST contain the command to be executed by a device. |
+| *content-type* | no | *properties* | *symbol*  | SHOULD be set to *application/octet-stream* if the message payload is to be considered *opaque* binary data. In most cases though, the client should be able to set a more specific content type indicating the type and characteristics of the data contained in the payload, e.g. `text/plain; charset="utf-8"` for a text message or `application/json` etc. |
```
sophokles73: better replace *client* with *sender* to prevent confusion

site/content/api/Command-And-Control-API.md
```diff
@@ -12,43 +12,39 @@ THIS IS A FIRST DRAFT VERSION OF THE API SPEC AND THUS STILL SUBJECT TO DISCUSSI
...
-| *device_id* | yes | *application-properties* | UTF-8 *string* | MUST contain the ID of the device the data in the payload has been reported by. |
-| *tenant_id* | yes | *application-properties* | UTF-8 *string* | MAY be set by the client to indicate which tenant the the device belongs to. |
-| *command* | no | *application-properties* | UTF-8 *string* | MUST contain the command to be executed by a device. |
-|*ttl* | no | *properties* | *int* | Indicates the timeout value (in milliseconds) indicating when server can assume that command has not been executed because of the lack of the response from the client. Default timeout is *60000* (1 minute). |
-|*durable* | no | *properties* | *boolean* | Indicates if the message should be persisted (*store and forward*) by the underlying message transport middleware. By default command&control messages are durable.  |
+| *message-id* | yes | *properties* | *string* | MUST contain the message ID used to match incoming control message with the response message containing result of execution. |
+| *subject* | yes | *properties* | *string* | MUST contain the command to be executed by a device. |
```
sophokles73: What is the purpose of this property? Do we map this to a transport protocol specific property in the protocol adapters?

site/content/api/Command-And-Control-API.md
```diff
@@ -12,43 +12,39 @@ THIS IS A FIRST DRAFT VERSION OF THE API SPEC AND THUS STILL SUBJECT TO DISCUSSI
...
+| *content-type* | no | *properties* | *symbol*  | SHOULD be set to *application/octet-stream* if the message payload is to be considered *opaque* binary data. In most cases though, the client should be able to set a more specific content type indicating the type and characteristics of the data contained in the payload, e.g. `text/plain; charset="utf-8"` for a text message or `application/json` etc. |
+| *ttl* | no | *properties* | *int* | Indicates the timeout value (in milliseconds) indicating when server can assume that command has not been executed because of the lack of the response from the client. |
```
sophokles73: what is the *server* here? is this the *sender* of the command (i.e. the application)? Or is this Hono? If so, what happens once the ttl has expired?

site/content/api/Command-And-Control-API.md
```diff
@@ -12,43 +12,39 @@ THIS IS A FIRST DRAFT VERSION OF THE API SPEC AND THUS STILL SUBJECT TO DISCUSSI
...
-LWM2M, etc). Devices and *Protocol Adapters* can also poll Hono periodically in order check for new command & control messages without
-maintaining a constant connection to Hono infrastructure.
+The Command & Control API is defined by means of AMQP 1.0 message exchanges, i.e. a client needs to connect to Hono using AMQP 1.0 in order to invoke operations of the API as described in the following sections. Throughout the remainder of this page we will simply use AMQP when referring to AMQP 1.0.
 
-The Command & Control API is defined by means of AMQP 1.0 message exchanges, i.e. a client needs to connect to Hono using an AMQP 1.0 client in 
-order to invoke operations of the API as described in the following sections.
+# Operations
 
-# Southbound Operations
+The following API defines operations that can be used by *Business Applications* to send control messages to *Protocol Adapters* (and by *Protocol Adapters* to receive them) as well as the response from the *Protocol Adapters* back to the *Business Applications*.
```
sophokles73: IMHO a *Business Application* uses this API to send a command to a device. An application should not need to know about the fact, that a thing as a *Protocol Adapter* exists, should it?

It is a little tricky to find the right scope here. My current feeling is that we do not have a southbound C & C API at all but that the involvement of the protocol adapters with the AMQP Network is merely an implementation detail. In the same spirit, we also do not define the northbound operations of the Event and Telemetry APIs as an interaction between a consuming application and the AMQP Network, do we?

site/content/api/Command-And-Control-API.md
```diff
@@ -57,22 +53,23 @@ After execution of a control message on a device, device MUST send a response me
 
 **Preconditions**
 
-1. Client has established an AMQP connection with Hono.
-2. Client has established an AMQP link in role *sender* with Hono using target address `control-reply/${tenant_id}/{$device_id}`. Where `${device_id}` is an ID of the
+1. *Protocol Adapter* and *Business Application* have established an AMQP connection with the AMQP 1.0 Network.
```
sophokles73: as stated above, I have strong doubts whether we actually want to define this implementation strategy as being the API.

site/content/api/Command-And-Control-API.md
```diff
@@ -57,22 +53,23 @@ After execution of a control message on a device, device MUST send a response me
...
 
-**Message Flow**
-
-*TODO* add sequence diagram
+(The message flow is shown in the sequence diagram above to show the whole sequence in one place)
 
 **Message Format**
 
-The following table provides an overview of the properties a client needs to set on a *Control* message.
+The following table provides an overview of the properties the *Protocol Adapter* needs to set on a *Control-Reply* message.
```
sophokles73: IMHO this should be the spec of the response message an application gets in return to sending a command.

site/content/api/Command-And-Control-API.md
pellmann: It is the command, that must be known by the device. It could work for itself or indicates what to do with the data of the command. How me map this to the device protocols is out of scope for this API, IMPOV. In MQTT this could be the last part of the topic.

site/content/api/Command-And-Control-API.md
pellmann: I agree that it is not clear. Since this is a standard property maybe reuse some of that description like:Duration in milliseconds for which the message is to be considered “live”. 
WDYT?

site/content/api/Command-And-Control-API.md
```diff
@@ -12,43 +12,39 @@ THIS IS A FIRST DRAFT VERSION OF THE API SPEC AND THUS STILL SUBJECT TO DISCUSSI
...
-
-## Receive Control Message
+## Send and Receive a Command
 
 **Preconditions**
 
-1. Client has established an AMQP connection with Hono.
-2. Client has established an AMQP link in role *receiver* with Hono using target address `control/${tenant_id}/${device_id}`. Where `${device_id}` is the ID of the device to receive control messages for and `${tenant_id}` is the ID of the tenant the device belongs to.
+1. *Protocol Adapter* and *Business Application* have established an AMQP connection with the AMQP 1.0 Network.
+2. *Protocol Adapter* has established an AMQP link in role *receiver* with the AMQP 1.0 Network using target address `control/${tenant_id}/${device_id}`. Where `${device_id}` is the ID of the device to receive control messages for and `${tenant_id}` is the ID of the tenant the device belongs to.
```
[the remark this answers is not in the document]
pellmann: Thats true for the AMQP links. For a typical reader I find it counter-intuitiv to call this source address, since it is the address to which the sender wants to send the message. But it is also confusing to call it target address, because that is wrong in terms of AMQP. Maybe I just write 'address'?

pellmann:
I changed the descriptions, to show the API strictly from an applications point of view.

site/content/api/Command-And-Control-API.md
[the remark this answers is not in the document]
sophokles73: Since the whole API is defined in terms of AMQP 1.0, I would really appreciate if we could use at least the correct terminology. BTW we are already doing so in all other APIs.

site/content/api/Command-And-Control-API.md
sophokles73: My question was more if we make sure that this information is conveyed to the device so that it *can* make use of it.

site/content/api/Command-And-Control-API.md
```diff
@@ -16,19 +16,18 @@ The Command & Control API is defined by means of AMQP 1.0 message exchanges, i.e
 
 # Operations
 
-The following API defines operations that can be used by *Business Applications* to send control messages to *Protocol Adapters* (and by *Protocol Adapters* to receive them) as well as the response from the *Protocol Adapters* back to the *Business Applications*.
+The following API defines operations that can be used by *Business Applications* to send control messages to devices over an *AMQP 1.0 Network* and how to get the response.
```
sophokles73: what about

> ... to send commands and control messages to devices over an *AMQP 1.0 Network* and get a response in return.

site/content/api/Command-And-Control-API.md
```diff
@@ -16,19 +16,18 @@ The Command & Control API is defined by means of AMQP 1.0 message exchanges, i.e
...
-## Send and Receive a Command
+## Send a Command
 
 **Preconditions**
 
-1. *Protocol Adapter* and *Business Application* have established an AMQP connection with the AMQP 1.0 Network.
-2. *Protocol Adapter* has established an AMQP link in role *receiver* with the AMQP 1.0 Network using target address `control/${tenant_id}/${device_id}`. Where `${device_id}` is the ID of the device to receive control messages for and `${tenant_id}` is the ID of the tenant the device belongs to. 
-3. *Business Application* has established an AMQP link in role *sender* (also using target address `control/${tenant_id}/${device_id}`). 
+1. The *Business Application* has established an AMQP connection with the AMQP 1.0 Network.  
+2. *Business Application* has established an AMQP link in role *sender* with the address `control/${tenant_id}/${device_id}`. Where `${device_id}` is the ID of the device to send control messages for and `${tenant_id}` is the ID of the tenant the device belongs to.
```
sophokles73: target address

sophokles73: what about

> ... is the ID of the device to send messages to and `${tenant_id}` is the ID of the tenant that the device belongs to.

site/content/api/Command-And-Control-API.md
```diff
@@ -16,19 +16,18 @@ The Command & Control API is defined by means of AMQP 1.0 message exchanges, i.e
...
-2. *Protocol Adapter* has established an AMQP link in role *receiver* with the AMQP 1.0 Network using target address `control/${tenant_id}/${device_id}`. Where `${device_id}` is the ID of the device to receive control messages for and `${tenant_id}` is the ID of the tenant the device belongs to. 
-3. *Business Application* has established an AMQP link in role *sender* (also using target address `control/${tenant_id}/${device_id}`). 
+1. The *Business Application* has established an AMQP connection with the AMQP 1.0 Network.  
+2. *Business Application* has established an AMQP link in role *sender* with the address `control/${tenant_id}/${device_id}`. Where `${device_id}` is the ID of the device to send control messages for and `${tenant_id}` is the ID of the tenant the device belongs to. 
 
 **Link establishment and Message Flow**
 
-The following sequence diagram shows the whole message exchange, together with the reply from the *Protocol Adapter* to the *Business Application* (which is defined above in detail).
+The following sequence diagram shows the whole message exchange, together with the reply to the *Business Application*.
```
sophokles73: what about

> The following sequence diagram shows the whole message exchange including the response being sent back to the *Business Application*.

site/content/api/Command-And-Control-API.md
```diff
@@ -40,33 +39,24 @@ The following table provides an overview of the properties the *Business Applica
...
 | *subject* | yes | *properties* | *string* | MUST contain the command to be executed by a device. |
-| *content-type* | no | *properties* | *symbol*  | SHOULD be set to *application/octet-stream* if the message payload is to be considered *opaque* binary data. In most cases though, the client should be able to set a more specific content type indicating the type and characteristics of the data contained in the payload, e.g. `text/plain; charset="utf-8"` for a text message or `application/json` etc. |
-| *ttl* | no | *properties* | *int* | Indicates the timeout value (in milliseconds) indicating when server can assume that command has not been executed because of the lack of the response from the client. |
+| *content-type* | no | *properties* | *symbol*  | SHOULD be set to *application/octet-stream* if the message payload is to be considered *opaque* binary data. In most cases though, the sender should be able to set a more specific content type indicating the type and characteristics of the data contained in the payload, e.g. `text/plain; charset="utf-8"` for a text message or `application/json` etc. |
```
sophokles73: I have the same question as for *ttl*: is this information conveyed to the device?

site/content/api/Command-And-Control-API.md
```diff
@@ -40,33 +39,24 @@ The following table provides an overview of the properties the *Business Applica
...
+| *ttl* | no | *properties* | *int* | Duration in milliseconds for which the message is to be considered 'live'. Messages that live longer than their expiration time will be discarded (or dead lettered). |
```
sophokles73: I still don't understand the purpose of this property. Does this indicate that Hono is supposed to *buffer* the request for that amount of time? Is this the minimum (?) time that Hono should wait for a reply from the device until it returns an error to the application?

site/content/api/Command-And-Control-API.md
```diff
@@ -40,33 +39,24 @@ The following table provides an overview of the properties the *Business Applica
...
-**Preconditions**
-
-1. *Protocol Adapter* and *Business Application* have established an AMQP connection with the AMQP 1.0 Network.
-2. *Protocol Adatper* has established an AMQP link in role *sender* with Hono using target address `control-reply/${tenant_id}/{$device_id}`. Where `${device_id}` is an ID of the
-device that received the control message and `${tenant_id}` is the ID of the tenant the device belongs to.
-3. *Business Application* has established an AMQP link in role *receiver* with the AMQP 1.0 Network. 
+## Receiving the Commands execution result
 
-(The message flow is shown in the sequence diagram above to show the whole sequence in one place)
+After execution of a control message on a device, the device MUST send a response message indicating a status of that execution.
```
sophokles73: is there a difference between a *command* and a *control* message? If so, IMHO we need to define these terms at the top of the document. It is quite confusing to read about *command* in one place and *control message* in the other without knowing whether these terms are synonyms to each other not.

site/content/api/Command-And-Control-API.md
```diff
@@ -40,33 +39,24 @@ The following table provides an overview of the properties the *Business Applica
...
 
 **Message Format**
 
-The following table provides an overview of the properties the *Protocol Adapter* needs to set on a *Control-Reply* message.
+The following table provides an overview of the properties set on a *Control-Reply* message.
 
 | Name | Mandatory | Location | Type | Description |
 | :--- | :-------: | :------- | :-------- | :---------- |
-| *correlation-id* | yes | *properties* | *string* | MUST contain the correlation ID used to match incoming control message with the response message containing result of execution. |
+| *correlation-id* | yes | *properties* | *string* | MUST contain the correlation ID used to match control message with the response message containing result of execution. |
```
sophokles73: what about

> MUST contain the correlation ID used to match the control message with the response message containing the result of execution on the device.

site/content/api/Command-And-Control-API.md
pellmann: There is no difference. As it was in the existing API description I understood it as sending a command which includes a control message. But I am open to always speak of a command.

sophokles73: I think that would be a good idea and make the whole document more consistent.

FMPOV we should also explain the difference between commands to which the application expects a response vs. *one-way* commands and if and how an application knows whether the command has been delivered to the device.

site/content/api/Command-And-Control-API.md
pellmann: I introduced this at the time I defined the API between Business Application and the Adapter to show, that this standard AMQP 1.0 properties are there and could be used. Now it is between Business Application and a device - without a concrete idea of all device specific implementations of this side of the communication. So it might be better to not list them here since I am not sure at the moment if they will be forwarded to the device or only interpreted by the adapter. So I would remove ttl and content-type from the API definition. WDYT?

site/content/api/Command-And-Control-API.md
pellmann: I decided to just start with commands, which always expects a response.

site/content/api/Command-And-Control-API.md
sophokles73: Unless we really know what they are needed for, yes, I agree, we should remove them for the time being.

site/content/api/Command-And-Control-API.md
pellmann: The response message, send by the device which at least includes the status code.

site/content/api/Command-And-Control-API.md
```diff
@@ -16,50 +16,46 @@ The Command & Control API is defined by means of AMQP 1.0 message exchanges, i.e
...
+The following API defines operations that can be used by *Business Applications* to send commands to devices over an AMQP 1.0 Network and get a response in return.
 
 ## Send a Command
 
 **Preconditions**
 
-1. The *Business Application* has established an AMQP connection with the AMQP 1.0 Network.  
-2. *Business Application* has established an AMQP link in role *sender* with the address `control/${tenant_id}/${device_id}`. Where `${device_id}` is the ID of the device to send control messages for and `${tenant_id}` is the ID of the tenant the device belongs to. 
+1. *Business Application* has established an AMQP connection with the AMQP 1.0 Network.  
+2. *Business Application* has established an AMQP link in role *sender* with the target address `control/${tenant_id}/${device_id}`. Where `${device_id}` is the ID of the device to send messages to and `${tenant_id}` is the ID of the tenant that the device belongs to.
```
sophokles73: The application also needs to open a receiver link on the *reply-to* source address in order to receive the response, doesn't it?

site/content/api/Command-And-Control-API.md
```diff
@@ -16,50 +16,46 @@ The Command & Control API is defined by means of AMQP 1.0 message exchanges, i.e
...
-| *message-id* | yes | *properties* | *string* | MUST contain the message ID used to match incoming control message with the response message containing result of execution. |
-| *subject* | yes | *properties* | *string* | MUST contain the command to be executed by a device. |
-| *content-type* | no | *properties* | *symbol*  | SHOULD be set to *application/octet-stream* if the message payload is to be considered *opaque* binary data. In most cases though, the sender should be able to set a more specific content type indicating the type and characteristics of the data contained in the payload, e.g. `text/plain; charset="utf-8"` for a text message or `application/json` etc. |
-| *ttl* | no | *properties* | *int* | Duration in milliseconds for which the message is to be considered 'live'. Messages that live longer than their expiration time will be discarded (or dead lettered). |
+| *message-id* | yes | *properties* | *string* | MUST contain the message ID used to match incoming command message with the response message containing the result of execution. |
+| *subject* | yes | *properties* | *string* | MUST contain the command name to be executed by a device. |
```
sophokles73: what about *reply-to*?

site/content/api/Command-And-Control-API.md
```diff
@@ -16,50 +16,46 @@ The Command & Control API is defined by means of AMQP 1.0 message exchanges, i.e
...
 
 | Name | Mandatory | Location | Type      | Description |
 | :--- | :-------: | :------- | :-------- | :---------- |
-| *message-id* | yes | *properties* | *string* | MUST contain the message ID used to match incoming control message with the response message containing result of execution. |
-| *subject* | yes | *properties* | *string* | MUST contain the command to be executed by a device. |
-| *content-type* | no | *properties* | *symbol*  | SHOULD be set to *application/octet-stream* if the message payload is to be considered *opaque* binary data. In most cases though, the sender should be able to set a more specific content type indicating the type and characteristics of the data contained in the payload, e.g. `text/plain; charset="utf-8"` for a text message or `application/json` etc. |
-| *ttl* | no | *properties* | *int* | Duration in milliseconds for which the message is to be considered 'live'. Messages that live longer than their expiration time will be discarded (or dead lettered). |
+| *message-id* | yes | *properties* | *string* | MUST contain the message ID used to match incoming command message with the response message containing the result of execution. |
```
sophokles73: do we also allow to specify a separate *correlation-id* like we do in the other request/response APIs?

site/content/api/Command-And-Control-API.md
```diff
@@ -16,50 +16,46 @@ The Command & Control API is defined by means of AMQP 1.0 message exchanges, i.e
...
 
 ## Receiving the Commands execution result
 
-After execution of a control message on a device, the device MUST send a response message indicating a status of that execution.
+After execution of a command on a device, the device MUST send a response message indicating a status of that execution.
 
 **Message Format**
 
-The following table provides an overview of the properties set on a *Control-Reply* message.
+The following table provides an overview of the properties set on a Commands response message.
```
sophokles73: should be *command's*

site/content/api/Command-And-Control-API.md
```diff
@@ -72,4 +68,4 @@ The status property must contain a valid HTTP status code:
 
 For status codes indicating an error (codes in the `400 - 499` range) the message body MAY contain a detailed description of the error that occurred.
 
-If a control message response contains some additional diagnostic information, the body of the message MUST consist of a single AMQP *Data* section containing a control message data. 
+If a command message response contains some additional diagnostic information, the body of the message MUST consist of a single AMQP *Data* section containing the response message data.
```
sophokles73: I wouldn't call it *diagnostic information* but simply the response payload, right?

site/content/api/Command-And-Control-API.md
pellmann: I agree, that we should handle this consistent. Why do we allow this in the other API's? Is there a need for an own ID?

sophokles73: IIRC we wanted to make sure that a client can exlicitly set a correlation ID on an outgoing message in order to prevent a race condition when matching the response. This problem could occur when sending a request using JMS where the message ID cannot be set by the client code but is set by the JMS provider and thus is only available *after* the message has been sent ...

pellmann: I see, that makes sense - so I will add it to Command and Control.

pellmann:
Integrated all comments.

site/content/api/Command-And-Control-API.md
```diff
@@ -12,66 +12,52 @@ THIS IS A FIRST DRAFT VERSION OF THE API SPEC AND THUS STILL SUBJECT TO DISCUSSI
...
-The following operations can be used by *Devices* and/or *Protocol Adapters* to receive command & control messages sent by *Business Applications*.
+## Send a Command
 
-## Receive Control Message
+### Preconditions
 
-**Preconditions**
+1. *Business Application* has established an AMQP connection with the AMQP 1.0 Network.  
+2. *Business Application* has established an AMQP link in role *sender* with the target address `control/${tenant_id}/${device_id}`. Where `${device_id}` is the ID of the device to send messages to and `${tenant_id}` is the ID of the tenant that the device belongs to.
+3. *Business Application* has established an AMQP link in role *receiver* with the source address `control-reply/${tenant_id}/${device_id}`. Where `${device_id}` is the ID of the device to receive the messages from and `${tenant_id}` is the ID of the tenant that the device belongs to.
```
sophokles73: this sounds like the reply-to address is fixed, which FMPOV will not work in a scenario where multiple applications want to send a command to the same device because the response(s) will be delivered to *any one* of the applications (*balanced* message distribution). IMHO the client should therefore use an individual reply-to address that contains an element of uniqueness, e.g. a UUID, as we demand for the other request response APIs.

site/content/api/Command-And-Control-API.md
```diff
@@ -12,66 +12,52 @@ THIS IS A FIRST DRAFT VERSION OF THE API SPEC AND THUS STILL SUBJECT TO DISCUSSI
...
-message data. The format and encoding of the data is indicated by the *command*. It is up to the device to interpret the incoming body by taking `command` into 
-consideration.
-
-## Sending back control message execution result
-
-After execution of a control message on a device, device MUST send a response message indicating a status of that execution.
-
-**Preconditions**
+| *subject*        | yes       | *properties*             | *string*     | MUST contain the command name to be executed by a device. |
+| *correlation-id* | no        | *properties*             | *message-id* | MAY contain an ID used to correlate a response message to the original request. If set, it is used as the *correlation-id* property in the response, otherwise the value of the *message-id* property is used. |
```
sophokles73: this property has no default value in the request. The request's message-id is used as the correlation-id in the *response*, if no specific *correlation-id* has been set in the request.

site/content/api/Command-And-Control-API.md
```diff
@@ -12,66 +12,52 @@ THIS IS A FIRST DRAFT VERSION OF THE API SPEC AND THUS STILL SUBJECT TO DISCUSSI
...
+| *message-id*     | yes       | *properties*             | *string*     | MUST contain an identifier that uniquely identifies the message at the sender side. |
+| *reply-to*       | yes       | *properties*             | *string*     | MUST contain the source address that the client wants to received response messages from. This address MUST be the same as the source address used for establishing the client's receive link (see [Preconditions]({{< relref "#preconditions" >}})). |
 
-1. Client has established an AMQP connection with Hono.
-2. Client has established an AMQP link in role *sender* with Hono using target address `control-reply/${tenant_id}/{$device_id}`. Where `${device_id}` is an ID of the
-device that received the control message and `${tenant_id}` is the ID of the tenant the device belongs to.
+If a command message requires additional data to be properly executed (like command parameters), the body of the message MUST consist of a single AMQP *Data* section containing a control 
+message data. The format and encoding of the data is indicated by the *subject*. It is up to the device to interpret the incoming body by taking it into 
+consideration.
```
sophokles73: what about
> A command message MAY contain arbitrary payload to be sent to the device in an AMQP *Data* section. The value of the command message's *subject* value may provide a hint to the device regarding the format, encoding and semantics of the payload data.

site/content/api/Command-And-Control-API.md
```diff
@@ -12,66 +12,52 @@ THIS IS A FIRST DRAFT VERSION OF THE API SPEC AND THUS STILL SUBJECT TO DISCUSSI
...
 
-**Message Flow**
+## Receiving the Command's execution result
 
-*TODO* add sequence diagram
+After execution of a command on a device, the device MUST send a response message indicating a status of that execution.
```
sophokles73: ... indicating the outcome of executing the command ?

site/content/api/Command-And-Control-API.md
pellmann: That's right and I do not see a reason to not use a UUID. I think in the beginning with potential queuing for a longer time we had an other expectation but now this fits well.

site/content/api/Command-And-Control-API.md
pellmann: I do not really understand your point?

site/content/api/Command-And-Control-API.md
pellmann: OK

site/content/api/Command-And-Control-API.md
pellmann: OK

site/content/api/Command-And-Control-API.md
sophokles73: the table defines the properties that can be contained in a command request, right? If so, then the *correlation-id* has no default value. The request message's *correlation-id* property is empty by default. The device, however, **MUST** set the *correlation-id* property of the response message to the request's *message-id* if the request **did not** contain a *correlation-id*.

sophokles73: the row for the *correlation-id* contains *message-id* in the fourth column, which is used to indicate the default value if not set explicitly ...

pellmann: No, it is the type column.

site/content/api/Command-And-Control-API.md
sophokles73: So, does this mean that you are going to update this sentence accordingly?

site/content/api/Command-And-Control-API.md
```diff
@@ -12,66 +12,50 @@ THIS IS A FIRST DRAFT VERSION OF THE API SPEC AND THUS STILL SUBJECT TO DISCUSSI
...
-|*durable* | no | *properties* | *boolean* | Indicates if the message should be persisted (*store and forward*) by the underlying message transport middleware. By default command&control messages are durable.  |
+| *subject*        | yes       | *properties*             | *string*     | MUST contain the command name to be executed by a device. |
+| *correlation-id* | no        | *properties*             | *message-id* | MAY contain an ID used to correlate a response message to the original request. If set, it is used as the *correlation-id* property in the response, otherwise the value of the *message-id* property is used. |
+| *message-id*     | yes       | *properties*             | *string*     | MUST contain an identifier that uniquely identifies the message at the sender side. |
+| *reply-to*       | yes       | *properties*             | *string*     | MUST contain the source address that the client wants to received response messages from. This address MUST be the same as the source address used for establishing the client's receive link (see [Preconditions]({{< relref "#preconditions" >}})). |
```
sophokles73: ... wants to receive response messages ...

site/content/api/Command-And-Control-API.md
```diff
@@ -12,66 +12,50 @@ THIS IS A FIRST DRAFT VERSION OF THE API SPEC AND THUS STILL SUBJECT TO DISCUSSI
...
 
 **Message Format**
 
-The following table provides an overview of the properties a client needs to set on a *Control* message.
+The following table provides an overview of the properties set on a Command's response message.
 
 | Name | Mandatory | Location | Type | Description |
 | :--- | :-------: | :------- | :-------- | :---------- |
-| *correlation-id* | yes | *properties* | UTF-8 *string* | MUST contain the correlation ID used to match incoming control message with the response message containing result of execution. |
+| *correlation-id* | yes | *properties* | *string* | MUST contain the correlation ID used to match the command message with the response message containing the result of execution on the device. |
```
sophokles73: in order to be consistent with the table for the request, the type should probably be *message-id*, right?

site/content/api/Command-And-Control-API.md
```diff
@@ -12,66 +12,50 @@ THIS IS A FIRST DRAFT VERSION OF THE API SPEC AND THUS STILL SUBJECT TO DISCUSSI
...
 | *status* | yes | *application-properties* | *integer* | MUST indicates the status of the execution. See table below for possible values. |
```
sophokles73: MUST indicate the status ...

site/content/api/Command-And-Control-API.md
```diff
@@ -85,25 +69,4 @@ The status property must contain a valid HTTP status code:
```
sophokles73: what is the code returned if the device is not connected? what about 503 ?

pellmann: Could that happen? So the device is disconnected after sending the command and before the adapter has send it to the device - in my picture this part is in "sync" so if the adapter could not forward the command the disposition would not be successful.

sophokles73: this seems to be part of the question of how an application can know, if its request has reached the device or not. With our current implementation strategy, the application would not get any credits on its sender link, right? If this is the *specified behavior* then we need to put it into the spec.

Issue #586, opened 2018-04-18
http-client OSGi bundle doesn't allow instantiation of HonoClient


avgustinmm:
AFAIK the only way to use HoneClient is to instantiate it via calling the direct constructor of org.eclipse.hono.client.impl.HonoClientImpl.
The org.eclipse.hono.client.impl, however, is not expored by the hono-client bundle. Only the org.eclipse.hono.client is exported.
So, the hono-client shall export org.eclipse.hono.client.impl (exporting impl, IMHO, is not recommended) or the org.eclipse.hono.client package shall provide meant to instantiate HonoClient without directly refer org.eclipse.hono.client.impl

ctron:
I will look into this.

sophokles73:
Maybe we should add a default factory method to the `HonoClient` interface, very much like vert.x does it all over the place ...

avgustinmm:
Yes, this will solve the issue and will improve even non-OSGi usage. Then there will be no need of using impl classes directly - better API

sophokles73:
I was more thinking about just adding something like the following to `HonoClient`

```
HonoClient createClient(Vertx vertx, ClientConfigProperties config) {
  return new HonoClientImpl(vertx, config);
}
```

I do not think that we need to change anything else for this, do we?

ctron:
Well that would ignore the case of using an explicit `ConnectionFactory`.

ctron:
I just pushed the (unfinished!) branch of the changes I had in mind.

ctron:
Ok, pushed again without the code formatter changes.

ctron:
The javadoc part and the refactoring of the "ConnectionBuilder" is still missing of course.

ctron:
This should be fixed by #593

PR #593, opened 2018-04-20
#586 Fix OSGi client instantiation

ctron:
This PR fixes the OSGi client instatiation issue reported in #586. I also aligns both pattern of the `HonoClient` and `ConnectionFactory`.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/HonoClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/HonoClient.java
@@ -17,10 +17,12 @@ import java.util.function.Consumer;
 
 import org.apache.qpid.proton.amqp.Symbol;
 import org.apache.qpid.proton.message.Message;
-
+import org.eclipse.hono.client.impl.HonoClientImpl;
+import org.eclipse.hono.config.ClientConfigProperties;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
+import io.vertx.core.Vertx;
 import io.vertx.proton.ProtonClientOptions;
 import io.vertx.proton.ProtonConnection;
 import io.vertx.proton.ProtonDelivery;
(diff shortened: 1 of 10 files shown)


client/src/main/java/org/eclipse/hono/client/HonoClient.java
```diff
@@ -281,4 +286,109 @@
      *         AMQP <em>open</em> frame, {@code false} otherwise.
      */
     boolean supportsCapability(Symbol capability);
+
+    /**
+     * Builder for for HonoClient instances.
+     * <p>
+     * This class is not thread safe.
+     */
+    class HonoClientBuilder {
```
sophokles73: what value does this builder add? it accepts the same properties as the standard constructor ...

core/src/main/java/org/eclipse/hono/connection/ConnectionFactory.java
```diff
@@ -88,4 +93,168 @@ void connect(
             Handler<AsyncResult<ProtonConnection>> closeHandler,
             Handler<ProtonConnection> disconnectHandler,
             Handler<AsyncResult<ProtonConnection>> connectionResultHandler);
+
+    /**
+     * Builder for ConnectionFactory instances.
+     */
+    final class ConnectionFactoryBuilder {
```
sophokles73: FMPOV we should rather get rid of the builder altogether since we do not really use it anyway. People usually create a `HonoClient` using the constructor that accepts a vert.x and `ClientConfigProperties` instance which creates the connection factory under the hood. This builder has been around since the beginning but turned out to be superfluous FMPOV.

client/src/main/java/org/eclipse/hono/client/HonoClient.java
ctron: Yes, but builder can be extended without breaking the API. Constructors cannot and thus you end up with a bunch of constructors over time. But yes, we could collapse the two builders into one.

core/src/main/java/org/eclipse/hono/connection/ConnectionFactory.java
ctron: See above. We could collapse both builders into one.

client/src/main/java/org/eclipse/hono/client/HonoClient.java
sophokles73: FMPOV the purpose of the *builder* is already covered by the config properties. This, BTW, seems to also be the way pursued by vert.x in all of its APIs ...

ctron: Ok, do you want me to drop both builders then?

sophokles73: I'd appreciate that

ctron:
Ok, I did drop the builders in favor of two "new" methods.

If that is ok, I would squash the commits before merging.

Issue #570, opened 2018-04-11
Add OpenShift S2I deployment as example


ctron:
I would like to add the OpenShift S2I templates I did create for Hono 0.5 to the Hono master branch. The 0.5 version of this can be seen in the feature branch at my cloned repository: https://github.com/ctron/hono/tree/feature/support_s2i_05x

I do understand that this duplicates code which currently is in a different set of YAML files in different locations of Hono. However I do see this single template more like one big example deployment scenario for Hono.

I am willing to maintain this and ensure that this is working at least for every final release.

In the provided feature branch the code is located under the `openshift` folder in the root of the repository, I would migrate this to `hono/example/src/main/deploy/openshift_s2i`.

Let me know what you think. Any objections or suggestions?

ctron:
Closed by PR #594

sophokles73:
@ctron can we close this issue?

PR #594, opened 2018-04-23
[#570] Add OpenShift S2I deployment

ctron:


site/content/deployment/openshift_s2i.md
```diff
@@ -0,0 +1,358 @@
...
+All examples in the following sub-sections assume that you are located in the
+`examples` directory.
+
+### Extract certificates
+
+In order to connect the external consumer to EnMasse, we need to use a
+proper SSL certificate. We can extract one from the OpenShift using the
+following command:
+
+    oc extract secret/external-certs-messaging --to=target/keys -n enmasse
```
dejanb: target/keys doesn't work by default ... maybe using target/config/hono-demo-certs-jar/ as we do in other document is a better option

site/content/deployment/openshift_s2i.md
```diff
@@ -34,6 +34,21 @@ the version of your OpenShift cluster. This guide was tested with
 OpenShift 3.7.2. It should work with older or newer versions as well, but that
 is untested.
 
+{{% warning title="Curl on Mac OS X" %}}
+The `curl` binary on Mac OS X suffers from an issue with TLS SNI. Also see: https://github.com/curl/curl/issues/1533
+
+As the use of SNI is required for Kubernetes/OpenShift when it comes to routing
```
sophokles73: My understanding is that this is just an issue on OpenShift when using routes (which do not exist on Kubernetes, do they?)

ctron: This effects all cases where TLS SNI is used. In the issue mentioned I saw some references to kubernetes as well. As I don't have any hands on experience with plain kubernetes, I don't know. I can remove it, if you think it is better.

site/content/deployment/openshift_s2i.md
ctron: Well the following commands all have the changed paths. Which makes the commands a bit shorter. But I agree, I you want to re-use the commands from the other tutorials, you might run into issued. I will sync this with the other examples.

Issue #500, opened 2018-02-21
Add HonoClient#disconnect() method


calohmn:
Currently there is a `HonoClient#connect()` method, but no `HonoClient#disconnect()` method.

A "disconnect" method is for example needed in a scenario where the open frame of the connection contained permissions from an authorization service and these permissions in the service have changed since opening the connection, making it necessary to disconnect and re-open the connection.

I could imagine a simple method with no parameters and return value here:
``
void disconnect();
``
Notification of a finished disconnect will be done on the `disconnectHandler` supplied to the `connect` method.

sophokles73:
What would/should be the difference to invoking `HonoClient.shutdown()`?

calohmn:
A `HonoClient.shutdown()` invocation causes all subsequent `HonoClient.connect` invocations to fail. I.e. the HonoClient can't be used anymore after that.

A `disconnect` should allow further `connect` invocations.

sophokles73:
should be fixed in 1550e8950be4f695cce929584fd35361fe0b039c

@calohmn can you verify and then close this issue, if it works for you?

calohmn:
Re-test successful.

What is perhaps somewhat unexpected, is that the `disconnect completionHandler` is called directly after having sent the "close" frame, without waiting for a "close" frame response or for the underlying transport connection to be disconnected.

Log output where the `disconnect completionHandler` is doing a new `connect` - see the late `closed connection to server` log line:
````
16:08:25.569 INFO  o.e.hono.example.ExampleReceiver - ##### call disconnect
16:08:25.574 INFO  o.e.hono.client.impl.HonoClientImpl - closing connection to server [192.168.99.100:30071]...

  16:08:25.576 INFO  o.e.hono.example.ExampleReceiver - ##### disconnect completionHandler succeeded; now reconnect...

16:08:25.582 DEBUG o.e.hono.client.impl.HonoClientImpl - starting attempt [#1] to connect to server [192.168.99.100:30071]
16:08:25.582 DEBUG o.e.h.config.ClientConfigProperties - using certificates from file [[..]trusted-certs.pem] as trust anchor
16:08:25.583 DEBUG o.e.h.c.impl.ConnectionFactoryImpl - enabling secure protocol [TLSv1.2]
16:08:25.583 DEBUG o.e.h.c.impl.ConnectionFactoryImpl - connecting to AMQP 1.0 container [amqps://192.168.99.100:30071]

[...]

PR #595, opened 2018-04-23
Provide support for HonoClient#disconnect()

Alfusainey:
Initial implementation of issue #500. the implementation mimics how the `shutdown` call is implemented.
@sophokles73 @calohmn

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/HonoClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/HonoClient.java
@@ -67,6 +67,30 @@ public interface HonoClient {
     Future<HonoClient> connect();
 
     /**
+     * Disconnects the connection to the Hono server. Upon terminating the connection to the server,
+     * this method does not automatically try to reconnect to the server again. To connect to the server,
+     * an explicit call to {@code HonoClient#connect()} should be made. Unlike {@code HonoClient#shutdown()},
+     * which does not allow to connect back to the server, this method allows to connect back to the server.
+     *
+     * Disconnecting from the Hono server is necessary when, for instance, the open frame of the connection contains
+     * permission information from an authorization service. If after connecting to the server the permissions
+     * from the service have changed, then it will be necessary to drop the connection and connect back to the server
+     * to retrieve the updated permissions.
+     *
+     */
+    void disconnect();
+
+    /**
+     * Similar to {@code HonoClient#disconnect()} but takes a handler to notify the caller about the result
+     * of the disconnect operation. The caller can use the handler to determine if the operation succeeded or failed.
+     *
+     * @param completionHandler The completion handler to notify about the success or failure of the operation. A failure could occur
+     * if this method is called in the middle of a disconnect operation.
+     * @throws NullPointerException if the completionHandler is {@code null}.
+     */
+    void disconnect(Handler<AsyncResult<Void>> completionHandler);
+
+    /**
      * Connects to the Hono server using given options.
      * <p>
      * The number of times that the client tries to (re-)connect to the peer is determined by the
--- a/client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
@@ -73,6 +73,7 @@ public class HonoClientImpl implements HonoClient {
     private final List<Handler<Void>> creationRequests = new ArrayList<>();
     private final AtomicBoolean connecting = new AtomicBoolean(false);
     private final AtomicBoolean shuttingDown = new AtomicBoolean(false);
+    private final AtomicBoolean disconnecting = new AtomicBoolean(false);
     private final ConnectionFactory connectionFactory;
     protected final ClientConfigProperties clientConfigProperties;
     private final Vertx vertx;
@@ -953,38 +954,78 @@ public class HonoClientImpl implements HonoClient {
      */
     @Override
     public final void shutdown(final Handler<AsyncResult<Void>> completionHandler) {
-
+        Objects.requireNonNull(completionHandler);
         if (shuttingDown.compareAndSet(Boolean.FALSE, Boolean.TRUE)) {
-            context.runOnContext(shutDown -> {
-                if (isConnectedInternal()) {
-                    shutdownConnection(completionHandler);
-                } else {
-                    LOG.info("connection to server [{}:{}] already closed", connectionFactory.getHost(),
-                            connectionFactory.getPort());
-                    completionHandler.handle(Future.succeededFuture());
-                }
+            context.runOnContext(shutDownResult -> {
+                closeConnection(completionHandler);
             });
         } else {
-            completionHandler.handle(Future.failedFuture(new IllegalStateException("already shutting down")));
+            completionHandler.handle(Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_CONFLICT, "already in the middle of a shutdown operation")));
         }
     }
 
-    private void shutdownConnection(final Handler<AsyncResult<Void>> completionHandler) {
-
-        LOG.info("closing connection to server [{}:{}]...", connectionFactory.getHost(), connectionFactory.getPort());
-        connection.disconnectHandler(null); // make sure we are not trying to re-connect
-        connection.closeHandler(closedCon -> {
-            if (closedCon.succeeded()) {
-                LOG.info("closed connection to server [{}:{}]", connectionFactory.getHost(),
-                        connectionFactory.getPort());
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public final void disconnect() {
+        final CountDownLatch countDown = new CountDownLatch(1);
+        disconnect(disconnectResult -> {
+            if (disconnectResult.succeeded()) {
+                LOG.info("successfully disconnected from the server [{}:{}]", connectionFactory.getHost(), connectionFactory.getPort());
+                countDown.countDown();
             } else {
-                LOG.info("closed connection to server [{}:{}]", connectionFactory.getHost(),
-                        connectionFactory.getPort(), closedCon.cause());
+                LOG.error("could not disconnect from the server [{}:{}]", connectionFactory.getHost(), connectionFactory.getPort());
             }
-            connection.disconnect();
-            if (completionHandler != null) {
-                completionHandler.handle(Future.succeededFuture());
+        });
+        try {
+            if (!countDown.await(5, TimeUnit.SECONDS)) {
+                LOG.error("Disconnecting from the server [{}:{}] timed out after 5 seconds", connectionFactory.getHost(), connectionFactory.getPort());
             }
-        }).close();
+        } catch (final InterruptedException e) {
+            Thread.currentThread().interrupt();
+        }
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public final void disconnect(final Handler<AsyncResult<Void>> completionHandler) {
+        Objects.requireNonNull(completionHandler);
+        if (disconnecting.compareAndSet(Boolean.FALSE, Boolean.TRUE)) {
+            context.runOnContext(disconnectResult -> {
+                closeConnection(completionHandler);
+            });
+        } else {
+            completionHandler.handle(Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_CONFLICT, "already in the middle of a disconnect operation")));
+        }
+    }
+
+    //-----------------------------------< private methods >---
+
+    private void closeConnection(final Handler<AsyncResult<Void>> completionHandler) {
+        if (isConnectedInternal()) {
+            LOG.info("closing connection to server [{}:{}]...", connectionFactory.getHost(), connectionFactory.getPort());
+            connection.disconnectHandler(null); // make sure we are not trying to re-connect
+            connection.closeHandler(closedCon -> {
+                if (closedCon.succeeded()) {
+                    LOG.info("closed connection to server [{}:{}]", connectionFactory.getHost(),
+                            connectionFactory.getPort());
+                } else {
+                    LOG.info("closed connection to server [{}:{}]", connectionFactory.getHost(),
+                            connectionFactory.getPort(), closedCon.cause());
+                }
+                connection.disconnect();
+                disconnecting.compareAndSet(Boolean.TRUE, Boolean.FALSE);
+
+                completionHandler.handle(Future.succeededFuture());
+            }).close();
+        } else {
+            LOG.info("connection to server [{}:{}] already closed", connectionFactory.getHost(), connectionFactory.getPort());
+            disconnecting.compareAndSet(Boolean.TRUE, Boolean.FALSE);
+            completionHandler.handle(Future.succeededFuture());
+        }
+    }
+
 }
--- a/client/src/test/java/org/eclipse/hono/client/impl/HonoClientImplTest.java
+++ b/client/src/test/java/org/eclipse/hono/client/impl/HonoClientImplTest.java
@@ -59,7 +59,7 @@ public class HonoClientImplTest {
      * Global timeout for each test case.
      */
     @Rule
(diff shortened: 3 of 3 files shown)


client/src/main/java/org/eclipse/hono/client/HonoClient.java
```diff
@@ -68,6 +68,18 @@
      */
     Future<HonoClient> connect();
 
+    /**
+     * Disconnects from the Hono server.
```
sophokles73: what is the purpose of *disconnecting*? We should provide some information to the client as to what might be possible reasons why a client would/should disconnect. It would also help to know if the client can be re-connected again once it has been disconnected.

client/src/main/java/org/eclipse/hono/client/HonoClient.java
```diff
@@ -68,6 +68,18 @@
...
+     */
+    void disconnect();
+
+    /**
+     * Disconnects from the Hono server.
+     *
+     * @param completionHandler The completion handler to invoke with the result of the operation.
```
sophokles73: what are the possible results?

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -907,8 +908,43 @@ public final void shutdown() {
...
+    public final void disconnect(final Handler<AsyncResult<Void>> completionHandler) {
+        closeConnection(completionHandler, false);
+    }
+
+    //-----------------------------------< private methods >---
+
+    private void closeConnection(final Handler<AsyncResult<Void>> completionHandler, final Boolean isShuttingDown) {
+
+        Objects.requireNonNull(completionHandler, "Result handler cannot be null");
+        AtomicBoolean flag = (isShuttingDown) ? shuttingDown : disconnecting;
```
sophokles73: what if *isShuttingDown* is `null`
flag should be declared final

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -907,8 +908,43 @@ public final void shutdown() {
...
+     * {@inheritDoc}
+     */
+    @Override
+    public final void disconnect(final Handler<AsyncResult<Void>> completionHandler) {
+        closeConnection(completionHandler, false);
+    }
+
+    //-----------------------------------< private methods >---
+
+    private void closeConnection(final Handler<AsyncResult<Void>> completionHandler, final Boolean isShuttingDown) {
```
sophokles73: what is `isShuttingDown` good for?

client/src/main/java/org/eclipse/hono/client/HonoClient.java
Alfusainey: thanks! `disconnecting` is a flag that indicates whether or not the client is in the middle of a disconnect operation. my understanding is that the `disconnect` method is called directly on a client instance to explicitly disconnect from the server (for reasons described in #500). 
i also think that a `HonoClient#disconnect()` call should be followed by a subsequent `HonoClient#connect()` if the client wants to connect back to the server after disconnecting. is that a valid assumption?

client/src/main/java/org/eclipse/hono/client/HonoClient.java
Alfusainey: the result should indicate whether the disconnect action succeeded or failed.

sophokles73: I know that. But the user of this API doesn't and should be informed about it by means of - you already guessed it - the JavaDoc comment ;-)

client/src/main/java/org/eclipse/hono/client/HonoClient.java
sophokles73: please document this in the JavaDoc comment so that users of the API have a chance to learn about this as well ...

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
Alfusainey: but the flag is already declared final or?

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
Alfusainey: to determine whether the `closeConnection` method is called during a `HonoClient#shutdown` call (i.e `isShuttingDown` is true) or `HonoClient#disconnect` call(i.e `isShuttingDown` is false). the `shuttingDown` global flag, once set to true (during shutdown), is never reset back again. this is the reason why a subsequent connect/reconnect always fails. but in the case of the disconnect, once the disconnect is successful we need to reset the flag back (to allow subsequent calls to disconnect again). so `isShuttingDown` determine whether the flag needs to be reset or not

client/src/main/java/org/eclipse/hono/client/HonoClient.java
Alfusainey: ok, thanks

client/src/main/java/org/eclipse/hono/client/HonoClient.java
Alfusainey: ok, thanks! will fix

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
calohmn: I think, deriving only one flag from `shuttingDown` and `disconnecting`, and using that flag for the subsequent check, could cause issues here and makes the code harder to understand.
For example the line below with the "already shutting down" exception doesn't fit if `isShuttingDown` is false and `disconnect()` is invoked concurrently.
On the other hand, calling `shutdown()` and subsequently having `disconnect()` be called, could trigger `shutdownConnection()` twice.

And I would rename the `isShuttingDown` method parameter (e.g. to `isShuttingDownParam`) to have it be more distinguishable from the `shuttingDown` field.

In general, having the `closeConnection()` method change the `shuttingDown` field could be considered an unexpected sideeffect of the method. So, I think it would be better to change the field inside the `shutdown()` method,

Alfusainey: thanks for the comments! i will update the PR

client/src/main/java/org/eclipse/hono/client/HonoClient.java
calohmn: I think it would also be important to write about the fact that - as it is implemented in this PR - none of the "usual" disconnectHandlers are called here. 

With "usual" I mean the `disconnectHandler` param given to the `connect()` method or, if that was not set, the implicit disconnectHandler that invokes `reconnect()` (see `handleConnectionLoss()`).

Alfusainey: ok thanks. however, i tried very hard to understand how can the `disconnectHandler` local parameter of the `connect()` method be invoked in the `disconnect()` method. if you consider the following usage of the honoclient: `connect()` -> `disconnect()` -> `connect()` -> `disconnect()`. i assume here that a call to `disconnect` does not automatically connects to the server until an explicit call to `connect()` is made.

could you please clarify?

calohmn: The reason why I suggested using the `disconnectHandler` supplied to the `connect` method in the #500 ticket text is that I considered it convenient for the API user to have the disconnect logic  (usually, I guess: reconnect and when done, re-establish AMQP-Links on the new connection) only be defined in one place.

But thinking more about it, I think that might not be intuitive (and more complicated in the HonoClientImpl, having to keep the `disconnectHandler` supplied to the `connect` method in a new HonoClientImpl field). 
That's why I prefer your approach now, letting the user supply a `disconnectHandler` to the `disconnect()` method.

Assuming that in most cases a HonoClient should always be connected and have AMQP-links established on the connection, I think it would be good to point out in the `disconnect` javadoc, that an appropriate "reconnect" logic (if needed) should be put in the `disconnectHandler` param of the `disconnect()` method.

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -916,19 +917,52 @@ public final void shutdown() {
...
+    @Override
+    public final void disconnect(final Handler<AsyncResult<Void>> completionHandler) {
+        if (disconnecting.compareAndSet(Boolean.FALSE, Boolean.TRUE)) {
+            context.runOnContext(disconnectResult -> {
+                disconnecting.compareAndSet(Boolean.TRUE, Boolean.FALSE);
+                closeConnection(completionHandler);
             });
         } else {
-            completionHandler.handle(Future.failedFuture(new IllegalStateException("already shutting down")));
+            completionHandler.handle(Future.failedFuture(new ClientErrorException(1, "already in the middle of a disconnect operation")));
```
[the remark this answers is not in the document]
Alfusainey: oops, good catch ;-) thanks

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -916,19 +917,52 @@ public final void shutdown() {
...
+        }
+    }
+
+    //-----------------------------------< private methods >---
+
+    private void closeConnection(final Handler<AsyncResult<Void>> completionHandler) {
+        if (isConnectedInternal()) {
+            shutdownConnection(completionHandler);
```
calohmn: I think the `shutdownConnection` code can be inlined here (and `shutdownConnection()`removed) to avoid naming confusion concerning these methods.

client/src/main/java/org/eclipse/hono/client/HonoClient.java
```diff
@@ -66,6 +66,29 @@
...
+    /**
+     * Disconnects the connection to the Hono server. Upon terminating the connection to the server,
+     * this method does not automatically try to reconnect to the server again. To connect to the server,
+     * an explicit call to {@code HonoClient#connect()} should be made. Unlike {@code HonoClient#shutdown()},
+     * which does not allow to connect back to the server, this method allows to connect back to the server.
+     *
+     * Disconnecting from the Hono server is necessary when, for instance, the open frame of the connection contains
+     * permission information from an authorization service. If after connecting to the server the permissions
+     * from the service has changed, then it will be necessary to drop the connection and connect back to the server
+     * to retrieved the updated permissions.
```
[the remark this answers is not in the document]
Alfusainey: :+1:

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -916,19 +917,52 @@ public final void shutdown() {
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public final void disconnect() {
+        disconnect(disconnectResult -> {
+            if (disconnectResult.succeeded()) {
+                LOG.info("successfully disconnected from the server");
+            } else {
+                LOG.error("could not disconnect from the server", disconnectResult.cause());
```
calohmn: Would be good to add Host and Port here, like in the line
`LOG.debug("already connected to server [{}:{}]", connectionFactory.getHost(), connectionFactory.getPort());`

Same applies to the log statement two lines above.

sophokles73:
@Alfusainey are you still working on this PR?

Alfusainey:
i paused working on it since the last comments. i will update the PR later this week

Alfusainey:
@sophokles73 pull request updated

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -953,38 +954,76 @@ public final void shutdown() {
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public final void disconnect(final Handler<AsyncResult<Void>> completionHandler) {
+        if (disconnecting.compareAndSet(Boolean.FALSE, Boolean.TRUE)) {
+            context.runOnContext(disconnectResult -> {
+                disconnecting.compareAndSet(Boolean.TRUE, Boolean.FALSE);
```
sophokles73: IMHO we need to move this into the handler passed into the *closeConnection* invocation.

client/src/test/java/org/eclipse/hono/client/impl/HonoClientImplTest.java
```diff
@@ -479,6 +479,26 @@ public void testConnectFailsAfterShutdown(final TestContext ctx) {
...
+    /**
+     * Verifies that if a client disconnects from the server, then an attempt to connect again will be succcessful.
+     *
+     * @param ctx The test execution context.
+     */
+    @Test
+    public void testConnectSucceedsAfterDisconnect(final TestContext ctx) {
+
+        // If a client disconnects from the server
+        Async disconnect = ctx.async();
```
sophokles73: final

client/src/test/java/org/eclipse/hono/client/impl/HonoClientImplTest.java
```diff
@@ -479,6 +479,26 @@ public void testConnectFailsAfterShutdown(final TestContext ctx) {
...
+        client.disconnect();
```
sophokles73: this needs to be the variant with a handler so that we know, when it's time to reconnect ...

Alfusainey:
this commit incorporates the latest comments

client/src/test/java/org/eclipse/hono/client/impl/HonoClientImplTest.java
```diff
@@ -59,7 +59,7 @@
      * Global timeout for each test case.
      */
     @Rule
-    public Timeout timeout = Timeout.seconds(5);
+    public Timeout timeout = new Timeout(2, TimeUnit.SECONDS);
```
sophokles73: is there a reason why you changed this?

client/src/main/java/org/eclipse/hono/client/HonoClient.java
```diff
@@ -66,6 +66,29 @@
...
+    void disconnect();
+
+    /**
+     * Similar to {@code HonoClient#disconnect()} but takes a handler to notify the caller about the result
+     * of the disconnect operation. The caller can use the handler to determine if the operation succeeded or failed.
+     *
+     * @param completionHandler The completion handler to notify about the success or failure of the operation. A failure could occur
+     * if this method is called in the middle of a disconnect operation.
+     */
+    void disconnect(Handler<AsyncResult<Void>> completionHandler);
```
sophokles73: it looks like completionHandler can also be `null`, or am I mistaken? If not, then you should document that in this case a `NullPointerException` is thrown so that users know that the parameter is mandatory

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -953,38 +954,77 @@ public final void shutdown() {
...
+                connection.disconnect();
+                disconnecting.compareAndSet(Boolean.TRUE, Boolean.FALSE);
+                if (completionHandler != null) {
+                    completionHandler.handle(Future.succeededFuture());
+                }
+            }).close();
+        } else {
+            LOG.info("connection to server [{}:{}] already closed", connectionFactory.getHost(), connectionFactory.getPort());
+            disconnecting.compareAndSet(Boolean.TRUE, Boolean.FALSE);
+            completionHandler.handle(Future.succeededFuture());
```
sophokles73: if completionHandler can indeed be null, then you should check for null here as well ..

client/src/test/java/org/eclipse/hono/client/impl/HonoClientImplTest.java
Alfusainey: no. this format works for me if i want to change the timeunit (during debugging). forgot to change it back afterwards

Alfusainey:
thanks. PR updated!

sophokles73:
@calohmn any additional comments?

calohmn:
@sophokles73 no, LGTM

PR #596, opened 2018-04-23
Allow devices to indicate readiness-to-receive-messages (impl).

sysexcontrol:
Http clients of the HTTP protocol adapter can set the `ttd` AMQP property
for any command by using the `hono-ttd` HTTP header (scaled in seconds).
Northbound applications can register a specific callback that is invoked
if at the current time the device indicated it will still be connected.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/EventConstants.java
+++ b/core/src/main/java/org/eclipse/hono/util/EventConstants.java
@@ -35,4 +35,10 @@ public final class EventConstants {
 
     private EventConstants() {
     }
+
+    /**
+     * The content type that is defined for empty events without any payload.
+     */
+    public static final String CONTENT_TYPE_EMPTY_NOTIFICATION = "application/vnd.eclipse-hono-empty-notification";
+
 }
--- a/core/src/main/java/org/eclipse/hono/util/MessageHelper.java
+++ b/core/src/main/java/org/eclipse/hono/util/MessageHelper.java
@@ -56,6 +56,15 @@ public final class MessageHelper {
      */
     public static final String APP_PROPERTY_DEVICE_ID              = "device_id";
     /**
+     * The name of the AMQP 1.0 message application property containing the time until disconnect of the device that is available
+     * for receiving an upstream message for the given number of seconds (short for <em>Time Til Disconnect</em>).
+     */
+    public static final String APP_PROPERTY_DEVICE_TTD              = "ttd";
+    /**
+     * Devices that are always ready to receive an upstream message can use this as value for {@link #APP_PROPERTY_DEVICE_TTD} to indicate it in messages.
+     */
+    public static final int TTD_VALUE_ALWAYS_READY = -1;
+    /**
      * The name of the AMQP 1.0 message application property containing the id of the gateway
      * that wants to report data on behalf of another device.
      */
@@ -396,6 +405,31 @@ public final class MessageHelper {
     }
 
     /**
+     * Adds the <em>time until disconnect</em> property to an AMQP 1.0 message.
+     * <p>
+     * The value is put to the message's <em>application-properties</em> under key
+     * {@link #APP_PROPERTY_DEVICE_TTD}.
+     *
+     * @param msg The message to add the property to.
+     * @param timeUntilDisconnect The value of the property.
+     * @throws NullPointerException if timeUntilDisconnect is {@code null}.
+
+     */
+    public static void addTimeUntilDisconnect(final Message msg, final Integer timeUntilDisconnect) {
+        addProperty(msg, APP_PROPERTY_DEVICE_TTD, timeUntilDisconnect);
+    }
+
+    /**
+     * Gets the value of a message's {@link #APP_PROPERTY_DEVICE_TTD} application property.
+     *
+     * @param msg The message to get the property from.
+     * @return The property value or {@code null} if not set.
+     */
+    public static Integer getTimeUntilDisconnect(final Message msg) {
+        return getApplicationProperty(msg.getApplicationProperties(), APP_PROPERTY_DEVICE_TTD, Integer.class);
+    }
+
+    /**
      * Adds a property to an AMQP 1.0 message.
      * <p>
      * The property is added to the message's <em>application-properties</em>.
@@ -527,6 +561,61 @@ public final class MessageHelper {
         if (msg.getCreationTime() == 0) {
             msg.setCreationTime(Instant.now().toEpochMilli());
         }
+    }
+
+    /**
+     * Verify if a message is signalling that the
+     * device for which it was received should be ready to receive a command.
+     * This is evaluated at the point in time this method is invoked.
+     *
+     * @param msg Message that is evaluated.
+     * @return Boolean {@code true} if the message signals that the device now should be ready to receive a command, {@code false}
+     * otherwise.
+     * @throws NullPointerException If msg is {@code null}.
+     */
+    public static Boolean isDeviceCurrentlyReadyForCommands(final Message msg) {
+
+        return Optional.ofNullable(MessageHelper.getTimeUntilDisconnect(msg)).map(ttd -> {
+            if (ttd == MessageHelper.TTD_VALUE_ALWAYS_READY) {
+                return Boolean.TRUE;
+            } else if (ttd == 0) {
+                return Boolean.FALSE;
+            } else {
+                final Instant creationTime = Instant.ofEpochMilli(msg.getCreationTime());
+                return Instant.now().isBefore(creationTime.plusSeconds(ttd));
+            }
+        }).orElse(Boolean.FALSE);
+    }
+
+    /**
+     * Provide an instance of {@link TimeUntilDisconnectNotification} if a message indicates that the device sending it
+     * is currently ready to receive an upstream message.
+     * <p>
+     * If this is not the case, the returned {@link Optional} will be empty.
+     *
+     * @param msg Message that is evaluated.
+     * @return Optional containing an instance of the class {@link TimeUntilDisconnectNotification} if the device is considered
+     * being ready to receive an upstream message or is empty otherwise.
+     * @throws NullPointerException If msg is {@code null}.
+     */
+    public static Optional<TimeUntilDisconnectNotification> fromMessage(final Message msg) {
+
+        if (isDeviceCurrentlyReadyForCommands(msg)) {
+            final String tenantId = MessageHelper.getTenantIdAnnotation(msg);
+            final String deviceId = MessageHelper.getDeviceId(msg);
+
+            if (tenantId != null && deviceId != null) {
+                final Integer ttd = MessageHelper.getTimeUntilDisconnect(msg);
+                final Instant creationTime = Instant.ofEpochMilli(msg.getCreationTime());
+                final Instant deviceCommandReadyUntil = creationTime.plusSeconds(ttd);
+
+                final TimeUntilDisconnectNotification notification =
+                        new TimeUntilDisconnectNotification(tenantId, deviceId, deviceCommandReadyUntil);
+                return Optional.of(notification);
+            }
+        }
+
+        return Optional.empty();
     }
 
 }
--- a/core/src/main/java/org/eclipse/hono/util/MessageHelper.java
+++ b/core/src/main/java/org/eclipse/hono/util/MessageHelper.java
@@ -412,10 +412,8 @@ public final class MessageHelper {
      *
      * @param msg The message to add the property to.
      * @param timeUntilDisconnect The value of the property.
-     * @throws NullPointerException if timeUntilDisconnect is {@code null}.
-
      */
-    public static void addTimeUntilDisconnect(final Message msg, final Integer timeUntilDisconnect) {
+    public static void addTimeUntilDisconnect(final Message msg, final int timeUntilDisconnect) {
         addProperty(msg, APP_PROPERTY_DEVICE_TTD, timeUntilDisconnect);
     }
 
@@ -564,16 +562,17 @@ public final class MessageHelper {
     }
 
     /**
-     * Verify if a message is signalling that the
-     * device for which it was received should be ready to receive a command.
+     * Verify if a device is currently connected to a protocol adapter.
      * This is evaluated at the point in time this method is invoked.
+     * <p>
+     * This could be used as a trigger condition for an attempt to send a command upstream to the device.
      *
      * @param msg Message that is evaluated.
      * @return Boolean {@code true} if the message signals that the device now should be ready to receive a command, {@code false}
      * otherwise.
      * @throws NullPointerException If msg is {@code null}.
      */
-    public static Boolean isDeviceCurrentlyReadyForCommands(final Message msg) {
+    public static Boolean isDeviceCurrentlyConnected(final Message msg) {
 
         return Optional.ofNullable(MessageHelper.getTimeUntilDisconnect(msg)).map(ttd -> {
             if (ttd == MessageHelper.TTD_VALUE_ALWAYS_READY) {
@@ -585,37 +584,6 @@ public final class MessageHelper {
                 return Instant.now().isBefore(creationTime.plusSeconds(ttd));
             }
         }).orElse(Boolean.FALSE);
-    }
-
-    /**
-     * Provide an instance of {@link TimeUntilDisconnectNotification} if a message indicates that the device sending it
-     * is currently ready to receive an upstream message.
-     * <p>
-     * If this is not the case, the returned {@link Optional} will be empty.
-     *
-     * @param msg Message that is evaluated.
-     * @return Optional containing an instance of the class {@link TimeUntilDisconnectNotification} if the device is considered
-     * being ready to receive an upstream message or is empty otherwise.
-     * @throws NullPointerException If msg is {@code null}.
-     */
-    public static Optional<TimeUntilDisconnectNotification> fromMessage(final Message msg) {
-
-        if (isDeviceCurrentlyReadyForCommands(msg)) {
-            final String tenantId = MessageHelper.getTenantIdAnnotation(msg);
-            final String deviceId = MessageHelper.getDeviceId(msg);
-
-            if (tenantId != null && deviceId != null) {
-                final Integer ttd = MessageHelper.getTimeUntilDisconnect(msg);
-                final Instant creationTime = Instant.ofEpochMilli(msg.getCreationTime());
-                final Instant deviceCommandReadyUntil = creationTime.plusSeconds(ttd);
-
-                final TimeUntilDisconnectNotification notification =
-                        new TimeUntilDisconnectNotification(tenantId, deviceId, deviceCommandReadyUntil);
-                return Optional.of(notification);
-            }
-        }
-
-        return Optional.empty();
     }
 
 }
--- a/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
@@ -679,7 +679,7 @@ public abstract class AbstractProtocolAdapterBase<T extends ProtocolAdapterPrope
             msg.setBody(new Data(new Binary(payload.getBytes())));
         }
         if (timeUntilDisconnect != null) {
-            MessageHelper.addTimeUntilDisconnect(msg, timeUntilDisconnect);
+            MessageHelper.addTimeUntilDisconnect(msg, timeUntilDisconnect.intValue());
         }
 
         MessageHelper.setCreationTime(msg);
--- a/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
@@ -32,6 +32,7 @@ import org.eclipse.hono.service.auth.device.Device;
 import org.eclipse.hono.service.monitoring.ConnectionEventProducer;
 import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.CredentialsConstants;
+import org.eclipse.hono.util.EventConstants;
 import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.RegistrationConstants;
 import org.eclipse.hono.util.Strings;
@@ -647,6 +648,8 @@ public abstract class AbstractProtocolAdapterBase<T extends ProtocolAdapterPrope
      * @param payload The message payload.
      * @param registrationInfo The device's registration information as retrieved by the <em>Device Registration</em>
      *            service's <em>assert Device Registration</em> operation.
+     * @param timeUntilDisconnect The time to deliver in seconds to indicate that a device is ready to receive an upstream message.
+     *                      Maybe {@code null}.
      * @return The message.
      * @throws NullPointerException if address, device ID or registration info are {@code null}.
      */
@@ -656,7 +659,8 @@ public abstract class AbstractProtocolAdapterBase<T extends ProtocolAdapterPrope
             final String publishAddress,
             final String contentType,
             final Buffer payload,
-            final JsonObject registrationInfo) {
+            final JsonObject registrationInfo,
+            final Integer timeUntilDisconnect) {
 
         Objects.requireNonNull(address);
         Objects.requireNonNull(deviceId);
@@ -674,6 +678,9 @@ public abstract class AbstractProtocolAdapterBase<T extends ProtocolAdapterPrope
         if (payload != null) {
             msg.setBody(new Data(new Binary(payload.getBytes())));
         }
+        if (timeUntilDisconnect != null) {
+            MessageHelper.addTimeUntilDisconnect(msg, timeUntilDisconnect);
+        }
 
         MessageHelper.setCreationTime(msg);
 
@@ -683,7 +690,6 @@ public abstract class AbstractProtocolAdapterBase<T extends ProtocolAdapterPrope
 
     /**
      * Trigger the creation of a <em>connected</em> event.
-     * 
      * @param remoteId The remote ID.
      * @param authenticatedDevice The (optional) authenticated device.
      * @return A failed future if an event producer is set but the event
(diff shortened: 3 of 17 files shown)


sysexcontrol:
Contained is the implementation for
- setting the time-til-disconnect via HTTP header
- callback in northbound applications if a device indicates current readiness
- some definitions for the device connection/disconnection notification (as preparation)

Soon following will 
- the additional possibility to include the time-til-disconnect in the URL

sysexcontrol:
> the additional possibility to include the time-til-disconnect in the URL

Contained in additional commit 0966f81 now.

In another implementation PR following will be addressed:
- the delay of an HTTP response to allow an application to send down a command
- the retrieval of the value for `ttd` from the Device Registry or via the Tenant API
- MQTT implementation

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -55,6 +57,11 @@
      * belongs to.
      */
     public static final String APP_PROPERTY_DEVICE_ID              = "device_id";
+    /**
+     * The name of the AMQP 1.0 message application property containing the time until disconnect of the device that is available
+     * for receiving an upstream message for the given number of seconds (shortened as <em>Time Til Disconnect</em>).
```
sophokles73: short for <em> ... </em>

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -393,6 +400,29 @@ public static String getCacheDirective(final Message msg) {
...
+    /**
+     * Adds the <em>time until disconnect</em> property to an AMQP 1.0 message.
+     * <p>
+     * The value is put to the message's <em>application-properties</em> under key
+     * {@link #APP_PROPERTY_DEVICE_TTD}.
+     *
+     * @param msg The message to add the property to.
+     * @param timeUntilDisconnect The value of the property.
+     */
+    public static void addTimeUntilDisconnect(final Message msg, final Integer timeUntilDisconnect) {
```
sophokles73: what if ttd is `null`

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -527,4 +557,44 @@ public static void setCreationTime(final Message msg) {
...
+     * </ul>
+     *
+     * @param tenantId The tenant this handler is scoped to.
+     * @param messageConsumer Message consumer that handles generic messages.
+     * @param notificationReadyToDeliverConsumer BiConsumer&lt;String,String&gt; that is called when the message indicates the current readiness
+     *                                           to receive an upstream message. Maybe {@code null}. Provided is the tenantId
+     *                                           and deviceId.
+     * @return The message consumer that invokes the specific handlers depending on the event type.
+     */
+    public static Consumer<Message> createMessageDemuxHandler(final String tenantId,
```
sophokles73: IMHO this should go into a separate class (`MessageDemuxer`?) that implements `Consumer<Message>` and has this method as its static factory method.

BTW what happened to @ctron 's idea of providing a map of content-types to handlers as the param?

core/src/main/java/org/eclipse/hono/util/NotificationConstants.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.apache.qpid.proton.message.Message;
+
+import java.util.Arrays;
+import java.util.Objects;
+
+/**
+ * Constants used for notification events.
+ */
+public abstract class NotificationConstants {
```
sophokles73: Do we really need this class?

core/src/main/java/org/eclipse/hono/util/NotificationDeviceCommandReadyConstants.java
```diff
@@ -0,0 +1,73 @@
...
+     * device for which it was received should be ready to receive a command.
+     * This is evaluated at the point in time this method is invoked.
+     *
+     * @param tenantId Tenant this message is scoped to.
+     * @param deviceNotification Message that is evaluated.
+     * @return Boolean {@code true} if the event signals that the device now should be ready to receive a command, {@code false}
+     * otherwise.
+     * @throws NullPointerException If deviceNotification is {@code null}.
+     */
+    public static Optional<NotificationDeviceReadyToDeliver> isDeviceCurrentlyReadyForCommands(final String tenantId, final Message deviceNotification) {
```
sophokles73: IMHO we do not need this in a separate class, instead this method should go into `MessageHelper` where we already have means to read and set the ttd. then remove this class ...

core/src/main/java/org/eclipse/hono/util/ProtocolAdapterMessageHelper.java
```diff
@@ -0,0 +1,68 @@
...
+import io.vertx.core.buffer.Buffer;
+
+/**
+ * Utility methods for working with messages that arrive at a protocol adapter.
+ * <p>
+ * Methods contained here are provided for the protocol adapters and do not cover any Proton {@code Message}s.
+ * See {@link MessageHelper} for utility methods that are related to these kind of messages.
+ *
+ */
+public final class ProtocolAdapterMessageHelper {
```
sophokles73: we already have a place for utility methods used by all protocol adapters: `AbstractProtocolAdapterBase`

why do we need to introduce an additional place?

core/src/main/java/org/eclipse/hono/util/EventConstants.java
```diff
@@ -29,4 +29,15 @@
...
+
+    /**
+     * The content type that is defined for empty events without any payload.
+     */
+    public static final String CONTENT_TYPE_EMPTY_NOTIFICATION = "application/vnd.eclipse-hono-empty-notification";
+
+    /**
+     * The content type that is defined for device connection notification events.
+     */
+    public static final String CONTENT_TYPE_DEVICE_CONNECTION_NOTIFICATION = "application/vnd.eclipse-hono-dc-notification+json";
```
sophokles73: is this relevant for devices indicating their readiness to receive a command?

core/src/main/java/org/eclipse/hono/util/NotificationDeviceCommandReadyConstants.java
```diff
@@ -0,0 +1,73 @@
...
+    public static final int TTD_VALUE_ALWAYS_READY = -1;
+
+    /**
+     * Verify if a deviceNotification event (as a JSON object that is defined in the Event API) is signalling that the
+     * device for which it was received should be ready to receive a command.
+     * This is evaluated at the point in time this method is invoked.
+     *
+     * @param tenantId Tenant this message is scoped to.
+     * @param deviceNotification Message that is evaluated.
+     * @return Boolean {@code true} if the event signals that the device now should be ready to receive a command, {@code false}
```
sophokles73: this method currently doesn't return a Boolean (but it should IMHO). There is also no apparent reason for passing in a tenant ID FMPOV. in fact, the method seems to be nothing more like 

```
return Optional.ofNullable(MessageHelper.getTimeUntilDisconnect(deviceNotification)).map(ttd -> {
  if (ttd == TTD_VALUE_ALWAYS_READY) {
    return Boolean.TRUE;  
  } else if (deviceNotification.getCreationTime() == 0) {
    return Boolean.FALSE;
  } else {
    final Instant creationTime = Instant.ofEpochMilli(deviceNotification.getCreationTime());
    return Instant.now().isBefore(creationTime.plusSeconds(ttd));
}).orElse(Boolean.FALSE);
```

core/src/main/java/org/eclipse/hono/util/NotificationDeviceConnectionConstants.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package org.eclipse.hono.util;
+
+import io.vertx.core.json.JsonObject;
+
+/**
+ * Constants used for command readiness notifications of a device.
+ */
+public final class NotificationDeviceConnectionConstants extends NotificationConstants {
```
sophokles73: this doesn't seem to have anything to do with a device indicating readiness to receive a command, does it?

core/src/main/java/org/eclipse/hono/util/NotificationDeviceReadyToDeliver.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+package org.eclipse.hono.util;
+
+import java.time.Instant;
+
+/**
+ * Contains all information about a device that is indicating being ready to receive an upstream message.
+ */
+public final class NotificationDeviceReadyToDeliver {
```
sophokles73: what is this good for?

sysexcontrol: This defines the values passed to the callback method in the business application for convenience in one class.
The idea is to have everything contained to enable the application to send a command upstream.

Like
```
private void handleCommandReadinessNotification(final NotificationDeviceReadyToDeliver notificationDeviceReadyToDeliver) {
  // do something here, e.g. send a command
}```

File: core/src/main/java/org/eclipse/hono/util/ProtocolAdapterMessageHelper.java (hunk 6 above)
Comment (sysexcontrol) [reply core/src/main/java/org/eclipse/hono/util/ProtocolAdapterMessageHelper.java]: You are right, this fitted well into the `AbstractProtocolAdapterBase`, will eliminate the helper class.

File: core/src/main/java/org/eclipse/hono/util/NotificationDeviceCommandReadyConstants.java (hunk 5 above)
Comment (sysexcontrol) [reply core/src/main/java/org/eclipse/hono/util/NotificationDeviceCommandReadyConstants.java]: Good point, will move it to the `MessageHelper`.

File: core/src/main/java/org/eclipse/hono/util/MessageHelper.java (hunk 2 above)
Comment (sysexcontrol) [reply core/src/main/java/org/eclipse/hono/util/MessageHelper.java]: Right, javadoc for NPE was missing, thx!

File: core/src/main/java/org/eclipse/hono/util/NotificationDeviceReadyToDeliver.java (hunk 10 above)
Comment (sophokles73) [reply core/src/main/java/org/eclipse/hono/util/NotificationDeviceReadyToDeliver.java]: since there is no such thing as a *CommandReadinessNotification* I do not see how this is supposed to work ...

Comment (sysexcontrol) [reply core/src/main/java/org/eclipse/hono/util/NotificationDeviceReadyToDeliver.java]: This is generated by the demuxer that is evaluating the `piggy-backed` message and then invokes
the callback with this object (which is constructed from the message).

```
Optional.ofNullable(notificationReadyToDeliverConsumer).map(notificationConsumer -> {
                getReadyForCommandsNotification(tenantId, msg).ifPresent(
                        notificationObject -> notificationConsumer.accept(notificationObject));
. . .
```

This is how such an object is put into the callback implemented in the example already, without having a specific `CommandReadinessNotification`.
The application itself does not have to demux/switch after receiving a message by calling helper methods, but instead only needs to provide the callback method to the demuxer.

By this Hono can internally determine  when it thinks a device is ready to receive a command and notify the application  - no business coding involved.

Comment (sophokles73) [reply core/src/main/java/org/eclipse/hono/util/NotificationDeviceReadyToDeliver.java]: Ok, I think I get it now. The demuxer is used to detect all messages that contain a ttd property and if they do, creates an instance of this class and forwards it to the notificationConsumer while forwarding the original message to the downstream consumer, right?

However, we definitely need to find a new (better) name for this class. There is nothing being delivered here, is there?  What about `TimeTilDisconnectNotification`?

File: core/src/main/java/org/eclipse/hono/util/MessageHelper.java [hunk 11]
```diff
@@ -527,4 +557,44 @@ public static void setCreationTime(final Message msg) {
...
+            }
+
+            Optional.ofNullable(notificationReadyToDeliverConsumer).map(notificationConsumer -> {
+                NotificationDeviceCommandReadyConstants.isDeviceCurrentlyReadyForCommands(tenantId, msg).ifPresent(
+                        notificationObject -> notificationConsumer.accept(notificationObject));
+
+                messageConsumer.accept(msg);
+
+                return notificationConsumer;
+            }).orElse(v -> messageConsumer.accept(msg));
```
sophokles73: what's the purpose of passing in `null` as the notificationConsumer? Wouldn't the client in that case simply register a Consumer<Message> using HonoClient?

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
sophokles73: Now that I understand the purpose of this, I do not think that it is a demuxer anymore. It is more of a filter or a *tap*.

sysexcontrol: Yes, that I also like better, and it reduces the already big MessageHelper class.
Will move it to an own class.

To @ctron 's idea:
it seemed to me a good point when still thinking to only "demux" on content-types. But since this is not the case anymore, I do not think it is still matching, at least not currently.
So in this PR I decided to let the signature as it was - passing in explicit callback methods.

core/src/main/java/org/eclipse/hono/util/NotificationDeviceReadyToDeliver.java
sysexcontrol: > Ok, I think I get it now. The demuxer is used to detect all messages that contain a ttd property and if they do, creates an instance of this class and forwards it to the notificationConsumer while forwarding the original message to the downstream consumer, right?

Exactly!

> However, we definitely need to find a new (better) name for this class.

:+1: 

> There is nothing being delivered here, is there? What about TimeTilDisconnectNotification?

Thanks for the proposal, I wasn't too happy with the name either...
`TimeUntilDisconnectNotification` currently looks best for me - reducing the 'til' only to the protocol header names. WDYT?

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -527,4 +557,44 @@ public static void setCreationTime(final Message msg) {
         }
     }
 
+    /**
+     * Construct a message consumer for event messages that detects an indication that a device is ready to receive an upstream
+     * message and delegates to the passed callback handlers.
+     * <p>
+     * Note that there are two different ways how this can be detected:
```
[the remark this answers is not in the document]
sysexcontrol: Yes, I agree.
And the `... for event messages` is still wrong as well (can as well be telemetry messages).
I will provide a corrected javadoc.

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -527,4 +557,44 @@ public static void setCreationTime(final Message msg) {
...
+     * <ul>
+     *     <li>a specific event (that has the Content-Type {@link EventConstants#CONTENT_TYPE_EMPTY_NOTIFICATION}</li>
+     *     <li>any other event that carries the AMQP 1.0 application property {@link MessageHelper#APP_PROPERTY_DEVICE_TTD}</li>
+     * </ul>
+     *
+     * @param tenantId The tenant this handler is scoped to.
+     * @param messageConsumer Message consumer that handles generic messages.
+     * @param notificationReadyToDeliverConsumer BiConsumer&lt;String,String&gt; that is called when the message indicates the current readiness
```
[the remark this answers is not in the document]
sysexcontrol: Right - will update the javadoc.

core/src/main/java/org/eclipse/hono/util/NotificationDeviceCommandReadyConstants.java
sysexcontrol: I will split the code to
- make the evaluation if a message indicates command readiness (returning a Boolean like you outlined above)
- and to construct a callback object that is provided to the application. 

For this second method the tenantId is needed - it cannot be retrieved from the message AFAIK. Or do I miss something?
The application itself DOES know its tenantId and provides it, so FMPOV there is no problem here (concerns?).

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
sysexcontrol: Good question - I thought of extending this to more callbacks when we merge this with the "connected/disconnected" notification. And then it might be good to only provide the callbacks that an application is interested in.

But thinking about it now it seems to make things more complicated and others will ask the same question as you. So it probably is better to enforce callbacks to be always not null, and in some cases to have empty implementations instead.
Will remove the "maybe null" from the code.

core/src/main/java/org/eclipse/hono/util/NotificationDeviceCommandReadyConstants.java
sophokles73: > For this second method the tenantId is needed - it cannot be retrieved from the message AFAIK. Or do I miss something?

In fact we do add it to the message annotations (see `MessageHelper.annotate()`). However, this is currently done as part of the filters of HonoMessaging, i.e. when we connect the adapters to the AMQP Network directly, this would not happen. But this might only mean that we should add the annotations in the protocol adapters themselves ...

sysexcontrol: Right - if we do that the tenantId can be eliminated. Which would be definitely better for the usability inside the application code.
I will have a look into it and if it is simple to achieve, I will eliminate tenantId from the signature here.

sysexcontrol:
> Now that I understand the purpose of this, I do not think that it is a demuxer anymore. It is more of a filter or a tap.

I think `tap` is matching best - a filter would mean that some messages are not propagated at all ("filtered"), which is not the case.
So I will rename it to `MessageTap` (instead of `MessageDemuxer`).

core/src/main/java/org/eclipse/hono/util/NotificationDeviceCommandReadyConstants.java
sysexcontrol: I looked into it now - the `tenantId` is currently NOT annotated at the message.
If I see things correctly,  the `to` property of the AMQP message is the only part that contains the tenant, but as part of the endpoint (like `telemetry/DEFAULT_TENANT`).

Technically I think it is correct that the tenantId is not contained, since the endpoint is already scoped to it.

IMHO we have several options here:
- annotate the message with the tenantId (one additional property at each message)
- extend the `MessageConsumer` interface by a `getTenant()` method and implicitly get the tenant internally
- parse it from the endpoint (ugly :-1: )
- leave the signature like it is

For simplicitly I would currently vote for option 1 - (message contains tenantId in application property, just as deviceId).
Any concerns (my own concern would be that we continue to increase the message by more and more properties, which increases traffic and may reduce performance)  ?

sophokles73: All downstream telemetry and event messages get the tenant ID and device ID added to the message annotations. We even verify that this is the case (see `org.eclipse.hono.tests.client.ClientTestBase.assertMessageProperties()`

sysexcontrol: Yes, you are right - the tenantId is present in the annotations (not in the application properties, this was the wrong assumption).
I can eliminate the tenantId in the signature then :+1: 

BTW: why do we double the `deviceId` currently in the message annotations *and* in the application properties?

sophokles73: the idea was to have all information relevant for e.g. routing the message in one place (in the annotations) and not requiring e.g. a router to gather the information from arbitrary different places.

core/src/main/java/org/eclipse/hono/util/EventConstants.java
sysexcontrol: No, it was meant as minimal step to integrate with the `connected/disconnected` event.

I will eliminate it for clarity and we introduce it later.

core/src/main/java/org/eclipse/hono/util/NotificationDeviceConnectionConstants.java
sysexcontrol: Was meant as an integration with the `connected/disconnected` event - will eliminate it for clarity.

sysexcontrol:
@sophokles73 Thanks for the good review, I pushed a commit which should reflect all review comments.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -665,4 +666,37 @@ protected final Message newMessage(
...
     }
+
+    /**
+     * Approve a message defined by a Content-Type and a payload for protocol adapters.
+     * @param contentType The Content-Type of the payload, as received by the protocol adapter.
+     * @param payload The payload received by the protocol adapter.
+     * @return A String containing an error message (for further processing), or an empty Optional if the message was
+     *         approved successfully.
+     */
+    protected Optional<String> approveMessage(final Optional<String> contentType, final Buffer payload) {
```
sophokles73: FMPOV introducing this method was counter-productive. Code that was easy to read and understand before (the checks in the request handler methods) is now very hard to understand, not the least because the name of this method suggests some sort of *terminal* handling of the request.

FMPOV we should keep the original code for the sake of readability even if it means that there is some duplication.

sysexcontrol: Resulting in something like 

```
        if (contentType == EventConstants.CONTENT_TYPE_EMPTY_NOTIFICATION && payload != null && payload.length() > 0) {
            HttpUtils.badRequest(ctx, String.format("Content-Type %s does not allow non empty payload", contentType));
        } else if (contentType != EventConstants.CONTENT_TYPE_EMPTY_NOTIFICATION && (payload == null || payload.length() == 0)) {
            HttpUtils.badRequest(ctx, String.format("Content-Type %s does not allow empty payload", contentType));
        } else {
```

I do not find that too readable either, to be honest, but I can live with that.
After rebase there will be several more else, maybe afterwards a simple extract refactoring makes sense.
Will adopt that then...

sysexcontrol: Looking into it a little further I find that keeping the check in a separate method is much better testable. Now there are some easy to understand test methods for this that I would like to keep somehow.

So my question is:
do you dislike mainly the name of the method? Then it may help to just rename it to `checkPayloadByType`.
or do you dislike the usage of `Optional` for the return value to contain the error message? Then it could help to return boolean instead and let the adapter print a little more generic message for that.

Or do you think of a third option?

sophokles73: I think it's a combination of both the name and the usage of the optional as the return value. When returning a boolean, the method should be named starting with an `is...` or `has...` like when asking a question, e.g. `hasPayloadOfIndicatedType`  ...

sysexcontrol: This I can understand (although I liked the idea of using an Optional to return a `boolean` together with a possible error message, would be great if Java supported multiple return values like a lot of other languages now do...).

So I will use a `boolean isPayloadOfIndicatedType(...)` instead.

sysexcontrol:
@sophokles73 : I substituted the `approveMessage` method in the last commit.

sysexcontrol:
@sophokles73 : FMPOV this PR is done (did a force update after rebase), a final look would be appreciated before merging, thanks!

sophokles73:
build on travis failed due to a JavaDoc error.
it would also be easier to take a final look, if you would squash the commits so that they reflect the code you want to push. currently, there are changes being applied to the same code spots in multiple commits ...

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
sophokles73: why is this an `Integer` instead of an `int` anyway if setting it to `null` is not allowed? Is it allowed to set it to a negative value or to 0? this would probably require adding some checks and throwing an `IllegalArgumentException` instead of a `NullPointerException` then ...

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -529,4 +563,59 @@ public static void setCreationTime(final Message msg) {
...
+     * Verify if a message is signalling that the
+     * device for which it was received should be ready to receive a command.
+     * This is evaluated at the point in time this method is invoked.
+     *
+     * @param msg Message that is evaluated.
+     * @return Boolean {@code true} if the message signals that the device now should be ready to receive a command, {@code false}
+     * otherwise.
+     * @throws NullPointerException If msg is {@code null}.
+     */
+    public static Boolean isDeviceCurrentlyReadyForCommands(final Message msg) {
```
sophokles73: this method checks if a device is (supposedly) currently connected. Whether this is used in order to send a command or not has nothing to do with this method, does it?

IMHO it should be renamed and the JavaDoc should be updated accordingly.

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -529,4 +563,59 @@ public static void setCreationTime(final Message msg) {
...
+     * is currently ready to receive an upstream message.
+     * <p>
+     * If this is not the case, the returned {@link Optional} will be empty.
+     *
+     * @param msg Message that is evaluated.
+     * @return Optional containing an instance of the class {@link TimeUntilDisconnectNotification} if the device is considered
+     * being ready to receive an upstream message or is empty otherwise.
+     * @throws NullPointerException If msg is {@code null}.
+     */
+    public static Optional<TimeUntilDisconnectNotification> fromMessage(final Message msg) {
```
sophokles73: I would expect a static factory method of this name to create an instance of `Constants` because it is not qualified with a name of the type of object it creates. Can't we simply move this method to `TimeUntilDisconnectNotification` or is there another dependency from `MessageHelper` on `TimeUntilDisconnectNotification`? Otherwise, can we at least rename this method to e.g. `newTtdNotification`?

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
sysexcontrol: I agree, I will rename it :+1:

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
sysexcontrol: Looking at it again I fully see your concern - I think it is best moved to `TimeUntilDisconnectNotification` where it belongs to as a static factory method.
The dependency introduced for `TimeUntilDisconnectNotification` to `MessageHelper` I do not consider being critical.

Issue #601, opened 2018-04-25
Response for delete tenant is broken in BaseTenantService


DanielMaier-BSI:
My tenant client does not receive response for delete tenant requests. 

I am not too deep in this topic but I think I have found the issue: In
https://github.com/eclipse/hono/blob/617a2353dd5b9cd1bf42bdaea73468158a1ca681/service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java#L198-L217

response gets not created out of the request (in contrast to the other CRUD methods). I guess correlation id is not set in this case and requester can not receive response.

Can anybody please confirm this issue and cause? I can provide a PR if analysis was correct.

sophokles73:
Indeed, that seems to be the problem. Thanks for spotting and reporting :+1: 
Looking forward to your PR ...

DanielMaier-BSI:
Fixed with #603

PR #603, opened 2018-04-26
Fix for #601 - Response for delete tenant is broken in BaseTenantService

DanielMaier-BSI:
Create response for delete tenant out of request in order to include correlation-id.

I am a little bit lost how to test this. I would favour to have an integration test, but as removing tenants is not part of public API you cannot use tenant client. However, I tested it manually and it works. Any suggestions how to test this in a realistic way?


sophokles73:
We already create and delete tenants as part of the HTTP adapter integration tests. However, these tests use the HTTP endpoint of the Device Registry which uses direct replies to the requests sent over the event bus. I would therefore suggest to add a corresponding unit test that verifies that the response `EventBusMessage` contains the *reply-to* property value from the request.

PR #611, opened 2018-05-07
Adapter side of the Command and Control API.

pellmann:
- Abstract adapter can open receivers for Command and Control and send back a response.
- QPID dispatcher and MQTT/HTTP adapters in example configured.

### DIFF
--- a/example/src/main/config/hono-adapter-http-vertx-config.yml
+++ b/example/src/main/config/hono-adapter-http-vertx-config.yml
@@ -33,6 +33,14 @@ hono:
     port: 5671
     credentialsPath: ${secret.path}/http-adapter.credentials
     trustStorePath: ${secret.path}/trusted-certs.pem
+  command:
+    name: 'Hono HTTP Adapter'
+    amqpHostname: hono-http-internal
+    host: hono-dispatch-router.hono
+    port: 5673
+    keyPath: ${secret.path}/http-adapter-key.pem
+    certPath: ${secret.path}/http-adapter-cert.pem
+    trustStorePath: ${secret.path}/trusted-certs.pem
   metric:
     reporter:
       graphite:
--- a/example/src/main/config/hono-adapter-mqtt-vertx-config.yml
+++ b/example/src/main/config/hono-adapter-mqtt-vertx-config.yml
@@ -33,6 +33,14 @@ hono:
     port: 5671
     credentialsPath: ${secret.path}/mqtt-adapter.credentials
     trustStorePath: ${secret.path}/trusted-certs.pem
+  command:
+    name: 'Hono MQTT Adapter'
+    amqpHostname: hono-mqtt-internal
+    host: hono-dispatch-router.hono
+    port: 5673
+    keyPath: ${secret.path}/mqtt-adapter-key.pem
+    certPath: ${secret.path}/mqtt-adapter-cert.pem
+    trustStorePath: ${secret.path}/trusted-certs.pem
   metric:
     reporter:
       graphite:
--- a/example/src/main/config/qpid/qdrouterd-with-broker.json
+++ b/example/src/main/config/qpid/qdrouterd-with-broker.json
@@ -90,7 +90,8 @@
           "users": "consumer@HONO",
           "remoteHosts": "*",
           "maxSessions": 10,
-          "sources": "telemetry/*, event/*"
+          "sources": "telemetry/*, event/*, control/*",
+          "targets": "control/*"
         },
         "DEFAULT_TENANT": {
           "users": "user1@HONO",
@@ -116,6 +117,34 @@
           "targets": "telemetry/*, event/*"
         }
       }
+  }],
+
+  ["vhost", {
+    "id": "hono-mqtt-internal",
+    "maxConnections": 30,
+    "groups": {
+      "Hono": {
+        "users": "Eclipse IoT;Hono;mqtt-adapter",
+        "remoteHosts": "*",
+        "allowUserIdProxy": true,
+        "sources": "control/*",
+        "targets": "control/*"
+      }
+    }
+  }],
+
+  ["vhost", {
+    "id": "hono-http-internal",
+    "maxConnections": 30,
+    "groups": {
+      "Hono": {
+        "users": "Eclipse IoT;Hono;http-adapter",
+        "remoteHosts": "*",
+        "allowUserIdProxy": true,
+        "sources": "control/*",
+        "targets": "control/*"
+      }
+    }
   }],
 
   ["log", {
--- a/service-base/src/main/java/org/eclipse/hono/service/command/Command.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/command/Command.java
@@ -47,26 +47,25 @@ public class Command {
         this.replyAddress = message.getReplyTo();
         this.messageData = MessageHelper.getPayload(message);
         this.contentType = message.getContentType();
-        ApplicationProperties applicationProperties = message.getApplicationProperties();
+        final ApplicationProperties applicationProperties = message.getApplicationProperties();
         if (applicationProperties != null) {
             this.applicationProperties = applicationProperties.getValue();
         }
     }
 
     void sendResponse(final Buffer data, final Map<String, Object> properties, final Handler<ProtonDelivery> update) {
-        Message message = ProtonHelper.message();
+        final Message message = ProtonHelper.message();
         message.setCorrelationId(correlationId);
-        if(data!=null) {
+        if (data != null) {
             message.setBody(new Data(new Binary(data.getBytes())));
         }
         if (properties != null) {
             message.setApplicationProperties(new ApplicationProperties(properties));
         }
         MessageHelper.addProperty(message, MessageHelper.APP_PROPERTY_STATUS, HttpURLConnection.HTTP_OK);
-        if(update!=null) {
+        if (update != null) {
             responder.getSender().send(message, update);
-        }
-        else {
+        } else {
             responder.getSender().send(message);
         }
     }
--- a/service-base/src/main/java/org/eclipse/hono/service/command/Command.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/command/Command.java
@@ -1,0 +1,113 @@
+/**
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 1.0 which is available at
+ * https://www.eclipse.org/legal/epl-v10.html
+ *
+ * SPDX-License-Identifier: EPL-1.0
+ */
+
+package org.eclipse.hono.service.command;
+
+import java.net.HttpURLConnection;
+import java.util.Map;
+
+import io.vertx.core.buffer.Buffer;
+import org.apache.qpid.proton.amqp.Binary;
+import org.apache.qpid.proton.amqp.messaging.ApplicationProperties;
+import org.apache.qpid.proton.amqp.messaging.Data;
+import org.apache.qpid.proton.message.Message;
+import org.eclipse.hono.util.MessageHelper;
+
+import io.vertx.core.Handler;
+import io.vertx.proton.ProtonDelivery;
+import io.vertx.proton.ProtonHelper;
+
+/**
+ * Represents a Command and Control API command on the receiver side (adapter).
+ * <p>
+ * Received command includes the message data and all application properties.
+ */
+public class Command {
+
+    private Object correlationId;
+    private String replyAddress;
+    private String contentType;
+    private Buffer messageData;
+    private Map<String, Object> applicationProperties;
+    private CommandResponder responder;
+
+    Command(final CommandResponder responder, final Message message) {
+        this.responder = responder;
+        this.correlationId = message.getMessageId();
+        this.replyAddress = message.getReplyTo();
+        this.messageData = MessageHelper.getPayload(message);
+        this.contentType = message.getContentType();
+        ApplicationProperties applicationProperties = message.getApplicationProperties();
+        if (applicationProperties != null) {
+            this.applicationProperties = applicationProperties.getValue();
+        }
+    }
+
+    void sendResponse(final Buffer data, final Map<String, Object> properties, final Handler<ProtonDelivery> update) {
+        Message message = ProtonHelper.message();
+        message.setCorrelationId(correlationId);
+        if(data!=null) {
+            message.setBody(new Data(new Binary(data.getBytes())));
+        }
+        if (properties != null) {
+            message.setApplicationProperties(new ApplicationProperties(properties));
+        }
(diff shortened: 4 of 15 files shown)


service-base/src/main/java/org/eclipse/hono/service/command/CommandResponder.java
```diff
@@ -0,0 +1,55 @@
...
+package org.eclipse.hono.service.command;
+
+import io.vertx.proton.ProtonReceiver;
+import io.vertx.proton.ProtonSender;
+
+/**
+ * A command responder with a receiver/sender pair.
+ */
+class CommandResponder {
+
```
sysexcontrol: Clearer for me would be a name like `CommandLinkPair` or something similar. Why is that a `Responder`?

service-base/src/main/java/org/eclipse/hono/service/command/CommandConnectionImpl.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    public Future<Void> sendCommandResponse(final Command command, final Buffer data, final Map<String, Object> properties,
+                                            final Handler<ProtonDelivery> update) {
+        Objects.requireNonNull(command);
+        LOG.debug("create a command responder (sender link) for [replyAddress: {}]", command.getReplyAddress());
+        Future<Void> result = Future.future();
+        synchronized (command) {
```
sysexcontrol: Why is that `synchronized`?

service-base/src/main/java/org/eclipse/hono/service/command/Command.java
```diff
@@ -0,0 +1,113 @@
...
+        this.messageData = MessageHelper.getPayload(message);
+        this.contentType = message.getContentType();
+        ApplicationProperties applicationProperties = message.getApplicationProperties();
+        if (applicationProperties != null) {
+            this.applicationProperties = applicationProperties.getValue();
+        }
+    }
+
+    void sendResponse(final Buffer data, final Map<String, Object> properties, final Handler<ProtonDelivery> update) {
+        Message message = ProtonHelper.message();
```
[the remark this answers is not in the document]
sysexcontrol: Why is that `package local`?

service-base/src/main/java/org/eclipse/hono/service/command/Command.java
```diff
@@ -0,0 +1,113 @@
...
+public class Command {
+
+    private Object correlationId;
+    private String replyAddress;
+    private String contentType;
+    private Buffer messageData;
+    private Map<String, Object> applicationProperties;
+    private CommandResponder responder;
+
+    Command(final CommandResponder responder, final Message message) {
```
sysexcontrol: IMHO it would be better to keep the `message` as a member instead of copying some values into separate members.
It would make the class much smaller, and keeps it more flexible. Or would you think of serializing this class and thus rather prefer the `basic` types inside this class?

service-base/src/main/java/org/eclipse/hono/service/command/CommandConnectionImpl.java
```diff
@@ -0,0 +1,126 @@
...
+            if(h.succeeded()) {
+                getConnection().createReceiver(ResourceIdentifier.from(CommandConstants.COMMAND_ENDPOINT, tenantId, deviceId).toString()) // TODO
+                        .openHandler(oh -> {
+                            if (oh.succeeded()) {
+                                LOG.debug("command receiver successfully opened for [tenant: {}, device-id: {}]", tenantId, deviceId);
+                                ProtonReceiver protonReceiver = oh.result();
+                                CommandResponder responder = new CommandResponder(protonReceiver);
+                                protonReceiver.handler((delivery, message) -> {
+                                    LOG.debug("command message received on [address: {}]", message.getAddress());
+                                    commandHandler.handle(new Command(responder, message));
```
sysexcontrol: Do I see it correctly that
- the `Command` gets the responder (effectively a pair of links)
- and internally only uses the `sender` link of the responder?

I find it hard to understand here how the `Command`, the `CommandResponder` and the two Proton links interact with each other. Maybe the `Command` should only receive the sender link, since it does not use the receiver link.
In the end I do not fully understand for what the `CommandResponder` is for - maybe there is no real need for this class? WDYT?

service-base/src/main/java/org/eclipse/hono/service/command/CommandConnectionImpl.java
```diff
@@ -0,0 +1,126 @@
...
+        Future<Void> result = Future.future();
+        synchronized (command) {
+            ProtonSender sender = command.getResponder().getSender();
+            if (sender == null || !sender.isOpen()) {
+                connect().setHandler(h -> {
+                    getConnection().createSender(command.getReplyAddress())
+                            .openHandler(oh -> {
+                                if (oh.succeeded()) {
+                                    LOG.debug("command reply sender opened successful for [replyAddress: {}]", command.getReplyAddress());
+                                    command.getResponder().setSender(oh.result());
```
sysexcontrol: I propose to avoid using a setter on an internal member of the command (better use a direct method on command for that to avoid changing the internals of the class...), like `setSender(...)` at the command.

example/src/main/config/hono-adapter-http-vertx-config.yml
```diff
@@ -33,6 +33,14 @@ hono:
     port: 5671
     credentialsPath: ${secret.path}/http-adapter.credentials
     trustStorePath: ${secret.path}/trusted-certs.pem
+  command:
+    name: 'Hono HTTP Adapter'
+    amqpHostname: hono-http-internal
```
sysexcontrol: Wouldn't `hono-http-command` be a better name?

example/src/main/config/hono-adapter-mqtt-vertx-config.yml
```diff
@@ -33,6 +33,14 @@ hono:
     port: 5671
     credentialsPath: ${secret.path}/mqtt-adapter.credentials
     trustStorePath: ${secret.path}/trusted-certs.pem
+  command:
+    name: 'Hono MQTT Adapter'
+    amqpHostname: hono-mqtt-internal
```
sysexcontrol: Wouldn't `hono-mqtt-command` be a better name?

example/src/main/config/qpid/qdrouterd-with-broker.json
```diff
@@ -118,6 +119,34 @@
...
   }],
 
+  ["vhost", {
+    "id": "hono-mqtt-internal",
+    "maxConnections": 30,
+    "groups": {
+      "Hono": {
+        "users": "Eclipse IoT;Hono;mqtt-adapter",
+        "remoteHosts": "*",
+        "allowUserIdProxy": true,
```
sysexcontrol: what about maxFrameSize here? Would that allow an application to send a full video e.g. upstream to the device?

example/src/main/config/qpid/qdrouterd-with-broker.json
```diff
@@ -118,6 +119,34 @@
...
+      }
+    }
+  }],
+
+  ["vhost", {
+    "id": "hono-http-internal",
+    "maxConnections": 30,
+    "groups": {
+      "Hono": {
+        "users": "Eclipse IoT;Hono;http-adapter",
```
sysexcontrol: The defined name of the HTTP adapter is `hono-http` - maybe this should be used here as well?
Are these users additionally contained in a user database (e.g. `qdrouterd.sasldb`)?

service-base/src/test/java/org/eclipse/hono/service/command/CommandConnectionTest.java
```diff
@@ -0,0 +1,65 @@
...
+        commandConnection = new CommandConnectionImpl(vertx, new ClientConfigProperties());
+    }
+
+    /**
+     * Tests the command responder.
+     * @param ctx The vert.x test context.
+     */
+    @Test
+    public void testCreateCommandResponder(final TestContext ctx) {
+        // TODO
```
sysexcontrol: TODO :-))

tests/src/test/java/org/eclipse/hono/tests/command/CommandIT.java
```diff
@@ -0,0 +1,26 @@
...
+import io.vertx.ext.unit.junit.VertxUnitRunner;
+import org.junit.runner.RunWith;
+
+/**
+ * Integration tests for send commands to an adapter.
+ *
+ */
+@RunWith(VertxUnitRunner.class)
+public class CommandIT {
+
```
sysexcontrol: So far empty ;-)

service-base/src/main/java/org/eclipse/hono/service/command/CommandConnection.java
```diff
@@ -0,0 +1,53 @@
...
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.proton.ProtonDelivery;
+import org.eclipse.hono.client.HonoClient;
+
+import java.util.Map;
+
+/**
+ * A connection between an Adapter and the AMQP 1.0 network to receive commands and send a response.
```
sysexcontrol: Maybe add the word `bidirectional` at the beginning? Would clarify it for me ...

service-base/src/main/java/org/eclipse/hono/service/command/Command.java
```diff
@@ -0,0 +1,113 @@
...
+        message.setCorrelationId(correlationId);
+        if(data!=null) {
+            message.setBody(new Data(new Binary(data.getBytes())));
+        }
+        if (properties != null) {
+            message.setApplicationProperties(new ApplicationProperties(properties));
+        }
+        MessageHelper.addProperty(message, MessageHelper.APP_PROPERTY_STATUS, HttpURLConnection.HTTP_OK);
+        if(update!=null) {
+            responder.getSender().send(message, update);
```
sysexcontrol: Maybe this method should return the `ProtonDelivery` instance after sending the message, to give the caller some information if the send was successful at all - WDYT?

pellmann: I do not see a value of the local ProtonDelivery. For my understanding this would not give any information about success.

pellmann:
I integrated you recommended changed.

example/src/main/config/qpid/qdrouterd-with-broker.json
sysexcontrol: Ah, I see, thanks for the explanation :-)

tests/src/test/java/org/eclipse/hono/tests/command/CommandIT.java
sysexcontrol: Yes, that is probably true :+1:

service-base/src/main/java/org/eclipse/hono/service/command/CommandConnection.java
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Creates a receiver link and a command handler for Command and Control API messages.
+     *
+     * @param tenantId The tenant to consume commands from.
+     * @param deviceId The device for which this receiver will be created.
+     * @param commandHandler The message handler for the received commands.
+     * @return A future indicating the outcome. The future will succeed if the receiver link has been created.
+     * @throws NullPointerException if tenantId, deviceId or commandHandler is {@code null};
+     */
+    Future<Void> createCommandResponder(String tenantId, String deviceId,
```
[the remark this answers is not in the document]
pellmann: Would be createCommandReceiver better?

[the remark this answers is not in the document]
sysexcontrol: Yes, that sounds much better to me :+1:

service-base/src/main/java/org/eclipse/hono/service/command/CommandConnectionImpl.java
pellmann: It is the pair of links and it implements the closing of the sender as soon as the receiver is closed. Also I could imagine to close the sender link more often - we need to test this. For me it feels natural to pair them together in a class, since they are always needed as a pair.

example/src/main/config/hono-adapter-http-vertx-config.yml
pellmann: My intention was that this is the connection from the adapter to the AMQP 1.0 network, which is at the moment used for command and control.

example/src/main/config/qpid/qdrouterd-with-broker.json
pellmann: I have no idea how big the data will be. For telemetry and events I would not expect very big messages but for commands I can image e.g. a firmware update.

service-base/src/main/java/org/eclipse/hono/service/command/Command.java
sysexcontrol: Yes, I understand that.
My point is: it looks like a lot (not all) parts of a message is "copied out" to the members, which might be just fine - it just feels strange that e.g. the application properties are fully copied, some other values as well, but others not. And by this defining a class that is something like "half a message".
But thinking about it now it might be what we need, if we somehow persist a command in the future (which I think could come quite soon). Then it will be fine to copy out only the releveant parts.

So how about explaining this a little in the class javadoc, so others won't ask the same questions? If it is written that this intentionally pulls part of the message to ease persistence, I would not have asked myself this question.

service-base/src/main/java/org/eclipse/hono/service/command/Command.java
sysexcontrol: But at least about the local and/or remote state.
Since this is void return, it could just as well return what Proton itself considers a valuable return value (i.e. ProtonDelivery). WDYT?

service-base/src/main/java/org/eclipse/hono/service/command/CommandConnectionImpl.java
sysexcontrol: My point is that I feel the current classes should be separated much better.
Looking at the code 

```
ProtonSender sender = command.getResponder().getSender();
            if (sender == null || !sender.isOpen()) {
                connect().setHandler(h -> {
                    getConnection().createSender(command.getReplyAddress())
                            .openHandler(oh -> {
                                if (oh.succeeded()) {
                                    LOG.debug("command reply sender opened successful for [replyAddress: {}]", command.getReplyAddress());
                                    command.getResponder().setSender(oh.result());
                                    command.sendResponse(data, properties, update);
                                    result.complete();
                                } else {
                                    LOG.debug("command reply failed opening for [replyAddress: {}] : {}", command.getReplyAddress(), oh.cause());
                                }
                            }).open();
                });
```

it mangles the ProtonReceiver with the Command class and also involves the CommandResponder which is changed outside of the Command class by invoking the `setSender` method and thus feels like a code smell to me.

I would at least avoid retrieving the CommandResponder and change this implementation detail of `Command` from the outside, and rather do something like
`command.setSender(oh.result());` (which internally may delegate that to the contained responder).

And if additionally the `CommandResponder` would be called something like `CommandPairManager` I would be much happier :-)

PR #616, opened 2018-05-13
Adapter side of the Command and Control API.

pellmann:
- Abstract adapter can open receivers for Command and Control and send back a response.
- QPID dispatcher and MQTT/HTTP adapters in example configured.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/command/CommandConnectionImpl.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/command/CommandConnectionImpl.java
@@ -17,6 +17,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.function.Consumer;
 
+import org.eclipse.hono.client.MessageConsumer;
 import org.eclipse.hono.client.impl.HonoClientImpl;
 import org.eclipse.hono.config.ClientConfigProperties;
 import org.eclipse.hono.connection.ConnectionFactory;
@@ -53,7 +54,7 @@ public class CommandConnectionImpl extends HonoClientImpl implements CommandConn
     /**
      * {@inheritDoc}
      */
-    public Future<CommandAdapter> createCommandConsumer(
+    public final Future<MessageConsumer> createCommandConsumer(
             final String tenantId,
             final String deviceId,
             final Consumer<Command> messageConsumer,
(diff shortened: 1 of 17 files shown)


service-base/src/main/java/org/eclipse/hono/service/command/CommandConnectionImpl.java
```diff
@@ -0,0 +1,105 @@
...
+            final String deviceId,
+            final Consumer<Command> messageConsumer,
+            final Handler<Void> closeHandler) {
+
+        Objects.requireNonNull(tenantId);
+        Objects.requireNonNull(deviceId);
+        Objects.requireNonNull(messageConsumer);
+        return checkConnected().compose(con -> {
+            final Future<CommandAdapter> result = Future.future();
+            CommandAdapter.create(context, clientConfigProperties, connection, tenantId, deviceId,
```
sysexcontrol: Possibly the `create` method would be better located in the `Command` class?
In the end, the creation of the adapter results here in the subsequent creation of the `Command` anyway.

Maybe I miss something, but the dependencies between `Command`, `CommandAdapter` and `CommandConnection` seem to me more complicated as they should - WDYT?

pellmann: The CommandConnection does the connection handling. On that connection are n link pairs, which are included in the CommandAdapter. On this receiver links there could arrive n commands, which need to send back a response on the sender link (included in the CommandAdapter).

What do you suggest to simplification?

sysexcontrol: After having a second look at the code, let me first ask the question:

why would a protocol adapter ever deal with a `CommandAdapter` directly?
IMHO it handles a received command (instance of `Command`), and it may handle a response to it (by passing the instance of `Command` to the `sendCommandResponse` method of `CommandConnection`). 
So it works with `Command` and `CommandConnection`. 

Is that right?

pellmann:
I committed some changes to simplify the response sending.

service-base/src/main/java/org/eclipse/hono/service/command/CommandConnectionImpl.java
pellmann: The needed way to close a receiver (and the sender) is now available with a MessageConsumer.

PR #620, opened 2018-05-16
Concept and current state of Command & Control

pellmann:
Added 'Command & Control' in 'Concepts' in the documentation


site/content/concepts/command-and-control.md
```diff
@@ -3,29 +3,74 @@ title = "Command & Control"
...
 
-With the `ttd`-[notification]({{< relref "device-notifications.md" >}}) in (1) the device indicates it will stay connected for max. 30 seconds. This notification will be consumed by the application (2) and it now tries to send a command (3) to the device at the given address `control/TENANT/4711` (If the device would not be connected and the adapter has not established a link for it, the application would got no credits so send the command).
+With the `hono-ttd`-[notification]({{< relref "device-notifications.md" >}}) in (1) the device indicates it will stay connected for max. 30 seconds. This notification will be consumed by the application (2) and it now tries to send a command (3) to the device at the given address `control/TENANT/4711` (If the device is not connected, the adapter has not established a link for it and the application would get no credits so send the command).
 
-The HTTP Adapter gets the command and writes it in the response of the devices `send event` (4).
+The HTTP Adapter gets the command and writes it in the response of the devices `send event` (4). The HTTP Adapter sets the following response headers and optionally a payload.
```
sophokles73: does the HTTP adapter only include the command in a response to a successful request (i.e. if the response contains a 2xx status code) or can the command also be included in a response to an unsuccessful request? If so then how does the device know that the response payload is *not* a description of the error that occurred but is a command that it should process?

site/content/concepts/command-and-control.md
```diff
@@ -3,29 +3,74 @@ title = "Command & Control"
...
+* URI: `/control/res/${tenantId}/${deviceId}` 
+* Method: `PUT`
+
+Further:
+
+* Request Headers:
+  * `hono-cmd-req-id`: The value of the response-header (with same name) of the command event.
+* Request Body:
+  * (optional) Arbitrary payload as response data.
+* Status Codes:
```
sophokles73: are these the status codes contained in the response that the adapter sends back to the device or are these status codes that the device should include in its HTTP request to indicate the outcome of the command?

site/diagrams/sequenceDiagrams/command_control_concept_http.txt
```diff
@@ -0,0 +1,18 @@
+participant Device
+participant HTTP Adapter
+participant AMQP 1.0 Network
```
sophokles73: the network doesn't seem to be involved, so remove it

site/diagrams/sequenceDiagrams/command_control_concept_http.txt
```diff
@@ -0,0 +1,18 @@
...
+participant Application
+
+Device->HTTP Adapter: 1. send event (hono-ttd=30)
```
sophokles73: IMHO it is relevant to point out that we are talking about an HTTP request here to which the device will later get a response. So I think this should be something like
`HTTP request ("POST", "/event", "hono-ttd=30")`

site/diagrams/sequenceDiagrams/command_control_concept_http.txt
```diff
@@ -0,0 +1,18 @@
...
+activate HTTP Adapter
+HTTP Adapter->Application: 2. connected notification (ttd=30)
+activate Application
+Application->HTTP Adapter: 3. transfer("control/TENANT/4711","light","ON")
+HTTP Adapter->Device: 4. event response("light","ON",${reqId})
```
sophokles73: according to my comment above:
`HTTP response ("202", "hono-cmd=light, hono-cmd-req-id=${reqId})", "ON")`

site/diagrams/sequenceDiagrams/command_control_concept_http.txt
```diff
@@ -0,0 +1,18 @@
...
+deactivate HTTP Adapter
+note over Device
+Command response in next operation
+end note over
+Device->HTTP Adapter: 5. /control/res("hono-cmd=light","ON",${reqId})
```
sophokles73: IMHO this request will need to include the request ID (in the header or as a path parameter) and will include the response in the payload, right?
Something like:
`HTTP request("POST", "/control/res/${reqId}", "lumen=100")`

site/diagrams/sequenceDiagrams/command_control_concept_http.txt
```diff
@@ -0,0 +1,18 @@
...
+HTTP Adapter->Application: 6. transfer("control/TENANT/4711/${replyId}",200,"lumen=100")
```
sophokles73: `${replyTo}`?
Where does the `200` come from? If this is the result status of executing the command on the device then it needs to be passed in from the device in its request, e.g.
`HTTP request("POST", "/control/res/${reqId}/200", "lumen=100")`

site/diagrams/sequenceDiagrams/command_control_concept_http.txt
```diff
@@ -0,0 +1,18 @@
...
+deactivate Application
```
sophokles73: IMHO we should also show the HTTP response that is returned to the device ...

site/diagrams/sequenceDiagrams/command_control_concept_http.txt
```diff
@@ -0,0 +1,18 @@
+participant Device
+participant HTTP Adapter
+participant AMQP 1.0 Network
+participant Application
+
+Device->HTTP Adapter: 1. send event (hono-ttd=30)
+activate HTTP Adapter
+HTTP Adapter->Application: 2. connected notification (ttd=30)
```
sophokles73: this is not a specific *notification* but simply the transfer of the event to the application, isn't it?

site/content/concepts/command-and-control.md
pellmann: The latter.

site/diagrams/sequenceDiagrams/command_control_concept_mqtt.txt
```diff
@@ -0,0 +1,18 @@
...
+participant AMQP 1.0 Network
+participant Application
+
+Device->MQTT Adapter: 1. send event
+note over Device
+content-type=application/vnd.eclipse-hono-empty-notification
+-> MQTT Adapter will convert this in tdd=-1 (infinitely)
+end note over
+activate MQTT Adapter
+MQTT Adapter->Application: 2. connected notification (ttd=-1)
```
sophokles73: this is simply the transfer of the (empty) event ...

site/diagrams/sequenceDiagrams/command_control_concept_mqtt.txt
```diff
@@ -0,0 +1,18 @@
...
+activate Application
+Application->MQTT Adapter: 3. transfer("control/TENANT/4711","light","ON")
+MQTT Adapter->Device: 4. "control/light/${reqId}":"ON"
+Device->MQTT Adapter: 5. "control/res/200/${reqId}":"lumen=100"
```
sophokles73: I believe the request ID should be *before* the status code in the topic name

site/diagrams/sequenceDiagrams/command_control_concept_http.txt
pellmann: The only thing is, that the business app connects to the network. But it adds nothing important to the sequence. But if I remove it, the reader might think the application connects directly to the adapter.

site/diagrams/sequenceDiagrams/command_control_concept_http.txt
pellmann: Yes, that makes it clearer.

site/diagrams/sequenceDiagrams/command_control_concept_http.txt
pellmann: Do you prefer path over parameter?

sophokles73: I believe that devices will have an easier time putting the information into the path than setting headers. This would also be more consistent with the way the info is conveyed in the topic name when using MQTT, wouldn't it?

site/diagrams/sequenceDiagrams/command_control_concept_http.txt
pellmann: Yes, I forgot the status in the diagram - thanks!

site/diagrams/sequenceDiagrams/command_control_concept_http.txt
pellmann: It could be an event with notification or telemetry with notification.. but only the notification part is relevant for the following seqence.

site/diagrams/sequenceDiagrams/command_control_concept_mqtt.txt
pellmann: Hmm - yes, and I shown it differently in the both diagrams. I will rework this - but I need to make it visible, that the ttd header is the relevant part.

site/diagrams/sequenceDiagrams/command_control_concept_mqtt.txt
pellmann: For what reason? The status is a part of the response from the device, the reqId is just needed for transport.

site/diagrams/sequenceDiagrams/command_control_concept_http.txt
pellmann: I agree with the latter and also the reqId feels like a recourse part.

site/diagrams/sequenceDiagrams/command_control_concept_http.txt
sophokles73: not sure what you mean by *notification*. FMPOV there is no such thing.

site/diagrams/sequenceDiagrams/command_control_concept_mqtt.txt
sophokles73: FMPOV this better reflects the *natural semantic ordering* of the information:
> this is the response to request X and it has outcome 200

as opposed to:
> this is a response to a request which has a status of 200, oh and by the way, the request that this is the response to is ${reqId} ...

site/content/concepts/command-and-control.md
sysexcontrol: @sophokles73 :+1: IMHO this is a very clever comment/question.

Due to the one-way character of the downstream messages (meaning that there is no answer to them that could be unsuccessful) I see mainly problems with sending the message at all (network problems downstream, QoS level problems with missing acks, ...).

Since currently we can only send a command when a downstream message successfully reached an application, the unsuccessful cases should never have a command for now.

But if we buffer or broker commands in the future, this would change, so this is only an assumption for 0.6 IMHO.

Summarized:
- IMHO we need to decide this question in the future
- but for 0.6 I see no possible occurence of a command in unsuccessful HTTP requests

WDYT?

pellmann: The command should only appear in a successful request. At least for 0.6. But I do not see why it should be in a a response to an unsuccessful request at all? I will add a hint, that the command will only be included in a successful requests response (2xx).

pellmann:
Included the review feedback.

PR #622, opened 2018-05-18
Hono HTTP adapter Command and Control.

sysexcontrol:
This PR contains the fully working version of "void" commands for the HTTP adapter.
Included also is the extended example consumer, which now reacts to a notification callback (having been empty before) by trying to send a command upstream.
The result can be watched e.g. by using `curl`, like the following:

```
$ curl -X POST -i -u sensor1@DEFAULT_TENANT:hono-secret -H 'Content-Type: application/json' --data-binary '{"temp": 5}' http://localhost:8080/telemetry?hono-ttd=10
```

resulting in

```
HTTP/1.1 202 Accepted
hono-command: setBrightness
hono-cmd-req-id: 47#cmd-client-f577a018-32ba-4b14-bc1f-902acc77da12/28c2b33a-a5f7-4962-8dcf-316725583cb9
Content-Length: 82

{
  "brightness" : 6,
  "correlationId" : "0315ad62-32a1-497e-80d6-950dbf6d8f02"
}
```

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -29,6 +29,7 @@ import org.eclipse.hono.client.MessageConsumer;
 import org.eclipse.hono.client.MessageSender;
 import org.eclipse.hono.service.AbstractProtocolAdapterBase;
 import org.eclipse.hono.service.auth.device.Device;
+import org.eclipse.hono.service.command.CommandResponseSender;
 import org.eclipse.hono.service.http.HttpUtils;
 import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.EventConstants;
@@ -607,7 +608,7 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
      * @return The BiConsumer to pass to the command and control consumer.
      */
     private BiConsumer<ProtonDelivery, Message> createCommandMessageConsumer(final RoutingContext ctx,
-            final String tenant, final String deviceId, final Handler<Void> commandReceivedHandler) {
+            final String tenant, final String deviceId, final Handler<Message> commandReceivedHandler) {
         return (delivery, commandMessage) -> {
 
             final Optional<String> commandSubject = Optional.ofNullable(commandMessage.getProperties().getSubject());
@@ -621,7 +622,7 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
 
                 HttpUtils.setResponseBody(ctx.response(), MessageHelper.getPayload(commandMessage));
 
-                commandReceivedHandler.handle(null);
+                commandReceivedHandler.handle(commandMessage);
 
             } else {
                 LOG.info("Received command with invalid reply-to endpoint - ignoring.");
@@ -654,7 +655,7 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
             final AtomicLong timerId = new AtomicLong();
 
             // create a handler for being invoked if a command was received
-            final Handler<Void> commandReceivedHandler = v -> {
+            final Handler<Message> commandReceivedHandler = commandMessage -> {
                 if (downstreamMessageSent.get()) {
                     cancelResponseTimer(timerId.get());
                     closeCommandReceiverLink(messageConsumerRef);
@@ -662,6 +663,24 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
                         ctx.response().end();
                     }
                 }
+                final String replyId = getReplyToIdFromCommand(tenant, deviceId, commandMessage).get();
+
+                // send answer to caller via sender link
+                final Future<CommandResponseSender> responseSender = createCommandResponseSender(tenant, deviceId, replyId);
+                responseSender.map(commandResponseSender ->
+                    commandResponseSender.sendCommandResponse(tenant, deviceId, replyId, getCorrelationIdFromMessage(commandMessage),
+                            null, null, HttpURLConnection.HTTP_OK)
+                ).map(protonDeliveryFuture -> {
+                    LOG.debug("Command acknowledged to sender.");
+                    responseSender.result().close(v -> {});
+                    return null;
+                }).otherwise(t -> {
+                    LOG.debug("Could not acknowledge command to sender", t);
+                    if (responseSender.succeeded()) {
+                        responseSender.result().close(v -> {});
+                    }
+                    return null;
+                });
             };
 
             // create the commandMessageConsumer that handles an incoming command message
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -616,7 +616,7 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
 
             if (commandSubject.isPresent() && commandRequestId.isPresent()) {
 
-                ctx.response().putHeader(Constants.HEADER_COMMAND, commandSubject.get().toString());
+                ctx.response().putHeader(Constants.HEADER_COMMAND, commandSubject.get());
                 ctx.response().putHeader(Constants.HEADER_COMMAND_REQUEST_ID, commandRequestId.get());
 
                 HttpUtils.setResponseBody(ctx.response(), MessageHelper.getPayload(commandMessage));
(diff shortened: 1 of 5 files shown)


adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -543,6 +576,133 @@ private void doUploadMessage(final RoutingContext ctx, final String tenant, fina
...
+     * Opens a command receiver link for a device by creating a command and control consumer.
+     *
+     * @param ctx The routing context of the HTTP request.
+     * @param tenant The tenant of the device for that a command may be received.
+     * @param deviceId The id of the device for that a command may be received.
+     * @param downstreamMessageSent An AtomicBoolean indicating if the (piggy backed) downstream message of the device
+     *                              was sent already.
+     * @param messageConsumerRef An AtomicReference to the {@link MessageConsumer} that is set for further usage
+     *                           by the caller of this method.
+     * @return Optional An optional timerId if there was a
```
pellmann: Sentence does not end.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -543,6 +576,133 @@ private void doUploadMessage(final RoutingContext ctx, final String tenant, fina
...
+            final String tenant, final String deviceId, final Handler<Void> commandReceivedHandler) {
+        return (delivery, commandMessage) -> {
+
+            final Optional<String> commandSubject = Optional.ofNullable(commandMessage.getProperties().getSubject());
+            final Optional<String> commandRequestId = validateAndGenerateCommandRequestId(tenant, deviceId,
+                    commandMessage);
+
+            if (commandSubject.isPresent() && commandRequestId.isPresent()) {
+
+                ctx.response().putHeader(Constants.HEADER_COMMAND, commandSubject.get().toString());
```
pellmann: toString() is not needed.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -543,6 +576,133 @@ private void doUploadMessage(final RoutingContext ctx, final String tenant, fina
...
+                        // create a timer that is invoked if no command was received until timeToDelayResponse is expired
+                        timerId.set(getVertx().setTimer(timeToDelayResponse * 1000L,
+                                delay -> {
+                                    closeCommandReceiverLink(messageConsumerRef);
+                                    // if the downstream message was sent already, finish the request now
+                                    if (downstreamMessageSent.get()) {
+                                        ctx.response().end();
+                                    }
+                                }));
+                        // responseTimerId.set(timerId.longValue());
```
pellmann: Old code?

example/src/main/java/org/eclipse/hono/vertx/example/base/HonoConsumerBase.java
```diff
@@ -144,14 +147,39 @@ private void printMessage(final String tenantId, final Message msg, final String
...
+        System.out.println(String.format("Device is ready to receive a command : <%s>.", notification.toString()));
+
+        final String tenantId = notification.getTenantId();
+        final String deviceId = notification.getDeviceId();
+
+        honoClient.getOrCreateCommandClient(tenantId, deviceId).map(commandClient -> {
+            // generate a correlationId for the application in the payload
+            final UUID correlationIdForPayload = UUID.randomUUID();
+            final JsonObject jsonCmd = new JsonObject().put("brightness", (int)(Math.random() * 100)).
+                    put("correlationId", correlationIdForPayload.toString());
```
pellmann: Did we not wanted to show a direct case without an internal correlationId (for later events) as our first case? I think this might be confusing, now?!

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -447,6 +449,37 @@ protected TrustOptions getServerTrustOptions() {
...
+     * @return Optional An Optional with the combined String, or an empty Optional if the validation failed.
+     * @throws NullPointerException If any of the parameters are null.
+     */
+    protected final Optional<String> validateAndGenerateCommandRequestId(final String tenantId, final String deviceId, final Message commandMessage) {
+        Objects.requireNonNull(tenantId);
+        Objects.requireNonNull(deviceId);
+        Objects.requireNonNull(commandMessage);
+
+        // example of commandReplyId: control/DEFAULT_TENANT/4711/33fe70fd-5a2e-4095-83db-00101bf74a07
+        final String commandReplyId = commandMessage.getProperties().getReplyTo();
```
pellmann: commandMessage.getReplyTo() might be better.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -496,6 +504,10 @@ private void doUploadMessage(final RoutingContext ctx, final String tenant, fina
                 final Future<JsonObject> tokenTracker = getRegistrationAssertion(tenant, deviceId, authenticatedDevice);
                 final Future<TenantObject> tenantConfigTracker = getTenantConfiguration(tenant);
 
+                final AtomicBoolean downstreamMessageSent = new AtomicBoolean(false);
```
pellmann: The code reads not very fluent with this marker, which is set and checked deep inside different code blocks. Maybe we can find a better solution in a future version.

sysexcontrol: Full ack to this - this serves as the now working version which is already cleaned up compared to the previous prototype, but still needs some atomic `helper` objects.

I am sure in the first milestone of 0.7 there will be an improved, better readable solution, but for now this is what we have here and what works in all cases:

- the command is received **before** the downstream message was sent
- the command is received completely parallel while the downstream message is being sent
- the command is received strictly **after** the downstream message was sent to the AMQP network

So, I propose to keep it for 0.6 like this and come up with a better readable version afterwards.

example/src/main/java/org/eclipse/hono/vertx/example/base/HonoConsumerBase.java
sysexcontrol: Yes, you are right, I will remove it!

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
sysexcontrol: Thanks for spotting :+1:

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
sysexcontrol: Correct, thanks for spotting!

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
sysexcontrol: Will correct it.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
sysexcontrol: :+1:

sysexcontrol:
@pellmann : thanks for the review comments!

I pushed a commit with all changes, and one commit that additionally implements the current solution for device responses:
instead of waiting for the device, the HTTP protocol adapter answers the request successfully as soon as the command was received.

Can you have a look into the new commit as well, please? Thx.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -654,14 +655,32 @@ private void closeCommandReceiverLink(final AtomicReference<MessageConsumer> mes
...
                         ctx.response().end();
                     }
                 }
+                final String replyId = getReplyToIdFromCommand(tenant, deviceId, commandMessage).get();
+
+                // send answer to caller via sender link
+                final Future<CommandResponseSender> responseSender = createCommandResponseSender(tenant, deviceId, replyId);
+                responseSender.map(commandResponseSender ->
+                    commandResponseSender.sendCommandResponse(tenant, deviceId, replyId, getCorrelationIdFromMessage(commandMessage),
+                            Buffer.buffer("ok"), null, HttpURLConnection.HTTP_OK)
```
pellmann: A response without a body (null) would be more straightforward FMPOV.

sysexcontrol: Fine for me too - will change it.

sysexcontrol:
@pellmann : pushed changes.

PR #628, opened 2018-05-25
Define revision in parent pom

pho-enix:
* Make it possible to override version via mvn property
* Reduce changes introduced by release

### DIFF
--- a/adapters/http-vertx-base/pom.xml
+++ b/adapters/http-vertx-base/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.eclipse.hono</groupId>
     <artifactId>hono-adapters</artifactId>
-    <version>0.7-M1-SNAPSHOT</version>
+    <version>${revision}</version>
   </parent>
   <artifactId>hono-adapter-http-vertx-base</artifactId>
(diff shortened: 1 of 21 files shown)


pho-enix:
I understand that you have the version update well automated for the Hono project.

The motivation for this PR comes from a vendoring perspective. We maintain a fork of Hono with some custom adoptions. We have our own schema for build numbers which we want to set externally and not in the pom.xml. With this change, we can simply `-Drevision=2018-05-08-41pre`.

sophokles73:
> With this change, we can simply -Drevision=2018-05-08-41pre

in order to do what exactly?

pho-enix:
Let's say I want Hono jars with a `2018-05-08-41pre` versioning.
I want subsequent build numbers because of security patches.

pho-enix:
Maybe you can elaborate how hono project updates the versions of the packages.

If this is some neat functionality I will be happily using this for my very purpose.

sophokles73:
In order to set a new version:
mvn versions:set -DnewVersion=1.0-MY_VERSION

reverting the changes:
mvn versions:revert

pho-enix:
Tried it out. Works perfectly for us :)

Thanks a lot for the support.

Issue #635, opened 2018-05-29
Upgrade Grafana to 5.x


ctron:
* Upgrade deployments
* Ensure that existing dashboards still work
* Try using datasource and dashboard pre-provisioning

sophokles73:
@ctron do you also want to use this issue for making the changes to the S2I deployment (i.e. keep this issue open) or will you do that separately (i.e. close this issue)?

PR #637, opened 2018-05-30
[#635] Upgrade to Grafana 5.1.3

ctron:
This PR upgrades Grafana to version 5.1.3 and makes use of the new pre-provisioning features of Grafana 5.

I did test this with OpenShift (+S2i) with `minishift`, Kubernetes (`minikube`), Docker (Ubuntu 16.04 VM). But I really encourage you to do your own testing.


example/pom.xml
```diff
@@ -304,4 +303,44 @@ Also contains scripts for deploying a Hono instance to a Docker Swarm, a Kuberne
...
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>io.fabric8</groupId>
+            <artifactId>fabric8-maven-plugin</artifactId>
+            <configuration>
+              <images>
+                <image>
+                  <name>${docker.image.org-name}/hono-grafana:%v</name>
```
sophokles73: we do not want to build our own image for grafana but instead want to use the standard image ...

ctron: Well the standard image suffers from a few issues:
*  It still has permission problems so you still require "root" permissions, which may not be possible to have in some OpenShift cluster setups. Thus the S2I based deployment already used its own Grafana build.
* The official image doesn't seem to provide an entrypoint for pre-provisioning
* From what I know the standard image isn't refreshed, so you will have stale images

All of this, in combination with the fact that the build for this image is super simple, I think it makes sense to provide a custom version, as it really offers a benefit over the official image.

What would speak against a custom image?

sophokles73:
> What would speak against a custom image?

My concern is that we are then no longer talking about a works-with dependency because the image would become part of the Hono release ...

example/pom.xml
sophokles73: That we need to maintain it and it is *not* in Hono's scope.

I understand your pain point regarding OpenShift deployment but frankly, I do not see why the Hono project should be responsible for fixing this problem. Why don't you try to get your fixes incorporated into the *official* Grafana image instead?

ctron:
So the response from the Eclipse IP team and EMO was that we can roll our own images as long as they are simply a convenience and are not required to provide Hono core functionality.

For OpenShift I don't see any problems as we already do that. OpenShift S2I already builds custom images.

However I would also like to use the pre-provisioning process for Grafana on both Kubernetes and Docker Swarm. And taking this upstream to Grafana is most likely a way more time consuming process than providing a small custom Grafana docker image in the context of Hono.

In a nutshell: from an Eclipse Foundation perspective we are good to go. We simply need to decide if we want to do that in the context of Hono. Or stick to the limits of the upstream image.

sophokles73:
@ctron I have upgraded Docker Swarm, Kubernetes and (standard) OpenShift deployment to use the standard Grafana 5.1.4 Docker image. I have not (yet) seen any limitations in using it for our purposes. I have simply added the provisioning files as configs (Swarm) or config maps (Kubernetes).
Works out of the box on Minikube and Minishift. I haven't tested with OpenShift S2I but personally, I do not see any reason for us to roll our own Grafana image. Maybe you can take a look and see if this works for you.

It is good to know that the EF has no problem with us building custom images but still. I would rather use simply start existing standard images than maintaining our own versions of them. In particular when it comes to images that are not core functionality of Hono. Does this make sense to you?

ctron:
I will have a look at it.

ctron:
Yes, that looks good to me for Swarm, Kubernetes and the simple OpenShift. I was assuming that Docker requires the "VOLUME" declaration for mapping in config files.

However for OpenShift I would prefer to keep the (currently deleted) grafana build template, as the whole idea behind the S2I approach is to build your own images. But we could move this to the `openshift_s2i` folder instead to make it more clear.

sophokles73:
> But we could move this to the openshift_s2i folder instead to make it more clear.

Sounds good to me. Are you updating this PR accordingly or will you create a new one?

ctron:
As long as this only impacts S2I I would directly commit. Otherwise I would open a new PR.

PR #639, opened 2018-06-04
Allow usage of native tls implementation in protocol adapters

Christian-Schmid:
With this PR it is possible to enable the usage of native TLS implementations in protocol adapters.
It does however not yet contain any dependency to a specific implementation.

To use native TLS a user still  has to add a concrete dependency to a specific `netty-tcnative` in the adapters maven dependencies.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/config/ServiceConfigProperties.java
+++ b/core/src/main/java/org/eclipse/hono/config/ServiceConfigProperties.java
@@ -104,10 +104,12 @@ public class ServiceConfigProperties extends AbstractConfig {
     /**
      * Checks if this server requires the usage of a native TLS implementation.
      * Native TLS implementations offer in general a better performance but may not be available on all platforms.
-     * If true, the server will require the usage of a native TLS implementation.
+     * If {@code true}, the server will require the usage of a native TLS implementation.
      * Server will not start if native implementation is not available on the current system.
-     * If false, the adapter will try to use a native TLS implementation. If no native implementation is available the
+     * If {@code false}, the adapter will try to use a native TLS implementation. If no native implementation is available the
      * default Java platform independent TLS implementation will be used.
+     * <p>
+     * The default value of this property is {@code false}.
      *
      * @return {@code true} if the server requires native TLS implementation.
      */
@@ -118,10 +120,12 @@ public class ServiceConfigProperties extends AbstractConfig {
     /**
      * Sets if this server should require the usage of a native TLS implementation.
      * Native TLS implementations offer in general a better performance but may not be available on all platforms.
-     * If true, the server will require the usage of a native TLS implementation.
+     * If {@code true}, the server will require the usage of a native TLS implementation.
      * Server will not start if native implementation is not available on the current system.
-     * If false, the adapter will try to use a native TLS implementation. If no native implementation is available the
+     * If {@code false}, the adapter will try to use a native TLS implementation. If no native implementation is available the
      * default Java platform independent TLS implementation will be used.
+     * <p>
+     * The default value of this property is {@code false}.
      *
      * @param nativeTlsRequired {@code true} if the server requires the usage of a native TLS implementation.
      */
@@ -131,8 +135,8 @@ public class ServiceConfigProperties extends AbstractConfig {
 
     /**
      * Checks if this server is configured to listen on an insecure port (i.e. without TLS) at all.
-     * If false, it is guaranteed by the server that no opened port is insecure.
-     * If true, it enables the definition of an insecure port (as the only port <u>or</u> additionally to the secure port).
+     * If {@code false}, it is guaranteed by the server that no opened port is insecure.
+     * If {@code true}, it enables the definition of an insecure port (as the only port <u>or</u> additionally to the secure port).
      *
      * @return {@code true} if the server guarantees that no opened port is insecure.
      */
@@ -142,8 +146,8 @@ public class ServiceConfigProperties extends AbstractConfig {
 
     /**
      * Sets if this server should support insecure AMQP 1.0 ports (i.e. without TLS) at all.
-     * If false, it is guaranteed by the server that no opened port is insecure.
-     * If true, it enables the definition of an insecure port (as the only port <u>or</u> additionally to the secure port).
+     * If {@code false}, it is guaranteed by the server that no opened port is insecure.
+     * If {@code true}, it enables the definition of an insecure port (as the only port <u>or</u> additionally to the secure port).
      *
      * @param insecurePortEnabled {@code true} if the server shall guarantee that no opened port is insecure.
      */
--- a/core/src/main/java/org/eclipse/hono/config/ServiceConfigProperties.java
+++ b/core/src/main/java/org/eclipse/hono/config/ServiceConfigProperties.java
@@ -33,6 +33,7 @@ public class ServiceConfigProperties extends AbstractConfig {
     private boolean waitForDownstreamConnection = false;
     private String bindAddress = LOOPBACK_DEVICE_ADDRESS;
     private int port = Constants.PORT_UNCONFIGURED;
+    private boolean nativeTlsRequired = false;
     private boolean insecurePortEnabled = false;
     private String insecurePortBindAddress = LOOPBACK_DEVICE_ADDRESS;
     private int insecurePort = Constants.PORT_UNCONFIGURED;
@@ -98,6 +99,34 @@ public class ServiceConfigProperties extends AbstractConfig {
         } else {
             throw new IllegalArgumentException("invalid port number");
         }
+    }
+
+    /**
+     * Checks if this server requires the usage of a native TLS implementation.
+     * Native TLS implementations offer in general a better performance but may not be available on all platforms.
+     * If true, the server will require the usage of a native TLS implementation.
+     * Server will not start if native implementation is not available on the current system.
+     * If false, the adapter will try to use a native TLS implementation. If no native implementation is available the
+     * default Java platform independent TLS implementation will be used.
+     *
+     * @return {@code true} if the server requires native TLS implementation.
+     */
+    public final boolean isNativeTlsRequired() {
+        return nativeTlsRequired;
+    }
+
+    /**
+     * Sets if this server should require the usage of a native TLS implementation.
+     * Native TLS implementations offer in general a better performance but may not be available on all platforms.
+     * If true, the server will require the usage of a native TLS implementation.
+     * Server will not start if native implementation is not available on the current system.
+     * If false, the adapter will try to use a native TLS implementation. If no native implementation is available the
+     * default Java platform independent TLS implementation will be used.
+     *
+     * @param nativeTlsRequired {@code true} if the server requires the usage of a native TLS implementation.
+     */
+    public final void setNativeTlsRequired(final boolean nativeTlsRequired) {
+        this.nativeTlsRequired = nativeTlsRequired;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


service-base/src/main/java/org/eclipse/hono/service/AbstractServiceBase.java
```diff
@@ -315,6 +317,24 @@ protected final void addTlsTrustOptions(final NetServerOptions serverOptions) {
         }
     }
 
+    /**
+     * Configures the TLS engine for the given set of server options.
+     *
+     * @param serverOptions The options to add configuration to.
+     */
+    protected final void addTlsEngineOptions(final NetServerOptions serverOptions) {
```
sophokles73: I think we should do this in the existing *addTlsKeyCertOptions* method so that we do not need to add another method call to all the adpters/components.

service-base/src/main/java/org/eclipse/hono/service/AbstractServiceBase.java
```diff
@@ -315,6 +317,24 @@ protected final void addTlsTrustOptions(final NetServerOptions serverOptions) {
...
+
+        if(getConfig().isNativeTlsEnabled()) {
```
sophokles73: we should also check for `io.netty.handler.ssl.OpenSsl.isAvailable()` here to fall back to JDK ssl if OpenSSL cannot be found.

Christian-Schmid: I was asking myself the same question when implementing, but decided against the fallback. 
When configured with ``true`` a user would have to browse through the logs to find out that actually he isn't using the native implementation. Without the fallback there is an immediate feedback to the user. 

If you like the fallback more we, could then also think about enable the native implementation by default. 


WDYT?

service-base/src/main/java/org/eclipse/hono/service/AbstractServiceBase.java
Christian-Schmid: Agreed, should I also consider changing the name or just put the lines to the existing method?

sophokles73: the latter, please

Christian-Schmid: Adapted change

service-base/src/main/java/org/eclipse/hono/service/AbstractServiceBase.java
Christian-Schmid: Yep, that's true. I renamed the property to {{nativeTlsRequired}}

core/src/main/java/org/eclipse/hono/config/ServiceConfigProperties.java
```diff
@@ -100,6 +101,34 @@ public final void setPort(final int port) {
...
     }
 
+    /**
+     * Checks if this server requires the usage of a native TLS implementation.
+     * Native TLS implementations offer in general a better performance but may not be available on all platforms.
+     * If true, the server will require the usage of a native TLS implementation.
+     * Server will not start if native implementation is not available on the current system.
+     * If false, the adapter will try to use a native TLS implementation. If no native implementation is available the
+     * default Java platform independent TLS implementation will be used.
+     *
```
sophokles73: please use the `{@code true}` formatting for the literal boolean values and also document the default value.

core/src/main/java/org/eclipse/hono/config/ServiceConfigProperties.java
```diff
@@ -100,6 +101,34 @@ public final void setPort(final int port) {
...
+        return nativeTlsRequired;
+    }
+
+    /**
+     * Sets if this server should require the usage of a native TLS implementation.
+     * Native TLS implementations offer in general a better performance but may not be available on all platforms.
+     * If true, the server will require the usage of a native TLS implementation.
+     * Server will not start if native implementation is not available on the current system.
+     * If false, the adapter will try to use a native TLS implementation. If no native implementation is available the
+     * default Java platform independent TLS implementation will be used.
```
sophokles73: please use the `{@code true}` formatting for the literal boolean values and also document the default value.

service-base/src/main/java/org/eclipse/hono/service/AbstractServiceBase.java
```diff
@@ -386,7 +386,8 @@ protected final void addTlsKeyCertOptions(final NetServerOptions serverOptions)
 
             final boolean isOpenSslAvailable = OpenSsl.isAvailable();
             final boolean supportsKeyManagerFactory =  OpenSsl.supportsKeyManagerFactory();
-            final boolean useOpenSsl = isOpenSslAvailable && supportsKeyManagerFactory;
+            final boolean useOpenSsl =
+                    getConfig().isNativeTlsRequired() || (isOpenSslAvailable && supportsKeyManagerFactory);
```
sophokles73: have you tested what happens when *nativeTlsRequired* is `true` but OpenSSL is not available?

Christian-Schmid: Yes, the adapter tries to load OpenSSL anyways and quits then with an `VertxException`: "OpenSSL is not available"

This would be an expected behavior in this case for me...

Stack Trace: 
```
10:01:14.194 [vert.x-eventloop-thread-0] INFO  o.e.h.a.h.v.VertxBasedHttpProtocolAdapter - using OpenSSL [version: null] instead of JDK's default SSL engine
10:01:14.195 [vert.x-eventloop-thread-0] INFO  o.e.h.a.h.v.VertxBasedHttpProtocolAdapter - enabling secure protocol [TLSv1.2]
10:01:14.197 [vert.x-eventloop-thread-0] INFO  o.e.h.a.h.v.VertxBasedHttpProtocolAdapter - enabling client authentication using certificates [org.eclipse.hono.service.auth.TenantApiTrustOptions]
10:01:14.268 [main] ERROR o.e.h.a.h.v.Application$$EnhancerBySpringCGLIB$$8afe9019 - exception occurred during startup, shutting down ...
java.util.concurrent.ExecutionException: io.vertx.core.VertxException: OpenSSL is not available
	at java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:357)
	at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1915)
	at org.eclipse.hono.service.AbstractApplication.run(AbstractApplication.java:148)
	at org.springframework.boot.SpringApplication.callRunner(SpringApplication.java:723)
	at org.springframework.boot.SpringApplication.callRunners(SpringApplication.java:713)
	at org.springframework.boot.SpringApplication.afterRefresh(SpringApplication.java:703)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:304)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:1118)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:1107)
	at org.eclipse.hono.adapter.http.vertx.Application.main(Application.java:34)
Caused by: io.vertx.core.VertxException: OpenSSL is not available
	at io.vertx.core.net.impl.SSLHelper.resolveEngineOptions(SSLHelper.java:74)
	at io.vertx.core.net.impl.SSLHelper.<init>(SSLHelper.java:163)
	at io.vertx.core.http.impl.HttpServerImpl.<init>(HttpServerImpl.java:143)
	at io.vertx.core.impl.VertxImpl.createHttpServer(VertxImpl.java:289)
	at org.eclipse.hono.adapter.http.AbstractVertxBasedHttpProtocolAdapter.bindSecureHttpServer(AbstractVertxBasedHttpProtocolAdapter.java:361)
	at org.eclipse.hono.adapter.http.AbstractVertxBasedHttpProtocolAdapter.lambda$doStart$1(AbstractVertxBasedHttpProtocolAdapter.java:180)
	at io.vertx.core.Future.lambda$compose$1(Future.java:265)
	at io.vertx.core.impl.FutureImpl.setHandler(FutureImpl.java:76)
	at io.vertx.core.Future.compose(Future.java:261)
	at org.eclipse.hono.adapter.http.AbstractVertxBasedHttpProtocolAdapter.doStart(AbstractVertxBasedHttpProtocolAdapter.java:168)
	at org.eclipse.hono.service.AbstractProtocolAdapterBase.startInternal(AbstractProtocolAdapterBase.java:259)
	at org.eclipse.hono.service.AbstractServiceBase.start(AbstractServiceBase.java:66)
	at io.vertx.core.impl.DeploymentManager.lambda$doDeploy$8(DeploymentManager.java:483)
	at io.vertx.core.impl.ContextImpl.lambda$wrapTask$2(ContextImpl.java:339)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:404)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:463)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:886)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.ClassNotFoundException: io.netty.internal.tcnative.SSL
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:348)
	at io.netty.handler.ssl.OpenSsl.<clinit>(OpenSsl.java:77)
	at org.eclipse.hono.connection.impl.ConnectionFactoryImpl.addTlsTrustOptions(ConnectionFactoryImpl.java:218)
	at org.eclipse.hono.connection.impl.ConnectionFactoryImpl.addOptions(ConnectionFactoryImpl.java:195)
	at org.eclipse.hono.connection.impl.ConnectionFactoryImpl.connect(ConnectionFactoryImpl.java:127)
	at org.eclipse.hono.connection.impl.ConnectionFactoryImpl.connect(ConnectionFactoryImpl.java:105)
	at org.eclipse.hono.client.impl.HonoClientImpl.lambda$connect$5(HonoClientImpl.java:324)
	... 7 common frames omitted
```

sophokles73: Hmm, and this is the desired behavior, that the adapter doesn't start at all in this case?
Why shouldn't it be better to start using the JDK's default SSL engine, thus providing a more *graceful degradation*? I understand that you want to verify that the adapter indeed starts up with OpenSSL but you can also find that out by taking a look at the logs during startup ...

ctron: Well I would say that if you want OpenSSL and it isn't available it should fail. And then you can check the logs. People tend not to look into logs unless something bad happens.

So if you have "bad performance" most likely no one will check the logs for that entry. However if the pod fails to start, then people will take a look.

It would be different if you want a "default" behavior, in which case I would agree that downgrading the a different implementation is a good idea, but not if you explicitly ask for a feature.

Christian-Schmid: I agree with ctron. Also from a DevOps perspective I like to have an immediate feedback, that I messed up something with my deployment. If the connector starts it will take a while till this failure is detected ... (E.g. after running the load tests). 

Everyone else who doesn't care about this will anyways not set the `nativeTlsRequired` property.

sophokles73: ok, just wanted to be sure you really mean it :-)

PR #640, opened 2018-06-04
[#643] AMQP Protocol Adapter

Alfusainey:
This is a minimal initial implementation of the amqp adapter (with minimal tests). with this setup, i was able to manually test sending telemetry messages to the downstream example receiver using:
```shell
qpid-send -b localhost:4040 --connection-options "{protocol: amqp1.0}" -a "'telemetry/DEFAULT_TENANT/4712'" --content-string "'{"temp": 5}'
```

what i now want to consider is having different sessions (and flow control) for events, separate from telemetry data and to add more tests.

/cc: @sophokles73 @sysexcontrol @ppatierno


adapters/hono-adapter-amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,382 @@
...
+            this.insecureServer = server;
+        }
+    }
+    /**
+     * This method is called when an AMQP BEGIN frame is received from a remote client. This method sets the incoming
+     * capacity in its BEGIN Frame to be communicated to the remote peer
+     *
+     */
+    private void handleSessionOpen(final ProtonConnection conn, final ProtonSession session) {
+        LOG.debug(String.format("opening new session with client [container: {}]", conn.getRemoteContainer()));
```
sophokles73: no need to use String.format in a logging statement. Logger.debug supports placeholder substitution natively, e.g.
`LOG.debug("opening new session with client [container: {}]", conn.getRemoteContainer());`
In fact, your statement does not work because String.format expects placeholders like `%s` for Strings or `%d` for Numbers

adapters/hono-adapter-amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,382 @@
...
+        ResourceIdentifier address = ResourceIdentifier.fromString(receiver.getRemoteTarget().getAddress());
+        switch (EndpointType.fromString(address.getEndpoint())) {
+        case TELEMETRY:
+            if (!ProtonQoS.AT_LEAST_ONCE
+                    .equals(receiver.getRemoteQoS()) && !ProtonQoS.AT_MOST_ONCE.equals(receiver.getRemoteQoS())) {
+                result.fail(new AmqpErrorException("unsupported-delivery-mode",
+                        "Telemetry endpoint only supports AT_LEAST_ONCE and AT_MOST_ONCE delivery semantics."));
+            } else {
+                // This adapter only supports AT_MOST_ONCE for telemetry?
+                receiver.setQoS(ProtonQoS.AT_MOST_ONCE);
```
sophokles73: IMHO you should set the QoS desired by the device

adapters/hono-adapter-amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,382 @@
...
+                        "Event endpoint only supports AT_LEAST_ONCE delivery semantics."));
+            } else {
+                receiver.setQoS(ProtonQoS.AT_LEAST_ONCE);
+                result.complete(address);
+            }
+            break;
+        default:
+            LOG.error(String.format("Endpoint with [name: %s] is not supported by this adapter ",
+                    address.getEndpoint()));
+            result.fail(new AmqpErrorException("unsupported-delivery-mode",
```
sophokles73: unsupported-delivery-mode ? This should probably result in the standard `amqp:not-found` error, shouldn't it?

adapters/hono-adapter-amqp-vertx/src/main/resources/logback-spring.xml
```diff
@@ -0,0 +1,39 @@
...
+      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+    </encoder>
+  </appender>
+
+  <root level="INFO">
+    <appender-ref ref="STDOUT" />
+  </root>
+
+  <springProfile name="dev">
+    <logger name="org.eclipse.hono.adapter" level="INFO"/>
```
sophokles73: at least the adapter classes should log at DEBUG level in *dev* profile ...
please set analogously to the other protocol adapters

adapters/hono-adapter-amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -0,0 +1,284 @@
...
+        startup.await();
+        verify(server).connectHandler(any(Handler.class));
+        verify(server).listen(any(Handler.class));
+    }
+
+    /**
+     * Verifies that the AMQP Adapter does not support anonymous relay.
+     */
+    @Test
+    public void anonymousRelyNotSupported() {
```
sophokles73: Relay

adapters/hono-adapter-amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -0,0 +1,284 @@
...
+        // GIVEN an AMQP adapter with a configured server.
+        final ProtonServer server = getAmqpServer();
+        final VertxBasedAmqpProtocolAdapter adapter = getAdapter(server);
+
+        // WHEN the adapter receives a request from a link
+        // that does not specify a target address
+        ProtonReceiver receiver = mock(ProtonReceiver.class);
```
sophokles73: final

adapters/hono-adapter-amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -0,0 +1,284 @@
...
+        adapter.handleRemoteReceiverOpen(receiver, mock(ProtonConnection.class));
+
+        // THEN the adapter rejects the message and closes the link.
```
sophokles73: no message involved here

adapters/hono-adapter-amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -0,0 +1,284 @@
...
+
+    /**
+     * Verifies that the adapter fails the upload of a telemetry message if the tenant for which a device belongs to has
+     * been disabled.
+     */
+    @Test
+    public void testUploadTelemetryFailsForDisabledTenant() {
+        // Given an adapter
+        final ProtonServer server = getAmqpServer();
+        MessageSender sender = givenAmessageSenderForAnyTenant();
```
sophokles73: final

adapters/hono-adapter-amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -0,0 +1,284 @@
...
+        // THAT is disabled for a tenant
+        givenAConfiguredTenant("disabled-tenant", false);
+        final VertxBasedAmqpProtocolAdapter adapter = getAdapter(server);
+
+        // When a device that belongs to such a tenant uploads a telemetry message
+        final ProtonDelivery delivery = mock(ProtonDelivery.class);
+
+        final Message message = getFakeMessage();
+        final ResourceIdentifier resource = ResourceIdentifier.from("telemetry", "disabled-tenant", "the-device");
+        AmqpContext context = spy(new AmqpContext(delivery, message, resource));
```
sophokles73: final

adapters/hono-adapter-amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -0,0 +1,284 @@
...
+    public void uploadTelemetryMessageWithSettledDeliverySemantics() {
+        // GIVEN an AMQP adapter with a configured server
+        final VertxBasedAmqpProtocolAdapter adapter = givenAnAmqpAdapter();
+        final MessageSender sender = givenAmessageSenderForAnyTenant();
+
+        // which is enabled for a tenant
+        givenAConfiguredTenant("some-tenant", true);
+
+        // IF a device sends a 'fire and forget' telemetry message
+        ProtonDelivery delivery = mock(ProtonDelivery.class);
```
sophokles73: final

adapters/hono-adapter-amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -0,0 +1,284 @@
...
+        adapter.setHonoMessagingClient(messagingServiceClient);
+        adapter.setRegistrationServiceClient(registrationServiceClient);
+        adapter.setCredentialsServiceClient(credentialsServiceClient);
+        adapter.setCommandConnection(commandConnection);
+        return adapter;
+    }
+
+    @SuppressWarnings("unchecked")
+    private ProtonServer getAmqpServer() {
+        ProtonServer server = mock(ProtonServer.class);
```
sophokles73: final

adapters/hono-adapter-amqp-vertx/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,25 @@
...
+      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+    </encoder>
+  </appender>
+
+  <root level="INFO">
+    <appender-ref ref="STDOUT" />
+  </root>
+
+  <logger name="org.eclipse.hono.service" level="INFO"/>
+  <logger name="org.eclipse.hono.service" level="DEBUG"/>
```
sophokles73: the default should be INFO IMHO

adapters/pom.xml
```diff
@@ -32,6 +32,7 @@
     <module>mqtt-vertx</module>
     <module>mqtt-vertx-base</module>
     <module>kura</module>
+    <module>hono-adapter-amqp-vertx</module>
```
sophokles73: can you name the module in analogy to the other adapter modules, please, i.e. `amqp-vertx`

adapters/hono-adapter-amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapterTest.java
ctron: Should we add `final` to checkstyle?

sophokles73: > Should we add final to checkstyle?

I believe so.

> maybe also adding license/copyright headers to checkstyle too?

One step after the other ;-) But seriously, I am not sure if we can come up with a good regex that is both lenient enough to allow for slight variations but strict enough to find the real positives.

ctron: > > Should we add final to checkstyle?
> I believe so.

Ok I will work on this.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Gets the content type of the AMQP 1.0 message.
+     * 
+     * @return The content type of the AMQP 1.0 message.
+     */
+    String getMessageContentType() {
+        return message.getContentType();
+    }
+
+    ProtonDelivery delivery() {
```
sophokles73: are these methods intentionally declared package private?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/Config.java
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Exposes the configuration properties of the AMQP adapter as a Spring bean.
+     * 
+     * @return The configuration properties.
+     */
+    @Bean
+    @ConfigurationProperties(prefix = "hono.amqp")
+    public ProtocolAdapterProperties adapterProperties() {
+        ProtocolAdapterProperties config = new ProtocolAdapterProperties();
```
sophokles73: final

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/Config.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    /**
+     * Exposes a factory for creating AMQP adapter instances.
+     * 
+     * @return The factory bean.
+     */
+    @Bean
+    public ObjectFactoryCreatingFactoryBean serviceFactory() {
+        ObjectFactoryCreatingFactoryBean factory = new ObjectFactoryCreatingFactoryBean();
```
sophokles73: final

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,379 @@
...
+    protected String getTypeName() {
+        return Constants.PROTOCOL_ADAPTER_TYPE_AMQP;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected void doStart(final Future<Void> startFuture) {
+        ProtonServerOptions options = new ProtonServerOptions();
```
sophokles73: final

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,379 @@
...
+
+    }
+
+    @Override
+    protected void doStop(final Future<Void> stopFuture) {
+        stopInsecureServer().compose(stopped -> stopFuture.complete(), stopFuture);
+    }
+
+    private Future<Void> stopInsecureServer() {
+        Future<Void> result = Future.future();
```
sophokles73: final

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,379 @@
...
+        return result;
+    }
+
+    private Future<ProtonServer> bindServer(final ProtonServer server, final ProtonServerOptions options) {
+        final Future<ProtonServer> result = Future.future();
+        final ProtonServer createdServer = (server == null) ? ProtonServer.create(this.vertx, options) : server;
+
+        createdServer.connectHandler(this::connectionRequestHandler).listen(res -> {
+            if (res.succeeded()) {
+                LOG.info(String.format("AMQP server running on %s:%d", getConfig().getInsecurePortBindAddress(),
```
sophokles73: remove String.format and use logger's capabilities for placeholder injection

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,379 @@
...
+    private void connectionRequestHandler(final ProtonConnection connRequest) {
+        LOG.info("Received connection request from client");
+        connRequest.disconnectHandler(conn -> {
+            LOG.error("Connection disconnected " + conn.getCondition().getDescription());
+        });
+        // when a BEGIN frame from a client
+        connRequest.sessionOpenHandler(session -> handleSessionOpen(connRequest, session));
+        // when an OPEN is received
+        connRequest.openHandler(remoteOpen -> {
+            ProtonConnection conn = remoteOpen.result();
```
sophokles73: final

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,379 @@
...
+            case EVENT:
+                LOG.info(String.format("Received request to upload events at {}", context.getEndpoint()));
+                doUploadMessage(context, getEventSender(context.getTenantId()), EventConstants.EVENT_ENDPOINT);
+                break;
+            default:
+                return Future.failedFuture(new AmqpErrorException(AmqpError.NOT_ALLOWED, "unknown endpoint"));
+            }
+            return Future.succeededFuture();
+        }).otherwise(t -> {
+            AmqpErrorException error = (AmqpErrorException) t;
```
sophokles73: final

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,379 @@
...
+                        receiver.closeHandler(remoteDetach -> onLinkDetach(receiver));
+                        receiver.open();
+                        return Future.succeededFuture();
+                    });
+        }
+    }
+
+    protected void uploadMessage(final ProtonConnection con, final AmqpContext context) {
+        final Future<Void> formalCheck = Future.future();
+        String contentType = context.getMessageContentType();
```
sophokles73: final

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,379 @@
...
+                    // client uses AT_MOST_ONCE delivery semantics -> fire and forget
+                    return sender.send(downstreamMessage);
+                } else {
+                    // client uses AT_LEAST_ONCE delivery semantics
+                    return sender.sendAndWaitForOutcome(downstreamMessage);
+                }
+            } else {
+                // this adapter is not enabled for tenant
+                return Future.failedFuture(
+                        new AmqpErrorException(AmqpError.NOT_ALLOWED, "This adapter is not enabled for tenant."));
```
sophokles73: IMHO we should use a ClientErrorException with a 403 code here in order to be consistent with the other adapters' forwarding code.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,379 @@
...
+                context.delivery().disposition(downstreamDelivery.getRemoteState(),
+                        downstreamDelivery.remotelySettled());
+            }
+            return Future.<Void> succeededFuture();
+        }).recover(t -> {
+            LOG.info("Cannot process message for [deviceId: {}] of [tenantId: {}]", context.getDeviceId(),
+                    context.getTenantId());
+            if (AmqpErrorException.class.isInstance(t)) {
+                AmqpErrorException error = (AmqpErrorException) t;
+                context.reject(error);
```
sophokles73: I like the idea of encapsulating the message disposition handling within the AmqpContext class. However, IMHO you should then either move all of the disposition handling code into this class or have that class provide methods for accepting, releasing and sending disposition frames as well

adapters/amqp-vertx/src/main/resources/logback-spring.xml
```diff
@@ -0,0 +1,43 @@
...
+  <springProfile name="prod">
+    <logger name="org.eclipse.hono" level="INFO"/>
+  </springProfile>
+
+  <logger name="io.netty.handler.logging.LoggingHandler" level="INFO"/>
+
+  <logger name="io.vertx.proton.impl" level="INFO"/>
+  <logger name="io.vertx.core.net.impl" level="INFO"/>
+
+</configuration>
```
sophokles73: missing EOL

Alfusainey:
yes, that make sense! thank you.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
Alfusainey: yes, at least for now.

sophokles73: Ok, a JavaDoc would still be helpful, I guess :-)

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
Alfusainey: ok

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
Alfusainey: yes

Alfusainey:
@sophokles73 updated the PR. this change contains a new test class that simulates a device, using a protonclient, for testing

sophokles73:
@Alfusainey it is not helpful to add more (and different) functionality while we are still discussing the original PR. Can you please remove the added test class and instead address the issues I had raised in the original PR? You can then later create additional PRs for adding more test cases etc.

Alfusainey:
@sophokles73 Okay, i will remove the added test class. But, this pull request also fixes all the issues you raised in the pull request (in addition to the new test class).

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,384 @@
...
+        if (isPayloadOfIndicatedType(context.getMessagePayload(), contentType)) {
+            formalCheck.complete();
+        } else {
+            formalCheck.fail(new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST,
+                    String.format("Content-Type: %s does not match payload", contentType)));
+        }
+        formalCheck.compose(ok -> {
+            switch (EndpointType.fromString(context.getEndpoint())) {
+            case TELEMETRY:
+                LOG.info("Received request to upload telemetry data to endpoint [with name: {}]",
```
sophokles73: this should be TRACE level

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,384 @@
...
+                        context.getEndpoint());
+                doUploadMessage(context, getTelemetrySender(context.getTenantId()),
+                        TelemetryConstants.TELEMETRY_ENDPOINT);
+                break;
+            case EVENT:
+                LOG.info("Received request to upload events to endpoint [with name: {}]", context.getEndpoint());
```
sophokles73: TRACE

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,384 @@
...
+                    return sender.sendAndWaitForOutcome(downstreamMessage);
+                }
+            } else {
+                // this adapter is not enabled for tenant
+                return Future.failedFuture(
+                        new ClientErrorException(HttpURLConnection.HTTP_NOT_FOUND,
+                                "This adapter is not enabled for tenant."));
+            }
+        }).compose(downstreamDelivery -> {
+            LOG.info("Successfully process message for Device [deviceId: {}] with Tenant [tenantId: {}]",
```
sophokles73: TRACE

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,384 @@
...
+                    // client uses AT_MOST_ONCE delivery semantics -> fire and forget
+                    return sender.send(downstreamMessage);
+                } else {
+                    // client uses AT_LEAST_ONCE delivery semantics
+                    return sender.sendAndWaitForOutcome(downstreamMessage);
+                }
+            } else {
+                // this adapter is not enabled for tenant
+                return Future.failedFuture(
+                        new ClientErrorException(HttpURLConnection.HTTP_NOT_FOUND,
```
sophokles73: we use 403 for this case in the other adapters

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,384 @@
...
+                context.accept();
+            } else {
+                // client uses AT_LEAST_ONCE delivery semantics
+                // forward disposition received from downstream peer back to the client device.
+                context.updateDelivery(downstreamDelivery.getRemoteState(),
+                        downstreamDelivery.remotelySettled());
+            }
+            return Future.<Void> succeededFuture();
+        }).recover(t -> {
+            LOG.info("Cannot process message for Device [deviceId: {}] with Tenant [tenantId: {}]",
```
sophokles73: DEBUG

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -0,0 +1,384 @@
...
+                    context.getDeviceId(),
+                    context.getTenantId());
+            if (!context.remotelySettled()) {
+                // client wants to be informed that the message cannot be processed.
+                context.released();
```
sophokles73: shouldn't we reject the message in this case and include an error description?

Alfusainey: Oay! but maybe by releasing it, we are telling the client that the message is okay but the adapter cannot process it (because the sender queue was full, ...). The client can then try to redeliver the message.
IMHO, rejecting the message could signal to the client that something is actually wrong with the message. WDYT?

sophokles73: Yes, I had the same thought. So it seems like we should distinguish between these two cases and then either release or reject the message, right? A ClientErrorException is used to indicate that the client has made an error (-> reject ?) while a ServerErrorException is used to indicate that there is a processing problem not caused by the client (-> release ?) ...

Alfusainey: > A ClientErrorException is used to indicate that the client has made an error (-> reject ?) while a ServerErrorException is used to indicate that there is a processing problem not caused by the client (-> release ?)

Yes, exactly! will update the PR

Alfusainey:
@sophokles73: PR updated..

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -238,7 +240,7 @@ protected void uploadMessage(final AmqpContext context) {
             return Future.succeededFuture();
         }).otherwise(t -> {
             final ClientErrorException error = (ClientErrorException) t;
-            context.reject(error);
+            context.reject(AmqpError.NOT_ALLOWED, error);
```
sophokles73: It seems like the ClientErrorException will always contain a 400 status code. Do we really want to map this code to AMQP's not-allowed error? The AMQP spec defines `amqp:not-allowed` as
> The peer tried to use a frame in a manner that is inconsistent with the semantics defined in
the specification.

This doesn't seem to be a good match, does it? Maybe we should create our own error here. What about `hono:bad-request`?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -289,12 +292,26 @@ private void doUploadMessage(final AmqpContext context, final Future<MessageSend
...
+
+                if (ServerErrorException.class.isInstance(t)) {
+                    errorMsg.append("Releasing message: {} - {}");
+                    context.released();
+                } else {
+                    errorMsg.append("Rejecting message : {} - {}");
+                    context.reject(AmqpError.NOT_FOUND, (ClientErrorException) error);
+                }
+
+                LOG.debug(errorMsg.toString(), error.getErrorCode(), error.getMessage());
```
sophokles73: please create only one log statement for the situation. Using multiple statements will make it close to impossible to correlate them if we are receiving more than one message per second ...

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -289,12 +292,26 @@ private void doUploadMessage(final AmqpContext context, final Future<MessageSend
...
+                // t can only be a ServiceInvocationException
+                final ServiceInvocationException error = (ServiceInvocationException) t;
+                final StringBuilder errorMsg = new StringBuilder();
+
+                if (ServerErrorException.class.isInstance(t)) {
+                    errorMsg.append("Releasing message: {} - {}");
+                    context.released();
+                } else {
+                    errorMsg.append("Rejecting message : {} - {}");
+                    context.reject(AmqpError.NOT_FOUND, (ClientErrorException) error);
```
sophokles73: IMHO we should try to have only a single place in the code where we map ServiceInvocationExceptions (or Throwables in general) to AmqpErrors. That place should probably be the AmqpContext's *reject* method.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
Alfusainey: yes, i thought so too. thanks

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
Alfusainey: i thought we are trying to reject only in the case of a client error not a server error? so we exclude mapping server errors and just release?

the possible error codes that may be reported while uploading a message:
1. Bad Request (400) maps to `hono:bad-request` (client-error)
2. Forbidden (403) maps to `amqp:not-found` (adapter is disabled for tenant - client error)
3. Internal Error (500) maps to `amqp:internal-error` (peer did not settle message - server error)
4. Unavailable (503) maps to `amqp:precondition-failed` (no credit available)

WDYT?

sophokles73: Sure, we only reject messages that caused a ClientErrorException and we release messages that caused a ServerErrorException. Still, we could have a single method in `AmqpContext` which handles a `ServiceInvocationException` and thus hides this policy and the concrete mapping to AMQP errors, right? BTW, we should probably map 403 to the standard `amqp:unauthorized-access`, don't you think? 500 and 503 do not need to be mapped because we are simply releasing messages for status codes in the 500s, right?

Alfusainey: > Still, we could have a single method in AmqpContext which handles a ServiceInvocationException and thus hides this policy and the concrete mapping to AMQP errors, right?

I see! If I understand correctly, we now only need one call, say `AmqpContext#handleFailure(SIE)`, and do not need explicit `reject` and `release` calls for dispositions?

> amqp:unauthorized-access -> 403

Sure. i was unable to relate the `security-settings` wording in the spec to a disabled adapter.

sophokles73: that would make sense to me :-)

Alfusainey: ok, PR update

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -136,4 +140,24 @@ ProtonDelivery updateDelivery(final DeliveryState state, final boolean settle) {
...
+     */
+    private ErrorCondition getErrorCondition(final ServiceInvocationException error) {
+        switch (error.getErrorCode()) {
+        case HttpURLConnection.HTTP_BAD_REQUEST:
+            return ProtonHelper.condition(Constants.AMQP_BAD_REQUEST, error.getMessage());
+        case HttpURLConnection.HTTP_FORBIDDEN:
+            return ProtonHelper.condition(AmqpError.UNAUTHORIZED_ACCESS, error.getMessage());
+        default:
+            // TODO: we default to null?
+            return null;
```
sophokles73: we need to return an ErrorCondition in any case. I am sure you will be able to come up with a reasonable default.

core/src/main/java/org/eclipse/hono/util/Constants.java
```diff
@@ -57,6 +57,12 @@
      */
     public static final String PROTOCOL_ADAPTER_TYPE_AMQP= "hono-amqp";
 
+    /**
+     * Indicates that the AMQP adapter cannot process the request due to a client error. A bad inbound
```
sophokles73: if this constant is intended to be used by the AMQP adapter only, then why don't we define the constant there? Or is this constant intended to be used by any type of server exposing an AMQP endpoint instead?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -87,15 +95,21 @@ ResourceIdentifier getResourceIdentifier() {
...
      */
-    void reject(final Symbol condition, final ClientErrorException error) {
-        final ErrorCondition errorCondition = ProtonHelper.condition(condition, error.getMessage());
-        MessageHelper.rejected(delivery, errorCondition);
+    void handleFailure(final ServiceInvocationException error) {
+        if (ServerErrorException.class.isInstance(error)) {
+            ProtonHelper.released(delivery, true);
+        } else {
+            final ErrorCondition condition = getErrorCondition(error);
+            if (condition != null) {
```
sophokles73: we need to reject the message in any case, don't we?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
Alfusainey: ok, i was thinking of using `amqp:precondition-failed` since if there is a client error, then there is something the adapter expects which the client is not providing?

core/src/main/java/org/eclipse/hono/util/Constants.java
Alfusainey: ok, sure

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
sophokles73: makes sense

core/src/main/java/org/eclipse/hono/util/Constants.java
sophokles73: If it is not intended to be used by other components then move it to the AMQP adapter, otherwise adapt the JavaDoc.

Alfusainey:
@sophokles73: PR updated!

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -87,15 +95,19 @@ ResourceIdentifier getResourceIdentifier() {
...
+     * the error condition value and description as reason for rejection.
      *
-     * @param condition The error condition value to include in the Rejected disposition frame.
-     * @param error The client exception providing a reason for rejecting the message.
+     * @param error The service invocation exception.
      */
-    void reject(final Symbol condition, final ClientErrorException error) {
-        final ErrorCondition errorCondition = ProtonHelper.condition(condition, error.getMessage());
-        MessageHelper.rejected(delivery, errorCondition);
+    void handleFailure(final ServiceInvocationException error) {
```
sophokles73: what if error is `null`

Alfusainey: `handleFailure` expects that the throwable is not `null`. so i can add a null-check at the beginning and modify the JavaDoc. 
question: how can it ever be `null`?

sophokles73: > question: how can it ever be null?

well, we cannot prevent anyone from creating a new class in the same package and invoke the method passing in `null`, can we? Just because *we* are not doing it, doesn't mean that anybody else isn't doing it. Documenting the behavior in case someone does is a safe way of encouraging people to actually care, because they know what to expect from the method.

Alfusainey: :+1:

Alfusainey:
@sophokles73: PR updated!

Alfusainey:
ok, will do that now

sophokles73:
You can squash all commits but please reference the issue number in the head line and don't call it *initial commit* but instead describe what the commit contains.

Alfusainey:
@sophokles73: its done

sophokles73:
Okay, I'll wait for the Travis build to succeed and then merge tomorrow morning ...

sophokles73:
Thanks for contributing :-)

Alfusainey:
sure :+1:

PR #641, opened 2018-06-05
Check for final local variables

ctron:


ctron:
Merged this is manually.

Issue #645, opened 2018-06-07
Document current metrics


ctron:
We want to document the current metrics which Hono considers an API and which can be used by users to build upon them. This includes the structure, naming, and what to expect from them metrics.

That information should be part of the 0.7 release.

sophokles73:
@ctron can this be closed?

ctron:
Yes

Issue #642, opened 2018-06-06
Migrate metrics to Prometheus


ctron:
For internal metrics we want to use Prometheus.

ctron:
Also see: https://dev.eclipse.org/mhonarc/lists/hono-dev/msg00852.html

ctron:
Decisions from today's project call:

* We need to keep the current metrics structure in the InfluxDB for at least the next release of Hono
  * The first step is to document the existing metrics – #645
* We do not necessarily need the graphite reporter for that as long as the InfluxDB structure can be provided in the same way
* The upgrade to Spring Boot 2 will be postponed and not done before this topic
* We will investigate if we can switch to Micrometer and drop the workaround of the current metrics naming and re-labeling using the graphite exporter. Possibly using a direct Micrometer to InfluxDB (and Prometheus) reporter
* The best possible solution would be to ship Hono 0.7 with two reporter options, InfluxDB and Prometheus
* We will re-visit this topic in the next project calls and continue discussions as we learn more

sophokles73:
fixed in https://github.com/eclipse/hono/commit/dbffcfaa68cc26c4a37cb66f57216722df4420a4

PR #648, opened 2018-06-08
[#645]: Document current metrics structure

ctron:
This is a first version of documenting the existing Hono metrics which are considered API.

Currently it leaves out:
* Everything from the Kura adapter
* The AMQP specific metrics from the messaging service (e.g. `connections.downstream`, `link.downstream.credits`, …)

Leaving those metrics out doesn't mean we drop them right away, but we might change them between releases without prior notice.

Please let me know if you need any of the above (left out) metrics in the documented list or if I overlooked something.

### DIFF
--- a/site/content/api/Metrics.md
+++ b/site/content/api/Metrics.md
@@ -20,9 +20,9 @@ this document. However they are not considered part of the interface to Hono and
 
 When metrics via InfluxDB is enabled, then Hono offers the following metrics
 in the Hono-local InfluxDB. Metrics are sent to InfluxDB via the Graphite
-reporter, however the Graphite reporter is not part of the guarantueed Hono
-interfacea. Only the structure of the InfluxDB is. This means that in a future
-version Hono might use a different mechanism to feed metrics in the InfluxDB,
+reporter, however the Graphite reporter is not part of the guaranteed Hono
+interface. Only the structure of the InfluxDB is. This means that a future
+version Hono might use a different mechanism to feed metrics into the InfluxDB,
 but the structure inside the InfluxDB will stay stable.
 
 ### Tags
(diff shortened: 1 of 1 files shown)


pellmann:
I am not sure we should define our current metrics names, which are reported with Graphite as "API". What is the value with this, if at the end the data in the different data stores looks very different and the queries also do?

The "old" long names with the point based delimiters, are not good to stay the same over time. This is one reason I like a tag based system more.

I propose to declare only the new tags as "API", when we introduce them.

If we do not name the tags, that map to influxDB, we need to show the whole names and mark, which parts will change (the parts, that maps to the tags).

ctron:
My understanding the last call (as documented in https://github.com/eclipse/hono/issues/642#issuecomment-395451476) was that we define the format as currently stored in the InfluxDB as API. So that we could drop Dropwizard, re-use something based on Micrometer, and don't rely on Graphite. As the whole dropwizard-to-graphite-to-influx is only a workaround around the missing tags feature. My understanding was also that Graphite is not important, but InfluxDB is.

Of course we can switch that to define Graphite as API and drop the InfluxDB structure in that document, which would however have an impact on the next steps in a way that we need to support Graphite instead of InfluxDB. And also could not directly write to InfluxDB.

For Prometheus the plan is to actually use the Prometheus wire format. I am not sure if that has a name of its own, so keeping it "Prometheus" sounds reasonable to me.

pellmann:
I also think the important part is that the structure in the InfluxDB remains the same in 0.7. And for this it is a combination of the graphite metric names and the tag mapping. I just think to call this "API" is a little strange and would propose to just be sure the actual data, written in InfluxDB stays as it is and we call the tags of the next version, the "API". WDYT?

ctron:
Well I am open to changing the term "API" to something different. It however is some kind of interface which Hono offers and on which people can rely on.

The problem is, on what can people rely? On the InfluxDB format? Or on the graphite reporter format? I think it will get pretty messy in the Hono code if we try to stick to the current Graphite format and add Prometheus in a reasonable way.

So I am not sure I get what you are aiming for. My goal (up until now) is to describe the current metrics structure as they are stored in the InfluxDB. And we can keep this stable over a number of Hono releases.

Right now there is no abstraction layer in the Hono metrics components. So we can't really switch to something else without adding one. Which would be fine as well IMHO.

pellmann:
> So I am not sure I get what you are aiming for. My goal (up until now) is to describe the current metrics structure as they are stored in the InfluxDB. And we can keep this stable over a number of Hono releases.

We want the same :) - my comment was more against the Graphite format as the API because this would not be enough to get the same data in InfluxDB (I think that proposal was in a comment, that is gone now). So thumbs up from my side to describe the format in the InfluxDB.

ctron:
@pellmann I did rephrase a few things. Does that work better for you now? The last section "Metrics API" is obviously something to provide in the future.

site/content/api/Metrics.md
```diff
@@ -0,0 +1,77 @@
...
+this document. However they are not considered part of the interface to Hono and
+ may change over time without notice.
+{{%/note%}}
+
+## InfluxDB
+
+When metrics via InfluxDB is enabled, then Hono offers the following metrics
+in the Hono-local InfluxDB. Metrics are sent to InfluxDB via the Graphite
+reporter, however the Graphite reporter is not part of the guarantueed Hono
+interfacea. Only the structure of the InfluxDB is. This means that in a future
```
pellmann: Typo: interfacea

pellmann: Also the 'in' is obsolete.

site/content/api/Metrics.md
```diff
@@ -0,0 +1,77 @@
...
+| meter.hono.mqtt.messages.processed.m1_rate | host, tenant, type | Messages processed by the MQTT protocol adapter. One minute, exponentially weighted, moving average. |
+| meter.hono.mqtt.messages.processed.m5_rate | host, tenant, type | Messages processed by the MQTT protocol adapter. Five minute, exponentially weighted, moving average. |
+| meter.hono.mqtt.messages.processed.m15_rate | host, tenant, type | Messages processed by the MQTT protocol adapter. Fifteen minute, exponentially weighted, moving average. |
+| meter.hono.mqtt.messages.processed.mean_rate | host, tenant, type | Messages processed by the MQTT protocol adapter. Mean rate of messages since the application start. |
+
+### Messaging Metrics
+
+| Metric | Tags | Description |
+| ------ | ---- | ----------- |
+| counter.hono.messaging.messages.discarded.count | host, tenant, type | Messages which could not be processed by the messaging component. Total count since application startup. |
```
pellmann: We have 'discarded' and 'undeliverable' depending on the settled state (event/telemetry).

ctron: True. However we discussed that we want to remove the messaging component altogether. So I think we actually should remove the whole section, as we will no longer support this. Right?

site/content/api/Metrics.md
ctron: Done.

sophokles73:
@pellmann can you take a look again and approve?

PR #651, opened 2018-06-11
Enable JDK 10 build on Travis in addition to JDK 8

ctron:
This change also contains the necessary upgrades and fixes for
successfully running the unit tests on Travis.

---

Please see this a a preview PR, which we can use for discussions.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -37,6 +37,8 @@
     <influxdb.version>1.5.4-alpine</influxdb.version>
     <jackson.version>2.9.5</jackson.version>
     <java-base-image.name>openjdk:8u171-jre-slim</java-base-image.name>
+    <jaxb.api.version>2.2.12</jaxb.api.version>
+    <javax.annotation.api.version>1.3.2</javax.annotation.api.version>
     <jjwt.version>0.7.0</jjwt.version>
     <jmeter.version>3.3</jmeter.version>
     <junit.version>4.12</junit.version>
@@ -242,6 +244,12 @@
         <version>${jjwt.version}</version>
       </dependency>
       <dependency>
+        <!-- required as a (missing) transient dependency for JJWT -->
+        <groupId>javax.xml.bind</groupId>
+        <artifactId>jaxb-api</artifactId>
+        <version>${jaxb.api.version}</version>
+      </dependency>
+      <dependency>
         <groupId>org.apache.jmeter</groupId>
         <artifactId>ApacheJMeter_core</artifactId>
         <version>${jmeter.version}</version>
@@ -271,6 +279,11 @@
         <groupId>org.eclipse.hono</groupId>
         <artifactId>hono-dispatch-router</artifactId>
         <version>${project.version}</version>
+      </dependency>
+      <dependency>
+          <groupId>javax.annotation</groupId>
+          <artifactId>javax.annotation-api</artifactId>
+          <version>${javax.annotation.api.version}</version>
       </dependency>
       <dependency>
         <groupId>com.google.guava</groupId>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -60,6 +60,11 @@
       <artifactId>jjwt</artifactId>
     </dependency>
     <dependency>
+      <!-- required as a (missing) transient dependency for JJWT -->
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
       <version>${jackson.version}</version>
--- a/service-base/pom.xml
+++ b/service-base/pom.xml
@@ -132,6 +132,10 @@
       <groupId>io.opentracing.contrib</groupId>
       <artifactId>opentracing-tracerresolver</artifactId>
     </dependency>
+    <dependency>
+      <groupId>javax.annotation</groupId>
+      <artifactId>javax.annotation-api</artifactId>
+    </dependency>
   </dependencies>
 
   <build>
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,15 @@
 dist: trusty
 sudo: required
 language: java
+
 jdk:
   - openjdk8
+  - oraclejdk10
+
 script:
   - mvn install verify -B -Dtest.env=true -DcreateJavadoc=true -Ddocker.host=unix:///var/run/docker.sock -Pbuild-docker-image,run-tests -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
 before_install:
+
 env:
   global:
   - COMMIT=${TRAVIS_COMMIT::8}
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -58,6 +58,10 @@
             <artifactId>hono-demo-certs</artifactId>
         </dependency>
         <dependency>
+            <groupId>javax.annotation</groupId>
+            <artifactId>javax.annotation-api</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-test</artifactId>
             <version>${spring-boot.version}</version>


ctron:
So the basic changes I had to made are:

* Enable matrix build support with Java 10 on Travis
* Upgrade a bunch of Maven plugins
* Add explicit dependencies to "javax." dependencies. This was already a thing for OSGi, but it seems this has also hit Java 9+ now. The JAXB one was actually a transient dependency from JJWT.
* Upgrade mockito, seemed like an old version anyway.

Currently both Java 8 and Java 10 builds run in parallel. Which could help us find issues which are caused by Java 10.

There still seems to be one last issue in the unit tests. I see a few IT tests failing with "timeout". However I have seen this a few times in the last days already, so I am not sure if that has anything to do with Java 10. But I haven't analyzed this much.

sophokles73:
> There still seems to be one last issue in the unit tests. I see a few IT tests failing with "timeout". However I have seen this a few times in the last days already, so I am not sure if that has anything to do with Java 10. But I haven't analyzed this much.

No,this is due to the fact that the VM on Travis has close to zero processor time ;-) The tests for uploading telemetry to MQTT adapter using QoS 0 sometimes (often) fail. I have added a flag which basically disables this IT on Travis.

bom/pom.xml
```diff
@@ -38,11 +38,13 @@
     <influxdb.version>1.2.4-alpine</influxdb.version>
     <jackson.version>2.9.0</jackson.version>
     <java-base-image.name>openjdk:8u162-jre-slim</java-base-image.name>
-    <jjwt.version>0.7.0</jjwt.version>
+    <jaxb.api.version>2.3.0</jaxb.api.version>
+    <javax.annotation.api.version>1.3.2</javax.annotation.api.version>
+    <jjwt.version>0.9.0</jjwt.version>
```
sophokles73: we cannot simply update this to 0.9.0 without a CQ

bom/pom.xml
```diff
@@ -266,6 +268,12 @@
         <artifactId>jjwt</artifactId>
         <version>${jjwt.version}</version>
       </dependency>
+      <dependency>
+        <!-- required as a (missing) transient dependency for JJWT -->
+        <groupId>javax.xml.bind</groupId>
+        <artifactId>jaxb-api</artifactId>
```
sophokles73: is this no longer part of Java?

pom.xml
```diff
@@ -277,20 +277,20 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-javadoc-plugin</artifactId>
-          <version>2.10.4</version>
+          <version>3.0.1</version>
```
sophokles73: this version has an issue with doing a multiproject build, that's why we are still using the old version

bom/pom.xml
ctron: I know.

bom/pom.xml
ctron: Seems so. I need to find an explicit source for that. But that is my understanding from what I saw googling the issue. Seems like jaxb or javax stuff is no longer part of core Java.

pom.xml
ctron: Can you provide a way to reproduce this issue.

ctron:
Just to be clear. I am not suggesting to merge this ASAP. I simply wanted to check which blockers we hit. But I looks like not too much trouble.

Let's fix the commented issues and then discuss what we actually want to do.

pom.xml
sophokles73: IIRC a simple

mvn javadoc:aggregate

should already fail

ctron: Yes that fails with:

~~~
ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.0.2:unpack-dependencies (copy_demo_certs) on project hono-core: Artifact has not been packaged yet. When used on reactor artifact, unpack should be executed after packaging: see MDEP-98. -> [Help 1]
~~~

But I think this is a bug in the Hono build which simply didn't manifest until now in that plugin (but in the Eclipse IDE) and which can be prevented by:

~~~
mvn package javadoc:aggregate
~~~

 I think the proper way to fix this would be make of use of the [remote resources plugin](https://maven.apache.org/plugins/maven-remote-resources-plugin/).

So would this workaround above the fine for the moment? Or should I try to replace the "dependency" plugin with the "remove resources" plugin? As I am pretty sure the Maven Javadoc Plugin will not change their behavior.

sophokles73:
@ctron I have upgraded Mockito to 2.19.0 in fa68ca2d005035eb3145d84fcba444be423830dc

pom.xml
sophokles73: No, I think we can live that shortcoming. We are using the aggregate goal in our release builds but we always use it in conjunction with a another goal like *install*, so it shouldn't be a problem.

sophokles73:
@ctron I have upgraded Surefire/Failsafe plugins to 2.22.0 in 660e7574dc9c8e13acc3f8cbb6863c2d05281014

bom/pom.xml
```diff
@@ -38,12 +38,14 @@
     <influxdb.version>1.2.4-alpine</influxdb.version>
     <jackson.version>2.9.0</jackson.version>
     <java-base-image.name>openjdk:8u171-jre-slim</java-base-image.name>
-    <jjwt.version>0.7.0</jjwt.version>
+    <jaxb.api.version>2.3.0</jaxb.api.version>
+    <javax.annotation.api.version>1.3.2</javax.annotation.api.version>
+    <jjwt.version>0.9.0</jjwt.version>
```
sophokles73: doesn't it work with 0.7.0?

sophokles73:
@ctron can you rebase?

ctron:
Here is a link to a successful build on travis: https://travis-ci.org/ctron/hono/builds/400028757

sophokles73:
> All unit tests work on my local machine. However it looks like as currently the tests on travis are rather unstable

Which ones do you mean? Can you give an example?

bom/pom.xml
sophokles73: @ctron have you tried using 0.7.0?

ctron:
e.g. https://travis-ci.org/eclipse/hono/jobs/399992628 … the job is simply stuck and gets terminated after the 50 minute limit. I've seen this the same way for Java 8. Restarting the job often "fixes" the issue.

I guess something doesn't properly shut down and gets stuck in the shutdown sequence.

sophokles73:
however, this is a problem that we have in the integration tests. This is a threading issue in the MQTT protocol adapter which I am currently working on ...

bom/pom.xml
ctron: I re-test with 0.7.0 and it does work. So I did change the version back to 0.7.0. However it still requires the extra dependency on jaxb.

ctron:
So I think what is left to do is to:

* File CQs for JAXB API and javax.annotations
* Merge this branch

bom/pom.xml
```diff
@@ -38,6 +38,8 @@
     <influxdb.version>1.2.4-alpine</influxdb.version>
     <jackson.version>2.9.5</jackson.version>
     <java-base-image.name>openjdk:8u171-jre-slim</java-base-image.name>
+    <jaxb.api.version>2.3.0</jaxb.api.version>
+    <javax.annotation.api.version>1.3.2</javax.annotation.api.version>
```
sophokles73: this dependency is not needed when I compile and run integration tests using Java 10

service-base/pom.xml
```diff
@@ -120,6 +120,10 @@
       <groupId>io.opentracing.contrib</groupId>
       <artifactId>opentracing-tracerresolver</artifactId>
     </dependency>
+    <dependency>
+      <groupId>javax.annotation</groupId>
+      <artifactId>javax.annotation-api</artifactId>
+    </dependency>
```
sophokles73: this dependency is not needed when I compile and run integration tests using Java 10

bom/pom.xml
```diff
@@ -273,6 +281,11 @@
         <artifactId>hono-dispatch-router</artifactId>
         <version>${project.version}</version>
       </dependency>
+      <dependency>
+          <groupId>javax.annotation</groupId>
+          <artifactId>javax.annotation-api</artifactId>
+          <version>${javax.annotation.api.version}</version>
+      </dependency>
```
sophokles73: this dependency is not needed when I compile and run integration tests using Java 10

bom/pom.xml
```diff
@@ -38,6 +38,8 @@
     <influxdb.version>1.2.4-alpine</influxdb.version>
     <jackson.version>2.9.5</jackson.version>
     <java-base-image.name>openjdk:8u171-jre-slim</java-base-image.name>
+    <jaxb.api.version>2.3.0</jaxb.api.version>
```
sophokles73: version 2.2.12 also does the trick and we could simply piggy back https://dev.eclipse.org/ipzilla/show_bug.cgi?id=13115

core/pom.xml
```diff
@@ -61,6 +61,11 @@
       <groupId>io.jsonwebtoken</groupId>
       <artifactId>jjwt</artifactId>
     </dependency>
+    <dependency>
+      <!-- required as a (missing) transient dependency for JJWT -->
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
+    </dependency>
```
sophokles73: I would rather put this into a profile which is activated by the java version:
```
  <profiles>
    <profile>
      <id>java-10-deps</id>
      <activation>
        <jdk>10</jdk>
      </activation>
      <dependencies>
        <dependency>
          <groupId>javax.xml.bind</groupId>
          <artifactId>jaxb-api</artifactId>
        </dependency>
      </dependencies>
    </profile>
  </profiles>
```
This way the jar gets only included in the Spring Boot jar when we build on Java 10. WDYT?

tests/pom.xml
```diff
@@ -61,6 +61,7 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
     <dependency>
       <groupId>io.vertx</groupId>
       <artifactId>vertx-unit</artifactId>
+      <scope>test</scope>
```
sophokles73: this is already declared with scope *test* in the bom/pom.xml ...

ctron:
Interesting .... I do know that I had class loader issues back when I created the PR, but it seems they are gone now. So I will change the PR, dropping the extras and see what Travis has to say about that.

bom/pom.xml
ctron: Well I wanted to be a good citizen and create a CQ for the most recent version for other to PB on. :wink:

core/pom.xml
ctron: I don't think it hurts the Java 8 deployment. And it would spare as the extra profile. So I would vote for keeping it simple and not adding any conditionals unless absolutely necessary.

ctron:
So it looks like this is the javadoc plugin:

~~~
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.0.1:jar (attach-javadocs) on project hono-service-base: MavenReportException: Error while generating Javadoc: 
[ERROR] Exit code: 1 - /home/travis/build/ctron/hono/service-base/src/main/java/org/eclipse/hono/service/AbstractApplication.java:27: error: package javax.annotation is not visible
[ERROR] import javax.annotation.PreDestroy;
[ERROR]             ^
[ERROR]   (package javax.annotation is declared in module java.xml.ws.annotation, which is not in the module graph)
[ERROR] 
[ERROR] Command line was: /home/travis/oraclejdk10/bin/javadoc @options @packages
[ERROR] 
[ERROR] Refer to the generated Javadoc files in '/home/travis/build/ctron/hono/service-base/target/apidocs' dir.
[ERROR] -> [Help 1]
~~~

Can you try again with `-DcreateJavadoc=true`.

sophokles73:
@ctron 

Ah, I see. I didn't try generating JavaDoc. We are indeed using the annotation in `AbstractApplication`. However, I wonder why it compiles without the annotation class on the classpath? Anyway, let's add the dependency again.

ctron:
We got checkin approval for both dependencies. So I am going to merge this in the next few days.

sophokles73:
@ctron Sharon indicated that we might not get all CQs approved on time for 0.7. So I would rather wait with merging until we know for sure, ok?

ctron:
> @ctron Sharon indicated that we might not get all CQs approved on time for 0.7. So I would rather wait with merging until we know for sure, ok?

Ok, in that case, let's wait.

ctron:
I just double checked .. and it looks like both CQs required for this PR are approved:

https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16910
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16911

Or did I miss anything?

sophokles73:
@ctron That's good news :-) Then Sharon actually was able to do the check in the meantime ...

If those are the only two CQs this PR depends on, then I would say we're good to go :+1:

ctron:
> Maybe it would still be worthwhile to rebase and see if the Travis build works ...

With this one, for sure!

ctron:
Rebased

ctron:
Travis gave a green light ... so I going to merge.

Issue #319, opened 2017-08-28
Add device specific message flow debugging


pellmann:
**Problem**

It is often hard to follow specific problems from end-to-end. This is true in particular with many different devices talking to Hono with different adapters on one side and also many different consumers and tenants on the other side.

It is hard for the solution developer to get an idea of the real message flow and what happens. It is also hard to identify a problem if a customers calls the DevOps with "My message never arrives..".

**Solution**

Allow an option in  the Device Registry to set a device (or a class of devices or a tenant) in a debug state. This could trigger different actions (dependent an the specific state):

- Extend the normal tracing/logging of the components, that are involved. Typically at least the Adapter and Hono Messaging.
- Add debugging information of each involved component to the message. This could be done with specified message properties.

WDYT?

pellmann:
Dispatch Router uses x-opt-qd.trace (https://qpid.apache.org/releases/qpid-dispatch-0.8.0/book/book.html#amqp-mapping) as list of strings with all routers. So maybe something like x-opt-hono.trace might be nice.

sophokles73:
I would rather employ existing technology like http://opentracing.io for this purpose. Maybe in conjunction with https://jaegertracing.io as the implementation.

pellmann:
For overall traceability it would be great to have such a tracing technology as opentracing.

But I still see a value for tracing information in the message itself (e.g. in annotations).

When there will be a system with all components involved, and all of them use such a system and also each customer/tenant is able to see (only) his message traces it might substitute the in-message information. 

But I would assume it might be easier in a lot of installations to also have such a in-message tracing.

sophokles73:
> But I still see a value for tracing information in the message itself (e.g. in annotations).

What kind of information would that be for example?

pellmann:
- Component (adapter, messaging, router) and instance - to see the path, the message took
- Timestamp - to see the time consumed in each component/network part
- Configuration info - ssl in network and from device to adapter, ..
- System snapshot - open connections/cpu/ram

I think this all might be interesting in all variants of tracing. To put it in the message (if configured) is just one simple variant of bringing this information together and to scope it, because the message already has this 'trace'.

sophokles73:
Well, I believe we can do all that with opentracing ...

sophokles73:
fixed in https://github.com/eclipse/hono/commit/985fba492d37ca79aadcdfbaddb408f152ae68b5

PR #652, opened 2018-06-11
[#319] Add high level tracing to HTTP and MQTT adapters.

sophokles73:
All standard adapters are instrumented using OpenTracing in order to be
able to track the processing of messages that have been uploaded by
devices.

In this first step, only the handlers for incoming HTTP requests and
published MQTT messages are instrumented, i.e. we only trace the
incoming request and the outcome of processing the request at a very
high level.

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -13,15 +13,16 @@
 package org.eclipse.hono.adapter.http;
 
 import java.net.HttpURLConnection;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 
-import io.vertx.core.Handler;
-import io.vertx.core.http.HttpHeaders;
-import io.vertx.proton.ProtonDelivery;
 import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.client.ClientErrorException;
 import org.eclipse.hono.client.MessageConsumer;
@@ -30,6 +31,7 @@ import org.eclipse.hono.service.AbstractProtocolAdapterBase;
 import org.eclipse.hono.service.auth.device.Device;
 import org.eclipse.hono.service.command.CommandResponseSender;
 import org.eclipse.hono.service.http.HttpUtils;
+import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.CommandConstants;
 import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.EventConstants;
@@ -41,15 +43,22 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 
+import io.opentracing.Span;
+import io.opentracing.contrib.vertx.ext.web.TracingHandler;
+import io.opentracing.contrib.vertx.ext.web.WebSpanDecorator;
+import io.opentracing.tag.Tags;
 import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
+import io.vertx.core.Handler;
 import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.HttpHeaders;
 import io.vertx.core.http.HttpServer;
 import io.vertx.core.http.HttpServerOptions;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.Router;
 import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.handler.BodyHandler;
+import io.vertx.proton.ProtonDelivery;
 
 /**
  * Base class for a Vert.x based Hono protocol adapter that uses the HTTP protocol.
@@ -167,6 +176,7 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
                     return Future.failedFuture("no router configured");
                 } else {
                     addRoutes(router);
+                    addTracingHandler(router);
                     return CompositeFuture.all(bindSecureHttpServer(router), bindInsecureHttpServer(router));
                 }
             }).compose(s -> {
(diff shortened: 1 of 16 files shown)


adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -507,6 +562,13 @@ private void doUploadMessage(final RoutingContext ctx, final String tenant, fina
             } else {
 
                 final Device authenticatedDevice = getAuthenticatedDevice(ctx);
+                Optional.ofNullable((Span) ctx.get(TracingHandler.CURRENT_SPAN)).map(span -> {
+                    span.setOperationName("upload " + endpointName);
+                    TracingHelper.TAG_TLS.set(span, ctx.request().isSSL());
+                    TracingHelper.TAG_AUTHENTICATED.set(span, authenticatedDevice != null);
+                    return span.context();
+                }).orElse(null);
```
calohmn: I would have expected an assignment to a `SpanContext` variable here - otherwise the `.orElse(null)` at the end is superfluous. And in that case a simple `if (span != null)` block would look more readable imho (also as `return span.context()` is ignored).

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -556,9 +587,15 @@ private void onAuthenticationSuccess(final MqttEndpoint endpoint, final Device a
...
             return Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST,
-                    String.format("Content-Type %s does not match with the payload", ctx.contentType())));
+                    String.format("Content-Type %s does not match with payload", ctx.contentType())));
         } else {
 
+            Optional.ofNullable(getCurrentSpan(ctx)).map(span -> {
+                span.setOperationName("upload " + endpointName);
+                TracingHelper.TAG_AUTHENTICATED.set(span, ctx.authenticatedDevice() != null);
+                return span.context();
+            }).orElse(null);
```
calohmn: I would have expected an assignment to a `SpanContext` variable here - otherwise the `.orElse(null)` at the end is superfluous. And in that case a simple `if (span != null)` block would look more readable imho (also as `return span.context()` is ignored).

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
sophokles73: will be assigned to a local variable in one of the next steps. Then it will all make sense ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -451,6 +461,27 @@ private void onAuthenticationSuccess(final MqttEndpoint endpoint, final Device a
...
+            .ignoreActiveSpan()
+            .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_SERVER)
+            .withTag(Tags.MESSAGE_BUS_DESTINATION.getKey(), context.message().topicName())
+            .withTag(TracingHelper.TAG_QOS.getKey(), context.message().qosLevel().toString())
+            .withTag(Tags.COMPONENT.getKey(), getTypeName())
+            .start();
+        context.put(KEY_CURRENT_SPAN, span);
+        onPublishedMessage(context).setHandler(processing -> {
+            if (processing.succeeded()) {
+                span.setTag(Tags.HTTP_STATUS.getKey(), HttpURLConnection.HTTP_ACCEPTED);
```
ppatierno: Just a minor thing, isn't it weird to read "http.status_code" (the `HTTP_STATUS` value) in a trace related to MQTT request? Can we add a custom MQTT tag for that?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -358,7 +368,7 @@ private void handleConnectionRequestResult(final MqttEndpoint endpoint,
             LOG.debug("connection to unauthenticated device [clientId: {}] closed", endpoint.clientIdentifier());
             metrics.decrementUnauthenticatedMqttConnections();
         });
-        endpoint.publishHandler(message -> onPublishedMessage(new MqttContext(message, endpoint)));
+        endpoint.publishHandler(message -> handlePublishedMessage(new MqttContext(message, endpoint)));
```
ppatierno: the "handlePublishedMessage" should be used in the Javadoc as well.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
sophokles73: In general, I agree. However, the OpenTracing documentation advises to use the standard tags wherever possible. Since we are indeed using HTTP status code (semantics) all over Hono, I thought it would make sense to use this standard tag as well so it is clear that the code indeed has HTTP status code semantics. WDYT?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
sophokles73: well, *handlePublishedMessage* is a private method which doesn't show up in the JavaDocs anyway and its only responsibility is to start and end the corresponding span. It does indeed invoke *onPublishedMessage*, though, so IMHO the JavaDoc comment is still correct and reflects the information that is relevant to a user/adopter, i.e. it points out the method he can plug into. WDYT?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
ppatierno: I'm fine with that, just hoping that the tracing reader won't be get confused about it. Let's stick with it.

Issue #643, opened 2018-06-06
AMQP Protocol Adapter


Alfusainey:
This is a container issue for the AMQP protocol adapter. All PRs for the adapter will reference this issue

Alfusainey:
@sophokles73: i just want to follow-up on your previous comment regarding integration tests. 
what do you think about adding the integration test for the amqp adapter (unauthenticated devices) before moving to the next step? like this i can have an initial test setup in the `tests` module before moving further. WDYT?

sophokles73:
Absolutely. Having an initial integration test that you can extend with additional test cases makes total sense to me.

Alfusainey:
> sounds reasonable, however, it would also be nice to check if it also works with e.g. the Qpid JMS provider 

@sophokles73: Am now working on this task

sophokles73:
@Alfusainey I think this can wait. Based on the fact that GSoC ends mid August, we should probably talk about prioritization. What about using our Thursday call to discuss?

Alfusainey:
> What about using our Thursday call to discuss?

Yes sure, we can meet this thursday

pho-enix:
Great to see the progress and that there is authentication for AMQP PA now :)

When will it be ready to use? With 0.7?

sophokles73:
We plan to have an initial version in 0.7. However, I think that it will most likely not yet support C&C..

sophokles73:
fixed in https://github.com/eclipse/hono/commit/f502009e47b7f84db852a319605217591485dd82

PR #663, opened 2018-06-20
[#643] AMQP Adapter integration tests

Alfusainey:
@sophokles73: first try on the AMQP adapter integration test. for some reason some of the tests are failing, so i temporarily marked them as `ignore` (while i investigate and allow you to look at the overall setup of the tests).

### DIFF
--- a/tests/pom.xml
+++ b/tests/pom.xml
@@ -232,7 +232,7 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch 
                     <env>
                       <LOGGING_CONFIG>file:///etc/hono/logback-spring.xml</LOGGING_CONFIG>
                       <SPRING_CONFIG_LOCATION>file:///etc/hono/</SPRING_CONFIG_LOCATION>
-                      <SPRING_PROFILES_ACTIVE>prod</SPRING_PROFILES_ACTIVE>
+                      <SPRING_PROFILES_ACTIVE>dev</SPRING_PROFILES_ACTIVE>
                     </env>
                     <log>
                       <prefix>REGISTRY</prefix>
@@ -568,6 +568,62 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch 
                     </wait>
                   </run>
                 </image>
+                <!-- ##### AMQP adapter ##### -->
+                <image>
+                  <name>${docker.image.org-name}/hono-adapter-amqp-test</name>
+                  <alias>amqp-adapter</alias>
+                  <build>
+                    <from>${docker.image.org-name}/hono-adapter-amqp-vertx:${project.version}</from>
+                    <assembly>
+                      <mode>dir</mode>
+                      <basedir>/etc/hono</basedir>
+                      <inline>
+                        <id>config</id>
+                        <fileSets>
+                          <fileSet>
+                            <directory>${project.basedir}/src/test/resources/amqp</directory>
+                            <outputDirectory>.</outputDirectory>
+                            <includes>
+                              <include>*</include>
+                            </includes>
+                          </fileSet>
+                          <fileSet>
+                            <directory>${project.build.directory}/certs</directory>
+                            <outputDirectory>certs</outputDirectory>
+                          </fileSet>
+                        </fileSets>
+                      </inline>
+                    </assembly>
+                  </build>
+                  <run>
+                    <ports>
+                      <port>+adapter.amqp.ip:adapter.amqp.port:4040</port>
+                      <port>+adapter.amqp.ip:adapter.amqp.health.port:8088</port>
+                    </ports>
+                    <network>
+                      <mode>custom</mode>
+                      <name>hono</name>
+                      <alias>hono-adapter-amqp-vertx.hono</alias>
+                    </network>
+                    <env>
+                      <LOGGING_CONFIG>file:///etc/hono/logback-spring.xml</LOGGING_CONFIG>
+                      <SPRING_CONFIG_LOCATION>file:///etc/hono/</SPRING_CONFIG_LOCATION>
+                      <SPRING_PROFILES_ACTIVE>prod</SPRING_PROFILES_ACTIVE>
+                    </env>
+                    <log>
+                      <prefix>AMQP</prefix>
+                      <color>RED</color>
+                    </log>
+                    <wait>
+                      <time>${service.startup.timeout}</time>
+                      <http>
+                        <method>GET</method>
+                        <url>http://${adapter.amqp.ip}:${adapter.amqp.health.port}/readiness</url>
+                        <status>200..299</status>
+                      </http>
+                    </wait>
+                  </run>
+                </image>
               </images>
             </configuration>
             <executions>
@@ -627,6 +683,8 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch 
                 <https.port>${https.port}</https.port>
                 <mqtt.host>${mqtt.ip}</mqtt.host>
                 <mqtt.port>${mqtt.port}</mqtt.port>
+                <adapter.amqp.host>${adapter.amqp.ip}</adapter.amqp.host>
+                <adapter.amqp.port>${adapter.amqp.port}</adapter.amqp.port>
                 <trust-store.path>${project.build.directory}/certs/trusted-certs.pem</trust-store.path>
                 <test.env>${test.env}</test.env>
                 <!-- javax.net.debug>ssl:handshake</javax.net.debug-->
--- a/tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
@@ -15,8 +15,6 @@ package org.eclipse.hono.tests;
 import java.net.InetAddress;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
@@ -48,6 +46,7 @@ public final class IntegrationTestSupport {
     public static final int    DEFAULT_HTTP_PORT = 8080;
     public static final int    DEFAULT_HTTPS_PORT = 8443;
     public static final int    DEFAULT_MQTT_PORT = 1883;
+    public static final int    DEFAULT_AMQP_PORT = 4040;
 
     public static final String PROPERTY_AUTH_HOST = "auth.host";
     public static final String PROPERTY_AUTH_PORT = "auth.amqp.port";
@@ -67,6 +66,8 @@ public final class IntegrationTestSupport {
     public static final String PROPERTY_HTTPS_PORT = "https.port";
     public static final String PROPERTY_MQTT_HOST = "mqtt.host";
     public static final String PROPERTY_MQTT_PORT = "mqtt.port";
+    public static final String PROPERTY_AMQP_HOST = "adapter.amqp.host";
+    public static final String PROPERTY_AMQP_PORT = "adapter.amqp.port";
     public static final String PROPERTY_TENANT = "tenant";
 
     public static final String AUTH_HOST = System.getProperty(PROPERTY_AUTH_HOST, DEFAULT_HOST);
@@ -93,6 +94,8 @@ public final class IntegrationTestSupport {
     public static final int    HTTPS_PORT = Integer.getInteger(PROPERTY_HTTPS_PORT, DEFAULT_HTTPS_PORT);
     public static final String MQTT_HOST = System.getProperty(PROPERTY_MQTT_HOST, DEFAULT_HOST);
     public static final int    MQTT_PORT = Integer.getInteger(PROPERTY_MQTT_PORT, DEFAULT_MQTT_PORT);
+    public static final String AMQP_HOST = System.getProperty(PROPERTY_AMQP_HOST, DEFAULT_HOST);
+    public static final int    AMQP_PORT = Integer.getInteger(PROPERTY_AMQP_PORT, DEFAULT_AMQP_PORT);
 
     public static final String PATH_SEPARATOR = System.getProperty("hono.pathSeparator", "/");
     public static final int    MSG_COUNT = Integer.getInteger("msg.count", 1000);
@@ -117,7 +120,7 @@ public final class IntegrationTestSupport {
      */
     public HonoClient downstreamClient;
 
-    private final List<String> tenantsToDelete = new LinkedList<>();
+    private final Set<String> tenantsToDelete = new HashSet<>();
     private final Map<String, Set<String>> devicesToDelete = new HashMap<>();
     private final Vertx vertx;
 
--- a/tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
@@ -1,0 +1,332 @@
+/**
+ * Copyright (c) 2018 Bosch Software Innovations GmbH and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ */
+package org.eclipse.hono.tests.amqp;
+
+import java.net.HttpURLConnection;
+import java.util.concurrent.CountDownLatch;
+
+import org.apache.qpid.proton.message.Message;
+import org.eclipse.hono.client.ClientErrorException;
+import org.eclipse.hono.client.HonoClient;
+import org.eclipse.hono.client.MessageConsumer;
+import org.eclipse.hono.client.MessageSender;
+import org.eclipse.hono.client.impl.HonoClientImpl;
+import org.eclipse.hono.config.ClientConfigProperties;
+import org.eclipse.hono.tests.IntegrationTestSupport;
+import org.eclipse.hono.tests.client.ClientTestBase;
+import org.eclipse.hono.util.Constants;
+import org.eclipse.hono.util.TenantObject;
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestName;
+
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.Vertx;
+import io.vertx.ext.unit.Async;
+import io.vertx.ext.unit.TestContext;
+import io.vertx.proton.ProtonClientOptions;
+import io.vertx.proton.ProtonHelper;
+
+/**
+ * Base class for the AMQP Adapter integration tests.
+ */
+public abstract class AmqpAdapterBase extends ClientTestBase {
+
+    private static final String DEVICE_PASSWORD = "device-password";
+
+    @Rule
+    public final TestName testName = new TestName();
+
+    /**
+     * The Vert.x instance to run on.
+     */
+    private static Vertx vertx;
+
+    /**
+     * A helper for accessing the AMQP 1.0 messaging network and for managing devices tenants, devices and credentials.
+     */
+    protected static IntegrationTestSupport helper;
+
+    /**
+     * A client for connecting to the AMQP protocol adapter.
+     */
+    protected HonoClient adapterClient;
+
+    private MessageSender sender;
+    private MessageConsumer consumer;
+
+    /**
+     * Create a HTTP client for accessing the device registry (for registering devices and credentials) and
+     * an AMQP 1.0 client for consuming messages from the messaging network.
+     * 
+     * @param context The Vert.x test context.
+     */
+    @BeforeClass
+    public static void init(final TestContext context) {
+        vertx = Vertx.vertx();
+        helper = new IntegrationTestSupport(vertx);
+        helper.init(context);
+    }
+
+    /**
+     * Shut down the client connected to the messaging network.
+     * 
+     * @param context The Vert.x test context.
+     */
+    @AfterClass
+    public static void shutdown(final TestContext context) {
+        helper.disconnect(context);
+    }
+
+    /**
+     * Logs a message before running a test case.
+     */
+    @Before
+    public void setUp() {
+        LOGGER.info("running {}", testName.getMethodName());
+    }
+
+    /**
+     * Disconnect the AMQP 1.0 client connected to the AMQP Adapter and close senders and consumers.
+     * Also delete all random tenants and devices generated during the execution of a test case.
+     * 
+     * @param context The Vert.x test context.
+     */
+    @After
+    public void postTest(final TestContext context) {
+        disconnect(context);
+        helper.deleteObjects(context);
+    }
+
+    /**
+     * Verifies that a BAD inbound message (i.e when the message content-type does not match the payload) is rejected by
+     * the adapter. This test case uses the credentials on record for device 4711 of the default tenant.
+     * 
+     * @param context The Vert.x context for running asynchronous tests.
+     */
+    @Test(timeout = 4000)
+    public void testAdapterRejectsBadInboundMessage(final TestContext context) {
+
+        final String tenantId = helper.getRandomTenantId();
+        final String deviceId = helper.getRandomDeviceId(tenantId);
+
+        final Async setup = context.async();
+        setupProtocolAdapter(tenantId, deviceId, false, context)
+        .compose(s -> {
+           sender = s;
+           return Future.succeededFuture();
+        }).setHandler(context.asyncAssertSuccess(ok -> setup.complete()));
+        setup.await();
+
+        final Message msg = ProtonHelper.message("msg payload");
+
+        msg.setContentType("application/vnd.eclipse-hono-empty-notification");
+
+        sender.send(msg).setHandler(context.asyncAssertFailure(t -> {
+
+            context.assertTrue(ClientErrorException.class.isInstance(t));
+
+            context.assertEquals(HttpURLConnection.HTTP_BAD_REQUEST, ((ClientErrorException) t).getErrorCode());
+        }));
+    }
+
+    /**
+     * Verifies that the AMQP adapter rejects messages from a device that belongs to a tenant for which it has been
+     * disabled. This test case only considers events since telemetry senders do not wait for response from the adapter.
+     * 
+     * @param context The Vert.x test context.
+     */
+    @Test(timeout = 4000)
+    public void testUploadMessageFailsForTenantDisabledAdapter(final TestContext context) {
+
+        final String tenantId = helper.getRandomTenantId();
+        final String deviceId = helper.getRandomDeviceId(tenantId);
+
+        final Async setup = context.async();
+        setupProtocolAdapter(tenantId, deviceId, true, context)
+        .compose(s -> {
+           sender = s;
+           return Future.succeededFuture();
+        }).setHandler(context.asyncAssertSuccess(ok -> setup.complete()));
+        setup.await();
+
+        final Message msg = ProtonHelper.message("some payload");
+        sender.send(msg).setHandler(context.asyncAssertFailure(t -> {
+
+            context.assertTrue(ClientErrorException.class.isInstance(t));
+
+            // HonoClient maps AmqpError.UNAUTHORIZED_ACCESS to HttpURLConnection.HTTP_BAD_REQUEST
+            // -> so we test the message
+
+            final String expectedErrorMsg = String.format("This adapter is not enabled for tenant [tenantId: %s].",
+                    tenantId);
+
+            context.assertEquals(expectedErrorMsg, ((ClientErrorException) t).getMessage());
+        }));
+    }
+
+    /**
+     * Verifies that the AMQP Adapter will fail to authenticate a device whose username does not match the expected pattern
+     * {@code [<authId>@<tenantId>]}.
+     * 
+     * @param context The Vert.x test context.
+     */
+    @Test(timeout = 1000)
+    public void testAuthenticationWithInvalidUsernamePattern(final TestContext context) {
+        connectToAdapter("invalidaUsername", DEVICE_PASSWORD, true)
+        .setHandler(context.asyncAssertFailure(t -> {
+            // SASL handshake failed
+            context.assertTrue(ClientErrorException.class.isInstance(t));
+            context.assertEquals(HttpURLConnection.HTTP_UNAUTHORIZED, ((ClientErrorException) t).getErrorCode());
+        }));
+    }
+
+    /**
+     * Verifies that a number of messages published through the AMQP adapter can be successfully consumed by
+     * applications connected to the AMQP messaging network.
+     *
+     * @param context The Vert.x test context.
+     * @throws InterruptedException Exception.
+     */
+    @Test
+    public void testUploadingXnumberOfMessages(final TestContext context) throws InterruptedException {
+
+        final String tenantId = helper.getRandomTenantId();
+        final String deviceId = helper.getRandomDeviceId(tenantId);
+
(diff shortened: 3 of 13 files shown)


tests/src/test/java/org/eclipse/hono/tests/amqp/AdapterAmqpIT.java
```diff
@@ -0,0 +1,407 @@
...
+            received.countDown();
+
+        }).setHandler(context.asyncAssertSuccess(ok -> consumerTracker.complete()));
+        consumerTracker.await();
+
+        final AtomicInteger messageCount = new AtomicInteger(0);
+        client = ProtonClient.create(vertx);
+        while (messageCount.get() < MESSAGES_TO_SEND) {
+            final Async sending = context.async();
+            client.connect(IntegrationTestSupport.AMQP_HOST, IntegrationTestSupport.AMQP_PORT, ar -> {
```
sophokles73: why do you want to connect for each message individually?

tests/src/test/resources/amqp/application.yml
```diff
@@ -0,0 +1,49 @@
...
+    port: 5672 # AMQP port of the device registry
+    username: amqp-adapter@HONO
+    password: amqp-secret
+  tenant:
+    name: 'Hono AMQP Adapter'
+    host: hono-service-device-registry.hono
+    port: 5672 # AMQP port of the device registry
+    username: amqp-adapter@HONO
+    password: amqp-secret
+  metric:
```
sophokles73: you should also add the command properties

tests/src/test/java/org/eclipse/hono/tests/amqp/AdapterAmqpIT.java
Alfusainey: otherwise the consumer receives a single message.. trying to understand why that is so

tests/src/test/resources/amqp/application.yml
Alfusainey: ok

Alfusainey:
@sophokles73: should i separate the AMQP bug fix from the IT? i want to have bug fix on master so that i proceeed to the next step. WDYT?

sophokles73:
> should i separate the AMQP bug fix from the IT? i want to have bug fix on master so that i proceed to the next step. WDYT?

Yes, that feels better ...

tests/src/test/java/org/eclipse/hono/tests/amqp/AdapterAmqpIT.java
```diff
@@ -64,17 +48,14 @@
...
-
     /**
      * Sets up and initializes the integration test setup.
      * 
      * @param context The Vert.x test context.
      */
     @BeforeClass
     public static void init(final TestContext context) {
-        vertx = Vertx.vertx();
+        vertx = Vertx.vertx(new VertxOptions().setBlockedThreadCheckInterval(1000 * 60 * 60));
```
sophokles73: what is the purpose of this?

tests/src/test/resources/amqp/logback-spring.xml
```diff
@@ -0,0 +1,39 @@
...
+  <springProfile name="prod">
+    <logger name="org.eclipse.hono" level="INFO"/>
+  </springProfile>
+
+  <logger name="io.netty.handler.logging.LoggingHandler" level="INFO"/>
+
+  <logger name="io.vertx.proton.impl" level="INFO"/>
+  <logger name="io.vertx.core.net.impl" level="INFO"/>
+
+</configuration>
```
sophokles73: missing EOL

tests/src/test/java/org/eclipse/hono/tests/amqp/AdapterAmqpIT.java
Alfusainey: sorry, i forgot to remove. i only use it when debugging

tests/src/test/resources/amqp/logback-spring.xml
Alfusainey: ok

sophokles73:
@Alfusainey any news on this one?

Alfusainey:
will work on this now.. I realize that i can use the AMQP 1.0 client (i.e `HonoClient`) to simulate an amqp device sending messages to the adapter. i think the api of honoclient will make the code more readable than directly using `ProtonClient`.

sophokles73:
sounds reasonable

Alfusainey:
@sophokles73 : PR updated.

as i was working on the integration tests, there was need to adapt the amqp adapter code for it to properly work for authenticated devices. this is the reason why this updated PR contains changes to the adapter codebase. i do not want to create a separate PR for that (since i will have to wait for that to be merged first). i hope that's okay?

tests/pom.xml
```diff
@@ -232,7 +232,7 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
                     <env>
                       <LOGGING_CONFIG>file:///etc/hono/logback-spring.xml</LOGGING_CONFIG>
                       <SPRING_CONFIG_LOCATION>file:///etc/hono/</SPRING_CONFIG_LOCATION>
-                      <SPRING_PROFILES_ACTIVE>prod</SPRING_PROFILES_ACTIVE>
+                      <SPRING_PROFILES_ACTIVE>dev</SPRING_PROFILES_ACTIVE>
```
sophokles73: you may, of course, set this to dev on your local machine, but in the repo, we want to keep this set to prod.

tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
```diff
@@ -224,6 +227,7 @@ public String getRandomTenantId() {
      * @see #deleteObjects(TestContext)
      */
     public String getRandomDeviceId(final String tenantId) {
+        tenantsToDelete.add(tenantId);
```
sophokles73: we do not add the tenant to the set of tenants to delete in order to be able to safely use the `DEFAULT_TENANT` (which shouldn't be deleted)

tests/src/test/resources/amqp/logback-spring.xml
sophokles73: EOL still missing

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
```diff
@@ -0,0 +1,289 @@
...
+                LOG.info("message {} received", MESSAGES_TO_SEND - received.getCount());
+            }
+        })).setHandler(context.asyncAssertSuccess(ok -> setup.complete()));
+        setup.await();
+
+        final AtomicInteger messageCount = new AtomicInteger(0);
+
+        while (messageCount.get() < MESSAGES_TO_SEND) {
+            final Async sendingComplete = context.async();
+            getOrCreateSender(tenantId, client).compose(sender -> {
```
sophokles73: I do not think that you need to create the sender for each message to send

Alfusainey:
> Can't we simply use the integration tests that we originally created for Hono Messaging (in package org.eclipse.hono.tests.client)?

just looked at that code. yes i think we can use it :-)

tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
Alfusainey: oh i see..

tests/pom.xml
Alfusainey: thanks. forgot to change it back

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
Alfusainey: > EOL still missing

question: what is the purpose of adding an EOL to the file?

sophokles73: https://stackoverflow.com/a/729795/6138465

Alfusainey:
@sophokles73 : is there a reason for not exposing a call to assert a device's registration status through the registry's `HttpEndpoint`? i see that the call is only exposed in the `AmqpEndpoint`

sophokles73:
who should invoke this operation? The Device Registration API is defined for the purpose of the protocol adapters' needs. They use AMQP 1.0 to interact with it. So what good would it be?

Alfusainey:
the integration tests for the protocol adapters accesses the device registry through http (to register devices, add credentials, tenants etc).however, the ITs in `org.eclipse.hono.tests.client` uses the amqp client to acccess the device registry (to register devices and to assert registration status).

i was thinking of having one client to do all that (that why the test for the messaging and the amqp adapter will use the http client to query the device registry). and have this client shared among the two ITs.

Alfusainey:
@sophokles73 i updated the patch to reuse the integration test for hono messaging.

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
```diff
@@ -1,289 +1,151 @@
...
-            final String expectedErrorMsg = String
-                    .format("This adapter is not enabled for tenant [tenantId: %s].", tenantId);
-
-            context.assertEquals(expectedErrorMsg, ((ClientErrorException) t).getMessage());
+            context.assertEquals(HttpURLConnection.HTTP_BAD_REQUEST, ((ClientErrorException) t).getErrorCode());
 
             completionTracker.complete();
         }));
-        completionTracker.awaitSuccess();
+        completionTracker.await(1000);
```
sophokles73: there is no need to create an `Async` and wait for it to complete. The test runner will already wait for the `context.asyncAssertFailure()` handler to fail and then perform the additional assertions you added there.

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
```diff
@@ -1,289 +1,151 @@
...
-            context.fail("did not receive all messages");
-        }
-    }
-
-    // ------------------------------< private methods >---
-
-    private static HonoClient prepareAmqpAdapterClient(final String username, final String password) {
         final ClientConfigProperties config = new ClientConfigProperties();
-        config.setName("test");
+        config.setName("EventAmqpIT");
```
sophokles73: since this is a generic class, does it make sense to use a name here that suggests we are sending events?

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestBase.java
```diff
@@ -230,7 +193,29 @@ private void disconnect(final TestContext ctx) {
...
+    /**
+     * Prepares the AMQP 1.0 client for connecting to one of Hono's services.
+     * 
+     * @param ctx The Vert.x test context.
+     * 
+     * @return The client for connecting to a Hono service.
+     */
+    protected abstract HonoClient prepareHonoClient(TestContext ctx);
+
+    protected abstract Future<Void> registerDevice();
```
sophokles73: I would rather create and register random tenants and devices as part of the individual tests (e.g. as done in the HTTP and MQTT adapter ITs).

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestBase.java
```diff
@@ -50,15 +49,12 @@
  */
 public abstract class ClientTestBase {
 
-    /**
-     * The identifier of the device used throughout the test cases.
-     */
-    protected static final String DEVICE_ID = "device-0";
+    protected static final String TEST_TENANT_ID = System.getProperty(IntegrationTestSupport.PROPERTY_TENANT, Constants.DEFAULT_TENANT);
+    protected static final Vertx  vertx = Vertx.vertx(new VertxOptions().setBlockedThreadCheckInterval(1000 * 60 * 60));
```
sophokles73: ???

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestBase.java
```diff
@@ -50,15 +49,12 @@
...
+    static final long   DEFAULT_TEST_TIMEOUT = 20000; // ms
```
sophokles73: is this used somewhere?

tests/src/test/java/org/eclipse/hono/tests/client/AbstractMessagingClientTest.java
```diff
@@ -0,0 +1,219 @@
...
+            LOGGER.debug("deregistering devices");
+            registrationClient.deregister(DEVICE_ID).setHandler(r -> done.complete());
+            done.await();
+        }
+
+        disconnect(ctx);
+    }
+
+    /**
+     * Assert properties of a number uploaded to Hono's Telemetry or Event API via the AMQP Messaging Network.
```
sophokles73: there seem to be some words missing ...

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
```diff
@@ -1,289 +1,151 @@
...
     @BeforeClass
-    public static void init(final TestContext context) {
-        vertx = Vertx.vertx();
+    public static void init(final TestContext ctx) {
         helper = new IntegrationTestSupport(vertx);
-        helper.init(context);
+        helper.initRegistryClient(ctx);
+
+        final Async setup = ctx.async();
+        final TenantObject tenant = TenantObject.from(AMQP_ADAPTER_TEST_TENANT, true);
```
sophokles73: I would rather have the individual tests register random tenants and devices ...

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
Alfusainey: i see, thanks

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
Alfusainey: not at all :+1:

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestBase.java
Alfusainey: yes, it use in the test case for that class when calculating the time a thread should wait for all messages to be consumed

tests/src/test/java/org/eclipse/hono/tests/client/AbstractMessagingClientTest.java
Alfusainey: :+1:

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
Alfusainey: thanks. will work on this now

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestBase.java
Alfusainey: ok

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestBase.java
Alfusainey: :+1:

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestBase.java
Alfusainey: @sophokles73 am having an issue with this proposition i.e creating a random tenant for the hono messaging tests. during the test, clients connect to the messaging service using the user `hono-client`, however, this user is only authorize to send events/telemetry for the default tenant.

this is different though for the protocol adapters -> they are authorize to publish telemetry/events for any tenant (so generating random tenantIds works in this case).

Can i change the permissions on record for the `hono-client` user so that I can create a sender for a tenant? WDYT?

Alfusainey: or maybe generate random tenants only for the amqp adapter and use the default tenant for the hono-messaging tests?

sophokles73: On a second thought, we could also use one of the protocol adapter users for the messaging tests, couldn't we?

Alfusainey: yes sure. i will try that in the next commit

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestBase.java
```diff
@@ -230,7 +201,58 @@ private void disconnect(final TestContext ctx) {
...
+     * @param ctx The Vert.x test context.
+     */
+    protected abstract void sendMessage(String payload, TestContext ctx);
+
+    /**
+     * The id of the tenant under test.
+     * 
+     * @return The tenant identifier created during the test.
+     */
+    protected abstract String getTenantId();
```
sophokles73: why do you want to store this in a class level field? I would assume that the tenant(s) and device(s) required by the test cases are created as part of the test cases themselves, using the helper to get random identifiers ...

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestBase.java
```diff
@@ -230,7 +201,58 @@ private void disconnect(final TestContext ctx) {
...
+
+    /**
+     * Gets the random device-id that was generated during a test execution.
+     * 
+     * @return The random device-id.
+     */
+    protected abstract String getRandomDevice();
```
sophokles73: same here

Alfusainey: please note that there is a `@Test` in the `ClientTestBase` that needs access to the device that was generated during connection to the adapter. for hono-messaging, we can generate a new one but not for the adapter(since a connection was already established before the test starts to execute)

Alfusainey:
> This PR is getting way too big.

Yes, that's true

> Can you please separate out the changes you want/need to make on the AMQP adapter implementation? This PR should then only be about adding the integration tests.

Then IMHO, its better to not have the adapter reuse tests for hono-messaging. we can try to factor out commonalities into a utility class to achieve code sharing.. but using inheritance will certainly increase the patch size for this PR

Alfusainey:
@sophokles73. I updated the pull request. hopefully this commit is closer to something that is ready to accept :-)

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
```diff
@@ -202,88 +175,110 @@ public void testUploadingXnumberOfMessages(final TestContext context) throws Int
...
+     * <li>Create an AMQP 1.0 client and authenticate it to the protocol adapter with username: {@code [device-id@tenantId]}.</li>
+     * <li>After a successful connection, create a producer/sender for sending messages to the protocol adapter.</li>
+     * </ul>
+     *
+     * @param tenantId The tenant to register with the device registry.
+     * @param deviceId The device to add to the tenant identified by tenantId.
+     * @param disableTenant If true, disable the protocol adapter for the tenant.
+     * @param ctx The Vert.x test context.
+     */
+    private void setupProtocolAdapter(final String tenantId, final String deviceId, final boolean disableTenant, final TestContext ctx) {
```
sophokles73: why not return a `Future<MessageSender>`?

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
```diff
@@ -147,47 +127,40 @@ public void testAdapterRejectsBadInboundMessage(final TestContext context) {
...
+     * Verifies that the AMQP Adapter will fail to authenticate a device whose username does not match the expected pattern
+     * {@code [<authId>@<tenantId>]}.
+     * 
+     * @param context The Vert.x test context.
+     */
+    @Test(timeout = 1000)
+    public void testAuthenticationWithInvalidUsernamePattern(final TestContext context) {
+        connectToAdapter("invalidaUsername", DEVICE_PASSWORD, true)
+        .setHandler(context.asyncAssertFailure(t -> {
+            context.assertTrue(ServiceInvocationException.class.isInstance(t));
```
sophokles73: why not check for ClientErrorException and status code 401?

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestHelper.java
```diff
@@ -0,0 +1,121 @@
+package org.eclipse.hono.tests.client;
```
sophokles73: missing license header

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
```diff
@@ -1,21 +1,15 @@
 package org.eclipse.hono.tests.amqp;
```
sophokles73: missing license header

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
Alfusainey: ok sure..

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
Alfusainey: yes, make sense

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
Alfusainey: humm.. just see that this will make the tests less more readable :-)

Alfusainey:
@sophokles73 just update the PR

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
```diff
@@ -220,13 +255,14 @@ private void setupProtocolAdapter(final String tenantId, final String deviceId,
...
                     .map(s -> {
-                        sender = s;
+                        result.complete(s);
                         return s;
                     });
 
                 }).setHandler(ctx.asyncAssertSuccess(ok -> setup.complete()));
 
         setup.await(2000);
+        return result;
```
sophokles73: It doesn't make much sense to wait for an Async now anymore, does it? Instead, the invoking code should simply compose the returned `Future<MessageSender>` with the code that uses the sender ...

Alfusainey: :+1:

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
```diff
@@ -0,0 +1,317 @@
...
+                    return connectToAdapter(username, DEVICE_PASSWORD, false)
+                    .compose(connectedClient -> createProducer(tenantId))
+                    .map(s -> {
+                        result.complete(s);
+                        return s;
+                    });
+
+                });
+
+        return result;
```
sophokles73: what about
```
return helper.registry..addDeviceForTenant(tenant, deviceId, DEVICE_PASSWORD)
  .compose(ok -> connectToAdapter(username, DEVICE_PASSWORD, false))
  .compose(client -> createProducer(tenantId));
```
?

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestHelper.java
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * A logger to be used by subclasses.
+     */
+    protected final Logger LOGGER = LoggerFactory.getLogger(getClass());
+
+    private static final long   DEFAULT_TEST_TIMEOUT = 15000; // ms
+
+    protected MessageSender sender;
+    protected MessageConsumer consumer;
+
```
sophokles73: these fields are never used locally, are they? Which method is responsible for initializing them and closing the links later?

Alfusainey: > these fields are never used locally, are they?

Yes, they are used by subclasses.

> Which method is responsible for initializing them

They are initialized either before a test executes (`@Before` in the `ClientTestBase`) or during test execution (that's what the `setupProtocolAdapter` does). the creation and initialization is left to subclasses

> and closing the links later?

Every subclass runs its own `@After` for closing the links. so the closing of the links is left to subclasses

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterBase.java
Alfusainey: thanks. PR updated

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestHelper.java
sophokles73: How is the implementor of a subclass supposed to know about all of this? The *createConsumer* and *createProducer* methods seem to indicate that a subclass is only responsible for creating these objects. However, the JavaDocs of those methods do not indicate that the respective result should be used to initialize the fields. And neither does it instruct implementors to close those objects once they are done. Adding these instructions to the JavaDoc, on the other hand, will still not make for a strong contract, because everything is dependent on the implementor's ability to follow the rules. IMHO neither the fields not the methods should be declared on ClientTestHelper. Instead, the *doUploadMessages* method should probably accept a sender and consumer as params ...

Alfusainey: yep, i see now. actually, the reason i placed them in the parent is to avoid declaring the same fields in both subclass (but i didnt see the bigger picture :+1: ). thanks..will update the PR

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestHelper.java
```diff
@@ -0,0 +1,115 @@
...
+    private static final long   DEFAULT_TEST_TIMEOUT = 15000; // ms
+
+    /**
+     * Creates a test specific message consumer.
+     *
+     * @param tenantId        The tenant to create the consumer for.
+     * @param messageConsumer The handler to invoke for every message received.
+     * @return A future succeeding with the created consumer.
+     */
+    protected abstract Future<MessageConsumer> createConsumer(String tenantId, Consumer<Message> messageConsumer);
```
sophokles73: this method is never invoked from this class, is it?

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestHelper.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * Creates a test specific message sender.
+     *
+     * @param tenantId     The tenant to create the sender for.
+     * @return A future succeeding with the created sender.
+     */
+    protected abstract Future<MessageSender> createProducer(String tenantId);
```
sophokles73: this method is never invoked from this class, is it?

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestHelper.java
```diff
@@ -0,0 +1,115 @@
...
+import org.slf4j.LoggerFactory;
+
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.ext.unit.TestContext;
+
+/**
+ * Base class for integration tests for Hono's AMQP 1.0 services.
+ */
+public abstract class ClientTestHelper {
```
sophokles73: please rename to e.g. ClientTestBase. We usually use *Helper* for classes that are mere utility classes providing static methods only.

sophokles73:
@Alfusainey please split up this PR into a part that deals with integration tests and another one for dealing with the rest. It is very hard to review the code ...

Alfusainey:
@sophokles73. ok i will do that. the only part that does not deal with ITs is the changes made to the amqp adapter codebase (i.e in `AmqpAdapterSaslAuthenticatorFactory.java`, `AmqpContext.java`, and `VertxBasedAmqpProtocolAdapter.java`)

If i should split this PR, then i will need those changes to first be on master before I can run the ITs

Alfusainey:
@sophokles73. i created PR #714. once we have that on  master, i will continue with this PR

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestHelper.java
Alfusainey: Yes true! i figured by putting in the parent class, we avoid duplicating code. isn't that better than defining the methods in both subclasses?

sophokles73: Well, FMPOV it only makes sense to declare abstract methods if those methods are use/required by some code in the same class that declares them. In your case it would make sense to declare the methods in the abstract class, if all test cases were defined in the abstract class as well and would use those methods to create the (subclass specific) senders and consumers. The subclasses would then e.g. be used to provide variants of sender and consumer and maybe implement additional test cases.

On the other hand, why should a subclass be forced to implement methods that are not used anywhere? So, unless you redesign the abstract class to actually invoke those methods, they should not be declared there IMHO.

Alfusainey: i see your point. will update the PR... thanks!

Alfusainey:
@sophokles73: PR updated

tests/src/test/java/org/eclipse/hono/tests/client/ClientTestBase.java
```diff
@@ -63,13 +40,13 @@
      * @param senderHandler The handler to call for sending messages to the messaging network.
      * @throws InterruptedException if test execution is interrupted.
      */
-    protected void doUploadMessages(final TestContext context, final Handler<CountDownLatch> consumerHandler, final Handler<String> senderHandler) throws InterruptedException {
+    protected void doUploadMessages(final TestContext context, final Handler<CountDownLatch> consumerCreationHandler, final Handler<String> senderHandler) throws InterruptedException {
```
sophokles73: `@param consumerHandler` does not match param name used in signature. The description of the param is also wrong. The desription of the params should also indicate, what the type argument is representing, e.g. I suppose that the senderHandler will handle a message's payload, right?

FMPOV it would be more intuitive to use `java.util.function.Consumer` instead of `Handler` and then call the parameters *receiver* and *sender*, don't you think?

tests/src/test/java/org/eclipse/hono/tests/client/MessagingClientTestBase.java
```diff
@@ -52,7 +55,7 @@
     private static final String TEST_TENANT_ID = System.getProperty(IntegrationTestSupport.PROPERTY_TENANT, Constants.DEFAULT_TENANT);
     private static final String CONTENT_TYPE_TEXT_PLAIN = "text/plain";
     private static final Vertx  vertx = Vertx.vertx();
-    private static final long   DEFAULT_TEST_TIMEOUT = 15000; // ms
+    //private static final long   DEFAULT_TEST_TIMEOUT = 15000; // ms
```
sophokles73: remove whole line, if not used

Alfusainey:
@sophokles73. i updated the PR to incorporate your latest comments

sophokles73:
@Alfusainey can you squash down to the desired/reasonable number of commits and rebase? I will then merge the PR.

Alfusainey:
@sophokles73: this is the first time that i had to jump in celebration seeing that the PR is accepted. waaw, great!

what do you think: squash to two meaning commits (separate changes to adapter from changes to hono-messaging tests) or a single commit?

sophokles73:
@Alfusainey yeah, this one took some time because there was so much re-factoring involved. Future PRs should be much faster to discuss/merge.

I think two commits would do fine.

Alfusainey:
@sophokles73. squashed and rebased. thanks!

PR #667, opened 2018-06-24
Improve service configuration

Alfusainey:
* The `400 (Bad Request)` status code is not correctly formatted in the HTTP Adapter documentation. It should have a formatting similar to the `403 (Forbidden)` status code

* Bootstrapping a protocol adapter requires doing the ff. minimal checks:
  * check if ports are configured correctly
  * check if either a secure or non-secure port is enabled.

All these checks relies on making a call to `AbstractConfig#getKeyCertOptions()`. Each time such a call is made,we instantiate a new `KeyCertPathOptions` object and set the path to `key and `cert` files. 

This patch caches the instance so that it is used by subsequent calls.
@sophokles73

### DIFF
--- a/site/content/user-guide/http-adapter.md
+++ b/site/content/user-guide/http-adapter.md
@@ -41,9 +41,9 @@ NB: There is a subtle difference between the *device identifier* (*device-id*) a
   * 202 (Accepted): The telemetry data has been accepted for processing. Note that if the `QoS-Level` request header is missing, the adapter does not *guarantee* successful delivery to potential consumers.
   However, if the QoS-Level header is set to `AT_LEAST_ONCE(1)`, then the adapter waits for the message to be delivered and accepted by a downstream peer before responding with a 202 status code to the device.
   * 400 (Bad Request): The request cannot be processed. Possible reasons for this might be that:
-    * The content type header is missing
-    * The request body is empty
-    * The QoS header value is invalid
+        * The content type header is missing.
+        * The request body is empty.
+        * The QoS header value is invalid.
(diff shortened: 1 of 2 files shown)


site/content/user-guide/http-adapter.md
```diff
@@ -41,9 +41,9 @@ NB: There is a subtle difference between the *device identifier* (*device-id*) a
   * 202 (Accepted): The telemetry data has been accepted for processing. Note that if the `QoS-Level` request header is missing, the adapter does not *guarantee* successful delivery to potential consumers.
   However, if the QoS-Level header is set to `AT_LEAST_ONCE(1)`, then the adapter waits for the message to be delivered and accepted by a downstream peer before responding with a 202 status code to the device.
   * 400 (Bad Request): The request cannot be processed. Possible reasons for this might be that:
-    * The content type header is missing
-    * The request body is empty
-    * The QoS header value is invalid
+        * The content type header is missing.
```
sophokles73: thanks for spotting (and fixing). However, this seems to be completely unrelated to the improvements regarding configuration. So, please put this into a separate commit.

Alfusainey: should i use the same PR and have 2 commits or put it in a separate PR?

sophokles73: In general, it would be preferrable if you would nto mix up separate issues in the same PR. In this case, however, it will be sufficient to simply create a separate commit for the docs (because it is so trivial).

Alfusainey: thanks. PR update!

PR #668, opened 2018-06-25
Improve error handling during command delivery.

sophokles73:
- The adapter rejects malformed command messages.
- The adapter acknowledges commands only after the response has been
sent to the device.
- The adapter handles errors when sending the response to the device.
- The command specific message handling has been factored out into a
dedicated set of classes (Command and CommandResponse).


sysexcontrol:
After the http replies are implemented in the master, we decided to close this PR and I will take over to create a new rebased PR in which also the new Response classes will be used for the replies.

Issue #633, opened 2018-05-28
Command and Control support for MQTT adapter


sysexcontrol:
Analogous to the HTTP adapter.

sophokles73:
@pellmann are you still working on this or can we close the issue?

PR #672, opened 2018-06-27
[#633] Command and Control support for MQTT adapter.

pellmann:
- response from device is not included right now.

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
@@ -14,11 +14,18 @@
 package org.eclipse.hono.adapter.mqtt;
 
 import java.net.HttpURLConnection;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 
+import io.vertx.mqtt.MqttTopicSubscription;
+import io.vertx.mqtt.messages.MqttPublishMessage;
+import io.vertx.proton.ProtonDelivery;
 import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.client.ClientErrorException;
+import org.eclipse.hono.client.MessageConsumer;
 import org.eclipse.hono.client.MessageSender;
 import org.eclipse.hono.client.ServerErrorException;
 import org.eclipse.hono.client.ServiceInvocationException;
@@ -34,6 +41,7 @@ import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.EndpointType;
 import org.eclipse.hono.util.EventConstants;
 import org.eclipse.hono.util.ExecutionContext;
+import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.ResourceIdentifier;
 import org.eclipse.hono.util.TelemetryConstants;
 import org.eclipse.hono.util.TenantObject;
@@ -73,6 +81,10 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends ProtocolAd
     private static final int IANA_MQTT_PORT = 1883;
     private static final int IANA_SECURE_MQTT_PORT = 8883;
 
+    private static final String COMMAND_REQUEST_PREFIX = "control/req";
+    private static final String COMMAND_RESPONSE_PREFIX = "control/res";
+    private static final String COMMAND_REQUEST_SUBSCRIPTION = COMMAND_REQUEST_PREFIX + "/#";
+
     /**
      * A logger to be used by concrete subclasses.
      */
@@ -83,6 +95,8 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends ProtocolAd
     private MqttServer server;
     private MqttServer insecureServer;
     private HonoClientBasedAuthProvider usernamePasswordAuthProvider;
+
+    private HashMap<String, MqttEndpoint> endpoints = new HashMap();
 
     /**
      * Sets the provider to use for authenticating devices based on a username and password.
@@ -442,8 +456,67 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends ProtocolAd
 
     private void onAuthenticationSuccess(final MqttEndpoint endpoint, final Device authenticatedDevice) {
 
+        // subscription handler for commands
+        endpoint.subscribeHandler(subscribe -> {
+            if(subscribe.topicSubscriptions().size()==1) {
+                final MqttTopicSubscription subscription = subscribe.topicSubscriptions().get(0);
+                final List<MqttQoS> grantedQosLevels = new ArrayList<>();
+                if (subscription.topicName().equals(COMMAND_REQUEST_SUBSCRIPTION)
+                            && subscription.qualityOfService() != MqttQoS.EXACTLY_ONCE) { // TODO: define allowed QoS 1 or 0 ?
+                    LOG.debug("subscription for [topic:{}, QoS:{}] from [tenant:{}, device:{}]",
+                            subscription.topicName(), subscription.qualityOfService(),
+                            authenticatedDevice.getTenantId(), authenticatedDevice.getDeviceId());
+                    grantedQosLevels.add(subscription.qualityOfService());
+                    // remember the endpoint for this device
+                    endpoints.put(Device.asAddress(authenticatedDevice), endpoint);
+                    endpoint.subscribeAcknowledge(subscribe.messageId(), grantedQosLevels);
+                    createCommandConsumer(authenticatedDevice).setHandler(h-> {
+                        if(h.succeeded()) {
+                            h.result().flow(1);
+                            sendTtdEvent(authenticatedDevice, -1);
+                        }
+                        else {
+                            LOG.error("command consumer could not be created for [tenant:{}, device:{}]: {}",
+                                    authenticatedDevice.getTenantId(), authenticatedDevice.getDeviceId(), h.cause());
+                        }
+                    });
+                }
+                else {
+                    LOG.error("denied subscription other then [{}] for [topic:{}, QoS:{}] from [tenant:{}, device:{}]",
+                            COMMAND_REQUEST_SUBSCRIPTION, subscription.topicName(), subscription.qualityOfService(),
+                            authenticatedDevice.getTenantId(), authenticatedDevice.getDeviceId());
+                }
+            }
+            else {
+                LOG.error("subscription for more/other topics then [{}] is not allowed - [tenant:{}, device:{}]",
+                        COMMAND_REQUEST_SUBSCRIPTION, authenticatedDevice.getTenantId(),
+                        authenticatedDevice.getDeviceId());
+            }
+        }).unsubscribeHandler(unsubscribe -> {
+            if(unsubscribe.topics().size()==1) {
+                final String topic = unsubscribe.topics().get(0);
+                if(topic.equals(COMMAND_REQUEST_SUBSCRIPTION)) {
+                    LOG.debug("unsubscribe for [topic:{}] from [tenant:{}, device:{}]",
+                            topic, authenticatedDevice.getTenantId(), authenticatedDevice.getDeviceId());
+                    endpoint.unsubscribeAcknowledge(unsubscribe.messageId());
+                    endpoints.remove(Device.asAddress(authenticatedDevice));
+                    sendTtdEvent(authenticatedDevice,0);
+                }
+                else {
+                    LOG.debug("denied unsubscribe for [topic:{}] from [tenant:{}, device:{}]",
+                            topic, authenticatedDevice.getTenantId(), authenticatedDevice.getDeviceId());
+                }
+            }
+            else {
+                LOG.error("unsubscribe for more/other topics then [{}] is not allowed - [tenant:{}, device:{}]",
+                        COMMAND_REQUEST_SUBSCRIPTION, authenticatedDevice.getTenantId(),
+                        authenticatedDevice.getDeviceId());
+            }
+        });
+
         endpoint.closeHandler(v -> {
             close(endpoint, authenticatedDevice);
+            endpoints.remove(Device.asAddress(authenticatedDevice));
             LOG.debug("connection to device [tenant-id: {}, device-id: {}] closed",
                     authenticatedDevice.getTenantId(), authenticatedDevice.getDeviceId());
             metrics.decrementMqttConnections(authenticatedDevice.getTenantId());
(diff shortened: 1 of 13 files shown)


service-base/src/main/java/org/eclipse/hono/service/command/CommandConnection.java
```diff
@@ -40,10 +40,20 @@
...
 
+    /**
+     * Close the command receiver (and removes it from internal map).
+     *
+     * @param tenantId The tenant to consume commands from.
+     * @param deviceId The device for which the consumer will be created.
+     * @return A future indicating the result of the closing operation.
+     * @throws NullPointerException if tenantId or deviceId is {@code null}.
+     */
+    Future<Void> closeCommandReceiver(String tenantId, String deviceId);
```
sysexcontrol: Either `consumer`, or `receiver`.
Currently these two words are used at the same time, while denoting the same, right?

sysexcontrol: Since we have used `sender` below, `receiver` might be the better counterpart. WDYT?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -442,8 +456,67 @@ private void handleConnectionRequestResult(final MqttEndpoint endpoint,
 
     private void onAuthenticationSuccess(final MqttEndpoint endpoint, final Device authenticatedDevice) {
 
+        // subscription handler for commands
+        endpoint.subscribeHandler(subscribe -> {
+            if(subscribe.topicSubscriptions().size()==1) {
+                final MqttTopicSubscription subscription = subscribe.topicSubscriptions().get(0);
+                final List<MqttQoS> grantedQosLevels = new ArrayList<>();
+                if (subscription.topicName().equals(COMMAND_REQUEST_SUBSCRIPTION)
```
sysexcontrol: Probably `subscription.topicName()` can never be null, but I would find it more defensive to turn around the check to 
`COMMAND_REQUEST_SUBSCRIPTION.equals(subscription.topicName())`.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -453,6 +526,25 @@ private void onAuthenticationSuccess(final MqttEndpoint endpoint, final Device a
         metrics.incrementMqttConnections(authenticatedDevice.getTenantId());
     }
 
+    private Future<MessageConsumer> createCommandConsumer(final Device authenticatedDevice) {
+        return getCommandConnection().getOrCreateCommandConsumer(authenticatedDevice.getTenantId(), authenticatedDevice.getDeviceId(), (delivery, message) -> {
+            LOG.debug("command received for [{}] with delivery ", authenticatedDevice);
+            onCommandReceived(authenticatedDevice.getTenantId(), authenticatedDevice.getDeviceId(), delivery, message);
+        }, close -> {
+            LOG.debug("command receiver link closed remotely"); // TODO reopen?
+            getCommandConnection().closeCommandReceiver(authenticatedDevice.getTenantId(), authenticatedDevice.getDeviceId());
```
sysexcontrol: IMHO better introduce a `CompositeFuture` for the close calls and collect the outcomes.

sysexcontrol: But why are all reponse senders implicitly closed when the command receiver link was closed (remotely)?
Is that resonable?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -453,6 +526,25 @@ private void onAuthenticationSuccess(final MqttEndpoint endpoint, final Device a
...
+            getCommandConnection().closeCommandResponseSenders(authenticatedDevice.getTenantId(), authenticatedDevice.getDeviceId());
```
sysexcontrol: But why are all reponse senders implicitly closed when the command receiver link was closed (remotely)?
Is that resonable?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -453,6 +526,25 @@ private void onAuthenticationSuccess(final MqttEndpoint endpoint, final Device a
...
+        }).map(consumer -> {
+            consumer.flow(1);
+            return consumer;
+        });
+    }
+
+    private void closeCommandLinks(final Device authenticatedDevice) {
```
sysexcontrol: This method should be used instead of the code above.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -797,4 +889,56 @@ protected void onMessageSent(final MqttContext ctx) {
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected void onCommandReceived(final String tenantId, final String deviceId, final ProtonDelivery delivery, final Message message) {
+        super.onCommandReceived(tenantId, deviceId, delivery, message);
+
+        final MqttEndpoint endpoint = endpoints.get(Device.asAddress(tenantId, deviceId));
+        if(endpoint!=null) {
```
sysexcontrol: spaces

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -797,4 +889,56 @@ protected void onMessageSent(final MqttContext ctx) {
...
+        }
+    }
+
+    /**
+     * Sends the event of a subscribed device, that is now online (ttd==-1), or disconnects (ttd=0).
+     *
+     * @param authenticatedDevice The device, that wants to handle commands.
+     * @param ttd The time til disconnect.
+     */
+    protected void sendTtdEvent(final Device authenticatedDevice, final Integer ttd) {
```
sysexcontrol: IMHO we have retrieved the TenantObject, the registrationAssertion and the event sender already at this point. Why do that again? Better would be to pass them in?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -442,8 +456,67 @@ private void handleConnectionRequestResult(final MqttEndpoint endpoint,
...
+                        COMMAND_REQUEST_SUBSCRIPTION, authenticatedDevice.getTenantId(),
+                        authenticatedDevice.getDeviceId());
+            }
+        }).unsubscribeHandler(unsubscribe -> {
+            if(unsubscribe.topics().size()==1) {
+                final String topic = unsubscribe.topics().get(0);
+                if(topic.equals(COMMAND_REQUEST_SUBSCRIPTION)) {
+                    LOG.debug("unsubscribe for [topic:{}] from [tenant:{}, device:{}]",
+                            topic, authenticatedDevice.getTenantId(), authenticatedDevice.getDeviceId());
+                    endpoint.unsubscribeAcknowledge(unsubscribe.messageId());
```
sysexcontrol: Shouldn't be the `commandConsumer` be closed here?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -442,8 +456,67 @@ private void handleConnectionRequestResult(final MqttEndpoint endpoint,
...
+            else {
+                LOG.error("unsubscribe for more/other topics then [{}] is not allowed - [tenant:{}, device:{}]",
+                        COMMAND_REQUEST_SUBSCRIPTION, authenticatedDevice.getTenantId(),
+                        authenticatedDevice.getDeviceId());
+            }
+        });
+
         endpoint.closeHandler(v -> {
             close(endpoint, authenticatedDevice);
+            endpoints.remove(Device.asAddress(authenticatedDevice));
```
sysexcontrol: And additionally close the `commandConsumer` ?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -73,6 +81,10 @@
     private static final int IANA_MQTT_PORT = 1883;
     private static final int IANA_SECURE_MQTT_PORT = 8883;
 
+    private static final String COMMAND_REQUEST_PREFIX = "control/req";
+    private static final String COMMAND_RESPONSE_PREFIX = "control/res";
+    private static final String COMMAND_REQUEST_SUBSCRIPTION = COMMAND_REQUEST_PREFIX + "/#";
```
sysexcontrol: IMHO we should move these three constants to either `Constants` and make them public,
or at least make them protected, so that sub classes of this abstract class can use them.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -84,6 +96,8 @@
     private MqttServer insecureServer;
     private HonoClientBasedAuthProvider usernamePasswordAuthProvider;
 
+    private HashMap<String, MqttEndpoint> endpoints = new HashMap();
```
sysexcontrol: IMHO this should be `protected final`.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
sysexcontrol: This occurs at some more places in the code - just toggle the both sides of `equals`.

example/src/main/deploy/docker/swarm_deploy.sh
```diff
@@ -76,7 +76,7 @@ echo Deploying Qpid Dispatch Router ...
 docker secret create -l project=$NS qdrouter-key.pem $CERTS/qdrouter-key.pem
 docker secret create -l project=$NS qdrouter-cert.pem $CERTS/qdrouter-cert.pem
 docker secret create -l project=$NS qdrouterd.json $SCRIPTPATH/qpid/qdrouterd-with-broker.json
-docker service create $CREATE_OPTIONS --name hono-dispatch-router -p 15671:5671 -p 15672:5672 \
+docker service create $CREATE_OPTIONS --name hono-dispatch-router -p 15671:5671 -p 15672:5672 -p 15673:5673 \
```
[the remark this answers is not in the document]
pellmann: It was an accident to commit it.

service-base/src/main/java/org/eclipse/hono/service/command/CommandConnection.java
```diff
@@ -40,10 +40,20 @@
      *         the link cannot be established, e.g. because this client is not connected.
      * @throws NullPointerException if tenantId, deviceId or messageConsumer is {@code null}.
      */
-    Future<MessageConsumer> createCommandConsumer(String tenantId, String deviceId,
+    Future<MessageConsumer> getOrCreateCommandConsumer(String tenantId, String deviceId,
```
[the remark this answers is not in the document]
pellmann: Of course it should be used also in the HTTP adapter! I understand that it would be easier with separate PRs.

service-base/src/main/java/org/eclipse/hono/service/command/CommandConnection.java
pellmann: At other parts of Hono "consumer" is used and also the existing method was called with consumer.., so I would stick with it.

sysexcontrol: Fine, but shouldn't then this method be renamed to `closeCommandConsumer` ?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
pellmann: I think the check for the tenant could be removed. The assertion comes from cache, so this should not be a problem. I do not think, that the sender exist in any case?

service-base/src/main/java/org/eclipse/hono/service/command/CommandConnection.java
pellmann: Sure - that is what I did. :)

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
pellmann: I agree that it would be better to stay with the response senders until the device disconnects. Or close after each send message - this depends on link creation vs. link count costs.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
pellmann: The question is if this is only MQTT or also other. Since I will need this in the ResourceIdentifier later, I agree to move it to Constants.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
sysexcontrol: The `TenantObject` comes from the cache as well. So basically we could remove the tenant, but:
I thought we want to make the behaviour if the MQTT adapter auto-sends a `ttd` event configurable per tenant in the future, and then the TenantObject would be needed here again.

The sender indeed is probably not present at this point of time (when the device just connected), so this cannot be eliminated, I agree.

But I have another proposal: 
very recently @sophokles73 introduced the `MqttContext` class, which could be used here to register key-value pairs.
For the authenticated device there is already a constructor available, but the generic `put` method is good for any pairs.
So why not put the TenantObject and the assertionToken into the context (which currently is constructed at the very end of `onAuthenticationSuccess` and pass that around?
This would IMHO solve the problem in an elegant way without repeating code...
WDYT?

pellmann: That might be a good solution but I think it is an optimization, that should be done in a second step.

sysexcontrol: Sounds reasonable :+1: . We'll do this in another step then.
IMHO the code can remain unchanged for the first step then.

pellmann:
I close this PR. The part with the cached command consumers is already done in an other PR and the MQTT part will be in a new one.

PR #674, opened 2018-06-28
[#643] Amqp Adapter - support for device authentication

Alfusainey:
@sophokles73 

This patch extends the Amqp adapter to provide support for authenticating client devices. Currently, only the `SASL PLAIN` mechanism is supported.

### DIFF
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
@@ -2,6 +2,7 @@ package org.eclipse.hono.adapter.amqp;
 
 import java.net.HttpURLConnection;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.qpid.proton.amqp.transport.AmqpError;
 import org.apache.qpid.proton.message.Message;
@@ -10,6 +11,9 @@ import org.eclipse.hono.client.MessageSender;
 import org.eclipse.hono.client.ServiceInvocationException;
 import org.eclipse.hono.config.ProtocolAdapterProperties;
 import org.eclipse.hono.service.AbstractProtocolAdapterBase;
+import org.eclipse.hono.service.auth.device.Device;
+import org.eclipse.hono.service.auth.device.HonoClientBasedAuthProvider;
+import org.eclipse.hono.service.auth.device.UsernamePasswordAuthProvider;
 import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.EndpointType;
 import org.eclipse.hono.util.EventConstants;
@@ -30,6 +34,7 @@ import io.vertx.proton.ProtonReceiver;
 import io.vertx.proton.ProtonServer;
 import io.vertx.proton.ProtonServerOptions;
 import io.vertx.proton.ProtonSession;
+import io.vertx.proton.sasl.ProtonSaslAuthenticatorFactory;
 
 /**
  * A Vert.x based Hono protocol adapter for publishing messages to Hono's Telemetry and Event APIs using AMQP.
@@ -42,14 +47,31 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
     private static final int DEFAULT_MAX_SESSION_WINDOW = 100 * DEFAULT_MAX_FRAME_SIZE;
 
     /**
-     * The default insecure port that this adapter binds to.
+     * The default insecure port that this adapter binds to for unencrypted connections.
      */
     private static final int DEFAULT_INSECURE_PORT = 4040;
+
+    /**
+     * The default secure port that this adapter binds to for TLS encrypted secure connections.
+     */
+    private static final int DEFAULT_SECURE_PORT = 4041;
+
+    /**
+     * The AMQP server instance that maps to a secure port.
+     */
+    private ProtonServer secureServer;
 
     /**
      * The AMQP server instance that listens for incoming request from an insecure port.
      */
     private ProtonServer insecureServer;
+
+    private AtomicBoolean secureListening = new AtomicBoolean(false);
+
+    /**
+     * This adapter's custom SASL authenticator factory for handling the authentication process for devices.
+     */
+    private ProtonSaslAuthenticatorFactory authenticatorFactory;
 
     // -----------------------------------------< AbstractProtocolAdapterBase >---
     /**
@@ -65,26 +87,26 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
      */
     @Override
     protected void doStart(final Future<Void> startFuture) {
-        final ProtonServerOptions options = new ProtonServerOptions();
-        options.setHost(getConfig().getInsecurePortBindAddress());
-        options.setPort(determineInsecurePort());
-        // options.setHeartbeat(6000); // set local-idle-timeout expiry
-        // TODO: communicate max-frame size
-        options.setMaxFrameSize(DEFAULT_MAX_FRAME_SIZE);
         checkPortConfiguration()
-                .compose(check -> {
-                    // TODO: check that adapter is connected to all its services.
-                    return bindServer(insecureServer, options);
-                }).compose(server -> {
-                    insecureServer = server;
+                .compose(success -> {
+                    if (authenticatorFactory == null && getConfig().isAuthenticationRequired()) {
+                        final HonoClientBasedAuthProvider usernamePasswordAuthProvider = new UsernamePasswordAuthProvider(getCredentialsServiceClient(), getConfig());
+                        authenticatorFactory = new AmqpAdapterSaslAuthenticatorFactory(usernamePasswordAuthProvider, getConfig());
+                    }
+                    return Future.succeededFuture();
+                }).compose(succcess -> {
+                    return CompositeFuture.all(bindSecureServer(), bindInsecureServer());
+                }).compose(success -> {
                     startFuture.complete();
                 }, startFuture);
 
     }
 
+
     @Override
     protected void doStop(final Future<Void> stopFuture) {
-        stopInsecureServer().compose(stopped -> stopFuture.complete(), stopFuture);
+        CompositeFuture.all(stopSecureServer(), stopInsecureServer())
+        .compose(ok -> stopFuture.complete(), stopFuture);
     }
 
     private Future<Void> stopInsecureServer() {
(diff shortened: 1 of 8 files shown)


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -97,24 +124,91 @@ protected void doStop(final Future<Void> stopFuture) {
...
+        } else {
+         // use proton's default authenticator -> SASL ANONYMOUS
+            createdServer.saslAuthenticatorFactory(null);
+        }
+        return createdServer;
+    }
+
     private void connectionRequestHandler(final ProtonConnection connRequest) {
+
         LOG.info("Received connection request from client");
```
sophokles73: better log at debug level

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/sasl/AmqpAdapterSaslAuthenticatorFactory.java
```diff
@@ -0,0 +1,214 @@
...
+                }
+                return Future.future();
+            }).recover(t -> {
+                completer.handle(Future.failedFuture(t));
+                return Future.failedFuture(t);
+            });
+
+        }
+
+        private Future<String[]> getFields(final byte[] saslResponse) {
```
sophokles73: IMHO this method should rather return a `Future<UsernamePasswordCredentials>` ...

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/sasl/AmqpAdapterSaslAuthenticatorFactory.java
```diff
@@ -0,0 +1,214 @@
...
+            // Get the authentication identity and password (ignoring authorization id)
+            final String authcid = new String(saslResponse, authzidEndPosition + 1,
+                    authcidEndPosition - (authzidEndPosition + 1), StandardCharsets.UTF_8);
+            final String passwd = new String(saslResponse, authcidEndPosition + 1,
+                    saslResponse.length - (authcidEndPosition + 1));
+
+            result.complete(new String[] { authcid, passwd });
+
+            return result;
+        }
```
sophokles73: all of this code for parsing the client's response data is a duplication of existing code (org.eclipse.hono.service.auth.AbstractHonoAuthenticationService), isn't it? Can't we find a way to put it into a place where we can re-use it properly?

adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorTest.java
```diff
@@ -0,0 +1,146 @@
...
+                    ProtonClient.create(vertx).connect("localhost", server.result().actualPort(), VALID_USERNAME,
+                            PASSWORD,
+                            clientConnection -> {
+                                // verify that credentials were successfully verified and the connection was successful.
+                                context.assertTrue(clientConnection.succeeded());
+                                clientConnection.result().disconnect();
+                                completionTracker.complete();
+                            });
+                });
+        completionTracker.awaitSuccess();
```
sophokles73: are you sure that you need to start a real AMQP server and client in order to verify the authenticator's behavior?

adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -264,6 +268,13 @@ private MessageSender givenATelemetrySenderForAnyTenant() {
         return sender;
     }
 
+    private Device getAuthenticatedDevice() {
+        final Device authenticatedDevice = mock(Device.class);
```
sophokles73: Is there a reason why you want to create a mock instead of just instantiating a `new Device(TEST_TENANT_ID, TEST_DEVICE)`?

core/src/main/java/org/eclipse/hono/util/Constants.java
```diff
@@ -88,6 +88,12 @@
      * attachments.
      */
     public static final String KEY_CLIENT_PRINCIPAL = "CLIENT_PRINCIPAL";
+    /**
+     * They key that an authenticated client of a protocol adapter (representing a device)
+     * is stored under in a {@code ProtonConnection}'s attachments.
+     */
+    public static final String KEY_CLIENT_DEVICE = "CLIENT_DEVICE";
```
sophokles73: this seems to be very specific to the AMQP adapter, i.e. it is of no relevance to any other component, is it? So, maybe you can find a more appropriate place for this constant?

adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorTest.java
Alfusainey: IMHO, having a client and server can trigger the SASL authentication process. and since the authenticator is used together with the proton server, i decided to test using a real server and client. WDYT?

adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapterTest.java
Alfusainey: :+1:

core/src/main/java/org/eclipse/hono/util/Constants.java
Alfusainey: yes sure..

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/sasl/AmqpAdapterSaslAuthenticatorFactory.java
sophokles73: Hmm, have you done any testing regarding the performance of both variants? I am usually cautious with statements regarding the runtime performance of code because you most of the times do not know what the VM actually makes out of the code.

Anyway, I do not think that we need multiple implementations of parsing the SASL response.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/sasl/AmqpAdapterSaslAuthenticatorFactory.java
```diff
@@ -0,0 +1,214 @@
...
+            if (authcidEndPosition < 0) {
+                result.fail(new CredentialException("PLAIN response must contain an authentication ID"));
+            }
+            if (authcidEndPosition == saslResponse.length - 1) {
+                result.fail(new CredentialException("PLAIN response must contain a password"));
+            }
+            // Get the authentication identity and password (ignoring authorization id)
+            final String authcid = new String(saslResponse, authzidEndPosition + 1,
+                    authcidEndPosition - (authzidEndPosition + 1), StandardCharsets.UTF_8);
+            final String passwd = new String(saslResponse, authcidEndPosition + 1,
```
sophokles73: which charset is used here?

adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorTest.java
sophokles73: FMPOV a unit test should try to narrow down the scope of the test to a single class. Running a client and server with real networking involved sounds more like an integration test to me. Maybe you can move this to the IT module?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/sasl/AmqpAdapterSaslAuthenticatorFactory.java
Alfusainey: good catch! should also be `StandardCharsets.UTF_8`
> Anyway, I do not think that we need multiple implementations of parsing the SASL response.

yes, that's true. i will drop this and use the implementation present with the `AuthenticationService`. can i move it to `AuthenticationConstants`?

Alfusainey:
ok, will do that!

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/sasl/AmqpAdapterSaslAuthenticatorFactory.java
Alfusainey: github comment boxes are a bit confusing - some of your comments are visible below with no reply-box

Alfusainey:
@sophokles73: PR updated!

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/sasl/AmqpAdapterSaslAuthenticatorFactory.java
```diff
@@ -167,48 +164,24 @@ private void verifyPlain(final String remoteMechanism, final byte[] saslResponse
...
+    }
 
-            return result;
-        }
+    /**
+     * Represent constants used throughout the Amqp adapter code base.
+     * 
+     * TODO: this class will be moved to a dedicated java file once the list of constants grows.
+     */
+    public final class AmqpAdapterConstants {
```
sophokles73: Hmm, why not define this as a class in the main package?
In fact, is it really helpful to have a separate sasl package instead of just the top level package? How many classes are there, five?

Alfusainey: currently there are 4 classes in total. adding it and the authenticator class will make it 6.

Alfusainey:
yes sure, i will. i mentioned that in a previous comment

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/sasl/AmqpAdapterSaslAuthenticatorFactory.java
sophokles73: so, do you think having them all in one package would be a problem?

Alfusainey: no not at all, i will update the PR and put them in one package

sophokles73: I think that the AMQP_BAD_REQUEST constant might be useful in all services that implement an AMQP server, not only in the AMQP adapter.

Alfusainey:
updated the PR! once approved, i will squashed the commits into one

sophokles73:
@Alfusainey the code doesn't seem to build on Travis ...

Alfusainey:
thanks! updated the PR

sophokles73:
still doesn't pass the travis build ...

adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -236,7 +236,7 @@ public void testUploadTelemetryMessageFailsForDisabledTenant() {
         final ProtonDelivery delivery = mock(ProtonDelivery.class);
         when(delivery.remotelySettled()).thenReturn(false);
         final ResourceIdentifier resource = ResourceIdentifier.from(TelemetryConstants.TELEMETRY_ENDPOINT, TEST_TENANT_ID, TEST_DEVICE);
-        final AmqpContext context = spy(new AmqpContext(delivery, getFakeMessage(), resource, mock(Device.class)));
+        final AmqpContext context = spy(new AmqpContext(delivery, getFakeMessage(), resource, null));
```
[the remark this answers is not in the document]
Alfusainey: we are testing for a disabled tenant here.. i realized for this test case, it is unnecessary to mock the device and can just pass in `null` (for unauthenticated device)

Alfusainey:
the travis build failing is a bit strange. i can locally build the sources, docker images and run integration tests with no problem i.e `mvn clean install -Pbuild-docker-image,run-tests`. so i wonder what the problem really is

Alfusainey:
oh i see what travis is complaining about (forgot to update the javadoc with the new constants class)

Alfusainey:
@sophokles73: any update on this PR?

sophokles73:

Alfusainey:
done!

PR #682, opened 2018-07-02
On all close calls to ProtonLink instances resources are freed.

sysexcontrol:
All calls to close that are not controlled by HonoClient additionally
call the helper method to free resources. This is true for error cases
and for closeHandlers (where defined). Resources are NOT freed in
onDetach methods (with closed=false).

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/amqp/AbstractAmqpEndpoint.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/amqp/AbstractAmqpEndpoint.java
@@ -18,6 +18,7 @@ import org.apache.qpid.proton.amqp.transport.ErrorCondition;
 import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.service.AbstractEndpoint;
 import org.eclipse.hono.util.Constants;
+import org.eclipse.hono.util.LinkHelper;
 import org.eclipse.hono.util.ResourceIdentifier;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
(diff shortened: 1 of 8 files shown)


sysexcontrol:
@sophokles73 : can you please review before we merge it (or not)?
Thanks!

service-base/src/main/java/org/eclipse/hono/service/amqp/AbstractAmqpEndpoint.java
```diff
@@ -108,13 +109,15 @@ public void onLinkAttach(final ProtonConnection con, final ProtonReceiver receiv
         logger.info("Endpoint [{}] does not support data upload, closing link.", getName());
         receiver.setCondition(ProtonHelper.condition(AmqpError.NOT_IMPLEMENTED, "resource cannot be written to"));
         receiver.close();
+        LinkHelper.freeLinkResources(receiver);
```
sophokles73: I would rather have two methods `LinkHelper.closeLink(final ProtonLink link, final boolean releaseResources)` and `LinkHelper.closeLink(final ProtonLink link)` which invokes the former with `true`. In our existing code we would then replace all occurrences of `ProtonLink.close` with the latter variant. Later, when vert.x 3.5.3 is released, we need to update only a single place in our code (LinkHelper), instead of removing the (then obsolete) call to LinkHelper.freeLinkResources`.

Additionally, we also should take care of sessions being closed in a similar (i.e. the same) way. However, for that, we should probably rename `LinkHelper` to `ProtonHelper` or something like that ...

sophokles73:
obsolete with 658fc04724652098939b0422f0ce25838bf5e168

Issue #688, opened 2018-07-03
Support more hash functions in Credentials API


dejanb:
Currently Credential API supports SHA-256 and SHA-512. For integration purposes, it'd be good to support a wider variety of algorithms, especially BCrypt.

dejanb:
This is now merged to the master

PR #689, opened 2018-07-03
#688 - add support for more flexible password encoding

dejanb:
This is an implementation based on Spring Security Crypto module with some modification to be better suited for Hono.

I'll add CQ and docs when we decide to proceed with this.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -82,6 +82,10 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.springframework.security</groupId>
+      <artifactId>spring-security-crypto</artifactId>
+    </dependency>
   </dependencies>
 
   <build>
@@ -128,10 +132,12 @@
             -->
             <Export-Package>
               io.opentracing*;version=${opentracing.version},
+              org.springframework.security.crypto*;version=${spring-security-crypto.version},
               {local-packages}
             </Export-Package>
             <Embed-Dependency>
-              opentracing-api;inline=true
+              opentracing-api;inline=true,
+              spring-security-crypto;inline=true
             </Embed-Dependency>
           </instructions>
         </configuration>
--- a/core/src/main/java/org/eclipse/hono/util/CredentialsObject.java
+++ b/core/src/main/java/org/eclipse/hono/util/CredentialsObject.java
@@ -12,9 +12,6 @@
  *******************************************************************************/
 package org.eclipse.hono.util;
 
-import java.nio.charset.StandardCharsets;
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
 import java.security.cert.X509Certificate;
 import java.time.Instant;
 import java.time.OffsetDateTime;
@@ -361,26 +358,15 @@ public final class CredentialsObject {
     }
 
     /**
-     * Creates a salted hash for a password.
-     * <p>
-     * Gets the password's UTF-8 bytes, prepends them with the salt (if not {@code null}
-     * and returns the output of the hash function applied to the byte array.
+     * Uses {@link ClearTextPassword#encode(String, byte[], String)} to return salted hash for a password.
      * 
      * @param hashFunction The hash function to use.
      * @param salt The salt to prepend the password bytes with.
      * @param password The password to hash.
-     * @return The hashed password.
-     * @throws NoSuchAlgorithmException if the given hash function is not supported on
-     *           the JVM.
+     * @return The hashed password as a Base64 formatted String.
      */
-    public static byte[] getHashedPassword(final String hashFunction, final byte[] salt, final String password) throws NoSuchAlgorithmException {
-
-        final MessageDigest digest = MessageDigest.getInstance(hashFunction);
-        if (salt != null) {
-            digest.update(salt);
-        }
-        digest.update(password.getBytes(StandardCharsets.UTF_8));
-        return digest.digest();
+    public static String getHashedPassword(final String hashFunction, final byte[] salt, final String password) {
+        return ClearTextPassword.encode(hashFunction, salt, password);
     }
 
     /**
@@ -445,21 +431,17 @@ public final class CredentialsObject {
         Objects.requireNonNull(password);
         Objects.requireNonNull(hashAlgorithm);
 
-        try {
-            final JsonObject secret = emptySecret(notBefore, notAfter);
-            secret.put(CredentialsConstants.FIELD_SECRETS_HASH_FUNCTION, hashAlgorithm);
-            if (salt != null) {
-                secret.put(
-                        CredentialsConstants.FIELD_SECRETS_SALT,
-                        Base64.getEncoder().encodeToString(salt));
-            }
+        final JsonObject secret = emptySecret(notBefore, notAfter);
+        secret.put(CredentialsConstants.FIELD_SECRETS_HASH_FUNCTION, hashAlgorithm);
+        if (salt != null) {
             secret.put(
-                    CredentialsConstants.FIELD_SECRETS_PWD_HASH,
-                    Base64.getEncoder().encodeToString(getHashedPassword(hashAlgorithm, salt, password)));
-            return secret;
-        } catch (final NoSuchAlgorithmException e) {
-            throw new IllegalArgumentException("unsupported hash algorithm");
+                    CredentialsConstants.FIELD_SECRETS_SALT,
+                    Base64.getEncoder().encodeToString(salt));
         }
+        secret.put(
+                CredentialsConstants.FIELD_SECRETS_PWD_HASH,
+                getHashedPassword(hashAlgorithm, salt, password));
+        return secret;
 
     }
 
--- a/core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
+++ b/core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
@@ -1,0 +1,103 @@
+/*******************************************************************************
+ * Copyright (c) 2018 Red Hat Inc and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat - initial creation
+ *******************************************************************************/
+package org.eclipse.hono.util;
+
+import io.vertx.core.json.JsonObject;
+import org.springframework.security.crypto.password.PasswordEncoder;
+
+import java.util.Base64;
+import java.util.Objects;
+
+/**
+ * Encoded password representation. It's used to parse Strings in the form of {Base64(salt)}password-hash into the value object, like
+ *
+ * <pre>
+ * EncodedPassword password = new EncodedPassword("{VGhlU2FsdA==}1L/qmnQ8kbgckAodOCbtyJAhoiK4k0rBtBBN+WD+TIE=");
+ * System.out.println(password.salt + " " + password.password);
+ * </pre>
+ *
+ * You can also use it to match passwords, like
+ *
+ * <pre>
+ * password.matches("ThePassword");
+ * </pre>
+ *
+ * This class also support matching passwords encrypted using BCrypt hashing function
+ *
+ * <pre>
+ * EncodedPassword password = new EncodedPassword("$2a$12$BjLeC/gqcnEyk.XNo2qorul.a/v4HDuOUlfmojdSZXRSFTjymPdVm");
+ * password.hashFunction = "bcrypt";
+ * password.matches("kapua-password");
+ * </pre>
+ *
+ */
+public class EncodedPassword extends BasePassword {
+
+    /**
+     * Creates an instance from the {Base64(salt)}password-hash formatted String.
+     *
+     * @param formattedPassword Password hash in the {Base64(salt)}password-hash format
+     */
+    public EncodedPassword(final String formattedPassword) {
+        parse(formattedPassword);
+    }
+
+    /**
+     * Creates an empty instance.
+     */
+    public EncodedPassword() {
+    }
+
+    /**
+     * Creates a new instance from Hono Secret.
+     *
+     * The secret is expected to be of type <em>hashed-password</em> as defined by
+     * <a href="https://www.eclipse.org/hono/api/credentials-api/#hashed-password">Hono's Credentials API</a>.
+     *
+     * @param honoSecret JSON object that contains hono-formatted secret
+     * @return Encoded password
+     * @throws IllegalArgumentException if candidate hashed-password secret does not contain a password hash
+     */
+    public static EncodedPassword fromHonoSecret(final JsonObject honoSecret) throws IllegalArgumentException {
+        Objects.requireNonNull(honoSecret);
+
+        final String pwdHash = honoSecret.getString(CredentialsConstants.FIELD_SECRETS_PWD_HASH);
+        if (pwdHash == null) {
+            throw new IllegalArgumentException("candidate hashed-password secret does not contain a pwd hash");
+        }
+
+        final String hashFunction = honoSecret.getString(
+                CredentialsConstants.FIELD_SECRETS_HASH_FUNCTION,
+                CredentialsConstants.DEFAULT_HASH_FUNCTION);
+
+        final String encodedSalt = honoSecret.getString(CredentialsConstants.FIELD_SECRETS_SALT);
+        final EncodedPassword encodedPassword = new EncodedPassword();
+        encodedPassword.hashFunction = hashFunction;
+        encodedPassword.password = pwdHash;
+        if (encodedSalt != null) {
+            encodedPassword.salt = Base64.getDecoder().decode(encodedSalt);
+        }
+        return encodedPassword;
+    }
+
+    /**
+     * Checks if provided password (in clear text form) matches this encoded password.
+     *
+     * @param password Password to match
+     * @return {@code true} if the password matches the stored secret.
+     */
+    public boolean matches(final String password) {
+        Objects.requireNonNull(password);
+        final PasswordEncoder encoder = PasswordEncoderFactory.createDelegatingPasswordEncoder();
+        return encoder.matches(password, format());
+    }
+
+}
--- a/core/src/main/java/org/eclipse/hono/util/MessageDigestPasswordEncoder.java
+++ b/core/src/main/java/org/eclipse/hono/util/MessageDigestPasswordEncoder.java
@@ -1,0 +1,91 @@
+/*******************************************************************************
+ * Copyright (c) 2018 Red Hat Inc and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *     Red Hat - initial creation
+ *******************************************************************************/
+package org.eclipse.hono.util;
+
+import org.springframework.security.crypto.password.PasswordEncoder;
+
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+import java.util.Base64;
+
+/**
+ * Implementation of {@link PasswordEncoder} that uses {@link MessageDigest} to encode passwords.
+ * This class creates SHA-256 or SHA-512 encoded passwords according to the type <em>hashed-password</em> as defined by
+ * <a href="https://www.eclipse.org/hono/api/credentials-api/#hashed-password">Hono's Credentials API</a>.
+ */
+public class MessageDigestPasswordEncoder implements PasswordEncoder {
+
+    private final MessageDigest messageDigest;
+
+    /**
+     * Creates message digest password encoder with specified hash function.
+     *
+     * @param hashFunction - hash function to be used
+     * @throws IllegalArgumentException if hash function is not valid
+     */
+    public MessageDigestPasswordEncoder(final String hashFunction) {
+        try {
+            messageDigest = MessageDigest.getInstance(hashFunction);
+        } catch (NoSuchAlgorithmException e) {
+            throw new IllegalArgumentException("No such hash function: " + e);
+        }
+    }
+
+    /**
+     *  Creates a salted hash for a password. The password is provided in {Base64(salt)}password format, where the salt is optional.
+     *  After the value is properly parsed,
+     *
+     *  @param rawPassword Clear text password to be hashed in {Base64(salt)}password format
+     *  @return Password hash
+     */
+    @Override
+    public String encode(final CharSequence rawPassword) {
+        final ClearTextPassword password = new ClearTextPassword(rawPassword.toString());
+
+        return Base64.getEncoder().encodeToString(digest(password.salt, password.password));
+
+    }
+
+    /**
+     * Verifies that password provided matches the one we have in the store of record. The stored password is in the {Base64(salt)}passwordHash format.
+     * The method will extract the salt (if present) and use it to encode the raw provided password. Then it will match the result with the
+     * hash on the record.
+     *
+     * @param rawPassword Password to verify in plain text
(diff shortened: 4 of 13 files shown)


core/src/main/java/org/eclipse/hono/util/MessageDigestPasswordEncoder.java
```diff
@@ -0,0 +1,86 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Red Hat Inc and others.
```
ctron: Welcome to 2018 :stuck_out_tongue_winking_eye:

ctron:
Just as an idea, wouldn't it make sense to make use of the Java Service Loader functionality and allow people to drop in their own encoders as well?

sophokles73:
In general I like the idea of using an existing (and hopefully well tested) implementation of high-quality hash functions for this purpose. However, I am a little concerned regarding the dependencies that are pulled in when using Spring Security. The reason is that we have made some efforts of keeping Hono Core and Client *Spring free* so that it can be used as an OSGi bundle more easily ...

core/src/main/java/org/eclipse/hono/util/MessageDigestPasswordEncoder.java
dejanb: Time flies 🙂

dejanb:
@ctron I was thinking about that at first, but figured it all needs to be containerized in the end anyways, so not sure if it's worth the effort. Also, I don't think it will be that common.

On the other hand, there are a couple of short-comings in the current interfaces that I'd like to eventually improve, so we can push it in that direction then.

sophokles73:
>  It's a self-contained jar with just codecs and interfaces to use them. So it shouldn't be dependency/CQ mess.

Then we should probably be able to inline it in the Hono core artifact, very much like we do with OpenTracing.

core/pom.xml
```diff
@@ -130,10 +134,12 @@
             -->
```
dejanb: @sophokles73 Jar and manifest now looks good to me

core/src/main/java/org/eclipse/hono/util/CredentialsObject.java
```diff
@@ -369,17 +367,16 @@ public static JsonObject emptySecret(final Instant notBefore, final Instant notA
...
      * @return The hashed password.
-     * @throws NoSuchAlgorithmException if the given hash function is not supported on
-     *           the JVM.
      */
-    public static byte[] getHashedPassword(final String hashFunction, final byte[] salt, final String password) throws NoSuchAlgorithmException {
 
-        final MessageDigest digest = MessageDigest.getInstance(hashFunction);
+    public static String getHashedPassword(final String hashFunction, final byte[] salt, final String password) {
+        //TODO Delegating encoder support encoding only using default encoder. Maybe we can enhace the interface to support more hash functions
+        final PasswordEncoder encoder = new MessageDigestPasswordEncoder(hashFunction);
```
sophokles73: why aren't you using the `PasswordEncoderFactory` that you have created in the same package?

core/src/main/java/org/eclipse/hono/util/CredentialsObject.java
```diff
@@ -369,17 +367,16 @@ public static JsonObject emptySecret(final Instant notBefore, final Instant notA
...
+        String hashedPassword = password;
```
sophokles73: this is not the hashed password, is it? It is the plain text password, isn't it?

service-base/src/main/java/org/eclipse/hono/service/auth/device/UsernamePasswordCredentials.java
```diff
@@ -114,27 +115,31 @@ public final String getPassword() {
     @Override
     public boolean matchesCredentials(final JsonObject candidateSecret) {
 
+        encoder = PasswordEncoderFactory.createDelegatingPasswordEncoder();
```
sophokles73: I would hope that we can re-use the same instance for all invocations, i.e. that the `PasswordEncoders` are thread safe. Can you verify?

service-base/src/main/java/org/eclipse/hono/service/auth/device/UsernamePasswordCredentials.java
```diff
@@ -114,27 +115,31 @@ public final String getPassword() {
...
                 LOG.debug("candidate hashed-password secret does not contain a pwd hash");
                 return false;
             }
 
-            final byte[] hashedPasswordOnRecord = Base64.getDecoder().decode(pwdHash);
+            final String hashFunction = candidateSecret.getString(
+                    CredentialsConstants.FIELD_SECRETS_HASH_FUNCTION,
+                    CredentialsConstants.DEFAULT_HASH_FUNCTION);
+
+            String encodedPassword = "{" + hashFunction + "}";
```
sophokles73: I think in this case it would be advisable to use a `StringBuilder` instead of concatenating strings.

service-base/src/main/java/org/eclipse/hono/service/auth/device/UsernamePasswordCredentials.java
```diff
@@ -114,27 +115,31 @@ public final String getPassword() {
...
 
-            byte[] salt = null;
             final String encodedSalt = candidateSecret.getString(CredentialsConstants.FIELD_SECRETS_SALT);
             // the salt is optional so decodedSalt may stay null if salt was not found
             if (encodedSalt != null) {
-                salt = Base64.getDecoder().decode(encodedSalt);
+                final byte[] salt = Base64.getDecoder().decode(encodedSalt);
+                encodedPassword += "{" + new String(salt, StandardCharsets.UTF_8) + "}";
```
sophokles73: I am not sure if this works for all hashed passwords that have been created with a byte array that does not represent valid UTF_8 encoded characters. When creating String from such a byte array, you will probably run into an exception or some bytes will be replaced by *default* UTF8 chars. However, this is a migration issue ...

service-base/src/test/java/org/eclipse/hono/service/auth/device/UsernamePasswordCredentialsTest.java
```diff
@@ -122,13 +116,30 @@ public void testMatchesCredentialsSucceedsForMatchingPassword() throws NoSuchAlg
...
 
+    /**
+     * Verifies that credentials support Bcrypt hash function.
+     */
+    @Test
+    public void testMatchesCredentialsSucceedsForMatchingBCryptPassword() {
+
+        // GIVEN a secret on record that uses bcrypt as the hash function
+        final JsonObject candidateSecret = CredentialsObject.emptySecret(null, null);
+        candidateSecret.put(CredentialsConstants.FIELD_SECRETS_PWD_HASH, "$2a$12$BjLeC/gqcnEyk.XNo2qorul.a/v4HDuOUlfmojdSZXRSFTjymPdVm");
```
sophokles73: Why don't you use `CredentialsObject.hashedPasswordSecret()` for creating the candidate secret?

sophokles73:
you should also rebase to HEAD of current master in order to get stable integration tests ...

service-base/src/test/java/org/eclipse/hono/service/auth/device/UsernamePasswordCredentialsTest.java
dejanb: I wanted to test it against default hash stored in Kapua, just to avoid any surprises later on.

service-base/src/main/java/org/eclipse/hono/service/auth/device/UsernamePasswordCredentials.java
dejanb: I found nothing that indicate this. This credentials objects are not reusable and encoders are not heavy-stated, so I think it's not too big of an issue. We can add that to the list of future improvements in this area.

core/src/main/java/org/eclipse/hono/util/MessageDigestPasswordEncoder.java
```diff
@@ -0,0 +1,89 @@
...
+            messageDigest.update(password.salt);
+        }
+        final byte[] hashedPassword = messageDigest.digest(rawPassword.toString().getBytes(StandardCharsets.UTF_8));
+        return Arrays.equals(hashedPassword, Base64.getDecoder().decode(password.password));
+    }
+
+    /**
+     * Value object class for parsing passwords in {salt}password format.
+     */
+    final class EncodedPassword {
```
sophokles73: IMHO we should refactor this into a standalone class and expose methods to do the un-/packaging of the salt bytes and password. This way we would not need to implement it in multiple different places (as we currently do).

dejanb: @sophokles73 It's refactored now. Give it a look and let me know if you see anything else missing.

core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
```diff
@@ -0,0 +1,114 @@
...
+ * Helper class for parsing and encoding passwords.
+ * <p>
+ * The encoded password have three main elements:
+ * <ul>
+ *     <li><em>Hash function</em> that are used to encode the password</li>
+ *     <li><em>Salt</em> which is optionally used for encoding</li>
+ *     <li><em>Password</em> we want to encode</li>
+ * </ul>
+ * <p>
+ * Encoded password is returned in the {salt}password format. It can also be represented in {hashFunction}{salt}password format.
```
sophokles73: which one is it? the former or the latter?

core/src/main/java/org/eclipse/hono/util/PasswordEncoderFactory.java
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @return delegating password encoder
+     */
+    public static PasswordEncoder createDelegatingPasswordEncoder() {
+        final String encodingId = "sha-256";
+        final Map<String, PasswordEncoder> encoders = new HashMap<String, PasswordEncoder>();
+        encoders.put(encodingId, new MessageDigestPasswordEncoder("SHA-256"));
+        encoders.put("ldap", new LdapShaPasswordEncoder());
+        encoders.put("MD4", new Md4PasswordEncoder());
+        encoders.put("MD5", new MessageDigestPasswordEncoder("MD5"));
```
sophokles73: IMHO we should not support MD4 and MD5, they are known to be completely unsafe, aren't they?

core/src/main/java/org/eclipse/hono/util/PasswordEncoderFactory.java
```diff
@@ -0,0 +1,55 @@
...
+        encoders.put("noop", NoOpPasswordEncoder.getInstance());
+        encoders.put("pbkdf2", new Pbkdf2PasswordEncoder());
+        encoders.put("scrypt", new SCryptPasswordEncoder());
+        encoders.put("bcrypt", new BCryptPasswordEncoder());
+        encoders.put("sha-512", new MessageDigestPasswordEncoder("SHA-512"));
+        encoders.put("sha256", new StandardPasswordEncoder());
```
sophokles73: we can also omit this one, can't we? it seems quite confusing to distinguish between *sha256* and *sha-256* ...

core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+    }
+
+    /**
+     * Encode the password based on the provided hash function and salt.
+     *
+     * @return Encoded password
+     */
+    String encode() {
```
sophokles73: why is this method here? isn't this supposed to be the job of `PasswordEncoder`?

dejanb: PasswordEncoder.encode() expects password in {salt}password format, so this helper function formats it. It also finds an appropriate encoder.

dejanb:
@sophokles73 I refactored EncodedPassword class a bit according to your suggestions. Let me know how it looks to you now.

core/src/main/java/org/eclipse/hono/util/CredentialsObject.java
```diff
@@ -370,17 +367,11 @@ public static JsonObject emptySecret(final Instant notBefore, final Instant notA
...
      */
-    public static byte[] getHashedPassword(final String hashFunction, final byte[] salt, final String password) throws NoSuchAlgorithmException {
 
-        final MessageDigest digest = MessageDigest.getInstance(hashFunction);
-        if (salt != null) {
-            digest.update(salt);
-        }
-        digest.update(password.getBytes(StandardCharsets.UTF_8));
-        return digest.digest();
+    public static String getHashedPassword(final String hashFunction, final byte[] salt, final String password) {
```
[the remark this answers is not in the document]
dejanb: It fits better with the PasswordEncoder interface and it has been converted back to String in most places anyway. So, I'm inclined to leave it changed.

core/src/main/java/org/eclipse/hono/util/CredentialsObject.java
```diff
@@ -361,26 +358,15 @@ public static JsonObject emptySecret(final Instant notBefore, final Instant notA
...
-    public static byte[] getHashedPassword(final String hashFunction, final byte[] salt, final String password) throws NoSuchAlgorithmException {
-
-        final MessageDigest digest = MessageDigest.getInstance(hashFunction);
-        if (salt != null) {
-            digest.update(salt);
-        }
-        digest.update(password.getBytes(StandardCharsets.UTF_8));
-        return digest.digest();
+    public static String getHashedPassword(final String hashFunction, final byte[] salt, final String password) {
+        return EncodedPassword.encode(hashFunction, salt, password);
```
sophokles73: This method's purpose has been to create a password hash as defined by the [Credentials API](https://www.eclipse.org/hono/api/credentials-api/#hashed-password). Now it seems to create a String of undefined format (neither this method's nor EncodedPassword.encode's JavaDoc even mention the produced format).

I still do not understand what the purpose of this change is, so please either elaborate or revert the change.

core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
```diff
@@ -0,0 +1,170 @@
...
+    byte[] salt;
+    String rawPassword;
+    String passwordHash;
+    String hashFunction =  CredentialsConstants.DEFAULT_HASH_FUNCTION;
+
+    /**
+     * Creates a new instance from Hono Secret.
+     *
+     * The secret is expected to be of type <em>hashed-password</em> as defined by
+     * <a href="https://www.eclipse.org/hono/api/credentials-api/">Hono's Credentials API</a>.
```
sophokles73: https://www.eclipse.org/hono/api/credentials-api/#hashed-password

core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
```diff
@@ -0,0 +1,170 @@
...
+    public String format() {
+        final StringBuilder result = new StringBuilder(PREFIX).append(hashFunction).append(SUFFIX);
+        append(salt, result);
+        result.append(passwordHash);
+
+        return result.toString();
+    }
+
+    /**
+     * Helper function that appends value (if not {@code null}) to the provided String builder, using "{value}" format.
```
sophokles73: it actually appends the Base64 encoding of the given value ...

core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
```diff
@@ -0,0 +1,170 @@
...
+        //TODO Delegating encoder support encoding only using default encoder. Maybe we can enhance the interface to support more hash functions
+        final PasswordEncoder encoder = new MessageDigestPasswordEncoder(hashFunction);
+        final StringBuilder passwordToEncode = new StringBuilder();
+        append(salt, passwordToEncode);
+        passwordToEncode.append(password);
+        return encoder.encode(passwordToEncode.toString());
+    }
+
+    /**
+     * Checks is provided password matches stored secret.
```
sophokles73: this is the clear text password as given by a client, right?

core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Encode the password based on the provided hash function and salt.
+     *
+     * @param hashFunction Hash function used to encode this password
+     * @param salt Salt in the form of byte array
+     * @param password Password to be encoded
+     * @return Encoded password as a Base64 formatted String
+     */
+    public static String encode(final String hashFunction, final byte[] salt, final String password) {
```
sophokles73: I have no clue what the result of this method is. Please document what the returned string looks like and how it is is created, otherwise nobody (including me) will ever understand what this method does...

core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
```diff
@@ -0,0 +1,170 @@
...
+        }
+        return encodedPassword;
+    }
+
+    /**
+     * Creates a new instance from the String in {salt}password format. The password can be in raw or hashed form.
+     * This method is used by Spring Security password encoders, to parse submitted values.
+     *
+     * @param formattedPassword Password in the {salt}password format
+     * @param raw {@true} if password is in raw form, {@false} if it's hashed
```
sophokles73: I do not understand what *raw* vs *hashed* means. Examples in the JavaDoc would *really* be helpful to understand what this method actually does. Apart from that I strongly believe that using a parameter to modify how another parameter needs to be interpreted is very bad practice and makes it very hard to both document and understand the method. I would therefore suggest to split this up into two dedicated methods whose purpose can then be easily described. WDYT?

core/src/main/java/org/eclipse/hono/util/CredentialsObject.java
dejanb: I don't understand what the problem is. The method used to return byte array that we manually encoded to base64

Base64.getEncoder().encodeToString(getHashedPassword(hashAlgorithm, salt, password)))

Now it returns Base64 encoded String. Nothing else changed.

core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
dejanb: I added some more description, but I don't now what else I can write (it's there in the "@return" description). A simple Base64 encoded password hash.

core/src/main/java/org/eclipse/hono/util/CredentialsObject.java
sophokles73: The method originally was used to create a password hash as it would be returned from the Credentials service. This is obviously no longer the case, so I wonder why we no longer need to perform tests based on the hashed password format that the Credentials API defines but instead do tests based on a hash format that has no relevance within Hono.

core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
sophokles73: The point is that there is no such thing as *a simple Base64 encoded password hash* because the result mainly depends on the input bytes that you apply the hash function to. So, if I do not know what how the hash has been computed, I have no clue what I am supposed to do with the result returned by this method.

core/src/main/java/org/eclipse/hono/util/MessageDigestPasswordEncoder.java
```diff
@@ -0,0 +1,64 @@
...
+    public MessageDigestPasswordEncoder(final String hashFunction) {
+        try {
+            messageDigest = MessageDigest.getInstance(hashFunction);
+        } catch (NoSuchAlgorithmException e) {
+            throw new IllegalArgumentException("No such hash function: " + e);
+        }
+    }
+
+    @Override
+    public String encode(final CharSequence rawPassword) {
```
sophokles73: IMHO we should also add a JavaDoc to this method which describes that the returned hash is created as defined by Hono's Credentials API.

core/src/main/java/org/eclipse/hono/util/MessageDigestPasswordEncoder.java
```diff
@@ -0,0 +1,64 @@
...
+
+        if (password.salt != null) {
+            messageDigest.update(password.salt);
+        }
+
+        return Base64.getEncoder().encodeToString(messageDigest.digest(password.password.getBytes(StandardCharsets.UTF_8)));
+    }
+
+    @Override
+    public boolean matches(final CharSequence rawPassword, final String encodedPassword) {
```
sophokles73: a JavaDoc comment which explains which is which and i what format it is expected to be would be very, very helpful ...

core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
```diff
@@ -0,0 +1,166 @@
...
+        encodedPassword.hashFunction = hashFunction;
+        encodedPassword.password = pwdHash;
+        if (encodedSalt != null) {
+            encodedPassword.salt = Base64.getDecoder().decode(encodedSalt);
+        }
+        return encodedPassword;
+    }
+
+    /**
+     * Creates a new instance from the String in {salt}password format. The password can be in raw or hashed form.
```
sophokles73: Does *raw* mean *clear text*? If so, how can it not matter if the password is either the *clear text* password or the password hash?

core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
```diff
@@ -0,0 +1,166 @@
...
+        final PasswordEncoder encoder = new MessageDigestPasswordEncoder(hashFunction);
+        // Prepare password in the "{salt}password" format as that's expected by PasswordEncoder class
+        final StringBuilder passwordToEncode = new StringBuilder();
+        append(salt, passwordToEncode);
+        passwordToEncode.append(password);
+        return encoder.encode(passwordToEncode.toString());
+    }
+
+    /**
+     * Checks is provided password (in clear text form) matches stored secret.
```
sophokles73: What about
> Checks if a given clear text password is the same as this instance's encoded password.

core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
```diff
@@ -0,0 +1,166 @@
...
+ *     <li><em>Password</em> we want to encode in plain text form or hashed value which is used for matching</li>
+ * </ul>
+ * <p>
+ */
+public class EncodedPassword {
+    private static final String PREFIX = "{";
+    private static final String SUFFIX = "}";
+
+    byte[] salt;
+    String password;
```
sophokles73: is this the *clear text* password or is this the password hash?

core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
```diff
@@ -0,0 +1,91 @@
...
+ *
+ *  You can also use it to match passwords
+ *
+ * password.matches("ThePassword");
+ *
+ */
+public class EncodedPassword extends BasePassword {
+
+    /**
+     * Creates an instance from the {Base64(salt)}password formatted String.
```
sophokles73: `{Base64(salt)}password-hash`, isn't it?

core/src/main/java/org/eclipse/hono/util/EncodedPassword.java
```diff
@@ -0,0 +1,91 @@
...
+        encodedPassword.hashFunction = hashFunction;
+        encodedPassword.password = pwdHash;
+        if (encodedSalt != null) {
+            encodedPassword.salt = Base64.getDecoder().decode(encodedSalt);
+        }
+        return encodedPassword;
+    }
+
+    /**
+     * Checks is provided password (in clear text form) matches this encoded password.
```
sophokles73: `Checks if ...`

core/src/main/java/org/eclipse/hono/util/MessageDigestPasswordEncoder.java
```diff
@@ -0,0 +1,91 @@
...
+     */
+    public MessageDigestPasswordEncoder(final String hashFunction) {
+        try {
+            messageDigest = MessageDigest.getInstance(hashFunction);
+        } catch (NoSuchAlgorithmException e) {
+            throw new IllegalArgumentException("No such hash function: " + e);
+        }
+    }
+
+    @Override
```
sophokles73: this should probably go below the JavaDoc ...

Issue #691, opened 2018-07-04
Optimized command link handling at adapter side


pellmann:
- reuse links for receiver and sender
- handle multiple senders per device at adapter instance

pellmann:
Fixed with https://github.com/eclipse/hono/commit/ebf7193cd708710c459c0345537bc264274d0df2

PR #692, opened 2018-07-04
[#691] Optimized command link handling at adapter side.

pellmann:
- reuse links for receiver and sender
- handle multiple senders per device at adapter instance

As mentioned by @sysexcontrol this is the preparation for the MQTT support of Command and Control as a separate RP, since this should also be used by HTTP etc.


pellmann:
I integrated your suggestions.

PR #698, opened 2018-07-09
Compile netty-tcnative with a provided OpenSSL version in S2I

ctron:
This change enables netty-tcnative and compiles the JNI layer so that
it matches the container image.


adapters/mqtt-vertx/pom.xml
```diff
@@ -21,6 +21,10 @@
       <artifactId>hono-adapter-mqtt-vertx-base</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-tcnative</artifactId>
```
sophokles73: it should be sufficient to add this to `adapters/pom.xml` instead of each adapter individually ...

services/messaging/pom.xml
```diff
@@ -29,6 +29,10 @@
       <groupId>org.eclipse.hono</groupId>
       <artifactId>hono-service-base</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-tcnative</artifactId>
```
sophokles73: it should be sufficient to add this to `services/pom.xml` instead of each service individually ...

adapters/mqtt-vertx/pom.xml
ctron: Fine with me. My intention was to not force it on every derived project. I will update the PR.

ctron:
I made the changes and switched to the most recent version of `netty-tcnative` (2.0.12.Final) and will file a CQ now for this component.

ctron:
It looks like BoringSSL is broken starting with version 2.0.11.Final … So I will switch back to version 2.0.10.Final … which the be the last working version with BoringSSL.

ctron:
The CQ for this has been approved. -> http://dev.eclipse.org/ipzilla/show_bug.cgi?id=16909

I will re-base and merge this PR.

tests/pom.xml
```diff
@@ -463,20 +468,28 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
...
                     <assembly>
                       <mode>dir</mode>
-                      <basedir>/etc/hono</basedir>
+                      <basedir>/</basedir>
                       <inline>
                         <id>config</id>
+                        <dependencySets>
+                          <dependencySet>
+                            <includes>
+                              <include>io.netty:netty-tcnative-boringssl-static</include>
```
sophokles73: did you create a CQ for this test dependency?

ctron: Irgs … no, I did not. That is a works-with anyway, as it is build & test only. But that will most likely be too late for this release.

I will try to get a CQ for that ASAP, if it doesn't work out, we can always remove that unit test and re-add it a later time.

sophokles73: ok

Issue #708, opened 2018-07-13
Add the ability to specify an arbitrary "reply-to" address for command replies


thjaeckle:
Currently the reply-to address for sent commands is in the following format:
```
control/TENANT/4711/${replyId}
```
The `${replyId}` is created by the Hono-Client (or AMQP-client).

For Eclipse Ditto it would be really helpful if when sending a command the whole "reply-to" address could be specified, or at least the part after `control/TENANT`.

That way we could pass in for each command a common "catch all" address like for example:
```
control/TENANT/replies
```
Ditto would only have to attach to this address in order to get all command replies.

Background:
Ditto's connectivity is clustered among several cluster nodes. When a failover happens (e.g. because one node crashes, a Hono connection which was established on this node will be opened on a still "living" cluster node. The new node however can have no knowledge of the `${replyId}`s which were created on the other cluster node which crashed.

Would that be possible?
Feels like if the client already specifies the `${replyId}` it could also specify the whole address.

sysexcontrol:
@thjaeckle : 
To complete the picture, there are currently already two variants:
- the Hono client generates a `replyId` and a receiver link is opened for `control/TENANT/4711/${replyd}`
- the application provides it's own `replyId`, that is appended to `control/TENANT/4711`

Probably it makes sense to change variant 2 to
- the application provides it's own `replyId`, that is appended to `control/TENANT`. To skip the tenant and let the application define this part as well seems to be more problematic, since the authentication to the AMQP network is on the level of the tenant. But this should be working well for Ditto, shouldn't it?

The `old` variant 2 would be substituted by the new one then, so there wouldn't be three variants.

thjaeckle:
> But this should be working well for Ditto, shouldn't it?
Yes, it should. No need to skip the `TENANT`

pellmann:
This would be a change in our API for the reply-to and the sender link address used by the adapter. But I think it makes sense to change it this way, as long as we stay with `control/tenant/` for authorization.

Before

`control/${tenant_id}/${device_id}/${reply-to}`

After

`control/${tenant_id}/${reply-to}`

and our Hono client by default will set  `${reply-to} ` as `${device-id}/something`

sophokles73:
@pellmann @thjaeckle can this be closed?

pellmann:
@thjaeckle - I think you should take look if this works for you.

thjaeckle:
According to the adjusted Command-And-Control-API.md it will work for us.

[...]

PR #715, opened 2018-07-16
[#708] Allow command responses reply-to without device id.

pellmann:
- authorization to tenant is still there
- information 'device-id' encoded in 'reply-id' by 1/0 at start


sophokles73:
@pellmann  some unit tests are failing ...

pellmann:
@sophokles73 - ups.. fixed that.

service-base/src/main/java/org/eclipse/hono/service/command/Command.java
```diff
@@ -260,11 +260,14 @@ public String getCorrelationId() {
...
-
+    static String getRequestId(final String correlationId, final String replyToId, final String deviceId) {
         final String stringOne = Optional.ofNullable(correlationId).orElse("");
-        final String stringTwo = Optional.ofNullable(replyToId).orElse("");
-        return String.format("%02x%s%s", stringOne.length(), stringOne, stringTwo);
+        String stringTwo = Optional.ofNullable(replyToId).orElse("");
+        final boolean removeDeviceFromReplyTo = stringTwo.startsWith(deviceId + "/");
+        if(removeDeviceFromReplyTo) {
+            stringTwo = stringTwo.substring(deviceId.length()+1);
+        }
```
sophokles73: missing whitespaces

service-base/src/main/java/org/eclipse/hono/service/command/Command.java
```diff
@@ -123,7 +123,7 @@ public static Command from(final ProtonDelivery delivery, final Message message,
                         Objects.requireNonNull(tenantId),
                         Objects.requireNonNull(deviceId),
                         correlationId,
-                        replyTo.getResourcePath()[3]);
+                        replyTo.toString().substring(replyTo.getBasePath().length()+1));
```
sophokles73: I would rather add a new method (e.g. *asString(int startIdx)*) to `ResourceIdentifier`for this purpose. Or maybe even add something more specific like *getReplyToId()*. WDYT?

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -485,38 +485,33 @@ protected void doStop(final Future<Void> stopFuture) {
...
             final String tenantId,
-            final String deviceId,
             final CommandResponse response) {
 
         Objects.requireNonNull(tenantId);
-        Objects.requireNonNull(deviceId);
         Objects.requireNonNull(response);
 
-        return createCommandResponseSender(tenantId, deviceId, response.getReplyToId())
+        return createCommandResponseSender(tenantId, response.getReplyToId())
```
sophokles73: wouldn't it make sense to add a *sendCommandResponse(CommandResponse response)* method to `CommandResponseSender`?

service-base/src/main/java/org/eclipse/hono/service/command/Command.java
pellmann: From the `ResourceIdentifier`s POV I would suggest `getWithoutBasePath()`?

pellmann: Or `getPathWithoutBase()`?

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
pellmann: Since only the tenantId and replyId are needed for this I think it is clearer to stay with this signature -also only the CommandResponse would not offer the tenantId.

core/src/main/java/org/eclipse/hono/util/ResourceIdentifier.java
```diff
@@ -230,13 +230,25 @@ public String toString() {
...
-     * 
+     *
      * @return A string consisting of the properties separated by a forward slash.
      */
     public String getBasePath() {
         return basePath;
     }
 
+    /**
+     * Gets a string representation of ths resource identifiers's parts without the base path {@see getBasePath()}.
```
sophokles73: typo, should read `identifiers'`. Also, we usually put an empty line between the description and the tags. `@see` should also go down to the tags section FMPOV.

core/src/main/java/org/eclipse/hono/util/ResourceIdentifier.java
```diff
@@ -230,13 +230,25 @@ public String toString() {
...
+     * @return A string with all parts after the base bath.
+     */
+    public String getPathWithoutBase() {
+        if (resource.length() > basePath.length()) {
```
sophokles73: `return createStringRepresentation(2);` should do the trick ...

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
sophokles73: ??? I am talking about *CommandResponseSender.sendCommandResponse()*

pellmann: Ah, I see - yes, that makes sense.

site/content/api/Command-And-Control-API.md
```diff
@@ -26,7 +26,9 @@ The following API defines operations that can be used by *Business Applications*
...
+3. *Business Application* has established an AMQP link in role *receiver* with the source address `control/${tenant_id}/${reply-id}`. Where `${reply-id}` may be any arbitrary string chosen by the client. This link is used by the client to receive the response of the executed command. This link’s source address is also referred to as the reply-to address for the request messages. 
+
+In cases where the response should be scoped for a device (one response link per device) we suggest to add the device-id as the first part of the reply-id (`control/${tenant_id}/${device_id}/..`). This allows the adapters to minimize the information needed correlate the requests of the device with the responses, since this information (`${device_id}`) is already known and does not need to be encoded in such an request-id.
```
sophokles73: can we rephrase this to not use fourth person (`we suggest`)? we are not using this style anywhere in the API specs.
IMHO we also should not hint in the spec at the optimization that an adapter might or might not perform. However, it would be nice to have an example of a tenant scoped *reply-id* vs. a device scoped *reply-id* ...

site/diagrams/sequenceDiagrams/command_control_Success.txt
```diff
@@ -2,9 +2,9 @@ participant AMQP 1.0 Network
 participant Business Application
 
 Business Application->AMQP 1.0 Network: attach("sender","control/${tenant_id}/${device_id}")
-Business Application->AMQP 1.0 Network: attach("receiver","control/${tenant_id}/${device_id}/${reply-to}")
+Business Application->AMQP 1.0 Network: attach("receiver","control/${tenant_id}/${reply-to}")
```
sophokles73: what about the PNG generated from this?

sophokles73:
@pellmann I also think that this deserves mentioning in the release notes ...

site/diagrams/sequenceDiagrams/command_control_Success.txt
pellmann: Good question. I changed it, but now it is gone. Thanks for spotting!

Issue #716, opened 2018-07-17
NPE in error handler


ctron:
While testing I saw the following exception:

~~~
09:19:16.238 [vert.x-eventloop-thread-0] ERROR i.v.e.w.impl.RoutingContextImplBase - Unexpected exception in route
java.lang.NullPointerException: null
	at io.vertx.core.buffer.impl.BufferImpl.<init>(BufferImpl.java:52)
	at io.vertx.core.buffer.impl.BufferImpl.<init>(BufferImpl.java:56)
	at io.vertx.core.buffer.impl.BufferFactoryImpl.buffer(BufferFactoryImpl.java:35)
	at io.vertx.core.buffer.Buffer.buffer(Buffer.java:70)
	at io.vertx.core.http.impl.HttpServerResponseImpl.end(HttpServerResponseImpl.java:308)
	at org.eclipse.hono.service.http.DefaultFailureHandler.sendError(DefaultFailureHandler.java:84)
	at org.eclipse.hono.service.http.DefaultFailureHandler.handle(DefaultFailureHandler.java:65)
	at org.eclipse.hono.service.http.DefaultFailureHandler.handle(DefaultFailureHandler.java:40)
	at io.vertx.ext.web.impl.RouteImpl.handleFailure(RouteImpl.java:223)
	at io.vertx.ext.web.impl.RoutingContextImplBase.iterateNext(RoutingContextImplBase.java:117)
	at io.vertx.ext.web.impl.RoutingContextImpl.next(RoutingContextImpl.java:133)
	at io.opentracing.contrib.vertx.ext.web.TracingHandler.handlerFailure(TracingHandler.java:90)
	at io.opentracing.contrib.vertx.ext.web.TracingHandler.handle(TracingHandler.java:44)
	at io.opentracing.contrib.vertx.ext.web.TracingHandler.handle(TracingHandler.java:25)
	at io.vertx.ext.web.impl.RouteImpl.handleFailure(RouteImpl.java:223)
	at io.vertx.ext.web.impl.RoutingContextImplBase.iterateNext(RoutingContextImplBase.java:117)
	at io.vertx.ext.web.impl.RoutingContextImpl.next(RoutingContextImpl.java:133)
	at io.vertx.ext.web.impl.RoutingContextImpl.doFail(RoutingContextImpl.java:465)
	at io.vertx.ext.web.impl.RoutingContextImpl.fail(RoutingContextImpl.java:166)
[...]

PR #717, opened 2018-07-17
[#716] Fix possible NPE when no message is present

ctron:


service-base/src/main/java/org/eclipse/hono/service/http/DefaultFailureHandler.java
```diff
@@ -39,6 +39,7 @@
  */
 public class DefaultFailureHandler implements Handler<RoutingContext> {
 
+    private static final String MSG_NO_ERROR = "<no error>";
```
sophokles73: I would rather send an empty string in this case. Getting a response with an error code that contains a body saying "no error" would be quite confusing, wouldn't it?

service-base/src/main/java/org/eclipse/hono/service/http/DefaultFailureHandler.java
```diff
@@ -75,6 +76,32 @@ public void handle(final RoutingContext ctx) {
...
     }
 
+    private String extractMessage(final HttpServerResponse response) {
+        if ( response == null ) {
+            return MSG_NO_ERROR;
+        }
+
+        final String message = response.getStatusMessage();
+        if ( message == null ) {
+            return Integer.toString(response.getStatusCode());
```
sophokles73: this information is already given by the response's status code. I would rather use the empty string here again (i.e. an empty body)

service-base/src/main/java/org/eclipse/hono/service/http/DefaultFailureHandler.java
```diff
@@ -75,6 +76,32 @@ public void handle(final RoutingContext ctx) {
...
+    }
+
+    private String extractMessage(final Throwable cause) {
+        if (cause == null) {
+            return MSG_NO_ERROR;
+        }
+
+        final String message = cause.getMessage();
+        if (message == null) {
+            return cause.getClass().getName();
```
sophokles73: I would rather not expose the exception class name here in order not to encourage people to actually rely on the class names...

ctron: I had this at first ... the problem with that seems to me that you would have no clue at all what went wrong. IMHO known the class name of an Exception doesn't really help you get any insight at all.

service-base/src/main/java/org/eclipse/hono/service/http/DefaultFailureHandler.java
ctron: ok

service-base/src/main/java/org/eclipse/hono/service/http/DefaultFailureHandler.java
ctron: Maybe "unknown error" is better. But having nothing at all again makes it hard to figure out what went wrong. Having some kind of string helps you find the right spot in the source code easily.

service-base/src/main/java/org/eclipse/hono/service/http/DefaultFailureHandler.java
sophokles73: Which view are you taking? A Hono developer? A user of a hosted Hono instance?

service-base/src/main/java/org/eclipse/hono/service/http/DefaultFailureHandler.java
sophokles73: How does this help?

service-base/src/main/java/org/eclipse/hono/service/http/DefaultFailureHandler.java
```diff
@@ -39,6 +39,7 @@
  */
 public class DefaultFailureHandler implements Handler<RoutingContext> {
 
+    private static final String MSG_UNKNOWN_ERROR = "<unknown error>";
```
sophokles73: still, IMHO this is not helpful at all and I would rather have an empty body in this case

service-base/src/main/java/org/eclipse/hono/service/http/DefaultFailureHandler.java
```diff
@@ -75,6 +76,32 @@ public void handle(final RoutingContext ctx) {
         }
     }
 
+    private String extractMessage(final HttpServerResponse response) {
+        if ( response == null ) {
+            return MSG_UNKNOWN_ERROR;
```
sophokles73: this code is unreachable, we will have run into an NPE in line 56 before we arrive here ...

service-base/src/main/java/org/eclipse/hono/service/http/DefaultFailureHandler.java
sophokles73: A *user* shouldn't see the internal class names. A *developer* can take a look at the (server side) logs.

PR #721, opened 2018-07-18
JMeter plugin: Improve sender log output

calohmn:
Improve log output by adding info about remaining credit.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/MessageConsumer.java
+++ b/client/src/main/java/org/eclipse/hono/client/MessageConsumer.java
@@ -42,23 +42,13 @@ public interface MessageConsumer {
     void flow(int credits) throws IllegalStateException;
 
     /**
-     * Retrieves the current amount of credit.
-     * <p>
-     * The value returned will still include the credits that will be used by any queued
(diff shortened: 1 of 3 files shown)


client/src/main/java/org/eclipse/hono/client/MessageConsumer.java
```diff
@@ -40,4 +40,25 @@
...
+     * Retrieves the current amount of credit.
+     * <p>
+     * The value returned will still include the credits that will be used by any queued
+     * incoming messages, use {@link #getQueued()} to assess the number of credits that will be used by queued messages.
+     * <p>
+     * Note that the value returned is valid during execution of the current vert.x handler only.
+     *
+     * @return The number of messages.
+     */
+    int getCredit();
```
sophokles73: since this interface is supposed to represent a consumer at a higher level of abstraction than a generic *Link*, I think it would make more sense to provide a *getRemainingCredit()* method only, which returns *getCredit* - *getQueued* without requiring the user to understand the underlying concept of buffering incoming messages etc, which, BTW, is an implementation detail of proton-j ...

calohmn: Ok, I've changed that.

PR #722, opened 2018-07-18
[#712] JMeter Sender: add options 'waitForDeliveryResult', 'messageCountPerSamplerRun'

### LINKED ISSUE #712: JMeter plugin: Add option "wait for delivery result" to sender
The JMeter `HonoSender` currently waits for the delivery result after having sent the message. That means it waits for the `Disposition` frame, usually with the `Accepted` state in it.

There may be cases, where such a synchronous message processing is not wanted and the next message should already be sent before having received the disposition frame. 
With this, more load can be generated.

Therefore there should be an option "wait for delivery result". By default, it should be enabled to produce the current behaviour.

calohmn:
This fixes #712:
Adds the options "Wait for delivery result" and "Number of messages per sampler run" for the JMeter Sender.

### DIFF
--- a/jmeter/src/main/java/org/eclipse/hono/jmeter/HonoSenderSampler.java
+++ b/jmeter/src/main/java/org/eclipse/hono/jmeter/HonoSenderSampler.java
@@ -143,12 +143,7 @@ public class HonoSenderSampler extends HonoSampler implements ThreadListener {
      * @return The registry port as integer, {@code 0} if the value cannot be parsed as an integer.
      */
     public int getRegistryPortAsInt() {
-        final String portString = getRegistryPort();
-        try {
-            return Integer.parseInt(portString);
-        } catch (final NumberFormatException e) {
-            return 0;
-        }
+        return getIntValueOrDefault(getRegistryPort(), 0);
     }
 
     public String getRegistryPort() {
@@ -171,12 +166,7 @@ public class HonoSenderSampler extends HonoSampler implements ThreadListener {
      * if the value cannot be parsed as integer.
      */
     public int getWaitForReceiversAsInt() {
-        final String value = getPropertyAsString(WAIT_FOR_RECEIVERS);
-        try {
-            return Integer.parseInt(value);
-        } catch (final NumberFormatException e) {
-            return 0;
-        }
+        return getIntValueOrDefault(getPropertyAsString(WAIT_FOR_RECEIVERS), 0);
     }
 
     public String getWaitForReceivers() {
@@ -199,12 +189,7 @@ public class HonoSenderSampler extends HonoSampler implements ThreadListener {
      * if the value cannot be parsed as integer.
      */
     public int getWaitForReceiversTimeoutAsInt() {
-        final String value = getPropertyAsString(WAIT_FOR_RECEIVERS_TIMEOUT);
-        try {
-            return Integer.parseInt(value);
-        } catch (final NumberFormatException e) {
-            return 0;
-        }
+        return getIntValueOrDefault(getPropertyAsString(WAIT_FOR_RECEIVERS_TIMEOUT), 0);
     }
 
     public String getWaitForReceiversTimeout() {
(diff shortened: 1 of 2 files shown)


jmeter/src/main/java/org/eclipse/hono/jmeter/HonoSenderSampler.java
```diff
@@ -295,6 +297,64 @@ public void setWaitForCredits(final boolean isWaitForCredits) {
         setProperty(WAIT_FOR_CREDITS, isWaitForCredits);
     }
 
+    /**
+     * Sets whether to wait for the result of sending the message.
```
sophokles73: `Gets` ...

jmeter/src/main/java/org/eclipse/hono/jmeter/HonoSenderSampler.java
```diff
@@ -295,6 +297,64 @@ public void setWaitForCredits(final boolean isWaitForCredits) {
...
+     * <p>
+     * If this option has not been set, {@code true} is returned.
+     * 
+     * @return true in order to wait for the delivery result.
```
sophokles73: can you please use `{@code true}` to format the boolean literal

jmeter/src/main/java/org/eclipse/hono/jmeter/HonoSenderSampler.java
```diff
@@ -295,6 +297,64 @@ public void setWaitForCredits(final boolean isWaitForCredits) {
...
+        }
+    }
+
+    /**
+     * Sets the number of messages to send per sample run.
+     *
+     * @param messageCountPerSamplerRun number of messages.
+     */
+    public void setMessageCountPerSamplerRun(final String messageCountPerSamplerRun) {
+        setProperty(MESSAGE_COUNT_PER_SAMPLER_RUN, messageCountPerSamplerRun);
```
sophokles73: this could use some range check, couldn't it?

jmeter/src/main/java/org/eclipse/hono/jmeter/ui/HonoSenderSamplerUI.java
```diff
@@ -65,7 +67,9 @@ public HonoSenderSamplerUI() {
         assertion = new JLabeledTextField("Registration Assertion");
         contentType = new JLabeledTextField("Content type");
         data = new JLabeledTextArea("Message data");
-        waitForCredits = new JCheckBox("Wait for credits");
+        waitForCredits = new JCheckBox("Wait for credits if none is left");
```
sophokles73: `Wait for credits if none left after sending`?

jmeter/src/main/java/org/eclipse/hono/jmeter/ui/HonoSenderSamplerUI.java
```diff
@@ -65,7 +67,9 @@ public HonoSenderSamplerUI() {
...
+        waitForDeliveryResult = new JCheckBox("Wait for delivery result");
+        waitForDeliveryResult.setToolTipText("<html>Deselecting this option increases throughput, especially of <em>event</em> messages.</html>");
```
sophokles73: however, in particular for events you will usually **not** want to disable this. However, the way it is written, it sounds like *good tuning advice* which is quite misleading FMPOV ...

calohmn:
> FMPOV the approach for increasing throughput by not considering the disposition is not very practical because it requires the user to take a look at the logs in order to determine if any of the messages sent has actually been received, or am I mistaken?

If just looking at the sender side, a look at the logs would be required, yes. But looking at the number of messages at the receiver side and comparing it to the sender message count also reveals any messages that were not accepted.

The main point of adding the option to skip waiting for the disposition is to enable putting a higher load on the tested system with fewer sender threads. 
To simulate a load on the tested system where there are 250 messages in-flight, I would as of now have to use 250 sender threads. I can theoretically achieve the same load when using just one thread and not waiting for the disposition (provided the link capacity is 250).

> I do like the idea of batch sending the messages, however. Why not send multiple messages directly after each other, but keep track of the dispositions (e.g. in a map) and mark the sampler result successful based on the aggregated dispositions? This way you would not introduce too much latency while still getting reasonable results. WDYT?

If I understand correctly you mean sending a batch of say 100 messages, wait for all to be accepted and then send the next batch. I think this would put an uneven kind of load on the target system, because the load is only applied in bursts, then waiting for all to have settled down again (quite literally ;)).

I could only imagine such a solution if there would be just _one_ sample invocation with this (using batch size of 1 million for example). With this, there would be the need for an extra throttling feature within the `Sender` though.

In general, I can understand the wish for including the disposition results in the sample results. It is difficult to combine with the `Sender` sampler invocations though, which serve as the clock for defining the sending rate (including any throttling put in place by a "Constant Throughput Timer" for example) and which are the only place where the disposition result samples can be collected.

Another option would be to let the disposition results be collected via the vertx-threads and have only all _currently available_ results be included after each batch "send" operation. That would cause the first sampler invocation to possibly return an empty result (no dispositions received yet). That's not a problem. But with the last sampler invocation, there would still be some messages in-flight, which would not be counted on the sender side but would still arrive (and be counted) at the receiver side.
This is because there doesn't seem to be a way to give JMeter some `SampleResult`s outside of the `sample()` invocations in the JMeter thread.

calohmn:
@sophokles73 I've committed changes concerning the above points.

PR #723, opened 2018-07-20
Support content type in command messages

sophokles73:
So far, the C&C API doesn't explicitly support the usage of Content-Type. IMHO it is worth doing so, even if not all protocol adapters will be able to forward the type information hence and forth due to limitations of the transport protocols.


pellmann:
We discussed the use of content-type and decided that it is not needed, since the subject (the command) should include all information needed to semantically and syntactically understand and parse the command. This is than also true for the response. Of course, the content-type COULD  be set if really needed, since it is an AMQP standard property.

However I would be fine to define it as an optional property also in the API. But in this case we should implement it also in the client, to send the command and not only in the response. WDYT?

sophokles73:
> However I would be fine to define it as an optional property also in the API. But in this case we should implement it also in the client, to send the command and not only in the response. WDYT?

Absolutely. I stumbled across the fact that the HTTP adapter already considers the content type of the command response submitted by a device in its HTTP request and also sets it on the AMQP response message being forwarded to the application. So FMPOV it would be consistent to also allow for the other direction (sending a command to the device) to use a content type (optionally). And yes, the CommandClient should then also support to set a content type as well.

sophokles73:
@pellmann I have reverted the changes regarding the reply-to-id. However, in order to take advantage of the content-type in command responses, the return type of `CommandClient.sendMessage` had to be changed (last commit).

PR #727, opened 2018-07-22
[#643] AMQP adapter - support only anonymous relay for publishing messages

Alfusainey:
Add check for empty tenant identifier in amqp address for unauthenticated devices. I notice this as I was manually testing using the **qpid-send** command.

@sophokles73


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -63,21 +63,21 @@ ProtonDelivery delivery() {
...
 
     /**
-     * Gets the tenant identifier of this context's resource.
+     * Gets the tenant identifier for this context.
      *
      * @return The tenant identifier.
      */
     String getTenantId() {
-        return resource.getTenantId();
+        return isDeviceAuthenticated() ? authenticatedDevice.getTenantId() : resource.getTenantId();
```
sophokles73: While this will probably yield the intended result, I would rather try to make sure in the receiver open handler that the receiver link's target address actually contains a tenant **and** device if the device is not authenticated. For an authenticated device, IMHO it should also check, if the target address contains either no tenant and device ID at all or a tenant ID that matches the authenticated device (this is relevant for gateways connecting to the adapter).
The receiver open handler seems to suggest, that anonymous relay mode is not supported. However, I think that for gateway mode, it would actually be very helpful to support this because otherwise a gateway would need to create separate links for each device it wants to publish data for.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -63,21 +63,21 @@ ProtonDelivery delivery() {
...
 
     /**
-     * Gets the device identifier of this context's resource.
+     * Gets the device identifier for this context.
      *
      * @return The device identifier.
      */
     String getDeviceId() {
-        return resource.getResourceId();
+        return isDeviceAuthenticated() ? authenticatedDevice.getDeviceId() : resource.getResourceId();
```
sophokles73: For a gateway connecting to the adapter, this will probably not return the intended value. A gateway usually will publish data *on behalf of* another device (of the same tenant), therefore the device ID from the link's target address will usually **not** be the same as the device ID of the authenticated device.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
Alfusainey: > However, I think that for gateway mode, it would actually be very helpful to support this because otherwise a gateway would need to create separate links for each device it wants to publish data for.

But shouldn't the gateway not indicate at least the `endpoint` that it wants to publish data to (either telemetry or event)?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
Alfusainey: humm.. ok, i will have to think about this a little bit

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
sophokles73: what would be the benefit?

Alfusainey: to identify if the gateway wants to publish an event or a telemetry message. or maybe am missing something?

sophokles73: the endpoint, tenant and device ID will be contained in the message's *to* property in that case, i.e. you can determine with each message being received, whether it is a telemetry message or an event ...

Alfusainey: ok, i see now.. thanks! 

i think i will have to find a different CLI client for testing with gateways, since `qpid-send` mandates that the link address be provided via the `-a` option

Alfusainey:
@sophokles73: PR updated with a17fab9. i plan to add 3+ test cases once this looks okay

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -77,6 +86,10 @@ String getTenantId() {
      * @return The device identifier.
      */
     String getDeviceId() {
+        if (receiverTarget == null) {
```
sophokles73: don't we have all necessary information available in the constructor to properly initialize the fields? Why do we need to do additional computations in the accessors? This class is supposed to be immutable, isn't it?
Doing all initializaiton and computation in the constructor would help very much with documenting and understanding the different cases that are possible, don't you think?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -27,12 +28,20 @@
...
+    AmqpContext(final ProtonDelivery delivery, final Message message, final Target receiverTarget, final Device authenticatedDevice) {
         this.delivery = delivery;
         this.message = message;
-        this.resource = resource;
         this.authenticatedDevice = authenticatedDevice;
+        this.receiverTarget = receiverTarget;
+        if (receiverTarget == null) {
+            // link possibly opened by an "authenticated" gateway component via anonymous relay
+            // TODO: validate the address in the message's _to property
+            this.resource = ResourceIdentifier.fromString(message.getProperties().getTo());
```
sophokles73: what if *to* is empty?

Alfusainey: @sophokles73: this is really an amazing feedback, thanks! i will not respond to all the comments since I agree with everything said :-) will update the PR

Alfusainey:
@sophokles73: PR updated

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -510,7 +470,41 @@ private void onLinkDetach(final ProtonReceiver receiver) {
             }
 
         }
-        return result;
+        return result.compose(address -> {
+            return validateAddress(address, authenticatedDevice);
+        });
+    }
+
+    private Future<Void> validateAddress(final ResourceIdentifier address, final Device authenticatedDevice) {
```
sophokles73: this should be *validateTargetAddress*

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -510,7 +470,41 @@ private void onLinkDetach(final ProtonReceiver receiver) {
...
+        final Future<Void> addressCheck = Future.future();
+        if (address == null) {
+            addressCheck.complete();
```
sophokles73: this is only here for the case that the device didn't provide an address at all. IMHO this should be handled in *validateEndpoint* already.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -283,79 +283,32 @@ private void handleSessionOpen(final ProtonConnection conn, final ProtonSession
...
-        if (authenticatedDevice == null && receiver.getRemoteTarget() == null) {
-            // anonymous relay only supported for authenticated clients (gateways in particular)
-            receiver.setCondition(
-                    ProtonHelper.condition(AmqpError.NOT_ALLOWED, "anonymous relay ONLY supported for authenticated clients"));
+        validateEndpoint(receiver, authenticatedDevice).recover(t -> {
+            LOG.debug(
+                    "fail to establish a receiving link with client [{}] due to an invalid endpoint [{}]."
+                            + " closing the link",
+                    receiver.getName(), receiver.getRemoteQoS());
+            receiver.setCondition(AmqpContext.getErrorCondition((ServiceInvocationException) t));
```
sophokles73: how do you know that it is a `ServiceInvocationException`? 
`AmqpContext.getErrorCondition()` should be able to do this check ...

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -194,4 +201,26 @@ private ErrorCondition getErrorCondition(final ServiceInvocationException error)
...
         }
     }
+
+    /**
+     * Validates the address contained in an AMQP 1.0 message property.
+     * 
+     * @param to The property to validate.
+     * @return A future indicating the outcome of the validation operation.
+     */
+    Future<Void> validateMessageProperty(final String to) {
```
sophokles73: this doesn't seem to be a generic method for validating all sorts of properties but instead it looks like it is specifically written to check the message's address, right?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -35,10 +37,15 @@
         this.message = message;
         this.authenticatedDevice = authenticatedDevice;
         this.receiverTarget = receiverTarget;
-        if (receiverTarget == null) {
+        if (receiverTarget == null || receiverTarget.getAddress() == null) {
             // link possibly opened by an "authenticated" gateway component via anonymous relay
-            // TODO: validate the address in the message's _to property
-            this.resource = ResourceIdentifier.fromString(message.getProperties().getTo());
+            validateMessageProperty(message.getProperties().getTo()).compose(ok -> {
+                resource = ResourceIdentifier.fromString(message.getProperties().getTo());
```
sophokles73: `message.getAddress()`

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -35,10 +37,15 @@
...
+                return Future.succeededFuture();
+            }).recover(t -> {
+               handleFailure((ServiceInvocationException) t);
```
sophokles73: can't *handleFailure* make a check if it's a `ServiceInvocationException`?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -184,7 +191,7 @@ boolean isRemotelySettled() {
      * @param error The service invocation error.
      * @return The ErrorCondition.
      */
-    private ErrorCondition getErrorCondition(final ServiceInvocationException error) {
+    static ErrorCondition getErrorCondition(final ServiceInvocationException error) {
```
sophokles73: param should better be a `Throwable` ...

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
sophokles73: this method should not return a Future. The instance created by the constructor should be initialized once the constructor returns, shouldn't it?
why not simply return void and declare a `ServiceInvocationException` being thrown in case of a failing check?

Alfusainey:
@sophokles73: thanks! a02d085 incorporates your latest comments

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -30,22 +29,22 @@
...
-            });
+            try {
+                validateMessageAddress(message.getAddress());
+                this.resource = ResourceIdentifier.fromString(message.getAddress());
+            } catch (ServiceInvocationException sie) {
+                handleFailure(sie);
+            }
+
+        } else if (isDeviceAuthenticated()) {
+            this.resource = ResourceIdentifier.from(receiverTarget.getAddress(), authenticatedDevice.getTenantId(), authenticatedDevice.getDeviceId());
```
sophokles73: is there a reason why the target address can only contain the endpoint name in this case?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -141,16 +136,19 @@ boolean isDeviceAuthenticated() {
...
+    void handleFailure(final Throwable t) {
+        Objects.requireNonNull(t);
+        if (ServiceInvocationException.class.isInstance(t)) {
+            final ServiceInvocationException error = (ServiceInvocationException) t;
+            if (ServerErrorException.class.isInstance(error)) {
+                ProtonHelper.released(delivery, true);
+            } else {
+                final ErrorCondition condition = getErrorCondition(error);
+                MessageHelper.rejected(delivery, condition);
+            }
```
sophokles73: what if it's not a `ServiceInvocationException`? Don't we need to do anything in that case, e.g. reject the message?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -185,42 +183,44 @@ boolean isRemotelySettled() {
...
-     * @param to The property to validate.
+     * @param address The message address to validate.
      * @return A future indicating the outcome of the validation operation.
+     * @throws ServiceInvocationException if the message address is not valid.
      */
-    Future<Void> validateMessageProperty(final String to) {
-        final Future<Void> propertyCheck = Future.future();
-        if (to == null || to.isEmpty()) {
-            propertyCheck.fail(new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST, "Message _to propery cannot be null or empty"));
+    void validateMessageAddress(final String address) throws ServiceInvocationException {
```
sophokles73: is there a reason for this method to be package private instead of private?
I am asking because if this is package private then the JavaDocs are basically useless because they do not give any clue regarding the validation that is actually taking place. In particular, the first null check only makes sense in the context of the target address being `null` as well, right? So, for this method to return a reasonable result, you actually need to have done some checks before as well, right? This would probably be fine, though, if the method was private, so that it cannot be invoked out of the context of this class.

Alfusainey: sorry, this method was intended to be a private method of the class

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
Alfusainey: yes, for authenticated devices(non-gateway), the `endpointName` is specified in the address

Alfusainey:
> It seems to be much easier to follow the pattern that the MQTT adapter uses, i.e. a device connects and is either authenticated or not

Okay, actually this was what I thought we were trying to achieve using the link's target address. for me, the link's target address was treated like a topic name..

> However, we always do the check based on the message's identity and the message address (analogously to the topic name in MQTT)

why the message's identity? shouldn't we only consider the message's address?

> instead of also checking the link's target address

am confused by this statement. you mentioned previously that we should **not** allow the link's target to be set at all, right? so why also check the link's target?

sophokles73:
> why the message's identity? shouldn't we only consider the message's address?

typo, meant *device's identity*

sophokles73:
> Okay, actually this was what I thought we were trying to achieve using the link's target address. for me, the link's target address was treated like a topic name..

That doesn't match the MQTT pattern, though. In MQTT there is no such thing as a link. There is a connection (not scoped to any topic) and there are messages being published to arbitrary topics. So, in order to get closer to the MQTT pattern, we should try to get the link's target address out of the way, agreed?

sophokles73:
> am confused by this statement. you mentioned previously that we should not allow the link's target to be set at all, right? so why also check the link's target?

You are *currently* checking the target address, aren't you? I was suggesting that we should *not* do that anymore, thus, *instead of checking the link's target address* (as we do now), we should *only check the device's identity and the message address* instead.

Alfusainey:
> So, in order to get closer to the MQTT pattern, we should try to get the link's target address out of the way, agreed?

:+1: 

> I was suggesting that we should not do that anymore, thus, instead of checking the link's target address (as we do now), we should only check the device's identity and the message address instead.

Okay, but the `device identity` will make sense only for authenticated devices, right? Also, for authenticated gateway components, the real device-id will be in the message's address, is that correct?

another thing: if the link's target is specified, we simply ignore but not reject a link? or should we reject and close the link when the target address is specified?

sophokles73:
> Okay, but the device identity will make sense only for authenticated devices, right? 

Yes, for an unauthenticated device we take the tenant and device ID from the message address. We have no way of authorizing the message in any way in this case. But this is what the operator of the protocol adapter actually wanted to do: allow every device to publish data on behalf of every device. So be it.

> Also, for authenticated gateway components, the real device-id will be in the message's address, is that correct?

Yes, but we still need to authorize the publishing in the context of the *authenticated device's* identity (which in this case is the gateway). We already do this in the other protocol adapters as well ...

sophokles73:
> if the link's target is specified, we simply ignore but not reject a link? or should we reject and close the link when the target address is specified?

what's *your* opinion on that?

Alfusainey:
> what's your opinion on that?

i think we can ignore and log a message (at info level) that the target address is not the address used for publishing messages. we only reject if the message's address is empty (or null). if both the target and message address is specified, we only pick the message address.. i think this will simplify the check on the adapter side. WDYT?

let's summarize the different cases:

* Authenticated devices:
  * message address: `telemetry` or `event` (address should only contain the name of the endpoint). the `tenant` and `device` IDs will be retrieve from the authenticated device's identity

* Authenticated gateway:
  * message address: `telemetry/DEFAULT_TENANT/4711` or `event/DEFAULT_TENANT/4711`. the authenticated device identifies the gateway component i.e `DEFAULT_TENANT/gw-1`. The AmqpContext#getDeviceId() should return `4711` in this case.

* Unauthenticated components:
  * message address: `telemetry/DEFAULT_TENANT/4711` or `event/DEFAULT_TENANT/4711`. the message address should all three components required to publish the message. 

Is this correct for all three cases?

sophokles73:
> i think we can ignore and log a message (at info level) that the target address is not the address used for publishing messages. we only reject if the message's address is empty (or null). if both the target and message address is specified, we only pick the message address.. i think this will simplify the check on the adapter side. WDYT?

I don't get it. when a device opens a link, there is no message (yet). Do you want to accept (open) a link that contains a target address or reject (close) the link?

Alfusainey:
> Do you want to accept (open) a link that contains a target address or reject (close) the link?

yes, we accept (open) the link even if it **does not** contain a target address. we simply ignore whatever address was set as the target address

Alfusainey:
@sophokles73: maybe we should not completely ignore the link's target address. what am deciding now is this: 

* If the message address is set, we use that and ignore the link's target address. 
* If the message address is not set, we consider the link's target address (this will make the current ITs for the adapter to pass)
* If both are not set, we reject message (at this stage the link will already be opened)

WDYT?

sophokles73:
> yes, we accept (open) the link even if it does not contain a target address. we simply ignore whatever address was set as the target address

FMPOV it should be the other way around: we reject links that contain a target address because accepting such a link might encourage the client (device) to believe that the given target address will have any meaning. I think we should only accept links that do **not** contain a target address.

Alfusainey:
> I think we should only accept links that do not contain a target address.

Ok, thanks! you can safely ignore my previous comment :-)

Alfusainey:
@sophokles73: i just realized that using `HonoClient` will now not be possible for the ITs, since it always create senders with a target address. I think i will have to use `ProtonClient` directly in this case.

this patch might grow a bit larger if it should include those changes. should i create a separate PR for the ITs (and remove the ITs for this patch)?

sophokles73:
> this patch might grow a bit larger if it should include those changes. should i create a separate PR for the ITs (and remove the ITs for this patch)?

yes, that would probably make sense.

Alfusainey:
@sophokles73: PR updated with adc1344 . i still have to add unit tests for the message address for authenticated clients

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -286,64 +286,60 @@ private void handleSessionOpen(final ProtonConnection conn, final ProtonSession
...
-            LOG.debug(
-                    "fail to establish a receiving link with client [{}] due to an invalid endpoint [{}]."
-                            + " closing the link",
-                    receiver.getName(), receiver.getRemoteQoS());
-            receiver.setCondition(AmqpContext.getErrorCondition(t));
+        if (receiver.getRemoteTarget() != null) {
+            if (receiver.getRemoteTarget().getAddress() != null && !receiver.getRemoteTarget().getAddress().isEmpty()) {
+                LOG.debug("Closing link due to the present of Target [address : {}]", receiver.getRemoteTarget().getAddress());
+            }
+            receiver.setCondition(
```
sophokles73: why don't you allow a Target without an *address*?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -286,64 +286,60 @@ private void handleSessionOpen(final ProtonConnection conn, final ProtonSession
...
+            receiver.setTarget(null);
             receiver.setQoS(receiver.getRemoteQoS());
             if (ProtonQoS.AT_LEAST_ONCE.equals(receiver.getRemoteQoS())) {
                 // disable auto-accept for this transfer model.
                 // in this case, the adapter will apply the required disposition
                 receiver.setAutoAccept(false);
             }
             receiver.handler((delivery, message) -> {
-                uploadMessage(new AmqpContext(delivery, message, receiver.getRemoteTarget(), authenticatedDevice));
+                uploadMessage(receiver.getRemoteQoS(), new AmqpContext(delivery, message, authenticatedDevice));
```
sophokles73: the device/gateway will use this link to send telemetry and/or event messages using both *at least once* and *at most once*. The QoS indicated on link establishment therefore is not very useful. Instead, you need to determine the QoS for each message based on whether the message is already settled or not ...

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
Alfusainey: sorry, i was sleeping here..

Alfusainey:
@sophokles73: PR updated with 0939fae

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
Alfusainey: now i understand what you mean :+1:

Alfusainey:
@sophokles73: thanks for all the feedback! would be nice if we can have this patch out of the way before end of working day today :-)

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpContext.java
```diff
@@ -37,14 +38,20 @@
...
+    AmqpContext(final ProtonDelivery delivery, final Message message, final Device authenticatedDevice) {
         this.delivery = delivery;
         this.message = message;
-        this.resource = resource;
         this.authenticatedDevice = authenticatedDevice;
+        try {
+            validateMessageAddress(message.getAddress());
+            this.resource = ResourceIdentifier.fromString(message.getAddress());
+        } catch (ServiceInvocationException sie) {
+            handleFailure(sie);
```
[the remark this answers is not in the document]
Alfusainey: yes, indeed! will update the PR

Alfusainey:
@sophokles73: PR updated with 3904c2e

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -287,71 +287,69 @@ private void handleSessionOpen(final ProtonConnection conn, final ProtonSession
...
-            formalCheck.fail(new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST,
+        if (!isPayloadOfIndicatedType(context.getMessagePayload(), contentType)) {
+            contentTypeCheck.fail(new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST,
                     String.format("Content-Type: [%s] does not match payload", contentType)));
+        } else {
+            contentTypeCheck.complete();
         }
 
-        formalCheck.compose(ok -> {
+        CompositeFuture.all(contentTypeCheck, validateEndpoint(context)).compose(ok -> {
```
sophokles73: is there a reason why you want to validate the endpoint as part of the uploadMessage method instead of doing it as part of the message address check like the MQTT adapter does it? It helps people to understand the different adapters more easily if we use similar patterns in the adapters. So, unless there are any compelling reasons to do otherwise, IMHO you should follow the pattern used in the MQTT adapter, i.e. check the endpoint (and QoS) of the incoming message and then validate the message's address before handing it over to the uploadMessage method. WDYT?

Alfusainey: the problem is that validating the endpoint requires a context instance which is already created after validating the address. let me see how to refactor that

Alfusainey:
@sophokles73 : PR updated with caa7de0

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -287,71 +288,72 @@ private void handleSessionOpen(final ProtonConnection conn, final ProtonSession
...
      */
     protected void handleRemoteReceiverOpen(final ProtonReceiver receiver, final ProtonConnection conn) {
-        if (receiver.getRemoteTarget() == null) {
+        if (receiver.getRemoteTarget() != null && receiver.getRemoteTarget().getAddress() != null) {
+            if (!receiver.getRemoteTarget().getAddress().isEmpty()) {
+                LOG.debug("Closing link due to the present of Target [address : {}]", receiver.getRemoteTarget().getAddress());
+            }
             receiver.setCondition(
-                    ProtonHelper.condition(AmqpError.NOT_ALLOWED, "anonymous relay not supported by this adapter"));
+                    AmqpContext.getErrorCondition(new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST,
```
sophokles73: shouldn't this be in an `else` branch?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -390,7 +390,7 @@ protected void uploadMessage(final AmqpContext context) {
             if (tenantObject.isAdapterEnabled(getTypeName())) {
 
                 final MessageSender sender = senderFuture.result();
-                final ResourceIdentifier resource = ResourceIdentifier.from(endpointName, context.getTenantId(), deviceId);
+                final ResourceIdentifier resource = ResourceIdentifier.from(endpointName, context.getTenantId(), context.getDeviceId());
```
sophokles73: why not simply use `context.getResourceIdentifier()` instead of instantiating a new one?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -428,7 +428,7 @@ protected void uploadMessage(final AmqpContext context) {
         }).recover(t -> {
             LOG.debug("Cannot process message for Device [tenantId: {}, deviceId: {}, endpoint: {}] due to Exception [Class: {}, Message: {}]",
                     context.getTenantId(),
-                    deviceId,
+                    context.getDeviceId(),
```
sophokles73: omit Class and Message and simply log the exception:
```
LOG.debug("Cannot process message for Device [tenantId: {}, deviceId: {}, endpoint: {}]",
context.getTenantId(), context.getDeviceId(), context.getEndpoint(), e);
```

adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -65,7 +68,7 @@
 public class VertxBasedAmqpProtocolAdapterTest {
 
     @Rule
-    public Timeout globalTimeout = new Timeout(5, TimeUnit.SECONDS);
+    public Timeout globalTimeout = new Timeout(5, TimeUnit.HOURS);
```
sophokles73: seconds

adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -158,25 +161,43 @@ public void testStartUsesClientProvidedAmqpServer(final TestContext ctx) {
     }
 
     /**
-     * Verifies that the AMQP Adapter does not support anonymous relay.
+     * Verifies that the AMQP Adapter support anonymous relay both for authenticated and unauthenticated clients.
      */
     @Test
-    public void testAnonymousRelayNotSupported() {
+    public void testAnonymousRelaySupportedForAllClients() {
```
sophokles73: IMHO this should be `testAnonymousRelayRequired` and verify that a client cannot open a link that contains a target address

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
Alfusainey: humm.. i don't think so. once we enter that block, then a target address is already set

Alfusainey:
@sophokles73: patch updated with commit 160abb0

sophokles73:

Alfusainey:
:man_dancing:

Alfusainey:
@sophokles73: the travis build passes for this PR

PR #731, opened 2018-07-25
Coap adapter 0.7

boaks:
coap adapter update to work with hono 0.7-SNAPSHOT after hono 07-M2.
Supports x.509 client authentication.
Obsoletes my old PR #599 

### DIFF
--- a/adapters/coap-vertx-base/pom.xml
+++ b/adapters/coap-vertx-base/pom.xml
@@ -1,0 +1,40 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.eclipse.hono</groupId>
+    <artifactId>hono-adapters</artifactId>
+    <version>0.7-SNAPSHOT</version>
+  </parent>
+  <artifactId>hono-adapter-coap-vertx-base</artifactId>
+  <name>Hono COAP Adapter base</name>
+  <url>https://www.eclipse.org/hono</url>
+
+  <dependencies>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.vertx</groupId>
+      <artifactId>vertx-unit</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.californium</groupId>
+      <artifactId>californium-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.californium</groupId>
+      <artifactId>element-connector</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.californium</groupId>
+      <artifactId>scandium</artifactId>
+    </dependency>
+  </dependencies>
+
+</project>
--- a/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
+++ b/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
@@ -1,0 +1,429 @@
+/**
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.hono.adapter.coap;
+
+import java.io.File;
+import java.net.HttpURLConnection;
+import java.net.InetSocketAddress;
+import java.util.Objects;
+
+import org.apache.qpid.proton.message.Message;
+import org.eclipse.californium.core.CoapServer;
+import org.eclipse.californium.core.coap.CoAP;
+import org.eclipse.californium.core.coap.CoAP.ResponseCode;
+import org.eclipse.californium.core.coap.MediaTypeRegistry;
+import org.eclipse.californium.core.network.CoapEndpoint;
+import org.eclipse.californium.core.network.Endpoint;
+import org.eclipse.californium.core.network.config.NetworkConfig;
+import org.eclipse.californium.core.server.resources.CoapExchange;
+import org.eclipse.hono.client.ClientErrorException;
+import org.eclipse.hono.client.MessageSender;
+import org.eclipse.hono.client.ServerErrorException;
+import org.eclipse.hono.service.AbstractProtocolAdapterBase;
+import org.eclipse.hono.service.auth.device.Device;
+import org.eclipse.hono.util.Constants;
+import org.eclipse.hono.util.EventConstants;
+import org.eclipse.hono.util.ResourceIdentifier;
+import org.eclipse.hono.util.TelemetryConstants;
+import org.eclipse.hono.util.TenantObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * Base class for a Vert.x based Hono protocol adapter that uses CoAP. It provides access to the Telemetry and Event
+ * API.
+ * 
+ * @param <T> The type of configuration properties used by this service.
+ */
+public abstract class AbstractVertxBasedCoapAdapter<T extends CoapAdapterProperties>
+        extends AbstractProtocolAdapterBase<T> {
+
+    protected final Logger log = LoggerFactory.getLogger(getClass());
+
+    /**
+     * COAP server. Created from blocking execution, therefore use volatile.
+     */
+    private CoapServer server;
+    private volatile Endpoint insecureEndpoint;
+    private CoapAdapterMetrics metrics;
+
+    /**
+     * Sets the metrics for this service.
+     *
+     * @param metrics The metrics
+     */
+    @Autowired
+    public final void setMetrics(final CoapAdapterMetrics metrics) {
+        this.metrics = metrics;
+    }
+
+    /**
+     * @return {@link CoAP#DEFAULT_COAP_SECURE_PORT}
+     */
+    @Override
+    public final int getPortDefaultValue() {
+        return CoAP.DEFAULT_COAP_SECURE_PORT;
+    }
+
+    /**
+     * @return {@link CoAP#DEFAULT_COAP_PORT}
+     */
+    @Override
+    public final int getInsecurePortDefaultValue() {
+        return CoAP.DEFAULT_COAP_PORT;
+    }
+
+    @Override
+    protected final int getActualPort() {
+        return Constants.PORT_UNCONFIGURED;
+    }
+
+    @Override
+    protected final int getActualInsecurePort() {
+        int port = Constants.PORT_UNCONFIGURED;
+        final Endpoint endpoint = insecureEndpoint;
+        if (endpoint != null) {
+            port = endpoint.getAddress().getPort();
+        }
+        return port;
+    }
+
+    /**
+     * Called during {@link #doStart(Future)}.
+     * 
+     * Note: The call is in the scope of "executeBlocking" and therefore special care accessing this instance must be
+     * obeyed.
+     * 
+     * @param adapterContext context of this adapter. Intended for schedule calls back into this vertical.
+     * @param server coap server to add resources.
+     */
+    protected abstract void addResources(Context adapterContext, CoapServer server);
+
+    /**
+     * Sets the coap server instance configured to serve requests.
+     * <p>
+     * If no server is set using this method, then a server instance is created during startup of this adapter based on
+     * the <em>config</em> properties.
+     * 
+     * @param server The coap server.
+     * @throws NullPointerException if server is {@code null}.
+     */
+    @Autowired(required = false)
+    public final void setCoapServer(final CoapServer server) {
+        Objects.requireNonNull(server);
+        this.server = server;
+    }
+
+    @Override
+    public final void doStart(final Future<Void> startFuture) {
+        checkPortConfiguration().compose(s -> preStartup()).compose(s -> {
+            final Future<Void> deployFuture = Future.future();
+            // access environment using the context of this vertx and
+            // load it into finals to access them within the executeBlocking
+            final CoapAdapterProperties config = getConfig();
+            final NetworkConfig insecureNetworkConfig = getInsecureNetworkConfig();
+            final Context adapterContext = this.context;
+            final CoapServer server = this.server;
+
+            // delegate for blocking execution
+            getVertx().executeBlocking(future -> {
+                // Call some potential blocking API
+                final CoapServer startingServer = server == null ? new CoapServer(insecureNetworkConfig) : server;
+                addResources(adapterContext, startingServer);
+                Endpoint insecureEndpoint = null;
+                if (config.isInsecurePortEnabled()) {
+                    if (config.isAuthenticationRequired()) {
+                        LOG.warn("ambig configuration! authenticationRequired is not supported for insecure endpoint!");
+                    } else {
+                        final CoapEndpoint.CoapEndpointBuilder builder = new CoapEndpoint.CoapEndpointBuilder();
+                        builder.setNetworkConfig(insecureNetworkConfig);
+                        builder.setInetSocketAddress(new InetSocketAddress(config.getInsecurePortBindAddress(),
+                                config.getInsecurePort(getInsecurePortDefaultValue())));
+                        insecureEndpoint = builder.build();
+                        startingServer.addEndpoint(insecureEndpoint);
+                    }
+                }
+                startingServer.start();
+                this.insecureEndpoint = insecureEndpoint;
+                future.complete(startingServer);
+            }, res -> {
+                if (res.succeeded()) {
+                    this.server = (CoapServer) res.result();
+                    deployFuture.complete();
+                } else {
+                    deployFuture.fail(res.cause());
+                }
+            });
+            return deployFuture;
+        }).compose(s -> {
+            try {
+                onStartupSuccess();
+                startFuture.complete();
+            } catch (Exception e) {
+                LOG.error("error in onStartupSuccess", e);
+                startFuture.fail(e);
+            }
+        }, startFuture);
+    }
+
+    /**
+     * Invoked before the coap server is started.
+     * <p>
+     * May be overridden by sub-classes to provide additional startup handling.
+     * 
+     * @return A future indicating the outcome of the operation. The start up process fails if the returned future
+     *         fails.
+     */
+    protected Future<Void> preStartup() {
+
+        return Future.succeededFuture();
+    }
+
+    /**
+     * Invoked after this adapter has started up successfully.
+     * <p>
+     * May be overridden by sub-classes.
+     */
+    protected void onStartupSuccess() {
+        // empty
+    }
+
+    /**
+     * Get the coap network configuration for the insecure endpoint.
+     * 
+     * Creates a coap network configuration setup with defaults. If the {@link CoapAdapterProperties} provides a
+     * filename in "networkConfig", load that available values from that file overwriting existing values. If the
+     * {@link CoapAdapterProperties} provides a filename in "insecureNetworkConfig", load that available values from
+     * that file also overwriting existing values.
+     * 
+     * @return coap network configuration for the insecure endpoint.
+     */
+    protected NetworkConfig getInsecureNetworkConfig() {
+        final CoapAdapterProperties config = getConfig();
+        final NetworkConfig networkConfig = new NetworkConfig();
+        networkConfig.setInt(NetworkConfig.Keys.PROTOCOL_STAGE_THREAD_COUNT, config.getCoapThreads());
+        networkConfig.setInt(NetworkConfig.Keys.NETWORK_STAGE_RECEIVER_THREAD_COUNT, config.getConnectorThreads());
+        networkConfig.setInt(NetworkConfig.Keys.NETWORK_STAGE_SENDER_THREAD_COUNT, config.getConnectorThreads());
+        loadNetworkConfig(config.getNetworkConfig(), networkConfig);
+        loadNetworkConfig(config.getInsecureNetworkConfig(), networkConfig);
+        return networkConfig;
+    }
+
+    protected void loadNetworkConfig(final String fileName, final NetworkConfig networkConfig) {
+        if (fileName != null && !fileName.isEmpty()) {
+            final File file = new File(fileName);
+            String cause = null;
+            if (!file.exists()) {
+                cause = "File doesn't exists!";
+            } else if (!file.isFile()) {
+                cause = "Isn't a file!";
+            } else if (!file.canRead()) {
+                cause = "Can't read file!";
+            } else {
+                networkConfig.load(file);
+            }
+            if (cause != null) {
+                LOG.warn("Can't read NetworkConfig from {}! {}", fileName, cause);
+            }
+        }
+    }
+
+    /**
+     * Invoked before the message is sent to the downstream peer.
+     * <p>
+     * Subclasses may override this method in order to customize the message before it is sent, e.g. adding custom
+     * properties.
+     * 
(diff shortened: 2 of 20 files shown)


adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/PreSharedKeyCredentialsProvider.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Authentication provider supporting pre-shared-key handshake.
+ * 
+ * Extends the {@link CredentialsApiAuthProvider} with a public
+ * {@link #getCredentialsForDevice(PreSharedKeyDeviceIdentity)} to provided the stored secret key for the device. In
+ * difference to the authentication model supported direct by the {@link CredentialsApiAuthProvider}, which compares a
+ * already available key with the stored one, a pre-shared-key handshake requires to access the secret key to generate a
+ * proper finish message.
+ */
+public final class PreSharedKeyCredentialsProvider extends CredentialsApiAuthProvider {
```
sophokles73: I am not sure I understand the purpose of this class. I would have simply created an implementation of Scandium's PskStore that uses a CredentialsClient to retrieve the pre-shared key from the Credentials service.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapCertificateHandler.java
```diff
@@ -0,0 +1,322 @@
...
+        return result;
+    }
+
+    @Override
+    public Class<X509CertPath> getType() {
+        return X509CertPath.class;
+    }
+
+    @Override
+    public Future<Device> getAuthorizedDevice(final CoapExchange exchange) {
```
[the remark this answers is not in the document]
boaks: OK.

boaks:
> Oh, and BTW, we have migrated to EPL 2.0 so it would be great if you could do a global replace for the license header ...

Any "tool" for that? Or manual update?

boaks:
> There is a lot of functionality included (multiple auth schemes, even a newly invented one) which I would prefer to be added gradually. Is it possible that you strip this down to the bare minimum to simply publish a telemetry or event message from an unauthenticated device? That would make it much easier to review and would also allow me to merge the PR without creating a CQ ...

That's (hopefully) possible to start with PSK or x.509 only, but I have my doubts, that this would make it under the 1000 lines.

For "plain" I'm not sure, how the device should be identified (or authenticated). In my previous PR I provided a URL base mechanism, but you seem to have your doubts (https://github.com/eclipse/hono/pull/599#discussion_r184040013).

So I see three variants:
1. plain, providing the devce id or auth id + cred. Maybe using the URI-PATH or URI-QUERY or?
2. server side authorized dtls (x.509, server cert only), providing the devce id or auth id + cred. Maybe using the URI-PATH or URI-QUERY or?
3. dtls authorized (x.509 or PSK).

1. and 2. would require to define, how the device id or auth id is provided.
2. would require, if used together with 3 in the future, a californium implementation of
`sslEngine.setWantClientAuth(true);` currently only `sslEngine.setNeedClientAuth(true);` is implemented. See https://github.com/eclipse/californium/issues/703#issuecomment-407434174 
3. is the variant I chose, because it seems to be the most defined one. But, as mentioned above, I have my doubts, that limit it to PSK or x.509 only will not shrink it to 1000 lines.

adapters/coap-vertx-base/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+  <parent>
+    <groupId>org.eclipse.hono</groupId>
+    <artifactId>hono-adapters</artifactId>
+    <version>0.7-SNAPSHOT</version>
+  </parent>
+  <artifactId>hono-adapter-coap-vertx-base</artifactId>
+  <name>Hono COAP Adapter base</name>
+  <url>https://www.eclipse.org/hono</url>
+
+  <dependencyManagement>
```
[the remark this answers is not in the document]
boaks: OK

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
```diff
@@ -0,0 +1,527 @@
...
+                    try {
+                        final char[] pass = config.getKeyStorePassword().toCharArray();
+                        final SslContextUtil.Credentials serverCredentials = SslContextUtil.loadCredentials(
+                                config.getKeyStorePath(),
+                                config.getKeyStoreAlias(), pass, pass);
+                        dtlsConfig.setIdentity(serverCredentials.getPrivateKey(),
+                                serverCredentials.getCertificateChain(),
+                                config.isPreferRawPublicKey());
+                    } catch (GeneralSecurityException e) {
+                        e.printStackTrace();
```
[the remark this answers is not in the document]
boaks: OK

adapters/pom.xml
```diff
@@ -27,6 +27,8 @@
   <packaging>pom</packaging>
 
   <modules>
+    <module>coap-vertx-base</module>
+    <module>coap-vertx</module>
```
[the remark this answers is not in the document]
boaks: OK.

adapters/coap-vertx/src/main/resources/logback-spring.xml
```diff
@@ -0,0 +1,37 @@
...
+  </springProfile>
+
+  <logger name="io.netty.handler.logging.LoggingHandler" level="INFO"/>
+
+  <logger name="io.vertx.proton.impl" level="INFO"/>
+  <logger name="io.vertx.core.net.impl" level="INFO"/>
+  <logger name="org.eclipse.hono.client.impl.TelemetrySenderImpl" level="INFO"/>
+  <logger name="org.eclipse.hono.client.impl.EventSenderImpl" level="INFO"/>
+
+</configuration>
```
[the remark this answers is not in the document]
boaks: OK

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapErrorResponse.java
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * Response the coap exchange with the provide error cause.
+     * 
+     * Convert http-code into coap-code, if available. Add cause message as payload for response.
+     * 
+     * @param exchange coap exchange to be responded
+     * @param cause error cause
+     * @param defaultCode default response code, if a more specific response code is not available.
+     */
+    public static void response(final CoapExchange exchange, final Throwable cause, final ResponseCode defaultCode) {
```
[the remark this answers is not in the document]
boaks: OK

sophokles73:
> Any "tool" for that? Or manual update?

Well, you can use Eclipse's search and replace functionality ... shouldn't be much of an issue.

sophokles73:
> For "plain" I'm not sure, how the device should be identified (or authenticated).

Please take a look at the HTTP adapter User Guide. The resource documentation should give you a pretty good idea of how this might work. In fact, I do not see why the CoAP adapter shouldn't expose the same resources, given that CoAP follows the same request/response pattern as HTTP ...

sophokles73:
> In my previous PR I provided a URL base mechanism, but you seem to have your doubts

Indeed, I would rather keep this out of the PR for the time being. We can add this later, if we consider it necessary. But again, I would rather start supporting a DHE PSK cipher in Californium ...

boaks:
> Please take a look at the HTTP adapter User Guide. The resource documentation should give you a pretty good idea of how this might work. In fact, I do not see why the CoAP adapter shouldn't expose the same resources, given that CoAP follows the same request/response pattern as HTTP ...

Cool, for me it's vice versa :-). 

HTTP basic auth => I'm not sure, what this should match in coap
Client Certificate => that's 3 not 1

So, I just don't know, what you mean.

boaks:
> Indeed, I would rather keep this out of the PR for the time being. We can add this later, if we consider it necessary. But again, I would rather start supporting a DHE PSK cipher in Californium ...

You make me really wondering. How can you use the term "again" with DHE PSK? and why shouldn't it work already with the CoapPreSharedKeyHandler?

sophokles73:
> So, I just don't know, what you mean. 

FMPOV the first step should be to be able to have an unauthenticated device POST telemetry data or an event to the adapter. You asked how we know the tenant and device ID in that case. IMHO we should follow the same approach as the HTTP adapter takes in that case: extract the tenant and device ID from the URI.

sophokles73:
> You make me really wondering. How can you use the term "again" with DHE PSK? and why shouldn't it work already with the CoapPreSharedKeyHandler?

I was referring to the mechanism of including the credentials in the URI, which, IIRC you proposed as an alternative to using a DHE based PSK authentication which exposes the auth-id in cleartext. Or am I confusing things?

adapters/coap-vertx-base/pom.xml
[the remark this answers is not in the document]
boaks: Any intended order?

adapters/coap-vertx/src/main/resources/logback-spring.xml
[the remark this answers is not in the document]
boaks: Just to mention, though I copied the file from the http adapter, the EOL seems to be missing also there.

adapters/pom.xml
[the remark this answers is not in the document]
boaks: Should I correct the "wrong" order for the http-modules (which I simply copied) also?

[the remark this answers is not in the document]
boaks: and kura and mqtt?
sorry, if lexical order is hard to recognize, if the already contained stuff is not in that order :-)

boaks:
fits into 1700 ;-)

adapters/coap-vertx/src/main/java/org/eclipse/hono/adapter/coap/vertx/VertxBasedCoapAdapter.java
```diff
@@ -0,0 +1,92 @@
...
+                final Device device = new Device(path.get(identityLevel), path.get(identityLevel + 1));
+                final ExtendedDevice extendedDevice = new ExtendedDevice(device, device);
+                if (extendedDevice != null) {
+                    log.debug("use {}", extendedDevice);
+                    handler.handle(extendedDevice);
+                    return;
+                }
+            }
+        }
+        exchange.respond(ResponseCode.INTERNAL_SERVER_ERROR);
```
sophokles73: we usually try to return a more specific error, e.g. if an unauthenticated device's request URI does not contain a tenant and device ID, the HTTP adapter returns a 404

adapters/coap-vertx/src/main/java/org/eclipse/hono/adapter/coap/vertx/VertxBasedCoapAdapter.java
```diff
@@ -0,0 +1,92 @@
...
+     * @param exchange coap exchange with URI.
+     * @param identityLevel path level to extract URI identity. e.g. for "/telemetry/tenant/authid" the level is 1.
+     * @param handler handler for determined extended device
+     */
+    public void getExtendedDevice(final CoapExchange exchange, final int identityLevel,
+            final Handler<ExtendedDevice> handler) {
+        if (identityLevel > 0) {
+            final List<String> path = exchange.getRequestOptions().getUriPath();
+            if (path.size() == identityLevel + 2) {
+                final Device device = new Device(path.get(identityLevel), path.get(identityLevel + 1));
```
sophokles73: you may want to consider using the `org.eclipse.hono.util.ResourceIdentifier` class which provides semantic access to the URI's segments.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
```diff
@@ -0,0 +1,419 @@
...
+                            ResourceIdentifier.from(endpointName, authenticatedDevice.getTenantId(),
+                                    device.getDeviceId()),
+                            sender.isRegistrationAssertionRequired(),
+                            "/" + exchange.getRequestOptions().getUriPathString(),
+                            contentType,
+                            payload,
+                            tokenTracker.result(),
+                            null);
+                    customizeDownstreamMessage(downstreamMessage, exchange);
+                    // wait for outcome, ensure message order, if CoAP NSTART-1 is used.
```
sophokles73: The specific method to use for sending a telemetry message downstream mainly depends on the QoS that the device wants to use. If its QoS 0 (fire-and-forget, the default), then `sender.send()` is more appropriate because it allows the response to be sent immediately to the device (always 2.04 in CoAP I assume).
If the device wants to use QoS 1, then `sender.sendAndWaitForOutcome` is appropriate because the returned Future will only be completed successfully, once the downstream peer has accepted the message.

For events we only support QoS 1, thus `sender.sendAndWaitForOutcome` is always the right choice.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
```diff
@@ -0,0 +1,419 @@
...
+    }
+
+    @Override
+    public final void doStart(final Future<Void> startFuture) {
+        checkPortConfiguration().compose(s -> preStartup()).compose(s -> {
+            final Future<Void> deployFuture = Future.future();
+            // access environment using the context of this vertx and
+            // load it into finals to access them within the executeBlocking
+            final CoapAdapterProperties config = getConfig();
+            final Context adapterContext = getVertx().getOrCreateContext();
```
sophokles73: you can simply refer to this adapter's vert.x context by the `context` field, which is set by vert.x on startup of the Verticle.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
```diff
@@ -0,0 +1,419 @@
...
+    /**
+     * Called during {@link #doStart(Future)}.
+     * 
+     * Note: The call is in the scope of "executeBlocking" and therefore special care accessing this instance must be
+     * obeyed.
+     * 
+     * @param adapterContext context of this adapter. Intended for schedule calls back into this vertical.
+     * @param server coap server to add resources.
+     */
+    protected abstract void addResources(Context adapterContext, CoapServer server);
```
sophokles73: no need to pass in the adapter's vert.x context. a subclass can simply refer to the class's `context` field for that purpose because all adapters are `Verticle`s.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
boaks: I'm not really sure, what you want to say.

In CoAP the most common pattern, as far as I know, is a CON request. That seems to be closest to QoS 1, or do you have a different opinion? 

NON request seems to be rarely used, so I just didn't pay attention to them.
I'm not sure, what the target should now be, try to make a smallest possible PR?
Or to support rare use-cases?

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
boaks: Will that `context` field be visible for other threads?
I could see that in the implementation nor did I find it in the docu.

```
final Context adapterContext = context; 
```

so that would be the resulting solution. Or do you have more information about using the context from other threads?

boaks:
Maybe, before we continue, its 1700 line not below 1000. 
So do you still want the "plain coap only" PR?

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
boaks: >     * Note: The call is in the scope of "executeBlocking" and therefore special care accessing this instance must be
>     * obeyed.

As mentioned above, if you are sure, that accessing the `context` from other threads is save, OK.
But FMPOV, the behaviour according other threads seems to be not documented.

adapters/coap-vertx/src/main/java/org/eclipse/hono/adapter/coap/vertx/VertxBasedCoapAdapter.java
boaks: I'm not sure, if you just want to exchange the extraction of the tenant and device id, or to replace the usage of `Device` or the usage of the `ExtendedDevice`. 

I introduced the usage of the `ExtendedDevice`, because it seems for me to be the "most straight forward" approach to pass the information to the abstract adapter, especially, if prinicpals are used to identify the "authorized device".

Or should it only be used to replace the usage of `Device` for the origin?
Or the extracting the origin device from the URL?

sophokles73: No, I just wanted to suggest using `ResourceIdentifier.from(String)` in order to parse the URI and access the endpoint name, tenant and device IDs using `ResourceIdentifier`'s getters, instead of working on a bare String array ...

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
sophokles73: This is not about the QoS of the CoAP request itself. I agree that we should use CON CoAP requests. The question here is, when do we return the CoAP response to the device. When a device uploads telemetry data, we do not make any guarantees as to whether the data is actually processed or gets lost while sending it downstream (this is what we refer to as QoS 0 in Hono). When processing the CoAP request, the adapter therefore doesn't need to wait for the acknowledgment of the downstream AMQP 1.0 peer that it has received the message. This behavior is reflected by the `MessageSender.send()` method. This method immediately returns a succeeded Future, if it was at all possible to send the message downstream, i.e. a connection exists and we have some credits. In constrast, the `MessageSender.sendAndWaitForOutcome()` method sends the message, but completes the returned Future only when the downstream peers acks the reception of the message. We refer to this behavior with QoS 1 in Hono. So, depending on what type of message the device uploads, the adapters invoke *send* (telemetry) or  *sendAndWaitForOutcome* (event) and return a response to the device based on the Future's outcome.

sophokles73:
> So do you still want the "plain coap only" PR?

Yes, please. It makes it much easier for me to review.

adapters/coap-vertx/src/main/java/org/eclipse/hono/adapter/coap/vertx/VertxBasedCoapAdapter.java
boaks: Coap split already the uri path into several string parts. If you want to reuse the indexing, I can use the `ResourceIdentifier.fromPath(final String[] path)`. But this implementation seems not validating the path too much :-), so I'm not sure, if the validation should not implemented or external implemented, which would make it more fragile for changes.
My feeling, this  `ResourceIdentifier` is too generic to work her well.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
boaks: ??? What's that ???

First you state:
"The specific method to use for sending a telemetry message downstream mainly depends on the QoS that the device wants to use. "

And now:

"So, depending on what type of message the device uploads, the adapters invoke send (telemetry) or sendAndWaitForOutcome (event) and return a response to the device based on the Future's outcome."

If it takes just a few days, to change your mind, would it be required to make it configurable?

sophokles73: I have not changed my mind, but I might not have been precise enough about the whole thing in my previous comments.

I think the confusion stems from the fact that a device may send telemetry using either QoS 0 (which is the default) or QoS 1 (which it needs to indicate explicitly, e.g. in the HTTP adapter by including the *qos-level* header or query parameter in the request).

On the other hand, events published by devices are *always* processed with QoS 1. You may want to refer to the other protocol adapters' corresponding *doUploadMessage* methods for how they determine the intended QoS and then use the corresponding methods to send the messages downstream.

boaks: I'm not sure, why a client should retransmit its message, if the server then drops it :-).
I can add support for NON, if you think that's used (we are anyway above the 1000 lines).
Or just define the way, you want to mark the CoAP request to be either processed as QoS 0 or 1.

boaks:
> Oh, and BTW, we have migrated to EPL 2.0 so it would be great if you could do a global replace for the license header ...

Done.

boaks:
Should I also add the commit with the additional credentials?
The example deployment seems to be moved. Do you think, that's the variant, which is worth to move to?

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
sophokles73: I am not trying to prescribe any particular way of implementing it. I merely tried to explain the overall idea of the different QoS levels and what their purpose within Hono is. Mapping these concepts to the transport protocol at hand, is the responsibility of the protocol adapter. That's why we are discussing this in the first place. I would welcome any proposals you have for doing so and if I understand correctly, you propose to use the (CoAP) request type to indicate the intended QoS. It sounds reasonable to me to support this, as it wouldn't require transmission of an additional header or query param. However, if I am not mistaken, you also seemed to have pointed out that some clients would never want to use a NON message at all (or may not even support sending them). If so, then it would still make sense to also support a corresponding option or query param, wouldn't it?

So, we would end up with something like this:
* if the request is for the telemetry endpoint and is a NON, use `MessageSender.send()`
* if the request is for the telemetry endpoint, is a CON and includes an option/query param indicating QoS 0,  use `MessageSender.send()` and send the response based on the outcome of the future returned by the send method
* if the request is for the event endpoint and is a CON, use `MessageSender.sendAndWaitForOutcome()` and send the response based on the outcome of the future

Does this make sense from your point of view?

sophokles73:
> The example deployment seems to be moved.

Yes, we now have a separate module (`deploy`) which contains all the example deployment variants.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
boaks: I wouldn't make it too complex, without having experience, how it is really used.
So I would got for NON => QoS 0 and CON => QoS 1,
 as long, as someone has no real use case, where something additional is required.

Basically the question behind that seems to be: 
is "use telemetry and the server may drop messages by random" really one major flavour,
or "when messages must be dropped, let the application/client decide which messages are dropped" would be more appreciated.
(My personal forecast is, many will start with "telemetry and random drop" but will end up in "selective drop".)

sophokles73: sounds good to me :-)

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
```diff
@@ -0,0 +1,527 @@
...
+            final Future<JsonObject> tokenTracker = getRegistrationAssertion(device.getTenantId(), device.getDeviceId(),
+                    authenticatedDevice);
+            final Future<TenantObject> tenantConfigTracker = getTenantConfiguration(authenticatedDevice.getTenantId());
+
+            CompositeFuture.all(tokenTracker, senderTracker, tenantConfigTracker).compose(ok -> {
+
+                if (tenantConfigTracker.result().isAdapterEnabled(getTypeName())) {
+                    final MessageSender sender = senderTracker.result();
+                    final Message downstreamMessage = newMessage(
+                            ResourceIdentifier.from(endpointName, authenticatedDevice.getTenantId(),
```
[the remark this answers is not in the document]
sophokles73: can you change this?

boaks:
>can you change this?

The usage of the `ResourceIdentifier`? I don't get it ;-(. 
May be you elaborate more what you want?

sophokles73:
> May be you elaborate more what you want?

You latest update has overwritten the original comment I made :-) Never mind, I'll add a new one ...

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
```diff
@@ -0,0 +1,430 @@
...
```
sophokles73: for reasons of clarity, this should be device.getTenantId(), IMHO

boaks:
OK, got it (and I also found the old comment :-) ).

sophokles73:
@boaks CQ has been approved, will merge now ...

Issue #331, opened 2017-09-07
Make connection timeout params configurable


balsmn:
In the following scenario's there could be connection problem leading to Hono client not able to connect to the Hono messaging

a. If DNS resolve takes a litte longer (for ex. currently statically set in [1])
b. If connection to resolved ip takes a litte longer. (for ex. currently set statically in Hono Receiver [2]) 

[1] https://github.com/eclipse/hono/blob/master/example/src/main/java/org/eclipse/hono/example/AppConfiguration.java#L48
[2] https://github.com/eclipse/hono/blob/master/example/src/main/java/org/eclipse/hono/example/AbstractExampleClient.java#L37

It would be better if they can be configured.

PR #737, opened 2018-07-27
[#331] Make connection timeout params configurable

kaniyan:
With reference to #331, now connection timeout parameter and QueryResolverOptions are configurable.

Example:mvn spring-boot:run -Drun.arguments=--hono.client.host=localhost,--hono.client.username=consumer@HONO,--hono.client.password=verysecret,**--address.resolver.queryTimeout=5000**,**--connection.timeout=3000**

### DIFF
--- a/cli/src/main/resources/application.yml
+++ b/cli/src/main/resources/application.yml
@@ -18,9 +18,21 @@ device:
 
 spring:
   profiles: receiver
+
 hono:
   client:
     port: 15671
+
+address:
+  resolver:
+    cacheNegativeTimeToLive: 0 #discard failed DNS lookup results immediately
+    cacheMaxTimeToLive: 0 #support DNS based service resolution
+    rotateServers: true
+    queryTimeout: 2000
+
+connection:
+  timeout: 1000
+
 message:
   type: all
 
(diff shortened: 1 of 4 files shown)


cli/src/main/resources/application.yml
```diff
@@ -18,9 +18,21 @@ device:
...
+
 hono:
   client:
     port: 15671
+
+address:
+  resolver:
+    cacheNegativeTimeToLive: 0 #discard failed DNS lookup results immediately
+    cacheMaxTimeToLive: 0 #support DNS based service resolution
+    rotateServers: true
```
pellmann: Are these properties really something, that a user need to configure?

kaniyan: Similar to the Address Resolution Timeout, there are many more properties in Vert.x hostname resolver (QueryResolverOptions).  For example, from the Operation-experts point of view, parameters such as searchDomains, maxQueries and queryTimeout are useful to tune based on the environment.  Now those can be configured.  IMHO users now have more flexibility to configure the hostname resolver if needed. Else default values are used as defined in the profile.

pellmann: My point was, that if we changed the defaults in a way that it is good for the application, it might be good to set these defaults in code. So the user CAN change them in the Yaml but it does not look like he should do it. But after thinking about - it might be good to see theses changed values directly in the configuration to be aware what happens.

kaniyan: I see so too, that it is better to set default values in spring application.yml rather in the code.

Issue #735, opened 2018-07-27
Half dead HTTP adapter instances


ctron:
Every now and then it happens to me that an HTTP protocol adapter instance is half dead. It is being reported as healthy, yet does not send any messages to the downstream Qpid router. HTTP requests return 503.

The following two exceptions keep repeating over and over:

~~~
08:21:33.966 [vert.x-eventloop-thread-0] ERROR i.v.e.w.impl.RoutingContextImplBase - Unexpected exception in route
java.lang.IllegalStateException: Response is closed
	at io.vertx.core.http.impl.HttpServerResponseImpl.checkValid(HttpServerResponseImpl.java:548)
	at io.vertx.core.http.impl.HttpServerResponseImpl.end0(HttpServerResponseImpl.java:401)
	at io.vertx.core.http.impl.HttpServerResponseImpl.end(HttpServerResponseImpl.java:319)
	at io.vertx.core.http.impl.HttpServerResponseImpl.end(HttpServerResponseImpl.java:308)
	at io.vertx.ext.web.impl.RoutingContextImplBase.unhandledFailure(RoutingContextImplBase.java:166)
	at io.vertx.ext.web.impl.RoutingContextImplBase.iterateNext(RoutingContextImplBase.java:132)
	at io.vertx.ext.web.impl.RoutingContextImpl.next(RoutingContextImpl.java:133)
	at io.opentracing.contrib.vertx.ext.web.TracingHandler.handlerFailure(TracingHandler.java:90)
	at io.opentracing.contrib.vertx.ext.web.TracingHandler.handle(TracingHandler.java:44)
	at io.opentracing.contrib.vertx.ext.web.TracingHandler.handle(TracingHandler.java:25)
	at io.vertx.ext.web.impl.RouteImpl.handleFailure(RouteImpl.java:223)
	at io.vertx.ext.web.impl.RoutingContextImplBase.iterateNext(RoutingContextImplBase.java:117)
	at io.vertx.ext.web.impl.RoutingContextImpl.next(RoutingContextImpl.java:133)
	at io.vertx.ext.web.impl.RoutingContextImpl.doFail(RoutingContextImpl.java:465)
	at io.vertx.ext.web.impl.RoutingContextImpl.fail(RoutingContextImpl.java:166)
	at org.eclipse.hono.service.auth.device.HonoAuthHandler.processException(HonoAuthHandler.java:233)
[...]

PR #738, opened 2018-07-27
[#735]: Fix issue with requests not being completed properly

ctron:
The message was inserted into the map using either the correlation ID
or the message ID. However are only removed from the map using the
correlation ID, which is never set.


ctron:
No .. that is completely wrong. Time for weekend!

Issue #632, opened 2018-05-28
Add metrics for Command and Control to HTTP adapter


sysexcontrol:
We need Grafana dashboards (scoped to tenant) for:

- command delivered to device before ttd expires
- ttd expired with no command received from applications
- command response delivered from device to application

kaniyan:
I made changes to collect below metrics for Command & Control and store them in Influx-db. 

- Commands delivered (command delivered to device before ttd expires)
- TTD expiries (ttd expired with no command received from applications)
- Commad responses delivered (command response delivered from device to application)

Find below the screenshots of Dashboard with Command and Control metrics. Its just an idea, how command and  control metrics can be showed in the dashboard.  @ctron @pellmann @sophokles73 @sysexcontrol any opinion on this?  :)  

![honodashboard-1](https://user-images.githubusercontent.com/18659982/43309931-75e9003a-9186-11e8-8a3a-8d833a8cb5d3.jpg)
From #711 , I understood that the graphs at the top show the no. of messages per tenant (aggregated across all adapter types). In this case, it shows the no. of commands (delivered to devices) per tenant (aggregated across all adapter types).

![honodashboard-2](https://user-images.githubusercontent.com/18659982/43309971-8dccfac6-9186-11e8-9015-3f957a379ac3.jpg)

![honodashboard-3](https://user-images.githubusercontent.com/18659982/43309974-8ff9c32e-9186-11e8-9844-be6236cf6c34.jpg)

pellmann:
I like the boards! Have you already included the queries for MQTT?

kaniyan:
Thank you for the feedback. Queries for MQTT are already included in this Dashboard.

sophokles73:
@kaniyan are you also going to add these metrics to the MQTT adapter? If so, we could re-use this issue and simply change its title accordingly.

kaniyan:
@sophokles73  yes, I will continue with metrics for MQTT adapter. We can re-use this issue.

PR #741, opened 2018-07-31
[#632] Add metrics for Command and Control to HTTP adapter

kaniyan:
Below metrics are added for Command and Control (HTTP).
- Commands delivered (command delivered to device before ttd expires)
- TTD expiries (ttd expired with no command received from applications)
- Commad responses delivered (command response delivered from device to application)

Also the Grafana dashboard has been changed to accommodate the above metrics from command and control for both HTTP and MQTT adapters. For more info refer #632.

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/HttpAdapterMetrics.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/HttpAdapterMetrics.java
@@ -23,10 +23,21 @@ import org.springframework.stereotype.Component;
 public class HttpAdapterMetrics extends Metrics {
 
     private static final String SERVICE_PREFIX = "hono.http";
+    private static final String COMMANDS = ".commands";
 
     @Override
     protected String getPrefix() {
         return SERVICE_PREFIX;
     }
+    void incrementCommandDeliveredToDevice(final String tenantId) {
+        counterService.increment(METER_PREFIX + getPrefix() + mergeAsMetric(COMMANDS, tenantId, "device", "delivered"));
+    }
 
+    void incrementNoCommandReceivedAndTTDExpired(final String tenantId) {
+        counterService.increment(METER_PREFIX + getPrefix() + mergeAsMetric(COMMANDS, tenantId, "ttd", "expired"));
+    }
+
+    void incrementCommandResponseDeliveredToApplication(final String tenantId) {
+        counterService.increment(METER_PREFIX + getPrefix() + mergeAsMetric(COMMANDS, tenantId, "response", "delivered"));
+    }
 }
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/HttpAdapterMetrics.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/HttpAdapterMetrics.java
@@ -23,7 +23,6 @@ import org.springframework.stereotype.Component;
 public class HttpAdapterMetrics extends Metrics {
 
     private static final String SERVICE_PREFIX = "hono.http";
-    private static final String COMMANDS = ".commands";
 
     @Override
     protected String getPrefix() {
--- a/site/content/api/Metrics.md
+++ b/site/content/api/Metrics.md
@@ -45,7 +45,7 @@ extracting the following *tags* from metric names transmitted via the Graphite r
 | *host*     | The name of the host that the service reporting the metric is running on. |
 | *type*     | The type of message that the metric is being processed for (either `telemetry` or `event`) |
 | *tenant*   | The name of the tenant that the metric is being reported for. |
-| *protocol* | The protocol of message that the metric is being processed for (example: `http`, `mqtt`) |
+| *protocol* | The protocol of message that the metric is being processed for (example: `http`, `mqtt`). This is used only in Command and Control metrics. |
 
 The following sections describe which of these tags are extracted for which metrics specifically.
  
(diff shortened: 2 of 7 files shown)


adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/HttpAdapterMetrics.java
```diff
@@ -23,6 +23,7 @@
 public class HttpAdapterMetrics extends Metrics {
 
     private static final String SERVICE_PREFIX = "hono.http";
+    private static final String COMMANDS = ".commands";
```
pellmann: Since we have commands also in other components, I think this could go a level up. For the other messages we use a singular and also the endpoint names (telemetry, event). So maybe "control" and the constant from CommandConstants would be good? WDYT?

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/HttpAdapterMetrics.java
```diff
@@ -37,4 +38,15 @@ void incrementUndeliverableHttpMessages(final String resourceId, final String te
         counterService.increment(getPrefix() + MESSAGES + mergeAsMetric(resourceId, tenantId) + UNDELIVERABLE);
     }
 
+    void incrementCommandDeliveredToDevice(final String tenantId) {
+        counterService.increment(METER_PREFIX + getPrefix() + mergeAsMetric(COMMANDS, tenantId, "device", "delivered"));
```
pellmann: We should add this constants to the Metrics class, to have them reusable for other components.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/HttpAdapterMetrics.java
```diff
@@ -37,4 +38,15 @@ void incrementUndeliverableHttpMessages(final String resourceId, final String te
...
+    }
+
+    void incrementNoCommandReceivedAndTTDExpired(final String tenantId) {
+        counterService.increment(METER_PREFIX + getPrefix() + mergeAsMetric(COMMANDS, tenantId, "ttd", "expired"));
```
pellmann: Same as above.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/HttpAdapterMetrics.java
```diff
@@ -37,4 +38,15 @@ void incrementUndeliverableHttpMessages(final String resourceId, final String te
...
+    }
+
+    void incrementCommandResponseDeliveredToApplication(final String tenantId) {
+        counterService.increment(METER_PREFIX + getPrefix() + mergeAsMetric(COMMANDS, tenantId, "response", "delivered"));
```
pellmann: Same as above.

site/content/api/Metrics.md
```diff
@@ -40,11 +40,12 @@ a way to transmit metrics to InfluxDB resulting in the same database structure.
...
-| -------- | ----------- |
-| *host*   | The name of the host that the service reporting the metric is running on. |
-| *type*   | The type of message that the metric is being processed for (either `telemetry` or `event`) |
-| *tenant* | The name of the tenant that the metric is being reported for. |
+| Tag        | Description |
+| ---------- | ----------- |
+| *host*     | The name of the host that the service reporting the metric is running on. |
+| *type*     | The type of message that the metric is being processed for (either `telemetry` or `event`) |
+| *tenant*   | The name of the tenant that the metric is being reported for. |
+| *protocol* | The protocol of message that the metric is being processed for (example: `http`, `mqtt`) |
```
pellmann: I think you have added this protocol tag only for the new command metrics and not for the old events/telemetry?! Since we plan to change the metrics to a tag and micrometer based approach in the near future I also would not change it. But then we should mark, that this in only used for commands.

kaniyan: I have added only for the command metrics. Documentation has also been updated.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/HttpAdapterMetrics.java
kaniyan: I have pulled the COMMANDS to one level up to the Metrics.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/HttpAdapterMetrics.java
kaniyan: I have pulled the COMMANDS to one level up to the Metrics.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/HttpAdapterMetrics.java
kaniyan: I have pulled the COMMANDS to one level up to the Metrics.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/HttpAdapterMetrics.java
kaniyan: Thank you for the comments.

1. I have pulled the COMMANDS to one level up to the Metrics.
2. Currently the word "messages" (.hono..messages*) is part of the metrics name to denote metrics corresponding to telemetry or event messages. To differentiate commands from them, I used the word "commands" as part of the metrics name (.hono.commands). If I change to "control" the series would look like ".hono.control". IMHO "commands" follows the similar naming convention as of "messages". Or do you think that commands should also be part of ".hono..messages*" with type as "control". I thought of this approach too. But the current approach is optimal also for the graph at the top in dashboard which shows no. of commands per tenant (aggregated across all adapter types). WDYT?

sophokles73:
@kaniyan can you rebase on master so that we see if the Travis build works now?

kaniyan:
@sophokles73 I have rebased.

PR #742, opened 2018-07-31
Allow to customize vertx options and set option to prefer native sockets

ctron:
This PR introduces a common way to configure vertx specific options. As a first option is allows to set whether to prefer native TCP sockets or not.

Using native sockets still requires e.g. the netty epoll or kqueue implementation. However simply adding this dependency doesn't enable it unless the flag in the vertx option is set.

This is independent of the native TLS implementation!

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/AbstractServiceBase.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/AbstractServiceBase.java
@@ -224,6 +224,10 @@ public abstract class AbstractServiceBase<T extends ServiceConfigProperties> ext
      */
     protected final Future<Void> checkPortConfiguration() {
 
+        if (vertx != null) {
+            LOG.info("Vertx native support: {}", vertx.isNativeTransportEnabled());
+        }
+
         final Future<Void> result = Future.future();
 
         if (getConfig().getKeyCertOptions() == null) {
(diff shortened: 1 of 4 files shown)


service-base/src/main/java/org/eclipse/hono/service/AbstractServiceBase.java
```diff
@@ -399,7 +403,7 @@ protected final void addTlsKeyCertOptions(final NetServerOptions serverOptions)
             final boolean isOpenSslAvailable = OpenSsl.isAvailable();
             final boolean supportsKeyManagerFactory =  OpenSsl.supportsKeyManagerFactory();
             final boolean useOpenSsl =
-                    getConfig().isNativeTlsRequired() || (isOpenSslAvailable && supportsKeyManagerFactory);
+                    getConfig().isNativeTlsRequired() || isOpenSslAvailable && supportsKeyManagerFactory;
```
sophokles73: is there a reason why you removed the brackets?

ctron: Eclipse did that for me, as they are unnecessary.

sophokles73: But they improve readability a lot, don't you think? I have to admit that I would have to look up operator preference in order to predict the expression's value ...

ctron: For me more brackets are more confusing … but I can restore them and disable the cleanup action.

sophokles73: I would appreciate it :+1:

ctron: Restored by 85df64f15fd7b5ebea012b23b011140324d88e67

PR #744, opened 2018-08-01
[#643] Add support for client certificate based authentication

Alfusainey:
@sophokles73: this patch adds support for client certificate authentication. now the auth providers are known only to the SaslAuthenticator class. i plan to test feature in the ITs. 

### DIFF
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorFactory.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorFactory.java
@@ -12,20 +12,30 @@
  *******************************************************************************/
 package org.eclipse.hono.adapter.amqp;
 
+import java.security.cert.Certificate;
+import java.security.cert.X509Certificate;
+import java.util.Collections;
 import java.util.Objects;
 
+import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.security.auth.login.CredentialException;
+import javax.security.auth.x500.X500Principal;
 
 import org.apache.qpid.proton.engine.Sasl;
 import org.apache.qpid.proton.engine.Sasl.SaslOutcome;
 import org.apache.qpid.proton.engine.Transport;
+import org.eclipse.hono.client.HonoClient;
 import org.eclipse.hono.config.ProtocolAdapterProperties;
 import org.eclipse.hono.service.auth.device.Device;
+import org.eclipse.hono.service.auth.device.DeviceCertificateValidator;
 import org.eclipse.hono.service.auth.device.DeviceCredentials;
 import org.eclipse.hono.service.auth.device.HonoClientBasedAuthProvider;
+import org.eclipse.hono.service.auth.device.SubjectDnCredentials;
 import org.eclipse.hono.service.auth.device.UsernamePasswordAuthProvider;
 import org.eclipse.hono.service.auth.device.UsernamePasswordCredentials;
+import org.eclipse.hono.service.auth.device.X509AuthProvider;
 import org.eclipse.hono.util.AuthenticationConstants;
+import org.eclipse.hono.util.TenantObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -57,29 +67,31 @@ import io.vertx.proton.sasl.ProtonSaslAuthenticatorFactory;
  */
 public class AmqpAdapterSaslAuthenticatorFactory implements ProtonSaslAuthenticatorFactory {
 
-    private final HonoClientBasedAuthProvider authProvider;
     private final ProtocolAdapterProperties config;
+    private final HonoClient tenantServiceClient;
+    private final HonoClient credentialsServiceClient;
 
     /**
      * Creates a new SASL authenticator factory for an authentication provider. If the AMQP adapter supports
      * multi-tenancy, then the authentication identifier contained in the SASL response should have the pattern
      * {@code [<authId>@<tenantId>]}.
      *
-     * @param authProvider The authentication provider to use when creating the SASL authenticator instance through
-     *            {@link #create()}.
+     * @param tenantServiceClient The Tenant Service client of the SASL authenticator.
+     * @param credentialsServiceClient The Credentials servicec client of the SASL authenticator.
      * @param config The protocol adapter configuration object.
      *
-     * @throws NullPointerException if either authProvider or config is null.
+     * @throws NullPointerException if any of the parameters is null.
      */
-    public AmqpAdapterSaslAuthenticatorFactory(final HonoClientBasedAuthProvider authProvider,
+    public AmqpAdapterSaslAuthenticatorFactory(final HonoClient tenantServiceClient, final HonoClient credentialsServiceClient,
             final ProtocolAdapterProperties config) {
-        this.authProvider = Objects.requireNonNull(authProvider, "Authentication provider cannot be null");
+        this.tenantServiceClient = Objects.requireNonNull(tenantServiceClient, "Tenant client cannot be null");
+        this.credentialsServiceClient = Objects.requireNonNull(credentialsServiceClient, "Credentials client cannot be null");
         this.config = Objects.requireNonNull(config, "configuration cannot be null");
     }
 
     @Override
     public ProtonSaslAuthenticator create() {
-        return new AmqpAdapterSaslAuthenticator(authProvider, config);
+        return new AmqpAdapterSaslAuthenticator(tenantServiceClient, credentialsServiceClient, config);
     }
 
     /**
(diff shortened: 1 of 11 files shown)


Alfusainey:
@sophokles73: can you please take a look at this?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorFactory.java
```diff
@@ -109,10 +126,17 @@ public void init(final NetSocket socket, final ProtonConnection protonConnection
             this.sasl = transport.sasl();
             sasl.server();
             sasl.allowSkip(false);
-            // Only username/password authentication is supported for now...
-            sasl.setMechanisms(AuthenticationConstants.MECHANISM_PLAIN);
+            sasl.setMechanisms(AuthenticationConstants.MECHANISM_PLAIN, AuthenticationConstants.MECHANISM_EXTERNAL);
             if (socket.isSsl()) {
-                LOG.debug("Client connected through a secured port. Ignoring peer certificates (not supported)");
+                LOG.debug("Client connected through a secured port");
```
sophokles73: trace

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorFactory.java
```diff
@@ -109,10 +126,17 @@ public void init(final NetSocket socket, final ProtonConnection protonConnection
...
+                try {
+                    peerCertificateChain = socket.sslSession().getPeerCertificates();
+                    clientCertAuthProvider = new X509AuthProvider(credentialsServiceClient, config);
+                    certValidator = new DeviceCertificateValidator();
```
sophokles73: do we need to create new instances for each device connecting?

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterTestBase.java
```diff
@@ -358,6 +359,7 @@ public void testUploadingXnumberOfMessages(final TestContext context) throws Int
 
         final Future<ProtonConnection> result = Future.future();
         final ProtonClientOptions options = new ProtonClientOptions();
+        options.addEnabledSaslMechanism(ProtonSaslPlainImpl.MECH_NAME);
```
sophokles73: what about a test for a client using SASL_EXTERNAL?

Alfusainey: i want to add that later for a device that wants to authenticate using a client certificate. the current test code base does not use any certificate.. 

`SASL_PLAIN` is added mainly so that the current ITs passes. WDYT?

sophokles73: I think that it would be good to have an integration test that uses SASL_EXTERNAL :-)

sophokles73: BTW you should rebase so that the Travis build works again ...

Alfusainey:
> I think that it would be good to have an integration test that uses SASL_EXTERNAL :-)

ok, i will include that in this PR

> BTW you should rebase so that the Travis build works again ...

just rebased, thanks!

Alfusainey:
@sophokles73: patch updated with c340b99. this commit also includes minor refactoring of the HTTP adapter code to reuse its trust options config.

adapters/http-vertx/src/main/java/org/eclipse/hono/adapter/http/vertx/VertxBasedHttpProtocolAdapter.java
```diff
@@ -85,27 +83,9 @@ public void setClientCertAuthProvider(final HonoClientBasedAuthProvider provider
...
-
-        return Optional.ofNullable(getConfig().getTrustOptions())
-                .orElseGet(() -> {
-                    if (getConfig().isAuthenticationRequired()) {
-                        return new ValidityBasedTrustOptions();
-                    } else {
-                        return null;
-                    }
-                });
+        return getTrustOptions();
```
sophokles73: I do not see why we need to introduce another level of indirection. Can't we simply pull up this method *as is*?

tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
```diff
@@ -101,6 +106,7 @@
     public static final int    MQTT_PORT = Integer.getInteger(PROPERTY_MQTT_PORT, DEFAULT_MQTT_PORT);
     public static final String AMQP_HOST = System.getProperty(PROPERTY_AMQP_HOST, DEFAULT_HOST);
     public static final int    AMQP_PORT = Integer.getInteger(PROPERTY_AMQP_PORT, DEFAULT_AMQP_PORT);
+    public static final int    AMQPS_PORT = Integer.getInteger(PROPERTY_AMQPS_PORT, 4041);
```
sophokles73: for consistency, can we use a constant for the default AMQPS port as well?

tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
```diff
@@ -130,14 +136,14 @@
 
     private final Set<String> tenantsToDelete = new HashSet<>();
     private final Map<String, Set<String>> devicesToDelete = new HashMap<>();
-    private final Vertx vertx;
+    private static Vertx VERTX;
```
sophokles73: we use upper case naming for constants only

sophokles73:
@Alfusainey  why doesn't the Travis build succeed?

adapters/http-vertx/src/main/java/org/eclipse/hono/adapter/http/vertx/VertxBasedHttpProtocolAdapter.java
Alfusainey: if you say "pull up", you mean having it as a default implementation for all protocol adapters?

Alfusainey:
> @Alfusainey why doesn't the Travis build succeed?

humm... looks like the vertx thread has been blocked for a long time... am suspecting that it does not directly relate to my changes:

> 	at java.base@10.0.2/java.util.concurrent.CompletableFuture.waitingGet(CompletableFuture.java:1823)
	at java.base@10.0.2/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1998)
	at app//io.vertx.ext.unit.impl.CompletionImpl.await(CompletionImpl.java:56)
	at app//org.eclipse.hono.tests.IntegrationTestSupport.lambda$deleteObjects$4(IntegrationTestSupport.java:206)
	at app//org.eclipse.hono.tests.IntegrationTestSupport$$Lambda$245/1193841543.accept(Unknown Source)
	at java.base@10.0.2/java.lang.Iterable.forEach(Iterable.java:75)
	at app//org.eclipse.hono.tests.IntegrationTestSupport.deleteObjects(IntegrationTestSupport.java:203)
	at app//org.eclipse.hono.tests.mqtt.MqttTestBase.lambda$postTest$1(MqttTestBase.java:121)

i think when deleting the temp devices, the call to `deletion.await()` does not return on time

adapters/http-vertx/src/main/java/org/eclipse/hono/adapter/http/vertx/VertxBasedHttpProtocolAdapter.java
sophokles73: Don't you think that would be a reasonable default?

Alfusainey:
@sophokles73: patch updated with f4ca7ce. now the `vertx` instance in the `IntegrationTestSupport` class is not static. i suspect this was failing the java10 build (but am not entirely sure)

Alfusainey:
@sophokles73: something very strange is happening with the jdk10 travis build. it takes longer to run and the MQTT adapter ITs is failing the build. i was thinking that my changes to the `IntegrationTestSupport` class fails the build but that is not the case

Alfusainey:
@sophokles73: i rebased this patch with current master to include 008e20c. hopefully the fix passes the build :+1:

sophokles73:
yeah, let's hope so ...

Alfusainey:
@sophokles73: the build passes :man_dancing:

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterTestBase.java
```diff
@@ -369,32 +426,53 @@ public void testUploadingXnumberOfMessages(final TestContext ctx) throws Interru
...
-                        result.fail(remoteOpen.cause());
-                    }
-                });
-                connection.closeHandler(remoteClose -> {
-                    connection.close();
-                });
-                connection.open();
-            }
-        });
+        if (!Strings.isNullOrEmpty(username) && !Strings.isNullOrEmpty(password)) {
```
sophokles73: please split this up into two methods: *connectToAdapter(String, String)* and *connectToAdapter(SelfSignedCertificate)* to make it more transparent in the client code which mechanism is used. There is basically no duplicated code so it shouldn't be an issue ...

Alfusainey: ok :+1:

sophokles73:

Alfusainey:
@sophokles73 : done!

Issue #696, opened 2018-07-06
Deadlock in HTTP Adapter of Hono 0.6-Final 


Christian-Schmid:
Hi Hono Devs,

we're running Hono 0.6 (Release) in a kubernetes cluster.
Sometimes we are encountering problems with an unresponsive Hono HTTP Adapter. 

The last time we took a thread dump of the pod, which shows a java deadlock:

```
Found one Java-level deadlock:
=============================
"vert.x-eventloop-thread-4":
  waiting to lock monitor 0x00007fc1c000d0a8 (object 0x00000000f07a7c40, a io.vertx.ext.web.impl.RouteImpl),
  which is held by "vert.x-eventloop-thread-0"
"vert.x-eventloop-thread-0":
  waiting to lock monitor 0x00007fc1c408c318 (object 0x00000000f07a8fc0, a io.vertx.ext.web.impl.RouteImpl),
  which is held by "vert.x-eventloop-thread-4"
```
I'm attaching the log file and thread dump:
[logs-from-hono-adapter-http-vertx-in-hono-adapter-http-vertx.txt](https://github.com/eclipse/hono/files/2170657/logs-from-hono-adapter-http-vertx-in-hono-adapter-http-vertx.txt)
[threadDump.log](https://github.com/eclipse/hono/files/2170659/threadDump.log)
 
Sadly I don't have concrete steps to reproduce the issue yet. But maybe someone else experienced  something similar or got a hunch what could be the problem.

sophokles73:
I have started to look into this. I am quite sure what the problem is (thanks to the thread dump) but I am not (yet) sure how to properly address the problem.

sophokles73:
The problem seems to be that we *switch* the vert.x `Context` while processing requests. When a request comes in it is handled on the event loop context that has been created when the HTTP adapter verticle gets deployed. This Context is associated with one of the event loop threads, say *event-loop-thread-0*.

When Spring instantiates the HonoClient instances for connecting to the registry and AMQP network, these instances all create a new vert.x Context to run on in their constructor. So, let's say the client for accessing the Tenant service runs on a Context that is associated with *event-loop-thread-1*.

[...]

PR #749, opened 2018-08-02
Protocol adapter liveness check to detect blocked vert.x event loop.

chapitos:
Signed-off-by: Marko Pascan <marko.pascan@bosch-si.com>

Currently no liveness checks are registered for protocol adapters. In our use case, where Hono services are running as part of Kubernetes cluster, this could lead to problems if a adapter runs into a deadlock (see https://github.com/eclipse/hono/issues/696). The expectation here is that a liveness check would detect a blocked/deadlocked protocol adapter and restart the Kubernetes POD.

The implemented liveness check procedure tries to execute an action on the protocol adapter context an in this way detect whether the vert.x event loop is blocked. If the event loop is blocked, the procedure will not complete within the defined timeout, thus causing failure of the liveness check.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
+++ b/core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
@@ -22,6 +22,7 @@ public class ProtocolAdapterProperties extends ServiceConfigProperties {
     private boolean authenticationRequired = true;
     private boolean jmsVendorPropsEnabled = false;
     private boolean defaultsEnabled = true;
+    private long eventLoopBlockedCheckTimeout = 5000L;
 
     /**
      * Checks whether the protocol adapter always authenticates devices using their provided credentials as defined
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
```diff
@@ -130,4 +131,27 @@ public boolean isDefaultsEnabled() {
...
+
+    /**
+     * Gets the timeout value used by protocol adapter liveness check,
+     * which determines if protocol adapter vert.x event loop is blocked.
+     * <p>
+     * Default value of the timeout is 5000 milliseconds.
+     *
+     * @return The timeout value in milliseconds.
+     */
+    public final long getEventLoopBlockedCheckTimeout() {
```
sysexcontrol: IMHO the final must be eliminated here. Or is there any reason to prevent specifically this method to be overwritten by childs?

chapitos: No specific reason. Removed 'final'.

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
```diff
@@ -140,7 +140,7 @@ public void setDefaultsEnabled(final boolean flag) {
      *
      * @return The timeout value in milliseconds.
      */
-    public final long getEventLoopBlockedCheckTimeout() {
+    public long getEventLoopBlockedCheckTimeout() {
```
sysexcontrol: @chapitos : I had another look into the class, and sorry, I think it should be the other way round:
both get and set should be final (matching all the other getters and setters in the property-classes).
The only exception is the `defaultsEnabled` property, but shouldn't be part of this PR.
It usually just doesn't make sense to overwrite a simple getter and setter.

Can you please make both of them final and push a new commit?
Thanks!

chapitos: Done.

chapitos:
@sysexcontrol : commits squashed.

PR #755, opened 2018-08-03
[#643] AMQP adapter CLI

Alfusainey:
@sophokles73 This integrates the command-line for the amqp adapter into the hono `cli` module.

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -30,6 +30,11 @@
 
     <dependencies>
         <dependency>
+            <groupId>commons-cli</groupId>
+            <artifactId>commons-cli</artifactId>
+            <version>1.4</version>
+        </dependency>
+        <dependency>
             <groupId>org.eclipse.hono</groupId>
             <artifactId>hono-client</artifactId>
         </dependency>
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -30,11 +30,6 @@
 
     <dependencies>
         <dependency>
-            <groupId>commons-cli</groupId>
-            <artifactId>commons-cli</artifactId>
-            <version>1.4</version>
-        </dependency>
-        <dependency>
             <groupId>org.eclipse.hono</groupId>
             <artifactId>hono-client</artifactId>
         </dependency>
--- a/cli/src/main/resources/application.yml
+++ b/cli/src/main/resources/application.yml
@@ -2,7 +2,7 @@ spring:
   jmx:
     enabled: false
   profiles:
-    active: receiver,ssl
+    active: receiver,ssl,amqp-adapter-cli
 
 hono:
   client:
--- a/cli/src/main/resources/application.yml
+++ b/cli/src/main/resources/application.yml
@@ -44,3 +44,9 @@ hono:
   client:
     hostnameVerificationRequired: false
     trustStorePath: target/config/hono-demo-certs-jar/trusted-certs.pem
+
+---
+
+spring:
+  profiles: amqp-adapter-cli
+    
--- a/cli/src/main/resources/application.yml
+++ b/cli/src/main/resources/application.yml
@@ -49,4 +49,4 @@ hono:
 
 spring:
   profiles: amqp-adapter-cli
-    
+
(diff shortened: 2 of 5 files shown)


cli/pom.xml
```diff
@@ -29,6 +29,11 @@
     <description>Hono Command line Interface</description>
 
     <dependencies>
+        <dependency>
+            <groupId>commons-cli</groupId>
+            <artifactId>commons-cli</artifactId>
```
sophokles73: I do not really understand what we need this for. Spring Boot supports binding command line params to fields or beans out-of-the-box ...

Alfusainey:
thanks... 183c8c2 removes the `commons-cli` dependency

cli/src/main/resources/application.yml
```diff
@@ -44,3 +44,9 @@ hono:
   client:
     hostnameVerificationRequired: false
     trustStorePath: target/config/hono-demo-certs-jar/trusted-certs.pem
+
+---
+
+spring:
+  profiles: amqp-adapter-cli
+
```
sophokles73: missing EOL

Alfusainey:
@sophokles73: PR updated

cli/src/main/resources/application.yml
```diff
@@ -2,7 +2,7 @@ spring:
   jmx:
     enabled: false
   profiles:
-    active: receiver,ssl
+    active: receiver,ssl,amqp-adapter-cli
```
sophokles73: My understanding is that this property defines the profiles to be active by default. FMPOV it doesn't make sense to activate receiver together with adapter client, does it?

cli/src/main/resources/application.yml
```diff
@@ -44,3 +44,9 @@ hono:
   client:
     hostnameVerificationRequired: false
     trustStorePath: target/config/hono-demo-certs-jar/trusted-certs.pem
+
+---
+
+spring:
+  profiles: amqp-adapter-cli
```
sophokles73: you should add some defaults to be used when running the application as AMQP adapter client, e.g. the default port to connect to ...

sophokles73:
@Alfusainey can you also paste a command line here for how you would actually run the client?

cli/src/main/resources/application.yml
Alfusainey: actually, i was wondering about this... i want to completely remove the `amqp-adapter-cli` profile and activate it using the `--spring.profiles.active` switch. WDYT?

cli/src/main/resources/application.yml
Alfusainey: ok

Alfusainey:
> @Alfusainey can you also paste a command line here for how you would actually run the client?

spring-boot:run -Drun.arguments=--message.address=telemetry/DEFAULT_TENANT/4712,--amqp.host=localhost,--amqp.port=4040,--username=sensor1@DEFAULT_TENANT,--password=hono-secret,--message='{"temp": 5}'

yes, i agree.. its a bit verbose :-)

sophokles73:
@Alfusainey you could put all of those values as default values into the yaml file ...

cli/src/main/resources/application.yml
sophokles73: >  i want to completely remove the amqp-adapter-cli profile and activate it using the --spring.profiles.active switch. WDYT?

FMPOV you simply need to remove `amqp-adapter-cli` from this line and then you will be able to do exactly what you want to do, i.e. using the `--spring.profiles.active` switch ...

Alfusainey:
> @Alfusainey you could put all of those values as default values into the yaml file ...

sure.

sophokles73:
This actually starts up a maven build, which is quite slow. A better alternative is to run the jar file created in the target folder:
```
java target/hono-cli-0.7-SNAPSHOT.jar --message.address=telemetry/DEFAULT_TENANT/4712 --amqp.host=localhost --amqp.port=4040 --username=sensor1@DEFAULT_TENANT --password=hono-secret --message='{"temp": 5}'
```

Alfusainey:
@sophokles73: with 5ad5e8d defaults values for the client is added in the YAML file

Alfusainey:
> This actually starts up a maven build, which is quite slow

okay, i will use the bundled JAR archive in the example docs. the `amqp-adapter-cli` profile is not active by default. how about having users add it via the `--spring.profiles.active` switch?

Issue #751, opened 2018-08-03
getTypesafeValueForField not working as intended


pho-enix:
I just dropped some code there, illustrating my issue:
https://github.com/eclipse/hono/pull/745

Probably nobody spotted it cause the build is (intentionally) broken.

sophokles73:
In fact, your test case illustrates that `EventBusService.getRequestPayload()` does not work as expected. The root cause for this indeed is an erroneous assumption about the way that generics work in Java in *getTypesafeValueForField*.

However, the question is: do you need a working *getRequestPayload* method or do you need a working *getTypesafeValueForField* method? Could you then please adapt the title of this issue accordingly?

pho-enix:
The function `getTypesafeValueForField`makes a promise: To get the value for a field in a typesafe way.
But it does not keep the promise and `getRequestPayload` (and other callers as well) will suffer from it.

I like the idea of `getTypesafeValueForField` a lot, as it makes the caller's code much easier and separates concerns. I very much prefer a solution that fixes `getTypesafeValueForField` so all callers can benefit from the improvement.

Does that answer your questions?

sophokles73:
No, it doesn't. My question is: do you have any code that fails due to it invoking *getRequestPayload* or do you have code that fails due to it (directly) invoking *getTypesafeValueForField*?

pho-enix:
I have code that fails invoking `getRequestPayload` (please have a look on the PR).

sophokles73:
I have taken a look at the PR, that's the reason why I ask. The title of this issue suggests that you have trouble invoking *getTypesafeValueForField* ...

ctron:
I any case, I think that `getTypesafeValueForField` is broken. I also do think it should be a static method.

The following test case would fail:

~~~java
@Test
public void testInvalidType() {
    final JsonObject device = new JsonObject().put("device-id", "someValue");

    final String stringValue = service.getTypesafeValueForField(device, "device-id");
    Assert.assertEquals("someValue", stringValue);

    final Integer intValue = service.getTypesafeValueForField(device, "device-id");
    Assert.assertNull(intValue);
}
~~~

According to the JavaDoc this should be `null`:

> @return The property value or {@code null} if no such property exists or is not of the expected type.

I think a proper solution would be:

~~~java
protected final <T> T getTypesafeValueForField(final Class<T> clazz, final JsonObject payload, final String field) {

    Objects.requireNonNull(clazz);
    Objects.requireNonNull(payload);
[...]

PR #757, opened 2018-08-03
[#751] Actually do a type check and add a unit test

ctron:


service-base/src/main/java/org/eclipse/hono/service/EventBusService.java
```diff
@@ -176,17 +176,23 @@ private void processRequestMessage(final Message<JsonObject> msg) {
...
 
+        Objects.requireNonNull(clazz);
         Objects.requireNonNull(payload);
         Objects.requireNonNull(field);
 
-        try {
-            return (T) payload.getValue(field);
-        } catch (final ClassCastException e) {
+        final Object result = payload.getValue(field);
+        if (result == null) {
```
SadatAnwar: This check is unnecessary, the `clazz.isInstance(result)` will return false if the object is null.

ctron: It indeed is.

balsmn:
Yes we have two service API's Credentials and Registration service based on `EventBusService`.

But I think the changes proposed here are ok and we can adapt our code. It should not be a problem.

sophokles73:
@ctron Based on @balsmn 's comment I think we  (i.e. you :-)) can merge this ...

Issue #756, opened 2018-08-03
Services can be flooded with requests


ctron:
Pushing "too many" requests towards the Hono HTTP protocol adapter make it behave rather poorly. My expectation would be that Hono would still be able to handle the amount it can handle, but it looks to me as if the HTTP adapter mostly breaks down.

I tried to dig a little bit into this and I think I know one reason why this could be. Pushing HTTP requests  make the HTTP protocol adapter issues requests towards the device registry. If such a request cannot be handled in a specific amount of time (default 200ms) then it will be aborted and the HTTP request failed.

My understanding of the message flow is the following:

![request handling](https://user-images.githubusercontent.com/202474/43648168-f907dc12-9739-11e8-9054-6c62f032e8d9.png)

<details>
title Request handling

Client->+Endpoint: Send request
note right of Endpoint: Validate Request
Endpoint->EventBus: Publish request
Endpoint-->-Client: Accept request

note over  Client: New credits are available

note right of EventBus: Process request
EventBus->Endpoint: Publish response

Endpoint->+Client: Send response
Client-->-Endpoint: Accept response
</details>

This would mean that the client (the protocol adapter) can push at a higher rate than the EventBus/RequestProcessor/Verticle can handle. Building up requests and slowing down the process.

That would also mean that, if requests stay longer than 200ms in some kind of queue on the EventBus side, then they would time out on the Protocol adapter side. So once your are in state that the EventBus queue is at a size that requests have to wait more than 200ms, then all requests will timeout. Which in fact means that no HTTP requests will be accepted any more.

A possible solution for that problem would be to delay the accept of the request until the response was send. The reject or release could still be sent directly after the validation of the request.

Another possible solution would be to get rid of the EventBus completely. As some initial profiler run showed that most of the time is spent processing EventBus messages.

This may as well relate to issue #735.

sophokles73:
> A possible solution for that problem would be to delay the accept of the request until the response was send. The reject or release could still be sent directly after the validation of the request.

I do not quite understand how this would help. If  the request messages take longer than 200 ms to process in the device registry, then the requests towards the registry will still time out in the adapter, because the response does not arrive in time (within 200 ms). The result should be the same: the adapter will reject HTTP requests with a 503, or am I mistaken?

Once the adapter runs out of credits from the registry, it should actually immediately return a 503 to clients without trying to invoke the registry.

Did you experience this same behavior using the "load test mode" of the registry, where it simply returns a response without looking up the data?

[...]

Issue #753, opened 2018-08-03
Memory leak in the RequestResponseEndpoint


ctron:
To me it looks like there is a memory leak in the device registry. After running for a while the device registry gets slower and slower. And the protocol adapters fail with 503, originating from the communication with the device registry.

Restarting the pod fixes the issue:

![image](https://user-images.githubusercontent.com/202474/43636358-23b0fffc-9712-11e8-8d63-c37bb2142399.png)

And running leak analysis on the heap dump seems to confirm this:

![image](https://user-images.githubusercontent.com/202474/43636381-3663c21a-9712-11e8-9452-7b51a9a792e0.png)

Further analysis is required how the happens.

ctron:
One possible cause for this could be `org.eclipse.hono.service.amqp.RequestResponseEndpoint.onLinkAttach(ProtonConnection, ProtonSender, ResourceIdentifier)`. It does register with the event bus and unregisters when the link is closed:

~~~java
final MessageConsumer<JsonObject> replyConsumer = vertx.eventBus().consumer(replyToAddress.toString(), message -> {
…
});
…
HonoProtonHelper.setCloseHandler(sender, senderClosed -> {
    logger.debug("client [{}] closed sender link, removing associated event bus consumer [{}]", sender.getName(), replyConsumer.address());
    replyConsumer.unregister();
    if (senderClosed.succeeded()) {
        senderClosed.result().close();
    }
});
sender.open();
~~~

However it doesn't seem to unregister when the session or transport gets closed. Which would happen every 10 minutes, when the service closes the connection due to the token expiration. Or, when the connection breaks due to other reasons.

Also does the method `org.eclipse.hono.service.amqp.AmqpServiceBase.closeExpiredConnection(ProtonConnection)` reset both the close and the disconnect handler.

sophokles73:
@ctron are you going to work on this? Otherwise, I could take a look, if you are stuck in the middle of something else. If you want me to try to fix this, simply assign the issue to me ...

ctron:
Fixed with PR #760

PR #760, opened 2018-08-06
[#735, #756] Fix service flooding and memory leak

ctron:
This PR will fix both #753 and #756.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/amqp/AmqpEndpoint.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/amqp/AmqpEndpoint.java
@@ -42,4 +42,15 @@ public interface AmqpEndpoint extends Endpoint {
      * @param sourceAddress The (remote) source address from the client's AMQP <em>ATTACH</em> message.
      */
     void onLinkAttach(ProtonConnection connection, ProtonSender sender, ResourceIdentifier sourceAddress);
+
+    /**
+     * Handles a closed connection.
+     * <p>
+     * This method is called whenever a connection got closed. Either actively but a call to disconnect, or by a
+     * broken/lost connection.
+     * </p>
+     * 
+     * @param connection The connection which got closed.
+     */
+    void onConnectionClosed(ProtonConnection connection);
 }
--- a/service-base/src/main/java/org/eclipse/hono/service/amqp/AmqpServiceBase.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/amqp/AmqpServiceBase.java
@@ -672,13 +672,15 @@ public abstract class AmqpServiceBase<T extends ServiceConfigProperties> extends
     }
 
     /**
-     * Is called whenever a proton connection was closed. The implementation is intentionally empty.
+     * Is called whenever a proton connection was closed.
      * <p>
-     * Subclasses should override this method to publish this as an event on the vertx bus if desired.
+     * Subclasses may override this method to publish this as an event on the vertx bus if desired. If they choose to
+     * override, they must however call this super method.
      *
      * @param con The connection that was closed.
      */
     protected void publishConnectionClosedEvent(final ProtonConnection con) {
+        endpoints.values().forEach(endpoint -> endpoint.onConnectionClosed(con));
     }
 
     /**
--- a/service-base/src/main/java/org/eclipse/hono/service/amqp/AmqpServiceBase.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/amqp/AmqpServiceBase.java
@@ -675,7 +675,8 @@ public abstract class AmqpServiceBase<T extends ServiceConfigProperties> extends
      * Is called whenever a proton connection was closed.
      * <p>
      * Subclasses may override this method to publish this as an event on the vertx bus if desired. If they choose to
-     * override, they must however call this super method.
+     * override, they must however call this super method as this method forwards the call to
+     * {@link AmqpEndpoint#onConnectionClosed(ProtonConnection)} of all endpoints managed by this instance.
      *
      * @param con The connection that was closed.
      */
--- a/service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
@@ -12,8 +12,11 @@
  *******************************************************************************/
 package org.eclipse.hono.service.amqp;
 
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
 import java.util.Objects;
-
+import java.util.Set;
 import org.apache.qpid.proton.amqp.transport.AmqpError;
 import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.auth.HonoUser;
@@ -28,6 +31,9 @@ import org.eclipse.hono.util.HonoProtonHelper;
 import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.ResourceIdentifier;
 import org.springframework.beans.factory.annotation.Autowired;
+
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
 
 import io.vertx.core.Future;
 import io.vertx.core.Vertx;
@@ -51,6 +57,11 @@ import io.vertx.proton.ProtonSender;
 public abstract class RequestResponseEndpoint<T extends ServiceConfigProperties> extends AbstractAmqpEndpoint<T> {
 
     private AuthorizationService authorizationService = new ClaimsBasedAuthorizationService();
+
+    private final Map<String, ProtonReceiver> replyToReceiverMap = new HashMap<>();
+    private final Multimap<ProtonConnection, MessageConsumer<?>> replyConsumerMap = HashMultimap.create();
+    private final Multimap<ProtonConnection, String> replyAddressMap = HashMultimap.create();
+    private final Set<String> replyAddresses = new HashSet<>();
 
     /**
      * Creates an endpoint for a Vertx instance.
@@ -124,14 +135,31 @@ public abstract class RequestResponseEndpoint<T extends ServiceConfigProperties>
         } else {
 
             logger.debug("establishing link for receiving messages from client [{}]", receiver.getName());
+
             receiver.setQoS(ProtonQoS.AT_LEAST_ONCE);
             receiver.setAutoAccept(true); // settle received messages if the handler succeeds
-            receiver.setPrefetch(config.getReceiverLinkCredit());
+
+            /*
+             * We do manual flow control, credits are replenished after responses have been sent.
+             */
+            receiver.setPrefetch(0);
+
+            // set up handlers
+
             receiver.handler((delivery, message) -> {
                 handleMessage(con, receiver, targetAddress, delivery, message);
             });
             HonoProtonHelper.setCloseHandler(receiver, clientDetached -> onLinkDetach(receiver));
+
+            // acknowledge the remote open
+
             receiver.open();
+
+            // send out initial credits, after opening
+
+            receiver.flow(config.getReceiverLinkCredit());
+
+            logger.debug("Flowing {} credits to the sender", config.getReceiverLinkCredit());
         }
     }
 
@@ -140,10 +168,10 @@ public abstract class RequestResponseEndpoint<T extends ServiceConfigProperties>
      * <p>
      * The message gets rejected if
      * <ul>
-     * <li>the message does not pass {@linkplain #passesFormalVerification(ResourceIdentifier, Message) formal verification}
-     * or</li>
-     * <li>the client is not {@linkplain #isAuthorized(HonoUser, ResourceIdentifier, Message) authorized to execute the operation}
-     * indicated by the message's <em>subject</em> or</li>
+     * <li>the message does not pass {@linkplain #passesFormalVerification(ResourceIdentifier, Message) formal
+     * verification} or</li>
+     * <li>the client is not {@linkplain #isAuthorized(HonoUser, ResourceIdentifier, Message) authorized to execute the
+     * operation} indicated by the message's <em>subject</em> or</li>
      * <li>its payload cannot be parsed</li>
      * </ul>
      * 
@@ -162,34 +190,65 @@ public abstract class RequestResponseEndpoint<T extends ServiceConfigProperties>
         } else {
             formalCheck.fail(new AmqpErrorException(AmqpError.DECODE_ERROR, "malformed payload"));
         }
+
         final HonoUser clientPrincipal = Constants.getClientPrincipal(con);
-        formalCheck.compose(ok -> isAuthorized(clientPrincipal, targetAddress, message)).compose(authorized -> {
 
-            logger.debug("client [{}] is {}authorized to {}:{}", clientPrincipal.getName(), authorized ? "" : "not ",
-                    targetAddress, message.getSubject());
-            if (authorized) {
-                try {
-                    processRequest(message, targetAddress, clientPrincipal);
-                    ProtonHelper.accepted(delivery, true);
+        final String replyTo = message.getReplyTo();
+
+        formalCheck
+
+                .compose(ok -> {
+
+                    if (!replyAddresses.contains(replyTo)) {
+                        return Future
+                                .failedFuture(new AmqpErrorException(AmqpError.ILLEGAL_STATE,
+                                        "unsubscribed reply-to address"));
+                    }
+
+                    allocateReceiverForReplyTo(replyTo, receiver);
+
                     return Future.succeededFuture();
-                } catch (final DecodeException e) {
-                    return Future.failedFuture(new AmqpErrorException(AmqpError.DECODE_ERROR, "malformed payload"));
-                }
-            } else {
-                return Future.failedFuture(new AmqpErrorException(AmqpError.UNAUTHORIZED_ACCESS, "unauthorized"));
-            }
+                })
 
-        }).otherwise(t -> {
+                .compose(ok -> isAuthorized(clientPrincipal, targetAddress, message))
 
-            if (t instanceof AmqpErrorException) {
-                final AmqpErrorException cause = (AmqpErrorException) t;
-                MessageHelper.rejected(delivery, cause.asErrorCondition());
-            } else {
-                logger.debug("error processing request [resource: {}, op: {}]: {}", targetAddress, message.getSubject(), t.getMessage());
-                MessageHelper.rejected(delivery, ProtonHelper.condition(AmqpError.INTERNAL_ERROR, "internal error"));
-            }
-            return null;
-        });
+                .compose(authorized -> {
+
+                    logger.debug("client [{}] is {}authorized to {}:{}", clientPrincipal.getName(),
+                            authorized ? "" : "not ", targetAddress, message.getSubject());
+
+                    if (authorized) {
+                        try {
+                            processRequest(message, targetAddress, clientPrincipal);
+                            ProtonHelper.accepted(delivery, true);
+                            return Future.succeededFuture();
+                        } catch (final DecodeException e) {
+                            return Future
+                                    .failedFuture(new AmqpErrorException(AmqpError.DECODE_ERROR, "malformed payload"));
+                        }
+                    } else {
+                        return Future
+                                .failedFuture(new AmqpErrorException(AmqpError.UNAUTHORIZED_ACCESS, "unauthorized"));
+                    }
+
+                })
+
+                .otherwise(t -> {
+
+                    flowCreditToRequestor(replyTo);
+
+                    if (t instanceof AmqpErrorException) {
+                        final AmqpErrorException cause = (AmqpErrorException) t;
+                        MessageHelper.rejected(delivery, cause.asErrorCondition());
+                    } else {
+                        logger.debug("error processing request [resource: {}, op: {}]: {}", targetAddress,
+                                message.getSubject(), t.getMessage());
+                        MessageHelper.rejected(delivery,
+                                ProtonHelper.condition(AmqpError.INTERNAL_ERROR, "internal error"));
+                    }
+
+                    return null;
+                });
     }
 
     /**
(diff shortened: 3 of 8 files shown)


service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
```diff
@@ -162,34 +179,95 @@ protected final void handleMessage(final ProtonConnection con, final ProtonRecei
...
-                MessageHelper.rejected(delivery, cause.asErrorCondition());
-            } else {
-                logger.debug("error processing request [resource: {}, op: {}]: {}", targetAddress, message.getSubject(), t.getMessage());
-                MessageHelper.rejected(delivery, ProtonHelper.condition(AmqpError.INTERNAL_ERROR, "internal error"));
-            }
-            return null;
-        });
+
+        final String replyTo = message.getReplyTo();
+        allocateReceiverForReplyTo(replyTo, receiver);
```
sophokles73: In theory, a client should always use the same reply-to address for messages sent over this link. However, nothing prevents the client from opening multiple receiver links and sending messages using arbitrary reply-to addresses that map to its multiple receiver links. We should either try to prevent that (how?) or we should also keep track of the reply-to addresses per receiver so that we can remove the entries from the reply-to/receiver map more efficiently when a receiver is closed/disconnected, i.e. without having to iterate over all of the map's entries.
WDYT?

ctron: If a client would use multiple reply-to addresses, then right now, I think it should still work fine. As each receiver will replenish credits properly once the request has been processed.

Of course it might get interesting when the client starts re-using the same reply-to address with several connections. In this case it wouldn't be possible anymore look up the correct receiver instance to replenish.

I am not so sure another map will help much here. You would need to re-lookup all elements again in the reply-to/receiver map. On the other hand, you could simply iterate over the entrySet of the map and kick out receivers by their connection:

~~~java
private void deallocateAllReceiversForConnection(final ProtonConnection connection) {
    replyToReceiverMap
            .entrySet()
            .removeIf(entry -> entry.getValue().getSession().getConnection() == connection);
}
~~~

ctron: On theory client could also re-use the same reply-to address. That would probably cause much more problems, as they could steal credits from each other.

The problem is, you could only really mitigate this by doing a check, if the reply-to address doesn't have any existing receiver set with a different connection. Otherwise rejecting the request. But that would make each request a bit more expensive.

sophokles73: > The problem is, you could only really mitigate this by doing a check, if the reply-to address doesn't have any existing receiver set with a different connection. Otherwise rejecting the request. But that would make each request a bit more expensive.

This would only need to be checked when the client opens its receiver link for the reply-to address, so it would need to be done only seldomly ....

sophokles73: > I am not so sure another map will help much here. You would need to re-lookup all elements again in the reply-to/receiver map. On the other hand, you could simply iterate over the entrySet of the map and kick out receivers by their connection

You are probably right, since we are not talking about too many clients anyways. the clients are basically protocol adapter instances, so we are probably talking dozens to several hundred entries only...

ctron: > > The problem is, you could only really mitigate this by doing a check, if the reply-to address doesn't have any existing receiver set with a different connection. Otherwise rejecting the request. But that would make each request a bit more expensive.
>
> This would only need to be checked when the client opens its receiver link for the reply-to address, so it would need to be done only seldomly ....

I don't think that is correct, or maybe I didn't understand what you had in mind.

When you have multiple receivers (on the endpoint side), receiving messages with the same reply-to address, then only one of the receivers will receive new credits, and too many. All others will be stuck. And this can only be checked when the incoming request is received.

ctron: Ok ... I understand ... well, that assumes that the client wouldn't send any request if the open of its receiver link failed. However, we are currently talking about a misbehaving client. So IMHO the problem should be handled at the earliest possible point, when the request is received. At that point it should be checked if the request may be processed, or rejected.

sophokles73: then we also need to keep track of the client's receiver(s) and the source address they have been opened for. Then you could check for every incoming message whether there is a corresponding receiver link ...

service-base/src/main/java/org/eclipse/hono/service/amqp/AmqpEndpoint.java
```diff
@@ -42,4 +42,11 @@
      * @param sourceAddress The (remote) source address from the client's AMQP <em>ATTACH</em> message.
      */
     void onLinkAttach(ProtonConnection connection, ProtonSender sender, ResourceIdentifier sourceAddress);
+
+    /**
+     * Handles a lost connection.
```
sophokles73: IMHO it would be helpful if we could describe in more detail under which circumstances this method will be invoked.

service-base/src/main/java/org/eclipse/hono/service/amqp/AmqpEndpoint.java
```diff
@@ -42,4 +42,11 @@
...
+     * 
+     * @param connection The connection which is lost.
+     */
+    void onConnectionLost(ProtonConnection connection);
```
sophokles73: Can we rename this method to something more specific to the circumstances under which it will be called, e.g. `onClientDisconnect`?

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
```diff
@@ -237,56 +324,98 @@ protected final void handleMessage(final ProtonConnection con, final ProtonRecei
...
+        deallocateAllReceiversForConnection(connection);
+    }
+
+    private void registerConsumerForConnection(final ProtonConnection connection,
+            final MessageConsumer<JsonObject> replyConsumer) {
+        replyConsumerMap.put(connection, replyConsumer);
+    }
+
+    private void unregisterConsumerForConnection(final ProtonConnection connection,
+            final MessageConsumer<JsonObject> replyConsumer) {
```
sophokles73: `MessageConsumer<?>` ?

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
```diff
@@ -237,56 +324,98 @@ protected final void handleMessage(final ProtonConnection con, final ProtonRecei
...
+    private final Multimap<ProtonConnection, MessageConsumer<?>> replyConsumerMap = HashMultimap.create();
+
+    @Override
+    public void onConnectionLost(final ProtonConnection connection) {
+        unregisterAllConsumersForConnection(connection);
+        deallocateAllReceiversForConnection(connection);
+    }
+
+    private void registerConsumerForConnection(final ProtonConnection connection,
+            final MessageConsumer<JsonObject> replyConsumer) {
```
sophokles73: `MessageConsumer<?>`?

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
ctron: I am not sure I understand what you have in mind. Assuming the following scenario:

* Client A opens receiver to "reply-to-A"
* Client A opens sender to "request"
* Client B open sender to "request" with reply-to="reply-to-A"
* Client B sends request

How would that help?

service-base/src/main/java/org/eclipse/hono/service/amqp/AmqpEndpoint.java
ctron: I did change the name to `onConnectionClosed`, which aligns with the rest of the methods named like `publishConnectionClosedEvent`, which effectively calls this method.

`onClientDisconnect` sounds like some active action of a requested disconnection. Which may or may not be the case. "connection closed" or "connection lost" indicates a more passive scenario to me. Of course "lost" would most likely imply an unexpected loss of connection. Where "closed" is more neutral.

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
ctron: I implemented that change anyway. However I am not sure that it helps much in the described scenario.

sophokles73: It helps with the following scenario 

>  Ok ... I understand ... well, that assumes that the client wouldn't send any request if the open of its receiver link failed. However, we are currently talking about a misbehaving client. So IMHO the problem should be handled at the earliest possible point, when the request is received. At that point it should be checked if the request may be processed, or rejected.

1. Client A opens receiver for *reply-to-A*
2. Client A opens sender with target address *registration/my-tenant*
3. Client A sends message with reply-to address *no-receiver-available*

sophokles73: > But I don't see how to fix the other issue

what is *the other issue*?

ctron: > what is the other issue?

* Client A opens receiver to "reply-to-A"
* Client A opens sender to "request"
* Client B open senders to "request"
* Client B sends request with reply-to="reply-to-A"

sophokles73: IMHO we could address this by checking if the corresponding links belong to the same session.

ctron: I am not sure if I would do that. If we decide to put the service behind a qpid router, then that would no longer work, would it?

sophokles73: maybe not

Can we try to focus on the original problem that we wanted to solve. We currently do not check if the links belongs together as well, so I would rather have a solution for the problem at hand in the 0.7 release and deal with other improvements later, ok?

ctron: Yes, I agree, we should clean that up at a later time.

So the PR should be fine in general. I think all the code style, documentation issues should be addresses as well. The only thing left is fixing the unit tests, or any issues which they bring up.

service-base/src/main/java/org/eclipse/hono/service/amqp/AmqpServiceBase.java
```diff
@@ -672,13 +672,15 @@ protected void handleSessionOpen(final ProtonConnection con, final ProtonSession
     }
 
     /**
-     * Is called whenever a proton connection was closed. The implementation is intentionally empty.
+     * Is called whenever a proton connection was closed.
      * <p>
-     * Subclasses should override this method to publish this as an event on the vertx bus if desired.
+     * Subclasses may override this method to publish this as an event on the vertx bus if desired. If they choose to
```
sophokles73: In order to know if I should override this method, it would be helpful to know, what the default implementation actually does

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
```diff
@@ -237,54 +296,179 @@ protected final void handleMessage(final ProtonConnection con, final ProtonRecei
...
+            return;
+        }
+
+        final String replyTo = replyToAddress.toString();
+
+        if (this.replyAddresses.contains(replyTo)) {
+            logger.debug("client [{}] wanted to subscribe to already subscribed reply-to address [{}]",
+                    sender.getName(), replyTo);
+            sender.setCondition(ProtonHelper.condition(AmqpError.ILLEGAL_STATE,
+                    String.format("reply-to address [%s] must not be subscribed yet", replyTo)));
```
sophokles73: can we change this to e.g.
> reply-to address [%s] already in use

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
```diff
@@ -237,54 +296,179 @@ protected final void handleMessage(final ProtonConnection con, final ProtonRecei
...
+                                final int status = ServiceInvocationException.extractStatusCode(t);
+                                return Future.succeededFuture(response.getResponse(status));
+                            })
+                            .map(filteredResponse -> {
+                                final Message amqpReply = getAmqpReply(filteredResponse);
+                                sender.send(amqpReply);
+                                flowCreditToRequestor(replyTo);
+                                return null;
+                            })
+                            .recover(t -> {
```
sophokles73: in which case is this supposed to be executed?

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
```diff
@@ -237,54 +296,179 @@ protected final void handleMessage(final ProtonConnection con, final ProtonRecei
...
+
+        registerConsumerForConnection(con, replyTo, replyConsumer);
+
+        sender.setQoS(ProtonQoS.AT_LEAST_ONCE);
+
+        HonoProtonHelper.setCloseHandler(sender, senderClosed -> {
+            logger.debug("client [{}] closed sender link, removing associated event bus consumer [{}]",
+                    sender.getName(), replyConsumer.address());
+
+            deallocateReceiverForReplyTo(replyToAddress.toString());
```
sophokles73: *replyTo* already contains the string representation

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
```diff
@@ -237,54 +296,179 @@ protected final void handleMessage(final ProtonConnection con, final ProtonRecei
...
+        replyAddresses.add(replyTo);
+
+    }
+
+    private void unregisterConsumerForConnection(final ProtonConnection connection,
+            final String replyTo, final MessageConsumer<?> replyConsumer) {
+
+        // unregister the consumer
+
+        replyConsumer.unregister();
```
sophokles73: AFAIK the `org.eclipse.hono.client.MessageConsumer` interface has no such method, or am I mistaken?

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
```diff
@@ -237,54 +296,179 @@ protected final void handleMessage(final ProtonConnection con, final ProtonRecei
...
+        replyAddresses.remove(replyTo);
+    }
+
+    private void unregisterAllConsumersForConnection(final ProtonConnection connection) {
+
+        // unregister all consumers the connection had registered
+
+        replyConsumerMap
+                .removeAll(connection)
+                .forEach(MessageConsumer::unregister);
```
sophokles73: unregister?

services/auth/src/main/java/org/eclipse/hono/service/auth/impl/AuthenticationEndpoint.java
```diff
@@ -83,6 +83,11 @@ public final void onLinkAttach(final ProtonConnection con, final ProtonReceiver
         super.onLinkAttach(con, receiver, targetResource);
     }
 
+    @Override
+    public void onConnectionClosed(final ProtonConnection connection) {
+        // we have nothing to clean up
```
sophokles73: IMHO this should go into a JavaDoc comment

services/messaging/src/main/java/org/eclipse/hono/messaging/MessageForwardingEndpoint.java
```diff
@@ -268,4 +268,8 @@ public void registerReadinessChecks(final HealthCheckHandler handler) {
             }
         });
     }
+
+    @Override
+    public void onConnectionClosed(final ProtonConnection connection) {
```
sophokles73: JavaDoc?

services/auth/src/main/java/org/eclipse/hono/service/auth/impl/AuthenticationEndpoint.java
ctron: None of the other public methods have Javadoc … I just tried to blend in :grin: … I'll add something.

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
ctron: That may be, but that is a `io.vertx.core.eventbus.MessageConsumer`, and it does compile.

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
ctron: Yes, unregister detaches the consumer from the event bus.

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
sophokles73: sure, but which method do you expect/suspect to fail there?

ctron: If I would expect a method to fail then I would do it differently :grin: 
But yes, the main reason is `getAmqpReply`.

ctron:
I am currently thinking about the section which replenishes the credits:

~~~java
.map(filteredResponse -> {
  final Message amqpReply = getAmqpReply(filteredResponse);
  sender.send(amqpReply);
  flowCreditToRequestor(replyTo);
  return null;
})
~~~

That actually doesn't wait for the disposition of the response:

~~~java
sender.send(amqpReply, dispostion -> {
  if (dispostion.remotelySettled()) {
    flowCreditToRequestor(replyTo);
  }
});
return null;
~~~

sophokles73:
@ctron what's the status of this PR? Do you need anything from me/others?

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
sophokles73: My understanding is that you want to make sure that we flow credit to the sender even if we are not able to create a response message properly (i.e. getAmypReply fails), right? If that is the case then I would really prefer to put the call to *getAmqpReply* and sender.send() into a try/catch block. That would make the intent much clearer FMPOV

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
```diff
@@ -237,54 +296,182 @@ protected final void handleMessage(final ProtonConnection con, final ProtonRecei
...
+        sender.setQoS(ProtonQoS.AT_LEAST_ONCE);
+
+        HonoProtonHelper.setCloseHandler(sender, senderClosed -> {
+            logger.debug("client [{}] closed sender link, removing associated event bus consumer [{}]",
+                    sender.getName(), replyConsumer.address());
+
+            deallocateReceiverForReplyTo(replyTo);
+            unregisterConsumerForConnection(con, replyTo, replyConsumer);
+
+            if (senderClosed.succeeded()) {
```
sophokles73: I don't see why we shouldn't close the sender if it contains an error message (which we don't seem to care about anyway).

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
```diff
@@ -237,54 +296,182 @@ protected final void handleMessage(final ProtonConnection con, final ProtonRecei
...
+    }
+
+    private void allocateReceiverForReplyTo(final String replyTo, final ProtonReceiver receiver) {
+
+        final ProtonReceiver oldReceiver = replyToReceiverMap.put(replyTo, receiver);
+
+        if (oldReceiver == null || oldReceiver == receiver) {
+            logger.debug("Allocated receiver [{}] for replies to [{}]", receiver, replyTo);
+        } else {
+            logger.info("Allocated receiver [{}] for replies to [{}] - Had existing receiver: [{}]", receiver, replyTo,
```
sophokles73: this means that we have multiple clients using the same reply-to address, right?

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
ctron: Like that?

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
ctron: Yes.

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
ctron: Well that comes from the original code. I kept it the same way.

sophokles73: can you change it anyway?

ctron: Done

service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
```diff
@@ -367,12 +366,10 @@ public final void onLinkAttach(final ProtonConnection con, final ProtonSender se
...
                     sender.getName(), replyConsumer.address());
 
-            deallocateReceiverForReplyTo(replyToAddress.toString());
+            deallocateReceiverForReplyTo(replyTo);
             unregisterConsumerForConnection(con, replyTo, replyConsumer);
 
-            if (senderClosed.succeeded()) {
-                senderClosed.result().close();
-            }
+            senderClosed.result().close();
```
sophokles73: IMHO this will yield a `NullPointerException` if the peer included an error in its detach frame. You will need to use `sender.close()` instead ...

ctron: 🤦 … True, I completely overlooked that.

PR #762, opened 2018-08-06
Add thank you page

ctron:


sophokles73:
Nice :-)

What about also adding the nice guys from Travis who provide us with free continuous integration services?

PR #773, opened 2018-08-09
docu: Correct disabled device for http adapter

pho-enix:
It is actually 404 NOT FOUND

### DIFF
--- a/site/content/user-guide/http-adapter.md
+++ b/site/content/user-guide/http-adapter.md
@@ -53,7 +53,7 @@ NB: There is a subtle difference between the *device identifier* (*device-id*) a
   * 401 (Unauthorized): The request cannot be processed because the request does not contain valid credentials.
   * 403 (Forbidden): The request cannot be processed because the device's registration status cannot be asserted. Possible reasons for this include:
         * The given tenant is not allowed to use this protocol adapter.
-        * The given device is disabled.
+  * 404 (Not Found): The request cannot be processed because the device is disabled or does not exist.
   * 503 (Service Unavailable): The request cannot be processed because there is no consumer of telemetry data for the given tenant connected to Hono.
 
 This is the preferred way for devices to publish telemetry data. It is available only if the protocol adapter is configured to require devices to authenticate (which is the default).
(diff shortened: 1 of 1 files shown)


site/content/user-guide/http-adapter.md
```diff
@@ -183,7 +183,7 @@ Publish some JSON data for device `4711`, indicating that the device will wait f
         * The tenant that the gateway belongs to is not allowed to use this protocol adapter.
         * The device belongs to another tenant than the gateway.
         * The gateway is not authorized to act *on behalf of* the device.
-        * The gateway is disabled.
+  * 404 (Not Found): The request cannot be processed because the device is disabled or does not exist.
```
sophokles73: have you actually tried the case of a disabled gateway? According to the Device Registration API this should indeed result in a 403. However, adding the 404 and its description is still correct...

site/content/user-guide/http-adapter.md
```diff
@@ -330,7 +330,7 @@ Publish some JSON data for device `4711`:
...
```
sophokles73: see above

site/content/user-guide/http-adapter.md
```diff
@@ -475,7 +475,7 @@ Send a response to a previously received command with the command-request-id `re
   * 403 (Forbidden): The request cannot be processed because the device's registration status cannot be asserted. Possible reasons for this might be:
         * The given tenant is not allowed to use this protocol adapter.
         * The given device does not belong to the given tenant.
-        * The given device is disabled.
+  * 404 (Not Found): The request cannot be processed because the device is disabled or does not exist.
```
sophokles73: see above

site/content/user-guide/http-adapter.md
pho-enix: No, I didn't check this with gateways.

sophokles73: Can you then please update the PR to keep the case of a disabled gateway in the list for the 403 status code? And please  don't forget the other two places where this applies as well ...

pho-enix: For gateway mode:
Gateway is disabled: 403
Device is disabled: 404

Changed PR accordingly.

Issue #785, opened 2018-08-15
Cannot use recent version of netty-tcnative-boringssl


sophokles73:
It turns out that the changes made in #698 lead to a situation where you can only use a concrete netty-tcnative library of the same version as is now included in the Docker images. The version currently included is netty-tcnative-2.0.10.Final. This means that if you want to use the BoringSSL implementation with the image, you are stuck to version 2.0.10.Final of that library as well. Using the most recent version (2.0.12) yields a segfault of the VM during startup. This is actually an issue because one of the advantages of using the PropertyLoader for adding (user provided) jars to the classpath is the flexibility it brings wrt the used version of the added libraries ...

ctron:
Well, that is an issue with version 2.0.11 of netty-tcnative. They did some kind of super-breaking change between micro version changes.

What I could imagine is that we enable the use of netty-tcnative (all variants) using a maven profile. I did use that when doing the scale testing, enabling the epoll netty implementation, and I think the same approach as also suggested for enabling jaeger. I think this would work here as well.

sophokles73:
@ctron but that means that we *do not* include `netty-tcnative` in the Docker images by default anymore, right?

ctron:
Yes. The profile would add whatever version of netty-tcnative you want. By default I wouldn't enable any profile regarding netty-tcnative. So you would have to explicitly opt-in to any of the variants.

One variant would be the current variant. But if you don't select anything (as the docker images would) then nothing of netty-tcnative would get added.

sophokles73:
@ctron  So, as a first step, in order to allow people to use a version > 2.0.11 we should first remove netty-tcnative from the Docker images. My understanding is that you only need/want it in the S2I builds for now. And there you are custom building netty-tcnative for the OopenSSL version at hand anyway, right?

ctron:
The OpenShift S2I only provides the JNI layer of netty-tcnative. So that alone wouldn't be enough.

sophokles73:
@ctron So, assuming that I remove the `netty-tcnative` dependency from hono-service-base, would you be able to do whatever it is you want to do in the S2I build?

ctron:
> @ctron So, assuming that I remove the netty-tcnative dependency from hono-service-base, would you be able to do whatever it is you want to do in the S2I build?

No.

I created PR #787 to show what would be required. The PR would remove netty-tcnative from the docker images, but would also allow it to be enabled using profiles. Which S2I could then make use of.

ctron:
[...]

PR #787, opened 2018-08-16
[#785] Move netty-tcnative into a profile

ctron:


ctron:
> now that the dependency is pulled in by a custom build only, wouldn't it make sense to allow the user to specify the version using a version as well?

You should be able to do that by using:

~~~
mvn … -Dnetty.tcnative.version=2.0.10.Final
~~~

As we use properties in the BOM. But I didn't try that.

sophokles73:
@ctron LGTM

Issue #783, opened 2018-08-14
Separate interfaces for implementing mandatory and optional service operations


dejanb:
Currently service interfaces define all operation that are defined by the API. Most of third-party registry implementations will only be interested in mandatory operations (as they have their own mechanisms for maintaining the data). It'd be good to separate mandatory and optional (CRUD) operations in separate interfaces, to ease implementation of third-party registers.

sophokles73:
@dejanb can you close this issue if you think it is fixed with https://github.com/eclipse/hono/commit/98e2c573310bf997809d51bbe55e25c1cd78fac0

jbtrystram:
maybe some other APIs that could benefit the same approach ?

sophokles73:
@dejanb @jbtrystram Oh, my bad, I thought this issue was only about the RegistrationService ...

@jbtrystram sure, additional PRs for the other service interfaces are welcome :-)

dejanb:
@sophokles73 @jbtrystram Yeah, I think we should do it for all "service" APIs:

* Credentials
* Tenants
* Authentication

Let's keep this issue as a placeholder for the work and do the work in the additional PRs

It'd be good to change the docs a bit as well, to better separate what's mandatory and what's optional for every API?

jbtrystram:
Credentials, tenants and registration API have been refactored. (PRs #802 and #795 are still to be merged.)
However, the Authentication API only expose one operation, which is mandatory. So nothing to separate there. 
Should we close this ?

sophokles73:
@jbtrystram Before closing this issue, we also need an entry in the [release notes](https://github.com/eclipse/hono/blob/master/site/content/release-notes.md) describing this API change.

jbtrystram:
@sophokles73 would that work ? 
#806

sophokles73:
@dejanb all changes have been made, so FMPOV you can close this issue ...

dejanb:
@sophokles73 Great. Thanks @jbtrystram

PR #789, opened 2018-08-20
[WIP]Device registry api refactor

jbtrystram:
This is some initial work toward https://github.com/eclipse/hono/issues/783
here is what I have came up with
Device Registry API refactor

This is WIP, as the relevant tests need to be adjusted still. Please feel free to comment and suggest improvements!

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/registration/CompleteBaseRegistrationService.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/registration/CompleteBaseRegistrationService.java
@@ -1,0 +1,184 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.hono.service.registration;
+
+import java.net.HttpURLConnection;
+import java.util.Objects;
+
+import org.eclipse.hono.client.ClientErrorException;
+import org.eclipse.hono.util.EventBusMessage;
+import org.eclipse.hono.util.RegistrationConstants;
+import org.eclipse.hono.util.RegistrationResult;
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * Base class for implementing {@link CompleteRegistrationService}s.
+ * <p>
+ * In particular, this base class provides support for receiving service invocation request messages
+ * via vert.x' event bus and route them to specific methods corresponding to the operation indicated
+ * in the message.
+ *
+ * @param <T> The type of configuration properties this service requires.
+ */
+public abstract class CompleteBaseRegistrationService<T> extends BaseRegistrationService<T> implements CompleteRegistrationService {
+
+    /**
+     * Processes a device registration API request received via the vert.x event bus.
+     * <p>
+     * This method validates the request parameters against the Device Registration API
+     * specification before invoking the corresponding {@code RegistrationService} methods.
+     *
+     * @param requestMessage The request message.
+     * @return A future indicating the outcome of the service invocation.
+     * @throws NullPointerException If the request message is {@code null}.
+     */
+    @Override
+    public final Future<EventBusMessage> processRequest(final EventBusMessage requestMessage) {
+
+        Objects.requireNonNull(requestMessage);
+
+        switch (requestMessage.getOperation()) {
+        case RegistrationConstants.ACTION_REGISTER:
+            return processRegisterRequest(requestMessage);
+        case RegistrationConstants.ACTION_ASSERT:
+            return super.processRequest(requestMessage);
+        case RegistrationConstants.ACTION_GET:
+            return processGetRequest(requestMessage);
+        case RegistrationConstants.ACTION_UPDATE:
+            return processUpdateRequest(requestMessage);
+        case RegistrationConstants.ACTION_DEREGISTER:
+            return processDeregisterRequest(requestMessage);
+        default:
+            return processCustomRegistrationMessage(requestMessage);
+        }
+    }
+
+    private Future<EventBusMessage> processRegisterRequest(final EventBusMessage request) {
+
+        final String tenantId = request.getTenant();
+        final String deviceId = request.getDeviceId();
+        final JsonObject payload = getRequestPayload(request.getJsonPayload());
+
+        if (tenantId == null || deviceId == null) {
+            return Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST));
+        } else {
+            log.debug("registering device [{}] for tenant [{}]", deviceId, tenantId);
+            final Future<RegistrationResult> result = Future.future();
+            addDevice(tenantId, deviceId, payload, result.completer());
+            return result.map(res -> {
+                return request.getResponse(res.getStatus())
+                        .setDeviceId(deviceId)
+                        .setCacheDirective(res.getCacheDirective());
+            });
+        }
+    }
+
+    private Future<EventBusMessage> processGetRequest(final EventBusMessage request) {
+
+        final String tenantId = request.getTenant();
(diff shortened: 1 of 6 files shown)


sophokles73:
@jbtrystram thanks for the effort. The diagram looks like most (all?) relations are pointing in the wrong direction ...

service-base/src/main/java/org/eclipse/hono/service/registration/CompleteBaseRegistrationService.java
```diff
@@ -0,0 +1,217 @@
...
+/**
+ * Base class for implementing {@link CompleteRegistrationService}s.
+ * <p>
+ * In particular, this base class provides support for receiving service invocation request messages
+ * via vert.x' event bus and route them to specific methods corresponding to the operation indicated
+ * in the message.
+ *
+ * @param <T> The type of configuration properties this service requires.
+ */
+public abstract class CompleteBaseRegistrationService<T> extends BaseRegistrationService implements CompleteRegistrationService {
```
dejanb: This should be `extends BaseRegistrationService<T>` to avoid erasure errors

service-base/src/main/java/org/eclipse/hono/service/registration/CompleteRegistrationService.java
```diff
@@ -0,0 +1,85 @@
...
+import io.vertx.core.Verticle;
+import io.vertx.core.json.JsonObject;
+import org.eclipse.hono.util.RegistrationResult;
+
+/**
+ * A service for keeping record of device identities.
+ * 
+ * @see <a href="https://www.eclipse.org/hono/api/device-registration-api/">Device Registration API</a>
+ */
+public interface CompleteRegistrationService extends Verticle, RegistrationService {
```
dejanb: We don't need to extend `Verticle` here as it's already implied from the `RegistrationService`

service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationService.java
```diff
@@ -117,40 +78,22 @@
...
+     *             is registered for the tenant.</li>
      *             </ul>
      * @throws NullPointerException if any of the parameters is {@code null}.
-     * @see <a href="https://www.eclipse.org/hono/api/device-registration-api/#deregister-device">
-     *      Device Registration API - Deregister Device</a>
+     * @see <a href="https://www.eclipse.org/hono/api/device-registration-api/#get-registration-information">
+     *      Device Registration API - Get Registration Information</a>
      */
-    void removeDevice(String tenantId, String deviceId, Handler<AsyncResult<RegistrationResult>> resultHandler);
+    void getDevice(String tenantId, String deviceId, Handler<AsyncResult<RegistrationResult>> resultHandler);
```
sophokles73: this operation is defined as optional in the API. Is there a particular reason why you want to keep it in this interface?

service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationService.java
```diff
@@ -22,53 +22,14 @@
 
 /**
  * A service for keeping record of device identities.
- * 
+ *
```
sophokles73: This description should probably be modified to reflect the fact, that the interface only contains the mandatory operations (if that is what you have in mind) ...

service-base/src/main/java/org/eclipse/hono/service/registration/CompleteRegistrationService.java
jbtrystram: yep, that make sense

service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationService.java
jbtrystram: it is used in the assertion process to retrieve the device and verify that a device is properly asserted. https://github.com/eclipse/hono/blob/fb942feacf74171d8b7507782bcbf3d4c1726c90/service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java#L311

We can put the method in the optional interface, but as it would be implemented anyway by the base class. In that case the Complete base class would respect the complete interface only because it extends the base class. (or duplicate code).
I don't know which case is better ?

sophokles73: FMPOV it belongs in the `CompleteRegistrationService` interface. We then simply remove the `@Override` annotation from *getDevice* in `BaseRegistrationService` and we are done, right?
I don't think it's a problem that CompleteBaseRegistrationService requires extension of BaseRegistrationService in order to fulfill its contract because we have designed it to an extension of it in the first place. It already needs to extend BaseRegistrationService in order to implement *assertRegistration* as well ...

jbtrystram: Sounds logical to me !

jbtrystram:
thanks @sophokles73 and @dejanb for the comments. 
I have made the according changes.

I'll push the according tests shortly

jbtrystram:
Here are the associated tests. The tests for `CompleteBaseRegistrationService` are somewhat dummy, as no optional API resources are really implemented (just returns HTTP error 501).

service-base/src/main/java/org/eclipse/hono/service/registration/CompleteBaseRegistrationService.java
```diff
@@ -0,0 +1,201 @@
...
+     * Subclasses should override this method in order to support additional, custom
+     * operations that are not defined by Hono's Device Registration API.
+     * <p>
+     * This default implementation simply returns a future that is failed with a
+     * {@link ClientErrorException} with an error code <em>400 Bad Request</em>.
+     *
+     * @param request The request to process.
+     * @return A future indicating the outcome of the service invocation.
+     */
+    protected Future<EventBusMessage> processCustomRegistrationMessage(final EventBusMessage request) {
```
sophokles73: since this is the same as in `BaseRegistrationService` I don't think that we need to override it, do we?

jbtrystram:
That make the _getDevice_  somewhat mandatory to implement then. Should we change the API accordingly?

sophokles73:
> That make the getDevice somewhat mandatory to implement then. Should we change the API accordingly?

That is the whole point, isn't it? With the `BaseRegistrationService`'s default implementation of the *getDevice* method, a subclass that doesn't override it will not work properly, i.e. any client invoking *assertRegistration* will always get a 501 in return, but never an assertion. Declaring the *getDevice* method as abstract simply makes this fact more transparent and allows the compiler to point out the problem. 

> Should we change the API accordingly?

Which API are your talking about? Hono's Device Registration API? I don't think so. The *getDevice* method of `BaseRegistrationService` is simply necessary to retrieve registration information from the persistence store (note that this method's name is totally arbitrary). However, Hono's protocol adapters are not interested in retrieving this information per se. They are only interested in getting a registration assertion.

jbtrystram:
It make complete sense. 
I was suggesting to change Hono's Device Registration API to avoid the confusion that may happen when someone would want to implement only the mandatory operations, and discovers that an optional operation need to be implemented in order to have the mandatory operation working. But as you pointed out, it's just needed to retrieve information internally, so it's almost a naming conflict somewhat.

jbtrystram:
Actually if one override the _assertRegistration_ method without the need of _getDevice_ there is no added-value in implementing it. (as in  https://github.com/eclipse/hono/blob/139ce8f020aeaf2ff1c05374258ca5ea91802e17/services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/DummyRegistrationService.java#L39 )

sophokles73:
@jbtrystram not sure what this is supposed to mean

sophokles73:
@jbtrystram  Ok, I see what you mean. So, we end up with keeping the (non-functional) default implementation of *getDevice*, right? In this case we need to make it very, very clear, that subclasses need to override (and implement) the *getDevice* method in order to get a functioning `RegistrationService` implementation. We should document this in `BaseregistrationService`'s class level JavaDoc. WDYT?

jbtrystram:
If the subclass do not override the original _assertRegistration_  then yes, _getDevice_ should be implemented (as in [the file based Registry](https://github.com/eclipse/hono/blob/139ce8f020aeaf2ff1c05374258ca5ea91802e17/services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedRegistrationService.java)
However, if they want to override _assertRegistration_ in a way that don't need _getDevice_, why not implement it would make the service non-functional ?

sophokles73:
In any case: it needs to be explicitly explained to users of `BaseRegistrationService` what they need to do in which case.

jbtrystram:
>  can you squash and rebase?
done !

PR #791, opened 2018-08-21
Free link resources for HonoClient if remote peer did not send detach.

sysexcontrol:
To prevent a link based memory leak, sender and receiver links are
currently freed in their close handlers. However, the close handler
is only called if the remote peer reacts with a detach frame after the
client closed the link. Now a timer is created additionally to
supervise if there should be no detach frame from the remote to free
the link resources then.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
+++ b/core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
@@ -13,6 +13,7 @@
 package org.eclipse.hono.util;
 
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicLong;
 
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Context;
@@ -26,6 +27,11 @@ import io.vertx.proton.ProtonSession;
  * Utility methods for working with Proton objects.
  */
 public final class HonoProtonHelper {
+
+    /**
+     * The default number of milliseconds to wait for a remote peer to send a detach frame after client closed a link.
+     */
+    public static final long DEFAULT_FREE_LINK_AFTER_CLOSE_INTERVAL_MILLIS = 3000;
 
     private HonoProtonHelper() {
         // prevent instantiation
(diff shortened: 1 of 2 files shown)


sysexcontrol:
@sophokles73 It seems that the call to free a link should not be done if there are still resources necessary (e.g. to handle a remote detach frame, which should be the normal behaviour).

The implemented solution to prevent a memory leak in the case that the remote peer for any reason does NOT send a detach (and thus the close handler is not invoked), is to supervise the close with a timer. Currently set to 3 seconds (which possibly must be configurable).
With this implementation all tests run fine.

What do you think of such a solution? It is more complex than hoped, but it should be safe against 
"misbehaving" peers.

core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
```diff
@@ -164,4 +165,44 @@ public static void setDefaultCloseHandler(final ProtonSession session) {
...
+            // but add a timer to free resources if peer did not detach
+            closeSenderTimerId.set(context.owner().setTimer(Constants.DEFAULT_FREE_LINK_AFTER_CLOSE_INTERVAL_MILLIS, timerId -> {
+                executeOrRunOnContext(context, v -> {
+                    freeLinkHandler.handle(null);
+                });
+            }));
+        } else {
+            // trigger handler manually to make sure that
+            // resources are freed up
+            freeLinkHandler.handle(null);
```
sophokles73: if we pass in the context then we should probably use it to run the handler on it, right?

core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
```diff
@@ -164,4 +165,44 @@ public static void setDefaultCloseHandler(final ProtonSession session) {
...
+     * Close the passed link and later call its {@link ProtonLink#free()} method in an attached closeHandler.
+     * <p>
+     * Additionally supervise if this handler is called in the number of {@link Constants#DEFAULT_FREE_LINK_AFTER_CLOSE_INTERVAL_MILLIS}
+     * milliseconds. If not, the timer is triggered and the link will be freed inside the timer code.
+     *
+     * @param context The context to run the timer code on.
+     * @param link The link to close and free.
+     * @param closeHandler The handler to invoke after the link was freed.
+     */
+    public static void closeAndFree(final Context context, final ProtonLink link, final Future<Void> closeHandler) {
```
sophokles73: I would appreciate an overloaded version that accepts the timeout in milliseconds as a parameter. FMPOV we should then also move the default timeout from Constants to this class since it will not need to be used anywhere else.

core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
```diff
@@ -164,4 +165,44 @@ public static void setDefaultCloseHandler(final ProtonSession session) {
...
+        link.closeHandler(closeAttempt -> {
+            context.owner().cancelTimer(closeSenderTimerId.get());
+            freeLinkHandler.handle(null);
+        });
+
+        if (link.isOpen()) {
+            // wait for peer's detach frame to trigger the handler
+            link.close();
+            // but add a timer to free resources if peer did not detach
+            closeSenderTimerId.set(context.owner().setTimer(Constants.DEFAULT_FREE_LINK_AFTER_CLOSE_INTERVAL_MILLIS, timerId -> {
```
sophokles73: IMHO we should start the timer *before* sending our detach frame in order to make sure that the timer ID is set when the remote detach frame arrives ...

core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
sysexcontrol: +1

sysexcontrol:
@sophokles73 added all changes to last commit.

PR #795, opened 2018-08-23
split credential API regarding mandatory and optional operations.

jbtrystram:
Same architecture as the device Registration API.


jbtrystram:
here it is !

Issue #771, opened 2018-08-08
Add Command and Control support in CLI


kaniyan:
Currently Hono-Cli doesn't support sending commands and receiving command-responses. Add this feature to CLI.

PR #796, opened 2018-08-23
[#771] Add Command and Control support in CLI

kaniyan:


cli/src/main/resources/application.yml
```diff
@@ -31,7 +31,7 @@ address:
     queryTimeout: 2000
 
 connection:
-  timeout: 1000
+  retryInterval: 1000
```
sophokles73: looks like you can pull this up to the top level

cli/src/main/java/org/eclipse/hono/cli/Commander.java
```diff
@@ -0,0 +1,162 @@
...
+    protected int requestTimeoutInSecs;
+    private WorkerExecutor workerExecutor;
+
+    /**
+     * Starts this component.
+     *
+     */
+    @PostConstruct
+    void start() {
+        workerExecutor = vertx.createSharedWorkerExecutor("user-input-pool", 3, TimeUnit.HOURS.toNanos(1));
```
sophokles73: out of curiosity, what's wrong with the default worker threads provided by vert.x?

cli/src/main/resources/application.yml
kaniyan: It makes sense and I pulled it up.

cli/src/main/java/org/eclipse/hono/cli/Commander.java
kaniyan: Default maximum execution time of default worker threads is 6000 ms. In this usecase of getting input from user using System.in, I wanted to increase it. Of course it can be changed using the vertxOptions. Then this changed value will be used by all the default working threads and named worker executors (that may be created). So I created a separate named worker executor and changed the maximum execution time of it, so that it remains within this scope.

Issue #607, opened 2018-05-03
Change AMQP data type in all APIs to Data (vs. AmqpValue)


pellmann:
We do not use the AMQP type system and also according to the AMQP spec. we should use Data when we want to set the content-type (as we do for JSON data). We already use Data for telementry and events.

This change will will affect all these APIs:
- Tenant API
- Device Registration API
- Credentials API
- Command & Control API

ctron:
So is there an agreement that we will do that? In that case I would suggest we fix the remaining parts for 0.7, as #765 shows this has a significant performance impact.

sophokles73:
@ctron I appreciate your enthusiasm for improving things all over the place. However, we already have enough on our plate for 0.7 (see project board https://github.com/eclipse/hono/projects/7). I would really appreciate, if we could focus on getting the stuff done that we already scheduled for 0.7 instead of adding more and more stuff. I have no problem discussing this issue *after* 0.7 is released and we find some time to think about what should be in 0.8 ...

ctron:
I simply think that the performance improvements may be significant. But ok, let's move this past 0.7.

sophokles73:
so, FMPOV there are two things we need to (re-)consider
1. The spec seems to advise to *only* set the content type to *opaque/binary* if the payload is explicitly known to contain *truly opaque* data,  i.e. if we know that the type of the data will be irrelevant to the consumer (my interpretation). 
2. The spec advises to *not* set a content type if the payload is *not* contained in a *Data* section.

The former should be addressed by another issue. Regarding the latter, we need to consider
* that this would constitute a breaking change of the Credentials, Tenant and Device Registration APIs
* that we want to allow those APIs to be implemented using a JMS provider, which has some implications regarding the supported message properties and payload types

sophokles73:
Digging a little deeper into the AMQP-JMS Mapping specification, it seems that a JMS provider is required to map a String valued AMQP Value into a JMS TextMessage (that's what we rely on today). It is also required to map a Data section to a TextMessage, if it contains one of the *well known* content types indicating a text message. *application/json* is one of those content types. So, my understanding is that if we want to change the API then we should do both change the section's type from *AMQP Value* to *Data* and require the content-type to be set to *application/json*.

ctron:
> … then we should do both change the section's type from *AMQP Value* to *Data* and require the content-type to be set to *application/json*.

[...]

PR #803, opened 2018-08-27
[#607]: Use 'Data' type for AMQP body section

ctron:
This PR switches to always use the the 'Data' type for AMQP body sections. As of now, it still can receive AmqpValue body sections of type byte[] or String in addition. However I am not sure if we should keep this.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
@@ -519,11 +519,11 @@ public abstract class AbstractRequestResponseClient<R extends RequestResponseRes
     /**
      * Creates a request message for a payload and sends it to the peer.
      * <p>
-     * This method simply invokes {@link #createAndSendRequest(String, Map, Buffer, Handler)}
-     * with {@code null} for the properties parameter.
+     * This method simply invokes {@link #createAndSendRequest(String, Map, Buffer, Handler)} with {@code null} for the
+     * properties parameter.
      * 
      * @param action The operation that the request is supposed to trigger/invoke.
-     * @param payload The payload to include in the request message as a an AMQP Value section.
+     * @param payload The payload to include in the request message as an AMQP Value section.
      * @param resultHandler The handler to notify about the outcome of the request.
      * @throws NullPointerException if any of action or result handler is {@code null}.
      */
@@ -538,11 +538,11 @@ public abstract class AbstractRequestResponseClient<R extends RequestResponseRes
     /**
      * Creates a request message for a payload and sends it to the peer.
      * <p>
-     * This method simply invokes {@link #createAndSendRequest(String, Map, Buffer, String, Handler, Object, Span)}
-     * with {@code null} for the properties, content type and cache key parameters.
+     * This method simply invokes {@link #createAndSendRequest(String, Map, Buffer, String, Handler, Object, Span)} with
+     * {@code null} for the properties, content type and cache key parameters.
      * 
      * @param action The operation that the request is supposed to trigger/invoke.
-     * @param payload The payload to include in the request message as a an AMQP Value section.
+     * @param payload The payload to include in the request message as an AMQP Value section.
      * @param resultHandler The handler to notify about the outcome of the request.
      * @param currentSpan The <em>Opentracing</em> span used to trace the request execution.
      * @throws NullPointerException if any of action, result handler or current span is {@code null}.
--- a/client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
@@ -22,7 +22,6 @@ import java.util.Optional;
 import java.util.UUID;
 
 import org.apache.qpid.proton.amqp.messaging.Accepted;
-import org.apache.qpid.proton.amqp.messaging.AmqpValue;
 import org.apache.qpid.proton.amqp.messaging.Rejected;
 import org.apache.qpid.proton.amqp.transport.DeliveryState;
 import org.apache.qpid.proton.message.Message;
@@ -699,12 +698,7 @@ public abstract class AbstractRequestResponseClient<R extends RequestResponseRes
 
         if (isOpen()) {
             final Message request = createMessage(action, properties);
-            if (payload != null) {
-                if (contentType != null) {
-                    request.setContentType(contentType);
-                }
-                request.setBody(new AmqpValue(payload.getBytes()));
-            }
+            MessageHelper.setPayload(request, contentType, payload);
             sendRequest(request, resultHandler, cacheKey, currentSpan);
         } else {
             TracingHelper.logError(currentSpan, "sender and/or receiver link is not open");
--- a/core/src/main/java/org/eclipse/hono/util/MessageHelper.java
+++ b/core/src/main/java/org/eclipse/hono/util/MessageHelper.java
@@ -173,6 +173,11 @@ public final class MessageHelper {
      */
     public static final String JMS_VENDOR_PROPERTY_CONTENT_TYPE = "JMS_AMQP_CONTENT_TYPE";
 
+    /**
+     * The MIME type representing the String representation of a JSON Object.
+     */
+    public static final String CONTENT_TYPE_APPLICATION_JSON = "application/json";
+
     private static final Logger LOG = LoggerFactory.getLogger(MessageHelper.class);
 
     private MessageHelper() {
@@ -682,39 +687,41 @@ public final class MessageHelper {
     }
 
     /**
-     * Set the payload of the message.
+     * Set the JSON payload of the message.
      * <p>
      * If the payload is {@code null}, then neither the payload, nor content type will be set.
      * </p>
      * 
      * @param message The message to update.
-     * @param contentType An optional content type.
      * @param payload The optional message payload.
      * 
      * @throws NullPointerException If the parameter {@code message} was {@code null}.
      */
-    public static void setPayload(final Message message, final String contentType, final JsonObject payload) {
+    public static void setJsonPayload(final Message message, final JsonObject payload) {
         Objects.requireNonNull(message);
 
-        setPayload(message, contentType, payload != null ? payload.toBuffer() : null);
+        setPayload(message, CONTENT_TYPE_APPLICATION_JSON, payload != null ? payload.toBuffer() : null);
     }
 
     /**
-     * Set the payload of the message.
+     * Set the JSON payload of the message.
      * <p>
      * If the payload is {@code null}, then neither the payload, nor content type will be set.
      * </p>
+     * <p>
+     * <b>Note:</b> No formal check is done if the payload actually is a JSON string.
+     * </p>
      * 
      * @param message The message to update.
-     * @param contentType An optional content type.
      * @param payload The optional message payload.
      * 
      * @throws NullPointerException If the parameter {@code message} was {@code null}.
      */
-    public static void setPayload(final Message message, final String contentType, final String payload) {
+    public static void setJsonPayload(final Message message, final String payload) {
         Objects.requireNonNull(message);
 
-        setPayload(message, contentType, payload != null ? payload.getBytes(StandardCharsets.UTF_8) : null);
+        setPayload(message, CONTENT_TYPE_APPLICATION_JSON,
+                payload != null ? payload.getBytes(StandardCharsets.UTF_8) : null);
     }
 
     /**
--- a/core/src/main/java/org/eclipse/hono/util/MessageHelper.java
+++ b/core/src/main/java/org/eclipse/hono/util/MessageHelper.java
@@ -12,18 +12,23 @@
  *******************************************************************************/
 package org.eclipse.hono.util;
 
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.HashMap;
 import java.util.Objects;
 import java.util.Optional;
 
 import io.vertx.core.buffer.Buffer;
+
+import org.apache.qpid.proton.amqp.Binary;
 import org.apache.qpid.proton.amqp.Symbol;
 import org.apache.qpid.proton.amqp.messaging.AmqpValue;
 import org.apache.qpid.proton.amqp.messaging.ApplicationProperties;
 import org.apache.qpid.proton.amqp.messaging.Data;
 import org.apache.qpid.proton.amqp.messaging.MessageAnnotations;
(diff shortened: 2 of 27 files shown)


core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -338,6 +344,44 @@ public static Buffer getPayload(final Message msg) {
         return null;
     }
 
+    /**
+     * Gets a message's body as Buffer object as {@link String}.
```
sophokles73: this seems to be garbled
also, there is no need to link to the String class, is there?

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -590,4 +634,133 @@ public static Boolean isDeviceCurrentlyConnected(final Message msg) {
...
+     * 
+     * @param message The message to test.
+     * @param allowAmqpValue Whether or now to allow AmqpValue of type {@code String} or {@code byte[]} to pass as well.
+     * @return {@code true} if the body was not {@code null} and an instance of {@link Data}, {@code false} otherwise.
+     *         If the parameter {@code allowAmqpValue} was {@code true}, then the method would also return {@code true}
+     *         if the body contains an {@link AmqpValue} of type {@code String} or {@code byte[]}.
+     * 
+     * @throws NullPointerException If the parameter {@code message} was {@code null}.
+     */
+    public static boolean hasDataBody(final Message message, final boolean allowAmqpValue) {
```
sophokles73: this is very helpful :+1:

service-base/src/main/java/org/eclipse/hono/service/command/CommandResponseSenderImpl.java
```diff
@@ -144,9 +143,7 @@ private static Message createResponseMessage(
         if (contentType != null) {
             msg.setContentType(contentType);
         }
-        if (payload != null) {
-            msg.setBody(new Data(new Binary(payload.getBytes())));
-        }
+        MessageHelper.setPayload(msg, contentType, payload);
```
sophokles73: we can also remove the three lines above, can't we?

services/messaging/src/main/java/org/eclipse/hono/messaging/HonoMessagingMessageFilter.java
```diff
@@ -58,7 +57,7 @@ public static boolean verify(final ResourceIdentifier linkTarget, final Message
          if (msg.getContentType() == null) {
              LOG.trace("message [{}] has no content type", msg.getMessageId());
              return false;
-         } else if (msg.getBody() != null && !(msg.getBody() instanceof Data)) {
+        } else if (!MessageHelper.hasDataBody(msg, false)) {
```
sophokles73: I think we still need the `null` check because an *empty notification* event may have an empty payload (as the name suggests) ...

service-base/src/main/java/org/eclipse/hono/service/command/CommandResponseSenderImpl.java
ctron: You mean:

~~~java
if (contentType != null) {
  msg.setContentType(contentType);
}
~~~

I will re-test this. But IIRC then a some integration test failed. As this would mean that the content type is set. `setPayload` only sets the content type, if the body is not `null`.

Hono uses two approaches for the content type. Sometimes it is set when the body is `null`, sometimes it isn't.

We could always set this. Or add a parameter to the `setPayload` method and explicitly tell it how to handle this situation.

sophokles73: So far, the policy is that we *always* set a content type on event and telemetry messages being forwarded downstream. There is only one case where we accept an empty body: the *empty-notification* as defined by the Event API. So, for the time being we should make sure that the content type gets set.

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -598,4 +640,133 @@ public static Boolean isDeviceCurrentlyConnected(final Message msg) {
         }).orElse(Boolean.FALSE);
     }
 
+    /**
+     * Set the payload of the message.
```
sophokles73: IMHO this should also point out that the given payload will be added to the message in a *Data* section

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -598,4 +640,133 @@ public static Boolean isDeviceCurrentlyConnected(final Message msg) {
...
+        if (payload != null) {
+            if (contentType != null) {
+                message.setContentType(contentType);
+            }
+            message.setBody(new Data(new Binary(payload)));
+        }
+    }
+
+    /**
+     * Set the payload of the message.
```
sophokles73: same here

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -598,4 +640,133 @@ public static Boolean isDeviceCurrentlyConnected(final Message msg) {
...
+     * @throws NullPointerException If the parameter {@code message} was {@code null}.
+     */
+    public static void setPayload(final Message message, final String contentType, final Buffer payload) {
+        Objects.requireNonNull(message);
+
+        setPayload(message, contentType, payload != null ? payload.getBytes() : null);
+    }
+
+    /**
+     * Set the payload of the message.
```
sophokles73: and here

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -598,4 +640,133 @@ public static Boolean isDeviceCurrentlyConnected(final Message msg) {
...
+     * @throws NullPointerException If the parameter {@code message} was {@code null}.
+     */
+    public static void setPayload(final Message message, final String contentType, final JsonObject payload) {
+        Objects.requireNonNull(message);
+
+        setPayload(message, contentType, payload != null ? payload.toBuffer() : null);
+    }
+
+    /**
+     * Set the payload of the message.
```
sophokles73: and here

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -699,12 +698,7 @@ protected final void createAndSendRequest(
 
         if (isOpen()) {
             final Message request = createMessage(action, properties);
-            if (payload != null) {
-                if (contentType != null) {
-                    request.setContentType(contentType);
-                }
-                request.setBody(new AmqpValue(payload.getBytes()));
-            }
+            MessageHelper.setPayload(request, contentType, payload);
```
sophokles73: did you already verify that a JMS provider based peer will be able to correctly create a JMS TextMessage from this? If we make this change in the code then we also need to change the Credentials, Tenant and Device Registration API as well and document the API change in the release notes ...

ctron: I didn't create a dedicated test for that yet. However looking at the tests for JMS, I think this is already covered:

https://github.com/eclipse/hono/blob/5527cbb83b5e02bf5fcc8894381e2e33d184f543/tests/src/test/java/org/eclipse/hono/tests/jms/RegistrationTestSupport.java#L168-L178

sophokles73: The existing JMS based integration tests only verify if a JMS based *client* can interact with our (non-JMS based) default registry implementation. So, we do not verify if a JMS based implementation of the current Device Registration API would be able to process an AMQP message that contains a Data section instead of an AMQP Value ...

service-base/src/main/java/org/eclipse/hono/service/command/CommandResponseSenderImpl.java
sophokles73: Meaning, keep the existing logic for setting the content type.

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
ctron: But that would use the same JMS functionality and APIs. And in case the JMS implementation would send an AMQP encoded String or byte array instead of a Data section, then Hono would still accept that.

ctron:
I think all comments and feedback should be taken care of.

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
sophokles73: Forget my comment on JMS based service implementations. There will not be such a thing because JMS is a client-only technology, i.e. you cannot *implement* an AMQP 1.0 based service endpoint using a JMS provider ...

So I guess you're right in that our existing integration test should be suitable for verifying this. However, we will also need to change the APIs and the device registry so that the response payload is also contained in *Data* sections and an *application/json* content type is set.

sophokles73:
> I think all comments and feedback should be taken care of.

I think what's still missing is the changes to the API specifications and the service implementation side (also using *Data* sections for the response payload) ...

ctron:
> I think what's still missing is the changes to the API specifications and the service implementation side (also using Data sections for the response payload) ...

True, documentation is not updated. … But everyone can read the Java code :grin:  … just kidding…

The response payload already is encoded as a Data section, the part in `org.eclipse.hono.util.RequestResponseApiConstants.getAmqpReply(String, EventBusMessage)`: https://github.com/eclipse/hono/pull/803/files#diff-941d733d5ca12f6bcee1f076fc88f1e2R106

sophokles73:
> The response payload already is encoded as a Data section

I see, I missed that ... then it's just the documentation (and an entry in the release notes).

ctron:
Ok, just pushed that.

site/content/api/Credentials-API.md
```diff
@@ -255,7 +257,7 @@ Hono uses the following AMQP message delivery states when receiving request mess
 
 Most of the operations of the Credentials API allow or require the inclusion of credential data in the payload of the
 request or response messages of the operation. Such payload is carried in the body of the corresponding AMQP 
-messages as part of a single *AMQP Value* section.
+messages as part of a single *Data* section. The content type must be set to `application/json`.
```
sophokles73: IMHO we should explicitly define the content-type property as mandatory in the definition of each operation that requires such payload. We also define the *subject* there explicitly ...

site/content/api/Device-Registration-API.md
```diff
@@ -279,7 +283,7 @@ Hono uses the following AMQP message delivery states when receiving request mess
 
 Most of the operations of the Device Registration API allow or require the inclusion of registration data in the payload of the
 request or response messages of the operation. Such payload is carried in the body of the corresponding AMQP 
-messages as part of a single *AMQP Value* section.
+messages as part of a single *Data* section. The content type must be set to `application/json`.
```
sophokles73: IMHO we should explicitly define the content-type property as mandatory in the definition of each operation that requires such payload. We also define the subject there explicitly ...

site/content/api/Tenant-API.md
```diff
@@ -269,7 +270,7 @@ Hono uses the following AMQP message delivery states when receiving request mess
 
 Most of the operations of the Tenant API allow or require the inclusion of tenant data in the payload of the
 request or response messages of the operation. Such payload is carried in the body of the corresponding AMQP 
-messages as part of a single *AMQP Value* section.
+messages as part of a single *Data* section. The content type must be set to `application/json`.
```
sophokles73: IMHO we should explicitly define the content-type property as mandatory in the definition of each operation that requires such payload. We also define the subject there explicitly ...

site/content/release-notes.md
```diff
@@ -14,6 +14,10 @@ title = "Release Notes"
...
 * The `org.eclipse.hono.service.tenant.TenantService` interface now describes only the mandatory operations of the API. The complete API is offered in `org.eclipse.hono.service.tenant.CompleteTenantService`. These interfaces are implemented in `org.eclipse.hono.service.tenant.BaseTenantService` and `org.eclipse.hono.service.tenant.CompleteBaseTenantService` respectfully. Tenant services implementations can offer the mandatory only or the full API by extending the according base class.
 * The `org.eclipse.hono.service.credentials.CredentialsService` interface now describes only the mandatory operations of the API. The complete API is offered in `org.eclipse.hono.service.credentials.CompleteCredentialsService`. These interfaces are implemented in `org.eclipse.hono.service.credentials.BaseCredentialsService` and `org.eclipse.hono.service.credentials.CompleteBaseCredentialsService` respectfully. Credentials services implementations can offer the mandatory only or the full API by extending the according base class.
+* All messages containing JSON objects as payload are now encoded using *Data*
+  sections and are required to have the content type `application/json`. When
```
sophokles73: FMPOV we should explicitly mention that the Tenant, Credentials and Device Registration APIs have been changed to require this, because that is what we did.

sophokles73:
The JMS based integration test currently do not include any payload in the request messages. FMPOV we should change them to do so in order to verify if our service implementations can retrieve the data from the request messages ...

site/content/api/Credentials-API.md
ctron: I did add this to the "standard" section of each API, as this affects all messages.

site/content/api/Device-Registration-API.md
ctron: I did add this to the "standard" section of each API, as this affects all messages.

site/content/api/Tenant-API.md
ctron: I did add this to the "standard" section of each API, as this affects all messages.

site/content/release-notes.md
ctron: Done.

ctron:
Ok, I just pushed a new unit test in the integration tests which sets and retrieves payload. Using a JMS text message.

core/src/main/java/org/eclipse/hono/util/MessageTap.java
```diff
@@ -25,7 +23,7 @@
  */
 public final class MessageTap implements Consumer<Message> {
 
-    private Consumer<Message> tapConsumer;
+    private final Consumer<Message> tapConsumer;
```
sysexcontrol: Given that
- the class is `final` already
- the member is `private`
why make it additionally `final` itself?
Not that it is wrong or particularly bad, but IMHO this should be superflous, shouldn't it?

ctron: `final` on the member makes it `const`, where `final` on the class only prevents any derived classes. So what I do here, lets the compiler check if the member gets re-assign and prevents that.

PR #807, opened 2018-08-30
Implement notification commands for HTTP adapter.

sysexcontrol:
This PR is the first implementation of notification commands (commands that do not expect any response from the device).
It is provided for the HTTP adapter first.

Missing in this PR:
- integration tests
- documentation
Will follow in the next step, as well as the MQTT implementation.


sophokles73:
While reviewing the code I got the feeling that we are trying to implement a *notification* message exchange pattern (MEP) over a mechanism that has been implemented for something different, i.e. the *request/response* MEP. While this is technically possible, I have strong doubts that in three months we will be able to understand why we did this in the first place.

I would therefore suggest, that we first go back to the C&C API and add a new operation for sending a (one-way) notification to a device, based on a sender link only (no response expected). In this context it  seems desirable for the client to be able to know whether its notification has been forwarded to the device. For that purpose I would suggest to use the different outcomes of the AMQP 1.0 transfer that is used to send the request message or notification from the command client to the protocol adapter instead of the adapters *abusing* the client's receiver link to send back a *fake* response on behalf of the device indicating that e.g. the device is no longer connected. For that purpose we would need to document the semantics of the outcome of the transfer of a request/notification to the C&C API. Note that we will also need to adapt the existing implementation in the protocol adapters accordingly, i.e. releasing a message instead of sending back a 503 response if the device is not connected (anymore).

Based on the definition of such a new *send notification* operation in the C&C API, we can then proceed with creating a proper implementation ...

sysexcontrol:
That makes perfect sense to me and I fully agree on that. Including that the current implementation for the already defined Command MEP (i.e. with a required response) also needs to be adopted to use the transfer outcomes (e.g. for 503) instead of using the receiver link top report that.
After this change, the receiver link for responses (in the application) will only be used for a response message that originated from the device - and never from the protocol adapter.

For the `notification` MEP I will provide a sequence diagram as a discussion base inside this PR first, and then prepare a specification PR (and close this one without merging).

Additionally, I will create a new issue for changing the implementation of the current Command and Control implementation to consider the transfer outcome instead of using the receiver link of the application for failure reports from the protocol adapter.

sysexcontrol:
@sophokles73 added two basic sequence diagrams how the notification commands should be handled IMHO.
WDYT?

sysexcontrol:
Thinking about it I changed my mind: the endpoint should probably be still "/control", and not "/notify".
Otherwise we have additional complexity regarding the permissions in the AMQP network which we should avoid if not having a good reason...

sysexcontrol:
I currently see following definition of transfer outcomes for notification commands (status codes would be used for the Hono C&C client, but not for the AMQP API definition):

| Processing in adapter | AMQP outcome |  HonoClient StatusCode |
|-----------------------------|----------------------| ---------------- |
| Bad command            | Rejected ("hono:bad-request") | HTTP_BAD_REQUEST |
| Device not connected (anymore) | Release | HTTP_UNAVAIL |
| Command delivered | Accepted | HTTP_ACCEPTED |

Additionally there are the AMQP "client" errors that could happen and are mapped by the Hono Client to a status code:

| Client error | HonoClient StatusCode |
| ---------------|----------------|
| Resource limit exceeded | HTTP_FORBIDDEN |
| unauthorized access | HTTP_FORBIDDEN |
| no link found | HTTP_NOT_FOUND |


@sophokles73 does the definition of "Processing in adapter" to "AMQP outcome" in the C&C API (for notification commands) make sense to you?

For commands, there probably is no "Accepted" to "HTTP_ACCEPTED" mapping, since this will later be responded by the device itself.

sophokles73:
@sysexcontrol IMHO it would make more sense to create a PR for updating/changing the C&C API and have this discussion on that PR ...

sysexcontrol:
@sophokles73 yes, that is the plan. This was only for getting the rough direction clear before making the new PR with the specification.
So, any comments on that? I will then put it into the new PR and close this one here.

sophokles73:
looks reasonable

however, you should also include the corresponding changes to the *send Command* operation in the C&C API...

sysexcontrol:
Closing this PR after the specification PR #813 was now provided.

Issue #794, opened 2018-08-23
Replace dropwizard with micrometer as default metric framework


sysexcontrol:
We have to use the micrometer backport for Spring Boot 1.5.

Both metric formats need to be supported : 
- existing format
- tagged format (as proposed on mailing list)

ctron:
As there are CQs still pending, I will move this to 0.8.0-M2.

ctron:
Close by PR #809

PR #809, opened 2018-08-31
[#794]: Replace Dropwizard with Micrometer

ctron:
This change replaces the internal usage of Dropwizard in favor of
Micrometer. This also makes use of using tags for meters.

Micrometer is integrated using Spring Boot, and vertx is only configured
to contribute to that existing meter registry. This allows to take full
advantage of the Spring Boot Micrometer integration, and simply drop in
any Micrometer adapter, including the use of auto-configuration.

With this change Hono can be operated in two modes, either the "legacy"
mode, which provides at least the same metrics as before, publishing
over the same Graphite reporter, with exactly the same metrics names.

Or using the new format, which is the default and makes use of any
Micrometer backend implementation that is dropped into the classpath.

---

**Note:** With this PR it would already be possible to use prometheus. However the Dashboards and deployments aren't updated and I would like to put that in a second PR.


adapters/http-vertx/pom.xml
```diff
@@ -1,16 +1,11 @@
...
-    information regarding copyright ownership.
-   
-    This program and the accompanying materials are made available under the
-    terms of the Eclipse Public License 2.0 which is available at
-    http://www.eclipse.org/legal/epl-2.0
-   
-    SPDX-License-Identifier: EPL-2.0
- -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+<!-- Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation See
```
sophokles73: can you please undo this *formatting*?

core/src/main/java/org/eclipse/hono/util/Hostnames.java
```diff
@@ -0,0 +1,53 @@
...
+    static {
+
+        // works on Linux
+        String hostname = System.getenv("HOSTNAME");
+
+        if (hostname == null) {
+            // this can produce all kinds of unexpected results
+            // but better than nothing
+            try {
+            final InetAddress localhost = InetAddress.getLocalHost();
```
sophokles73: indentation

ctron:
All CQs have checkin permssion now, most of them are already approved. So I think it makes sense to continue here.

I re-based the branch and would encourage you to take a look.

core/src/test/java/org/eclipse/hono/util/HostnamesTest.java
```diff
@@ -11,21 +11,21 @@
...
-
-    @Override
-    protected String getScope() {
-        return SERVICE_PREFIX;
+    /**
+     * The result must never be {@code null}.
+     */
+    @Test
+    public void testNonNull() {
+        Assert.assertNotNull(Hostnames.getHostname());
```
sophokles73: this is not a reproducible test because it doesn't test the specific ways for determining the hostname, does it?

adapters/coap-vertx/src/main/java/org/eclipse/hono/adapter/coap/vertx/Config.java
```diff
@@ -87,6 +93,20 @@ public CoapAdapterProperties adapterProperties() {
...
+     * 
+     * @return The new meter registry customizer.
+     */
+    @Bean
+    public MeterRegistryCustomizer<MeterRegistry> commonTags() {
+        return r -> r.config().commonTags(
+                Tags.of(
+                        Tag.of("host", Hostnames.getHostname()),
+                        Tag.of("component", "adapter"),
+                        Tag.of("protocol", "coap")));
```
sophokles73: IMHO these tag names should be defined as constants (e.g. in MicrometerBasedMetrics ?) so that we can be sure to always use the same values, don't you think?

core/src/test/java/org/eclipse/hono/util/HostnamesTest.java
ctron: Well this test only tests if "not null" is returned, and that should be reproducible.

service-base/src/main/java/org/eclipse/hono/service/metric/legacy.properties
```diff
@@ -0,0 +1,2 @@
+management.metrics.export.graphite.protocol=plaintext
+management.metrics.export.graphite.tagsAsPrefix=host
```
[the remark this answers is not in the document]
ctron: Fixed.

adapters/coap-vertx/src/main/java/org/eclipse/hono/adapter/coap/vertx/Config.java
ctron: Absolutely. I put it alongside `MicrometerBasedMetrics` though, to make it less coupled and more visible.

core/src/test/java/org/eclipse/hono/util/HostnamesTest.java
sophokles73: You're right in that it always succeeds because the method always returns a non-null value. But that is a no-brainer, isn't it? A reasonable test should actually check if it returns a value based on the different ways the getHostname method tries to determine the hostname, shouldn't it?

ctron: >  But that is a no-brainer, isn't it?

At the moment yes. But the actual lookup of the value is not. And if someone changes the logic of that method, accidentally returning `null`, this test would catch that.

Yes, that unit test could have additional methods. But I am not sure how feasible it will be creating them. Modifying the environment variable HOSTNAME inside a unit test might be bad idea.

deploy/src/main/config/hono-adapter-amqp-vertx-config.yml
```diff
@@ -43,15 +43,16 @@ hono:
...
+  metrics:
+    legacy: true
+
+# disable spring boot actuator endpoints
 endpoints:
   enabled: false
-  metrics:
-    enabled: true
 
+management:
```
sophokles73: it looks  like configuration of the reporter has changed when compared to https://www.eclipse.org/hono/admin-guide/monitoring-tracing-config/ 

IMHO you should update the documentation accordingly (and add a corresponding note to the release notes as well)

site/content/admin-guide/monitoring-tracing-config.md
```diff
@@ -16,21 +16,22 @@ Metrics usually provide insights into the past and current status of an individu
...
-| `HONO_METRIC_REPORTER_GRAPHITE_PREFIX`<br><nobr>`--hono.metric.reporter.graphite.prefix`</nobr> | no  | - | An arbitrary string that all metric names will be prefixed with. |
-| `HONO_METRIC_JVM_MEMORY`<br><nobr>`--hono.metric.jvm.memory`</nobr>                             | no  | `false` | Enables reporting of the JVM's memory metrics (from the Dropwizard JVM Instrumentation). The metric name is `hono.http.jvm.memory`. |
-| `HONO_METRIC_JVM_THREAD`<br><nobr>`--hono.metric.jvm.thread`</nobr>                             | no  | `false` | Enables reporting of the JVM's thread metrics (from the Dropwizard JVM Instrumentation). The metric name is `hono.http.jvm.thread`.|
-| `HONO_METRIC_VERTX`<br><nobr>`--hono.metric.vertx`</nobr>                                       | no  | `false` | Enables reporting of Vert.x' standard metrics (from the Vert.x metrics project). The metric name is `hono.http.vertx`. |
+| `HONO_METRICS_LEGACY`<br><nobr>`--hono.metrics.legacy`</nobr>   | no  | `true` | Enable legacy metrics using graphite. |
```
sophokles73: My understanding is that this will not be enough but you will also need to configure the InfluxDB host and port etc. Or am I mistaken?

Issue #624, opened 2018-05-22
Command and control in JMeter client to allow load tests


pellmann:
To load test command and control it should be possible to
- send a command and receive the response in the role of a business application
- receive a command and send a response in the role of an adapter

kaniyan:
@pellmann Can this issue be closed or anything more need to be done?

Issue #808, opened 2018-08-30
Specify notification commands as new MEP


sysexcontrol:
Currently, our Command and Control API requires that a device must send a response to a received command.

For `notification commands` we need an API extension that reflects that there is no response expected or supported.
This especially means that there must be no receiver link opened in the sending application, but instead the outcome of the transfer frame is used to find out if the notification command was accepted by the protocol adapter.

sysexcontrol:
Available in 0.8-M1_1 milestone (specification, not the implementation).

Issue #669, opened 2018-06-25
Support "one way" command pattern


sysexcontrol:
Currently a command sent by an application waits for a response sent by the device to become successful.

Since some devices are just not capable of ever sending a response, we need an additional pattern for commands without a response:

- the application should be able to signal that there is no response expected (in this case no receiver link for a response would be opened at all). 
- on the AMQP 1.0 message level this probably is best handled by a missing `reply-to`.
- on the API level there must be some extension implemented that lets the application control that the command is to be considered "one-way".

sophokles73:
> * the application should be able to signal that there is no response expected (in this case no receiver link for a response would be opened at all)

The application should then simply not set the *reply-to* property in the command message.

> * the device should be able to signal for downstream messages in addition to the ttd parameter that there will be no response sent to a potentially received command (called e.g. nor for no response

I do not think that I understand the purpose of this.

> * the adapter should be able to detect a default value for the device regarding the nor parameter

What is this supposed to be good for?

sophokles73:
We already have added the concept of a *one-way* command to the [C&C API](https://www.eclipse.org/hono/api/command-and-control-api/#send-a-one-way-command).

However, the protocol adapters need to be adapted accordingly, e.g. the HTTP adapter needs to be changed to not include the *hono-cmd-id* header in the HTTP response if no *reply-to* property is set in the command.

sophokles73:
@sysexcontrol are you currently working on this issue?  It is assigned to you but it is not in progress (yet) ...

sysexcontrol:
> > * the application should be able to signal that there is no response expected (in this case no receiver link for a response would be opened at all)
> 
> The application should then simply not set the _reply-to_ property in the command message.

That is a reasonable implementation detail. However, the current API needs to be somehow extended as well, since for the client the _reply-to_ property is completely transparent currently (and thus would always be set).
I will extend this part in the description.

sysexcontrol:
> > * the device should be able to signal for downstream messages in addition to the ttd parameter that there will be no response sent to a potentially received command (called e.g. nor for no response
> 
> I do not think that I understand the purpose of this.

This was meant as the possibility for a device to signal that it will (or cannot) reply to any command. At the same time as the `ttd` parameter is given.

The first option demands that the application itself needs to know if a command can be responded by the device or not. This 2nd option would enable an application to send a command, and the underlying API can decide itself if there should be a response from the device or not - the application would be free from this decision.

[...]

PR #812, opened 2018-09-03
[#624] Jmeter client for Command and Control to enable load tests

kaniyan:
This PR contains HonoCommanderSampler, using which settings required to perform load test for feature Command and Control can be configured. Users can choose the type of protocol being used (http or mqtt) for communication between devices and hono. This sampler establishes connection to hono, get devices that are ready to receive commands under a particular tenant, send commands and receive command responses.This sampler can be combined with standard Jmeter samplers (for example HTTP Request Sampler) to create a test plan for C&C.

### DIFF
--- a/jmeter/src/main/java/org/eclipse/hono/jmeter/HonoCommanderSampler.java
+++ b/jmeter/src/main/java/org/eclipse/hono/jmeter/HonoCommanderSampler.java
@@ -1,0 +1,155 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.jmeter;
+
+import static java.lang.String.format;
+import static org.eclipse.hono.jmeter.HonoSamplerUtils.getIntValueOrDefault;
+
+import java.util.concurrent.CompletionException;
+
+import org.apache.jmeter.samplers.Entry;
+import org.apache.jmeter.samplers.SampleResult;
+import org.apache.jmeter.testelement.ThreadListener;
+import org.eclipse.hono.jmeter.client.HonoCommander;
+import org.eclipse.hono.jmeter.sampleresult.HonoCommanderSampleResult;
+import org.slf4j.LoggerFactory;
+
+/**
+ * JMeter creates an instance of a sampler class for every occurrence of the element in every thread. [some additional
+ * copies may be created before the test run starts]
+ *
+ * Thus each sampler is guaranteed to be called by a single thread - there is no need to synchronize access to instance
+ * variables.
+ *
+ * However, access to class fields must be synchronized.
+ */
+public class HonoCommanderSampler extends HonoSampler implements ThreadListener {
+
+    private static final org.slf4j.Logger LOG = LoggerFactory.getLogger(HonoCommanderSampler.class);
+    private static final String RECONNECT_ATTEMPTS = "reconnectAttempts";
+    private static final String COMMAND = "command";
+    private static final String COMMAND_PAYLOAD = "commandPayload";
+    private static final String COMMAND_TIMEOUT = "commandTimeOut";
+    private static final String TRIGGER_TYPE = "triggerType";
+    private static final int DEFAULT_COMMAND_REQUEST_TIMEOUT = 1000;
+    private HonoCommander honoCommander;
+
+    @Override
+    public SampleResult sample(final Entry entry) {
+        final HonoCommanderSampleResult sampleResult = new HonoCommanderSampleResult();
+        sampleResult.setDataType(SampleResult.TEXT);
+        sampleResult.setSampleLabel(getName());
+        honoCommander.sample(sampleResult);
+        return sampleResult;
+    }
+
+    @Override
+    public void threadStarted() {
+        LOG.info("Sampler thread started");
+        try {
+            honoCommander = new HonoCommander(this);
+            honoCommander.start().join();
+        } catch (final CompletionException e) {
+            throw new RuntimeException(format("Error starting commander sampler for tenant %s", getTenant()),
+                    e.getCause());
+        }
+    }
+
+    @Override
+    public void threadFinished() {
+        LOG.info("Sampler thread finished");
+        if (honoCommander != null) {
+            try {
+                honoCommander.close().join();
+            } catch (final CompletionException e) {
+                LOG.error("error during shut down of command application", e);
+            }
+        }
+
+    }
+
+    public int getReconnectAttemptsAsInt() {
+        return getIntValueOrDefault(getPropertyAsString(RECONNECT_ATTEMPTS), 0);
+    }
+
+    public String getReconnectAttempts() {
+        return getPropertyAsString(RECONNECT_ATTEMPTS);
+    }
+
+    /**
+     * Sets the number of re-connect attempts.
+     *
+     * @param reconnectAttempts The number of attempts as string.
+     */
+    public void setReconnectAttempts(final String reconnectAttempts) {
+        setProperty(RECONNECT_ATTEMPTS, reconnectAttempts);
+    }
+
+    public String getCommand() {
+        return getPropertyAsString(COMMAND, "ArbitraryCommand");
+    }
+
+    /**
+     * Sets the command to send to device.
+     *
+     * @param command Command to send to device
+     */
+    public void setCommand(final String command) {
+        setProperty(COMMAND, command);
+    }
+
+    public String getCommandPayload() {
+        return getPropertyAsString(COMMAND_PAYLOAD, "{\"set\": \"arbitrary value\"}");
+    }
+
+    /**
+     * Sets the command payload to send to device.
+     *
+     * @param commandPayload Command payload to send to device or the default timeout if the value is empty or cannot be
+     *            parsed as integer.
+     */
+    public void setCommandPayload(final String commandPayload) {
+        setProperty(COMMAND_PAYLOAD, commandPayload);
+    }
+
+    public int getCommandTimeoutAsInt() {
+        return getIntValueOrDefault(getPropertyAsString(COMMAND_TIMEOUT), DEFAULT_COMMAND_REQUEST_TIMEOUT);
+    }
+
+    public String getCommandTimeout() {
+        return getPropertyAsString(COMMAND_TIMEOUT);
+    }
+
+    /**
+     * Sets the timeout for waiting for a command response by the application.
+     *
+     * @param commandTimeout The timeout in milliseconds encoded as string.
+     */
+    public void setCommandTimeout(final String commandTimeout) {
+        setProperty(COMMAND_TIMEOUT, commandTimeout);
+    }
+
+    public String getTriggerType() {
+        return getPropertyAsString(TRIGGER_TYPE);
+    }
+
+    /**
+     * Sets the type of trigger for the command sampler.
+     * 
+     * @param triggerType The trigger type (device or sampler)
+     */
+    public void setTriggerType(final String triggerType) {
+        setProperty(TRIGGER_TYPE, triggerType);
+    }
+}
--- a/jmeter/src/main/java/org/eclipse/hono/jmeter/HonoSamplerUtils.java
+++ b/jmeter/src/main/java/org/eclipse/hono/jmeter/HonoSamplerUtils.java
@@ -1,0 +1,44 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.jmeter;
+
+/**
+ * A collection of utility methods for Jmeter tests.
+ *
+ */
+public final class HonoSamplerUtils {
+
+    private HonoSamplerUtils() {
+        // prevent instantiation
+    }
+
+    /**
+     * Parses string value and returns an integer value. If input string value is null, then provided default integer
+     * value is returned.
+     * 
+     * @param stringValue The string value that need to be converted as an integer.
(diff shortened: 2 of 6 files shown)


jmeter/src/main/java/org/eclipse/hono/jmeter/client/HonoCommander.java
```diff
@@ -0,0 +1,252 @@
...
+ *  * information regarding copyright ownership.
+ *  *
+ *  * This program and the accompanying materials are made available under the
+ *  * terms of the Eclipse Public License 2.0 which is available at
+ *  * http://www.eclipse.org/legal/epl-2.0
+ *  *
+ *  * SPDX-License-Identifier: EPL-2.0
+ *  ******************************************************************************
+ *
+ */
```
pellmann: Javadoc

jmeter/src/main/java/org/eclipse/hono/jmeter/sampleresult/HonoCommanderSampleResult.java
```diff
@@ -0,0 +1,39 @@
...
+ *  *
+ *  * This program and the accompanying materials are made available under the
+ *  * terms of the Eclipse Public License 2.0 which is available at
+ *  * http://www.eclipse.org/legal/epl-2.0
+ *  *
+ *  * SPDX-License-Identifier: EPL-2.0
+ *  ******************************************************************************
+ *
+ */
+
```
pellmann: Javadoc

jmeter/src/main/java/org/eclipse/hono/jmeter/ui/HonoCommanderSamplerUI.java
```diff
@@ -0,0 +1,156 @@
...
+ *  * information regarding copyright ownership.
+ *  *
+ *  * This program and the accompanying materials are made available under the
+ *  * terms of the Eclipse Public License 2.0 which is available at
+ *  * http://www.eclipse.org/legal/epl-2.0
+ *  *
+ *  * SPDX-License-Identifier: EPL-2.0
+ *  ******************************************************************************
+ *  
+ */
```
pellmann: Javadoc

jmeter/src/main/java/org/eclipse/hono/jmeter/ui/HonoCommanderSamplerUI.java
```diff
@@ -0,0 +1,156 @@
...
+
+    private String getDescriptionTextForSelectedProtocol() {
+        final String prototypeSelected = ((String) protocol.getSelectedItem()).trim();
+        if (prototypeSelected.equals("http")) {
+            return "Devices use Http event/telemetry messages to inform their command readiness by setting hono-ttd value. In this case, commands are triggered by this sampler on arrival of those messages.";
+        } else if (prototypeSelected.equals("mqtt")) {
+            return "Devices subscribe to a specified mqtt topic to inform their command readiness. In this case, this sampler send one command per sample to the devices.";
+        }
+        return "No description available.";
+    }
```
pellmann: In general I think it would be better not to distinguish between MQTT/HTTP, but between the modes of devices with permanent connection (e.g. MQTT, AMQP) and devices that occasionally report (e.g. with HTTP calls).

The whole point of Hono is that the protocol doesn't really matter. However, the basic type of connection of the Devices may already play a role for the case.

jmeter/src/main/java/org/eclipse/hono/jmeter/HonoCommanderSampler.java
```diff
@@ -0,0 +1,207 @@
...
+ *  *
+ *  * This program and the accompanying materials are made available under the
+ *  * terms of the Eclipse Public License 2.0 which is available at
+ *  * http://www.eclipse.org/legal/epl-2.0
+ *  *
+ *  * SPDX-License-Identifier: EPL-2.0
+ *  ******************************************************************************
+ *
+ */
+
```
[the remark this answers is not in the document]
kaniyan: I will change them.

jmeter/src/main/java/org/eclipse/hono/jmeter/HonoSamplerUtils.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ */
+public final class HonoSamplerUtils {
+
+    private HonoSamplerUtils() {
+        // prevent instantiation
+    }
+
+    /**
+     * @param stringValue The string value that need to be converted as an integer.
```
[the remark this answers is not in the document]
kaniyan: I didn't get it. Already a description for this class is available. What do you mean by general description?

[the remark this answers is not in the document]
pellmann: I meant for method - only the parameters are described.

jmeter/src/main/java/org/eclipse/hono/jmeter/ui/HonoCommanderSamplerUI.java
pellmann: I think with your MQTT example it is the opposite. If a device makes a MQTT connection or subscription only for a short time, the subscription (ttd event) should be the trigger and it need to be used the same way as the HTTP right now.

As I understand it, the technical difference is whether you can assume that the device is completely accessible for commands during the test or the device is only accessible if it has sent a ttd event. How this maps to the protocols should not matter from the point of view. For example, there could be an HTTP adapter that supports sockets. Or just that by MQTT is always only briefly connected. You would also want to test such a scenario via JMeter. For example, to check the costs of MQTT connection setups.

I am not sure what you mean with uni/bidirectional as the commands in the moment are always bidirectional FMPOV, since we always expect a command response to be sure, the command arrived. In the future we will maybe have unidirectional calls. (#669, #808)

jmeter/src/main/java/org/eclipse/hono/jmeter/HonoSamplerUtils.java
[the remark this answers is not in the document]
kaniyan: Got it now :)

jmeter/src/main/java/org/eclipse/hono/jmeter/HonoCommanderSampler.java
```diff
@@ -0,0 +1,207 @@
...
+     * Sets whether or not to use the sender time.
+     *
+     * @param useSenderTime {@code true} in order to use the sender time.
+     */
+    public void setUseSenderTime(final boolean useSenderTime) {
+        setProperty(USE_SENDER_TIME, useSenderTime);
+    }
+
+    @Override
+    public String getEndpoint() {
```
ctron: Why is this method overridden? It doesn't seem to add anything? Also see `setEndpoint`.

jmeter/src/main/java/org/eclipse/hono/jmeter/client/HonoCommander.java
```diff
@@ -0,0 +1,252 @@
...
+                } else if ("mqtt".equals(sampler.getProtocol())) {
+                    LOG.trace("A device [{}] got registered over mqtt", notification.getDeviceId());
+                    allConnectedDevices.add(notification.getDeviceId());
+                }
+            }
+        }
+    }
+
+    private void sendCommandAndReceiveResponse(final String tenantId, final String deviceId) {
+        synchronized (lock) {
```
ctron: I don't think the synchronized does what you had in mind here. The following calls build up a chain of calls, but doesn't actually execute them. So e.g. the call to `client.getOrCreateCommandClient` will hold the lock, but the lambda of the following `compose`might not.

jmeter/src/main/java/org/eclipse/hono/jmeter/HonoCommanderSampler.java
kaniyan: Thank you for bringing it to notice. I will remove them.

jmeter/src/main/java/org/eclipse/hono/jmeter/ui/HonoCommanderSamplerUI.java
kaniyan: 1. As we discussed, instead of grouping based on protocols, I group now based on triggers (initiated by device or sampler) as in the attached screenshot. WDYT?

image

2. As you noticed, there is a chance that after sending commands and before the devices receive them, devices can unsubscribe. In this case, it is recorded as error. It can be overcome, by  checking if the device is still subscribed, before counting it as an error.

pellmann: I like this much more then MQTT/HTTP, because it shows the cases from the application point of view (for which the protocol should not matter).

jmeter/src/main/java/org/eclipse/hono/jmeter/client/HonoCommander.java
kaniyan: Thank you for the input. My intention was not to ensure the execution of vertx future, rather to ensure that only one thread operates on this at a time. Inside the handler, there is a piece of code to increment success or error counts (which are field variables) and the idea is to ensure that this part of code is handled by only one thread at a time. Ofcourse only this particular part can be placed inside synchronised.

As part of the other changes, currently I am replacing this part of code to use vertx eventbus to notify the success or the failure. You can have have a look once I pushed the code.

kaniyan: Pls ignore my last comment, I interpreted your comment differently. :) I understood now what you meant in your last comment with lock and compose. In the last push, I synchronized the block inside the mapper instead of doing it as before.

jmeter/src/main/java/org/eclipse/hono/jmeter/client/HonoCommander.java
```diff
@@ -0,0 +1,259 @@
...
+
+    private static final org.slf4j.Logger LOG = LoggerFactory.getLogger(HonoCommander.class);
+    private final HonoClient client;
+    private final HonoCommanderSampler sampler;
+    private final List<String> devicesReadyToReceiveCommands = new CopyOnWriteArrayList<>();
+    private final AtomicInteger successCount = new AtomicInteger(0);
+    private final AtomicInteger errorCount = new AtomicInteger(0);
+    private final AtomicLong bytesSent = new AtomicLong(0);
+    private final AtomicLong bytesReceived = new AtomicLong(0);
+    private final AtomicLong sampleStart = new AtomicLong(0);
```
pellmann: I don't see what the Atomic's helping with. FMPOV you need to to be sure, that the sequence of writing the values to the sample and then set them to zero at each call of sample is synchronized against the increments of the values from the command receiving. If you would just use Atomics, you stay with a possible leak between setting in sample and resetting to zero. Und if you use synchronized for this sequence you need the same synchronization lock for the increments.

jmeter/src/main/java/org/eclipse/hono/jmeter/client/HonoCommander.java
```diff
@@ -0,0 +1,259 @@
...
+                                bytesReceived.addAndGet(sampler.getCommandPayload().getBytes().length);
+                                bytesSent.addAndGet(commandResponseText.getBytes().length);
+                            }
+                            return commandResponse;
+                        })
+                        .map(x -> closeCommandClient(commandClient, tenantId, deviceId))
+                        .recover(error -> {
+                            if (sampler.getTriggerType().equals("device")
+                                    || devicesReadyToReceiveCommands.contains(deviceId)) {
+                                errorCount.addAndGet(1);
```
pellmann: Not synchronized.

kaniyan: synchronized (lock) {
	successCount.addAndGet(1);
	bytesReceived.addAndGet(sampler.getCommandPayload().getBytes().length);
	bytesSent.addAndGet(commandResponseText.getBytes().length);
}
In the above part, I used synchronized block to let one thread to work on all the three lines at a time. In case of errorCount.addAndGet(1) there is only one line and hence there is no synchronised is needed and only vertx threads work on these parts.

jmeter/src/main/java/org/eclipse/hono/jmeter/client/HonoCommander.java
kaniyan: During every sampling , sampler resets those counts to 0 and on the other side vertx threads set them to actual value on success or failure. In order to ensure writing by these threads, Atomic is used. When vertx threads set these count, synchronized locks were used.

pellmann: You use synchronized on the same lock object in the sample (JMeter thread) and in the command response. You do not use it for errorCount. I think this is wrong. The errorCount could be incremented while the sample method is invoked by the JMeter thread. And this could happen between `result.setErrorCount(errorCount.get());` and `errorCount.set(0);` since the atomic does not help here. In this case one error would be gone.

jmeter/src/main/java/org/eclipse/hono/jmeter/client/HonoCommander.java
pellmann: See comment above - FMPOV it need to be synchronized with the code in the sample() method.

jmeter/src/main/java/org/eclipse/hono/jmeter/client/HonoCommander.java
kaniyan: Thank you. I got your point. Yes, I see now the chance of counts getting missed out. I would like to replace `result.setErrorCount(errorCount.get())` and `errorCount.set(0)` with `result.setErrorCount(errorCount.getAndSet(0))`. With this, it can be ensured that error counts won't be missed. Also with this way, I think we can avoid `synchronized(lock)`block inside `sample()` method. What do you say?

jmeter/src/main/java/org/eclipse/hono/jmeter/client/HonoCommander.java
kaniyan: pls see comment above.

jmeter/src/main/java/org/eclipse/hono/jmeter/client/HonoCommander.java
pellmann: There are still the two operations `errorCount.getAndSet(0)` and `result.setErrorCount(..)` as a sequence in which the `errorCount.set(0)` can happen. I do not see how the synchronized could be avoided.

kaniyan: getAndSet(0) is atomic, it returns the value and the set the value to 0. Hence errorCount.set(0) is no more needed.

pellmann: Sorry, I meant `errorCount.addAndGet(1);`, not `errorCount.set(0)`. 

`getAndSet(0)` is atomic. This is called.. and after it `result.setErrorCount()` with the value from `getAndSet(0)` is called. And in between, `errorCount.addAndGet(1);` could be called from the vert.x thread.

kaniyan:
@pellmann I have rebased this PR.

kaniyan:
@pellmann Done :)

PR #813, opened 2018-09-05
[#808] Specify notification commands as message exchange pattern.

sysexcontrol:
This PR extends the Command and Control API by specifying *notification commands* as new message exchange pattern (commands that cannot be responded by the device).
This will be the base of the implementation that is provided in the next PR after we have discussed this one.


site/content/api/Command-And-Control-API.md
```diff
@@ -3,38 +3,129 @@ title = "Command & Control API"
...
+| Processing in protocol adapter   | AMQP outcome                  |
+|----------------------------------|------------------------------ |
+| Bad command                      | Rejected ("hono:bad-request") |
+| Device not connected (anymore)   | Release                       | 
+| Command delivered                | Accepted                      | 
+
+For any outcome except for `Accepted` the *Business application* should immediately close the response receiver link again, 
+if it was scoped to exactly one command response. If the receiver link is scoped to the tenant, it usually should remain 
+open.
+
```
pellmann: Why should the application close the receiver link for e.g. device 4711 if a release happened or the command was bad? It could be a device, which is connected by MQTT, but was unconnected for a short time or one command was wrong. In both cases it would be fine to stay with the link.

I think you have the HTTP adapter case in mind were only one command could be send per ttd event from the device. 

It seems that we cannot distinguish this case on the side of the application at the moment from the variant that basically any number of commands can be sent after one ttd.

site/content/api/Command-And-Control-API.md
```diff
@@ -3,38 +3,129 @@ title = "Command & Control API"
...
+To scope a response to the device the `${reply-id}` should consist of the device-id plus an arbitrary string to scope the command sender instance. If all responses shall be received just in the context of the tenant, the device-id has to be omitted.   
+
+**Link establishment**
+
+The following sequence diagram shows the necessary links that need to be established to enable sending back a response to the Business Application:
+
+![Send Command](../command_control_receive_preconditions.png)
+
+# Outcomes for sending a command or notification command
+Commands and notification commands that are forwarded by the AMQP 1.0 network to a protocol adapter need to be responded with following outcomes:
```
pellmann: I think "with one of the following" might be better?

site/content/api/Command-And-Control-API.md
```diff
@@ -3,38 +3,129 @@ title = "Command & Control API"
...
+
+For any outcome except for `Accepted` the *Business application* should immediately close the response receiver link again, 
+if it was scoped to exactly one command response. If the receiver link is scoped to the tenant, it usually should remain 
+open.
+
+
 # Operations
 
-The following API defines operations that can be used by *Business Applications* to send commands to devices over an AMQP 1.0 Network and get a response in return.
+The following API defines operations that can be used by *Business Applications* to send commands or notification commands to devices over an AMQP 1.0 Network and get a response in return to a sent command.
```
pellmann: This sounds like notifications may also get responses.. maybe something like:

> The following API defines operations that can be used by *Business Applications* to send commands (wich gets a response in return) or notification commands to devices over an AMQP 1.0 .

site/content/api/Command-And-Control-API.md
```diff
@@ -3,38 +3,129 @@ title = "Command & Control API"
...
+The following sequence diagram shows the message flow with a successful outcome:
 
-**Link establishment and Message Flow**
+![Send Command successful outcome](../command_control_Success.png)
 
-The following sequence diagram shows the whole message exchange including the response being sent back to the Business Application.
+The following sequence diagram shows the message flow with an unsuccessful outcome:
 
-![Send Command](../command_control_Success.png)
+![Send Command unsuccessful outcome](../command_control_Unsuccessful_outcome.png)
```
pellmann: Same as above.. I do not think that the application needs to close the links on refejected or released.

sysexcontrol: Agreed.

sysexcontrol:
@pellmann I pushed a new commit with changes.

site/content/api/Command-And-Control-API.md
```diff
@@ -3,38 +3,138 @@ title = "Command & Control API"
...
+![Send Command](../command_control_receive_preconditions.png)
+
+# Outcomes for sending a command or notification command
+Commands and notification commands that are forwarded by the AMQP 1.0 network to a protocol adapter need to be responded with one of the following outcomes:
+
+| Processing in protocol adapter   | AMQP outcome                  |
+|----------------------------------|------------------------------ |
+| Bad command                      | Rejected ("hono:bad-request") |
+| Device not connected (anymore)   | Release                       | 
+| Command delivered                | Accepted                      |
```
ctron: I think we should be a bit more explicit in what "command delivered" means. Is that a guarantee that the command did reach the device?

sysexcontrol: Good point - it means "it reached the protocol adapter". More guarantees we cannot give, if an application needs to be absolutely sure that the command reached the device, it
- must not use a "notification command", but a "command"
- wait for the reply from the device

I will clarify this here that only the protocol adapter received the command.

site/content/api/Command-And-Control-API.md
```diff
@@ -40,9 +40,18 @@ the *Business Application* needs to open a receiver link as precondition:
...
-2. The `${reply-id}` may be any arbitrary string chosen by the client.  
+2. The `${reply-id}` may be any arbitrary string chosen by the client. 
+
+### Choosing a strategy for building the receiver link address
+
+While the `${reply-id}` can be completely chosen by the client, the specific use case for sending commands may influence 
+how it should be set.
+Thus following hints for choosing the right `${reply-id}` may be helpful:
+
+- if only one command response is expected from the device, the `${reply-id}` could be set to a UUID and closed again after receiving the response.
```
pellmann: Since the reply-id is something that need to be given to the device and will be send back I do not think that the advice for a UUID is the best. The replyTo is already scoped to tenant and device, so in most cases a UUID is much more then needed. Your phrase "arbitrary string to scope the command sender instance" would also be true.

sysexcontrol:
@pellmann : is your concern addressed by the new commit?
Thanks!

Issue #784, opened 2018-08-14
AMQP Adapter: Implement Command & Control API


Alfusainey:
This is a container issue for implementing the `Command & Control` API for the AMQP protocol adapter. All pull requests relating to `C&C` for the AMQP adapter will reference this issue.

Alfusainey:
@sophokles73 

I have spent sometime looking at how C&C is implemented for both the HTTP and MQTT adapters to get a feel for how to approach it for the AMQP adapter. I see the implementation for the AMQP adapter following a similar pattern to how it is done for the MQTT adapter (as summarized below):

1. Device indicates that it is ready to receive commands (until further notice)
   - device publishes a message to an address encapsulating the `subscribe` operation in the message's subject.
     * message.address (authenticated device): `control/+/+/req/#` (i.e # is a placeholder for additional parameters e.g property bags)
     * message.address (unauthenticated device): `control/[tenant-id]/[device-id]/req/#`
     * message.subject = `subscribe`

     Device can only subscribe to receiving commands using one valid subscription address. This is 
     different from the MQTT adapter which allows devices to subscribe to a list of topics.

  - Upon the adapter receiving the message to subscribe for commands:
      * it sends a `device ready for command` notification downstream (with a ttd=-1).
  - Upon the adapter receiving a command from applications:
      * it forwards the command to the device connected to the adapter.
        * message.address (authenticated device): `control///req/<req-id>`
        * message.address (unauthenticated device): `control/[tenant]/[device-id]/req/<req-id>`
        * message.subject: `<command>` (command encapsulated in the message's subject)

2. Device indicates that it can no longer receive commands from applications
   - device publishes a message to an address encapsulating the `unsubscribe` operation in the message's subject.
     * message.address (authenticated device): `control/+/+/req/#`
     * message.address (unauthenticated device): `control/[tenant-id]/[device-id]/req/#`
     * message.subject = `unsubscribe`

3. Device publishes a message in response to a command received from applications.
   * message.address (authenticated) = `control///res/<req-id>/<status>`
   * message.address (unauthenticated) = `control/[tenant]/[device-id]/res/<req-id>/<status>`

WDYT?

sophokles73:
> device publishes a message to an address encapsulating the subscribe operation in the message's subject.

No, the device opens a receiver link for that purpose, using an appropriate source address, e.g. `control/$tenant/$device/req` (where $tenant stands for the device's real tenant ID).

> Device publishes a message in response to a command received from applications.

[...]

PR #815, opened 2018-09-07
[#784] Command & Control Concepts over AMQP Adapter

Alfusainey:
@sysexcontrol: This patch describes the concepts for Command & Control over AMQP adapter 


Alfusainey:
@sysexcontrol any update on this?

sysexcontrol:
@Alfusainey I will review it tomorrow!

site/content/concepts/command-and-control.md
```diff
@@ -66,3 +66,24 @@ The response of the command will be sent by the device to
...
 
 ![Command & Control over MQTT Adapter](../command_control_concept_mqtt.png) 
+
+## Command & Control over AMQP Adapter
+
+When an authenticated device connected to the AMQP adapter wants to receive commands from the adapter, it opens a receiver link at the source address:
+
+* `control`
+
+while unauthenticated devices have to fully specify their `tenant` and `device-id` as source address.
```
sysexcontrol: It would be better to show the source address here as well (instead of putting it into words), like
`control/tenant/device-id`.

site/content/concepts/command-and-control.md
```diff
@@ -66,3 +66,24 @@ The response of the command will be sent by the device to
...
+
+## Command & Control over AMQP Adapter
+
+When an authenticated device connected to the AMQP adapter wants to receive commands from the adapter, it opens a receiver link at the source address:
+
+* `control`
+
+while unauthenticated devices have to fully specify their `tenant` and `device-id` as source address. 
+
+Once the link is opened, the device can receive command messages from the adapter. The AMQP adapter sets the
```
sysexcontrol: ... the receiver link is opened, ...

site/content/concepts/command-and-control.md
```diff
@@ -66,3 +66,24 @@ The response of the command will be sent by the device to
...
+`reply-to` and `subject` properties of the message as follows:
+
+* `reply-to`: `control///req/<req-id>`(authenticated device) or `control/tenant/device-id/<req-id>` (unauthenticated device). The `<req-id>` can be used by the device when sending the command response to the adapter.
```
sysexcontrol: I think this is wrong: I would set the `req-id` as property, which is then set again by the device when responding and  using it's `sender` link that has opened the `control` endpoint.
And the AMQP protocol adapter uses it to send the message to the application by using the endpoint `control/tenant/device-id/${reply-id}`. Remark: the reply-id is constructed from the req-id.
WDYT?

site/content/concepts/command-and-control.md
```diff
@@ -66,3 +66,24 @@ The response of the command will be sent by the device to
...
+* `subject` Contains the actual `<command>` to be executed by the device.
+
+Devices can publish command response messages (similar to how `telemetry` and `event` messages are published) using the following message address:
+
+* `control///res/<req-id>/<status>` (authenticated device)
```
sysexcontrol: IMHO this must be only `control`.
Containing the other params as properties.

site/content/concepts/command-and-control.md
```diff
@@ -66,3 +66,24 @@ The response of the command will be sent by the device to
...
+* `control/tenant/device-id/res/<req-id>/<status>` (unauthenticated device)
```
sysexcontrol: IMHO this must be `control/tenant/device-id`.
Containing the other params as properties.

site/content/concepts/command-and-control.md
Alfusainey: yes, i see! however, when publishing messages (events, telemetry, command responses) to the adapter, devices open an anonymous sender link (with no target address). The address is set in the AMQP message address

site/content/concepts/command-and-control.md
Alfusainey: yes, that make sense :+1:

site/content/concepts/command-and-control.md
Alfusainey: ok :+1:

Alfusainey:
@sysexcontrol: thanks for the feedback!

> in the sequence diagram the receiver link is opened from the device to the "control" endpoint. The transfer frame notes "control///req/${reqId}, though. IMHO these two must match, so the transfer goes to "control", and carry the "${reqId}" as property - or am I missing something?

Sure! i was including the entire address as `replyTo` property of the message. i will update the patch to incorporate your comments

site/content/concepts/command-and-control.md
Alfusainey: > Remark: the reply-id is constructed from the req-id.
WDYT?

yes, i think this is the default for all protocol adapters i.e command responses are sent to `control/tenant/device-id/${reply-id}`?

sysexcontrol: Ok, I am probably not deep enough into the AMQP protocol adapter yet, but I still do not understand why the complete `reply-to` property is important for the device?

IMHO only the `req-id` and the `subject` need to be set, right?
No matter if the sender link is anonymous or set with an address...

Alfusainey: > I still do not understand why the complete reply-to property is important for the device?
IMHO only the req-id and the subject need to be set, right?

yes, that's correct. it makes a lot of sense to not set the `reply-to` property and only set the `req-id` and `subject`..  I was thinking you meant from device -> adapter

Alfusainey:
@sysexcontrol: patch updated with 583bb9b

site/content/concepts/command-and-control.md
```diff
@@ -66,3 +66,31 @@ The response of the command will be sent by the device to
...
 ![Command & Control over MQTT Adapter](../command_control_concept_mqtt.png) 
+
+## Command & Control over AMQP Adapter
+
+When a device connected to the AMQP adapter wants to receive commands from the adapter, it opens a receiver link specifying the following source address:
+
+* `control` for authenticated devices
+* `control/tenant/device-id` for unauthenticated devices
+
+Once the receiver link is opened, the AMQP adapter sends command messages to devices through the link. The `subject` property of the message contains the actual command to be executed on the device while the `<request-id>`,used to correlate the command request with the response, is specified as an application property under key `APP_PROPERTY_REQUEST_ID`.
```
sophokles73: property names should be marked up using italics (i.e. wrapping them in asterisks). 
I do not think that the name of the property to convey the request ID is *APP_PROPERTY_REQUEST_ID*, or is it?

site/content/concepts/command-and-control.md
```diff
@@ -66,3 +66,31 @@ The response of the command will be sent by the device to
...
+
+Devices can publish command response messages by opening an **anonymous** sender link (similar to how `telemetry` and `event` messages are published) using the following message address:
+
+* `control` (authenticated device)
+* `control/tenant/device-id` (unauthenticated device)
```
sophokles73: please mark up the variables like this:
> `control/${tenant}/${device-id}`

site/content/concepts/command-and-control.md
```diff
@@ -66,3 +66,31 @@ The response of the command will be sent by the device to
 If the device is authenticated, the `tenant` and `device-id` are left empty (resulting in 3 subsequent `/`s).
 
 ![Command & Control over MQTT Adapter](../command_control_concept_mqtt.png) 
+
+## Command & Control over AMQP Adapter
+
+When a device connected to the AMQP adapter wants to receive commands from the adapter, it opens a receiver link specifying the following source address:
+
+* `control` for authenticated devices
+* `control/tenant/device-id` for unauthenticated devices
```
sophokles73: please mark up the variables like this:
> `control/${tenant}/${device-id}`

site/content/concepts/command-and-control.md
Alfusainey: welcome back @sophokles73. This is already a good start :+1: 

> I do not think that the name of the property to convey the request ID is APP_PROPERTY_REQUEST_ID, or is it?

I defined it as a constant in the `AmqpAdapterConstants.java` class. I think its better to omit mentioning the application key property from the docs, not so?

sophokles73: no

site/content/concepts/command-and-control.md
```diff
@@ -66,3 +66,31 @@ The response of the command will be sent by the device to
...
 ![Command & Control over MQTT Adapter](../command_control_concept_mqtt.png) 
+
+## Command & Control over AMQP Adapter
+
+When a device connected to the AMQP adapter wants to receive commands from the adapter, it opens a receiver link specifying the following source address:
+
+* `control` for authenticated devices
+* `control/${tenant}/${device-id}` for unauthenticated devices
+
+Once the receiver link is opened, the AMQP adapter sends command messages to devices through the link. The `subject` property of the message contains the actual command to be executed on the device while the `<request-id>`, used to correlate the command request with the response, is specified as an application property under key **APP_PROPERTY_REQUEST_ID**.
```
sophokles73: We should either explicitly name the properties used to convey the information, e.g. 

> The *subject* property of the message contains the name of the command to be executed on the device while the *request_id* application property contains the identifier used to correlate the command request with the response.

or we should simply mention that the message contains the command's name and a request identifier in (unspecific) properties.

site/content/concepts/command-and-control.md
```diff
@@ -66,3 +66,31 @@ The response of the command will be sent by the device to
...
+
+Devices can publish command response messages by opening an **anonymous** sender link (similar to how `telemetry` and `event` messages are published) using the following message address:
+
+* `control` (authenticated device)
+* `control/${tenant}/${device-id}` (unauthenticated device)
+
+The `<request-id>` and `<status>` are set as application properties of the AMQP 1.0 message sent over the sender link.
```
sophokles73: See above, either name the properties explicitly or do it like you already did (but then remove the brackets and formatting from the properties ...

site/content/concepts/command-and-control.md
```diff
@@ -66,3 +66,31 @@ The response of the command will be sent by the device to
...
+The device should set the `request_id` application property and the `status` application property on the command response message sent over the sender link. The `status` application property should contain the outcome of executing the command on the device.
+
+**Command Message Properties**
+
+The following table provides an overview of the properties sets on a command message.
+
+| Name            | Mandatory       | Location                 | Type        | Description |
+| :-------------- | :-------------: | :----------------------- | :---------- | :---------- |
+| *subject*       | yes             | *properties*             | *string*    | MUST contain the command name to be executed by a device. This property is set by the AMQP adapter when sending a command to a device. |
+| *status*        | yes             | *application-properties*               | *string*    | The status code indicating the outcome of processing the command by the device. MUST be set by the device after executing the command. |
```
sophokles73: this seems to be relevant for the response message only, isn't it?

Alfusainey:
@sophokles73 : thanks! i also squashed the commits into one

PR #818, opened 2018-09-10
Added provision to get application properties from command.

kaniyan:
while sending commands using a command client, there is a provision available in Hono to set application properties to a message. This PR provides a provision to retrieve application properties from a command, if it is set. So that the command consumer can make use of  them.

### DIFF
--- a/service-base/src/test/java/org/eclipse/hono/service/command/CommandTest.java
+++ b/service-base/src/test/java/org/eclipse/hono/service/command/CommandTest.java
@@ -14,17 +14,22 @@
 package org.eclipse.hono.service.command;
 
 import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.hamcrest.CoreMatchers.nullValue;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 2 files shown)


kaniyan:
Thank you for the comments. Changes have been made correspondingly.

service-base/src/test/java/org/eclipse/hono/service/command/CommandTest.java
```diff
@@ -79,17 +79,14 @@ public void testFromMessageSucceedsWithApplicationProperties() {
...
-        assertThat(cmd.getApplicationProperties(), is(notNullValue()));
         assertThat(cmd.getApplicationProperties(), is(notNullValue()));
+        assertThat(cmd.getApplicationProperties().size(), is(2));
         assertThat(cmd.getApplicationProperties().get("deviceId"), is("4711"));
         assertThat(cmd.getApplicationProperties().get("tenantId"), is("DEFAULT_TENANT"));
     }
 
     /**
-     * Verifies that a command can be created from a valid message with no application properties.
+     * Verifies that a command can be created from a valid message with no application properties, is valid.
```
sysexcontrol: ... a command created from a valid message without application properties is valid.

kaniyan: done :)

PR #820, opened 2018-09-11
Fix formatString errors

calohmn:


sysexcontrol:
@calohmn Thanks for contribution!

PR #831, opened 2018-10-01
Update Command & Control API to reflect usage of disposition frames.

sophokles73:


site/content/api/Command-And-Control-API.md
```diff
@@ -3,140 +3,106 @@ title = "Command & Control API"
...
-To send a notification command to a device, the first precondition needs to be fulfilled.
+**Examples**
 
-The following sequence diagram shows the message flow with a successful outcome:
+The following sequence diagram shows the successful delivery of a command to a device:
 
 ![Send Notification Command successful outcome](../command_control_notification_Success.png)
 
-The following sequence diagram shows the message flow with an unsuccessful outcome:
+The following sequence diagram shows how the delivery of a command fails because the device is not connected:
```
sysexcontrol: Later the third case is illustrated (delivery of a command failing since the command is malformed), so all cases are covered.
But IMHO it would be even clearer to have this third case covered here already, and then later only refer to the two failing deliveries (since they do not differ for both types of commands).
WDYT?

site/diagrams/sequenceDiagrams/command_control_Success.txt
```diff
@@ -1,9 +1,18 @@
...
+note over AMQP 1.0 Network
+forward command
+to device
+end note
+AMQP 1.0 Network->Business Application: disposition("accepted")
+note over AMQP 1.0 Network
+wait for response
+from device
+end note
+AMQP 1.0 Network->Business Application: transfer("control/${tenant_id}/${reply_id}", status)
```
sysexcontrol: `reply-id`

site/content/api/Command-And-Control-API.md
```diff
@@ -3,140 +3,106 @@ title = "Command & Control API"
...
 
-**Command** messages can be used to execute operations on devices, e.g. updating a configuration property, installing a software component or activating an actuator.
-Commands are bidirectional, i.e. a device receiving a command is expected to send a response to it.
-
-**Notification command** messages are unidirectional and cannot be responded.
+Commands can be used to trigger actions on devices. Examples include updating a configuration property, installing a software component or switching the state of an actuator.
 <!--more-->
 
+Hono distinguishes two types of commands. The first type is *one-way* only. In this case the sender of the command does not expect the device to send back a message in response to the command. Such commands may be used to e.g. *notify* a device about a change of state.
+
```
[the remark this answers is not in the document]
sophokles73: I am fine with using `command` for the default case (request/response) and `one-way command` for the other (special) case. I will change it accordingly throughout the API docs.

site/content/api/Command-And-Control-API.md
```diff
@@ -3,140 +3,106 @@ title = "Command & Control API"
...
-The following API defines operations that can be used by *Business Applications* to send commands (which get a response 
-in return) or notification commands to devices over an AMQP 1.0 Network.
+Hono indicates the outcome of the operation by means of the following AMQP delivery states:
 
-## Send a Notification Command
+| Delivery State | Description                                      |
+| :------------- | :----------------------------------------------- |
+| `accepted`    | The command has been delivered to the device for processing. |
+| `released`    | The command has not been delivered to the device because the device is (currently) not connected. |
+| `rejected`    | The command has not been delivered to the device because it does not contain all required information. |
```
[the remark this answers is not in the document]
sophokles73: I tend to agree but on the other hand, it is just a small table and I also like to have the information in place instead of having to click and jump to another place in the document from where you need to return then in order to continue reading. We also have two variants of the send command only so it doesn't seem to wasteful to have the same table in two places, does it?

site/content/api/Command-And-Control-API.md
sophokles73: sure, I can move the rejected case here as well.

site/content/api/Command-And-Control-API.md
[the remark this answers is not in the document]
sysexcontrol: Accepted - I can live with the two duplicated, but small tables... :+1:

sophokles73:
@sysexcontrol I have made the suggested changes ...

site/content/api/Command-And-Control-API.md
```diff
@@ -157,17 +160,18 @@ If a command message response contains a payload, the body of the message MUST c
 
 The following sequence diagram illustrates how a *Business Application* sends a command and receives the response from the device:
 
-![Send Command successful outcome](../command_control_Success.png)
+![Successfully send a Command](../command_control_success.png)
 
-The following sequence diagram illustrates how a malformed command sent by a *Business Application* gets rejected:
+The sending of a command may fail for the same reasons as those illustrated for sending a one-way command. Additonally, the sending of a command may be considered unsuccessful by an application if it does not receive the response from the device in a reasonable amount of time:
```
sysexcontrol: Additionally

Issue #825, opened 2018-09-20
New Bcrypt credentials enables DoS attack on the protocol adapters.


balsmn:
As per Credentials API, Hono is using Spring security for Bcrypt password encoding.

Spring security supports bcrypt password encoding with log rotations from 4 to 31, 10 being default (i.e:- 2^10 hash iterations). Once hashed, the log rotation used is part of the password like 

`$2a$21$LQdzaDSINBpHmYRiGEVtkOVt2GScZ6bPvEyoAIpiVIE6F4rviemdm`

where $21 represents the log rotation used. There is currently no max limit defined in Hono. If a user, say stores the password with higher rotation, then it could enable a DOS attack on the adapters.

I tested the time taken for `BCryptPasswordEncoder.matches` with higher log rotation on my laptop (Intel Core i7 - 2.7GHz 8 core, 32 GB RAM) it took

- 10 - 105 milliseconds
- 12 - 360 milliseconds
- 18 - 21 seconds
- 21 - 178 seconds
- 31 - more than 1 hour and I had to abort it.

Setting 31 as log rotation will definitely block the protocol adapters.

As a best practise, I would restrict the log rotation to 12 or 14, document them as part of credentials API and have this checked in 

org.eclipse.hono.util.CredentialsObject.hasValidSecrets()

sysexcontrol:
@balsmn Thanks a lot for this detailed report!
IMHO very reasonable, I will schedule a fix for the next milestone 0.8-M2.

sophokles73:
Nice catch @balsmn :+1: 
There seem to be two issues to consider in this context:
1. Make sure that we do not allow a DOS attack by registering a bcrypted password with more iteratins than seem reasonable.
2. Make sure that we do not block the event loop of protocol adapters that validate a password against a bcrypt hash.

The first item should probably be addressed by introducing a (configurable) upper limit to the Credentials endpoint implementation included in Hono. However, other Credentials API implementations that do not make use of the base classes provided by Hono will need to enforce such a limit themselves.

The second item needs to be addressed in Hono's `AbstractDeviceCredentials` class by scheduling the hash validation on a worker thread instead of running it on the event loop.

WDYT?

balsmn:
>> The first item should probably be addressed by introducing a (configurable) upper limit to the Credentials endpoint implementation included in Hono. However, other Credentials API implementations that do not make use of the base classes provided by Hono will need to enforce such a limit themselves.

[...]

PR #833, opened 2018-10-02
[#825] Run password hash validation on vert.x worker thread.

sophokles73:
The computation of a password hash can be quite time consuming if e.g.
bcrypt is used as the hash algorithm. Thus it seems feasible to move the
password hash validation from the event loop to a worker thread.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/auth/device/UsernamePasswordAuthProvider.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/auth/device/UsernamePasswordAuthProvider.java
@@ -13,12 +13,18 @@
 
 package org.eclipse.hono.service.auth.device;
 
+import java.net.HttpURLConnection;
 import java.util.Objects;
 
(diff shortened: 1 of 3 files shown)


service-base/src/main/java/org/eclipse/hono/service/auth/device/UsernamePasswordAuthProvider.java
```diff
@@ -72,4 +78,25 @@ protected DeviceCredentials getCredentials(final JsonObject authInfo) {
         }
     }
 
+    @Override
+    protected Future<Device> validateCredentials(
+            final DeviceCredentials deviceCredentials,
+            final CredentialsObject credentialsOnRecord) {
+
+        final Context currentContext = Vertx.currentContext();
+        if (currentContext == null) {
```
sysexcontrol: Since both branches (if and else) return at the end, it could be a perfect candidate to be written with the help of `Optional`. But more a matter of taste than a highly important improvement.

PR #834, opened 2018-10-02
Use OpenJDK 11 instead of OracleJDK 10 on Travis.

sophokles73:
This PR simply replaces the currently used OracleJDK 11 with OpenJDK 11 because OpenJDK is what we also use in our Docker images.


sophokles73:
@ctron would you mind taking a quick look at his PR?

Issue #557, opened 2018-04-03
Implement Jenkins build jobs as pipelines


sophokles73:
Hono's build jobs on Jenkins are currently configured using Jenkins web UI. This makes it hard to migrate the jobs to other build infrastructure. Re-implemening the jobs based on the Jenkins Pipeline plugin would allow us to keep the job definitions in GitHub and version them properly.

sophokles73:
fixed in https://github.com/eclipse/hono/commit/2c9cce242db7db6f75966a9a8c85dfea4b3bbba5

PR #839, opened 2018-10-09
[ #557] Added pipeline for Hono-Website build job

kaniyan:
Pipeline for Hono-Website build job

### DIFF
--- a/jenkins/Hono-Website-Pipeline.groovy
+++ b/jenkins/Hono-Website-Pipeline.groovy
@@ -1,0 +1,94 @@
+#!/usr/bin/env groovy
+
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+/**
+ * Builds Hono's web site using Hugo every night (between 3 and 4 AM) and publishes to Eclipse.
+ *
+ */
+
+node {
+    def utils = evaluate readTrusted("jenkins/Hono-PipelineUtils.groovy")
+    properties([buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '3')),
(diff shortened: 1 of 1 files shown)


jenkins/Hono-Website-Pipeline.groovy
```diff
@@ -0,0 +1,99 @@
...
+def build() {
+
+    stage('Checkout to sub-directory "hono"') {
+        echo "Check out master to subdirectory [hono]"
+        checkout([$class                           : 'GitSCM',
+                  branches                         : [[name: 'master']],
+                  doGenerateSubmoduleConfigurations: false,
+                  extensions                       : [[$class: 'RelativeTargetDirectory', relativeTargetDir: 'hono']],
+                  submoduleCfg                     : [],
+                  userRemoteConfigs                : [[credentialsId: 'bebef3c5-da22-425c-8554-f62d0b3a9608',
```
sophokles73: where does the credentialsId come from? Is this really something that we want to include in the code instead of e.g. passing it in as a configuration parameter?

kaniyan:
An internal unique ID by which credentials are identified from jobs and other configuration. Normally left blank, in which case an ID will be generated, which is fine for jobs created using visual forms. Useful to specify explicitly when using credentials from scripted configuration (Reference: CredentialsPlugin).

The credentials Id that I used was already available and I think it was configured by Eclipse. I think its a good idea to parameterize the Id too.

jenkins/Hono-Website-Pipeline.groovy
kaniyan: An internal unique ID by which credentials are identified from jobs and other configuration. Normally left blank, in which case an ID will be generated, which is fine for jobs created using visual forms. Useful to specify explicitly when using credentials from scripted configuration (Reference: CredentialsPlugin).

The credentials Id that I used was already available and I think it was configured by Eclipse. I think its a good idea to parameterize the Id too.

jenkins/Hono-Website-Pipeline.groovy
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Build stages for Hono's web site pipeline using Hugo
+ *
+ */
+def build() {
+
+    stage('Checkout to sub-directory "hono"') {
+        echo "Check out master to subdirectory [hono]"
+        checkout([$class                           : 'GitSCM',
```
sophokles73: Hmmm, I think we do not need this block at all and can instead simply do an anonymous checkout using a shell command as in the next stage. We do not change the content of the GitHub repo, thus IMHO we do not need to authenticate ...

sophokles73:
@kaniyan did you check if the pipeline runs successfully with your latest changes?

kaniyan:
@sophokles73 yes I did. https://ci.eclipse.org/hono/job/Hono-Web-Site_Pipeline/

PR #843, opened 2018-10-09
Checkout git repo without credentials for nightly build pipeline

kaniyan:
Hono repo can be checked out without credentials for the nightly build and this PR contains changes for that.


kaniyan:
@sophokles73 Done :-)

Issue #777, opened 2018-08-13
Add maven profile to include Jaeger client


calohmn:
It would be convenient to have a maven profile in Hono, with which Docker images of the Hono components get built that have the Jaeger OpenTracing client included.

sophokles73:
Sounds reasonable

ctron:
Yes, that would be great.

sophokles73:
@calohmn can this be closed with https://github.com/eclipse/hono/commit/4361c352d96ecbccd7f860268ec2d86606cb6247 ?

sophokles73:
@calohmn can this be closed now? If so, please do :-)

PR #844, opened 2018-10-11
[WIP] [#777] Add a build profile for Jaeger

ctron:


bom/pom.xml
```diff
@@ -346,6 +347,12 @@
         <version>${netty.tcnative.version}</version>
         <scope>runtime</scope>
       </dependency>
+      <!-- Jaeger support -->
+      <dependency>
+        <groupId>io.jaegertracing</groupId>
+        <artifactId>jaeger-client</artifactId>
```
sophokles73: I think that we will need a works-with CQ for this, won't we?

ctron: I guess this should be sufficient yes. I wouldn't really enable anything be default, as without a Jaeger server, it isn't really useful anyway.

sophokles73:
@ctron I am not sure if we are on the same page here. My understanding is that we add a profile for adding the Jaeger Java client to the Docker images we build for our services. This PR is **not** about adding a Jaeger backend to the deployment scripts.

I am asking because the CQ you have created seems to indicate otherwise ...

ctron:
Following up on the discussion from Eclipse Con … we indeed need a second works-with CQ for the Java Jaeger client. The reason for this CQ was the Jager agent, which I finally found in that repository: https://github.com/jaegertracing/jaeger/tree/master/cmd/agent … no too obvious from the root.

So I do think we need both. And I will create a CQ for the second one right now.

PR #848, opened 2018-10-11
[#557] Added pipeline for Hono Release

kaniyan:
Pipeline for Hono-Release build job

### DIFF
--- a/jenkins/Hono-Release-Pipeline.groovy
+++ b/jenkins/Hono-Release-Pipeline.groovy
@@ -80,7 +80,7 @@ def checkOut() {
 def setReleaseVersionAndBuild() {
     stage('Build') {
         withMaven(maven: 'apache-maven-latest', jdk: 'jdk1.8.0-latest', options: [jacocoPublisher(disabled: true), artifactsPublisher(disabled: true)]) {
-            sh "mvn versions:set -DgenerateBackupPoms=false -DnewVersion=${params.RELEASE_VERSION}"
+            sh "mvn versions:set -DgenerateBackupPoms=false -DnewVersion=${RELEASE_VERSION}"
             sh 'mvn clean install javadoc:aggregate -Dmaven.test.failure.ignore=false -DenableEclipseJarSigner=true -DsnapshotDependencyAllowed=false -Ddocker.skip.build=true'
         }
     }
--- a/jenkins/Hono-Release-Pipeline.groovy
+++ b/jenkins/Hono-Release-Pipeline.groovy
@@ -1,0 +1,143 @@
+#!/usr/bin/env groovy
+
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+/**
+ * Jenkins pipeline script for Hono release.
+ *
+ */
+
+node {
+    def utils = evaluate readTrusted("jenkins/Hono-PipelineUtils.groovy")
+    properties([buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '3')), parameters([
+            string(defaultValue: '',
+                    description: "The branch to build and release from.\nExamples: \n master\n1.0.x\n1.x",
+                    name: 'BRANCH',
+                    trim: true),
+            string(defaultValue: '',
+                    description: "The version identifier to use for the artifacts built and released by this job. \nExamples:\n1.0.0-M6\n1.0.0-RC1\n2.1.0",
+                    name: 'RELEASE_VERSION',
+                    trim: true),
+            string(defaultValue: '',
+                    description: "The version identifier to use during development of the next version.\nExamples:\n2.0.0-SNAPSHOT\n1.1.0-SNAPSHOT",
+                    name: 'NEXT_VERSION',
+                    trim: true),
+            string(defaultValue: '',
+                    description: "The version identifier to use during development of the next version.\nExamples:\n2.0.0-SNAPSHOT\n1.1.0-SNAPSHOT",
+                    name: 'CREDENTIALS_ID',
+                    trim: true)])])
+    try {
+        checkOut()
+        setReleaseVersionAndBuild()
+        commitAndTag()
+        setNextVersion()
+        commitAndPush()
+        copyArtifacts()
+        utils.publishJavaDoc('target/site/apidocs', true)
+        currentBuild.result = 'SUCCESS'
+    } catch (err) {
+        currentBuild.result = 'FAILURE'
+        echo "Error: ${err}"
+    }
+    finally {
+        echo "Build status: ${currentBuild.result}"
+        utils.notifyBuildStatus()
+    }
+}
+
+/**
+ * Checks out the specified branch from git repo
+ *
+ */
+def checkOut() {
+    stage('Checkout') {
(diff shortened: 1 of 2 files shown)


jenkins/Hono-Release-Pipeline.groovy
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * Jenkins pipeline script for Hono release.
+ *
+ */
+
+node {
+    def utils = evaluate readTrusted("jenkins/Hono-PipelineUtils.groovy")
+    echo "Start"
+    properties([buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '3')),
+                pipelineTriggers([cron('TZ=Europe/Berlin \n # every night between 3 and 4 AM \n H 3 * * *')])])
```
sophokles73: this job is triggered manually only

jenkins/Hono-Release-Pipeline.groovy
```diff
@@ -0,0 +1,142 @@
...
+    properties([buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '3')), parameters([
+            string(defaultValue: '',
+                    description: "The branch to build and release from.\nExamples: \n master\n1.0.x\n1.x",
+                    name: 'BRANCH',
+                    trim: false),
```
sophokles73: it would actually make sense if the input would be trimmed, I guess. the version identifiers and branch names do not start nor end with whitespace

jenkins/Hono-Release-Pipeline.groovy
```diff
@@ -0,0 +1,142 @@
...
+ */
+def checkOut() {
+    stage('Checkout') {
+        echo "Check out branch: origin/${params.BRANCH}"
+        checkout([$class                           : 'GitSCM',
+                  branches                         : [[name: "origin/${params.BRANCH}"]],
+                  doGenerateSubmoduleConfigurations: false,
+                  extensions                       : [[$class: 'WipeWorkspace'],
+                                                      [$class: 'LocalBranch']],
+                  userRemoteConfigs                : [[credentialsId: 'bebef3c5-da22-425c-8554-f62d0b3a9608', url: 'ssh://github.com/eclipse/hono.git']]])
```
sophokles73: Do we need to keep the credentialsId in the code?

jenkins/Hono-Release-Pipeline.groovy
```diff
@@ -0,0 +1,142 @@
...
+           '''
+    }
+}
+
+/**
+ * Copy the artifacts so that the binaries are available for download
+ *
+ */
+def deploy() {
+    stage("Deploy Artifacts") {
```
sophokles73: IMHO this should be rather named *Copy Artifacts` because *deploy* has a different meaning in the context of Maven (where it is used to upload the artifacts to a repository).

jenkins/Hono-Release-Pipeline.groovy
kaniyan: Its a mistake. I will change it.

jenkins/Hono-Release-Pipeline.groovy
kaniyan: During credentials configuration , it seems eclipse guys didn't set an Id and left it blank. Thereby jenkins created an ID as above. It is just an Id to identify the credentials stored in jenkins. I can parametrize it. But I don't see any real value in it. For example I would have created this ID as GitCredentialsId. During parametrization , I would set the paramter name as credentialsId and set the default value as GitCredentialsId. ($credentialsId=GitCredentialsId).

But I see some value in parametrization, if we may have in future more than one credentials in stored in jenkins and we can set the credentials Id accordingly.

sophokles73: I do not really understand. Is this just a logical identifier which we can make up ourselves or does this have to match the identifier that exists on our JIPP instance (and which has been created by the Eclipse folks)?

kaniyan: Credentials are stored in JIPP instance for hono by the Eclipse folks. An unique ID (by which these credentials are identified from jobs and other configuration) can be provided explicitly during storing of credentials. If left blank, then jenkins will create an ID as we see here (bebef3c5-da22-425c-8554-f62d0b3a9608)

Sample screenshot of a credentials configuration page is attached below.
image

jenkins/Hono-Release-Pipeline.groovy
```diff
@@ -0,0 +1,143 @@
...
+            string(defaultValue: '',
+                    description: "The version identifier to use for the artifacts built and released by this job. \nExamples:\n1.0.0-M6\n1.0.0-RC1\n2.1.0",
+                    name: 'RELEASE_VERSION',
+                    trim: true),
+            string(defaultValue: '',
+                    description: "The version identifier to use during development of the next version.\nExamples:\n2.0.0-SNAPSHOT\n1.1.0-SNAPSHOT",
+                    name: 'NEXT_VERSION',
+                    trim: true),
+            string(defaultValue: '',
+                    description: "The version identifier to use during development of the next version.\nExamples:\n2.0.0-SNAPSHOT\n1.1.0-SNAPSHOT",
```
sophokles73: I do not think that this is the correct description of the parameter, is it?
How will the user who starts the pipeline know the value to set for this parameter? When creating a build job manually, you can select the credentials from a drop down box. However, since this seems to be a build parameter, the user will probably need to enter the value manually ...

kaniyan: I think you are referring to the credentials parameter. I will use that instead of the current one.

sophokles73: ???

sophokles73: Ok, so you change the *type* of the parameter from `String` to `Credentials`, right?

jenkins/Hono-Release-Pipeline.groovy
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Commit and push to the Hono repo
+ *
+ */
+def commitAndPush() {
+    stage("Commit and push") {
+        sh ''' 
+            git add pom.xml \\*/pom.xml
+            git commit -m "Bump version to ${NEXT_VERSION}"
```
sophokles73: is there a reason, why the parameters are sometimes referred to as `${params.NEXT_VERSION}` and sometimes as just `${NEXT_VERSION}`?

jenkins/Hono-Release-Pipeline.groovy
kaniyan: yes, along with suitable arguments

jenkins/Hono-Release-Pipeline.groovy
kaniyan: Whenever I used sh command to execute shell scripts, I directly invoked the parameter without  prefix `params`.  The parameters are exposed as environment variables of the same name into shell.

jenkins/Hono-Release-Pipeline.groovy
```diff
@@ -0,0 +1,144 @@
...
+}
+
+/**
+ * Set version to RELEASE_VERSION and build using maven
+ *
+ */
+def setReleaseVersionAndBuild() {
+    stage('Build') {
+        withMaven(maven: 'apache-maven-latest', jdk: 'jdk1.8.0-latest', options: [jacocoPublisher(disabled: true), artifactsPublisher(disabled: true)]) {
+            sh "mvn versions:set -DgenerateBackupPoms=false -DnewVersion=${params.RELEASE_VERSION}"
```
sophokles73: shouldn't this be `${RELEASE_VERSION}` as well? Does it matter at all, i.e. do both versions work?

kaniyan: In this place it works fine. Today I tested with various scenarios and I found issues using ${params.XXX} inside multi-line shell scripts (sh ''' <content> '''). I would like to keep it uniform  throughout all shell script invocations, I will change ${params.RELEASE_VERSION} to ${RELEASE_VERSION}

PR #852, opened 2018-10-12
Hono command receiver links reconnects

sysexcontrol:
Contained is the recreation of command receiver links after connection losses in protocol adapters
Previously a connection loss recovered but did not reopen the command receivers again, so applications could not send commands anymore.


sysexcontrol:
Closing this PR due to concerns regarding bursts of link (re)creations if there were many links opened.

Favoured now is rather a protocol dependent solution, so e.g. for MQTT it may make more sense to close the corresponding MQTT connection and get the receiver link creation for free if the device reconnects and subscribes to the command topic.
Other protocol adapters may need different strategies, but the generic solution implemented here would need at least detailed optimizations, like strictly sequential recreation of links...

PR #856, opened 2018-10-17
Separate the mandatory tests from optinal for the device registry

jbtrystram:
Also, I removed the use of the REST API used in the test to create the resources before asserting them, and replaced it with a resource JSON file, as it is done in the other tests. In addition, if the REST call would fail for whatever reason, the assertion unit test would fail, causing some confusion.

This PR follows a discussion during the last community calls. This is early work to later offer some abstraction to third party developers that would want to use the hono test suite against their own device registry implementation.


tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationMandatoryAmqpIT.java
```diff
@@ -0,0 +1,144 @@
...
+import org.junit.runner.RunWith;
+
+import java.net.HttpURLConnection;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Tests verifying the behavior of the Device Registry component's Device Registration AMQP endpoint.
+ */
+@RunWith(VertxUnitRunner.class)
+public class DeviceRegistrationMandatoryAmqpIT {
```
sophokles73: I would rather have the classes named the other way around, i.e. have the mandatory tests in `DeviceRegistrationAmqpIT` and the optional ones in `DeviceRegistrationOptionalAmqpIT`. This way, the optional tests become the *special* case which is more appropriate FMPOV.

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationMandatoryAmqpIT.java
```diff
@@ -0,0 +1,144 @@
...
+            .compose(ok -> deviceRegistryclient.getOrCreateRegistrationClient(Constants.DEFAULT_TENANT))
+            .compose(client -> client.assertRegistration(deviceId))
+            .setHandler(ctx.asyncAssertFailure(t -> assertErrorCode(t, HttpURLConnection.HTTP_NOT_FOUND, ctx)));
+    }
+
+    private static void assertErrorCode(final Throwable error, final int expectedErrorCode, final TestContext ctx) {
+        ctx.assertTrue(error instanceof ServiceInvocationException);
+        ctx.assertEquals(expectedErrorCode, ((ServiceInvocationException) error).getErrorCode());
+    }
+}
```
sophokles73: missing EOL

tests/src/test/resources/deviceregistry/application.yml
```diff
@@ -39,5 +41,4 @@ management:
     enabled: false
     export:
       graphite:
-        enabled: false
-    
\ No newline at end of file
+        enabled: false
```
sophokles73: missing EOL

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationMandatoryAmqpIT.java
jbtrystram: make sense to me !

jbtrystram:
There is a conflict with `tests/src/test/resources/deviceregistry/application.yml`, I  don't get what is blocking the merging.. I've tried to merge but that lead to re-commiting a bunch of changes which is odd. Need a bit of help from a git master :)

sophokles73:
The tests regarding the assertion of a device connected via a gateway are not optional. They are part of the mandatory operation *assert Device Registration*. So IMHO they need to remain in the `DeviceRegistrationAmqpIT` class ....

sophokles73:
Can you rebase and resolve the conflict in the application.yml file?

jbtrystram:
> The tests regarding the assertion of a device connected via a gateway are not optional. They are part of the mandatory operation _assert Device Registration_. So IMHO they need to remain in the `DeviceRegistrationAmqpIT` class ....

The gateway field is described as not mandatory in the API documentation, and that the implementation "SHOULD" verify that the gateway is authorized. that was why I moved them.

dejanb:
I agree, let's keep at least testAssertRegistrationFailsForNonExistingGateway() as mandatory and make sure implementations properly handle this case.

jbtrystram:
Thanks for your comments. Done !

dejanb:
One last thing is that we could also preload devices for gateway tests in the device-identities.json, just to be consistent for all tests.

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationAmqpIT.java
```diff
@@ -144,79 +140,22 @@ public void testAssertRegistrationFailsForDisabledDevice(final TestContext ctx)
...
-
-        final String deviceId = helper.getRandomDeviceId(Constants.DEFAULT_TENANT);
-        final String gatewayId = helper.getRandomDeviceId(Constants.DEFAULT_TENANT);
-        helper.registry.registerDevice(Constants.DEFAULT_TENANT, gatewayId)
-            .compose(ok -> helper.registry.registerDevice(Constants.DEFAULT_TENANT, deviceId, new JsonObject()
-                    .put(PROPERTY_VIA, "not-" + gatewayId)))
-            .compose(ok -> deviceRegistryclient.getOrCreateRegistrationClient(Constants.DEFAULT_TENANT))
-            .compose(client -> client.assertRegistration(deviceId, gatewayId))
-            .setHandler(ctx.asyncAssertFailure(t -> assertErrorCode(t, HttpURLConnection.HTTP_FORBIDDEN, ctx)));
+                .compose(ok -> deviceRegistryclient.getOrCreateRegistrationClient(Constants.DEFAULT_TENANT))
```
sophokles73: I guess you won't want to register a device here using the helper because the registry under test will probably not expose a corresponding HTTP based API, right?

Instead, the test should probably use the *pre-configured* device 4711 (which, BTW, should be defined as a constant IMHO).

jbtrystram: I think I noticed and removed this as you typed this comment :)

jbtrystram: @sophokles73 As i updated the JSON file to use pre-configured devices I added a few more devices to cover the various test cases. 
Do you suggest I define constants for all the devices ?

sophokles73: > Do you suggest I define constants for all the devices ?

I think that would be a good idea. Maybe you can name the constants according to the characteristics of the device(s), e.g. *DEVICE_ID_ENABLED* and *DEVICE_ID_DISABLED*?

I think that the mandatory tests should also include a test for asserting a non-existing device ID.

jbtrystram: > I think that the mandatory tests should also include a test for asserting a non-existing device ID.
There is `testAssertRegistrationFailsForUnknownDevice` which verifies that the service reply 404 for a non-existing device.

sophokles73: oh, sure. My bad ...

jbtrystram: >Maybe you can name the constants according to the characteristics of the device(s), e.g. _DEVICE_ID_ENABLED_ and _DEVICE_ID_DISABLED_?

I updated the PR with that.

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationOptionalAmqpIT.java
```diff
@@ -0,0 +1,142 @@
...
+     * Verifies that the registry fails a gateway's request to assert a
+     * device's registration status for which it is not authorized with a 403 error code.
+     *
+     * @param ctx The vert.x test context.
+     */
+    @Test
+    public void testAssertRegistrationFailsForUnauthorizedGateway(final TestContext ctx) {
+
+        deviceRegistryclient.getOrCreateRegistrationClient(Constants.DEFAULT_TENANT)
+            .compose(client -> client.assertRegistration(DEVICE_ID_ENABLED_WITH_ENABLED_GATEWAY, NON_EXISTING_GATWAY_ID))
```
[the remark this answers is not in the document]
jbtrystram: This had me thinking as well, I agree with your opinion on this. I'll update the PR :)
For reference though, I took the same route as the existing tests which prefix an existing gateway ID with "not" ( see https://github.com/eclipse/hono/blob/1597fdd4fe1cdf42213a5fbc630ce197d5e10d93/tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationAmqpIT.java#L212 )

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationOptionalAmqpIT.java
```diff
@@ -0,0 +1,142 @@
...
+ * Tests verifying the behavior of the Device Registry component's Device Registration AMQP endpoint.
+ */
+@RunWith(VertxUnitRunner.class)
+public class DeviceRegistrationOptionalAmqpIT {
+
+    private static final String DEVICE_ID_ENABLED_WITH_ENABLED_GATEWAY = "4712";
+    private static final String GATEWAY_ID_ENABLED = "gw-1";
+    private static final String DEVICE_ID_ENABLED_WITH_DISABLED_GATEWAY = "4713";
+    private static final String GATEWAY_ID_DISABLED = "gw-2";
+    private static final String NON_EXISTING_GATWAY_ID = "non-existing-gateway";
```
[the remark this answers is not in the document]
jbtrystram: We could but then `DeviceRegistrationOptionalAmqpIT` would need to extends `DeviceRegistrationAmqpIT`. Would not this results in duplicated tests when running the test suite ?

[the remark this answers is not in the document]
sophokles73: No, we would simply need to define the constant in `DeviceRegistrationAmqpIT` with package private scope:
```
static final String NON_EXISTING_GATEWAY_ID = "";
```
We can then refer to it by means of `DeviceRegistrationAmqpIT.NON_EXISTING_GATEWAY_ID` ...

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationOptionalAmqpIT.java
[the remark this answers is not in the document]
jbtrystram: I finally added a third gateway for this case, as testing with `GATEWAY_ID_DISABLED` would be ambiguous.

Issue #855, opened 2018-10-16
Command and Control classes need to be contained and OSGi-exported in hono-client


sysexcontrol:
Currently, the Command and Control classes for offering this feature in protocol adapters are contained in the `hono-service-base` module, which is intentionally not an OSGi-bundle.

To enable writing a protocol adapter that needs to run in an OSGi environment these classes need to be available in the `hono-client` module with the correct packaging to export the necessary classes.

sysexcontrol:
Fixed by ffca900.

PR #857, opened 2018-10-17
[#855] Provide Command and Control classes in hono-client.

sysexcontrol:
Additionally, all necessary classes are exported to OSGi
classloaders, so a protocol adapter with support for
Command and Control may be running in an OSGi environment.


core/src/main/java/org/eclipse/hono/auth/Device.java
```diff
@@ -22,23 +22,18 @@
...
  * An authenticated client of a protocol adapter representing a device.
  * <p>
  * The device's identity and authorities are contained in a Java Web Token.
  */
-public class Device implements User {
+public class Device {
 
     private final JsonObject principal;
-    private final Set<Object> authorities = new HashSet<>();
+    protected final Set<Object> authorities = new HashSet<>();
```
sophokles73: I would rather not give subclasses access to this field because it contains the security relevant information that should be immutable once created. Instead, I propose to add a method for checking if the device has a particular authority here in this class. Subclasses can then use this method for implementing e.g. the `User` interface.

site/content/release-notes.md
```diff
@@ -2,6 +2,23 @@
...
+* The `hono-client` module now contains all classes necessary to implement Command and Control in protocol adapters.
+  Previously this has only been the case for the sending of a command, as it is typically done by an application, while
+  the classes to receive commands, typically used by protocol adapters, were located in the `hono-service-base` module.
+  Additionally, the package structure was reworked to allow for implementing protocol adapters that run in an OSGi 
+  environment, so several classes are not in the same package anymore.
+  Custom protocol adapters thus may need to be slightly refactored to import the Command classes
+  from their new packages - the functionality has not changed.
+  The only exception to this is the `Device` class that was moved to `hono-core` with a specific
+  subclass `DeviceUser`. This subclass is needed in code lines that implement the authentication as defined by the `vertx-auth-common` 
+  module, so for this class there might be some very few changes to the custom code necessary (the adopted standard protocol
```
sophokles73: *adopted* ?

site/content/release-notes.md
```diff
@@ -2,6 +2,23 @@
 title = "Release Notes"
 +++
 
+## 0.8
```
sophokles73: can you append a `(not released yet)`?

site/content/release-notes.md
sysexcontrol: will change :+1:

core/src/main/java/org/eclipse/hono/auth/Device.java
sysexcontrol: :+1: will change this.

sysexcontrol:
@sophokles73 : pushed changes according to review, rebased to recent master changes.

core/src/main/java/org/eclipse/hono/auth/Device.java
```diff
@@ -85,48 +80,36 @@ private JsonObject getPrincipal(final String tenantId, final String deviceId) {
      * In order for the check to succeed, the JWT must
      * <ul>
      * <li>not be expired</li>
-     * <li>contain the given authorities in its <em>aut</em> claim</li>
+            * <li>contain the given authorities in its <em>aut</em> claim</li>
```
sophokles73: indentation

client/src/main/java/org/eclipse/hono/client/AbstractConsumer.java
```diff
@@ -11,14 +11,13 @@
  * SPDX-License-Identifier: EPL-2.0
  *******************************************************************************/
 
-package org.eclipse.hono.client.impl;
+package org.eclipse.hono.client;
```
sophokles73: I wonder if it is necessary to move the abstract base classes from impl to the top level package. In the end they are implementation detail and are not referenced in any of the interfaces/types, right? So, if possible, IMHO we should keep them in the impl package. The concrete implementations are created using factory methods in the exported interfaces anyway ....

sysexcontrol: Probably you are right - I hope that this does not cause visibility problems in an OSGi environment.
Will soon push an update where the classes are moved, if this does not work we can move them back anyway.

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -413,7 +413,7 @@ public final void setRequestTimeout(final long timoutMillis) {
      * @param delivery The handle for accessing the message's disposition.
      * @param message The response message.
      */
-    protected final void handleResponse(final ProtonDelivery delivery, final Message message) {
+    public final void handleResponse(final ProtonDelivery delivery, final Message message) {
```
sophokles73: is there a particular reason why this method needs to become public?

sysexcontrol: After I moved the abstract base classes (back) to the impl package: no, not anymore.
Previously some test classes needed the `public` modifier.
I will switch it back to `protected`.

sophokles73:
@sysexcontrol I also think that we should ask @ctron and @dejanb about their opinion on this change, e.g. by letting them review this PR ...

sysexcontrol:
@sophokles73 Thanks for the detailed review. I fully agree to your proposals and will ask for your a review after I squashed everything into a single commit.

sysexcontrol:
@ctron @dejanb : may I ask you to take a look to this (not entirely small) refactoring of the `hono-client` module?
The background is to make the necessary classes for protocol adapters available to OSGi-environments, so there are several classes moved, and/or made available in the top package of the `hono-client` to obtain class loader visibility.
Thanks!

PR #861, opened 2018-10-19
[#557] Added pipelines to deploy artifacts to maven central and eclipse repos

kaniyan:
Added pipelines to build and deploy hono artifacts  to the maven central and the eclipse repos.

### DIFF
--- a/jenkins/Hono-Deploy-Eclipse-Pipeline.groovy
+++ b/jenkins/Hono-Deploy-Eclipse-Pipeline.groovy
@@ -1,0 +1,62 @@
+#!/usr/bin/env groovy
+
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+/**
+ * Jenkins pipeline script that checks out ${RELEASE_VERSION}, builds all artifacts to deploy,
+ * signs them and creates PGP signatures for them and deploys artifacts to Eclipse Release repo.
+ *
+ */
+
+node {
+    def utils = evaluate readTrusted("jenkins/Hono-PipelineUtils.groovy")
+    properties([buildDiscarder(logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '3')), parameters([
+            string(defaultValue: '',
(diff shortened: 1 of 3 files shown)


jenkins/Hono-Deploy-Eclipse-Pipeline.groovy
```diff
@@ -0,0 +1,62 @@
...
+ * Build and deploy with maven.
+ *
+ */
+def buildAndDeploy() {
+    stage('Build and deploy to maven central') {
+        withMaven(maven: 'apache-maven-latest',
+                jdk: 'jdk1.8.0-latest',
+                mavenLocalRepo: '.repository',
+                options: [jacocoPublisher(disabled: true), artifactsPublisher(disabled: true)]) {
+            sh "mvn --projects :hono-service-auth,:hono-service-messaging,:hono-service-device-registry,:hono-adapter-http-vertx,:hono-adapter-mqtt-vertx,:hono-adapter-kura,:hono-adapter-amqp-vertx,:hono-example,:hono-cli -am deploy -DskipTests=true -DcreateJavadoc=true -DenableEclipseJarSigner=true -DskipStaging=true"
```
sophokles73: is there another, probably more structured, way of setting the maven options or do we need to use this *shell* command?

kaniyan: withMaven(...) method provides a way to set mavenOpts as an argument and it is applicable to all the shell commands within this block. I will this try this out.

sophokles73: well, if it only means that *all* options end up in a single separate string, then I guess we can stick with the existing approach. I was hoping that there would be named parameters like *projects*, *properties* and/or *targets* which would help in making the command more readable ...

kaniyan: Based on the Pipeline Maven Integration Plugin documentation, I don't find any named parameters like _projects_ or _targets_. _MavenOpts_ can be used to specify JVM specific options as in the below example, so that only this part can be removed from the _shell_ script.

``` withMaven(maven: 'apache-maven-latest',
                jdk: 'jdk1.8.0-latest',
                mavenLocalRepo: '.repository',
                mavenOpts: "-DskipTests=true -DcreateJavadoc=true -DenableEclipseJarSigner=true -DskipStaging=true",
                options: [jacocoPublisher(disabled: true), artifactsPublisher(disabled: true)]) {
            sh "mvn --projects :hono-service-auth,:hono-service-messaging,:hono-service-device-registry,:hono-adapter-http-vertx,:hono-adapter-mqtt-vertx,:hono-adapter-kura,:hono-adapter-amqp-vertx,:hono-example,:hono-cli -am deploy "
        }
```

sophokles73: That's not really helpful, is it?

kaniyan: Not really. I would stick with the existing approach.

sophokles73:
@kaniyan you need to sign-off your commit ...

Issue #832, opened 2018-10-01
MQTT adapter should support commands to be published using QoS 1


sophokles73:
The MQTT adapter currently only supports publishing of commands to devices using QoS 0. The adapter therefore gets no indication whether the command has been accepted by the device and thus immediately sends back an *accepted* disposition to the application as soon as it has published the command (using QoS 0).

We already allow devices to publish their response to a command using QoS 1 so it seems to make sense to support QoS 1 for the other direction as well. Allowing the device to indicate a QoS level of 1 in its SUBSCRIBE packet would give the adapter a chance to use the PUBACK from the device as a trigger to send back the *accepted* disposition to the application. If no PUBACK arrives from the device the adapter could either simply do nothing, i.e. do not send back a disposition frame at all or it could send back a *released* disposition after a reasonable amount of time.

sophokles73:
depends on #921

PR #869, opened 2018-10-29
[#832] Add support for publishing commands with QoS 1 in MQTT adapter

kaniyan:
With this PR, MQTT adapter supports publishing commands with QoS 1 [#832](https://github.com/eclipse/hono/issues/832). If a device subscribes with QoS 1 to a control endpoint, then MQTT adapter also publishes commands to that device with QoS 1. When a device subscribes with QoS 0, then commands are published using QoS 0.


adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1209,24 +1214,34 @@ protected final void onCommandReceived(
         // example: control/DEFAULT_TENANT/4711/req/xyz/light
         final String topic = String.format("%s/%s/%s/%s/%s/%s", subscription.getEndpoint(), tenantId, deviceId,
                 subscription.getRequestPart(), command.getRequestId(), command.getName());
+        LOG.debug("Publishing command to device [tenant-id: {}, device-id: {}, MQTT client-id: {}, QoS: {}]",
+                subscription.getTenant(), subscription.getDeviceId(), endpoint.clientIdentifier(), qos);
         endpoint.publish(topic, command.getPayload(), qos, false, false);
+        endpoint.publishAcknowledgeHandler(
```
sophokles73: I think this should go into an `else if` block of the following `if` statement ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1209,24 +1214,34 @@ protected final void onCommandReceived(
...
+                messageId -> afterCommandPublished(endpoint, qos, subscription, topic, commandContext));
```
sophokles73: this is not how the *publishAcknowledgeHandler* works, is it? The handler is supposed to process the message ID that is being acknowledged by the peer. So I guess it needs to be passed in to the handler method as a parameter. This, however, requires that you keep track of the unacknowledged message IDs somewhere ...

kaniyan: I got it. I will make use of the message ID. Currently there is no provision in vertx-mqtt library to get the message ID while publishing a message so that it can be matched with the message ID during acknowledgement. There is a method `lastMessageId()`, which can be used to retrieve the message ID of the last sent message. In some cases, its not sure which message ID will be return. 
A new method `publish(String topic, Buffer payload, MqttQoS qosLevel, boolean isDup, boolean isRetain, Handler<AsyncResult<Integer>> publishSentHandler)` is commited into vertx-mqtt master from [PR#86](https://github.com/vert-x3/vertx-mqtt/pull/86), which can be used. Should we wait for the release of next release of vertx-mqtt (3.6.0)? WDYT.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: From the functionality perspective, it behaves the same if it is here or inside `else if` block. But I agree that we need not register the handler always, I will change it.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
sophokles73: I think that would be the easiest way. In the meantime we will live with QoS 0 for publishing commands to devices. I was actually surprised to discover that this method isn't in the MqttEndpoint interface already. It is in the MqttClient and I thought that it should be in the endpoint as well ...

sophokles73: now that we have upgraded to vert.x 3.6, you may want to continue to work on this?

kaniyan: I am already working on it :-)

kaniyan:
`CommandSubscription` can be used to further encapsulate `CommandContext` to acheive that. I will refactor it.

Regarding the timeout, FMPOV I think it makes sense if we use the request timeout value set during sending of commands by the business applications. WDYT?

sophokles73:
Hm, I do not really think that we need to encapsulate the `CommandContext`. My intention was more to put the map for keeping track of the sent commands and the expected PUBACKs into the `CommandSubscription`.

> I think it makes sense if we use the request timeout value set during sending of commands by the business applications. WDYT?

The request timeout covers the whole round-trip, i.e. sending the command to the device as well as receiving the response (if not a one-way command). I think that we should use a much more *aggressive*, i.e. small, timeout for sending the command to the device because we should be able to indicate to the command client rather sooner than later that we weren't (even) able to send the command. I would therefore propose to use a separate configuration property for this on the MQTT adapter and use quite a small default value.

kaniyan:
Done :+1:

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1148,14 +1149,92 @@ protected final void onCommandReceived(
...
+                } else {
+                    afterCommandPublished(subscription, commandContext);
+                }
+            } else {
+                LOG.error(
+                        "Error publishing command to device [tenant-id: {}, device-id: {}, MQTT client-id: {}, QoS: {}]",
+                        subscription.getTenant(), subscription.getDeviceId(), endpoint.clientIdentifier(),
+                        subscription.getQos(),
+                        publishSentHandler.cause());
+                TracingHelper.logError(commandContext.getCurrentSpan(), publishSentHandler.cause());
```
sophokles73: IMHO the command needs to be released in this case, doesn't it?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1148,14 +1149,92 @@ protected final void onCommandReceived(
         final String commandRequestId = (command.isOneWay() ? "" : command.getRequestId());
         final String topic = String.format("%s/%s/%s/%s/%s/%s", subscription.getEndpoint(), tenantId, deviceId,
                 subscription.getRequestPart(), commandRequestId, command.getName());
-        endpoint.publish(topic, command.getPayload(), qos, false, false);
+        LOG.debug("Publishing command to device [tenant-id: {}, device-id: {}, MQTT client-id: {}, QoS: {}]",
+                        subscription.getTenant(), subscription.getDeviceId(), subscription.getClientId(), subscription.getQos());
+        commandContext.getCurrentSpan().log(String.format("Publishing command to device [tenant-id: %s, device-id: %s, MQTT client-id: %s, QoS: %s]",
```
sophokles73: the tenant and device ID will already be present in the span as tags. Please use a Map to log the event, client ID and QoS in a structured way

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1148,14 +1149,92 @@ protected final void onCommandReceived(
...
+                subscription.getTenant(), subscription.getDeviceId(), subscription.getClientId(), subscription.getQos()));
+
+        if (MqttQoS.AT_LEAST_ONCE.equals(subscription.getQos())) {
+            registerPublishAcknowledgeHandler(endpoint, subscription);
```
sophokles73: My understanding of vertx-mqtt is that the PUBACK handler is registered once per endpoint only, or is there a particular reason why we need to do this *for each* command? What if an application sends two commands in direct succession, i.e. without waiting for the first command's response to arrive before sending the second one?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: yes, it needs to be released. I have updated the code.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: Done

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: It makes sense to register once per endpoint. I will push the changed version.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -502,6 +502,11 @@ protected final void onSubscribe(
                     // we do not support subscribing to commands using QoS 2
                     result = Future.failedFuture(new IllegalArgumentException("QoS 2 not supported for command subscription"));
                 } else {
+
+                    if (MqttQoS.AT_LEAST_ONCE.equals(subscription.qualityOfService())) {
```
sophokles73: Now we are registering the handler for each topic that the device subscribes to, right? IMHO this means that we won't be able to handle PUBACKs received for messages sent in the context of different topics, right?

I think that we need to register a PUBACK handler at the endpoint level, i.e. a (single) handler that will be responsible for handling *all* PUBACKs we receive from the device. Maybe a new class `Subscriptions` would be helpful in which we keep track of all topics that the device has subscribed to and which can be used as the PUBACK handler ...

kaniyan: In order to handle subscription of multiple topics by devices, it makes sense to register PUBACK handler at the endpoint level. Only `msg-id` is available during PUBACKs. I didn't get it how the `Subscriptions` with track of all the subscribed topics would help. IMHO the `msg-id` of the published messages (with Qos 1) along with corresponding subscription info, command context and `timer-id` need to be stored. When any PUBACK is received from the device, the `msg-id`can be used to retrieve the stored values and respond accordingly. WDYT?

kaniyan: In this approach with `Subscriptions`,  the topics that are being unsubscribed should also be taken care of. IMHO its more straightforward if we keep track of all the published msg-ids (with Qos1) along with respective CommandSubscription and CommandContext. For example in a class `PublishAcknowledgementHandler` which also can be used as the PUBACK handler. WDYT?
```
class PublishAcknowledgementHandler implements Handler<Integer> {
/**Map<msgId, TriTuple<TimerId, CommandSubscription, CommandContext>>**/
private Map<Integer, TriTuple<Long, CommandSubscription, CommandContext>> publishedWithQoS1AndWaitingForAcknowledgement = new ConcurrentHashMap<>();
```

sophokles73: > In this approach with Subscriptions, the topics that are being unsubscribed should also be taken care of.

I agree, that is why I proposed to use it. The Subscriptions class can provide a method that accepts a message ID and handles all PUBACKs of the device. There is no need to implement `Handler<Integer>` because you can simply register the `Subscriptions.handlePubAck` method with the endpoint.

kaniyan: @sophokles73 During unsubscribe, the topic entries in Subscriptions class cannot be removed if there are any active acknowledgement timers. Setting a timer (with ack timeout value) and removing the entries should work. In order to avoid such overheads, I have also already worked on a slightly different approach. Can you pls take a quick look at the `AbstractVertxBasedMqttProtocolAdapter` class and provide your comments?

sophokles73: I think this goes in the right direction :-)
I wonder if we can add the `CommandConsumer` instance to the `CommandSubscription` class so that we have a handle on the corresponding consumer when a device unsubscribes from a topic.
I also wonder if we can also use `CommandAcknowledgementHandler` to encapsulate the device's subcriptions and then only pass around the `CommandAcknowledgementHandler` instance instead of `CommandAcknowledgementHandler` + `CommandSubscription` ...

kaniyan: > I think this goes in the right direction :-)

:+1: 

> I wonder if we can add the `CommandConsumer` instance to the `CommandSubscription` class so that we have a handle on the corresponding consumer when a device unsubscribes from a topic. I would like to get rid of `CommandConnection.closeCommandConsumer()` ...

> I also wonder if we can also use `CommandAcknowledgementHandler` to encapsulate the device's subcriptions and then only pass around the `CommandAcknowledgementHandler` instance instead of `CommandAcknowledgementHandler` + `CommandSubscription` ...

IMHO I would rather keep `CommandAcknowledgementHandler` only to handle the acknowledgement of published commands with Qos 1. Currently it tracks only the published commands (only Qos 1) using MsgId (as map key) and the corresponding  CommandSubscription, CommandContext and TimerIds (as map value).  So it can't be used to keep track of subscriptions or command consumers. For example, a published message (with QoS 0) will have no entry in the  map of `CommandAcknowledgementHandler`.

sophokles73: > So it can't be used to keep track of subscriptions or command consumers. For example, a published message (with QoS 0) will have no entry in the map of CommandAcknowledgementHandler.

I know. That's why I think it should be named differently (e.g. `Subscriptions`) and be extended to also contain a list of the device's subscriptions (or just the device's `CommandSubscription` for the time being). It can still also contain the map of message IDs that are pending to be ACKed, but it could also be used to keep track of the subscriptions.

kaniyan: @sophokles73 I have named it to `CommandHandler` (since it has handlers to subscribe, unsubscribe and puback for commands). Apart from tracking pubacks, it also tracks all the subscriptions to a device.

> I wonder if we can add the `CommandConsumer` instance to the `CommandSubscription` class so that we have a handle on the corresponding consumer when a device unsubscribes from a topic. I would like to get rid of `CommandConnection.closeCommandConsumer()` ...

This `CommandHandler` also contains reference to the `CommandConsumer` instances, so I am able to get rid of the `CommandConnection.closeCommandConsumer()`

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1227,4 +1008,406 @@ private static MqttConnectReturnCode getConnectReturnCode(final Throwable e) {
             return MqttConnectReturnCode.CONNECTION_REFUSED_NOT_AUTHORIZED;
         }
     }
+
+
+    /**
+     * A class that contains handlers for command subscription, unsubscription and sending of commands.
+     */
+    protected class CommandHandler {
```
sophokles73: The way this class is currently implemented, it is very tightly coupled to its enclosing class because it accesses many of its private fields and methods.
IMHO this is not very desirable because it makes it hard(er) to test. Maybe we can keep the OpenTracing related stuff and the logic for processing the MQTT messages in the enclosing class and use the `CommandHandler` only for keeping the state. WDYT?

kaniyan: I will use the `CommandHandler `only for keeping the state (subscriptions and pubacks).  I will move back `onSubscribe`, `onUnsubscribe`, `onCommandReceived` and related methods to the enclosed class as before.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttProtocolAdapterProperties.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.eclipse.hono.config.ProtocolAdapterProperties;
+
+/**
+ * Properties for configuring an Mqtt based protocol adapter.
+ *
+ */
+public class MqttProtocolAdapterProperties extends ProtocolAdapterProperties {
+
+    private static final long DEFAULT_COMMAND_ACKNOWLEDGEMENT_TIME_OUT = 100L;
```
sophokles73: this needs to be at least *protected* in order for the link to this field from the method JavaDoc to succeed ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -470,13 +475,15 @@ private void handleConnectionRequestResult(final MqttEndpoint endpoint,
     protected final void onSubscribe(
             final MqttEndpoint endpoint,
             final Device authenticatedDevice,
-            final MqttSubscribeMessage subscribeMsg) {
+            final MqttSubscribeMessage subscribeMsg,
+            final CommandHandler cmdHandler) {
```
sophokles73: missing JavaDoc

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -604,7 +614,8 @@ private Span newSpan(final String operationName, final MqttEndpoint endpoint, fi
     protected final void onUnsubscribe(
             final MqttEndpoint endpoint,
             final Device authenticatedDevice,
-            final MqttUnsubscribeMessage unsubscribeMsg) {
+            final MqttUnsubscribeMessage unsubscribeMsg,
+            final CommandHandler cmdHandler) {
```
sophokles73: missing JavaDoc

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1137,14 +1164,13 @@ protected void onMessageUndeliverable(final MqttContext ctx) {
     protected final void onCommandReceived(
             final MqttEndpoint endpoint,
             final CommandSubscription subscription,
-            final CommandContext commandContext) {
+            final CommandContext commandContext,
+            final CommandHandler cmdHandler) {
```
sophokles73: missing JavaDoc

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1226,4 +1300,110 @@ private static MqttConnectReturnCode getConnectReturnCode(final Throwable e) {
...
+
+        /**
+         * Stores the published message id along with command subscription and command context.
+         *
+         * @param msgId The id of the command (message) that has been published.
+         * @param commandContext The commandContext of the command sent.
+         *
+         */
+        protected final void addToWaitingForAcknowledgement(final Integer msgId,
+                final CommandSubscription commandSubscription,
```
sophokles73: missing JavaDoc

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1226,4 +1300,110 @@ private static MqttConnectReturnCode getConnectReturnCode(final Throwable e) {
...
+                final MessageConsumer commandConsumer) {
+            subscriptions.put(subscription.getTopic(), TriTuple.of(subscription, commandConsumer, null));
+        }
+
+        /**
+         * Removes the subscription entry for the given topic.
+         * 
+         * @return The trituple object containing command subscription and command consumer for the given topic.
+         **/
+        protected final TriTuple<CommandSubscription, MessageConsumer, Object> removeSubscription(final String topic) {
```
sophokles73: missing JavaDoc for param

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1226,4 +1300,110 @@ private static MqttConnectReturnCode getConnectReturnCode(final Throwable e) {
...
+         * Invoked when a device sends an MQTT <em>PUBACK</em> packet.
+         *
+         * @param msgId The msgId of the command published with QoS 1.
+         */
+        protected void handlePubAck(final Integer msgId) {
+            LOG.trace("Acknowledgement received for command [Msg-id: {}] that has been sent to device.", msgId);
+            Optional.ofNullable(removeFromWaitingForAcknowledgement(msgId)).ifPresent(value -> {
+                cancelTimer(value.one());
+
+                final CommandSubscription subscription = value.two();
```
sophokles73: what if the timer has already fired? Do you still need to do all of the following in that case?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1226,4 +1300,110 @@ private static MqttConnectReturnCode getConnectReturnCode(final Throwable e) {
...
+        }
+
+        /**
+         * Removes the entry from the waitingForAcknowledgement map for the given msgId.
+         *
+         * @param msgId The id of the command (message) that has been published.
+         * @return The trituple object containing timer-id, Commandsubscription and CommandContext for the given msgId.
+         *
+         */
+        protected final TriTuple<Long, CommandSubscription, CommandContext> removeFromWaitingForAcknowledgement(
```
sophokles73: is there a need for this to have more visibility than *private*?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1226,4 +1300,110 @@ private static MqttConnectReturnCode getConnectReturnCode(final Throwable e) {
...
+         * @param msgId The msgId of the command published with QoS 1.
+         */
+        protected void handlePubAck(final Integer msgId) {
+            LOG.trace("Acknowledgement received for command [Msg-id: {}] that has been sent to device.", msgId);
+            Optional.ofNullable(removeFromWaitingForAcknowledgement(msgId)).ifPresent(value -> {
+                cancelTimer(value.one());
+
+                final CommandSubscription subscription = value.two();
+                final CommandContext commandContext = value.three();
+                afterCommandPublished(subscription, commandContext);
```
sophokles73: this class is still tightly coupled to its surrounding class by invoking this method, isn't it?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -659,6 +671,21 @@ protected final void onUnsubscribe(
                 livenessCheckInterval);
     }
 
+    private void closeCommandConsumer(final CommandSubscription commandSubscription,
```
sophokles73: I would have expected to find this inside the `CommandHandler` class ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
sophokles73: It would probably be a good idea to factor out this class into its own file so that you are no longer tempted to tightly couple them together ;-)

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: - If the timer has already fired then there will be no entry for the msgId in map and hence ``ifPresent(value ->...)`` won't be invoked.

- After entering the ``ifPresent(value ->...)`` block and if the timer expires, the timer block finds no  entry for the msgId in map. I don't see any issue.

sophokles73: Ah, I see

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttProtocolAdapterProperties.java
kaniyan: Changed.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: Added.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: Added.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: Added.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: Added.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: Added.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandHandler.java
```diff
@@ -0,0 +1,170 @@
...
+                        subscription.getDeviceId());
+            } else {
+                LOG.error("Error closing command consumer [tenant-it: {}, device-id :{}]", subscription.getTenant(),
+                        subscription.getDeviceId(), cls.cause());
+            }
+        });
+    }
+
+    private long startTimer(final Integer msgId) {
+        LOG.debug("Start a timer for [{}] ms", config.getCommandAcknowledgementTimeOut());
```
sophokles73: IMHO this should also be logged at TRACE level

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandHandler.java
```diff
@@ -0,0 +1,170 @@
...
+        return vertx.setTimer(config.getCommandAcknowledgementTimeOut(), timerId -> {
+            LOG.trace("Timer [{}] expired", timerId);
+            Optional.ofNullable(removeFromWaitingForAcknowledgement(msgId)).ifPresent(value -> {
+                final CommandSubscription subscription = value.two();
+                final CommandContext commandContext = value.three();
+                LOG.trace(
```
sophokles73: this, however, seems to be useful DEBUG information

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandHandler.java
```diff
@@ -0,0 +1,170 @@
...
+     * @param commandConsumer A client for consuming messages.
+     */
+    private void closeCommandConsumer(final CommandSubscription subscription,
+            final MessageConsumer commandConsumer) {
+        commandConsumer.close(cls -> {
+            if (cls.succeeded()) {
+                LOG.trace("Command consumer closed [tenant-it: {}, device-id :{}]", subscription.getTenant(),
+                        subscription.getDeviceId());
+            } else {
+                LOG.error("Error closing command consumer [tenant-it: {}, device-id :{}]", subscription.getTenant(),
```
sophokles73: this should not be logged at ERROR level. In fact, we don't really care about what the peer thinks about closing the link. We consider it closed in any case. So, IMHO this should be DEBUG ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandSubscription.java
```diff
@@ -159,4 +197,24 @@ public static CommandSubscription fromTopic(final String topic, final Device aut
...
+     * fields if not given.
+     *
+     * @param topic The topic to subscribe for commands.
+     * @param authenticatedDevice The authenticated device or {@code null}.
+     * @param qos The qos to subscribe for commands.     *
+     * @param clientId The the client identifier as provided by the remote MQTT client.     *
+     * @return The CommandSubscription object or {@code null} if the topic does not match the rules.
+     * @throws NullPointerException if topic is {@code null}.
+     */
+    public static CommandSubscription fromTopic(final String topic, final Device authenticatedDevice, final MqttQoS qos, final String clientId) {
```
sophokles73: why don't we simply pass in the `MqttSubscribeMessage` as an argument. It contains all of the other information and `CommandSubscription` is already dependent on vertx-mqtt.
We then only need one factory method, e.g. *fromSubscribe* instead of two, WDYT?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -408,9 +410,10 @@ private void handleConnectionRequestResult(final MqttEndpoint endpoint,
 
         endpoint.closeHandler(v -> close(endpoint, null));
         endpoint.publishHandler(message -> handlePublishedMessage(MqttContext.fromPublishPacket(message, endpoint)));
-
-        endpoint.subscribeHandler(subscribeMsg -> onSubscribe(endpoint, null, subscribeMsg));
-        endpoint.unsubscribeHandler(unsubscribeMsg -> onUnsubscribe(endpoint, null, unsubscribeMsg));
+        final CommandHandler cmdHandler = new CommandHandler(vertx, getConfig());
```
sophokles73: cmdHandler should be parameterized

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -459,24 +462,27 @@ private void handleConnectionRequestResult(final MqttEndpoint endpoint,
...
+     * @param cmdHandler The commandHandler to track command subscriptions, unsubscriptions and handle PUBACKs.
      * @throws NullPointerException if any of endpoint or subscribe message are {@code null}.
      */
     @SuppressWarnings("rawtypes")
     protected final void onSubscribe(
             final MqttEndpoint endpoint,
             final Device authenticatedDevice,
-            final MqttSubscribeMessage subscribeMsg) {
+            final MqttSubscribeMessage subscribeMsg,
+            final CommandHandler cmdHandler) {
```
sophokles73: cmdHandler should be parameterized

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -504,15 +511,15 @@ protected final void onSubscribe(
...
+                    result = createCommandConsumer(endpoint, cmdSub, cmdHandler).map(consumer -> {
                         // make sure that we close the consumer and notify downstream
                         // applications when the connection is closed
                         endpoint.closeHandler(c -> {
                             // do not use the current span for sending the disconnected event
                             // because that span will (usually) be finished long before the
                             // connection is closed
                             final Span closeHandlerSpan = newSpan("Send Disconnected Event", endpoint, authenticatedDevice);
-                            sendDisconnectedTtdEvent(cmdSub.getTenant(), cmdSub.getDeviceId(), authenticatedDevice, closeHandlerSpan.context()) 
+                            sendDisconnectedTtdEvent(cmdSub.getTenant(), cmdSub.getDeviceId(), authenticatedDevice, closeHandlerSpan.context())
```
sophokles73: when the client closes the connection we need to clean up the state in the `CommandHandler`, in particular, we need to close all consumers of all of the client's subscriptions and remove the subscriptions. I believe that any pending PUBACKs will be cleaned up by the timers, right?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
sophokles73: cmdHandler should be parameterized

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -635,7 +646,7 @@ protected final void onUnsubscribe(
         span.finish();
     }
 
-    private Future<MessageConsumer> createCommandConsumer(final MqttEndpoint mqttEndpoint, final CommandSubscription sub) {
+    private Future<MessageConsumer> createCommandConsumer(final MqttEndpoint mqttEndpoint, final CommandSubscription sub, final CommandHandler cmdHandler) {
```
sophokles73: cmdHandler should be parameterized

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
sophokles73: cmdHandler should be parameterized

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1159,14 +1170,45 @@ protected final void onCommandReceived(
...
+
+        endpoint.publish(topic, command.getPayload(), subscription.getQos(), false, false, sentHandler -> {
+            if (sentHandler.succeeded()) {
+                if (MqttQoS.AT_LEAST_ONCE.equals(subscription.getQos())) {
+                    cmdHandler.addToWaitingForAcknowledgement(sentHandler.result(), subscription, commandContext);
+                } else {
+                    afterCommandPublished(subscription, commandContext);
+                }
+            } else {
+                LOG.error(
```
sophokles73: this is an expected outcome (imagine the device suddenly disappearing due to network failure or power loss). No need to wake up the system admins ;-) simply log at DEBUG level

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandSubscription.java
kaniyan: In case of subscribe it is ` MqttSubscribeMessage` and in case of unsubscribe it is `MqttUnsubscribeMessage`. IMHO it will be still 2 methods.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: Do you mean use type parameter as below?
`CommandHandler<MqttProtocolAdapterProperties> cmdHandler = new CommandHandler(vertx, getConfig());`

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandSubscription.java
sophokles73: that's true. However, we might still pass the message itself instead of multiple params which are contained in the message. WDYT?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
sophokles73: Inside of `AbstractVertxBasedMqttProtocolAdapter` you should be able to simply re-use its `T` parameter instead of explicitly specifying `MqttProtocolAdapterProperties` ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: Inside `protected final void close(final MqttEndpoint endpoint, final Device authenticatedDevice)` the consumers can be closed and entries can be removed. Yes PUBACKs will be cleaned by the timers.

kaniyan: @sophokles73 In `onSubscribe` method the `closeHandler ` as given in the below code is being registered. I see consumer is being closed here but it is not right when there are more than one subscriptions.  IMHO this part `endpoint.closeHandler(` should be removed. I will close the consumers for all subscriptions within the `close` method and remove the subscription entries from `CommandHandler`. 
```
                        endpoint.closeHandler(c -> {
                            // do not use the current span for sending the disconnected event
                            // because that span will (usually) be finished long before the
                            // connection is closed
                            final Span closeHandlerSpan = newSpan("Send Disconnected Event", endpoint, authenticatedDevice);
                            sendDisconnectedTtdEvent(cmdSub.getTenant(), cmdSub.getDeviceId(), authenticatedDevice, closeHandlerSpan.context())
                            .setHandler(sendAttempt -> {
                                consumer.close(null);
                                close(endpoint, authenticatedDevice, cmdHandler);
                                closeHandlerSpan.finish();
                            });
                        });
```

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandSubscription.java
kaniyan: `MqttSubscribeMessage ` and `MqttUnsubscribeMessage` messages contain list of topics. So this cannot be directly passed to _fromSubscribe_ . In case of subscribe, `MqttTopicSubscription` can be sent instead of topic and qos. In case of unsubscribe, it contains list of topic strings, so in this case it remains unchanged.

sophokles73: yes, I forgot about that.

kaniyan:
@sophokles73 I have incorporated the above comments.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -395,22 +397,22 @@ private void handleConnectionRequestResult(final MqttEndpoint endpoint,
...
+     * Registers a close handler on the endpoint which invokes {@link #close(MqttEndpoint, Device, CommandHandler)}. Registers a publish
      * handler on the endpoint which invokes {@link #onPublishedMessage(MqttContext)} for each message being published
      * by the client. Accepts the connection request.
      * 
      * @param endpoint The MQTT endpoint representing the client.
      */
     private Future<Device> handleEndpointConnectionWithoutAuthentication(final MqttEndpoint endpoint) {
-
-        endpoint.closeHandler(v -> close(endpoint, null));
+        final CommandHandler<T> cmdHandler = new CommandHandler(vertx, getConfig());
```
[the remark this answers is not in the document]
kaniyan: yes, it can be refactored to use _registerHandlers_.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandHandler.java
```diff
@@ -0,0 +1,187 @@
...
+     *
+     * @param msgId The id of the command (message) that has been published.
+     * @param commandSubscription The device's command subscription.
+     * @param commandContext The commandContext of the command sent.
+     *
+     */
+    public void addToWaitingForAcknowledgement(final Integer msgId,
+            final CommandSubscription commandSubscription,
+            final CommandContext commandContext) {
+        waitingForAcknowledgement.put(msgId, TriTuple.of(startTimer(msgId), commandSubscription, commandContext));
```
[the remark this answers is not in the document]
kaniyan: Null checks have been included for public methods in CommandHandler.  If null is allowed then this info is provided in the java doc.

kaniyan:
@sophokles73 Thank you for your support and patience :-)

PR #870, opened 2018-10-30
Deprecate Kura protocol adapter

ctron:


site/content/release-notes.md
```diff
@@ -25,6 +25,13 @@ title = "Release Notes"
   context. The default implementation of the *connection events* still defaults
   to the logging producer.
 
+### Deprecations
+
+* The Kura protocol adapter is being deprecated with 0.8. It will still be part
+  Hono 0.8, but may be removed in a future version. As Kura and Hono can not be
+  used together due to modifications on both sides, it doesn't make sense
+  anymore to keep a custom Kura protocol adapter.
```
sophokles73: I am not sure if this is true. The existing Kura adapter can be used with *older* (i.e. < 4.0.0) versions of Kura, can't it? Not every Kura user will immediately upgrade to 4.0.0 so there might still be value in having the Kura Adapter around for some time. WDYT?

ctron: Well, I am not saying to remove this right now. Are people using that currently? Do we want that in Hono 1.0?

That is why I was proposing to deprecate that and remove it, maybe before 1.0.

sophokles73: > Are people using that currently? Do we want that in Hono 1.0?

I have no idea :-) I am fine with deprecating it now but I think we can postpone the discussion on when we actually want to remove it.

sophokles73: @ctron any news on this?

site/content/release-notes.md
```diff
@@ -25,6 +25,13 @@ title = "Release Notes"
   context. The default implementation of the *connection events* still defaults
   to the logging producer.
 
+### Deprecations
+
+* The Kura protocol adapter is being deprecated with 0.8. It will still be part
+  Hono 0.8, but may be removed in a future version. Both Kura and Hono have been
```
sophokles73: IMHO it would be very beneficial if you could indicate which versions this applies to, i.e. Kura >= 4 and Hono >= 0.8. WDYT?

site/content/release-notes.md
```diff
@@ -25,6 +25,13 @@ title = "Release Notes"
   context. The default implementation of the *connection events* still defaults
   to the logging producer.
 
+### Deprecations
+
+* The Kura protocol adapter is being deprecated with 0.8. It will still be part
```
sophokles73: there is an `of` missing a the end of the line

Issue #750, opened 2018-08-02
Not all service instances are "healthchecked"


ctron:
To me it looks like that only the first instance of a service is being monitored:

Looking at `org.eclipse.hono.service.AbstractApplication.deployServiceVerticles()`:
~~~java
for (final ObjectFactory<? extends AbstractServiceBase<?>> serviceFactory : serviceFactories) {

    AbstractServiceBase<?> serviceInstance = serviceFactory.getObject();

    // registered
    healthCheckServer.registerHealthCheckResources(serviceInstance);

    final Future<String> deployTracker = Future.future();
    vertx.deployVerticle(serviceInstance, deployTracker.completer());
    deploymentTracker.add(deployTracker);

    for (int i = 1; i < maxInstances; i++) { // first instance has already been deployed
        serviceInstance = serviceFactory.getObject();

        // NOT registered

        log.debug("created new instance of service: {}", serviceInstance);
        final Future<String> tracker = Future.future();
        vertx.deployVerticle(serviceInstance, tracker.completer());
        deploymentTracker.add(tracker);
    }
}
~~~

Wouldn't it make sense to use the "instances" option from the `DeploymentOptions` and let vertx handle the number of instances. And then invert the way the services find the health check server by using Spring functionality (injection).

sophokles73:
We are talking about *Verticle* instances running inside the same vert.x instance, right?

b-abel:
The general question for me is how to define the health state of the application with respect to multiple instances of a Verticle: Is the whole application DOWN if one Verticle instance is DOWN? And, if yes, how is this determined.
If different Verticle instances report a different health state, the health check would not provide a consistent result. I don't see a simple solution for this problem and would argue that it is beyond the scope of this issue. The current behaviour of having only the first Verticle instance checked is not more accurate or reliable.
So, **TL;DR** I would go with @ctron and register the health checks for every Verticle instance.

Other opinions on that?

sophokles73:
> So, TL;DR I would go with @ctron and register the health checks for every Verticle instance.

sounds good to me

b-abel:
I opened a PR, but I would like to provide another one with a different solution. The idea there is that the Verticles actively notify the `HealthCheckServer` via EventBus about health changes, the `HealthCheckServer` holds the health state of all verticle instances and can immediately answer all health check requests.
We could then compare and discuss the two PRs.

sophokles73:
[...]

PR #873, opened 2018-11-02
 [#750] Health Check for all Verticle instances.

b-abel:
A minimal implementation of ctron's proposal in #750:
1. the mechanics of the deployment of Verticles in `AbstractApplication` are simplified
2. the `HealthCheckServer` gets autowired into the Verticles, i.e. `AbstractServiceBase`

All other changes are only noise coming from HealthcheckServer being now a Bean.


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/Config.java
```diff
@@ -120,4 +121,14 @@ public ConnectionEventProducer connectionEventProducer() {
         return new LoggingConnectionEventProducer();
     }
 
+    /**
+     * Exposes the health check server as a Spring bean.
+     *
+     * @return the health check server.
+     */
+    @Bean
+    public HealthCheckServer healthCheckServer() {
```
sophokles73: how about moving this into `AbstractAdapterConfig`?

b-abel: Good point. Will do.

service-base/src/main/java/org/eclipse/hono/service/AbstractServiceBase.java
```diff
@@ -65,6 +67,17 @@ public final void setTracer(final Tracer opentracingTracer) {
         this.tracer = Objects.requireNonNull(opentracingTracer);
     }
 
+    /**
+     * Sets the HealthCheckServer to be used.
+     * 
+     * @param healthCheckServer The HealthCheckServer.
+     * @throws NullPointerException if healthCheckServer is {@code null}.
+     */
+    @Autowired
```
sophokles73: I wonder if this could/should be marked `required = false`. If no health check server is registered, the service will still work properly, wouldn't it?

service-base/src/main/java/org/eclipse/hono/service/AbstractApplication.java
```diff
@@ -104,6 +105,17 @@ public final void setApplicationConfiguration(final ApplicationConfigProperties
         this.config = Objects.requireNonNull(config);
     }
 
+    /**
+     * Sets the HealthCheckServer for this application.
+     * 
+     * @param healthCheckServer The HealthCheckServer.
+     * @throws NullPointerException if healthCheckServer is {@code null}.
+     */
+    @Autowired
```
sophokles73: I wonder if this could/should be marked required = false. If no health check server is registered, the service will still work properly, wouldn't it?

b-abel: Would be possible but require some minor adjustments to prevent NullPointers. Currently there is only one situation where you don't have health checks. That is when no port for the `HealthCheckServer` is configured. In that case it logs "No health check configured.". Making the bean not required would result two possible situations where you don't have health checks: if you don't provide the bean and if you do provide it but have no port configured. If we change this, I would suggest that we add an additional log statement that informs that no health check is provided.
I don't think one of the options is clearly better. What do you think?

sophokles73: Before digging into the details, do we agree that making the health check server optional is a good idea?

ctron: I am not so sure about making the health checks optional. If you explicitly doen't want one, then this may be a valid decision. If however, by accident, there is no health server present, but the setup assumes one to be present, then you would never know, as nothing fails.

So maybe having a noop-implementation of a health server would be a better approach.

sophokles73:
> If however, by accident, there is no health server present, but the setup assumes one to be present, then you would never know, as nothing fails.

I think that is the reason why @b-abel wants to print a log statement indicating that there is no health check server configured. In general, this shouldn't be an issue anyway, because the config of our adapters includes a health check server bean by default.

service-base/src/main/java/org/eclipse/hono/service/AbstractApplication.java
b-abel: > I am not so sure about making the health checks optional. If you explicitly doen't want one, then this may be a valid decision. If however, by accident, there is no health server present, but the setup assumes one to be present, then you would never know, as nothing fails.

I agree.

> So maybe having a noop-implementation of a health server would be a better approach.

IMHO this is what we currently have.

sophokles73:
I am not sure if I understood the arguments exchanged. Here's what I think I heard:
* A Hono component should always be required to start a health check server -> The *AbstractApplication.setHealthCheckServer* method should therefore be mandatory.
* `AbstractServiceBase` is an abstract base class which may or may not be used for implementing a Hono component that should be health checked. Keeping its *setHealthCheckServer* method optional should not do any harm because if it is used for a Hono component (which derives from `AbstractApplication` then the component will fail to start anyway if no health check server is configured, right?

service-base/src/main/java/org/eclipse/hono/service/AbstractServiceBase.java
```diff
@@ -75,6 +88,9 @@ public final void setTracer(final Tracer opentracingTracer) {
      */
     @Override
     public final void start(final Future<Void> startFuture) {
+        if (healthCheckServer != null) {
```
sophokles73: based on our discussions, we should add a log statement here for the `null` case, right?

service-base/src/main/java/org/eclipse/hono/service/AbstractApplication.java
b-abel: > So maybe having a noop-implementation of a health server would be a better approach.

@ctron Are you good with the current situation, where it is mandatory to have a  `HealthCheckServer` bean, but it can be disabled by just not setting a port for the health check (plus the log statement informing you that you don't have a health check configured)? Or do you want a noop-implementation of the `HealthCheckServer` bean?

b-abel:
Looks like I messed up the rebase. I will create a new PR instead of this one.

b-abel:
Replaced by PR  #959.

PR #875, opened 2018-11-05
Add hooks invoked during Command connection re-establishment.

sophokles73:
Added two hooks to AbstractProtocolAdapterBase:
- onCommandConnectionLost which is invoked when the connection is lost
- onCommandConnectionEstablished which is invoked when the connection is
(re-)established.

These methods can be overridden by subclasses (protocol adapters) to implement
adapter specific behavior for handling the loss and (re-)establishment of the connection
used for receiving commands.


sophokles73:
@Christian-Schmid @DanielMaier-BSI can you guys take a look at this PR and see if it helps you with implementing your protocol adapter?

DanielMaier-BSI:
Works like a charm for us. Thanks a lot.

Issue #876, opened 2018-11-05
MQTT/AMQP adapter: Deleted/disabled devices should not be able to connect


kaniyan:
A device has been registered and credentials for that device has also been added.  Then the device is deleted but not the corresponding credentials. In this case, the deleted device is still able to subscribe to control endpoint, receive commands and send command responses.

sophokles73:
If a device is deleted (I assume you are referring to the process of removing its registration information from the Device Registration service) then the device should no longer be able to connect to the MQTT adapter at all, and thus not be able to subscribe to the command topic.

I suspect that the behavior you experience is merely because the registration information (the assertion, to be precise) is cached in the protocol adapter for a configurable amount of time. We (currently) do not invalidate the cache if the underlying information is removed from the registry. Can you validate this by e.g. restarting the MQTT adapter after deleting the registration information and see if the device can still connect? You should also be able to simply wait for the cache to expire (5 minutes?) instead of restarting the adapter.

kaniyan:
> If a device is deleted (I assume you are referring to the process of removing its registration information from the Device Registration service) then the device should no longer be able to connect to the MQTT adapter at all, and thus not be able to subscribe to the command topic.
> 
> I suspect that the behavior you experience is merely because the registration information (the assertion, to be precise) is cached in the protocol adapter for a configurable amount of time. We (currently) do not invalidate the cache if the underlying information is removed from the registry. Can you validate this by e.g. restarting the MQTT adapter after deleting the registration information and see if the device can still connect? You should also be able to simply wait for the cache to expire (5 minutes?) instead of restarting the adapter.

I tried by restarting the MQTT adapter and I am still able to connect to the Mqtt adapter, subscribe to the control endpoint, receive commands and send responses.

kaniyan:
Currently for telemetry and event messages, this check is being done on receiving those messages. I agree that it should be done as part of establishing the connection.

sophokles73:
Would you like to create a PR for fixing this?

kaniyan:
sure.

kaniyan:
As per this solution, we check the device registration at the time of the connection establishment.  Assume a device, which is connected to the Mqtt adapter, is deleted. In this case, the deleted device, 
 that remains connected, can still perform telemetry, event and command operations until the Mqtt adapter is restarted or the device reconnects.  I think that this scenario (Handling deletion of connected devices) is also applicable to other adapters.

sysexcontrol:
[...]

PR #877, opened 2018-11-06
[#876] Deleted devices should not be able to connect to the Mqtt adapter

kaniyan:
When devices try to connect to the Mqtt adapter with credentials, then the Mqtt adapter checks for the device registration in addition to the credentials validation, before establishing a connection.

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
@@ -64,7 +64,6 @@ import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.json.JsonObject;
-import io.vertx.mqtt.MqttAuth;
 import io.vertx.mqtt.MqttConnectionException;
 import io.vertx.mqtt.MqttEndpoint;
 import io.vertx.mqtt.MqttServer;
@@ -408,9 +407,7 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends ProtocolAd
      */
     private Future<Device> handleEndpointConnectionWithoutAuthentication(final MqttEndpoint endpoint) {
 
-        endpoint.closeHandler(v -> {
-            close(endpoint, null);
-        });
+        endpoint.closeHandler(v -> close(endpoint, null));
         endpoint.publishHandler(message -> handlePublishedMessage(new MqttContext(message, endpoint)));
 
         endpoint.subscribeHandler(subscribeMsg -> onSubscribe(endpoint, null, subscribeMsg));
@@ -421,55 +418,27 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends ProtocolAd
         return accepted();
     }
 
-    private Future<Device> handleEndpointConnectionWithAuthentication(final MqttEndpoint endpoint, final Span currentSpan) {
+    private Future<Device> handleEndpointConnectionWithAuthentication(final MqttEndpoint endpoint,
+            final Span currentSpan) {
 
-        if (endpoint.auth() == null) {
-
-            LOG.debug("connection request from device [clientId: {}] rejected: {}",
-                    endpoint.clientIdentifier(), "device did not provide credentials in CONNECT packet");
-
-            return rejected(MqttConnectReturnCode.CONNECTION_REFUSED_BAD_USER_NAME_OR_PASSWORD);
-
-        } else {
-
-            final DeviceCredentials credentials = getCredentials(endpoint.auth());
-
-            if (credentials == null) {
-
-                LOG.debug("connection request from device [clientId: {}] rejected: {}",
-                        endpoint.clientIdentifier(), "device provided malformed credentials in CONNECT packet");
-                return rejected(MqttConnectReturnCode.CONNECTION_REFUSED_BAD_USER_NAME_OR_PASSWORD);
-
-            } else {
-
-                return getTenantConfiguration(credentials.getTenantId(), currentSpan.context()).compose(tenantConfig -> {
-                    if (tenantConfig.isAdapterEnabled(getTypeName())) {
-                        LOG.debug("protocol adapter [{}] is enabled for tenant [{}]",
-                                getTypeName(), credentials.getTenantId());
-                        return Future.succeededFuture(tenantConfig);
-                    } else {
-                        LOG.debug("protocol adapter [{}] is disabled for tenant [{}]",
-                                getTypeName(), credentials.getTenantId());
-                        return Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_FORBIDDEN,
-                                "adapter disabled for tenant"));
+        final Future<DeviceCredentials> credentialsTracker = getCredentials(endpoint);
+        return credentialsTracker
+                .compose(credentials -> authenticateCredentials(credentials))
+                .compose(device -> CompositeFuture.all(
+                        getTenantConfiguration(device.getTenantId(), currentSpan.context())
+                                .compose(tenant -> isAdapterEnabled(tenant)),
+                        isDeviceRegistered(device, credentialsTracker.result(), currentSpan))
+                        .map(result -> device))
+                .compose(device -> createLink(endpoint, device, currentSpan))
+                .recover(t -> {
+                    if (credentialsTracker.result() != null) {
+                        LOG.debug("cannot establish connection with device [tenant-id: {}, auth-id: {}]",
+                                credentialsTracker.result().getTenantId(), credentialsTracker.result().getAuthId(), t);
                     }
-                }).compose(tenantConfig -> {
-                    final Future<DeviceUser> result = Future.future();
-                    usernamePasswordAuthProvider.authenticate(credentials, result.completer());
-                    return result;
-                }).compose(authenticatedDevice -> {
-                    currentSpan.log(String.format("device authenticated"));
-                    LOG.debug("successfully authenticated device [tenant-id: {}, auth-id: {}, device-id: {}]",
-                            authenticatedDevice.getTenantId(), credentials.getAuthId(),
-                            authenticatedDevice.getDeviceId());
-                    return triggerLinkCreation(authenticatedDevice.getTenantId()).map(done -> {
-                        currentSpan.log(String.format("opened downstream links"));
-                        onAuthenticationSuccess(endpoint, authenticatedDevice);
-                        return null;
-                    }).compose(ok -> accepted(authenticatedDevice));
-                }).recover(t -> {
-                    LOG.debug("cannot establish connection with device [tenant-id: {}, auth-id: {}]",
-                            credentials.getTenantId(), credentials.getAuthId(), t);
+                    if (t instanceof MqttConnectionException) {
+                        return rejected(((MqttConnectionException) t).code());
+                    }
+
                     if (t instanceof ServerErrorException) {
                         // one of the services we depend on might not be available (yet)
                         return rejected(MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE);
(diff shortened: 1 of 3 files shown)


adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1233,4 +1193,66 @@ private static void addRetainAnnotation(final MqttContext context, final Message
...
+        return getRegistrationAssertion(authenticatedDevice.getTenantId(),
+                authenticatedDevice.getDeviceId(),
+                authenticatedDevice, currentSpan.context())
+                        .compose(ok -> accepted(authenticatedDevice));
+    }
+
+    private Future<Device> createLink(final MqttEndpoint endpoint, final Device authenticatedDevice,
+            final Span currentSpan) {
+        currentSpan.log(String.format("device registration verified"));
+        LOG.debug("verified device registration [tenant-id: {}, device-id: {}]",
```
sophokles73: this log statement assumes that this method has been invoked *after* the device's registration has been verified. However, there is nothing that prevents this method from being invoked in a completely different context. IMHO not *every* few statements need to be refactored into a separate method ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -438,64 +443,30 @@ private void handleConnectionRequestResult(final MqttEndpoint endpoint,
...
-                    } else {
-                        // validation of credentials has failed
-                        return rejected(MqttConnectReturnCode.CONNECTION_REFUSED_NOT_AUTHORIZED);
-                    }
-                });
-
+                return getTenantConfiguration(credentials.getTenantId(), currentSpan.context())
+                        .compose(tenantConfig -> isAdapterEnabled(credentials, tenantConfig))
+                        .compose(tenantConfig -> authenticateCredentials(credentials))
+                        .compose(authenticatedDevice -> isDeviceRegistered(authenticatedDevice, credentials,
```
sophokles73: running all of these checks sequentially will increase latency (a lot). IMHO we should retrieve all relevant data and make the checks in parallel, failing the (composite) future as soon as one of the checks fails ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1233,4 +1193,66 @@ private static void addRetainAnnotation(final MqttContext context, final Message
             MessageHelper.addAnnotation(downstreamMessage, MessageHelper.ANNOTATION_X_OPT_RETAIN, Boolean.TRUE);
         }
     }
+
+    private Future<TenantObject> isAdapterEnabled(final DeviceCredentials credentials, final TenantObject tenantConfig) {
```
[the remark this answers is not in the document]
kaniyan: There is an already existing debug statement that prints authId and only for that sake, credentials parameter is needed

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -73,6 +73,11 @@
 import io.vertx.mqtt.messages.MqttSubscribeMessage;
 import io.vertx.mqtt.messages.MqttUnsubscribeMessage;
 
+import static io.netty.handler.codec.mqtt.MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE;
```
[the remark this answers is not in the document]
kaniyan: I will change them

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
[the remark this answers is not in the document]
sophokles73: hm, I do not seem to be able to find the location, can you help me out?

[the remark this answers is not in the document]
kaniyan: I referred to `isDeviceRegistered` instead of `isAdapterEnabled`. As you suggested, the ID of the tenantConfig object can be used instead from the credentials parameter.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: FMPOV `isAdapterEnabled()` and `authenticateCredentials()` can be checked in parallel and the code snippet is as below.  And the rest have dependency on the result of another future and I don't see many options for composite future. Technically it is possible to execute `createLink()` in parallel with the rest. FMPOV why should we even create links until all the checks are successful.
```
final Future<DeviceUser> authenticatedDeviceTracker = authenticateCredentials(credentials);
                
getTenantConfiguration(credentials.getTenantId(), currentSpan.context())
     .compose(tenantObject -> CompositeFuture.all(isAdapterEnabled(tenantObject), authenticatedDeviceTracker))
     .compose(x->isDeviceRegistered(authenticatedDeviceTracker.result(), credentials, currentSpan))
     .compose(authenticatedDevice->createLink(endpoint, authenticatedDevice, currentSpan))
```

sophokles73: The point is that the *retrieval* of the data is very expensive (i.e. slow) compared to the logical checks we need to perform. So, instead of doing it sequentially, we can at least *retrieve* the data in parallel
* tenant
* credentials
* device registration
If any of those retrieval operations fails, then the device cannot connect. Otherwise, we need to do the logical checks (which are cheap to perform), i.e.
* is tenant enabled?
* are the credentials correct?
* is the device enabled?

Sometimes we might not actually validate the retrieved registration information anymore because the tenant has already been determined to be disabled, but that is no issue FMPOV.

kaniyan: I fully agree with it with the benefits of _retrieval_ of data  in parallel.  IMHO, at least the credentials should be retrieved and verified first before retrieving the rest (tenant or device registration). If we retrieve tenants, devices etc even before verifying credentials, then in case of _invalid_ credentials or any _attack_, it adds unnecessary overload to the system or prone to attack.

sophokles73: sounds like a plan

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: I consolidated into one method `createLink`.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1226,11 +1185,75 @@ protected final void onCommandReceived(
...
+        usernamePasswordAuthProvider.authenticate(credentials, result.completer());
+        return result;
+    }
+
+    private Future<Device> isDeviceRegistered(final Device authenticatedDevice, final DeviceCredentials credentials,
+            final Span currentSpan) {
+        currentSpan.log(String.format("device authenticated"));
+        LOG.debug("successfully authenticated device [tenant-id: {}, auth-id: {}, device-id: {}]",
+                authenticatedDevice.getTenantId(), credentials.getAuthId(),
+                authenticatedDevice.getDeviceId());
```
sophokles73: the five lines above do not belong into this method IMHO. They should be moved to the *authenticateCredentials* method, don't you think?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1226,11 +1185,75 @@ protected final void onCommandReceived(
...
+                        });
+    }
+
+    private Future<Device> createLink(final MqttEndpoint endpoint, final Device authenticatedDevice,
+            final Span currentSpan) {
+        final Future<MessageSender> telemetrySender = getTelemetrySender(authenticatedDevice.getTenantId());
+        final Future<MessageSender> eventSender = getEventSender(authenticatedDevice.getTenantId());
+
+        return CompositeFuture
+                .all(getRegistrationClient(authenticatedDevice.getTenantId()), telemetrySender, eventSender)
```
sophokles73: I believe that at this point we already have created a registration client because we have already verified that the device exists and is enabled, right? It shouldn't have too much of an impact because the client will be retrieved from the cache. However, for clarity I think you should remove it from here and just create the senders.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1226,11 +1185,75 @@ protected final void onCommandReceived(
...
+                .map(ok -> {
+                    currentSpan.log(String.format("opened downstream links"));
+                    LOG.debug(
+                            "providently opened downstream links [credit telemetry: {}, credit event: {}] for tenant [{}]",
+                            telemetrySender.result().getCredit(), eventSender.result().getCredit(),
+                            authenticatedDevice.getTenantId());
+                    onAuthenticationSuccess(endpoint, authenticatedDevice);
```
sophokles73: this should be invoked from within *handleEndpointConnectionWithAuthentication* as it is the last thing to do after the device has been authenticated etc.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: These debug statements already existed and I understood that the intention was to execute them once authentication succeeds. You are right and IMHO it doesn't belong here or `authenticateCredentials` method. I will change so that these statements are executed after successful authentication.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: Handlers are being registered inside this method. Hence I have refactored this method name to `registerHandlers` and it will be invoked within `handleEndpointConnectionWithAuthentication`

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: Done

PR #889, opened 2018-11-08
Drop the vert.x metrics options, and allow disabling vertx metrics.

ctron:
This disables the external vertx metrics configurations, as Spring Boot is the main driver of this now. Vertx itself can be enabled, using the vertx options of Hono. However it is disabled by default, as some of the vertx metrics can really blow the metrics backends.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/config/VertxProperties.java
+++ b/core/src/main/java/org/eclipse/hono/config/VertxProperties.java
@@ -14,6 +14,7 @@
 package org.eclipse.hono.config;
 
 import io.vertx.core.VertxOptions;
+import io.vertx.core.metrics.MetricsOptions;
 
 /**
  * Vertx properties.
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/eclipse/hono/config/VertxProperties.java
```diff
@@ -34,6 +35,22 @@ public void setPreferNative(final boolean preferNative) {
...
 
+    private boolean enableMetrics = MetricsOptions.DEFAULT_METRICS_ENABLED;
+
+    /**
+     * Enable the vert.x metrics system, or not.
+     * 
+     * <p>
+     * This decides if the vert.x metrics system will be enabled. Hono uses the Spring Boot integration with Micrometer,
+     * so enabling the vert.x metrics will only enable the contribution of vert.x internal metrics to this system.
+     * </p>
```
sophokles73: IMHO you should also indicate the default value of this property.

PR #891, opened 2018-11-08
Do not return cached command consumers

### LINKED ISSUE #754: Multiple ttd based downstream messages may make command sending impossible
In our current implementation we seem to have a problem with following scenario:
- a device sends a downstream message with ttd = 10 seconds
- the same device sends another downstream message (e.g. one second later) with ttd = 60 seconds
- application tries to send a command after 30 seconds (it has a good reason to do so, since it got a notification about the 60 second timeframe)

In this case:
- the application gets an error when trying to send the command after 30 seconds (no credits, since the link to the device was closed after 10 seconds)
- the device never gets a command, although the second request indicated that it would be ready for a command

If things go really bad, this could make the sending of commands always impossible, if the link is closed already due to timer expiries of other messages with a ttd.

Solution could be to not close the link if there is at least one timer related to this link that expires later than the current timer.

sophokles73:
The CommandConnection.getOrCreatedCommandConsumer method has been renamed to
createCommandConsumer and will no longer return an already existing
instance of a command consumer for a given device but will instead fail
the returned future with an
org.eclipse.hono.client.ResourceConflictException to indicate that a
consumer for the given device is already in use.

The original behavior allowed an implementation to return a consumer
that was scoped to another message handler than the one passed into the
method as an argument. However, client code had no chance to determine
whether it got back a newly created instance or an existing one. This
has been resolved with the new method semantics.

There is now also an integration test for the HTTP adapter which
verifies the correct behavior when a device issues multiple requests
with a TTD without waiting for a response before sending the next
request. The HTTP adapter will in this case ignore the TTD in the subsequent
requests and simply return a 202.

This should also implicitly fix #754

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/CommandConnection.java
+++ b/client/src/main/java/org/eclipse/hono/client/CommandConnection.java
@@ -27,22 +27,35 @@ import org.eclipse.hono.config.ClientConfigProperties;
 public interface CommandConnection extends HonoClient {
 
     /**
-     * Gets a command consumer for a device.
+     * Creates a command consumer for a device.
      * <p>
-     * Implementations may choose to cache consumers for devices and
-     * return a cached instance instead of creating a new consumer on
-     * each invocation.
+     * For each device only one command consumer may be active at any given time.
+     * It is the responsibility of the calling code to properly close a consumer
+     * once it is no longer needed. The preferred way of doing so is to invoke the
+     * instance's {@link CommandConsumer#close(Handler)} method. Alternatively, if
+     * no reference to the instance is held, the {@link #closeCommandConsumer(String, String)}
+     * method can be used instead.
      * 
      * @param tenantId The tenant to consume commands from.
      * @param deviceId The device for which the consumer will be created.
      * @param commandHandler The handler to invoke with every command received.
      * @param remoteCloseHandler A handler to be invoked after the link has been closed
      *                     at the peer's request.
-     * @return A future that will complete with the consumer once the link has been established. The future will fail if
-     *         the link cannot be established, e.g. because this client is not connected.
-     * @throws NullPointerException if tenantId, deviceId or messageConsumer is {@code null}.
+     * @return A future indicating the outcome of the operation.
+     *         <p>
+     *         The future will be completed with the newly created consumer once the link
+     *         has been established.
+     *         <p>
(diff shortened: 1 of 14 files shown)


client/src/main/java/org/eclipse/hono/client/impl/CommandConnectionImpl.java
```diff
@@ -95,12 +96,16 @@ protected void clearState() {
...
-            return Future.succeededFuture(messageConsumer);
+            log.debug("cannot create concurrent command consumer [tenant: {}, device-id: {}]", tenantId, deviceId);
+            return Future.failedFuture(new ResourceConflictException("message consumer already in use"));
         } else {
             return createConsumer(
                     tenantId,
-                    () -> newCommandConsumer(tenantId, deviceId, commandConsumer, closeHandler));
+                    () -> newCommandConsumer(tenantId, deviceId, commandConsumer, closeHandler))
+               .map(consumer -> {
+                    commandReceivers.put(Device.asAddress(tenantId, deviceId), consumer);
```
ctron: Is that in any way synchronized? What would happen if two threads would call into this method at the same time?

sophokles73:
@sysexcontrol @ctron 
I have made some changes on master which allow the proper usage of `CommandConsumer.close()` to close a consumer instead of CommandConnection.closeCommandConsumer()`. Based on these 
I have pushed some updates which make sure that the access to the consumer cache is synchronized (i.e. run on the correct event loop context).
Can you take another look?

client/src/main/java/org/eclipse/hono/client/CommandConnection.java
```diff
@@ -27,22 +27,32 @@
...
     /**
-     * Gets a command consumer for a device.
+     * Creates a command consumer for a device.
      * <p>
-     * Implementations may choose to cache consumers for devices and
-     * return a cached instance instead of creating a new consumer on
-     * each invocation.
+     * For each device only one command consumer may be active at any given time.
+     * The {@link #closeCommandConsumer(String, String)} method can be used to close
+     * an active consumer.
```
ctron: I think the wording should be a bit stronger, more in the direction "it is the responsibility of the caller to close the instance…"

sophokles73:
@ctron WDYT?

PR #892, opened 2018-11-08
[#669] Provide one-way pattern for commands.

sysexcontrol:
This is step 1 of a sequence of PRs for implementing #669 : it provides the base for sending one-way commands by using the slightly extended AbstractRequestResponseClient.
The commandClient can decide for each command if a response from the device is expected or not by choosing different send methods:
`sendCommand` or `sendOneWayCommand`.

It was tested with the HTTP adapter and an extended example app - both are not part of this PR though and follow in a subsequent PR.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/Command.java
+++ b/client/src/main/java/org/eclipse/hono/client/Command.java
@@ -111,9 +111,7 @@ public final class Command {
 
         String replyToId = null;
 
-        if (message.getReplyTo() == null) {
-            valid = false;
-        } else {
+        if (message.getReplyTo() != null) {
             try {
                 final ResourceIdentifier replyTo = ResourceIdentifier.fromString(message.getReplyTo());
                 if (!CommandConstants.COMMAND_ENDPOINT.equals(replyTo.getEndpoint())) {
@@ -152,6 +150,15 @@ public final class Command {
      */
     public Message getCommandMessage() {
         return message;
+    }
+
+    /**
+     * Checks if this command is a <em>one-way</em> command (meaning there is no response expected).
+     *
+     * @return {@code true} if this is a valid command.
+     */
+    public boolean isOneWay() {
+        return replyToId == null;
     }
 
     /**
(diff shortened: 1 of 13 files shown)


sysexcontrol:
> It looks like you are trying to use the request/response mechanism for sending a one-way message. I do not quite get, why you want to do that.

The point to do it was to have the same code running that sends the request before it waits for a response, and then only change the part of waiting for the response.
Since the CommandClient inherits from AbstractRequestResponseClient this was IMHO the obvious way to go.



> The overall intention of using the disposition frames for signalling the outcome of sending a command (request) to a device, was that the protocol adapter does not need to send back a response on behalf of the device when we want to send a one-way command.

Which it doesn't have to do with this implementation, either. It only needs to set the disposition to `accepted`.

>FMPOV we should simply create an AMQP message containing the one-way command and send it using the ProtonSender available in the CommandClientImpl. Then simply check for the disposition in order to complete or fail the returned Future. 
This is another approach of doing it, which IMHO has pros and cons:
- it would remove the changes to the request response code, since a missing reply does not have to be handled
- but it needs the same handling for the send part as already implemented. So at least a refactoring of the send code should be done first, to not duplicate the code.

Since I kind of understand your idea of doing it, my proposal is: let me provide an implementation with directly using the sender (as you proposed) in a different PR and then we can decide which way to go. Hopefully this should be not difficult, and if it works out, it could be cleaner regarding the application property handling that is coded in this PR.

> The only change we need to make in the protocol adapters would be to handle a potentially missing reply-to property in the command message, right?

In both implementations: right, that is the only thing to do for the protocol adapter.

sysexcontrol:
@sophokles73 I provided the alternative implementation (which looks much better IMHO, so I think your proposal was a good one :+1: ) in the PR #893 .
If we agree to use this direction, this PR will be closed again (until then I leave it open).

sysexcontrol:
We do not need this PR anymore, since we favor the clearer implementation in PR #893 .

PR #893, opened 2018-11-09
[#669] Alternative implementation for one-way pattern for commands.

sysexcontrol:
This is step 1 of a sequence of PRs for implementing #669 : it provides the base for sending one-way commands in a different implementation (not using `createAndSend` methods anymore, but still sharing the `sendRequest` method with the AbstractRequestResponseClient).
The commandClient can decide for each command if a response from the device is expected or not by choosing different send methods:
`sendCommand` or `sendOneWayCommand`.

It was tested with the HTTP adapter and an extended example app - both are not part of this PR though and follow in a subsequent PR.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/Command.java
+++ b/client/src/main/java/org/eclipse/hono/client/Command.java
@@ -107,9 +107,7 @@ public final class Command {
 
         String replyToId = null;
 
-        if (message.getReplyTo() == null) {
-            valid = false;
-        } else {
+        if (message.getReplyTo() != null) {
             try {
                 final ResourceIdentifier replyTo = ResourceIdentifier.fromString(message.getReplyTo());
                 if (!CommandConstants.COMMAND_ENDPOINT.equals(replyTo.getEndpoint())) {
@@ -148,6 +146,15 @@ public final class Command {
      */
     public Message getCommandMessage() {
         return message;
+    }
+
+    /**
+     * Checks if this command is a <em>one-way</em> command (meaning there is no response expected).
+     *
+     * @return {@code true} if this is a valid command.
+     */
+    public boolean isOneWay() {
+        return replyToId == null;
     }
 
     /**
@@ -293,11 +300,11 @@ public final class Command {
      * @param correlationId The identifier to use for correlating the response with the request.
      * @param replyToId An arbitrary identifier to encode into the request ID.
      * @param deviceId The target of the command.
-     * @return The request identifier or {@code null} if any of the parameters are {@code null}.
+     * @return The request identifier or {@code null} if any the correlationId or the deviceId is {@code null}.
      */
     public static String getRequestId(final String correlationId, final String replyToId, final String deviceId) {
 
-        if (correlationId == null || replyToId == null || deviceId == null) {
+        if (correlationId == null || deviceId == null) {
             return null;
         }
 
--- a/client/src/main/java/org/eclipse/hono/client/CommandClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/CommandClient.java
@@ -68,4 +68,42 @@ public interface CommandClient extends RequestResponseClient {
      * @see RequestResponseClient#setRequestTimeout(long)
      */
     Future<BufferResult> sendCommand(String command, String contentType, Buffer data, Map<String, Object> properties);
+
+    /**
+     * Sends a <em>one-way command</em> to a device, i.e. there is no response expected from the device.
+     * <p>
+     * A device needs to be (successfully) registered before a client can upload
+     * any data for it. The device also needs to be connected for a successful delivery.
+     *
+     * @param command The one-way command name.
+     * @param data The command data to send to the device or {@code null} if the one-way command has no input data.
+     * @return A future indicating the result of the operation.
+     *         <p>
+     *         If the one-way command was accepted, the future will succeed.
+     *         <p>
+     *         The future will fail with a {@link ServiceInvocationException} if the one-way command could not be forwarded to the device.
+     * @throws NullPointerException if command is {@code null}.
+     * @see RequestResponseClient#setRequestTimeout(long)
+     */
+    Future<Void> sendOneWayCommand(String command, Buffer data);
+
+    /**
+     * Sends a <em>one-way command</em> to a device, i.e. there is no response from the device expected.
+     * <p>
+     * A device needs to be (successfully) registered before a client can upload
+     * any data for it. The device also needs to be connected for a successful delivery.
+     *
+     * @param command The one-way command name.
+     * @param contentType The type of the data submitted as part of the one-way command or {@code null} if unknown.
+     * @param data The command data to send to the device or {@code null} if the command has no input data.
+     * @param properties The headers to include in the one-way command message as AMQP application properties.
+     * @return A future indicating the result of the operation:
+     *         <p>
+     *         If the one-way command was accepted, the future will succeed.
+     *         <p>
+     *         The future will fail with a {@link ServiceInvocationException} if the one-way command could not be forwarded to the device.
+     * @throws NullPointerException if command is {@code null}.
+     * @see RequestResponseClient#setRequestTimeout(long)
+     */
+    Future<Void> sendOneWayCommand(String command, String contentType, Buffer data, Map<String, Object> properties);
 }
--- a/client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
@@ -719,7 +719,7 @@ public abstract class AbstractRequestResponseClient<R extends RequestResponseRes
      * @param cacheKey The key to use for caching the response (if the service allows caching).
      * @param currentSpan The <em>Opentracing</em> span used to trace the request execution.
      */
-    private void sendRequest(
+    protected final void sendRequest(
             final Message request,
             final Handler<AsyncResult<R>> resultHandler,
             final Object cacheKey,
@@ -770,6 +770,12 @@ public abstract class AbstractRequestResponseClient<R extends RequestResponseRes
                         LOG.trace("service has accepted request [target address: {}, subject: {}, correlation ID: {}]",
                                 targetAddress, request.getSubject(), correlationId);
                         currentSpan.log("request accepted by peer");
+                        // if no reply-to is set, the request is assumed to be one-way (no response is expected)
+                        if (request.getReplyTo() == null) {
+                            Tags.HTTP_STATUS.set(currentSpan, HttpURLConnection.HTTP_ACCEPTED);
+                            replyMap.remove(correlationId);
+                            resultHandler.handle(Future.succeededFuture());
+                        }
                     } else {
                         LOG.debug("service did not accept request [target address: {}, subject: {}, correlation ID: {}]: {}",
                                 targetAddress, request.getSubject(), correlationId, remoteState);
--- a/client/src/main/java/org/eclipse/hono/client/impl/CommandClientImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/CommandClientImpl.java
@@ -13,15 +13,22 @@
 
 package org.eclipse.hono.client.impl;
 
+import java.net.HttpURLConnection;
 import java.util.Map;
 import java.util.Objects;
 
+import io.opentracing.Span;
+import io.vertx.proton.ProtonHelper;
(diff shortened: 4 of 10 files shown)


client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -154,6 +152,15 @@ public Message getCommandMessage() {
         return message;
     }
 
+    /**
+     * Checks if this command is a <em>one-way</em> command (meaning there is no response expected).
+     *
+     * @return {@code true} if this is a valid command.
+     */
+    public boolean isOneWay() {
```
sophokles73: do we need this?

core/src/main/java/org/eclipse/hono/util/Constants.java
```diff
@@ -216,6 +216,11 @@ public Authorities getAuthorities() {
      */
     public static final String HEADER_COMMAND = "hono-command";
 
+    /**
+     * The header name defined for setting the <em>one-way command</em> that is sent to the device.
+     */
+    public static final String HEADER_ONE_WAY_COMMAND = "hono-one-way-command";
```
sophokles73: do we need this?

client/src/test/java/org/eclipse/hono/client/impl/CommandClientImplTest.java
```diff
@@ -56,7 +57,7 @@
      * Time out test cases after 3 seconds.
      */
     @Rule
-    public Timeout globalTimeout = Timeout.seconds(3);
+    public Timeout globalTimeout = Timeout.seconds(300);
```
sophokles73: this doesn't look right

client/src/main/java/org/eclipse/hono/client/impl/CommandClientImpl.java
```diff
@@ -152,6 +161,49 @@ protected BufferResult getResult(final int status, final String contentType, fin
...
+
+        final Future<BufferResult> responseTracker = Future.future();
+        final Span currentSpan = newChildSpan(null, command);
+
+        Optional.ofNullable(sender).ifPresent(s -> {
+            if (s.isOpen()) {
+                final Message request = ProtonHelper.message();
+
+                // set a unique correlationId for one-way requests, since there is no unique reply-id
+                request.setCorrelationId(UUID.randomUUID().toString());
```
sophokles73: why do we need to set a correlation ID? There is no response expected so there will be nothing to correlate, right?

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -770,6 +770,12 @@ private void sendRequest(
                         LOG.trace("service has accepted request [target address: {}, subject: {}, correlation ID: {}]",
                                 targetAddress, request.getSubject(), correlationId);
                         currentSpan.log("request accepted by peer");
+                        // if no reply-to is set, the request is assumed to be one-way (no response is expected)
+                        if (request.getReplyTo() == null) {
+                            Tags.HTTP_STATUS.set(currentSpan, HttpURLConnection.HTTP_ACCEPTED);
+                            replyMap.remove(correlationId);
```
sophokles73: this is a clever trick to be able to use all of the existing time-out behavior etc :+1:

client/src/main/java/org/eclipse/hono/client/Command.java
sysexcontrol: We will need it in the adapter implementation to decide if a command was one way (for convenience).
So, it is not strictly necessary in this commit, but in the directly following step 2 then, so I decided to provide it here already.

core/src/main/java/org/eclipse/hono/util/Constants.java
sysexcontrol: Good point: this is only needed in the HTTP adapter implementation, so I will remove it from this PR (and provide it again in the next PR).

client/src/test/java/org/eclipse/hono/client/impl/CommandClientImplTest.java
sysexcontrol: No, for sure not. Will revert this.

client/src/main/java/org/eclipse/hono/client/impl/CommandClientImpl.java
sysexcontrol: No response, this is right - but to have a correlationId on the receiver (i.e. the device) should be reasonable, right?
And the here set correlationId would be the siginificant part of the `commandId`, that is sent to the device.

One could argue, though, that a one-way command does not need a commandId at all (to keep the footprint as low as possible).
Was that your concern?

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
sysexcontrol: Thx, that was the idea of it :-)

client/src/main/java/org/eclipse/hono/client/impl/CommandClientImpl.java
sophokles73: > One could argue, though, that a one-way command does not need a commandId at all (to keep the footprint as low as possible). Was that your concern?

yes, I do not see what it would be good for.

core/src/main/java/org/eclipse/hono/util/Constants.java
sophokles73: What does the HTTP adapter need this for?

client/src/main/java/org/eclipse/hono/client/Command.java
sysexcontrol: Right.

core/src/main/java/org/eclipse/hono/util/Constants.java
sysexcontrol: If we skip the `command-request-id` for one-way commands, we do not need it (and send one-way commands only by omitting the `command-request-id`).
So, I will remove it completely.

client/src/main/java/org/eclipse/hono/client/impl/CommandClientImpl.java
sysexcontrol: Ok, we then handle one-way commands exactly the same as request-response commands - and only omit the `command-request-id`.

I will see what this will mean for MQTT (regarding the topic structure), but it seems be one valid way to go, which additionally has some advantages (low footprint).

core/src/main/java/org/eclipse/hono/util/Constants.java
sophokles73: sounds good to me

client/src/main/java/org/eclipse/hono/client/impl/CommandClientImpl.java
sophokles73: Maybe the MQTT adapter can simply use an empty string in that case, e.g. publishing to a topic like
`control///req//setBrightness`
analogous to the way we omit the tenant and device id ...

sysexcontrol: Yes, this was my first though as well.
Looks good, probably the way to implement it for MQTT.

sysexcontrol:
@sophokles73 : pushed the changes from the review.

client/src/main/java/org/eclipse/hono/client/impl/CommandClientImpl.java
```diff
@@ -152,6 +160,44 @@ protected BufferResult getResult(final int status, final String contentType, fin
...
+     * {@inheritDoc}
+     */
+    @Override
+    public Future<Void> sendOneWayCommand(final String command, final String contentType, final Buffer data, final Map<String, Object> properties) {
+        Objects.requireNonNull(command);
+
+        final Future<BufferResult> responseTracker = Future.future();
+        final Span currentSpan = newChildSpan(null, command);
+
+        Optional.ofNullable(sender).ifPresent(s -> {
```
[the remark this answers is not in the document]
sysexcontrol: That is true, thanks for pointing out :+1: 
I will update the method accordingly.

client/src/main/java/org/eclipse/hono/client/CommandClient.java
```diff
@@ -68,4 +68,44 @@
...
+     * A device needs to be (successfully) registered before a client can upload
+     * any data for it. The device also needs to be connected for a successful delivery.
+     *
+     * @param command The one-way command name.
+     * @param data The command data to send to the device or {@code null} if the one-way command has no input data.
+     * @return A future indicating the result of the operation.
+     *         <p>
+     *         If the one-way command was accepted, the future will succeed.
+     *         <p>
+     *         If the one-way command could not be processed by the protocol adapter, the future will fail with a {@link ServiceInvocationException} containing
```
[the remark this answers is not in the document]
sysexcontrol: :+1: will update the description.

client/src/main/java/org/eclipse/hono/client/CommandClient.java
```diff
@@ -68,4 +68,44 @@
...
+     *
+     * @param command The one-way command name.
+     * @param contentType The type of the data submitted as part of the one-way command or {@code null} if unknown.
+     * @param data The command data to send to the device or {@code null} if the command has no input data.
+     * @param properties The headers to include in the one-way command message as AMQP application properties.
+     * @return A future indicating the result of the operation:
+     *         <p>
+     *         If the one-way command was accepted, the future will succeed.
+     *         <p>
+     *         If the one-way command could not be processed by the protocol adapter, the future will fail with a {@link ServiceInvocationException} containing
```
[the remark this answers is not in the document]
sysexcontrol: Oh, of course!

client/src/main/java/org/eclipse/hono/client/CommandClient.java
```diff
@@ -68,4 +68,44 @@
...
+     *         the (error) status code. Status codes are defined at <a href="https://www.eclipse.org/hono/api/command-and-control-api">Command and Control API</a>.
+     * @throws NullPointerException if oneWayCommand is {@code null}.
```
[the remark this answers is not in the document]
sysexcontrol: Oh, of course!

sysexcontrol:
@sophokles73 : I addressed the proposed changes in the last commit - please have a another look, thanks!

client/src/main/java/org/eclipse/hono/client/CommandClient.java
```diff
@@ -81,8 +81,8 @@
      *         <p>
      *         If the one-way command was accepted, the future will succeed.
      *         <p>
-     *         If the one-way command could not be processed by the protocol adapter, the future will fail with a {@link ServiceInvocationException} containing
-     *         the (error) status code. Status codes are defined at <a href="https://www.eclipse.org/hono/api/command-and-control-api">Command and Control API</a>.
+     *         The future will fail with a {@link ServiceInvocationException} if the one-way command could not be forwarded to the device.
+     *         The contained status code is defined at <a href="https://www.eclipse.org/hono/api/command-and-control-api">Command and Control API</a>.
```
sophokles73: it is not, there is no such definition in the API for one-way commands because there is no response containing a status code. If there are any specific error codes then they will need to be defined here.

client/src/main/java/org/eclipse/hono/client/CommandClient.java
```diff
@@ -102,9 +102,9 @@
      *         <p>
      *         If the one-way command was accepted, the future will succeed.
      *         <p>
-     *         If the one-way command could not be processed by the protocol adapter, the future will fail with a {@link ServiceInvocationException} containing
-     *         the (error) status code. Status codes are defined at <a href="https://www.eclipse.org/hono/api/command-and-control-api">Command and Control API</a>.
-     * @throws NullPointerException if oneWayCommand is {@code null}.
+     *         The future will fail with a {@link ServiceInvocationException} if the one-way command could not be forwarded to the device.
+     *         The contained status code is defined at <a href="https://www.eclipse.org/hono/api/command-and-control-api">Command and Control API</a>.
```
sophokles73: see above

client/src/main/java/org/eclipse/hono/client/CommandClient.java
sysexcontrol: Yes, you are right - the only error code is `503` (which is not generated by a response), so we do not need to specify it here IMHO.
Will remove this reference.

sysexcontrol:
@sophokles73 : pushed a new commit reflecting last comments.
If you approve it, I will merge it and continue with the next step.

PR #897, opened 2018-11-15
Update Jaeger version to 0.32.0

calohmn:


calohmn:
Not to be merged yet!
Just saw there are some further changed needed.

sophokles73:
We cannot simply update dependencies without getting them through the Eclipse IP process first. So, regardless of whether there are other changes pending, this will take some time. I will schedule it to 0.9 ...

calohmn:
I've amended the commit - an update of opentracing-tracerresolver was needed for the newer jaeger version.

sophokles73:
created [CQ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=18372) for opentracing-tracerresolver 0.1.5

PR #900, opened 2018-11-15
Fix issues and improve MQTT Adapter integration tests.

kaniyan:


tests/src/test/java/org/eclipse/hono/tests/mqtt/MqttConnectionIT.java
```diff
@@ -107,8 +111,8 @@ public void testConnectFailsForDisabledAdapter(final TestContext ctx) {
         final String password = "secret";
         final TenantObject tenant = TenantObject.from(tenantId, true);
         final JsonObject adapterDetailsMqtt = new JsonObject()
-                .put(TenantConstants.FIELD_ADAPTERS_TYPE, Constants.PROTOCOL_ADAPTER_TYPE_HTTP)
-                .put(TenantConstants.FIELD_ENABLED, Boolean.TRUE);
+                .put(TenantConstants.FIELD_ADAPTERS_TYPE, Constants.PROTOCOL_ADAPTER_TYPE_MQTT)
+                .put(TenantConstants.FIELD_ENABLED, Boolean.FALSE);
```
sophokles73: nice catch :+1:

tests/src/test/java/org/eclipse/hono/tests/mqtt/MqttConnectionIT.java
```diff
@@ -143,4 +147,41 @@ public void testConnectFailsForDeletedDevices(final TestContext ctx) {
...
+
+
+    /**
+     * Verifies that the adapter rejects connection attempts from devices
+     * that belong to a tenant for which the adapter has been disabled.
+     *
+     * @param ctx The test context
+     */
+    @Test
+    public void testConnectFailsForDisabledTenant(final TestContext ctx) {
```
sophokles73: this seems to be a duplicate of *testConnectFailsForDisabledAdapter*, isn't it?
However, it might make sense to have a test that actually verifies that a device belonging to a disabled *tenant* cannot connect, i.e. 
```
final TenantObject tenant = TenantObject.from(tenantId, false);
```

kaniyan: Its a good catch. That was an existing test case and I just moved to MqttConnectionIT class. 
I will enable the adapter and disable the tenant.

Issue #830, opened 2018-09-27
Command receiver links do not reopen after reconnection


sysexcontrol:
Protocol adapters do not automatically reopen command receiver links if the connection to the AMQP 1.0 network fails.
For long living command receiver links that can occur for:
- MQTT connected devices
- HTTP connected devices with a high ttd value
- tenant scoped command receiver links for custom protocol adapters (soon being available)

this can be a major problem since the commands cannot be sent from applications anymore after temporary network problems.

sysexcontrol:
@sophokles73 : With having PR #875 we now have a solution so that custom protocol adapters can reopen command receiver links themselves.
So I propose to remove this issue from the 0.8 project.

What remains a *fundamental* problem IMHO though are command receiver links for MQTT devices that stay connected for long.

Maybe the default implementation of the `connectionLost` handler in the MQTT adapter should close the MQTT connection (instead of inheriting the NOP implementation of the base class), or am I missing something?
This should be small work and could be done instead of fixing this issue in 0.8.

WDYT?

sophokles73:
> Maybe the default implementation of the connectionLost handler in the MQTT adapter should close the MQTT connection (instead of inheriting the NOP implementation of the base class), or am I missing something?

That should do the trick, yes. In fact, only those connections to devices need to be closed, which have subscribed to the control topic. BTW the same holds true for the AMQP adapter.

If we implement this default behavior in the MQTT and AMQP adapter, don't you think we could close this issue as fixed?

sysexcontrol:
> In fact, only those connections to devices need to be closed, which have subscribed to the control topic. 

:+1: this is true.

> If we implemented this default behavior in the MQTT and AMQP adapter, don't you think we could close this issue as fixed?

Yes, after the MQTT and AMQP adapters are equipped with such a connection-loss handler, this issue can be closed IMHO.

sophokles73:
@sysexcontrol I think this has been fixed with #875 and #902 
if you agree, can you close this issue?

sysexcontrol:
Yes, I agree - this is now fixed. Closing the issue.

PR #902, opened 2018-11-16
Handle command connection loss

sophokles73:
This PR addresses #830 

A straightforward approach for solving the issue would be to simply re-create all links once the connection is re-restablished. However, with a large number of devices connected to the adapter instance, this might result in a *burst* of link establishment traffic whose outcome is hard to predict.

The approach taken by this PR is therefore a different one: when a device subscribes to commands and a corresponding CommandConsumer is created, a device specific liveness check is registered which periodically checks for the link to be *alive* and, if not, tries to re-create the link. This should provide for much better distribution of re-creation attempts while being able to automatically recover from the connection loss.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/CommandConnectionImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/CommandConnectionImpl.java
@@ -17,6 +17,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.eclipse.hono.auth.Device;
 import org.eclipse.hono.client.CommandConnection;
@@ -38,10 +39,21 @@ import io.vertx.core.Vertx;
 public class CommandConnectionImpl extends HonoClientImpl implements CommandConnection {
 
     /**
+     * The minimum number of milliseconds to wait between checking a
+     * command consumer link's liveness.
+     */
+    public static final long MIN_LIVENESS_CHECK_INTERVAL_MILLIS = 2000;
+
+    /**
      * The consumers that can be used to receive command messages.
      * The device, which belongs to a tenant is used as the key, e.g. <em>DEFAULT_TENANT/4711</em>.
      */
-    private final Map<String, MessageConsumer> commandReceivers = new HashMap<>();
+    private final Map<String, MessageConsumer> commandConsumers = new HashMap<>();
+    /**
+     * A mapping of command consumer addresses to vert.x timer IDs which represent the
+     * liveness checks for the consumers.
+     */
+    private final Map<String, Long> livenessChecks = new HashMap<>();
 
     /**
      * Creates a new client for a set of configuration properties.
@@ -78,7 +90,7 @@ public class CommandConnectionImpl extends HonoClientImpl implements CommandConn
     @Override
     protected void clearState() {
         super.clearState();
-        commandReceivers.clear();
+        commandConsumers.clear();
     }
 
     /**
@@ -96,7 +108,7 @@ public class CommandConnectionImpl extends HonoClientImpl implements CommandConn
 
         return executeOrRunOnContext(result -> {
             final String key = Device.asAddress(tenantId, deviceId);
-            final MessageConsumer messageConsumer = commandReceivers.get(key);
+            final MessageConsumer messageConsumer = commandConsumers.get(key);
             if (messageConsumer != null) {
                 log.debug("cannot create concurrent command consumer [tenant: {}, device-id: {}]", tenantId, deviceId);
                 result.fail(new ResourceConflictException("message consumer already in use"));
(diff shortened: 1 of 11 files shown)


client/src/main/java/org/eclipse/hono/client/impl/CommandConnectionImpl.java
```diff
@@ -105,14 +116,67 @@ protected void clearState() {
...
+        }
+
+        return createCommandConsumer(tenantId, deviceId, commandConsumer, remoteCloseHandler).map(c -> {
+
+            final String address = Device.asAddress(tenantId, deviceId);
+            final long effectiveCheckInterval = Math.max(MIN_LIVENESS_CHECK_INTERVAL_MILLIS, checkInterval);
+            final long livenessCheckId = vertx.setPeriodic(effectiveCheckInterval, timerId -> {
+                if (isShutdown()) {
+                    vertx.cancelTimer(timerId);
+                } else if (isConnectedInternal() && !commandConsumers.containsKey(address)) {
```
sysexcontrol: IMHO this is a critical part of the implementation:
while the check for `commandConsumers.containsKey(address)` should work for the standard behaviour, I fear there is a danger that a recreation of a command consumer could take very long, or even hang (especially with previous connection losses).
In this case, this periodic timer could fire again and again and trigger more and more creations of command consumers, that possibly later all would succeed. Resulting in suddenly multiple command consumers, recreated out of one.

How about using something like an AtomicBoolean explicitly in this code, and check for this? By such an implementation there would be no dependency to the duration of the command consumer creation that could result in unpredictable behaviour.

sophokles73: That's a very good catch :+1: 
We are already doing something similar when we create senders ...

client/src/main/java/org/eclipse/hono/client/impl/CommandConnectionImpl.java
```diff
@@ -105,14 +116,67 @@ protected void clearState() {
...
+            final String tenantId,
+            final String deviceId,
+            final Handler<CommandContext> commandConsumer,
+            final Handler<Void> remoteCloseHandler,
+            final long checkInterval) {
+
+        Objects.requireNonNull(tenantId);
+        Objects.requireNonNull(deviceId);
+        Objects.requireNonNull(commandConsumer);
+        if (checkInterval < 0) {
```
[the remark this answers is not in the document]
sophokles73: I think that if someone provides a negative value or zero here, he/she has a wrong understanding of the API altogether and should be notified of this wrong-being by means of an exception. On the other hand, providing a (too) small interval length indicates that he/she has basically understood the meaning of the parameter but simply intends to perform the check too frequently. Thus, the difference in handling. Does this make sense?

sophokles73:
@sysexcontrol I have pushed some changes. Can you take another look?

client/src/main/java/org/eclipse/hono/client/impl/CommandConnectionImpl.java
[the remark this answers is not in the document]
sysexcontrol: Yes, kind of.
And after having read this part of the MQTT Spec:
>A Keep Alive value of zero (0) has the effect of turning off the keep alive mechanism. This means that, in this case, the Server is not required to disconnect the Client on the grounds of inactivity.

it seems to be that `0` is a valid keepAlive time (which is fed into the checkInterval here), but negative values cannot occur, so :+1: for keeping it like that.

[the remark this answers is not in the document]
sysexcontrol: One point though:
>throw new IllegalArgumentException("liveness check interval must be > 0");

should be changed to
>throw new IllegalArgumentException("liveness check interval must be >= 0");

client/src/main/java/org/eclipse/hono/client/impl/CommandConnectionImpl.java
```diff
@@ -147,34 +148,55 @@ protected void clearState() {
...
+                    // set a lock in order to prevent spawning multiple attempts
+                    // to re-create the consumer
+                    log.debug("trying to re-create command consumer [tenant: {}, device-id: {}]",
                             tenantId, deviceId);
                     // we try to re-create the link using the original parameters
                     // which will put the consumer into the cache again, if successful
                     createCommandConsumer(tenantId, deviceId, commandConsumer, remoteCloseHandler)
                     .map(consumer -> {
-                        log.debug("successfully re-established command consumer [tenant: {}, device-id: {}]",
+                        log.debug("successfully re-created command consumer [tenant: {}, device-id: {}]",
```
sysexcontrol: Shouldn't it be `recreate`, instead of `re-create`?

client/src/main/java/org/eclipse/hono/client/impl/CommandConnectionImpl.java
```diff
@@ -147,34 +148,55 @@ protected void clearState() {
...
+
+    Handler<Long> newLivenessCheck(
+            final String tenantId,
+            final String deviceId,
+            final String key,
+            final Handler<CommandContext> commandConsumer,
+            final Handler<Void> remoteCloseHandler) {
+
+        final AtomicBoolean recreating = new AtomicBoolean(false);
+        return timerId -> {
```
sysexcontrol: Looks correct now :+1:

PR #907, opened 2018-11-21
[#895] Don't create error span if no client certs were given

### LINKED ISSUE #895: HTTP-Adapter: misleading error trace if client certificate check failed (but basic auth succeeded)
Using the Hono HTTP adapter with Jaeger tracing enabled, we see traces like this:
http-adapter-trace

We use the device authentication via Basic Auth (which succeeded here) - the error above is about the failed attempt to verify the device using a client certificate (done before the basic auth check).

Having this marked as an error in the trace is somewhat misleading and makes it harder to filter the traces list for actual errors.

calohmn:
This fixes #895.

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/X509AuthHandler.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/X509AuthHandler.java
@@ -146,16 +146,16 @@ public class X509AuthHandler extends HonoAuthHandler {
         Objects.requireNonNull(handler);
 
         if (context.request().isSSL()) {
-            final SpanContext currentSpan = TracingHandler.serverSpanContext(context);
-            final Span span = tracer.buildSpan("verify device certificate")
-                    .asChildOf(currentSpan)
-                    .ignoreActiveSpan()
-                    .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_CLIENT)
-                    .withTag(Tags.COMPONENT.getKey(), getClass().getSimpleName())
-                    .start();
-
             try {
                 final Certificate[] path = context.request().sslSession().getPeerCertificates();
+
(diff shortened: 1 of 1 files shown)


calohmn:
I've amended the commit to minimize the changes.

With my first commit I wanted to make it clearer, where the `SSLPeerUnverifiedException` may actually get thrown (that is only in the line with the `getPeerCertificates()` invocation) by reducing the try-catch scope accordingly. This lead to introducing early exits for the common cases "no ssl used" and "no certs given".

Omitting the early exits but keeping the minimized try-catch block, I would suggest this further change:
```
            Certificate[] peerCerts = null;
            try {
                peerCerts = context.request().sslSession().getPeerCertificates();
            } catch (SSLPeerUnverifiedException e) {
                LOG.debug("could not get client certificates: " + e.getMessage());
            }
            if (peerCerts != null) {
                 //..
            } else {
                handler.handle(Future.failedFuture(UNAUTHORIZED));
            }
```

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/X509AuthHandler.java
```diff
@@ -190,8 +190,7 @@ public final void parseCredentials(final RoutingContext context, final Handler<A
                 });
             } catch (SSLPeerUnverifiedException e) {
                 // client certificate has not been validated
-                TracingHelper.logError(span, e);
-                span.finish();
+                LOG.debug("could not get client certificates: " + e.getMessage());
```
sophokles73: can you use the usual placeholder syntax for log statements, i.e.
`LOG.debug("could not get client certificates: {}", e.getMessage())`

calohmn: yes, I've changed that.

PR #911, opened 2018-11-22
[#909] Check tenant config when processing command response

### LINKED ISSUE #909: HTTP Adapter: missing adapter-enabled-for-tenant check when processing command response messages
When processing a command response message, the HTTP protocol adapter currently doesn't check whether the adapter is enabled for the given tenant.

calohmn:
This fixes #909.

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -936,22 +936,40 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
                     .withTag(TracingHelper.TAG_AUTHENTICATED.getKey(), authenticatedDevice != null)
                     .start();
 
-            // send response message to application via sender link
-            sendCommandResponse(tenant, commandResponse, currentSpan.context())
-            .map(delivery -> {
-                metrics.incrementCommandResponseDeliveredToApplication(tenant);
-                LOG.trace("delivered command response [command-request-id: {}] to application", commandRequestId);
-                currentSpan.log("delivered command response to application");
-                ctx.response().setStatusCode(HttpURLConnection.HTTP_ACCEPTED);
-                ctx.response().end();
-                return delivery;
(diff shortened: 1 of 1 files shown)


adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -936,22 +936,40 @@ public final void uploadCommandResponseMessage(
...
+                            }).setHandler(c -> {
+                                currentSpan.finish();
+                            });
+
+                    return Future.succeededFuture();
+                } else {
+                    // this adapter is not enabled for the tenant
+                    return Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_FORBIDDEN,
+                            "adapter is not enabled for tenant"));
+                }
```
sophokles73: I would rather have this whole block moved to a subsequent `compose` method invocation, i.e. if the tenant enabled check succeeds, simply return a succeeded future and then do the sending in the next composed block ...

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -936,22 +936,40 @@ public final void uploadCommandResponseMessage(
...
-            sendCommandResponse(tenant, commandResponse, currentSpan.context())
-            .map(delivery -> {
-                metrics.incrementCommandResponseDeliveredToApplication(tenant);
-                LOG.trace("delivered command response [command-request-id: {}] to application", commandRequestId);
-                currentSpan.log("delivered command response to application");
-                ctx.response().setStatusCode(HttpURLConnection.HTTP_ACCEPTED);
-                ctx.response().end();
-                return delivery;
+            final Future<TenantObject> tenantConfigTracker = getTenantConfiguration(tenant, currentSpan.context());
+            tenantConfigTracker.compose(tenantObj -> {
```
sysexcontrol: In contrast to the `doUploadMessage` method we do not need the `tenantConfigTracker` for further handling,
so I propose to just code it like 

`getTenantConfiguration(tenant, ...).compose(tenantObj -> {
. . . 
});`

calohmn:
I've applied both changes and amended the commit.

PR #912, opened 2018-11-22
[#669] Handle one-way commands in the MQTT adapter.

sysexcontrol:
The implementation and documentation is contained in this PR.
Release notes for one-way commands are contained as well.

After this PR has been merged, the #669 issue can be closed.


site/content/release-notes.md
```diff
@@ -4,13 +4,19 @@ title = "Release Notes"
 
 ## 0.8 (not released yet)
 
+### New Features
+
+* The already specified new message exchange pattern - called *one-way commands* - is now available to *business applications*. Therefore the `CommandClient` class was extended by a method `sendOneWayCommand` that does not expect a response from the device. See the `HonoExampleApplicationBase` class for how to use this new command pattern.
+  On the adapter side, this pattern is supported for HTTP and MQTT so far.
```
sophokles73: why not AMQP?

site/content/user-guide/mqtt-adapter.md
```diff
@@ -194,6 +198,13 @@ For example, if the [HonoExampleApplication]({{< relref "dev-guide/java_client_c
        "brightness" : 79
     }
 
+If the command is a *one-way* command, it will arrive as follows:
+
+    $ c///q//setBrightness {
```
sophokles73: are your sure about the `c/`? IMHO the MQTT adapter always publishes to `control/`

site/content/release-notes.md
sysexcontrol: Not implemented yet - no technical reasons.
Do you think it is being essential for 0.8 Release?

site/content/user-guide/mqtt-adapter.md
sysexcontrol: Actually, IMHO it depends on to which topic the client originally subscribed.
It was like that in the already existing example, but it does not match with all the other topics on that page (which I didn't check).

So, good point - I will change this to `control` instead of the (also valid) `c`. This should eliminate the confusion.

site/content/release-notes.md
sophokles73: There is nothing to do for the AMQP adapter because it simply forwards the original message, i.e. if the original message does not contain a reply-to, then it is a one-way command. My point is that you can safely mention the AMQP adapter also supporting one-way commands :-)

sysexcontrol: You are right, very good :+1: (haven't thought about that until now).
I will adapt the release notes.

Issue #914, opened 2018-11-22
Add OpenTracing support in the base classes for the device registration, tenant and credentials API


calohmn:
Implementations of the device registration, tenant and credentials API should be able to create trace spans based on the `SpanContext` propagated by the protocol adapters.

Therefore the base classes for these APIs should extract the `SpanContext` from the incoming AMQP message and provide it to the implementation classes.

calohmn:
Steps to take here:
- extract `SpanContext` and put in `EventBusMessage` -> #915
- provide `Tracer` instance in `EventBusService` -> #917
- add `SpanContext` param in `CredentialsService`, `TenantService` and `RegistrationService` interface methods and base implementation classes
- create spans in base implementation classes

calohmn:
Necessary changes have been made now.
What's missing still is adding tracing support in `CredentialsClient` - that will be handled in #974.

PR #915, opened 2018-11-22
[#914] TenantAmqpEndpoint: extract SpanContext and put in EventBusMessage

calohmn:
Put SpanContext extracted from TenantAmqpEndpoint message into EventBusMessage.

This is for #914.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/tracing/JsonObjectExtractAdapter.java
+++ b/core/src/main/java/org/eclipse/hono/tracing/JsonObjectExtractAdapter.java
@@ -1,0 +1,74 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.tracing;
+
+import java.util.AbstractMap;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.Map.Entry;
+import java.util.Objects;
+
+import io.opentracing.propagation.TextMap;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * An adapter for extracting properties from a JSON object.
+ *
+ */
+public class JsonObjectExtractAdapter implements TextMap {
+
+    private final JsonObject jsonObject;
+    private final String propertiesObjectKey;
+
+    /**
+     * Creates an adapter for a JSON object.
+     * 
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/eclipse/hono/tracing/JsonObjectExtractAdapter.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Objects;
+
+import io.opentracing.propagation.TextMap;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * An adapter for extracting properties from a JSON object.
+ *
+ */
+public class JsonObjectExtractAdapter implements TextMap {
```
sophokles73: to be honest,I do not really like the idea of putting the tracing related information into the message payload. What about injecting/extracting the tracing context to/from a vert.x MultiMap which can then be passed along in the headers of the event bus message? It would also be much easier to do because MultiMap already implements `Iterable<Map.Entry<String, String>>` ...

calohmn: yes, good idea. I've created PR #935 with such a change.
One thing to note here: This requires the follow-up commits (where the `SpanContext` will be added to the Tenant/Registration/Credentials Services) to be more invasive, having to add a `SpanContext` param alongside the `EventBusMessage` param in several methods.
On the plus side, this could be seen as making things more explicit.

sophokles73: or maybe simply include the extracted SpanContext in the `EventBusMessage` instance itself ...

calohmn: Keeping the `SpanContext` in a field of the `EventBusMessage` would require defining a custom `io.vertx.core.eventbus.MessageCodec` for the `EventBusMessage` class, I guess (so that vertx can marshall it across the event bus). 
Then I would rather use the solution of this PR, keeping things more straightforward. The way the `SpanContext` is stored in the JSON is not really revealed outside of the `EventBusMessage` class anyway, so it is more like an implementation detail.

sophokles73: I don't think that's necessary. We can simply add the `SpanContext` instance that is extracted from the event bus message headers to the `EventBusMessage` instance that is created from the event bus message's content **after** it has been created but **before** handing it over to the service implementation class. The span context would then simply be a *transient* field in the `EventBusMessage`. Since the context needs to be propagated in the request message only, there shouldn't be a problem, right?

calohmn: Ah, you mean, choosing the newer approach from PR #935 (using the headers MultiMap), but after having received the `EventBusMessage`, adding the `SpanContext` in there, so that method signatures can stay the same (without an added `SpanContext` param)?
Yes, right, that sounds good.

calohmn:
Closed in favour of  #935.

Issue #871, opened 2018-10-31
Add possibility to register credentials with cleartext password


b-abel:
For prototyping and learning it would be helpful to have an operation in the Credentials (REST) API which allows to register new credentials without the need to hash the password. The `CredentialsHttpEndpoint` could be extended with an operation that takes a cleartext password and handles the hashing.

sophokles73:
That would be highly appreciated :+1: 
It would be cool if the hash function to use could be configured by means of an environment variable with a default of *bcrypt*. WDYT?

b-abel:
Implemented by PR  #949.

PR #923, opened 2018-11-26
[#871] Add hashing of plain text password to CredentialsHttpEndpoint.

b-abel:

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsHttpEndpoint.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsHttpEndpoint.java
@@ -19,10 +19,13 @@ import java.util.Objects;
 import io.vertx.core.http.HttpHeaders;
 import org.eclipse.hono.config.ServiceConfigProperties;
 import org.eclipse.hono.service.http.AbstractHttpEndpoint;
+import org.eclipse.hono.util.ClearTextPassword;
 import org.eclipse.hono.util.CredentialsConstants;
 import org.eclipse.hono.util.EventBusMessage;
 import org.springframework.beans.factory.annotation.Autowired;
 
+import io.vertx.core.Context;
+import io.vertx.core.Future;
 import io.vertx.core.Vertx;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.Router;
@@ -41,6 +44,8 @@ public final class CredentialsHttpEndpoint extends AbstractHttpEndpoint<ServiceC
     // path parameters for capturing parts of the URI path
     private static final String PARAM_TYPE = "type";
     private static final String PARAM_AUTH_ID = "auth_id";
+    private static final String DEFAULT_HASH_FUNCTION = System.getProperty("pwd-hash-function",
+            CredentialsConstants.HASH_FUNCTION_BCRYPT);
 
     /**
      * Creates an endpoint for a Vertx instance.
@@ -73,6 +78,7 @@ public final class CredentialsHttpEndpoint extends AbstractHttpEndpoint<ServiceC
         // add credentials
         router.post(pathWithTenant).handler(bodyHandler);
         router.post(pathWithTenant).handler(this::extractRequiredJsonPayload);
+        router.post(pathWithTenant).handler(this::hashPasswordIfPLain);
         router.post(pathWithTenant).handler(this::addCredentials);
 
         // get credentials by auth-id and type
@@ -239,6 +245,52 @@ public final class CredentialsHttpEndpoint extends AbstractHttpEndpoint<ServiceC
         sendAction(ctx, requestMsg, getDefaultResponseHandler(ctx,
                 status -> status == HttpURLConnection.HTTP_OK,
                 null));
+    }
+
+    private void hashPasswordIfPLain(final RoutingContext ctx) {
+        final JsonObject payload = (JsonObject) ctx.get(KEY_REQUEST_BODY);
+
+        if (CredentialsConstants.SECRETS_TYPE_PLAIN_PASSWORD.equals(payload.getString(CredentialsConstants.FIELD_TYPE))) {
+            final JsonObject payloadHashedPwd = payload.copy();
+            vertx.executeBlocking(blockingCodeHandler -> {
+                logger.debug("hashing password on vert.x worker thread [{}]", Thread.currentThread().getName());
+                blockingCodeHandler.complete(hashPwdAndUpdateSecrets(payloadHashedPwd));
+            }, ar -> {
+                ctx.put(KEY_REQUEST_BODY, ar.result());
+                ctx.next();
+            });
+
+        } else {
+            ctx.next();
+        }
+    }
+
+    private JsonObject hashPwdAndUpdateSecrets(final JsonObject payload) {
+        payload.put(CredentialsConstants.FIELD_TYPE, CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD);
+        payload.getJsonArray(CredentialsConstants.FIELD_SECRETS).forEach(
+                s -> {
+                    final JsonObject secret = (JsonObject) s;
+                    final String pwdHash = hashPassWord(secret);
+                    updateSecretWithPwdHash(secret, pwdHash);
+                });
+        return payload;
+    }
+
+    private String hashPassWord(final JsonObject secret) {
+        final String pwd = secret.getString(CredentialsConstants.FIELD_SECRETS_PWD_PLAIN);
+
+        if (DEFAULT_HASH_FUNCTION.equals(CredentialsConstants.HASH_FUNCTION_BCRYPT)) {
+            return ClearTextPassword.encodeBCrypt(pwd);
+        } else {
+            final byte[] salt = secret.getBinary(CredentialsConstants.FIELD_SECRETS_SALT);
+            return ClearTextPassword.encode(DEFAULT_HASH_FUNCTION, salt, pwd);
+        }
+    }
+
+    private void updateSecretWithPwdHash(final JsonObject secret, final String pwdHash) {
+        secret.put(CredentialsConstants.FIELD_SECRETS_PWD_HASH, pwdHash);
+        secret.put(CredentialsConstants.FIELD_SECRETS_HASH_FUNCTION, DEFAULT_HASH_FUNCTION);
+        secret.remove(CredentialsConstants.FIELD_SECRETS_PWD_PLAIN);
     }
 
 }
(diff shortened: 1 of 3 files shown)


service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsHttpEndpoint.java
```diff
@@ -241,4 +247,50 @@ private void getCredentialsForDevice(final RoutingContext ctx) {
                 null));
     }
 
+    private void hashPasswordIfPLain(final RoutingContext ctx) {
+        final JsonObject payload = (JsonObject) ctx.get(KEY_REQUEST_BODY);
+
+        if (CredentialsConstants.SECRETS_TYPE_PLAIN_PASSWORD.equals(payload.getString(CredentialsConstants.FIELD_TYPE))) {
```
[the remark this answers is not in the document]
b-abel: I can do that.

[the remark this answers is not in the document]
sophokles73: to me that would seem more natural :+1:

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsHttpEndpoint.java
```diff
@@ -241,4 +247,50 @@ private void getCredentialsForDevice(final RoutingContext ctx) {
...
+                    updateSecretWithPwdHash(secret, pwdHash);
+                });
+        return payload;
+    }
+
+    private String hashPassWord(final JsonObject secret) {
+        final String pwd = secret.getString(CredentialsConstants.FIELD_SECRETS_PWD_PLAIN);
+
+        if (DEFAULT_HASH_FUNCTION.equals(CredentialsConstants.HASH_FUNCTION_BCRYPT)) {
+            return ClearTextPassword.encodeBCrypt(pwd);
```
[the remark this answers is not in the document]
sophokles73: ah, my bad, didn't see that :+1:

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsHttpEndpoint.java
```diff
@@ -41,6 +44,8 @@
     // path parameters for capturing parts of the URI path
     private static final String PARAM_TYPE = "type";
     private static final String PARAM_AUTH_ID = "auth_id";
+    private static final String DEFAULT_HASH_FUNCTION = System.getProperty("pwd-hash-function",
```
sophokles73: we usually do not read environment variables directly but instead have them mapped to config object properties

b-abel: That would require to define a new ConfigProperties class, right? It feels a little bit like an overkill. 
An alternative idea would be to let the client request a hash function (and salt for shaX), which would be even closer to the current syntax. Example:

```
$ curl -i -X POST -H 'Content-Type: application/json' --data-binary '{
    "device-id": "4720",
    "type": "hashed-password",
    "auth-id": "sensor20",
    "secrets": [{
        "hash-function" : "bcrypt",
        "pwd-plain": "mylittlesecret"
    }]
  }' http://localhost:28080/credentials/DEFAULT_TENANT```

WDYT?

Comment (sophokles73) [reply service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsHttpEndpoint.java]: letting the user select a hash function sounds good, but we shouldn't allow the client to specify the salt. That should be handled on the server side where we can guarantee the use of a secure random number generator.

File: service-base/src/test/java/org/eclipse/hono/service/credentials/CredentialsPlainPasswordHelperTest.java [hunk 4]
```diff
@@ -0,0 +1,179 @@
...
+/**
+ * Unit tests for {@link CredentialsPlainPasswordHelper}.
+ */
+public class CredentialsPlainPasswordHelperTest {
+
+    private static final String PASSWORD = "mylittlesecret";
+    private static final byte[] SALT = "saltnpepper".getBytes(StandardCharsets.UTF_8);
+
+    /**
+     * Tests that password is correctly hashed with sha-256.
```
sophokles73: FMPOV the description of the test methods should start with `Verifies that`, implying that we expect the test to turn out positively, right?

service-base/src/test/java/org/eclipse/hono/service/credentials/CredentialsPlainPasswordHelperTest.java
```diff
@@ -0,0 +1,179 @@
...
+    public void secretIsUpdatedForBcrypt() {
+        final String hashFunction = CredentialsConstants.HASH_FUNCTION_BCRYPT;
+
+        final JsonObject secret = newPlainPwdSecret(hashFunction);
+
+        hashPwdAndUpdateSecret(secret);
+
+        assertNull(secret.getString(CredentialsConstants.FIELD_SECRETS_PWD_PLAIN));
+        assertNotNull(secret.getString(CredentialsConstants.FIELD_SECRETS_PWD_HASH));
+        assertNull(secret.getString(CredentialsConstants.FIELD_SECRETS_SALT));
```
sophokles73: maybe add a comment like `// salt is contained in pwd hash for bcrypt`?

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsPlainPasswordHelper.java
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * The cryptographically maximal strong salt for SHA-256 in bytes.
+     */
+    private static final int SALT_LENGTH_SHA256 = 32;
+    /**
+     * The cryptographically maximal strong salt for SHA-512 in bytes.
+     */
+    private static final int SALT_LENGTH_SHA512 = 64;
+
+    private static final String DEFAULT_HASH_FUNCTION = CredentialsConstants.HASH_FUNCTION_BCRYPT;
```
sophokles73: what about making this configurable in a property?

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsPlainPasswordHelper.java
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * Hashes a plain text password in the field @link{CredentialsConstants.FIELD_SECRETS_PWD_PLAIN} and updates the
+     * secret accordingly.
+     *
+     * @param secret The secret to be updated.
+     * @return The updated secret.
+     */
+    public static JsonObject hashPwdAndUpdateSecret(final JsonObject secret) {
+
+        if (secret.getString(CredentialsConstants.FIELD_SECRETS_PWD_PLAIN, "").isEmpty()) {
```
sophokles73: what happens if a client provides an empty cleartext pwd?
```
{ "pwd-plain": "" }
```

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsPlainPasswordHelper.java
b-abel: Since we decided to let the client choose the hash function, I suggest to completely remove this and **require** that the client does so. 
One could argue that it is a feature to allow that the client omits the `"hash-function"` field and lets the server decide. But FMPOV this small "feature" does not justify the additional complexity in code, API, documentation, error handling, ... (making this configurable would require to introduce a new `ConfigProperties` class for `CredentialsHttpEndpoint`). 

Another point is, that there is already a constant `DEFAULT_HASH_FUNCTION` in `CredentialsConstants`, which is used for validating passwords and which set to `sha-256`. Defining a second default hash function would IMHO lead to a semantic mess.

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsPlainPasswordHelper.java
b-abel: The client gets a HTTP 400 as for all other client errors in the Credentials REST API. We could at least log explicitly that the password is empty, which is consistent to the current state of client errors in the Credentials REST API.

sophokles73: ok, so we let this case *fall through* to the service implementation which will (hopefully) detect that the *pwd-hash* field is missing and will return a 400, right?

b-abel: In my last comment I meant that we should throw an exception if the clear text password is empty.

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsPlainPasswordHelper.java
sophokles73: Most, i.e. > 95%, of the clients will have no clue about hash functions and their security properties. Consequently, a service provider should in-fact not allow clients to provide a hash function at all but instead enforce the usage of a particular (safe) hash function of his own choice. 

I would therefore propose to ignore the *hash-function* property (or even return a 400 if present) and instead *always* use the configured hash function.

sophokles73: Not really *change* but *evolve*, I would say ...

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsPlainPasswordHelper.java
sophokles73: I understand, however, I think we can safely let the service implementation handle this problem. It needs to do that anyway, don't you think?

b-abel: So, you are saying we could leave this line as it is in the snippet above?

sophokles73: I think so.

b-abel:
At the current state of the discussion here, this is no longer the small hack in the REST endpoint of the credentials API. For me it would be better to put the hashing into the service instead of the `CredentialsHttpEndpoint`. This implies that it would be available over AMQP (and potential other protocols to be added later) too.
The question remains what the right level is: in the base services (`BaseCredentialsService` or `CompleteBaseCredentialsService`) or in the Device Registry application (`FileBasedCredentialsService`)? **My suggestion** is to add the hashing into `CompleteBaseCredentialsService`. I would prepare a new PR for that.

Issue #932, opened 2018-11-30
Issue compiling Hono


ctron:
I noticed this a while back, but didn't really look into it, as it only sporadically popped up in a CI instance.

However now I run into this when deploying Hono 0.8 (the released version) on OpenShift. I didn't run into this before though.

Compiling Hono gives:

~~~
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.7.0:testCompile (default-testCompile) on project hono-client: Compilation failure
[ERROR] /tmp/src/client/src/test/java/org/eclipse/hono/client/impl/CommandConnectionImplTest.java:[243,25] cannot find symbol
[ERROR]   symbol:   method close(io.vertx.core.Handler<io.vertx.core.AsyncResult<java.lang.Void>>)
[ERROR]   location: variable consumer of type java.lang.Object
[ERROR] -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.7.0:testCompile (default-testCompile) on project hono-client: Compilation failure
/tmp/src/client/src/test/java/org/eclipse/hono/client/impl/CommandConnectionImplTest.java:[243,25] cannot find symbol
symbol:   method close(io.vertx.core.Handler<io.vertx.core.AsyncResult<java.lang.Void>>)
location: variable consumer of type java.lang.Object
~~~

ctron:
I actually can reproduce this locally when doing `mvn clean test` in the `client` directory.

Running with Java:
~~~
openjdk version "1.8.0_191"
OpenJDK Runtime Environment (build 1.8.0_191-b12)
OpenJDK 64-Bit Server VM (build 25.191-b12, mixed mode)
~~~

ctron:
Same result with:

~~~
java version "1.8.0_151"
Java(TM) SE Runtime Environment (build 1.8.0_151-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.151-b12, mixed mode)
~~~

sophokles73:
We are running into the same issue. However, since the code seems to be correct (at least nobody who took a look so far could find something suspicious), I would attribute this to the (growing) list of issues with OpenJDK 8 ... I (we at Bosch) have moved on to OpenJDK 11.0.1 which behaves as expected ...

ctron:
[...]

PR #934, opened 2018-11-30
[#932] Fix compile error with Java 8 for S2I

ctron:


deploy/src/main/deploy/openshift_s2i/hono-template.yml
```diff
@@ -407,7 +407,7 @@ objects:
           name: fabric8-s2i-java-custom:2.3
         env:
         - name: MAVEN_ARGS_APPEND
-          value: -B -pl org.eclipse.hono:hono-adapter-amqp-vertx --also-make -Pnetty-tcnative
+          value: -B -pl org.eclipse.hono:hono-adapter-amqp-vertx --also-make -Pnetty-tcnative -Dmaven.test.skip=true
```
sophokles73: hmm, is that really what you want to do, i.e. **not** run the tests?

ctron: That is what I need to do, in order to work around this issue.

On the other side, this is a released version of Hono. So I am not sure if re-running the unit tests, for every local installation, is necessary.

sophokles73: I see. Just wanted be sure we know what we're doing ;-)

PR #935, opened 2018-11-30
[#914] Put extracted SpanContext into EventBusMessage delivery headers

calohmn:
Put the `SpanContext`, extracted from an AMQP message received at a `RequestResponseEndpoint`, into the delivery headers of the vert.x event bus message.

This is for #914.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
@@ -17,6 +17,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+
 import org.apache.qpid.proton.amqp.transport.AmqpError;
 import org.apache.qpid.proton.message.Message;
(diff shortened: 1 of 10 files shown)


service-base/src/main/java/org/eclipse/hono/service/amqp/RequestResponseEndpoint.java
```diff
@@ -496,4 +527,32 @@ protected boolean isValidReplyToAddress(final ResourceIdentifier replyToAddress)
...
+            deliveryOptions.setHeaders(multiMap);
+        }
+        return deliveryOptions;
+    }
+
+    /**
+     * Extracts a {@code SpanContext} out of the delivery annotations of the given {@code Message}.
+     * 
+     * @param msg The AMQP message.
+     * @return The extracted {@code SpanContext} (may be {@code null}).
```
sophokles73: should document that msg must not be `null`

calohmn: I've adapted the javadoc accordingly.

Issue #927, opened 2018-11-29
Support limiting the number of simultaneously connected devices


sysexcontrol:
Protocol adapters should support limiting the number of simultaneously connected devices per tenant.
The limit should apply to all devices of the tenant connected to any protocol adapter instance (regardless of transport protocol).
The limit must be configurable per tenant.

sophokles73:
@kaniyan can we close this issue from your point of view?

kaniyan:
@sophokles73 With the last merged PRs, now this feature is available in Hono. Yes, we can close this issue.

sophokles73:
fixed in https://github.com/eclipse/hono/commit/bba1d9d6e9b16b218c2806ac05b000ec5b8b784d

Issue #931, opened 2018-11-29
Limit number of MQTT connections according to resource availability


sophokles73:
Currently, the MQTT protocol adapter accepts any incoming connection request from devices. For each connection and message received, the adapter needs to maintain some state which is usually kept in memory. In order to not run out of resources during runtime, the MQTT adapter should only accept a reasonable number of connections.

It seems desirable that the adapter determines this number dynamically during startup based on the  resources (RAM, CPU) available to the JVM.

b-abel:
To my knowledge there is no option on Vert.x's [NetServer](https://vertx.io/docs/apidocs/io/vertx/core/net/NetServer.html) to limit to number of connections. We need to count the connections and closeevery connection above the limit, e.g. in a connection handler as suggested in [this issue](https://github.com/eclipse-vertx/vert.x/issues/2542).

b-abel:
**Note regarding the HTTP adapter**
It supports [HTTP 1.1 persistent connections](https://tools.ietf.org/html/rfc7230#section-6.3):
```
$ curl -v -X POST -i -u sensor1@DEFAULT_TENANT:hono-secret -H 'Content-Type: application/json' --data-binary '{"temp": 5}' http://localhost:8080/telemetry -X POST -i -u sensor1@DEFAULT_TENANT:hono-secret http://localhost:8080/telemetry                                                                
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8080 (#0)
* Server auth using Basic with user 'sensor1@DEFAULT_TENANT'
> POST /telemetry HTTP/1.1
> Host: localhost:8080
> Authorization: Basic c2Vuc29yMUBERUZBVUxUX1RFTkFOVDpob25vLXNlY3JldA==
> User-Agent: curl/7.58.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 11
> 
* upload completely sent off: 11 out of 11 bytes
< HTTP/1.1 202 Accepted
HTTP/1.1 202 Accepted
< content-length: 0
content-length: 0

< 
* Connection #0 to host localhost left intact
Note: Unnecessary use of -X or --request, POST is already inferred.
* Found bundle for host localhost: 0x55ad961886a0 [can pipeline]
* Re-using existing connection! (#0) with host localhost
* Connected to localhost (127.0.0.1) port 8080 (#0)
[...]

Issue #929, opened 2018-11-29
Protocol adapters should support limiting the data volume per time period per tenant


sysexcontrol:

sophokles73:
fixed in https://github.com/eclipse/hono/commit/84eb7b0e7c704fcbcc9c688a4efac91b02a79ecb

PR #937, opened 2018-11-30
[#927] Limit number of connections per tenant

kaniyan:
This PR introduces an interface `TenantPlanValidator` which contains methods to check if further connections are allowed. A default Implementation of this interface based on the metrics data retrieved from the prometheus back end is provided. This validates the number of connections by comparing with the configured limit. Mqtt adapter uses the above method to limit the number of connections.

### DIFF
--- a/adapters/mqtt-vertx-base/src/test/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapterTest.java
+++ b/adapters/mqtt-vertx-base/src/test/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapterTest.java
@@ -51,6 +51,7 @@ import org.eclipse.hono.service.auth.DeviceUser;
 import org.eclipse.hono.service.auth.device.AuthHandler;
 import org.eclipse.hono.service.metric.MetricsTags;
 import org.eclipse.hono.service.metric.MetricsTags.EndpointType;
+import org.eclipse.hono.service.plan.ResourceLimitChecks;
 import org.eclipse.hono.util.EventConstants;
 import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.RegistrationConstants;
@@ -111,6 +112,7 @@ public class AbstractVertxBasedMqttProtocolAdapterTest {
     private RegistrationClient regClient;
     private TenantClient tenantClient;
     private AuthHandler<MqttContext> authHandler;
+    private ResourceLimitChecks resourceLimitChecks;
     private MqttProtocolAdapterProperties config;
     private MqttAdapterMetrics metrics;
     private CommandConnection commandConnection;
@@ -175,6 +177,8 @@ public class AbstractVertxBasedMqttProtocolAdapterTest {
         when(commandConnection.connect(any(Handler.class))).thenReturn(Future.succeededFuture(commandConnection));
 
         authHandler = mock(AuthHandler.class);
+        resourceLimitChecks = mock(ResourceLimitChecks.class);
+        when(resourceLimitChecks.isConnectionLimitExceeded(any(TenantObject.class))).thenReturn(Future.succeededFuture());
     }
 
     /**
@@ -1000,6 +1004,32 @@ public class AbstractVertxBasedMqttProtocolAdapterTest {
         verify(metrics).decrementUnauthenticatedConnections();
     }
 
+    /**
+     * Verifies that the connection is rejected due to the limit exceeded.
+     */
+    @Test
+    public void testConnectionsLimitExceeded() {
+
+        // GIVEN an adapter requiring devices to authenticate endpoint
+        final MqttServer server = getMqttServer(false);
+        config.setAuthenticationRequired(true);
+        final AbstractVertxBasedMqttProtocolAdapter<MqttProtocolAdapterProperties> adapter = getAdapter(server);
+        forceClientMocksToConnected();
+
+        // WHEN a device tries to establish a connection
+        when(authHandler.authenticateDevice(any(MqttContext.class)))
+                .thenReturn(Future.succeededFuture(new DeviceUser("DEFAULT_TENANT", "4711")));
+        when(resourceLimitChecks.isConnectionLimitExceeded(any(TenantObject.class)))
+                .thenReturn(Future.failedFuture("Connections limit exceeded"));
+        final MqttEndpoint endpoint = getMqttEndpointAuthenticated();
+        adapter.handleEndpointConnection(endpoint);
+
+        // THEN the adapter has tried to authenticate the device
+        verify(authHandler).authenticateDevice(any(MqttContext.class));
+        // THEN the connection request is rejected
+        verify(endpoint).reject(MqttConnectReturnCode.CONNECTION_REFUSED_NOT_AUTHORIZED);
+    }
+
     private void forceClientMocksToConnected() {
         when(tenantServiceClient.isConnected()).thenReturn(Future.succeededFuture());
         when(messagingClient.isConnected()).thenReturn(Future.succeededFuture());
@@ -1064,6 +1094,7 @@ public class AbstractVertxBasedMqttProtocolAdapterTest {
         adapter.setCredentialsServiceClient(credentialsServiceClient);
         adapter.setCommandConnection(commandConnection);
         adapter.setAuthHandler(authHandler);
+        adapter.setResourceLimitChecks(resourceLimitChecks);
 
         if (server != null) {
             adapter.setMqttInsecureServer(server);
--- a/core/src/main/java/org/eclipse/hono/util/TenantObject.java
+++ b/core/src/main/java/org/eclipse/hono/util/TenantObject.java
@@ -496,4 +496,18 @@ public final class TenantObject extends JsonBackedValueObject {
                 .put(TenantConstants.FIELD_ADAPTERS_TYPE, type)
                 .put(TenantConstants.FIELD_ENABLED, enabled);
     }
+
+    /**
+     * Gets the connections limit for the tenant if configured, else returns
+     * {@link TenantConstants#DEFAULT_MAX_CONNECTIONS}.
+     *
+     * @return The connections limit.
+     */
+    @JsonIgnore
+    public long getConnectionsLimit() {
+        return Optional.ofNullable(getProperty(TenantConstants.LIMITS))
+                .map(limits -> getProperty((JsonObject) limits, TenantConstants.MAX_CONNECTIONS,
+                        (Number) TenantConstants.DEFAULT_MAX_CONNECTIONS).longValue())
+                .orElse(TenantConstants.DEFAULT_MAX_CONNECTIONS);
+    }
 }
--- a/service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -25,12 +25,15 @@ import org.eclipse.hono.config.ApplicationConfigProperties;
 import org.eclipse.hono.config.ClientConfigProperties;
 import org.eclipse.hono.config.VertxProperties;
 import org.eclipse.hono.service.cache.SpringCacheProvider;
+import org.eclipse.hono.service.plan.ResourceLimitChecks;
+import org.eclipse.hono.service.plan.PrometheusBasedResourceLimitChecks;
 import org.eclipse.hono.util.CommandConstants;
 import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.CredentialsConstants;
 import org.eclipse.hono.util.RegistrationConstants;
 import org.eclipse.hono.util.TenantConstants;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.cache.guava.GuavaCacheManager;
 import org.springframework.context.annotation.Bean;
@@ -55,7 +58,7 @@ public abstract class AbstractAdapterConfig {
      * The Tracer will be resolved by means of a Java service lookup.
      * If no tracer can be resolved this way, the {@code NoopTracer} is
      * returned.
-     * 
+     *
      * @return The tracer.
      */
     @Bean
@@ -71,7 +74,7 @@ public abstract class AbstractAdapterConfig {
      * This method creates new Vert.x default options and invokes
      * {@link VertxProperties#configureVertx(VertxOptions)} on the object returned
      * by {@link #vertxProperties()}.
-     * 
+     *
      * @return The Vert.x instance.
      */
     @Bean
@@ -272,7 +275,7 @@ public abstract class AbstractAdapterConfig {
 
     /**
      * Exposes the provider for caches as a Spring bean.
-     * 
+     *
      * @return The provider instance.
      */
     @Bean
@@ -307,7 +310,7 @@ public abstract class AbstractAdapterConfig {
 
     /**
      * Exposes configuration options for vertx.
-     * 
+     *
      * @return The Properties.
      */
     @ConfigurationProperties("hono.vertx")
@@ -318,7 +321,7 @@ public abstract class AbstractAdapterConfig {
 
     /**
      * Create a new cache provider based on Guava and Spring Cache.
-     * 
+     *
      * @param config The configuration to use as base for this cache.
      * @return A new cache provider or {@code null} if no cache should be used.
      */
@@ -361,5 +364,17 @@ public abstract class AbstractAdapterConfig {
     @Bean
     public HealthCheckServer healthCheckServer() {
         return new VertxBasedHealthCheckServer(vertx(), applicationConfigProperties());
+    }
+
+    /**
+     * Creates a new instance of {@link ResourceLimitChecks} based on prometheus metrics data.
+     * 
+     * @return A ResourceLimitChecks instance.
+     */
+    @Bean
+    @ConfigurationProperties(prefix = "hono.plan.prometheusBased")
+    @ConditionalOnProperty(name = "hono.plan.prometheusBased.host")
+    public ResourceLimitChecks resourceLimitChecks() {
+        return new PrometheusBasedResourceLimitChecks(vertx());
     }
 }
--- a/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
@@ -37,6 +37,8 @@ import org.eclipse.hono.config.ProtocolAdapterProperties;
 import org.eclipse.hono.service.auth.ValidityBasedTrustOptions;
 import org.eclipse.hono.service.limiting.ConnectionLimitManager;
 import org.eclipse.hono.service.monitoring.ConnectionEventProducer;
+import org.eclipse.hono.service.plan.NoopResourceLimitChecks;
+import org.eclipse.hono.service.plan.ResourceLimitChecks;
 import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.CredentialsConstants;
 import org.eclipse.hono.util.EventConstants;
@@ -93,7 +95,7 @@ public abstract class AbstractProtocolAdapterBase<T extends ProtocolAdapterPrope
     private ConnectionLimitManager connectionLimitManager;
 
     private ConnectionEventProducer connectionEventProducer;
-
+    private ResourceLimitChecks resourceLimitChecks = new NoopResourceLimitChecks();
     private final ConnectionEventProducer.Context connectionEventProducerContext = new ConnectionEventProducer.Context() {
 
         @Override
@@ -105,7 +107,7 @@ public abstract class AbstractProtocolAdapterBase<T extends ProtocolAdapterPrope
 
     /**
      * Adds a Micrometer sample to a command context.
-     * 
+     *
      * @param ctx The context to add the sample to.
      * @param sample The sample.
      * @throws NullPointerException if ctx is {@code null}.
@@ -118,7 +120,7 @@ public abstract class AbstractProtocolAdapterBase<T extends ProtocolAdapterPrope
     /**
      * Gets the Micrometer used to track the processing
      * of a command message.
-     * 
+     *
      * @param ctx The command context to extract the sample from.
      * @return The sample or {@code null} if the context does not
      *         contain a sample.
@@ -317,6 +319,26 @@ public abstract class AbstractProtocolAdapterBase<T extends ProtocolAdapterPrope
     }
 
     /**
+     * Sets the ResourceLimitChecks instance used to check if the number of connections exceeded the limit or not.
+     *
+     * @param resourceLimitChecks The ResourceLimitChecks instance
+     * @throws NullPointerException if the resourceLimitChecks is {@code null}.
+     */
+    @Autowired(required = false)
+    public final void setResourceLimitChecks(final ResourceLimitChecks resourceLimitChecks) {
+        this.resourceLimitChecks = Objects.requireNonNull(resourceLimitChecks);
+    }
+
+    /**
+     * Gets the ResourceLimitChecks instance used to check if the number of connections exceeded the limit or not.
+     *
+     * @return The ResourceLimitChecks instance.
+     */
+    protected final ResourceLimitChecks getResourceLimitChecks() {
+        return this.resourceLimitChecks;
+    }
+
+    /**
      * Establishes the connections to the services this adapter depends on.
      * <p>
      * Note that the connections will most likely not have been established when the
@@ -447,7 +469,7 @@ public abstract class AbstractProtocolAdapterBase<T extends ProtocolAdapterPrope
 
     /**
      * Validates a message's target address for consistency with Hono's addressing rules.
-     * 
+     *
      * @param address The address to validate.
      * @param authenticatedDevice The device that has uploaded the message.
      * @return A future indicating the outcome of the check.
@@ -934,7 +956,7 @@ public abstract class AbstractProtocolAdapterBase<T extends ProtocolAdapterPrope
      * <p>
      * This method creates a new {@code Message}, sets its content type and payload as an AMQP <em>Data</em> section
      * and then invokes {@link #addProperties(Message, ResourceIdentifier, boolean, String, JsonObject, Integer)}.
-     * 
+     *
      * @param target The resource that the message is targeted at.
      * @param regAssertionRequired {@code true} if the downstream peer requires the registration assertion to
      *            be included in the message.
--- a/service-base/src/main/java/org/eclipse/hono/service/plan/NoopResourceLimitChecks.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/plan/NoopResourceLimitChecks.java
@@ -1,0 +1,27 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.hono.service.plan;
+
+import io.vertx.core.Future;
+import org.eclipse.hono.util.TenantObject;
+
+/**
+ * A no-op implementation for the limit check which always passes all checks.
+ */
+public class NoopResourceLimitChecks implements ResourceLimitChecks {
+
+    @Override
+    public Future<?> isConnectionLimitExceeded(final TenantObject tenantObject) {
+        return Future.succeededFuture(tenantObject);
+    }
+}
--- a/service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
@@ -1,0 +1,173 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.hono.service.plan;
+
+import io.vertx.core.Future;
+import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.client.HttpResponse;
+import io.vertx.ext.web.client.WebClient;
+import io.vertx.ext.web.client.predicate.ResponsePredicate;
+import org.eclipse.hono.client.ClientErrorException;
+import org.eclipse.hono.service.metric.MicrometerBasedMetrics;
+import org.eclipse.hono.util.PortConfigurationHelper;
+import org.eclipse.hono.util.TenantObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.HttpURLConnection;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * An implementation of {@link ResourceLimitChecks} which uses metrics data from the prometheus backend to check if
+ * further connections or messages are allowed by comparing with the configured limits.
+ */
+public final class PrometheusBasedResourceLimitChecks implements ResourceLimitChecks {
+    private static final String CONNECTIONS_METRIC_NAME = MicrometerBasedMetrics.METER_CONNECTIONS_AUTHENTICATED
+            .replace(".", "_");
+    private final Logger log = LoggerFactory.getLogger(getClass());
+    private final WebClient client;
+    private String host;
+    private int port = 9090;
+
+    /**
+     * Creates a new PrometheusBasedResourceLimitChecks instance.
+     *
+     * @param vertx The Vertx instance to use.
+     * @throws NullPointerException if any of the parameters are {@code null}.
+     */
+    public PrometheusBasedResourceLimitChecks(final Vertx vertx) {
+        this.client = WebClient.create(Objects.requireNonNull(vertx));
+    }
+
+    @Override
+    public Future<?> isConnectionLimitExceeded(final TenantObject tenant) {
+        Objects.requireNonNull(tenant);
+        log.trace("Connections limit for tenant [{}] is [{}]", tenant.getTenantId(), tenant.getConnectionsLimit());
+        if (tenant.getConnectionsLimit() != -1) {
+            return queryForSumInPrometheus(CONNECTIONS_METRIC_NAME, tenant)
+                    .recover(failure -> Future.succeededFuture(0L))
+                    .compose(noOfConnections -> {
+                        if (Optional.ofNullable(noOfConnections).orElse(0L) < tenant.getConnectionsLimit()) {
+                            return Future.succeededFuture();
+                        } else {
+                            log.trace(
+                                    "Connections limit exceeded for tenant. [tenant: {}, no.-Connections:{}, connections-limit:{}]",
+                                    tenant.getTenantId(), noOfConnections, tenant.getConnectionsLimit());
+                            return Future
+                                    .failedFuture(new ClientErrorException(HttpURLConnection.HTTP_FORBIDDEN,
+                                            String.format("Connections limit exceeded for tenant: [%s]",
+                                                    tenant.getTenantId())));
+                        }
+                    });
+        }
+        return Future.succeededFuture(tenant);
+    }
+
+    /**
+     * Gets the host of the prometheus backend.
+     *
+     * @return host The host of the Prometheus backend.
+     */
+    public String getHost() {
+        return host;
+    }
+
+    /**
+     * Sets the host of the prometheus backend. The default value of this property is empty.
+     *
+     * @param host The host of the Prometheus backend.
+     */
+    public void setHost(final String host) {
+        this.host = Objects.requireNonNull(host);
+    }
+
+    /**
+     * Gets the port of the prometheus backend.
+     *
+     * @return port The port of the Prometheus backend.
+     */
+    public int getPort() {
+        return port;
+    }
+
+    /**
+     * Sets the port of the prometheus backend. The default value of this property is 9090.
+     *
+     * @param port The port of the Prometheus backend.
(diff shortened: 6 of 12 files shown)


sophokles73:
@kaniyan It looks like you are trying to enforce the limit *per adapter instance* based on the *local* value of the corresponding metric. However, the limit needs to be enforced across all adapter instances and regardless of protocol. I therefore do not see how this approach could work.

sophokles73:
@kaniyan maybe this approach would work for #931 ?

kaniyan:
@sophokles73 Thank you for the feedback. #931 corresponds to limiting the number of connections at adapters level and this approach should work there.

kaniyan:
@sophokles73 This PR is based on the mailing list discussion. At the moment, only mqtt adapter is equipped with this limit check. Can you pls take a look?

deploy/src/main/config/hono-adapter-mqtt-vertx-config.yml
```diff
@@ -43,3 +43,6 @@ hono:
     keyPath: ${secret.path}/mqtt-adapter-key.pem
     certPath: ${secret.path}/mqtt-adapter-cert.pem
     trustStorePath: ${secret.path}/trusted-certs.pem
+  plan:
+    prometheusBased:
+      queryURL: http://prometheus-operated.hono:9090/api/v1/query
```
sophokles73: if this is a property for a Prometheus specific implementation then FMPOV the only thing that needs to be defined is the Prometheus host name, isn't it? The relative URI of the query endpoint should not need to be configured, should it?

service-base/src/main/java/org/eclipse/hono/service/plan/TenantPlanValidator.java
```diff
@@ -0,0 +1,30 @@
...
+ *******************************************************************************/
+package org.eclipse.hono.service.plan;
+
+import io.vertx.core.Future;
+import org.eclipse.hono.util.TenantObject;
+
+/**
+ * Interface to check if further connections or messages are allowed based on the configured tenant limits.
+ */
+public abstract class TenantPlanValidator {
```
sophokles73: why isn't this an interface?

service-base/src/main/java/org/eclipse/hono/service/plan/TenantPlanValidator.java
```diff
@@ -0,0 +1,30 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2017 Contributors to the Eclipse Foundation
```
sophokles73: this file is new, right? so it should be 2019 only

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusSumQueryResultObject.java
```diff
@@ -0,0 +1,89 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
```
sophokles73: 2019

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedTenantPlanValidatorProperties.java
```diff
@@ -0,0 +1,41 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
```
sophokles73: 2019

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedTenantPlanValidator.java
```diff
@@ -0,0 +1,125 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
```
sophokles73: 2019

service-base/src/main/java/org/eclipse/hono/service/plan/TenantPlanValidator.java
sophokles73: IMHO the overall purpose of this interface should not be tied to the notion of a *tenant*. In fact, the limits that need to be checked could as well be scoped to the whole system. I would therefore prefer a more neutral name. What about `ResourceLimitChecks`?

service-base/src/main/java/org/eclipse/hono/service/plan/TenantPlanValidator.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+public abstract class TenantPlanValidator {
+
+    /**
+     * Check if further connections are allowed.
+     * 
+     * @param tenantObject The TenantObject that contains the tenant plans for a given tenant.
+     * @return A future indicating the outcome of the check.
+     */
+    public abstract Future<?> isFurtherConnectionsAllowed(TenantObject tenantObject);
```
sophokles73: what about *isConnectionLimitExceeded* instead?

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusSumQueryResultObject.java
```diff
@@ -0,0 +1,89 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Encapsulates the JSON response for the prometheus query found by the get operation of the Prometheus HTTP API.
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public final class PrometheusSumQueryResultObject {
```
sophokles73: I am not sure I understand why we need to bind the query result to a class. It looks as if the relevant data can be easily extracted from the result JSON and it doesn't look like the other data contained in the result is used anywhere, or is it?

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedTenantPlanValidator.java
```diff
@@ -0,0 +1,125 @@
...
+        }
+        return Future.succeededFuture();
+    }
+
+    private Future<Long> queryForSumInPrometheus(final String metric, final TenantObject tenant) {
+        final Future<Long> result = Future.future();
+        final String query = String.format("sum(%s{tenant=\"%s\"})", metric, tenant.getTenantId());
+        log.debug("Prometheus backend url: {}, query: {}", config.getQueryURL(), query);
+        client.getAbs(config.getQueryURL())
+                .addQueryParam("query", query)
```
sophokles73: you can make your life a little easier here by adding an expectation on the result status
```
.expect(ResponsePredicate.SC_OK)
```

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedTenantPlanValidator.java
```diff
@@ -0,0 +1,125 @@
...
+                .send(sendResult -> {
+                    if (sendResult.succeeded()) {
+                        final HttpResponse<Buffer> response = sendResult.result();
+                        if (response.statusCode() == HttpResponseStatus.OK.code()) {
+                            final PrometheusSumQueryResultObject queryResult = response
+                                    .bodyAsJson(PrometheusSumQueryResultObject.class);
```
sophokles73: I am pretty sure this will throw a runtime exception if the JSON cannot be mapped to the class

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedTenantPlanValidator.java
```diff
@@ -0,0 +1,125 @@
...
+    private Future<Long> queryForSumInPrometheus(final String metric, final TenantObject tenant) {
+        final Future<Long> result = Future.future();
+        final String query = String.format("sum(%s{tenant=\"%s\"})", metric, tenant.getTenantId());
+        log.debug("Prometheus backend url: {}, query: {}", config.getQueryURL(), query);
+        client.getAbs(config.getQueryURL())
+                .addQueryParam("query", query)
+                .send(sendResult -> {
+                    if (sendResult.succeeded()) {
+                        final HttpResponse<Buffer> response = sendResult.result();
+                        if (response.statusCode() == HttpResponseStatus.OK.code()) {
```
sophokles73: you can make your life a little easier here by adding an expectation to the client on the result status
```
client.expect(ResponsePredicate.SC_OK)
```

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedTenantPlanValidator.java
```diff
@@ -0,0 +1,125 @@
...
+     * @param config The configuration properties.
+     * @throws NullPointerException if props is {@code null}.
+     */
+    @Autowired
+    public void setConfig(final PrometheusBasedTenantPlanValidatorProperties config) {
+        Objects.requireNonNull(config);
+        this.config = config;
+    }
+
+    private Future<?> checkConnectionsLimit(final TenantObject tenant) {
```
sophokles73: IMHO this should actually be the implementation of *isFurtherConnectionsAllowed* ...

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -538,4 +538,14 @@ public static JsonObject newAdapterConfig(final String type, final boolean enabl
...
     }
+
+    /**
+     * Gets the concurrent connections limit for a tenant.
+     *
+     * @return The concurrent connections limit.
+     */
+    @JsonIgnore
+    public long getConnectionsLimit() {
+        return json.getLong(TenantConstants.CONNECTIONS_LIMIT, TenantConstants.DEFAULT_CONNECTIONS_LIMIT);
```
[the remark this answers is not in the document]
kaniyan: Done.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedTenantPlanValidator.java
kaniyan: @sophokles73 Good tip. Thank you.

service-base/src/main/java/org/eclipse/hono/service/plan/TenantPlanValidator.java
kaniyan: @sophokles73 In the issue it is mentioned as per _tenant_ and so the scope. Hence I have named it so. In case of system level scope more neutral name like `ResourceLimitChecks` sounds good.

deploy/src/main/config/hono-adapter-mqtt-vertx-config.yml
kaniyan: @sophokles73 For example it can also be deployed or mapped to a different port. In order to give more flexibility, I have made the entire url configurable.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusSumQueryResultObject.java
kaniyan: @sophokles73 The value from the query result is not quite simple to extract. That's why I bind the query result to a class. IMHO it improves the readability. As an alternative, I have changed the existing code to extract the value from the result JSON instead of binding. You can take a look once I pushed the changes.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedTenantPlanValidator.java
kaniyan: Done

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusSumQueryResultObject.java
kaniyan: Done

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedTenantPlanValidatorProperties.java
kaniyan: Done

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedTenantPlanValidator.java
kaniyan: Done

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedTenantPlanValidator.java
kaniyan: I have made changes to handle it

service-base/src/main/java/org/eclipse/hono/service/plan/TenantPlanValidator.java
kaniyan: Done

service-base/src/main/java/org/eclipse/hono/service/plan/ResourceLimitChecks.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Interface to check if further connections or messages are allowed based on the configured limits.
+ */
+public interface ResourceLimitChecks {
+
+    /**
+     * Check if the number of connections exceeded the configured limit.
+     * 
+     * @param tenantObject The TenantObject that contains the connections limit configured.
+     * @return A future indicating the outcome of the check.
```
sophokles73: IMHO we should explicitly state the type of exception that the returned future will be failed with so that clients can make proper use of it. In this case that would probably be a `ClientErrorException` with status code 403, right?

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -0,0 +1,137 @@
...
+    public PrometheusBasedResourceLimitChecks(final Vertx vertx) {
+        this.client = WebClient.create(Objects.requireNonNull(vertx));
+    }
+
+    @Override
+    public Future<?> isConnectionLimitExceeded(final TenantObject tenant) {
+        Objects.requireNonNull(tenant);
+        log.trace("Connections limit for tenant [{}] is [{}]", tenant.getTenantId(), tenant.getConnectionsLimit());
+        if (tenant.getConnectionsLimit() != -1) {
+            return queryForSumInPrometheus(MicrometerBasedMetrics.METER_CONNECTIONS_AUTHENTICATED.replace(".", "_"),
```
sophokles73: the metric name should be configurable and have a reasonable default value which is not computed on every invocation ...

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -0,0 +1,137 @@
...
+import org.springframework.stereotype.Component;
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * An implementation of {@link ResourceLimitChecks} which uses metrics data from the prometheus backend to check if
+ * further connections or messages are allowed by comparing with the configured limits.
+ */
+@Component
```
sophokles73: Is there a reason why you want to mark this as a component (but do not mark the constructor args as `@Autowired`)?
The configuration explicitly creates an instance using the constructor ...

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -487,6 +508,23 @@ protected void doStop(final Future<Void> stopFuture) {
         return result;
     }
 
+    /**
+     * Checks if any further connections for the given tenant are allowed.
+     *
+     * @param tenantConfig The tenant to check for.
+     * @return A succeeded future if further connections are allowed otherwise a failed future with failure message.
+     */
+    protected final Future<TenantObject> isFurtherConnectionsAllowed(final TenantObject tenantConfig) {
```
sophokles73: can we please rename this to the same name used in `ResourceLimitChecks`, i.e. *isConnectionLimitExceeded*?

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -316,6 +317,26 @@ public final CommandConnection getCommandConnection() {
         return this.commandConnection;
     }
 
+    /**
+     * Sets the ResourceLimitChecks for this service.
+     *
+     * @param resourceLimitChecks The ResourceLimitChecks instance
+     */
+    @Autowired
```
sophokles73: is this required? what if I do not want to enforce limits defined in a *plan*? We should probably use a default `ResourceLimitChecks` implementation that always returns `true` and then allow users to set another implementation using this method, if they want to ...

deploy/src/main/config/hono-adapter-mqtt-vertx-config.yml
sophokles73: are you going to change this? at least, do not *require* users to provide the resource path, i.e. the `/api/v1/query`, by e.g. checking if the provided URI contains a resource path and, if not, append this default.
Please also rename the property to *queryUrl* so that it complies with our method naming strategy.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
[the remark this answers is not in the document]
sophokles73: I think that for consistency reasons, you should not access the json field directly but use `TenantObject.getProperty(String, Object)` so that we can be sure to get the same behavior whenever we retrieve data ...

deploy/src/main/config/hono-adapter-mqtt-vertx-config.yml
kaniyan: @sophokles73 I have made it to support two properties namely _host_ and _port_.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
kaniyan: @sophokles73 Do you want it to be configurable by the user as Prometheus _host_ and _port_?

sophokles73: That would make sense, wouldn't it?

kaniyan: @sophokles73 The metrics names in hono are clearly defined as in the [hono-docs](https://www.eclipse.org/hono/api/metrics/). IMHO making the metric name configurable won't be that much useful. If any user decides to implement his/her own metrics (for the number of connections) with a different metric name instead of the standard one shipped with hono, then he/she could use this feature.

sophokles73: Can we at least agree that the name should not be re-computed on every invocation?

kaniyan: @sophokles73 That for sure 👍

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
kaniyan: @sophokles73 I see the point regarding enforcing limit checks. How about setting `@Autowired(required = false)`?  Before checking the limits, ensure that `ResourceLimitChecks` instance is available. If available, use it else return `Future.succeededFuture(...)`. May be it easier to have a look at the changes. I will push the changes.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
kaniyan: @sophokles73 `@Component` is not needed anymore. I will remove it.

kaniyan:
@sophokles73 The changes are there now.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecksProperties.java
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.hono.service.plan;
+
+/**
+ * A class for managing configuration properties for PrometheusBasedResourceLimitChecks.
+ *
+ */
+public class PrometheusBasedResourceLimitChecksProperties {
```
sophokles73: if these properties are specific to `PrometheusBasedResourceLimitChecks` then I do not see any reason why we should keep these properties in a separate class and not simply add them to `PrometheusBasedResourceLimitChecks` itself.
If these properties are generic and are relevant for all types of `ResourceLImitChecks` then this class should be named generically, wouldn't you agree?
Personally, I believe that the former is true, though ...

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -496,4 +496,15 @@ public static JsonObject newAdapterConfig(final String type, final boolean enabl
...
                 .put(TenantConstants.FIELD_ENABLED, enabled);
     }
+
+    /**
+     * Gets the concurrent connections limit for a tenant.
+     *
+     * @return The concurrent connections limit.
+     */
+    @JsonIgnore
+    public long getConnectionsLimit() {
```
sophokles73: in anticipation of the next step (#929) I think that we should make the underlying data structure more generic, i.e. provide for means to add more limits. What about something like:
```
{ "limits": {
      "max-connections": 15000
      }
}
```
which could then be extended later to:
```
{ "limits": {
      "max-connections": 15000,
      "max-bytes-per-month": 20_000_000
      }
}
```
WDYT?

kaniyan: 👍 I also have the same in mind for the next PR. I can start with this PR.

sophokles73: I think it still makes sense to have dedicated getters, e.g. *getConnectionLimit* and (later) *getDataVolumeLimit* or something like that ...

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecksProperties.java
kaniyan: @sophokles73 Do you mean directly into `PrometheusBasedResourceLimitChecks` as below?
```
    @Value(value = "${hono.plan.prometheus.host}")
    private String host;
    @Value(value = "${hono.plan.prometheus.port}")
    private String port;
```

sophokles73: No, by moving all the properties along with the getters and setters to `PrometheusBasedResourceLimitChecks` and then simply move the `@ConfigurationProperties(prefix = "hono.plan.prometheusBased")` annotation from the  *resourceLimitChecksProperties* method to the *resourceLimitChecks* method in `AbstractAdapterConfig` (and, of course, removing the former method from the configuration) ...

core/src/main/java/org/eclipse/hono/util/TenantObject.java
kaniyan: @sophokles73 Done. I have pushed the changes.

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
```diff
@@ -362,4 +364,15 @@ public ApplicationConfigProperties applicationConfigProperties() {
...
+
+    /**
+     * Creates a new instance of {@link ResourceLimitChecks} based on prometheus metrics data.
+     * 
+     * @return A ResourceLimitChecks instance.
+     */
+    @Bean
+    @ConfigurationProperties(prefix = "hono.plan.prometheusBased")
+    public ResourceLimitChecks resourceLimitChecks() {
+        return new PrometheusBasedResourceLimitChecks(vertx());
```
sophokles73: I guess this means that it is no longer possible to run Hono without Prometheus because we always inject the Prometheus based checks which will always fail if Prometheus is not running, right?

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -316,6 +317,26 @@ public final CommandConnection getCommandConnection() {
...
+    public final void setResourceLimitChecks(final ResourceLimitChecks resourceLimitChecks) {
+        this.resourceLimitChecks = Objects.requireNonNull(resourceLimitChecks);
+    }
+
+    /**
+     * Gets the ResourceLimitChecks instance used to check if the number of connections exceeded the limit or not.
+     *
+     * @return The ResourceLimitChecks instance.
+     */
+    protected ResourceLimitChecks getResourceLimitChecks() {
```
sophokles73: final

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -487,6 +508,25 @@ protected void doStop(final Future<Void> stopFuture) {
...
     }
 
+    /**
+     * Checks if the number of connections exceeded the configured limit for the given tenant or not.
+     *
+     * @param tenantConfig The config of the tenant to be checked.
+     * @return If the limit is not exceeded, then a succeeded future containing the tenant configuration is returned.
+     *         Else a failed future with a {@link ClientErrorException} containing <em>403 Forbidden</em> status.
+     */
+    protected final Future<TenantObject> isConnectionLimitExceeded(final TenantObject tenantConfig) {
```
sophokles73: I would rather have a default implementation being injected (in particular in the case of no Prometheus being available) instead of this method ...

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
kaniyan: @sophokles73 How about a similar approach that we use for the metrics? During maven build we provide an profile `metrics-prometheus` so that prometheus will be used to store the metrics. And then corresponding beans are created based on condition `@ConditionalOnClass(name = "io.micrometer.prometheus.PrometheusMeterRegistry")`.

In this case it looks like this. If no prometheus is being used, then this bean won't be created.
```
    @Bean
    @ConditionalOnClass(name = "io.micrometer.prometheus.PrometheusMeterRegistry")
    @ConfigurationProperties(prefix = "hono.plan.prometheusBased")
    public ResourceLimitChecks resourceLimitChecks() {
        return new PrometheusBasedResourceLimitChecks(vertx());
    }
```

sophokles73: the problem with this approach is that it would still require Prometheus running. The Maven profile is (so far) only used to add the Prometheus specific Meter registry to the class path. That alone makes no assumption whether Prometheus will actually be running. this is because Prometheus actively *scrapes* the metrics from components instead of having the components report metrics to Prometheus.
FMPOV it would be better to use a `ConditionalOnProperty` and check for the Prometheus host name being set. The default yaml should then (of course) not contain the corresponding properties.

kaniyan: @sophokles73 `ConditionalOnProperty` sounds good.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
kaniyan: @sophokles73 Do you mean a default implementation of the `ResourceLimitChecks`? If Prometheus host name is not set, then inject this default implementation. Anyway the protocol adapters will invoke the `isConnectionLimitExceeded(...)` in `AbstractProtocolAdapterBase` class. If the prometheus host name is set or not, accordingly the default or the `PrometheusBasedResourceLimitChecks` implementation will be used.

Currently if no prometheus host name is set, then no implementation is being injected. Then `ResourceLimitChecks` instance will be null. If it is null, then simply `true` is returned when adapters invoke `isConnectionLimitExceeded(...)`. I try to understand the advantages that the default implementation brings.

sophokles73: Using a default implementation will prevent us from having to add more and more methods with ugly null checks to `AbstractProtocolAdapterBase` when `ResourceLimitChecks` is extended with new methods. Instead, we only need the *getResourceLimitChecks()` method and return the instance.
By default, this would be the dummy implementation (which always passes all checks). The dummy implementation can then be overwritten by using the setter. This is the same approach we also use for the OpenTracing `Tracer` ...

kaniyan: @sophokles73 I pushed my changes. Can you pls take a look?

sophokles73: sorry, but I do not see any changes having been pushed since three days ...

kaniyan: @sophokles73 Sorry, there was a mistake. Now it should be there.

adapters/mqtt-vertx-base/src/test/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapterTest.java
```diff
@@ -1000,6 +1004,32 @@ public void testUnauthenticatedConnectionMetrics() {
...
+                .thenReturn(Future.succeededFuture(new DeviceUser("DEFAULT_TENANT", "4711")));
+        when(resourceLimitChecks.isConnectionLimitExceeded(any(TenantObject.class)))
+                .thenReturn(Future.failedFuture("Connections limit exceeded"));
+        final MqttEndpoint endpoint = getMqttEndpointAuthenticated();
+        adapter.handleEndpointConnection(endpoint);
+
+        // THEN the adapter has tried to authenticate the device
+        verify(authHandler).authenticateDevice(any(MqttContext.class));
+        // THEN the connection request is rejected
+        verify(endpoint).reject(MqttConnectReturnCode.CONNECTION_REFUSED_NOT_AUTHORIZED);
```
sophokles73: we now have two different outcomes for a similar problem:
1. if the maximum number of connections for the adapter instance is exceeded (as indicated by `getConnectionLimitManager().isLimitExceeded()`, we return a `MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE`
2. if the overall maximum number of connections for the tenant is exceeded (as indicated by `ResourceLimitChecks.isConnectionLimitExceeded()`, we return a `MqttConnectReturnCode.CONNECTION_REFUSED_NOT_AUTHORIZED`

that doesn't seem right.

Also, IMHO it is a little confusing and cumbersome to have two different places where connection limits need to be configured/enforced. Maybe we can merge the existing `ConnectionLimitManager` with the `ResourceLimitChecks` so that we have only one place for doing so? But we can also do that in a second step ...

kaniyan: @sophokles73 Absolutely. I also noticed that inside the `service-base` module with package as `limiting` containing `ConnectionLimitManager` and another named `plan`containing `ResourceLimitChecks` confusing. I agree that its better to merge it. I would suggest a separate PR for it.

sophokles73: can we then please use the same return code?

kaniyan: @sophokles73 Sure. IMHO `MqttConnectReturnCode.CONNECTION_REFUSED_NOT_AUTHORIZED` makes more sense for this scenario than the `MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE`.  WDYT?

sophokles73: Is there also a reason for your assessment?

kaniyan: @sophokles73 As per the spec, the description for `CONNECTION_REFUSED_SERVER_UNAVAILABLE` is _The Network Connection has been made but the MQTT service is unavailable_. FMPOV  the mqtt service is available but not authorised to connect as the limit is exceeded. so I find the `CONNECTION_REFUSED_NOT_AUTHORIZED` more suited.

sophokles73: @b-abel @dejanb @ctron 
what do you guys think about this? I tend to agree with @kaniyan that `CONNECTION_REFUSED_NOT_AUTHORIZED` is a better fit here.

b-abel: Isn't there a real difference between those two scenarios? One is about the tenant-specific quota and -- as @kaniyan points out -- a authorization issue. The other one is a technical issue, where the server is not able to process more load. So having this difference reflected in the return codes IMO makes sense. 
And there are different strategies possible for the client to react. In one case this could be monitored and eventually lead to an upgrade of the quota, in the other case a simple retry might be the right thing to do, because other instances of the protocol adapter will hopefully accept the connection.

sophokles73: > And there are different strategies possible for the client to react. In one case this could be monitored and eventually lead to an upgrade of the quota, in the other case a simple retry might be the right thing to do, because other instances of the protocol adapter will hopefully accept the connection.

Simply retrying will often also work in the former case (where the quota is reached) because other devices of the tenant may already have disconnected in the meantime ...

b-abel: > > And there are different strategies possible for the client to react. In one case this could be monitored and eventually lead to an upgrade of the quota, in the other case a simple retry might be the right thing to do, because other instances of the protocol adapter will hopefully accept the connection.
> 
> Simply retrying will often also work in the former case (where the quota is reached) because other devices of the tenant may already have disconnected in the meantime ...

That was an easy target, since "retry" is the generic solution in MQTT for nearly every problem :laughing: 
But seriously, for me there is one situation where, if it persists, the user of the service (tenant) needs to act and another one, that should be solved by the operator of the service, if it happens too often. This is one of very few points where the MQTT spec has something like return codes and I don't see why we should not use them.

ctron: I like the initial definition of:

> 1. if the maximum number of connections for the adapter instance is exceeded (as indicated by getConnectionLimitManager().isLimitExceeded(), we return a MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE
> 2. if the overall maximum number of connections for the tenant is exceeded (as indicated by ResourceLimitChecks.isConnectionLimitExceeded(), we return a MqttConnectReturnCode.CONNECTION_REFUSED_NOT_AUTHORIZED

dejanb: I tend to agree as well. The current definition is clearer from the tenant/operator points of view.

sophokles73: @ctron @dejanb 
My understanding is that both of you think that returning different codes for the two different cases is the right thing to do. Is that correct?

dejanb: Yes

sophokles73: Ok, then we'll have two return codes indicating the similar, but different, cases. We only need to make sure that we document this properly in the MQTT adapter user guide.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
sophokles73: IMHO we should simply remove this method and let subclasses use *getResourceLimitChecks* directly instead of introducing more and more of these *delegating* methods which do not add value.

service-base/src/main/java/org/eclipse/hono/service/plan/NoopResourceLimitChecks.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.hono.service.plan;
+
+import io.vertx.core.Future;
+import org.eclipse.hono.util.TenantObject;
+
+/**
+ * A no-op implementation for the limit check.
```
sophokles73: > ... which always passes all checks.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * An implementation of {@link ResourceLimitChecks} which uses metrics data from the prometheus backend to check if
+ * further connections or messages are allowed by comparing with the configured limits.
+ */
+public final class PrometheusBasedResourceLimitChecks implements ResourceLimitChecks {
+    private static final String CONNECTIONS_METRIC_NAME = MicrometerBasedMetrics.METER_CONNECTIONS_AUTHENTICATED
+            .replace(".", "_");
+    private final Logger log = LoggerFactory.getLogger(getClass());
+    private final WebClient client;
+    private String host = "prometheus-operated.hono";
```
sophokles73: IMHO this default value is too specific and we should better not define a default value at all. In fact, we made the usage of this implementation dependent on the existence of the corresponding environment variable so we can be sure that in production, it will always be set explicitly.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -0,0 +1,173 @@
...
+    public String getHost() {
+        return host;
+    }
+
+    /**
+     * Sets the host of the prometheus backend.
+     *
+     * @param host The host of the Prometheus backend.
+     */
+    public void setHost(final String host) {
```
sophokles73: please document the default value (*empty*)

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -0,0 +1,173 @@
...
+        return port;
+    }
+
+    /**
+     * Sets the port of the prometheus backend.
+     *
+     * @param port The port of the Prometheus backend.
+     * @throws IllegalArgumentException if the port number is &lt; 0 or &gt; 2^16 - 1
+     */
+    public void setPort(final int port) {
```
sophokles73: please document the default value (`9090`)

kaniyan: @sophokles73 The default values are documented and I pushed the changes.

PR #938, opened 2018-12-01
Add command and control example to AMQP adpater user guide.

Alfusainey:
@sophokles73: This patch adds an example to the adapter user guide to test command and control

### DIFF
--- a/cli/src/main/java/org/eclipse/hono/cli/AmqpCliClient.java
+++ b/cli/src/main/java/org/eclipse/hono/cli/AmqpCliClient.java
@@ -15,53 +15,29 @@ package org.eclipse.hono.cli;
 import java.io.PrintWriter;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.concurrent.CancellationException;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
 
-import javax.annotation.PostConstruct;
 
-import org.apache.qpid.proton.amqp.messaging.Rejected;
-import org.apache.qpid.proton.amqp.transport.DeliveryState;
-import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.config.ClientConfigProperties;
 import org.eclipse.hono.util.Strings;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.beans.factory.annotation.Value;
-import org.springframework.context.annotation.Profile;
-import org.springframework.stereotype.Component;
 
 import io.vertx.core.Future;
 import io.vertx.proton.ProtonClient;
 import io.vertx.proton.ProtonClientOptions;
 import io.vertx.proton.ProtonConnection;
-import io.vertx.proton.ProtonDelivery;
-import io.vertx.proton.ProtonHelper;
 import io.vertx.proton.ProtonSender;
 import io.vertx.proton.sasl.impl.ProtonSaslPlainImpl;
 
 /**
- * A simple command-line client for interacting with the AMQP adapter.
+ * Abstract base CLI client for interacting with the AMQP adapter.
  */
-@Component
-@Profile("amqp-adapter-cli")
-public class AmqpSend extends AbstractCliClient {
+public abstract class AmqpCliClient extends AbstractCliClient {
+
+    protected PrintWriter writer = new PrintWriter(System.out);
+
+    protected ProtonConnection adapterConnection;
 
     private ClientConfigProperties properties = new ClientConfigProperties();
-
-    /**
-     * The address to set on the message being sent.
-     */
-    @Value(value = "${message.address}")
-    private String messageAddress;
-
-    /**
-     * The payload to send.
-     */
-    @Value(value = "${message.payload}")
-    private String payload;
-
-    private ProtonConnection adapterConnection;
 
     /**
      * Sets the configuration properties to use for connecting
@@ -75,71 +51,17 @@ public class AmqpSend extends AbstractCliClient {
         this.properties = Objects.requireNonNull(props);
     }
 
-    @PostConstruct
-    void start() {
-
-        final CompletableFuture<ProtonDelivery> messageSent = new CompletableFuture<>();
-        final Message message = ProtonHelper.message(messageAddress, payload);
-
-        ctx.runOnContext(go -> {
-            connectToAdapter()
-            .compose(con -> {
-                adapterConnection = con;
-                final Future<ProtonSender> senderTracker = Future.future();
-                final ProtonSender sender = adapterConnection.createSender(null);
-                sender.openHandler(senderTracker);
-                sender.open();
-                return senderTracker;
-            })
-            .map(sender -> {
-                sender.send(message, delivery -> {
-                    adapterConnection.close();
-                    messageSent.complete(delivery);
-                });
-                return sender;
-            })
-            .otherwise(t -> {
-                messageSent.completeExceptionally(t);
-                return null;
-            });
-        });
-
-        final PrintWriter pw = new PrintWriter(System.out);
-
-        try {
-            final ProtonDelivery delivery = messageSent.join();
-            // Logs the delivery state to the console
-            pw.println();
-            printDelivery(delivery, pw);
-        } catch (CompletionException e) {
-            pw.println(e.getCause());
-            pw.flush();
-        } catch (CancellationException e) {
-            // do-nothing
-        }
-        System.exit(0);
-    }
-
-    private void printDelivery(final ProtonDelivery delivery, final PrintWriter pw) {
-
-        final DeliveryState state = delivery.getRemoteState();
-        pw.println(state.getType());
-        switch(state.getType()) {
-        case Rejected:
-            final Rejected rejected = (Rejected) state;
-            if (rejected.getError() != null) {
-                pw.println(rejected.getError().getCondition() + ": " + rejected.getError().getDescription());
-            }
-            break;
-        default:
-            break;
-        }
-        pw.flush();
+    protected Future<ProtonSender> createSender() {
+        final Future<ProtonSender> senderTracker = Future.future();
+        final ProtonSender sender = adapterConnection.createSender(null);
+        sender.openHandler(senderTracker);
+        sender.open();
+        return senderTracker;
     }
 
     // ----------------------------------< Vertx-proton >---
 
-    private Future<ProtonConnection> connectToAdapter() {
+    protected Future<ProtonConnection> connectToAdapter() {
 
         final Future<ProtonConnection> connectAttempt = Future.future();
         final ProtonClientOptions options = new ProtonClientOptions();
--- a/cli/src/main/java/org/eclipse/hono/cli/CommandAndControlClient.java
+++ b/cli/src/main/java/org/eclipse/hono/cli/CommandAndControlClient.java
@@ -1,0 +1,108 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.cli;
+
+import java.net.HttpURLConnection;
+
+import javax.annotation.PostConstruct;
+
+import org.apache.qpid.proton.amqp.messaging.Data;
+import org.apache.qpid.proton.message.Message;
+import org.eclipse.hono.util.CommandConstants;
+import org.eclipse.hono.util.MessageHelper;
+import org.springframework.context.annotation.Profile;
+import org.springframework.stereotype.Component;
+
+import io.vertx.core.Future;
+import io.vertx.proton.ProtonHelper;
+import io.vertx.proton.ProtonMessageHandler;
+import io.vertx.proton.ProtonQoS;
+import io.vertx.proton.ProtonReceiver;
+import io.vertx.proton.ProtonSender;
+
+/**
+ * A client for accessing Hono's Command and Control APIs to
+ * receive commands and sends command response messages via the AMQP adapter.
+ */
+@Component
+@Profile("amqp-command")
+public class CommandAndControlClient extends AmqpCliClient {
+
+    private ProtonSender sender;
+
+    @PostConstruct
+    void start() {
+        startCommandReceiver((d, m) -> {
+
+            String commandPayload = null;
+            if (m.getBody() instanceof Data) {
+                final byte[] body = (((Data) m.getBody()).getValue()).getArray();
+                commandPayload = new String(body);
+            }
+            writer.printf("Received Command Message : [Command name: %s, Command payload: %s] %n", m.getSubject(),
+                    commandPayload).flush();
+
+            final Message commandResponse = ProtonHelper.message(m.getReplyTo(), "OK: " + m.getSubject());
+            commandResponse.setCorrelationId(m.getCorrelationId());
+            MessageHelper.addProperty(commandResponse, MessageHelper.APP_PROPERTY_STATUS, HttpURLConnection.HTTP_OK);
+            commandResponse.setContentType(m.getContentType());
+            sender.send(commandResponse, delivery -> {
+                if (delivery.remotelySettled()) {
+                    writer.printf("Command response sent [outcome: %s] %n", delivery.getRemoteState());
+                } else {
+                    writer.println("Application did not settle command response message");
+                }
+                writer.flush();
+            });
+        })
+        .otherwise(t -> {
+            writer.printf("Fail to create command receiver link [reason: %s] %n", t.getMessage()).flush();
+            System.exit(1);
+            return null;
+        });
+    }
+
+    private Future<ProtonReceiver> startCommandReceiver(final ProtonMessageHandler msgHandler) {
+        return connectToAdapter()
+        .compose(con -> {
+            adapterConnection = con;
+            return createSender();
+        }).compose(s -> {
+            sender = s;
+            return subscribeToCommands(msgHandler);
+        });
+    }
+
+    /**
+     * Opens a receiver link to receive commands from the adapter.
+     *
+     * @param msgHandler The handler to invoke when a command message is received.
+     * @return A succeeded future with the created receiver link or a failed future
+     *         if the receiver link cannot be created.
+     *
+     */
+    private Future<ProtonReceiver> subscribeToCommands(final ProtonMessageHandler msgHandler) {
+        final Future<ProtonReceiver> result = Future.future();
+        final ProtonReceiver receiver = adapterConnection.createReceiver(CommandConstants.COMMAND_ENDPOINT);
+        receiver.setQoS(ProtonQoS.AT_LEAST_ONCE);
+        receiver.handler(msgHandler);
+        receiver.openHandler(result);
+        receiver.open();
+        return result.map(recver -> {
+            writer.println("Device is now ready to receive commands (Press Ctrl + c to terminate)");
+            writer.flush();
+            return recver;
+        });
+    }
+}
--- a/cli/src/main/java/org/eclipse/hono/cli/TelemetryAndEventCli.java
+++ b/cli/src/main/java/org/eclipse/hono/cli/TelemetryAndEventCli.java
@@ -1,0 +1,120 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.cli;
+
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+
+import javax.annotation.PostConstruct;
+
+import org.apache.qpid.proton.amqp.messaging.Rejected;
+import org.apache.qpid.proton.amqp.transport.DeliveryState;
+import org.apache.qpid.proton.message.Message;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Profile;
+import org.springframework.stereotype.Component;
+
+import io.vertx.proton.ProtonDelivery;
+import io.vertx.proton.ProtonHelper;
+
+/**
+ * A Client for accessing Hono's Telemetry and Event APIs to send
+ * telemetry data respectively events via the AMQP adapter.
+ */
+@Component
+@Profile("amqp-send")
+public class TelemetryAndEventCli extends AmqpCliClient {
+
+    /**
+     * The address to set on the message being sent.
+     */
+    @Value(value = "${message.address}")
+    private String messageAddress;
+
+    /**
+     * The payload to send.
+     */
+    @Value(value = "${message.payload}")
+    private String payload;
+
+    @PostConstruct
+    void start() {
+
+        final CompletableFuture<ProtonDelivery> messageSent = new CompletableFuture<>();
+
+        sendMessage(messageSent);
+
+        try {
+
+            final ProtonDelivery delivery = messageSent.join();
+            // Logs the delivery state to the console
+            printDelivery(delivery);
+            System.exit(0);
+        } catch (CompletionException e) {
+            writer.printf("Sending message failed [reason: %s] %n", e.getCause());
+            writer.flush();
+        } catch (CancellationException e) {
+            // do-nothing
+        }
+        System.exit(1);
+    }
+
+    /**
+     * Connects to the AMQP adapter and send a telemetry/event message to the adapter.
+     *
+     * @param messageTracker The future to notify when the message is sent. The future is completed with the delivery
+     *            upon success or completed with an exception.
+     */
+    private void sendMessage(final CompletableFuture<ProtonDelivery> messageTracker) {
+
+        ctx.runOnContext(go -> {
+            connectToAdapter()
+            .compose(con -> {
+                adapterConnection = con;
+                return createSender();
+            })
+            .map(sender -> {
+                final Message message = ProtonHelper.message(messageAddress, payload);
+                sender.send(message, delivery -> {
+                    adapterConnection.close();
+                    messageTracker.complete(delivery);
+                });
+                return sender;
+            })
+            .otherwise(t -> {
+                messageTracker.completeExceptionally(t);
+                return null;
+            });
+        });
+     }
+
+    private void printDelivery(final ProtonDelivery delivery) {
+        final DeliveryState state = delivery.getRemoteState();
+        writer.printf("[Delivery State: %s] %n", state.getType()).flush();
+        switch(state.getType()) {
+        case Rejected:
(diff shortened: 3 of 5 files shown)


cli/src/main/java/org/eclipse/hono/cli/AmqpSend.java
```diff
@@ -89,7 +101,19 @@ void start() {
...
+                    final Message commandResponse = ProtonHelper.message(m.getReplyTo(), "OK: " + m.getSubject());
+                    commandResponse.setCorrelationId(m.getCorrelationId());
+                    MessageHelper.addProperty(commandResponse, MessageHelper.APP_PROPERTY_STATUS, HttpURLConnection.HTTP_OK);
+                    commandResponse.setContentType(m.getContentType());
+                    sender.send(commandResponse, delivery -> messageSent.complete(delivery));
+                });
+            } else {
+                final Message message = ProtonHelper.message(messageAddress, payload);
+                sender.send(message, delivery -> messageSent.complete(delivery));
+            }
```
sophokles73: I think it would improve readability if we could factor out this whole block into a separate method that accepts a `ProtonSender`, checks the *subscribe* flag, creates the command consumer and returns a `Future<ProtonSender>`, don't you think? That way we could also add decent error handling for the case that creating the consumer doesn't work ...

site/content/user-guide/amqp-adapter.md
```diff
@@ -63,17 +63,17 @@ The command-line client supports the following parameters (with default values):
...
 * `--hono.client.host`: The hostname that the AMQP adapter is running on (default: `localhost`)
-* `--hono.client.port`: The port that the adapter is listening for incoming connections (default: `5672`)
+* `--hono.client.port`: The port that the adapter is listening for incoming connections (default: `4040`)
 * `--hono.client.username`: The username to use for authenticating to the adpter (default: `sensor1@DEFAULT_TENANT`)
 * `--hono.client.password`: The password to use for authenticating to the adapter (default: `hono-secret`)
 
 To run the client using the above default values, open a terminal and execute the following:
 
-    $ /hono/cli/target$ java -jar hono-cli-0.8-SNAPSHOT-exec.jar --spring.profiles.active=amqp-adapter-cli
+    $ /hono/cli/target$ java -jar hono-cli-0.9-M1-SNAPSHOT-exec.jar --spring.profiles.active=amqp-adapter-cli
```
sophokles73: what about this:
```
/hono/cli$ java -jar target/hono-cli-*-exec.jar --spring.profiles.active=amqp-adapter-cli
```
That way we wouldn't need to update the text with every new release ...

site/content/user-guide/amqp-adapter.md
```diff
@@ -275,6 +275,35 @@ The table below provides an overview of the properties that must be set on a com
...
+The AMQP Command-line client can be used to receive and send a command response message from/to a consumer application. The client is used to simulate a device to receive commands and for sending command responses back to the application. The command consumer client is used to simulate an application sending a command to the device and receiving a command response from the device.
+
+Start the AMQP Command-line client, as follows:
+
+    $ /hono/cli/target$ java -jar hono-cli-0.9-M1-SNAPSHOT-exec.jar --spring.profiles.active=amqp-adapter-cli --subscribe=true
+    $ 04:13:49.651 [vert.x-eventloop-thread-0] INFO  org.eclipse.hono.cli.AmqpSend - Device is now ready to receive commands
+
+After successfully starting the client, a message indicating that the device is ready to receive commands will be printed to standard output. The device is now waiting to receive commands from applications. To send a command to the device, open a new terminal shell and start the command consumer, as shown below:
+
+    $ /hono/cli/target$ mvn spring-boot:run -Drun.arguments=--hono.client.host=localhost,--hono.client.username=consumer@HONO,--hono.client.password=verysecret -Drun.profiles=command,ssl
```
sophokles73: FMPOV it is very confusing that we start the client using `java -jar` in the first step and then start the client using `mvn spring-boot:run` in the second step. I would actually favor using `java -jar` everywhere ...

site/content/user-guide/amqp-adapter.md
```diff
@@ -275,6 +275,35 @@ The table below provides an overview of the properties that must be set on a com
...
+
+Once the command consumer starts successfully, enter a command name, payload and content type of the command to send to the device.
+
+After sending the command, the device (i.e AMQP Command-line client) will printout the command message received from the consumer application and send back the response message. The command consumer prints the payload of the command response message to standard output, as shown below:
+
+    $ 04:15:40.210 [vert.x-eventloop-thread-0] INFO  org.eclipse.hono.cli.Commander - Received Command response : OK: setBrightness
+
+Finally, the device shows that the command response message has been received by the application by printing out the disposition to standard output:
+
+    $ Accepted{}
```
sophokles73: the `$` character is a suffix to the current working directory path in the shell, i.e. if I am in directory `/home/hono` then the shell usually prints `/home/hono$` before the prompt.
If you assume that the user is working in directory `/hono/cli/target` (as you do above), then the shell will print
```
/hono/cli/target$ 
```
but not 
```
$ /hono/cli/target$
```
can you please fix this accordingly in the example?

site/content/user-guide/amqp-adapter.md
Alfusainey: you mean starting the command consumer using `java -jar`? using the `mvn spring-boot:run` approach, for starting the command consumer, is how it is documented on the hono website

cli/src/main/java/org/eclipse/hono/cli/AmqpSend.java
Alfusainey: yes, that make sense!

site/content/user-guide/amqp-adapter.md
sophokles73: both ways work, no doubt. I just propose to start the consumer consistently using the same approach throughout the example and my personal preference would be to use the `java -jar` form because it is more compact and intuitive.

Alfusainey: ok 👍

Alfusainey:
@sophokles73 patch is updated..

cli/src/main/java/org/eclipse/hono/cli/AmqpSend.java
```diff
@@ -116,6 +126,49 @@ void start() {
...
 
+    private Future<ProtonSender> sendMessage(final ProtonSender sender, final Consumer<ProtonDelivery> consumer) {
+        final Future<ProtonSender> result = Future.future();
+        if (!subscribe) {
+            final Message message = ProtonHelper.message(messageAddress, payload);
+            sender.send(message, delivery -> consumer.accept(delivery));
+            result.complete(sender);
+            return result;
+        }
+        subscribeToCommands((delivery, message) -> {
```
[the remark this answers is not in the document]
Alfusainey: the idea is to have the client automatically respond to command request messages it receives (when the subscribe flag is set). but yes, you are right.. let me think about how to refactor this properly

Alfusainey:
@sophokles73 how about this refactoring?

sophokles73:
@Alfusainey I am in a two-day workshop and won't find the time to review until Thursday ...

Alfusainey:
sure, enjoy the workshop..

Alfusainey:
@sophokles73: any news on this :-)

sophokles73:
@Alfusainey sorry, totally forgot about it. I'll take a look today ...

site/content/user-guide/amqp-adapter.md
```diff
@@ -258,6 +258,39 @@ The table below provides an overview of the properties that must be set on a com
...
+### Examples
+
+The AMQP Command-line client can be used to receive and send a command response message from/to a consumer application. The client is used to simulate a device to receive commands and for sending command responses back to the application. The command consumer client is used to simulate an application sending commands to devices and receiving command responses from devices.
+
+Start the AMQP Command-line client, as follows:
+
+    /hono/cli/target$ java -jar hono-cli-*-exec.jar --spring.profiles.active=amqp-adapter-cli --subscribe=true
+    $ 04:13:49.651 [vert.x-eventloop-thread-0] INFO  org.eclipse.hono.cli.AmqpSend - Device is now ready to receive commands
+
+After successfully starting the client, a message indicating that the device is ready to receive commands will be printed to standard output. The device is now waiting to receive commands from applications. To send a command to the device, open a new terminal shell and start the command consumer, as shown below:
```
sophokles73: > open a new terminal shell and start the command consumer, as shown below:

command consumer? this seems to be the client for *sending* the command, isn't it?

site/content/user-guide/amqp-adapter.md
```diff
@@ -258,6 +258,39 @@ The table below provides an overview of the properties that must be set on a com
...
+	>>>>>>>>> Enter command payload:
+	some-payload
+	>>>>>>>>> Enter content type:
+	text/plain
+
+{{% note %}}
+Change into the `cli` directory before running the command above to start the consumer. If you change into the target directory (i.e `cli/target`), then the client will not be able to locate to certificate needed to connect to the messaging network.
+{{% /note %}}
+
+Once the command consumer starts successfully, enter a command name, payload and content-type of the command to send to the device.
```
sophokles73: consumer?

site/content/user-guide/amqp-adapter.md
```diff
@@ -258,6 +258,39 @@ The table below provides an overview of the properties that must be set on a com
...
+
+After sending the command, the device (i.e AMQP Command-line client) will printout the command message received from the consumer application and send back the response message. The command consumer prints the payload of the command response message to standard output, as shown below:
```
sophokles73: print out
consumer application?

site/content/user-guide/amqp-adapter.md
```diff
@@ -65,12 +65,12 @@ The command-line client supports the following parameters (with default values):
 * `--message.payload`: The message payload body (default: `'{"temp": 5}'`)
 * `--hono.client.host`: The host name that the AMQP adapter is running on (default: `localhost`)
 * `--hono.client.port`: The port that the adapter is listening for incoming connections (default: `5672`)
+* `--hono.client.username`: The username to use for authenticating to the adpter (default: `sensor1@DEFAULT_TENANT`)
```
sophokles73: please remove these, we do not use default values for credentials

site/content/user-guide/amqp-adapter.md
Alfusainey: oops, that was a mistake..

Alfusainey:
> IMHO we should use separate classes (and separate Spring Boot profiles) for the two use cases sending a command vs. receiving and replying. it is very confusing to have the code for both in the same class called AmqpSend ...

indeed 👍

site/content/user-guide/amqp-adapter.md
Alfusainey: but we configured a default value for `username` and `password` in application.yml. can you please clarify?

sophokles73: we have removed them from the yaml file ...

Alfusainey: ok 👍

Alfusainey:
@sophokles73 i updated the PR

cli/src/main/java/org/eclipse/hono/cli/CommandAndControlClient.java
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Connects to the AMQP adapter and open a command receiver link for receiving commands from the adapter. Once a
+     * command request message is received, this method automatically sends a command response back to the adapter.
+     *
+     * @param messageTracker The future to notify when the message is sent. The future is completed with the delivery
+     *            upon success or complete the future with an exception.
+     */
+    @Override
+    protected void sendMessage(final CompletableFuture<ProtonDelivery> messageTracker) {
```
sophokles73: why is this method called *sendMessage* when it is about creating a command consumer?

cli/src/main/java/org/eclipse/hono/cli/CommandAndControlClient.java
```diff
@@ -0,0 +1,87 @@
...
+        .compose(con -> {
+            adapterConnection = con;
+            return createSender();
+        }).map(sender -> subscribeToCommands((d, m) -> {
+            LOG.info("Received Command Message : {}", m.getSubject());
+            final Message commandResponse = ProtonHelper.message(m.getReplyTo(), "OK: " + m.getSubject());
+            commandResponse.setCorrelationId(m.getCorrelationId());
+            MessageHelper.addProperty(commandResponse, MessageHelper.APP_PROPERTY_STATUS, HttpURLConnection.HTTP_OK);
+            commandResponse.setContentType(m.getContentType());
+            sender.send(commandResponse, delivery -> messageTracker.complete(delivery));
```
sophokles73: I do not see any relation between sending a command response and completing this future ...
The command consumer is supposed to run until the user hits ctrl-c and should reply to all commands that arrive, right?

cli/src/main/java/org/eclipse/hono/cli/AmqpCliClient.java
```diff
@@ -23,45 +23,41 @@
...
-
-    private ClientConfigProperties properties = new ClientConfigProperties();
+public abstract class AmqpCliClient extends AbstractCliClient {
 
     /**
      * The address to set on the message being sent.
      */
     @Value(value = "${message.address}")
-    private String messageAddress;
+    protected String messageAddress;
```
sophokles73: FMPOV this doesn't belong here but rather in `TelemetryAndEventCli`

cli/src/main/java/org/eclipse/hono/cli/AmqpCliClient.java
```diff
@@ -23,45 +23,41 @@
...
 
     /**
      * The payload to send.
      */
     @Value(value = "${message.payload}")
-    private String payload;
+    protected String payload;
```
sophokles73: FMPOV this doesn't belong here but rather in `TelemetryAndEventCli`

cli/src/main/java/org/eclipse/hono/cli/AmqpCliClient.java
```diff
@@ -23,45 +23,41 @@
...
+
+    protected ProtonConnection adapterConnection;
+
+    private ClientConfigProperties properties = new ClientConfigProperties();
 
-    private ProtonConnection adapterConnection;
+    protected abstract void sendMessage(CompletableFuture<ProtonDelivery> messageTracker);
```
sophokles73: I do not see why a command consumer should implement this method ...

cli/src/main/java/org/eclipse/hono/cli/CommandAndControlClient.java
Alfusainey: > The command consumer is supposed to run until the user hits ctrl-c and should reply to all commands that arrive, right?

what i had in mind was for the user to start a command consumer to receive -> send command response -> then exit

i will change this so the consumer runs until explicitly closed

sophokles73: > i will change this so the consumer runs until explicitly closed

don't you think that this is more practical?

Alfusainey: yes, way more practical and intuitive :-)

Alfusainey:
@sophokles73 PR updated!

cli/src/main/java/org/eclipse/hono/cli/CommandAndControlClient.java
```diff
@@ -0,0 +1,111 @@
...
+        });
+    }
+
+    /**
+     * Connects to the AMQP adapter and open a command receiver link for receiving commands from the adapter. Once a
+     * command request message is received, this method automatically sends a command response back to the adapter.
+     *
+     * @param messageTracker The future to notify when the message is sent. The future is completed with the delivery
+     *            upon success or complete the future with an exception.
+     */
```
sophokles73: this JavaDoc comment does not belong to this method, does it?

site/content/user-guide/amqp-adapter.md
```diff
@@ -56,21 +56,20 @@ For a client side error, the adapter settles the message transfer with the *reje
...
 * `--hono.client.host`: The host name that the AMQP adapter is running on (default: `localhost`)
 * `--hono.client.port`: The port that the adapter is listening for incoming connections (default: `5672`)
 
-To run the client using the above default values, open a terminal and execute the following:
+To run the client to send telemetry message to Hono, open a terminal and execute the following:
 
-    $ /hono/cli/target$ java -jar hono-cli-*-exec.jar --spring.profiles.active=amqp-adapter-cli --hono.client.username=sensor1@DEFAULT_TENANT --hono.client.password=hono-secret
-    
-    Accepted{}
+    $ /hono/cli/target$ java -jar hono-cli-*-exec.jar --spring.profiles.active=amqp-send
```
sophokles73: this will not work because it lacks the credentials

cli/src/main/resources/application.yml
```diff
@@ -48,11 +48,13 @@ hono:
 ---
 
 spring:
-  profiles: amqp-adapter-cli
+  profiles: amqp-send,amqp-command
 
 hono:
   client:
     port: 5672
+    username: sensor1@DEFAULT_TENANT
```
sophokles73: remove the credentials

site/content/user-guide/amqp-adapter.md
```diff
@@ -262,6 +261,36 @@ The table below provides an overview of the properties that must be set on a com
 | *correlation-id* | yes             | *properties*             | *string*     | MUST contain the value of the *correlation-id* property of the command request message. |
 | *status*         | yes             | *application-properties* | *integer*    | MUST contain a status code indicating the outcome of processing the command at the device (see [Command & Control API]({{< ref "/api/Command-And-Control-API.md" >}}) for details). |
 
+### Examples
+
+Using the `amqp-command` spring profile, the AMQP Command-line client can be used to receive and send a command response message from/to an application. The client is used to simulate a device to receive commands and for sending command responses back to the application. The command CLI client is used to simulate an application sending commands to devices and receiving command responses from devices.
```
sophokles73: > The AMQP adapter client can be used to simulate a device which receives commands and sends responses back to the application. The command line client is used to simulate an application sending commands to devices and receiving command responses from devices.

site/content/user-guide/amqp-adapter.md
```diff
@@ -262,6 +261,36 @@ The table below provides an overview of the properties that must be set on a com
...
+
+Start the AMQP Command client, as follows:
```
sophokles73: > Start the AMQP adapter client, as follows:

cli/src/main/java/org/eclipse/hono/cli/TelemetryAndEventCli.java
```diff
@@ -0,0 +1,116 @@
...
+     }
+
+    private void printDelivery(final ProtonDelivery delivery) {
+        final DeliveryState state = delivery.getRemoteState();
+        LOG.info("[Delivery State: {}]", state.getType());
+        switch(state.getType()) {
+        case Rejected:
+            final Rejected rejected = (Rejected) state;
+            if (rejected.getError() != null) {
+                LOG.info("[Error Condition: {}, Error Description: {}]", rejected.getError().getCondition(), rejected.getError().getDescription());
```
sophokles73: please undo all the changes you made regarding the replacement of the System out PrintWriter with Logger. We deliberately chose to print to stdout because this is a command line client and its output is expected to show up on stdout and stderr.

Alfusainey: Sure! it was SonarLint that was complaining so i decided to use a logger for it to be happy :-) i will change it back..

sophokles73: tools are now always smarter than we are, I guess ;-)

site/content/user-guide/amqp-adapter.md
Alfusainey: > The command line client is

dont't you think this could be confused with the `terminal` itself? how about the `command CLI` ?

Alfusainey:
@sophokles73 i updated the PR and squashed the commits into one..

cli/src/main/java/org/eclipse/hono/cli/CommandAndControlClient.java
```diff
@@ -0,0 +1,112 @@
...
+    private Future<ProtonReceiver> subscribeToCommands(final ProtonMessageHandler msgHandler) {
+        final Future<ProtonReceiver> result = Future.future();
+        final ProtonReceiver receiver = adapterConnection.createReceiver(CommandConstants.COMMAND_ENDPOINT);
+        receiver.setQoS(ProtonQoS.AT_LEAST_ONCE);
+        receiver.handler(msgHandler);
+        receiver.openHandler(result);
+        receiver.open();
+        return result.map(recver -> {
+            LOG.info("Device is now ready to receive commands (Press Ctrl + c to terminate)");
```
sophokles73: this should write to stdout ...

site/content/user-guide/amqp-adapter.md
```diff
@@ -262,6 +261,36 @@ The table below provides an overview of the properties that must be set on a com
...
 | *status*         | yes             | *application-properties* | *integer*    | MUST contain a status code indicating the outcome of processing the command at the device (see [Command & Control API]({{< ref "/api/Command-And-Control-API.md" >}}) for details). |
 
+### Examples
+
+The AMQP adapter client can be used to simulate a device which receives commands and sends responses back to the application. The command line client is used to simulate an application sending commands to devices and receiving command responses from devices.
+
+Start the AMQP adapter client, as follows:
+
+    /hono/cli/target$ java -jar hono-cli-*-exec.jar --spring.profiles.active=amqp-command --hono.client.username=sensor1@DEFAULT_TENANT --hono.client.password=hono-secret
+    $ 04:13:49.651 [vert.x-eventloop-thread-0] INFO  o.e.hono.cli.CommandAndControlClient - Device is now ready to receive commands (Press Ctrl + c to terminate)
```
sophokles73: IMHO this should be adapted to the real output, i.e. not a log statement anymore

Alfusainey: yes, but i i used a log statement in this PR. i will update it if i remove the log statement(as you mention previously)

cli/src/main/java/org/eclipse/hono/cli/CommandAndControlClient.java
Alfusainey: ok

cli/src/main/java/org/eclipse/hono/cli/CommandAndControlClient.java
```diff
@@ -0,0 +1,112 @@
...
+        })
+        .otherwise(t -> {
+            writer.printf("Fail to create command receiver link [reason: %s] %n", t.getMessage()).flush();
+            return null;
+        });
+    }
+
+    /**
+     * Connects to the AMQP adapter and open a command receiver link for receiving commands from the adapter. Once a
+     * command request message is received, this method automatically sends a command response back to the adapter.
```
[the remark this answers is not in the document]
Alfusainey: > it doesn't, it hands over the received message to the given msgHandler. This method itself has no clue about what the handler does, does it?

👍 

thanks, will remove the javadoc..

cli/src/main/java/org/eclipse/hono/cli/CommandAndControlClient.java
```diff
@@ -0,0 +1,112 @@
...
+                } else {
+                    writer.println("Application did not settle command response message");
+                }
+                writer.flush();
+            });
+        })
+        .otherwise(t -> {
+            writer.printf("Fail to create command receiver link [reason: %s] %n", t.getMessage()).flush();
+            return null;
+        });
```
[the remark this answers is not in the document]
Alfusainey: yes i did test it many many times and works as expected. it does not exit and continuously receives and respond to command messages until explicitly closed using ctrl + c.

however, it does not exit the JVM if the receiver cannot be created (e.g when authentication fails)

cli/src/main/java/org/eclipse/hono/cli/TelemetryAndEventCli.java
```diff
@@ -0,0 +1,117 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
```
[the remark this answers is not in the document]
Alfusainey: i think this should also just be 2019 (its a new file)

cli/src/main/java/org/eclipse/hono/cli/CommandAndControlClient.java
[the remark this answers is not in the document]
Alfusainey: what do you think: terminate the program if the receiver link cannot be created?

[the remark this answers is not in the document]
sophokles73: I would think so. What good is a stale client that doesn't do anything?

cli/src/main/java/org/eclipse/hono/cli/TelemetryAndEventCli.java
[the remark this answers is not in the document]
sophokles73: if it's new, then it should be just the current year

cli/src/main/java/org/eclipse/hono/cli/CommandAndControlClient.java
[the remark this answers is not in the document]
Alfusainey: then what do you think of using `System.exit(1)` to close the client whenever a failure is encountered?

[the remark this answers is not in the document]
sophokles73: makes sense to me

Alfusainey:
@sophokles73 i just pushed an extra commit

Alfusainey:
@sophokles73 just noticed there was a conflict with `amqp-adapter.md` file.. so i fixed and squashed the commits. can you take a look now?

site/content/user-guide/amqp-adapter.md
```diff
@@ -116,13 +117,13 @@ When a device publishes data to the `telemetry` address, the AMQP adapter automa
...
+    java -jar hono-cli-*-exec.jar --spring.profiles.active=amqp-send --hono.client.username=sensor1@DEFAULT_TENANT --hono.client.password=hono-secret
 
 Notice that we only supplied a new value for the message address, leaving the other default values.
 
 Publish some JSON data for device `4711` using a client certificate for authentication:
 
-    ~/hono/cli/target$ java -jar hono-cli-*-exec.jar --spring.profiles.active=amqp-adapter-cli --hono.client.port=5671 --hono.client.certPath=config/hono-demo-certs-jar/device-4711-cert.pem --hono.client.keyPath=config/hono-demo-certs-jar/device-4711-key.pem --hono.client.trustStorePath=config/hono-demo-certs-jar/trusted-certs.pem --hono.client.hostnameVerificationRequired=false
+    ~/hono/cli/target$ java -jar hono-cli-*-exec.jar --spring.profiles.active=amqp-send --hono.client.port=5671 --hono.client.certPath=config/hono-demo-certs-jar/device-4711-cert.pem --hono.client.keyPath=config/hono-demo-certs-jar/device-4711-key.pem --hono.client.trustStorePath=config/hono-demo-certs-jar/trusted-certs.pem --hono.client.hostnameVerificationRequired=false
```
sophokles73: please adapt to the new style we use for command line examples, i.e. add a comment above indicating the expected working directory and then only use the plain command:
```
java -jar hono-cli-*-exec.jar --spring.profiles.active=amqp-send --hono.client.port=5671 --hono.client.certPath=config/hono-demo-certs-jar/device-4711-cert.pem --hono.client.keyPath=config/hono-demo-certs-jar/device-4711-key.pem --hono.client.trustStorePath=config/hono-demo-certs-jar/trusted-certs.pem --hono.client.hostnameVerificationRequired=false
```

Alfusainey: OK. 
actually, it wasn't changed in the commit that introduced `comment for working directories` (so it was not affected by the conflict)

Alfusainey:
@sophokles73 done!

PR #939, opened 2018-12-02
trivial fixes/improvements

Alfusainey:
* In some tests, the arguments to `assertEqual(expected, actual)` are swapped, resulting in `assertEqual(actual, expected)`
* `AbstractServiceBase` defines a logger instance to be shared by all subclasses. child classes (esp. protocol adapters) can simply use that instance.
* `receiversPerConnection` and `activeSenders` instances can never be null
* ...


sophokles73:
@ctron are you going to merge this PR, now that you have approved it?

PR #942, opened 2018-12-05
Added metrics for AMQP adapter

kaniyan:


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -354,7 +369,7 @@ private void handleSessionOpen(final ProtonConnection conn, final ProtonSession
      * @param con The connection to close.
      * @param res The client's close frame.
      */
-    private void handleRemoteConnectionClose(final ProtonConnection con, final AsyncResult<ProtonConnection> res) {
+    private void handleRemoteConnectionClose(final ProtonConnection con, final AsyncResult<ProtonConnection> res, final Device device) {
```
sophokles73: there doesn't seem to be a need for the device param

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -674,6 +689,7 @@ protected void onCommandReceived(final ProtonSender sender, final CommandContext
         });
         final Map<String, Object> items = new HashMap<>(4);
         items.put(Fields.EVENT, "command sent to device");
+        metrics.incrementCommandDeliveredToDevice(command.getTenant());
```
sophokles73: IMHO this needs to be moved to the block handling the the *accepted* outcome.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -734,15 +750,27 @@ protected void onCommandReceived(final ProtonSender sender, final CommandContext
         })
         .map(downstreamDelivery -> {
             context.accept();
+            metrics.incrementProcessedMessages(context.getEndpoint(), context.getTenantId());
```
sophokles73: hmm, we also have the *hono.commands.response.delivered* metric which is supposed to track the number of responses being uploaded. However, it seems that we could as well simply use the *hono.messages.processed* metric for that purpose, we could then also record the response payload in *hono.messages.processed.payload* and get rid of *hono.commands.response.delivered* altogether.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
kaniyan: I will remove it.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
kaniyan: It sounds good to use the _hono.messages.processed.payload_ metric with type as  _command_ for the command response. This is already available and so I will remove the ` metrics.incrementProcessedMessages(context.getEndpoint(), context.getTenantId());`

kaniyan: It sounds good to use the _hono.messages.processed_ metrics for command responses too. I will remove the _hono.commands.response.delivered_ part.

kaniyan: :-) I will send a proposal to the mailing list.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
kaniyan: The _accepted_ outcome is validated with _if/else_ blocks as below.

 `if (Accepted.class.isInstance(remoteState)) {
...
}`. 

I understood that the _incrementCommandDeliveredToDevice_ part should be moved there in consideration with the outcome. Is my understanding right? If not can you please explain where exactly do you mean?

sophokles73:
obsolete with 13d21b34316e9b11e6fca3173d4e1ce393f2ebfa

PR #946, opened 2018-12-10
Deploy the OLM in a separate script

ctron:


ctron:
@sophokles73 do you have time at the moment to do a proper test? That would be great.

sophokles73:
@ctron sorry for the delay, I am quite busy with other PRs at the moment ...

deploy/src/main/deploy/kubernetes/olm_deploy.sh
```diff
@@ -0,0 +1,47 @@
...
+# CRD fails, and so a catalog source is missing. The next steps wait "a bit"
+# and then re-register the catalog source. Which again might fail, because
+# the previous step might have been successful.
+#
+
+sleep 5 # delay and retry CRD
+
+kubectl create -f operator-lifecycle-manager/deploy/upstream/manifests/latest/*rh-operators.catalogsource.yaml || true
+
+echo ... done
```
sophokles73: what about removing the local olm workspace cloned from GitHub? Shouldn't we remove it as well?

ctron: I think it is worth keeping the content. As re-downloading is might be an issue for someone. However I did add a short message in the "undeploy" script:

~~~
➜  kubernetes git:(feature/olm_fixup_1) ✗ ./olm_undeploy.sh 
UNDEPLOYING OPERATOR LIFECYCLE MANAGER FROM KUBERNETES
namespace "olm" deleted
namespace "operators" deleted
…
OPERATOR LIFECYCLE MANAGER UNDEPLOYED FROM KUBERNETES

The git repository of OLM is still kept in the local directory. If you would like to delete it as well, you can simply run:

    rm -Rf operator-lifecycle-manager

➜  kubernetes git:(feature/olm_fixup_1) ✗
~~~

deploy/src/main/deploy/kubernetes/common.sh
```diff
@@ -0,0 +1,19 @@
...
+# terms of the Eclipse Public License 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0
+#
+# SPDX-License-Identifier: EPL-2.0
+#*******************************************************************************
+
+set -e
+
+function is_olm_deployed {
+	kubectl get crd catalogsources.operators.coreos.com&>/dev/null && echo "YES" || echo "NO"
```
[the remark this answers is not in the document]
ctron: I did change the way this is being detected the way the master branch currently does it. I also added a bunch of additional deletes for other, cluster wide, resources.

deploy/src/main/deploy/kubernetes/olm_deploy.sh
```diff
@@ -0,0 +1,48 @@
...
+if [ $(is_olm_deployed) == "YES" ]; then
+	echo "OLM already deployed. Exiting ..."
+	exit 0
+fi
+
+
+echo
+echo "Deploying OLM ..."
+test -d operator-lifecycle-manager || git clone https://github.com/operator-framework/operator-lifecycle-manager.git
+kubectl create -f operator-lifecycle-manager/deploy/upstream/manifests/latest/ || true
```
sophokles73: I have some concerns here using `latest` since it might result in undeterministic behavior due to changes in the Git repository. I also wonder if we cannot replace the whole cloning of the repo with some explicit URLs to the resources in the 0.8.0 folder.

ctron: I agree, so I pinned the checkout to the specific commit that I tested. I am not sure though how the git clone operation could be replaced.

sophokles73: well, we could simply do something like
```
kubectl create -f https://github.com/operator-framework/operator-lifecycle-manager/blob/master/deploy/upstream/manifests/0.8.0/0000_30_00-namespace.yaml
kubectl create -f https://github.com/operator-framework/operator-lifecycle-manager/blob/master/deploy/upstream/manifests/0.8.0/0000_30_01-olm-operator.serviceaccount.yaml
...
```
since it is only 18 resources, it shouldn't be too bad to deploy them individually. This might also help with the timing problem that requires us to re-deploy some of the resources ...

ctron: Done

deploy/src/main/deploy/kubernetes/common.sh
```diff
@@ -0,0 +1,45 @@
...
+# SPDX-License-Identifier: EPL-2.0
+#*******************************************************************************
+
+set -e
+
+function is_olm_deployed {
+	kubectl get deployment olm-operator -n olm &> /dev/null && echo "YES" || echo "NO"
+}
+
+RESOURCES_BASE=https://raw.githubusercontent.com/operator-framework/operator-lifecycle-manager/4370fc33c4ff7b6ba3a23def82d4db837dfcbb43/deploy/upstream/manifests/0.8.0/
```
sophokles73: is there a particular reason why you want to keep this and the resource groups in this commons script? As far as I can see they are only used by the `olm_deploy.sh` script, right?

Does the `4370fc33c4ff7b6ba3a23def82d4db837dfcbb43` represent a particular commit in the repo, so that it is even more specific than just *master*?

ctron: I initially had the plan to use those resources for deletion as well, however that doesn't seem to work, so I guess I can move them back to the `deploy` script.

It looks to me like as if the `0.8.0` directory itself gets updated over time, so I would really like to stick to a specific commit. This was the latest commit I saw for this directory.

ctron: I moved the resources back to the `deploy` script.

PR #948, opened 2018-12-10
[#914] Extract SpanContext in EventBusService; put in EventBusMessage

calohmn:
Followup of #935: 
Extract the `SpanContext` from the delivery headers of the vert.x event bus message and put the `SpanContext` in the `EventBusMessage`.

This is for #914.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/EventBusMessage.java
+++ b/core/src/main/java/org/eclipse/hono/util/EventBusMessage.java
@@ -22,6 +22,7 @@ import org.apache.qpid.proton.amqp.Binary;
 import org.apache.qpid.proton.amqp.UnsignedLong;
 import org.apache.qpid.proton.message.Message;
 
+import io.opentracing.SpanContext;
 import io.vertx.core.json.JsonObject;
 
 /**
@@ -35,6 +36,7 @@ public class EventBusMessage {
     private static final String FIELD_CORRELATION_ID_TYPE = "correlation-id-type";
 
     private final JsonObject json;
+    private transient SpanContext spanContext;
 
     private EventBusMessage(final String operation) {
         Objects.requireNonNull(operation);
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/eclipse/hono/util/EventBusMessage.java
```diff
@@ -535,6 +537,27 @@ public EventBusMessage setStringProperty(final String name, final Message msg) {
         }
     }
 
+    /**
+     * Sets a {@code SpanContext} on this message.
```
sophokles73: I think you can come up with a more meaningful description, can't you? The fact that this is the setter for a `SpanContext` is obvious. However, it might be interesting to know what this is good for ...

service-base/src/main/java/org/eclipse/hono/service/EventBusService.java
```diff
@@ -150,6 +153,8 @@ private void processRequestMessage(final Message<JsonObject> msg) {
         }
 
         final EventBusMessage request = EventBusMessage.fromJson(msg.body());
+        final SpanContext spanContext = tracer.extract(Format.Builtin.TEXT_MAP, new MultiMapExtractAdapter(msg.headers()));
```
sophokles73: does `Message.headers()` always returns a `MultiMap` instance, i.e. never `null`?

core/src/main/java/org/eclipse/hono/util/EventBusMessage.java
calohmn: ok, I suggest this:
````
* Sets a {@code SpanContext} on this message.
* This may be the {@code SpanContext} associated with the AMQP message for which this {@code EventBusMessage} was created.  
* When tracing any operations on this {@code EventBusMessage}, this {@code SpanContext} can be used as parent for the
* created <em>OpenTracing</em> spans.
* <p>
* Note: this {@code SpanContext} instance will not get serialized when sending
* the message over the vert.x event bus!
````

Comment (sophokles73) [reply core/src/main/java/org/eclipse/hono/util/EventBusMessage.java]: what about
```
* Sets an <em>OpenTracing</em> {@code SpanContext} on this message.
* <p>
* This may be the tracing context associated with the (request) message for which this {@code EventBusMessage} was created.
* The span context should be used as the parent context of each OpenTracing {@code Span} that is created as part of processing this event bus message.
* <p>
* Note: the span context instance will not get serialized when sending
* this event bus message over the vert.x event bus!
```
?

calohmn: Good, I've amended the commit accordingly.

PR #949, opened 2018-12-10
[#871] Add hashing of plaintext passwords to Credentials API.

b-abel:

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/ClearTextPassword.java
+++ b/core/src/main/java/org/eclipse/hono/util/ClearTextPassword.java
@@ -15,6 +15,8 @@ package org.eclipse.hono.util;
 import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
 import org.springframework.security.crypto.password.PasswordEncoder;
 
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
 import java.util.Objects;
 
 /**
@@ -32,6 +34,20 @@ import java.util.Objects;
  * </pre>
  */
 public class ClearTextPassword extends BasePassword {
+
+    /**
+     * The cryptographically maximal strong salt for SHA-256 in bytes.
+     */
+    private static final int SALT_LENGTH_SHA256 = 32;
+    /**
+     * The cryptographically maximal strong salt for SHA-512 in bytes.
+     */
+    private static final int SALT_LENGTH_SHA512 = 64;
+
+    /**
+     * The random generator for hashing passwords. Lazily initialized because it could be expensive.
+     */
+    private static SecureRandom secureRandom;
 
     /**
      * Creates an instance from the {Base64(salt)}password formatted String.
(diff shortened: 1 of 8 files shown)


service-base/src/main/java/org/eclipse/hono/service/credentials/CompleteBaseCredentialsService.java
```diff
@@ -89,20 +94,32 @@
...
-                });
-            } catch (IllegalStateException e) {
-                return Future.failedFuture(new ClientErrorException(
-                        HttpURLConnection.HTTP_BAD_REQUEST,
-                        e.getMessage()));
-            }
+            final Future<EventBusMessage> requestTracker = Future.future();
+            getVertx().executeBlocking(blockingCodeHandler -> {
+                log.debug("hashing password on vert.x worker thread [{}]", Thread.currentThread().getName());
+                hashPlainPasswords(payload);
```
sophokles73: IMHO we should only run the *hashPlainPasswords* method call on the vert.x worker thread pool
that would make the code more readable and would also keep most of the processing of the request on the original event loop (as before this change)

service-base/src/main/java/org/eclipse/hono/service/credentials/CompleteBaseCredentialsService.java
```diff
@@ -161,6 +189,15 @@ protected int getMaxBcryptIterations() {
         return DEFAULT_MAX_BCRYPT_ITERATIONS;
     }
 
+    /**
+     * Gets the hash function to be used to hash plaintext passwords.
+     * 
+     * @return The name of the hash function.
+     */
+    protected String getHashFunction() {
```
sophokles73: is this method supposed to overridden? If so, then the JavaDoc should state so and indicate why a user would want to do so. It would also be helpful to know, which function names are actually supported.

service-base/src/test/java/org/eclipse/hono/service/credentials/CompleteBaseCredentialsServiceTest.java
```diff
@@ -302,6 +303,11 @@ public void getAll(final String tenantId, final String deviceId, final Handler<A
             protected int getMaxBcryptIterations() {
                 return MAX_ITERATIONS;
             }
+
+            @Override
+            public Vertx getVertx() {
+                return Vertx.vertx();
```
sophokles73: this creates a new vert.x instance on each invocation. are you sure this is what you want?

b-abel: Right. I pushed a fix.

service-base/src/main/java/org/eclipse/hono/service/credentials/CompleteBaseCredentialsService.java
b-abel: Added JavaDoc.

service-base/src/main/java/org/eclipse/hono/service/credentials/CompleteBaseCredentialsService.java
b-abel: I tried that and it didn't look more readable to me. But I will add a commit and we can decide then.

b-abel: I added a change.

core/src/main/java/org/eclipse/hono/util/ClearTextPassword.java
```diff
@@ -102,4 +118,58 @@ public static String encodeBCrypt(final String password, final int strength) {
...
+    }
+
+    private static SecureRandom getSecureRandom() {
+        // this rather complicated setup is intended to prevent blocking, which is especially an issue in containers
+        // see: https://hackernoon.com/hack-how-to-use-securerandom-with-kubernetes-and-docker-a375945a7b21
+        // and: https://tersesystems.com/blog/2015/12/17/the-right-way-to-use-securerandom/
+
+        if (secureRandom == null) {
+            try {
+                secureRandom = SecureRandom.getInstance("NativePRNGNonBlocking"); // non-blocking UNIX
```
sophokles73: this is nice :+1:

PR #950, opened 2018-12-11
[#914] Add SpanContext param to RegistrationService methods

calohmn:
This is for #914.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationService.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationService.java
@@ -15,6 +15,7 @@ package org.eclipse.hono.service.registration;
 
 import org.eclipse.hono.util.RegistrationResult;
 
+import io.opentracing.SpanContext;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Handler;
 import io.vertx.core.Verticle;
@@ -32,6 +33,9 @@ public interface RegistrationService extends Verticle {
      *
      * @param tenantId The tenant the device belongs to.
      * @param deviceId The ID of the device to get the assertion for.
+     * @param spanContext The currently active OpenTracing span (may be {@code null}). An implementation
+     *                    should use this as the parent for any span it creates for tracing
+     *                    the execution of this operation.
      * @param resultHandler The handler to invoke with the result of the operation.
      *             The <em>status</em> will be
      *             <ul>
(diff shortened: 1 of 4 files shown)


service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationService.java
```diff
@@ -41,11 +45,12 @@
      *             <li><em>404 Not Found</em> if no device with the given identifier is
      *             registered for the tenant or its <em>enabled</em> property is {@code false}.</li>
      *             </ul>
-     * @throws NullPointerException if any of the parameters is {@code null}.
+     * @throws NullPointerException if any of the parameters (except spanContext) is {@code null}.
      * @see <a href="https://www.eclipse.org/hono/api/device-registration-api/#assert-device-registration">
      *      Device Registration API - Assert Device Registration</a>
      */
-    void assertRegistration(String tenantId, String deviceId, Handler<AsyncResult<RegistrationResult>> resultHandler);
+    void assertRegistration(String tenantId, String deviceId, SpanContext spanContext,
```
sophokles73: this will break the API
IMHO we should overload the method with a variant accepting a `SpanContext` instead of replacing it

service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationService.java
```diff
@@ -71,9 +79,11 @@
      *             <li><em>404 Not Found</em> if no device with the given identifier is
      *             registered for the tenant or its <em>enabled</em> property is {@code false}.</li>
      *             </ul>
-     * @throws NullPointerException if any of the parameters is {@code null}.
+     * @throws NullPointerException if any of the parameters (except spanContext) is {@code null}.
      * @see <a href="https://www.eclipse.org/hono/api/device-registration-api/#assert-device-registration">
      *      Device Registration API - Assert Device Registration</a>
      */
-    void assertRegistration(String tenantId, String deviceId, String gatewayId, Handler<AsyncResult<RegistrationResult>> resultHandler);
+    void assertRegistration(String tenantId, String deviceId, String gatewayId, SpanContext spanContext,
```
sophokles73: this will break the API
IMHO we should overload the method with a variant accepting a `SpanContext` instead of replacing it

calohmn:
@sophokles73 I've created a separate PR #952 that keeps the `RegistrationService` interface compatible. 
However, I don't see it as necessarily the better solution, because:
- the `RegistrationService` interface in PR #952 has 4 `assertRegistration` methods, making it rather confusing
- concerning implementations of the `RegistrationService` interface, I guess most would use the `BaseRegistrationService`. As the `BaseRegistrationService` explicitly makes use of the `SpanContext`, this would require custom `BaseRegistrationService` subclasses to adopt the `SpanContext` parameter as well, making this change incompatible in the same way as a change of the `RegistrationService` interface, as shown above, would be.

sophokles73:
> the RegistrationService interface in PR #952 has 4 assertRegistration methods, making it rather confusing

That is why we should probably deprecate the variants not accepting the span context

> concerning implementations of the RegistrationService interface, I guess most would use the BaseRegistrationService. As the BaseRegistrationService explicitly makes use of the SpanContext, this would require custom BaseRegistrationService subclasses to adopt the SpanContext parameter as well, making this change incompatible in the same way as a change of the RegistrationService interface, as shown above, would be.

I think we can fix this by simply implementing the variants that do not accept the span context by delegating to the variants accepting the span context, right?

PR #952, opened 2018-12-11
[#914] Add SpanContext param to RegistrationService methods (2)

calohmn:
This is for #914.

This is an alternative to #950 - see the discussion there.


service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationService.java
```diff
@@ -47,6 +48,34 @@
...
+     *             the tenant and its <em>enabled</em> property is {@code true}.
+     *             The <em>payload</em> will contain a JWT token asserting the registration status.</li>
+     *             <li><em>404 Not Found</em> if no device with the given identifier is
+     *             registered for the tenant or its <em>enabled</em> property is {@code false}.</li>
+     *             </ul>
+     * @throws NullPointerException if any of the parameters (except spanContext) is {@code null}.
+     * @see <a href="https://www.eclipse.org/hono/api/device-registration-api/#assert-device-registration">
+     *      Device Registration API - Assert Device Registration</a>
+     */
+    default void assertRegistration(String tenantId, String deviceId, SpanContext spanContext,
```
sophokles73: WDYT about deprecating the other variant?

calohmn: Yes, I think it makes sense to only keep the 2 variants with the `SpanContext` parameter in the long run.

sophokles73: I think so too but I guess we should wait a while with deprecating them anyway.
Can you please mark the arguments to the default implementations as final?

calohmn: Ok, I've marked them as final.

Issue #920, opened 2018-11-23
DefaultFailureHandler expects response message to be set correctly


mbaeuerle:
The `DefaultFailureHandler` uses the status message of the request if the context has a status code set: https://github.com/eclipse/hono/blob/c947ae6be19e27e24938335c8b508a1f3fc331e2/service-base/src/main/java/org/eclipse/hono/service/http/DefaultFailureHandler.java#L73
but in some cases the status message was not yet set and the default message ("OK") is used: 
https://github.com/eclipse-vertx/vert.x/blob/6aafd1733bb82ef71a69675c4414c1a0210d7fcf/src/main/java/io/vertx/core/http/impl/HttpServerResponseImpl.java#L89
For example this is the case when using the Vertx AuthHandler which will fail the context but doesn't set a status message: 
https://github.com/vert-x3/vertx-web/blob/af922fb2b6b46ccbb18568ca5ac4123be8a33491/vertx-web/src/main/java/io/vertx/ext/web/handler/impl/AuthHandlerImpl.java#L193-L200

In this case when the authorization fails, the status will be `401` but with a body of `OK`.

Maybe the DefaultFailureHandler should defer the status message from the status code?

sophokles73:
Once we have upgraded to vert.x 3.6.0, the AuthHandlers that Hono uses will override the *processException* method of `AuthHandlerImpl` to not only fail the context with the status code but always with either an `HttpStatusException` or a `ServiceInvocationException`. The response body will then be set to the message contained in the exception. However, what you propose still makes sense to me, i.e. deriving the body from the status code if the context has been failed with a status code only (and no exception).

sophokles73:
@mbaeuerle  would you like to create a PR for `DefaultFailureHandler`?

PR #953, opened 2018-12-11
[#920] Fix response message to be set correctly

kaniyan:
@mbaeuerle Thank you for the fix.


service-base/src/test/java/org/eclipse/hono/service/http/DefaultFailureHandlerTest.java
```diff
@@ -13,21 +13,28 @@
...
 
-import io.vertx.core.buffer.Buffer;
-import io.vertx.core.http.HttpServerRequest;
-import io.vertx.core.http.HttpServerResponse;
-import io.vertx.ext.web.RoutingContext;
+import java.net.HttpURLConnection;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.RETURNS_MOCKS;
```
sophokles73: please remove this static import

service-base/src/test/java/org/eclipse/hono/service/http/DefaultFailureHandlerTest.java
```diff
@@ -137,4 +144,34 @@ public void testHandlerWritesDetailMessageAndErrorCodeToBody() {
...
 
+    /**
+     * Verifies that the handler writes the status code and the status message to the response in case of failure and
+     * ctx.failure() returns null.
+     */
+    @Test
+    public void testHandlerOnDefaultWritesStatusCodeAsMessageToBody() {
+        final RoutingContext ctx = mock(RoutingContext.class);
+        final HttpServerRequest request = mock(HttpServerRequest.class, Mockito.RETURNS_MOCKS);
+        final HttpServerResponse response = mock(HttpServerResponseImpl.class, RETURNS_MOCKS);
```
sophokles73: we usually do not use partial mocks, so please use the interface type instead

service-base/src/test/java/org/eclipse/hono/service/http/DefaultFailureHandlerTest.java
kaniyan: Below static imports already exists. As part of this PR, I replaced  import `import <package>.*` as to exact classes to be used. Also I noticed that in our test classes often static imports are being used. 

```
import static org.hamcrest.CoreMatchers.is;
import static org.junit.Assert.*;
import static org.mockito.Mockito.*;
```

service-base/src/test/java/org/eclipse/hono/service/http/DefaultFailureHandlerTest.java
kaniyan: This particular issue with status code = 401 and status message ="OK" is closely related to the implementation of HttpServerResponseImpl in vertx. In order to test that, I used partial mocks involving those methods.

service-base/src/test/java/org/eclipse/hono/service/http/DefaultFailureHandlerTest.java
sophokles73: We always refer to constants defined in other classes using the defining class name and the constant name. You are doing it yourself in the code:

```
final HttpServerRequest request = mock(HttpServerRequest.class, Mockito.RETURNS_MOCKS);
final HttpServerResponse response = mock(HttpServerResponseImpl.class, RETURNS_MOCKS);
```

kaniyan: Agreed, I will append the class name too.

service-base/src/main/java/org/eclipse/hono/service/http/DefaultFailureHandler.java
```diff
@@ -70,6 +70,7 @@ public void handle(final RoutingContext ctx) {
                         sendError(ctx.response(), HttpURLConnection.HTTP_INTERNAL_ERROR, ctx.failure().getMessage());
                     }
                 } else if (ctx.statusCode() != -1) {
+                    ctx.response().setStatusCode(ctx.statusCode());
```
sophokles73: I do not think that we can/should fix this issue by trying to set the status code on the response like this.
The underlying problem is that we are (mis-)using the *ctx.response().setStatusMessage()* method together with *ctx.fail(int)* to convey detail information about a problem that has occurred while processing the request. However, we have a much more powerful tool at hand for doing this, which is using *ctx.fail(Throwable)* which we can invoke with a `ServiceInvocationException` containing the status code and detail message in one go. This is already handled correctly in the block above.
I therefore would suggest to simply pass `null` instead of *ctx.response().getStatusMessage()* to the *sendError()* method here and replace the erroneous usage of *ctx.fail(int)* in
* org.eclipse.hono.service.http.AbstractHttpEndpoint#extractRequiredJsonPayload
* org.eclipse.hono.service.credentials.CredentialsHttpEndpoint#updateCredentials
* org.eclipse.hono.service.tenant.TenantHttpEndpoint#checkPayloadForTenantId

with *ctx.fail(ClientErrorException)* ...

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsHttpEndpoint.java
```diff
@@ -140,11 +141,11 @@ private void updateCredentials(final RoutingContext ctx) {
 
         // auth-id and type from URI must match payload
         if (!authIdFromUri.equals(authId)) {
-            ctx.response().setStatusMessage("Non-matching authentication identifier");
-            ctx.fail(HttpURLConnection.HTTP_BAD_REQUEST);
+            ctx.fail(new ServiceInvocationException(HttpURLConnection.HTTP_BAD_REQUEST,
```
sophokles73: well, since 400 indicates a client error, we should create a `ClientErrorException` instead of a (generic) `ServiceInvocationException`
the same applies to multiple other locations ...

kaniyan: I replaced it with `ClientErrorException`

PR #954, opened 2018-12-11
Http adapter: Check device registration during command response upload

kaniyan:


adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapterTest.java
```diff
@@ -410,6 +411,42 @@ public void testUploadCommandResponseFailsForDisabledTenant() {
...
+                .put(TenantConstants.FIELD_ADAPTERS_TYPE, ADAPTER_TYPE)
+                .put(TenantConstants.FIELD_ENABLED, true));
+        when(tenantClient.get(eq("tenant"), (SpanContext) any())).thenReturn(Future.succeededFuture(to));
+
+        // which is connected to a Credentials service that has credentials on record for device 9999
+        when(adapter.getAuthenticatedDevice(ctx)).thenReturn(new DeviceUser("tenant", "9999"));
+
+        // but for which no registration information is available
+        when(regClient.assertRegistration((String) any(), (String) any(), (SpanContext) any()))
+                .thenReturn(Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_FORBIDDEN,
```
sophokles73: this should probably return a 404, reflecting the case that the device does not (or no longer) exist anymore, right?

adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapterTest.java
```diff
@@ -410,6 +411,42 @@ public void testUploadCommandResponseFailsForDisabledTenant() {
         verify(metrics, never()).incrementCommandResponseDeliveredToApplication(eq("tenant"));
     }
 
+    /**
+     * Verifies that the adapter fails the upload of a command response with a 403 if the device is not registered and
+     * the valid credentials for the device is available.
+     */
+    @Test
+    public void testUploadCommandResponseFailsForDeletedDevices() {
```
sophokles73: the generic case that the device does not exist, isn't it?
So, I would propose to name this method *testUploadCommandResponseFailsForNonExistingDevice*

adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapterTest.java
kaniyan: Device registration doesn't exist and where as device credentials exists. It seemed to me partially available and hence it led me to 403. I think so too, that 404 is more appropriate.

adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapterTest.java
kaniyan: Done

adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapterTest.java
```diff
@@ -410,6 +411,42 @@ public void testUploadCommandResponseFailsForDisabledTenant() {
         verify(metrics, never()).incrementCommandResponseDeliveredToApplication(eq("tenant"));
     }
 
+    /**
+     * Verifies that the adapter fails the upload of a command response with a 403 if the device is not registered and
```
sophokles73: now the JavaDoc needs to be adapted to the new status code as well ...

kaniyan: Done

PR #955, opened 2018-12-11
[#914] Create span in "assertRegistration" methods

calohmn:
This is for #914.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
@@ -17,14 +17,20 @@ import java.util.Objects;
 
 import org.eclipse.hono.client.ClientErrorException;
 import org.eclipse.hono.service.EventBusService;
+import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.CacheDirective;
 import org.eclipse.hono.util.EventBusMessage;
+import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.RegistrationConstants;
 import org.eclipse.hono.util.RegistrationResult;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 
+import io.opentracing.References;
+import io.opentracing.Span;
 import io.opentracing.SpanContext;
+import io.opentracing.Tracer;
+import io.opentracing.tag.Tags;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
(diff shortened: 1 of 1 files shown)


service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
```diff
@@ -212,16 +218,21 @@ public void assertRegistration(
         Objects.requireNonNull(deviceId);
         Objects.requireNonNull(resultHandler);
 
+        final Span currentSpan = buildSpan(spanContext, tenantId, deviceId, null);
```
sophokles73: hmm, given that this method is supposed to be overridden by users, wouldn't it make sense to create a Span in *processAssertRequest* already so that we have at least *some* kind of tracing in any case (i.e. regardless of whether this method is overridden or not)?

calohmn: As the `assertRegistration` method isn't necessarily overridden by users, I don't see this as strictly necessary.
For the case that the default `assertRegistration` method is used, an extra span in `processAssertRequest` doesn't provide much extra info. And creating such a Span in `processAssertRequest` (while making sure it is finished when the returned future is finised) would make the code more complex.
I would say, if a user wants to implement a custom `assertRegistration` method, the user should also implement tracing there if tracing is relevant for the user. I've make the method to create a span protected, so that adding tracing support in an overridden method should be straightforward.

sophokles73: FMPOV it is the other way around. Because the *assertRegistration* **may** be overridden, we should start (and finish) a span in *processAssertRequest* and pass in that span into *assertRegistration*. A subclass then may (or may not) log to that span or create child spans of it as it sees fit. But in any case we have some minimal trace of the *assertRegistration* method being executed (and of its outcome).

calohmn: The extra, kind of superfluous span being created in the (probably common) case that `assertRegistration` is _not_ overridden doesn't look that good, FMPOV.
I've added the span nonetheless and it at least has the benefit of capturing the case of a missing tenant/device id.

service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
```diff
@@ -212,16 +218,21 @@ public void assertRegistration(
         Objects.requireNonNull(deviceId);
         Objects.requireNonNull(resultHandler);
 
+        final Span currentSpan = newChildSpan("assert Device Registration", spanContext, tenantId, deviceId, null);
```
sophokles73: this is not what I meant. What I wanted to propose was to create the span in *processAssertRequest* and pass the span instead of a `SpanContext` into this method ...

calohmn: Ok, sorry, got that wrong. (I've removed the commit again here.)
That would mean changing the `assertRegistration` method in the `RegistrationService` interface (using a Span instead of SpanContext param). And, to keep things consistent, also adapting `TenantService` (and later on `CredentialsService`).

(Thinking of having consistence in the kind of parameters used, it is to be noted here that there are similar methods in the `RegistrationClient`, `TenantClient` (and soon `CredentialsClient`) interfaces, having the `SpanContext` param there as well. As these interfaces are not necessarily implemented by users, we could keep them as is, though.)

Although I would intuitively rather refrain from passing on a span to other methods (so as to prevent the other methods from being able to finish it), I see the mentioned advantages.

I've created a new PR #962 with the corresponding changes.
(TenantService should be adapted accordingly after that.)

calohmn:
Closed in favor of #962.

PR #957, opened 2018-12-12
[#914] Create span for CredentialsService.get

calohmn:
This is for #914.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/credentials/BaseCredentialsService.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/credentials/BaseCredentialsService.java
@@ -17,12 +17,21 @@ import java.util.Objects;
 import java.util.Optional;
 
 import org.eclipse.hono.client.ClientErrorException;
+import org.eclipse.hono.client.ServiceInvocationException;
 import org.eclipse.hono.service.EventBusService;
+import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.CredentialsConstants;
 import org.eclipse.hono.util.CredentialsResult;
 import org.eclipse.hono.util.EventBusMessage;
+import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.TenantConstants;
 
+import io.opentracing.References;
+import io.opentracing.Span;
+import io.opentracing.SpanContext;
+import io.opentracing.Tracer;
+import io.opentracing.noop.NoopSpan;
+import io.opentracing.tag.Tags;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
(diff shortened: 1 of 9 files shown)


calohmn:
I've rebased and adapted the commit to the recent changes for the vertx update.
(Failing CI build doesn't seem caused by changes here.)

I'm about to create a 2nd PR that will add the necessary changes to the `CredentialsClient`.

calohmn:
I've applied the signature change in `newChildSpan` and have refactored the `BaseCredentialsService.processGetRequest()` method.
Concerning the query parameters: I prefer having them as Tags, also to enable searching for them.

service-base/src/main/java/org/eclipse/hono/service/credentials/BaseCredentialsService.java
```diff
@@ -80,47 +93,110 @@ protected String getEventBusAddress() {
...
+            TracingHelper.logError(span, "missing tenant and/or payload");
+            resultFuture = Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST));
         } else {
             final String type = removeTypesafeValueForField(String.class, payload, CredentialsConstants.FIELD_TYPE);
             final String authId = removeTypesafeValueForField(String.class, payload,
                     CredentialsConstants.FIELD_AUTH_ID);
             final String deviceId = removeTypesafeValueForField(String.class, payload,
                     CredentialsConstants.FIELD_PAYLOAD_DEVICE_ID);
 
+            span = newChildSpan(SPAN_NAME_GET_CREDENTIALS, request.getSpanContext(), tenantId);
```
sophokles73: there doesn't seem to be a difference between the *if* block and the *else* block regarding the creation of the span, is there?

service-base/src/main/java/org/eclipse/hono/service/credentials/BaseCredentialsService.java
```diff
@@ -80,47 +93,110 @@ protected String getEventBusAddress() {
...
+                TracingHelper.logError(span, String.format(
+                        "invalid search criteria [type: %s, device-id: %s, auth-id: %s]", type, deviceId, authId));
                 log.debug("get credentials request contains invalid search criteria [type: {}, device-id: {}, auth-id: {}]",
                         type, deviceId, authId);
-                return Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST));
+                resultFuture = Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST));
+            }
+        }
+        return resultFuture.compose(eventBusMessage -> {
+            Tags.HTTP_STATUS.set(span, eventBusMessage.getStatus());
```
sophokles73: I think we'll want to set this tag in any case, right?

PR #959, opened 2018-12-12
[#750] Health check for all verticle instances.

b-abel:
This replaces the PR #873 and is based on the discussion there.

Vertx handles the number of instances for each Verticle.

HealthCheckServer is now a bean. If the bean is present, it must be configured (port given). If the bean is not present a Noop-implementation is used.

Refactor: the configuration of `ApplicationConfigProperties` beans is moved from individual protocol adapters to `AbstractAdapterConfig`.


service-base/src/main/java/org/eclipse/hono/service/HealthCheckRegistration.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.service;
+
+/**
+ * Provides a server for health checks.
+ */
+public interface HealthCheckRegistration extends Lifecycle {
```
sophokles73: I wonder why this interface isn't called `HealthCheckServer`. Isn't that what it is supposed to be (according to the JavaDoc)?

b-abel: To be honest, this was a compromise because that is already the name of the class implementing it and the diff would be horror if I rename the class and give it's name to the interface. If you are willing to go through diff hell I can rename to rename both :)

sophokles73: I do not see the issue, there are only very few places where HealthCheckServer is used. SO FMPOV we should rename the existing `HealthCheckServer` class to e.g. `VertxBasedHealthCheckServer` and rename `HealthCheckRegistration` to `HealthCheckServer`.

b-abel: There is no issue with the code, just with the tooling. The diff shown in github and probably every IDE will compare the new interface with the old implementation class, because they have the same name, making it harder for you to spot the difference in the code. But no problem, I will change it.

b-abel:
Contains now the changes that have been requested so far.

PR #962, opened 2018-12-13
[#914] Create span in "assertRegistration" methods (2)

calohmn:
Alternative to #955.
Create and pass on span in "processAssertRequest" method.
Use Span instead of SpanContext parameter in "assertRegistration" methods.

This is for #914.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
@@ -17,14 +17,21 @@ import java.util.Objects;
 
 import org.eclipse.hono.client.ClientErrorException;
 import org.eclipse.hono.service.EventBusService;
+import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.CacheDirective;
 import org.eclipse.hono.util.EventBusMessage;
+import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.RegistrationConstants;
 import org.eclipse.hono.util.RegistrationResult;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 
+import io.opentracing.References;
+import io.opentracing.Span;
 import io.opentracing.SpanContext;
+import io.opentracing.Tracer;
+import io.opentracing.noop.NoopSpan;
+import io.opentracing.tag.Tags;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.CompositeFuture;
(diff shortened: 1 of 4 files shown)


service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
```diff
@@ -124,28 +133,34 @@ protected String getEventBusAddress() {
...
+                log.debug("asserting registration of device [{}] with tenant [{}]", deviceId, tenantId);
+                assertRegistration(tenantId, deviceId, span, result.completer());
+            } else {
+                log.debug("asserting registration of device [{}] with tenant [{}] for gateway [{}]",
+                        deviceId, tenantId, gatewayId);
+                assertRegistration(tenantId, deviceId, gatewayId, span, result.completer());
+            }
             return result.map(res -> {
+                if (res.isError()) {
+                    TracingHelper.logError(span, "status " + res.getStatus());
```
sophokles73: we usually use `io.opentracing.tag.Tags.HTTP_STATUS` to record the resulting status code of invoking a service operation. Marking the Span as erroneous is still necessary, though, but could be done with simply invoking
`io.opentracing.tag.Tags.ERROR.set(span, true)`

service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
```diff
@@ -124,28 +133,34 @@ protected String getEventBusAddress() {
...
-                        .setCacheDirective(res.getCacheDirective());
-            });
         } else {
-            log.debug("asserting registration of device [{}] with tenant [{}] for gateway [{}]",
-                    deviceId, tenantId, gatewayId);
             final Future<RegistrationResult> result = Future.future();
-            assertRegistration(tenantId, deviceId, gatewayId, spanContext, result.completer());
+            if (gatewayId == null) {
+                log.debug("asserting registration of device [{}] with tenant [{}]", deviceId, tenantId);
+                assertRegistration(tenantId, deviceId, span, result.completer());
```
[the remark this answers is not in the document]
calohmn: Yes, I'm aware of that. Just thought that using `result.completer()` makes it more explicit that a Handler is used here - also because other places in Hono still use that as well.

service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
calohmn: Ok, I've added a tag for `HTTP_STATUS`.

calohmn:
I've added a note in the release notes.

PR #963, opened 2018-12-13
[#914] Create span for TenantService.get

calohmn:
Create and pass on span in "TenantService#processGetRequest" method.
Use Span instead of SpanContext parameter in "TenantService#get" methods.

This is for #914.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
@@ -21,11 +21,18 @@ import javax.security.auth.x500.X500Principal;
 import org.eclipse.hono.client.ClientErrorException;
 import org.eclipse.hono.client.ServerErrorException;
 import org.eclipse.hono.service.EventBusService;
+import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.EventBusMessage;
+import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.TenantConstants;
 import org.eclipse.hono.util.TenantResult;
 
+import io.opentracing.References;
+import io.opentracing.Span;
 import io.opentracing.SpanContext;
+import io.opentracing.Tracer;
+import io.opentracing.noop.NoopSpan;
+import io.opentracing.tag.Tags;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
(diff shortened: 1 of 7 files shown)


service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
```diff
@@ -94,53 +109,113 @@ protected final String getEventBusAddress() {
...
+     *
+     * @param operationName The operation name that the span should be created for.
+     * @param spanContext Existing span context.
+     * @param tenantId The tenant id.
+     * @return The new {@code Span}.
+     */
+    protected final Span newChildSpan(final String operationName, final SpanContext spanContext, final String tenantId) {
+        // we set the component tag to the class name because we have no access to
+        // the name of the enclosing component we are running in
+        final Tracer.SpanBuilder spanBuilder = tracer.buildSpan(operationName)
```
sophokles73: what if operation is `null`?

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedTenantService.java
```diff
@@ -228,23 +230,25 @@ public void get(final String tenantId, final SpanContext spanContext, final Hand
...
+    private TenantResult<JsonObject> getForCertificateAuthority(final X500Principal subjectDn, final Span span) {
 
         if (subjectDn == null) {
+            TracingHelper.logError(span, "missing subject DN");
             return TenantResult.from(HttpURLConnection.HTTP_BAD_REQUEST);
         } else {
             final TenantObject tenant = getByCa(subjectDn);
 
             if (tenant == null) {
+                TracingHelper.logError(span, "tenant not found");
```
sophokles73: I think it is clear from the status code that the tenant has not been found. IMHO the log message would be more helpful if it would explain the reason, in this case something like `no tenant found for subject DN`.

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
calohmn: I've added a null check and documented the NPE thrown in that case.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedTenantService.java
calohmn: I've changed the log message accordingly.

Issue #958, opened 2018-12-12
Headers device-id and tenant-id missing in command response


dguggemos:
When using Hono's command&control together with Ditto, I noticed that the command responses do not contain the `tenant-id` and `device-id` headers, as provided for example in telemetry messages.
This would be very useful information for us, as we would not need to remember which device sent a command and correlate the response using the `correlation-id` (across multiple nodes in the worst case).

kaniyan:
Currently as part of the telemetry and event messages, the application properties contain `device-id` and not `tenant-id`. I think `tenant-id` should also be part of it. Eventually the command responses should also have the both.

sophokles73:
> This would be very useful information for us, as we would not need to remember which device sent a command and correlate the response using the correlation-id (across multiple nodes in the worst case).

Including the device ID in a command response as an application property should be no big deal. Do you also need the message to contain the tenant ID or can you derive the tenant from the *connector* that the command response is received over? I would assume that you do the same for telemetry/events already, don't you?

sophokles73:
@dguggemos any news on this?

dguggemos:
We do not need the `tenant-id` currently, but maybe in the future. As the PR also contains the `tenant-id` for the command responses, I'm completely fine with the changes. 👍

PR #965, opened 2018-12-17
[#958] Part 1: Get the tenant and device ids into command response

kaniyan:


kaniyan:
This is PR for #958

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -807,8 +807,9 @@ protected void onCommandReceived(final ProtonSender sender, final CommandContext
 
         LOG.debug("creating command response [status: {}, correlation-id: {}, reply-to: {}]",
                 statusCode, correlationId, context.getResourceIdentifier().toString());
-        final CommandResponse commandResponse = CommandResponse.from(context.getMessagePayload(),
-                context.getMessageContentType(), statusCode, correlationId, context.getResourceIdentifier());
+        final CommandResponse commandResponse = CommandResponse.from(context.getTenantId(), context.getDeviceId(),
```
sophokles73: I am afraid that this will not work if the device is not authenticated and the reply-to address specified by the Command indeed contains the device ID. However, the C&C API does not strictly require this but instead allows the usage of any arbitrary string after the `command/$tenant-id` ...

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -992,6 +992,7 @@ public void registerLivenessChecks(final HealthCheckHandler handler) {
      * <li><em>to</em> will be set to the address consisting of the target's endpoint and tenant</li>
      * <li><em>content-type</em> will be set to content type</li>
      * <li><em>creation-time</em> will be set to the current number of milliseconds since 1970-01-01T00:00:00Z</li>
+     * <li>application property <em>tenant_id</em> will be set to the target's resourceId property</li>
```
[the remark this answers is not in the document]
kaniyan: Its tenantId instead of resourceId. I will change it.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
kaniyan: @sophokles73 Thank you. How about on arrival of the commands, check if the device is not authenticated, extract the reply-id from the reply-to parameter and compare with the device-id. If it doesn't match then append `:~device-id~:` to the reply-to parameter. On arrival of the command response, check for for the pattern `:~ ~:` in reply-id. Then extract the device id, remove the pattern and send the response. WDYT?

sophokles73: > How about on arrival of the commands, check if the device is not authenticated, extract the reply-id 

IMHO we should do this for *every* command and not rely on the fact that the device will also be authenticated when it sends the response ...

Apart from that, I think that your approach can work. However, I would prefer to use a similar approach as in the HTTP adapter, i.e. include the device ID at the position in the address that we usually expect it but e.g. prepend it with a single character (`0`/`1`) indicating whether the device ID needs to be removed from the address before forwarding the response downstream ...

kaniyan: Good to know more about the approach in Http Adapter. I also thinks its better to have an uniform approach. I will go ahead with it.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -633,6 +633,17 @@ protected void onCommandReceived(final ProtonSender sender, final CommandContext
             msg.setCorrelationId(msg.getMessageId());
         }
 
+        String replyToId = command.getReplyToId();
+        if (!command.isOneWay()) {
+            if (command.getRequestId().startsWith("0")) {
```
sophokles73: IMHO we should not make any assumptions regarding the structure of the request ID generated by the `Command` class. In the end, one of the reasons why we have the `Command` class, is to encapsulate exactly such logic. Why don't we simply *re-write* the (AMQP) message's *reply-to* property according to this logic when we create the `Command` instance from the message? We could then simply forward the result of `Command.getCommandMessage()` to the device.
We could do the reverse when we receive the command response from the device, i.e. encapsulate the restoring of the original address of the message as part of creating the `CommandResponse` instance. WDYT?

kaniyan: Initially I also thought of re-writing the (AMQP) message's reply-to property according to this logic when we create the Command instance from the messages. This re-writing of reply-id is needed only for the AMQP adpater to get the `device-id`. Encoding and decoding of `reply-to` for all commands irrespective of the adapters being used will add a bit overhead (time) to the total processing time of commands. So I have scoped down the Implementation only to the AMQP adapter.

kaniyan: @sophokles73 I have pushed my changes. Can you pls take a look?

client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -327,4 +330,11 @@ public String toString() {
             return "Invalid Command";
         }
     }
+
+    private static String processForDeviceId(final String replyToId, final String deviceId) {
+        if (replyToId.startsWith(deviceId + "/")) {
+            return replyToId.replaceFirst(deviceId + "/", deviceId + "/0");
```
sophokles73: it feels strange to me that we mark up the *next* segment in order to indicate whether the deviceId needs to be removed or not. Can we instead prepend the deviceId with a `o` or `1`? That would also be more in line with the way we do it for the requestId ...

kaniyan: > it feels strange to me that we mark up the _next_ segment in order to indicate whether the deviceId needs to be removed or not. Can we instead prepend the deviceId with a `o` or `1`? That would also be more in line with the way we do it for the requestId ...

@sophokles73 It also uses similar approach as of the _requestId_ for the _deviceId_ handling. In case of _requestId_ no _deviceId_ is added the _replyTo_ address. The _requestId_ is prepended with `1` or `0` to indicate if the _deviceId_ is to be later in the `CommandResponse` class added or not. But for the current problem, the _deviceId_ should be sent with the _replyTo_ address.

In this PR, the _replyTo_ address is checked for the _deviceId_ . If it is present, then the _replyId_ is prepended with `0` to indicate that later int the `CommandResponse` class the _deviceId_ should not be removed. If not present, then the _device-id_ is included in the _replyTo_ address and the _replyId_ is prepended with `1` to indicate the _deviceId_ to be removed later in the below example.
```
control/Tenant/Device/xyz    ->    control/Tenant/Device/0xyz 
control/Tenant/xyz   ->    control/Tenant/Device/1xyz
```

Prepending to the _deviceId_ instead to the _replyId_ would appear as below.
```
control/Tenant/Device/xyz    ->    control/Tenant/0Device/xyz 
control/Tenant/xyz     ->    control/Tenant/1Device/xyz
```
Also an advantage of prepending to the _replyId_ instead to the _deviceId_ is that the `ResourceIdentifier.getResourceId()` returns a valid _deviceId_ instead starting with `1` or `0`. For example `AMQPContext`  uses `getResourceId() `  to set the _deviceId_ in case of unauthenicated devices. IMHO its a matter of  IMHO its a matter of perspective, if the _deviceId_ being added to the _replyId_ or to the _replyTo_ address.

kaniyan:
> Can you please split up this PR into multiple ones?

@sophokles73 sure, I will split them.

kaniyan:
@sophokles73 I split the orginal PR. I pushed the first part containing changes required to get the tenant and device ids into the `CommandResponse`.

sophokles73:
@kaniyan can you please rebase?

kaniyan:
> @kaniyan can you please rebase?

@sophokles73 Done.

client/src/main/java/org/eclipse/hono/client/CommandResponse.java
```diff
@@ -35,7 +35,8 @@
     private final String replyToId;
     private final String correlationId;
 
-    private CommandResponse(final Buffer payload, final String contentType, final int status, final String correlationId, final String replyToId) {
+    private CommandResponse(final String tenantId, final String deviceId, final Buffer payload,
```
sophokles73: this doesn't look right, neither *tenantId* nor *deviceId* are used ...

client/src/main/java/org/eclipse/hono/client/CommandResponse.java
```diff
@@ -96,15 +101,16 @@ public static CommandResponse from(
      * @return The command response or {@code null} if any of correlationId, replyTo and status is null or if the
      *         status code is &lt; 200 or &gt;= 600.
      */
-    public static CommandResponse from(final Buffer payload, final String contentType, final Integer status,
+    public static CommandResponse from(final String tenantId, final String deviceId, final Buffer payload,
```
sophokles73: this seems to be the factory method to be used by the AMQP adapter, right? What about simply passing in the `Message` instead of all the params?

client/src/main/java/org/eclipse/hono/client/CommandResponse.java
kaniyan: @sophokles73 These _tenantId_ and _deviceId_ are used in the next part of the PR to set those values in amqp message.

sophokles73: then please remove the unused parameters

client/src/main/java/org/eclipse/hono/client/CommandResponse.java
kaniyan: yes. I think its a good idea. It makes the Adapter code more readable and it can be reused by the adapters.

client/src/main/java/org/eclipse/hono/client/CommandResponse.java
kaniyan: If I remove them here, then it will be still unused in `CommandResponse.from(..` and it has a chain effect.Then device and tenant ids cannot be brought into the `CommandResponse`. I will add one next step (setting inside properties), so that these ids are used.

client/src/main/java/org/eclipse/hono/client/CommandResponse.java
kaniyan: @sophokles73 I have refactored to accept message as a parameter and pushed my changes.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -465,6 +465,7 @@ private AmqpContext createContext(
             LOG.debug("adjusting message's address [orig: {}, updated: {}]", message.getAddress(), to);
             message.setAddress(to);
         }
+        LOG.trace("Received message: [{}]", message.toString());
```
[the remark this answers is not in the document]
kaniyan: Removed

client/src/main/java/org/eclipse/hono/client/CommandResponse.java
```diff
@@ -88,29 +96,39 @@ public static CommandResponse from(
...
+                MessageHelper.APP_PROPERTY_STATUS, Integer.class);
+
+        if (correlationId == null || message.getAddress() == null || status == null) {
             return null;
         } else if (INVALID_STATUS_CODE.test(status)) {
             return null;
         } else {
-            return new CommandResponse(tenantId, deviceId, payload, contentType, status, correlationId,
-                    getReplyId(deviceId, address));
+            final ResourceIdentifier resource = ResourceIdentifier.fromString(message.getAddress());
```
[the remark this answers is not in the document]
kaniyan: > I wonder if it would be helpful to replace the fields of `CommandResponse` with a `Message` for storing the properties. In the end, producing a `Message` to be sent to an application is the whole purpose of the `CommandResponse` class. That way we would also prevent the reading of properties from a response AMQP message received from a device just in order to later putting them into a new AMP message to be sent to the application. WDYT?

@sophokles73 Currently CommandResponseSender is responsible for creating a message out of the `CommandResponse` fields. I like the idea to replace the existing fields with a `Message`.
I would suggest a separate PR to replace the fields with a `Message`. I can raise it as a 3rd PR. WDYT?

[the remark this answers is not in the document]
kaniyan: > this can throw a runtime exception

@sophokles73 The `NullPointerException` and  `IllegalArgumentException` that may be thrown by `ResourceIdentifier.fromString(...` have been handled and the changes are pushed.

sophokles73:
@kaniyan can you squash and rebase?

kaniyan:
> @kaniyan can you squash and rebase?

@sophokles73 Done.

PR #968, opened 2018-12-18
Http/Mqtt adapter: Add getter for metrics.

calohmn:
Expose metrics to sub classes of `AbstractVertxBasedHttpProtocolAdapter`.


calohmn:
Added there as well.

PR #973, opened 2018-12-20
[#914] Add info about CredentialsService change in release notes

calohmn:
This is for #914.


sophokles73:
@calohmn can you rebase?

PR #978, opened 2019-01-08
Support X.509 cert based auth in MQTT

sophokles73:
This PR adds support for authenticating devices using an X.509 client certificate. It is modeled and implemented along the lines of how it is done in the HTTP adapter.


adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/X509AuthHandler.java
```diff
@@ -0,0 +1,233 @@
+/**
```
dejanb: This class copies a lot from the http/X509AuthHandler. Maybe with some refactoring we can reuse the existing code instead of duplicating it?

sophokles73:
I already have the integration tests in my local workspace already ;-) Will create a new PR for them once this is approved ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/X509AuthHandler.java
sophokles73: Yes, I thought so too. However, the only way I can imagine reusing the code would be to factor it out into some static helper methods in a utility class. Or do you have another idea?

dejanb: Yeah, a small utility class with parseCredentials(Certificate[] path) and other duplicated methods would be a good approach.

sophokles73: Ok, I will try that (tomorrow) ...

sophokles73:
@dejanb I have added a commit regarding factoring out common code for validating cert path

PR #981, opened 2019-01-08
Consolidate OpenShift deployments

ctron:


ctron:
Indeed, I just removed that.

PR #987, opened 2019-01-14
[#931] Limit number of connections in MQTT adapter

b-abel:
Connections are rejected once the connection limit is reached.
Currently the limit is fixed on Long.MAX_VALUE, needs to be made
configurable and/or auto-configurable later on.

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
@@ -326,11 +326,26 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends ProtocolAd
 
     private Future<Device> handleConnectionRequest(final MqttEndpoint endpoint, final Span currentSpan) {
 
+        if (isConnectionLimitExceeded()) {
+            LOG.debug("Connection limit ({}) exceeded, reject connection request", getConnectionLimit());
+            currentSpan.log(String.format("connection limit (%d) exceeded", getConnectionLimit()));
+            return Future.failedFuture(new MqttConnectionException(
+                    MqttConnectReturnCode.CONNECTION_REFUSED_SERVER_UNAVAILABLE));
+        }
+
         if (getConfig().isAuthenticationRequired()) {
             return handleEndpointConnectionWithAuthentication(endpoint, currentSpan);
         } else {
             return handleEndpointConnectionWithoutAuthentication(endpoint);
         }
+    }
+
+    private boolean isConnectionLimitExceeded() {
+        return metrics.getNumberOfConnections() >= getConnectionLimit();
+    }
+
+    private int getConnectionLimit() {
+        return getConfig().getMaxConcurrentConnections();
     }
 
     private void handleConnectionRequestResult(final MqttEndpoint endpoint,
--- a/core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
+++ b/core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
@@ -23,6 +23,7 @@ public class ProtocolAdapterProperties extends ServiceConfigProperties {
     private boolean jmsVendorPropsEnabled = false;
     private boolean defaultsEnabled = true;
     private long eventLoopBlockedCheckTimeout = 5000L;
+    private int maxConcurrentConnections = Integer.MAX_VALUE;
 
     /**
      * Checks whether the protocol adapter always authenticates devices using their provided credentials as defined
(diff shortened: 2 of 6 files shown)


adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -325,13 +326,23 @@ final void handleEndpointConnection(final MqttEndpoint endpoint) {
...
+        }
+
         if (getConfig().isAuthenticationRequired()) {
             return handleEndpointConnectionWithAuthentication(endpoint, currentSpan);
         } else {
             return handleEndpointConnectionWithoutAuthentication(endpoint);
         }
     }
 
+    private boolean connectionLimitExceeded() {
```
sophokles73: can we please rename this to `isConnectionLimitExceeded`?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -87,6 +87,7 @@
     private MqttServer server;
     private MqttServer insecureServer;
     private AuthHandler<MqttContext> authHandler;
+    private final long connectionLimit = Long.MAX_VALUE; // TODO make configurable
```
sophokles73: can you add the configuration property right now?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -325,13 +326,23 @@ final void handleEndpointConnection(final MqttEndpoint endpoint) {
 
     private Future<Device> handleConnectionRequest(final MqttEndpoint endpoint, final Span currentSpan) {
 
+        if (connectionLimitExceeded()) {
+            LOG.info("Connection limit ({}) exceeded, reject connection request", connectionLimit);
```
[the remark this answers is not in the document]
b-abel: > The log level should be debug FMPOV because this is an expected (and handled) case that will occur thousands of times in production, don't you think?

Probably one wants to see if the protocol adapter regularly hits the limit without constantly logging on debug level. But you are right: logging this for every connection request (above the limit) is probably an overkill.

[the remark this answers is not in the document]
sophokles73: I think that this would be an excellent use case for a Dashboard on top of the metrics back end :-)

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
b-abel: Sure. What do you prefer: adding it to `ProtocolAdapterProperties` or creating a new `MqttProtocolAdapterProperties` (and analogously the other protocol adapters)? I tend to the former one, even if it might not be used by e.g. the HTTP adapter.

sophokles73: I also think that `ProtocolAdapterProperties` would be the right place.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -339,8 +339,12 @@ final void handleEndpointConnection(final MqttEndpoint endpoint) {
         }
     }
 
-    private boolean connectionLimitExceeded() {
-        return metrics.getNumberOfConnections() >= connectionLimit;
+    private boolean isConnectionLimitExceeded() {
+        return metrics.getNumberOfConnections() >= connectionLimit();
+    }
+
+    private long connectionLimit() {
```
sophokles73: can we rename this to *getConnectionLimit*?

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
```diff
@@ -23,6 +23,7 @@
     private boolean jmsVendorPropsEnabled = false;
     private boolean defaultsEnabled = true;
     private long eventLoopBlockedCheckTimeout = 5000L;
+    private long concurrentConnectionsLimit = Long.MAX_VALUE;
```
sophokles73: I think an int will be more than enough, don't you think?

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
```diff
@@ -154,4 +155,28 @@ public final long getEventLoopBlockedCheckTimeout() {
...
     }
+
+    /**
+     * Gets the number of maximal concurrent connections the protocol adapter should accept.
+     * <p>
+     * Default is {@link Long#MAX_VALUE}.
+     * 
+     * @return The maximal number of concurrent connections.
+     */
+    public long getConcurrentConnectionsLimit() {
```
sophokles73: can we rename this to *getMaxConcurrentConnections*? We use this naming pattern in other places already ...

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
b-abel: Right. With 0 as "unlimited"?

sophokles73: or `Integer.MAX_VALUE`?

b-abel: `Integer.MAX_VALUE` is fine.

I just remembered where the `Long` comes from: I blame the class `MicrometerBasedMetrics` where open connections are counted with `AtomicLong` :)

b-abel:
Sorry, I added the test afterwards. Do you want me to squash again?

PR #988, opened 2019-01-14
Ensure OpenTracing span is finished if connection is closed prematurely

calohmn:
This prevents the "upload event" and "upload telemetry" OpenTracing spans to be left unfinished in case the client connected to the HTTP protocol adapter has prematurely closed the connection.

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -606,7 +606,7 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
                                 ttd);
                         customizeDownstreamMessage(downstreamMessage, ctx);
 
-                        addConnectionCloseHandler(ctx, commandConsumerTracker.result(), tenant, deviceId);
+                        addConnectionCloseHandler(ctx, commandConsumerTracker.result(), tenant, deviceId, currentSpan);
 
                         if (qos == null) {
                             return CompositeFuture.all(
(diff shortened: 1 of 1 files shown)


adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -674,6 +674,11 @@ private void doUploadMessage(final RoutingContext ctx, final String tenant, fina
                             Optional.ofNullable(commandConsumerTracker.result()).ifPresent(consumer -> consumer.close(null));
                         });
                         ctx.response().end();
+                    } else {
+                        LOG.debug("failed to send http response for [{}] message from device [tenantId: {}, deviceId: {}]: response already closed",
+                                endpointName, tenant, deviceId);
+                        TracingHelper.logError(currentSpan, "failed to send HTTP response to device: response already closed");
+                        currentSpan.finish();
```
sophokles73: in order to improve readability, I'd rather move this block to the top and check for `if (ctx.response().closed())` instead
WDYT?

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -733,17 +738,20 @@ protected long messagePayloadSize(final RoutingContext ctx) {
...
             final String tenantId,
-            final String deviceId) {
+            final String deviceId,
+            final Span currentSpan) {
 
         if (messageConsumer != null && !ctx.response().closed()) {
             ctx.response().closeHandler(v -> {
                 LOG.debug("device [tenant: {}, device-id: {}] closed connection before response could be sent",
                         tenantId, deviceId);
+                TracingHelper.logError(currentSpan, "device closed connection");
```
sophokles73: is this an error? Maybe the device has good reasons to do so ...

calohmn: yes, right. The error, that is probably following this, will be set on the span in the `if (ctx.response().closed()) {` block above.
I've changed it to a normal log statement.

PR #992, opened 2019-01-16
[#991] Make the 'heartbeat' ProtonClient option configurable

### LINKED ISSUE #991: Add configuration property to protocol adapters to configure AMQP heartbeat of their AMQP clients
Hi,

currently AMQP clients to messaging network (e.g. telemetry sender, event sender etc.) used in protocol adapters do not make use of AMQP connection idle timeout. This could be enabled by setting AMQP heartbeat in ProtonClientOptions.

Without this heartbeat it might happen that client does not discover broken connections in a timely manner (multiple minutes in case it relies on TCP timeouts). 

Would be great if AMQP heartbeat could be enabled and configured for protocol adapters. Perhaps it makes sense to enable them by default with some feasible default value.

calohmn:
ClientConfigProperties now supports configuring the heartbeat/idleTimeout value used for the ProtonClient.

Fixes #991.


calohmn:
As default for the heartbeat (and `idle-time-out` `open` frame value), I've used 8s for now. That is the same as the default in the Qpid dispatch router (on the listener side).

(To clarify: the default in the qdrouterd.conf docs is described as `idleTimeoutSeconds (integer, default=16)`. This 16s value is the *actual* timeout threshold. As per the AMQP spec, the `idle-time-out` advertised in the `open` frame is half of that, hence the 8s value.)

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -57,6 +61,7 @@
     private int connectTimeoutMillis = DEFAULT_CONNECT_TIMEOUT;
     private String credentialsPath;
     private long flowLatency = DEFAULT_FLOW_LATENCY;
+    private int heartbeatIntervalMillis = DEFAULT_HEARTBEAT_INTERVAL;
```
sophokles73: why don't we call it *idleTimeout* as the AMQP spec and QPid do?

calohmn: Ok, yes, makes sense to focus on the idleTimeout rather than the heartbeat. 

I would use the _actual_ idleTimeout here after which the connection gets closed.
The `idle-time-out` in the `open` frame and the heartbeat would be half of that then.

calohmn: I've amended the commit accordingly.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -593,4 +599,50 @@ public final void setConnectTimeout(final int connectTimeoutMillis) {
...
+    /**
+     * Sets the amount of time in milliseconds after which a connection will be closed
+     * when no frames have been received from the remote peer.
+     * <p>
+     * This property is also used to configure a heartbeat mechanism, checking that the connection is still alive.
+     * The corresponding heartbeat interval will be set to <em>idleTimeout/2</em> ms.
+     * <p>
+     * The default value of this property is {@link #DEFAULT_IDLE_TIMEOUT}.
+     *
+     * @param idleTimeoutMillis The idleTimeout in milliseconds.
```
sophokles73: please document the `IllegalArgumentException` as well

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -593,4 +599,50 @@ public final void setConnectTimeout(final int connectTimeoutMillis) {
...
+
+    /**
+     * Gets the interval in milliseconds in which frames are sent to the remote peer to check 
+     * that the connection is still alive.
+     * <p>
+     * This value is set to be half of {@link #getIdleTimeout()}.
+     *
+     * @return The heartbeatInterval in milliseconds.
+     */
+    public int getHeartbeatInterval() {
```
sophokles73: can you make this final?

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -593,4 +599,50 @@ public final void setConnectTimeout(final int connectTimeoutMillis) {
...
+     * when no frames have been received from the remote peer.
+     * <p>
+     * This property is also used to configure a heartbeat mechanism, checking that the connection is still alive.
+     * The corresponding heartbeat interval will be set to <em>idleTimeout/2</em> ms.
+     * <p>
+     * The default value of this property is {@link #DEFAULT_IDLE_TIMEOUT}.
+     *
+     * @return The idleTimeout in milliseconds.
+     */
+    public int getIdleTimeout() {
```
sophokles73: can you make this final?

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -593,4 +599,50 @@ public final void setConnectTimeout(final int connectTimeoutMillis) {
...
+     * when no frames have been received from the remote peer.
+     * <p>
+     * This property is also used to configure a heartbeat mechanism, checking that the connection is still alive.
+     * The corresponding heartbeat interval will be set to <em>idleTimeout/2</em> ms.
+     * <p>
+     * The default value of this property is {@link #DEFAULT_IDLE_TIMEOUT}.
+     *
+     * @param idleTimeoutMillis The idleTimeout in milliseconds.
+     */
+    public void setIdleTimeout(final int idleTimeoutMillis) {
```
sophokles73: can you make this final?

calohmn:
I've applied the changes concerning the above comments.

PR #993, opened 2019-01-16
Refactor device registry unit tests, so that they can be used against…

dejanb:
… multiple implementations

### DIFF
--- a/service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
+++ b/service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
@@ -1,0 +1,254 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.hono.service.credentials;
+
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.unit.Async;
+import io.vertx.ext.unit.TestContext;
+import org.eclipse.hono.util.CredentialsConstants;
+import org.junit.Test;
+
+import java.net.HttpURLConnection;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Abstract class used as a base for verifying behavior of {@link CompleteCredentialsService} in device registry implementations.
+ *
+ */
+public abstract class AbstractCompleteCredentialsServiceTest {
+
+    /**
+     * Gets credentials service being tested.
+     * @return The credentials service
+     */
+    public abstract CompleteCredentialsService getCompleteCredentialsService();
+
+    /**
+     * Verifies that only one set of credentials can be registered for an auth-id and type (per tenant).
+     *
+     * @param ctx The vert.x test context.
+     */
+    @Test
+    public void testAddCredentialsRefusesDuplicateRegistration(final TestContext ctx) {
+
+        register(getCompleteCredentialsService(), "tenant", "device", "myId", "myType", ctx);
+
+        final JsonObject payload2 = new JsonObject()
+                .put(CredentialsConstants.FIELD_PAYLOAD_DEVICE_ID, "other-device")
+                .put(CredentialsConstants.FIELD_AUTH_ID, "myId")
+                .put(CredentialsConstants.FIELD_TYPE, "myType")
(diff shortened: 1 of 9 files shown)


dejanb:
I would like to use tests in "abstract" classes against the new registry without duplicating them. Do you have a suggestion of a better way to do it?

dejanb:
Maybe just to split them, with a common small parent class that will init things?

sophokles73:
I think it's ok to use inheritance if we have only two or three subclasses in the same package. My understanding is that we will have one subclass per concrete registry implementation, right?

dejanb:
Yes. Although I think we should different modules for different registry implementation. I think that will give us more flexibility on how we package and deploy things and provides a good blueprint for others on how to do the same outside of our codebase.

sophokles73:
Hmm, then you are talking about putting the abstract test base classes into  a separate test artifact which you will then depend on from other modules?

dejanb:
New test artifact sounds a bit like an over-kill, but we could move them to the service-base where they will be tied to the interface definitions and other base classes. So that should be all you depend on and test against one specific version of service-base.

sophokles73:
But we would still new to create a test artifact in the service-base module which the other modules then depend on, right?

dejanb:
Any registry implementation needs to depend on service-base, so I don't think we need a separate artifact. With these couple of abstract test classes added, we're just giving people an easy way to write basic unit tests for their implementations. It won't be used outside this context.

sophokles73:
So you want to include abstract unit test base classes in the hono-service-base.jar artifact?

dejanb:
Sorry for the confusion, I was under impression that we already have one.

Based on all discussion, I think we should:

* Move base/abstract tests to hono-service-base module
* Create test-jar 
* Develop new registries in separate modules/projects that depend only on hono-service-base (and its test jar).

sophokles73:
sounds reasonable

dejanb:
@sophokles73 Done. Have a look at it when you find a moment.

service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCredentialsServiceTest.java
```diff
@@ -0,0 +1,259 @@
...
+ *
+ */
+@RunWith(VertxUnitRunner.class)
+abstract public class AbstractCredentialsServiceTest {
+
+    /**
+     * Gets credentials service being tested.
+     * @return The credentials service
+     */
+    abstract public CompleteBaseCredentialsService<Object> getCredentialsService();
```
sophokles73: `public abstract`?
It looks like using the interface `CompleteCredentialsService` as the return type should be ok, shouldn't it?

service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCredentialsServiceTest.java
```diff
@@ -0,0 +1,259 @@
...
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Tests verifying behavior of {@link CredentialsService}.
+ *
+ */
+@RunWith(VertxUnitRunner.class)
+abstract public class AbstractCredentialsServiceTest {
```
sophokles73: `public abstract`?

Should probably be renamed to `AbstractCompleteCredentialsServiceTest` because it tests all the optional functionality ...

service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
```diff
@@ -0,0 +1,257 @@
...
+import java.net.HttpURLConnection;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Abstract class used as a base for verifying behavior of {@link CompleteCredentialsService} in device registry implementations.
+ *
+ */
+@RunWith(VertxUnitRunner.class)
```
sophokles73: I would rather have the concrete test (sub-)classes contain this annotation in order to make it more explicit that it is really those (sub-)classes that implement the tests. WDYT?
Same applies to the timeout `@Rule` that is defined in one of the abstract base classes ...

service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
```diff
@@ -0,0 +1,257 @@
...
+                .put(CredentialsConstants.FIELD_AUTH_ID, authId)
+                .put(CredentialsConstants.FIELD_TYPE, type)
+                .put(CredentialsConstants.FIELD_SECRETS, secrets);
+
+        if (clientContext != null) {
+            data.mergeIn(clientContext);
+        }
+
+        final Async registration = ctx.async();
+       svc.add("tenant", data, ctx.asyncAssertSuccess(s -> {
```
sophokles73: indentation

service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
dejanb: Make sense. Pushed the changes.

PR #997, opened 2019-01-18
Fix wrong status code being used on released/modified delivery state

calohmn:


client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
```diff
@@ -447,7 +447,7 @@ public final boolean isRegistrationAssertionRequired() {
                         }
                     } else {
                         LOG.debug("message [message ID: {}] not accepted by peer: {}", messageId, remoteState);
-                        e = new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST);
+                        e = new ClientErrorException(HttpURLConnection.HTTP_UNAVAILABLE);
```
sophokles73: For the *released* outcome this is the right choice. However, if the transfer results in the *modified* outcome, the 400 code is quite accurate FMPOV. Can you add another `else if` branch for the *modified* outcome?

calohmn: Looking at the [spec](http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-modified) and for example this [qpid ticket](https://issues.apache.org/jira/browse/DISPATCH-366), I think we should look at the values of the _modified_ fields here:
````
delivery-failed	        count the transfer as an unsuccessful delivery attempt
undeliverable-here	prevent redelivery
````
So, if `undeliverable-here` is `true`, I think we should probably send back 400 `HTTP_BAD_REQUEST` (same as in `rejected` case). 
If `delivery-failed` is `true`, it looks to me as if we probably should return a status so that the client tries again - which would point to 503 `HTTP_UNAVAILABLE`. WDYT`?

sophokles73: I agree with using 503 for *delivery-failed* = `true`.
For *undeliverable-here* = `true` I can also imagine returning 404 to distinguish it from the case where the message itself is unprocessable (*rejected* case) and we return a 400 already.

calohmn: Ok, I've amended the commit. In the _modified_ _undeliverable-here_ = `true` case there's a 404, else for _modified_ there's 503.

PR #1001, opened 2019-01-22
Copy legal files to hono-service-base test jar

dejanb:


service-base/pom.xml
```diff
@@ -178,6 +178,30 @@
...
+      <!--
+        Copy legal documents from "target/classes" folder to "target/test-classes" folder
+        so that we make sure to include legal docs in test jar.
+       -->
+      <plugin>
+        <artifactId>maven-resources-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>copy_legal_docs_for_test_jar</id>
+            <phase>prepare-package</phase>
```
sophokles73: IMHO this execution relies on the legal docs already having been extracted to `${project.build.outputDirectory}/META-INF`. However, since both are bound to the *prepare-package* phase, I wonder if it is a fair assumption to make that the dependency plugin will *always* run before the resources plugin ...

dejanb: As far as I can see, the order is defined by the definition of the plugins and the parent ones will *always* be executed first, so it should be guaranteed to run in proper order.

Issue #961, opened 2018-12-13
Use revised metrics for command and control


kaniyan:
The names of metrics used when processing messages should be adapted as [discussed on the mailing list](https://www.eclipse.org/lists/hono-dev/msg01059.html)

sophokles73:
fixed in 13d21b34316e9b11e6fca3173d4e1ce393f2ebfa

PR #1005, opened 2019-01-25
[#961] Use "timer" meter for downstream messages.

sophokles73:
Use a Micrometer "timer" instead of a "counter" for the metrics reported
for downstream telemetry messages and events.
This way we can also monitor the avg time if takes to process messages
instead of tracking just the message rate itself.

The name of the metric has been changed to "hono.messages.received".
The following additional tags are used:
- status: indicates the processing status
- qos: the QoS level used for the message
- ttd: outcome of handling the TTD value provided by a device

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -35,15 +35,16 @@ import org.eclipse.hono.service.AbstractProtocolAdapterBase;
 import org.eclipse.hono.service.auth.DeviceUser;
 import org.eclipse.hono.service.http.DefaultFailureHandler;
 import org.eclipse.hono.service.http.HttpUtils;
+import org.eclipse.hono.service.metric.MetricsTags;
 import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.Constants;
-import org.eclipse.hono.util.EventConstants;
+import org.eclipse.hono.util.EndpointType;
 import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.ResourceIdentifier;
-import org.eclipse.hono.util.TelemetryConstants;
 import org.eclipse.hono.util.TenantObject;
 import org.springframework.beans.factory.annotation.Autowired;
 
+import io.micrometer.core.instrument.Timer.Sample;
 import io.opentracing.Span;
 import io.opentracing.contrib.vertx.ext.web.TracingHandler;
 import io.opentracing.contrib.vertx.ext.web.WebSpanDecorator;
@@ -72,9 +73,7 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
      */
     protected static final String DEFAULT_UPLOADS_DIRECTORY = "/tmp";
 
-    private static final int AT_LEAST_ONCE = 1;
-    private static final int HEADER_QOS_INVALID = -1;
-
+    private static final String KEY_MICROMETER_SAMPLE = "micrometer.sample";
     private static final String KEY_TIMER_ID = "timerId";
 
     private HttpServer server;
@@ -192,7 +191,8 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
     }
 
     /**
-     * Adds a handler for adding an OpenTracing Span to the routing context.
+     * Adds a handler for adding an OpenTracing {@code Span}
+     * and a Micrometer {@code Timer.Sample} to the routing context.
      *
      * @param router The router to add the handler to.
      * @param position The position to add the tracing handler at.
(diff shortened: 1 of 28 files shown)


site/content/api/Metrics.md
```diff
@@ -56,23 +56,26 @@ The names of Hono's standard components are as follows:
 
 Additional tags for protocol adapters are:
 
-| Tag        | Value                              | Description |
-| ---------- | ---------------------------------- | ----------- |
-| *tenant*   | *string*                           | The name of the tenant that the metric is being reported for |
-| *type*     | `telemetry`, `event`             | The type of message that the metric is being reported for |
+| Tag        | Value                                                  | Description |
+| ---------- | ------------------------------------------------------ | ----------- |
+| *qos*      | [ `N/A`, `0`, `1` ]                                   | The quality of service level used for a message. |
```
ctron: Does it make sense to add `N/A` as a value to this list? I would assume that if there is no QoS, then the tag would not be added.

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
```diff
@@ -13,11 +13,132 @@
...
 /**
  * A collector for metrics.
  */
 public interface Metrics {
 
+    /**
+     * The direction of a message.
+     *
+     */
+    enum Direction {
```
ctron: I would propose to use the convention of all uppercase enum literals.

sophokles73: Yes, I thought so too

site/content/api/Metrics.md
sophokles73: Not sure if this has any (negative) impact on e.g. query performance in Prometheus. I will try to find out ...

sophokles73: Prometheus has no problem with some time series instance vector not having a specific tag set while other instance vectors of the same metric name have it set. However, the Micrometer API does not allow to pass in a `null` value for a tag. Thus, it would require extra handling of unknown QoS value in the metrics code. IMHO there should be only very rare cases where a QoS value is unknown, so I'd rather keep the `N/A` value.

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
sophokles73: done

sophokles73:
@ctron I have pushed an update, can you take another look?

site/content/api/Metrics.md
ctron: So I think this could very easily be solved by adding an extra method to the enum, and letting it handle with the different cases.

Butting adding some arbitrary value (`N/A`) as a replacement of `null` sounds like a bad idea. I think not adding a value in the case there is none, is the right approach.

sophokles73: > So I think this could very easily be solved by adding an extra method to the enum, and letting it handle with the different cases.

Not sure what you mean by this. Something like `QoS.asTag()` which returns a `Tag` instance or `null`?

ctron: Either that or something like `QoS.applyTo(…)` if you don't want to add Micrometer APIs to the enum.

sophokles73: What would such a `QoS.applyTo()` method do and what input parameters would it have?

sophokles73:
@ctron I have made some changes which prevent the *qos* tag from being added if `UNKNOWN` ...

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -106,29 +119,80 @@ public long getNumberOfConnections() {
...
+        Objects.requireNonNull(timer);
+
+        if (!EndpointType.TELEMETRY.equals(type) && !EndpointType.EVENT.equals(type)) {
+            throw new IllegalArgumentException("invalid type, must be either telemetry or event");
+        }
 
+        timer.stop(this.registry.timer(
+                METER_MESSAGES_RECEIVED,
+                Tags.of(MetricsTags.getTypeTag(type))
+                    .and(MetricsTags.getTenantTag(tenantId))
```
ctron: I don't think calling "and" with `null` will work. Looking into the callstack of `Tags`, this seems to boil down to:

~~~java
Map<String, Tag> merged = new LinkedHashMap<>(this.tags);
tags.forEach(tag -> merged.put(tag.getKey(), tag));
return new Tags(merged);
~~~

Which should throw a NPE in the case of a `null` element.

sophokles73: Tags.and(Tag ...):
```
    /**
     * Return a new {@link Tags} instance by merging this collection and the specific tags.
     *
     * @param tags the tags to add
     * @return a new {@link Tags} instance
     */
    public Tags and(@Nullable Tag... tags) {
        if (tags == null || tags.length == 0) {
            return this;
        }
        return and(Arrays.asList(tags));
    }
```
this looks as if it would accept a `null` arg, doesn't it?

sophokles73: Yes, I ran the same test :-) With the same result (obviously).

sophokles73: @ctron  Based on this discussion I wonder if we also should not create a *ttd* tag for `TtdStatus.NONE`. WDYT?

sophokles73: @ctron Given that we are reporting every single message being sent, I think that we should also probably not instantiate new Tag instances for each message. Instead, we could create the Tag instances as part of the enums and re-use the instances, thus reducing GC pressure. Do you think that would be worth the effort?

ctron: I agree, that might be a good idea.

sophokles73:
@ctron Ok, I have made some changes based on our discussion:
* the *ttd* and *qos* tags are only added if they have a *real* value
* the enums now contain `Tag` instances which are reused when reporting meter values

can you take another look?

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -544,30 +560,37 @@ public final void uploadEventMessage(final RoutingContext ctx, final String tena
...
+            final EndpointType endpoint) {
 
         if (!isPayloadOfIndicatedType(payload, contentType)) {
             HttpUtils.badRequest(ctx, String.format("content type [%s] does not match payload", contentType));
         } else {
             final String qosHeaderValue = ctx.request().getHeader(Constants.HEADER_QOS_LEVEL);
-            final Integer qos = getQoSLevel(qosHeaderValue);
-            if (qos != null && qos == HEADER_QOS_INVALID) {
+            final MetricsTags.QoS qos = getQoSLevel(qosHeaderValue);
+            if (MetricsTags.QoS.UNKNOWN.equals(qos)) {
```
ctron: As enums are guaranteed to have only one instance per literal, you can do a comparison of the object reference, which should be faster than calling the equals method.

sophokles73: will do

sophokles73:
merged in b4a31e6bf23ed6b0bffae8d958ce5eb0365c64a5

PR #1009, opened 2019-01-28
Example device registry overwrites existing config when restarting in openshift

BobClaerhout:
Fix for issue https://github.com/eclipse/hono/issues/1007

PR #1011, opened 2019-01-29
k8s: Don't overwrite existing device registry config

ghys:
Related to #1009.

Signed-off-by: Yannick Schaus <eclipse@schaus.net>

### DIFF
--- a/deploy/src/main/deploy/resource-descriptors/hono-service-device-registry/hono-service-device-registry-deployment.yaml
+++ b/deploy/src/main/deploy/resource-descriptors/hono-service-device-registry/hono-service-device-registry-deployment.yaml
@@ -73,8 +73,8 @@ spec:
       - command:
         - sh
         - -c
-        - cp -u /tmp/hono/example-credentials.json /var/lib/hono/device-registry/credentials.json;
-          cp -u /tmp/hono/example-tenants.json /var/lib/hono/device-registry/tenants.json
+        - cp -n /tmp/hono/example-credentials.json /var/lib/hono/device-registry/credentials.json;
+          cp -n /tmp/hono/example-tenants.json /var/lib/hono/device-registry/tenants.json
         image: busybox
         name: copy-example-data
         volumeMounts:


deploy/src/main/deploy/resource-descriptors/hono-service-device-registry/hono-service-device-registry-deployment.yaml
```diff
@@ -73,8 +73,8 @@ spec:
       - command:
         - sh
         - -c
-        - cp -u /tmp/hono/example-credentials.json /var/lib/hono/device-registry/credentials.json;
-          cp -u /tmp/hono/example-tenants.json /var/lib/hono/device-registry/tenants.json
+        - cp -n /tmp/hono/example-credentials.json /var/lib/hono/device-registry/credentials.json;
```
sophokles73: this will not work because the `cp` command in the busybox image does not support the `-n` switch:
```
/tmp # cp -n test test2
cp: invalid option -- n
BusyBox v1.30.0 (2018-12-31 18:16:17 UTC) multi-call binary.

Usage: cp [OPTIONS] SOURCE... DEST

Copy SOURCE(s) to DEST

	-a	Same as -dpR
	-R,-r	Recurse
	-d,-P	Preserve symlinks (default if -R)
	-L	Follow all symlinks
	-H	Follow symlinks on command line
	-p	Preserve file attributes if possible
	-f	Overwrite
	-i	Prompt before overwrite
	-l,-s	Create (sym)links
	-T	Treat DEST as a normal file
	-u	Copy only newer files
```

sophokles73:
@ghys do you have any plans to update this PR or should I close it?

ghys:
@sophokles73 oh right, this fell off my radar... Sorry.
I just wanted to reproduce #1009 for k8s since I've been bitten by this as well, but I should have checked.
The workaround is to either:
- remove the initContainer part from the device registry deployment
- remove the example tenants & credentials from the secret (`hono-service-device-registry-conf`).

PR #1014, opened 2019-01-29
[#1013] Fix wrong exception being thrown on SASL handshake failure.

### LINKED ISSUE #1013: HonoClient: Wrong exception when using invalid SASL credentials
Since the update to vert.x 3.6, there is a different exception being thrown in case of a failed SASL authentication attempt. See https://github.com/vert-x3/vertx-proton/issues/69.

The corresponding change in Hono is still missing. This leads to a `ServerErrorException` (with `HTTP_UNAVAILABLE`/"failed to connect") being thrown. It should be a `ClientErrorException` (`HTTP_UNAUTHORIZED`) instead.

calohmn:
This fixes #1013.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -30,6 +30,8 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
+
+import javax.security.sasl.AuthenticationException;
 
 import org.apache.qpid.proton.amqp.Symbol;
 import org.apache.qpid.proton.message.Message;
@@ -65,6 +67,7 @@ import io.vertx.core.Vertx;
 import io.vertx.proton.ProtonClientOptions;
 import io.vertx.proton.ProtonConnection;
 import io.vertx.proton.ProtonDelivery;
+import io.vertx.proton.sasl.SaslSystemException;
 
 /**
  * A helper class for creating Vert.x based clients for Hono's arbitrary APIs.
@@ -388,7 +391,11 @@ public class HonoClientImpl implements HonoClient {
                         conAttempt -> {
                             connecting.compareAndSet(true, false);
                             if (conAttempt.failed()) {
-                                reconnect(conAttempt.cause(), connectionHandler, disconnectHandler);
+                                if (isConnectionFailureEntailingReconnectAttempt(conAttempt.cause())) {
+                                    reconnect(conAttempt.cause(), connectionHandler, disconnectHandler);
+                                } else {
+                                    failConnectionAttempt(conAttempt.cause(), connectionHandler);
+                                }
                             } else {
                                 final ProtonConnection newConnection = conAttempt.result();
                                 if (shuttingDown.get()) {
@@ -497,18 +504,7 @@ public class HonoClientImpl implements HonoClient {
             log.debug("max number of attempts [{}] to re-connect to peer [{}:{}] have been made, giving up",
                     clientConfigProperties.getReconnectAttempts(), connectionFactory.getHost(), connectionFactory.getPort());
             clearState();
-            if (connectionFailureCause == null) {
-                connectionHandler.handle(Future.failedFuture(
-                        new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "failed to connect")));
-            } else if (connectionFailureCause instanceof SecurityException) {
-                // SASL handshake failed continuously, maybe due to wrong credentials?
-                connectionHandler.handle(Future.failedFuture(
-                        new ClientErrorException(HttpURLConnection.HTTP_UNAUTHORIZED, "failed to authenticate with server")));
-            } else {
-                connectionHandler.handle(Future.failedFuture(
-                        new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "failed to connect",
-                                connectionFailureCause)));
-            }
+            failConnectionAttempt(connectionFailureCause, connectionHandler);
 
         } else {
             if (connectionFailureCause != null) {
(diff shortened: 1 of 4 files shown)


calohmn:
As a further differentiation, the `SaslSystemException.permanent` flag could be checked, maybe throwing a `ServerErrorException` with `HTTP_UNAVAILABLE` ("failed to connect") if `permanent` is false. But I'm not sure about that.

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -500,7 +501,7 @@ private void reconnect(
             if (connectionFailureCause == null) {
                 connectionHandler.handle(Future.failedFuture(
                         new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "failed to connect")));
-            } else if (connectionFailureCause instanceof SecurityException) {
+            } else if (connectionFailureCause instanceof SaslSystemException) {
```
sophokles73: it is true that the behavior in vertx-proton has changed. However, `ProtonSaslClientAuthenticatorImpl` uses different exceptions to indicate different types of failures:
```
  private void handleSaslFail() throws SaslException {
    switch(sasl.getOutcome()) {
    case PN_SASL_AUTH:
      throw new AuthenticationException("Failed to authenticate");
    case PN_SASL_SYS:
    case PN_SASL_TEMP:
      throw new SaslSystemException(false, "SASL handshake failed due to a transient error");
    case PN_SASL_PERM:
      throw new SaslSystemException(true, "SASL handshake failed due to an unrecoverable error");
    default:
      throw new SaslException("SASL handshake failed");
    }
  }
```
In particular, the most probable cause for failure (wrong credentials) is indicated by a `javax.security.sasl.AuthenticationException` (a subclass of `javax.security.sasl.SaslException`).

So, I think that we should catch `javax.security.sasl.SaslException` here instead of `SecurityException` and add another catch clause for `SaslSystemException` which is used to indicate cases where the server is not ready to process the handshake and fail with a `ServerErrorException` in that case.

calohmn: right, I now see I was concentrating on only one case, the one thrown in `handleSaslInit` if no SASL mechanism is found: 
````
throw new SaslSystemException(
            true, "Could not find a suitable SASL mechanism for the remote peer using the available credentials.");
````

A further note about this:
That case can also occur if the PLAIN mechanism is used and no username/password was given. 
In general, I would prefer to throw a `ClientErrorException` in that case - also keeping the behaviour from before the vert.x update. WDYT?

(Unfortunately, to distinguish that case from the `SASL handshake failed due to an unrecoverable error` case, a match on the exception text is required. We could add a small integration test to be safe concerning future vert.x changes here.)

Comment (sophokles73) [reply client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java]: Giving it some more thought, I think that the `reconnect()` method in general should
* fail the attempt with a `ClientErrorException` (e.g. 401) in case of an `AuthenticationException`
* retry to connect in case of a `SaslSystemException` with *permanent* being `false`
* fail the attempt with a `ServerErrorException` (e.g. 503) in case of a `SaslSystemException` with *permanent* being `true`
* retry to connect in case of another `SaslException`

Regarding the case where no SASL mechanism is found, I agree that this should cause a `ClientErrorException` and that we would need to match on the exception's description (for the time being). I think that proton-vertx should throw an `AuthenticationException` in this case instead of a `SaslSystemException`. I will create a PR for that. Thanks for spotting this :+1:

Comment (calohmn) [reply client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java]: I agree that it's better to throw `AuthenticationException` in `handleSaslInit`.

Regarding your list of cases above:
By "fail the attempt", do you mean fail the overall connection attempt, ignoring the `reconnectAttempts`, and doing no further connection attempts?

In the "retry to connect" cases, do you mean reconnect without taking the `reconnectAttempts` into account, potentially doing reconnect attempts indefinitely?

Comment (calohmn) [reply client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java]: I've updated the commit - now failing directly on `AuthenticationException` and permanent `SaslSystemException`.

File: client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java [hunk 2]
```diff
@@ -525,6 +521,33 @@ private void reconnect(
...
+    private void failConnectionAttempt(final Throwable connectionFailureCause, final Handler<AsyncResult<HonoClient>> connectionHandler) {
+        if (connectionFailureCause == null) {
+            connectionHandler.handle(Future.failedFuture(
+                    new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "failed to connect")));
+        } else if (connectionFailureCause instanceof AuthenticationException) {
+            // SASL handshake failed continuously, maybe due to wrong credentials?
+            connectionHandler.handle(Future.failedFuture(
+                    new ClientErrorException(HttpURLConnection.HTTP_UNAUTHORIZED, "failed to authenticate with server")));
+        } else if (connectionFailureCause instanceof SaslSystemException
+                && connectionFailureCause.getMessage().contains("Could not find a suitable SASL mechanism")) {
```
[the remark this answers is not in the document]
calohmn: ok, I've added a comment and additional checks in the unit tests for the exception message.

tests/src/test/java/org/eclipse/hono/tests/client/HonoClientImplIT.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Verifies that a connection attempt where no credentials are given fails
+     * immediately with a ClientErrorException with status code 401.
+     * 
+     * @param ctx The vert.x test context.
+     */
+    @Test
+    public void testConnectFailsWithClientErrorForNoSASLMechanismException(final TestContext ctx) {
```
sophokles73: IMHO we should move this test to `org.eclipse.hono.tests.amqp.AmqpConnectionIT` because
* we already perform similar tests there as well and
* the AMQP adapter is a component under our own control which we can make sure behaves in the way we expect it to

client/src/test/java/org/eclipse/hono/client/impl/HonoClientImplTest.java
```diff
@@ -150,6 +153,97 @@ public void testConnectFailsWithClientErrorForSecurityException(final TestContex
...
+    }
+
+    /**
+     * Verifies that the client fails with a ServerErrorException with status code 503
+     * if it cannot authenticate to the server because of a transient error.
+     *
+     * @param ctx The vert.x test client.
+     */
+    @Test
+    public void testConnectFailsWithClientErrorForImpermanentSaslSystemException(final TestContext ctx) {
```
sophokles73: `testConnectFailsWithClientErrorForTransientSaslSystemException`

calohmn: I've changed that.

tests/src/test/java/org/eclipse/hono/tests/client/HonoClientImplIT.java
calohmn: At first glance, I also thought so - seeing the already existing `testConnectFailsForWrongCredentials` method there.
But, this method is using the `ProtonClient` directly instead of the `HonoClient`. Of course this could be changed, but that would mean either having this one case (plus the new test method) where the `AmqpAdapterTestBase.connectToAdapter` method is not used in favour of a HonoClient-using method. Or it would mean adapting all invocations of `AmqpAdapterTestBase.connectToAdapter` - these are quite many.

So, I think this test doesn't fit that well there after all, since the focus in the `amqp` package is on the AMQP adapter.

Having a separate `HonoClientImplIT` in a separate package makes it clearer that this test is about `HonoClient` functionality. And it also offers the possibility of adding various other test methods for testing the `HonoClientImpl` there later on.
If the test class is already there, the hurdle of actually adding an integration test later on is lower.

calohmn:
A note about the change here of not doing a `reconnect` any more in all cases:
I think we have to take care to not cause issues with this in cases where code relies on infinite `reconnect` attempts and where there is the "Authentication failed" SASL outcome being returned although that is wrong in that particular case (e.g. in a timeout scenario).
One possible issue concerning `HonoSaslAuthenticator` is described here: https://github.com/eclipse/hono/issues/1015#issuecomment-459111757.

tests/src/test/java/org/eclipse/hono/tests/client/HonoClientImplIT.java
sophokles73: I still don't like it. In particular, we are not verifying any specific behavior of Hono here, so I do not see why we should have an *integration test* for `HonoClientImpl`. FMPOV all of the client's expected behavior can easily be tested using unit tests (which you already do).
Anyway, I will merge it for the time being. Let's see how it evolves ...

calohmn:
I've added a note in the release notes.

tests/src/test/java/org/eclipse/hono/tests/client/HonoClientImplIT.java
calohmn: > FMPOV all of the client's expected behavior can easily be tested using unit tests

The thing here is that the client's behavior is also defined by the underlying vert.x-proton.
The plain unit test relies on mocks to mimic the behavior of vert.x proton and qpid. If that behavior changes with an update (e.g. throwing an `AuthenticationException` in `handleSaslInit`) the unit test possibly is still green. The integration test won't be in that case.

PR #1016, opened 2019-01-31
Fix unpack hono-demo-certs docker deploy

kaizimmerm:
Hi,

I was running https://www.eclipse.org/hono/getting-started/ and the swarm deploy script did not work due to missing certificates.

It seems that by adding helm you overrit `outputDirectory`. I changed it by copy into helm directory instead.

Let me know if that works for you.

Kai


sophokles73:
Hi Kai,

this has been fixed in b0bf433aa58e3275768750b3ee542bf108cc3dac already ...
Thanks anyway for creating a CQ :+1:

PR #1017, opened 2019-02-01
[#1015] Use error codes for failed AuthenticationService requests

### LINKED ISSUE #1015: AuthService: use error codes for failed authentication attempts
This is relevant for a scenario where a non-standard `AuthenticationService` component is used (i.e. not the `FileBasedAuthenticationService`).

Imagine an `AuthenticationService` implementation based on a database. When that database is down temporarily, authentication requests should not be answered with the standard "authentication failed" response but instead with an exception that points out the temporary nature of the error.
For such a authentication exception, the `HonoClient` should retry the authentication attempt (i.e. retry the connection attempt with the SASL handshake).

This relates to the different SASL outcomes now taken into account by the `HonoClient` (see #1014) and the place in `HonoSaslAuthenticator` where the `AuthenticationService` is invoked and its response is used to set a SASL outcome - currently this is only PN_SASL_OK or PN_SASL_AUTH.

---
The one part of a potential fix for this could be in all classes implementing `AuthenticationService`, see the FIXME comment here:
````
     * @param authRequest The authentication information provided by the client in the SASL exchange.
     * @param authenticationResultHandler The handler to invoke with the authentication result. If authentication succeeds,
     *                                    the result contains the authenticated user.
     *                                    FIXME describe ServiceInvocationException (with its status field) as the preferred exception type here and use it in all implementing classes.
     */
    void authenticate(JsonObject authRequest, Handler<AsyncResult<HonoUser>> authenticationResultHandler);
````

The other part would be in `HonoSaslAuthenticator#process`:
````
authTracker.setHandler(s -> {
    if (s.succeeded()) {

        final HonoUser user = s.result();
        LOG.debug("authentication of client [authorization ID: {}] succeeded", user.getName());
        Constants.setClientPrincipal(protonConnection, user);
        succeeded = true;
        sasl.done(SaslOutcome.PN_SASL_OK);

    } else {
        // FIXME add support for setting these kinds of outcomes based on the exception (should be a ServiceInvocationException):
        //       PN_SASL_SYS failed due to a system error
        //       PN_SASL_TEMP impermanent error
        //       PN_SASL_PERM failed due to unrecoverable error
        //       see ProtonSaslClientAuthenticatorImpl

        LOG.debug("authentication failed: " + s.cause().getMessage());
        sasl.done(SaslOutcome.PN_SASL_AUTH);
    }
````

### PR DESCRIPTION
**Pull request by calohmn**
This fixes #1015.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/AuthenticationServerClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/AuthenticationServerClient.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -13,7 +13,10 @@
 
 package org.eclipse.hono.client;
 
+import java.net.HttpURLConnection;
 import java.util.Objects;
+
+import javax.security.sasl.AuthenticationException;
 
 import org.eclipse.hono.auth.HonoUser;
 import org.eclipse.hono.auth.HonoUserAdapter;
@@ -31,6 +34,7 @@ import io.vertx.proton.ProtonClientOptions;
 import io.vertx.proton.ProtonConnection;
 import io.vertx.proton.ProtonMessageHandler;
 import io.vertx.proton.ProtonReceiver;
+import io.vertx.proton.sasl.SaslSystemException;
 
 /**
  * A client for retrieving a token from an authentication service via AMQP 1.0.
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by calohmn** [comment]
I'll try to split this in multiple PRs.

First one is #1031.

File: client/src/main/java/org/eclipse/hono/client/AuthenticationServerClient.java [hunk 1]
```diff
@@ -90,7 +92,9 @@ public void verifyPlain(final String authzid, final String authcid, final String
         options.addEnabledSaslMechanism(AuthenticationConstants.MECHANISM_PLAIN);
         factory.connect(options, authcid, password, null, null, conAttempt -> {
             if (conAttempt.failed()) {
-                authenticationResultHandler.handle(Future.failedFuture("cannot connect to Authentication service"));
+                authenticationResultHandler
+                        .handle(Future.failedFuture(new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE,
```
sophokles73: Is there a reason why we *swallow* the root cause for the failure to connect? The attempt might have been failed with a `javax.security.sasl.AuthenticationException` or a `io.vertx.proton.sasl.SaslSystemException` so we might be able to fail the future with a much more specific exception ...

calohmn: Right, now we can differentiate further. The mechanism to get the correct `ServiceInvocationException` would be the same as in `HonoClientImpl#failConnectionAttempt`. 
I've adopted it from there, with modified exception messages.
(It might make sense to use a common method, I'm not sure where to put that though. And we might want to keep the different exception messages.)

sophokles73:
@calohmn can you please rebase?

Issue #775, opened 2018-08-13
Support tenant-scoped command response link in HonoClient


pellmann:
With #708 we allowed command response links, that are not scoped for a device but for a tenant (or min. for tenant). Currently we do not yet support this case in HonoClient.

In the current variant, the command is sent to a device and the device responds directly. From the HonoClient's point of view, the application instance that sent the command also receives the answer directly and can continue working with it or resend the command in the event of an error.

In the variant with one response-link pro tenant it is not clear which instance of the application receives the response. The delivery is the same as for events and any instance that listens for responses can get the response. This makes sense if the response can be delivered to a central component (such as a persistence layer), as e.g. in Ditto. The direct connection between request and response is lost in this case.

DanielMaier-BSI:
Fixed with #1028. I think we can close this issue.

PR #1028, opened 2019-02-07
Fix for #775 Support tenant-scoped command response link in HonoClient

DanielMaier-BSI:
This PR is not intended do be merged (yet), but to discuss API for fix of #775!

Idea is to have a consumer that receives all command responses to address control/<tenantId>/<replyId> and to have a sender to send commands with reply-to set to control/<tenantId>/<replyId>.

I am fine with the consumer part, but I am not happy with the sender part regarding 2 points myself:
- I not have found a better name as "sendCommandWithTenantScopedResponse" in the CommandClient interface to send commands with according reply-to
- If someone creates a command client without setting replyId explicitly he ends up with a CommandClient that sends commands with tenant scoped responses with a random reply-to and therefore is not able to receive the response at all.

WDYT?

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/AsyncCommandClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/AsyncCommandClient.java
@@ -1,0 +1,92 @@
+/**
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.hono.client;
+
+import java.util.Map;
+
+import io.vertx.core.Future;
+import io.vertx.core.buffer.Buffer;
+
+/**
+ * A client for sending asynchronous request response commands.
+ * <p>
+ * An instance of this interface is always scoped to a specific tenant and device.
+ * </p>
+ *
+ * @see org.eclipse.hono.client.HonoClient#createAsyncCommandResponseConsumer(String, String,
+ *      java.util.function.Consumer, io.vertx.core.Handler)
+ * @see org.eclipse.hono.client.HonoClient#createAsyncCommandResponseConsumer(String, String,
+ *      java.util.function.BiConsumer, io.vertx.core.Handler)
+ */
+public interface AsyncCommandClient extends MessageSender {
+
+    /**
+     * Sends an async command to a device, i.e. there is no immediate response expected from the device, but
+     * asynchronously via a separate consumer.
+     * <p>
+     * A device needs to be (successfully) registered before a client can upload any data for it. The device also needs
+     * to be connected for a successful delivery.
+     *
+     * @param command The command name.
+     * @param data The command data to send to the device or {@code null} if the command has no input data.
+     * @param correlationId The identifier to use for correlating the response with the request. Note: This parameter
+     * is security sensitive. To ensure secure request response mapping choose correlationId carefully, e.g.
+     * {@link java.util.UUID#randomUUID()}.
+     * @param replyId An arbitrary string which gets used for response link address in the form of <em>control/${
+     * tenantId}/${replyId}</em> Must match the {@code replyId} passed to the command response receiver, see also below.
+     * @return A future indicating the result of the operation.
+     *         <p>
+     *         If the command was accepted, the future will succeed.
+     *         <p>
+     *         The future will fail with a {@link ServiceInvocationException} if the command could not be forwarded to
+     *         the device.
+     * @throws NullPointerException if command, correlationId or replyId is {@code null}.
+     * @see org.eclipse.hono.client.HonoClient#createAsyncCommandResponseConsumer(String, String,
+     *      java.util.function.Consumer, io.vertx.core.Handler)
+     * @see org.eclipse.hono.client.HonoClient#createAsyncCommandResponseConsumer(String, String,
+     *      java.util.function.BiConsumer, io.vertx.core.Handler)
+     */
+    Future<Void> sendAsyncCommand(String command, Buffer data, String correlationId, String replyId);
+
+    /**
+     * Sends an async command to a device, i.e. there is no immediate response expected from the device, but
+     * asynchronously via a separate consumer.
+     * <p>
+     * A device needs to be (successfully) registered before a client can upload any data for it. The device also needs
+     * to be connected for a successful delivery.
+     *
+     * @param command The command name.
+     * @param contentType The type of the data submitted as part of the command or {@code null} if unknown.
+     * @param data The command data to send to the device or {@code null} if the command has no input data.
+     * @param correlationId The identifier to use for correlating the response with the request. Note: This parameter
+     * is security sensitive. To ensure secure request response mapping choose correlationId carefully, e.g.
+     * {@link java.util.UUID#randomUUID()}.
+     * @param replyId An arbitrary string which gets used for response link address in the form of <em>control/${
+     * tenantId}/${replyId}</em>. Must match the {@code replyId} passed to the command response receiver, see also below.
+     * @param properties The headers to include in the command message as AMQP application properties
+     * @return A future indicating the result of the operation.
+     *         <p>
+     *         If the command was accepted, the future will succeed.
+     *         <p>
+     *         The future will fail with a {@link ServiceInvocationException} if the command could not be forwarded to
+     *         the device.
+     * @throws NullPointerException if command, correlationId or replyId is {@code null}.
+     * @see org.eclipse.hono.client.HonoClient#createAsyncCommandResponseConsumer(String, String,
+     *      java.util.function.Consumer, io.vertx.core.Handler)
+     * @see org.eclipse.hono.client.HonoClient#createAsyncCommandResponseConsumer(String, String,
+     *      java.util.function.BiConsumer, io.vertx.core.Handler)
+     */
+    Future<Void> sendAsyncCommand(String command, String contentType, Buffer data, String correlationId, String replyId,
+            Map<String, Object> properties);
+}
--- a/client/src/main/java/org/eclipse/hono/client/HonoClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/HonoClient.java
@@ -214,7 +214,7 @@ public interface HonoClient extends ConnectionLifecycle,
 
     /**
      * {@inheritDoc}
-     * 
+     *
      * To re-connect to the server, an explicit call to {@code #connect()} should be made.
      * Unlike {@code #shutdown()}, which does not allow to connect back to the server,
      * this method allows to connect back to the server.
--- a/client/src/main/java/org/eclipse/hono/client/impl/AsyncCommandClientImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/AsyncCommandClientImpl.java
@@ -1,0 +1,150 @@
+/**
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.hono.client.impl;
+
+import java.util.Map;
+import java.util.Objects;
+
+import org.apache.qpid.proton.message.Message;
+import org.eclipse.hono.client.AsyncCommandClient;
+import org.eclipse.hono.config.ClientConfigProperties;
+import org.eclipse.hono.util.CommandConstants;
+import org.eclipse.hono.util.MessageHelper;
+
+import io.opentracing.Span;
+import io.opentracing.SpanContext;
+import io.opentracing.tag.Tags;
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.proton.ProtonConnection;
+import io.vertx.proton.ProtonDelivery;
+import io.vertx.proton.ProtonHelper;
+import io.vertx.proton.ProtonQoS;
+import io.vertx.proton.ProtonSender;
+
+/**
+ * Sender for commands with asynchronous responses.
+ */
+public class AsyncCommandClientImpl extends AbstractSender implements AsyncCommandClient {
+
+    private AsyncCommandClientImpl(
+            final ClientConfigProperties config,
+            final ProtonSender sender,
+            final String tenantId,
+            final String targetAddress,
+            final Context context) {
+        super(config, sender, tenantId, targetAddress, context, null);
+    }
+
+    @Override
+    protected Future<ProtonDelivery> sendMessage(final Message message, final Span currentSpan) {
+        return sendMessageAndWaitForOutcome(message, currentSpan);
+    }
+
+    @Override
+    protected Span startSpan(final SpanContext parent, final Message message) {
+        if (tracer == null) {
+            throw new IllegalStateException("no tracer configured");
+        } else {
+            final Span span = newFollowingSpan(parent, "sending async command");
+            Tags.SPAN_KIND.set(span, Tags.SPAN_KIND_PRODUCER);
+            return span;
+        }
+    }
+
+    @Override
+    protected String getTo(final String deviceId) {
+        return null;
+    }
+
+    @Override
+    public String getEndpoint() {
+        return CommandConstants.COMMAND_ENDPOINT;
+    }
+
+    static String getTargetAddress(final String tenantId, final String deviceId) {
+        return String.format("%s/%s/%s", CommandConstants.COMMAND_ENDPOINT, tenantId, deviceId);
+    }
+
+    @Override
+    public Future<ProtonDelivery> sendAndWaitForOutcome(final Message message) {
+        return send(message);
+    }
+
+    @Override
+    public Future<Void> sendAsyncCommand(final String command, final Buffer data, final String correlationId,
+            final String replyId) {
+        return sendAsyncCommand(command, null, data, correlationId, replyId, null);
+    }
+
+    @Override
+    public Future<Void> sendAsyncCommand(final String command, final String contentType, final Buffer data,
+            final String correlationId, final String replyId, final Map<String, Object> properties) {
+        Objects.requireNonNull(command);
+        Objects.requireNonNull(correlationId);
+        Objects.requireNonNull(replyId);
+
+        final Message message = ProtonHelper.message();
+        message.setCorrelationId(correlationId);
+        MessageHelper.setCreationTime(message);
+        MessageHelper.setPayload(message, contentType, data);
+        message.setSubject(command);
+        final String replyToAddress = String.format("%s/%s/%s", CommandConstants.COMMAND_ENDPOINT, tenantId, replyId);
+        message.setReplyTo(replyToAddress);
+
+        return sendAndWaitForOutcome(message).compose(ignore -> Future.succeededFuture());
+    }
+
+    /**
+     * Creates a new async command client for a tenant and device.
+     * <p>
+     * The instance created is scoped to the given device. In particular, the sender link's target address is set to
+     * <em>control/${tenantId}/${deviceId}</em>.
+     *
+     * @param context The vert.x context to run all interactions with the server on.
+     * @param clientConfig The configuration properties to use.
+     * @param con The AMQP connection to the server.
+     * @param tenantId The tenant that the device belongs to.
+     * @param deviceId The device to create the client for.
+     * @param closeHook A handler to invoke if the peer closes the sender link unexpectedly.
+     * @param creationHandler The handler to invoke with the outcome of the creation attempt.
+     * @throws NullPointerException if any of context, clientConfig, con, tenantId, deviceId or creationHandler are
+     *             {@code null}.
+     */
+    public static void create(
+            final Context context,
+            final ClientConfigProperties clientConfig,
+            final ProtonConnection con,
+            final String tenantId,
+            final String deviceId,
+            final Handler<String> closeHook,
+            final Handler<AsyncResult<AsyncCommandClient>> creationHandler) {
+
+        Objects.requireNonNull(context);
+        Objects.requireNonNull(clientConfig);
+        Objects.requireNonNull(con);
+        Objects.requireNonNull(tenantId);
+        Objects.requireNonNull(deviceId);
+        Objects.requireNonNull(creationHandler);
+
+        final String targetAddress = AsyncCommandClientImpl.getTargetAddress(tenantId, deviceId);
+        createSender(context, clientConfig, con, targetAddress, ProtonQoS.AT_LEAST_ONCE, closeHook)
+                .compose(sender -> Future.<AsyncCommandClient> succeededFuture(
+                        new AsyncCommandClientImpl(clientConfig, sender, tenantId, targetAddress, context)))
+                .setHandler(creationHandler);
+    }
+}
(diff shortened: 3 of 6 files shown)


client/src/main/java/org/eclipse/hono/client/CommandClient.java
```diff
@@ -106,4 +106,8 @@
      * @see RequestResponseClient#setRequestTimeout(long)
      */
     Future<Void> sendOneWayCommand(String command, String contentType, Buffer data, Map<String, Object> properties);
+
+    Future<Void> sendCommandWithTenantScopedResponse(String command, Buffer data);
```
sophokles73: what about *sendAsyncCommand*? We could add a parameter for the *replyId*. That way a client could not end up in the situation that you described, i.e. sending an async command using a random replyId ...

client/src/main/java/org/eclipse/hono/client/HonoClient.java
```diff
@@ -338,6 +338,13 @@
     Future<MessageConsumer> createEventConsumer(String tenantId, BiConsumer<ProtonDelivery, Message> eventConsumer,
             Handler<Void> closeHandler);
 
+    Future<MessageConsumer> createCommandResponseConsumer(String tenantId, String replyId, Consumer<Message> commandResponseConsumer,
```
sophokles73: *createAsyncCommandResponseConsumer*?

DanielMaier-BSI: I am fine with the naming :)

client/src/main/java/org/eclipse/hono/client/CommandClient.java
DanielMaier-BSI: I am fine with the naming :). 

If we pass the replyId here what do you suggest to do with clients that are created with specific replyId, just ignore?

sophokles73: Yes, we would need to document this behavior in this method's JavaDoc, of course.

sophokles73:
@DanielMaier-BSI are you going to update this PR accordingly?

DanielMaier-BSI:
yes.

DanielMaier-BSI:
Hi @sophokles73 ,

after looking into the implementation details the following questions have arisen:

- Do we also need to be able to provide correlation-id in sendAsyncCommand? Otherwise cosnumer cannot correlate request to response?
- I am struggling with the inheritance hierarchy. CommandClientImpl extends AbstractRequestResponseClient, but now we have commands without response from ComandClient point of view. We already have this case with one way commands. There the expected response gets just removed from reply map when no replyTo is set in message. I have no good idea for such a quick "workaround" for async commands. Refactoring AbstractRequestResponseClient to also be able to send requests without response sounds a little bit awkward. Is there already some kind of client or base class to just send request without expecting responses?

sophokles73:
@DanielMaier-BSI 

> Do we also need to be able to provide correlation-id in sendAsyncCommand? Otherwise cosnumer cannot correlate request to response?

Not necessarily. I can imagine two variants:
1. let a client supply a correlation ID explicitly which is then set on the outgoing AMQP message's *correlation-id* property.
1. implicitly use the outbound message's *message-id* as the correlation ID (this is done by the protocol adapters automatically if the command message has no *correlation-id* property.

In both cases, the future returned by *sendAsync* could be of type `String` instead of `Void` and contain the correlation ID being used so that the client can store it somewhere for correlating the response later on.

> * CommandClientImpl extends AbstractRequestResponseClient, but now we have commands without response from ComandClient point of view. We already have this case with one way commands. There the expected response gets just removed from reply map when no replyTo is set in message. I have no good idea for such a quick "workaround" for async commands. Refactoring AbstractRequestResponseClient to also be able to send requests without response sounds a little bit awkward. Is there already some kind of client or base class to just send request without expecting responses?

I agree. I have also stumbled across this while thinking a little more about it. I guess we could create a separate `AsyncCommandClient` which extends `AbstractSender` instead of `AbstractRequestResponseClient`. This way we could also (elegantly) get rid of the *replyId* which needs to be specified upon creation of the client but would be ignored when doing an *asyncSend*. WDYT?

DanielMaier-BSI:
Yes I think this is a step into the right direction. However I have some concerns regarding `AsyncCommandClientImpl`. To reuse all the sender handling in hono client `AsyncCommandClientImpl` would need to implement `MessageSender`, but `MessageSender` has a lot of methods that we do not need at all for `AsyncCommandClient`. Or do you suggest to also implement own `getOrCreateSender` equivalents?

sophokles73:
> To reuse all the sender handling in hono client `AsyncCommandClientImpl` would need to implement `MessageSender`, but `MessageSender` has a lot of methods that we do not need at all for `AsyncCommandClient`. Or do you suggest to also implement own `getOrCreateSender` equivalents?

I am not sure if I am following. If you extend `AbstractSender` then you do not need to bother about implementing `MessageSender`s methods, do you? I would propose that you also create an `AsyncCommandClient` interface which only defines the async sender methods. Very much along the lines of `CommandResponseSender` and `CommandResponseSenderImpl` ...

sophokles73:
> To prevent misunderstanding: We want to create these new senders via hono client and add it to the creation requests, activeSenders map etc?

Yes, I would think so. The creation process of the `AsyncCommandClient` would profit from all the safeguards and caching that we have implemented. Since the client is still scoped to a tenant, I do not see why it shouldn't. Do you see any reasons why it shouldn't?

DanielMaier-BSI:
No, makes sense. The only reason I ask is that I need to refactor things for this. For example I cannot put the new `AsyncCommandClient` to existing `activeSenders` because I need to retrieve specific type `AsyncCommandClient` later on. Wanted just to ensure that we are talking about the same things before I really start to implement.

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -1025,6 +1111,35 @@ protected void removeTenantClient() {
         });
     }
 
+    @Override
+    public Future<AsyncCommandClient> getOrCreateAsyncCommandClient(final String tenantId, final String deviceId) {
```
DanielMaier-BSI: I am still struggling with this one. Should it be allowed to have async and non-async command clients for same tenant and device in parallel? I yes, we need somehow to distinguish them in `activeSenders` map etc. or duplicate the code an put it into separate map. I also don't like the casting. Any ideas?

DanielMaier-BSI:
Hi @sophokles73 ,

I adapted the PR according to our discussion. See also my inline comments. Additionally I have two questions I stumbled across while playing around with command and control:
 - What happens if with this command and control pattern the replyId == deviceId? We have then two receiver links with the same address, one on protocol adapter side and one on response consumer side. 
- correlation-id needs to be unpredictable because every authenticated device can send responses on behalf of other devices if it is aware about the correlation-id. You proposed to use message-id as a fallback for correlation-id above. Shouldn't we enforce people to set correlation-id in order to force them to think about what this id is for and also advice users to use a cryptographically strong random value? 

I will provide some tests after we discussed implementation details.

sophokles73:
> What happens if with this command and control pattern the replyId == deviceId? We have then two receiver links with the same address, one on protocol adapter side and one on response consumer side.

Yeah, that would probably not work out too well, would it? I do not see how the AMQP Messaging Network could reliably distinguish between the adapter and an application at the time they attach such a receiver with the same source address. This probably means that we need to change the endpoint name for receiving the responses, e.g. using `response/${tenant_id}/${reply_id}` instead of `control/${tenant_id}/${reply_id}` for receiving the responses.

> * Shouldn't we enforce people to set correlation-id in order to force them to think about what this id is for and also advice users to use a cryptographically strong random value?

I am not sure if *enforcing* people to set the correlation-id makes the *correlation-id* any better ;-) However, your observation is correct and we could either point out this problem to users more explicitly, thus raising awareness and (hopefully) motivate people to use random identifiers. Or we could support the creation of a temporary (source) node when the application attaches its receiver link for the responses. Then the AMQP Messaging Network could make sure that the reply-to address is random. However, we would then also give up on the overall structure of the reply-to address, i.e. we could no longer assume that the second segment is the tenant ID etc. This would require some changes to the protocol adapters (at least the AMQP adapter) and the way they convey the reply-to address to the devices.

DanielMaier-BSI:
Should I create a separate issue for changing response address to `response/${tenant_id}/${reply_id}`? This would be an API change with no backwards compatibility, are we still able to do this?

For the second point: Should I create an issue to at least document more explicitly that `correlation-id`  property is security relevant for command and control?

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
DanielMaier-BSI: Probably using separate `creationRequests` and `creationLocks` for senders and request response clients could resolve this issue

sophokles73:
> Should I create a separate issue for changing response address to `response/${tenant_id}/${reply_id}`?

It would be great if you could create an issue pointing out the possibility of competing receiver links if an application uses the device ID as *reply-id*. We can then discuss what the potential fix to this is. Creating a distinct source address for the application to receive responses is one option.

DanielMaier-BSI:
> It would be great if you could create an issue pointing out the possibility of competing receiver links if an application uses the device ID as _reply-id_. We can then discuss what the potential fix to this is. Creating a distinct source address for the application to receive responses is one option.

Created #1073

client/src/main/java/org/eclipse/hono/client/AsyncCommandClient.java
```diff
@@ -0,0 +1,93 @@
...
+package org.eclipse.hono.client;
+
+import java.util.Map;
+
+import io.vertx.core.Future;
+import io.vertx.core.buffer.Buffer;
+
+/**
+ * A client for sending asynchronous request response commands, i.e. all command responses for a specific tenant and
+ * reply ID shall be received by the same consumer in an asynchronous fashion.
```
sophokles73: IMHO this is what we do not want to enforce by introducing the `AsyncCommandResponseConsumer`, isn't it?

client/src/main/java/org/eclipse/hono/client/AsyncCommandClient.java
```diff
@@ -0,0 +1,93 @@
...
+     * A device needs to be (successfully) registered before a client can upload any data for it. The device also needs
+     * to be connected for a successful delivery.
+     *
+     * @param command The command name.
+     * @param data The command data to send to the device or {@code null} if the command has no input data.
+     * @param correlationId The identifier to use for correlating the response with the request. Note: This parameter
+     * is security sensitive. To ensure secure request response mapping choose correlationId carefully, e.g.
+     * {@link java.util.UUID#randomUUID()}.
+     * @param replyId An arbitrary string which gets used for response link address in the form of <em>control/${
+     * tenantId}/${replyId}</em> Must match the {@code replyId} passed to the command response receiver, see also below.
```
sophokles73: below? where?
Didn't you want to add a warning that using *device ID* as the *replyId* is strongly discouraged and might lead to unexpected results?

client/src/main/java/org/eclipse/hono/client/CommandClient.java
sophokles73: this method should be removed with the advent of `AsyncCommandClient`, shouldn't it?

client/src/main/java/org/eclipse/hono/client/CommandClient.java
```diff
@@ -106,4 +106,8 @@
...
+
+    Future<Void> sendCommandWithTenantScopedResponse(String command, String contentType, Buffer data, Map<String, Object> properties);
```
sophokles73: this method should be removed with the advent of `AsyncCommandClient`, shouldn't it?

client/src/main/java/org/eclipse/hono/client/HonoClient.java
sophokles73: this should be removed, shouldn't it?

client/src/main/java/org/eclipse/hono/client/HonoClient.java
```diff
@@ -338,6 +338,49 @@
...
+            Handler<Void> closeHandler);
+
+    Future<MessageConsumer> createCommandResponseConsumer(String tenantId, String replyId,
```
sophokles73: this should be removed, shouldn't it?

client/src/main/java/org/eclipse/hono/client/HonoClient.java
```diff
@@ -338,6 +338,49 @@
...
+            BiConsumer<ProtonDelivery, Message> commandResponseConsumer,
+            Handler<Void> closeHandler);
+
+    /**
+     * Creates a client for consuming async command responses from Hono's north bound <em>Command API</em>.
+     * <p>
+     * The command responses passed in to the responses consumer will be settled automatically if the consumer does not
```
sophokles73: > The command responses passed in to the ~~responses~~ consumer will be settled automatically if the consumer does not throw an exception ~~and does not manually handle the message disposition using the passed in delivery~~.

There is no delivery being passed in ...

client/src/main/java/org/eclipse/hono/client/HonoClient.java
```diff
@@ -338,6 +338,49 @@
...
+     * throw an exception and does not manually handle the message disposition using the passed in delivery.
+     *
+     * @param tenantId The tenant to consume command responses for.
+     * @param replyId The replyId of commands to consume command responses for.
+     * @param asyncCommandResponseConsumer The handler to invoke with every command response received.
```
sophokles73: IMHO simply calling this *consumer* should do

client/src/main/java/org/eclipse/hono/client/HonoClient.java
```diff
@@ -432,6 +475,22 @@
      */
     void shutdown();
 
+    /**
+     * Gets a client for sending async commands, i.e. command responses get received by a separate receiver in an
```
sophokles73: > Gets a client for sending commands to a device asynchronously, i.e. command responses get received by a separate receiver.

client/src/main/java/org/eclipse/hono/client/impl/AsyncCommandClientImpl.java
```diff
@@ -0,0 +1,150 @@
...
+     * <em>control/${tenantId}/${deviceId}</em>.
+     *
+     * @param context The vert.x context to run all interactions with the server on.
+     * @param clientConfig The configuration properties to use.
+     * @param con The AMQP connection to the server.
+     * @param tenantId The tenant that the device belongs to.
+     * @param deviceId The device to create the client for.
+     * @param closeHook A handler to invoke if the peer closes the sender link unexpectedly.
+     * @param creationHandler The handler to invoke with the outcome of the creation attempt.
+     * @throws NullPointerException if any of the parameters are {@code null}.
```
sophokles73: except for the close hook ...

client/src/main/java/org/eclipse/hono/client/impl/CommandClientImpl.java
```diff
@@ -210,6 +212,48 @@ protected BufferResult getResult(
         }
     }
 
+    @Override
+    public Future<Void> sendCommandWithTenantScopedResponse(final String command, final Buffer data) {
```
sophokles73: this should be removed, shouldn't it?

client/src/main/java/org/eclipse/hono/client/impl/CommandClientImpl.java
```diff
@@ -210,6 +212,48 @@ protected BufferResult getResult(
...
+        return sendCommandWithTenantScopedResponse(command, null, data, null);
+    }
+
+    @Override
+    public Future<Void> sendCommandWithTenantScopedResponse(final String command, final String contentType, final Buffer data,
```
sophokles73: this should be removed, shouldn't it?

client/src/main/java/org/eclipse/hono/client/impl/CommandResponseConsumerImpl.java
```diff
@@ -0,0 +1,91 @@
...
+import io.vertx.core.Handler;
+import io.vertx.proton.ProtonConnection;
+import io.vertx.proton.ProtonDelivery;
+import io.vertx.proton.ProtonQoS;
+import io.vertx.proton.ProtonReceiver;
+
+/**
+ * A Vertx-Proton based client for consuming command response messages from a Hono server.
+ */
+public class CommandResponseConsumerImpl extends AbstractConsumer implements MessageConsumer {
```
sophokles73: do we still need this? Or is this obsolete with `AsycCommandResponseConsumerImpl`?

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -757,6 +758,91 @@ private void getOrCreateSender(
         });
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public final Future<MessageConsumer> createCommandResponseConsumer(
```
sophokles73: do we still need this?

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -757,6 +758,91 @@ private void getOrCreateSender(
...
+
+        return createCommandResponseConsumer(tenantId, replyId,
+                (delivery, message) -> createCommandResponseConsumer.accept(message), closeHandler);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public final Future<MessageConsumer> createCommandResponseConsumer(
```
sophokles73: do we still need this?

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -757,6 +758,91 @@ private void getOrCreateSender(
...
+            final String tenantId, final String replyId,
+            final BiConsumer<ProtonDelivery, Message> messageConsumer,
+            final Handler<Void> closeHandler) {
+
+        return createConsumer(
+                tenantId,
+                () -> newCommandResponseConsumer(tenantId, replyId, messageConsumer, closeHandler));
+    }
+
+    private Future<MessageConsumer> newCommandResponseConsumer(
```
sophokles73: do we still need this?

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
sophokles73: I  do not quite see the problem with creating consumers. As for the senders, we could simply prefix the keys used in the activeSenders and creationLocks with the simple class name of the sender to create, couldn't we?

client/src/main/java/org/eclipse/hono/client/CommandClient.java
DanielMaier-BSI: ups, so sorry. Something went terribly wrong with my local merging.

client/src/main/java/org/eclipse/hono/client/impl/CommandClientImpl.java
DanielMaier-BSI: ups, so sorry. Something went terribly wrong with my local merging.

client/src/main/java/org/eclipse/hono/client/impl/CommandClientImpl.java
DanielMaier-BSI: ups, so sorry. Something went terribly wrong with my local merging.

client/src/main/java/org/eclipse/hono/client/HonoClient.java
DanielMaier-BSI: ups, so sorry. Something went terribly wrong with my local merging.

client/src/main/java/org/eclipse/hono/client/HonoClient.java
DanielMaier-BSI: ups, so sorry. Something went terribly wrong with my local merging.

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
DanielMaier-BSI: ups, so sorry. Something went terribly wrong with my local merging.

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
DanielMaier-BSI: ups, so sorry. Something went terribly wrong with my local merging.

client/src/main/java/org/eclipse/hono/client/impl/CommandResponseConsumerImpl.java
DanielMaier-BSI: ups, so sorry. Something went terribly wrong with my local merging.

client/src/main/java/org/eclipse/hono/client/AsyncCommandClient.java
DanielMaier-BSI: I was referring to the @see links below. If this is confusing I can just remove the "see below" hint.

About the warning I am not sure. From my point of view this is a bug that will be fixed with #1073. If you like I can add a hint, but there will be the danger that we miss to remove it after the fix of #1073.

client/src/main/java/org/eclipse/hono/client/HonoClient.java
DanielMaier-BSI: I fixed this.

client/src/main/java/org/eclipse/hono/client/HonoClient.java
DanielMaier-BSI: done

client/src/main/java/org/eclipse/hono/client/HonoClient.java
DanielMaier-BSI: done

client/src/main/java/org/eclipse/hono/client/impl/AsyncCommandClientImpl.java
DanielMaier-BSI: done

client/src/main/java/org/eclipse/hono/client/AsyncCommandClient.java
DanielMaier-BSI: I don't get your point here.

sophokles73: Well, my point is that we specifically introduced the `AsyncCommandResponseConsumer` so that multiple instances of it can consume responses for arbitrary commands sent using the same replyId. So the intention is *not* that
> all command responses for a specific tenant and reply ID shall be received by the **same** consumer

client/src/main/java/org/eclipse/hono/client/AsyncCommandClient.java
sophokles73: that is true, then let's not add the hint.

client/src/main/java/org/eclipse/hono/client/AsyncCommandClient.java
DanielMaier-BSI: I see, I was referring to the same consumer from a logical point of view and not the very same instance, i.e. using the same replyId. I must admit wording is wording and confusing. I will change.

DanielMaier-BSI: I just deleted the i.e. part

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
DanielMaier-BSI: This would work. I just had some concerns because we are loosing type safety in some degree. But if you are fine with it I will provide the updated implementation soon.

DanielMaier-BSI:
I forgot to mention, I did push the changes for prefixing senders in HonoClientImpl

client/src/main/java/org/eclipse/hono/client/impl/HonoClientImpl.java
```diff
@@ -1025,6 +1068,35 @@ protected void removeTenantClient() {
...
+                    onSenderClosed -> {
+                        activeSenders.remove(getResourcesKeyForAsyncCommandClient(tenantId, deviceId));
+                    },
+                    result.completer());
+            return result.map(client -> (MessageSender) client);
+        });
+    }
+
+    private static String getResourcesKeyForAsyncCommandClient(final String tenantId, final String deviceId) {
+        // TODO is it fine to have an async command client and a non-async command client at the same time? If yes we
```
sophokles73: are you going to add the prefix?

sophokles73:
@DanielMaier-BSI 
> I forgot to mention, I did push the changes for prefixing senders in HonoClientImpl

doesn't look like you did ...

DanielMaier-BSI:
> doesn't look like you did ...
Now I did, too many remotes :( sorry.

sophokles73:
@DanielMaier-BSI can you please rebase on master?

DanielMaier-BSI:
> @DanielMaier-BSI can you please rebase on master?

Of course, did it.

DanielMaier-BSI:
oh, you already merged. Would it make sense to add some unit and integration tests for this?

sophokles73:
@DanielMaier-BSI I am already working on an integration test ...

DanielMaier-BSI:
OK.

PR #1031, opened 2019-02-07
[#1015] Move EventBusAuthenticationService class to top level

calohmn:
In preparation for adding a unit test for that class.


sophokles73:
@calohmn can you please rebase?

calohmn:
It was already rebased to the latest master commit. 
The integration test failure here seems to be caused by some kind of communication error between device registry and auth server, leading to an unspecified authentication exception.
This exception is treated as a permanent error and no further reconnect attempts are done.
The upcoming fix for #1015 should fix such scenarios.
I've pushed a whitespace change to trigger another test run.

PR #1036, opened 2019-02-08
Helm deployment documentation

dejanb:


site/content/deployment/kubernetes.md
```diff
@@ -45,6 +45,71 @@ From the directory `deploy/target/deploy/kubernetes` run:
...
+
+~~~sh
+~hono/deploy$ helm delete --purge eclipse-hono
+~hono/deploy$ kubectl delete crd prometheuses.monitoring.coreos.com
+~hono/deploy$ kubectl delete crd prometheusrules.monitoring.coreos.com
+~hono/deploy$ kubectl delete crd servicemonitors.monitoring.coreos.com
+~hono/deploy$ kubectl delete crd alertmanagers.monitoring.coreos.com
+~~~
+
+The additional delete commands are necessary to clean [Prometheus operator CRDs](https://github.com/helm/charts/tree/master/stable/prometheus-operator#uninstalling-the-chart).
```
sophokles73: Are the CRDs related to Prometheus only? Or do they belong to OLM?

dejanb: They are related to Prometheus operator only. OLM is not mandatory for the Helm deployment.

sophokles73: So, in this case you are deploying Prometheus *without* OLM, right?

dejanb: That is my understanding, yes. It's uses only Prometheus Operator for now. We can see how to approach all this in the future.

sophokles73:
@dejanb can you also add a note to the release notes about the new possibility to deploy?

dejanb:
Done

Issue #732, opened 2018-07-25
Any plans to support client certificate based authentication "out of the box" for the example deployment?


boaks:
Are there any plans to support client certificate based authentication "out of the box" for the example deployment? If so, which `trusted-ca` certificate is intended to be used? 
FMPOV the setup is not that easy, so I think, prepare the example deployment will do a good job to make some promotion for that feature.

sophokles73:
That sounds like a good idea :-) An obvious choice would be to use the `cn=ca` certificate as the trust anchor, I guess.

b-abel:
I think this is depended on #349.

b-abel:
I am working on this issue.

sophokles73:
fixed in 4e5902d5df62c93583d9ab7f6c7ec842a362c536

PR #1040, opened 2019-02-11
Proposal for #732

b-abel:
The script for creating the demo certs creates now a client certificate
for device 4711 as well.
The example tenant configuration contains now a the public key of the
CA cert as trust anchor for DEFAULT_TENANT.
The example credentials configuration now contain an entry for the
client certificate of device 4711.


b-abel:
The CA cert is the trust anchor of the default tenant, which makes the example nice and small. The script for creating the demo certs automatically updates the `example-tenants.json`.
The trade-offs I made for this to work are:
* The script requires the tool `jq` to be installed (parsing the json file with sed is IMO too error prone)
* The script in the module `demo-certs` is now coupled to the module `deploy` in order to change the tenant config file.

**WDYT?**

demo-certs/create_certs.sh
```diff
@@ -90,6 +98,14 @@ echo "creating JKS trust store ($DIR/$HONO_TRUST_STORE) containing CA certificat
 keytool -import -trustcacerts -noprompt -alias root -file $DIR/root-cert.pem -keystore $DIR/$HONO_TRUST_STORE -storepass $HONO_TRUST_STORE_PWD
 keytool -import -trustcacerts -noprompt -alias ca -file $DIR/ca-cert.pem -keystore $DIR/$HONO_TRUST_STORE -storepass $HONO_TRUST_STORE_PWD
 
+echo ""
+echo "updating example-tenants.json with CA public key "
```
sophokles73: I would rather not introduce this dependency on the deploy module here
Instead, we might be able to add the trust anchor as part of the initial creation of the `device-registry` volume when starting up/deploying the device registry?

demo-certs/create_certs.sh
```diff
@@ -100,3 +116,4 @@ create_cert kura-adapter $KURA_ADAPTER_KEY_STORE $KURA_ADAPTER_KEY_STORE_PWD
 create_cert artemis $ARTEMIS_KEY_STORE $ARTEMIS_KEY_STORE_PWD
 create_cert coap-adapter $COAP_ADAPTER_KEY_STORE $COAP_ADAPTER_KEY_STORE_PWD
 
+create_client_cert 4711
```
sophokles73: missing EOL
this should actually be detected by the checkstyle rules ...

b-abel: Dammit. Looks like the Plugin checks only Java files. I'll check what I can do.

demo-certs/create_certs.sh
b-abel: > Instead, we might be able to add the trust anchor as part of the initial creation of the `device-registry` volume when starting up/deploying the device registry?

This proved to be rather difficult, because it requires to install tools (jq and openssl) into the init container which is a pain with the busybox image. Instead I decoupled the modules by writing the CA's public key into a file and then use it to update example-tenants.json in the build process via resource filtering.

b-abel:
Linking issues in subjects obviously does not not work, therefore: this belongs to the issue #732.

deploy/pom.xml
```diff
@@ -131,6 +132,26 @@
...
+                            <outputDirectory>${project.build.directory}</outputDirectory>
+                            <resources>
+                                <resource>
+                                    <directory>${project.basedir}/src/main/deploy</directory>
+                                    <targetPath>deploy</targetPath>
+                                    <filtering>true</filtering>
+                                </resource>
+                            </resources>
+                            <filters>
+                                <filter>${project.build.directory}/config/hono-demo-certs-jar/trust-anchor.properties</filter>
```
sophokles73: I like the idea of injecting the values using the resources plugin :+1:

deploy/src/main/deploy/example-credentials.json
```diff
@@ -54,6 +54,17 @@
             "key" : "Y29hcC1zZWNyZXQ="
           }
         ]
+      },
+      {
+        "device-id": "4711",
+        "type": "x509-cert",
+        "auth-id": "CN=Device 4711,OU=Hono,O=Eclipse IoT,L=Ottawa,C=CA",
```
sophokles73: Since we already have the mechanism in place, can't we also use it to inject the device cert's subject DN to these credentials? IMHO that would make the whole process less error prone when changing or adding a device ...

deploy/src/main/deploy/example-tenants.json
```diff
@@ -1,6 +1,10 @@
 [
   {
     "tenant-id": "DEFAULT_TENANT",
+    "trusted-ca": {
+      "subject-dn": "CN=ca,OU=Hono,O=Eclipse IoT,L=Ottawa,C=CA",
```
sophokles73: could we also inject the CA's subject DN using the resources plugin?

sophokles73:
> Linking issues in subjects obviously does not not work, therefore: this belongs to the issue #732.

It does:
```
[#732] This commit does ...
```

deploy/src/main/deploy/example-credentials.json
b-abel: Of course. That's good.

deploy/src/main/deploy/example-tenants.json
b-abel: That would be more consistent.

b-abel:
> > Linking issues in subjects obviously does not not work, therefore: this belongs to the issue #732.
> 
> It does:
> 
> ```
> [#732] This commit does ...
> ```

I was talking about the title of this pull request:

> Proposal for #732

This did not cause a backlink on the issue page, so you could not see there that a PR already existed.

demo-certs/create_certs.sh
```diff
@@ -99,7 +106,9 @@ keytool -import -trustcacerts -noprompt -alias root -file $DIR/root-cert.pem -ke
 keytool -import -trustcacerts -noprompt -alias ca -file $DIR/ca-cert.pem -keystore $DIR/$HONO_TRUST_STORE -storepass $HONO_TRUST_STORE_PWD
 
 echo ""
-echo "writing CA public key into properties file"
+echo "extracting trust anchor information from CA cert"
+CA_SUBJECT=$(openssl x509 -in $DIR/ca-cert.pem -noout -subject -nameopt RFC2253 | sed s/^subject=//)
+sed -i "s%\(trusted-ca\.subject-dn=\).*\$%\1${CA_SUBJECT}%" $DIR/trust-anchor.properties
```
sophokles73: why don't we simply overwrite or create the file here, e.g. using something like
```
echo "trusted-ca.subject-dn=$CA_SUBJECT" > $DIR/trust-anchor.properties
echo "trusted-ca.public-key=$PK" >> $DIR/trust-anchor.properties
```
looks a little easier to understand than the (highly sophisticated) `sed` command for replacing the value ...

b-abel: Done.

demo-certs/certs/trust-anchor.properties
```diff
@@ -0,0 +1,11 @@
...
+#**
+#******************************************************************************
+#
+# The CN of the certificate used as trust anchor of a tenant for client certificate based authentication.
+trusted-ca.subject-dn=CN=ca,OU=Hono,O=Eclipse IoT,L=Ottawa,C=CA
+#
+# The public key of the CA in the format that the tenant API expects for setting up a trust anchor.
+trusted-ca.public-key=MIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEApK5L6yUknQnj4FREQqs/G+klIhEKc1X4q4KrSVvFa7Vc4NW2ITTZ9X89i1AGz3gqUUZNOpN1EMNMeJkqewoYI5Bqbr4hiXC6gfUHpgiNxOBq84u/xgqNvOqrQyrz2MMivLDRfwPLbkS+/dXR33Ebh0iFJFHtceH8Vi7Fq7QIKG90+bNkzf+u3r57uBOvEjUZ96tMXaadDGmpjcaNEHp2XSZRQ/vdhSqMYKS28IRdDYuBTXguSxA7MTAtGhaqaIfyqLLufzoqc8FEyNnHKK1QQLGKIv0kvvN1OZUaaA2M0pjZh+ELs3omtBTaw03TfuqAF4bUwtv5MnPXrc25hEU1UhcCc1YOoB7KMZVUP7uP1Ua4nDBChgdRw3hDhQP+IPe5JL92780tFoQ+aFHMtp7ydyhGJGHo00mW7+5h8cGFvAkQbGf+i/tmA5zCdhBsH1f+BOknVMsQiyKb+FQk725qhe1HLZQfiU4vAjmwZLWKPESs0AquGGdFcjrjmdt2WUQ0l2dRaJLNOLxYHaxEYjvko4KPdPLpJIoKKLy/8uLrES4ervczJUlHRhbGDaWMD4NhBueyfrM1NNGtnexLTEzn8SEDwyoP9pv2OzkvdZiFKUcAl6HSV+Davh8lyb0fjQNCDKG2d+kLyxzUyhuChXTCt46seF3JjZwVmHBbHbQrBkkCAwEAAQ==
```
sophokles73: Is this value consistent with the current content of the ca-cert.pem file?
If the key pair that you have created for the device has been created using the script, then the other keys and certs have been re-created as well and would need to be included in this PR ...

b-abel: Yes, it is. I took care that the current CA cert is used and that I did not commit any changed cert or key files to keep this PR as small as possible
But: I found that the CA cert expires in a week. So I hoped to get this PR merged soon and then I would provide another one with new certificates.

sophokles73: Given that the certs expire anyway, FMPOV it would make sense to re-generate all keys and certs using the script and include everything in this PR. Then we can also be sure that everything is in a consistent state ...

b-abel: Oh, doing that I found that the script does not create a cert for the AMQP adapter. I will fix this as well.

b-abel: The cert for Hono Messaging can be removed?

b-abel: Done.

b-abel:
I added examples to the user guide for sending telemetry data with the HTTP and the MQTT adapters using client certificates. For the AMQP adapter we probably need this as well (have to check if this is easily possible with the command line client).
Do we want to include this topic to the getting started guide?

sophokles73:
I think we can add an example for the AMQP adapter later. Let's try to finish this PR up for now ...

sophokles73:
@b-abel can you please squash and rebase?

b-abel:
Done.

sophokles73:
@b-abel can you also please fix the header of the commit message to refer to the issue number using a hash, i.e. `[#732] .... ` instead of `[732] ...` so that the commit gets linked to the issue automatically?

b-abel:
Yes, sorry. Fixed it.

PR #1042, opened 2019-02-12
Replace the fields of CommandResponse with a message for storing the properties.

kaniyan:


client/src/main/java/org/eclipse/hono/client/CommandResponse.java
```diff
@@ -179,16 +162,17 @@ public String getContentType() {
...
     /**
-     * Gets a map of the properties to include in the message as application properties.
+     * Gets the AMQP 1.0 message representing this command response.
      *
-     * @return The properties.
+     * @return The command response message.
      */
-    public Map<String, Object> getProperties() {
-        return properties;
+    public Message getCommandResponseMessage() {
```
sophokles73: IMHO something like `toMessage()` would better reflect the purpose of transforming this object to an AMQP message. The client doesn't know (and doesn't need to know) that, in fact, we keep an instance of `Message` internally all the time ...

client/src/main/java/org/eclipse/hono/client/CommandResponse.java
```diff
@@ -34,22 +33,24 @@
...
-        this.contentType = contentType;
-        this.status = status;
+        message = ProtonHelper.message();
+        message.setCorrelationId(correlationId);
+        MessageHelper.setCreationTime(message);
+        MessageHelper.addProperty(message, MessageHelper.APP_PROPERTY_TENANT_ID, tenantId);
+        MessageHelper.addProperty(message, MessageHelper.APP_PROPERTY_DEVICE_ID, deviceId);
+        MessageHelper.addProperty(message, MessageHelper.APP_PROPERTY_STATUS, status);
+        MessageHelper.setPayload(this.message, contentType, payload);
+        this.replyToId = replyToId;
```
sophokles73: why do we need to keep this property? Can't we determine the message's address from the replyToId and the other parameters?

client/src/main/java/org/eclipse/hono/client/CommandResponse.java
kaniyan: In the command class, it is named as `getCommandMessage` and I thought of keeping the same naming convention. I agree with `toMessage()`.

client/src/main/java/org/eclipse/hono/client/CommandResponse.java
kaniyan: @sophokles73 Below method in `CommandConnection` takes _replyId_ as an argument.
`Future<CommandResponseSender> getCommandResponseSender(String tenantId, String replyId);` 

`AbstractProtocolAdapterBase` uses the above method to get a sender and it needs the _replyId_.

sophokles73: I see :+1:

client/src/main/java/org/eclipse/hono/client/impl/CommandResponseSenderImpl.java
```diff
@@ -113,19 +113,12 @@ static final String getTargetAddress(final String tenantId, final String replyId
     public Future<ProtonDelivery> sendCommandResponse(final CommandResponse commandResponse,
             final SpanContext context) {
 
+        Objects.requireNonNull(targetAddress);
```
sophokles73: *targetAddress* is not a parameter and should therefore not be checked using `Objects.requireNull()`. In fact, the *create* method which is used to create instances of this class, makes sure that targetAddress will not be null already.

kaniyan: @sophokles73 Done.

kaniyan:
> CommandResponseSenderImpl

@sophokles73 If a client uses the above method, then the client should set the `deviceId `and `tenantId` as part of the properties map. This properties map is later added into the application properties of the message. Or as an alternative the deviceId and tenantId can be added as parameters and this can be added to the message in the next PR.
Also I am wondering that currently this method is not being used anywhere and is there any reason why it is there?

sophokles73:
> Also I am wondering that currently this method is not being used anywhere and is there any reason why it is there?

Good question. It looks obsolete anyway. I will mark it as deprecated ...

Issue #1022, opened 2019-02-04
Hono client may become stalled while connecting because of stalled state in AMQP connection stack


DanielMaier-BSI:
There are some states in AMQP connection stack used by hono where hono client ends up in a stalled state while connecting.

Examples are: AMQP Server does not respond to AMQP open frame, Server does not respond while SASL handshake (see also https://github.com/vert-x3/vertx-proton/issues/92) and maybe some others.

To protect hono client against such states, caused by bugs or missing features in used libraries, I would suggest to introduce some kind of overall connection timeout for connection attempts as a fallback. This timeout should abort the running connection attempt and trigger, if configured, a connection retry. This timeout needs not necessarily to be configurable, perhaps it is sufficient and avoids confusion if hono calculates an appropriate value from the other timeouts like TCP timeout etc. 

WDYT?

ctron:
I agree, the client side should always time out after it can't successfully connect. You never know what happens.

I am not so sure about the automatic retry, though. I think if it fails, then it fails. And the client (device) will re-trigger at some point in time anyways.

ctron:
> Such an error causes the reconnect handling in HonoClientImpl to retry opening the connection (if there's reconnectAttempts left).

Ok, that is fine. I was assuming you wanted to add some additional re-connect handling.

calohmn:
Concerning the configuration property for this new timeout feature:
We already have `connectTimeout` in `ClientConfigProperties`. This currently refers just to the time it takes for the TCP/TLS connection to be established.

Looking at the way the `ProtonSaslClientAuthenticatorImpl` class, handling the SASL handshake in the client, is hardwired to be used in vert.x-proton, it looks to me there's no good way to apply a timeout setting just for the SASL handshake.

Therefore I suggest changing the meaning of the existing `connectTimeout` property so that it refers to the time for TCP/TLS connection establishment _plus_ the time for the SASL handshake.

sophokles73:
[...]

PR #1045, opened 2019-02-12
[#1022] Include SASL handshake in ClientConfigProperties connectTimeout

calohmn:
Fix for #1022.


calohmn:
Right, extending the timeout to cover also the _open_ frame makes sense. 
I'd keep the change in `ConnectionFactoryImpl` as the `open` method is invoked there. And because the `ClientConfigProperties` with the `connectTimeout` property is already given to the `ConnectionFactoryImpl` constructor, I think it makes sense to handle that timeout in that class (so that other classes besides `HonoClient`, e.g. `AuthenticationServerClient`, have that feature as well).

core/src/main/java/org/eclipse/hono/connection/impl/ConnectionFactoryImpl.java
```diff
@@ -130,28 +131,53 @@ public void connect(
         final ProtonClient client = protonClient != null ? protonClient : ProtonClient.create(vertx);
         logger.debug("connecting to AMQP 1.0 container [{}://{}:{}]", clientOptions.isSsl() ? "amqps" : "amqp",
                 config.getHost(), config.getPort());
+
+        final AtomicBoolean connectionTimeoutReached = new AtomicBoolean(false);
+        final Long connectionTimeoutTimerId = config.getConnectTimeout() > 0
+                ? vertx.setTimer(config.getConnectTimeout(), id -> {
+                    if (connectionTimeoutReached.compareAndSet(false, true)) {
+                        failConnectionAttempt(clientOptions, connectionResultHandler, new TimeoutException(
```
sophokles73: Using `java.util.concurrent.TimeoutException` here feels a little strange to me, in particular since its description states
> Exception thrown when a blocking operation times out. Blocking operations for which a timeout is specified need a means to indicate that the timeout has occurred.

In absence of another, better suited *standard* exception, we should IMHO define our own exception for this purpose ...

calohmn: Ok, I've used a newly defined `ConnectTimeoutException` now.

Issue #926, opened 2018-11-29
Allow devices to connect via multiple gateways


sysexcontrol:
Currently there is only support for one gateway per device.
We now need to support any number of gateways instead.

Commands must be possible to be sent no matter which gateway the device is connected to.

Use case example:
- a bike could be connected via different smartphones (e.g. in a family context)

Issue #930, opened 2018-11-29
Support gateway agnostic addressing of devices


sysexcontrol:
Applications must be able to send commands to devices without knowing the particular gateway they may be connected to.

calohmn:
To be achieved here is that the consumer application can send the command message only knowing the device id, not the gateway id.
However, on the side of the protocol adapter, the receiver link for receiving such commands may have been created with the gateway id in the address ("control/tenantId/gatewayId"), not the device id.
That means that either in the consumer application or somewhere between consumer application and protocol adapter, a mapping from device id to corresponding gateway id has to be applied.

The place, where such mapping data is kept (the "mapper component"), could be a new component or in the device registry for example. Whenever there is a new request from a device via a gateway, this mapping data could be updated.


As to the way this mapping can get applied so that the command request reaches the right protocol adapter, I see these alternatives:

1. Mapping applied in the consumer application:
The consumer app first makes a request to the new mapping data component (via qpid) to get the current gateway id for the device and then uses that gateway id in the command request.
This functionality could be encapsulated in the HonoClient and therefore be mostly transparent to the consumer app.

2. Define Qpid waypoint that applies mapping and redirects the request:
Using the router feature of waypoints and auto-links, the command request with the device id from the consumer application is first handled by the "waypoint mapper" component (i.e. "mapper component", could be the same as the device registry). In this component the mapping is applied and the request gets redirected to the "control/tenantId/gatewayId" address to be consumed by the protocol adapter.
Whether this approach is actually feasible and whether it would work together with EnMasse or a non-Qpid based AMQP messaging network remains to be seen.

3. Use an intermediate "control/tenantId" dispatcher component:
With this approach, there is a new "Command dispatcher" (or "Gateway-to-device Mapper") component in between the protocol adapter and Qpid.
This new component creates an AMQP receiver on "control/tenantId" addresses in connection with Qpid, so that it will receive the command requests from the consumer application.
The device id of the received requests is mapped to the gateway id and the command request is sent to the protocol adapter.
This means the protocol adapters create the receiver link on the "control/tenantId/gatewayId" address not with Qpid but with the new "Command dispatcher" component.

Option 1 is probably the most straightforward to implement but introduces further complexity on non-HonoClient based consumer applications.
With Option 2 it is unclear, how this could work exactly and how support would look like with EnMasse or a non-Qpid based messaging network.
Option 3 seems the most flexible to me. (I'm not sure yet about all the implications, though.)

Thoughts on this?

[...]

PR #1047, opened 2019-02-18
[#926] Support authorization of multiple gateways per device.

sophokles73:
Device registration service now supports multiple gateway identifiers to
be specified in a device's registration information's "via" property.
The "via" property can be set to either a string or a JSON array of
strings.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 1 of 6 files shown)


sophokles73:
@BobClaerhout is this extension useful for your use cases? Note, though, that this PR does not allow the device registry nor applications to determine, via which particular gateway a device is *currently* connected. For that, we have #930 in the pipeline ...

BobClaerhout:
We bypassed the gateway feature because we do not know which gateway will be sending the messages for a specific device (i.e.: we have a pool of devices and a pool of gateways on premise which send data for the devices). However, currently hono does not know of the actual devices because only one gateway was possible per device. Now it gets tricky: this means that the physical gateway is actually  a device in hono. Which gateway has received which device last, is currently kept in the business application to make sure to send downlink commands to the correct hono device/physical gateway. This PR is a step in the right direction for us.

sophokles73:
> This PR is a step in the right direction for us.

Does this mean that you then would not need to *bypass the gateway feature* altogether anymore?

BobClaerhout:
@sophokles73 Eventually, yes. The real benefit for us is https://github.com/eclipse/hono/issues/930 as we no longer have to keep the latest gateway-device combination in our business application.
However, we also think it would be a lot more easier to configure if we could configure a pool of gateways for a device instead of a list of gateways for each device. This will allow us to add/remove gateways without having to persist this for each device. Ultimately a pool of gateways ways can refer to other pools to make this as dynamical as possible (e.g.: some devices roaming to other gateway pools).

PR #1048, opened 2019-02-18
[#931] Add connection limit in AMQP adapter.

b-abel:
Add connection limit to AMQP adapter (#931). The limit is enforced before authentication happens. The recommended and potentially autoconfigured limit is calculated based on the given maximum amount of memory and the AMQP session window size.

A next step for me would be to refactor the limit handling of the MQTT and the AMQP adapter in order to have a interface `ConnectionLimitManager` that is implemented by the protocal adapers.


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorFactory.java
```diff
@@ -171,46 +179,53 @@ public void process(final Handler<Boolean> completionHandler) {
...
-                if (AuthenticationConstants.MECHANISM_PLAIN.equals(remoteMechanism)) {
-                    verifyPlain(saslResponse, deviceAuthTracker.completer());
-                } else if (AuthenticationConstants.MECHANISM_EXTERNAL.equals(remoteMechanism)){
-                    verifyExternal(deviceAuthTracker.completer());
+                return;
+            }
+
+            if (connectionLimitManager.isConnectionLimitExceeded()) {
+                LOG.debug("Connection limit exceeded, reject connection request");
+                completionHandler.handle(Boolean.TRUE);
```
sophokles73: IMHO this is not enough to signal to the client that the SASL handshake has failed. we need to explicitly invoke `sasl.done()` with an appropriate state. In this case, `SaslOutcome.PN_SASL_TEMP` seems to be appropriate in order to signal an unrecoverable but temporary problem.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/ConnectionLimitManager.java
```diff
@@ -0,0 +1,113 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class manages a limit of concurrent connections. The limit can be configured in
+ * {@link ProtocolAdapterProperties#setMaxConnections(int)}. If no value is configured, it will set a limit based on the
+ * given maximum amount of memory and the AMQP session window size (
+ * {@link org.eclipse.hono.adapter.amqp.AmqpAdapterProperties#getMaxSessionWindowSize()}).
+ * 
+ */
+public class ConnectionLimitManager {
```
sophokles73: Hmmm, we already use something similar for the MQTT adapter, don't we? This smells like a strategy interface with different implementations for different protocol adapters, don't you think?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -142,7 +142,8 @@ protected void doStart(final Future<Void> startFuture) {
                                     .ignoreActiveSpan()
                                     .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_SERVER)
                                     .withTag(Tags.COMPONENT.getKey(), getTypeName())
-                                    .start());
+                                        .start(),
```
sophokles73: indentation

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/ConnectionLimitManager.java
b-abel: Absolutely. I thought of doing the alignment between both modules (and moving common code to module `service-base`) in a separate PR. But I can do it here if you prefer.

sophokles73: Yes, I would rather have it in this PR and not add code that we immediately remove again.

sophokles73:
@b-abel what's the status of this CQ? Will you be pushing additional commits for the refactoring?

b-abel:
> @b-abel what's the status of this CQ? Will you be pushing additional commits for the refactoring?

Yes.

b-abel:
> @b-abel what's the status of this CQ? Will you be pushing additional commits for the refactoring?

Done.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpConnectionLimitStrategy.java
```diff
@@ -0,0 +1,83 @@
...
+     * should be reserved for the startup of the adapter. Per connection the maximum AMQP session window size + 20KB are
+     * needed.
+     * <p>
+     *
+     * @param maxMemory The maximum amount of memory for the JVM in bytes as returned by {@link Runtime#maxMemory()}.
+     * @return The recommended maximum connection limit.
+     */
+    @Override
+    protected long calculateRecommendedLimit(final long maxMemory) {
+        return (maxMemory - MINIMAL_MEMORY) / (MEMORY_PER_CONNECTION + maxSessionWindowSize);
```
sophokles73: isn't this formula always the same for all connection oriented protocols?
I wonder if `MemoryBasedConnectionLimitStrategy` couldn't implement this method in a generic way:
```
return (maxMemory - MINIMAL_MEMORY) / (getMemoryRequiredPerConnection());
```
and let subclasses only implement *getMemoryRequiredPerConnection* ...

service-base/src/main/java/org/eclipse/hono/service/limiting/ConnectionLimitManager.java
```diff
@@ -0,0 +1,92 @@
...
+     * connections and the adapter's properties for the configured limit.
+     * <p>
+     * If the protocol adapter properties are null or no limit is configured, the recommended limit of the strategy will
+     * be set as the limit.
+     * 
+     * @param strategy The strategy to configure the connection limit.
+     * @param metrics The metrics of the adapter.
+     * @param config The configuration of the adapter or {@code null}.
+     */
+    public ConnectionLimitManager(final ConnectionLimitStrategy strategy, final Metrics metrics,
```
sophokles73: I'd rather not tie this class to `Metrics`. In fact, we only ever need to read the current number of connections, right?
So, why not use something like
```
public ConnectionLimitManager(final ConnectionLimitStrategy strategy, final Supplier<Integer> currentConnections, final ProtocolAdapterProperties config)
```
instead.
A supplier based on Metrics could then be passed in like this:
```
new ConnectionLimitManager(strategy, () -> metrics.getNumberOfConnections(), config) ...
```
That way we could more easily swap out the source for the number of connections if necessary. WDYT?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpConnectionLimitStrategy.java
b-abel: You are right. Then I would go one step further: there is no need for`MemoryBasedConnectionLimitStrategy` to be extended, since the calculation is done there based on static values (for minimal memory and memory required per connection), which could be constructor parameters.

service-base/src/main/java/org/eclipse/hono/service/limiting/ConnectionLimitManager.java
b-abel: Yes, that is very good.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpConnectionLimitStrategy.java
sophokles73: sounds even better :+1:

b-abel:
Pushed a new version.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -101,7 +102,7 @@
     private MqttServer insecureServer;
     private AuthHandler<MqttContext> authHandler;
     private final BiConsumer<CommandSubscription, CommandContext> afterCommandPubAckedConsumer = this::afterCommandPublished;
-    private int connectionLimit = Integer.MAX_VALUE;
+    private ConnectionLimitManager connectionLimitManager;
```
sophokles73: what about pulling up this property to `AbstractProtocolAdapterBase` and providing a setter for it as well? That would make testing the enforcing of the limit in the adapters much easier ...

b-abel: Done.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -281,8 +280,8 @@ public void doStart(final Future<Void> startFuture) {
 
         final MemoryBasedConnectionLimitStrategy strategy = new MemoryBasedConnectionLimitStrategy(MINIMAL_MEMORY,
                 MEMORY_PER_CONNECTION);
-        connectionLimitManager = new ConnectionLimitManager(strategy, () -> metrics.getNumberOfConnections(),
-                getConfig());
+        setConnectionLimitManager(new ConnectionLimitManager(strategy, () -> metrics.getNumberOfConnections(),
```
sophokles73: why not use the same approach as for the AMQP adapter?

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -1456,4 +1458,11 @@ protected TrustOptions getServerTrustOptions() {
                 });
     }
 
+    public ConnectionLimitManager getConnectionLimitManager() {
```
sophokles73: should be final and needs JavaDoc

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -1456,4 +1458,11 @@ protected TrustOptions getServerTrustOptions() {
...
+        return connectionLimitManager;
+    }
+
+    public void setConnectionLimitManager(final ConnectionLimitManager connectionLimitManager) {
```
sophokles73: should be final and needs JavaDoc

b-abel:
Done.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -396,9 +353,9 @@ final void handleEndpointConnection(final MqttEndpoint endpoint) {
 
     private Future<Device> handleConnectionRequest(final MqttEndpoint endpoint, final Span currentSpan) {
 
-        if (isConnectionLimitExceeded()) {
-            LOG.debug("Connection limit ({}) exceeded, reject connection request", connectionLimit);
-            currentSpan.log(String.format("connection limit (%d) exceeded", connectionLimit));
+        // the ConnectionLimitManager is null in some unit tests
```
sophokles73: would it make sense to initialize the connectionLimitManager as part of the constructor?

b-abel: The Constructor of the protocol adapter? The ConnectionLimitManager can be provided by a setter and depends on properties that are not present during construction as well (metrics, config), so I don't see how this could be done in the constructor.  The other option I see would be to leverage one of Spring's initialization options. If the goal is to prevent  connectionLimitManager  from being null, then it could be provided as a Bean and initialized with a NoOp implementation or Optional. Not sure if this would be more obvious.

sophokles73: oh sure, I forgot about the properties that it depends on  ...

sophokles73:
@b-abel can you squash and rebase?

b-abel:
> @b-abel can you squash and rebase?

Done.

PR #1050, opened 2019-02-19
Minor fixes and improvements

calohmn:
Adopting some inspection results, etc.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
@@ -21,8 +21,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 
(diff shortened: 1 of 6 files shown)


client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -345,7 +343,7 @@ protected static final void setApplicationProperties(final Message msg, final Ma
                     // and will send a detach frame for closing the link very shortly
                     // see AMQP 1.0 spec section 2.6.3
                     LOG.debug("peer did not create terminus for target [{}] and will detach the link", targetAddress);
-                    result.tryFail(new ServerErrorException(HttpsURLConnection.HTTP_UNAVAILABLE));
+                    result.tryFail(new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE));
```
sophokles73: Nice catch :+1:

PR #1051, opened 2019-02-19
Add missing @Override annotations

calohmn:


calohmn:
There is the `MissingOverride` module. It only checks methods where the the `{@inheritDoc} javadoc tag`is present though.
Anyway, adding this won't hurt. -> PR #1052.

Issue #1046, opened 2019-02-15
Fix encoding of OpenTracing span in AMQP messages


ctron:
My understanding is that the current encoding of OpenTracing information in AMQP messages could cause some issues it is not compliant with the AMQP spec.

About annotations the spec says (http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-annotations):

> The annotations type is a map where the keys are restricted to be of type symbol or of type ulong. All ulong keys, and all symbolic keys except those beginning with "x-" are reserved. Keys beginning with "x-opt-" MUST be ignored if not understood. On receiving an annotation key which is not understood, and which does not begin with "x-opt", the receiving AMQP container MUST detach the link with a not-implemented error. 

What seems to happen right now is that the `MessageAnnotationsInjectAdapter` is being used in combination with the `TEXT_MAP` format, which would directly inject `uber-trace-id` and `uberctx-*` into the message annotation map (see: `io.jaegertracing.internal.propagation.TextMapCodec`):

So it might happen, that a receiver simply closes this link when the annotation is present in the map.

The proposal would be to use something like `x-opt-opentracing-trace-id` and `x-opt-opentracing-ctx-` instead.

ctron:
Looking into the code, it actually might be tricky to get the proposal implemented. Maybe a different way to do it would be to keep the current keys of `uber-*` but put them into the annotations under a single key, using an AMQP map.

sophokles73:
The `MessageAnnotations*Adapter` actually injects to/extracts from the AMQP message's *delivery annotations* at the moment. The naming of the class does not reflect this properly (bad :-1: ). However, this does not change the overall problem here.

calohmn:
In general I think it's a good idea to keep an eye on the AMQP related work in https://github.com/w3c/trace-context/. There's a draft in https://github.com/w3c/trace-context/pull/217 there.
There it is proposed to use a binary format. As this doesn't seem to be supported by Jaeger yet, I guess it's premature to adopt it here.
In any case I think it would be good to also address https://github.com/eclipse/hono/issues/1012 when doing changes on the Adapter classes here.

calohmn:
When following the approach suggested in #1012 (also trying to read the trace-context from the _application-properties_ of the bare message), it has to be noted that an AMQP map is not supported as value type there. I would therefore suggest using the plain property names (e.g. _uber-trace-id_) in the _application-properties_ case.

[...]

Issue #1012, opened 2019-01-29
Reconsider putting trace-context into message annotations


calohmn:
Currently the OpenTracing context used across messaging components is transported via AMQP delivery annotations in Hono.

However, the W3C Trace Context Specification plans on using application properties and/or message annotations.
See the corresponding [draft proposal](https://github.com/w3c/trace-context/pull/217):
> AMQP defines message as a payload with the additional annotations sections. There are two annotation sections this specification refers - "application-properties" and "message-annotations". See 3.2 of OASIS Advanced Message Queuing Protocol (AMQP) Version 1.0, Part 3: Messaging (AMQP specification).
>
> AMQP message section "application-properties" is immutable collection of properties that is defined by message publisher and can be read by the message consumer. Message brokers cannot mutate those properties. Section "message-annotations" is designed for message brokers to use and can be mutated during the message processing.
>
> Fields traceparent and tracestate SHOULD be added to the message in the application-properties section by message publisher. Once the message has been created, it’s no longer permissible to edit the bare message. So if it were necessary to annotate the message inside the middleware as it transits, that MUST happen in the “message-annotations” section, using the “application-properties” as a base. Message reader SHOULD construct the full trace context by reading traceparent and tracestate fields from the “message-annotations” first and if not exist - from “application-properties”.

See also this [comment](https://github.com/w3c/trace-context/pull/217#pullrequestreview-189826628).

Based on this proposal, it would be good to reconsider the current approach in Hono.

The alternative would be for the protocol adapters to _write_ the trace context into message annotations and for the receiving end (e.g. AMQP protocol adapter, device registry service, etc.) to _read_ the trace context first from message annotations and, if not found there, from the message application properties.

calohmn:
PR is merged now.

To sum up the change:
The trace-context is read from / written to message annotations now (instead of delivery annotations).

If not found there, currently no further attempt is made to read the trace-context from the application-properties of the bare message. We can add such a feature later on when the w3c trace-context spec is finalized (and when jaeger has support for the binary format if that is used in the spec).

PR #1054, opened 2019-02-20
[#1046][#1012] Fix encoding of OpenTracing span context in AMQP message

calohmn:
The span context will now be injected in and extracted from the map value of a "x-opt-trace-context" message annotation.

Fix for #1046 and #1012.


calohmn:
We _could_ of course. However, as the use case of each class is distinct and as the given example in opentracing-java also follows this pattern of separate classes, I'd rather keep it like it is.

PR #1058, opened 2019-02-21
Allow using netty epoll on Linux

ctron:


ctron:
This PR required the following CQs:

* [X] https://dev.eclipse.org/ipzilla/show_bug.cgi?id=18640
* [ ] https://dev.eclipse.org/ipzilla/show_bug.cgi?id=18641

sophokles73:
@ctron looks like you have a go for [CQ18641](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=18641) as well. is there anything else, preventing you from merging this?

ctron:
From what I know, you need to wait for the approval before doing the release.

I would say to wait a few more days, and then ask. But not merge it before we have the full approval.

sophokles73:
Hono 0.9 is due for release on Feb 25th.

ctron:
Time flies … e-mail sent.

ctron:
The CQs are approved.

ctron:
So the intention is not to have this as a S2I only feature. However I think it may be good idea to make conscious decision to enable this.

However document this might be good idea :)

site/content/admin-guide/common-config.md
```diff
@@ -0,0 +1,27 @@
...
+
+Most Hono components support a common set of configuration options. This section
+describes those options. <!--more-->
+
+Each component which supports the following options explicitly states so. If
+it doesn't, then these options are not supported by this component.
+
+## Using epoll() on Linux 
+
+Netty allows the use of `epoll()` on Linux x86_64 based systems. Hono provides
```
sophokles73: IMHO it would be good to explain what `epoll` is about and/or provide a link to a resource where I can learn about it. I do not expect many people to actually know why they should activate/use this feature without further explanation.

site/content/admin-guide/amqp-adapter-config.md
```diff
@@ -10,6 +10,8 @@ The adapter is implemented as a Spring Boot application. It can be run either di
 
 ## Service Configuration
 
+In addition the following options, this component supports the options described in [Common Configuration]({{< relref "common-config.md" >}}).
```
sophokles73: IMHO this should read
`In addition to the following ...`

ctron: Indeed. Should be fixed.

site/content/admin-guide/common-config.md
ctron: I added a short section about that.

site/content/admin-guide/common-config.md
```diff
@@ -0,0 +1,31 @@
...
+
+Using `epoll()` on Linux may provide better performance for application which
+have a high I/O throughput. Especially when the application supports an
+asynchronous I/O model. This is true for most Hono applications.
+
+Netty allows the use of `epoll()` on Linux x86_64 based systems. Hono provides
+a build profile for enabling this support, and allows to configure Vert.x to
+make use of this.
+
+Only when the Maven profile was active during the build **and** the Vert.x
```
sophokles73: why do we need to activate the profile? The CQ should allow us to include the libs by default, shouldn't it? Using the property, the user can still decide if he wants to enable it ...

ctron: Yes, we could enable this for the docker images explicitly. However we don't know if the user builds this on 64bit linux. So, when releasing the images, yes we can. By default, I wouldn't do that.

Issue #1056, opened 2019-02-20
Device Registry Integration Tests assume non-empty device registry


jbtrystram:
In the process of creating a device registry based on Infinispan for production use, I wanted to get the Hono integration tests to run against our implementation. 

The integration tests expect the DR to already hold some data, which is given in the `resources` of the test module, in JSON files.
This works well with the file based device registry since it use the same persistence mechanism. However, a third party device registry would reasonably be empty. 

I started to write some code to load the json files, similarly to how it is done in the file based registry. But this creates some code that is dedicated to the integration tests in the device registry, which hurts readability and doesn't really make sense.

Maybe we could rework the tests to have some code that "prepare" the device registry before the integration tests run, the same way it's done with the unit tests. That would reduce code duplication and also help future implementations. The downside I see with this approach is that we would have to send HTTP or AMQP messages to do that, so the preparation steps would already been into the testing territory of some sort.

@ctron take on this is a good idea imo : "maybe we can introduce some kind of "IGNORE_JSON" env-var, which would, when the container is started up, ignore the provided JSON content. so we could have an empty DR for tests, but still have a pre-filled one for demo deployments"

What do you guys think ?

sophokles73:
We certainly started to implement the integration tests with the idea to pre-load the registry with data from files, based on the example data we use in the example deployment. However, in the meantime, and with the advent of the `DeviceRegistryHttpClient` I do not think that we actually need to do that at all. Instead, we could create the fixture (i.e. the necessary data in the registry) as part of the integration tests themselves. We are doing this in many tests already and I do not see why we shouldn't make this the general pattern.

Another question is whether we want the test cases to put the data to the registry entity-by-entity or if we want to support *bulk import* of the fixture based on JSON files. I can imagine the `DeviceRegistryHttpClient` to support both ...

sophokles73:
@jbtrystram what's the status of this? With the recent changes you contributed, can we (i.e. you) close this issue?

jbtrystram:
I suppose we can close it, yes !

PR #1060, opened 2019-02-21
[#1056] Populate the device registry with the expected data before running Integration Tests

jbtrystram:
Hi, 

This is a quick draft to raise some discussion on how to implement a way to populate the device registry with the data needed for the integration tests, as stated in issue #1056. 
Here I used a `@Before` method and used the `DeviceRegistryHttpClient` to push the data in the device registry. 

Maybe a it would be better to directly load the Json file and push it as it is in the DR, to have a single source of data. Imo, that would improve maintainability (and readibility, as creating those JSON object programatically isn't very pleasant)..
What do you guys thinks ?

### DIFF
--- a/services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsServiceTest.java
+++ b/services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsServiceTest.java
@@ -15,10 +15,7 @@ package org.eclipse.hono.deviceregistry;
 
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
-import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
 
 import java.net.HttpURLConnection;
 import java.nio.charset.StandardCharsets;
@@ -228,6 +225,34 @@ public class FileBasedCredentialsServiceTest extends AbstractCompleteCredentials
         startup.await();
         assertRegistered(svc, Constants.DEFAULT_TENANT, "sensor1", CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, ctx);
     }
+
+    /**
+     * Verifies that credentials are ignored if the startEmpty property is set.
+     *
+     * @param ctx The test context.
+     */
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    @Test
+    public void testDoStartIgnoreCredentialIfStartEmptyIsSet(final TestContext ctx) {
+
+        // GIVEN a service configured with a file name and startEmpty set to true
+        props.setFilename(FILE_NAME);
+        props.setStartEmpty(true);
+        when(fileSystem.existsBlocking(props.getFilename())).thenReturn(Boolean.TRUE);
+
+        // WHEN the service is started
+        final Async startup = ctx.async();
+        final Future<Void> startFuture = Future.future();
+        startFuture.setHandler(ctx.asyncAssertSuccess(s -> {
+            startup.complete();
+        }));
+        svc.doStart(startFuture);
+
+        // THEN the credentials from the file are not loaded
+        startup.await();
+        verify(fileSystem, never()).readFile(anyString(), any(Handler.class));
+    }
+
 
     /**
      * Verifies that the file written by the registry when persisting the registry's contents can
--- a/services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/FileBasedRegistrationServiceTest.java
+++ b/services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/FileBasedRegistrationServiceTest.java
@@ -262,6 +262,33 @@ public class FileBasedRegistrationServiceTest extends AbstractCompleteRegistrati
     }
 
     /**
+     * Verifies that device identities in file are ignored if startEmpty is set to true.
+     *
+     * @param ctx The test context.
+     */
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    @Test
+    public void testDoStartIgnoreIdentitiesIfStartEmptyIsSet(final TestContext ctx) {
+
+        // GIVEN a service configured with a file name and startEmpty set to true
+        props.setFilename(FILE_NAME);
+        props.setStartEmpty(true);
+        when(fileSystem.existsBlocking(props.getFilename())).thenReturn(Boolean.TRUE);
+
+        // WHEN the service is started
+        final Async startup = ctx.async();
+        final Future<Void> startFuture = Future.future();
+        startFuture.setHandler(ctx.asyncAssertSuccess(s -> {
+            startup.complete();
+        }));
+        registrationService.doStart(startFuture);
+
+        // THEN the device identities from the file are not loaded
+        startup.await();
+        verify(fileSystem, never()).readFile(anyString(), any(Handler.class));
+    }
+
+    /**
      * Verifies that the registry enforces the maximum devices per tenant limit.
      */
     @Test
--- a/services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/FileBasedTenantServiceTest.java
+++ b/services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/FileBasedTenantServiceTest.java
@@ -214,6 +214,33 @@ public class FileBasedTenantServiceTest extends AbstractCompleteTenantServiceTes
     }
 
     /**
+     * Verifies that tenants are successfully loaded from file during startup.
+     *
+     * @param ctx The test context.
+     */
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    @Test
+    public void testDoStartwithStartEmptyIgnoreTenants(final TestContext ctx) {
+
+        // GIVEN a service configured with a file name
+        props.setFilename(FILE_NAME);
+        props.setStartEmpty(true);
+        when(fileSystem.existsBlocking(props.getFilename())).thenReturn(Boolean.TRUE);
+
+        // WHEN the service is started
+        final Async startup = ctx.async();
+        final Future<Void> startFuture = Future.future();
+        startFuture.setHandler(ctx.asyncAssertSuccess(s -> {
+            startup.complete();
+        }));
+        svc.doStart(startFuture);
+
+        // THEN the credentials from the file are loaded
+        startup.await();
+        verify(fileSystem, never()).readFile(anyString(), any(Handler.class));
+    }
+
+    /**
      * Verifies that the tenants file written by the registry when persisting the contents can
      * be loaded in again.
      *
--- a/tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpConnectionIT.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpConnectionIT.java
@@ -89,13 +89,19 @@ public class AmqpConnectionIT extends AmqpAdapterTestBase {
     @Test
     public void testConnectFailsForNonExistingDevice(final TestContext ctx) {
 
-        // GIVEN an adapter
-        // WHEN an unknown device tries to connect
-        connectToAdapter(IntegrationTestSupport.getUsername("non-existing", Constants.DEFAULT_TENANT), "secret")
-        .setHandler(ctx.asyncAssertFailure(t -> {
-            // THEN the connection is refused
-            ctx.assertTrue(t instanceof SaslException);
-        }));
+        // GIVEN an existing tenant
+        final String tenantId = helper.getRandomTenantId();
+        final TenantObject tenant = TenantObject.from(tenantId, true);
+
+        helper.registry.addTenant(JsonObject.mapFrom(tenant)).compose( r ->
+
+            // GIVEN an adapter
+            // WHEN an unknown device tries to connect
+            connectToAdapter(IntegrationTestSupport.getUsername("non-existing", tenantId), "secret")
+            .setHandler(ctx.asyncAssertFailure(t -> {
+                // THEN the connection is refused
+                ctx.assertTrue(t instanceof SaslException);
+            })));
     }
 
     /**
--- a/tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
@@ -21,9 +21,11 @@ import static org.junit.Assert.assertTrue;
 
 import java.net.HttpURLConnection;
 import java.nio.charset.StandardCharsets;
+import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.Base64;
+import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 
 import org.eclipse.hono.client.CredentialsClient;
@@ -33,10 +35,11 @@ import org.eclipse.hono.tests.IntegrationTestSupport;
 import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.CredentialsConstants;
 import org.eclipse.hono.util.CredentialsObject;
-import org.junit.AfterClass;
+import org.junit.After;
 import org.junit.BeforeClass;
-import org.junit.Rule;
+import org.junit.AfterClass;
 import org.junit.Test;
+import org.junit.Rule;
 import org.junit.rules.Timeout;
 import org.junit.runner.RunWith;
 
@@ -57,12 +60,12 @@ public class CredentialsAmqpIT {
     private static final String CREDENTIALS_AUTHID2 = "little-sensor2";
     private static final String CREDENTIALS_USER_PASSWORD = "hono-secret";
     private static final byte[] CREDENTIALS_PASSWORD_SALT = "hono".getBytes(StandardCharsets.UTF_8);
-    private static final String DEFAULT_DEVICE_ID = "4711";
 
     private static final Vertx vertx = Vertx.vertx();
 
     private static HonoClient client;
     private static CredentialsClient credentialsClient;
+    private static IntegrationTestSupport helper;
 
     /**
      * Global timeout for all test cases.
@@ -72,11 +75,14 @@ public class CredentialsAmqpIT {
 
     /**
      * Starts the device registry and connects a client.
-     * 
+     *
      * @param ctx The vert.x test context.
      */
     @BeforeClass
     public static void prepareDeviceRegistry(final TestContext ctx) {
+
+        helper = new IntegrationTestSupport(vertx);
+        helper.initRegistryClient(ctx);
 
         client = DeviceRegistryAmqpTestSupport.prepareDeviceRegistryClient(vertx,
                 IntegrationTestSupport.HONO_USER, IntegrationTestSupport.HONO_PWD);
@@ -89,8 +95,18 @@ public class CredentialsAmqpIT {
     }
 
     /**
+     * Remove the fixture from the device registry if the test had set up any.
+     *
+     * @param ctx The vert.x test context.
+     */
+    @After
+    public void cleanupDeviceRegistry(final TestContext ctx){
+        helper.deleteObjects(ctx);
+    }
+
+    /**
      * Shuts down the device registry and closes the client.
-     * 
+     *
      * @param ctx The vert.x test context.
      */
     @AfterClass
@@ -102,7 +118,7 @@ public class CredentialsAmqpIT {
 
     /**
      * Verify that a not existing authId is responded with HTTP_NOT_FOUND.
-     * 
+     *
      * @param ctx The vert.x test context.
      */
     @Test
@@ -119,35 +135,41 @@ public class CredentialsAmqpIT {
 
     /**
      * Verifies that the service returns credentials for a given type and authentication ID.
-     * 
+     *
      * @param ctx The vert.x test context.
      */
     @Test
     public void testGetCredentialsReturnsCredentialsTypeAndAuthId(final TestContext ctx) {
 
-        credentialsClient
-            .get(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, CREDENTIALS_AUTHID1)
-            .setHandler(ctx.asyncAssertSuccess(result -> {
-                ctx.assertEquals(CREDENTIALS_AUTHID1, result.getAuthId());
-                ctx.assertEquals(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, result.getType());
-            }));
+        final CredentialsObject payload = getRandomHashedPasswordCredential();
+
+        helper.registry.addCredentials(Constants.DEFAULT_TENANT, JsonObject.mapFrom(payload))
+                .compose(ok -> credentialsClient.get(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, payload.getAuthId()))
+                .setHandler(ctx.asyncAssertSuccess(result -> {
+                    ctx.assertEquals(payload.getAuthId(), result.getAuthId());
+                    ctx.assertEquals(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, result.getType());
+                }));
     }
 
-    /**
-     * Verifies that the service returns credentials for a given type, authentication ID and matching client context.
-     *
-     * @param ctx The vert.x test context.
-     */
+/**
+ * Verifies that the service returns credentials for a given type, authentication ID and matching client context.
+ *
+ * @param ctx The vert.x test context.
+ */
     @Test
     public void testGetCredentialsExistingClientContext(final TestContext ctx) {
+
+        final CredentialsObject payload = getRandomHashedPasswordCredential();
 
         final JsonObject clientContext = new JsonObject()
                 .put("client-id", "gateway-one");
 
-        credentialsClient
-                .get(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, "gw", clientContext)
+        payload.setProperty("client-id", "gateway-one");
+
+        helper.registry.addCredentials(Constants.DEFAULT_TENANT, JsonObject.mapFrom(payload))
+                .compose(ok -> credentialsClient.get(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, payload.getAuthId(), clientContext))
                 .setHandler(ctx.asyncAssertSuccess(result -> {
-                    ctx.assertEquals("gw", result.getAuthId());
+                    ctx.assertEquals(payload.getAuthId(), result.getAuthId());
                     ctx.assertEquals(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, result.getType());
                 }));
     }
@@ -160,12 +182,16 @@ public class CredentialsAmqpIT {
     @Test
     public void testGetCredentialsNotMatchingClientContext(final TestContext ctx) {
 
+        final CredentialsObject payload = getRandomHashedPasswordCredential();
+
         final JsonObject clientContext = new JsonObject()
                 .put("client-id", "gateway-two");
 
-        credentialsClient
-                .get(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, "gw", clientContext)
-                .setHandler(ctx.asyncAssertFailure(t -> {
+        payload.setProperty("client-id", "gateway-one");
+
+        helper.registry.addCredentials(Constants.DEFAULT_TENANT, JsonObject.mapFrom(payload))
+                .compose(ok ->  credentialsClient.get(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, payload.getAuthId(), clientContext))
+                .setHandler(ctx.asyncAssertFailure( t -> {
                     ctx.assertEquals(
                             HttpURLConnection.HTTP_NOT_FOUND,
                             ((ServiceInvocationException) t).getErrorCode());
@@ -180,12 +206,14 @@ public class CredentialsAmqpIT {
     @Test
     public void testGetCredentialsNotExistingClientContext(final TestContext ctx) {
 
+        final CredentialsObject payload = getRandomHashedPasswordCredential();
+
         final JsonObject clientContext = new JsonObject()
                 .put("client-id", "gateway-one");
 
-        credentialsClient
-                .get(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, CREDENTIALS_AUTHID1, clientContext)
-                .setHandler(ctx.asyncAssertFailure(t -> {
+        helper.registry.addCredentials(Constants.DEFAULT_TENANT, JsonObject.mapFrom(payload))
+                .compose(ok -> credentialsClient.get(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, payload.getAuthId(), clientContext))
+                .setHandler(ctx.asyncAssertFailure( t -> {
                     ctx.assertEquals(
                             HttpURLConnection.HTTP_NOT_FOUND,
                             ((ServiceInvocationException) t).getErrorCode());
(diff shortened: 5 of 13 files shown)


sophokles73:
@jbtrystram FMPOV this goes in the right direction. It might be tricky to remove the data after running the test case from the registry again when you load it from a JSON file, though. In the test case code you will need to keep track of the identifiers of the objects created. It might get even more tricky if you want to use random identifiers. In that case you may want to use placeholders in the JSON files and have the test case replace them with random identifiers (and keep track of them).

sophokles73:
@jbtrystram how do you mark a PR as a draft? That's a nice feature which I haven't been aware of yet :-)

jbtrystram:
Here is an update considering @sophokles73 comment. I chose a different approach though : each test sets up it's own fixture , then the `@After` method does a cleanup. 

It seems to work properly but i am getting am issue compiling `test/java/org/eclipse/hono/tests/CrudHttpClient`

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
```diff
@@ -86,11 +91,32 @@ public static void prepareDeviceRegistry(final TestContext ctx) {
...
+    }
+
+    /**
+     * Remove the fixture from the device registry if the test had set up any.
+     *
+     */
+    @After
+    public void cleanupDeviceRegistry(){
+        if (testData != null) {
+            registryHttpClient.removeCredentials(
```
dejanb: @jbtrystram you should handle the result of this call and await for its execution to avoid async problems. Take a look at https://github.com/eclipse/hono/blob/master/tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsHttpIT.java#L104 for example on how to do it.

jbtrystram: thanks for the tip !

jbtrystram:
It looks in shape now. What do you think ? 
If you are OK with the approach i'll apply it to the other tests and update the PR.

On a side note : I discovered that `testGetCredentialsExistingClientContext` fails because the client context is not handed over all the way to the device registry (receiving an empty JsonObject.). So it looks like it's not correctly parsed  / eliminated along the way ...

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
```diff
@@ -119,37 +134,84 @@ public void testGetCredentialsNotExistingAuthId(final TestContext ctx) {
...
     @Test
     public void testGetCredentialsExistingClientContext(final TestContext ctx) {
 
+        // Prepare the credential to insert
+        final String deviceId = helper.getRandomDeviceId(Constants.DEFAULT_TENANT);
+        final JsonObject payload = new JsonObject()
+                .put(CredentialsConstants.FIELD_PAYLOAD_DEVICE_ID, deviceId)
+                .put(CredentialsConstants.FIELD_TYPE, CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD)
+                .put(CredentialsConstants.FIELD_AUTH_ID, "gw")
+                .put("client-id", "gateway-one")
```
jbtrystram: The context is not handed all the way through the device registry, so the entry is created without it and the subsequent test fails. The issue is most likely into `helper.registry.addCredentials()`

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsServiceTest.java
```diff
@@ -229,6 +229,40 @@ public void testDoStartLoadsCredentials(final TestContext ctx) {
...
+        final Async startup = ctx.async();
+        final Future<Void> startFuture = Future.future();
+        startFuture.setHandler(ctx.asyncAssertSuccess(s -> {
+            startup.complete();
+        }));
+        svc.doStart(startFuture);
+
+        // THEN the credentials from the file are not loaded
+        startup.await();
+        assertNotRegistered(svc, Constants.DEFAULT_TENANT, "sensor1", CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, ctx);
```
sophokles73: this is a very weak assertion since it will also be true if the file loaded does not contain such credentials.
instead, you should verify that the *readFile* operation has not been invoked at all. For that, you can omit the `doAnswer` block above and add

```
verify(filesystem, never()).readFile(anyString(), any(Handler.class));
```
down here.

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/FileBasedRegistrationServiceTest.java
```diff
@@ -261,6 +261,39 @@ public void testDoStartLoadsDeviceIdentities(final TestContext ctx) {
...
+        final Async startup = ctx.async();
+        final Future<Void> startFuture = Future.future();
+        startFuture.setHandler(ctx.asyncAssertSuccess(s -> {
+            startup.complete();
+        }));
+        registrationService.doStart(startFuture);
+
+        // THEN the device identities from the file are not loaded
+        startup.await();
+        assertThat(registrationService.getDevice(TENANT, DEVICE).getStatus(), is(HttpURLConnection.HTTP_NOT_FOUND));
```
sophokles73: see above

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/FileBasedTenantServiceTest.java
```diff
@@ -213,6 +213,39 @@ public void testDoStartLoadsTenants(final TestContext ctx) {
...
+        final Async startup = ctx.async();
+        final Future<Void> startFuture = Future.future();
+        startFuture.setHandler(ctx.asyncAssertSuccess(s -> {
+            startup.complete();
+        }));
+        svc.doStart(startFuture);
+
+        // THEN the credentials from the file are loaded
+        startup.await();
+        assertTenantDoesNotExist(svc, Constants.DEFAULT_TENANT, ctx);
```
sophokles73: see above

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpConnectionIT.java
```diff
@@ -89,13 +89,19 @@ public void testConnectSucceedsForRegisteredDevice(final TestContext ctx) {
...
-        connectToAdapter(IntegrationTestSupport.getUsername("non-existing", Constants.DEFAULT_TENANT), "secret")
-        .setHandler(ctx.asyncAssertFailure(t -> {
-            // THEN the connection is refused
-            ctx.assertTrue(t instanceof SaslException);
-        }));
+        // GIVEN an existing tenant
+        final String tenantId = helper.getRandomTenantId();
+        final TenantObject tenant = TenantObject.from(tenantId, true);
+
+        helper.registry.addTenant(JsonObject.mapFrom(tenant)).compose( r ->
```
sophokles73: what purpose does it serve to register the random tenant when it is not used later on?

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
```diff
@@ -119,37 +134,84 @@ public void testGetCredentialsNotExistingAuthId(final TestContext ctx) {
...
+                .put(CredentialsConstants.FIELD_PAYLOAD_DEVICE_ID, deviceId)
+                .put(CredentialsConstants.FIELD_TYPE, CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD)
+                .put(CredentialsConstants.FIELD_AUTH_ID, CREDENTIALS_AUTHID1)
+                .put(CredentialsConstants.FIELD_SECRETS, getAuthId1Secrets());
+
+        final Async done = ctx.async();
+        // Insert it into the device Registry
+        helper.registry.addCredentials(Constants.DEFAULT_TENANT, payload).setHandler(r -> {
+
+            // it's successfully inserted, run the test.
```
sophokles73: simply *compose()* the next step instead of using *setHandler* and checking manually if the previous step succeeded.
at the end, you can then use *setHandler(ctx.asyncAssertSuccess())* to make sure that the test succeeds. No need for the *done* `Async` as well ...

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
```diff
@@ -119,37 +134,84 @@ public void testGetCredentialsNotExistingAuthId(final TestContext ctx) {
...
 
-        credentialsClient
-            .get(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, CREDENTIALS_AUTHID1)
-            .setHandler(ctx.asyncAssertSuccess(result -> {
-                ctx.assertEquals(CREDENTIALS_AUTHID1, result.getAuthId());
-                ctx.assertEquals(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, result.getType());
-            }));
+        // Prepare the credential to insert
+        final String deviceId = helper.getRandomDeviceId(Constants.DEFAULT_TENANT);
+        final JsonObject payload = new JsonObject()
```
sophokles73: you should be able to use `CredentialsObject`'s factory methods for doing this ...

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
sophokles73: > The issue is most likely into `helper.registry.addCredentials()`

I cannot find anything in there that would cause the issue.

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
```diff
@@ -119,37 +134,84 @@ public void testGetCredentialsNotExistingAuthId(final TestContext ctx) {
...
+ * Verifies that the service returns credentials for a given type, authentication ID and matching client context.
+ *
+ * @param ctx The vert.x test context.
+ */
     @Test
     public void testGetCredentialsExistingClientContext(final TestContext ctx) {
 
+        // Prepare the credential to insert
+        final String deviceId = helper.getRandomDeviceId(Constants.DEFAULT_TENANT);
+        final JsonObject payload = new JsonObject()
```
sophokles73: again, you should/could use the factory methods of `CredentialsObject`, I guess.

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
```diff
@@ -119,37 +134,84 @@ public void testGetCredentialsNotExistingAuthId(final TestContext ctx) {
...
-                .setHandler(ctx.asyncAssertSuccess(result -> {
-                    ctx.assertEquals("gw", result.getAuthId());
-                    ctx.assertEquals(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, result.getType());
-                }));
+        final Async done = ctx.async();
+        // Insert it into the device Registry
+        helper.registry.addCredentials(Constants.DEFAULT_TENANT, payload).setHandler(r -> {
+
+            // it's successfully inserted, run the test.
+            if (r.succeeded()) {
```
sophokles73: see above

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
jbtrystram: I found the issue, opened a PR :)

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpConnectionIT.java
jbtrystram: if no corresponding tenant exists, the adapters is blocked to connect, and cannot check if the device exists.

sophokles73: that's true, but then you should also use that tenant in the device credentials later on (instead of DEFAULT_TENANT), right?

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsServiceTest.java
jbtrystram: oh you can even monitor calls, TIL ! Thanks for the suggestion :smile:

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/FileBasedRegistrationServiceTest.java
jbtrystram: changed as suggested.

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/FileBasedTenantServiceTest.java
jbtrystram: changed as suggested.

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
dejanb: There a couple of more places in this tests that this could be improved

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationOptionalAmqpIT.java
```diff
@@ -96,10 +101,26 @@ public static void shutdown(final TestContext ctx) {
...
-            .compose(client -> client.assertRegistration(DeviceRegistrationAmqpIT.DEVICE_ID_ENABLED_WITH_ENABLED_GATEWAY,
-                    DeviceRegistrationAmqpIT.GATEWAY_ID_ENABLED))
-            .setHandler(ctx.asyncAssertSuccess());
+        // Prepare the identities to insert
+        final String deviceId = helper.getRandomDeviceId(Constants.DEFAULT_TENANT);
+        final String gatewayId = helper.getRandomDeviceId(Constants.DEFAULT_TENANT);
+
+        final Async done = ctx.async();
+        // Insert into the device Registry
+        final Future fut1 = helper.registry.registerDevice(Constants.DEFAULT_TENANT, deviceId,
```
dejanb: Perhaps there two calls can be better merged with the existing calls (using compose())?

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationOptionalAmqpIT.java
```diff
@@ -111,10 +132,26 @@ public void testAssertRegistrationSucceedsForGateway(final TestContext ctx) {
     @Test
     public void testAssertRegistrationFailsForDisabledGateway(final TestContext ctx) {
 
-        deviceRegistryclient.getOrCreateRegistrationClient(Constants.DEFAULT_TENANT)
-            .compose(client -> client.assertRegistration(DeviceRegistrationAmqpIT.DEVICE_ID_ENABLED_WITH_DISABLED_GATEWAY,
-                    DeviceRegistrationAmqpIT.GATEWAY_ID_DISABLED))
-            .setHandler(ctx.asyncAssertFailure(t -> assertErrorCode(t, HttpURLConnection.HTTP_FORBIDDEN, ctx)));
+        // Prepare the identities to insert
```
dejanb: Same as above?

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationOptionalAmqpIT.java
```diff
@@ -126,10 +163,29 @@ public void testAssertRegistrationFailsForDisabledGateway(final TestContext ctx)
     @Test
     public void testAssertRegistrationFailsForUnauthorizedGateway(final TestContext ctx) {
 
-        deviceRegistryclient.getOrCreateRegistrationClient(Constants.DEFAULT_TENANT)
-            .compose(client -> client.assertRegistration(DeviceRegistrationAmqpIT.DEVICE_ID_ENABLED_WITH_ENABLED_GATEWAY,
-                    DeviceRegistrationAmqpIT.GATEWAY_ID_ENABLED_2))
-            .setHandler(ctx.asyncAssertFailure(t -> assertErrorCode(t, HttpURLConnection.HTTP_FORBIDDEN, ctx)));
+        // Prepare the identities to insert
```
dejanb: And here as well

tests/src/test/java/org/eclipse/hono/tests/registry/TenantAmqpIT.java
```diff
@@ -93,11 +110,28 @@ public static void shutdown(final TestContext ctx) {
...
-            .get(Constants.DEFAULT_TENANT)
-            .setHandler(ctx.asyncAssertSuccess(tenantObject -> {
-                ctx.assertEquals(Constants.DEFAULT_TENANT, tenantObject.getTenantId());
-            }));
+        // Prepare the identities to insert
+        final String tenantId = Constants.DEFAULT_TENANT;
+
+        final Async done = ctx.async();
+        // Insert into the device Registry
+        helper.registry.addTenant(new JsonObject()
```
dejanb: We could try to reuse TenantObject class here (and in couple more places in this class) to make this JSON Object formatting easier.

sophokles73:
@jbtrystram I have fixed #1070 so you might want to rebase to HEAD of master ...

jbtrystram:
Hi @sophokles73 and @dejanb .

I squased and rebased on master.
I made changes according to your suggestions. However i could not figure out howto use `.compose()` in the tests to chain the actions, and ran into asynchronous issues when removing the `Async` object to block the test methods. I'd gladly accept some help here.

I also kept some constants (i.e. `Constants.DEFAULT_TENANT`) in some places, as the Authentication service is also hardcoded with some well-known values. That could be changed as well but requires more features in `AuthenticationServerClient`, so that will be for another PR.

jbtrystram:
Update : I found out how to compose the calls properly . It indeed allows to remove the `async`  :)
so it should be OK now

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
```diff
@@ -26,17 +26,19 @@
...
 import org.eclipse.hono.util.CredentialsConstants;
 import org.eclipse.hono.util.CredentialsObject;
-import org.junit.AfterClass;
-import org.junit.BeforeClass;
-import org.junit.Rule;
+import org.junit.After;
 import org.junit.Test;
+import org.junit.Rule;
+import org.junit.BeforeClass;
+import org.junit.AfterClass;
```
sophokles73: this does not comply with the import ordering we use in Hono. Please configure your IDE to use the order defined in `eclipse/hono.importorder`

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
```diff
@@ -119,37 +134,55 @@ public void testGetCredentialsNotExistingAuthId(final TestContext ctx) {
...
+ *
+ * @param ctx The vert.x test context.
+ */
     @Test
     public void testGetCredentialsExistingClientContext(final TestContext ctx) {
 
+        final String deviceId = helper.getRandomDeviceId(Constants.DEFAULT_TENANT);
+        final CredentialsObject payload = new CredentialsObject(deviceId, "gw", CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD);
+
+        payload.addSecret( new JsonObject()
```
sophokles73: simply use a corresponding factory method, e.g. `CredentialsObject.fromClearTextPassword()` which saves a lot of boiler plate code ...

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
```diff
@@ -119,37 +134,55 @@ public void testGetCredentialsNotExistingAuthId(final TestContext ctx) {
...
+        payload.setProperty("client-id", "gateway-one");
+
+        helper.registry.addCredentials(Constants.DEFAULT_TENANT, JsonObject.mapFrom(payload))
+                .compose( r -> credentialsClient.get(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, "gw", clientContext))
+                .compose(result -> {
                     ctx.assertEquals("gw", result.getAuthId());
                     ctx.assertEquals(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, result.getType());
-                }));
+                    return Future.succeededFuture();
+                }).setHandler(ctx.asyncAssertSuccess());
```
sophokles73: better use
```
.setHandler(ctx.asyncAssertSuccess(result -> {
   ctx.assertEquals("gw", result.getAuthId());
   ctx.assertEquals(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, result.getType());
}));
```

jbtrystram: done

jbtrystram:
@sophokles73 updated with your latest comments

jbtrystram:
and rebased on master as well

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
```diff
@@ -119,33 +135,48 @@ public void testGetCredentialsNotExistingAuthId(final TestContext ctx) {
...
-                ctx.assertEquals(CREDENTIALS_AUTHID1, result.getAuthId());
-                ctx.assertEquals(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, result.getType());
-            }));
+        final String deviceId = helper.getRandomDeviceId(Constants.DEFAULT_TENANT);
+        final CredentialsObject payload = new CredentialsObject(deviceId, CREDENTIALS_AUTHID1, CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD);
+        getAuthId1Secrets().forEach( secret -> payload.addSecret((JsonObject) secret));
+
+        helper.registry.addCredentials(Constants.DEFAULT_TENANT, JsonObject.mapFrom(payload))
+                .compose(r -> credentialsClient.get(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, CREDENTIALS_AUTHID1))
+                .compose(result -> {
```
sophokles73: replace with
```
.setHandler(ctx.asyncAssertSuccess(result -> {
  ctx.assertEquals(CREDENTIALS_AUTHID1, result.getAuthId());
  ctx.assertEquals(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, result.getType());
});
```

jbtrystram:
@sophokles73 OK, got you. Adjusted as per your comment. The general shape now looks like : 
```
registry.insert(data)
.compose(amqpClient.get(data))
.setHandler(ctx.asyncAssertSuccess( t -> {
    ctx.Assert(....)
)
```

tests/src/test/java/org/eclipse/hono/tests/registry/TenantAmqpIT.java
```diff
@@ -87,17 +110,23 @@ public static void shutdown(final TestContext ctx) {
...
-        tenantClient
-            .get(Constants.DEFAULT_TENANT)
-            .setHandler(ctx.asyncAssertSuccess(tenantObject -> {
-                ctx.assertEquals(Constants.DEFAULT_TENANT, tenantObject.getTenantId());
-            }));
+        final String tenantId = Constants.DEFAULT_TENANT; //helper.getRandomTenantId();
+        final TenantObject payload = new TenantObject()
+                .setTenantId(tenantId)
+                .setEnabled(true)
+                .setTrustAnchor(getRandomPublicKey(), new X500Principal("CN=ca, OU=Hono, O=Eclipse"));
```
sophokles73: no need to  slow down test execution by creating a key here, simply omit the trust anchor ...

tests/src/test/java/org/eclipse/hono/tests/registry/TenantAmqpIT.java
```diff
@@ -109,13 +138,24 @@ public void testGetTenant(final TestContext ctx) {
     @Test
     public void testGetNotConfiguredTenantReturnsUnauthorized(final TestContext ctx) {
 
-        tenantClient
-                .get("HTTP_ONLY")
+        final String tenantId = helper.getRandomTenantId();
+        final X500Principal subjectDn = new X500Principal("CN=ca, OU=Hono, O=Eclipse");
```
sophokles73: no need for a trust anchor ...

tests/src/test/java/org/eclipse/hono/tests/registry/TenantAmqpIT.java
```diff
@@ -109,13 +138,24 @@ public void testGetTenant(final TestContext ctx) {
...
+
+        helper.registry.addTenant(JsonObject.mapFrom(payload))
+                .compose(r -> tenantClient.get(tenantId))
                 .setHandler(ctx.asyncAssertFailure(t -> {
-                    ctx.assertEquals(
-                            HttpURLConnection.HTTP_FORBIDDEN,
-                            ((ServiceInvocationException) t).getErrorCode());
-                }));
+                        ctx.assertEquals(
+                                HttpURLConnection.HTTP_FORBIDDEN,
```
sophokles73: I am not sure if this test case should be included in the suite because it relies on the (static) configuration of the client's permissions in the AuthServer. So, unless you also can configure the permissions in AuthServer dynamically as part of the fixture, it probably doesn't make sense to include this test here.

tests/src/test/java/org/eclipse/hono/tests/registry/TenantAmqpIT.java
```diff
@@ -168,13 +214,36 @@ public void testGetByCaSucceeds(final TestContext ctx) {
...
+
+        helper.registry.addTenant(JsonObject.mapFrom(payload))
+                .compose(r ->  tenantClient.get(subjectDn))
                 .setHandler(ctx.asyncAssertFailure(t -> {
-                    ctx.assertEquals(
-                            HttpURLConnection.HTTP_FORBIDDEN,
-                            ((ServiceInvocationException) t).getErrorCode());
-                }));
+                        ctx.assertEquals(
+                                HttpURLConnection.HTTP_FORBIDDEN,
```
sophokles73: see above, this test case also depends on permissions configuration ...

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
```diff
@@ -180,96 +221,147 @@ public void testGetCredentialsNotMatchingClientContext(final TestContext ctx) {
...
+                .setHandler(ctx.asyncAssertSuccess(result -> {
+                    assertFalse(checkPayloadGetCredentialsContainsDeviceIdAndReturnEnabled(result, deviceId));
+                }));
+    }
+
+    private JsonArray getAuthId1Secrets(){
+
+        return new JsonArray()
+                .add( new JsonObject()
+                        .put( CredentialsConstants.FIELD_SECRETS_NOT_BEFORE, "2017-05-01T14:00:00+01:00")
```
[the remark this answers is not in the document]
jbtrystram: Look like `hashedPasswordSecretForClearTextPassword` simply stores the password in plain text https://github.com/eclipse/hono/blob/b66119adfc4f06ef68b11eb936fc01e76c32c50f/core/src/main/java/org/eclipse/hono/util/CredentialsObject.java#L642
`

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
```diff
@@ -180,96 +221,147 @@ public void testGetCredentialsNotMatchingClientContext(final TestContext ctx) {
...
     public void testGetCredentialsReturnsCredentialsDefaultDeviceIdAndIsEnabled(final TestContext ctx) {
 
-        credentialsClient
-            .get(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, CREDENTIALS_AUTHID1)
-            .setHandler(ctx.asyncAssertSuccess(result -> {
-                assertTrue(checkPayloadGetCredentialsContainsDefaultDeviceIdAndReturnEnabled(result));
-            }));
+        final String deviceId = helper.getRandomDeviceId(Constants.DEFAULT_TENANT);
+        final CredentialsObject payload = new CredentialsObject(deviceId, CREDENTIALS_AUTHID1, CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD);
+        getAuthId1Secrets().forEach( secret -> payload.addSecret((JsonObject) secret));
```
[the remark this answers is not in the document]
jbtrystram: Now the tests use `getRandomHashedPasswordCredential()`

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
```diff
@@ -160,16 +189,28 @@ public void testGetCredentialsExistingClientContext(final TestContext ctx) {
...
-                .get(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, "gw", clientContext)
-                .setHandler(ctx.asyncAssertFailure(t -> {
-                    ctx.assertEquals(
-                            HttpURLConnection.HTTP_NOT_FOUND,
-                            ((ServiceInvocationException) t).getErrorCode());
-                }));
+
+        helper.registry.addCredentials(Constants.DEFAULT_TENANT, JsonObject.mapFrom(payload))
+                .compose(r -> {
+                    return credentialsClient.get(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD, "gw", clientContext)
```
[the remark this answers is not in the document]
jbtrystram: Yep, a couple slipped through ! Changed it so they use the same structure as other tests.

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
```diff
@@ -119,33 +134,47 @@ public void testGetCredentialsNotExistingAuthId(final TestContext ctx) {
...
+ */
     @Test
     public void testGetCredentialsExistingClientContext(final TestContext ctx) {
 
+        final String deviceId = helper.getRandomDeviceId(Constants.DEFAULT_TENANT);
+        final CredentialsObject payload = CredentialsObject.fromClearTextPassword(
+                deviceId,
+                "gw",
+                "hono-secret",
+                Instant.parse("2018-01-01T00:00:00Z"),
```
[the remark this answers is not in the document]
jbtrystram: `testGetCredentialsFirstSecretCurrentlyActiveTimeInterval` fails when using `null`

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationAmqpIT.java
```diff
@@ -151,9 +152,15 @@ public void testAssertRegistrationFailsForDisabledDevice(final TestContext ctx)
     @Test
     public void testAssertRegistrationFailsForNonExistingGateway(final TestContext ctx) {
 
-        deviceRegistryclient.getOrCreateRegistrationClient(Constants.DEFAULT_TENANT)
-                .compose(client -> client.assertRegistration(DEVICE_ID_ENABLED, NON_EXISTING_GATWAY_ID))
-                .setHandler(ctx.asyncAssertFailure(t -> assertErrorCode(t, HttpURLConnection.HTTP_FORBIDDEN, ctx)));
+        final String deviceId = helper.getRandomDeviceId(Constants.DEFAULT_TENANT);
+
+        helper.registry.registerDevice(Constants.DEFAULT_TENANT, deviceId,
+                    new JsonObject().put(RegistrationConstants.FIELD_ENABLED, true))
```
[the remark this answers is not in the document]
jbtrystram: Did not know. I removed the unneeded JsonObject accordingly, thanks.

tests/src/test/java/org/eclipse/hono/tests/registry/TenantAmqpIT.java
jbtrystram: I think it make sense in the Integration Tests context as it reflects a sensible user story. Changing the Authentication HTTP client so we can dynamically create the complete fixture is definitely a good objective, and I wrote earlier, this could be dealt with in another PR. 
But in the meantime why decrease the test coverage ?

tests/src/test/java/org/eclipse/hono/tests/registry/TenantAmqpIT.java
jbtrystram: removed

tests/src/test/java/org/eclipse/hono/tests/registry/TenantAmqpIT.java
jbtrystram: removed

tests/src/test/java/org/eclipse/hono/tests/registry/TenantAmqpIT.java
dejanb: I ageee with @jbtrystram. This is not a regression and we can tackle it in the future improvements.

tests/src/test/java/org/eclipse/hono/tests/registry/TenantAmqpIT.java
```diff
@@ -168,13 +208,35 @@ public void testGetByCaSucceeds(final TestContext ctx) {
...
+                    }));
+    }
+
+    private PublicKey getRandomPublicKey(){
+        try {
+            final KeyPairGenerator keyGen = KeyPairGenerator.getInstance("RSA");
+            keyGen.initialize(1024);
+            final KeyPair keypair = keyGen.genKeyPair();
+            return keypair.getPublic();
+        } catch (NoSuchAlgorithmException e){
```
dejanb: Why not just propagate the exception from this method, so that test fails if it cannot generate PK?

tests/src/test/java/org/eclipse/hono/tests/registry/TenantAmqpIT.java
```diff
@@ -145,11 +180,16 @@ public void testGetConfiguredButNotCreatedTenantReturnsNotFound(final TestContex
     @Test
     public void testGetByCaSucceeds(final TestContext ctx) {
 
+        final String tenantId = Constants.DEFAULT_TENANT; //helper.getRandomTenantId();
```
dejanb: Should this be random tenant id?

jbtrystram: Nope, as it is tied to the hard coded values in the Auth service. See the discussion with @sophokles73 above :)

tests/src/test/java/org/eclipse/hono/tests/registry/TenantAmqpIT.java
jbtrystram: Good suggestion !

tests/src/test/java/org/eclipse/hono/tests/registry/TenantAmqpIT.java
dejanb: OK. The comment is a bit confusing, as it implies it's on TODO. Maybe remove the comment and we'll change this in another pr

jbtrystram: i removed it

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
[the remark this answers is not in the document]
sophokles73: > Look like hashedPasswordSecretForClearTextPassword simply stores the password in plain text

That is true, however, the Credentials endpoint will replace it with a hash when you add the credentials. This brings us to an interesting aspect: using the Credentials HTTP endpoint for setting up the fixture will only make sense if we actually can expect the system under test (the Credentials API implementation) to support this HTTP endpoint and its resource, won't it?

tests/src/test/java/org/eclipse/hono/tests/registry/TenantAmqpIT.java
sophokles73: ok, then let's keep it for the time being

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsAmqpIT.java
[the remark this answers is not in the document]
jbtrystram: As it is Integration Tests, I suppose it make sense to expect the system to be complete, to some extent. Plus, it's so easy for to write an implementation supporting HTTP,  just need to extend the proper class :)

sophokles73:
@jbtrystram can you please squash and rebase?

jbtrystram:
Done :)

PR #1061, opened 2019-02-21
Replace Spring Boot Actuator with vertx-web based Prometheus endpoint.

sophokles73:
Use the vertx-web based HealthCheckServer's HTTP server to expose a
resource for scraping the PrometheusMeterRegistry of Hono components.

This way we do not need to spawn up a jetty server just for serving the Prometheus scraping endpoint and can prevent including a bunch of additional jars when we already have a (running) HTTP server in place.

In this initial version, the scraping and health check endpoints are exposed on an unsecured port (no TLS) without authentication. However, this can be added in a subsequent PR.

### DIFF
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/Application.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/Application.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2018, 2019 Contributors to the Eclipse Foundation
  *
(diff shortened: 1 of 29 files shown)


ctron:
For me the AMQP metrics are not working. Think this is due to the missing package `org.eclipse.hono.service.metric` in the `@ComponentScan` annotation of the class `org.eclipse.hono.adapter.amqp.Application`.

ctron:
Also does the "auth service" no longer provide metrics now, as it does not have a health check endpoint, and thus doesn't register the promtheus alongside.

So if we re-use the health check for the prometheus scraping, then I think also the auth service should follow that pattern.

sophokles73:
@ctron I have pushed corresponding changes ...

PR #1065, opened 2019-02-22
Set Java 11 as source+target version

ctron:
As Hono has problems building on Java 8, and we already ship Java 11
images, we decided to fully switch to Java 11. But keeping the
classpath model.

This PR switches to Java 11 for source+target, and also makes use
of a few Java 11-only features, just test it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: required
 language: java
 
 jdk:
-#  - openjdk8
   - openjdk11
 
 script:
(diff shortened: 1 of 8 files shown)


sophokles73:
@ctron IIRC our conversation, we wanted to do this switch in the next release, didn't we?

ctron:
> @ctron IIRC our conversation, we wanted to do this switch in the next release, didn't we?

I thought we decided to enable that even for the 0.9 release, I am not sure though … But in any case, if you want to postpone it until next week, that is fine with me.

sophokles73:
> if you want to postpone it until next week, that is fine with me.

I'd rather do. We do not have any issue with this currently and I'd rather not make that major change that close to the release date.

ctron:
@sophokles73 Do you think it is ok to merge that now?

ctron:
> Are these changes required by source level 11?

No, they are just there for ensuring that "something" from the Java 11 API is being used.

sophokles73:
@ctron LGTM

avgustinmm:
Hi all, 
are these java 8 problems related to hono-client and its dependencies (hono-core)?
If not, maybe it'll be better to keep them java 8 compatible.
It's probably ok to force  hono backend side to java 11, But do you feel it's ok to force client side (e.g. backend application) to go to java 11?

PR #1069, opened 2019-02-27
DR - forward the complete payload to credential service

jbtrystram:
when a request to create credentials containing client context is forwarded to tje device registry, the payload is mapped into a `CredentialObject`, which prune the additional properties. The mapping is useful to do some checks on the payload.

This simple fix simply forward the original payload once it have been checked.

Signed-off-by: Trystram Jean-Baptiste <jbtrystram@redhat.com>

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/credentials/CompleteBaseCredentialsService.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/credentials/CompleteBaseCredentialsService.java
@@ -111,7 +111,7 @@ public abstract class CompleteBaseCredentialsService<T> extends BaseCredentialsS
         try {
             payload.checkValidity(this::checkSecret);
             final Future<CredentialsResult<JsonObject>> result = Future.future();
-            add(tenantId, JsonObject.mapFrom(payload), result.completer());
+            add(tenantId, request.getJsonPayload(), result.completer());
             return result.map(res -> {
                 return request.getResponse(res.getStatus())
                         .setDeviceId(payload.getDeviceId())


service-base/src/main/java/org/eclipse/hono/service/credentials/CompleteBaseCredentialsService.java
```diff
@@ -111,7 +111,7 @@ protected CompleteBaseCredentialsService(final HonoPasswordEncoder pwdEncoder) {
         try {
             payload.checkValidity(this::checkSecret);
             final Future<CredentialsResult<JsonObject>> result = Future.future();
-            add(tenantId, JsonObject.mapFrom(payload), result.completer());
+            add(tenantId, request.getJsonPayload(), result.completer());
```
sophokles73: The *processAddRequest* creates the CredentialsObject instance from the payload and then invokes *hashPlainPasswords()* with the CredentialsObject. So, with your change, we would be losing the password hashes that have been created from any plain text passwords contained in the original payload.
A fix will probably require a change to `CredentialsObject` to also allow additional properties to be included ...

sophokles73:
@jbtrystram it would be great if you could create an issue for this problem so that we can better track it.

service-base/src/main/java/org/eclipse/hono/service/credentials/CompleteBaseCredentialsService.java
jbtrystram: Sorry, I submitted that too fast, should have thought of that.
we could also insert the hashed password in the Json object, in the end the DR expect JSON, so why not stick to it ? having extra properties is very easy.

jbtrystram:
> @jbtrystram it would be great if you could create an issue for this problem so that we can better track it.

created issue #1070

sophokles73:
fixed in 9c48973686c45b30868108dbc3ad31a755479445

jbtrystram:
Nice ! Thanks!

Issue #1074, opened 2019-03-01
Align main Application class names


ctron:
The current situtation:

* `org.eclipse.hono.adapter.amqp.Application`
* `org.eclipse.hono.adapter.coap.vertx.Application`
* `org.eclipse.hono.adapter.http.vertx.Application`
* `org.eclipse.hono.adapter.mqtt.impl.Application`

I think we should align this.

calohmn:
Commit where the Http adapter `Application` was moved to the `vertx` package: https://github.com/eclipse/hono/commit/19011f6c5c491fcae4df28f24ac4982d557e4606.
There were Eclipse jar-signer conflicts before.

ctron:
Maybe then align that all to `org.eclipse.hono.adapter.<foo>.impl.Application`?!

sophokles73:
@ctron can we (you) close this?

PR #1076, opened 2019-03-04
[#1074]: Align names of protocol adapter main application classes

ctron:
This is proposal to fix issue #1074 … by aligning all protocol adapter main application classes to the pattern of `org.eclipse.hono.adapter.<foo>.impl.Application`.


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/Application.java
```diff
@@ -21,7 +21,9 @@
 /**
  * The Hono AMQP main application class.
  */
-@ComponentScan(basePackages = { "org.eclipse.hono.adapter.amqp", "org.eclipse.hono.service.metric" })
+@ComponentScan("org.eclipse.hono.adapter.amqp")
+@ComponentScan("org.eclipse.hono.adapter.amqp.impl")
+@ComponentScan("org.eclipse.hono.service.metric")
```
sophokles73: I am not quite convinced that this is the same as specifying the packages in a single annotation. Do you have any source where the semantics of multiple instances of the same annotation on a type is defined, in particular in the Spring context?

ctron: I think it is:

* https://docs.oracle.com/javase/tutorial/java/annotations/repeating.html
* https://github.com/spring-projects/spring-framework/blob/bc283ed7d70570bffba0f2a8c82b8e666a87d9f4/spring-context/src/main/java/org/springframework/context/annotation/ComponentScan.java#L58
* https://github.com/spring-projects/spring-framework/blob/4.3.x/spring-context/src/main/java/org/springframework/context/annotation/ComponentScans.java

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/Application.java
```diff
@@ -22,7 +22,9 @@
 /**
  * The Hono MQTT adapter main application class.
  */
-@ComponentScan(basePackages = { "org.eclipse.hono.adapter.mqtt", "org.eclipse.hono.service.metric" })
+@ComponentScan("org.eclipse.hono.adapter.mqtt")
+@ComponentScan("org.eclipse.hono.adapter.mqtt.impl")
```
sophokles73: This should already be covered by *base package* `org.eclipse.hono.adapter.mqtt`, shouldn't it?

ctron: I wasn't able to find any explicit documentation, but running a test confirmed that this does work. So I dropped the annotations for "impl".

ctron:
@sophokles73 Is it ok to merge, from your perspective?

PR #1078, opened 2019-03-11
Add Http based LoRa Adapter with support for different providers

mbaeuerle:
A new protocol adapter which provides connectivity to LoRa devices via different providers. The adapter is based on the AbstractVertxBasedHttpProtocolAdapter. Each provider has a different endpoint.
Telemetry/Events can be received for all listed providers. The Kerlink provider also supports C&C.

Currently implemented providers:
- Actility
- Everynet
- Kerlink
- Objenious
- Proximus
- The Things Network

Signed-off-by: Marian Bäuerle <marian.baeuerle@bosch-si.com>
Also-by: Schmid Christian <christian.schmid3@bosch-si.com>
Also-by: Daniel Maier <daniel.maier4@bosch-si.com>
Also-by: Bala Azhagappan <Balasubramanian.Azhagappan@bosch-si.com>
Also-by: Sebastian Poehn <sebastian.poehn@bosch-si.com>
Also-by: Pascan Marko <Marko.Pascan@bosch-si.com>
Also-by: Klein Marita <marita.klein@bosch-si.com>
Also-by: Lohmann Carsten <carsten.lohmann@bosch-si.com>
Also-by: Meier Sandy <Sandy.Meier@bosch-si.com>


adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraProtocolAdapter.java
```diff
@@ -0,0 +1,636 @@
...
+    private static final String ERROR_MSG_MISSING_OR_UNSUPPORTED_CONTENT_TYPE = "missing or unsupported content-type";
+    private static final Logger LOG = LoggerFactory.getLogger(LoraProtocolAdapter.class);
+    private static final String LORA_COMMAND_CONSUMER_DEVICE_ID = "lora";
+    private static final int LORA_COMMAND_CONSUMER_RETRY_INTERVAL = 2_000;
+    private static final String TAG_LORA_DEVICE_ID = "lora_device_id";
+    private static final String TAG_LORA_PROVIDER = "lora_provider";
+    private static final String JSON_MISSING_REQUIRED_FIELDS = "JSON Body does not contain required fields";
+    private static final String INVALID_PAYLOAD = "Invalid payload";
+    @Autowired
+    private final List<LoraProvider> loraProviders = new ArrayList<>();
```
sophokles73: please add a setter for this property and move the annotation to the setter to reduce coupling with Spring field injection.

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraProtocolAdapter.java
```diff
@@ -0,0 +1,636 @@
...
+    private HonoClientBasedAuthProvider usernamePasswordAuthProvider;
+    private HonoClientBasedAuthProvider clientCertAuthProvider;
+    @Autowired
+    private LoraCommandProperties loraCommandProperties;
```
sophokles73: please add a setter for this property and move the annotation to the setter to reduce coupling with Spring field injection.

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraProtocolAdapter.java
```diff
@@ -0,0 +1,636 @@
...
+     * @param provider The provider to use.
+     * @throws NullPointerException if provider is {@code null}.
+     */
+    public void setClientCertAuthProvider(final HonoClientBasedAuthProvider provider) {
+        this.clientCertAuthProvider = Objects.requireNonNull(provider);
+    }
+
+    @Override
+    protected String getTypeName() {
+        return "hono-lora";
```
sophokles73: please define this name as a constant in `org.eclipse.hono.util.Constants`

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraProtocolAdapter.java
```diff
@@ -0,0 +1,636 @@
...
+    }
+
+    @Override
+    protected void addRoutes(final Router router) {
+        setupAuthorization(router);
+
+        for (final LoraProvider current : loraProviders) {
+            router.route(HttpMethod.OPTIONS, current.pathPrefix()).handler(this::handleOptionsRoute);
```
sophokles73: what if several providers return the same *pathPrefix*?

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraProtocolAdapter.java
```diff
@@ -0,0 +1,636 @@
...
+                    handle202(ctx);
+                }
+            } catch (final ClassCastException | LoraProviderMalformedPayloadException e) {
+                LOG.debug("Got invalid payload '{}' which leads to exception: {}", loraMessage, e);
+                TracingHelper.logError(currentSpan,
+                        "Received message of type '" + type + "' has invalid payload; error: " + e);
+                handle400(ctx, INVALID_PAYLOAD);
+            }
+        } else {
+            LOG.debug("Supplied credentials are not an instance of the user. Returning 401");
```
sophokles73: I think the log message needs to be rephrased. It doesn't make much sense this way ...

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraProtocolAdapter.java
```diff
@@ -0,0 +1,636 @@
...
+                    doUpload(ctx, gatewayDevice, deviceId, payloadBase64);
+                } else {
+                    LOG.debug("Received message '{}' of type [{}] for device [{}], will discard message.", loraMessage,
+                            type, deviceId);
+                    currentSpan.log(
+                            "Received message of type '" + type + "' for device '" + deviceId + "' will be discarded.");
+                    // Throw away the message but return 202 to not cause errors on the LoRa provider side
+                    handle202(ctx);
+                }
+            } catch (final ClassCastException | LoraProviderMalformedPayloadException e) {
```
sophokles73: I assume that `ClassCastException` is thrown if the payload does not contain fields of the expected type. Wouldn't it be much cleaner to also throw  `LoraProviderMalformedPayloadException` in that case?

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraProtocolAdapter.java
```diff
@@ -0,0 +1,636 @@
...
+
+    private Span getCurrentSpan(final RoutingContext ctx) {
+        return (ctx.get(TracingHandler.CURRENT_SPAN) instanceof Span) ? ctx.get(TracingHandler.CURRENT_SPAN)
+                : NoopSpan.INSTANCE;
+    }
+
+    void handleOptionsRoute(final RoutingContext ctx) {
+        LOG.debug("Handling options method");
+
+        if (ctx.user() instanceof Device) {
```
sophokles73: IMHO this check does not add much value because you never use the `Device` data. The request will already have been rejected with a 401 if the provider didn't provide valid credentials, won't it?

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraProtocolAdapter.java
```diff
@@ -0,0 +1,636 @@
...
+    private void commandConsumer(final String tenantId, final CommandContext receivedCommandContext) {
+        Tags.COMPONENT.set(receivedCommandContext.getCurrentSpan(), getTypeName());
+        final Command command = receivedCommandContext.getCommand();
+        final CommandData commandData = new CommandData();
+        commandData.setCommand(command);
+
+        getTenantConfiguration(tenantId, receivedCommandContext.getCurrentSpan().context()).compose(tenantObject -> {
+            if (tenantObject.isAdapterEnabled(getTypeName())) {
+                final String logMsg = "Adapter " + getTypeName() + " is enabled for tenant " + tenantId;
+                LOG.debug(logMsg);
```
sophokles73: please construct the message only if DEBUG logging is enabled

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraProtocolAdapter.java
```diff
@@ -0,0 +1,636 @@
...
+                    closeCommandConsumer(tenantId, LORA_COMMAND_CONSUMER_DEVICE_ID);
+                }, LORA_COMMAND_CONSUMER_RETRY_INTERVAL);
+    }
+
+    @Override
+    protected void onStartupSuccess() {
+        scheduleStartLoraCommandConsumers();
+    }
+
+    private void commandConsumer(final String tenantId, final CommandContext receivedCommandContext) {
```
sophokles73: you seem to already have the command direction implemented as well, is that right?

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraProtocolAdapter.java
```diff
@@ -0,0 +1,636 @@
...
+                        .otherwise(loraGatewayException -> {
+                            LOG.error("Error getting lora device gateway. Sending error response to application with "
+                                    + "code [{}]", HTTP_INTERNAL_ERROR, loraGatewayException);
+                            TracingHelper.logError(receivedCommandContext.getCurrentSpan(), loraGatewayException);
+                            sendResponseToApplication(command, loraDeviceId,
+                                    getHttpResponseWithCode(HTTP_INTERNAL_ERROR,
+                                            loraGatewayException.getMessage()),
+                                    receivedCommandContext.getCurrentSpan().context());
+                            return null;
+                        });
```
sophokles73: I hope that we can make this more readable ...

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraProtocolAdapter.java
```diff
@@ -0,0 +1,636 @@
...
+            receivedCommandContext
+                    .reject(new ErrorCondition(Constants.AMQP_BAD_REQUEST, validationException.getMessage()), 1);
+            return null;
+        });
+    }
+
+    private static boolean isValidLoraCommand(final Command command) {
+        try {
+            final String payload = command.getPayload().toJsonObject()
+                    .getString(LoraConstants.FIELD_LORA_DOWNLINK_PAYLOAD);
```
sophokles73: FMPOV the command's payload *is* the payload to be sent. Why would it be encapsulated in a JSON String property?

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraProtocolAdapter.java
```diff
@@ -0,0 +1,636 @@
...
+                final String logMsg = "Adapter " + getTypeName() + " is not enabled for tenant " + tenantId;
+                LOG.error(logMsg);
+                return Future.failedFuture(logMsg);
+            }
+        }).compose(adapterEnabledForTenant -> {
+            // we accept the message before sending it to the device in order to provide a consistent behavior
+            // across all protocol adapters, i.e. accepting the message only means that the message contains
+            // a valid command which we are willing to deliver see
+            // org.eclipse.hono.adapter.http.AbstractVertxBasedHttpProtocolAdapter.createCommandConsumer
+            receivedCommandContext.accept(1);
```
sophokles73: IMHO we should at least try to send the command to the provider and accept or reject based on the provider's response ...

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraCommandProperties.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.List;
+
+/**
+ * Properties for LoRaWAN commands.
+ */
+public class LoraCommandProperties {
+
+    private final List<String> commandEnabledTenants = new LinkedList<>();
+
+    public List<String> getCommandEnabledTenants() {
```
calohmn: A javadoc comment would be good here.

The fact that the Command&Control feature of the LoRa adapter is restricted to these tenants should be pointed out in the (to be written) documentation and could also be mentioned in the commit comment.

bom/pom.xml
```diff
@@ -31,6 +31,8 @@
     <artemis.image.name>enmasseproject/activemq-artemis:2.2.0-4</artemis.image.name>
     <assertj-core.version>3.11.1</assertj-core.version>
     <californium.version>2.0.0-M13</californium.version>
+    <commons-codec.version>1.10</commons-codec.version>
+    <commons-lang.version>2.6</commons-lang.version>
```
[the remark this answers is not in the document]
mbaeuerle: `Lang` is used for `StringUtils.isBlank`.
`Codec` is used for `Base64.decode` etc. but we should be able to use the one from the std lib.

bom/pom.xml
```diff
@@ -59,6 +61,7 @@
     <spring-security-crypto.version>4.2.11.RELEASE</spring-security-crypto.version>
     <truth.version>0.28</truth.version>
     <vertx.version>3.6.3</vertx.version>
+    <wiremock.version>2.19.0</wiremock.version>
```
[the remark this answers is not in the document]
mbaeuerle: It is used to test the LoRa http requests by mocking the actual request.

bom/pom.xml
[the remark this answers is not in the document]
sophokles73: if `StringUtils.isBlank` basically does
```
s == null || s.length() = 0
```
then you can use `org.eclipse.hono.util.Strings.isNullOrEmpty()` instead ...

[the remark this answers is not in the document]
sophokles73: Please do use Java8's standard Base64 encoder instead.

sophokles73:
@mbaeuerle any news on this PR?

bom/pom.xml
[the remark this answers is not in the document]
mbaeuerle: StringUtils.isBlank also is `true` for whitespace only strings (i.e. "   ") but I can replace it with something similar in the LoraUtils.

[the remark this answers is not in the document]
sophokles73: so is `org.eclipse.hono.util.Strings.isNullOrEmpty()`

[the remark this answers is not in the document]
mbaeuerle: The Commons-codec dependency is also used for converting hex i.e. `Hex.encodeString` which is not that easy to replace with the Java standard lib. Do you know an alternative approach?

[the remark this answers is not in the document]
sophokles73: well, we could implement two simple methods for doing that in Hono's `Strings` class. Hex encoding is not rocket science, is it?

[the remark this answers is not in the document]
mbaeuerle: I have implemented the required hex conversion methods to remove common-codec!

adapters/lora-vertx/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+            <artifactId>vertx-web-client</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.github.tomakehurst</groupId>
+            <artifactId>wiremock</artifactId>
```
sophokles73: please remove the tests that require this library and this dependency for the time being. We can add the tests later in a second PR, but we need to create a CQ for this library and have it approved first ...

sophokles73:
@mbaeuerle the Travis build fails constantly for this PR. Can you please check? May be rebasing to head of master would already do the trick ...

mbaeuerle:
Well the last time I checked on Monday, the build was still in waiting state, seems like it took Travis nearly a day to build it at all. Checking now!

PR #1080, opened 2019-03-13
Upgrade micrometer to 1.1.x, specifically 1.1.3

ctron:


bom/pom.xml
```diff
@@ -44,7 +44,7 @@
     <jmeter.version>3.3</jmeter.version>
     <junit.version>4.12</junit.version>
     <logback.version>1.2.3</logback.version>
-    <micrometer.version>1.0.9</micrometer.version>
+    <micrometer.version>1.1.3</micrometer.version>
```
sophokles73: there's already version 1.1.4 available ...

ctron:
> just out of curiosity: do you intend to use any of the new features in 1.1.x?

Not really. However we would like to use 1.1.x internally. And there is a small glitch in the API, which is already fixed in a compatible way.

But I think it makes sense to stay up to date, on 1.1.x.

bom/pom.xml
ctron: Looks like this came out after I filed the CQs, I will update the PR.

PR #1081, opened 2019-03-13
Language level 11 in code style.

b-abel:
Changed properties setting to Java 1.8 language level to 11.

Needs testing in Eclipse.
I did not examine if there are other properties that should be changed or added for Java 11.

### DIFF
--- a/eclipse/hono-code-style.xml
+++ b/eclipse/hono-code-style.xml
@@ -176,7 +176,6 @@
 <setting id="org.eclipse.jdt.core.formatter.parentheses_positions_in_try_clause" value="common_lines"/>
 <setting id="org.eclipse.jdt.core.formatter.alignment_for_selector_in_method_invocation" value="16"/>
 <setting id="org.eclipse.jdt.core.formatter.never_indent_block_comments_on_first_column" value="false"/>
-<setting id="org.eclipse.jdt.core.compiler.source" value="1.8"/>
 <setting id="org.eclipse.jdt.core.formatter.insert_space_after_opening_paren_in_synchronized" value="do not insert"/>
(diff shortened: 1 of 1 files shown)


eclipse/hono-code-style.xml
```diff
@@ -176,7 +176,7 @@
 <setting id="org.eclipse.jdt.core.formatter.parentheses_positions_in_try_clause" value="common_lines"/>
 <setting id="org.eclipse.jdt.core.formatter.alignment_for_selector_in_method_invocation" value="16"/>
 <setting id="org.eclipse.jdt.core.formatter.never_indent_block_comments_on_first_column" value="false"/>
-<setting id="org.eclipse.jdt.core.compiler.source" value="1.8"/>
+<setting id="org.eclipse.jdt.core.compiler.source" value="11"/>
```
sophokles73: I have set the compliance level in Eclipse to Java 11. When I then export our code style, the file does not contain any of the lines indicating a particular Java version. I suspect that these lines are only added if the compliance level is different from the level of the JDK used for building the project. Can you verify if simply removing these lines (instead of replacing `1.8` with `11` works with your plugin in IDEA as well?

b-abel: Yes. That works. 
Note: In the setting of Eclipse Code Formatter plugin for IntelliJ IDEA the "Java Formatter Version" must be > Eclipse 4.5.0

sophokles73: Can you then please update your PR accordingly, i.e. remove those lines completely?

b-abel: Sure. Have you checked if anything else is changed when you export the code style?

sophokles73: No, I haven't.

Issue #1066, opened 2019-02-22
[User Guide] Add Examples for AMQP with Client Certificates


b-abel:
The User Guide for the AMQP adapter does not contain examples for the usage of client certificate based authentication (as it does for the MQTT and the HTTP adapter). 
It is probably necessary to add this feature to the CLI application first.

sophokles73:
@b-abel can we (i.e. you) close this?

PR #1082, opened 2019-03-13
[#1066] Add client certificate based authentication to CLI.

b-abel:
The AMQP CLI application can now be used for client certificate based
authentication against the AMQP protocol adapter.
+ Added example to User Guide.

### DIFF
--- a/cli/src/main/java/org/eclipse/hono/cli/AmqpSend.java
+++ b/cli/src/main/java/org/eclipse/hono/cli/AmqpSend.java
@@ -164,7 +164,15 @@ public class AmqpSend extends AbstractCliClient {
                     properties.getPassword(),
                     connectAttempt);
         } else {
-            // SASL ANONYMOUS auth
+            if (properties.getKeyCertOptions() != null && properties.getTrustOptions() != null) {
+                // SASL EXTERNAL auth
+                options.setSsl(true);
+                options.setKeyCertOptions(properties.getKeyCertOptions());
+                options.setTrustOptions(properties.getTrustOptions());
+            } else {
+                // SASL ANONYMOUS auth
+            }
+
             LOG.info("connecting to AMQP adapter [host: {}, port: {}]", properties.getHost(), properties.getPort());
             client.connect(options, properties.getHost(), properties.getPort(), connectAttempt);
         }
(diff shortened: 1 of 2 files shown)


site/content/user-guide/amqp-adapter.md
```diff
@@ -118,6 +118,10 @@ Publish some JSON data for device `4711`:
 
 Notice that we only supplied a new value for the message address, leaving the other default values.
 
+Publish some JSON data for device `4711` using a client certificate for authentication:
+
+    ~/hono/cli/target$ java -jar hono-cli-*-exec.jar --spring.profiles.active=amqp-adapter-cli --hono.client.port=5671 --hono.client.certPath=config/hono-demo-certs-jar/device-4711-cert.pem --hono.client.keyPath=config/hono-demo-certs-jar/device-4711-key.pem --hono.client.trustStorePath=config/hono-demo-certs-jar/trusted-certs.pem --hono.client.hostnameVerificationRequired=false
```
sophokles73: is there a reason why the `client-cert` profile is not used here?

b-abel: Like in the other examples in the User Guide of the AMQP adapter, the relevant configuration is given as arguments. I just could not resist to add a profile for client certificates to the application.yml. I would rather remove the profile `client-cert`, than having an example which is implicit as : `~/hono/cli/target$ java -jar hono-cli-*-exec.jar --spring.profiles.active=client-cert`.

sophokles73: I agree, in particular since the profile is not really an *add-on* like the *ssl* profile. So, can you please remove the *client-cert* profile then?

b-abel: Done

cli/src/main/java/org/eclipse/hono/cli/AmqpSend.java
```diff
@@ -44,7 +44,7 @@
  * A simple command-line client for interacting with the AMQP adapter.
  */
 @Component
-@Profile("amqp-adapter-cli")
+@Profile({ "amqp-adapter-cli", "client-cert" })
```
sophokles73: this should go as well, I guess ...

b-abel: Uh, sorry. I removed it now.

sophokles73:
@b-abel can you please squash and rebase?

b-abel:
> @b-abel can you please squash and rebase?

Done

PR #1090, opened 2019-03-18
Fix Javadoc

calohmn:


calohmn:
@sophokles73 Was the intention of [this commit](https://github.com/bsinno/hono/commit/6bbbe7d96951ce5ed8481cc213835dd9f7b8f729#diff-8494b204d6ebc117cf63926d61d702dbR299) to actually return a failed future with an exception in `handleUnimplementedOperation`?
That would be more intuitive (not sure about all the implications though).

Issue #780, opened 2018-08-14
Simplify and split HonoClient for applications versus protocol adapters


sysexcontrol:
Originating from a discussion on the dev mailing list, I think we should split the HonoClient to at least two different use cases:

- what typical **applications** need
  - telemetry and event consumers
  - command senders

- what (custom) **protocol adapters** need
  - telemetry and event senders
  - tenant, credentials and registration clients

I am sure this needs some discussion - I can imagine that the **tenant, credentials and registration** clients might be desirable to have available outside a protocol adapter as well, so that we would have three different clients.
A use case for that could be a device registry client application IMHO.

The point of this issue is to come to a more focussed HonoClient, so that an application developer should not have to ask a question like "why do I have a telemetry sender available, but I cannot use it since I can only consume telemetry messages?". The same question could be asked for the tenant, credentials and registration clients, for which the remote APIs are usually not accessible for applications.

After having discussed this issue, we can rewrite it to describe what we want to achieve in the future with HonoClient (and possibly split it up).

Remark: for Command and Control we already made a small step into this direction : the `CommandConnection` is an extension of `HonoClient` and thus does not provide it's methods to applications.

avgustinmm:
I think that current approach is not consistent - e.g; having telemetry & event senders but not command consumers and command response senders. 
The Hono Client should take one of the following approaches:
* contain ALL APIs, including protocol adapter C&C. E.g. hono client will stand for all-in-one API/util for accessing Hono core.
* contain only application related APIs. I.e. remove telemetry&event senders. It may or may not keep device/credential and tenant APIs. In this case a protocol adapter API shall be implemented that contains telemetry&event sender as well as device/credential and tenant APIs. If the message senders and admin API will be separated is up not a big difference to me.

sysexcontrol:
> I think that current approach is not consistent - e.g; having telemetry & event senders but not command consumers and command response senders.

Hi @avgustinmm : as I wrote we have the `CommandConnection` available (which is a `HonoClient` already). This separates a little on the adapter side.

However, we agree on that the hono client needs to be more clearly separated for the needs of
- applications
- and protocol adapters

Now it is separated by the _features_ which undoubtedly is not good.

I propose to have an interface hierarchy for this purpose:

`HonoApplicationClient extends HonoClient`

`HonoAdapterClient extends HonoClient`

IMHO this would lead to three maven modules (instead of one that we have now):

- `hono-client` : containing generic parts (like the `connect` method and so on)
[...]

PR #1092, opened 2019-03-18
[#780] Extract dedicated client factory interfaces.

sophokles73:
The factory methods for creating the arbitrary service clients have been
refactored into dedicated interfaces.

This is the first step in splitting up `HonoClient` into multiple factories with dedicated scope. In subsequent PRs, I intend to
* rename `HonoClient` to `HonoConnection` and limit its responsibility to just the establishment and maintenance of the AMQP connection and serve as a factory for sender and receiver links
* let the dedicated factories *wrap* around a `HonoConnection`
* refactor protocol adapters to use the dedicated factories instead of `HonoConnection` and be no longer directly involved with connection management

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/ConnectionLifecycle.java
+++ b/client/src/main/java/org/eclipse/hono/client/ConnectionLifecycle.java
@@ -1,0 +1,71 @@
+/**
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+
+package org.eclipse.hono.client;
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+
+/**
+ * Provides access to the life cycle events of a connection to a Hono service.
+ *
+ */
+public interface ConnectionLifecycle {
+
+    /**
+     * Adds a listener to be notified when the connection is lost unexpectedly.
+     * 
+     * @param listener The listener to add.
+     */
+    void addDisconnectListener(DisconnectListener listener);
+
+    /**
+     * Adds a listener to be notified when the connection has been re-established after
+     * it had been lost unexpectedly.
+     * 
+     * @param listener The listener to add.
+     */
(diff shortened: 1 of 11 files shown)


client/src/main/java/org/eclipse/hono/client/ConnectionLifecycle.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ */
+public interface ConnectionLifecycle {
+
+    /**
+     * Adds a handler to be notified when the connection is lost unexpectedly.
+     * 
+     * @param handler The handler.
+     */
+    void addDisconnectHandler(Handler<Void> handler);
```
ctron: I am not so fond of those generic `Handler<Void>` types, Couldn't we add our own ones, providing a proper interfaces. And mapping this internally to `Handler<Void>`?

sophokles73: You mean an interface like 
```
interface DisconnectHandler {
  void onDisconnect();
}
```
?

sophokles73:
@ctron I have added two functional listener interfaces to replace the generic handlers. WDYT?

client/src/main/java/org/eclipse/hono/client/DisconnectListener.java
```diff
@@ -0,0 +1,30 @@
...
+
+
+package org.eclipse.hono.client;
+
+
+/**
+ * A listener to be notified when a connection is lost unexpectedly.
+ *
+ */
+public interface DisconnectListener {
```
ctron: I would annotate those with `@FunctionalInterface` … which lets the compiler enforce all requirements:

```suggestion
@FunctionalInterface
public interface DisconnectListener {
```

client/src/main/java/org/eclipse/hono/client/ReconnectListener.java
```diff
@@ -0,0 +1,32 @@
...
+
+package org.eclipse.hono.client;
+
+
+/**
+ * A listener to be notified when a connection is re-established after
+ * it has been lost unexpectedly.
+ *
+ */
+public interface ReconnectListener {
```
ctron: Same here:

```suggestion
@FunctionalInterface
public interface ReconnectListener {
```

client/src/main/java/org/eclipse/hono/client/ConnectionLifecycle.java
ctron: Yes. I would prefer that over `Handler<Void>`. Alternatively you could also use something more generic like `VoidHandler` (or some other name). My primary concern here is the useless `Void` object type.

sophokles73:
@ctron WDYT?

PR #1093, opened 2019-03-18
[#930] Add "last-via" property to device data on "assertRegistration"

calohmn:
First part of the solution for #930:
When "assertRegistration" is called on a device that has multiple "via"
gateways configured, the id of the gateway doing this request is put
into a "last-via" property in the device's registration data.

Example of the updated device registration data:
````
      {
        "device-id": "4712",
        "data": {
          "enabled": true,
          "via": ["gw-1", "gw-2"],
          "last-via": {
            "device-id": "gw-1",
            "update-date": "2019-03-18T15:19:28.186970Z"
          }
        }
      }
````

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
@@ -16,6 +16,7 @@ import java.net.HttpURLConnection;
 import java.util.Objects;
 
 import org.eclipse.hono.client.ClientErrorException;
+import org.eclipse.hono.client.ServiceInvocationException;
 import org.eclipse.hono.service.EventBusService;
 import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.CacheDirective;
@@ -61,9 +62,19 @@ public abstract class BaseRegistrationService<T> extends EventBusService<T> impl
 
     /**
      * The name of the field in a device's registration information that contains
-     * the identifier of the gateway that it is connected to.
+     * the identifier of the gateway that it is connected to (either as string value or inside a JSON array). 
      */
     public static final String PROPERTY_VIA = "via";
+    /**
+     * The name of the field in a device's registration information that contains
+     * the identifier of the gateway that it was last connected to as well as the date when this information was updated.
+     */
+    public static final String PROPERTY_LAST_VIA = "last-via";
+    /**
+     * The name of the field in a device's registration information that contains
+     * the date when the 'last-via' device id was last updated.
+     */
+    public static final String PROPERTY_LAST_VIA_UPDATE_DATE = "update-date";
 
     private static final String SPAN_NAME_ASSERT_DEVICE_REGISTRATION = "assert Device Registration";
 
@@ -292,10 +303,16 @@ public abstract class BaseRegistrationService<T> extends EventBusService<T> impl
                 final JsonObject gatewayData = gatewayResult.getPayload().getJsonObject(RegistrationConstants.FIELD_DATA, new JsonObject());
 
                 if (isGatewayAuthorized(gatewayId, gatewayData, deviceId, deviceData)) {
-                    return Future.succeededFuture(RegistrationResult.from(
-                        HttpURLConnection.HTTP_OK,
-                        getAssertionPayload(tenantId, deviceId, deviceData),
-                        CacheDirective.maxAgeDirective(assertionFactory.getAssertionLifetime())));
+                    return updateDeviceLastViaIfNeeded(tenantId, deviceId, gatewayId, deviceData).map(res -> {
+                        return RegistrationResult.from(
+                                HttpURLConnection.HTTP_OK,
+                                getAssertionPayload(tenantId, deviceId, deviceData),
+                                CacheDirective.maxAgeDirective(assertionFactory.getAssertionLifetime()));
+                    }).recover(t -> {
+                        log.error("update of the 'last-via' property failed", t);
+                        TracingHelper.logError(span, "update of the 'last-via' property failed: " + t.toString());
+                        return Future.succeededFuture(RegistrationResult.from(ServiceInvocationException.extractStatusCode(t)));
+                    });
                 } else {
                     TracingHelper.logError(span, "gateway not authorized");
                     return Future.succeededFuture(RegistrationResult.from(HttpURLConnection.HTTP_FORBIDDEN));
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java [hunk 1]
```diff
@@ -304,6 +322,50 @@ public void assertRegistration(
...
+     * that have multiple gateway entries in their 'via' definition.
+     *
+     * @param tenantId The tenant id.
+     * @param deviceId The device id.
+     * @param gatewayId The gateway id.
+     * @param deviceData The current data associated with the device.
+     * @param resultHandler The handler to invoke with the result of the operation.
+     */
+    protected void updateLastVia(final String tenantId, final String deviceId, final String gatewayId,
+            final JsonObject deviceData, final Handler<AsyncResult<RegistrationResult>> resultHandler) {
```
sophokles73: I do not quite understand why this method requires a `RegistrationResult` handler. FMPOV this method should simply return a `Future<Void` so that it can easily be composed with other futures ....

service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
```diff
@@ -304,6 +322,50 @@ public void assertRegistration(
         }).setHandler(resultHandler);
     }
 
+    private Future<RegistrationResult> updateDeviceLastViaIfNeeded(final String tenantId, final String deviceId,
```
sophokles73: Why not simply return `Future<Void>`? Why is this method required to create the result of the overall operation it is invoked as part of (and which this operation doesn't need to know anything about)?

service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
calohmn: My motivation to add the Handler as parameter (instead of returning a Future) was to keep the method signature consistent with the other methods of this class (addDevice/updateDevice/removeDevice). But I'm also ok with returning a Future instead.

calohmn: It's now changed to return `Future<Void>`.

service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
calohmn: It's now changed to return `Future<Void>`.

service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
```diff
@@ -304,6 +322,43 @@ public void assertRegistration(
...
+     *
+     * @param tenantId The tenant id.
+     * @param deviceId The device id.
+     * @param gatewayId The gateway id.
+     * @param deviceData The current data associated with the device.
+     * @return A future indicating whether the operation succeeded or not.
+     */
+    protected Future<Void> updateDeviceLastVia(final String tenantId, final String deviceId, final String gatewayId,
+            final JsonObject deviceData) {
+        return Future.failedFuture(new ServerErrorException(HttpURLConnection.HTTP_NOT_IMPLEMENTED));
```
sophokles73: I wonder if the default implementation should not return a succeeded future instead. The authorization check has already been performed so this method basically represents *optional* behavior. If a subclass wants to support it, then it needs to override it with an appropriate implementation. However, if not, then nothing would need to be done.
On the other hand, if we want to force people to think about this feature when creating a subclass, then we should define this method as abstract. WDYT?

calohmn: I would rather make the method abstract then, describing in the javadoc that a proper implementation is needed for getting command & control to work in the scenario described in #930 (ie. device with multiple `via` gateways; gateways subscribing to commands without specifying the device id).
If such a scenario is not important for someone implementing the device registry service, the method can then be implemented by just returning a succeeded future.

calohmn: I've amended the commit accordingly.

service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
```diff
@@ -61,9 +62,19 @@
...
+    /**
+     * The name of the field in a device's registration information that contains
+     * the identifier of the gateway that it was last connected to as well as the date when this information was updated.
+     */
+    public static final String PROPERTY_LAST_VIA = "last-via";
+    /**
+     * The name of the field in a device's registration information that contains
+     * the date when the 'last-via' device id was last updated.
+     */
+    public static final String PROPERTY_UPDATE_DATE = "update-date";
```
BobClaerhout: Wouldn't it be better to name the variable so that it is clear that this update_date is meta info about the last-via property? I'm not good in naming but maybe something like: PROPERTY_UPDATE_DATE_LAST_VIA?

calohmn: yes, right, that also fits better to the javadoc content. I've named it PROPERTY_LAST_VIA_UPDATE_DATE now.

service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationService.java
```diff
@@ -115,6 +115,10 @@ default void assertRegistration(final String tenantId, final String deviceId, fi
      * Implementing classes should verify, that the gateway is authorized to get an assertion for the device.
      * Such a check might be based on a specific role that the client needs to have or on an
      * explicitly defined relation between the gateway and the device(s).
+     * <br>
+     * In the case of a device configured with multiple <em>via</em> gateways, implementing classes should
+     * update the device's registration information with the given gateway in the form of a <em>last-via</em>
+     * property.
```
sophokles73: isn't this relevant to note in the JavaDoc of the other *assertRegistration* variant as well?

calohmn: right, I've added it there as well.

calohmn:
When thinking about the implementation of the gateway mapping logic for #930, I came upon this question:
Suppose a device is configured with one or more `via` gateways. Currently, such a device can also connect **directly** to a protocol adapter (not using a gateway) and receive commands.
Do we still want to support this?

I tend to think it would be preferably to support this. WDYT? 

In any case, changes are needed concerning this: either preventing direct connections or changing the logic regarding storing the `last-via` information.

BobClaerhout:
I do not think that devices which are connecting trough gateways, also connect to the adapter. E.g.: a device connects using LoRaWAN to the gateway, the gateway communicates with the LoRaWAN server which communicates with the hono adapter. That specific device will not connect to the hono adapter itself, it does not have the credentials nor the capabilities to connect with the adapter.
I think this is more or less the same for all cases where devices are connecting through gateways. They probably do not have the credentials, communication abilities, technology abilities nor software to directly communicate with the adapter. Maybe you have another opinion on that?
So FMPOV it is allowed to prevent direct connections. On the other hand, when allowing these direct connections, you support more use cases (also the ones I do not thin about right now) which is probably better.

sophokles73:
> Currently, such a device can also connect directly to a protocol adapter (not using a gateway) and receive commands. Do we still want to support this?

I do not quite understand what would need to be changed in order to (still) support this.

calohmn:
> I do not quite understand what would need to be changed in order to (still) support this.

Currently we don't write the `last-via` property in case of a request directly from the device (not through a gateway). If a device previously was connecting via a gateway and now connects directly, we would have to save this info somehow. Otherwise there would be no way of knowing that an incoming command for this device should now **not** be mapped to the gateway anymore.

So, to be changed:
The `last-via` property would be written for _every_ `assertRegistration` request concerning a device that has multiple `via` entries.
In case of a direct connection, the device-id would be kept empty:
````
{
  "device-id": "4712",
  "data": {
    "enabled": true,
    "via": ["gw-1", "gw-2"],
    "last-via": {
      "device-id": "",
      "update-date": "2019-03-18T15:19:28.186970Z"
    }
  }
}
````

BobClaerhout:
I suggest filling in the device-id in the "last-via" property with the device-id if the device send messages directly. An empty field could confuse someone while filling it in with the device-id states that it was last connected by itself which is correct.

calohmn:
Right. It's not really a "via" then, but better to have a value in there.

calohmn:
PR for that is #1107.

Issue #623, opened 2018-05-22
Support multiple trusted-ca objects in Tenant API


pho-enix:
Tenant API should support multiple trusted-ca objects. This is needed for those use-cases:

* Within a tenant, devices with certificates from different CAs should be supported
* Key Rotation: If a CA certificate expires and a new one gets deployed, during a certain amount of time, both certificates should be accepted

I would recommend to change the data schema to something like:

```
"trusted-ca": [
  {
    "subject-dn": "CN=devices,O=ACME Corporation",
    "public-key": "aGFsbG8gT21hCg==",
    "not-before": "2015-01-01T00:00:00+0000",
    "not-after": "2025-01-01T00:00:00+0000"
  },
  {
    "subject-dn": "CN=devices,O=ACME Corporation",
    "public-key": "aGFsbG8gT21hMgo=",
    "not-before": "2024-01-01T00:00:00+0000",
    "not-after": "2034-01-01T00:00:00+0000"
  },
}
```

boaks:
I would like to have at least the possibility to use a `trusted-ca` specific for an connector. 
Especially, if "none-tls" connectors are in use, the may introduce their own "most common used" asymmetric variant, e.g. ECC for "coap over DTLS".

Alfusainey:
@sophokles73 @pho-enix In the example above, the two `trusted-cas` have the same `subject-dn` i.e are issued by the same issuer. Is this right?

I was just wondering whether it is possible for two (or more) trust anchors to have the same `subject-dn` configured for a tenant?

sophokles73:
@Alfusainey I do not see why not. In fact, the trust anchor itself is just the public key. The subject DN is simply the identifier that we use to determine the tenant.

b-abel:
> I would like to have at least the possibility to use a `trusted-ca` specific for an connector.
> Especially, if "none-tls" connectors are in use, the may introduce their own "most common used" asymmetric variant, e.g. ECC for "coap over DTLS".

I'm trying to understand the requirements here. 
@boaks Is it really necessary to restrict those specific `trusted-ca` to the corresponding protocol adapter? Or would it be OK to add it to the `trusted-ca` array?
Or is your point about the schema of such a specific `trusted-ca`? Maybe you can provide an example...

Alfusainey:
@b-abel are you also working on this? I already started working on this and might file a PR for it this weekend :-)

b-abel:
[...]

PR #1095, opened 2019-03-19
[#623] TenantService: Add support for multiple trusted certificate au…

Alfusainey:
This is an **initial PR** that addresses #623 . i disabled about two tests for now in the IT test module (i.e in the http and amqp adapters) because now the base tenant implementation rejects a tenant add request with payload containing trust configurations whose public key or cert property values are not base64 encoded

i decided to file this PR to get initial feedback before changing the http/amqp adapters (or adding integration tests)

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/TenantConstants.java
+++ b/core/src/main/java/org/eclipse/hono/util/TenantConstants.java
@@ -64,9 +64,9 @@ public final class TenantConstants extends RequestResponseApiConstants {
      */
     public static final String FIELD_PAYLOAD_PUBLIC_KEY = "public-key";
     /**
-     * The name of the property that contains the list of trusted certificate authorities configured for this tenant.
+     * The name of the property that contains the list of trusted certificate authorities configured for a tenant.
      */
-    public static final String FIELD_PAYLOAD_TRUSTED_CA = "trust-ca";
+    public static final String FIELD_PAYLOAD_TRUSTED_CA = "trusted-ca";
 
     /**
      * The name of the Tenant API endpoint.
--- a/core/src/main/java/org/eclipse/hono/util/TenantConstants.java
+++ b/core/src/main/java/org/eclipse/hono/util/TenantConstants.java
@@ -64,9 +64,9 @@ public final class TenantConstants extends RequestResponseApiConstants {
      */
     public static final String FIELD_PAYLOAD_PUBLIC_KEY = "public-key";
     /**
-     * The name of the property that contains the trusted certificate authority configured for a tenant.
+     * The name of the property that contains the list of trusted certificate authorities configured for this tenant.
      */
-    public static final String FIELD_PAYLOAD_TRUSTED_CA = "trusted-ca";
+    public static final String FIELD_PAYLOAD_TRUSTED_CA = "trust-ca";
 
     /**
      * The name of the Tenant API endpoint.
--- a/core/src/main/java/org/eclipse/hono/util/TenantObject.java
+++ b/core/src/main/java/org/eclipse/hono/util/TenantObject.java
@@ -25,13 +25,17 @@ import java.security.cert.TrustAnchor;
 import java.security.cert.X509Certificate;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.X509EncodedKeySpec;
+import java.util.ArrayList;
 import java.util.Base64;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import javax.security.auth.x500.X500Principal;
 
@@ -54,7 +58,14 @@ public final class TenantObject extends JsonBackedValueObject {
     @JsonIgnore
     private Map<String, JsonObject> adapterConfigurations;
     @JsonIgnore
-    private TrustAnchor trustAnchor;
+    private List<JsonObject> trustedCaList;
+    @JsonIgnore
+    private List<TrustAnchor> trustAnchors;
+    @JsonIgnore
+    private List<X500Principal> subjectNames;
+    //<subjectDn, configs>
+    @JsonIgnore
+    private Map<String, List<JsonObject>> trustConfigurations;
 
     /**
      * Adds a property to this tenant.
@@ -114,28 +125,23 @@ public final class TenantObject extends JsonBackedValueObject {
 
     /**
      * Gets the subject DN of this tenant's configured trusted
-     * certificate authority.
+     * certificate authority. This method assumes that this tenant is
+     * configured with a single trusted CA.
      * 
      * @return The DN or {@code null} if no CA has been set.
      */
     @JsonIgnore
     public X500Principal getTrustedCaSubjectDn() {
-
-        final JsonObject trustedCa = getProperty(TenantConstants.FIELD_PAYLOAD_TRUSTED_CA);
-        if (trustedCa == null) {
-            return null;
-        } else {
-            return Optional.ofNullable((String) getProperty(trustedCa, TenantConstants.FIELD_PAYLOAD_SUBJECT_DN))
-                    .map(dn -> new X500Principal(dn)).orElse(null);
-        }
+        final List<X500Principal> subjectdns = Optional.ofNullable(getTrustedCaSubjectNames()).orElse(null);
+        return subjectdns == null ? null : subjectdns.get(0);
     }
 
     /**
-     * Sets the trusted certificate authority to use for authenticating
-     * devices of this tenant.
+     * Sets the trusted anchor for this tenant's Trusted CA using the given public key and subject DN.
      * 
-     * @param publicKey The CA's public key.
-     * @param subjectDn The CA's subject DN.
+     * @param publicKey The trusted CA's public key.
+     * @param subjectDn The trusted CA's subject DN.
+     * 
      * @return This tenant for command chaining.
      * @throws NullPointerException if any of the parameters is {@code null}.
      */
@@ -150,11 +156,14 @@ public final class TenantObject extends JsonBackedValueObject {
         trustedCa.put(TenantConstants.FIELD_PAYLOAD_PUBLIC_KEY, publicKey.getEncoded());
         trustedCa.put(TenantConstants.FIELD_PAYLOAD_KEY_ALGORITHM, publicKey.getAlgorithm());
         return setProperty(TenantConstants.FIELD_PAYLOAD_TRUSTED_CA, trustedCa);
+
+        // return addTrustedCaConfiguration(trustedCa);
     }
 
     /**
      * Sets the trusted certificate authority to use for authenticating
-     * devices of this tenant.
+     * devices of this tenant. The certificate must contain a {@code NonNull} and non-empty
+     * subject DN value.
      * 
      * @param certificate The CA certificate.
      * @return This tenant for command chaining.
@@ -168,15 +177,17 @@ public final class TenantObject extends JsonBackedValueObject {
 
         try {
             final JsonObject trustedCa = new JsonObject()
+                    .put(TenantConstants.FIELD_PAYLOAD_SUBJECT_DN, certificate.getSubjectX500Principal().getName())
                     .put(TenantConstants.FIELD_PAYLOAD_CERT, certificate.getEncoded());
-            return setProperty(TenantConstants.FIELD_PAYLOAD_TRUSTED_CA, trustedCa);
-        } catch (CertificateEncodingException e) {
+            return addTrustedCaConfiguration(trustedCa);
+        } catch (final CertificateEncodingException e) {
             throw new IllegalArgumentException("cannot encode certificate");
         }
     }
 
     /**
      * Gets the trusted certificate authority configured for this tenant.
+     * This method assumes that the tenant is configured with a single Trusted CA.
      * <p>
      * This method tries to extract the certificate from the data contained in
      * the JSON object of the <em>trusted-ca</em> property.
@@ -185,39 +196,31 @@ public final class TenantObject extends JsonBackedValueObject {
      * value as the <em>printable</em> form but without the leading
      * {@code -----BEGIN CERTIFICATE-----} and trailing {@code -----END CERTIFICATE-----}.
      * 
+     * 
      * @return The certificate or {@code null} if no certificate authority
      *         has been set.
      * @throws CertificateException if the value of <em>cert</em> property cannot be parsed into
-     *             an X.509 certificate.
+     *              an X.509 certificate.
      */
     @JsonIgnore
     public X509Certificate getTrustedCertificateAuthority() throws CertificateException {
 
-        final JsonObject trustedCa = getProperty(TenantConstants.FIELD_PAYLOAD_TRUSTED_CA);
-        if (trustedCa == null) {
+        final List<X509Certificate> certs = getTrustedCertificateAuthorities();
+        if (certs == null) {
             return null;
         } else {
-            final Object obj = trustedCa.getValue(TenantConstants.FIELD_PAYLOAD_CERT);
-            if (obj instanceof String) {
-                try {
-                    final byte[] derEncodedCert = Base64.getDecoder().decode(((String) obj));
-                    final CertificateFactory factory = CertificateFactory.getInstance("X.509");
-                    return (X509Certificate) factory.generateCertificate(new ByteArrayInputStream(derEncodedCert));
-                } catch (final IllegalArgumentException e) {
-                    // Base64 decoding failed
-                    throw new CertificateParsingException("cert property does not contain valid Base64 scheme", e);
-                }
-            } else {
-                return null;
-            }
+            return certs.get(0);
         }
     }
 
     /**
-     * Gets the trust anchor for this tenant.
+     * Gets the trust anchor for this tenant. This method assumes that this tenant is
+     * configured with a single trusted CA.
      * <p>
      * This method tries to create the trust anchor based on the information
-     * from the JSON object contained in the <em>trusted-ca</em> property.
+     * from the JSON array contained in the <em>trusted-ca-store</em> property. The JSON
+     * array should contain a single JSON object, which stores the config. info for this
+     * tenant's trusted CA.
      * <ol>
      * <li>If the object contains a <em>cert</em> property then its content is
      * expected to contain the Base64 encoded (binary) DER encoding of the
@@ -235,23 +238,12 @@ public final class TenantObject extends JsonBackedValueObject {
      * @return The trust anchor or {@code null} if no trusted certificate authority
      *         has been set.
      * @throws GeneralSecurityException if the value of the <em>trusted-ca</em> property
-     *             cannot be parsed into a trust anchor, e.g. because of unsupported
-     *             key type, malformed certificate or public key encoding etc. 
+     *              cannot be parsed into a trust anchor, e.g. because of unsupported
+     *              key type, malformed certificate or public key encoding etc.
      */
     @JsonIgnore
     public TrustAnchor getTrustAnchor() throws GeneralSecurityException {
-
-        if (trustAnchor != null) {
-            return trustAnchor;
-        } else {
-            final X509Certificate cert = getTrustedCertificateAuthority();
-            if (cert != null) {
-                trustAnchor = new TrustAnchor(cert, null);
-                return trustAnchor;
-            } else {
-                return getTrustAnchorForPublicKey(getProperty(TenantConstants.FIELD_PAYLOAD_TRUSTED_CA));
-            }
-        }
+        return (getTrustAnchors() == null ? null : getTrustAnchors().get(0));
     }
 
     @JsonIgnore
@@ -262,6 +254,7 @@ public final class TenantObject extends JsonBackedValueObject {
         } else {
             final String subjectDn = getProperty(keyProps, TenantConstants.FIELD_PAYLOAD_SUBJECT_DN);
             final String encodedKey = getProperty(keyProps, TenantConstants.FIELD_PAYLOAD_PUBLIC_KEY);
+
             if (subjectDn == null || encodedKey == null) {
                 return null;
             } else {
@@ -270,8 +263,7 @@ public final class TenantObject extends JsonBackedValueObject {
                     final X509EncodedKeySpec keySpec = new X509EncodedKeySpec(Base64.getDecoder().decode(encodedKey));
                     final KeyFactory factory = KeyFactory.getInstance(type);
                     final PublicKey publicKey = factory.generatePublic(keySpec);
-                    trustAnchor = new TrustAnchor(subjectDn, publicKey, null);
-                    return trustAnchor;
+                    return new TrustAnchor(subjectDn, publicKey, null);
                 } catch (final IllegalArgumentException e) {
                     // Base64 decoding failed
                     throw new InvalidKeySpecException("cannot decode Base64 encoded public key", e);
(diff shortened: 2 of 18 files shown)


sophokles73:
@Alfusainey thanks for the PR :+1: 

Can you please create a separate PR for adding the Base64 sanity checks to the *addTenant* operation? IMHO that is a valuable addition which also makes sense outside the scope of this PR.

Also, please rebase your PR to current master ...

Alfusainey:
@sophokles73 i just rebased with current master

> Can you please create a separate PR for adding the Base64 sanity checks to the addTenant operation? IMHO that is a valuable addition which also makes sense outside the scope of this PR.

sure 👍

sophokles73:
@Alfusainey what's the status on this one? are you going to continue work on this PR? If so, a first step would be to rebase, I guess ...

Alfusainey:
> @Alfusainey what's the status on this one? are you going to continue work on this PR? If so, a first step would be to rebase, I guess ...

sure, its in my todo list for this week (until Sunday)

Alfusainey:
@sophokles73 patch updated!

Alfusainey:
@sophokles73 just pushed an extra commit.. any feedback before the weekend starts would be awesome 👍

core/src/main/java/org/eclipse/hono/util/TenantConstants.java
```diff
@@ -60,9 +60,9 @@
      */
     public static final String FIELD_PAYLOAD_PUBLIC_KEY = "public-key";
     /**
-     * The name of the property that contains the trusted certificate authority configured for a tenant.
+     * The name of the property that contains the list of trusted certificate authorities configured for this tenant.
```
sophokles73: > ... for a tenant.

In the constants, there is no *this tenant*, is there?

service-base/src/main/java/org/eclipse/hono/service/tenant/CompleteBaseTenantService.java
```diff
@@ -251,21 +251,41 @@ private boolean isEncodedPublicKey(final Object encodedKey, final String algorit
...
+                return false;
+            } else {
+                return isValidTrustedCaSpec(trustConfigs);
+            }
+        }
+    }
+
+    private boolean isValidTrustedCaSpec(final JsonArray trustConfigs) {
+        final boolean containsInvalidTrustedCa = trustConfigs.stream()
+                .anyMatch(obj -> !JsonObject.class.isInstance(obj));
```
sophokles73: IMHO this would be easier to read (and faster to execute) if both conditions were *or*ed together ...

service-base/src/test/java/org/eclipse/hono/service/tenant/AbstractCompleteTenantServiceTest.java
```diff
@@ -133,15 +134,16 @@ public void testGetForCertificateAuthoritySucceeds(final TestContext ctx) {
...
 
         addTenant("tenant", tenant).map(ok -> {
             getCompleteTenantService().get(subjectDn, null, ctx.asyncAssertSuccess(s -> {
                 assertThat(s.getStatus(), is(HttpURLConnection.HTTP_OK));
                 final TenantObject obj = s.getPayload().mapTo(TenantObject.class);
                 assertThat(obj.getTenantId(), is("tenant"));
-                final JsonObject ca = obj.getProperty(TenantConstants.FIELD_PAYLOAD_TRUSTED_CA);
-                assertThat(ca, is(trustedCa));
+                final List<JsonObject> cas = obj.getTrustConfigurations();
+                assertThat(1, is(cas.size()));
```
sophokles73: `assertThat(cas.size(), is(1));`

site/content/api/Tenant-API.md
```diff
@@ -282,7 +282,7 @@ The table below provides an overview of the standard members defined for the JSO
...
 | *enabled*                | *no*      | *boolean*     | `true`       | If set to `false` the tenant is currently disabled. Protocol adapters MUST NOT allow devices of a disabled tenant to connect and MUST NOT accept data published by such devices. |
-| *trusted-ca*             | *no*      | *JSON object* | `-`          | The trusted certificate authority to use for validating certificates presented by devices of the tenant for authentication purposes. See [Trusted Certificate Authority Format]({{< relref "#trusted-ca-format" >}}) for a definition of the content model of the object. |
+| *trust-store*             | *no*      | *JSON array* | `-`          | The list of trusted certificate authorities to use for validating certificates presented by devices of the tenant for authentication purposes. The JSON array consist of JSON objects, each describing the configuration information of a trusted CA. See [Trusted Certificate Authority Format]({{< relref "#trusted-ca-format" >}}) for a definition of the content model of the object and an example tenant configured to use two trusted certification authorities. |
```
sophokles73: > ...for a definition of the content model of the object.

No need to mention the examples.

site/content/api/Tenant-API.md
```diff
@@ -362,6 +362,28 @@ The table below provides an overview of the members defined for the *trusted-ca*
...
+Below is an example for a payload of the response to a *get* request for tenant `TEST_TENANT` configured with two trusted certificate authorities.
+
+~~~json
+ {
+   "tenant-id" : "TEST_TENANT",
+   "enabled" : true,
+   "trust-store": [ {
+	    "subject-dn": "CN=ca,OU=Hono,O=Eclipse",
+	    "public-key": "NOTAPUBLICKEY",
+	    "not-before": "2015-01-01T00:00:00+0000",
```
sophokles73: neither *not-before* nor *not-after* are part of the content model, are they? And a trust store usually does not contain such properties as well.

core/src/main/java/org/eclipse/hono/util/TenantConstants.java
Alfusainey: i like this comment :-)

Alfusainey:
just pushed an extra commit

Alfusainey:
@sophokles73 if this looks good to you then i will squash the commits :-)

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -262,16 +259,17 @@ private TrustAnchor getTrustAnchorForPublicKey(final JsonObject keyProps) throws
         } else {
             final String subjectDn = getProperty(keyProps, TenantConstants.FIELD_PAYLOAD_SUBJECT_DN);
             final String encodedKey = getProperty(keyProps, TenantConstants.FIELD_PAYLOAD_PUBLIC_KEY);
+
             if (subjectDn == null || encodedKey == null) {
                 return null;
             } else {
+                final String algorithmName = Optional
+                        .ofNullable((String) getProperty(keyProps, TenantConstants.FIELD_ADAPTERS_TYPE)).orElse("RSA");
+                final X509EncodedKeySpec keySpec = new X509EncodedKeySpec(Base64.getDecoder().decode(encodedKey));
```
sophokles73: what's the purpose of pulling this out of the try block?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -510,4 +508,305 @@ public long getConnectionsLimit() {
                         (Number) TenantConstants.DEFAULT_MAX_CONNECTIONS).longValue())
                 .orElse(TenantConstants.DEFAULT_MAX_CONNECTIONS);
     }
+
+    /**
+     * Sets the configuration information for this tenant's configured trusted CA's. The RSA public key type is assumed
```
sophokles73: Please put additional sentences into a new paragraph, i.e.
```
Sets the configuration information for this tenant's configured trusted CAs.
<p>
The RSA public key type is assumed ...
```

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -510,4 +508,305 @@ public long getConnectionsLimit() {
...
+     * if the trusted CA JSON object does not contain a type property.
+     * 
+     * @param configurations A list of configuration properties, one set of properties for each configured trusted CA.
+     * @return This tenant for command chaining.
+     * 
+     * @throws IllegalArgumentException if the any of the trusted CA configuration is missing the required
```
sophokles73: several typos ...

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -168,15 +178,17 @@ public TenantObject setTrustAnchor(final X509Certificate certificate) {
...
-        } catch (CertificateEncodingException e) {
+            return addTrustedCaConfiguration(trustedCa);
+        } catch (final CertificateEncodingException e) {
             throw new IllegalArgumentException("cannot encode certificate");
         }
     }
 
     /**
      * Gets the trusted certificate authority configured for this tenant.
+     * This method assumes that the tenant is configured with a single Trusted CA.
```
sophokles73: actually, it only requires to have *at least one* CA configured

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -232,26 +238,17 @@ public X509Certificate getTrustedCertificateAuthority() throws CertificateExcept
...
+     * @deprecated Replaced by {@link #getTrustAnchors()}
      * @return The trust anchor or {@code null} if no trusted certificate authority
      *         has been set.
      * @throws GeneralSecurityException if the value of the <em>trusted-ca</em> property
-     *             cannot be parsed into a trust anchor, e.g. because of unsupported
-     *             key type, malformed certificate or public key encoding etc. 
+     *              cannot be parsed into a trust anchor, e.g. because of unsupported
+     *              key type, malformed certificate or public key encoding etc.
      */
+    @Deprecated(since="1.0", forRemoval=true)
```
sophokles73: I do not think that we should deprecate this method. Instead, we should simply replace it with *getTrustAnchors*.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -55,7 +59,14 @@
     @JsonIgnore
     private Map<String, JsonObject> adapterConfigurations;
     @JsonIgnore
-    private TrustAnchor trustAnchor;
+    private List<JsonObject> trustedCaList;
+    @JsonIgnore
+    private List<TrustAnchor> trustAnchors;
```
sophokles73: FMPOV this is the only collection we need. What are the others for?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -115,28 +126,25 @@ public TenantObject setEnabled(final boolean flag) {
...
-                    .map(dn -> new X500Principal(dn)).orElse(null);
-        }
+        final List<X500Principal> subjectdns = Optional.ofNullable(getTrustedCaSubjectNames()).orElse(null);
+        return subjectdns == null ? null : subjectdns.get(0);
     }
 
     /**
-     * Sets the trusted certificate authority to use for authenticating
-     * devices of this tenant.
+     * Sets the trusted anchor for this tenant's Trusted CA using the given public key and subject DN.
```
sophokles73: this method should probably be changed to support *adding* a trust anchor instead of setting one, i.e. *addTrustAnchor*

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -146,15 +154,17 @@ public TenantObject setTrustAnchor(final PublicKey publicKey, final X500Principa
...
+                .put(TenantConstants.FIELD_PAYLOAD_SUBJECT_DN, subjectDn.getName(X500Principal.RFC2253))
+                .put(TenantConstants.FIELD_PAYLOAD_PUBLIC_KEY, publicKey.getEncoded());
+
+        return addTrustedCaConfiguration(trustedCa);
     }
 
     /**
      * Sets the trusted certificate authority to use for authenticating
-     * devices of this tenant.
+     * devices of this tenant. The certificate must contain a {@code NonNull} and non-empty
```
sophokles73: this method should probably be changed to support *adding* a trust anchor instead of setting one, i.e. *addTrustAnchor*

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -185,39 +197,33 @@ public TenantObject setTrustAnchor(final X509Certificate certificate) {
...
      * 
+     * @deprecated Replaced by {@link #getTrustedCertificateAuthorities()}
+     * 
      * @return The certificate or {@code null} if no certificate authority
      *         has been set.
      * @throws CertificateException if the value of <em>cert</em> property cannot be parsed into
-     *             an X.509 certificate.
+     *              an X.509 certificate.
      */
+    @Deprecated(since="1.0", forRemoval=true)
```
sophokles73: I do not think that we should deprecate this method. It is only used internally for deserializing a trust anchor that has been specified in the form of an X.509 certificate. Thus, I guess we should make it private and change its signature to accept a JsonObject that contains the cert.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: if i should have replaced it in this PR, then i will have to change the *-adapters code base that uses it when authenticating client devices using x.509 certificates. i decided to do that in a separate PR (and to also add some integration tests for it). This is the reason why i deprecated it in this PR.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: a client might be interested to get all the trusted CAs in a tenant's trust-store. the `trustedCaList` is a cache for subsequent calls to `getTrustConfigurations()`

likewise, the `subjectNames` is a cached for subsequent calls to `getTrustedCaSubjectNames()`

i find them worthwhile to have. Don't you agree?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: > Thus, I guess we should make it private and change its signature to accept a JsonObject that contains the cert.

yes, such a method already exist. 

Like i said before, the reason why i deprecated it for now is so that I do not touch the *-adapters code base in this PR. i will remove it entirely in a follow up PR

sophokles73:
@Alfusainey any news on this?

Alfusainey:
> @Alfusainey any news on this?

sorry, i have been quite busy since the last update.. but i will take a look and incorporate your comments tomorrow (public holiday)

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: since this method will eventually be replaced with `getTrustedCertificateAuthorities()`, i will remove this sentence

Alfusainey:
@sophokles73 PR updated with b6b0a25

Alfusainey:
@sophokles73 @dejanb I just resolved the merge conflict caused by commit 014d7c7a8.

Alfusainey:
any news on this? i think the longer it takes to get feedback on this, the more merge conflicts i will have to resolve :-)

sophokles73:
@Alfusainey , sorry for the delay. I'll take a look today ...

core/src/main/java/org/eclipse/hono/util/TenantConstants.java
```diff
@@ -60,9 +60,9 @@
      */
     public static final String FIELD_PAYLOAD_PUBLIC_KEY = "public-key";
     /**
-     * The name of the property that contains the trusted certificate authority configured for a tenant.
+     * The name of the property that contains the list of trusted certificate authorities configured for a tenant.
      */
-    public static final String FIELD_PAYLOAD_TRUSTED_CA = "trusted-ca";
+    public static final String FIELD_PAYLOAD_TRUST_STORE = "trust-store";
```
sophokles73: I wonder if we can keep the original name but change its content model to either support a single JSON *object* or an *array* of objects. That way we could more easily keep backwards compatibility.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
sophokles73: FMPOV we should keep the use cases in mind that the `TenantObject` is relevant for:
1. create/update a tenant in the device registry and
2. verify a client certificate in a protocol adapter

So, for use case 1, the registry implementation needs be check if any of the subject DNs are already in use by another tenant. So, having a *getTrustedCaSubjectDns* method returning a list of Strings should do. However, this is done only once when the request to create/update is received, so I do not quite see why we should cache this list instead of just iterating over the *trusted-ca* property content and retrieve the subject DNs ...

For use case 2, the protocol adapter needs to
1. look up the tenant by subject DN, getting back the tenant data,
2. retrieve the list of TrustAnchors from the tenant data and
3. pass the list of trust anchors to the CertificateValidator which uses it to veryfy the client certificate.

Looks like for that we only need the *trustAnchors*, don't we?
Based on this, maybe you can re-assess all the methods and fields that you have added?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -114,68 +125,69 @@ public TenantObject setEnabled(final boolean flag) {
...
     public X500Principal getTrustedCaSubjectDn() {
-
-        final JsonObject trustedCa = getProperty(TenantConstants.FIELD_PAYLOAD_TRUSTED_CA);
-        if (trustedCa == null) {
-            return null;
-        } else {
-            return Optional.ofNullable((String) getProperty(trustedCa, TenantConstants.FIELD_PAYLOAD_SUBJECT_DN))
-                    .map(dn -> new X500Principal(dn)).orElse(null);
-        }
+        final List<X500Principal> subjectdns = Optional.ofNullable(getTrustedCaSubjectNames()).orElse(null);
```
sophokles73: IMHO we should remove this method altogether.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -114,68 +125,69 @@ public TenantObject setEnabled(final boolean flag) {
...
         Objects.requireNonNull(certificate);
 
         try {
             final JsonObject trustedCa = new JsonObject()
+                    .put(TenantConstants.FIELD_PAYLOAD_SUBJECT_DN, certificate.getSubjectX500Principal().getName())
                     .put(TenantConstants.FIELD_PAYLOAD_CERT, certificate.getEncoded());
-            return setProperty(TenantConstants.FIELD_PAYLOAD_TRUSTED_CA, trustedCa);
-        } catch (CertificateEncodingException e) {
+            return addTrustedCaConfiguration(trustedCa);
+        } catch (final CertificateEncodingException e) {
```
sophokles73: this method should probably just delegate to *addTrustAnchor(PublicKey, X500Principal)*, shouldn't it?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -231,26 +237,17 @@ public X509Certificate getTrustedCertificateAuthority() throws CertificateExcept
...
-        } else {
-            final X509Certificate cert = getTrustedCertificateAuthority();
-            if (cert != null) {
-                trustAnchor = new TrustAnchor(cert, null);
-                return trustAnchor;
-            } else {
-                return getTrustAnchorForPublicKey(getProperty(TenantConstants.FIELD_PAYLOAD_TRUSTED_CA));
-            }
-        }
+        return (getTrustAnchors() == null ? null : getTrustAnchors().get(0));
```
sophokles73: IMHO we should remove this method.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: thanks for this elaborate feedback 👍

> Based on this, maybe you can re-assess all the methods and fields that you have added?

Sure

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: i think if we should maintain the current contain model for backwards compatibility (as you mention previously), then we should keep this method too. Don't you think?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: as i mention in a previous comment, we can keep this for the current content model

core/src/main/java/org/eclipse/hono/util/TenantConstants.java
Alfusainey: @sophokles73 how about we then keep the currently available `setTrustAnchor`and `getTrustAnchor`methods? These calls can assume that the `trusted-ca` property contains JSON object.

sophokles73: My main concern is backwards compatibility, i.e. it would be nice, if existing database content would not need to be migrated, i.e. it should be possible to map a JSON structure containing a single trusted ca as a JSON object into the `TenantObject` and then use e.g. *getTrustAnchor` to retrieve the *list* of trust anchors (which in this case would contain a single element only).

core/src/main/java/org/eclipse/hono/util/TenantObject.java
sophokles73: this will probably be needed by the service implementation when adding/updating the trust anchor. However, the method should then return a `Set<X500Principal>`, don't you think?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
sophokles73: as I mentioned above, should probably return `List<TrustAnchor>`.

Alfusainey:
@sophokles73 I have not forgotten about this patch.. I will work on it this weekend

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: sure, but even if we were to return a list, it will not contain duplicate elements since the `trusted-ca` is internally implemented using a map with the `subject-dn` property serving as key.

will fix

sophokles73: @Alfusainey I have added a [comment on the original issue](https://github.com/eclipse/hono/issues/623#issuecomment-496389453) which might shine some light on this ...

This basically goes back to whether we need to record a life span for the public keys or not.

Alfusainey: yes, am following the discussions there. however, I do not see a problem if the life span of the trusted-ca(s) are not registered with the tenant (but I might be wrong). 

Since a protocol adapter will retrieve a list of trusted CAs configured for a tenant, the adapter can try to validate a client's certificate using the first public key, if that fails, it tries the second public key, if that fails, it tries the third keyand so on, until it can finally find one that can valid a client's certificate.

This is because if one of the trusted CAs expires in the meantime, the tenant would have already  been configured with another (new) trusted CA which will succeed in authenticating the device.

Maybe you want protocol adapters be able to determine (upfront) which trusted CA can authenticate a client?

Alfusainey:
@sophokles73 I update the PR

now a JSON payload with a trusted-ca that uses either the old or new content model can be mapped to a `TenantObject`

sophokles73:
@Alfusainey this PR has gotten way to big to follow/review properly.
I think we need to break it down into several steps. I propose the following:
1. create a PR for changing the Tenant API and `TenantObject` to support an array of trusted CAs instead of just a single one. Please allow the TenantObject to de-serialize a single trusted CA that is kept in JSON object (instead of an array).
1. extend the data model of the trusted CAs in the Tenant API with (optional) *not-before* and *not-after* properties. At the same time, remove support for storing a full X.509 certificate in the *cert* property.
1. create additional PRs for e.g. adapting the HTTP based Tenant management API.

Alfusainey:
> @Alfusainey this PR has gotten way to big to follow/review properly.
> I think we need to break it down into several steps. 
> I propose the following:
> 
> 1. create a PR for changing the Tenant API and `TenantObject` to support an array of trusted CAs instead of just a single one. Please allow the TenantObject to de-serialize a single trusted CA that is kept in JSON object (instead of an array).
> 2. extend the data model of the trusted CAs in the Tenant API with (optional) _not-before_ and _not-after_ properties. At the same time, remove support for storing a full X.509 certificate in the _cert_ property.
> 3. create additional PRs for e.g. adapting the HTTP based Tenant management API.

sure 👍

sophokles73:
@Alfusainey can you close this PR? I assume it will be replaced by several others, right?

PR #1107, opened 2019-03-21
[#930][#1103] Fixes regarding handling of "last-via" property.

### LINKED ISSUE #1103: Fix cache handling in RegistrationClientImpl (concerning multiple gateways)
Relates to #930, #1093:
In order to support the scenario of a device connecting via different gateways (not at the same time but alternating between gateways), the caching logic concerning `assertRegistration` requests has to be adapted.

Scenario:
- device `dev-A` connects via gateway `gw-1` => no cache hit, assertRegistration is invoked (last-via is updated), new cache entry with key `(assert, dev-A, gw-1)` is created
- device `dev-A` connects via gateway `gw-2` => no cache hit, assertRegistration is invoked (last-via is updated), new cache entry with key `(assert, dev-A, gw-2)` is created
- device `dev-A` again connects via gateway `gw-1` => **cache hit** => assertRegistration is NOT invoked and last-via is NOT updated - that is an error

Possible solution:
When writing the cache entry `(assert, dev-A, gw-2)` above, we have to delete all other cache entries concerning `assert` and `dev-A`.

EDIT: We shouldn't cache `RegistrationResult` objects representing devices with multiple `via` entries at all (see below).

calohmn:
This fixes #1103 and relates to #930:
The "last-via" property is now also written if the device connects directly (not through a gateway). 
It is now made sure that the 'assertRegistration' results are not cached for devices having multiple 'via' entries.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
@@ -237,15 +237,17 @@ public abstract class BaseRegistrationService<T> extends EventBusService<T> impl
         final Future<RegistrationResult> getResultTracker = Future.future();
         getDevice(tenantId, deviceId, getResultTracker.completer());
 
-        getResultTracker.map(result -> {
+        getResultTracker.compose(result -> {
             if (isDeviceEnabled(result)) {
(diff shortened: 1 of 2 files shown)


BobClaerhout:
From a functional perspective this looks good to me.

PR #1108, opened 2019-03-21
Draft: new Theme DocDook for the Documentation.

b-abel:
First steps to make the site run with the theme "DocDock".
Run `./build-site.sh` in hono/site/ to try out.

The new theme provides a search, which is really hard to integrate into the old one. 
The old theme has not seen any changes for two years.
A new look could be a good idea for the release of the upcoming version1.0.

**Draft: this is just a preview. A couple of adjustments need to be made before using this.**

I just want to get some feedback before investing more time in this.

**WDYT?**


b-abel:
This theme one proposal for #1109.

The are some visual glitches to fix and the footer of the menu needs to be reworked.

hono_site_theme_docdock01

jbtrystram:
The navigation box is awesome :love_letter:

ctron:
I absolutely like the new theme. I only would prefer the navigation sidebar on the left to be a little bit more bright.

One thing to note, and I didn't check for that, maybe you already did, we had some issues with the current theme and the web fonts. IIRC some custom font was being used, but not imported using CSS, so it only worked on Ubuntu :wink:

b-abel:
# How to try out

```
~/hono/site$ ./build-site.sh
~/hono/site$ hugo server -v
```

Then open the displayed link.

dejanb:
I also like this theme better. As Jens said, a bit brighter sidebar would make it perfect.

sophokles73:
at least on my machine, the rendering of pages is veeeeery slow. In the browser's status bar I can read that it waits for api.github.com to complete. Any idea what this is needed for?

Apart from that, I do not see any real difference between this one and the *Lean* #1111 theme. Can you point out some relevant diffferences?

b-abel:
> at least on my machine, the rendering of pages is veeeeery slow. In the browser's status bar I can read that it waits for api.github.com to complete. Any idea what this is needed for?

That URL is probably used to fetch the number of stars from Github. But should be async... I'm pretty sure that there are a couple of things that we could optimize.

b-abel:
> Apart from that, I do not see any real difference between this one and the _Lean_ #1111 theme. Can you point out some relevant diffferences?

I summed up some differences in the ticket: https://github.com/eclipse/hono/issues/1109#issuecomment-477564314

b-abel:
We just decided in the Community Call to close this PR in favor of #1111 and proceed with theme "Learn".

Issue #1109, opened 2019-03-22
New Theme for the Homepage


b-abel:
On the occasion of the release of version 1.0 the homepage could get some love. 
A new theme should provide a search feature and a fresh look and feel.

Related tasks are:
* a new landing page
* restructuring of the menu
* a FAQ (#1035)

b-abel:
The PR #1108 is one idea based on the theme [DocDock](https://themes.gohugo.io/docdock/). I am working on another one with the (quite similar) theme [Learn](https://github.com/matcornic/hugo-theme-learn/).

jbtrystram:
The main feature I miss on the website is an index when navigating the API resources :) Is that difficult to add ? It would be a nice addition !

sophokles73:
+1 for having a search field and a better navigation bar that supports > 2 levels ...

kaniyan:
Also 👍  for the search feature.

kaniyan:
I tried out `bsinno:PR/site_theme_learn` and `bsinno:PR/site_theme_docdock`. Both looks good. I like the theme `learn` more that the `docbook`. 

The icons in learn `looks` cool. In `docbook`, I found that the navigation from one page to another is not smooth as `learn`.
Example:  It takes some seconds for correct rendering of code highlights. Check the below 2 images. 

![image](https://user-images.githubusercontent.com/18659982/55091825-c9684c80-50b1-11e9-91ba-48980b5ed9c3.png)


![image](https://user-images.githubusercontent.com/18659982/55091965-046a8000-50b2-11e9-8385-1cdf3fe78025.png)

b-abel:
## Differences between themes *DocDock* and *Learn*

DocDock is a fork of Learn. And a couple of additions in DocDock went back into Learn. So the differences are in detail (apart from features that we don't use, e.g. you can create revealJS presentations from DocDock). 

DocDock provides "**excerpts**", that is the generated overview page for each content folder. The current Hono homepage uses this. But until now nobody insisted on keeping that and I would prefer to have an `_index.md` in each content folder with an explicit descriptive text.
DocDock provides support for a custom footer, in Learn we need to add that by ourselves.

**Visually**, there are details in the styling of the sidebar and the fonts. 

**Technically**: Learn provides a simple mechanism to customize colors. 

**Common**: Things that are IMO easy to adjust on both themes are colors and icons.

**Community**: Learn seems to be more actively developed.

sophokles73:
[...]

PR #1111, opened 2019-03-26
[#1109] new Theme "Learn"

b-abel:
Draft of the site with theme 'learn' (for #1109). There are a couple of open points.
Try out with:
```
~/hono/site$ mvn clean install
~/hono/site$ hugo server -v
```

The theme [DocDock](https://themes.gohugo.io/docdock/) in #1108 is a fork of this theme [Learn](https://github.com/matcornic/hugo-theme-learn/). Therefore probability is higher that "learn" passes the license checks. Since we don't need the additional features which DocDock provides, I would prefer "Learn". 

Draft: this is just a preview. A couple of adjustments need to be made before using this.
I just want to get some feedback before investing more time in this.

WDYT?

hono_site_theme_learn03


b-abel:
## Open points 

* the icons for the guides are only to show what is possible. I think we could restructure the menu (IMHO too many at top level) and then later on think about having an icon per top level topic.
* The legal stuff at the bottom of the menu bar should go IMHO into an top level topic "Legal" (like it is on the Eclipse Homepage)
* Each "Chapter" (e.g. "Concepts", "User Guide", "Admin Guide", ...) needs a descriptive sentence. This is instead of the "excerpts" that the current theme shows. I added dummy examples for "User Guide" and "Admin Guide" (image below).

hono_site_theme_learn02

sophokles73:
I like the navigation bar at the left, in particular the fact, that it supports expanding sublevels and they remain opened for navigation. The search bar is also very nice. Regarding the colors, we definitely need to work on that ;-)
Does this theme support a common footer for the content pages? I have to admit that I do not like putting all the stuff from the original theme's footer into the navigation bar.

sophokles73:
IMHO we should create a landing page that contains just the index page and a high level (horizontal?) navigation bar with entries for the content that is not specific to a particular version of Hono, e.g. *Getting Started*, *Download*, *Community* , *License*. This landing page could then be designed independently from the chosen theme and strictly focus on *visual appeal* (maybe we can get some help from a web designer from our marketing here).
We should then use the new theme to create the documentation for the specific Hono version(s) we support and add a link them to the landing page's menu bar (e.g. a dropdown list for *Documentation*).
WDYT?

b-abel:
> Does this theme support a common footer for the content pages? I have to admit that I do not like putting all the stuff from the original theme's footer into the navigation bar.

No. Learn does not support this. We could of course add it ourselves.
The stuff can clearly not stay in the navigation bar. Does everything of this have to be displayed on every page? I would like to have the 2 Eclipse logos only on the landing page. And we could put the links in a section "Legal".

sophokles73:
> I would like to have the 2 Eclipse logos only on the landing page. And we could put the links in a section "Legal".

FMPOV there will be no section *Legal* in the navigation bar because we only use this theme for the documentation. We could, however, have a *Legal* drop down on the landing page's menu bar ...

I am not sure, if we need to have these links on *every* page or if it is sufficient to have them on the landing page.

b-abel:
I changed the background of the sidebar to light gray: 

**WDYT?**

hono_site_theme_learn05

ctron:
I would definitely prefer the light gray. Maybe lightening up the darker areas a bit as well … but in any case, I this would be my preferred theme.

b-abel:
> I would definitely prefer the light gray. Maybe lightening up the darker areas a bit as well … but in any case, I this would be my preferred theme.

Good news for you: yesterday in the community call we decided to go with this theme :smiley:

b-abel:
The PR is now ready to be reviewed. 

## Main changes

* The content files are only slightly adapted for the new theme, to leverage some of it's features and to make it look better. 
* I added a custom (color) "theme variant" with a light gray sidebar and orange elements. 
* The theme does not contain partials for the "Note" and "Warning" boxes, so I wrote my own.
* Since the theme does not provide one, I added a simple one for the copyright stuff. I would like to move the content somewhere else soon and remove the footer then.

## Open Questions

* Keep the Icons for the sections? If yes, keep this icons? As discussed elsewhere some of the sections will be removed from the documentation anyways and show up only on the landing page (with a probably very different appearance).
* How to organize multiple versions of the documentation (options below)? 

## Versioning of documentation

I see 2 options with theme learn:

1. use the internationalization feature, e.g. languages "v1" and "v2", and have a dropdown menu to select the version like in https://learn.netlify.com/en/cont/i18n/ In contrast to the example there, the content can be organized in separate folders per language (version), as documented here: https://gohugo.io/content-management/multilingual#translation-by-content-directory (Successfully tested with theme learn and versions instead of real languages)
2. don't use any feature of Hugo and the theme, but configure different directories in the web server and link to them from the landing page. This would probably the preferred option, if multiple versions of Hono are maintained in different git branches.


Having versions simply in sub-folders of "content" is not a great option, because "learn" always expands the active section including all subsections, so if a version is open, all pages of that version are expanded in the menu bar.

sophokles73:
@b-abel Thanks a lot for all the work :+1: 
I think that this can indeed be the basis for refactoring the Hono web site. Regarding the approach for keeping multiple versions of the documentation, here's what I think.
(Mis-)Using the internationalization feature of the theme doesn't feel right to me. We might achieve our (short term) goal of hosting multiple versions this way, but what if we then want to start providing multiple language versions of the documentation, or if the theme's maintainers cease support for multiple languages as nobody really needs it, or if the way that internationalization works in the theme is changed and no longer uses separate folders for the different languages? I therefore think that we would be better off, if we used separate, independent folders for each set of documentation for a particular version and have a landing page in front of it, where we add links to the supported versions manually. This would probably mean that we run a separate `hugo server` for each folder containing the documentation of a particular version, but since we probably only support two or three versions at a time, this shouldn't be much of an issue. With this approach we could also easily switch the theme with a new version without needing to adapt the old versions.

b-abel:
@sophokles73 I agree. Keeping the the documentation for different versions separate and manually link to them is preferable. 
Probably it will be best to create branches for new releases (starting with 1.0) to be able to add fixes to multiple versions. That way we can take the documentation of each version from it's branch and have no need to manage multiple versions of the documentation in the `master` branch.

b-abel:
@sophokles73  How do you want to proceed? Merge this PR and we'll do the other planned changes in new PRs?

b-abel:
I pushed a big update today. Main changes:

* The module `site` is now split into `homepage` and `documentation` which are separate hugo projects with different themes
* The part in `documentation` is intended to be versioned with the Hono releases, so that a separate documentation is online for each supported version (starting with the 1.0 release)
* Prepared the Jenkins pipelines for versioned documentation (should work with versioning by git tags as well as branches)

The idea is to bring this online on the day of the 1.0 release.

**Please take a look at the preview of the site at http://hono-website.s3-website.eu-central-1.amazonaws.com/hono/**

**Remarks**

* Please note the menu "Documentation", this is where you choose the version. "Stable" is the last stable release (selected in the release pipeline). The idea is to guide users to this version (all other versions contain a warning banner) and to have it in the URL, so that one can link to individual pages in the current stable version. "Development" is build nightly from master branch, as it is the case today.
* The section "Who uses Eclipse Hono?" can be easily disabled with a config option. I would enable it once we clarified which logos can be added and how we may present them.  
* The links between documents are broken in a couple of places (will fix them soon).
* The "Eclipse Incubation" logo can -- as far as I know -- completely be removed with the release of Hono 1.0
* We should probably decide about the usage of the Hono Twitter account before the 1.0 release

sophokles73:
@b-abel this looks great :+1: 

I like the idea of having the *stable* and *latest* documentation links in the drop down.

> The "Eclipse Incubation" logo can -- as far as I know -- completely be removed with the release of Hono 1.0

Yes, we can get rid of that once we had our graduation review (which will be done with the 1.0 release review).

> We should probably decide about the usage of the Hono Twitter account before the 1.0 release

Let's talk about that in our next community call.

One thing about the landing page. Is it possible to get rid of the black frame around the picture and the white space right next to the picture? It would also be nice to scale up the picture so that it becomes wider and takes up more space. On a large screen, the landing page otherwise looks a little *empty* ...

mahbe:
@b-abel the picture on the start page (or more precise the fonts in the picture) look a bit strange in Firefox on Windows. Is it intended that the font is not a sans serif font?

b-abel:
I pushed fixes/changes for the feedback I already got.
 
> @b-abel the picture on the start page (or more precise the fonts in the picture) looks a bit strange in Firefox on Windows. Is it intended that the font is not a sans serif font?

Should be fixed now.

mahbe:
Looks great now 👍

Alfusainey:
@b-abel the new site looks really nice 👍 

I just have a few comments:

* what do you think about reducing the amount of empty space to `left` and `right` of the site's body? maybe having a left (vertical) navigation menu and shifting all the rest to the right of it?

* `landing page`: how about keeping the same (white) background color on all the sections of the page? or maybe reducing the thickness of the orange background?

* `documentation`: I was even thinking the documentation for specific versions will only discuss/mention things related to a specific version of Hono. However, I see that a lot of the docs are repeated across all the versions or maybe this is not ready yet?. 

    If not, then how about splitting the documentation into a base doc (does not change frequently, not version dependent) from a version dependent doc (e.g API changes etc).. I don't know if this makes any sense?

sophokles73:
@b-abel this looks really great now :+1: 

Using latest Chrome on Linux, the picture on the landing page is not horizontally centered. Is this intended? IMHO it should be centered ...

With the space available in the footer, I would like to also include the iot.eclipse.org logo there in order *pledge our allegiance* ;-) WDYT?

In contrast to @Alfusainey I do like the strong contrast between the sections of the landing page created by the orange background very much :-)

b-abel:
@Alfusainey Thank you very much for your feedback!

>     * what do you think about reducing the amount of empty space to `left` and `right` of the site's body? maybe having a left (vertical) navigation menu and shifting all the rest to the right of it?

This depends on the aspect ratio of your screen. This is, like the menu part of the Hugo theme, which is a lot effort to change (and then adjust to our needs). To change this, we would basically throw away the layout of the Hugo theme that we selected and build our own. That is an effort I would not take unless there are very good reasons to do so. Especially because I know that creating a coherent design with good usability, accessibility and moving around the pitfalls of browser incompatibilities requires a lot more front-end developer know-how than I have.

>     * `landing page`: how about keeping the same (white) background color on all the sections of the page? or maybe reducing the thickness of the orange background?

I understand your point. I see the strong contrasts and the bold orange parts as a integral part of the "corporate design" of Hono up to now. So, I think you need to collect a lot of votes to change that ;)

b-abel:
**Versioning of the Documentation** 

What I tried to achieve here was to align the versioning very tightly with the versioning of the code to maximize consistency while minimize manual effort. I expect that -- starting from the release of Hono 1.0 -- we will create a Git branch for every release. The Jenkins pipeline that builds the web site (running every night) iterates over a list of all supported versions of Hono and for each version checks out the branch of that version (the implementation in this PR works with Git tags as well) and builds the content of the directory `site/documentation/` (and additionally the ones for `stable` and `development`). The list of all supported versions is maintained in file on the master branch (`site/homepage/versions_supported.csv`), as well as the current stable version. Both files are updated in the release pipeline. The whole process does not require manual tasks (besides selecting 2 check-boxes when performing a release and removing old versions from the list, once they are no longer supported). 

>     * `documentation`: I was even thinking the documentation for specific versions will only discuss/mention things related to a specific version of Hono. However, I see that a lot of the docs are repeated across all the versions or maybe this is not ready yet?.
>       If not, then how about splitting the documentation into a base doc (does not change frequently, not version dependent) from a version dependent doc (e.g API changes etc).. I don't know if this makes any sense?

Separating content which is likely to change from parts that (hopefully) would stay the same, would require extra effort on our side to describe the differences and for the reader to put those information parts together. To look up difference between versions I would consult the release notes.

kaniyan:
@b-abel The site looks nice  :+1:. Good work.

I have attached below screen shots of different pages from the website. What I noticed is that the font and the size of the texts are different. FMPOV it would be nicer to have an uniform font and size. 

##### Documentation
image

##### Home page
image

##### Getting started
image

b-abel:
> What I noticed is that the font and the size of the texts are different. 

The complete layout between the homepage part and the documentation part is different, which includes the fonts, sizing and nearly every other aspect of the visual appearance.  I see them as two separate sites (although they are available under the same domain and path prefix). That is why we decided to make them technically independent, use different Hugo themes and even open the documentation in a new tab.
Trying to change the fonts and font sizes is FMPOV a very fundamental change for the layout, having impact on spacing, readability and the overall visual appearance and the consequences may be not obvious (for future pages or menu entries, other browsers or screen sizes). Technically it is not about changing two properties in the CSS, but instead the CSS files contain thousands of lines. So, I would rather not fiddle with that.


> ##### Home page
> 
> image

Regarding the difference on the landing page and other pages on the homepage: the difference is intended. I made the punchline and the "greeting sentence" bigger. I see it not as a normal heading and text. But I can change that if this irritates some of you.

mbaeuerle:
Some more input from me:
- The different fonts on website and documentation were already mentioned. I can see that changing the fonts would change the whole theme, but still it's quite a big difference (documentation having much less font weight, font sizes are different)
- The summary below _Connect. Command. Control._ on the home page should imo be centered, like the rest
- The six key features on the home page are looking quite good, the problem I had with them is this "zoom" animation, which suggests, that the items are clickable. I would prefer a static image here.
- The getting started guide imo needs a bit more of padding inbetween so it doesn't overwhelm the user
- Also in the getting started and maybe also in other places: When I open the guide, the first thing that pops right into my eyes is this red warning, which is a little off-putting. Maybe the alert color can be changed to yellow, when it's not a don't-do-this-ever alert
- The Hono logo in the documentation top left corner is quite big which leads to scroll bars in the sidebar for almost every open navigation topic (using a 1080p display), maybe something like that would fit better: https://learn.netlify.com/en/cont/markdown/
- In info boxes sometimes a heading is missing, so the text looks a bit off, also it needs some padding to the bottom:
2019-06-12_10h35_17
Maybe we can get some inspiration from the box of the [theme documentation](https://learn.netlify.com/en/cont/markdown/) 
2019-06-12_10h40_38

PR #1117, opened 2019-03-27
Added docs and scripts to deploy Hono on Azure Kubernetes Services

spoplavskiy:
Added "DeployToAzure.md" file with tutorial about Azure deployment of Eclipse Hono
Added yaml files of external services, that will be exposed to public
Added "azure_deploy.sh" and "azure_undeploy.sh" scripts to automate azure deployment or deleting. Those scripts using original olm_deploy.sh, olm_undeploy.sh, kubernetes_deploy.sh, kubernetes_undeploy.sh, and additional commands to add external endpoints.
Added "push_hono_images_private.sh". It based on original "push_hono_images.sh” but require one more parameter with private container registry address. This can be used to push containers in private repo.

Signed-off-by: Sergii Poplavskyi <spopla@microsoft.com>


docs/DeployOnAzure.md
```diff
@@ -0,0 +1,295 @@
+# Eclipse Hono deployment on Microsoft Azure
```
sophokles73: we maintain the deployment guides in the `site/content/deployment/` folder. Can you please move this page there? I will then add a corresponding header.

deploy/src/main/deploy/kubernetes/azure_deploy.sh
```diff
@@ -0,0 +1,19 @@
...
+echo DEPLOYING ECLIPSE HONO TO AZURE AKS
+set +e
+source $SCRIPTPATH/olm_deploy.sh
+source $SCRIPTPATH/kubernetes_deploy.sh
+set -e
+
+echo CREATE AZURE EXTERNAL SERVICES
+kubectl delete service hono-dispatch-router-ext -n hono || true
+
+kubectl create -f $RESOURCES/azure --namespace $NS
```
sophokles73: my understanding is that you overwrite the service definitions in order to use the LoadBalancer type instead of defining a NodePort, right?

docs/DeployOnAzure.md
spoplavskiy: Hi @sophokles73 , Thank you for your comment. Added commit with those changes: https://github.com/eclipse/hono/pull/1117/commits/f80793136cf65c1f286a54d53a7163f9d66c5dea

deploy/src/main/deploy/kubernetes/azure_deploy.sh
spoplavskiy: Yes. By preparing tutorial and scripts I tried to avoid any changes in original files. That is why I have this "workaround".

sophokles73: I see. We have deprecated the script based installation to kubernetes some weeks ago and want to focus on Helm based installation in the future. A next step could therefore be, to incorporate Azure deployment into the existing Helm templates by means of another deployment target variant like we already do for distinguishing between *plain* kubernetes and Openshift. But I think we can do that in a separate PR (if you are interested in doing it).

spoplavskiy: This is something I'm focused on right now.

kartben:
Could someone amend the PR with a "chmod +x" on the push_hono_images_private.sh file, which is not executable at the moment?

dejanb:
@kartben done

PR #1122, opened 2019-03-28
Add Kerlink Provider tests

mbaeuerle:
For the tests to work the `wiremock` dependency is needed.
I also had to change the visibility of two fields, so they are accessible by tests.


sophokles73:
@mbaeuerle my understanding is that wiremock is used to create mocks implementing the LoRa provider HTTP API so that we can run tests for sending commands to devices, right?

mbaeuerle:
@sophokles73 yes that is correct.

sophokles73:
Created CQ for wiremock http://dev.eclipse.org/ipzilla/show_bug.cgi?id=19445

PR #1124, opened 2019-03-28
Fix IT test for disabled amqp adapter.

kaniyan:


tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpConnectionIT.java
```diff
@@ -153,8 +153,8 @@ public void testConnectFailsForDisabledAdapter(final TestContext ctx) {
         // GIVEN a tenant for which the AMQP adapter is disabled
         final TenantObject tenant = TenantObject.from(tenantId, true);
         final JsonObject adapterConfiguration = new JsonObject()
-                .put(TenantConstants.FIELD_ADAPTERS_TYPE, Constants.PROTOCOL_ADAPTER_TYPE_HTTP)
-                .put(TenantConstants.FIELD_ENABLED, Boolean.TRUE);
+                .put(TenantConstants.FIELD_ADAPTERS_TYPE, Constants.PROTOCOL_ADAPTER_TYPE_AMQP)
+                .put(TenantConstants.FIELD_ENABLED, Boolean.FALSE);
```
sophokles73: the existing configuration does the trick already, i.e. it enables the HTTP adapter only for the tenant which results in the AMQP adapter being disabled. The configuration you propose, on the other hand, actually results in a tenant for which no protocol adapter at all is enabled.

kaniyan: It doesn't fail. IMHO it makes the code easy to understand when we directly say that the AMQP adapter is not enabled by setting it to false. FMPOV, this test case is to check what happens when the amqp adapter is disabled, so I think its not relevant if other adapters are enabled or not.  We also use this approach in Mqtt adapter IT too.

sophokles73: ok, in this case I would propose to *add* the two lines for explicitly disabling the AMQP adapter instead of replacing the configuration for the HTTP adapter ...

kaniyan: I have pushed the changes.

PR #1125, opened 2019-03-28
Check if amqp adapter is enabled or not only after successful authentication.

kaniyan:


sophokles73:
what's the purpose of this?

kaniyan:
Unlike other adapters, in the Amqp Adapter it is being checked first that if it is enabled or not followed by the authentication of the device. To check if the Amqp Adapter is enabled or not, `TenantObject` is retrieved and the check is done. FMPOV this need not be done unless the credentials are already verified. 

In the next PR for connection limits, I like to check if the adapter is enabled or not, in parallel with connection limits using `CompositeFuture`. I think it makes sense to do these checks after the credentials are verified.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorFactory.java
```diff
@@ -302,14 +302,15 @@ private void verifyExternal(final Handler<AsyncResult<DeviceUser>> completer) {
                             getCertificateAuthProvider().authenticate(credentials, currentSpan.context(), user);
                             return user;
                         })
+                        .compose(user -> checkTenantIsEnabled(tenantTracker.result())
```
sophokles73: why is this check only done for SASL EXTERNAL?

kaniyan: For SASL PLAIN, this check is also being done and it is part of the method `getTenantObject(credentials.getTenantId())`. 

For SASL EXTERNAL, the tenant object is first retrieved to get the `TrustAnchor` and thats why the check `checkTenantIsEnabled(tenantTracker.result()` is not part of the `getTenantObject(...)`.

sophokles73: ahhh, now I see. This is very confusing FMPOV. Can you please align the behavior of the two methods so that they follow the same pattern?

kaniyan: Done

PR #1126, opened 2019-03-29
Add LoRaWAN adapter build profile

BobClaerhout:


adapters/lora-vertx/pom.xml
```diff
@@ -48,4 +48,64 @@
             <version>${jackson.version}</version>
         </dependency>
     </dependencies>
+
```
sophokles73: You also need to add the two following plugins to the POM's build:

```
  <build>
    <plugins>
      <plugin>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-maven-plugin</artifactId>
      </plugin>
      <plugin>
        <artifactId>maven-dependency-plugin</artifactId>
      </plugin>
    </plugins>
  </build>
```
otherwise, the Spring Boot jar will not be created and the docker build fails ...

BobClaerhout: I've added this as well

PR #1129, opened 2019-04-02
[#927] Add an IT test for connection limit

kaniyan:
When Prometheus backend is not reachable, check that the connections to Mqtt and Amqp adapters are accepted irrespective of the limits configured.


sophokles73:
IMHO this behavior can be verified by a unit test on `PrometheusBasedResourceLimitChecks` in a more concise way instead of using the *connectToAdapter* method to simulate two distinct connections to the adapter (for which it hasn't been designed).

PR #1130, opened 2019-04-02
Device Registry HTTP API

ctron:
Documented:
 * [X] Registration API
 * [x] Credentials API
 * [x] Tenant API
 * [x] Authentication

This PR contains an OpenAPI 3 model, which can be rendered and tested using the following command:

~~~
# cd hono/site/static/api
docker run -ti --rm -p 8080:8080 -e SWAGGER_JSON=/foo/device-registry.yaml  -v "$(pwd):/foo:z" swaggerapi/swagger-ui
~~~

Or directly from GitHub (without the option to make local changes):

~~~
docker run -ti --rm -p 8080:8080 -e URL=https://raw.githubusercontent.com/ctron/hono/feature/devreg_api_1/site/static/api/device-registry.yaml swaggerapi/swagger-ui
~~~

You can also directly use the hosted version here:

* Current API – https://current-api-hono-device-registry-api.wonderful.iot-playground.org
* Revised API – https://revised-api-hono-device-registry-api.wonderful.iot-playground.org

You can actually use the "Try it out" feature, as currently the API describes the "as is" situation. However I would like to change this for the final version. You can already see in the definition why, but let's wait with the discussion until all the current APIs are mapped.

### DIFF
--- a/site/static/api/device-registry-v1.yaml
+++ b/site/static/api/device-registry-v1.yaml
@@ -1,0 +1,745 @@
+openapi: 3.0.1
+info:
+   title: Eclipse Hono™ Device Registry API
+   description: |
+      This API defines how to manage *Tenants*, *Devices*, and *Credentials*.
+      It acts as a common basis which all Hono device registries should
+      implement.
+
+      ## Required APIs
+
+      All operations, except the `tenants` resource are required. The tenant
+      management might be outside of the scope of the device registry and
+      managed by a higher level system. In this case all calls should simply
+      return `404`. However, if the `tenants` resource is implemented, then all
+      operations of it must be implemented.
+
+      ## Security
+
+      This specification explicitly leaves out the part of authenticating and
+      authorizing users with the device registry. It is assumed that some form
+      of token exchange between the user agent and the backend service will
+      take place. Like for example HTTP basic authentication, or a bearer token.
+
+      ## Code generation
+
+      This model is not optimized for generating code from it. Code generators
+      try to understand the model and then translate this into the require
+      programming language. Even if the there would be no bugs in the code
+      generators, that process would be already only be an approximation. So
+      this model does focus in the description of the API, and doesn't tweak
+      the specification in a way to please code generators.
+
+   contact:
+      name: Contact details
+      url: https://www.eclipse.org/hono/community/get-in-touch/
+   license:
+      name: EPL-2.0
+      url: https://www.eclipse.org/legal/epl-2.0/
+   version: 1.0.0
+
+externalDocs:
+   description: Eclipse Hono™ web page
+   url: https://eclipse.org/hono
+
+tags:
+   - name: tenants
+     description: Tenant Management (optional)
+     externalDocs:
+        description: Hono Multi-Tenancy
+        url: https://www.eclipse.org/hono/concepts/tenancy/
+   - name: devices
+     description: Device registration
+     externalDocs:
+        description: Hono device identity
+        url: https://www.eclipse.org/hono/concepts/device-identity/
+   - name: credentials
+     description: Device credentials
+     externalDocs:
+        description: Hono device identity
+        url: https://www.eclipse.org/hono/concepts/device-identity/
+
+servers:
+   - url: '{corsProxy}{server}/v1'
+     variables:
+        server:
+           default: http://hono.eclipse.org:28080
+        corsProxy:
+           default: https://cors-anywhere.herokuapp.com/
+
+security:
+   - BearerAuth: []
+   - BasicAuth: []
+
+paths:
+
+# Tenant API
+
+   /tenants:
+
+      post:
+         tags:
+            - tenants
+         summary: Create new tenant with auto-generated ID
+         operationId: createTenant
+         requestBody:
+            description: New tenant information
+            content:
+               application/json:
+                  schema:
+                     $ref: '#/components/schemas/Tenant'
+            required: true
+         responses:
+            201:
+               $ref: '#/components/responses/Created'
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
+
+   /tenants/{tenantId}:
+
+      parameters:
+         - $ref: '#/components/parameters/tenantId'
+
+      post:
+         tags:
+            - tenants
+         summary: Create new tenant
+         operationId: createTenantWithId
+         requestBody:
+            description: New tenant registration
+            content:
+               application/json:
+                  schema:
+                     $ref: '#/components/schemas/Tenant'
+            required: true
+         responses:
+            201:
+               $ref: '#/components/responses/Created'
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
+            409:
+               $ref: '#/components/responses/AlreadyExists'
+
+      get:
+         tags:
+            - tenants
+         summary: Get tenant information
+         operationId: getTenant
+         responses:
+            200:
+               description: operation successful
+               content:
+                  application/json:
+                     schema:
+                        $ref: '#/components/schemas/Tenant'
+               headers:
+                  ETag:
+                     description: Version of the resource
+                     schema:
+                        type: string
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            404:
+               $ref: '#/components/responses/NotFound'
+
+      put:
+         tags:
+            - tenants
+         summary: Update tenant information
+         operationId: updateTenant
+         parameters:
+            - $ref: '#/components/parameters/resourceVersion'
+         requestBody:
+            description: Tenant information
+            content:
+               application/json:
+                  schema:
+                     $ref: '#/components/schemas/Tenant'
+            required: true
+         responses:
+            204:
+               $ref: '#/components/responses/Updated'
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
+            404:
+               $ref: '#/components/responses/NotFound'
+            412:
+               $ref: '#/components/responses/ResourceVersionMismatch'
+
+      delete:
+         tags:
+            - tenants
+         summary: Delete tenant
+         operationId: deleteTenant
+         parameters:
+            - $ref: '#/components/parameters/resourceVersion'
+         responses:
+            204:
+               description: object deleted
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
+            404:
+               $ref: '#/components/responses/NotFound'
+            412:
+               $ref: '#/components/responses/ResourceVersionMismatch'
+
+# Device API
+
+   /devices/{tenantId}:
+
+      parameters:
+         - $ref: '#/components/parameters/tenantId'
+
+      post:
+         tags:
+            - devices
+         summary: Create new device registration with auto-generated ID
+         operationId: createDeviceRegistration
+         requestBody:
+            description: New device
+            content:
+               application/json:
+                  schema:
+                     $ref: '#/components/schemas/Device'
+            required: true
+         responses:
+            201:
+               $ref: '#/components/responses/Created'
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
+
+   /devices/{tenantId}/{deviceId}:
+
+      parameters:
+         - $ref: '#/components/parameters/tenantId'
+         - $ref: '#/components/parameters/deviceId'
+
+      post:
+         tags:
+            - devices
+         summary: Create new device registration
+         operationId: createDeviceRegistrationWithId
+         requestBody:
+            description: New device
+            content:
+               application/json:
+                  schema:
+                     $ref: '#/components/schemas/Device'
+            required: true
+         responses:
+            201:
+               $ref: '#/components/responses/Created'
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
+            409:
+               $ref: '#/components/responses/AlreadyExists'
+
+      get:
+         tags:
+            - devices
+         summary: Get device registration information
+         operationId: getRegistration
+         responses:
+            200:
+               description: operation successful
+               content:
+                  application/json:
+                     schema:
+                        $ref: '#/components/schemas/Device'
+               headers:
+                  ETag:
+                     description: Version of the resource
+                     schema:
+                        type: string
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            404:
+               $ref: '#/components/responses/NotFound'
+
+      put:
+         tags:
+            - devices
+         summary: Update existing device registration
+         operationId: updateRegistration
+         parameters:
+            - $ref: '#/components/parameters/resourceVersion'
+         requestBody:
+            description: Updated device registration
+            content:
+               application/json:
+                  schema:
+                     $ref: '#/components/schemas/Device'
+            required: true
+         responses:
+            204:
+               $ref: '#/components/responses/Updated'
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
+            404:
+               $ref: '#/components/responses/NotFound'
+            412:
+               $ref: '#/components/responses/ResourceVersionMismatch'
+
+      delete:
+         tags:
+            - devices
+         summary: Delete device registration
+         operationId: deleteRegistration
+         parameters:
+            - $ref: '#/components/parameters/resourceVersion'
+         responses:
+            204:
+               $ref: '#/components/responses/Delete'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
+            404:
+               $ref: '#/components/responses/NotFound'
+            412:
+               $ref: '#/components/responses/ResourceVersionMismatch'
+
+   /credentials/{tenantId}/{deviceId}:
+
+      parameters:
+         - $ref: '#/components/parameters/tenantId'
+         - $ref: '#/components/parameters/deviceId'
+
+      get:
+         tags:
+            - credentials
+         summary: Get credentials set of a device.
+         description: |
+            Get the credentials set of a device. As long as the device is
+            registered and the user has read access to it, this call should
+            never return "not found".
+         operationId: getAllCredentials
+         responses:
+            200:
+               description: Operation successful
+               content:
+                  application/json:
+                     schema:
+                        $ref: '#/components/schemas/CredentialsSet'
+               headers:
+                  ETag:
+                     description: Version of the resource
+                     schema:
+                        type: string
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            404:
+               $ref: '#/components/responses/NotFound'
+
+      put:
+         tags:
+            - credentials
+         summary: Update credentials set for registered device
+         operationId: setAllCredentials
+         parameters:
+            - $ref: '#/components/parameters/resourceVersion'
+         requestBody:
+            content:
+               application/json:
+                  schema:
+                     $ref: '#/components/schemas/CredentialsSet'
+            required: true
+         responses:
+            204:
+               $ref: '#/components/responses/Updated'
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
+            404:
+               $ref: '#/components/responses/NotFound'
+            412:
+               $ref: '#/components/responses/ResourceVersionMismatch'
+
+      delete:
+         tags:
+            - credentials
+         summary: Delete all credentials of a device
+         description: |
+             Deletes all credentials from a device registration. The device is
+             still registered, and a following call to *get* credentials must
+             not return `404`, as the device is still registered. Instead an
+             empty result must be returned.
+         operationId: deleteAllCredentials
+         parameters:
+            - $ref: '#/components/parameters/resourceVersion'
+         responses:
+            204:
+               $ref: '#/components/responses/Delete'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
+            404:
+               $ref: '#/components/responses/NotFound'
+            412:
+               $ref: '#/components/responses/ResourceVersionMismatch'
+
+components:
+
+   schemas:
+
+# Common schema
+
+      Error:
+         type: object
+         additionalProperties: true
+         required:
+            - error
+         properties:
+            "error":
+               type: string
+               description: A human readable error message of what went wrong.
+
+      Extensions:
+         type: object
+         additionalProperties: true
+         description: Allows arbitrary properties as extension to the ones
+                      specified by the Hono API.
+
+# Tenant schema
+
+      Tenant:
+         type: object
+         additionalProperties: false
+         properties:
+            "enabled":
+               type: boolean
+               default: true
+            "ext":
+               $ref: '#/components/schemas/Extensions'
+            "adapters":
+               type: array
+               description: Only a single entry per type is allowed. If multiple entries for the same type are present it is handled as an error.
+               items:
+                  $ref: '#/components/schemas/Adapter'
+            "limits":
+               $ref: '#/components/schemas/ResourceLimit'
+            "trusted-ca":
+               $ref: '#/components/schemas/TrustedCA'
+
+      TrustedCA:
+         type: object
+         additionalProperties: false
+         required:
+            - "subject-dn"
+         properties:
+            "subject-dn":
+               type: string
+               description: The subject DN of the trusted root certificate in
+                            the format defined by RFC 2253.
+            "public-key":
+               type: string
+               format: byte
+               description: The Base64 encoded binary DER encoding of the
+                            trusted root certificate’s public key.
+                            Either this property or `cert` must be set.
+            "cert":
+               type: string
+               format: byte
+               description: The Base64 encoded binary DER encoding of the
+                            trusted root certificate. Either this property
+                            or `public-key` must be set.
+
+      Adapter:
+         type: object
+         additionalProperties: true
+         required:
+            - type
+         properties:
+            "type":
+               type: string
+            "enabled":
+               type: boolean
+               default: false
+            "device-authentication-required":
+               type: boolean
+               default: true
+            "ext":
+               $ref: '#/components/schemas/Extensions'
+
+      ResourceLimit:
+         type: object
+         additionalProperties: false
+         properties:
+            "max-connections":
+               type: integer
+               default: -1
+               description: The maximum number of concurrent connections allowed
+                            from devices of this tenant. The default value -1
+                            indicates that no limit is set.
+            "ext":
+               $ref: '#/components/schemas/Extensions'
+
+# Devices schema
+
+      Device:
+         type: object
+         additionalProperties: false
+         properties:
+            "enabled":
+               type: boolean
+               default: true
+            "ext":
+               $ref: '#/components/schemas/Extensions'
+
+# Credentials
+
+      CredentialsSet:
+         type: array
+         items:
+            $ref: '#/components/schemas/TypedSecret'
+
+      TypedSecret:
+         additionalProperties: false
+         oneOf:
+            - $ref: '#/components/schemas/PasswordSecret'
+            - $ref: '#/components/schemas/PSKSecret'
+            - $ref: '#/components/schemas/X509CertificateSecret'
+         discriminator:
+            propertyName: type
+            mapping:
+               "hashed-password": '#/components/schemas/PasswordSecret'
+               "psk": '#/components/schemas/PSKSecret'
+               "x509-cert": '#/components/schemas/X509CertificateSecret'
+
+      CommonSecret:
+         type: object
+         additionalProperties: false
+         required:
+            - auth-id
+            - type
+         properties:
+            "type":
+               type: string
+            "auth-id":
+               type: string
+            "enabled":
+               type: boolean
+               default: true
+            "not-before":
+               type: string
+               format: date-time
+            "not-after":
+               type: string
+               format: date-time
+            "comment":
+               type: string
+
+      X509CertificateSecret:
+         additionalProperties: false
+         allOf:
+            - $ref: '#/components/schemas/CommonSecret'
+
+      PasswordSecret:
+         additionalProperties: false
+         allOf:
+            - $ref: '#/components/schemas/CommonSecret'
+            - type: object
+              additionalProperties: false
+              properties:
+                 "hash-function":
+                    type: string
+                    example: sha-512
+                 "pwd-hash":
+                    type: string
+                    format: byte
+                 "salt":
+                    type: string
+                    format: byte
+
+      PSKSecret:
+         additionalProperties: false
+         allOf:
+            - $ref: '#/components/schemas/CommonSecret'
+            - type: object
+              additionalProperties: false
+              required:
+                 - key
+              properties:
+                 "key":
+                    type: string
+                    format: byte
+
+   parameters:
+
+      resourceVersion:
+         name: If-Match
+         in: header
+         description: The expected resource version
+         required: false
+         schema:
+            type: string
+
+      tenantId:
+         name: tenantId
+         in: path
+         description: The ID of the tenant
+         required: true
+         schema:
+            type: string
+         example: DEFAULT_TENANT
+
+      deviceId:
+         name: deviceId
+         in: path
+         description: The ID of the device
+         required: true
+         schema:
+            type: string
+         example: 4711
+
+      authId:
+         name: authId
+         in: path
+         description: The authentication ID of the device
+         required: true
+         schema:
+            type: string
+         example: sensor1
+
+      type:
+         name: type
+         in: path
+         description: The credentials type
+         required: true
+         schema:
+            type: string
+         example: sha-256
+
+   responses:
+
+      Unauthorized:
+         description: Authentication credentials are required, but missing.
+         headers:
+            "WWW-Authenticate":
+               schema:
+                  type: string
+
+      Created:
+         description: Object created.
+         headers:
+            Location:
+               description: URL to the resource
+               schema:
+                  type: string
+                  format: uri
+            ETag:
+               description: The new version of the resource
+               schema:
+                  type: string
+         content:
+            application/json:
+               schema:
+                  type: object
+                  additionalProperties: false
+                  required:
+                     - id
+                  properties:
+                     id:
+                        type: string
+                        description: The ID of the created object
+
+      Updated:
+         description: Object updated.
+         headers:
+            ETag:
+               description: The new version of the resource
+               schema:
+                  type: string
+
+      Deleted:
+         description: Object deleted.
+
+      NotFound:
+         description: |
+            Object not found. This may also be returned for some operations
+            if the user misses read access for the object.
+         content:
+            application/json:
+               schema:
+                  $ref: '#/components/schemas/Error'
+
+      NotAllowed:
+         description: |
+            Operation not allowed. If the user does not have read access
+            for this object, then `404` will be returned instead.
+         content:
+            application/json:
+               schema:
+                  $ref: '#/components/schemas/Error'
+
+      MalformedRequest:
+         description: Malformed request
+         content:
+            application/json:
+               schema:
+                  $ref: '#/components/schemas/Error'
+
+      AlreadyExists:
+         description: |
+            Object already exists. If the user has no read access for
+            the existing object, then `403` should be returned instead.
+         content:
+            application/json:
+               schema:
+                  $ref: '#/components/schemas/Error'
+
+      ResourceVersionMismatch:
+         description: |
+            Expected resource version does not match current.
+            This can only happen when the request header `If-Match`
+            was set.
+         content:
+            application/json:
+               schema:
+                  $ref: '#/components/schemas/Error'
+
+   securitySchemes:
+
+      BearerAuth:
+         type: http
+         scheme: bearer
+
+      BasicAuth:
+         type: http
+         scheme: basic
--- a/site/content/release-notes.md
+++ b/site/content/release-notes.md
@@ -18,6 +18,10 @@ title = "Release Notes"
   type *hashed-password* and *x509-cert* for a configurable amount of time. Please refer to the
   [Device Registry Admin Guide]({{< ref "/admin-guide/device-registry-config.md" >}}) for details
   regarding the configuration properties to use.
+* There is now an official HTTP API for managing devices served by the
+  device registry. The API is defined using an OpenAPI v3 specification and is
+  part of the Git repository, and served at
+  https://eclipse.org/hono/api/device-registry-v1.yaml.
 
 ### API Changes
 


ctron:
I added the credentials API, to the best of my understanding.

I also started to work on a revised API, the current process is focused on the basics of tenant + registration. The main changes are:

* Provide a common error response mechanism
* Allow versioning data sets
  * GET requests can return a *Resource Version*
  * PUT requests can provide a *Resource Version* (from a GET) and the server must return 409 (Conflict) when the current resource version has changed, the client must re-try
* Data Model
  * Re-use the same object, instead of partially nesting
  * Drop the *ID* from the payload and put it into the path for create operations
  * TO DISCUSS: Add a few *common* properties (like `enabled`), but keep "allow additional"

This is not a final version, but a work in progress in order to discuss and refine it further. Feedback is welcome!

Points to discuss:

* Versioning the API – My current take on this would be to simply use a base prefix on the URL … e.g. `https://foo.bar/v1/`, but don't write it down in the spec.
* The use of "additional properties" – We could allow additional properties on the top level objects (like the current proposal) or require additional properties to go into (e.g.) a `data` child property. I think there are pros and cons, as always. Having a child property might make it easier to implement a new device registry, while additional properties on the root level might make it easier to support the current state. In any case I think we should define a few "well known" properties, like `enabled`, that the current Hono components use.
* … I guess a bunch more


Inspiration came from:

* https://blog.florimondmanca.com/restful-api-design-13-best-practices-to-make-your-users-happy
* https://restfulapi.net/rest-api-design-tutorial-with-example/
* https://blog.mwaysolutions.com/2014/06/05/10-best-practices-for-better-restful-api/

jbtrystram:
I am not sure about dropping the ID from the payload :  having tthe AMQP and the HTTP APIs aligned is a nice feature imo.

b-abel:
I like the improvements so far!
One remark regarding the Credentials:
In #871 we added the possibility to let the Device Registry do the hashing of the password (instead of leaving this to the client). So there is an alternative schema of the secret for POST and PUT operations that contains the property `"pwd-plain"` instead of `"hash-function"`, `"pwd-hash"` and `"salt"`.

sophokles73:
In fact, allowing the user to select a (potentially weak) hash function himself looks like a mistake to me in hindsight. I would therefore propose to **not** allow the user to provide the hashed password anymore at all but **require** him/her to provide the plain text password instead so that the registry can enforce proper usage of a strong hash function ...

ctron:
> In fact, allowing the user to select a (potentially weak) hash function himself looks like a mistake to me in hindsight. I would therefore propose to not allow the user to provide the hashed password anymore at all but require him/her to provide the plain text password instead so that the registry can enforce proper usage of a strong hash function ...

I fully agree. I think the device registry should be configured, by the administrator, to use whatever they think is reasonable. And then let them enforce the use of specific hash algorithm **and** salt. As right now you can still use an un-salted hash.

ctron:
> I am not sure about dropping the ID from the payload : having tthe AMQP and the HTTP APIs aligned is a nice feature imo.

So I think the APIs for AMQP and HTTP service different purposes. And have different behaviors.

For HTTP, you have the ID in the URL/path. Having them, in addition, in the payload only makes things confusing IMHO. Currently the device registry actually removes them before processing internally. So I think it should only live in one play, and for a RESTful API, that sounds like the path to me.

jbtrystram:
+1 for removing the client-side hash support :+1:  Transmitting a hashed password in the payload doesn't add any security anyway.

> For HTTP, you have the ID in the URL/path. Having them, in addition, in the payload only makes things confusing IMHO. Currently the device registry actually removes them before processing internally. So I think it should only live in one play, and for a RESTful API, that sounds like the path to me.

you're right, I did not notice the ID was already in the URL, so yeah that make sense to remove it from the payload.

sophokles73:
Thanks for putting up the swagger UIs :+1:  It's really convenient to review the API this way.

**POST** /tenants/{tenant}
* It is unclear whether the *tenant* path variable is required or not. IMHO it should be optional to also allow for implementations to create a (unique) ID. That way, clients would not be required to take part in the *guessing game* of picking an unused ID.
* The *Location* response header should be documented to contain the path to the created object (incl. the potentially created tenant ID).
* All properties should be optional. The *enabled * property should have a default value of `true`.

**GET** /tenants/{tenant}
* Not sure how a request could be malformed (400), given that it does not contain any payload.
* I like the idea of returning an object version which clients can use when PUTting an update. However, I wonder why we wouldn't use the standard HTTP *ETag* header for that purpose.

**PUT** /tenants/{tenant}
* I wonder why we wouldn't use the standard HTTP *ETag* header instead of *X-Resource-Version*

**POST** /registrations/{tenant}/{deviceId}
* Using the plural of noun *registration* sounds strange to me in this context. What about changing this to *devices* altogether?
* IMHO the *deviceId* parameter should be optional to also allow for implementations to create a (unique) ID. That way, clients would not be required to take part in the *guessing game* of picking an unused ID.
* The *Location* response header should be documented to contain the path to the created object (incl. the potentially created ID).
* The *enabled * property should be optional and have a default value of `true`.

**GET** /registrations/{tenant}/{deviceId}
* Not sure how a request could be malformed (400), given that it does not contain any payload.
* Same concerns regarding *X-Resource-Version*

ctron:
> POST /tenants/{tenant}
> * It is unclear whether the tenant path variable is required or not. IMHO it should be optional to also allow for implementations to create a (unique) ID. That way, clients would not be required to take part in the guessing game of picking an unused ID.

Currently it is required. But indeed, it might be a good idea to have an automatically generated tenant ID. If that is what you meant.

> GET /tenants/{tenant}, GET /registrations/{tenant}/{deviceId}
> * Not sure how a request could be malformed (400), given that it does not contain any payload.

Yes, there is no payload, but the request could be invalid otherwise. e.g. when we have something like paging, (start > end -> invalid request).

> PUT /tenants/{tenant}
> * I wonder why we wouldn't use the standard HTTP ETag header instead of X-Resource-Version

Great idea!

> POST /registrations/{tenant}/{deviceId}
> * Using the plural of noun registration sounds strange to me in this context. What about changing this to devices altogether?

The suggestions I read many times is to simply **always** use the plural form. I kept registration as I wanted to be close the current. But renaming this to "devices" would make absolute sense to me.

ctron:
Ok, I pushed a change. This should cover most of the feedback.

dejanb:
I thought a couple of times about changing of registrations to devices and now might be the right time to do it. Hierarchy tenants->devices->credentials look much clearer IMHO

sophokles73:
Looks good, just one more thing to fix. Clients need to specify the expected version of the resource in a PUT request in the *If-Match* header, not the *ETag* header (that is only used to convey the version of the resource in a response).
Accordingly, the status code to return in case the version of the resource to update is not as expected, is 412 (Precondition Failed) and not 409 (Conflict).

ctron:
I didn't know about `If-Match` either :grin: … I made the necessary changes. I also added the `If-Match` to the DELETE operation.

I will move on then to the *credentials* section.

sophokles73:
> I also added the If-Match to the DELETE operation.

So, you would *only* want to delete the object if its version was `A` but you would not want to delete it if its version was `B`? I wonder what the use case for that would be ....

ctron:
> So, you would only want to delete the object if its version was A but you would not want to delete it if its version was B? I wonder what the use case for that would be ....

Correct.

The reason for this would be to detect/prevent and accidental deletion of a re-created device. Assuming two operations in parallel:

a) delete device "A", create device "A"
b) get device "A", make decision to delete, delete device "A"

Depending on the order of operations the outcome might be different each time. And the intention of deleting "A" in b) would be based on the original "A", not the re-created "A".

If the user doesn't care in b), then the header parameter can simply be omitted and the re-created A gets deleted, but if the user cares, then there would be a way now.

Christian-Schmid:
Hi,

thank you very much for the API draft. Having a well defined Hono rest API is something I always wanted :-) 

One of the big benefits for such I see for writing web UIs for device / tenant management. 

For such operations one method however I sill would wish to be well-defined as-well: A _list_ operation on the devices (and probably for the tenants). I already implemented such method as a draft for our hono deployment (with pagination / querying). If you're interested I could post our signature here. 

Thanks
Chris

ctron:
> A list operation on the devices (and probably for the tenants)

Yes, I think that is a good idea. If you have something already, I would be glad to take a look at this.

Christian-Schmid:
Hi, 

I'm currently a bit busy, so here the api draft just as text (and not as a nice openapi doc):

GET call on `"/registration/<tenantId>"`

Query Parameters:
 - `q`: Perfoms a starts-with based search on the device-id
 - `limit`: Limits the maximum number of returned results per page (e.g. default with 25 and a maximum of 100) 
 - `skip` : The number of entries to "skip" in the result. Used to page through the result set.


Example:
`GET /registration/<tenantId>?q=dummy-dev&limit=25&skip=0`

Response:
```
{
   total: 2 
   "registrations": [
        {
            "enabled": true,
            "device-id": "dummy-dev-1"
        },
        {
            "enabled": true,
            "device-id": "dummy-dev-2"
        }
    ]
}
```

* The used pagination is basically described here: https://www.moesif.com/blog/technical/api-design/REST-API-Design-Filtering-Sorting-and-Pagination/#offset-pagination

* None of the parameters are required.

sophokles73:
> Yes, I think that is a good idea. 

I am a little more reluctant than @ctron. The *cons* listed in the provided link pose some interesting questions. Do we expect to impose some particular order for the objects, e.g. based on ID? That would probably alleviate the issue a little. And before the question comes up: I am quite strictly opposed to having a query API in addition to *get by ID* .... :-)

ctron:
> >  Yes, I think that is a good idea.
>
> I am a little more reluctant than @ctron. The cons listed in the provided link pose some interesting questions. Do we expect to impose some particular order for the objects, e.g. based on ID? That would probably alleviate the issue a little. And before the question comes up: I am quite strictly opposed to having a query API in addition to get by ID .... :-)

Well, I think it comes down to the fact that we will implement that anyway. And it looks like you did the same. So I would prefer to have a common API, rather than two different.

But I agree that this is definitely more complex than a simple get/post/put/delete operation. So we could see this as an add-on, and progress with the credentials API anyway. As a list operation should look the same for all APIs IMHO.

DanielMaier-BSI:
> The use of "additional properties" – We could allow additional properties on the top level objects (like the current proposal) or require additional properties to go into (e.g.) a data child property. I think there are pros and cons, as always. Having a child property might make it easier to implement a new device registry, while additional properties on the root level might make it easier to support the current state. In any case I think we should define a few "well known" properties, like enabled, that the current Hono components use.

I would prefer to have additional properties not on top level. We definitely will have some custom properties. If Hono now introduces properties with the same name later on we will have a name clash.

ctron:
> I would prefer to have additional properties not on top level. We definitely will have some custom properties. If Hono now introduces properties with the same name later on we will have a name clash.

That is why I offered the idea of a `data` section … have the "standard" Hono properties in the top level, and "extras" under something like `data`.

DanielMaier-BSI:
> That is why I offered the idea of a data section … have the "standard" Hono properties in the top level, and "extras" under something like data.

Yes I would prefer this style. At least in tenant and device object we already have custom properties. I am not sure at the moment if we also need it in credentials.

ctron:
> Yes I would prefer this style. At least in tenant and device object we already have custom properties. I am not sure at the moment if we also need it in credentials.

Who else would prefer this? Does GitHub support voting? :grinning:

Christian-Schmid:
Hi,

two more ideas, as general food for thought :-): 

* **ErrorFormat**:  Would it make sense also to define the error message format? In the AMQ API sometimes it's stated "message body MAY contain a detailed description of the error that occurred". For validation errors e.g. a json structure could be useful for human and technical clients. E.g. 
```
{
  "validationErrors": [
    {
      "message": "The input is missing the required field 'auth-id'",
      "error-code": "123"
    }
  ]
}
```
* **Response of Device Creating**: Recently we're discussing if it would be nice when we would return the `endpoint urls`, `tenant id` and `auth id` as a response to the `add credential` operation. One could take the result json directly to use it for it's device configuration. We didn't came to a conclusion ;-)

b-abel:
Another topic came to my mind and I'm not sure if it affects the REST API design  (or if it should). That is the character set of the different IDs and their encoding. The Hono APIs impose very little restrictions. 
For example following the [blog post](https://blog.bosch-si.com/developer/x-509-based-device-authentication-in-eclipse-hono/) about client certificate authenticatioon we get a credentials ID containing whitespaces. This is probably without noticing it, because it sneaks in when configuring a certificate. E.g.:
```
$ curl -i -H 'Content-Type: application/json' --data-binary @credentials.json https://hono.eclipse.org:28443/credentials/DEFAULT_TENANT
HTTP/1.1 201 Created
location: /credentials/DEFAULT_TENANT/CN=Super Hot Fuzz Device,O=ACME Inc./x509-cert
content-length: 0
```

ctron:
@b-abel I would assume this specific case gets URL encoded, (IIRC `%20` or `+`) … so that shouldn't be an issue.

ctron:
I would like to summarize today's community call:

* We will go for a "data" section, were implementations can store arbitrary information in their own format. That field will be named `ext` though (meaning "extension") as this is an extension to the core model
* We will also define a "search" or "query" operation. However, as the task is more complex than the rest of the API, we will work on this as a separate topic, and add this in something like a version "1.1" of the API. In the meantime we finish the other APIs.

As I will be out-of-office the next week, it might take a bit longer until I update the PR.

sophokles73:
Going through the Swagger UI again:
* I would rather not return a 403 in response to GET, PUT, DELETE requests for specific entities as this might be used for determining valid identifiers. Instead, I think we should always return 404 in cases where the client is not authorized to perform the operation on that entity. This approach is also explicitly mentioned in the [HTTP 1.1 RFC](https://tools.ietf.org/html/rfc7231#section-6.5.3).
* The description of the 412 status code for the PUT and DELETE operations erroneously mention the *ETag* request header. However, the expected resource version is actually conveyed in the `If-Match` request header.

site/static/api/device-registry-revised.yaml
```diff
@@ -0,0 +1,448 @@
...
+         properties:
+            "error":
+               type: string
+               description: A human readable error message of what went wrong.
+
+# Tenant schema
+
+      Tenant:
+         type: object
+         properties:
```
sophokles73: we need to add the *trusted-ca*, *adapters* and *resource-limits* properties as defined by the Tenant API

site/static/api/device-registry-revised.yaml
```diff
@@ -0,0 +1,448 @@
...
+            "enabled":
+               type: boolean
+               default: true
+            "subject-dn":
```
sophokles73: this property is not part of the tenant but is defined as part of *trusted-ca* (for which we should add another type definition)

ctron:
> I would rather not return a 403 in response to GET, PUT, DELETE requests for specific entities as this might be used for determining valid identifiers. Instead, I think we should always return 404 in cases where the client is not authorized to perform the operation on that entity. This approach is also explicitly mentioned in the HTTP 1.1 RFC.

At first I agreed with that, on a second thought I would suggest a bit more diverse handling of that situation:

I think it is a good approach to hide existence on the entity, however I would only do that if the user doesn't have "visibility" anyway.

So my proposal would be (if the user lacks "modify" of the entity):

* if the user can "read" the entity: return 403 for PUT and DELETE
* if the user cannot "read" the entity: return 404 for GET, PUT and DELETE

In the code this would be a quick check: if the user cannot "read" the entry -> return 404, then check for "modify".

I think this makes sense because otherwise you might run into the issue of a successful GET, followed by "Not found" when you want to delete. And that would be odd.

site/static/api/device-registry-revised.yaml
```diff
@@ -0,0 +1,448 @@
...
+
+      Device:
+         type: object
+         properties:
+            "enabled":
+               type: boolean
+               default: true
+            "ext":
+               type: object
+               additionalProperties: true
```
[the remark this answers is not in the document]
ctron: True, there is now need. I rather wanted to be explicit about that fact. Does it hurt to have it in addition? I would say "no".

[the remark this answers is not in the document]
sophokles73: it doesn't hurt but, FMPOV, the thing that needs to be made explicit is that at some points we *do not* allow additional properties ....

site/static/api/device-registry-revised.yaml
ctron: Updated.

site/static/api/device-registry-revised.yaml
ctron: Updated.

site/static/api/device-registry-revised.yaml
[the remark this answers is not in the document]
ctron: I did use an explicit type (`Extension`) now, which also contains a description.

ctron:
Following up on the 404 handling, we might consider something similar for the creation of entities.

Assuming you create a new tenant "foo", but don't have "read" access to "foo". In that case I would suggest to return 403 (Not allowed) instead of 409 (Already exists).

ctron:
Ok, I added a bit about security: simple added the bearer token type, added 401 responses in case credentials are required but missing.

I also added the missing types for the tenant.

Tomorrow I will check of types for the Device/Registration are missing and implement the "credentials" part.

ctron:
I added the basic set of credentials operations.

As a first step I added the basic concept of PUT, GET and DELETE for "all credentials" of a device.

PUT is actually *set* (create or update), as what always bothered me with the current API, is the requirement to "create" a new struct for e.g. "hashed-password". IMHO credentials should only be set to a device that was registered before. When the device gets deleted (unregistered) all credentials should be removed by the device registry itself. That means also that a PUT operation should only be successful if the device was already registered.

The question is, do we want to have such a generic *set* operation? I would like to add some more specialized versions (like password rollover, or setting only a specific type+authId combination). Having a generic *set all* operation would easily allow to implement all the other operations, based on it. On the other side, it may be difficult (or maybe even impossible) to implement this *set all* operation for some device registry implementations. One way to deal with this would be to make this optional.

With the data structure I moved away from the *array* concept, using maps instead. So instead of having an array of objects with `authId` and `type`, there is now a map structure of `type` -> `authId` -> `secrets`. I think this maps more to what we actually have, and removes the problem of duplicate keys. I put the `authId` beneath `type`, as I think that an *auth id* is specific to a *type* and not the other way round. Meaning: "for passwords the device is known as 'foo' and for PSK the device is known as 'bar'".

ctron:
I added, as an example, the method to get by "type + authId".

ctron:
The more I am playing around with this, the more I am thinking of changing the model.

The AMQP side of the API does fetch by *auth id* only, the type is more or less just a filter to skip irrelevant entries on the server side. The protocol adapters look up by *auth id* and then authentication the device. Getting the *device id* in the process.

The HTTP side of the API will focus on the *device id*. My assumption is, that you will look up a device and manage credentials.

So we do need two access paths to the set of credentials: "by auth id" and "by device id". However, "by auth id" will be more frequent.

Looking at that from the HTTP side the following would be fine:

~~~yaml
device4711:
  authSensor1:
    secrets:
      - type: psk
        key: ...
      - type: password
        password: ...
~~~

However, the *auth id* `authSensor1` must be unique for the whole tenant, looking from the AMQP side:

~~~yaml
authSensor1:
  deviceId: device4711
  secrets:
    - type: psk
      key: ...
    - type: password
      password: ...
~~~

I also think that we should drop the `enabled` property from the AMQP API. If the credentials are disabled, then they should not be offered to the protocol adapters. For HTTP this is different, as the HTTP API should be aware of the *enablement* and allow to manage that.

ctron:
I did refactor the credentials model a bit. I think this makes it much simpler, working with the HTTP API.

The *credentials set* is now seen as an attachment to the device itself. So from the HTTP side, you would "GET" all credentials and "PUT" all credentials. As long as the device is registered and the user has *read access* the call to  "GET" will never return 404, instead it will return an empty credentials set if there are none.

A lookup *by auth id* for the credentials (for the HTTP API!) is currently not specified, as that could be part of the follow up task, of creating a query API.

ctron:
I think it is time to finish up on this. All required parts seem to be there, so I removed the "draft" state.

sophokles73:
> However, the auth id authSensor1 must be unique for the whole tenant.

That is not true, it needs to be unique in combination with the credential type, i.e. the following is possible as well:

```
deviceA:
  popular-auth-id:
    - type: "psk"
      key: ...

deviceB
  popular-auth-id:
    - type: "hashed-password"
      pwd-hash: "..."

File: site/static/api/device-registry-revised.yaml [hunk 4]
```diff
@@ -0,0 +1,712 @@
...
+                        $ref: '#/components/schemas/CredentialsSet'
+               headers:
+                  ETag:
+                     description: Version of the resource
+                     schema:
+                        type: string
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
```
sophokles73: under which circumstances will this be returned in response to a read request?

site/static/api/device-registry-revised.yaml
```diff
@@ -0,0 +1,712 @@
...
+               "x509-cert": '#/components/schemas/X509CertificateSecret'
+
+      CommonSecret:
+         type: object
+         required:
+            - type
+         properties:
+            "type":
+               type: string
+            "not-before":
```
sophokles73: the *auth-id* property is missing

ctron: The `auth-id` is not part of the object itself, but the key of the map which holds this object.

site/static/api/device-registry-revised.yaml
ctron: Indeed never. That was a copy&paste issue. Same for the 412 on the GET call.

ctron:
> > However, the auth id authSensor1 must be unique for the whole tenant.
> 
> That is not true, it needs to be unique in combination with the credential type, i.e. the following is possible as well:
> 
> ```
> deviceA:
>   popular-auth-id:
>     - type: "psk"
>       key: ...
> 
> deviceB
>   popular-auth-id:
>     - type: "hashed-password"
>       pwd-hash: "..."
> ```

Hm ... I wasn't aware of that. I will change it.

site/static/api/device-registry-revised.yaml
```diff
@@ -0,0 +1,713 @@
...
+servers:
+   - url: '{corsProxy}{server}/v1'
+     variables:
+        server:
+           default: http://hono.eclipse.org:28080
+        corsProxy:
+           default: https://cors-anywhere.herokuapp.com/
+
+security:
+   - BearerAuth: []
```
sophokles73: what about Basic auth?

site/static/api/device-registry-revised.yaml
sophokles73: where is this defined? I cannot find it in the schema ...

site/static/api/device-registry-revised.yaml
```diff
@@ -0,0 +1,713 @@
...
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            404:
+               $ref: '#/components/responses/NotFound'
+
+      put:
+         tags:
+            - credentials
+         summary: Update credentials set for registered device
+         operationId: setAllCredentials
```
wistefan: Hope its ok to jump into this discussion. I really would like to use this api.

Shouldn't the credentialsSet be a parameter somewhere here?

site/static/api/device-registry-revised.yaml
```diff
@@ -0,0 +1,713 @@
...
+                     description: Version of the resource
+                     schema:
+                        type: string
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
+            409:
```
jbtrystram: if no read access it should answer 404 as well.. Having 403 to mask conflict and 404 to mask 403 allows someone to poke around and find outs existing devices ID, if i am correct.

site/static/api/device-registry-revised.yaml
```diff
@@ -0,0 +1,713 @@
...
+               $ref: '#/components/responses/AlreadyExists'
+
+   /tenants/{tenant}:
+
+      post:
+         tags:
+            - tenants
+         summary: Create new tenant
+         operationId: createTenantWithId
+         parameters:
```
wistefan: Even thought this naming is valid in terms of rest and openapi in general, its problematic for most of the open-api code generators(at least the java once). Since both, the path-parameter and the body are called "tenant", the generated interfaces would have duplicate variable names in the updateTenant and createTenantWithId methods. 
I would suggest to call the pathParameter something like tenantName or tenantId (similar to deviceId in the DevicesApi) to avoid such problems.

ctron: Interesting point, I just tried that:

~~~
$ npx @openapitools/openapi-generator-cli generate -i device-registry-revised.yaml -g java -o build/
npx: installed 1 in 1.252s
Exception in thread "main" org.openapitools.codegen.SpecValidationException: There were issues with the specification. The option can be disabled via validateSpec (Maven/Gradle) or --skip-validate-spec (CLI).
 | Error count: 9, Warning count: 0
Errors: 
	-attribute paths.'/devices/{tenant}/{deviceId}'(put).parameters.There are duplicate parameter values
	-attribute paths.'/credentials/{tenant}/{deviceId}'(get).parameters.There are duplicate parameter values
	-attribute paths.'/devices/{tenant}/{deviceId}'(delete).parameters.There are duplicate parameter values
	-attribute paths.'/devices/{tenant}/{deviceId}'(get).parameters.There are duplicate parameter values
	-attribute paths.'/credentials/{tenant}/{deviceId}'(put).parameters.There are duplicate parameter values
	-attribute paths.'/credentials/{tenant}/{deviceId}'(delete).parameters.There are duplicate parameter values
	-attribute paths.'/tenants/{tenant}'(put).parameters.There are duplicate parameter values
	-attribute paths.'/tenants/{tenant}'(delete).parameters.There are duplicate parameter values
	-attribute paths.'/devices/{tenant}/{deviceId}'(post).parameters.There are duplicate parameter values

	at org.openapitools.codegen.config.CodegenConfigurator.toClientOptInput(CodegenConfigurator.java:626)
	at org.openapitools.codegen.cmd.Generate.run(Generate.java:367)
	at org.openapitools.codegen.OpenAPIGenerator.main(OpenAPIGenerator.java:60)
~~~

So that is indeed something that needs fixing. Thanks!

ctron: Ok, I did clean up the parameter definitions. However I think the remaining validation errors are due to:

* https://github.com/OpenAPITools/openapi-generator/issues/2631
* https://github.com/OpenAPITools/openapi-generator/pull/2775

However that shouldn't be a real problem. I will check the body issue now.

ctron: Ok, I see your point now. Code generation is always fun :roll_eyes:

ctron: Should be fixed now.

site/static/api/device-registry-revised.yaml
```diff
@@ -0,0 +1,725 @@
...
+            - $ref: '#/components/schemas/PSKSecret'
+            - $ref: '#/components/schemas/X509CertificateSecret'
+         discriminator:
+            propertyName: type
+            mapping:
+               "hashed-password": '#/components/schemas/PasswordSecret'
+               "psk": '#/components/schemas/PSKSecret'
+               "x509-cert": '#/components/schemas/X509CertificateSecret'
+
+      CommonSecret:
```
wistefan: Maybe I'm missing something, but I think the CommonSecret should extend the TypedSecret. 
E.g.:
```YAML
CommonSecret:
      allOf:
        - $ref: '#/components/schemas/TypedSecret'
        - type: object
          properties:
            "type":
              type: string
            "enabled":
              type: boolean
              default: true
            "not-before":
              type: string
              format: date-time
            "not-after":
              type: string
              format: date-time
            "comment":
              type: string
```
Again, in terms of code generation for java, this would lead to a decoupled TypedSecret and would be hard to get back together with the concrete secrets. Currently you would get something like:

```Java
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type", visible = true)
@JsonSubTypes({
  @JsonSubTypes.Type(value = PasswordSecret.class, name = "hashed-password"),
  @JsonSubTypes.Type(value = PSKSecret.class, name = "psk"),
  @JsonSubTypes.Type(value = X509CertificateSecret.class, name = "x509-cert"),
})
public class TypedSecret { ... }

public class CommonSecret { ... }

public class X509CertificateSecret extends CommonSecret { ... }
```
where the "oneOf" information is (in best case) included as annotations, but not in the object hirarchy. The example is for spring(with jackson), if you use the plain java generator (or f.e. the vertx-generator) you get no connection between TypedSecret and Common/PasswordSecret/etc.  at all.

site/static/api/device-registry-revised.yaml
wistefan: Yes, looks good now.:)

site/static/api/device-registry-revised.yaml
ctron: I tried to explain this in previous comment. The idea was the following (and I didn't check for the generated code, but then again, code generation always misses something):

Each device has a list of authentication identities, which is defined as a `Map<String,AuthenticationIdentity>`. Each `AuthenticationIdentity` has a list of secrets, which could either be `PasswordSecret`, `PSKSecret`, or some other type.

So the `AuthenticationIdentity` should look like:

~~~java
class AuthenticationIdentity {
  List<TypedSecret> secrets;
}
~~~

The `TypedSecret` is the "union type" for a single secret entry. It is required to have `type`, but none of the other properties from `CommonSecret`.

The other secrets however share the fields from the `CommonSecret`, just because the choose to do so.

So again, looking at Java I would see this as:

~~~java
interface TypedSecret {
   String getType();
   String setType(String type);
}

class CommonSecret {
   public String getType() {
    return type;
  }

  public void setType(String type) {
    this.type = type;
  }
  // ...
}

class PassswordSecret extends CommonSecret {
}
~~~

My experience with the code generators from OpenAPI and Swagger showed that you will always have something that doesn't really work. And if it works fine in language A, it fails in language B. So, I am not sure how far we should push the spec file to produce code. I would see it more as a documentation.

ctron: But just to be clear. I am open to suggestions on how to improve this!

wistefan: Thanks for the explanation. It seems that this is a shortcoming of the generators, it makes totally sense the way you defined it.

ctron: @wistefan I am glad to hear that. I did in the meantime play around with the model, based on your proposal and what OpenAPI documents. I ended up with something that worked fine-ish in Java, was messed up both in Go and the actual Specification then (and I didn't test beyond that).

I guess we have to still wait a few centuries, for a modelling approach that just works :wink: 

So for the moment I would focus in the spec itself, and less on the code generation.

ctron: And just because I didn't say it explicitly. I am thankful for your input! Keep it coming!

ctron:
I added a note about the non-focus on code generators. If you think that is too grumpy, just let me know.

site/static/api/device-registry-revised.yaml
wistefan: Especially for java, there are quite some discussions on how to implement the support for one-of(and for that matter all-of and any-of) in the generators. I extended my generator now to produce it as: 
```Java
interface TypedSecret {
   String getType();
   String setType(String type);
}

class CommonSecret {
   public String getType() {
    return type;
  }

  public void setType(String type) {
    this.type = type;
  }
  // ...
}

class PassswordSecret extends CommonSecret implements TypedSecret {
}
```
which seems to work best at the moment. 
I also think it will take a while to for a working modelling approach(if there will be one at all).

site/static/api/device-registry-revised.yaml
```diff
@@ -0,0 +1,734 @@
...
+            412:
+               $ref: '#/components/responses/ResourceVersionMismatch'
+
+components:
+
+   schemas:
+
+# Common schema
+
+      Error:
```
wistefan: Since you already define a specific error response object, I think it would be good to include a (optional) error-code field. A frontend might use it to show localized messages or an automated client might use it for a more specific errorhandling.

jbtrystram: @wistefan which partsdid you implement ? I wanted to start this work during this week but there is no need to duplicate the effort if you've worked on it. Would you like some help ?

wistefan: The whole Rest-Part of the api, including persistence in a sql-database. Since I implemented it using micronaut I'm not totally sure if it will fit into the whole hono project and I still have some work to do until I can put it on github(f.e. resolving the dependencies to our internal openapi-generator). It will probably take a couple of weeks(~4) until I can prepare a PR.

jbtrystram: So you did not update the `HttpEndpoint` classes ?

wistefan: Correct.

ctron:
@sophokles73 I made the change requested by you of putting the "auth-id" explicitly into the secrets as a field. Please have a look if that is ok for you now.

jbtrystram:
I've started to work on implementing this in #1240 .

I noticed two things on the tenant API :
- `If-match` headers for resource version are missing in the GET and POST operations for tenants. Is it intentionnal ? If yes, what's the point of creating versionned resources if I cannot request them afterwards ?
- I expect this versionning feature to be optionnal. If a GET request comes in with a given _ tenantID_ without a version specified, and several versions exist for this value. Which one is used in the response ?
- I think there is a security issue in the conflict / no read acess response : a non authorized user receive `403 Unauthorized` insead of `409 Conflict` when trying to _ POST_ (add operation) a value while having no read access. That means I can discover the existing tenantID values by poking around even i I have no read access : i'll get `403 Unauthorized` instead of `404 Not found`, hence knowing the _tenantID_ I just tried exists.

ctron:
>     * `If-match` headers for resource version are missing in the GET and POST operations for tenants. Is it intentionnal ? If yes, what's the point of creating versionned resources if I cannot request them afterwards ?

For the `GET`method yes. You already get the most recent version, but you also, in the response headers (the etag) the version you can use for modifying operations.

>     * I expect this versionning feature to be optionnal. If a GET request comes in with a given _tenantID_ without a version specified, and several versions exist for this value. Which one should be used in the response ?

The versioning is indeed optional. If you don't provide it in a modifying operation, the version check will be skipped.

>     * I think there is a security issue in the conflict / no read acess response : a non authorized user receive `403 Unauthorized` insead of `409 Conflict` when trying to _ POST_ (add operation) a value while having no read access. That means I can discover the existing tenantID values by poking around even i I have no read access : i'll get `403 Unauthorized` instead of `404 Not found`, hence knowing the _tenantID_ I just tried exists.

Looks like it, I will check again.

ctron:
> > * I think there is a security issue in the conflict / no read acess response : a non authorized user receive `403 Unauthorized` insead of `409 Conflict` when trying to _ POST_ (add operation) a value while having no read access. That means I can discover the existing tenantID values by poking around even i I have no read access : i'll get `403 Unauthorized` instead of `404 Not found`, hence knowing the _tenantID_ I just tried exists.
> 
> Looks like it, I will check again.

After reading through it once again, I think it is complicated :grin: 

So yes, the scenario you described would give out the information if the tenant exists or not. I was assuming that the user either has "create tenant" access, in which case you would get `409` or `201`, depending on the actual outcome. And in the case of "not having create tenant access", you would only get `409` if he also has "read this tenant" access. If he hasn't, he would get `403`, like he would get when trying to create any other tenant (as he doesn't have "create tenant" access).

Now you could have the situation that a user has "create tenant" access, but doesn't have access for this specific tenant. Honestly I don't know what to do in that case, because the tenant name is a unique element, and you simply have to reject the request.

On the other hand I think it is fair to say that, if the user has a global "create tenant" access role, then he can also know if that specific tenant exists or not. He still woulnd't see the content of that tenant though.

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
+openapi: 3.0.1
+info:
+   title: Eclipse Hono™ Device Registry API
+   description: |
+      This API allows one to interact with the Eclipse Hono Device Registry
+      API. It acts as a common basis which all Hono device registries should
```
sophokles73: there is no such thing as the *Device Registry API*. Instead, we have  *Tenant*, *Device Registration* and *Credentials* APIs.

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+
+      This model is not optimized for generating code from it. Code generators
+      try to understand the model and then translate this into the require
+      programming language. Even if the there would be no bugs in the code
+      generators, that process would be already only be an approximation. So
+      this model does focus in the description of the API, and doesn't tweak
+      the specification in a way to please code generators.
+
+   contact:
+      email: "hono-dev@eclipse.org"
```
sophokles73: this seems a little problematic, as you can only send to this address, if you have subscribed to the list ...

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+            201:
+               $ref: '#/components/responses/Created'
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
+            409:
+               $ref: '#/components/responses/AlreadyExists'
```
sophokles73: IMHO this cannot happen as the ID is auto generated, can it?

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+paths:
+
+# Tenant API
+
+   /tenants:
+
+      post:
+         tags:
+            - tenants
+         summary: Create new tenant with auto-generated ID
```
sophokles73: is the identifier called *name* (as below) or *ID*, as here?

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+      parameters:
+         - $ref: '#/components/parameters/tenantName'
+
+      post:
+         tags:
+            - tenants
+         summary: Create new tenant
+         operationId: createTenantWithId
+         requestBody:
+            description: New device registration
```
sophokles73: > New tenant information

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+            required: true
+         responses:
+            204:
+               $ref: '#/components/responses/Updated'
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
```
sophokles73: remind me again: I get a 403 if I am not allowed to update the tenant but I am allowed to read the tenant, and I get a 404 if I am not allowed to read the tenant as well, right?
If so, I find the advantage of having a separate 403 very questionable in comparison to the confusion this seems to create (with me at least) and effort required to implement this distinction ...

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+         operationId: deleteTenant
+         parameters:
+            - $ref: '#/components/parameters/resourceVersion'
+         responses:
+            204:
+               description: object deleted
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
```
sophokles73: same here, FMPOV we should simply always return 404

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+         requestBody:
+            description: New device
+            content:
+               application/json:
+                  schema:
+                     $ref: '#/components/schemas/Device'
+            required: true
+         responses:
+            201:
+               description: device registered
```
sophokles73: shouldn't this be a `#ref` to the *Created* response?

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+            required: true
+         responses:
+            204:
+               $ref: '#/components/responses/Updated'
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
```
sophokles73: again, IMHo always returning 404 would be easier to understand and implement

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+         operationId: deleteRegistration
+         parameters:
+            - $ref: '#/components/parameters/resourceVersion'
+         responses:
+            204:
+               $ref: '#/components/responses/Delete'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
```
sophokles73: and again

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+         - $ref: '#/components/parameters/tenantName'
+         - $ref: '#/components/parameters/deviceId'
+
+      get:
+         tags:
+            - credentials
+         summary: Get credentials set of a device.
+         description: |
+            Get the credentials set of a device. As long as the device is
+            registered and the user has read access to is, this call should
```
sophokles73: > .. read access to it, ...

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+         operationId: getAllCredentials
+         responses:
+            200:
+               description: Operation successful
+               content:
+                  application/json:
+                     schema:
+                        $ref: '#/components/schemas/CredentialsSet'
+               headers:
+                  ETag:
```
sophokles73: this needs to be marked as *optional*, doesn't it?

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+         operationId: getRegistration
+         responses:
+            200:
+               description: operation successful
+               content:
+                  application/json:
+                     schema:
+                        $ref: '#/components/schemas/Device'
+               headers:
+                  ETag:
```
sophokles73: this needs to be marked as *optional*, doesn't it?

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+         operationId: getTenant
+         responses:
+            200:
+               description: operation successful
+               content:
+                  application/json:
+                     schema:
+                        $ref: '#/components/schemas/Tenant'
+               headers:
+                  ETag:
```
sophokles73: this needs to be marked as *optional*, doesn't it?

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+            required: true
+         responses:
+            204:
+               $ref: '#/components/responses/Updated'
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
```
sophokles73: use 404 instead FMPOV

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+         operationId: deleteAllCredentials
+         parameters:
+            - $ref: '#/components/parameters/resourceVersion'
+         responses:
+            204:
+               $ref: '#/components/responses/Delete'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            403:
+               $ref: '#/components/responses/NotAllowed'
```
sophokles73: 404

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+               $ref: '#/components/schemas/TrustedCA'
+
+      TrustedCA:
+         type: object
+         additionalProperties: false
+         required:
+            - "subject-dn"
+         properties:
+            "subject-dn":
+               type: string
```
sophokles73: > description: The subject DN of the trusted root certificate in the format defined by RFC 2253.

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+            "public-key":
```
sophokles73: > description: The Base64 encoded binary DER encoding of the trusted root certificate’s public key. Either this property or "cert" must be set.

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+               type: string
+               format: byte
+            "cert":
```
sophokles73: > description: The Base64 encoded binary DER encoding of the trusted root certificate. Either this property or "public-key" must be set.

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+            "device-authentication-required":
+               type: boolean
+               default: true
+            "ext":
+               $ref: '#/components/schemas/Extensions'
+
+      ResourceLimit:
+         type: object
+         additionalProperties: false
+         properties:
```
sophokles73: IMHO we should allow registration of additional/custom limits here via the *ext* mechanism

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+            "max-connections":
```
sophokles73: > description: The maximum number of concurrent connections allowed from devices of this tenant. The default value -1 indicates that no limit is set.

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+            "not-before":
+               type: string
+               format: date-time
+            "not-after":
+               type: string
+               format: date-time
+            "comment":
+               type: string
+
+      X509CertificateSecret:
```
sophokles73: can this secret have additional properties?

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+         allOf:
+            - $ref: '#/components/schemas/CommonSecret'
+
+      PasswordSecret:
```
sophokles73: can this secret have additional properties?

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,724 @@
...
+                    type: string
+                    example: sha-512
+                 "pwd-hash":
+                    type: string
+                    format: byte
+                 "salt":
+                    type: string
+                    format: byte
+
+      PSKSecret:
```
sophokles73: can this secret have additional properties?

site/static/api/device-registry-v1.yaml
ctron: This should be clear from the description of the `NotAllowed` schema:

> Operation not allowed. If the user does not have read access for this object, then `404` will be returned instead.

So if the user has read access the object `403` should be returned, otherwise `404.

site/static/api/device-registry-v1.yaml
ctron: I am open to suggestions.

site/static/api/device-registry-v1.yaml
ctron: Indeed. Fixed.

site/static/api/device-registry-v1.yaml
ctron: Changed it to "ID" in order to be consistent also with "device ID"

site/static/api/device-registry-v1.yaml
jbtrystram: being responded with 404 when you know the tenant exists but are not allowed to modify it would be very confusing IMHO. How would you understand that you lack the proper permissions with a 404 ? If I was confronted with this (i.e. can read the tenant fine but given 404 when trying to edit) I would loose a lot of time and probably think the service doesn't work properly

site/static/api/device-registry-v1.yaml
ctron: From the OpenAPI documentation:

> By default, OpenAPI treats all request parameters as optional. You can add required: true to mark a parameter as required. Note that path parameters must have required: true, because they are always required.

site/static/api/device-registry-v1.yaml
ctron: See above.

site/static/api/device-registry-v1.yaml
ctron: See above.

site/static/api/device-registry-v1.yaml
ctron: Done.

site/static/api/device-registry-v1.yaml
ctron: Done.

site/static/api/device-registry-v1.yaml
ctron: Not to my knowledge. However it is a distinct type via the "discriminator" value in `TypedSecret`.

site/static/api/device-registry-v1.yaml
ctron: Not to my knowledge.

site/static/api/device-registry-v1.yaml
ctron: Not to my knowledge.

site/static/api/device-registry-v1.yaml
ctron: Fixed.

site/static/api/device-registry-v1.yaml
ctron: Changed wording.

site/static/api/device-registry-v1.yaml
ctron: > remind me again: I get a 403 if I am not allowed to update the tenant but I am allowed to read the tenant, and I get a 404 if I am not allowed to read the tenant as well, right?

Correct.

I think it is important to make a distinct error here. Assuming you have read access to the object. You are going ahead and try to modify it. Your first do a get (OK) and then to an update, which results in "not found".  That seems weird to me. I think it is more important to prevent confusing the end user, considering the developers.

Of course if you (as implementer of this API) can always do the shortcut of returning 404 in any case, if you think that is appropriate to your implementation.

ctron: The original comment on that would be: https://github.com/eclipse/hono/pull/1130#issuecomment-487567019

site/static/api/device-registry-v1.yaml
sophokles73: I would replace the *email* property with a *url* property pointing to https://www.eclipse.org/hono/community/get-in-touch/

site/static/api/device-registry-v1.yaml
sophokles73: I was wondering, if the way that the object is defined, would allow clients to set additional properties on it ...

site/static/api/device-registry-v1.yaml
ctron: Done.

site/static/api/device-registry-v1.yaml
ctron: Ok, now I understand what you meant. Done.

site/static/api/device-registry-v1.yaml
ctron: See above.

site/static/api/device-registry-v1.yaml
ctron: See above.

ctron:
> BTW I haven't taken a look at the spec for the existing API. Do you really want to add it as well?

No, before merging, I would like to drop the "old" API spec, and squash the PR. I would still recommend to keep the existing API implemented for a release or two. Just to let people continue to work with existing documentation and not break those.

site/static/api/device-registry-v1.yaml
ctron: Fixed.

site/static/api/device-registry-v1.yaml
ctron: Done.

site/static/api/device-registry-v1.yaml
ctron: Done.

site/static/api/device-registry-v1.yaml
ctron: Done.

site/static/api/device-registry-v1.yaml
```diff
@@ -0,0 +1,745 @@
...
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            404:
+               $ref: '#/components/responses/NotFound'
+
+      put:
+         tags:
+            - credentials
+         summary: Update credentials set for registered device
```
jbman: The only way to add a credentials seems to be getting all credentials, then modifying, deleting or adding an entry and then putting the result back. This requires some client side logic  for JSON parsing to find the right entry. Dedicated URLS for putting and deleting an entry would add much more convenience for the client. A use case like credential rotation can then be implemented without any parsing.

PUT /credentials/{tenantId}/{deviceId}/{auth-id}
DELETE /credentials/{tenantId}/{deviceId}/{auth-id}

Updating a set at once (PUT /credentials/{tenantId}/{deviceId}}) may not be needed when individual URLs are provided.
Deleting all at once would still be convenient for cleaning up all credentials at once. 

Was the option of providing dedicated auth id update considered? 
This is just an API consumers view, there might be some reasons internally, why this isn't a good idea :-)

sophokles73: makes sense to me. However, the resources would need to include the credentials type as well, i.e.
```
PUT /credentials/{tenantId}/{deviceId}/{type}/{auth-id}
DELETE /credentials/{tenantId}/{deviceId}/{type}/{auth-id}
```

jbtrystram: > Was the option of providing dedicated auth id update considered?

> However, the resources would need to include the credentials type

That is how the current API look like : https://current-api-hono-device-registry-api.wonderful.iot-playground.org/
I am not sure why those operation have been removed. Also : how does one create a credential if none exist yet ?

ctron: I touched that point in the following comment: https://github.com/eclipse/hono/pull/1130#issuecomment-487902225

I agree, I think it makes sense to have such specialized operations. However they can all be implemented on top of this API. And I think it makes sense to add them in a later version of the API. However I don't think it conflicts with the current API definition.

Assuming you have all kinds of specialized secrets (Password, PSK, client cert, custom protocol adapter secrets) I think it makes sense to support them in the way currently described. And then adding more specific operations on top of it, like:

* Cleanup expired passwords
* Password / Key rollover
* …

So, let's finish this first. And then do version 1.1, including search and specialized secret operations.

PR #1133, opened 2019-04-03
[#927] Add user guide for limiting connections per tenant.

kaniyan:

PR #1139, opened 2019-04-04
[#929 Part-1]  Add data volume limits to the tenant config

kaniyan:
1. This is the first part of the PR for #929. This PR adds provision to set the volume limit (max-bytes-per-month) to the tenant configuration.
2. As [discussed](https://github.com/eclipse/hono/pull/1133#discussion_r273019726), the tenant configuration parameter `limits` has been renamed to `resource-limits`.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/TenantConstants.java
+++ b/core/src/main/java/org/eclipse/hono/util/TenantConstants.java
@@ -75,19 +75,9 @@ public final class TenantConstants extends RequestResponseApiConstants {
     public static final String EVENT_BUS_ADDRESS_TENANT_IN = "tenant.in";
 
     /**
-     * The name of the property that contains the configuration options for limits.
+     * The name of the property that contains the configuration options for the resource limits.
      */
-    public static final String LIMITS = "limits";
-
-    /**
-     * The default value for the maximum number of connections to be allowed is -1, which implies no limit.
-     */
-    public static final long DEFAULT_MAX_CONNECTIONS = -1;
-
-    /**
-     * The name of the property that contains the maximum number of connections to be allowed for a tenant.
-     */
-    public static final String MAX_CONNECTIONS = "max-connections";
+    public static final String FIELD_RESOURCE_LIMITS = "resource-limits";
 
     /**
      * Request actions that belong to the Tenant API.
--- a/core/src/main/java/org/eclipse/hono/util/TenantObject.java
+++ b/core/src/main/java/org/eclipse/hono/util/TenantObject.java
@@ -44,7 +44,6 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 
-
 /**
  * Encapsulates the tenant information that was found by the get operation of the
  * <a href="https://www.eclipse.org/hono/api/tenant-api/">Tenant API</a>.
@@ -498,16 +497,25 @@ public final class TenantObject extends JsonBackedValueObject {
     }
 
     /**
-     * Gets the connections limit for the tenant if configured, else returns
-     * {@link TenantConstants#DEFAULT_MAX_CONNECTIONS}.
+     * Gets the resource limits for the tenant.
      *
-     * @return The connections limit.
+     * @return The resource limits or {@code null} if not set.
      */
     @JsonIgnore
-    public long getConnectionsLimit() {
-        return Optional.ofNullable(getProperty(TenantConstants.LIMITS))
-                .map(limits -> getProperty((JsonObject) limits, TenantConstants.MAX_CONNECTIONS,
-                        (Number) TenantConstants.DEFAULT_MAX_CONNECTIONS).longValue())
-                .orElse(TenantConstants.DEFAULT_MAX_CONNECTIONS);
+    public JsonObject getResourceLimits() {
+        return getProperty(TenantConstants.FIELD_RESOURCE_LIMITS);
+    }
+
+    /**
+     * Sets the resource limits for the tenant.
+     *
+     * @param resourceLimits The resource limits configuration to add.
+     * @return The TenantObject.
+     * @throws NullPointerException if resource limits to be set is {@code null}.
+     */
+    @JsonIgnore
+    public TenantObject setResourceLimits(final JsonObject resourceLimits) {
+        Objects.requireNonNull(resourceLimits);
+        return setProperty(TenantConstants.FIELD_RESOURCE_LIMITS, resourceLimits);
     }
 }
--- a/core/src/test/java/org/eclipse/hono/util/TenantObjectTest.java
+++ b/core/src/test/java/org/eclipse/hono/util/TenantObjectTest.java
@@ -14,6 +14,7 @@
 package org.eclipse.hono.util;
 
 import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.nullValue;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertThat;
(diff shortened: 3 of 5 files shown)


core/src/main/java/org/eclipse/hono/util/TenantConstants.java
```diff
@@ -89,6 +89,16 @@
...
 
+    /**
+     * The default value for the maximum data volume to be allowed for a tenant per month, which implies no limit.
+     */
+    public static final long DEFAULT_MAX_BYTES_PER_MONTH = -1;
+
+    /**
+     * The name of the property that contains the maximum data volume to be allowed for a tenant per month.
+     */
+    public static final String MAX_BYTES_PER_MONTH = "max-bytes-per-month";
```
sophokles73: This should better be renamed to `FIELD_MAX_BYTES_PER_MONTH` to make clear that this constant defienes the name of the field only and not the value
This also holds true for the *MAX_CONNECTIONS* constant ...

kaniyan: @sophokles73 Done :-)

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -505,9 +505,22 @@ public static JsonObject newAdapterConfig(final String type, final boolean enabl
...
                 .orElse(TenantConstants.DEFAULT_MAX_CONNECTIONS);
     }
+
+    /**
+     * Gets the data volume limit to be allowed for the tenant per month.
+     *
+     * @return The maximum bytes per month or {@link TenantConstants#DEFAULT_MAX_BYTES_PER_MONTH} if not set.
+     */
+    @JsonIgnore
+    public long getMaxBytesPerMonthLimit() {
```
sophokles73: I wonder if this is should better be more generic when it comes to the underlying time frame that is to be considered.
I could imagine to store the *maximum number of bytes* as well as the *number of days* that should be considered when determining if the limit has been reached. WDYT?

kaniyan: @sophokles73 It would be good to make it more generic. IMHO just the _number of days_ wouldn't make much sense without knowing _from which point of time_ to calculate the _data consumed_. 
Assume an user would like to limit the data volume to 200 MB every month for a tenant and lets say he/she configures as below.  
```
    "limits": {
      "max-bytes": 209_715_200,
      "max-bytes-period-in-days": 30
    }
```
IMHO calculating the data consumed for the last 30 days wouldn't make any sense. Calculating from the first of the month till the current day is one way. In this case, I can imagine something like the below configuration. If it is quarterly, then calculate from the beginning of this quarter till current date and so for half-yearly and yearly.
```
    "limits": {
      "max-bytes": 209_715_200,
      "max-bytes-period": weekly/monthly/quarterly/half-yearly/yearly
    }
```

As an alternative, the tenant registration date or an explicit user defined date can be used. In the below sample configuration, the time period is set to 90 days. The data consumed every 90 days starting from the device registration date can be used for the limit checks. But the reference device registry implementation doesn't have the information on which date a tenant is registered.
```
    "limits": {
      "max-bytes": 209_715_200,
      "max-bytes-period-in-days": 90
    }
```
WDYT?

sophokles73: > IMHO just the number of days wouldn't make much sense without knowing from which point of time to calculate the data consumed.

I agree.

I like the idea of using
```
"max-bytes-period-in-days": 90
```
to indicate the time period over which the limit is enforced. The limit on the data volume is enforced at the tenant level, not the device level. Thus, we only need to know the date at which the tenant has been created or the current plan has been chosen and can the deduce the *current period's starting date* based on that date. We will also need to add properties for tracking these dates to the tenant and/or its *limits* property.

kaniyan: @sophokles73 Sorry, my intention was also the tenant registration date and not that of the device. 

>  Thus, we only need to know the date at which the tenant has been created or the current plan has been chosen and can the deduce the _current period's starting date_ based on that date. We will also need to add properties for tracking these dates to the tenant and/or its _limits_ property.

A user can change a tenant plan to another plan, for example from a monthly to a quarterly cycle. I  think it makes more sense to use _when the current plan has been chosen_ instead of the _fixed tenant creation date_ .  I can imagine a property named `plan-start-date` or `effective-date`.  WDYT?

```
    "limits": {
      "max-connections": 100,
      "max-bytes": 209_715_200,
      "max-bytes-period-in-days": 90,
      "effective-date": "2019-04-23"
    }
```

If the above property is not set, the _Protocol Adapters_ consider the `max-bytes-...` config as invalid and refrain from checking the message limits based on data volume. This can be notified through a warn message in logs. WDYT?

sophokles73: sounds reasonable to me. I would prefer *effective-since* over *effective-date*, though.
We would then end up with three properties needed to define the data volume limit which to me indicates that we should introduce a new JSON object for it, don't you think?

kaniyan: @sophokles73 WDYT about the below config?

   ```
    "resource-limits": {
      "max-connections": 2,
      "max-messages": {
        "bytes-limit": 209_715_200,
        "period-in-days": 90,
        "effective-since": "2019-04-23"
      }
    }
```

Comment (sophokles73) [reply core/src/main/java/org/eclipse/hono/util/TenantObject.java]: well, we do not really care about the number of messages, do we? So FMPOV the name should not suggest that we do. What about simply calling it *transferred-data* or *data-volume* and rename the *bytes-limit* to *max-bytes*?

Comment (kaniyan) [reply core/src/main/java/org/eclipse/hono/util/TenantObject.java]: My intention was to keep it generic, so that any custom implementation of the `ResourceLimitChecks`  are not restricted to use only data volume to limit the incoming messages. Custom implementations should be able to use any attributes (no. of messages, data volume, etc.) or combination of them. Giving a more generic method name for the message limit check in `ResourceLimitChecks` is sufficient for that purpose. The limit configurations are used only within the `ResourceLimitChecks` implementation and they need not be generic. _data-volume_ and _max-bytes_ sounds good.

File: core/src/main/java/org/eclipse/hono/util/TenantObject.java [hunk 3]
```diff
@@ -505,9 +507,56 @@ public static JsonObject newAdapterConfig(final String type, final boolean enabl
...
+     * @throws DateTimeParseException if the date string fails to comply with the
+     *             {@link DateTimeFormatter#ISO_LOCAL_DATE} format.
+     */
+    public LocalDate getEffectiveSince() {
+        return Optional.ofNullable(getProperty(TenantConstants.FIELD_RESOURCE_LIMITS))
+                .map(limits -> getProperty((JsonObject) limits, TenantConstants.FIELD_DATA_VOLUME))
+                .map(dataVolumeLimits -> getProperty((JsonObject) dataVolumeLimits, TenantConstants.FIELD_EFFECTIVE_SINCE))
+                .map(effectiveSince -> LocalDate.parse((String) effectiveSince, DateTimeFormatter.ISO_LOCAL_DATE))
+                .orElse(null);
+    }
```
sophokles73: I am not sure if this is helpful. A client needs to retrieve all these properties individually and then pass them into the constructor of a `ResourceLimitChecks` instance. Wouldn't it make more sense to just use a *getResourceLimits* method to retrieve the `resource-limits* JSON object and then pass that in to the constructor?

kaniyan: These properties are not passed into the constructor of a `ResourceLimitChecks` instance. `ResourceLimitChecks` methods take a `TenantObject` instance as an argument and retrieve the individual properties.  Its makes more sense to have one _getResourceLimits_ method in `TenantObject`, so that any custom implementation of `ResourceLimitChecks` need not worry about the `TenantObject` class.

I will add one method _getResourceLimits_ to the `TenantObject ` and move the above methods to the default `ResourceLimitChecks` implementation which is `PrometheusBasedResourceLimitChecks`.

sophokles73: yes, I think that would be better, do you agree?

kaniyan: yes. Also there is no need to define properties names _(starting with FIELD\_*)_ and default values in `TenantConstants`.

kaniyan: As we discussed, I have made changes and pushed it.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -498,16 +497,12 @@ public static JsonObject newAdapterConfig(final String type, final boolean enabl
...
-     * @return The connections limit.
+     * @return The resource limits or {@code null} if not set.
      */
     @JsonIgnore
-    public long getConnectionsLimit() {
-        return Optional.ofNullable(getProperty(TenantConstants.LIMITS))
-                .map(limits -> getProperty((JsonObject) limits, TenantConstants.MAX_CONNECTIONS,
-                        (Number) TenantConstants.DEFAULT_MAX_CONNECTIONS).longValue())
-                .orElse(TenantConstants.DEFAULT_MAX_CONNECTIONS);
+    public JsonObject getResourceLimits() {
```
sophokles73: what about a setter?

kaniyan: Done.

kaniyan:
@sophokles73 Would you mind merging this.

kaniyan:
Thank you.

PR #1144, opened 2019-04-05
[#927] Improve Prometheus based limits implementation.

kaniyan:
1. Instead of framing url on every request now it is changed during @PostConstruct.
2. Method `queryForSumInPrometheus` is refactored to `executeQuery` so that it can be reused by the next PR to check limits based on data volume.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
@@ -27,6 +27,7 @@ import org.eclipse.hono.util.TenantObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.annotation.PostConstruct;
 import java.net.HttpURLConnection;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -118,11 +121,14 @@ public void setPort(final int port) {
         }
     }
 
-    private Future<Long> queryForSumInPrometheus(final String metricName, final TenantObject tenant) {
+    @PostConstruct
+    private void init() {
+        queryUrl = String.format("http://%s:%s/api/v1/query", getHost(), getPort());
+    }
+
+    private Future<Long> executeQuery(final String queryParams) {
```
sophokles73: My understanding of the code is that the parameter *is* the actual query, not  *queryParams*, isn't it?.

kaniyan: From the perspective of _PromQL_, it is query. In connection to the _httpRequest_ it is _queryParams_. I think in the above case, _query_ suits better than the _queryParams_. I will update it in one of the future PRs for _resource-limits_.

PR #1145, opened 2019-04-05
[#930] Add tenant-scoped command consumer

calohmn:
For #930:
This introduces a new command consumer on a tenant-scoped address.

This consumer handler will map the device id in the command message 'To' address to a corresponding 'via' device id (if applicable) and delegate the command either to a local command handler or to a consumer on another protocol adapter instance by way of the AMQP network.
As the existing command consumer address scheme isn't changed (yet), this change doesn't cause any incompatibilities.

Additional changes for getting this to work are in #1156 and #1157.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/CommandConsumerFactory.java
+++ b/client/src/main/java/org/eclipse/hono/client/CommandConsumerFactory.java
@@ -28,12 +28,13 @@ public interface CommandConsumerFactory extends ConnectionLifecycle {
     /**
      * Creates a new factory for an existing connection.
      *
-     * @param connection The connection to use.
+     * @param connection The connection to the AMQP network.
+     * @param gatewayMapper The component mapping a command device id to the corresponding gateway device id.
      * @return The factory.
-     * @throws NullPointerException if connection is {@code null}
+     * @throws NullPointerException if connection or gatewayMapper is {@code null}.
      */
-    static CommandConsumerFactory create(final HonoConnection connection) {
-        return new CommandConsumerFactoryImpl(connection);
+    static CommandConsumerFactory create(final HonoConnection connection, final GatewayMapper gatewayMapper) {
+        return new CommandConsumerFactoryImpl(connection, gatewayMapper);
     }
 
     /**
--- a/client/src/main/java/org/eclipse/hono/client/DelegatedCommandSender.java
+++ b/client/src/main/java/org/eclipse/hono/client/DelegatedCommandSender.java
@@ -1,0 +1,50 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.client;
+
+import org.eclipse.hono.config.ClientConfigProperties;
+
+import io.opentracing.SpanContext;
+import io.vertx.core.Future;
+import io.vertx.proton.ProtonDelivery;
+
+/**
+ * A sender to send command messages that are delegated to be processed by another command consumer.
+ * <p>
+ * This usually involves command messages first retrieved via a tenant-scoped consumer and then delegated back to the
+ * downstream peer so that they can be consumed by the device-specific consumer.
+ */
+public interface DelegatedCommandSender extends MessageSender {
+
+    /**
+     * Sends a command message to the downstream peer to be consumed by a device-specific consumer.
+     *
+     * @param command The command to send.
+     * @param context The currently active OpenTracing span or {@code null} if no
+     *         span is currently active. An implementation should use this as the
+     *         parent for any new span(s) it creates for tracing the execution of
+     *         this operation.
+     * @return A future indicating the outcome of the operation.
+     *         <p>
+     *         The future will succeed if the message has been accepted (and settled)
+     *         by the consumer.
+     *         <p>
+     *         The future will be failed with a {@link ServiceInvocationException} if the
+     *         message could not be sent or if no delivery update
+     *         was received from the peer within the configured timeout period
+     *         (see {@link ClientConfigProperties#getSendMessageTimeout()}).
+     * @throws NullPointerException if command is {@code null}.
+     */
+    Future<ProtonDelivery> sendCommandMessage(Command command, SpanContext context);
+}
--- a/client/src/main/java/org/eclipse/hono/client/GatewayMapper.java
+++ b/client/src/main/java/org/eclipse/hono/client/GatewayMapper.java
@@ -1,0 +1,42 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.client;
+
+import io.vertx.core.Future;
+
+/**
+ * A component that maps a given device to the gateway through which data was last published for the given device.
+ *
+ */
+public interface GatewayMapper extends ConnectionLifecycle {
+
+    /**
+     * Determines the gateway device id for the given device id (if applicable).
+     * <p>
+     * The value of the returned Future can be either
+     * <ul>
+     * <li>the gateway device id</li>
+     * <li>{@code null} if the device is configured to be accessed via a gateway (i.e. one or more 'via' devices are
+     * set), but no 'last-via' device is set, meaning that no message has been sent yet for this device.</li>
+     * <li>the given device id if the device is not configured to be accessed via a gateway</li>
+     * </ul>
+     *
+     * @param tenantId The tenant identifier.
+     * @param deviceId The device identifier.
+     * @return A succeeded Future containing the mapped gateway device id or {@code null};
+     *         or a failed Future if there was an error determining the mapped gateway device.
+     */
+    Future<String> getMappedGatewayDevice(String tenantId, String deviceId);
+
+}
--- a/client/src/main/java/org/eclipse/hono/client/impl/DelegatedCommandSenderImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/DelegatedCommandSenderImpl.java
@@ -1,0 +1,170 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.client.impl;
+
+import java.net.HttpURLConnection;
+import java.util.Objects;
+
+import org.apache.qpid.proton.message.Message;
+import org.eclipse.hono.client.Command;
+import org.eclipse.hono.client.DelegatedCommandSender;
+import org.eclipse.hono.client.HonoConnection;
+import org.eclipse.hono.client.ServerErrorException;
+import org.eclipse.hono.client.ServiceInvocationException;
+import org.eclipse.hono.tracing.TracingHelper;
+import org.eclipse.hono.util.CommandConstants;
+import org.eclipse.hono.util.MessageHelper;
+
+import io.opentracing.Span;
+import io.opentracing.SpanContext;
+import io.opentracing.tag.Tags;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.proton.ProtonDelivery;
+import io.vertx.proton.ProtonQoS;
+import io.vertx.proton.ProtonSender;
+
+/**
+ * A Vertx-Proton based sender to send command messages to the downstream peer as part of delegating them to be
+ * processed by another command consumer.
+ */
+public class DelegatedCommandSenderImpl extends AbstractSender implements DelegatedCommandSender {
+
+    DelegatedCommandSenderImpl(
+            final HonoConnection connection,
+            final ProtonSender sender) {
+
+        super(connection, sender, "", "");
+    }
+
+    @Override
+    protected String getTo(final String deviceId) {
+        return null;
+    }
+
+    @Override
+    public String getEndpoint() {
+        return CommandConstants.COMMAND_ENDPOINT;
+    }
+
+    @Override
+    protected Future<ProtonDelivery> sendMessage(final Message message, final Span currentSpan) {
+        return runSendAndWaitForOutcomeOnContext(message, currentSpan);
+    }
+
+    @Override
+    public Future<ProtonDelivery> sendAndWaitForOutcome(final Message message) {
+        return sendAndWaitForOutcome(message, null);
+    }
+
+    @Override
+    public Future<ProtonDelivery> sendAndWaitForOutcome(final Message rawMessage, final SpanContext parent) {
+        Objects.requireNonNull(rawMessage);
+
+        final Span span = startSpan(parent, rawMessage);
+        span.setTag(MessageHelper.APP_PROPERTY_TENANT_ID, MessageHelper.getTenantId(rawMessage));
+        span.setTag(MessageHelper.APP_PROPERTY_DEVICE_ID, MessageHelper.getDeviceId(rawMessage));
+        TracingHelper.injectSpanContext(connection.getTracer(), span.context(), rawMessage);
+
+        return runSendAndWaitForOutcomeOnContext(rawMessage, span);
+    }
+
+    private Future<ProtonDelivery> runSendAndWaitForOutcomeOnContext(final Message rawMessage, final Span span) {
+        return connection.executeOrRunOnContext(result -> {
+            if (sender.sendQueueFull()) {
+                final ServiceInvocationException e = new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "no credit available");
+                logError(span, e);
+                span.finish();
+                result.fail(e);
+            } else {
+                // TODO improve error-handling: we have to use a different kind of sendMessageAndWaitForOutcome() here
+                //  where a non-Accepted updatedProtonDelivery.getRemoteState() does not fail the returned Future
+                //  (we want to transfer such errors as is, not by way of translating them to/from ServiceInvocationExceptions)
+                sendMessageAndWaitForOutcome(rawMessage, span).setHandler(result);
+            }
+        });
+    }
+
+    @Override
+    public Future<ProtonDelivery> sendCommandMessage(final Command command, final SpanContext spanContext) {
+        Objects.requireNonNull(command);
+        final String tenantId = command.getTenant();
+        final String deviceId = command.getDeviceId();
+        final String targetAddress = getTargetAddress(tenantId, deviceId);
+        final String replyToAddress = command.isOneWay() ? null
+                : String.format("%s/%s/%s",
+                CommandConstants.COMMAND_ENDPOINT, tenantId, command.getReplyToId());
+        return sendAndWaitForOutcome(
+                createDelegatedCommandMessage(command.getCommandMessage(), targetAddress, replyToAddress),
+                spanContext);
+    }
+
+    /**
+     * Gets the AMQP <em>target</em> address to use for sending the delegated command messages to.
+     *
+     * @param tenantId The tenant identifier.
+     * @param deviceId The device identifier.
+     * @return The target address.
+     * @throws NullPointerException if tenant or device id is {@code null}.
+     */
+    static String getTargetAddress(final String tenantId, final String deviceId) {
+        return String.format("%s/%s/%s", CommandConstants.COMMAND_ENDPOINT, Objects.requireNonNull(tenantId),
+                Objects.requireNonNull(deviceId));
+    }
+
+    private static Message createDelegatedCommandMessage(final Message originalMessage, final String targetAddress,
+            final String replyToAddress) {
+        Objects.requireNonNull(targetAddress);
+        Objects.requireNonNull(originalMessage);
+        // copy original message
+        final Message msg = MessageHelper.getShallowCopy(originalMessage);
+        // set target address
+        msg.setAddress(targetAddress);
+        msg.setReplyTo(replyToAddress);
+        // use original message id as correlation id
+        msg.setCorrelationId(originalMessage.getMessageId());
+        return msg;
+    }
+
+    /**
+     * Creates a new sender for sending the delegated command messages to the AMQP network.
+     *
+     * @param con The connection to the AMQP network.
+     * @param closeHook A handler to invoke if the peer closes the link unexpectedly.
+     * @return A future indicating the result of the creation attempt.
+     * @throws NullPointerException if con is {@code null}.
+     */
+    public static Future<DelegatedCommandSender> create(
+            final HonoConnection con,
+            final Handler<String> closeHook) {
+
+        Objects.requireNonNull(con);
+
+        final String targetAddress = ""; // use anonymous relay (ie. use empty address)
+        return con.createSender(targetAddress, ProtonQoS.AT_LEAST_ONCE, closeHook)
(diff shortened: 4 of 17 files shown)


calohmn:
Rebased and resolved merge conflicts.
I'll still refactor this a bit (aiming to replace the new `registrationClientFactory` param in the `CommandConsumerFactory`).

BobClaerhout:
Any particular reason why this is not yet merged?
@calohmn: is the refactor big?

calohmn:
I've done some refactoring:
- The mapping logic is now in a separate class (`GatewayMapper` and `GatewayMapperImpl`) that also implements the `ConnectionLifecycle` interface. An object of this class is exposed as a bean and autowired in `AbstractProtocolAdapterBase`. That allows for potentially exchanging the implementation.
- The `GatewayMapperImpl` uses the `RegistrationClientFactory` internally to query the device registry. The `HonoConnection` for that is a separate one compared to the connection for the existing `AbstractProtocolAdapterBase.commandConsumerFactory` instance.
- The command handling logic for mapping the device id and for delegating the command request back to the AMQP network is in separate Handler classes now (`GatewayMappingCommandHandler` and `DelegatingCommandHandler`). I'll add unit tests for these once we have agreed on the general class structure.
- I had to add a new `init(GatewayMapper)` method to the `CommandConsumerFactory` interface. I wanted to add the GatewayMapper instance in the constructor, but that constructor is called from the `AbstractAdapterConfig` class - the `AbstractProtocolAdapterBase.gatewayMapper` instance can't be used from there.

client/src/main/java/org/eclipse/hono/client/CommandConsumerFactory.java
```diff
@@ -36,6 +36,16 @@ static CommandConsumerFactory create(final HonoConnection connection) {
...
 
+    /**
+     * Initializes the CommandConsumerFactory with the given GatewayMapper.
+     * <p>
+     * This needs to be called before calling any of the createCommandConsumer() methods.
+     *
+     * @param gatewayMapper The component mapping a command device id to the corresponding gateway device id.
+     * @throws NullPointerException if gatewayMapper is {@code null}.
+     */
+    void init(GatewayMapper gatewayMapper);
```
sophokles73: why can't we pass in the gateway mapper to the factory method and the `CommandConsumerFactoryImpl` constructor? You can then override `AbstractHonoClientFactory`'s *connect()* and *disconnect()* method to start/sop both connections, or wouldn't that work?

client/src/main/java/org/eclipse/hono/client/DelegatedCommandSender.java
```diff
@@ -0,0 +1,21 @@
...
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.client;
+
+/**
+ * A sender to send delegated command messages.
```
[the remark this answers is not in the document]
calohmn: That is a command message, first retrieved from the *tenant-scoped* consumer in the protocol adapter, and then *delegated* back to the AMQP message network so that it can be consumed by the *device-specific* consumer later.
I can extend the javadoc there.

client/src/main/java/org/eclipse/hono/client/GatewayMapper.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ */
+public interface GatewayMapper extends ConnectionLifecycle {
+
+    /**
+     * Determines the gateway device id for the given device id (if applicable).
+     * The value of the returned Future can be either
+     * <ul>
+     * <li>the gateway device id</li>
+     * <li>{@code null} if the device is configured to be accessed via a gateway but none is set</li>
```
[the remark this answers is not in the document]
calohmn: That is the case where the device is configured with a `via` property (ie. it is configured to be accessed via a gateway), but the `last-via` property is not set. 
That means that there hasn't been a message sent for this device yet.

client/src/main/java/org/eclipse/hono/client/GatewayMapper.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.client;
+
+import io.vertx.core.Future;
+
+/**
+ * A mapping component that determines the corresponding gateway device id for a given device id.
```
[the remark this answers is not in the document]
calohmn: Ok, I would rephrase it like this:
`A component that maps a given device to the gateway through which data was last published for the given device.`

client/src/main/java/org/eclipse/hono/client/TenantScopedCommandConsumer.java
```diff
@@ -0,0 +1,101 @@
...
+
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.proton.ProtonMessageHandler;
+import io.vertx.proton.ProtonQoS;
+import io.vertx.proton.ProtonReceiver;
+
+/**
+ * A wrapper around an AMQP receiver link for consuming commands that will be delegated to the appropriated
+ * handler/consumer.
```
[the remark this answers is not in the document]
calohmn: Right, the info is rather vague here.
I would change the javadoc to:
````
A wrapper around an AMQP receiver link for consuming commands on a tenant-scoped address.
<p>
This class is used by the default {@link CommandConsumerFactory} implementation to receive commands from northbound applications.
````
What is done with the command message afterwards is out of the scope of this class.
I can extend the javadoc of the `CommandConsumerFactoryImpl` to describe the handling of command messages there.

File: core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java [hunk 6]
```diff
@@ -274,4 +278,24 @@ public static void closeAndFree(
...
     }
+
+    /**
+     * Returns a copy of the given message.
+     * This is a shallow copy of the <em>Message</em> object, except for the copied <em>Properties</em>.
+     *
+     * @param message The message to copy.
+     * @return The message copy.
+     */
+    public static Message getMessageCopyWithCopiedProperties(final Message message) {
```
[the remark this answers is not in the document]
calohmn: Ok, yes, will move it there.
Concerning the name: Well, we could use `getCopyWithCopiedProperties` (or `getCopyWithPropertiesCopy`). Not the prettiest, but I'd rather keep the `withCopiedProperties` part in there to emphasize what the method is doing.

client/src/main/java/org/eclipse/hono/client/CommandConsumerFactory.java
calohmn: I thought about this, but decided against it.
This would mean that all the `ConnectionLifecycle` members of `AbstractProtocolAdapterBase` are connected in the `AbstractProtocolAdapterBase.connectToService` method - **except** the `gatewayMapper` field, where the connect invocation is hidden in the `CommandConsumerFactoryImpl` class.
This fact plus needing to override the `AbstractHonoClientFactory` methods makes this solution IMHO inferior to just adding the `init(GatewayMapper)` method.

calohmn: Or do you mean this along with removing the GatewayMapper field from the `AbstractProtocolAdapterBase` class?
I guess that would make exchanging the GatewayMapper implementation in a protocol adapter different from the way it would be done for the other factories.

sophokles73: I guess we do not need (direct) access to the gateway mapper from the protocol adapters anyway, do we? Why not make it an implementation detail of the `CommandConsumerFactory`?

core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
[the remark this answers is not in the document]
sophokles73: why not simply *getShallowCopy* and let users read the JavaDoc? in the end that is why you wrote it, isn't it?

client/src/main/java/org/eclipse/hono/client/GatewayMapper.java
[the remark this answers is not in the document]
sophokles73: Ah, can you somehow incorporate that into the JavaDoc?

core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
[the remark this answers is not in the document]
calohmn: ok, I can live with that.

client/src/main/java/org/eclipse/hono/client/DelegatedCommandSender.java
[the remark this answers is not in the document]
calohmn: See [commit](https://github.com/eclipse/hono/compare/08dfcff597695d2a40c589f4d25a36d52a5d470a..dcca29de3f945fbec1b912f5d6626174c4b5f341) below.

client/src/main/java/org/eclipse/hono/client/GatewayMapper.java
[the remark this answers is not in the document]
calohmn: See [commit](https://github.com/eclipse/hono/compare/08dfcff597695d2a40c589f4d25a36d52a5d470a..dcca29de3f945fbec1b912f5d6626174c4b5f341) below.

client/src/main/java/org/eclipse/hono/client/GatewayMapper.java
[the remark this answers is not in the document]
calohmn: See [commit](https://github.com/eclipse/hono/compare/08dfcff597695d2a40c589f4d25a36d52a5d470a..dcca29de3f945fbec1b912f5d6626174c4b5f341) below.

client/src/main/java/org/eclipse/hono/client/TenantScopedCommandConsumer.java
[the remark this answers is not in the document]
calohmn: See [commit](https://github.com/eclipse/hono/compare/08dfcff597695d2a40c589f4d25a36d52a5d470a..dcca29de3f945fbec1b912f5d6626174c4b5f341) below.

client/src/main/java/org/eclipse/hono/client/CommandConsumerFactory.java
calohmn: I've explored this solution a bit more and think we can go with it. I've amended the commit with the corresponding changes.

calohmn:
Rebased to resolve merge conflict.

client/src/main/java/org/eclipse/hono/client/impl/DelegatedCommandSenderImpl.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private Future<ProtonDelivery> runSendAndWaitForOutcomeOnContext(final Message rawMessage, final Span span) {
+        return connection.executeOrRunOnContext(result -> {
+            if (sender.sendQueueFull()) {
+                final ServiceInvocationException e = new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "no credit available");
+                logError(span, e);
+                span.finish();
+                result.fail(e);
+            } else {
+                sendMessageAndWaitForOutcome(rawMessage, span).setHandler(result.completer());
```
sophokles73: `Future.completer()` will be removed in vert.x 4 so I wouldn't use it anymore ...

client/src/main/java/org/eclipse/hono/client/impl/DelegatingCommandHandler.java
```diff
@@ -0,0 +1,157 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(DelegatingCommandHandler.class);
+
+    private final Function<String, Handler<CommandContext>> deviceSpecificCommandHandlerSupplier;
+    private final Function<String, Future<DelegatedCommandSender>> delegatedCommandSenderSupplier;
+
+    /**
+     * Creates a new DelegatingCommandHandler.
+     *
+     * @param deviceSpecificCommandHandlerSupplier Function to get an existing command handler for a given device.
```
sophokles73: so, the function's parameter is the device ID?

client/src/main/java/org/eclipse/hono/client/impl/DelegatingCommandHandler.java
```diff
@@ -0,0 +1,157 @@
...
+     * @param delegatedCommandSenderSupplier Supplies a sender for a given tenant to send the delegated command via the
```
sophokles73: so, the function's parameter is the tenant ID?
can the returned Future fail? what if there is no sender for the given tenant?

client/src/main/java/org/eclipse/hono/client/impl/DelegatingCommandHandler.java
```diff
@@ -0,0 +1,157 @@
...
+        delegatedCommandSender.setHandler(cmdSenderResult -> {
+            if (cmdSenderResult.succeeded()) {
+                final MessageSender sender = cmdSenderResult.result();
+                final String targetAddress = DelegatedCommandSenderImpl.getTargetAddress(tenantId, deviceId);
+                final String replyToAddress = command.isOneWay() ? null
+                        : String.format("%s/%s/%s",
+                                CommandConstants.COMMAND_ENDPOINT, tenantId, command.getReplyToId());
+                final Future<ProtonDelivery> protonDeliveryFuture = sender.sendAndWaitForOutcome(
+                        createDelegatedCommandMessage(command.getCommandMessage(), targetAddress, replyToAddress),
+                        commandContext.getTracingContext());
```
sophokles73: Isn't this the specific command context handling that `DelegatedCommandSender` could/should encapsulate? Why don't we define a method `DelegatedCommandSender.send(CommandContext)`?

client/src/main/java/org/eclipse/hono/client/impl/DelegatingCommandHandler.java
```diff
@@ -0,0 +1,157 @@
...
+                                new Exception("failed to send command message to downstream peer",
+                                        sendResult.cause()));
+                        commandContext.release();
+                    }
+                });
+            } else {
+                // failed to create sender
+                LOG.error("failed to create sender for sending command message to downstream peer", cmdSenderResult.cause());
+                TracingHelper.logError(commandContext.getCurrentSpan(),
+                        new Exception("failed to create sender for sending command message to downstream peer",
```
sophokles73: I do not really like using `Exception` ...

client/src/main/java/org/eclipse/hono/client/impl/GatewayMappingCommandHandler.java
```diff
@@ -0,0 +1,97 @@
...
+    public GatewayMappingCommandHandler(final GatewayMapper gatewayMapper, final Handler<CommandContext> nextCommandHandler) {
+        this.gatewayMapper = gatewayMapper;
+        this.nextCommandHandler = nextCommandHandler;
+    }
+
+    @Override
+    public void handle(final CommandContext originalCommandContext) {
+        final Command originalCommand = originalCommandContext.getCommand();
+        if (!originalCommand.isValid()) {
+            originalCommandContext.reject(new ErrorCondition(Constants.AMQP_BAD_REQUEST, "malformed command message"));
```
sophokles73: do we still need to execute the rest of the method if the command is being rejected?
I think we also should re-issue some credit so that the application can send another command, don't you think?

client/src/main/java/org/eclipse/hono/client/impl/GatewayMappingCommandHandler.java
```diff
@@ -0,0 +1,97 @@
...
+                        }
+                        final Command command = Command.from(originalCommand.getCommandMessage(), tenantId, lastViaDeviceId);
+                        commandContext = CommandContext.from(command, originalCommandContext.getDelivery(), originalCommandContext.getReceiver(), originalCommandContext.getCurrentSpan());
+                    }
+                    nextCommandHandler.handle(commandContext);
+                } else {
+                    // lastViaDeviceId is null - device hasn't connected yet
+                    LOG.error("last-via for device {} is not set", originalDeviceId);
+                    TracingHelper.logError(originalCommandContext.getCurrentSpan(), "last-via for device " + originalDeviceId + " is not set");
+                    originalCommandContext.release();
```
sophokles73: I think we also should re-issue some credit so that the application can send another command, don't you think?

client/src/main/java/org/eclipse/hono/client/impl/GatewayMappingCommandHandler.java
```diff
@@ -0,0 +1,97 @@
...
+                }
+            } else {
+                // failed to get last-via
+                LOG.error("error getting last-via for device {}", originalDeviceId, deviceIdFutureResult.cause());
+                TracingHelper.logError(originalCommandContext.getCurrentSpan(),
+                        new Exception("error getting last-via for device", deviceIdFutureResult.cause()));
+                originalCommandContext.release();
```
sophokles73: I think we also should re-issue some credit so that the application can send another command, don't you think?

client/src/main/java/org/eclipse/hono/client/CommandConsumerFactory.java
```diff
@@ -63,6 +64,7 @@ static CommandConsumerFactory create(final HonoConnection connection) {
      *         the cause of the failure</li>
      *         </ul>
      * @throws NullPointerException if any of tenant, device ID or command handler are {@code null}.
+     * @throws IllegalStateException if init() hasn't been called before.
```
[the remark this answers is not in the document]
calohmn: No, I've removed it.

client/src/main/java/org/eclipse/hono/client/CommandConsumerFactory.java
```diff
@@ -106,6 +108,7 @@ static CommandConsumerFactory create(final HonoConnection connection) {
      *         </ul>
      * @throws NullPointerException if tenant, device ID or command handler are {@code null}.
      * @throws IllegalArgumentException if the checkInterval is negative.
+     * @throws IllegalStateException if init() hasn't been called before.
```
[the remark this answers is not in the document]
calohmn: No, I've removed it.

client/src/main/java/org/eclipse/hono/client/DelegatedCommandSender.java
```diff
@@ -0,0 +1,24 @@
...
+
+package org.eclipse.hono.client;
+
+/**
+ * A sender to send command messages that are delegated to be processed by another command consumer.
+ * <p>
+ * This usually involves command messages first retrieved via a tenant-scoped consumer and then delegated back to the
+ * downstream peer so that they can be consumed by the device-specific consumer.
+ */
+public interface DelegatedCommandSender extends MessageSender {
```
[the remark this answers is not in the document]
calohmn: Good point, I've added `sendCommandMessage(Command, SpanContext)`. Better structure now.

client/src/main/java/org/eclipse/hono/client/GatewayMapper.java
```diff
@@ -0,0 +1,41 @@
...
+     * <ul>
+     * <li>the gateway device id</li>
+     * <li>{@code null} if the device is configured to be accessed via a gateway (i.e. one or more 'via' devices are
+     * set), but no 'last-via' device is set, meaning that no message has been sent yet for this device.</li>
+     * <li>the given device id if the device is not configured to be accessed via a gateway</li>
+     * </ul>
+     *
+     * @param tenantId The tenant identifier.
+     * @param deviceId The device identifier.
+     * @return Future containing the mapped gateway device id or {@code null}.
```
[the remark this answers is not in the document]
calohmn: I've adapted the javadoc.

client/src/main/java/org/eclipse/hono/client/impl/DelegatingCommandHandler.java
calohmn: I've adapted the javadoc.

client/src/main/java/org/eclipse/hono/client/impl/DelegatingCommandHandler.java
calohmn: I've adapted the javadoc.

client/src/main/java/org/eclipse/hono/client/impl/DelegatingCommandHandler.java
calohmn: Yes, added a method in the interface (see above).

client/src/main/java/org/eclipse/hono/client/impl/GatewayMappingCommandHandler.java
calohmn: Oh, thanks for spotting, missed the `return` statement there.
As for the credits: The `TenantScopedCommandConsumer` has its receiver link being created with the default (ie. non-empty) `prefetchSize`, so the credit handling should be automatic (in contrast to the `DeviceSpecificCommandConsumer` with `prefetchSize=0`).

client/src/main/java/org/eclipse/hono/client/impl/DelegatingCommandHandler.java
calohmn: I've removed creating an Exception there.

client/src/main/java/org/eclipse/hono/client/impl/GatewayMappingCommandHandler.java
calohmn: See above, credit-handling being automatic on that link.

client/src/main/java/org/eclipse/hono/client/impl/GatewayMappingCommandHandler.java
calohmn: See above, credit-handling being automatic on that link.

client/src/main/java/org/eclipse/hono/client/impl/DelegatedCommandSenderImpl.java
calohmn: Ok, I've changed that.

calohmn:
I've pushed the corresponding changes. 
I've also added a `TODO` in `DelegatedCommandSenderImpl.runSendAndWaitForOutcomeOnContext` concerning error-handling. To not further increase the size of this PR, I would fix this in a separate PR.

client/src/main/java/org/eclipse/hono/client/impl/DelegatingCommandHandler.java
```diff
@@ -0,0 +1,133 @@
...
+                commandContext.modify(modified.getDeliveryFailed(), modified.getUndeliverableHere(), 0);
+                break;
+            case Released:
+                commandContext.release();
+                break;
+            default:
+                LOG.warn("got unexpected delivery outcome; remote state: {}", delegatedMsgDelivery.getRemoteState());
+                TracingHelper.logError(commandContext.getCurrentSpan(), "got unexpected delivery outcome; remote state: "
+                        + delegatedMsgDelivery.getRemoteState());
+                commandContext.getCurrentSpan().finish();
```
sophokles73: My understanding is that at this point the application that sent the message is still expecting a disposition frame indicating the outcome of the transfer, isn't it? IMHO we need to invoke any of the `CommandContext`'s methods in order to finish the original message transfer here, i.e. even if we received an unexpected outcome, we should send a disposition with a supported outcome to the application, don't you think?.

It also feels strange that we finish the Span in this case but not in the other cases ...

client/src/main/java/org/eclipse/hono/client/impl/GatewayMappingCommandHandler.java
```diff
@@ -0,0 +1,98 @@
...
+                    // lastViaDeviceId is null - device hasn't connected yet
+                    LOG.error("last-via for device {} is not set", originalDeviceId);
+                    TracingHelper.logError(originalCommandContext.getCurrentSpan(), "last-via for device " + originalDeviceId + " is not set");
+                    originalCommandContext.release();
+                }
+            } else {
+                // failed to get last-via
+                LOG.error("error getting last-via for device {}", originalDeviceId, deviceIdFutureResult.cause());
+                TracingHelper.logError(originalCommandContext.getCurrentSpan(),
+                        new Exception("error getting last-via for device", deviceIdFutureResult.cause()));
```
sophokles73: `Exception`?

My understanding is that we run into this block also, if the GatewayMapper`s connection to the Device Registration service is lost, right?

client/src/main/java/org/eclipse/hono/client/impl/DelegatingCommandHandler.java
calohmn: Right, I've put a `commandContext.release()` in there. 
Besides the transaction related outcomes (`Declared` and `Transactional`), this block can only be reached with a `Received` delivery state, indicating partial transfer. I guess the latter case would be handled by the proton-j layer below and not be exposed here.

> It also feels strange that we finish the Span in this case but not in the other cases ...

In the other cases the span is finished by the `CommandContext#accept/reject/etc.` methods.

client/src/main/java/org/eclipse/hono/client/impl/GatewayMappingCommandHandler.java
calohmn: Yes, that would be a common cause for an error here. I've added a comment indicating this.

sophokles73:
@dejanb do you also want to take a look?

Issue #1055, opened 2019-02-20
Add Jaeger tracing component in Docker Swarm / Kubernetes deployment


calohmn:
In order to have a Hono setup for testing that includes a tracing component, it would be useful to include the [Jaeger All-in-one image](https://www.jaegertracing.io/docs/1.10/getting-started/#all-in-one) in the Docker Swarm and Kubernetes/Openshift deployment scripts.

Usage of the Jaeger component should be made optional there.

sophokles73:
@calohmn can you close this issue?

PR #1148, opened 2019-04-08
[#1055] Add option to deploy Jaeger in docker swarm

calohmn:
This is for #1055.


calohmn:
Sure, I've added #1175 for that.

PR #1077, opened 2019-03-07
Device Registry Implementation backed with infinispan caches

jbtrystram:
This is an implementation of the device registry APIs based on Infinispan caches. It currently used embedded in-momory caches, and it should be able to be distributed as the number of instances of the DR are increased.

Signed-off-by: Jean-Baptiste Trystram <jbtrystram@redhat.com>

PR #1151, opened 2019-04-08
Split operations into mandatory and optional to implement.

sophokles73:
The API specifications now contain distinct sections for the mandatory
vs optional to implement operations.

I wonder if we shouldn't remove the *optional* operations from the API spec altogether. I do not think that anybody wants to implement them based on AMQP anyway. With #1130 under way, they probably do not add much value at all. WDYT?

### DIFF
--- a/site/content/api/Credentials-API.md
+++ b/site/content/api/Credentials-API.md
@@ -3,7 +3,7 @@ title = "Credentials API"
 weight = 430
 +++
 
-The *Credentials API* is used by *Protocol Adapters* to retrieve credentials used to authenticate *Devices* connecting to the adapter. In particular, the API supports the storage, look up and deletion of *shared secrets* which are often used by IoT devices by means of *username/password* based authentication schemes.
+Eclipse Hono&trade;'s *Credentials API* is used by *Protocol Adapters* to retrieve credentials used to authenticate *Devices* connecting to the adapter. In particular, the API supports the storage, look up and deletion of *shared secrets* which are often used by IoT devices by means of *username/password* based authentication schemes.
 <!--more-->
 
 Credentials are of a certain *type* which indicates which authentication mechanism the credentials can be used with. Each set of credentials also contains an *authentication identity* which is the identity claimed by the device during authentication. This authentication identity is usually different from the *device-id* the device has been registered with using Hono's [Device Registration API]({{< relref "api/Device-Registration-API.md" >}}). Multiple sets of credentials (including arbitrary *authentication identities*) can be registered for each *device-id*.
@@ -12,7 +12,7 @@ Note, however, that in real world applications the device credentials will proba
 
 The Credentials API is defined by means of AMQP 1.0 message exchanges, i.e. a client needs to connect to Hono using an AMQP 1.0 client in order to invoke operations of the API as described in the following sections.
 
-## Preconditions
+# Preconditions
 
 The preconditions for invoking any of the Credential API's operations are as follows:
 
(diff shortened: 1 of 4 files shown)


calohmn:
Note: With the changes in #930/#1145, the `Get Registration Information` operation of the `Device Registration API` will be become mandatory as well. 
After this PR and #1145 have been merged, I can create a separate PR with an update to the API spec and also with a change of the `RegistrationService` interface, moving the `getDevice` method from `CompleteRegistrationService` to `RegistrationService`.

sophokles73:
> the Get Registration Information operation of the Device Registration API will be become mandatory as well.

One could actually ask, if the *get Registration Information* operation will replace the *assert Registration Information* operation, given that we do not need/verify the registration token anymore anyway ...

calohmn:
> One could actually argue, that the get Registration Information operation should replace the assert Registration Information operation, given that we do not need/verify the registration token anymore anyway ...

I think the `assert Registration Information` operation is still needed for the check whether a device or gateway is enabled or the gateway is authorized, not requiring the protocol adapters to do these checks.

> the Get Registration Information operation of the Device Registration API will be become mandatory as well.

@jbtrystram, @ctron, @dejanb That info is also relevant for the new Device Registry implementation (#1077). Do you see an issue with that, also regarding moving the `getDevice` method from `CompleteRegistrationService` to `RegistrationService`?

jbtrystram:
@calohmn I don't see any problem with this change, as `CompleteRegistrationService` extends `RegistrationService` anyway :)

site/content/api/Credentials-API.md
```diff
@@ -3,7 +3,7 @@ title = "Credentials API"
 weight = 430
 +++
 
-The *Credentials API* is used by *Protocol Adapters* to retrieve credentials used to authenticate *Devices* connecting to the adapter. In particular, the API supports the storage, look up and deletion of *shared secrets* which are often used by IoT devices by means of *username/password* based authentication schemes.
+Eclipse Hono&trade;'s *Credentials API* is used by *Protocol Adapters* to retrieve credentials used to authenticate *Devices* connecting to the adapter. In particular, the API supports the storage, look up and deletion of *shared secrets* which are often used by IoT devices by means of *username/password* based authentication schemes.
```
ctron: Although it is not wrong, simply using Hono is sufficient. I also would prefer Eclipse Hono. But repeating the ™ is not necessary, as the document will be rendered by the Hono theme, which already shows that at the first and most prominent place. Also see: https://www.eclipse.org/legal/logo_guidelines.php#Proper_Notice_and_Attribution

sophokles73: I see your point. We are currently using `Eclipse Hono&trade;` on every page when we first mention Hono. FMPOV we should instead modify the theme to render Eclipse Hono with the TM symbol in the orange header of the page and then simply refer to Hono as just *Hono* on all content pages. WDYT?

sophokles73: I will do this in a separate commit/PR.

sophokles73:
@ctron @dejanb what do you think about removing the optional operations completely as indicated in the PR description?

ctron:
> @ctron @dejanb what do you think about removing the optional operations completely as indicated in the PR description?

I tend to agree here. I am not sure they bring in much value at the moment. On the other side, having them in 1.0 would mean that we would need to maintain them for a while. So maybe a soft +1 for removing them.

sophokles73:
> So maybe a soft +1 for removing them.

We could deprecate them in 1.0-M2 and then remove them in 1.0.

ctron:
> >  So maybe a soft +1 for removing them.
>
> We could deprecate them in 1.0-M2 and then remove them in 1.0.

Yes, so my assumption was what they will be gone in 1.0.

PR #1161, opened 2019-04-10
Fix checking logic of connection limit.

kaniyan:
AMQP and MQTT Protocol adapters check the _connection limit_, if configured, before accepting any new connections. The issue is that they allow one more connection above the configured limit. i.e if the _limit_ `max-connection` is set as 2, then the adapters allow 3 connections for that tenant. This PR fixes this issue.

### DIFF
--- a/adapters/mqtt-vertx-base/src/test/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapterTest.java
+++ b/adapters/mqtt-vertx-base/src/test/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapterTest.java
@@ -180,7 +180,7 @@ public class AbstractVertxBasedMqttProtocolAdapterTest {
 
         authHandler = mock(AuthHandler.class);
         resourceLimitChecks = mock(ResourceLimitChecks.class);
-        when(resourceLimitChecks.isConnectionLimitExceeded(any(TenantObject.class))).thenReturn(Future.succeededFuture(Boolean.FALSE));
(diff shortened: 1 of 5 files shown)


sophokles73:
I agree with the assessment. However, based on the naming of the method (*isConnectionLimitExceeded*) the current implementation is actually correct, isn't it? So, maybe we should change the name (and semantics) of the method accordingly, e.g. to *isConnectionLimitReached* which would then always allow for an additional connection to be established if it returns `false`, right?

kaniyan:
> I agree with the assessment. However, based on the naming of the method (_isConnectionLimitExceeded_) the current implementation is actually correct, isn't it? So, maybe we should change the name (and semantics) of the method accordingly, e.g. to _isConnectionLimitReached_ which would then always allow for an additional connection to be established if it returns `false`, right?

@sophokles73  _isConnectionLimitReached_ suits well the intended purpose. If it returns `true` connection requests are rejected, else accepted.

kaniyan:
> > I agree with the assessment. However, based on the naming of the method (_isConnectionLimitExceeded_) the current implementation is actually correct, isn't it? So, maybe we should change the name (and semantics) of the method accordingly, e.g. to _isConnectionLimitReached_ which would then always allow for an additional connection to be established if it returns `false`, right?
> 
> @sophokles73 _isConnectionLimitReached_ suits well the intended purpose. If it returns `true` connection requests are rejected, else accepted.

@sophokles73 Done.

sophokles73:
@ctron are you going to merge this PR?

sophokles73:
@kaniyan can you please rebase your PR so that the Travis build has a chance of succeeding?

PR #1162, opened 2019-04-10
[#929 Part-2] Extend ResourceLimitChecks API to verify the messages limit.

kaniyan:
`ResourceLimitChecks` API is now extended with the method `isMessageLimitExceeded(...)` to check the messages limit. This method is intended to be used by the _protocol adapters_ to verify the messages limit before accepting any telemetry or event messages. 

The _protocol adapters_ using this api to verify the messages limit will be part of the next PR.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
@@ -30,6 +30,7 @@ import org.eclipse.hono.config.ClientConfigProperties;
 import org.eclipse.hono.config.VertxProperties;
 import org.eclipse.hono.service.cache.SpringCacheProvider;
 import org.eclipse.hono.service.plan.PrometheusBasedResourceLimitChecks;
+import org.eclipse.hono.service.plan.PrometheusBasedResourceLimitChecksConfig;
 import org.eclipse.hono.service.plan.ResourceLimitChecks;
 import org.eclipse.hono.util.CommandConstants;
 import org.eclipse.hono.util.Constants;
@@ -384,7 +385,7 @@ public abstract class AbstractAdapterConfig {
 
     /**
      * Exposes configuration properties for vert.x.
-     * 
+     *
      * @return The properties.
      */
     @ConfigurationProperties("hono.vertx")
@@ -400,8 +401,17 @@ public abstract class AbstractAdapterConfig {
      * @return A new cache provider or {@code null} if no cache should be used.
      */
     private static CacheProvider newGuavaCache(final RequestResponseClientConfigProperties config) {
-        final int minCacheSize = config.getResponseCacheMinSize();
-        final long maxCacheSize = config.getResponseCacheMaxSize();
+        return newGuavaCache(config.getResponseCacheMinSize(), config.getResponseCacheMaxSize());
+    }
+
+    /**
+     * Create a new cache provider based on Guava and Spring Cache.
+     *
+     * @param minCacheSize The minimum size of the cache.
+     * @param maxCacheSize the maximum size of the cache.
+     * @return A new cache provider or {@code null} if no cache should be used.
+     */
+    private static CacheProvider newGuavaCache(final int minCacheSize, final long maxCacheSize) {
 
         if (maxCacheSize <= 0) {
             return null;
@@ -441,14 +451,27 @@ public abstract class AbstractAdapterConfig {
     }
 
     /**
+     * Exposes configuration properties for ResourceLimitChecks as a Spring bean.
+     *
+     * @return The properties.
+     */
+    @Bean
+    @ConfigurationProperties(prefix = "hono.plan.prometheusBased")
+    @ConditionalOnProperty(name = "hono.plan.prometheusBased.host")
+    public PrometheusBasedResourceLimitChecksConfig resourceLimitChecksConfig() {
+        return new PrometheusBasedResourceLimitChecksConfig();
+    }
+
+    /**
      * Creates a new instance of {@link ResourceLimitChecks} based on prometheus metrics data.
      * 
      * @return A ResourceLimitChecks instance.
      */
     @Bean
-    @ConfigurationProperties(prefix = "hono.plan.prometheusBased")
     @ConditionalOnProperty(name = "hono.plan.prometheusBased.host")
     public ResourceLimitChecks resourceLimitChecks() {
-        return new PrometheusBasedResourceLimitChecks(WebClient.create(vertx()));
+        final PrometheusBasedResourceLimitChecksConfig config = resourceLimitChecksConfig();
+        return new PrometheusBasedResourceLimitChecks(WebClient.create(vertx()), config,
+                newGuavaCache(config.getCacheMinSize(), config.getCacheMaxSize()));
     }
 }
--- a/service-base/src/main/java/org/eclipse/hono/service/plan/NoopResourceLimitChecks.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/plan/NoopResourceLimitChecks.java
@@ -24,4 +24,10 @@ public class NoopResourceLimitChecks implements ResourceLimitChecks {
     public Future<Boolean> isConnectionLimitReached(final TenantObject tenantObject) {
         return Future.succeededFuture(Boolean.FALSE);
     }
+
+    @Override
+    public Future<Boolean> isMessageLimitReached(final TenantObject tenantObject,
+            final long payloadSize) {
+        return Future.succeededFuture(Boolean.FALSE);
+    }
 }
--- a/service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
@@ -12,14 +12,20 @@
  *******************************************************************************/
 package org.eclipse.hono.service.plan;
 
+import static java.time.temporal.ChronoUnit.DAYS;
+
+import java.time.Duration;
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.util.Objects;
 import java.util.Optional;
 
+import org.eclipse.hono.cache.CacheProvider;
+import org.eclipse.hono.cache.ExpiringValueCache;
+import org.eclipse.hono.client.ServiceInvocationException;
+import org.eclipse.hono.service.metric.MetricsTags;
 import org.eclipse.hono.service.metric.MicrometerBasedMetrics;
-import org.eclipse.hono.util.PortConfigurationHelper;
 import org.eclipse.hono.util.TenantObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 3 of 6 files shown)


sophokles73:

kaniyan:
P.S.
The message size of the incoming message should also be taken into consideration before accepting or rejecting it. So I could well think of the below method for that purpose. 
`Future<Boolean> isMessageLimitExceeded(final TenantObject tenant, final long messageSize)` 

As in `ResourceLimitChecks.isConnectionLimitReached(tenanat)`, it is not sufficient just to check if the message limit has been already reached or not. In this case, the sum of the data volume already used up for that month plus the size of the incoming message  is being compared with the configured limit. If this sum is less than or equal to the configured limit, `false` is returned, if not then `true`.

sophokles73:
@kaniyan can you rebase this PR so that I can start reviewing?

kaniyan:
> @kaniyan can you rebase this PR so that I can start reviewing?

@sophokles73 Currently I am working on it. Once it is ready I will update you.

kaniyan:
@sophokles73 I have pushed the changes. Pls have a look :)

service-base/src/main/java/org/eclipse/hono/service/plan/NoopResourceLimitChecks.java
```diff
@@ -24,4 +24,10 @@
     public Future<Boolean> isConnectionLimitReached(final TenantObject tenantObject) {
         return Future.succeededFuture(Boolean.FALSE);
     }
+
+    @Override
+    public Future<Boolean> isMessageLimitExceeded(final TenantObject tenantObject,
```
sophokles73: Can we rename this to `isDataVolumeLimitReached` and align its semantics with those of *isConnectionLimitReached*?

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -125,6 +131,52 @@ public PrometheusBasedResourceLimitChecks(final WebClient webClient) {
...
+            final long dataUsagePeriod = calculateDataUsagePeriod(effectiveSince, periodInDays);
+            if (dataUsagePeriod <= 0) {
+                return Future.succeededFuture(Boolean.FALSE);
+            }
+            final String queryParams = String.format("floor(sum(increase(%s{status=\"%s\", tenant=\"%s\"} [%sd])))",
+                    MESSAGES_PAYLOAD_SIZE_METRIC_NAME,
+                    MetricsTags.ProcessingOutcome.FORWARDED.asTag().getValue(),
+                    tenant.getTenantId(),
+                    dataUsagePeriod);
+            return executeQuery(queryParams)
```
sophokles73: we cannot execute this query for every message published by a device, because this is waaayyy too expensive to do. Instead, we could maintain a local counter which gets updated with the current message's payload size and which is periodically updated based on the value retrieved from the metrics back end.

service-base/src/main/java/org/eclipse/hono/service/plan/NoopResourceLimitChecks.java
kaniyan: The motive behind keeping _message_ instead of _dataVolume_ as part of the method name is to provide flexibility to the users to choose different attributes to perform the check. For example, the message limit can be based on the number of messages or data volume or combination of both.

Case 1: `bytesAlreadyConsumed + sizeOfIncomingMessage < max-bytes`: returns `false`. The limit is not reached neither exceeded.
Case 2: `bytesAlreadyConsumed + sizeOfIncomingMessage == max-bytes`: still returns `false`. The limit is reached but not exceeded. Thats why I had to name it as `isMessageLimitExceeded` instead of `isMessageLimitReached`.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
kaniyan: It is already intended for the next PR while equipping the _protocol adapters_ with this new method.
 I thought of a caching mechanism similar to the tenant and registration clients. Then the user can configure the cache timeout according to his/her needs. The method `isMessageLimitReached` is not aware, if the message will be later successfully accepted or not.  Where as protocol adapters know and the delta part can be updated by the protocol adapters to the cache. The query result can be updated by the `PrometheusBasedResourceLimitChecks`. WDYT?

service-base/src/main/java/org/eclipse/hono/service/plan/NoopResourceLimitChecks.java
sophokles73: > Case 2: bytesAlreadyConsumed + sizeOfIncomingMessage == max-bytes: still returns false. The limit is reached but not exceeded. Thats why I had to name it as isMessageLimitExceeded instead of isMessageLimitReached.

I don't get it. *isConnectionLimitReached* returns false, if another connection can be established, right? So, analogously, *isMessageLimitReached* should return false if another message can be accepted, right? It is not really important whether `bytesAlreadyConsumed` + `sizeOfIncomingMessage` is less than or equal to the limit as we do not have the exact value of `bytesAlreadyConsumed` for all adapters together anyway. However, the semantics of the two methods would be the same, i.e. if they return false, we accept the current connection attempt/message.

kaniyan: I see the point. I will rename it is `isMessageLimitReached`.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
sophokles73: IMHO a simple timer running the query towards the metrics backend periodically would do ...

kaniyan: Given a scenario: A device `D1` belonging to a tenant `Tenant1` sends an _event_ message. The metrics back-end is queried for the _consumed bytes_ for the tenant `Tenant1`. And a timer is set periodically to retrieve this data for `Tenant1`. Assume the next message from this device `D1` is 3 days later or even later. This timer periodically queries the metrics back-end, which IMHO is not needed. The number of timers will then be equal to the number of _tenants_ of the connected devices that have sent at-least one message.

IMHO a data structure to store the _consumed bytes_ grouped by tenant id is needed and should be configurable to say how long the old data can be used. Hence I think that caching mechanism may suit the purpose.

sophokles73: I agree, that would be even better. All I am saying is that the code in this PR should not retrieve the value for each request ...

kaniyan: I agree and it is intended as part of the next PR.

sophokles73: I understood that this is your plan. However, I will not approve this PR with the query being executed on every message. So you can either implement the caching in this PR already or (much easier) refactor the code executing the query into a separate method which is never invoked (and then always return `Boolean.FALSE`).

kaniyan: IMHO currently no protocol adapters are equipped with the method `isMessageLimitReached(...)` to check the message limit. On the contrary, this query is currently never being executed. If you think its still better to have caching mechanism or separate method for the query which is never invoked as part of this PR, I will add to this.

sophokles73: please do

kaniyan:
@sophokles73 I have implemented the caching mechanism as part of this PR.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -126,47 +141,60 @@ public PrometheusBasedResourceLimitChecks(final WebClient webClient) {
...
+
+            final String queryParams = String.format("floor(sum(increase(%s{status=\"%s\", tenant=\"%s\"} [%sd])))",
+                    MESSAGES_PAYLOAD_SIZE_METRIC_NAME,
+                    MetricsTags.ProcessingOutcome.FORWARDED.asTag().getValue(),
+                    tenant.getTenantId(),
+                    dataUsagePeriod);
+
+            final ExpiringValueCache<Object, Object> cache = Optional
+                    .ofNullable(cacheProvider)
+                    .map(provider -> provider.getCache(tenant.getTenantId()))
```
sophokles73: If I am not mistaken, this means that we create a separate cache for each tenant, which will then contain a *single* entry with key `BYTES_CONSUMED_CACHE_ID` each, right?
FMPOV we should need a single cache instance only (e.g. with name `bytes_consumed`) which we can use for *all* tenants by using the tenant identifier as the key, don't you think?

kaniyan: I created this having some more resource limits in the future such as _connection-time_ in the mind. I think it makes sense to create once cache instance (for `bytes_consumed`) and have `tenant-id` as the key.

sophokles73: I see, then maybe we name the cache instance `limits` and use e.g. `${tenant-id}_bytes_consumed` as the key? We can then later extend with e.g. `${tenant-id}_connection_time`. WDYT?

kaniyan: sounds good.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -126,47 +141,60 @@ public PrometheusBasedResourceLimitChecks(final WebClient webClient) {
...
         } else {
-            throw new IllegalArgumentException("invalid port number");
+            final long dataUsagePeriod = calculateDataUsagePeriod(effectiveSince, periodInDays);
+            if (dataUsagePeriod <= 0) {
+                return Future.succeededFuture(Boolean.FALSE);
+            }
+
+            final String queryParams = String.format("floor(sum(increase(%s{status=\"%s\", tenant=\"%s\"} [%sd])))",
+                    MESSAGES_PAYLOAD_SIZE_METRIC_NAME,
+                    MetricsTags.ProcessingOutcome.FORWARDED.asTag().getValue(),
```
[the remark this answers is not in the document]
kaniyan: Now the _payload size_ of both the  `FORWARDED` and the `UNPROCESSABLE` messages are taken into account.

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
```diff
@@ -440,15 +450,39 @@ public HealthCheckServer healthCheckServer() {
...
+    }
+
+    /**
+     * Exposes the provider for caches as a Spring bean.
+     *
+     * @return The provider instance.
+     */
+    @Bean
+    @ConditionalOnProperty(name = "hono.plan.prometheusBased.host")
+    public CacheProvider resourceLimitChecksCacheProvider() {
```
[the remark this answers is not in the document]
kaniyan: It has been moved to the `resourceLimitChecks` method.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -126,47 +143,58 @@ public PrometheusBasedResourceLimitChecks(final WebClient webClient) {
...
+                return Future.succeededFuture(Boolean.FALSE);
+            }
+
+            final String queryParams = String.format("floor(sum(increase(%s{status=~\"%s|%s\", tenant=\"%s\"} [%sd])))",
+                    MESSAGES_PAYLOAD_SIZE_METRIC_NAME,
+                    MetricsTags.ProcessingOutcome.FORWARDED.asTag().getValue(),
+                    MetricsTags.ProcessingOutcome.UNPROCESSABLE.asTag().getValue(),
+                    tenant.getTenantId(),
+                    dataUsagePeriod);
+            final String cacheId = String.format("%s_bytes_consumed", tenant.getTenantId());
```
sophokles73: this is the key, isn't it?

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecksConfig.java
```diff
@@ -0,0 +1,178 @@
...
+package org.eclipse.hono.service.plan;
+
+import java.util.Objects;
+
+import org.eclipse.hono.util.PortConfigurationHelper;
+
+/**
+ * The configuration properties required for the PrometheusBasedResourceLimitChecks.
+ */
+public class PrometheusBasedResourceLimitChecksConfig {
```
sophokles73: if this class is not supposed to be extended, make it final, please. Otherwise, please make the methods final ...

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
kaniyan: yes. Do you suggest renaming it to something like `cacheKey` instead of `cacheId`?

sophokles73: Or just `key` ...

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecksConfig.java
kaniyan: Now it is final.

PR #1163, opened 2019-04-11
Fix: Do not start protocol adapter if not enough memory.

b-abel:
Throw a exception if the connection limit is auto-configured to 0. 
Currently the protocol adapter only prints a warning to the logs, but then completes the startup. So, possibly nobody would notice that it would not accept any connections.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/limiting/DefaultConnectionLimitManager.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/limiting/DefaultConnectionLimitManager.java
@@ -44,6 +44,8 @@ public class DefaultConnectionLimitManager implements ConnectionLimitManager {
      * @param currentConnections The supplier to invoke for getting the current number of connections.
      * @param config The configuration of the adapter or {@code null}.
      * @throws NullPointerException if strategy or currentConnections are {@code null}.
+     * @throws ConnectionLimitAutoConfigException if no connection limit is configured and the auto-configuration
+     *             calculates a limit of 0.
      */
     public DefaultConnectionLimitManager(final ConnectionLimitStrategy strategy, final Supplier<Integer> currentConnections,
             final ProtocolAdapterProperties config) {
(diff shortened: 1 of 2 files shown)


sophokles73:
I think it is reasonable to indicate this problem to the user more explicitly, as proposed in this PR.
I also wonder if the current strategy we follow, i.e. shutting down the service and exiting the JVM, if a problem occurs, is very helpful. Most container orchestration platforms will immediately try to restart the container which will most probably fail again and then over and over again. This makes it particularly hard for ops people to get to the cause of the problem because the log of the exited container might already be gone as well. Maybe it would therefore be better to start the container successfully, but never let it reach the *ready* state instead. WDYT? @ctron @dejanb

sophokles73:

dejanb:
@sophokles73 I agree. There's no point in restarting the container if the error will persist. It might be better to indicate this problem in ready/healthy status

b-abel:
Rebased.

service-base/src/main/java/org/eclipse/hono/service/limiting/DefaultConnectionLimitManager.java
```diff
@@ -61,6 +61,10 @@ private int autoconfigureConnectionLimit() {
 
         final int recommendedLimit = strategy.getRecommendedLimit();
 
+        if (recommendedLimit == 0) {
+            throw new RuntimeException("Not enough memory to handle connections. To override this check, configure a connection limit.");
```
sophokles73: Please throw a more specific exception and also document it in the constructor's JavaDoc.

b-abel: Like this?

sophokles73: Sorry, I didn't make myself clear enough. I would have been completely content with e.g. throwing an `IllegalStateException` here. However, using a newly designed exception works as well, of course :-)

PR #1165, opened 2019-04-11
Create tenant CA for DEFAULT_TENANT.

b-abel:
The example device cert is no longer issued by Hono's "server" CA.
Instead the script creates a separate CA for the tenant and issues
the client certificate from it.


sophokles73:

b-abel:
Done.

sophokles73:
can we also use this for the integration tests where we check authentication of a device using a client certificate?

b-abel:
@sophokles73 I don't see why it would not be possible to load the cert in the test. But currently the tests use random tenant and device IDs, which needs to be changed then.

sophokles73:
yeah, didn't think of that. We can maybe later change that ...

PR #1167, opened 2019-04-12
Configure nexus plugin in a dedicated profile

ctron:
I would like to put the nexus plugin into a dedicated profile. I know this, most likely has an impact in the Eclipse CI infrastructure as well.

The reason behind that is that the plugin completely takes control of the deployment. And causes some problems in our internal build. And currently there is no other way to remove this plugin, than to manually patch the main POM file.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -454,31 +454,6 @@
         </executions>
       </plugin>
       <plugin>
-        <!-- 
-          Use the Nexus Staging plugin as a full replacement for the standard
-          Maven Deploy plugin.
-          See https://github.com/sonatype/nexus-maven-plugins/tree/master/staging/maven-plugin
-          why this makes sense :-)
-          We can control whether we want to deploy to the Eclipse repo or Maven Central
-          by a combination of the version being a SNAPHOT or release version and property
-          skipStaging=true/false.
-          In any case we can take advantage of the plugin's "deferred deploy" feature which
-          makes sure that all artifacts of a multi-module project are deployed as a whole
-          at the end of the build process instead of deploying each module's artifacts
-          individually as part of building the module.
-         -->
-        <groupId>org.sonatype.plugins</groupId>
-        <artifactId>nexus-staging-maven-plugin</artifactId>
-        <version>1.6.8</version>
-        <extensions>true</extensions>
-        <configuration>
-          <serverId>ossrh</serverId>
-          <nexusUrl>https://oss.sonatype.org/</nexusUrl>
-          <autoReleaseAfterClose>false</autoReleaseAfterClose>
-        </configuration>
-      </plugin>
-
-      <plugin>
         <groupId>org.jacoco</groupId>
         <artifactId>jacoco-maven-plugin</artifactId>
         <version>${jacoco.version}</version>
@@ -515,6 +490,43 @@
   </build>
 
   <profiles>
+    <profile>
+      <id>nexus</id>
+      <activation>
+        <property>
+          <name>disableNexus</name>
+          <value>!true</value>
+        </property>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <!-- 
+              Use the Nexus Staging plugin as a full replacement for the standard
+              Maven Deploy plugin.
+              See https://github.com/sonatype/nexus-maven-plugins/tree/master/staging/maven-plugin
+              why this makes sense :-)
+              We can control whether we want to deploy to the Eclipse repo or Maven Central
+              by a combination of the version being a SNAPHOT or release version and property
+              skipStaging=true/false.
+              In any case we can take advantage of the plugin's "deferred deploy" feature which
+              makes sure that all artifacts of a multi-module project are deployed as a whole
+              at the end of the build process instead of deploying each module's artifacts
+              individually as part of building the module.
+             -->
+            <groupId>org.sonatype.plugins</groupId>
+            <artifactId>nexus-staging-maven-plugin</artifactId>
+            <version>1.6.8</version>
+            <extensions>true</extensions>
+            <configuration>
+              <serverId>ossrh</serverId>
+              <nexusUrl>https://oss.sonatype.org/</nexusUrl>
+              <autoReleaseAfterClose>false</autoReleaseAfterClose>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
     <profile>
       <!-- 
         This profile is activated by Eclipse IDE automatically.
--- a/pom.xml
+++ b/pom.xml
@@ -244,6 +244,10 @@
           </configuration>
         </plugin>
         <plugin>
+          <artifactId>maven-deploy-plugin</artifactId>
+          <version>2.7</version>
+        </plugin>
+        <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-resources-plugin</artifactId>
           <version>3.0.2</version>


sophokles73:
> The reason behind that is that the plugin completely takes control of the deployment. And causes some problems in our internal build. 

Can you elaborate what the problem is? Maybe we find another (simple) solution?

ctron:
The problem is that is the goal `mvn deploy` has to be run, however it must deploy to an alternate repository, with the credentials provided through an external settings file, and the repository has a specific ID for referencing this.

This works with the standard maven deploy plugin, but with the nexus plugin I found no working way to override the deployment repository and using the internal ID.

The ID in the Hono build is set to `<serverId>ossrh</serverId>`, and there seems to be no way to override that either.

sophokles73:
We are doing the same thing when deploying the release artifacts to the Eclipse Repository instead of Maven Central.
I would assume that you can use `-DskipLocalStaging=true` to use the standard `deploy` behavior, i.e. no local nor remote staging, and then use the [deploy plugin's `altDeploymentRepository` standard property](https://maven.apache.org/plugins/maven-deploy-plugin/deploy-mojo.html) to set the repository ID and URL.

ctron:
I assumed that as well, but still it failed:

I used `-DskipLocalStaging=true  -DaltReleaseDeploymentRepository=our-mvn::${OurURL}` with the result of:

```
Caused by: org.eclipse.aether.transfer.ArtifactTransferException: Could not transfer artifact org.eclipse.hono:hono-parent:pom:0.9.0.temporary-redhat-00001 from/to repo.eclipse.org (https://repo.eclipse.org/content/repositories/hono-releases/): PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
```

sophokles73:
The docs for the `altReleaseDeploymentRepository` property indicate a format of `id::layout::url`.
Are you sure that you can omit the layout part like you did?

ctron:
Yes, and in the docs of the Maven deploy plugin is says: "Note: Since 3.0.0 the layout part has been removed."

And that is the version used by our build system.

I think the problem is in the nexus plugin, the documentation says:

> The deployment repository is "sourced" from `pom.xml <distributionManagement>`.

See: https://help.sonatype.com/repomanager2/staging-releases/configuring-your-project-for-deployment#ConfiguringYourProjectforDeployment-DeploymentwiththeNexusStagingMavenPlugin (near "skipLocalStaging").

And that is the problem, because in our case this we would like to use the parameter you suggested, but is simply seems not be used by the nexus plugin. And the default `<distributionManagement>` section doesn't work either.

sophokles73:
I have taken a look at the staging plugin's source code [1]. The *getDeploymentRepository* method explicitly mentions the usage of `-DaltDeploymentRepository=id::layout::url` in order to set the deploy target.

Since we are also not using the maven-deploy-plugin, it might be worth a shot to actually use that parameter with the original format, e.g. `our-repo::default::http://our-repo.com`

WDYT?

[1] https://github.com/sonatype/nexus-maven-plugins/blob/master/staging/maven-plugin/src/main/java/org/sonatype/nexus/maven/staging/deploy/strategy/AbstractDeployStrategy.java

ctron:
So I just re-tried that, the settings are:

    -DskipLocalStaging=true -DaltDeploymentRepository=own-mvn::default::${OwnUrl}

The result is:

    Caused by: org.eclipse.aether.transfer.ArtifactTransferException: Could not transfer artifact org.eclipse.hono:hono-parent:pom:0.9.0.temporary-redhat-00003 from/to repo.eclipse.org (https://repo.eclipse.org/content/repositories/hono-releases/): PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target

ctron:
Digging a bit into the code you linked, it looks to me as if this functionality is actually broken. The maven default project implementation only takes the "distribution" element to fetch the target. While the deployment plugin does the "alt" handling itself:

https://github.com/apache/maven-deploy-plugin/blob/master/src/main/java/org/apache/maven/plugins/deploy/DeployMojo.java#L156

So I guess this message in the nexus plugin is just not correct anymore.

So still, the issue remains.

sophokles73:
So, I think that the right thing to do would be to create an issue in the [nexus staging plugin's bug tracker](https://issues.sonatype.org/browse/NEXUS-18170?jql=project%20%3D%20NEXUS%20AND%20component%20%3D%20%22Build%20Tooling%22)
and maybe also create a PR for fixing this. Moving the staging plugin *out of the way* may fix this issue short term, but it feels rather strange to do so, given that Hono itself doesn't even face this issue.
WDYT?

ctron:
Well I think that sticking to standard Maven workflow isn't a bad thing to do. And enabling the special handling of the Nexus plugin in case of a special context (releasing at Eclipse) makes sense. But that step should be optional (although it may be enabled by default).

However forcing this on everyone seems a bit weird to me.

Still I created the issue: https://issues.sonatype.org/browse/NEXUS-19789

sophokles73:
> However forcing this on everyone seems a bit weird to me.

Well, the nexus staging plugin actually advertises that it can be used as a complete replacement of the standard maven deploy plugin. So, based on that claim, I wouldn't call it *forcing it on everyone*, i.e. if the plugin would keep up with its promises, you wouldn't have this issue, would you.

Based on what we have found out, I agree that we should create a profile for using the nexus plugin. However, your PR lacks the usage of this profile in the deploy build scripts in the `jenkins` folder ...

ctron:
If everything would work as advertised :grin: 

I did make a change to the PR, so that by default everything stays as is. However you can disable the profile by passing `-DdisableNexus=true` to the maven build, which disables the profile. That should be sufficient for our use case. I will try to run a test with that and come back with the results.

ctron:
So this works, however I had to make a second change.

Current Hono doesn't manage the version of the plugin `maven-deploy-plugin`. This results in multipe (different versions) of the plugin being enabled during the run. Unfortunately one is 2.7, and the other 3.0 (when using Maven 3.5.x). Which results in the problem that the `DaltDeploymentRepository` is expected to be present in two different patterns. So either the 2.7 version fails, or the 3.0 one. See:

~~~
➜  hono git:(master) mvn help:effective-pom   | grep maven-deploy-plugin -A1
[INFO]   ... total of 26 executions of maven-deploy-plugin replaced with nexus-staging-maven-plugin
[INFO] ------------------------------------------------------------------------
--
          <artifactId>maven-deploy-plugin</artifactId>
          <version>2.7</version>
--
          <artifactId>maven-deploy-plugin</artifactId>
          <version>2.7</version>
--
          <artifactId>maven-deploy-plugin</artifactId>
          <version>2.7</version>
--
          <artifactId>maven-deploy-plugin</artifactId>
          <version>2.7</version>
--
          <artifactId>maven-deploy-plugin</artifactId>
        </plugin>
--
          <artifactId>maven-deploy-plugin</artifactId>
        </plugin>
--
          <artifactId>maven-deploy-plugin</artifactId>
          <version>2.7</version>
--
          <artifactId>maven-deploy-plugin</artifactId>
          <version>2.7</version>
…
~~~

The solution for that is simple, manage the plugin version. I used 2.7, as this seems to be version evaluated to be most modules in the Hono build.

With those both changes, the default behavior should be the same, so no Jenkins changes are requires. But there is also a way to work around the nexus staging bug for us.

ctron:
Following up as the issue I opened with Nexus was updated: https://issues.sonatype.org/browse/NEXUS-19789

To me it seems they acknowledged the issue, but won't fix it. At least, for a while.

PR #1169, opened 2019-04-12
Fix up travis build:

ctron:
* Remove sudo flag: https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration
* Enable Java 12 build
* Drop unnecessary before_install

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,13 @@
 dist: xenial
-sudo: required
+
 language: java
 
 jdk:
   - openjdk11
+  - openjdk12
 
 script:
   - mvn verify -B -e -Dmsg.count=400 -Dtest.env=true -DcreateJavadoc=true -Ddocker.host=unix:///var/run/docker.sock -Dservice.startup.timeout=600000 -Dmax.event-loop.execute-time=20000 -Pbuild-docker-image,run-tests -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
-before_install:
 
 env:
   global:


.travis.yml
```diff
@@ -1,13 +1,13 @@
 dist: xenial
-sudo: required
+
 language: java
 
 jdk:
   - openjdk11
+  - openjdk12
```
sophokles73: not sure about this, the build already takes around 11 minutes and running another JDK will double that time, won't it?

ctron: If you build this in the "eclipse" organization, then most likely yes. If you do this in your personal account, then this will, most likely, run in parallel.

sophokles73: Since we run it on Travis in order to integrate properly with the GitHub build checks, I do not think that we should add another variant. It is already slow enough ...

ctron: However we would gain some insights on upcoming issues.

sophokles73:
maybe we could configure the jdk12 based build to run once a day using a CRON build job ...

sophokles73:
I was thinking about using a Travis CRON job for this purpose.

ctron:
> I was thinking about using a Travis CRON job for this purpose.

That would require the help of "webmaster", that is why I was thinking we could simply do this on the Jenkins instance.

sophokles73:
???

On the [builds page](https://travis-ci.org/eclipse/hono/builds) on the right hand side you can drop down *More Options* and select *Trigger Build*. What do we need webmaster for?

ctron:
Ok … I wasn't aware we have access to that. However it seems there can only be one trigger for each branch, so I am not sure how we can switch the JDK in the case of a cron job … maybe there is a way.

PR #1179, opened 2019-04-16
Fix issue in device registry deployment using helm.

kaniyan:


kaniyan:
@calohmn Thank you for the support.

PR #1187, opened 2019-04-23
Minor Command improvements

calohmn:
This is for #930:
- EDIT: removed: `Command.from()` is changed to not manipulate the given `Message` object anymore.
- EDIT: removed: `Command.getReplyToId()` has been renamed to `Command.getOriginalReplyToId()`: For #930 this method will be needed when forwarding the command AMQP message Hono-internally (restoring the original reply-to id of the AMQP message). The new name reflects its sematics in a better way.
- EDIT: added `getInvalidCommandReason()` - useful for inclusion in trace span logs.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/Command.java
+++ b/client/src/main/java/org/eclipse/hono/client/Command.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -16,6 +16,7 @@ package org.eclipse.hono.client;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.StringJoiner;
 
 import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.util.CommandConstants;
@@ -30,7 +31,10 @@ import io.vertx.core.buffer.Buffer;
  */
 public final class Command {
 
-    private final boolean valid;
+    /**
+     * If present, the command is invalid.
+     */
+    private final Optional<String> validationError;
     private final Message message;
     private final String tenantId;
     private final String deviceId;
@@ -39,14 +43,14 @@ public final class Command {
     private final String requestId;
 
     private Command(
-            final boolean valid,
+            final Optional<String> validationError,
             final Message message,
             final String tenantId,
             final String deviceId,
             final String correlationId,
             final String replyToId) {
 
-        this.valid = valid;
+        this.validationError = validationError;
         this.message = message;
         this.tenantId = tenantId;
         this.deviceId = deviceId;
@@ -61,13 +65,17 @@ public final class Command {
      * The message is expected to contain
      * <ul>
      * <li>a non-null <em>subject</em></li>
-     * <li>a non-null <em>reply-to</em> address that matches the tenant and device IDs and consists
+     * <li>either a null <em>reply-to</em> address (for a one-way command)
+     * or a non-null <em>reply-to</em> address that matches the tenant and device IDs and consists
      * of four segments</li>
      * <li>a String valued <em>correlation-id</em> and/or <em>message-id</em></li>
      * </ul>
      * <p>
      * If any of the requirements above are not met, then the returned command's {@link Command#isValid()}
      * method will return {@code false}.
+     * <p>
+     * Note that, if set, the <em>reply-to</em> address of the given message will be adapted, making sure it contains
+     * the device id.
      *
      * @param message The message containing the command.
      * @param tenantId The tenant that the device belongs to.
@@ -84,7 +92,10 @@ public final class Command {
         Objects.requireNonNull(tenantId);
         Objects.requireNonNull(deviceId);
 
-        boolean valid = message.getSubject() != null;
+        final StringJoiner validationErrorJoiner = new StringJoiner(", ");
+        if (message.getSubject() == null) {
+            validationErrorJoiner.add("subject not set");
+        }
 
         final String correlationId = Optional.ofNullable(message.getCorrelationId()).map(obj -> {
             if (obj instanceof String) {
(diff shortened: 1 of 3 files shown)


client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -84,6 +85,8 @@ public static Command from(
         Objects.requireNonNull(tenantId);
         Objects.requireNonNull(deviceId);
 
+        final Message commandMessage = HonoProtonHelper.getMessageCopyWithCopiedProperties(message);
```
ctron: I think it would be good to also have a note in the javadoc, that this method will return a new instance.

calohmn: I've adapted the javadoc of the `from` method accordingly.

client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -61,13 +62,17 @@ private Command(
...
+     * <li>either a null <em>reply-to</em> address (for a one-way command)
+     * or a non-null <em>reply-to</em> address that matches the tenant and device IDs and consists
      * of four segments</li>
      * <li>a String valued <em>correlation-id</em> and/or <em>message-id</em></li>
      * </ul>
      * <p>
      * If any of the requirements above are not met, then the returned command's {@link Command#isValid()}
      * method will return {@code false}.
+     * <p>
+     * The returned command will return a copy of the given message, containing an adapted <em>reply-to</em> address
```
sophokles73: what about
> The returned command will contain a copy of the given message with an adapted <em>reply-to</em> address ...

calohmn: I've adopted this change.

calohmn:
@sophokles73 For the most part it's about removing a side effect of the method, which currently causes issues when the method is invoked twice with the same message object. The 2nd invocation could be with a mapped gateway device id (scenario for #930).

I have since refactored my changes for #930 so that I'm not strictly requiring this change anymore. But nonetheless I think it's cleaner to not manipulate the method param object.
However, I would also be fine with undoing the change and updating the javadoc concerning the `message` param, if you prefer not doing the copy.

sophokles73:
> However, I would also be fine with undoing the change and updating the javadoc concerning the message param, if you prefer not doing the copy.

My concern is that copying the message is quite expensive and (so far) unnecessary. If you need to jump through several hoops in order to work around it, then I think we should use the approach taken in this PR. Otherwise, I think it would be worthwhile to not copy the message ...

calohmn:
I don't think the copy is expensive (it's mostly a shallow copy).
Anyway, I've removed the copy and adapted the javadoc instead. Also made some further minor changes.

client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -169,6 +173,29 @@ public boolean isValid() {
...
+        if (message.getSubject() == null) {
+            joiner.add("subject not set");
+        }
+        if (message.getCorrelationId() == null && message.getMessageId() == null) {
+            joiner.add("message/correlation id not set");
+        }
+        if (message.getReplyTo() != null && originalReplyToId == null) {
+            joiner.add("invalid replyTo '" + message.getReplyTo() + "'");
+        }
+        return joiner.toString();
```
sophokles73: wouldn't it make sense to do this in the factory method(s) already?

calohmn: yes, I've updated the class accordingly.

client/src/test/java/org/eclipse/hono/client/CommandTest.java
```diff
@@ -212,4 +212,17 @@ public void testFromMessageFailsForNonMatchingReplyToAddress() {
...
+     * Verifies the return value of getInvalidCommandReason().
+     */
+    @Test
+    public void testGetInvalidCommandReason() {
+        final Message message = mock(Message.class);
+        when(message.getReplyTo()).thenReturn("invalid-reply-to");
+        final Command command = Command.from(message, Constants.DEFAULT_TENANT, "4712");
+        assertFalse(command.isValid());
+        assertThat(command.getInvalidCommandReason(),
+                is("subject not set, message/correlation id not set, replyTo not a command address: invalid-reply-to"));
```
sophokles73: I do not really see the value of doing such literal error message comparisons. It only makes it harder to refactor later ...

client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -102,42 +112,42 @@ public static Command from(
...
-        String replyToId = null;
+        String originalReplyToId = null;
 
         if (message.getReplyTo() != null) {
             try {
                 final ResourceIdentifier replyTo = ResourceIdentifier.fromString(message.getReplyTo());
                 if (!CommandConstants.COMMAND_ENDPOINT.equals(replyTo.getEndpoint())) {
                     // not a command message
-                    valid = false;
+                    validationErrorJoiner.add("replyTo not a command address: " + message.getReplyTo());
```
sophokles73: the field is called *reply-to* in the AMQP message

client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -102,42 +112,42 @@ public static Command from(
...
                 } else if (!tenantId.equals(replyTo.getTenantId())) {
                     // command response is targeted at wrong tenant
-                    valid = false;
+                    validationErrorJoiner.add("replyTo not with tenant " + tenantId + ": " + message.getReplyTo());
```
sophokles73: the field is called *reply-to* in the AMQP message

client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -102,42 +112,42 @@ public static Command from(
...
                 } else {
-                    replyToId = replyTo.getPathWithoutBase();
-                    if (replyToId == null) {
-                        valid = false;
+                    originalReplyToId = replyTo.getPathWithoutBase();
+                    if (originalReplyToId == null) {
+                        validationErrorJoiner.add("replyTo part after tenant not set: " + message.getReplyTo());
```
sophokles73: the field is called *reply-to* in the AMQP message

client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -102,42 +112,42 @@ public static Command from(
...
                     } else {
                         message.setReplyTo(
-                                String.format("%s/%s", replyTo.getBasePath(), processForDeviceId(replyToId, deviceId)));
+                                String.format("%s/%s", replyTo.getBasePath(), adaptReplyToIdForGivenDevice(originalReplyToId, deviceId)));
                     }
                 }
             } catch (IllegalArgumentException e) {
                 // reply-to could not be parsed
-                valid = false;
+                validationErrorJoiner.add("replyTo cannot be parsed: " + message.getReplyTo());
```
sophokles73: the field is called *reply-to* in the AMQP message

client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -30,29 +31,32 @@
  */
 public final class Command {
 
-    private final boolean valid;
+    /**
+     * If present, the command is invalid.
+     */
+    private final Optional<String> validationError;
```
sophokles73: is there a particular reason why you want to use an Optional here instead of just a simple String?

calohmn: That's just for making it more obvious that `validationError` may be not set, marking the important distinction between a valid and not valid Command.

client/src/test/java/org/eclipse/hono/client/CommandTest.java
calohmn: I've changed that:
In order to still have something to test in this method, I just try to match the different message field names there now.

client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -265,14 +288,14 @@ public String getContentType() {
...
 
     /**
-     * Gets this command's reply-to-id.
+     * Gets this command's original reply-to-id, as extracted from the message given to the <em>from</em> factory method.
      *
      * @return The identifier.
      * @throws IllegalStateException if this command is invalid.
      */
-    public String getReplyToId() {
+    public String getOriginalReplyToId() {
```
sophokles73: IMHO we should not change the name of this method. FMPOV it is sufficient to document that the original, unaltered *reply-to* value will be returned. I guess that is the behavior that clients expect anyways.
Otherwise we would also need to document this API change in the release notes.

calohmn: Ok, yes. I've reverted this name change and adapted the javadoc instead. Also added some further javadoc emphasizing the difference between reply-to as received from the application and the adapted reply-to used when forwarding the command to the device.

sophokles73:
@calohmn can you fix the failing unit test?

calohmn:
@sophokles73 done.

Issue #1138, opened 2019-04-04
Device registry doesn't provide CORS headers


ctron:
The current HTTP endpoint of the device registry does not provide CORS headers. Which makes it rather difficult to work with under some circumstances.

sophokles73:
In particular, it means that we cannot use the Swagger UI to directly access the Sandbox registry, which would be great.
It would also allow us to get rid of the (ugly) Heroku CORS proxy in the OpenAPI spec file.

dejanb:
Fixed

PR #1189, opened 2019-04-24
[#1138] Add CORS support for device registry HTTP API

ctron:
This enables CORS support for the current HTTP API of the device registry endpoint.

There is one thing still missing, which is the source of the CORS allowed origin pattern. Currently this is implemented simply allowing everything:

~~~java
protected String getCorsAllowedOriginPattern() {
   return "*";
}
~~~
See: https://github.com/eclipse/hono/pull/1189/files#diff-128d9180cb18f40e68e4d2203ae4c83aR88

This is fine for the moment, as we don't authenticate the device registry API anyway. However, I think this should be allowed to be configured, like for the HTTP adapter.

However, the endpoint currently does not have any specific configuration for HTTP endpoints. It only receives a single configuration object, which has to be of type `ServiceConfigProperties` in the case of the device registry. And that type is shared with the AMQP endpoints. There are two instances though, one marked with the qualifier `ampq`, the other with `rest`.

I don't think it makes sense to add the CORS configuration to the generic `ServiceConfigProperties` object, as CORS seems specific to HTTP for me.

My proposal would be to create a derived `HttpServiceConfigProperties` object and add the configuration there. Requiring the use of this class for all HTTP related methods and classes.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsHttpEndpoint.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsHttpEndpoint.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
  *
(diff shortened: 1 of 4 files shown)


ctron:
We can revisit this after the device registry work on HTTP.

Issue #1183, opened 2019-04-18
Internal server error when sending malformed authentication header


mbaeuerle:
When sending a malformed Authentication header like "Authentication: basic test test" Hono responds with a 500 and with an error message "Illegal base64 character 20" which I suppose is from the thrown Exception.

**Examples**
``` sh
curl -X POST \
  http://hono.eclipse.org:8080/event \
  -H 'Authorization: basic test test' \
  -H 'Content-Type: application/json' \
  -H 'cache-control: no-cache' \
  -d '{ "test-data": "3" }'
```
Response:
> Status 500: Last unit does not have enough valid bits

``` sh
curl -X POST \
  http://hono.eclipse.org:8080/event \
  -H 'Authorization: basic testtest1' \
  -H 'Content-Type: application/json' \
  -H 'cache-control: no-cache'
  -d '{ "test-data": "3" }'
```
Response:
> Status 500: Last unit does not have enough valid bits

**Expected behavior**
- Status either 400 or 422
- Meaningful error message like: `Authorization Header malformed`

sophokles73:
FMPOV this should result in a 400 because the request does not fail due to a problem with the data in the body but cannot be processed due a problem in an earlier stage (where the body's content is of no concern yet).

kaniyan:
@sophokles73 IMHO this should be fixed in _vertx_ (`BasicAuthHandlerImpl`). The PR #1194 fixes this issue in _Hono_ by using a `RoutingContextDecorator`.

sophokles73:
fixed in https://github.com/eclipse/hono/commit/5e5735f5b4ff89b5e896298c3d264d9b7f187636

PR #1194, opened 2019-04-24
[#1183] Fix internal server error when sending malformed authentication header

kaniyan:
Fix for issue #1183 by using a `RoutingContextDecorator`.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/http/HonoBasicAuthHandler.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/http/HonoBasicAuthHandler.java
@@ -13,6 +13,7 @@
 
 package org.eclipse.hono.service.http;
 
+import io.vertx.ext.web.impl.RoutingContextDecorator;
 import org.eclipse.hono.client.ServiceInvocationException;
 import org.eclipse.hono.tracing.TracingHelper;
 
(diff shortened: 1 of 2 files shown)


service-base/src/main/java/org/eclipse/hono/service/http/HonoBasicAuthHandler.java
```diff
@@ -70,7 +72,14 @@ protected void processException(final RoutingContext ctx, final Throwable except
 
     @Override
     public void parseCredentials(final RoutingContext context, final Handler<AsyncResult<JsonObject>> handler) {
-        super.parseCredentials(context, ar -> {
+        final RoutingContextDecorator routingContextDecorator = new RoutingContextDecorator(context.currentRoute(),
+                context) {
+
+            public void fail(final Throwable throwable) {
+                context.fail(HttpURLConnection.HTTP_UNAUTHORIZED, throwable);
```
sophokles73: wouldn't be sufficient to add this line in an `else` block in the handler below? Using the `RoutingContextDecorator` should probably work but I think we can achieve the same result with a little less *ammo*, don't you think?
Can you also add a corresponding unit test?

kaniyan: @sophokles73 Already I checked for the `else` block, it doesn't work. Vertx `BasicAuthHandlerImpl` invokes `context.fail(e)` when error occurs, thereby the `DefaultFailureHandler` is being invoked.

sophokles73: I see. Can you then please add a comment explaining why using the `RoutingContextDecorator` in this case is necessary?

kaniyan: @sophokles73 I have added a unit test and a comment for the above.

PR #1195, opened 2019-04-24
POC of junit5/jupiter test migration

jbtrystram:
This is a POC on the tests to experiment the junit 5 tests, as discussed in a recent community call. 
As expected, it's needed to rewrite a bunch of things, but nothing really complicated.

Also, there is currently no support for class-wide `@Rule` timeout, as we currently use in hono. Each test must have a Timeout specified. However, this feature should be introduced back in the next jupiter release, as this is a popular demand ( https://github.com/junit-team/junit5/issues/80).

On the plus side there are some nice features such as nested tests that would allow to improve the maintainability of the tests ( see  `FileBasedCredentialsServiceTest.testLoadCredentialsCanReadOutputOfSaveToFile` where I needed `countdownLatch` to replace the vertx `Async`. This could be factored the `@BeforeEach` method of a nested test class)
I also removed all the vertx assertions to use only the assertions provided by Junit. 
It is also nice that both Junit4 and jupiter can coexist without disrupting anything, so the migration can happen over time.

I also completely removed the junit4 dependency : all the tests can run through the vintage engine. I had to add hamcrest as i think it was pulled through junit4. Keeping both dependencies made the codecov bot miss out on the test coverage.

For more details on `VertxTestContext` see https://vertx.io/docs/vertx-junit5/java/

Signed-off-by: Trystram Jean-Baptiste <jbtrystram@redhat.com>

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -217,6 +217,11 @@
         <scope>test</scope>
       </dependency>
       <dependency>
+        <groupId>io.vertx</groupId>
+        <artifactId>vertx-junit5</artifactId>
+        <version>${vertx.version}</version>
+      </dependency>
+      <dependency>
         <groupId>org.junit.jupiter</groupId>
         <artifactId>junit-jupiter-engine</artifactId>
         <version>${junit.jupiter.version}</version>
@@ -225,6 +230,12 @@
       <dependency>
         <groupId>org.junit.vintage</groupId>
         <artifactId>junit-vintage-engine</artifactId>
+        <version>${junit.jupiter.version}</version>
+        <scope>test</scope>
+      </dependency>
+      <dependency>
+        <groupId>org.junit.jupiter</groupId>
+        <artifactId>junit-jupiter-params</artifactId>
         <version>${junit.jupiter.version}</version>
         <scope>test</scope>
       </dependency>
--- a/service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
+++ b/service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
@@ -12,17 +12,16 @@
  *******************************************************************************/
 package org.eclipse.hono.service.credentials;
 
+import static org.junit.Assert.assertEquals;
+
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
-import io.vertx.ext.unit.Async;
-import io.vertx.ext.unit.TestContext;
+import io.vertx.junit5.Checkpoint;
+import io.vertx.junit5.VertxTestContext;
 import org.eclipse.hono.util.CredentialsConstants;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
 
 import java.net.HttpURLConnection;
-
-import static org.hamcrest.CoreMatchers.is;
-import static org.junit.Assert.assertThat;
 
 /**
  * Abstract class used as a base for verifying behavior of {@link CompleteCredentialsService} in device registry implementations.
@@ -42,7 +41,7 @@ public abstract class AbstractCompleteCredentialsServiceTest {
      * @param ctx The vert.x test context.
      */
     @Test
-    public void testAddCredentialsRefusesDuplicateRegistration(final TestContext ctx) {
+    public void testAddCredentialsRefusesDuplicateRegistration(final VertxTestContext ctx) {
 
         register(getCompleteCredentialsService(), "tenant", "device", "myId", "myType", ctx);
 
@@ -51,12 +50,10 @@ public abstract class AbstractCompleteCredentialsServiceTest {
                 .put(CredentialsConstants.FIELD_AUTH_ID, "myId")
                 .put(CredentialsConstants.FIELD_TYPE, "myType")
                 .put(CredentialsConstants.FIELD_SECRETS, new JsonArray());
-        final Async add = ctx.async();
-        getCompleteCredentialsService().add("tenant", payload2, ctx.asyncAssertSuccess(s -> {
-            assertThat(s.getStatus(), is(HttpURLConnection.HTTP_CONFLICT));
-            add.complete();
-        }));
-        add.await();
+        getCompleteCredentialsService().add("tenant", payload2, ctx.succeeding(s -> ctx.verify(() -> {
+            assertEquals(HttpURLConnection.HTTP_CONFLICT, s.getStatus());
+            ctx.completeNow();
+        })));
     }
 
     /**
@@ -65,14 +62,12 @@ public abstract class AbstractCompleteCredentialsServiceTest {
      * @param ctx The vert.x test context.
      */
     @Test
-    public void testGetCredentialsFailsForNonExistingCredentials(final TestContext ctx) {
+    public void testGetCredentialsFailsForNonExistingCredentials(final VertxTestContext ctx) {
 
-        final Async get = ctx.async();
-        getCompleteCredentialsService().get("tenant", "myType", "non-existing", ctx.asyncAssertSuccess(s -> {
-            assertThat(s.getStatus(), is(HttpURLConnection.HTTP_NOT_FOUND));
-            get.complete();
-        }));
-        get.await();
+        getCompleteCredentialsService().get("tenant", "myType", "non-existing", ctx.succeeding(s -> ctx.verify(() -> {
+            assertEquals(HttpURLConnection.HTTP_NOT_FOUND, s.getStatus());
+            ctx.completeNow();
+        })));
     }
 
     /**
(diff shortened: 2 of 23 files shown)


service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
```diff
@@ -147,17 +136,15 @@ public void testGetCredentialsFailsForWrongClientContext(final TestContext ctx)
...
+    public void testRemoveCredentialsByAuthIdAndTypeSucceeds(final VertxTestContext ctx) {
 
         register(getCompleteCredentialsService(), "tenant", "device", "myId", "myType", ctx);
 
-        final Async remove = ctx.async();
-        getCompleteCredentialsService().remove("tenant", "myType", "myId", ctx.asyncAssertSuccess(s -> {
-            assertThat(s.getStatus(), is(HttpURLConnection.HTTP_NO_CONTENT));
+        getCompleteCredentialsService().remove("tenant", "myType", "myId", ctx.succeeding(s -> ctx.verify(() -> {
+            assertEquals(HttpURLConnection.HTTP_NO_CONTENT, s.getStatus());
             assertNotRegistered(getCompleteCredentialsService(), "tenant", "myId", "myType", ctx);
```
dejanb: Does this helps with "nested" calls issues we experienced with Infinispan client?

jbtrystram: did not tried yet !

bom/pom.xml
```diff
@@ -42,7 +43,7 @@
     <javax.annotation.api.version>1.3.2</javax.annotation.api.version>
     <jjwt.version>0.7.0</jjwt.version>
     <jmeter.version>3.3</jmeter.version>
-    <junit.version>4.12</junit.version>
+    <junit.jupiter.version>5.4.2</junit.jupiter.version>
```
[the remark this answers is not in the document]
jbtrystram: Why not. We'll have to use 5.5 if we want to get back the `Timer` rule anyway..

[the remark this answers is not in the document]
dejanb: @jbtrystram So timer rule is just ignored at the moment? Is there a workaround for that before 5.5? Or is it not worth bothering?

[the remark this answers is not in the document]
dejanb: Also, we might want to try 5.5.0-M1 if we can wait for the official release before merging it?

service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
```diff
@@ -51,12 +50,10 @@ public void testAddCredentialsRefusesDuplicateRegistration(final TestContext ctx
...
                 .put(CredentialsConstants.FIELD_TYPE, "myType")
                 .put(CredentialsConstants.FIELD_SECRETS, new JsonArray());
-        final Async add = ctx.async();
-        getCompleteCredentialsService().add("tenant", payload2, ctx.asyncAssertSuccess(s -> {
-            assertThat(s.getStatus(), is(HttpURLConnection.HTTP_CONFLICT));
-            add.complete();
-        }));
-        add.await();
+        getCompleteCredentialsService().add("tenant", payload2, ctx.succeeding(s -> ctx.verify(() -> {
+            assertEquals(HttpURLConnection.HTTP_CONFLICT, s.getStatus());
```
sophokles73: what's wrong with using `assertThat`? FMPOV it is easier to read and produces better error messages...

service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
```diff
@@ -51,12 +50,10 @@ public void testAddCredentialsRefusesDuplicateRegistration(final TestContext ctx
...
+            ctx.completeNow();
```
sophokles73: it looks strange to me that you would need to explicitly invoke this method. Have you checked if it works without?

service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
dejanb: No, I don't think it would. Just thinking maybe we can tackle this at the same time. You can always use your local builds (SNAPSHOTS) to test against Infinispan. But we can do it, one step at the time.

bom/pom.xml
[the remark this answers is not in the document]
jbtrystram: in tests using jupiter the timer ruel is ignored yeah. It's explained better here : https://github.com/junit-team/junit5/issues/80
trying  5.5.0-M1 make sense yes. As i'd like to migrate the complete module before merging anyway

.travis.yml
```diff
@@ -6,6 +6,7 @@ jdk:
   - openjdk11
 
 script:
+  - mvn clean package
```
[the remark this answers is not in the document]
jbtrystram: Nothing, i was trying to reseolve the jacoco issue. it was reverted.

service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
jbtrystram: `assertThat` is removed from junit5. IMHO, for comparing HTTP status codes `assertEquals` is more intuitive.

service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
jbtrystram: the `ctx.completeNow();` method ? Yes, I agree with you that it's a bit odd. 
I went to re-read the vertx doc and send you a link but it look like when the test expect a success I could use `ctx.completing()` which complete the test context on a success. I am going to try it !

ctron: So as @jbtrystram explained, with vertx it is necessary to let the async test cases know that the test is actually completed. As the idea is that the unit tests may still be running when it exited the method annotated with `@Test`.

One thing to note here, the call to `completeNow()` actually completes the test context. If the test has multiple steps, you need to use `Checkpoint` as done in the other tests.

service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
sophokles73: just wanted to make sure that we do not strictly replace all occurrences of `assertThat` with `assertEquals` because we think that we need to.

sophokles73:
@jbtrystram would it also be possible to keep all tests *as-is* for the time being and simply replace junit 4.12 with junit5 using the legacy engine?

jbtrystram:
@sophokles73 it's exactly what i've done here :)  See https://github.com/eclipse/hono/pull/1195/commits/b69291e8c0b8e3fedf6335e64c5997beb01d7964

service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
jbtrystram: I'll definitely give it a try :)

service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
jbtrystram: We absolutely don't have to. As long as `hamcrest-core` stays in the dependencies that will work fine. I simply wanted to try it because using the built-in assertions would remove a dependency. And also for the sake of trying something new :)

sophokles73:
@jbtrystram FMPOV it would be nice to create a separate PR for migrating to JUnit5 first, i.e. removing the dependency on JUnit4 altogether and instead run the existing tests using the legacy engine.
Then we could use this PR to actually start migrating the first tests to JUnit5.
WDYT? @dejanb @jbtrystram

dejanb:
@sophokles73 sounds reasonable to me. Did we decide which exact version are we going to use? 5.5.0-M1 for the timeout support?

sophokles73:
@dejanb @jbtrystram not sure if I fully understood the consequences of not having the *timeout* support. Would that mean that the `Timeout` Rule that we use in the existing tests wouldn't work any more or would it just mean that in newly created JUnit5 based tests we would need to specify the timeout on each method individually? The latter wouldn't be a problem FMPOV since in many cases we would like to do that in any case ...
So, my preference would be *not* to use a milestone but a released version, if possible.

jbtrystram:
> it just mean that in newly created JUnit5 based tests we would need to specify the timeout on each method individually

@dejanb @sophokles73  This is correct I think.

sophokles73:
@jbtrystram @dejanb can we then agree on using the most recent *released* version of JUnit5?

dejanb:
@sophokles73 yes. I'll get the CQs going.

sophokles73:
@dejanb we do not need CQs for that, we already have a *works-with* CQ for JUnit which covers version 4 and higher ...

dejanb:
@sophokles73 ok, great

sophokles73:
@jbtrystram can you rebase this PR now that we have migrated to JUnit5?

ctron:
@sophokles73 I am trying to sum up what is missing in order to merge this PR. It is rebased, we don't need new CQs. So the only thing missing is to move the `@Timeout` annotation to individual tests?!

sophokles73:
@ctron you probably mean: we do not need new **PR**s, right? FMPOV there is nothing missing apart from @dejanb and your approval ...

jbtrystram:
@ctron  the default 30sec timeout is still working!
I migrated the file based tests as well (as they extend the base tests), FMPOV it's good to go, and thoses tests can be looked by other guys that may want to write tests in the future :)

ctron:
> @ctron you probably mean: we do not need new PRs, right? FMPOV there is nothing missing apart from @dejanb and your approval ...

No, I did mean that we do not need additional CQs for merging this.

PR #1197, opened 2019-04-25
Upgrade to JJWT 0.10.6

ctron:
* Update to version 0.10.6
* Use -api, -impl, and -jackson modules
* Drop the use of deprecated JJWT APIs
* Fix a misuse of the key material (testing only)


ctron:
CQ for new dependency (covers all three modules):

* [ ] https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19596

PR #1199, opened 2019-04-25
Pass through original message in LoRaWAN adapter

### LINKED ISSUE #1190: (Actility) LoRaWAN adapter does not pass through the metadata
The actility LoRaWAN adapter does not passtrough the meta data received from the actility LoRaWAN provider. I think it would be nice to have at least the possibility to analyze meta data in the backend application. When it is not provided, you do not have the possibility to analyze it.
I suggest adding this data in the headers of the message posted on the AMQP network.
We will implement this anyway but I would rather push this to the upstream as well.
WDYT?

BobClaerhout:
Fixes #1190: LoRaWAN adapter does not pass through the metadata

### DIFF
--- a/adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/impl/LoraProtocolAdapter.java
+++ b/adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/impl/LoraProtocolAdapter.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2019, 2019 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -18,11 +18,13 @@ import static java.net.HttpURLConnection.HTTP_INTERNAL_ERROR;
 import static java.net.HttpURLConnection.HTTP_OK;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Collections;
 
+import io.vertx.core.buffer.impl.BufferImpl;
 import org.apache.qpid.proton.amqp.transport.ErrorCondition;
 import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.adapter.http.AbstractVertxBasedHttpProtocolAdapter;
@@ -76,6 +78,7 @@ import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.client.HttpResponse;
 import io.vertx.ext.web.handler.ChainAuthHandler;
 
+
 /**
  * A Vert.x based Hono protocol adapter for receiving HTTP push messages from and sending commands to LoRa backends.
  */
@@ -89,6 +92,8 @@ public final class LoraProtocolAdapter extends AbstractVertxBasedHttpProtocolAda
     private static final String TAG_LORA_PROVIDER = "lora_provider";
     private static final String JSON_MISSING_REQUIRED_FIELDS = "JSON Body does not contain required fields";
     private static final String INVALID_PAYLOAD = "Invalid payload";
+    private static final String FIELD_BINARY_PAYLOAD = "payload";
+    private static final String FIELD_ORIG_UPLINK_MESSAGE = "orig-message";
 
     private final List<LoraProvider> loraProviders = new ArrayList<>();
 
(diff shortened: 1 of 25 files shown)


adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/providers/LoraProvider.java
```diff
@@ -37,11 +37,20 @@
...
      */
     String pathPrefix();
 
+    /**
+     * The content type which identifies content of this provider. E.g. "application/json+myloraprovider".
+     *
+     * @return The content type of this LoRaWAN provider. E.g. "application/json+myloraprovider".
+     */
+    default String getContentType(){
+        return HttpUtils.CONTENT_TYPE_JSON + "+" + getProviderName();
```
sophokles73: I think we should try to adhere to common practices when [defining media types](https://tools.ietf.org/html/rfc6838). 
In particular, we should use a type defined in the *Vendor* tree, e.g. using `application/vnd.eclipse-hono` as the prefix. The suffix (the stuff after the `+` sign) should be a (registered) *Structured Syntax Name*.  Most likely `json` in our case.

We could then either add the provider name to the sub type name, e.g. 

`application/vnd.eclipse-hono.lora.actility+json`

or use a generic media type indicating that the data structure has been created by the LoRa protocol adapter, e.g.

`application/vnd.eclipse-hono.lora+json`

and then add the network provider's name as an application property to the AMQP message.

Personally, I think the latter one makes more sense, in particular if we want to also normalize a lot of the data.

sophokles73:
@BobClaerhout can you also fix the unit tests?

sophokles73:
@BobClaerhout any news on this?

BobClaerhout:
> @BobClaerhout any news on this?

I'm sorry. I was on business trip last week and I have to finish some work before I can continue on this. Will do my best to update this week.

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/providers/LoraProvider.java
BobClaerhout: I've changed it again.
As for your suggestion on the application property of the AMQP message, I've seen that this is already done in the "customizeDownstreamMessage" function:
```
@Override
    protected void customizeDownstreamMessage(final Message downstreamMessage, final RoutingContext ctx) {
        MessageHelper.addProperty(downstreamMessage, LoraConstants.APP_PROPERTY_ORIG_LORA_PROVIDER,
                ctx.get(LoraConstants.APP_PROPERTY_ORIG_LORA_PROVIDER));
    }
```
The context variable is set earlier on using
`ctx.put(LoraConstants.APP_PROPERTY_ORIG_LORA_PROVIDER, provider.getProviderName());`

BobClaerhout:
> @BobClaerhout can you also fix the unit tests?

Done

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/impl/LoraProtocolAdapter.java
```diff
@@ -265,8 +265,8 @@ private void doUpload(final RoutingContext ctx, final Device device, final Strin
         }
     }
 
-    private JsonObject createDownstreamJson(final String base64Payload) {
-        return new JsonObject().put("payload", base64Payload);
+    private JsonObject createDownstreamJson(final String base64Payload, final JsonObject originalMessage) {
+        return new JsonObject().put("payload", base64Payload).put("originalMessage", originalMessage);
```
sophokles73: can you please define the property names as constants, e.g. `FIELD_BINARY_PAYLOAD` and `FIELD_ORIG_UPLINK_MESSAGE`? We also usually do not use camel case for JSON properties but use a dash to separate parts, e.g. `orig-message` instead of `origMessage`.

BobClaerhout: Done

service-base/src/main/java/org/eclipse/hono/service/http/HttpUtils.java
```diff
@@ -42,6 +42,10 @@
      * The <em>application/json</em> content type.
      */
     public static final String CONTENT_TYPE_JSON = "application/json";
+    /**
+     * The <em>application/vnd.eclipse-hono.lora.actility+json</em> content type.
+     */
+    public static final String CONTENT_TYPE_LORA = "application/vnd.eclipse-hono.lora+json";
```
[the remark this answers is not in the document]
BobClaerhout: Right, replaced it.

BobClaerhout:
@sophokles73, the build failed on the AMQP adapter. Any idea what this might be? Can I trigger this again somehow?

sophokles73:
It looks like the build process was not able to retrieve some artifact from Maven Central. This looks like a network issue with Travis. I have triggered the PR build again ...

BobClaerhout:
@sophokles73 , tests are running just fine on my computer. Can you please trigger the build once more?

Issue #1203, opened 2019-04-26
Implement credentials caching mechanism in Hono Adapters to speedup credentials lookup


almheida:
Protocol adaptors use Hono Credentials API to lookup credentials for device authentication, each time the device is connected to the adapter. So far by messaging flow (specifically through HTTP protocol) a lookup operation against the credentials store is to be performed each time the device sends message through the adapter. That might turn to be costly in term of many resulting request timeouts when applying huge messaging load to the adapter. Therefore, implementing lookup caching mechanism for device credentials in the adapter may solve the issue here.

sophokles73:
The reason why the `CredentialsClientFactoryImpl` currently does not cache results is for security reasons. In general, it is considered bad practice to cache credentials. In addition to that, the (Credentials) service implementation uses the *cache-control* application property in its response messages to indicate to the client, if the data in the body is supposed to be cached, and if so, how long.

I can imagine that
1. we add general support for caching response data to the `CredentialsClientFactoryImpl` as we have it in all other client factories.
2. we change the Credentials service base classes to include a *cache-control* property in its responses which allows clients to cache the data. We should not make this the default behavior but allow it to be enabled explicitly by setting a configuration option. We also should allow caching of only those credential types that do not contain plain text credentials (like pre-shared key does).

DanielMaier-BSI:
What came into my mind regarding this topic:

Caching of hashed credentials should not be a security issue, so I am fine with that and I also think we should do this. Especially HTTP based adapters need to execute many authentication requests. 

Caching the credentials locally in the adapters would already improve the situation, however in our deployment setup this comes with some limitations:
- Our deployment on AWS uses ELB which load balances requests to different machines without any stickiness
- Our deployment uses Kubernetes which load balances service requests in a round robin manner between pods

For our deployment it would probably make sense to use some kind of distributed cache like infinispan or redis. The question is now, do you think it would be useful to discuss and implement such distributed caching in open source space or is this to deployment specific and hono should only define the caching API and provide a local caching implementation?

sophokles73:
> For our deployment it would probably make sense to use some kind of distributed cache like infinispan or redis.

Actually, I doubt that this would make much difference, because the expensive operation will not be the retrieval from the DB but the remote invocation of the Credentials service from the protocol adapters. If you replace the local cache in the adapters with a distributed cache, then you will also have remote calls for retrieving the values from another cache node in case of a cache-miss.
I would therefore try to keep it simple by using a local cache only ...

[...]

PR #1213, opened 2019-04-30
[#1203] Add caching mechanism to credentials client factory implementation.

kaniyan:
With reference to the issue #1203, this PR aims to support caching of the response data to the `CredentialsClientFactoryImpl` as in the other client factories.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientImpl.java
@@ -18,7 +18,10 @@ import java.net.HttpURLConnection;
 import java.util.Objects;
 import java.util.UUID;
 
+import io.vertx.proton.ProtonReceiver;
+import io.vertx.proton.ProtonSender;
 import org.apache.qpid.proton.amqp.messaging.ApplicationProperties;
+import org.eclipse.hono.cache.CacheProvider;
 import org.eclipse.hono.client.ClientErrorException;
 import org.eclipse.hono.client.CredentialsClient;
 import org.eclipse.hono.client.HonoConnection;
@@ -30,6 +33,7 @@ import org.eclipse.hono.util.CredentialsObject;
 import org.eclipse.hono.util.CredentialsResult;
 import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.RequestResponseApiConstants;
+import org.eclipse.hono.util.TriTuple;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -66,6 +70,19 @@ public class CredentialsClientImpl extends AbstractRequestResponseClient<Credent
      */
     CredentialsClientImpl(final HonoConnection connection, final String tenantId) {
         super(connection, tenantId);
+    }
+
+    /**
+     * Creates a new client for accessing the Credentials service.
+     *
+     * @param connection The connection to Hono.
+     * @param tenantId The identifier of the tenant for which the client should be created.
+     * @param sender The AMQP link to use for sending requests to the service.
+     * @param receiver The AMQP link to use for receiving responses from the service.
+     */
+    CredentialsClientImpl(final HonoConnection connection, final String tenantId, final ProtonSender sender,
+            final ProtonReceiver receiver) {
+        super(connection, tenantId, sender, receiver);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


kaniyan:
@sophokles73 Rebased and also changed the `AbstractAdapterConfig` to support this feature.

client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientImpl.java
```diff
@@ -180,22 +203,29 @@ public static final String getTargetAddress(final String tenantId) {
                 .put(CredentialsConstants.FIELD_TYPE, type)
                 .put(CredentialsConstants.FIELD_AUTH_ID, authId)
                 .mergeIn(clientContext);
+        final TriTuple<CredentialsConstants.CredentialsAction, String, String> key = TriTuple.of(
```
sophokles73: I think that we need to consider the client context in the key as well. IIRC the client context might actually be the *real* source of information used by the service to determine the data it puts into the response. For example with Kapua, all devices are using the same auth-id (conveyed in the username in an MQTT CONNECT packet) but all use different MQTT client IDs which are then used by the Credentials service implementation as the *real* auth-id and used to look up the password ...

kaniyan: @sophokles73 `TriTuple.of(CredentialsConstants.CredentialsAction.get, String.format("%s-%s", type, authId), clientContext)` WDYT?

sophokles73: I'd rather not keep a reference to the `JsonObject` in order to allow for it to be GCed more easily. Maybe we can use its hash code instead?

kaniyan: I have used the hash code instead and pushed the changes.

client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientImpl.java
```diff
@@ -180,22 +203,28 @@ public static final String getTargetAddress(final String tenantId) {
                 .put(CredentialsConstants.FIELD_TYPE, type)
                 .put(CredentialsConstants.FIELD_AUTH_ID, authId)
                 .mergeIn(clientContext);
+        final TriTuple<CredentialsConstants.CredentialsAction, String, Integer> key = TriTuple
+                .of(CredentialsConstants.CredentialsAction.get, String.format("%s-%s", type, authId), clientContext.hashCode());
```
sophokles73: what if client context is `null`?

In this context: can you also please fix the @throws tag of the JavaDoc of `CredentialsClient.get(String, String, JsonObject)` accordingly?

kaniyan: In javadoc  of `CredentialsClient.get(String, String, JsonObject)` it is mentioned that `@param clientContext Optional bag of properties that can be used to identify the device`. Since it is optional, I thing it is valid to set as also null.

May be it makes more sense: if client context is not `null`, then pass its `hashcode` as the third argument, else pass `null` instead. WDYT?

sophokles73: I was just hoping for you to fix it like this:
> @throws NullPointerException if any of the parameters other than client context are {@code null}

kaniyan: > @throws NullPointerException if any of the parameters other than client context are {@code null}

Sorry, I didn't get the intention behind the JavaDoc statement. In the current code (in PR), if the client context is `null`, then the `NullPointerException` is thrown due to `clientContext.hashCode()`. Then the above JavaDoc statement is contradictory.  If we fix the code with ` Optional.ofNullable(clientContext).map(cc-> cc.hashCode()).orElse(null);`, then the above JavaDoc statement is valid and be used. 

If we intend to leave the code as it is in this PR, then no need to change any JavaDoc statement. In that case, the below is valid. 
```
     * @see RequestResponseClient#setRequestTimeout(long)
     */
    Future<CredentialsObject> get(String type, String authId, JsonObject clientContext);
```

sophokles73: Yeah, I see what you mean. I misread `(except spanContext)` for `(except clientContext)`. So, it seems that the methods indeed require a non-null `JsonObject` as client context. However, the JSON may be empty.
So, the client context object itself should indeed never be `null`. However, I would prefer if we could add an explicit check at the top of the method then, instead of just running into a `NullPointerException` down here, ok?

kaniyan: I have added an explicit check at the top.

PR #1217, opened 2019-05-02
Support setting default TTL for downstream messages

sophokles73:
This PR extends the concept of *default properties* to the tenant level and then uses this concept to support setting a default *time-to-live* on downstream messages.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/MessageHelper.java
+++ b/core/src/main/java/org/eclipse/hono/util/MessageHelper.java
@@ -105,6 +105,27 @@ public final class MessageHelper {
     public static final String APP_PROPERTY_TENANT_ID = "tenant_id";
 
     /**
+     * The AMQP 1.0 <em>delivery-count</em> message header property.
+     */
+    public static final String SYS_HEADER_PROPERTY_DELIVERY_COUNT = "delivery-count";
+    /**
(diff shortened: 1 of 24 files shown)


core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -104,6 +104,27 @@
      */
     public static final String APP_PROPERTY_TENANT_ID = "tenant_id";
 
+    /**
```
dejanb: Should we add somewhere in the docs what properties can be set as defaults?

sophokles73: I tried to do this in the protocol adapter user guides, e.g. https://www.eclipse.org/hono/user-guide/http-adapter/#downstream-meta-data ...

dejanb: OK, now I see it. So only ttl and content-type can be set.

sophokles73: for the time being: yes. For most of the AMQP defined properties, a default value doesn't make much sense, or are there others you would like to support?

dejanb: No, it's all good. I just wanted to understand it.

PR #1215, opened 2019-05-02
[#930] Make getDevice() a mandatory operation

calohmn:
This is for #930:
The `Get Registration Information` operation is mandatory now.

PR #1222, opened 2019-05-06
Add 'get' method in RegistrationClient with SpanContext parameter

calohmn:
Adds an additional `RegistrationClient#get` method with a `SpanContext` parameter.

This relates to #1215 (`RegistrationClient#get` being now a mandatory operation).

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/RegistrationClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/RegistrationClient.java
@@ -121,6 +121,31 @@ public interface RegistrationClient extends RequestResponseClient {
     Future<JsonObject> get(String deviceId);
 
     /**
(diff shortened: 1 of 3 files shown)


ctron:
Wouldn't it make sense to turn the "default" around. Let the non-span-context method be a default, calling into `get` with a `null` context?!

calohmn:
@ctron I see that this would have the benefit of putting more focus on the spanContext method getting implemented. (And that way we could more easily deprecate the non-spanContext method at some point - although that's just an idea, not having been discussed yet).

On the other hand, this pattern of having the spanContext method be a default method is used throughout the Hono service/client interfaces. That's why I'd rather keep it the same in this PR.

We could switch to the other pattern in a separate PR, doing it for all service/client interfaces.

PR #1223, opened 2019-05-06
Abstract credentials test race condition

jbtrystram:
The credentials tests did not wait for the credential registration to be complete before retrieving the data.

### DIFF
--- a/service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
+++ b/service-base/src/test/java/org/eclipse/hono/service/credentials/AbstractCompleteCredentialsServiceTest.java
@@ -14,6 +14,8 @@ package org.eclipse.hono.service.credentials;
 
 import static org.junit.Assert.assertEquals;
 
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Future;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import io.vertx.junit5.VertxTestContext;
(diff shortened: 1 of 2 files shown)


jbtrystram:
in the process of writing that fix, I realised that there is a lot of redundancy in those tests.
Basically a test will add the credential to the registry, then get/delete/update it.

So we could factor the credential insertion in a `@BeforeEach` method. The issue here is that when extending a test class, if the extending class also defines a `@BeforeEach` method, then it's executed *before* the parent class' one. 
Typically, the extending class would spin up the device registry then run the tests. So using a `@BeforeEach` class here wouldn't work. Alternatively we could define an abstract method that must be implemented as the `BeforeEach` method in the extending classes. But it may be weird

ctron:
What would happen when the handler set by `setHandler` gets called with a *failed* result?

jbtrystram:
test fails : 
```
 svc.add("tenant", data, ctx.succeeding(s -> ctx.verify(() -> {
            assertEquals(HttpURLConnection.HTTP_CREATED, s.getStatus());
            future.complete();
        })));
```
the VertxTestContext expect a success on the assertion following the `add` operation.

ctron:
But for the following:

~~~java
registration.setHandler(r-> {});
~~~

I don't see any check there. The handler would always be called, not matter if the registration was successful or not.

jbtrystram:
Does the execution carry through even if the context is already failed ?

dejanb:
@jbtrystram I wouldn't bother with moving register() into `@BeforeEach`. This way is actually clearer what's being tested.

jbtrystram:
@ctron Completing the future with the `RegistrationResult` then 
```
        registration.setHandler(r-> {
           if ( ((RegistrationResult) r).isOk() ) {
                    // carry on...
           }
      });
```
WDYT ?

dejanb:
> WDYT ?

I was looking into this and the test stops properly on the first failure. I don't think we need extra checks.

jbtrystram:
> > WDYT ?
> 
> I was looking into this and the test stops properly on the first failure. I don't think we need extra checks.

nice. That was I assumed but i didn't double checked. Thanks :)

PR #1224, opened 2019-05-07
[#930] Fix command error handling; unify tracing/logging

calohmn:
For #930:
This resolves a TODO in `DelegatedCommandSenderImpl` concerning handling of non-accepted command messages.
Such scenarios now cause `DelegatedCommandSenderImpl.sendMessageAndWaitForOutcome()` to return a *succeeded* Future, so that the contained `ProtonDelivery` object can be passed on to the originating command sender.

Furthermore, tracing and logging of the received delivery update in the corresponding methods of `AbstractSender` is now done in a common method.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
@@ -41,6 +41,7 @@ import org.slf4j.LoggerFactory;
 
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
+import io.opentracing.log.Fields;
 import io.opentracing.tag.Tags;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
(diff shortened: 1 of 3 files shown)


client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
```diff
@@ -329,4 +318,65 @@ protected final Span startSpan(final Message message) {
             return Future.failedFuture(t);
         });
     }
+
+    /**
+     * Creates a log entry in the given span with the message id and information about the sender (credits, QOS).
+     * 
+     * @param currentSpan The current span to log to.
+     * @param messageId The message id.
+     */
```
sophokles73: can you also document the `NullPointerException` being thrown?

client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
```diff
@@ -329,4 +318,65 @@ protected final Span startSpan(final Message message) {
...
+    /**
+     * Creates a log entry in the given span with information about the message delivery outcome given in the delivery
+     * parameter. Sets the {@link Tags#HTTP_STATUS} as well.
+     * <p>
+     * Also corresponding log output is created.
+     *
+     * @param currentSpan The current span to log to.
+     * @param messageId The message id.
+     * @param delivery The updated delivery.
+     */
```
sophokles73: @throws NullPointerException ?

client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
calohmn: I've added that.

client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
calohmn: I've added that.

PR #1225, opened 2019-05-07
[#930] Rename DelegatingCommandHandler, add tests and improve tracing.

calohmn:
For #930:

This renames the `DelegatingCommandHandler` class to `DelegateViaDownstreamPeerCommandHandler`, making the class only responsible for one thing, namely delegating the handling of a command message (received on a tenant-scoped link) to a device-specific command consumer by sending the command message (with a modified `To` address) back to the downstream peer.

Furthermore, tests for the classes introduced in #1145 have been added. And tracing of command messages is improved by properly adding the tracing context in `CommandContext`.

The `CommandContext` class is not `final` anymore here, so that it can be mocked (configuring Mockito so that it can also mock final methods doesn't seem worth it regarding effort and the possible performance hit here, imho).

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/CommandConsumerFactoryImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/CommandConsumerFactoryImpl.java
@@ -137,7 +137,7 @@ public class CommandConsumerFactoryImpl extends AbstractHonoClientFactory implem
                 final Future<MessageConsumer> deviceSpecificConsumerFuture = Future.future();
                 deviceSpecificCommandConsumerFactory.getOrCreateClient(
                         key,
-                        () -> newCommandConsumer(tenantId, deviceId, commandHandler, remoteCloseHandler),
+                        () -> newDeviceSpecificCommandConsumer(tenantId, deviceId, commandHandler, remoteCloseHandler),
                         deviceSpecificConsumerFuture);
                 // create the tenant-scoped consumer that delegates/maps incoming commands to the right handler/consumer
                 final Future<MessageConsumer> tenantScopedCommandConsumerFuture = getOrCreateTenantScopedCommandConsumer(tenantId);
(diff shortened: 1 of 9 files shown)


client/src/main/java/org/eclipse/hono/client/impl/CommandConsumerFactoryImpl.java
```diff
@@ -167,12 +167,22 @@ private String getKey(final String tenantId, final String deviceId) {
...
+                    final Handler<CommandContext> commandHandler = deviceSpecificCommandHandlers
+                            .get(getKey(tenantId, deviceId));
+                    if (commandHandler != null) {
+                        log.trace("use local command handler for device {}", deviceId);
+                        commandHandler.handle(commandContext);
+                    } else {
+                        // delegate to matching consumer via downstream peer
+                        delegatingCommandHandler.handle(commandContext);
+                    }
+                });
```
sophokles73: I also prefer this to be done here instead of the delegating command handler :+1:

client/src/main/java/org/eclipse/hono/client/impl/GatewayMapperImpl.java
```diff
@@ -41,9 +42,9 @@ public GatewayMapperImpl(final RegistrationClientFactory registrationClientFacto
     }
 
     @Override
-    public Future<String> getMappedGatewayDevice(final String tenantId, final String deviceId) {
+    public Future<String> getMappedGatewayDevice(final String tenantId, final String deviceId, final SpanContext context) {
         return registrationClientFactory.getOrCreateRegistrationClient(tenantId).compose(client -> {
-            return client.get(deviceId);
+            return client.get(deviceId); // TODO use SpanContext here
```
sophokles73: I have merged the PR adding the *get* method variant accepting the span context. Do you want to rebase and update this right away?

calohmn: Yes, I'll do that.

calohmn:
@sophokles73 I've had a look at refactoring the `applyDelegatedMessageDeliveryResultToCommandContext` method, using a new `CommandContext#disposition` method instead.
One question about that:
In [this commit](https://github.com/bsinno/hono/commit/e8b2664c0b4a88c280bddba03bb49eec36821619#diff-3f9ca2cc5e5c5115d5950a75d5c45964R170), the `released` method was changed to *not* set the error tag on the span anymore (only an error log entry). What was the motivation behind this? Shouldn't we treat this as an overall error in the trace if the command message *was not and will not be acted upon* by the device?

client/src/main/java/org/eclipse/hono/client/impl/GatewayMapperImpl.java
calohmn: Done.

sophokles73:
@calohmn regarding the logging on the Span in case of the released outcome:
I cannot recall if I deliberately made the decision to not set the error tag. However, to me it actually looks as if I actually wanted to mark the Span as erroneous but used the wrong approach to do so, i.e. log the ERROR key instead of setting the ERROR tag. I agree that we should mark the Span as erroneous instead, though. Can you add a separate commit to your PR fixing this?

calohmn:
@sophokles73 ok, I've added a commit for this.
(I've also fixed tracing in `CommandContext#reject()` where the "rejected command for device" message was not ending up in the trace log).

calohmn:
I'll refactor `DelegateViaDownstreamPeerCommandHandler` (replacing `applyDelegatedMessageDeliveryResultToCommandContext` with a new `CommandContext#disposition` method) in a follow-up PR.

PR #1228, opened 2019-05-09
Add a basic Sigfox protocol adapter

ctron:
This adds a first version of a Sigfox protocol adapter. The current limitations are:

* ~~Telemetry only~~
* Supports only `DATA` callbacks
* Does not yet support command and control

I would consider this "tech preview" for the moment.

### DIFF
--- a/adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/Config.java
+++ b/adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/Config.java
@@ -20,6 +20,7 @@ import org.eclipse.hono.client.RequestResponseClientConfigProperties;
 import org.eclipse.hono.config.ClientConfigProperties;
 import org.eclipse.hono.service.AbstractAdapterConfig;
 import org.eclipse.hono.service.metric.MetricsTags;
+import org.eclipse.hono.util.Constants;
 import org.springframework.beans.factory.config.ObjectFactoryCreatingFactoryBean;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.context.annotation.Bean;
@@ -37,7 +38,6 @@ public class Config extends AbstractAdapterConfig {
 
     private static final String CONTAINER_ID_HONO_SIGFOX_ADAPTER = "Hono Sigfox Adapter";
     private static final String BEAN_NAME_SIGFOX_PROTOCOL_ADAPTER = "sigfoxProtocolAdapter";
-    private static final String METRICS_PROTOCOL_TAG = "sigfox";
 
     @PostConstruct
     void validateConfiguration() {
@@ -130,7 +130,7 @@ public class Config extends AbstractAdapterConfig {
     @Bean
     public MeterRegistryCustomizer<MeterRegistry> commonTags() {
         return r -> r.config().commonTags(
-                MetricsTags.forProtocolAdapter(METRICS_PROTOCOL_TAG));
+                MetricsTags.forProtocolAdapter(Constants.PROTOCOL_ADAPTER_TYPE_SIGFOX));
     }
 
 }
--- a/adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/Config.java
+++ b/adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/Config.java
@@ -1,0 +1,150 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.adapter.sigfox.impl;
+
+import javax.annotation.PostConstruct;
+
+import org.eclipse.hono.adapter.http.HttpProtocolAdapterProperties;
+import org.eclipse.hono.client.RequestResponseClientConfigProperties;
+import org.eclipse.hono.config.ClientConfigProperties;
+import org.eclipse.hono.service.AbstractAdapterConfig;
+import org.eclipse.hono.service.metric.MetricsTags;
+import org.springframework.beans.factory.config.ObjectFactoryCreatingFactoryBean;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.cache.CacheManager;
+import org.springframework.cache.annotation.EnableCaching;
+import org.springframework.cache.concurrent.ConcurrentMapCacheManager;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Scope;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.spring.autoconfigure.MeterRegistryCustomizer;
+
+/**
+ * Spring Boot configuration for the Sigfox adapter.
+ */
+@Configuration
+@EnableCaching
+public class Config extends AbstractAdapterConfig {
+
+    private static final String CONTAINER_ID_HONO_SIGFOX_ADAPTER = "Hono Sigfox Adapter";
+    private static final String BEAN_NAME_SIGFOX_PROTOCOL_ADAPTER = "sigfoxProtocolAdapter";
+    private static final String METRICS_PROTOCOL_TAG = "sigfox";
+
+    @PostConstruct
+    void validateConfiguration() {
+        final HttpProtocolAdapterProperties httpProtocolAdapterProperties = adapterProperties();
+        if (!httpProtocolAdapterProperties.isAuthenticationRequired()) {
+            throw new IllegalStateException(
+                    "LoRa Protocol Adapter does not support unauthenticated mode. Please change your configuration accordingly.");
+        }
+    }
+
+    /**
+     * Creates a new LoRa adapter instance.
+     * 
+     * @return The new instance.
+     */
+    @Bean(name = BEAN_NAME_SIGFOX_PROTOCOL_ADAPTER)
+    @Scope("prototype")
+    public SigfoxProtocolAdapter sigfoxProtocolAdapter() {
+        return new SigfoxProtocolAdapter();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected void customizeDownstreamSenderFactoryConfig(final ClientConfigProperties props) {
+        if (props.getName() == null) {
+            props.setName(CONTAINER_ID_HONO_SIGFOX_ADAPTER);
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected void customizeRegistrationClientFactoryConfig(final RequestResponseClientConfigProperties props) {
+        if (props.getName() == null) {
+            props.setName(CONTAINER_ID_HONO_SIGFOX_ADAPTER);
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected void customizeCredentialsClientFactoryConfig(final RequestResponseClientConfigProperties props) {
+        if (props.getName() == null) {
+            props.setName(CONTAINER_ID_HONO_SIGFOX_ADAPTER);
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected void customizeTenantClientFactoryConfig(final RequestResponseClientConfigProperties props) {
+        if (props.getName() == null) {
+            props.setName(CONTAINER_ID_HONO_SIGFOX_ADAPTER);
+        }
+    }
+
+    /**
+     * Exposes the LoRa adapter's configuration properties as a Spring bean.
+     *
+     * @return The configuration properties.
+     */
+    @Bean
+    @ConfigurationProperties(prefix = "hono.http")
+    public HttpProtocolAdapterProperties adapterProperties() {
+        return new HttpProtocolAdapterProperties();
+    }
+
+    /**
+     * Exposes a factory for creating LoRa adapter instances.
+     *
+     * @return The factory bean.
+     */
+    @Bean
+    public ObjectFactoryCreatingFactoryBean serviceFactory() {
+        final ObjectFactoryCreatingFactoryBean factory = new ObjectFactoryCreatingFactoryBean();
+        factory.setTargetBeanName(BEAN_NAME_SIGFOX_PROTOCOL_ADAPTER);
+        return factory;
+    }
+
+    /**
+     * Gets the cache manager.
+     *
+     * @return the cache manager
+     */
+    @Bean
+    public CacheManager cacheManager() {
+        return new ConcurrentMapCacheManager();
+    }
+
+    /**
+     * Customizer for meter registry.
+     *
+     * @return The new meter registry customizer.
+     */
+    @Bean
+    public MeterRegistryCustomizer<MeterRegistry> commonTags() {
+        return r -> r.config().commonTags(
+                MetricsTags.forProtocolAdapter(METRICS_PROTOCOL_TAG));
+    }
+
+}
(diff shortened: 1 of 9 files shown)


adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/Config.java
```diff
@@ -0,0 +1,150 @@
...
+import org.springframework.context.annotation.Scope;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.spring.autoconfigure.MeterRegistryCustomizer;
+
+/**
+ * Spring Boot configuration for the Sigfox adapter.
+ */
+@Configuration
+@EnableCaching
```
sophokles73: what is this good for?

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/SigfoxProtocolAdapter.java
```diff
@@ -0,0 +1,176 @@
...
+            downstreamMessage.getApplicationProperties().getValue().put(SIGFOX_PROPERTY_PREFIX + key, entry.getValue());
+        }
+
+    }
+
+    private static Buffer decode(final String data) {
+        if (data == null) {
+            return Buffer.buffer();
+        }
+        return Buffer.buffer(BaseEncoding.base16().decode(data.toUpperCase()));
```
sophokles73: This is nice :+1:  I didn't know that guava contains a hex decoder. We can then also use it in the LoRaWan adapter and get rid of `org.eclipse.hono.util.Strings.decodeHex` ...

site/content/user-guide/sigfox-adapter.md
```diff
@@ -0,0 +1,126 @@
...
+a username of `sigfox` and a password of `test12`:
+
+~~~json
+{
+  "auth-id": "sigfox",
+  "device-id": "sigfox-backend",
+  "enabled": true,
+  "secrets": [
+    {
+      "hash-function": "sha-512",
```
sophokles73: please use the clear text password and let the registry create the (bcrypt) hash instead

site/content/user-guide/sigfox-adapter.md
```diff
@@ -0,0 +1,126 @@
...
+
+Log in to the Sigfox backend at https://backend.sigfox.com and then open up
+the view `Device Type` -> `Callbacks`.
+
+Create a new "Custom" callback, with the following settings:
+
+* **Type**: `DATA` – `UPLINK`
+* **Channel**: `URL`
+* **Url pattern**: `https://iot-sigfox-adapter.my.hono/data/telemetry?device={device}&data={data}`
+* **Use HTTP Method**: `GET`
```
sophokles73: can't we PUT the data to `telemetry/{tenant}/{device} ?

site/content/user-guide/sigfox-adapter.md
```diff
@@ -0,0 +1,126 @@
...
+
+As a workaround, you can explicitly set the `Authorization` header to a
+value of `Basic <base64 encoded credentials>`. You can encode the
+credentials using:
+
+~~~sh
+echo -n "sigfox@tenant:password" | base64
+~~~
+
+The get the full value, including the `Basic` you may use:
```
sophokles73: To get the full ...

site/content/user-guide/sigfox-adapter.md
ctron: Sure we could. The default selection in the Sigfox UI is "GET" however. The protocol adapter actually implements all three (GET, PUT, POST) with the same operation. And the proposal is also to use query parameters.

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/Config.java
ctron: I have no idea, copied it over from the Lora adapter.

site/content/user-guide/sigfox-adapter.md
ctron: Actually it is not possible to send the `data` argument in the request body. So even using PUT, we would need to use the query parameter.

site/content/user-guide/sigfox-adapter.md
ctron: Done.

site/content/user-guide/sigfox-adapter.md
ctron: Done.

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/SigfoxProtocolAdapter.java
ctron: I will do that in another PR.

sophokles73: I have already pushed that change ...

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/Config.java
ctron: It seems to enable support for caching. https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-caching.html

site/content/user-guide/sigfox-adapter.md
sophokles73: I see :-(

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/SigfoxProtocolAdapter.java
ctron: Ok … just saw your commit :+1:

site/content/user-guide/sigfox-adapter.md
ctron: I have to correct myself, it looks like for `POST` and `PUT` it is possible to transmit this using `application/x-www-form-urlencoded`. Not sure this is an improvement though. I will play around with this a bit and see what is actually possible.

In any case, I think we could improve this to:

    /data/{device}/telemetry?data={data}

Considering other URLs, for other options/callbacks:

    /data/{device}/event?data={data}
    /error/{device}/event?data={data}
    /status/{device}/event?info={info}

I don't think we need the tenant information a second time, as it is already included in the authentication credentials.

ctron: So it turns you actually can specify in more detail what the "body" section may look like when choosing PUT or POST and putting in the different values for the content type.

However, that would mean you need to put in part of the information (device) in the URL and part into the body. But first remove all the default information from the Web UI.

So I think the most simple approach is to keep the current way of passing this via query parameters.

sophokles73: what is wrong with using
```
PUT /telemetry/{tenantId}/{deviceId}?data=....
```
This is what we defined for publishing data using the gateway mode (which is what the SigFox network provider is, from Hono's point of view).

ctron: Sure, I can change that. Why do we need the tenant a second time?

sophokles73: That's how we defined the URI for the gateway mode. A gateway can act on behalf of a device, which is identified uniquely by its tenant and device ID. We currently only allow gateways to act on behalf of devices of the same tenant as the gateway, but that is just current implementation ...

ctron: But the tenant is already part of the authentication information of the gateway.

ctron: What I also miss with this pattern is a way to specify which callback this covers. Sigfox has multiple callbacks, this currently only implement `DATA`, thus the `data` prefix.

sophokles73: as I said, we currently verify if the tenant of the authenticated device (the gateway, here: SigFox back end) matches the tenant of the device that the gateway wants to act on behalf of.

sophokles73: > What I also miss with this pattern is a way to specify which callback this covers. Sigfox has multiple callbacks, this currently only implement DATA, thus the data prefix.

The data callback seems to map nicely to telemetry, doesn't it?
Which other callbacks need to be handled?

ctron: DATA maps to telemetry and event.

There are currently 8 callback types. I think the others that may make sense at the moment are: ERROR, SERVICE/STATUS, SERVICE/DATA_ADVANCED.

ctron: > as I said, the protocol adapters verify that the tenant of the authenticated device (the gateway, here: SigFox back end) matches the tenant of the device that the gateway wants to act on behalf of. However, that may change in the future ...

Ok, but that is already done without the need for a second tenant id (`gatewayDevice.getTenantId()`). Offering a second place to configure the tenant ID will only allow misconfigurations.

sophokles73: The tenant that you get with `gatewayDevice.getTenantId()` is the tenant of the device that has authenticated. The tenant in the URI is the tenant of the device that the GW wants to act on behalf of. These are **not necessarily** the same ...

ctron: I just tried to configure the backend with `PUT`. However this requires to put in a JSON encoded payload, which has to be configuring using a JSON template in the backend. It is not possible to use `application/octet-stream`.

I don't see any benefit in wrapping the while hex encoded data into a JSON object, just to remove that later on.

Could we just stick to the default of using `GET` with the `data` parameter?

ctron: > The tenant that you get with `gatewayDevice.getTenantId()` is the tenant of the device that has authenticated. The tenant in the URI is the tenant of the device that the GW wants to act on behalf of. These are **not necessarily** the same ...

So just to be sure I understood correctly. You want me to add the tenant into the path, and then fail the message later on, if the tenant of the credentials doesn't match tenant of the URI?!  As currently we don't have any way of granting cross-tenant posting.

ctron: I just tried to using path parameters, that doesn't work with the backend either. So, the only way use this in a sane and working way is to use GET with query parameters.

sophokles73: > So just to be sure I understood correctly. You want me to add the tenant into the path, and then fail the message later on, if the tenant of the credentials doesn't match tenant of the URI?! As currently we don't have any way of granting cross-tenant posting.

I do not want you to do anything. I just wanted to explain the way that the gateway mode works. We use this pattern in all the other protocol adapters. You might even call this a *design principle* :-)

If this protocol adapter doesn't need it, e.g. because it should assume that the device ID provided in the path belongs to the same tenant as the authenticated device, then be my guest to omit it ...

sophokles73: > I don't see any benefit in wrapping the while hex encoded data into a JSON object, just to remove that later on.

Neither do I

> Could we just stick to the default of using GET with the data parameter?

Sure

sophokles73: Having the tenant in the URI path is also necessary, if the client is not authenticated. A use case that we still support by means of the corresponding configuration property in all protocol adapters.
If the client is not authenticated, then you definitely need to pass in the tenant of the device because you cannot determine it as part of the authentication process.
So, if the SigFox adapter does not support anonymous clients then I think we would not need the tenant in the path. However, in order to provide consistent behavior across all protocol adapters, I think it should ...

ctron: Ok, I will simply put it in.

ctron: Ok, the tenant ID is now part of the URL.

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/Config.java
```diff
@@ -0,0 +1,150 @@
...
+public class Config extends AbstractAdapterConfig {
+
+    private static final String CONTAINER_ID_HONO_SIGFOX_ADAPTER = "Hono Sigfox Adapter";
+    private static final String BEAN_NAME_SIGFOX_PROTOCOL_ADAPTER = "sigfoxProtocolAdapter";
+    private static final String METRICS_PROTOCOL_TAG = "sigfox";
```
sophokles73: in analogy to our other adapters, I think that this constant should be defined in `org.eclipse.hono.util.Constants` instead:
```
public static final PROTOCOL_ADAPTER_TYPE_SIGFOX = "hono-sigfox";
```

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/Config.java
```diff
@@ -0,0 +1,150 @@
...
+        return factory;
+    }
+
+    /**
+     * Gets the cache manager.
+     *
+     * @return the cache manager
+     */
+    @Bean
+    public CacheManager cacheManager() {
```
sophokles73: I do not see where this is used. Can you try to remove it (along with the `@EnableCaching` annotation at the class level) and see if it has any impact? FMPOV it shouldn't ...

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/SigfoxProtocolAdapter.java
```diff
@@ -0,0 +1,193 @@
...
+import io.vertx.ext.web.handler.ChainAuthHandler;
+import io.vertx.ext.web.handler.CorsHandler;
+
+/**
+ * A Vert.x based Hono protocol adapter for receiving HTTP push messages from and sending commands to the Sigfox
+ * backend.
+ */
+public final class SigfoxProtocolAdapter extends AbstractVertxBasedHttpProtocolAdapter<HttpProtocolAdapterProperties> {
+
+    private static final String TYPE_NAME = "hono-sigfox";
```
sophokles73: this should be defined in `org.eclipse.hono.util.Constants` (see previous comment) ...

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/Config.java
ctron: Done.

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/Config.java
```diff
@@ -0,0 +1,150 @@
...
+     * @return The configuration properties.
+     */
+    @Bean
+    @ConfigurationProperties(prefix = "hono.http")
+    public HttpProtocolAdapterProperties adapterProperties() {
+        return new HttpProtocolAdapterProperties();
+    }
+
+    /**
+     * Exposes a factory for creating LoRa adapter instances.
```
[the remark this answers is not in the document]
ctron: Done.

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/Config.java
```diff
@@ -0,0 +1,150 @@
...
+     */
+    @Override
+    protected void customizeTenantClientFactoryConfig(final RequestResponseClientConfigProperties props) {
+        if (props.getName() == null) {
+            props.setName(CONTAINER_ID_HONO_SIGFOX_ADAPTER);
+        }
+    }
+
+    /**
+     * Exposes the LoRa adapter's configuration properties as a Spring bean.
```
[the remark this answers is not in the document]
ctron: Done.

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/SigfoxProtocolAdapter.java
ctron: Done.

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/Config.java
ctron: Done. Indeed it doesn't.

Issue #1230, opened 2019-05-13
Remove token from assert Registration response


sophokles73:
The protocol adapters invoke the *assert Registration* operation on the Device Registration service in order to make sure that a given device is registered and enabled. In the past, the Device Registration service was required to return a signed JWT including the device and tenant ID in its response. This token had been forwarded to the former *Hono Messaging* component which used it to verify the device's registration status.

Since this component is no longer part of Hono, we should remove the token from the response message, thus reducing the computational overhead in implementations.

sophokles73:
fixed in https://github.com/eclipse/hono/commit/de1e04549fa9f6d0c2cc86eab93eec83acb6fff6

PR #1232, opened 2019-05-13
[#1230] Remove assertion token.

sophokles73:
The assert Device Registration of the Device Registration API no longer
includes an assertion token in its response message. The base classes
for implementing a Device Registration service therefore no longer
creates a JWT for inclusion in the message.


sophokles73:
@dejanb the places you are referring to are not about the token used to assert a device's registration status but are about the token issued for a Hono component when it connects to another component.
There was, however, one other place in the docs where I removed a reference to the assertion token.

Issue #1073, opened 2019-03-01
Potentially competing receiver links when using command and control API


DanielMaier-BSI:
Command and control API uses address `control/${tenant_id}/${reply_id}` for receiving responses in consumer applications. Users are free to choose `${reply_id}`. If an application wants to receive all responses of one tenant, hono docs recommend to use an arbitrary (short) string as `${reply_id}` without `${device_id}` and to keep the receiver link open. 

However, protocol adapters use a very similar addressing scheme to receive commands from applications in the form of `control/${tenant_id}/${device_id}`. If `${device_id}` == `${reply_id}` we now have two competing receiver links and messaging network can not distinguish if a message should be sent to protocol adapter or to the application.

One possible solution would be to change addressing scheme for responses to 
`response/${tenant_id}/${reply_id}`. Of course this would require an API change.

WDYT?

calohmn:
I would like to see the "control" keyword in the address.

Therefore, I think we could define as address schemes:

`control/request/${tenant_id}/${device_id}` for the application to send commands on.
`control/response/${tenant_id}/${reply_id}` for the application to receive replies on.

Thinking about the necessary permissions for such addresses in EnMasse/Qpid/Keycloak, these Keycloak groups would have to be assigned to the user:

`send_control/request/${tenant_id}`
`recv_control/response/${tenant_id}`

sophokles73:
All addresses in Hono are built around the assumption that the first segment represents the endpoint, the second segment indicates the tenant and the third segment indicates the device ID. If we want to diverge from this pattern in the C&C API  then we will have a lot of work to do, introducing special cases all over Hono. I'd rather not do that, so introducing an additional path segment in the addresses is not an option FMPOV.

DanielMaier-BSI:
What about something like this:
`control_request/${tenant_id}/${reply_id}`
`control_response/${tenant_id}/${reply_id}`
?

sophokles73:
I think we should wait for the results of the discussion around #930 as it might have an impact on the addressing scheme used for C&C as well.

calohmn:
As the discussion for #930 is mostly done, I think we can define the address patterns here now.

I would be ok with the last mentioned suggestion:
````
control_request/${tenant_id}/${device_id}
control_response/${tenant_id}/${reply_id}
````
Sender link for the command requests would be on `control_request/${tenant_id}`.

For the potential additional message transfer as shown in #930 (regarding routing of commands mapped to gateways), I would suggest:
````
[...]

PR #1237, opened 2019-05-14
[#930] Let command client use a tenant-scoped sender link address

calohmn:
This concludes the implementation for #930:
The command client will now create a sender link on the tenant-scoped address. The `To` property of the command message is set to the device-specific address.

Documentation changes and additions to the release notes will go in a separate PR.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
@@ -80,10 +80,15 @@ public abstract class AbstractRequestResponseClient<R extends RequestResponseRes
                             HttpURLConnection.HTTP_GONE
     };
 
+    /**
+     * Target address of the request message.
+     * Note that the target address of the sender link may be different, see {@link #getLinkTargetAddress()}.
+     */
+    protected final String targetAddress;
+
     private final Map<Object, TriTuple<Handler<AsyncResult<R>>, Object, Span>> replyMap = new HashMap<>();
     private Handler<Void> drainHandler;
     private final String replyToAddress;
-    private final String targetAddress;
     private final String tenantId;
 
     /**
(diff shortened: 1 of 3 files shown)


calohmn:
Note that the address prefix here is still `control` - not `control_request` as discussed in #1073. 
I would suggest doing that change in a subsequent PR, once the migration path (see https://github.com/eclipse/hono/issues/1073#issuecomment-492235224) has been agreed upon.

client/src/main/java/org/eclipse/hono/client/impl/AsyncCommandClientImpl.java
```diff
@@ -72,10 +72,27 @@ public String getEndpoint() {
         return CommandConstants.COMMAND_ENDPOINT;
     }
 
+    /**
+     * The target address of the command message.
```
sophokles73: What about 
> Gets the address to use for a command message to be sent to a device.

client/src/main/java/org/eclipse/hono/client/impl/AsyncCommandClientImpl.java
```diff
@@ -72,10 +72,27 @@ public String getEndpoint() {
...
+     * @param tenantId The tenant id.
+     * @param deviceId The device id.
+     * @return The target address of the command message.
+     */
     static String getTargetAddress(final String tenantId, final String deviceId) {
         return String.format("%s/%s/%s", CommandConstants.COMMAND_ENDPOINT, tenantId, deviceId);
     }
 
+    /**
+     * The target address of the sender link on which the command message is sent.
```
sophokles73: what about
> Gets the target address to use for creating this client's sender link.

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -486,6 +502,9 @@ private Message createMessage(final String subject, final Map<String, Object> ap
         final Message msg = ProtonHelper.message();
         final String messageId = createMessageId();
         AbstractHonoClient.setApplicationProperties(msg, appProperties);
+        if (!targetAddress.equals(getLinkTargetAddress())) {
```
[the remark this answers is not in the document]
calohmn: That is the case when `getLinkTargetAddress` has been overridden, like in the case of the `CommandClientImpl` subclass.

calohmn:
@sophokles73 I've updated the javadocs accordingly.

Issue #1238, opened 2019-05-15
Allow to enable SSL for Health Check Server by configuration


Christian-Schmid:
Hi, 
we use some additional endpoints in the ``VertxBasedHealthCheckServer`` which we want to secure by means of SSL like the main server. 

Therefore we extended the configuration options of the health check server to allow configuration of such. 

Maybe you are interested in such config option as well, I will create a PR :-)

sophokles73:
Not sure if I understand correctly. What "additional endpoints" are you talking about?

Christian-Schmid:
We want to use the Hono provided Prometheus metrics, they are exposed on that server. 
So far all our internal communication is secured by SSL, so we would like to secure the Prometheus metrics endpoint with SSL as well.

sophokles73:
I see, and yes, we would be very interested in a PR adding that functionality...

Christian-Schmid:
I created https://github.com/eclipse/hono/pull/1239 to show the functionality. If it is ok, i can add some integration tests.

sophokles73:
closed via #1349

PR #1239, opened 2019-05-15
[#1238]: Allow to enable SSL for Health Check Server by configuration

Christian-Schmid:
This pull request allows to configure the hono health check server to use SSL.

Signed-off-by: Christian Schmid <Christian.Schmid@bosch-si.com>

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/VertxBasedHealthCheckServer.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/VertxBasedHealthCheckServer.java
@@ -29,6 +29,7 @@ import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
 import io.vertx.core.http.HttpServer;
 import io.vertx.core.http.HttpServerOptions;
+import io.vertx.core.net.JksOptions;
 import io.vertx.ext.healthchecks.HealthCheckHandler;
 import io.vertx.ext.web.Router;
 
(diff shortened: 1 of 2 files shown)


service-base/src/main/java/org/eclipse/hono/service/VertxBasedHealthCheckServer.java
```diff
@@ -120,6 +121,15 @@ public void setAdditionalResources(final List<Handler<Router>> resourceProviders
         final HttpServerOptions options = new HttpServerOptions()
                 .setPort(config.getHealthCheckPort())
                 .setHost(config.getHealthCheckBindAddress());
+
+        if (config.isHealthCheckUseSsl()) {
+            options.setSsl(true);
+            final JksOptions keyOptions = new JksOptions();
```
ctron: It would be great if we could re-use the methods we already have for the key/cert material, and allow for PKS12 and PEM. And not hardcode JKS.

Christian-Schmid: True. I will have a look

mbaeuerle: Comming back to this:
At the moment the healthCheck related properties are set as children of `app` for example in the MQTT config, which is extended with the SSL properties:
```yaml
hono:
  app:
    maxInstances: 1
    healthCheckPort: ${vertx.health.port}
    healthCheckBindAddress: 0.0.0.0
    healthCheckUseSsl: true
    healthCheckKeyStorePath: "/etc/hono/auth-server-internal.p12"
    healthCheckKeyStorePassword: "auth-server-internal"
```
To make it work with different key store formats I suggest to move the healthCheck to its own property like so:
```yaml
hono:
  app:
    maxInstances: 1
  healthCheck:
    port: ${vertx.health.port}
    bindAddress: 0.0.0.0
    useSsl: true
    keyStorePath: "/etc/hono/auth-server.p12"
    keyStorePassword: "password"
```

Then we could introduce a new config `HealthCheckConfig` which would extend [`AbstractConfig`](https://github.com/eclipse/hono/blob/master/core/src/main/java/org/eclipse/hono/config/AbstractConfig.java).
With this in place setting the options would go something like that:
```java
if (healthCheckConfig.isUseSsl()) {
  options.setSsl(true);
  options.setKeyCertOptions(healthCheckConfig.getKeyCertOptions());
}
```

WDYT?

Christian-Schmid:
Closing this PR in favor of: https://github.com/eclipse/hono/pull/1342 (see discussion)

PR #1240, opened 2019-05-15
HTTP API changes and file based registry implementation

jbtrystram:
Hi,

This is a draft to implement the changes proposed by @ctron work in #1130 . This contains only the changes for the tenant service at the moment, so we can begin the discussion on the implementation details.
I added support for the HTTP header _If-Match_ in so the versions values can be passed to the device registry as parameters.

The other main change is that a tenant / device can now be created without specifying a _tenantID_ / _deviceID_. It make sense for the device registry to be responsible for picking a suitable value.

What's left to do : 

- [x] Tenant API
- [x] File Based Tenant Service
- [x] Device API
- [x] File Based Registration Service
- [x] Credential API
- [x] File Based Credential Service
- [x] Credential Service Unit tests
- [x] Tenant Service Unit tests
- [x] Registration Service Unit tests
- [ ] Documentation


service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -56,6 +57,10 @@
      * The name of the URI path parameter for the device ID.
      */
     protected static final String PARAM_DEVICE_ID = "device_id";
+    /**
+     * The key that is used to put the if-Match ETags values to the RoutingContext.
+     */
+    protected static final String KEY_ETAGS_VERSION = "KEY_ETAGS_VERSION";
```
ctron: We might want to name this a bit more neutral, maybe "resource-version".

jbtrystram: Note that it's only used internally in  the _HTTPEndpoint_ to store the values in the `RoutingContext`. I thought it made sense to match the header name, but both works :+1:

ctron: I don't have strong opinion here, both works for me.

ctron:
It looks like the `Signed-off-by` footer is missing.

jbtrystram:
> It looks like the `Signed-off-by` footer is missing.

i'll rebase, squash and sign anyway ;)

service-base/src/main/java/org/eclipse/hono/service/credentials/CompleteCredentialsService.java
```diff
@@ -105,37 +108,40 @@ default void getAll(final String tenantId, final String deviceId, final Span spa
...
      * @param resultHandler The handler to invoke with the result of the operation.
+     * @deprecated removed in hono 1.0 API.
      *         <p>
      *         The <em>status</em> will be
      *         <ul>
      *         <li><em>204 No Content</em> if the credentials matching the criteria have been removed.</li>
      *         <li><em>404 Not Found</em> if no credentials match the given criteria.</li>
      *         </ul>
-     * @throws NullPointerException if any of the parameters is {@code null}.
+     * @throws NullPointerException if any of the parameters except resourceVersion is {@code null}.
```
ctron: I think this is not completely correct. If the `resourceVersion` is `null` I would expect a NPE. However, if it is "empty" it should be "ok".

ctron:
Looks good to me. But I will leave this for others to comment.

service-base/src/main/java/org/eclipse/hono/service/credentials/CompleteCredentialsService.java
jbtrystram: Agreed. Since I changed to using Optional having a null value will raise NPE. It's updated.

core/src/main/java/org/eclipse/hono/util/RegistrationConstants.java
```diff
@@ -72,7 +72,7 @@
     /**
      * The name of the Device Registration API endpoint.
      */
-    public static final String REGISTRATION_ENDPOINT = "registration";
```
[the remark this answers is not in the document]
jbtrystram: I asked on the IRC channel expecting to trigger the discussion but nothing came out so far.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedTenantService.java
```diff
@@ -256,21 +259,29 @@ public void get(final X500Principal subjectDn, final Span span,
...
 
-    TenantResult<JsonObject> removeTenant(final String tenantId) {
+    TenantResult<JsonObject> removeTenant(final String tenantId, final Optional<String> resourceVersion) {
 
         Objects.requireNonNull(tenantId);
 
         if (getConfig().isModificationEnabled()) {
-            if (tenants.remove(tenantId) != null) {
+            if (tenants.containsKey(tenantId)) {
+                if (!tenants.get(tenantId).getResourceVersion().equals(resourceVersion.orElse(null))) {
```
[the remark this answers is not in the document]
jbtrystram: It does works from what I tested :)

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
```diff
@@ -453,6 +459,14 @@ public void update(final String tenantId, final JsonObject newCredentials, final
                     while (credentialsIterator.hasNext()) {
                         final JsonObject creds = (JsonObject) credentialsIterator.next();
                         if (creds.getString(CredentialsConstants.FIELD_TYPE).equals(type)) {
+                            final String currentVersion = creds.getString(CredentialsConstants.FIELD_RESOURCE_VERSION);
+                            if (resourceVersion.isPresent()){
+                                if (!currentVersion.equals(resourceVersion)) {
+                                    resultHandler.handle(Future.succeededFuture(CredentialsResult.from(HttpURLConnection.HTTP_NOT_FOUND)));
```
dejanb: 412 return code here as well?

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedTenantService.java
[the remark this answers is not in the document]
dejanb: My bad, it works as it's a negative comparison. Although it'd be a bit more readable if we used the same syntax with isPresent() like in the credentials service implementation.

core/src/main/java/org/eclipse/hono/util/RegistrationConstants.java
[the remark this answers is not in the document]
dejanb: I'm for it, as we do the refactoring now and it makes sense to align things toward the new spec. Maybe folks can put their objections here if they disagree.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedRegistrationService.java
```diff
@@ -331,7 +338,8 @@ public RegistrationResult addDevice(final String tenantId, final String deviceId
     }
 
     @Override
-    public void updateDevice(final String tenantId, final String deviceId, final JsonObject data, final Handler<AsyncResult<RegistrationResult>> resultHandler) {
+    public void updateDevice(final String tenantId, final String deviceId, final JsonObject data,
+            final Optional<String> resourceVersion, final Handler<AsyncResult<RegistrationResult>> resultHandler) {
```
dejanb: We're not handling resourceVersion here?

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedTenantService.java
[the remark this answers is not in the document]
jbtrystram: I struggled a lot to do something readable for this... any suggestion is welcome

core/src/main/java/org/eclipse/hono/util/RegistrationConstants.java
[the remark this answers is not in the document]
jbtrystram: I am for it as well , as it will help the API to match the concepts

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedTenantService.java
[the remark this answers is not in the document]
dejanb: Looks good

core/src/main/java/org/eclipse/hono/util/RegistrationConstants.java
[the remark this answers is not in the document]
dejanb: @sophokles73 any thoughts on this?

core/src/main/java/org/eclipse/hono/util/RegistrationConstants.java
```diff
@@ -72,7 +72,7 @@
...
+    public static final String REGISTRATION_ENDPOINT = "devices";
```
sophokles73: The value of this constant is defined by the Device Registration API. I do not think that we agreed to change that API, did we? Exposing the HTTP API on base URI `devices/` is another thing ...

jbtrystram: Ah, some comments on the PR mentioned how the tenant > devices > credentials hierarchy made more sense so I thought it was a general API change.

core/src/main/java/org/eclipse/hono/util/EventBusMessage.java
```diff
@@ -296,6 +296,15 @@ public String getDeviceId() {
         return getProperty(MessageHelper.APP_PROPERTY_DEVICE_ID);
     }
 
+    /**
+     * Gets the value of the object identifier property.
+     *
+     * @return The value or {@code null} if not set.
+     */
+    public String getObjectId() {
+        return getJsonPayload().getString(RequestResponseApiConstants.FIELD_OBJECT_ID);
```
jbtrystram: We could maybe refactor `setTenant` and  `setDeviceId` methods into this ?

service-base/src/main/java/org/eclipse/hono/service/tenant/CompleteBaseTenantService.java
```diff
@@ -300,15 +300,9 @@ private void addNotPresentFieldsWithDefaultValuesForAdapter(final JsonObject ada
...
-    /**
-     * {@inheritDoc}
-     *
-     * This default implementation simply returns an empty result with status code 501 (Not Implemented).
-     * Subclasses should override this method in order to provide a reasonable implementation.
-     */
     @Override
-    public void add(final Optional<String> tenantId, final JsonObject tenantObj, final Handler<AsyncResult<TenantResult<JsonObject>>> resultHandler) {
-        handleUnimplementedOperation(resultHandler);
+    public void add(Optional<String> tenantId, JsonObject tenantObj, Handler<AsyncResult<CreateResult>> resultHandler) {
```
sophokles73: final modifiers are missing

service-base/src/main/java/org/eclipse/hono/service/tenant/CompleteBaseTenantService.java
```diff
@@ -300,15 +300,9 @@ private void addNotPresentFieldsWithDefaultValuesForAdapter(final JsonObject ada
...
+        resultHandler.handle(Future.failedFuture(new ServerErrorException(HttpURLConnection.HTTP_NOT_IMPLEMENTED)));
```
sophokles73: what's wrong with *handleUnimplementedOperation*?

service-base/src/test/java/org/eclipse/hono/service/tenant/CompleteBaseTenantServiceTest.java
```diff
@@ -37,15 +35,16 @@
...
+import javax.security.auth.x500.X500Principal;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.util.Optional;
+import java.util.UUID;
```
sophokles73: this is the wrong import order, please use the order defined in `eclipse/hono.importorder`

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
```diff
@@ -37,6 +23,18 @@
...
+import org.eclipse.hono.service.EventBusService;
+import org.eclipse.hono.tracing.TracingHelper;
+import org.eclipse.hono.util.EventBusMessage;
+import org.eclipse.hono.util.MessageHelper;
+import org.eclipse.hono.util.TenantConstants;
+import org.eclipse.hono.util.TenantResult;
+
+import javax.security.auth.x500.X500Principal;
+import java.net.HttpURLConnection;
+import java.util.Objects;
```
sophokles73: this is the wrong import order, please use the order defined in `eclipse/hono.importorder`

core/src/main/java/org/eclipse/hono/util/CreateResult.java
```diff
@@ -0,0 +1,60 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
```
sophokles73: this is a new class, so it should only be 2019

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -105,6 +105,10 @@
      * reported the data belongs to.
      */
     public static final String APP_PROPERTY_TENANT_ID = "tenant_id";
+    /**
+     * The name of the AMQP 1.0 message application property containing the resource version expected by the client.
+     */
+    public static final String APP_PROPERTY_RESOURCE_VERSION = "resource-version";
```
sophokles73: please replace `-` with `_` so that it also works with JMS

core/src/main/java/org/eclipse/hono/util/RegistrationConstants.java
sophokles73: @jbtrystram please revert this change

core/src/main/java/org/eclipse/hono/util/TenantConstants.java
```diff
@@ -67,7 +67,7 @@
     /**
      * The name of the Tenant API endpoint.
      */
-    public static final String TENANT_ENDPOINT = "tenant";
+    public static final String TENANT_ENDPOINT = "tenants";
```
sophokles73: this endpoint is defined by the Tenant API and cannot be changed by the device registry's HTTP management API
please revert

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -112,6 +112,28 @@ public TenantObject setEnabled(final boolean flag) {
...
     }
 
+
+    /**
+     * Retrieve the resource version property.
+     *
+     * @return the resource version value as {@code String}.
+     */
+    @JsonIgnore
+    public String getResourceVersion() {
```
sophokles73: is the resource version also contained in the body?

service-base/src/main/java/org/eclipse/hono/service/tenant/CompleteBaseTenantService.java
jbtrystram: this has been fixed.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
jbtrystram: At this point yes.

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
jbtrystram: done

core/src/main/java/org/eclipse/hono/util/TenantConstants.java
jbtrystram: Done.

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
jbtrystram: fixed.

service-base/src/test/java/org/eclipse/hono/service/tenant/CompleteBaseTenantServiceTest.java
jbtrystram: fixed.

service-base/src/main/java/org/eclipse/hono/service/tenant/CompleteBaseTenantService.java
jbtrystram: @dejanb  defined a _CreateResult_ class to return the resource Version to the client, as defined in the new spec. This modifies the type of the handler to `Handler<AsyncResult<CreateResult>>` so _handleUnimplementedOperation_ is not compatible anymore as it requires `Handler<AsyncResult<TenantResult<JsonObject>>>_

core/src/main/java/org/eclipse/hono/util/RegistrationConstants.java
jbtrystram: done

core/src/main/java/org/eclipse/hono/util/CreateResult.java
jbtrystram: fixed.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
sophokles73: It looks like you also have the resource version in a dedicated field in the `EventBusMessage`. Why do you need to duplicate it? When the response arrives, the HTTP endpoint can simply read the resource version from that dedicated field, can't it?

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsHttpEndpoint.java
```diff
@@ -106,6 +98,9 @@ public String getName() {
         return CredentialsConstants.CREDENTIALS_ENDPOINT;
     }
 
+    /**
+     * @deprecated This is removed in the new API.
```
sophokles73: I do not think that anybody knows what *the new API* is ...

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsHttpEndpoint.java
```diff
@@ -135,6 +130,7 @@ private void updateCredentials(final RoutingContext ctx) {
         final String deviceId = payload.getString(CredentialsConstants.FIELD_PAYLOAD_DEVICE_ID);
         final String authId = payload.getString(CredentialsConstants.FIELD_AUTH_ID);
         final String type = payload.getString(CredentialsConstants.FIELD_TYPE);
+        final String resourceVersion = ctx.get(KEY_ETAGS_VERSION);
```
sophokles73: if we use the term *resource version* in our code, then IMHO we should use this name for the key as well, i.e. `KEY_RESOURCE_VERSION`. The fact that it is related to the ETAG value returned by the service and that its value has been retrieved from the If-Match header field should be an implementation detail of the *extractIfMatchVersionParam* method ...

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
sophokles73: FMPOV this should be named `KEY_RESOURCE_VERSION` as that is what we store under this key. The fact that it is determined from the If-Match header is an implementation details ...

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -227,4 +282,22 @@ protected final String getDeviceIdParam(final RoutingContext ctx) {
...
 
+    /**
+     * Check the ETags values from the HTTP if-Match header if they exist,
+     * and extract the values if this check was successful
+     * <p>
+     * The HTTP header "If-Match"is parsed and the values are put to the routingContext ctx with the
+     * key {@link #KEY_ETAGS_VERSION}.
+     *
+     * @param ctx The routing context of the request.
+     */
```
sophokles73: please declare the `NullPointerException` that it throws ..

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -227,4 +282,22 @@ protected final String getDeviceIdParam(final RoutingContext ctx) {
...
+    protected void extractIfMatchVersionParam(final RoutingContext ctx) {
+        final String ifMatchHeader = ctx.request().getHeader(HttpHeaders.IF_MATCH);
+        if (ifMatchHeader != null) {
```
sophokles73: we have a nice `Strings.isNullOrEmpty()` method for this kind of check ;-)

service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationHttpEndpoint.java
```diff
@@ -59,32 +58,41 @@ protected String getEventBusAddress() {
...
     @Override
     public void addRoutes(final Router router) {
 
         final String pathWithTenant = String.format("/%s/:%s", RegistrationConstants.REGISTRATION_ENDPOINT, PARAM_TENANT_ID);
-        // ADD device registration
-        router.route(HttpMethod.POST, pathWithTenant).consumes(HttpUtils.CONTENT_TYPE_JSON)
-                .handler(this::doRegisterDeviceJson);
-        router.route(HttpMethod.POST, pathWithTenant)
-                .handler(ctx -> HttpUtils.badRequest(ctx, "missing or unsupported content-type"));
+
```
sophokles73: my understanding is that you want to expose the API under base URI `/devices`, right? If so, then you should use the value returned by *getName* instead of `RegistrationConstants.REGISTRATION_ENDPOINT` when creating the routes ...

service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationHttpEndpoint.java
```diff
@@ -109,57 +117,41 @@ private void doGetDevice(final RoutingContext ctx) {
...
+            final JsonObject requestMsg = EventBusMessage.forOperation(RegistrationConstants.ACTION_REGISTER)
+                    .setTenant(tenantId)
+                    .setDeviceId(deviceId)
+                    .setJsonPayload(payload)
+                    .toJson();
+            sendAction(ctx, requestMsg, getDefaultResponseHandler(ctx,
+                    status -> status == HttpURLConnection.HTTP_CREATED,
+                    (response, responsePayload) -> response.putHeader(
+                            HttpHeaders.LOCATION,
+                            String.format("/%s/%s/%s", RegistrationConstants.REGISTRATION_ENDPOINT, tenantId,
```
sophokles73: use *getName* instead of `RegistrationConstants.REGISTRATION_ENDPOINT` when creating the *Location* value...

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantHttpEndpoint.java
```diff
@@ -72,45 +72,56 @@ public void addRoutes(final Router router) {
...
+        // ADD tenant with auto-generated ID
         router.post(path).handler(bodyHandler);
         router.post(path).handler(this::extractRequiredJsonPayload);
-        router.post(path).handler(this::checkPayloadForTenantId);
+        router.post(path).handler(this::extractIfMatchVersionParam);
         router.post(path).handler(this::addTenant);
 
         final String pathWithTenant = String.format("/%s/:%s", TenantConstants.TENANT_ENDPOINT, PARAM_TENANT_ID);
 
+        // ADD tenant
```
sophokles73: my understanding is that you want to expose the API under base URI `/tenants`, right? If so, then you should use the value returned by *getName* instead of `TenantConstants.TENANT_ENDPOINT` when creating the routes ...

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantHttpEndpoint.java
```diff
@@ -123,11 +134,12 @@ private void addTenant(final RoutingContext ctx) {
 
         final String tenantId = getTenantIdFromContext(ctx);
 
-        final String location = String.format("/%s/%s", TenantConstants.TENANT_ENDPOINT, tenantId);
+        final String location = String.format("/%s/", TenantConstants.TENANT_ENDPOINT);
```
sophokles73: use *getName* instead of `TenantConstants.TENANT_ENDPOINT` when creating *Location* value...

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantHttpEndpoint.java
```diff
@@ -123,11 +134,12 @@ private void addTenant(final RoutingContext ctx) {
...
 
         doTenantHttpRequest(ctx, tenantId, TenantConstants.TenantAction.add,
                 status -> status == HttpURLConnection.HTTP_CREATED,
-                response -> response.putHeader(HttpHeaders.LOCATION, location)
+                (response, payload) -> response.putHeader(HttpHeaders.LOCATION,
+                        location+payload.getString(TenantConstants.FIELD_PAYLOAD_TENANT_ID))
```
sophokles73: whitespaces

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
jbtrystram: updated to the suggested value

service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationHttpEndpoint.java
jbtrystram: Good catch, thanks !

core/src/main/java/org/eclipse/hono/util/TenantObject.java
jbtrystram: Yes. I was confused at the beginning between _RequestResponseResult_ and _EventBusMessage_ so i added the field in both. It's fixed now. That's said it was a nice to have for the _FileBasedTenantService_ but i worked around it.

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsHttpEndpoint.java
jbtrystram: what do you think about a different wording : 

@deprecated This is removed in the official HTTP API.

sophokles73: Do we currently have an *unofficial* API? This might be removed in Hono 1.0, if that is the version that you would like the new version of the HTTP API to be part of ...

jbtrystram: Yes it would be good to have this in Hono 1.0 as it will be a stable version, so I expect we won't be able to do a lot of change afterwards.

sophokles73:
@jbtrystram @dejanb 
not sure if you are aware of it, but the current implementation of the example registry is used for implementing both the AMQP 1.0 based API as well as the HTTP based API. As such, the payload returned by the service implementation via the event bus is the same for both API endpoints. When you now adapt the service implementation to use the payload format defined by the new HTTP API, you implicitly also change the payload format used by the AMQP 1.0 based API. The Device Registration API defines the *get registration info* operation and also defines the response payload format which I believe is different from the payload format defined by the new HTTP API ...

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedTenantService.java
```diff
@@ -460,11 +460,12 @@ public void update(final String tenantId, final JsonObject tenantSpec, final Opt
         }
     }
 
+    @SuppressWarnings("unchecked")
     private Tenant convertTenantObject(final TenantObject tenantObject) {
         return new Tenant()
                 .setEnabled(tenantObject.isEnabled())
                 .setExtensions((JsonObject) tenantObject.getProperty("ext"))
-                .setAdapters(tenantObject.getAdapterConfigurations())
+                .setAdapters(tenantObject.getAdapterConfigurations().getList())
```
jbtrystram: but if no adapter is specified you get NPE :) That why I added a second method in the first place

calohmn:
The merge conflicts result from [this commit](https://github.com/eclipse/hono/commit/b81f380fcfce3782b9bc72693d848913187440dc) which removes the update of the `last-via` device registration data property.

I've kept the changes as small as possible to reduce merge conflicts here. The remaining references to the `last-via` property can be removed once this PR here has been merged.

dejanb:
@sophokles73 This PR should be ready for review. Can you please start looking into it when you find some time?
I'd like to focus on the implementation of the current spec and features here. We can continue discussing further improvements in all areas in subsequent PRs after we finish with this one.

core/src/main/java/org/eclipse/hono/auth/SpringBasedHonoPasswordEncoder.java
```diff
@@ -45,8 +45,9 @@
  */
 public class SpringBasedHonoPasswordEncoder implements HonoPasswordEncoder {
 
+    public static final int DEFAULT_BCRYPT_STRENGTH = 10;
```
sophokles73: missing JavaDoc

core/src/main/java/org/eclipse/hono/util/CredentialsConstants.java
```diff
@@ -34,6 +34,10 @@
      * The name of the field that contains the secret(s) of the credentials.
      */
     public static final String FIELD_SECRETS                     = "secrets";
+    /**
+     * The name of the field that contains the extension fields.
+     */
+    public static final String FIELD_EXT                         = "ext";
```
sophokles73: the `CredentialsConstants` class is used to define the constants required for interacting with and implementing the Credentials API.
In the past, when we deliberately used the Credential API's content (payload) model for the management API as well, all the constants defined here were also used for implementing the management API.
However, with the clear distinction between the Credentials API and the HTTP management API, I would rather define the constants required for the HTTP management API in separate classes and not *mix* them with each other. That way we would remove all (unwanted) dependencies between these two APIs and would be free to evolve them separately without the need for refactoring the constant names and values.

WDYT?

core/src/main/java/org/eclipse/hono/util/CredentialsObject.java
```diff
@@ -94,7 +94,7 @@ public CredentialsObject setProperty(final String name, final Object value) {
      */
     @JsonIgnore
     public String getDeviceId() {
-        return (String) getProperty(CredentialsConstants.FIELD_PAYLOAD_DEVICE_ID);
+        return getProperty(CredentialsConstants.FIELD_PAYLOAD_DEVICE_ID, String.class);
```
sophokles73: same here: the CredentialsObject has been designed to represent the messages received from the Credentials API. Using it for the payload exchanged via the HTTP management API *might* work for now but will certainly require more and more special case handling once the two APIs divert even more. Therefore I would also suggest to use distinct value object classes for both APIs ...

core/src/main/java/org/eclipse/hono/util/DeviceManagementConstants.java
```diff
@@ -0,0 +1,71 @@
...
+package org.eclipse.hono.util;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Constants &amp; utility methods used throughout the Device Management API.
+ */
+public final class DeviceManagementConstants extends RequestResponseApiConstants {
```
sophokles73: this seems to be the place where I would like to see *all* constants relevant for the HTTP management API being defined ...

core/src/main/java/org/eclipse/hono/util/JsonBackedValueObject.java
```diff
@@ -93,17 +93,20 @@
...
+     * @param clazz The target type.
      * @param <T> The type of the property.
-     * @return The property value or the given default value if the property is not set
-     *         or is of an unexpected type.
+     * @return The property value or the given default value if the property is not set or is of an unexpected type.
      * @throws NullPointerException if any of parent or name are {@code null}.
      */
-    @SuppressWarnings("unchecked")
-    protected final <T> T getProperty(final JsonObject parent, final String name, final T defaultValue) {
+    protected static final <T> T getProperty(final JsonObject parent, final String name, final Class<T> clazz,
```
sophokles73: is there any advantage with always passing in the type?

core/src/main/java/org/eclipse/hono/util/CredentialsConstants.java
dejanb: Are you thinking also of duplicating all constants that are shared between two APIs?

core/src/main/java/org/eclipse/hono/util/CredentialsObject.java
dejanb: It's not used by design. For the purpose of HTTP API we created new value objects that matches the spec. Take a look at `CommonCredential`, `PasswordCredential`, etc in `org.eclipse.hono.service.management.credentials` package. If there's a specific case where they overlap we can take a look into that and refactor things.

core/src/main/java/org/eclipse/hono/util/CredentialsConstants.java
sophokles73: If we want a clean separation between the two, then I think we will have to, don't you think?
The fact that many of the fields used in the payload of the two APIs have the same name needs to be considered arbitrary/coincidental, doesn't it?

core/src/main/java/org/eclipse/hono/util/CredentialsObject.java
sophokles73: Then the changes to this class are not needed for the HTTP management API? If so, then I think they should go into a separate PR, right?

core/src/main/java/org/eclipse/hono/util/CredentialsConstants.java
dejanb: OK, sounds good to me. I'll do some refactoring in this area. Thanks

core/src/main/java/org/eclipse/hono/util/JsonBackedValueObject.java
ctron: Yes, the fact that it works. You cannot cast with a type argument.

sophokles73: I guess the problem with the existing code is that it never throws a `ClassCastException`, right?

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -100,6 +100,10 @@
      * reported the data belongs to.
      */
     public static final String APP_PROPERTY_TENANT_ID = "tenant_id";
+    /**
+     * The name of the AMQP 1.0 message application property containing the resource version expected by the client.
+     */
+    public static final String APP_PROPERTY_RESOURCE_VERSION = "resource_version";
```
sophokles73: do we actually ever have such an application property in an AMQP message anywhere?

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Constants &amp; utility methods used throughout the Device Management API.
+ */
+public final class RegistryManagementConstants extends RequestResponseApiConstants {
+
+    /**
+     * The AMQP 1.0 <em>subject</em> to use for the <em>assert device registration</em> operation.
```
sophokles73: this looks wrong

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public static final String ACTION_CREATE = "create";
+    /**
+     * The AMQP 1.0 <em>subject</em> to use for the <em>get registration information</em> operation.
```
sophokles73: this looks wrong

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public static final String ACTION_GET = "get";
+    /**
+     * The AMQP 1.0 <em>subject</em> to use for the <em>deregister device</em> operation.
```
sophokles73: this looks wrong

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public static final String ACTION_DELETE = "delete";
+    /**
+     * The AMQP 1.0 <em>subject</em> to use for the <em>update device registration</em> operation.
```
sophokles73: this looks wrong

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public static final String ACTION_UPDATE = "update";
+
+    /**
+     * The name of the Credentials Registration HTTP API endpoint.
```
sophokles73: this looks wrong

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public static final String DEVICES_HTTP_ENDPOINT = "devices";
+
+    /**
+     * The name of the Device Registration HTTP API endpoint.
```
sophokles73: this looks wrong

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public static final String CREDENTIALS_HTTP_ENDPOINT = "credentials";
+
+    /**
+     * The name of the that contains credentials object in event bus message.
```
sophokles73: name of the *property* that ...

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
```diff
@@ -369,7 +369,7 @@ public HonoConnection deviceConnectionServiceConnection() {
      * @return The properties.
      */
     @Qualifier(DeviceConnectionConstants.DEVICE_CONNECTION_ENDPOINT)
-    @ConfigurationProperties(prefix = "hono.device-connection")
+    @ConfigurationProperties(prefix = "hono.deviceConnection")
```
sophokles73: My understanding is that this needs to be kebab case with dashes.

service-base/src/main/java/org/eclipse/hono/service/AbstractApplication.java
```diff
@@ -95,212 +58,45 @@ public final void addServiceFactories(final Set<ObjectFactory<? extends Abstract
...
-    protected void preShutdown() {
-        // empty
+    @Override
+    protected Future<?> deployVerticles() {
+        // call into super ...
+        return super.deployVerticles()
+                // ... then deploy service verticles
+                .compose(s -> deployServiceVerticles(getConfig().getMaxInstances()))
+                // ... then deploy the required verticles
+                .compose(s -> deployRequiredVerticles(getConfig().getMaxInstances()));
```
sophokles73: shouldn't this be the other way around, i.e. first deploy the required verticles, then the service verticles?

service-base/src/main/java/org/eclipse/hono/service/AbstractBaseApplication.java
```diff
@@ -0,0 +1,277 @@
...
+import org.springframework.boot.ApplicationRunner;
+
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.Vertx;
+
+/**
+ * A base class for implementing Spring Boot applications.
+ * <p>
+ * Compared to the {@link AbstractApplication}, this class has no expectations towards implementations.
```
sophokles73: IMHO we should not add a dependency to a subclass here ...

service-base/src/main/java/org/eclipse/hono/service/credentials/BaseCredentialsService.java
```diff
@@ -45,165 +32,13 @@
...
  */
-public abstract class BaseCredentialsService<T> extends EventBusService<T> implements CredentialsService {
-
-    private static final String SPAN_NAME_GET_CREDENTIALS = "get Credentials";
-
-    @Override
-    protected String getEventBusAddress() {
-        return CredentialsConstants.EVENT_BUS_ADDRESS_CREDENTIALS_IN;
-    }
+@Deprecated
```
sophokles73: I think adding a `@deprecated` JavaDoc indicating what to use instead would be helpful

service-base/src/main/java/org/eclipse/hono/service/credentials/CompleteCredentialsService.java
```diff
@@ -26,6 +26,7 @@
  *
  * @see <a href="https://github.com/eclipse/hono/blob/1.0-M4/site/content/api/Credentials-API.md">Credentials API</a>
  */
+@Deprecated
```
sophokles73: I think adding a `@deprecated` JavaDoc indicating what to use instead would be helpful

service-base/src/main/java/org/eclipse/hono/service/credentials/CredentialsHttpEndpoint.java
```diff
@@ -37,6 +39,7 @@
  * It receives HTTP requests representing operation invocations and sends them to the address {@link CredentialsConstants#CREDENTIALS_ENDPOINT} on the vertx
  * event bus for processing. The outcome is then returned to the client in the HTTP response.
  */
+@Deprecated
```
sophokles73: I think adding a `@deprecated` JavaDoc indicating what to use instead would be helpful

service-base/src/main/java/org/eclipse/hono/service/credentials/EventBusCompleteCredentialsAdapter.java
```diff
@@ -0,0 +1,317 @@
...
+/**
+ * A base class for implementing {@link CredentialsService}s.
+ * <p>
+ * This base class provides support for receiving <em>Get</em> request messages
+ * via vert.x' event bus and routing them to specific methods accepting the
+ * query parameters contained in the request message.
+ *
+ * @param <T> The type of configuration class this service supports.
+ */
+@Deprecated
```
sophokles73: I think adding a `@deprecated` JavaDoc indicating what to use instead would be helpful

service-base/src/main/java/org/eclipse/hono/service/credentials/EventBusCredentialsAdapter.java
```diff
@@ -0,0 +1,208 @@
...
+import io.opentracing.Span;
+import io.opentracing.SpanContext;
+import io.opentracing.Tracer;
+import io.opentracing.tag.Tags;
+import io.vertx.core.Future;
+import io.vertx.core.Verticle;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * A base class for implementing {@link CredentialsService}s.
```
sophokles73: why doesn't the class implement `CredentialsService` then?

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -111,15 +122,54 @@ protected void extractRequiredJsonPayload(final RoutingContext ctx) {
...
+    /**
+     * Check the Content-Type of the request to be 'application/json' and extract the payload if this check was
+     * successful.
+     * <p>
+     * The payload is parsed to ensure it is valid JSON and is put to the RoutingContext ctx with the key
+     * {@link #KEY_REQUEST_BODY}.
+     *
+     * @param ctx The routing context to retrieve the JSON request body from.
+     */
+    protected void extractRequiredJsonPayload(final RoutingContext ctx) {
```
sophokles73: is this method supposed to be overrideen by subclasses? If yes, a corresponding comment would be nice, if not, it should be declared final ...

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -111,15 +122,54 @@ protected void extractRequiredJsonPayload(final RoutingContext ctx) {
...
+    /**
+     * Check the Content-Type of the request to be 'application/json' and extract the payload if this check was
+     * successful.
+     * <p>
+     * The payload is parsed to ensure it is valid JSON and is put to the RoutingContext ctx with the
+     * key {@link #KEY_REQUEST_BODY}.
+     *
+     * @param ctx The routing context to retrieve the JSON request body from.
+     */
+    protected void extractRequiredJsonArrayPayload(final RoutingContext ctx) {
```
sophokles73: is this method supposed to be overrideen by subclasses? If yes, a corresponding comment would be nice, if not, it should be declared final ...

service-base/src/main/java/org/eclipse/hono/service/management/Id.java
```diff
@@ -0,0 +1,48 @@
...
+ *******************************************************************************/
+
+package org.eclipse.hono.service.management;
+
+import java.util.Objects;
+
+/**
+ * An ID used in results.
+ */
+public class Id {
```
sophokles73: final ?

core/src/main/java/org/eclipse/hono/util/DeviceManagementConstants.java
dejanb: I tried to consolidate all constants here (and renamed it `RegistryManagementConstants` (better name ideas?). Note, that file based registry implementation will still use AMQP API constants and objects where applicable as underlying data structure is optmized for that.

core/src/main/java/org/eclipse/hono/util/JsonBackedValueObject.java
ctron: Correct. The type information `T` is gone during runtime and is effectively `Object`. Casting to `Object` will always work.

The compiler did complain about that "Type safety: Unchecked cast from Object to T", but that got suppressed by using `@SuppressWarnings("unchecked")`

This is basically the same case as #751 was, which got fixed by #757.

Searching through the code, I found at least one additional location, having the same issue.

sophokles73: Thanks for explaining. Given that this seems to be a bug and that it occurs in other places as well, I think we should fix this in a separate commit/PR. Would you like me to take care of that?

ctron: Fixing this in another PR will cause conflicts in this PR. I don't see any benefits in that.

My proposal would be to merge this PR, including this change here. And follow up with another PR, fixing the other locations.

sophokles73: I understand your desire for getting this PR over with. However, personally, I do not find it that easy to review a PR covering 170 files with that amount of changes.
I am therefore trying to offer some help in reducing the scope of this PR by means of pulling out changes that we need to make in any case and which are not strictly imposed by the functionality being added here. Hope that makes sense ...

ctron: I absolutely understand your concern. But rebasing this PR is quite time consuming as well.

So I am not looking to push approving this PR, but more reducing the effort that it would take to rebase it over and over again.

sophokles73:
@ctron @jbtrystram Am I right in assuming that all the existing integration tests for the registry have been moved unaltered to the *legacy* package?

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
ctron: This methods signature and concept has been copied over from `extractRequiredJsonPayload`. I can't find either (comment, final) there. What would it be?

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
ctron: That was an existing method, which hasn't been changed from an API perspective. So I can't really say.

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
ctron: Yes, that was brought in by the revert of commit 62e0555346215d49f446df5a763351822b7611a5, which did break the build and which also broke validation of this PR.

The next time I rebase this PR, I will drop the "revert" comment fb456eeaad2e81dea81b4e5151d4db31db9995b0.

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
ctron: It is used in the `EventBusMessage` and all constants there seem to come from the class `org.eclipse.hono.util.MessageHelper`. So it seems logical to add this there as well.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/Application.java
```diff
@@ -40,128 +36,52 @@
...
-        final Future<String> result = Future.future();
-        log.info("Starting registration service {}", registrationService);
-        getVertx().deployVerticle(registrationService, result);
-        return result;
-    }
+            for (final Verticle verticle : this.verticles) {
+                log.info("Deploying: {}", verticle);
+                final Future<String> result = Future.future();
+                getVertx().deployVerticle(verticle, result);
+            }
```
calohmn: A custom device registry component using this `Application` class as a blueprint might find itself in a situation where not *all* verticle beans in the componentScan should be deployed here.
Deploying the service instances one-by-one again in this method doesn't look so straightforward, the service interfaces (like `RegistrationService`) not being `Verticle` instances anymore (instead `BaseRegistrationService` needs to be used) and the service factory verticles (e.g. from `deviceRegistryAmqpServerFactory`) not being deployed via the base class anymore (the `Application` class here doesn't extend `AbstractApplication` with its `deployServiceVerticles` method).

I would find it clearer, if the `Application` class here would also extend `AbstractApplication` and not just deploy *all* verticle beans.

ctron: Extending from `AbstractApplication` does require at least one `ObjectFactory` to be present. Otherwise it will fail.

However the application at hand doesn't use anything of that. The question is, does this implementation service as a blueprint for writing Spring Boot applications? Or does it serve as an example device registry.

I think the previous model of the `AbstractApplication` was extremely hard to understand, and I hear quite some complaints about that.

Now I think: a) we have what is required for this specific task, and not over-engineer b) people can extend this in a spring-native way, as they want to use spring.

Right now it is simple to understand how this is wired up. Before it was not. And if people want more complexity, they can always introduce that. But I don't think we should create an uber-Application startup framework.

jbtrystram:
> @ctron @jbtrystram Am I right in assuming that all the existing integration tests for the registry have been moved unaltered to the _legacy_ package?

the assumption would be correct. That said, we may have made changes in the files before deciding to keep a _legacy_ version (this was intended to stay a reasonably sized PR ! :smile: ). We tested the legacy tests against the legacy registry and everything pass, though.

service-base/src/main/java/org/eclipse/hono/service/credentials/EventBusCompleteCredentialsAdapter.java
ctron: Honestly I am not sure what to put there. The explanation takes a bit longer, and is summarized in the release notes. Maybe putting a link to the release notes?

Issue #1245, opened 2019-05-20
Move update of "last_via" property out of "assertRegistration" method


calohmn:
The update of the `last_via` property is currently done as part of the `RegistrationService#assertRegistration` method (as defined in the javadoc).

It should be decoupled from that, allowing for more optimized implementations.

calohmn:
Without being tied to `DeviceRegistry#assertRegistration`, the update of the `last_via` property should then only be done when there is a command subscription request being handled.

In that regard, the term `last_via` may not fit very well anymore.

I would rather use `commandGateway` instead. That would mean adding these methods to `RegistrationService`:
````
Future<Void> updateCommandGateway(String tenantId, String deviceId, String gatewayId)
String getCommandGateway(String tenantId, String deviceId);
````

calohmn:
The solution here shall introduce a new `RegistrationService#updateDeviceLastVia` method and corresponding `RegistrationClient` methods, to be invoked by the Protocol adapters after the `assertRegistration` request was successful.

In order to be able to decide, whether the `updateDeviceLastVia` request is actually needed, `assertRegistration` should also return the `via` property of the device registration data.
Added #1248 for that.

calohmn:
The suggested changes for this issue concerning the Device Registration API (AMQP) are:
- New mandatory operation `Update last-used gateway for device` for the AMQP interface. 
In `RegistrationService` this looks like:
`Future<Void> updateLastUsedGateway(String tenantId, String deviceId, String gatewayId, Span span)`
- New mandatory operation `Get last-used gateway for device` for the AMQP interface. 
In `RegistrationService` this looks like:
`Future<JsonObject> getLastUsedGateway(String tenantId, String deviceId, Span span);`
The returned JSON value would contain the gateway device id and the update date.
- With the introduction of the above method, the existing `Get Registration Information` operation of the AMQP interface isn't needed anymore in Hono code, therefore we could mark it as deprecated in M4 and remove it in Hono 1.0 (along with removing all the other non-mandatory operations).

In the `RegistrationClient` interface 2 new corresponding methods get added.
(With the introduction of the Infinispan-based Device Registry we could use an adapted RegistrationClient that communicates via the HotRod protocol with the Infinispan backend for the implementation of these methods.)

@sophokles73, @ctron thoughts on this?

sophokles73:
sounds good to me

PR #1248, opened 2019-05-20
[#1245] Add indication of GW usage to assertRegistration response JSON

calohmn:
This is for #1245.

The intention of #1245 is that the update of the `last_via` property shall be done in a separate request. In order to decide whether we actually need to do that request, this PR changes the `assertRegistration` return JSON to include the `via` property value.

### DIFF
--- a/site/content/api/Device-Registration-API.md
+++ b/site/content/api/Device-Registration-API.md
@@ -61,12 +61,14 @@ The body of the response message consists of a single *Data* section containing 
 | Name             | Mandatory | JSON Type     | Description |
 | :--------------- | :-------: | :------------ | :---------- |
 | *device-id*      | *yes*     | *string*      | The ID of the device that is subject of the assertion. |
+| *gw-supported*   | *no*      | *boolean*     | If set to `true`, the device allows one or more gateways to act on its behalf. These gateways are configured in the device registration data by means of the `via` property. |
 | *defaults*       | *no*      | *object*      | Default values to be used by protocol adapters for augmenting messages from devices with missing information like a *content type*. It is up to the discretion of a protocol adapter if and how to use the given default values when processing messages published by the device. |
 
 Below is an example for a payload of a response to an *assert* request for device `4711` which also includes a default *content-type*:
 ~~~json
 {
   "device-id" : "4711",
+  "gw-supported": true,
   "defaults": {
     "content-type": "application/vnd.acme+json"
   }
--- a/core/src/main/java/org/eclipse/hono/util/RegistrationConstants.java
+++ b/core/src/main/java/org/eclipse/hono/util/RegistrationConstants.java
@@ -64,6 +64,12 @@ public final class RegistrationConstants extends RequestResponseApiConstants {
     public static final String FIELD_LAST_VIA_UPDATE_DATE = "update-date";
 
     /**
+     * The name of the field in a response payload that contains the boolean value
+     * defining whether one or more gateways may act on behalf of the device.
+     */
+    public static final String FIELD_PAYLOAD_GATEWAY_SUPPORTED = "gw-supported";
+
+    /**
      * The name of the Device Registration API endpoint.
      */
     public static final String REGISTRATION_ENDPOINT = "registration";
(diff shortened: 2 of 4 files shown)


site/content/api/Device-Registration-API.md
```diff
@@ -58,15 +58,17 @@ A response to an *assertion* request contains the [Standard Response Properties]
...
 
-| Name             | Mandatory | JSON Type     | Description |
-| :--------------- | :-------: | :------------ | :---------- |
-| *device-id*      | *yes*     | *string*      | The ID of the device that is subject of the assertion. |
-| *defaults*       | *no*      | *object*      | Default values to be used by protocol adapters for augmenting messages from devices with missing information like a *content type*. It is up to the discretion of a protocol adapter if and how to use the given default values when processing messages published by the device. |
+| Name             | Mandatory | JSON Type         | Description |
+| :--------------- | :-------: | :---------------- | :---------- |
+| *device-id*      | *yes*     | *string*          | The ID of the device that is subject of the assertion. |
+| *via*            | *yes*     | *string or array* | Information about which gateways may act on behalf of the device. The property may either contain a single gateway device id as a string or a JSON array containing zero or more gateway ids as strings. In case no gateways are involved, an empty JSON array is to be used. |
```
sophokles73: why do you need the value itself in the protocol adapter? The authorization decision is made in the Device Registration service. My understanding is that you only need to know, whether the device may connect via a gateway at all, i.e. a boolean would be sufficient, or am I mistaken?

calohmn: Right, the value itself isn't needed. I would use a boolean property then.
As name, I think `via-enabled` would do (`access-via-gateway-enabled` or `via-gateway-usage-enabled` being a bit long).

sophokles73: what about `gw-supported`?

calohmn: yes, good. I've changed it to that.

core/src/main/java/org/eclipse/hono/util/RequestResponseApiConstants.java
```diff
@@ -54,7 +54,11 @@
      * The name of the property that contains the identifier of a tenant.
      */
     public static final String FIELD_PAYLOAD_TENANT_ID = Constants.JSON_FIELD_TENANT_ID;
-
+    /**
+     * The name of the property that contains the boolean value defining whether
+     * one or more gateways may act on behalf of the device.
+     */
+    public static final String FIELD_PAYLOAD_GATEWAY_SUPPORTED = "gw-supported";
```
sophokles73: this looks like it would better be placed into `RegistrationConstants` since it is very specific to the Device Registration API, isn't it?

site/content/api/Device-Registration-API.md
```diff
@@ -61,12 +61,14 @@ The body of the response message consists of a single *Data* section containing
 | Name             | Mandatory | JSON Type     | Description |
 | :--------------- | :-------: | :------------ | :---------- |
 | *device-id*      | *yes*     | *string*      | The ID of the device that is subject of the assertion. |
+| *gw-supported*   | *yes*     | *boolean*     | If set to `true`, the device allows one or more gateways to act on its behalf. These gateways are configured in the device registration data by means of the `via` property. |
```
sophokles73: I wonder if we should make this optional with a default value of `false`. That way, a registry that doesn't support gateway mode at all would not need to implement it. WDYT?

sophokles73:
@calohmn are you going to make any more changes to this PR?

core/src/main/java/org/eclipse/hono/util/RequestResponseApiConstants.java
calohmn: Right, I've changed that.

site/content/api/Device-Registration-API.md
calohmn: Well, as of now, the standard `BaseRegistrationService.getAssertionPayload()` method (marked final) will always add such a property anyway.
I can make the method used to get the value of the `gw-supported` property overridable, so that a device registry implementation could always return `false` there.
In that sense, I see not so much value in making the property optional, but have no problem with it either.

calohmn: Made `gw-supported` optional and made `isGatewaySupportedForDevice()` protected in `BaseRegistrationService`.

sophokles73:
@calohmn don't forget to document this API change in the release notes

PR #1251, opened 2019-05-22
Fix getPathWithoutBase(), extend javadocs and tests

calohmn:
`ResourceIdentifier.fromString("control///req/cmd-req-id").getPathWithoutBase()` 
was returning `null/req/cmd-req-id`.

With this fix it will return `/req/cmd-req-id`.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/ResourceIdentifier.java
+++ b/core/src/main/java/org/eclipse/hono/util/ResourceIdentifier.java
@@ -117,7 +117,9 @@ public final class ResourceIdentifier {
 
         final StringBuilder b = new StringBuilder();
         for (int i = startIdx; i < resourcePath.length; i++) {
-            b.append(resourcePath[i]);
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/eclipse/hono/util/ResourceIdentifier.java
```diff
@@ -286,6 +294,9 @@ public String getBasePath() {
 
     /**
      * Gets a string representation of the resource identifiers' parts without the base path.
+     * <p>
+     * E.g. for a resource <em>control/myTenant/deviceId/some/path</em>, the return value will be
+     * <em>deviceId/some/path</em>.
```
sophokles73: we should also document that this method returns an empty string (instead of `null`), if the identifier doesn't contain any additional path segments.

calohmn: done

PR #1253, opened 2019-05-22
[#1245] Extract methods for handling the bit flag used in the Command class

calohmn:
For #1245:
Extract methods for the bit flag used in the Command request id (denoting whether the original reply-to address contained the device id).

Encoding and decoding of the bit flag used in `Command.getDeviceFacingReplyToId`
and `CommandResponse.getReplyToId` has been refactored to use the new
methods as well. Note that for this, the bit flag value in these methods has been switched, making it identical to the flag value used for the Command request id.

This is a preparation for storing an additional option in the bit flag,
denoting whether the old or new command response address pattern was
used.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/Command.java
+++ b/client/src/main/java/org/eclipse/hono/client/Command.java
@@ -32,6 +32,12 @@ import io.vertx.core.buffer.Buffer;
 public final class Command {
 
     /**
+     * Bit flag value for the boolean option that defines whether the original reply-to address of the command message
+     * contained the device id.
+     */
+    private static final byte FLAG_REPLY_TO_CONTAINED_DEVICE_ID = 1;
+
+    /**
      * If present, the command is invalid.
      */
     private final Optional<String> validationError;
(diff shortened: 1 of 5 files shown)


calohmn:
(I'll push some more changes to fix the failing `VertxBasedAmqpProtocolAdapterTest.testUploadCommandResponseSucceeds` test.)

calohmn:
There was an illegal address used in the test. I've fixed that and added further validation logic and tests.

client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -349,14 +355,45 @@ public static String getRequestId(final String correlationId, final String reply
...
+        if (replyToContainedDeviceId) {
+            bitFlag |= FLAG_REPLY_TO_CONTAINED_DEVICE_ID;
         }
-        return String.format("%s%02x%s%s", removeDeviceFromReplyTo ? "1" : "0", stringOne.length(), stringOne,
-                stringTwo);
+        return String.valueOf(bitFlag);
+    }
+
+    /**
+     * Returns {@code true} if the original reply-to address of the command message contained the device id.
```
sophokles73: Can we agree to not describe the return value in the description of the method's purpose and do that in the `@return` tag instead? The description would then be something like
> Checks if the original reply-to address of the command message contained the device id.

client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -381,10 +418,12 @@ public String toString() {
...
      * @return The reply-to-id, starting with the device id.
      */
-    static String getDeviceFacingReplyToId(final String replyToId, final String deviceId) {
-        if (replyToId.startsWith(deviceId + "/")) {
-            return replyToId.replaceFirst(deviceId + "/", deviceId + "/0");
+    public static String getDeviceFacingReplyToId(final String replyToId, final String deviceId) {
+        final boolean replyToContainedDeviceId = replyToId.startsWith(deviceId + "/");
+        final String bitFlagString = encodeReplyToOptions(replyToContainedDeviceId);
+        if (replyToContainedDeviceId) {
+            return replyToId.replaceFirst(deviceId + "/", deviceId + "/" + bitFlagString);
```
sophokles73: if I understand the code correctly, then you are now adding a `1` in case the reply-to-id contains the device-id instead of adding a `0` as it was before. Is this so that you are able to use the same *encode*/*decode* methods?

client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -381,10 +418,12 @@ public String toString() {
...
         }
-        return String.format("%s/1%s", deviceId, replyToId);
+        return String.format("%s/%s%s", deviceId, bitFlagString, replyToId);
```
sophokles73: it doesn't seem to be necessary to check the *replyToContainedDeviceId* flag, is it? we can always return
`String.format("%s/%s%s", deviceId, bitFlagString, replyToId)`, can't we?

client/src/main/java/org/eclipse/hono/client/CommandResponse.java
```diff
@@ -88,7 +88,8 @@ public static CommandResponse from(
             return null;
         } else {
             try {
-                final boolean addDeviceIdToReply = "1".equals(requestId.substring(0, 1));
+                final String replyToOptionsBitFlag = requestId.substring(0, 1);
+                final boolean addDeviceIdToReply = Command.isReplyToContainedDeviceIdOptionSet(replyToOptionsBitFlag);
```
sophokles73: we now have a mutual dependency between `Command` and `CommandResponse` ...

client/src/main/java/org/eclipse/hono/client/Command.java
calohmn: Yes, that's the intention (see also PR description above).

client/src/main/java/org/eclipse/hono/client/Command.java
calohmn: If `replyToContainedDeviceId` is true, we would end up with having the device id in there twice, so some string manipulation for the `replyToContainedDeviceId` case is needed anyway (either cutting off the device id or squeezing in the bitFlagString as it is done now).

client/src/main/java/org/eclipse/hono/client/CommandResponse.java
calohmn: The `Command` class depends on `CommandResponse` only because of the reference in the `getDeviceFacingReplyToId` method javadoc (pointing to `CommandResponse#getReplyToId`).
I put that reference in there some time ago mainly to connect the two places that dealt with this bit flag logic.
Now that this logic is handled by common methods, the reference isn't needed so much anymore, I can remove it.

client/src/main/java/org/eclipse/hono/client/Command.java
calohmn: Ok, I've changed that accordingly.

client/src/main/java/org/eclipse/hono/client/Command.java
sophokles73: if *replyToId* starts with `$deviceId/` then we return
`$deviceId/$bitFlagString$remainderOfReplyToId`
if *replyToId* **does not** start with `$deviceId/` then $remainderOfReplyToId == $replyToId and we return
`$deviceId/$bitFlagString$replyToId`
which is the same as in the first case, isn't it?

calohmn: I guess you mean it like [this](https://github.com/eclipse/hono/compare/6f59777aebf9e7a2b32f40d198e2d1014a14ccc7..75bdeb227f3ab5d878a7cc2f0ba9c26ce5679217)? That's more concise, yes.

PR #1257, opened 2019-05-24
[#1245] Add Registration API methods concerning the last-used gateway.

calohmn:
This is for #1245:

Adds the `setLastUsedGateway` and `getLastUsedGateway` methods to the Device Registration API.

These methods are not used yet. This will come in a follow-up PR.


calohmn:
The implementation introduced a few weeks ago relies on updating the "last-used gateway" information via `assertRegistration` and reading it via `getDevice` (as part of the complete device data).
This causes this single "write-heavy" and rather transient piece of information to be mixed with the mostly "read-only" rest of the device data.

The motivation for #1245 was therefore to decouple both pieces of information, allowing the "last-used gateway" information for example to be stored in a separate Infinispan cache, independent of the device data returned by `getDevice`.
And this is what is being made possible in this PR, separate `setLastUsedGateway` and `getLastUsedGateway` methods, but still under the umbrella of the Device Registration API.

I didn't understand our discussion like we would want to go even further and introduce `setLastUsedGateway` and `getLastUsedGateway` outside of the Device Registration API entirely.
That would require a new `Gateway (usage) API` (or something like that). Is this what you had in mind?
To me that seems a bit overkill, seeing that these new methods are device-related and therefore do fit in the Device Registration API.

---
An alternative I could imagine would be introducing `setLastUsedGateway` and `getLastUsedGateway` only in a new `GatewayMappingDataClient` interface. We would provide a basic non-production/non-data-grid implementation that just calls the Device Registry `updateDevice` and `getDevice` methods, storing the data along with the device data.
Then an alternative client implementation could be plugged into the Protocol Adapter that for example talks directly with the Data Grid. 
*Advantage* of this:
No `setLastUsedGateway` and `getLastUsedGateway` methods in the Device Registry (which might not *need* to be mandatory anyway, as discussed [here](https://github.com/eclipse/hono/pull/1248#discussion_r286335843)).
*Disadvantage*:
We still need Device Registry `updateDevice` and `getDevice` methods for the basic non-data-grid solution (or something similar).

Seeing that we could also plug a special talk-directly-to-data-grid `RegistrationClient` implementation in case of the first solution above, I think I would favor the first solution.

sophokles73:
> The motivation for #1245 was therefore to decouple both pieces of information, allowing the "last-used gateway" information for example to be stored in a separate Infinispan cache, independent of the device data returned by getDevice.

So, is the idea then that one would implement the Device Registration API's *assert Registration* operation based on a persistent store like a DB whereas the *update Last-GW* and *get Last-GW* would be implemented based on a data grid (on the service implementation side)?

I can understand that you would want to use one or the other, but I wonder what the benefit of doing both in combination would be in this scenario. Decoupling *assert Registration* from the *update last GW* method also seems to make some sense wrt reducing the number of updates required and allowing the results of *assert Registration* to be cached more aggressively.

So, I didn't want to drive into your parade, I just wanted to make sure that we are all on the same page here. My understanding of this PR then is that we are **not** envisioning the case where the last GW data is maintained in a data grid that is independent from the Registration service implementation, right?

ctron:
> IIRC the discussion we had at our F2F, we actually didn't want to require this kind of information to be stored in the Device Registration service, did we? The whole argument was about being able to e.g. keep this information in a data grid that was independent from the Device Registration service. So i wonder why these operations now show up in the Device Registration API as mandatory to implement.
Or am I mistaken?

I was under the same impression. The may reason for us to split is this off is to **not** have this in the device registry API. However, to my understanding, this PR would actually include this as a mandatory operation for the registration API.

At least this should be an optional operation, so that we ensure that we **do not** flood the device registry with update operations, that would require to actually persist this information.

The idea is to provide an additional implementation in the future, using some data grid/caching backend, which allows us to keep updated on the device registration to an absolute minimum.

calohmn:
@ctron If not putting the `setLastUsedGateway` and `getLastUsedGateway` methods in the device registry API, where should they go? For the basic implementation that doesn't need a data grid/caching backend, the data has to be stored somewhere. And the straightforward place to use while not introducing a new component would be the device registry.

And concerning the idea of *only* having these methods in some `GatewayMappingDataClient` code in the protocol adapter (see my comment above), I still don't see that as the better alternative - we would explicitly need to document this for the case that someone wants to use their own device registry implementation but doesn't want to use the default of letting the `GatewayMappingDataClient` store the last-used gateway along with the device data.
Letting the custom device registry implementation just override the two methods looks way more straightforward.

And for our future Infinispan device registry along with a separate data grid: We could just leave these two methods unimplemented in the device registry implementation then, if we want to use a special `RegistrationClient` that stores the last-used gateway directly in the data grid.

> The may reason for us to split is this off is to *not* have this in the device registry API. [...]
At least this should be an optional operation, so that we ensure that we do not flood the device registry with update operations, that would require to actually persist this information.

A device registry implementation could store the last-used gateway independent of the device data, preventing any "flooding" effect on the data structure of the device data.

ctron:
I do understand that for the "demo"/"PoC" file based device registry, we would still need a way to store this. My concern is, that it seems we are optimizing this for exactly this use case, which I don't think should be our primary implementation. And while it might be a bit more complicated to use the file based device registry, I think it makes more sense seeing this as a different use case: storing the state of the device connectivity.

> A device registry implementation could store the last-used gateway independent of the device data, preventing any "flooding" effect on the data structure of the device data.

That would actually try to solve the issue after it was already created. Using a dedicated API we would never need to fix this up at a later point in time. So having a dedicated API for "device state" would prevent us from running into this issue in the first place.

sophokles73:
@ctron then what's your proposal? How does this dedicated API look like? Do you want to introduce a new (fourth) service API next to the Device Registration API?

ctron:
> @ctron then what's your proposal? How does this dedicated API look like? Do you want to introduce a new (fourth) service API next to the Device Registration API?

On the pure Java side in the protocol adapter: yes. That should be an exchangeable API, dedicated to this use case.

I also think it is OK to amend the Device Registry API with those two methods. As long as they are marked *optional*. Because you would only need them, I your choose to use the device registry as a backing service. If you choose to use some alternative implementation for the *device state* then those two methods would no longer be required.

calohmn:
> I also think it is OK to amend the Device Registry API with those two methods. As long as they are marked optional. Because you would only need them, I your choose to use the device registry as a backing service.

But our default would be to choose the device registry as a backing service, not requiring the deployment of protocol adapters with a custom `RegistrationClientImpl` / `GatewayMappingDataClientImpl`. Therefore I think we cannot mark them optional in the sense that an `NotImplemented` exception is being thrown by default, otherwise the C&C feature wouldn't work with a custom device registry that hasn't implemented them.
To make things explicit, I would mark the methods mandatory and add documentation to say that if not wanting to support the gateway feature (while still wanting C&C to work), such and such values should be returned.

I think the topic of *device state* and whether/how to introduce an API for that opens an array of additional questions. Discussing this would take longer than our projected M4 build on Wednesday, imho.

Therefore I would suggest of going ahead with adding the methods in the Device Registry API here and then lead the discussion concerning *device state* towards maybe considering this for an extra M5 build.

ctron:
I agree that we should not rush things here.

Marking those new operations as *mandatory* would mean that a device registry *must* implement them right? And I am not saying to throw a "not implemented" exception by default.

However, how to we ensure that we can support a use case without using the device registry as a backing system of this device state?

I think we can achieve this (without too much trouble at the moment) by defining those new methods as "optional", and at the same time implement them. Leaving the possibility to swap out the device state implementation with a different solution than using the device registry, and not requiring a system, that doesn't use the device registry as a backend, to implement those methods.

In other words: If I choose to use a different strategy for the device state, then I would not like to be force to implement those new methods on the device registry, just because they are mandatory.

sophokles73:
> I think we can achieve this (without too much trouble at the moment) by defining those new methods as "optional", and at the same time implement them. Leaving the possibility to swap out the device state implementation with a different solution than using the device registry, and not requiring a system, that doesn't use the device registry as a backend, to implement those methods.
In other words: If I choose to use a different strategy for the device state, then I would not like to be force to implement those new methods on the device registry, just because they are mandatory.

FMPOV it is the other way around:
Making the operations *mandatory* allows protocol adapters using the default `RegistrationClientImpl` that comes with Hono to use *any* Device Registration implementation. That is, in fact, what we wanted to achieve with defining these APIs.
Making them optional to implement would mean that this promise may be broken.
You may still chose to use your own `RegistrationClient` implementation for the PAs which keeps the *device state* in a (local) data grid that is independent from the Device Registration service. In this case, it would be irrelevant how the Device Registration service implements those two methods, because they would never be invoked. However, in any case can the Device Registration service implementation be used with any protocol adapter ...

ctron:
But that is exactly my point: making this mandatory, makes it a must for all device registries to implement this, workaround, that we are currently putting in place because we want to quickly get this in.

This is a new use case that we want to bring in. A complete change to what was there before. And we are re-using some existing functionality to quickly implement this. And I think it is ok to take the shortcut for the moment. But forcing that shortcut on everyone seems wrong to me.

IMHO the "right" thing to do is to create a new API, which stores *device state*. Which is something different than *tenant*, *registration*, and *credentials*. So we do need a new API for this. And this API can be implemented in different ways.

Taking a shortcut now to get 1.0 out, sounds reasonable to me. Making that shortcut mandatory doesn't.

calohmn:
@ctron To further discuss this here first, I would postpone any kind of change to M5 then (along with these changes: https://github.com/eclipse/hono/issues/1073#issuecomment-496513633).
As for a `device state` API - would that be something that you have other potential operations in mind for already?

ctron:
> @ctron To further discuss this here first, I would postpone any kind of change to M5 then (along with these changes: [#1073 (comment)](https://github.com/eclipse/hono/issues/1073#issuecomment-496513633)).

That sounds like a good idea.

> As for a `device state` API - would that be something that you have other potential operations in mind for already?

Not really, however I think we could re-think the *connection events* in this context. Having a device state, would allow is to e.g record something like "last known message".

calohmn:
I think the main question is: 
Do we already see the definitive need for a `device state` API, so that putting the `last-used gateway` methods in the device registration API would indeed be temporary and kind of a workaround? Then I would prefer creating the `device state` API for 1.0 already.

Otherwise I would suggest defining the device registry as the official and default place where `last-used gateway` is stored right now and therefore would make these methods mandatory, because otherwise device registration API implementations wouldn't be interchangeable.

FMPOV, a `device state` API doesn't look necessary right now. It would make the Hono APIs and their fields of responsibility clearer and cleaner but it would also add more complexity, also for people using Hono (e.g. requiring extra configuration for the protocol adapters).
I would lean towards keeping things simple until there are more reasons for a `device state` API and therefore use the device registration API for the `last-used gateway` for now.

calohmn:
As discussed in our call, the `last-used gateway` methods should be put in a new API, allowing for easier extensibility and meaning a clearer separation of concerns.

I've created PR #1274 as the initial contribution for this.
(I've used the name `Device Connection API` instead of `Device State API` - see the PR description.)

Issue #1255, opened 2019-05-23
Support tenant-scoped CommandClient


calohmn:
With the changes for #930, the sender link for sending commands is scoped on the tenant now.

That means there only needs to be one sender link on which all commands for devices of one particular tenant can be sent.

However, the current `CommandClient` and `AsyncCommandClient` instances are geared towards a particular device. The underlying sender link can't be shared for sending commands for different devices.

Therefore we should support a tenant-scoped CommandClient, where the `sendCommand` method gets a device parameter.

sophokles73:
IMHO we should replace the existing methods with methods accepting the device ID, i.e. we should change the clients' underlying sender link to be always scoped to the tenant instead of the device.

calohmn:
@sophokles73 For the `AsyncCommandClient` class this looks like the straightforward solution.

For the `CommandClient` class however, we have to take into account that there is a command response receiver link *scoped on the device* being opened there. This requires some more thought.

calohmn:
With the `CommandClientImpl` being an `AbstractHonoClient` with one `ProtonSender` and one `ProtonReceiver` we would have to override quite some of the base class functionality in order to handle multiple *device-scoped* command response receiver links (not opening and closing them on every `sendCommand`). I would rather keep the `CommandClientImpl` device-scoped, but handle re-use of the tenant-scoped command sender link via the `ApplicationClientFactory` (on which `getOrCreateCommandClient` is to be called).

sophokles73:
isn't the intended behavior of the CommandClient the one that we already have in all the other Device Registry clients i.e. using a single set of sender and receiver links for multiple requests?

calohmn:
Yes, the CommandClient would have a single device-scoped Command receiver link and the tenant-scoped Command sender link, provided by and shared via the  ApplicationClientFactory / CommandClientFactory instance. Both would be used for multiple requests. Or what have you been aiming at?

sophokles73:
Well, I would not call it *scoped to a single device* but rather *scoped to a single UUID* c´generated on instantiation, right?

calohmn:
Right, the command response receiver links are indeed scoped to the `replyId` - which may be an ID shared between commands for *different* devices. 

In that sense the `CommandClient` instance can indeed keep just *one* command response receiver link to handle commands for multiple devices.

[...]

PR #1258, opened 2019-05-24
[#1073] Change northbound control endpoint names

calohmn:
This is for #1073:
The recently introduced `control/${tenant_id}` endpoint (providing the gateway mapping feature) is renamed to `control_request/${tenant_id}`.
For the command reply address, the pattern `control_response/${tenant_id}/${reply_id}` is now used by default. The old pattern `control/${tenant_id}/${reply_id}` is still supported as legacy control endpoint.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/Command.java
+++ b/client/src/main/java/org/eclipse/hono/client/Command.java
@@ -36,6 +36,11 @@ public final class Command {
      * contained the device id.
      */
     private static final byte FLAG_REPLY_TO_CONTAINED_DEVICE_ID = 1;
+    /**
+     * Bit flag value for the boolean option that defines whether the original reply-to address of the command message
+     * contained the legacy endpoint name.
+     */
+    private static final byte FLAG_REPLY_TO_LEGACY_ENDPOINT_USED = 2;
 
     /**
      * If present, the command is invalid.
@@ -46,6 +51,7 @@ public final class Command {
     private final String deviceId;
     private final String correlationId;
     private final String replyToId;
+    private final boolean replyToLegacyEndpointUsed;
     private final String requestId;
 
     private Command(
@@ -54,7 +60,8 @@ public final class Command {
             final String tenantId,
             final String deviceId,
             final String correlationId,
-            final String replyToId) {
+            final String replyToId,
+            final boolean replyToLegacyEndpointUsed) {
 
         this.validationError = validationError;
         this.message = message;
@@ -62,7 +69,8 @@ public final class Command {
         this.deviceId = deviceId;
         this.correlationId = correlationId;
         this.replyToId = replyToId;
-        this.requestId = getRequestId(correlationId, replyToId, deviceId);
+        this.replyToLegacyEndpointUsed = replyToLegacyEndpointUsed;
+        this.requestId = getRequestId(correlationId, replyToId, deviceId, replyToLegacyEndpointUsed);
     }
 
     /**
@@ -123,23 +131,25 @@ public final class Command {
         }
 
         String originalReplyToId = null;
-
+        boolean replyToLegacyEndpointUsed = false;
         if (message.getReplyTo() != null) {
             try {
                 final ResourceIdentifier replyTo = ResourceIdentifier.fromString(message.getReplyTo());
-                if (!CommandConstants.COMMAND_ENDPOINT.equals(replyTo.getEndpoint())) {
+                if (!CommandConstants.isNorthboundCommandResponseEndpoint(replyTo.getEndpoint())) {
                     // not a command message
                     validationErrorJoiner.add("reply-to not a command address: " + message.getReplyTo());
                 } else if (!tenantId.equals(replyTo.getTenantId())) {
                     // command response is targeted at wrong tenant
                     validationErrorJoiner.add("reply-to not targeted at tenant " + tenantId + ": " + message.getReplyTo());
                 } else {
+                    replyToLegacyEndpointUsed = CommandConstants.isNorthboundCommandLegacyEndpoint(replyTo.getEndpoint());
                     originalReplyToId = replyTo.getPathWithoutBase();
                     if (originalReplyToId == null) {
                         validationErrorJoiner.add("reply-to part after tenant not set: " + message.getReplyTo());
                     } else {
                         message.setReplyTo(
-                                String.format("%s/%s", replyTo.getBasePath(), getDeviceFacingReplyToId(originalReplyToId, deviceId)));
+                                String.format("%s/%s/%s", CommandConstants.COMMAND_ENDPOINT, tenantId,
+                                        getDeviceFacingReplyToId(originalReplyToId, deviceId, replyToLegacyEndpointUsed)));
                     }
                 }
             } catch (IllegalArgumentException e) {
@@ -154,7 +164,8 @@ public final class Command {
                 tenantId,
                 deviceId,
                 correlationId,
-                originalReplyToId);
+                originalReplyToId,
+                replyToLegacyEndpointUsed);
 
         return result;
     }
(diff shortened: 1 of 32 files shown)


sophokles73:
@calohmn what about keeping the `control` prefix for the response messages and only change the prefix for sending commands to, well, `command` (while still keeping the old `control` around for some time)?. Wouldn't that help us in preventing the name clash while allowing us to keep all of the existing code for the response direction?

calohmn:
@sophokles73 As discussed in #1073, I've now changed the 
`NORTHBOUND_COMMAND_REQUEST_ENDPOINT` from "control_request" to "command" and the 
`NORTHBOUND_COMMAND_RESPONSE_ENDPOINT` from "control_response" to "command_response".

Changing the (HTTP/MQTT/AMQP) device endpoints in the protocol adapters from "control" to "command" (or, more specifically, adding "command" endpoints and marking "control" deprecated) is something that involves quite a lot of further changes. I would do that in a separate subsequent PR then.

sophokles73:
@calohmn sounds good to me :+1:

calohmn:
Not increasing the size of this PR any further, I'll put the corresponding adaptations of the Hono documentation and release notes (along with a deprecation notice of the old endpoint names) in a separate PR.

client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -123,23 +129,25 @@ public static Command from(
...
                 final ResourceIdentifier replyTo = ResourceIdentifier.fromString(message.getReplyTo());
-                if (!CommandConstants.COMMAND_ENDPOINT.equals(replyTo.getEndpoint())) {
+                if (!CommandConstants.isNorthboundCommandResponseEndpoint(replyTo.getEndpoint())) {
                     // not a command message
                     validationErrorJoiner.add("reply-to not a command address: " + message.getReplyTo());
                 } else if (!tenantId.equals(replyTo.getTenantId())) {
                     // command response is targeted at wrong tenant
                     validationErrorJoiner.add("reply-to not targeted at tenant " + tenantId + ": " + message.getReplyTo());
                 } else {
+                    replyToLegacyEndpointUsed = CommandConstants.NORTHBOUND_COMMAND_LEGACY_ENDPOINT.equals(replyTo.getEndpoint());
```
sophokles73: since we already have a *CommandConstants.isNorthboundCommandResponseEndpoint* method, IMHO it would be nice if we could replace this explicit comparison with an invocation of *CommandConstants.isNorthboundLegacyEndpoint*. WDYT?

client/src/main/java/org/eclipse/hono/client/CommandConsumerFactory.java
```diff
@@ -123,11 +123,29 @@ static CommandConsumerFactory create(final HonoConnection connection, final Gate
...
+     * once the sender is no longer needed anymore.
+     *
+     * @param tenantId The ID of the tenant to send the command responses for.
+     * @param replyId The ID used to build the reply address as {@code control/tenantId/replyId}.
+     * @return A future that will complete with the sender once the link has been established.
+     *         The future will be failed with a {@link ServiceInvocationException} if
+     *         the link cannot be established, e.g. because this client is not connected.
+     * @throws NullPointerException if any of the parameters are {@code null}.
+     */
+    Future<CommandResponseSender> getLegacyCommandResponseSender(String tenantId, String replyId);
```
sophokles73: IMHO we should mark this method as `@deprecated` already ...

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -284,6 +284,19 @@ public final void sendQueueDrainHandler(final Handler<Void> handler) {
      */
     protected abstract String getName();
 
+    /**
+     * Get the name of the endpoint that will be used for the reply-to address.
```
sophokles73: Gets the name ...

client/src/main/java/org/eclipse/hono/client/impl/AsyncCommandClientImpl.java
```diff
@@ -117,7 +117,7 @@ static String getLinkTargetAddress(final String tenantId) {
         MessageHelper.setPayload(message, contentType, data);
         message.setSubject(command);
         message.setAddress(targetAddress);
-        final String replyToAddress = String.format("%s/%s/%s", CommandConstants.COMMAND_ENDPOINT, tenantId, replyId);
+        final String replyToAddress = String.format("%s/%s/%s", CommandConstants.NORTHBOUND_COMMAND_RESPONSE_ENDPOINT, tenantId, replyId);
```
sophokles73: IMHO we should use *getEndpoint* here instead of the constant, shouldn't we?

client/src/main/java/org/eclipse/hono/client/impl/CommandClientImpl.java
```diff
@@ -240,10 +245,11 @@ protected BufferResult getResult(
     /**
      * Creates a new command client for a tenant and device.
      * <p>
-     * The instance created is scoped to the given device.
-     * In particular, the sender link's target address is set to
-     * <em>control/${tenantId}/${deviceId}</em> and the receiver link's source
-     * address is set to <em>control/${tenantId}/${deviceId}/${replyId}</em>.
+     * The instance created is scoped to the given device. In particular, the target address of messages is set to
```
sophokles73: didn't we want to change this so that the client is scoped to the tenant only?

client/src/main/java/org/eclipse/hono/client/impl/DelegatedCommandSenderImpl.java
```diff
@@ -183,7 +183,7 @@ public String getEndpoint() {
         final String targetAddress = getTargetAddress(tenantId, deviceId);
         final String replyToAddress = command.isOneWay() ? null
                 : String.format("%s/%s/%s",
-                CommandConstants.COMMAND_ENDPOINT, tenantId, command.getReplyToId());
+                CommandConstants.NORTHBOUND_COMMAND_RESPONSE_ENDPOINT, tenantId, command.getReplyToId());
```
sophokles73: a comment explaining why this is *not* the legacy endpoint would be helpful ...

client/src/test/java/org/eclipse/hono/client/CommandResponseTest.java
```diff
@@ -324,7 +324,7 @@ public void testForNoDeviceIdInReplyToId() {
     @Test
     public void testForDeviceAndTenantIds() {
         final CommandResponse response = CommandResponse.from(
-                Command.getRequestId(CORRELATION_ID, REPLY_TO_ID, DEVICE_ID),
+                Command.getRequestId(CORRELATION_ID, REPLY_TO_ID, DEVICE_ID, false),
```
sophokles73: what about tests for the case where the flag is `true`?

client/src/main/java/org/eclipse/hono/client/impl/AsyncCommandClientImpl.java
calohmn: *getEndpoint()* would return the *request* endpoint. However, in the above line we need the *response* endpoint.

client/src/main/java/org/eclipse/hono/client/impl/CommandClientImpl.java
calohmn: This topic is discussed in #1255. I don't want to include a solution for that in this PR here.

client/src/main/java/org/eclipse/hono/client/Command.java
calohmn: Ok, I've added and used a `isNorthboundCommandLegacyEndpoint` method.

client/src/main/java/org/eclipse/hono/client/CommandConsumerFactory.java
calohmn: Right, I've added that.

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
calohmn: Done

client/src/main/java/org/eclipse/hono/client/impl/DelegatedCommandSenderImpl.java
calohmn: Thanks for pointing at that line. Always choosing the `NORTHBOUND_COMMAND_RESPONSE_ENDPOINT` was actually wrong here. I've added and used a `Command.getReplyToEndpoint` method here.

client/src/test/java/org/eclipse/hono/client/CommandResponseTest.java
calohmn: I've added tests for that.

PR #1260, opened 2019-05-27
Limit the use of "suppress unchecked"

ctron:


client/src/test/java/org/eclipse/hono/client/impl/VertxSupport.java
```diff
@@ -0,0 +1,39 @@
...
+package org.eclipse.hono.client.impl;
+
+import org.mockito.ArgumentMatchers;
+
+import io.vertx.core.Handler;
+
+/**
+ * Argument matchers for the use with vert.x.
+ */
+public final class VertxSupport {
```
sophokles73: If this class is supposed to provide ArgumentMatchers, then why not call it `ArgumentMatchers`?

ctron: That is what I named it at first. Looking at a few other locations I found a few more targets to work on (mock(), ArgumentCaptor), … I thought it might make more sense to put all of those into `VertxSupport`, rather than creating three different classes.

sophokles73: Then what about `VertxMockSupport` ?

ctron: Done. Also added two more variants of this.

Issue #1262, opened 2019-05-28
MQTT Adapter: Close device connection when a telemetry/event message cannot be processed.


kaniyan:
A device publishes an _event/telemetry_ message with _QoS 1_. The _MQTT adapter_ after successfully establishing the device connection, realizes that the message cannot be _processed_ e.g _no credit available_, _the message limit is exceeded_ etc.  In this case the device remains connected and keeps waiting for the `PUBACK` packet from the Hono. 

Similar scenario is when a device publishes a _telemtry/event_ message with _QoS 2_ (_not supported by Hono_). In this case, the device remain connected and it keep waiting for the `PUBREC` packet.

_MQTT 3.1.1 spec_ ([MQTT-3.3.5-2](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718042)) specifies that there is no way to inform the MQTT client that the _publish_ action is not accepted. The server must either make a positive acknowledgement or _close_ the _Network Connection._

sophokles73:
From the spec:
> If a Server implementation does not authorize a PUBLISH to be performed by a
Client; it has no way of informing that Client. It MUST either make a positive
acknowledgement, according to the normal QoS rules, or close the Network
Connection.

First, this seems to refer to a case where the server considers the client to *not be authorized* to PUBLISH the message. Your example is about insufficient credits, though, which seems to be a different case FMPOV. However, the problem with MQTT 3.1.1 lacking a mechanism to convey an error situation to a client arises in multiple situations. We have discussed this in the past and always came to the conclusion that there is no real use in closing the connection, because the client will (most likely) immediately re-open the connection and re-send the very same message again. On the other hand, the MQTT spec also advises clients to *wait for a reasonable amount of time* for the server's PUBACK and after that consider the message to be lost. Thus, a client needs to deal with this situation anyway, so FMPOV it doesn't really make much sense to force the client to go through the (expensive) TLS handshake again, just to end up in the same situation as it would be in without disconnecting the client in the first place.

Does this make sense to you?

ctron:
I made the experience that e.g. Paho doesn't handle missing ACKs very well. As it simply seems to follow the spec and waits/blocks until all ACKs have been received.

With QoS 1 the clients wait for an ACK, as the remote side either sends an ACK, or is expected to close the connection.

[...]

PR #1263, opened 2019-05-28
[#1262] Close mqtt connection when a telemetry/event message cannot be processed.

kaniyan:
This PR address the issue #1262 with reference to the [MQTT 3.1.1 spec](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718042). When a _telemetry/event_ message is not able to be processed successfully then the device connection is closed.

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
@@ -713,6 +713,12 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends MqttProtoc
                     } else {
                         onMessageUndeliverable(context);
                     }
+                        if (context.deviceEndpoint().isConnected()) {
+                            TracingHelper.logError(span,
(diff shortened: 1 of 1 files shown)


sophokles73:
@ctron please do not merge this PR before you have taken a look at my comment on #1262 
We deliberately chose to **not** close the MQTT connection in such cases so FMPOV this PR is obsolete ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -713,6 +713,9 @@ void handlePublishedMessage(final MqttContext context) {
                     } else {
                         onMessageUndeliverable(context);
                     }
+                    if (context.deviceEndpoint().isConnected()) {
+                        context.deviceEndpoint().close();
+                    }
```
sophokles73: I think we should also log any potential error that caused us to disconnect the client to the span ...

kaniyan: Now it logs the potential error to the span.

PR #1264, opened 2019-05-29
Remove kubernetes shell deployment

sophokles73:
- Removes script and resource descriptors for shell script based deployment to Kukbernetes
- Adds a page in Deployment section for Setting up a Kubernetes cluster
- Adds a page describing only Helm based deployment (regardless of cluster type)


sophokles73:
> I would still prefer to have AKS setup in a separate doc file as it's large enough on its own and we'll probably add other providers in the future.

I had the same concern. However, in general I would like to refer people to the standard guides provided in the Kubernetes docs instead of writing (and maintaining) guides ourselves. Thus, FMPOV the AKS guide already is an exceptional case and I would rather not add more guides. Does that make sense?

dejanb:
Yeah, it should work for now.

Issue #1272, opened 2019-06-05
Reduce number of Command & Control receiver links


calohmn:
Currently there is one Command & Control receiver link *for each device*, opened from the protocol adapter instance to the AMQP network.
This link is on the address `control/$tenant/$device` (or `control/$tenant/$gateway` respectively).

It would be good if we could reduce that potentially large number of links.

As proposed by @sophokles73, we could use receiver links for each protocol adapter instance (containing an instance uuid in the address) instead.

Taking on the example illustration from #930 
![Gateway-Mapper](https://user-images.githubusercontent.com/891931/58940526-f43de380-8779-11e9-970b-881176c4fc2c.png)
this means replacing the red `control/$tenant/$gateway` links with just one for each protocol adapter instance, e.g. `control_internal/$tenant/$protocolAdapterUUID`.

Whenever a device/gateway subscribes to a protocol adapter for receiving Command & Control messages, we will store the mapping of `device/gateway ID -> protocol adapter instance` via the new Device Connection API. We will also make sure there is a receiver link `control_internal/$tenant/$protocolAdapterUUID` on that protocol adapter instance.
Upon receiving a command on the tenant-scoped receiver link, the protocol adapter will query the Device State API to get the corresponding protocol adapter uuid and forward the command message to `control_internal/$tenant/$protocolAdapterUUID`.

---
Concerning when to implement this:
we currently have 2 kinds of receiver links in the protocol adapters: 
- `control/$tenant` (tenant-scoped link with gateway-mapping feature)
- `control/$tenant/$device` (legacy command link, also used now for forwarding the gateway-mapped commands internally)

The 2nd `control/$tenant/$device` link will be made into an only-used-internally kind of link at some point, removing the backwards compatibility for applications to send commands on that link.
When we do this, we could change the `control/$tenant/$device` link into the new 
 `control_internal/$tenant/$protocolAdapterUUID` link. That way, we don't have to deal with 3 kinds of receiver links at the same time.

ctron:
Wouldn't it make sense to allow the client side to also re-use that information in some way. Like passing along the protocol adapter ID in the telemetry messages and allow the client to send a command directly to that protocol adapter.

That would allow to forward the device command (after an inbound message) directly to the correct protocol adapter and prevent us from doing a lookup in order to figure out the "correct" protocol adapter.

calohmn:
I see that this would speed up routing of command messages.

[...]

PR #1274, opened 2019-06-05
[#1245] Add new Device Connection API documentation

calohmn:
This is for #1245:

As discussed in #1257, we need a new API to store and retrieve information concerning which gateway acted last on behalf of a given device.

This PR adds the documentation for such a new API.

I've called it `Device Connection API`, containing the `Set Last Known Gateway for Device` and `Get Last Known Gateway for Device` operations and using `device_con` as AMQP address prefix.

The name `Device State API`, as used beforehand in #1257, would imply that information regarding the state of devices is stored there which is not really the case (it is more about the connections of the devices). The other alternative name of `Connection State API` would point to connection states like "open", "closed" which is also not a good fit, IMHO.

The `Device Connection API` may then also be used to act as a serialization layer for a `ConnectionEventProducer` implementation.

A further future addition to this API might be a mapping of the gateway/device ID to the instance UUID of the protocol adapter it is connected to (`protocolAdapterForDevice`). See #1272 for that.


site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,148 @@
...
+
+# Preconditions
+
+The preconditions for performing any of the operations are as follows:
+
+1. Client has established an AMQP connection with Hono.
+2. Client has established an AMQP link in role *sender* with Hono using target address `device_con/${tenant_id}`. This link is used by the client to send commands concerning device connections to Hono.
+3. Client has established an AMQP link in role *receiver* with Hono using source address `device_con/${tenant_id}/${reply-to}` where *reply-to* may be any arbitrary string chosen by the client. This link is used by the client to receive responses to the device state requests it has sent to Hono. This link's source address is also referred to as the *reply-to* address for the request messages.
+
+This flow of messages is illustrated by the following sequence diagram (showing the AMQP performatives):
```
sophokles73: check out the changes I made to the Tenant API
can you align this section with the phrasing used there so that we have a consistent terminology?
I will change the Device Registration and Credentials API accordingly during the next days.

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,148 @@
...
+
+![Device_Connection message flow preconditions](../connectToDeviceConnection.png)
+
+# Mandatory Operations
```
sophokles73: there do not seem to be any optional operations so I would propose to rename this to just `Operations` ...

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,148 @@
...
+
+The operations described in the following sections are invoked by Hono's components during run time and are therefore mandatory to implement.
+
+## Set Last Known Gateway for Device
+
+Clients use this command to *set* the gateway that last acted on behalf of a given device. 
+
+The association between device and its last known gateway is used for scenarios where devices with multiple potential gateways are used, along with gateways subscribing to command messages only using their gateway id. In such scenarios, the value set here is needed to route command messages to the right gateway.
```
sophokles73: FMPOV this sentence is not particularly enlightening. IMHO we need to either elaborate (a lot) on the routing issue or (preferred) simply omit this sentence altogether.

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,148 @@
...
+
+This operation is *mandatory* to implement.
+
+**Message Flow**
+
+
+The following sequence diagram illustrates the flow of messages involved in a *Client* retrieving the last known gateway for a device.
```
sophokles73: a Client *setting* the last ...

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,148 @@
...
+
+The operations described in the following sections are invoked by Hono's components during run time and are therefore mandatory to implement.
+
+## Set Last Known Gateway for Device
+
+Clients use this command to *set* the gateway that last acted on behalf of a given device. 
+
+The association between device and its last known gateway is used for scenarios where devices with multiple potential gateways are used, along with gateways subscribing to command messages only using their gateway id. In such scenarios, the value set here is needed to route command messages to the right gateway.
+
+This operation is *mandatory* to implement.
```
sophokles73: no need to mention ...

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,148 @@
...
+
+
+The following sequence diagram illustrates the flow of messages involved in a *Client* retrieving the last known gateway for a device.
+
+![Set Last Known Gateway for Device message flow](../setLastKnownGatewayForDevice_Success.png)
+
+
+**Request Message Format**
+
+The following table provides an overview of the properties a client needs to set on a message to retrieve the last known gateway for a device in addition to the [Standard Request Properties]({{< relref "#standard-request-properties" >}}).
```
sophokles73: to set the last ...

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,148 @@
...
+
+| Name         | Mandatory | Location                 | AMQP Type | Description |
+| :----------- | :-------: | :----------------------- | :-------- | :---------- |
+| *subject*    | yes       | *properties*             | *string*  | MUST be set to `setLastKnownGatewayForDevice`. |
```
sophokles73: do we really need such a long subject? what about just `set-last-gw`?

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,148 @@
...
+| *gateway_id* | yes       | *application-properties* | *string*  | The identifier of the gateway that last acted on behalf of a given device. If a device connects directly instead of through a gateway, the device identifier itself is to be used here. |
```
sophokles73: > The identifier of the gateway that last acted on behalf of the device identified by the *device_id* property. If a device connects directly instead of through a gateway, the device's identifier MUST be specified here.

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,148 @@
...
+**Response Message Format**
+
+A response to a *set last known gateway for device* request contains the [Standard Response Properties]({{< relref "#standard-response-properties" >}}).
+
+The response message's *status* property may contain the following codes:
+
+| Code | Description |
+| :--- | :---------- |
+| *204* | OK, the last known gateway for the device was updated. |
+| *404* | Not Found, there is no device registered with the given *device_id* within the given *tenant_id*. |
```
sophokles73: or no gateway with the given ID, right?

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,148 @@
...
+
+For status codes indicating an error (codes in the `400 - 499` range) the message body MAY contain a detailed description of the error that occurred.
+
+## Get Last Known Gateway for Device
+
+Clients use this command to *retrieve* the gateway that last acted on behalf of a given device.
+
+This operation is *mandatory* to implement.
```
sophokles73: no need to mention

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,148 @@
...
+![Get Last Known Gateway for Device message flow](../getLastKnownGatewayForDevice_Success.png)
+
+
+**Request Message Format**
+
+The following table provides an overview of the properties a client needs to set on a message to retrieve the last known gateway for a device in addition to the [Standard Request Properties]({{< relref "#standard-request-properties" >}}).
+
+| Name        | Mandatory | Location                 | AMQP Type | Description |
+| :---------- | :-------: | :----------------------- | :-------- | :---------- |
+| *subject*   | yes       | *properties*             | *string*  | MUST be set to `getLastUsedGateway`. |
```
sophokles73: what about `get-last-gw` or `getLastGW`?

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,148 @@
...
+
+**Response Message Format**
+
+A response to a *get last known gateway for device* request contains the [Standard Response Properties]({{< relref "#standard-response-properties" >}}).
+
+The response message includes payload with a JSON object having the following properties:
+
+| Name             | Mandatory | JSON Type     | Description |
+| :--------------- | :-------: | :------------ | :---------- |
+| *device-id*      | *yes*     | *string*      | The ID of the last known gateway for the device, or the device ID itself if the device doesn't support usage via a gateway. |
```
sophokles73: then this should probably be named *gateway-id*, right?

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,148 @@
...
+| Code | Description |
+| :--- | :---------- |
+| *200* | OK, the payload contains the gateway or device ID. |
+| *404* | Not Found, there is no device registered with the given *device_id* within the given *tenant_id* or the device supports usage via a gateway but there was no last known gateway assigned to the device. |
+
+For status codes indicating an error (codes in the `400 - 499` range) the message body MAY contain a detailed description of the error that occurred.
+
+# Standard Message Properties
+
+Due to the nature of the request/response message pattern of the operations of the Device State API, there are some standard properties shared by all of the request and response messages exchanged as part of the operations.
```
sophokles73: Device State API?

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,148 @@
...
+| :--------------- | :-------: | :----------------------- | :----------- | :---------- |
+| *subject*        | yes       | *properties*             | *string*     | MUST be set to the value defined by the particular operation being invoked. |
+| *correlation-id* | no        | *properties*             | *message-id* | MAY contain an ID used to correlate a response message to the original request. If set, it is used as the *correlation-id* property in the response, otherwise the value of the *message-id* property is used. |
+| *message-id*     | yes       | *properties*             | *string*     | MUST contain an identifier that uniquely identifies the message at the sender side. |
+| *reply-to*       | yes       | *properties*             | *string*     | MUST contain the source address that the client wants to received response messages from. This address MUST be the same as the source address used for establishing the client's receive link (see [Preconditions]({{< relref "#preconditions" >}})). |
+| *content-type*   | yes       | *properties*             | *string*     | MUST be set to `application/json`. |
```
sophokles73: we should remove this because
1. the request messages do not contain any payload anyway and
1. this limits compatibility with JMS based clients as there is no way to set the content type in JMS

site/content/api/Device-Connection-API.md
calohmn: Ok, I'll omit it.

site/content/api/Device-Connection-API.md
calohmn: Right, I'll use `set-last-gw`.

site/content/api/Device-Connection-API.md
calohmn: Yes, `get-last-gw` it'll be.

site/content/api/Device-Connection-API.md
calohmn: I first thought that because the returned id could also be the device ID (if the device isn't connected to a gateway) the more general `device-id` should better be used.
But I agree that this could be confusing, also since in the setter operation the property is also named `gateway_id`.
I'll change it to `gateway-id`.

calohmn:
@sophokles73 I've amended the commit according to your comments.

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,138 @@
...
+**Response Message Format**
+
+A response to a *set last known gateway for device* request contains the [Standard Response Properties]({{< relref "#standard-response-properties" >}}).
+
+The response message's *status* property may contain the following codes:
+
+| Code | Description |
+| :--- | :---------- |
+| *204* | OK, the last known gateway for the device was updated. |
+| *404* | Not Found, there is no device registered with the given *device_id* within the given *tenant_id*, or no gateway registered with the given *gateway_id*. |
```
sophokles73: returning a 404 would actually require the service implementation to check with the registration service if the device/gw exists. This, however, introduces a new dependency on the registration service which we wanted to eliminate with the introduction of this service in the first place ...
So we will probably end up accepting all identifiers, right?

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,138 @@
...
+**Response Message Format**
+
+A response to a *get last known gateway for device* request contains the [Standard Response Properties]({{< relref "#standard-response-properties" >}}).
+
+The response message includes payload with a JSON object having the following properties:
+
+| Name             | Mandatory | JSON Type     | Description |
+| :--------------- | :-------: | :------------ | :---------- |
+| *gateway-id*     | *yes*     | *string*      | The ID of the last known gateway for the device, or the device ID itself if the device doesn't support usage via a gateway. |
+| *update-date*    | *no*      | *string*      | The date that the information about the last known gateway for the device was last updated. The date is formatted in ISO-8601 format, e.g. '2011-12-03T10:15:30Z'. |
```
sophokles73: we should provide a link to the specification of  the format here. See https://www.eclipse.org/hono/api/tenant-api/#data-volume-configuration-format

site/content/api/Device-Connection-API.md
calohmn: Right, that is still a leftover from defining the method in the registration API. I've removed both such places in the doc.

site/content/api/Device-Connection-API.md
calohmn: done

site/content/api/Device-Connection-API.md
sophokles73: FMPOV it is ok to remove the 404 and indeed *not* perform a check for existence in this service. However, I think we should explicitly mention that this means that every client can submit *any* data (within its tenant) and that implementors should therefore take care to e.g. restrict the overall amount of data that can be stored per tenant, in particular if the data is to be kept in memory only.

calohmn: Good point. I've added corresponding notes in the doc.

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,143 @@
...
+
+A response to a *set last known gateway for device* request contains the [Standard Response Properties]({{< relref "#standard-response-properties" >}}).
+
+The response message's *status* property may contain the following codes:
+
+| Code | Description |
+| :--- | :---------- |
+| *204* | OK, the last known gateway for the device was updated. |
+
+Implementors of this API may choose to check whether there is a device registered with the given *device_id* within the given *tenant_id* or a gateway registered with the given *gateway_id*. In such a case a *404* status code may be returned.
```
sophokles73: what about
> Implementors of this API may return a *404* status code in order to indicate that no device and/or gateway with the given identifiers exists for the tenant. However, performing such a check is optional.

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,143 @@
...
+| :--------------- | :-------: | :------------ | :---------- |
+| *gateway-id*     | *yes*     | *string*      | The ID of the last known gateway for the device. |
+| *update-date*    | *no*      | *string*      | The date that the information about the last known gateway for the device was last updated. The value MUST be an [ISO 8601 compliant *combined date and time representation in extended format*](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations). |
+
+The response message's *status* property may contain the following codes:
+
+| Code | Description |
+| :--- | :---------- |
+| *200* | OK, the payload contains the gateway ID. |
+| *404* | Not Found, there is no last known gateway assigned to the device. |
```
sophokles73: I assume this means: *we do not have any information for the given device ID on record*.
Is this also returned if there is no device with the given ID? Do we need to distinguish these cases?
Would it make sense to always return a response with *gateway-id* set to *device_id* if we do not have any information on record?

calohmn: > I assume this means: *we do not have any information for the given device ID on record*.

Yes.

> Is this also returned if there is no device with the given ID?

Yes, if such an optional check is performed.

> Do we need to distinguish these cases?

I don't think so.

> Would it make sense to always return a response with *gateway-id* set to *device_id* if we do not have any information on record?

I wouldn't do that, I would return a 404. 
For a device with a *single* gateway in its `via` device information entry, commands should by default be routed to that gateway, IMHO. This doesn't mean that the Device Connection API should return that gateway id as default - that would impose a dependency on the device registration API.
Instead, the `GatewayMapper` implementation should first invoke `assertRegistration`, get the list of `via` entries in its response (see #1281) and then invoke the `getLastKnownGateway` method of the Device Connection API. With no last known gateway for the device on record, a 404 response should be returned for that request and the `GatewayMapper` should choose the single `via` gateway as mapping target.

site/content/api/Device-Connection-API.md
calohmn: Ok, that's simpler. I've adopted that.

site/content/api/Device-Connection-API.md
sophokles73: sounds reasonable. The Gateway Mapper can then also use the information returned in the *assert Registration* response to verify if the GW returned by the Device Connection service is actually in the list of *allowed* gateways ...

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,143 @@
...
+| :---------- | :-------: | :----------------------- | :-------- | :---------- |
+| *subject*   | yes       | *properties*             | *string*  | MUST be set to `get-last-gw`. |
+
+The body of the message SHOULD be empty and will be ignored if it is not.
+
+**Response Message Format**
+
+A response to a *get last known gateway for device* request contains the [Standard Response Properties]({{< relref "#standard-response-properties" >}}).
+
+The response message includes payload with a JSON object having the following properties:
```
sophokles73: This needs to be more specific FMPOV:
> The result of a successful invocation is carried in a single Data section of the response message as a UTF-8 encoded string representation of a single JSON object. It is an error to include payload that is not of this type.

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,143 @@
...
+| *device_id*      | yes       | *application-properties* | *string*     | MUST contain the ID of the device that is subject to the operation. |
+
+### Standard Response Properties
+
+The following table provides an overview of the properties shared by all response messages regardless of the particular operation being invoked.
+
+| Name             | Mandatory | Location                 | AMQP Type    | Description |
+| :--------------- | :-------: | :----------------------- | :----------- | :---------- |
+| *correlation-id* | yes       | *properties*             | *message-id* | Contains the *message-id* (or the *correlation-id*, if specified) of the request message that this message is the response to. |
+| *content-type*   | yes       | *properties*             | *string*     | MUST be set to `application/json`. |
```
sophokles73: > MUST be set to `application/json` if the response message's status code is 200 and the body of the message contains the result of invoking the operation.

calohmn:
@sophokles73 I've changed both places accordingly.

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Standard Response Properties
+
+The following table provides an overview of the properties shared by all response messages regardless of the particular operation being invoked.
+
+| Name             | Mandatory | Location                 | AMQP Type    | Description |
+| :--------------- | :-------: | :----------------------- | :----------- | :---------- |
+| *correlation-id* | yes       | *properties*             | *message-id* | Contains the *message-id* (or the *correlation-id*, if specified) of the request message that this message is the response to. |
+| *content-type*   | yes       | *properties*             | *string*     | MUST be set to `application/json` if the response message's status code is 200 and the body of the message contains the result of invoking the operation. |
+| *device_id*      | yes       | *application-properties* | *string*     | Contains the ID of the device. |
```
sophokles73: IMHO we do not need this, do we?

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,145 @@
...
+| *device_id*      | yes       | *application-properties* | *string*     | MUST contain the ID of the device that is subject to the operation. |
+
+### Standard Response Properties
+
+The following table provides an overview of the properties shared by all response messages regardless of the particular operation being invoked.
+
+| Name             | Mandatory | Location                 | AMQP Type    | Description |
+| :--------------- | :-------: | :----------------------- | :----------- | :---------- |
+| *correlation-id* | yes       | *properties*             | *message-id* | Contains the *message-id* (or the *correlation-id*, if specified) of the request message that this message is the response to. |
+| *content-type*   | yes       | *properties*             | *string*     | MUST be set to `application/json` if the response message's status code is 200 and the body of the message contains the result of invoking the operation. |
```
sophokles73: I would rather move this to the operation that it is relevant for

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,145 @@
...
+| *status*         | yes       | *application-properties* | *int*        | Contains the status code indicating the outcome of the operation. Concrete values and their semantics are defined for each particular operation. |
+| *cache_control*  | no        | *application-properties* | *string*     | Contains an [RFC 2616](https://tools.ietf.org/html/rfc2616#section-14.9) compliant <em>cache directive</em>. The directive contained in the property MUST be obeyed by clients that are caching responses. |
+
+## Delivery States
+
+Hono uses the following AMQP message delivery states when receiving request messages from clients:
+
+| Delivery State | Description |
+| :------------- | :---------- |
+| *ACCEPTED*     | Indicates that Hono has successfully received and accepted the request for processing. |
```
sophokles73: can we adopt the wording used in Tenant API?

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,145 @@
...
+### Standard Response Properties
+
+The following table provides an overview of the properties shared by all response messages regardless of the particular operation being invoked.
+
+| Name             | Mandatory | Location                 | AMQP Type    | Description |
+| :--------------- | :-------: | :----------------------- | :----------- | :---------- |
+| *correlation-id* | yes       | *properties*             | *message-id* | Contains the *message-id* (or the *correlation-id*, if specified) of the request message that this message is the response to. |
+| *content-type*   | yes       | *properties*             | *string*     | MUST be set to `application/json` if the response message's status code is 200 and the body of the message contains the result of invoking the operation. |
+| *device_id*      | yes       | *application-properties* | *string*     | Contains the ID of the device. |
+| *tenant_id*      | yes       | *application-properties* | *string*     | Contains the ID of the tenant to which the device belongs. |
```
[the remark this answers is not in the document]
calohmn: No, it's not needed. I just wanted to keep the set of standard response properties consistent with the other APIs. That could have some benefit, e.g. in being able to apply the same kind of tracing mechanism, setting the same tags, or, in general, having the same kind of fields set in the internal response `EventBusMessage`.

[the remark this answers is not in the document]
sophokles73: Yes, I understand. However, I have removed all of those properties from the other APIs as well in the context of removing the optional operations ...

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,145 @@
...
+Due to the nature of the request/response message pattern of the operations of the Device Connection API, there are some standard properties shared by all of the request and response messages exchanged as part of the operations.
+
+### Standard Request Properties
+
+The following table provides an overview of the properties shared by all request messages regardless of the particular operation being invoked.
+
+| Name             | Mandatory | Location                 | AMQP Type    | Description |
+| :--------------- | :-------: | :----------------------- | :----------- | :---------- |
+| *subject*        | yes       | *properties*             | *string*     | MUST be set to the value defined by the particular operation being invoked. |
+| *correlation-id* | no        | *properties*             | *message-id* | MAY contain an ID used to correlate a response message to the original request. If set, it is used as the *correlation-id* property in the response, otherwise the value of the *message-id* property is used. |
```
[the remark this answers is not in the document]
calohmn: done

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,145 @@
...
+
+**Response Message Format**
+
+A response to a *set last known gateway for device* request contains the [Standard Response Properties]({{< relref "#standard-response-properties" >}}).
+
+The response message's *status* property may contain the following codes:
+
+| Code | Description |
+| :--- | :---------- |
+| *204* | OK, the last known gateway for the device was updated. |
```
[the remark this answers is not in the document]
calohmn: done

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,145 @@
...
+| Name             | Mandatory | JSON Type     | Description |
+| :--------------- | :-------: | :------------ | :---------- |
+| *gateway-id*     | *yes*     | *string*      | The ID of the last known gateway for the device. |
+| *update-date*    | *no*      | *string*      | The date that the information about the last known gateway for the device was last updated. The value MUST be an [ISO 8601 compliant *combined date and time representation in extended format*](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations). |
+
+The response message's *status* property may contain the following codes:
+
+| Code | Description |
+| :--- | :---------- |
+| *200* | OK, the payload contains the gateway ID. |
```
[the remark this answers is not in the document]
calohmn: done

site/content/api/Device-Connection-API.md
calohmn: ok, done.

site/content/api/Device-Connection-API.md
calohmn: done

site/content/api/Device-Connection-API.md
calohmn: Same argument as above, shouldn't we keep the set of standard response properties somewhat consistent?

site/content/api/Device-Connection-API.md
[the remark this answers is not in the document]
calohmn: Ok, I see. I've removed it then.

site/content/api/Device-Connection-API.md
calohmn: Just read your comment above - I've removed it accordingly.

site/content/api/Device-Connection-API.md
```diff
@@ -0,0 +1,148 @@
...
+| *content-type*   | no        | *properties*             | *string*     | MUST be set to `application/json` if the invocation of the operation was successful and the body of the response message contains payload as described below. |
+
+The result of a successful invocation is carried in a single Data section of the response message as a UTF-8 encoded string representation of a single JSON object. It is an error to include payload that is not of this type.
+
+The response message JSON object has the following properties:
+
+| Name             | Mandatory | JSON Type     | Description |
+| :--------------- | :-------: | :------------ | :---------- |
+| *gateway-id*     | *yes*     | *string*      | The ID of the last known gateway for the device. |
+| *update-date*    | *no*      | *string*      | The date that the information about the last known gateway for the device was last updated. The value MUST be an [ISO 8601 compliant *combined date and time representation in extended format*](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations). |
```
sophokles73: *update-date* sounds a little weird, doesn't it? What about just *updated* or *last-updated* instead?

calohmn: yes, I've changed it to `last-updated`.

PR #1279, opened 2019-06-07
Add missing "via" and "defaults" section

ctron:
It looks like we are missing the "via" field and the  "defaults" section on both the Tenant and Device entity.

### DIFF
--- a/site/static/api/device-registry-v1.yaml
+++ b/site/static/api/device-registry-v1.yaml
@@ -430,6 +430,11 @@ components:
                type: string
                description: A human readable error message of what went wrong.
 
+      DefaultProperties:
+         type: object
+         additionalProperties: true
+         description: Defaults for properties defined on the tenant and device level.
+
       Extensions:
          type: object
          additionalProperties: true
@@ -452,6 +457,8 @@ components:
                description: Only a single entry per type is allowed. If multiple entries for the same type are present it is handled as an error.
                items:
                   $ref: '#/components/schemas/Adapter'
+            "defaults":
+               $ref: '#/components/schemas/DefaultProperties'
             "limits":
                $ref: '#/components/schemas/ResourceLimit'
             "trusted-ca":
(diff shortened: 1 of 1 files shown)


site/static/api/device-registry-v1.yaml
```diff
@@ -430,6 +430,11 @@ components:
                type: string
                description: A human readable error message of what went wrong.
 
+      Settings:
```
sophokles73: *Settings* seems to suggest some specific semantics attached to the properties contained in it.
What about *Any* or *AnyProperties*?

site/static/api/device-registry-v1.yaml
```diff
@@ -519,6 +526,12 @@ components:
             "enabled":
                type: boolean
                default: true
+            "defaults":
+               $ref: '#/components/schemas/Settings'
+            "via":
+               type: string
```
sophokles73: this should be an array of Strings as you can allow a device to connect via any one of multiple gateways ...

ctron: Are you sure about that? Because in the `device-identities.json` used for testing it is only a `String`.

site/static/api/device-registry-v1.yaml
ctron: I am also not truly satisfied with "Settings". However part of this schema definition is the `description` and that gives a bit of explanation around this construct. So simply naming this "Any" wouldn't really fit with the description field anymore.

But I am open for suggestions to other names. Initially I had "Defaults".

site/static/api/device-registry-v1.yaml
sophokles73: Yes, absolutely sure. The example registry allows to read a single String for backwards compatibility reasons only. However, the external API already uses an array ...

site/static/api/device-registry-v1.yaml
sophokles73: My understanding is that we use this type to contain arbitrary properties whose semantics is up to the user. If the only place we use this in is the *defaults* property of Tenant and Device, then *DefaultProperties* would not be bad. Otherwise maybe *ArbitraryProperties*?
The description should be adapted as well in any case.

site/static/api/device-registry-v1.yaml
ctron: Updated.

site/static/api/device-registry-v1.yaml
ctron: Ok, I changed this to `DefaultProperties` and I tweaked the description.

PR #1281, opened 2019-06-11
[#1245] Replace 'gw-supported with 'via' in assertRegistration response

calohmn:
This is for #1245.

In the course of the discussion for #1274 it became apparent that we need the actual gateway information in the `assertRegistration` response when invoking it from the `GatewayMapper`.
The reason being that a single `via` entry should be the default return value of the `GatewayMapper` in case there is no known `last gateway` for the device the command is directed towards.

This PR replaces the `gw-supported` boolean field (introduced in #1248) with the actual `via` property.


sophokles73:
@ctron what impact does this have on the HTTP management API?

ctron:
> @ctron what impact does this have on the HTTP management API?

IIRC we have a `via` field there. I guess that should be aligned.

calohmn:
@ctron You mean the part in the HTTP API spec introduced [here](https://github.com/eclipse/hono/commit/c4a79b25e97e5c02d2887b3051976592970fd894#diff-675cc07421d1039320f67af7205e5a92R531), right?
That is already aligned in the sense that both `via` fields refer to a JSON array of String values.

But I don't see what else to put in `device-registry-v1.yaml`, since there is not `assertRegistration` method in there.

PR #1283, opened 2019-06-11
[#623] Part 1 - Multiple trusted CAs

Alfusainey:
@sophokles73 As discussed in PR #1095, this is the first part of splitting that PR. The major changes are in the `TenantObject`class, which you have almost "completely" reviewed. I also made minor changes to the `base` tenant service classes and the `file based` device registry in this PR. 

Changes to all other classes are required due to the changes in the `TenantObject` class.

In the next part, i will add the tests for the new content model in the tenant service test classes.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/TenantObject.java
+++ b/core/src/main/java/org/eclipse/hono/util/TenantObject.java
@@ -697,20 +697,6 @@ public final class TenantObject extends JsonBackedValueObject {
     }
 
     /**
-     * Get the list of configured trusted CAs for this tenant.
-     * 
-     * @return The list of configured trusted CAs or {@code null} if this tenant has no configured trusted CA.
-     * 
-     */
-    @JsonIgnore
-    public List<JsonObject> getTrustedCAs() {
-        if (trustConfigurations == null) {
-            return null;
-        } else {
-            return trustConfigurations.values().stream().flatMap(List::stream).collect(Collectors.toList());
-        }
-    }
-    /**
      * Gets all the trusted certificate authorities configured for this tenant.
      * 
      * @return The list of X.509 certificates or {@code null} if no certificate authority
--- a/core/src/main/java/org/eclipse/hono/util/TenantObject.java
+++ b/core/src/main/java/org/eclipse/hono/util/TenantObject.java
@@ -227,7 +227,7 @@ public final class TenantObject extends JsonBackedValueObject {
             return trustAnchors;
         } else {
             trustAnchors = new ArrayList<>();
-            final List<JsonObject> configs = Optional.ofNullable(getTrustConfigurationsAsList()).orElse(Collections.emptyList());
+            final List<JsonObject> configs = Optional.ofNullable(getTrustedCAs()).orElse(Collections.emptyList());
             for (JsonObject config : configs) {
                 final TrustAnchor anchor = getTrustAnchor(config);
                 if (anchor != null) {
@@ -618,13 +618,13 @@ public final class TenantObject extends JsonBackedValueObject {
     }
 
     /**
-     * Get all the trusted certificate authorities configured for this tenant.
+     * Get the list of configured trusted CAs for this tenant.
      * 
-     * @return A list of trusted CA configurations for this tenant 
-     *         or {@code null} if no trusted CA has been set for this tenant.
+     * @return The list of configured trusted CAs or {@code null} if this tenant has no configured trusted CA.
+     *
      */
     @JsonIgnore
-    private List<JsonObject> getTrustConfigurationsAsList() {
+    public List<JsonObject> getTrustedCAs() {
 
         if (trustConfigurations == null) {
             return null;
@@ -707,7 +707,7 @@ public final class TenantObject extends JsonBackedValueObject {
      */
     @JsonIgnore
     public List<X509Certificate> getTrustedCertificateAuthorities() throws CertificateException {
-        final List<JsonObject> configs = getTrustConfigurationsAsList();
+        final List<JsonObject> configs = getTrustedCAs();
         if (configs == null) {
             return null;
         } else {
--- a/core/src/main/java/org/eclipse/hono/util/TenantObject.java
+++ b/core/src/main/java/org/eclipse/hono/util/TenantObject.java
@@ -25,13 +25,19 @@ import java.security.cert.TrustAnchor;
 import java.security.cert.X509Certificate;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.X509EncodedKeySpec;
+import java.util.ArrayList;
 import java.util.Base64;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.security.auth.x500.X500Principal;
 
@@ -53,8 +59,12 @@ public final class TenantObject extends JsonBackedValueObject {
 
     @JsonIgnore
     private Map<String, JsonObject> adapterConfigurations;
+
     @JsonIgnore
-    private TrustAnchor trustAnchor;
+    private List<TrustAnchor> trustAnchors;
+
+    @JsonIgnore
+    private Map<String, List<JsonObject>> trustConfigurations;
 
     /**
      * Adds a property to this tenant.
@@ -113,35 +123,33 @@ public final class TenantObject extends JsonBackedValueObject {
     }
 
     /**
-     * Gets the subject DN of this tenant's configured trusted
-     * certificate authority.
+     * Get the subject DN of the configured trusted CA.
      * 
      * @return The DN or {@code null} if no CA has been set.
      */
     @JsonIgnore
+    @Deprecated
     public X500Principal getTrustedCaSubjectDn() {
 
-        final JsonObject trustedCa = getProperty(TenantConstants.FIELD_PAYLOAD_TRUSTED_CA, JsonObject.class);
-        if (trustedCa == null) {
-            return null;
+        final Set<X500Principal> subjectNames = getTrustedCaSubjectDns();
+        if (subjectNames == null) {
+           return null;
         } else {
-            return Optional
-                    .ofNullable(getProperty(trustedCa, TenantConstants.FIELD_PAYLOAD_SUBJECT_DN, String.class))
-                    .map(dn -> new X500Principal(dn)).orElse(null);
+            return subjectNames.stream().findFirst().orElse(null);
         }
     }
 
     /**
-     * Sets the trusted certificate authority to use for authenticating
+     * Adds the trusted certificate authority to use for authenticating
      * devices of this tenant.
+     * @param publicKey The trusted CA's public key.
+     * @param subjectDn The trusted CA's subject DN.
      * 
-     * @param publicKey The CA's public key.
-     * @param subjectDn The CA's subject DN.
      * @return This tenant for command chaining.
      * @throws NullPointerException if any of the parameters is {@code null}.
      */
     @JsonIgnore
-    public TenantObject setTrustAnchor(final PublicKey publicKey, final X500Principal subjectDn) {
+    public TenantObject addTrustAnchor(final PublicKey publicKey, final X500Principal subjectDn) {
 
         Objects.requireNonNull(publicKey);
         Objects.requireNonNull(subjectDn);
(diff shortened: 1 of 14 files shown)


sophokles73:
@Alfusainey can we please start with the Tenant API specification and the changes to the trusted CA config property? Can you please add the changes to this PR?

Alfusainey:
> @Alfusainey can we please start with the Tenant API specification and the changes to the trusted CA config property? Can you please add the changes to this PR?

@sophokles73: i modified the PR to include the tenant spec changes. i intend to remove the `cert` property of the trusted CA JSON object in a later PR

sophokles73:
@Alfusainey can you please first rebase and resolve conflicts?

Alfusainey:
> @Alfusainey can you please first rebase and resolve conflicts?

done

site/content/api/Tenant-API.md
```diff
@@ -107,7 +107,7 @@ The table below provides an overview of the standard members defined for the JSO
...
 | *resource-limits*        | *no*      | *object*      | Any resource limits that should be enforced for the tenant, e.g. the maximum number of concurrent connections and the maximum data volume for a given period. Refer to [Resource Limits Configuration Format]({{< relref "#resource-limits-configuration-format" >}}) for details. |
 | *tenant-id*              | *yes*     | *string*      | The ID of the tenant. |
-| *trusted-ca*             | *no*      | *object*      | The trusted certificate authority to use for validating certificates presented by devices of the tenant for authentication purposes. See [Trusted Certificate Authority Format]({{< relref "#trusted-ca-format" >}}) for a definition of the content model of the object. |
+| *trusted-ca*             | *no*      | *object* or *array*     | The trusted certificate authority(ies) to use for validating certificates presented by devices of the tenant for authentication purposes. The *trusted-ca* property can contain either a single JSON object  or an array of JSON objects. See [Trusted Certificate Authority Format]({{< relref "#trusted-ca-format" >}}) for a definition of the content model of the object. |
```
sophokles73: now that we have removed all optional (management) operations from the Tenant API (they have gone into the new HTTP based management API), we should only use an *array* here. It is a different question, if and how the default implementation supports loading existing tenant information from file that still contains a single object (instead of an array) ...
IMHO we should never have an empty array, should we?

site/content/api/Tenant-API.md
```diff
@@ -151,17 +151,39 @@ The JSON structure below contains example information for tenant `TEST_TENANT`.
 
 ## Trusted CA Format
 
-The table below provides an overview of the members defined for the *trusted-ca* JSON object:
+The *trusted-ca* property of the tenant (JSON) payload can contain either a single JSON object containing the trusted CA to use for validating certificates of devices belonging to the tenant.
```
sophokles73: it should always be an array

site/content/api/Tenant-API.md
```diff
@@ -151,17 +151,39 @@ The JSON structure below contains example information for tenant `TEST_TENANT`.
...
++The tenant can also be configured with multiple trusted certificates. In this case, the *trusted-ca* property can contain an array of JSON objects, were each JSON object represents a single trusted CA. The table below provides an overview of the members defined in a JSON object contained in the *trust-ca* property:
 
 | Name                     | Mandatory  | Type          | Default Value | Description |
 | :------------------------| :--------: | :------------ | :------------ | :---------- |
 | *subject-dn*             | *yes*      | *string*      |               | The subject DN of the trusted root certificate in the format defined by [RFC 2253](https://www.ietf.org/rfc/rfc2253.txt). |
 | *cert*                   | *no*       | *string*      |               | The Base64 encoded binary DER encoding of the trusted root X.509 certificate. |
 | *public-key*             | *no*       | *string*      |               | The Base64 encoded binary DER encoding of the trusted root certificate's public key. |
+| *not-before*             | *no*       | *string*      |               | The property indicating that the trusted root X.509 certificate is not valid before the given date. |
```
sophokles73: please provide a link to the RFC that defines the date/time format used for this (and the next) property.

site/content/api/Tenant-API.md
```diff
@@ -151,17 +151,39 @@ The JSON structure below contains example information for tenant `TEST_TENANT`.
...
+| *not-after*              | *no*       | *string*      |               | The property indicating that the trusted root X.509 certificate is not valid after the given date. |
 | *algorithm*              | *no*       | *string*      | `RSA`        | The name of the public key algorithm. Supported values are `RSA` and `EC`. This property is ignored if the *cert* property is used to store a certificate. |
 
 * The *subject-dn* MUST be unique among all registered tenants.
 * Either the *cert* or the *public-key* MUST be set.
+* If the *cert* property is used to store a certificate then the *not-before*, *not-after* and *algorithm* properties are ignored.
```
sophokles73: I would rather not support the *cert* property at all. It is up to the implementation of the service how this information is kept in persistent store, but we only return this information as public key plus not before and not after ...

site/content/api/Tenant-API.md
```diff
@@ -151,17 +151,39 @@ The JSON structure below contains example information for tenant `TEST_TENANT`.
...
+~~~json
+ {
+   "tenant-id" : "TEST_TENANT",
+   "enabled" : true,
+   "trusted-ca": [ {
+           "subject-dn": "CN=ca,OU=Hono,O=Eclipse",
+           "public-key": "NOTAPUBLICKEY"
+         }, {
+           "subject-dn": "CN=ca,OU=Hono,O=Eclipse",
+           "public-key": "NOTAPUBLICKEY"
```
sophokles73: it looks like you can use multiple public keys sharing the same subject DN. I agree, that this makes sense. However, the example should also define e.g. overlapping validity periods in order to illustrate a meaningful use case for this.
Also, we should explicitly state that a subject DN still needs to be unique across tenants. Otherwise we could not resolve a tenant based on the subject DN anymore...

site/content/api/Tenant-API.md
Alfusainey: > IMHO we should never have an empty array, should we?

yes, thats true. so you are saying we should explicitly mention that in the spec?

sophokles73:
@Alfusainey are you going to make any of the suggested changes?

Alfusainey:
> @Alfusainey are you going to make any of the suggested changes?

@sophokles73 i just updated the PR

site/documentation/content/api/Tenant-API.md
```diff
@@ -152,36 +152,37 @@ The JSON structure below contains example information for tenant `TEST_TENANT`.
...
-| *cert*                   | *no*       | *string*      |               | The Base64 encoded binary DER encoding of the trusted root X.509 certificate. |
+| *subject-dn*             | *yes*      | *string*      |               | The subject DN of the trusted root certificate in the format defined by [RFC 2253](https://www.ietf.org/rfc/rfc2253.txt). The subject DN MUST be unique across tenants. |
 | *public-key*             | *no*       | *string*      |               | The Base64 encoded binary DER encoding of the trusted root certificate's public key. |
-| *not-before*             | *no*       | *string*      |               | The property indicating that the trusted root X.509 certificate is not valid before the given date. |
-| *not-after*              | *no*       | *string*      |               | The property indicating that the trusted root X.509 certificate is not valid after the given date. |
+| *not-before*             | *no*       | *string*      |               | The property indicating that the trusted root X.509 certificate is not valid before the given date. The date format is defined by [RFC 5280](https://www.ietf.org/rfc/rfc5280.txt). |
```
sophokles73: The X.509 RFC is not helpful here. See the definition of the data volume limit for the correct link/reference ...

site/documentation/content/api/Tenant-API.md
```diff
@@ -152,36 +152,37 @@ The JSON structure below contains example information for tenant `TEST_TENANT`.
...
 | *algorithm*              | *no*       | *string*      | `RSA`        | The name of the public key algorithm. Supported values are `RSA` and `EC`. This property is ignored if the *cert* property is used to store a certificate. |
 
 * The *subject-dn* MUST be unique among all registered tenants.
 * Either the *cert* or the *public-key* MUST be set.
-* If the *cert* property is used to store a certificate then the *not-before*, *not-after* and *algorithm* properties are ignored.
 
 **Examples**
 
-Below is an example for a payload of the response to a *get* request for tenant `TEST_TENANT` configured with two trusted certificate authorities.
+Below is an example for a payload of the response to a *get* request for tenant `TEST_TENANT`. The tenant is configured with two valid trusted certificates issued by the same issuer and having overlapping validity periods.
```
sophokles73: we don't know anything about the issuer. But the public keys are associated with the same *subject*

Alfusainey: thanks. I was confusing this with the issuer of client certificates which (usually) matches the subject-dn of the trusted CA.

Alfusainey: fixed in  4380d8c

site/documentation/content/api/Tenant-API.md
Alfusainey: fix in 4380d8c

site/documentation/content/api/Tenant-API.md
```diff
@@ -152,36 +152,37 @@ The JSON structure below contains example information for tenant `TEST_TENANT`.
...
+If the *trusted-ca* property is provided, then it MUST contain a non-empty array of JSON objects, were each JSON object represents a single trusted CA. The table below provides an overview of the members defined in a JSON object contained in the *trust-ca* property:
 
 | Name                     | Mandatory  | Type          | Default Value | Description |
 | :------------------------| :--------: | :------------ | :------------ | :---------- |
-| *subject-dn*             | *yes*      | *string*      |               | The subject DN of the trusted root certificate in the format defined by [RFC 2253](https://www.ietf.org/rfc/rfc2253.txt). |
-| *cert*                   | *no*       | *string*      |               | The Base64 encoded binary DER encoding of the trusted root X.509 certificate. |
+| *subject-dn*             | *yes*      | *string*      |               | The subject DN of the trusted root certificate in the format defined by [RFC 2253](https://www.ietf.org/rfc/rfc2253.txt). The subject DN MUST be unique across tenants. |
```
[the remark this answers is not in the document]
Alfusainey: fix in 4380d8c

site/documentation/content/api/Tenant-API.md
```diff
@@ -152,36 +152,37 @@ The JSON structure below contains example information for tenant `TEST_TENANT`.
 
 ### Trusted CA Format
 
-The *trusted-ca* property of the tenant (JSON) payload can contain either a single JSON object containing the trusted CA to use for validating certificates of devices belonging to the tenant.
-+The tenant can also be configured with multiple trusted certificates. In this case, the *trusted-ca* property can contain an array of JSON objects, were each JSON object represents a single trusted CA. The table below provides an overview of the members defined in a JSON object contained in the *trust-ca* property:
+If the *trusted-ca* property is provided, then it MUST contain a non-empty array of JSON objects, were each JSON object represents a single trusted CA. The table below provides an overview of the members defined in a JSON object contained in the *trust-ca* property:
```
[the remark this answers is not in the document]
Alfusainey: fix in 4380d8c

Alfusainey:
@sophokles73 any news on this?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -113,34 +123,16 @@ public TenantObject setEnabled(final boolean flag) {
...
+     * @param subjectDn The trusted CA's subject DN.
      * 
-     * @param publicKey The CA's public key.
-     * @param subjectDn The CA's subject DN.
      * @return This tenant for command chaining.
      * @throws NullPointerException if any of the parameters is {@code null}.
      */
     @JsonIgnore
-    public TenantObject setTrustAnchor(final PublicKey publicKey, final X500Principal subjectDn) {
+    public TenantObject addTrustAnchor(final PublicKey publicKey, final X500Principal subjectDn) {
```
sophokles73: shouldn't it also accept a not-before and a not-after `Instant`?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -543,4 +512,181 @@ public TenantObject setDefaults(final JsonObject defaultProperties) {
...
+     *            of properties for each configured trusted CA.
+     * @return This tenant for command chaining.
+     * 
+     * @throws IllegalArgumentException if the any of the trusted CA configuration is missing the required
+     *             <em>subject-dn</em> property or either the {@code public-key} or {@code cert} property does not
+     *             contain a valid Base64 encoding.
+     */
+    @SuppressWarnings("unchecked")
+    @JsonProperty(TenantConstants.FIELD_PAYLOAD_TRUSTED_CA)
+    public TenantObject setTrustedCa(final Object configuration) {
```
sophokles73: I have to admit that I do not really like the `Object` type parameter. In which case is this method invoked? Can we maybe replace it with two variants, one accepting a `JsonArray`, the other a `JsonObject`?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: @sophokles73 : recall that you mention this in #1095 
> 1. ...
> 2. extend the data model of the trusted CAs in the Tenant API with (optional) not-before and not-after properties. At the same time, remove support for storing a full X.509 certificate in the cert property.
> 3. ...

So I intend to add this change in a second PR, ideally following that breakdown we agreed on

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: > In which case is this method invoked?

this method is invoked by the `Jackson` library when mapping a `JsonObject` to a `TenantObject` i.e during a call to `JsonObject.mapTo(TenantObject.class)`

> Can we maybe replace it with two variants, one accepting a JsonArray, the other a JsonObject?

I actually tried this first, however, If I should separate it into two methods, then I cannot annotate both methods using the same property, i.e `trusted-ca`. If the annotation is missing, then the method will not be called during the mapping process. 

This is why I decided to put them in a single method and annotate the single method.

Alfusainey:
> I do not think that it makes a lot of sense to continue working on this PR unless #1240 has been merged because it affects a lot of the code you are also working on in this PR ...

OK, then I will continue to work on this after July 8th (I have an exam on the 8th :-))

sophokles73:
> OK, then I will continue to work on this after July 8th (I have an exam on the 8th :-))

Good luck!!!

Alfusainey:
@sophokles73 i rebased this PR to fix the conflicts due to f98f3fc. for some reason, only the ITs tests for the legacy DR implementation (i.e `tests-legacy-device-registry`) succeeds. since this change to the tenant API affects the new DR, i will create a separate issue for it and not include the fix in this PR

Alfusainey:
@sophokles73 revision 91cf13d fixes the test failures in the ITs for the new management API. those changes are required due to the change in the `trusted-ca` object model

site/documentation/content/api/Tenant-API.md
```diff
@@ -153,17 +153,37 @@ The JSON structure below contains example information for tenant `TEST_TENANT`.
...
-| *public-key*             | *no*       | *string*      |               | The Base64 encoded binary DER encoding of the trusted root certificate's public key. |
+| *subject-dn*             | *yes*      | *string*      |               | The subject DN of the trusted root certificate in the format defined by [RFC 2253](https://www.ietf.org/rfc/rfc2253.txt). Trusted CAs of the same tenant MAY share the same subject DN (e.g. allowing for the definition of overlapping validity periods). However, trusted CAs of different tenants MUST NOT share the same subject DN in order to allow for the look up of a tenant by the subject DN of one of its trusted CAs. |
+| *public-key*             | *yes*       | *string*     |               | The Base64 encoded binary DER encoding of the trusted root certificate's public key. |
+| *not-before*             | *no*       | *string*      |               | The property indicating that the trusted root X.509 certificate is not valid before the given date. The value MUST be an **ISO 8601 compliant** [*combined date and time representation*](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations) in extended format. |
```
sophokles73: I do not think that it makes much sense to include trust anchors in the list that are not valid at this particular moment, right? IMHO the service implementation should only return trust anchors that an adapter can use to verify the client cert's signature without the need to filter out those that are *not yet* or *no longer* valid. WDYT?

Alfusainey: sure 👍 

i think it should be enough if we document this in the API docs of `TenantObject#getTrustAnchors()` method. should we include that in the tenant API spec?

sophokles73: FMPOV this *has* to be reflected particularly in the Tenant API. In fact, it means that the trust anchor objects do not (need to) contain a *not-before* nor a *not-after* property because every trust anchor included is *per definitionem* valid and usable for the validation.

Alfusainey: hmm.. I am confused here..  let me break it down with my little understanding :-)

During tenant creation, all trust anchors (MUST/MAY/COULD)? be valid, but that might not necessary be the case in the future

Lets assume two trust actors (`A` and `B`) are added for a tenant (during tenant creation). trust anchor `A` will expire in 1 year while `B` will expire in 5 years (relative to the time instant in which the tenant was created). From the point the tenant is created and before `A` expires, a call to retrieve the trust anchors for the tenant should return `two valid` trust anchors.

Now lets assume that trust anchor `A` expires but it is still registered with the tenant. Now a call to retrieve the valid trust anchors for the tenant should return a single trust anchor. I think this is where we might need the not-before/not-after properties to perform the validity check?

sophokles73: Yes, but I would like the filtering of the relevant trust anchors to be done in the Tenant service implementation, not in the protocol adapter. Therefore, we need to have these properties in the HTTP based management API for tenants, but we do not need it in the payload that is sent to a client of the Tenant API.

Alfusainey: > Therefore, we need to have these properties in the HTTP based management API for tenants, but we do not need it in the payload that is sent to a client of the Tenant API.

I see now, thanks 👍  I think this distinction is very important. what I find confusing sometimes is the usage of the `TenantObject` to encapsulate information for a client sending a tenant add/update request. I think we should really restrict its usage only for the get operation and remove its occurrence in all tests cases (both in unit and integration tests) that have to do with sending tenant requests

sophokles73: that is my opinion as well and that is why I created #1360

Alfusainey: @sophokles73 I filed #1375. once that is merged, then I will revert revision 91cf13d and continue working on this PR

Alfusainey:
@sophokles73 this patch was for the legacy device registry impl. and is no longer required

closing...

PR #1290, opened 2019-06-14
[#1245] Add new Device Connection API classes

calohmn:
This is for #1245:

Add interfaces and base implementation classes for the new Device Connection API.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/DeviceConnectionClient.java
+++ b/client/src/main/java/org/eclipse/hono/client/DeviceConnectionClient.java
@@ -1,0 +1,67 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.client;
+
+import io.opentracing.SpanContext;
+import io.vertx.core.Future;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * A client for accessing Hono's Device Connection API.
+ * <p>
+ * An instance of this interface is always scoped to a specific tenant.
+ * <p>
+ * See Hono's <a href="https://www.eclipse.org/hono/api/device-connection-api/">
+ * Device Connection API specification</a> for a description of the result codes returned.
+ */
+public interface DeviceConnectionClient extends RequestResponseClient {
+
+    /**
+     * Sets the given gateway as the last gateway that acted on behalf of the given device.
+     * <p>
+     * If a device connects directly instead of through a gateway, the device identifier itself is to be used as value
+     * for the <em>gatewayId</em> parameter.
+     *
+     * @param deviceId The device id.
+     * @param gatewayId The gateway id (or the device id if the last message came from the device directly).
+     * @param context The currently active OpenTracing span or {@code null} if no span is currently active.
+     *            An implementation should use this as the parent for any span it creates for tracing
+     *            the execution of this operation.
+     * @return A future indicating whether the operation succeeded or not.
+     * @throws NullPointerException if device id or gateway id is {@code null}.
+     */
+    Future<Void> setLastKnownGatewayForDevice(String deviceId, String gatewayId, SpanContext context);
+
+    /**
+     * Gets the gateway that last acted on behalf of the given device.
+     * <p>
+     * If no last known gateway has been set for the given device yet, a failed future with status <em>Not Found</em>
+     * is returned.
+     *
+     * @param deviceId The device id.
+     * @param context The currently active OpenTracing span or {@code null} if no span is currently active.
+     *            An implementation should use this as the parent for any span it creates for tracing
+     *            the execution of this operation.
+     * @return A future indicating the result of the operation.
+     *         <p>
+     *         The future will succeed if a response with status 200 has been received from the device connection service.
+     *         In that case the value of the future will contain a <em>gateway-id</em> property with the
+     *         gateway id.
+     *         <p>
+     *         In case a status other then 200 is received, the future will fail with a
+     *         {@link ServiceInvocationException} containing the (error) status code returned by the service.
+     * @throws NullPointerException if device id is {@code null}.
+     */
+    Future<JsonObject> getLastKnownGatewayForDevice(String deviceId, SpanContext context);
+}
--- a/client/src/main/java/org/eclipse/hono/client/impl/DeviceConnectionClientImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/DeviceConnectionClientImpl.java
@@ -1,0 +1,217 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.client.impl;
+
+import java.net.HttpURLConnection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.UUID;
+
+import org.apache.qpid.proton.amqp.messaging.ApplicationProperties;
+import org.eclipse.hono.client.HonoConnection;
+import org.eclipse.hono.client.DeviceConnectionClient;
+import org.eclipse.hono.client.StatusCodeMapper;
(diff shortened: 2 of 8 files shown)


client/src/main/java/org/eclipse/hono/client/DeviceConnectionClient.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Sets the given gateway as the last gateway that acted on behalf of the given device.
+     * <p>
+     * If a device connects directly instead of through a gateway, the device identifier itself is to be used as value
+     * for the <em>gatewayId</em> parameter.
+     *
+     * @param deviceId The device id.
+     * @param gatewayId The gateway id (or the device id if the last message came from the device directly).
+     * @param context The currently active OpenTracing span. An implementation
```
sophokles73: can this be `null`?

client/src/main/java/org/eclipse/hono/client/DeviceConnectionClient.java
```diff
@@ -0,0 +1,67 @@
...
+     *         should use this as the parent for any span it creates for tracing
+     *         the execution of this operation.
+     * @return A future indicating whether the operation succeeded or not.
+     * @throws NullPointerException if tenant id, device id or gateway id is {@code null}.
```
sophokles73: there is no *tenant ID* parameter

client/src/main/java/org/eclipse/hono/client/DeviceConnectionClient.java
```diff
@@ -0,0 +1,67 @@
...
+     *         the execution of this operation.
+     * @return A future indicating the result of the operation.
+     *         <p>
+     *         The future will succeed if a response with status 200 has been received from the device connection service.
+     *         In that case the value of the future will contain a <em>gateway-id</em> property with the
+     *         gateway id.
+     *         <p>
+     *         In case a status other then 200 is received, the future will fail with a
+     *         {@link ServiceInvocationException} containing the (error) status code returned by the service.
+     * @throws NullPointerException if tenant id or device id is {@code null}.
```
sophokles73: there is no *tenant ID* parameter

client/src/main/java/org/eclipse/hono/client/DeviceConnectionClient.java
```diff
@@ -0,0 +1,67 @@
...
+    Future<Void> setLastKnownGatewayForDevice(String deviceId, String gatewayId, SpanContext context);
+
+    /**
+     * Gets the gateway that last acted on behalf of the given device.
+     * <p>
+     * If no last known gateway has been set for the given device yet, a failed future with status <em>Not Found</em>
+     * is returned.
+     *
+     * @param deviceId The device id.
+     * @param context The currently active OpenTracing span. An implementation
```
sophokles73: can this be `null`?

client/src/main/java/org/eclipse/hono/client/impl/DeviceConnectionClientImpl.java
```diff
@@ -0,0 +1,222 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(DeviceConnectionClientImpl.class);
+
+    /**
+     * Creates a new client for accessing the Device Connection service.
+     * <p>
+     * The client will be ready to use after invoking {@link #createLinks()} or
+     * {@link #createLinks(Handler, Handler)} only.
+     *
+     * @param connection The connection to Hono.
```
sophokles73: the connection to the Device Connection service

client/src/main/java/org/eclipse/hono/client/impl/DeviceConnectionClientImpl.java
```diff
@@ -0,0 +1,222 @@
...
+     * @param tenantId The identifier of the tenant for which the client should be created.
+     */
+    protected DeviceConnectionClientImpl(final HonoConnection connection, final String tenantId) {
+        super(connection, tenantId);
+    }
+
+    /**
+     * Creates a new client for accessing the Device Connection service.
+     *
+     * @param connection The connection to Hono.
```
sophokles73: the connection to the Device Connection service

client/src/test/java/org/eclipse/hono/client/impl/DeviceConnectionClientImplTest.java
```diff
@@ -0,0 +1,328 @@
...
+import io.vertx.proton.ProtonHelper;
+import io.vertx.proton.ProtonReceiver;
+import io.vertx.proton.ProtonSender;
+
+
+/**
+ * Tests verifying behavior of {@link DeviceConnectionClientImpl}.
+ *
+ */
+@RunWith(VertxUnitRunner.class)
```
sophokles73: can you please use JUnit5 for new unit tests

service-base/src/main/java/org/eclipse/hono/service/deviceconnection/BaseDeviceConnectionService.java
```diff
@@ -0,0 +1,210 @@
...
+
+        Objects.requireNonNull(request);
+
+        switch (DeviceConnectionConstants.DeviceConnectionAction.from(request.getOperation())) {
+        case GET_LAST_GATEWAY:
+            return processGetLastGatewayRequest(request);
+        case SET_LAST_GATEWAY:
+            return processSetLastGatewayRequest(request);
+        default:
+            return processCustomStateMessage(request);
```
sophokles73: CustomState?
should be CustomCommand or CustomOperation, shouldn't it?

client/src/main/java/org/eclipse/hono/client/impl/DeviceConnectionClientImpl.java
```diff
@@ -0,0 +1,222 @@
...
+        final Map<String, Object> properties = createDeviceIdProperties(deviceId);
+        properties.put(MessageHelper.APP_PROPERTY_GATEWAY_ID, gatewayId);
+
+        final Future<DeviceConnectionResult> resultTracker = Future.future();
+        createAndSendRequest(
+                DeviceConnectionConstants.DeviceConnectionAction.SET_LAST_GATEWAY.getSubject(),
+                properties,
+                null,
+                resultTracker,
+                null,
```
sophokles73: the intention here is to **never** cache any result, right?

client/src/main/java/org/eclipse/hono/client/impl/DeviceConnectionClientImpl.java
```diff
@@ -0,0 +1,222 @@
...
+    public Future<JsonObject> getLastKnownGatewayForDevice(final String deviceId, final SpanContext context) {
+        Objects.requireNonNull(deviceId);
+        final Future<DeviceConnectionResult> resultTracker = Future.future();
+
+        createAndSendRequest(
+                DeviceConnectionConstants.DeviceConnectionAction.GET_LAST_GATEWAY.getSubject(),
+                createDeviceIdProperties(deviceId),
+                null,
+                resultTracker,
+                null,
```
sophokles73: again: **never** cache any results?

client/src/main/java/org/eclipse/hono/client/impl/DeviceConnectionClientImpl.java
calohmn: yes

client/src/main/java/org/eclipse/hono/client/impl/DeviceConnectionClientImpl.java
calohmn: yes

client/src/main/java/org/eclipse/hono/client/impl/DeviceConnectionClientImpl.java
sophokles73: then IMHO we should also ignore any `Cache-Directive` header when creating the `DeviceConnectionResult` from the response message and should not support supplying a `CacheProvider` when creating the client, don't you think?

client/src/test/java/org/eclipse/hono/client/impl/DeviceConnectionClientImplTest.java
calohmn: Seeing that the similar `RegistrationClientImplTest`, `TenantClientImplTest` and `CredentialsClientImplTest` classes are still not migrated to JUnit5 as well, I would rather migrate all 4 classes together in a subsequent PR.

sophokles73: no need for a PR, just migrate the tests ...

calohmn:
@sophokles73 I've applied the changes according to the above comments (apart from the JUnit5 migration - see above).

client/src/main/java/org/eclipse/hono/client/DeviceConnectionClient.java
calohmn: Yes, I've adapted the javadoc to point this out.

client/src/main/java/org/eclipse/hono/client/DeviceConnectionClient.java
calohmn: Yes, I've adapted the javadoc to point this out.

client/src/main/java/org/eclipse/hono/client/impl/DeviceConnectionClientImpl.java
calohmn: I've amended the commit accordingly.

PR #1292, opened 2019-06-17
Fix grafana dashboard for JVM details

kaniyan:
In Grafana dashboard for JVM details, the values corresponding to the `Threads` and `Open Files` are always empty. In this PR, this has been addressed by fixing the query expressions with appropriate time series.


sophokles73:
In future PRs that are supposed to fix an issue, it would be very helpful if you could provide some description of the problem it tries to address/fix. That will make it easier to review your PRs ...

kaniyan:
@sophokles73  sure I will include some more descriptions.

Issue #1288, opened 2019-06-13
flowLatency delays creation of senders unnecessarily long


DanielMaier-BSI:
Hi,

when someone configures flowLatency in hono client, hono client will delay creation of a sender by flowLatency if no credits are available immediately. Given the following setup:

- flowLatency of 200ms leads to no credit issues in our infrastructure from time to time
- So I thought flowLatency of 1000ms should be really enough even in our setup

We now get "always" credits, however the creation of the sender gets delayed very often by a second. This causes for example that the device needs to wait 1 second for a response to a telemetry message or a command sender for the command response.

Would it be possible from a technical point of view to wait for credits only as long as required (as CreditBasedSender does)? Do you think this would be an improvement for hono client or do you just say "get your infrastructure fixed"? 

Even with this in place we would still have the "issue" that sender links with no corresponding receiver links need to wait the maximum amount of time.

sophokles73:
> Do you think this would be an improvement for hono client or do you just say "get your infrastructure fixed"?

You should always get your infrastructure fixed in any case :-) That said, I think that we could (should) improve the code for establishing a sender link in `HonoConnectionImpl.createSender()` so that it registers a *send queue drain handler* in case no credit is available yet. The timer for the *flow latency* would then only act to limit the time to wait for credit to an upper limit. However, if the drain handler fires first, the timer would be canceled.

Shouldn't be too much of an issue. A PR for this would be very welcome :-)

DanielMaier-BSI:
I will try to provide a PR next week.

One additional question out of curiosity: This solution would work fine for command response clients as they get newly created for every single response. As you mentioned in another issue for telemetry connection in protocol adapters this is not true. If there are multiple messages for the same tenant concurrently all but the first will still get a 503 error. Did you discuss or did you think about in the past to not postpone the creation of the sender but postpone sending of the messages, i.e. introduce a small message buffer only until the first credits were received?

sophokles73:
> If there are multiple messages for the same tenant concurrently all but the first will still get a 503 error. Did you discuss or did you think about in the past to not postpone the creation of the sender but postpone sending of the messages, i.e. introduce a small message buffer only until the first credits were received?

We did not introduce a *message buffer* but a request to create a downstream sender will be retried for several times if a concurrent request is in progress already. This should improve the situation to some extent.

sophokles73:
fixed in https://github.com/eclipse/hono/commit/8f50891a1520bb02cca1aa4bd729ec7193f3351c

PR #1295, opened 2019-06-18
Fix for #1288 flowLatency delays creation of senders unnecessarily long

DanielMaier-BSI:
PR for #1288. If code looks fine to you I will try to provide a test for this, tested it so far in our deployment with system tests.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
@@ -635,10 +635,22 @@ public class HonoConnectionImpl implements HonoConnection {
                         log.debug("sender open [target: {}, sendQueueFull: {}]", targetAddress, sender.sendQueueFull());
                         // wait on credits a little time, if not already given
                         if (sender.getCredit() <= 0) {
-                            vertx.setTimer(clientConfigProperties.getFlowLatency(), timerID -> {
-                                log.debug("sender [target: {}] has {} credits after grace period of {}ms", targetAddress,
-                                        sender.getCredit(), clientConfigProperties.getFlowLatency());
-                                senderFuture.tryComplete(sender);
+                            final long waitOnCreditsTimerId = vertx.setTimer(clientConfigProperties.getFlowLatency(),
+                                    timerID -> {
+                                        log.debug("sender [target: {}] has {} credits after grace period of {}ms",
+                                                targetAddress,
+                                                sender.getCredit(), clientConfigProperties.getFlowLatency());
+                                        sender.sendQueueDrainHandler(null);
+                                        senderFuture.tryComplete(sender);
+                                    });
+                            sender.sendQueueDrainHandler(replenishedSender -> {
+                                log.debug("sender [target: {}] has received {} initial credits",
+                                        targetAddress, replenishedSender.getCredit());
+                                if (vertx.cancelTimer(waitOnCreditsTimerId)) {
+                                    result.tryComplete(replenishedSender);
+                                    replenishedSender.sendQueueDrainHandler(null);
+                                } // otherwise the timer has already completed the future and cleaned up
+                                  // sendQueueDrainHandler
                             });
                         } else {
                             senderFuture.tryComplete(sender);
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -633,10 +633,20 @@ public void closeAndFree(
...
+                                            targetAddress,
+                                            sender.getCredit(), clientConfigProperties.getFlowLatency());
+                                    sender.sendQueueDrainHandler(null);
+                                    result.tryComplete(sender);
+                                });
+                        sender.sendQueueDrainHandler(replenishedSender -> {
+                            log.debug("sender [target: {}] has received {} initial credits",
+                                    targetAddress, replenishedSender.getCredit());
+                            vertx.cancelTimer(waitOnCreditsTimerId);
+                            replenishedSender.sendQueueDrainHandler(null);
```
sophokles73: this is not necessary as the registered handler will be invoked *exactly once* only anyway.

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -633,10 +633,20 @@ public void closeAndFree(
...
+                            result.tryComplete(replenishedSender);
```
sophokles73: IMHO it would be cleaner to actually check for the time having been cancelled:
```
if (vertx.cancelTimer(waitOnCreditsTimerId) {
  result.tryComplete(replenishedSender);
} // otherwise the timer has already completed the future
```

DanielMaier-BSI: OK.

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
DanielMaier-BSI: Where do you have this from? Can't find this in proton java docs or from quick view into the implementation. In https://github.com/eclipse/hono/blob/cc645489961ace208449cbbcb0788267ff5897c6/client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java#L263-L278 this behaviour seems to be implemented on top?

sophokles73: https://github.com/eclipse/hono/blob/1ef2c6af172b66771b4c8d712a14907a01086622/client/src/main/java/org/eclipse/hono/client/CreditBasedSender.java#L32

DanielMaier-BSI: But I set it on a `ProtonSender`.

sophokles73: Good catch :+1: 
My bad. Then you should definitely clear the drain handler like you already do!

DanielMaier-BSI:
I did the check when cancel timer and I also added unit tests.

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -633,10 +633,21 @@ public void closeAndFree(
...
+                                    sender.sendQueueDrainHandler(null);
+                                    result.tryComplete(sender);
+                                });
+                        sender.sendQueueDrainHandler(replenishedSender -> {
+                            log.debug("sender [target: {}] has received {} initial credits",
+                                    targetAddress, replenishedSender.getCredit());
+                            if (vertx.cancelTimer(waitOnCreditsTimerId)) {
+                                result.tryComplete(replenishedSender);
+                            } // otherwise the timer has already completed the future
+                            replenishedSender.sendQueueDrainHandler(null);
```
sophokles73: the `replenishedSender.sendQueueDrainHandler(null);` should also go into the if block, don't you think?

client/src/test/java/org/eclipse/hono/client/impl/HonoConnectionImplTest.java
```diff
@@ -113,7 +115,7 @@ public void setUp() {
     /**
      * Verifies that the client tries to connect a limited
      * number of times only.
-     * 
+     *
```
sophokles73: can you please configure your code formatter to **not** make these changes?

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
DanielMaier-BSI: I though safety first :). But I don't insist on it. I changed it accordingly.

PR #1298, opened 2019-06-18
[#1073] Change southbound control endpoint names

calohmn:
This is for #1073:

The default prefix for the southbound command & control endpoints has been renamed from `control` to `command`. The old `control` endpoints are still exposed and working.

(For the AMQP adapter, there is still one "command" endpoint used both for receiving commands and sending back the command response. Usage of a "command_response" endpoint for the responses shall be introduced following this commit.)

### DIFF
--- a/adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
+++ b/adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
@@ -400,7 +400,7 @@ public class VertxBasedAmqpProtocolAdapterTest {
         when(deviceConnection.attachments()).thenReturn(mock(Record.class));
         when(commandConsumerFactory.createCommandConsumer(eq(TEST_TENANT_ID), eq(TEST_DEVICE), any(Handler.class), any(Handler.class), anyLong()))
             .thenReturn(Future.succeededFuture(mock(MessageConsumer.class)));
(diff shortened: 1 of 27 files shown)


adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -851,6 +851,22 @@ public void testMessageLimitExceededForAnEventMessage(final TestContext ctx) {
...
+
+    /**
+     * Checks whether the legacy Command & Control endpoint shall be used.
+     * <p>
+     * Returns {@code false} by default. Subclasses may return {@code true} here to perform tests using the legacy
+     * command endpoint.
+     *
+     * @return {@code true} if the legacy command endpoint shall be used.
+     */
+    protected boolean useLegacyCommandEndpoint() {
```
sophokles73: what about using [parameterized tests](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests) for running the tests with the legacy and the new endpoint name instead of using a subclass and introducing this method?

calohmn: I had a look into that but since there is no parameterized test definition at class level in JUnit 5 yet, I decided against it (I didn't want to annotate each method).
When we remove support for the legacy endpoints, we can just delete the subclasses.

sophokles73: fair enough

sophokles73:
@calohmn please don't forget to also update the protocol adapter user guides accordingly (including deprecation notices) and to update the release notes ...

Issue #1297, opened 2019-06-18
Upgrade to Spring Boot 2


sophokles73:
Spring Boot 1 will run out of support August 1st 2019 as indicated in [this blog post](https://spring.io/blog/2018/07/30/spring-boot-1-x-eol-aug-1st-2019)

IMHO we should include this upgrade in the 1.0 release already.

This will also entail upgrading to Spring 5 AFAIK

kaniyan:
The latest spring boot GA release is [2.1.5](https://spring.io/projects/spring-boot#learn). Spring Framework 5.1.7.RELEASE or above is [required](https://docs.spring.io/spring-boot/docs/current/reference/html/getting-started-system-requirements.html).

sophokles73:
CQs have been approved.

sophokles73:
fixed via https://github.com/eclipse/hono/pull/1303

PR #1303, opened 2019-06-19
[#1297] Upgrade to Spring Boot 2

kaniyan:
**Not to merge until necessary CQs are created and approved.**

1. The spring artifacts are upgraded as below.

| **Artifact** | **Version** |
| ------------- | ------------- |
| Spring Boot | 2.1.5.RELEASE |
| Spring  | 5.1.8.RELEASE |
| Spring  Security | 5.1.5.RELEASE |

2. The below dependencies/plug-ins are updated to be in compatible with the above spring version.

| **Artifact** | **Version** |
| ------------- | ------------- |
| snakeyaml | 1.23 |
| maven-deploy-plugin  | 2.8.2 |
| maven-assembly-plugin |3.1.0 |

3. Support for Guava has been [dropped](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.0-Migration-Guide#guavacachemanager) in Spring Framework 5. Instead of `GuavaCacheManager`,  `CaffeineCacheManager` is being used and the corresponding dependency ([com.github.ben-manes.caffeine:caffeine:2.7.0](https://github.com/ben-manes/caffeine/tree/v2.7.0)) is added. It uses Apache 2.0 license.

4. [Changed](https://docs.spring.io/spring-boot/docs/2.1.5.RELEASE/reference/html/boot-features-external-config.html#boot-features-external-config-relaxed-binding) the prefix value in annotations to be in kebab case (lowercase and separated by -)

5. Spring Boot Maven plug-in configuration attributes are to [start](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.0-Migration-Guide#spring-boot-maven-plugin) with `spring-boot` as in the below example. It has been updated in the documentation.
```
mvn spring-boot:run -Dspring-boot.run.arguments=--hono.client.username=consumer@HONO,--hono.client.password=verysecret -Dspring-boot.run.profiles=command,ssl
```

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -34,6 +34,7 @@
     <dispatch-router.image.name>quay.io/enmasse/qdrouterd-base:1.7.0</dispatch-router.image.name>
     <grafana.version>5.3.2</grafana.version>
     <guava.version>25.0-jre</guava.version>
+    <caffeine.version>2.6.2</caffeine.version>
     <hamcrest-core.version>2.1</hamcrest-core.version>
     <jackson.version>2.9.8</jackson.version>
     <jaeger.version>0.32.0</jaeger.version>
@@ -56,9 +57,9 @@
     <qpid-jms.version>0.42.0</qpid-jms.version>
     <slf4j.version>1.7.26</slf4j.version>
     <snakeyaml.version>1.17</snakeyaml.version>
-    <spring.version>4.3.23.RELEASE</spring.version>
-    <spring-boot.version>1.5.20.RELEASE</spring-boot.version>
-    <spring-security-crypto.version>4.2.12.RELEASE</spring-security-crypto.version>
+    <spring.version>5.1.8.RELEASE</spring.version>
+    <spring-boot.version>2.1.5.RELEASE</spring-boot.version>
+    <spring-security-crypto.version>5.1.5.RELEASE</spring-security-crypto.version>
     <truth.version>0.28</truth.version>
     <vertx.version>3.7.0</vertx.version>
     <wiremock.version>2.22.0</wiremock.version>
@@ -402,6 +403,11 @@
             <artifactId>animal-sniffer-annotations</artifactId>
           </exclusion>
         </exclusions>
+      </dependency>
+      <dependency>
+        <groupId>com.github.ben-manes.caffeine</groupId>
+        <artifactId>caffeine</artifactId>
+        <version>${caffeine.version}</version>
       </dependency>
       <dependency>
         <groupId>io.opentracing</groupId>
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -56,7 +56,7 @@
     <proton.version>0.31.0</proton.version>
     <qpid-jms.version>0.42.0</qpid-jms.version>
     <slf4j.version>1.7.26</slf4j.version>
-    <snakeyaml.version>1.17</snakeyaml.version>
+    <snakeyaml.version>1.23</snakeyaml.version>
     <spring.version>5.1.8.RELEASE</spring.version>
     <spring-boot.version>2.1.5.RELEASE</spring-boot.version>
     <spring-security-crypto.version>5.1.5.RELEASE</spring-security-crypto.version>
--- a/legal/legal/LICENSE.Spring.txt
+++ b/legal/legal/LICENSE.Spring.txt
@@ -1,6 +1,6 @@
                                  Apache License
                            Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+                        https://www.apache.org/licenses/
 
    TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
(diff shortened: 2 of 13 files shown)


bom/pom.xml
```diff
@@ -34,6 +34,7 @@
     <dispatch-router.image.name>quay.io/enmasse/qdrouterd-base:1.7.0</dispatch-router.image.name>
     <grafana.version>5.3.2</grafana.version>
     <guava.version>25.0-jre</guava.version>
+    <caffeine.version>2.7.0</caffeine.version>
```
sophokles73: the idea is to have the list in alphabetical order ...
BTW does it also work with version 2.6.2? That would be easier to get approved ...

legal/legal/LICENSE.Spring.txt
```diff
@@ -202,9 +202,9 @@
 
 =======================================================================
 
-SPRING FRAMEWORK 4.3.18.RELEASE SUBCOMPONENTS:
+SPRING FRAMEWORK 5.1.8.RELEASE SUBCOMPONENTS:
```
sophokles73: did you just change the version in the existing file or did you replace the existing file with the corresponding file from the 5.1.8 artifacts?

legal/legal/NOTICE.Spring.txt
```diff
@@ -1,5 +1,5 @@
-Spring Framework 4.3.18.RELEASE
-Copyright (c) 2002-2018 Pivotal, Inc.
+Spring Framework 5.1.8.RELEASE
+Copyright (c) 2002-2019 Pivotal, Inc.
```
sophokles73: did you just change the version in the existing file or did you replace the existing file with the corresponding file from the 5.1.8 artifacts?

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
```diff
@@ -455,8 +455,8 @@ public HealthCheckServer healthCheckServer() {
      * @return The properties.
      */
     @Bean
-    @ConfigurationProperties(prefix = "hono.plan.prometheusBased")
-    @ConditionalOnProperty(name = "hono.plan.prometheusBased.host")
+    @ConfigurationProperties(prefix = "hono.plan.prometheus-based")
```
sophokles73: these properties are defined in the documentation, which you'll need to adapt accordingly as well ...

kaniyan: The above changes are needed only in the annotation side. The property `hono.plan.prometheus-based.host` or  `hono.plan.prometheusBased.host` can all be used as Spring Boot supports relaxed binding. So the properties such as  `hono.plan.prometheusBased.host` are still valid.

sophokles73: That might be true. Still, our documentation should define the property name that we are actually using in the code, don't you think? It is an added bonus that people who (still) use a similar but different property also succeed ...

kaniyan: yes, I will update it in the documentation too.

legal/legal/NOTICE.Spring.txt
kaniyan: I followed the same as for the spring 4.3.12.RELEASE update. I changed the version and the year in the above file. Also I compared it with the _notice.txt_ from the `spring-core-5.1.8.RELEASE.jar` artifact.

sophokles73: My point is, we cannot just update the year but indeed need to replace the existing file with the one from the 5.1.8.RELEASE jar (if it differs in anything else than just the year).

kaniyan: There is no difference from the file in 5.1.8.RELEASE jar except the version and the date.

bom/pom.xml
kaniyan: It works with version 2.6.2. I will push the changes.

legal/legal/LICENSE.Spring.txt
kaniyan: I have modified the existing file. In case of license file, I will replace it with the file from 5.1.8 artifact.

sophokles73:
@kaniyan can you rebase to resolve conflicts?

kaniyan:
> @kaniyan can you rebase to resolve conflicts?

@sophokles73 rebased.

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
sophokles73: I wasn't aware that this is required by Spring 2. I thought you deliberately chose to change the property prefix. Based on my new knowledge I think that we should keep the original camel case notation in our documentation for consistency. Would you mind changing back the property names in the docs to the original values? Sorry for the inconvenience ...

kaniyan: I have removed the changes regarding property names in the docs.

sophokles73:
Great work, @kaniyan 
Thanks a lot!

PR #1304, opened 2019-06-19
Add missing "algorithm" field to the trust-ca object

ctron:
Currently we are missing the field "algorithm" in the trust anchor configuration of the tenant.

### DIFF
--- a/site/documentation/static/api/device-registry-v1.yaml
+++ b/site/documentation/static/api/device-registry-v1.yaml
@@ -486,6 +486,17 @@ components:
                description: The Base64 encoded binary DER encoding of the
                             trusted root certificate. Either this property
                             or `public-key` must be set.
+            "algorithm":
+               type: string
+               description: The algorithm used for the public key of the CA.
+                            If the `cert` property is used to provide an
+                            X.509 certificate then the algorithm is determined
+                            from the certificate and this property is ignored.
+                            Otherwise, i.e. if the `public-key` property is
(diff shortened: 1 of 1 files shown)


site/documentation/static/api/device-registry-v1.yaml
```diff
@@ -486,6 +486,10 @@ components:
                description: The Base64 encoded binary DER encoding of the
                             trusted root certificate. Either this property
                             or `public-key` must be set.
+            "algorithm":
+               type: string
+               description: The algorithm of the key material.
```
sophokles73: I think this can use some more detail:
> The algorithm used for the public key of the CA. If the "cert" property is used to provide an X.509 certificate then the algorithm is determined from the certificate and this property is ignored. Otherwise, i.e. if the "public-key" property is used, this property must be set to the algorithm used if other than "RSA" (the default value).

site/documentation/static/api/device-registry-v1.yaml
```diff
@@ -486,6 +486,10 @@ components:
...
+               example: RSA
```
sophokles73: RSA or EC

ctron: Unfortunately multiple examples are still unsupported by the swagger UI: https://github.com/swagger-api/swagger-ui/issues/2651

I will use "RSA" as default and "EC" as example.

Issue #1286, opened 2019-06-13
HonoConnectionImpl throws NPE


ctron:
I just noticed this in the log:

~~~
11:49:18.669 [vert.x-eventloop-thread-0] DEBUG o.e.h.s.http.DefaultFailureHandler - unexpected internal failure
java.lang.NullPointerException: null
~~~

<details>
<summary>Full stack trace</summary>

~~~
11:49:18.669 [vert.x-eventloop-thread-0] DEBUG o.e.h.s.http.DefaultFailureHandler - unexpected internal failure
java.lang.NullPointerException: null
	at org.eclipse.hono.client.impl.HonoConnectionImpl.lambda$createReceiver$19(HonoConnectionImpl.java:685)
	at org.eclipse.hono.util.HonoProtonHelper.executeOrRunOnContext(HonoProtonHelper.java:183)
	at org.eclipse.hono.client.impl.HonoConnectionImpl.executeOrRunOnContext(HonoConnectionImpl.java:226)
	at org.eclipse.hono.client.impl.HonoConnectionImpl.createReceiver(HonoConnectionImpl.java:684)
	at org.eclipse.hono.client.impl.HonoConnectionImpl.createReceiver(HonoConnectionImpl.java:666)
	at org.eclipse.hono.client.impl.AbstractRequestResponseClient.createReceiver(AbstractRequestResponseClient.java:369)
	at org.eclipse.hono.client.impl.AbstractRequestResponseClient.createLinks(AbstractRequestResponseClient.java:351)
	at org.eclipse.hono.client.impl.CredentialsClientImpl.create(CredentialsClientImpl.java:158)
	at org.eclipse.hono.client.impl.CredentialsClientFactoryImpl.lambda$getOrCreateCredentialsClient$1(CredentialsClientFactoryImpl.java:69)
	at org.eclipse.hono.client.impl.CachingClientFactory.getOrCreateClient(CachingClientFactory.java:173)
	at org.eclipse.hono.client.impl.CachingClientFactory.getOrCreateClient(CachingClientFactory.java:146)
	at org.eclipse.hono.client.impl.CredentialsClientFactoryImpl.lambda$getOrCreateCredentialsClient$2(CredentialsClientFactoryImpl.java:67)
	at org.eclipse.hono.util.HonoProtonHelper.executeOrRunOnContext(HonoProtonHelper.java:183)
[...]

PR #1305, opened 2019-06-20
[#1286] Handle error when creating client

calohmn:
Relates to #1286.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/CachingClientFactory.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/CachingClientFactory.java
@@ -170,20 +170,29 @@ class CachingClientFactory<T> extends ClientFactory<T> {
             creationLocks.put(key, Boolean.TRUE);
             log.debug("creating new client for [{}]", key);
 
-            clientInstanceSupplier.get().setHandler(creationAttempt -> {
+            try {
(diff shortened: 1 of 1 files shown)


client/src/main/java/org/eclipse/hono/client/impl/CachingClientFactory.java
```diff
@@ -170,20 +170,29 @@ private void getOrCreateClient(
...
+                        log.debug("successfully created new client for [{}]", key);
+                        activeClients.put(key, newClient);
+                        result.handle(Future.succeededFuture(newClient));
+                    } else {
+                        log.debug("failed to create new client for [{}]", key, creationAttempt.cause());
+                        activeClients.remove(key);
+                        result.handle(Future.failedFuture(creationAttempt.cause()));
+                    }
+                });
+            } catch (final Exception ex) {
```
sophokles73: I assume this is supposed to catch an (unexpected runtime) exception thrown by the `clientInstanceSupplier.get()` method invocation, right?

calohmn: Yes, like the one in the description of #1286. 
So this is mainly about preventing the followup `already creating client for key` errors and making the original error more obvious.
See also https://github.com/eclipse/hono/issues/1286#issuecomment-503895722.

PR #1306, opened 2019-06-20
Handle errors occurring in receiver messageHandler

calohmn:
This is to make handling of incoming messages more robust.

Otherwise errors like this don't appear in the logs at all:
````
AMQPjava.lang.NullPointerException: null
AMQP	at java.base/java.util.Objects.requireNonNull(Objects.java:221)
AMQP	at org.eclipse.hono.util.ResourceIdentifier.fromString(ResourceIdentifier.java:154)
AMQP	at org.eclipse.hono.client.impl.CommandConsumerFactoryImpl.lambda$newTenantScopedCommandConsumer$10(CommandConsumerFactoryImpl.java:194)
AMQP	at org.eclipse.hono.client.impl.HonoConnectionImpl.lambda$createReceiver$15(HonoConnectionImpl.java:699)
AMQP	at io.vertx.proton.impl.ProtonReceiverImpl.onDelivery(ProtonReceiverImpl.java:206)
AMQP	at io.vertx.proton.impl.ProtonTransport.handleSocketBuffer(ProtonTransport.java:163)
AMQP	at io.vertx.core.net.impl.NetSocketImpl$DataMessageHandler.handle(NetSocketImpl.java:393)
AMQP	at io.vertx.core.streams.impl.InboundBuffer.handleEvent(InboundBuffer.java:230)
AMQP	at io.vertx.core.streams.impl.InboundBuffer.write(InboundBuffer.java:120)
AMQP	at io.vertx.core.net.impl.NetSocketImpl.handleMessage(NetSocketImpl.java:371)
AMQP	at io.vertx.core.net.impl.ConnectionBase.handleRead(ConnectionBase.java:390)
````
(The bug shown by this stacktrace has been fixed already.)

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
@@ -52,6 +52,7 @@ import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
 import io.vertx.proton.ProtonClientOptions;
 import io.vertx.proton.ProtonConnection;
+import io.vertx.proton.ProtonHelper;
 import io.vertx.proton.ProtonLink;
 import io.vertx.proton.ProtonMessageHandler;
 import io.vertx.proton.ProtonQoS;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java [hunk 1]
```diff
@@ -708,11 +709,16 @@ public void closeAndFree(
...
+                    try {
+                        messageHandler.handle(delivery, message);
+                        if (log.isTraceEnabled()) {
+                            final int remainingCredits = receiver.getCredit() - receiver.getQueued();
+                            log.trace("handling message [remotely settled: {}, queued messages: {}, remaining credit: {}]",
+                                    delivery.remotelySettled(), receiver.getQueued(), remainingCredits);
+                        }
+                    } catch (final Exception ex) {
+                        log.warn("error handling message", ex);
+                        ProtonHelper.modified(delivery, true, true, false);
```
sophokles73: I'd rather release the message here since we have no clue why the processing failed and we cannot indicate to the sender, what needs to be changed in order for the next attempt to succeed.

calohmn: Right, I've changed that accordingly.

sophokles73:
@calohmn please don't forget to set proper labels, project and milestone before merging ...

PR #1311, opened 2019-06-20
Fix links to the documentation in Javadoc.

b-abel:


b-abel:
> I think, though, that we should refrain from referring to anchors within pages wehre it is not strictly necessary in order to have more stable links, e.g. it should be sufficient to refer to the API pages instead of individual sections with the API page. WDYT?

Makes sense. Everyone probably thinks longer before changing names or location of the the markdown files than before changing a heading (from which the anchors are generated). So FMPOV the anchor is nowhere strictly necessary.
Let's see what the other code owners say, who are notified by Github :smile:

sophokles73:
@ctron any opinion regarding the usage of anchors into pages or not?

ctron:
Well I don't mind having anchors. If they break, no harm is done, as they will still load the page properly. On the other hand, getting pointed directly to the correct paragraph it indeed a benefit.

PR #1312, opened 2019-06-20
[#1073] Use "command_response" endpoint for AMQP command responses

calohmn:
This is for #1073:
A device, sending a command response to the AMQP adapter, should now use `command_response` instead of `command` as prefix in the response message address (as also mandated now by the `reply-to` property of the command message).
The old `control` prefix is still supported as well.


calohmn:
@sophokles73 The documentation update is in #1313.

PR #1314, opened 2019-06-21
Change credentials model and drop "delete" operation

ctron:
This change alters the model of the device registry HTTP API in way
that redundant parts of the secrets (type, auth-id, enabled) get pulled
up one level. And that new level gets the "ext" extensions as well.

The "enabled" flag is still present on the "secrets" level, as it might
be beneficial to e.g. disable a single secret temporarily.

### DIFF
--- a/site/documentation/static/api/device-registry-v1.yaml
+++ b/site/documentation/static/api/device-registry-v1.yaml
@@ -390,30 +390,6 @@ paths:
             412:
                $ref: '#/components/responses/ResourceVersionMismatch'
 
-      delete:
-         tags:
-            - credentials
-         summary: Delete all credentials of a device
-         description: |
-             Deletes all credentials from a device registration. The device is
-             still registered, and a following call to *get* credentials must
-             not return `404`, as the device is still registered. Instead an
-             empty result must be returned.
-         operationId: deleteAllCredentials
-         parameters:
-            - $ref: '#/components/parameters/resourceVersion'
-         responses:
-            204:
-               $ref: '#/components/responses/Deleted'
-            401:
-               $ref: '#/components/responses/Unauthorized'
-            403:
-               $ref: '#/components/responses/NotAllowed'
-            404:
-               $ref: '#/components/responses/NotFound'
-            412:
-               $ref: '#/components/responses/ResourceVersionMismatch'
-
 components:
 
    schemas:
--- a/site/documentation/static/api/device-registry-v1.yaml
+++ b/site/documentation/static/api/device-registry-v1.yaml
@@ -527,23 +527,25 @@ components:
 
       CredentialsSet:
          type: array
+         description: A set of credentials. The entries in this list must be
+                      unique by the composite key of `auth-id` and `type`.
          items:
-            $ref: '#/components/schemas/TypedSecret'
+            $ref: '#/components/schemas/TypedCredentials'
 
-      TypedSecret:
+      TypedCredentials:
          additionalProperties: false
          oneOf:
-            - $ref: '#/components/schemas/PasswordSecret'
-            - $ref: '#/components/schemas/PSKSecret'
-            - $ref: '#/components/schemas/X509CertificateSecret'
+            - $ref: '#/components/schemas/PasswordCredentials'
+            - $ref: '#/components/schemas/PSKCredentials'
+            - $ref: '#/components/schemas/X509CertificateCredentials'
          discriminator:
             propertyName: type
             mapping:
-               "hashed-password": '#/components/schemas/PasswordSecret'
-               "psk": '#/components/schemas/PSKSecret'
-               "x509-cert": '#/components/schemas/X509CertificateSecret'
+               "hashed-password": '#/components/schemas/PasswordCredentials'
+               "psk": '#/components/schemas/PSKCredentials'
+               "x509-cert": '#/components/schemas/X509CertificateCredentials'
 
-      CommonSecret:
+      CommonCredentials:
          type: object
          additionalProperties: false
          required:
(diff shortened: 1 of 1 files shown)


calohmn:
Having a fresh look at the resulting swagger api docs, I still find it a bit odd that the `credentials` definitions are on the top level and not below `devices`. 
Just from seeing `credentials` at the top level, I would wonder why there is no operation to *create* (and *delete*) credentials. The condition, that a set of credentials can't exist without its device, doesn't become clear here.
Having `credentials` placed below `/devices` would make that relation obvious and would make it clear for implementors of the API that when creating/deleting a device, corresponding credentials have to be created/deleted as well.
I know it's late in the discussion, but I think there would still be time to reevaluate the design here.
From what I got from the call today, there was the argument that there being an independent `Credentials API` would mean that there should also be a prominent, top-level credentials resource definition. However, having a REST API structure that properly reflects the relationships between resources would be better, IMHO.
Are there more points speaking for the top-level placement of the `credentials` resource?

ctron:
> I know it's late in the discussion, 

Yes indeed. A that is why I would be asking more for reasons, why not to do it the way we all agreed upon!

> Are there more points speaking for the top-level placement of the credentials resource?

I think "credentials" are a top level construct in the Hono system. Credentials do have a relationship between each other. A set of credentials may be "belong" to a device, but it also has a relationship, based on the "auth-id"+"type" key to other credentials. And while it may have a clear owner, I think it is a construct of its own.

sophokles73:
> A set of credentials may be "belong" to a device, but it also has a relationship, based on the "auth-id"+"type" key to other credentials.

I agree that credentials *always* belong to exactly one device of one particular tenant. What other relationship between credentials do you see apart from that?

When you design an object model you usually consider the lifecycle of objects that have a *part-of* relationship to each other in order to determine whether it is an aggregation or a  composition. The way I think about credentials I would say we are talking about a composition, i.e. a device has exclusive ownership of (its) credentials and when the device no longer exists, there is also no reason for the credentials to exist anymore (because they are not shared among devices). 

This fact has always been a topic in the discussion about how we want to design the APIs that the protocol adapters need to interact with and it has constantly been brought up that the lifecycle of the credentials are tied to the devices and thus it was not clear why we should have separate APIs for accessing (and back then also managing) them.

Now that we have separated the two concerns
* accessing the data from the adapters vs
* managing the content of the registry

I see an opportunity to make the management aspect more intuitive to use and indeed also express the fact that the credentials lifecycle is tied to the devices by means of moving the resources for updating credentials below the `devices` prefix. 

I also think that the fact that we initially agreed to the current version of the HTTP API spec shouldn't mean that we cannot reason about it and come to the conclusion that we should better change it *before* we release it if we feel that it can be improved. In particular if any later change would require us to create a new major release ...

ctron:
So maybe you can then create a PR that will reflect what you have in mind.

site/documentation/static/api/device-registry-v1.yaml
```diff
@@ -527,23 +527,25 @@ components:
...
-      TypedSecret:
+      TypedCredentials:
          additionalProperties: false
          oneOf:
-            - $ref: '#/components/schemas/PasswordSecret'
-            - $ref: '#/components/schemas/PSKSecret'
-            - $ref: '#/components/schemas/X509CertificateSecret'
+            - $ref: '#/components/schemas/PasswordCredentials'
+            - $ref: '#/components/schemas/PSKCredentials'
+            - $ref: '#/components/schemas/X509CertificateCredentials'
```
sophokles73: This looks like it would restrict the possible content to just these three types, wouldn't it? The original idea was to allow for any kind of secrets to be supported. Can this be defined in a way that allows additional types to be used as well?

dejanb: I'm not sure we could have a useful "type" for some unknown future enhancement. I think the spec should evolve with the new requirements when we know them

sophokles73: Of course, once we know about a new type, we can add it, sure. I was just wondering about what to do in the mean time if somebody implementing the API wants to use a (custom?) credentials type. If this spec is just meant for (informal) reference then I think we do not have a problem. However, if some payload validation is supposed to be based on the JSON schema then I think we should provide for extensibility and/or toleration of unexpected content. What's your take?

sophokles73:
> So maybe you can then create a PR that will reflect what you have in mind.

fair enough. Let me try to come up with something :-)

dejanb:
> In particular if any later change would require us to create a new major release ...

I don't see a problem with this. 

The thing is that this has been in design/implementation for weeks (if not months) now. Fundamental changes are putting a serious delay on the already complicated piece of work. This API doesn't have now any fundamental flows. I'm sure it will evolve over time anyhow, so let's do these big changes then.

sophokles73:
@dejanb I have expressed my concerns and suggestions for improving the API. However, my arguments have obviously not been strong enough to convince you guys, which is sad (for me) but also a valid outcome of the discussion :-)

I will then refrain from creating a PR ...

dejanb:
Thinking more about this if we follow this logic about resources, then only tenant is a top-level resources and everything else should live under it. But that would make this API much less usable as the current approach allows finer granulation of how implementation should implement those concepts.

site/documentation/static/api/device-registry-v1.yaml
dejanb: I'm not sure that we can have proper validation without a defined type and how that would help someone implement a registry without defined credential type. I of course, could be wrong. Can you explain how would you define this?

dejanb:
@sophokles73 Discussing  everything is always more than welcome. The thing is that this has been on the table for so long and throwing in some random changes while we wrapping up the implementation is not good process-wise FMPOV.

sophokles73:
@dejanb I do not think that @calohmn nor I had thrown in *random changes*. May be we just have different ideas of how an HTTP based API should look like. I would still like to understand what role the OpenApi document should play in reality and if you envision it to be used for formal validation and/or code generation. Again, if not, then my comment is irrelevant.

But anyway, I will follow your lead on this one and be happy if you merged this PR as-is.

dejanb:
@sophokles73 "random" is a poor choice of words, sorry about that :)
I think the spec should be used formally.

dejanb:
I merged this so we can continue working on the current PR. Let's discuss all future improvements once we have the current work merged.

PR #1315, opened 2019-06-21
Fix Twitter preview card.

b-abel:
Twitter does not support SVG for the image. Changed it to the landing page graphic. I looked up the dimensions and sizes at https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/summary-card-with-large-image.html and this seems to be good fit. The descriptions are supposed to unique per page, but the same goes for the general description meta tag and we anyways need to add them per page for SEO purposes. This will them be picked up for twitter as well.


b-abel:
Once the change is online, we can check it with the Twitter validator tool: https://cards-dev.twitter.com/validator

PR #1316, opened 2019-06-21
[#1245] Add basic DeviceConnectionService implementation

calohmn:
This is for #1245.

Adds a basic implementation of the `Device Connection API` to the Hono device registry component.


ctron:
This looks good in general to me. But wouldn't it make more sense to use something like a cache instead of a plain map. That way it would be easier to manage the limits allow cleanups of older entries.

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/MapBasedDeviceConnectionServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+     * @param ctx The vert.x context.
+     */
+    @Test
+    public void testSetAndGetLastKnownGatewayForDevice(final VertxTestContext ctx) {
+        final String deviceId = "testDevice";
+        final String gatewayId = "testGateway";
+        final Future<DeviceConnectionResult> setLastGwResult = Future.future();
+        svc.setLastKnownGatewayForDevice(DEFAULT_TENANT, deviceId, gatewayId, span, setLastGwResult);
+        setLastGwResult.compose(deviceConnectionResult -> {
+            assertEquals(HttpURLConnection.HTTP_NO_CONTENT, deviceConnectionResult.getStatus());
```
sophokles73: this needs to go into a `ctx.verify()` call, I guess ...

services/device-registry/src/test/java/org/eclipse/hono/deviceregistry/MapBasedDeviceConnectionServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+     */
+    @Test
+    public void testSetLastKnownGatewayForDeviceFailsIfLimitReached(final VertxTestContext ctx) {
+        props.setMaxDevicesPerTenant(1);
+        final String deviceId = "testDevice";
+        final String gatewayId = "testGateway";
+        final Future<DeviceConnectionResult> setLastGwResult = Future.future();
+        svc.setLastKnownGatewayForDevice(DEFAULT_TENANT, deviceId, gatewayId, span, setLastGwResult);
+        setLastGwResult.compose(deviceConnectionResult -> {
+            assertEquals(HttpURLConnection.HTTP_NO_CONTENT, deviceConnectionResult.getStatus());
```
sophokles73: `ctx.verify()`

site/documentation/content/admin-guide/device-registry-config.md
```diff
@@ -70,6 +70,7 @@ The following table provides an overview of the configuration variables and corr
 | `HONO_TENANT_SVC_RECEIVER_LINK_CREDIT`<br>`--hono.tenant.svc.receiverLinkCredit` | no | `100` | The number of credits to flow to a client connecting to the Tenant endpoint. |
 | `HONO_TENANT_SVC_SAVE_TO_FILE`<br>`--hono.tenant.svc.saveToFile` | no | `false` | When set to `true` the server will periodically write out the registered tenants to the file specified by the `HONO_TENANTS_SVC_TENANT_FILENAME` property. |
 | `HONO_VERTX_DNS_QUERY_TIMEOUT`<br>`--hono.vertx.dnsQueryTimeout` | no | `5000` | The amount of time after which a DNS query is considered to be failed. Setting this variable to a smaller value may help to reduce the time required to establish connections to the services this service depends on. However, setting it to a value that is too small for any DNS query to succeed will effectively prevent any connections to be established at all. |
+| `HONO_DEVICE_CONNECTION_SVC_MAX_DEVICES_PER_TENANT`<br>`--hono.deviceConnection.svc.maxDevicesPerTenant` | no | `100` | The number of devices per tenant for which connection related data is stored. It is an error to set this property to a value <= 0. |
```
sophokles73: can you move this to a position based on lexical order?

calohmn:
> This looks good in general to me. But wouldn't it make more sense to use something like a cache instead of a plain map. That way it would be easier to manage the limits allow cleanups of older entries.

I had experimented with using a Guava Cache in order to set up expiration of entries. But that brought up the question on what expiration duration to set as default here. There could be scenarios (even just for testing) where devices send messages only very sporadically. Therefore an expiration duration here shouldn't just be minutes or even hours. 
Seeing that the `FileBasedRegistrationService` has the `maxDevicesPerTenant` setting, I adopted this setting as a means to limit the size of the map in the `MapBasedDeviceConnectionService`. I don't think that further mechanisms (like entry expiration) to prevent OutOfMemoryExceptions here are necessary. In any case this implementation isn't meant for production anyway.

As to using a Guava Cache just for its `maximumSize` feature: I don't see that as much of an advantage and would rather use the `Map` interface with its `computeIfAbsent` and `compute` methods.

calohmn:
@sophokles73 I've made the corresponding changes.

Issue #1317, opened 2019-06-24
Lower entry barrier for Getting Started guide


sophokles73:
We do not want to support Docker Swarm any longer in the guide and would also like to make it easier for a potential user to experience Hono's functionality without the need to check out the source code, compile it and set up a local Docker Swarm to deploy to.

Instead
* use the Sandbox if user can access it from local network
* otherwise use Minikube to run a local Kubernetes cluster to deploy to

Also consider #1064 and always let the user create tenant, device and credentials.

PR #1318, opened 2019-06-24
[#1317] Use Minikube in Getting Started guide.

sophokles73:
The Getting Started guide is based on the Sandbox (if the user can access it) or a local Minikube instance.
In a next step I would then like to remove "official" support for Docker Swarm as well and remove the Swarm deploy scripts.

### DIFF
--- a/site/documentation/content/deployment/create-kubernetes-cluster.md
+++ b/site/documentation/content/deployment/create-kubernetes-cluster.md
@@ -1,6 +1,6 @@
 +++
 title = "Setting up a Kubernetes Cluster"
-weight = 485
+weight = 480
 +++
 
 This guide describes how to set up a Kubernetes cluster which can be used to run Eclipse Hono.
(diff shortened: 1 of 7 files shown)


site/documentation/content/deployment/create-kubernetes-cluster.md
```diff
@@ -10,19 +10,31 @@ Hono can be deployed to any Kubernetes cluster running version 1.11 or newer. Th
...
+* **cpus**: 2
+* **memory**: 8192 (MB)
+* **kubernetes-version**: v1.13.6
+
+The command to start the VM will look something like this:
+```sh
+minikube start --cpus 2 --memory 8192 --kubernetes-version v1.13.6 --vm-driver kvm2 -p hono
+```
+
+You may want to use another VM driver, depending on your OS. After the Minikube VM has started successfully, the `minikube tunnel` command should be run in order to support Hono's services being deployed using the *LoadBalancer* type. Please refer to the [Minikube Networking docs](https://github.com/kubernetes/minikube/blob/master/docs/networking.md#loadbalancer-emulation-minikube-tunnel) for details.
```
calohmn: The anchor in the linked minikube docs page seems to have changed. It is now:
https://github.com/kubernetes/minikube/blob/master/docs/networking.md#access-to-loadbalancer-services-using-minikube-tunnel

site/documentation/content/deployment/create-kubernetes-cluster.md
```diff
@@ -10,19 +10,31 @@ Hono can be deployed to any Kubernetes cluster running version 1.11 or newer. Th
...
 This kind of setup is sufficient for evaluation and development purposes.
 Please refer to Kubernetes' [Minikube setup guide](https://kubernetes.io/docs/setup/minikube/) for instructions on how to set up a cluster locally.
 
-When using Minikube, the `minikube tunnel` command should be run in order to support Hono's services being deployed using the *LoadBalancer* type. Please refer to the [Minikube Networking docs](https://github.com/kubernetes/minikube/blob/master/docs/networking.md#loadbalancer-emulation-minikube-tunnel) for details.
+The recommended settings for a Minikube VM used for running Hono's example setup are as follows:
 
-## Production Environment
+* **cpus**: 2
+* **memory**: 8192 (MB)
+* **kubernetes-version**: v1.13.6
```
calohmn: Instead of recommending a specific version (which could get outdated soon in terms of security fixes), I would rather omit the `kubernetes-version` option and let the default be used. Same for the `vm-driver` option.

site/homepage/content/getting-started.md
```diff
@@ -72,123 +107,116 @@ The second command creates and starts up Docker Swarm *services* for all compone
...
-http --auth sensor1@DEFAULT_TENANT:hono-secret POST http://localhost:8080/telemetry temp:=5
+13:36:49.169 [vert.x-eventloop-thread-0] INFO  org.eclipse.hono.cli.Receiver - received telemetry message [device: my-device, content-type: application/json]: {"temp": 5}
+13:36:49.170 [vert.x-eventloop-thread-0] INFO  org.eclipse.hono.cli.Receiver - ... with application properties: {orig_adapter=hono-http, device_id=my-device, orig_address=/telemetry, JMS_AMQP_CONTENT_TYPE=application/json}
 ~~~
 
-The username and password used above for device `4711` are part of the example configuration that comes with Hono. See [Device Identity](https://www.eclipse.org/hono/docs/latest/concepts/device-identity/) for an explanation of how devices are identified in Hono and how device identity is related to authentication.
+You can publish more data simply by re-running the `curl` command above with arbitrary payload.
 
-When you first invoke any of the two commands above after you have started up your Hono instance, you may get the following response:
+{{% note title="Service Unavailabe" %}}
```
calohmn: typo: Unavailable

site/documentation/content/deployment/helm-based-deployment.md
```diff
@@ -106,58 +106,41 @@ prometheus-operated                     ClusterIP      None             <none>
...
+The following command can then be used to check for the existence of the *DEFAULT_TENANT* which is created as part of the installation:
 
 ~~~sh
-# in directory: eclipse-hono-$VERSION/deploy/
-eval $(services.sh)
-~~~
-
-In order to upload telemetry data to Hono, a device needs to be registered with the system first.
-The following command registers a device with ID `4711` in the example device registry that is deployed with Hono:
+curl -sIXGET http://$REGISTRY_IP:28080/tenant/DEFAULT_TENANT
```
calohmn: Although the command seems to work, I guess there should be a space after the `X`:
`curl -sIX GET`

site/homepage/content/getting-started.md
```diff
@@ -1,67 +1,102 @@
...
+Note that running the command line client requires a Java 11 runtime environment being installed locally.
 
-If you do not already have a working Maven installation on your system, please follow the [installation instructions on the Maven home page](https://maven.apache.org/).
+#### Hono Sandbox
 
-Then run the following from the project's root folder
+The most important prerequisite is, of course, a Hono instance that you can work with.
 
-    mvn clean install -Ddocker.host=tcp://${host}:${port} -Pbuild-docker-image,metrics-prometheus
+The most straight forward option to use for this guide is the [Hono Sandbox]({{< relref "sandbox" >}}) which is running on
```
calohmn: _straightforward_

site/documentation/content/deployment/create-kubernetes-cluster.md
sophokles73: I intended to set kind of a baseline regarding the minimum version we would be willing to support. If we do not set a particular version here, then Minikube will choose the version for us and we might end up with a version that we haven't test yet (too new) or we know does not work (too old?).
WDYT?

calohmn: Well, I wouldn't expect much issues because of a "too new" version, actually.
But, of course we can't be sure. So, ok, with specifying the kubernetes-version we are on the safe side. 
I still would remove the `--vm-driver` option though, letting the default be used.

PR #1320, opened 2019-06-24
[#1245] Use Device Connection API for the gateway mapping feature

calohmn:
This is for #1245.

With the changes here, the protocol adapters invoke the `setLastKnownGatewayForDevice` method of the Device Connection API whenever a device (that has one or more `via` gateways defined) sends a telemetry, event, command response or ttd event.

The gateway mapping logic (that is part of processing a command message in the protocol adapter) now uses the `getLastKnownGatewayForDevice` method.

The old code for storing a `last-via` property in the device registration data is still present and will be removed later.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/GatewayMapperImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/GatewayMapperImpl.java
@@ -13,64 +13,179 @@
 
 package org.eclipse.hono.client.impl;
 
-import java.util.Objects;
+import java.net.HttpURLConnection;
 
+import org.eclipse.hono.client.ClientErrorException;
+import org.eclipse.hono.client.ConnectionLifecycle;
+import org.eclipse.hono.client.DeviceConnectionClientFactory;
+import org.eclipse.hono.client.DisconnectListener;
 import org.eclipse.hono.client.GatewayMapper;
 import org.eclipse.hono.client.HonoConnection;
+import org.eclipse.hono.client.ReconnectListener;
 import org.eclipse.hono.client.RegistrationClientFactory;
-import org.eclipse.hono.util.Constants;
+import org.eclipse.hono.client.ServerErrorException;
+import org.eclipse.hono.client.ServiceInvocationException;
+import org.eclipse.hono.tracing.TracingHelper;
+import org.eclipse.hono.util.DeviceConnectionConstants;
+import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.RegistrationConstants;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import io.opentracing.References;
+import io.opentracing.Span;
 import io.opentracing.SpanContext;
+import io.opentracing.Tracer;
+import io.opentracing.tag.Tags;
+import io.vertx.core.AsyncResult;
+import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 
 /**
  * A component that maps a given device to the gateway through which data was last published for the given device.
  */
-public class GatewayMapperImpl extends ConnectionLifecycleWrapper<HonoConnection> implements GatewayMapper {
+public class GatewayMapperImpl implements GatewayMapper, ConnectionLifecycle<HonoConnection> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(GatewayMapperImpl.class);
 
     private final RegistrationClientFactory registrationClientFactory;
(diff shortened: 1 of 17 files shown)


service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -925,6 +973,82 @@ protected void onCommandConnectionEstablished(final HonoConnection commandConnec
...
+     *            <p>
+     *            If not {@code null} then the authenticated device is compared to the given tenant and device ID. If
+     *            they differ in the device identifier, then the authenticated device is considered to be a gateway
+     *            acting on behalf of the device.
+     * @param context The currently active OpenTracing span that is used to trace the operation.
+     * @return The assertion.
+     * @throws NullPointerException if any of tenant or device ID are {@code null}.
+     */
+    protected final Future<JsonObject> updateLastGateway(
+            final JsonObject assertRegistrationJson,
```
sophokles73: can we rename this param to *registrationAssertion*?

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -925,6 +973,82 @@ protected void onCommandConnectionEstablished(final HonoConnection commandConnec
...
+     *            they differ in the device identifier, then the authenticated device is considered to be a gateway
+     *            acting on behalf of the device.
+     * @param context The currently active OpenTracing span that is used to
+     *                trace the retrieval of the assertion.
+     * @return The assertion.
+     * @throws NullPointerException if any of tenant or device ID are {@code null}.
+     * @see #getRegistrationAssertion(String, String, Device, SpanContext)
+     * @see #updateLastGateway(JsonObject, String, String, Device, SpanContext)
+     */
+    protected final Future<JsonObject> getRegistrationAssertionWithLastGatewayUpdate(
```
sophokles73: why do we need this extra method? Can't we simply use/alter the existing *getRegistrationAssertion* method?

calohmn: Well, yes. We could use the existing `getRegistrationAssertion` method.
My idea here was to make it more explicit what is actually being done in this method. 
But, on second thought, seeing that the long name is somewhat unwieldy and that M5 migration for existing custom protocol adapters will be easier using the existing method, I can change that, adding a note in the javadoc about the extra functionality.

calohmn:
@sophokles73 I've amended the commit accordingly.

client/src/main/java/org/eclipse/hono/client/impl/GatewayMapperImpl.java
```diff
@@ -13,64 +13,179 @@
...
+    public void addReconnectListener(final ReconnectListener<HonoConnection> listener) {
+        registrationClientFactory.addReconnectListener(listener);
+        deviceConnectionClientFactory.addReconnectListener(listener);
+    }
+
+    @Override
+    public Future<Void> isConnected() {
+        final Future<Void> registrationFuture = registrationClientFactory.isConnected();
+        final Future<Void> deviceConnectionFuture = deviceConnectionClientFactory.isConnected();
+        return CompositeFuture.all(registrationFuture, deviceConnectionFuture).map(cf -> null);
```
ctron: Why not use `.mapEmpty()`?

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -922,7 +969,53 @@ protected void onCommandConnectionEstablished(final HonoConnection commandConnec
...
+            return Future.succeededFuture(registrationAssertion);
+        }
+
+        final Future<String> gatewayId = getGatewayId(tenantId, deviceId, authenticatedDevice);
+
+        return gatewayId
+                .compose(gwId -> getDeviceConnectionClient(tenantId))
+                .compose(client -> client.setLastKnownGatewayForDevice(deviceId,
+                        Optional.ofNullable(gatewayId.result()).orElse(deviceId), context))
+                .map(r -> registrationAssertion);
```
ctron: Why not use `map(value)` here?

client/src/main/java/org/eclipse/hono/client/impl/GatewayMapperImpl.java
calohmn: Good, I've changed that.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
calohmn: Good, I've changed that.

PR #1322, opened 2019-06-26
Implement command and control for Sigfox

ctron:


adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -935,6 +976,20 @@ private void setResponsePayload(final HttpServerResponse response, final Command
...
+     * <p>
+     * The default implementation will call {@link Command#isValid()}. Protocol adapters may override this, but should
+     * consider calling the super method.
+     * 
+     * @param command The command to validate, will never be {@code null}.
+     * @param currentSpan The current tracing span.
+     * @return {@code true} if the command is valid, {@code false} otherwise.
+     */
+    protected boolean isCommandValid(final Command command, final Span currentSpan) {
+        return command.isValid();
```
sophokles73: are you going to also log the outcome of the check to the span?

core/src/main/java/org/eclipse/hono/util/Constants.java
```diff
@@ -203,15 +203,22 @@ public Authorities getAuthorities() {
...
+
+    /**
+     * The header name defined for setting the <em>time to deliver</em> for device command readiness notification events.
+     */
+    public static final String HEADER_TIME_TILL_DISCONNECT = "hono-ttd";
+
     /**
      * The header name defined for setting the <em>time to deliver</em> for device command readiness notification events.
+     * 
+     * @deprecated Use {@link #HEADER_TIME_TILL_DISCONNECT} instead.
```
sophokles73: FMPOV *TIME_TIL_DISCONNECT* is the correct spelling, since it is the short form of *TIME_UNTIL_DISCONNECT* or is there another reason to change the name of the constant?

calohmn: It looks like there's the 3 choices `until`, `till` and `'til` (see [here](https://www.merriam-webster.com/words-at-play/should-you-use-until-or-till-or-til)). 
Since we can't put the apostrophe in the constant here, I would opt for `till` with 2 L as well.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
ctron: I simply refactored that method for now, but that could be an option. However, doing that here would be wrong place. As others, overriding this method, would need to actually do that, when they call into the super method here.

core/src/main/java/org/eclipse/hono/util/Constants.java
ctron: IMHO "til" is definitely wrong:

* https://www.lexico.com/en/definition/till
* https://www.lexico.com/en/definition/until
* https://www.lexico.com/en/definition/til

So I would go for "till" or  "until". I don't mind having typos in various places, but I think in the more official APIs, we should fix it.

ctron:
All of those changes are actually triggered by the development of the sigfox adapter. That is the motivation behind those changes. And thus it is a bit hard to make them, without the context of that adapter.

I agree that it might make sense to have those changes in any case. But pulling them out only makes sense after the PR is approved, as then all changes are in that were necessary.

The only exception to that are the changes around "til". I will pull those changes out of this PR, and create a new one, as they touch quite a few files.

ctron:
Created PR #1331 … I will rebase this PR as soon as the other is merged.

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/SigfoxProtocolAdapter.java
```diff
@@ -57,6 +64,12 @@
 
     private static final String SIGFOX_PARAM_TENANT = "tenant";
 
+    private static final String SIGFOX_HEADER_PARAM_ACK = "ack";
+
+    private static final String SIGFOX_QUERY_PARAM_ACK = "ack";
+
+    private static final String CONTENT_TYPE_JSON = "application/json";
```
sophokles73: you may want to use `HttpUtils.CONTENT_TYPE_JSON` instead ...

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/SigfoxProtocolAdapter.java
```diff
@@ -185,4 +199,95 @@ private static Buffer decodeData(final String data) {
...
+
+    @Override
+    protected boolean isCommandValid(final Command command, final Span span) {
+
+        if (!super.isCommandValid(command, span)) {
+            return false;
+        }
+
+        final String conentType = command.getContentType();
+        if (!CONTENT_TYPE_OCTET_STREAM.equals(conentType)) {
```
sophokles73: does SigFox require this particular content type?

ctron: Yes. The command/downstream format is actually pretty strict. It has to be exactly 8 bytes.

sophokles73: Then I would suggest to simply ignore whatever has been set on the command AMQP message and just always use this content type when forwarding to the SigFox provider.

ctron: Well … I wanted to ensure that users are aware of the requirements. Assuming someone would just stick e.g. JSON in there, having 8 bytes by accident, it would work, and the next message (not having 8 bytes) would not. In this case it is clear: 8 bytes of binary data.

And actually the Sigfox backend expects a very specific JSON structure. Which will wrap the payload for the device.

If you think there is a benefit in ignoring the content type, then I will do that. But from a "setting the right expectations", it seemed like a good idea.

sophokles73: Not sure if I am getting it correctly.
I understand that the payload is expected to be exactly 8 bytes, right? You are already verifying this, so either the payload is ok regarding size, or it is not. This doesn't seem to be related to whether the payload bytes represent a JSON string or something else, or is it?
My understanding is also that the SigFox adapter will expect to receive the (raw) payload bytes wrapped in a specific JSON structure. However, that JSON structure is created by the adapter, not the application that has sent the (raw) payload bytes in the body of a command, right?
So, the adapter will *always* take the (raw) bytes from the body of the command and (base64 ?) encode them and put them in the JSON structure it sends to the SigFox back end, right?

ctron: > So, the adapter will always take the (raw) bytes from the body of the command and (base64 ?) encode them and put them in the JSON structure it sends to the SigFox back end, right?

It is hex encoded, but yes.

sophokles73: Then I do not really see the issue that you describe with the content type. The payload is *always* 8 bytes, regardless of whether it represents bit vectors of JSON. The receiver of the data will need to know how to interpret this data. The content-type might give a clue to the receiver regarding the structure and encoding of the payload data. However, SigFox doesn't seem to allow you to provide a more specific content type than *it's just bytes*. So, FMPOV, there is nothing wrong with always setting the same content type expected by SigFox in this case.

A sender of a command will need to be aware of the fact that it talks to a SigFox device in any case, I guess ...

ctron: I simply dropped the check, as it seems that is important to you.

ctron:
Rebased and squashed.

adapters/sigfox-vertx/src/main/java/org/eclipse/hono/adapter/sigfox/impl/SigfoxProtocolAdapterProperties.java
```diff
@@ -0,0 +1,35 @@
...
+ * Enhanced configuration properties for {@link SigfoxProtocolAdapter}.
+ */
+public class SigfoxProtocolAdapterProperties extends HttpProtocolAdapterProperties {
+
+    private static final int DEFAULT_TTD_WHEN_ACK_REQUIRED = 20;
+
+    private int ttdWhenAckRequired = DEFAULT_TTD_WHEN_ACK_REQUIRED;
+
+    public void setTtdWhenAckRequired(final int ttdWhenAckRequired) {
+        this.ttdWhenAckRequired = ttdWhenAckRequired;
```
sophokles73: does it make sense to check for e.g. positive values only?
you might also want to declare the getter/setter final?

ctron: I added a check to reject negative values. The hono stack already handles that, but an additional check doesn't hurt. Zero is already handled.

I don't see a ready to make this final, if someone wants to override this, then this is fine to me.

PR #1323, opened 2019-06-26
[#1255] Let CommandClient instance be scoped to a tenant

calohmn:
This is for #1255.

With this PR, the `CommandClient` is scoped to a tenant, not a device anymore. The device id is given as a parameter to the individual `sendCommand` invocation.

Note: The `AsyncCommandClient` is not changed here - this will come in a follow-up commit.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/ApplicationClientFactory.java
+++ b/client/src/main/java/org/eclipse/hono/client/ApplicationClientFactory.java
@@ -135,53 +135,46 @@ public interface ApplicationClientFactory extends ConnectionLifecycle<HonoConnec
             BiConsumer<ProtonDelivery, Message> consumer, Handler<Void> closeHandler);
 
     /**
-     * Gets a client for sending commands to a device.
+     * Gets a client for sending commands to devices of the given tenant.
      * <p>
      * The client returned may be either newly created or it may be an existing
-     * client for the given device.
+     * client for the given tenant.
      * <p>
      * This method will use an implementation specific mechanism (e.g. a UUID) to create
-     * a unique reply-to address to be included in commands sent to the device. The protocol
-     * adapters need to convey an encoding of the reply-to address to the device when delivering
-     * the command. Consequently, the number of bytes transferred to the device depends on the
-     * length of the reply-to address being used. In situations where this is a major concern it
-     * might be advisable to use {@link #getOrCreateCommandClient(String, String, String)} for
-     * creating a command client and provide custom (and shorter) <em>reply-to identifier</em>
+     * a unique reply-to address to be included in commands sent to devices of the tenant.
+     * The protocol adapters need to convey an encoding of the reply-to address to the device
+     * when delivering the command. Consequently, the number of bytes transferred to the device
+     * depends on the length of the reply-to address being used. In situations where this is a
+     * major concern it might be advisable to use {@link #getOrCreateCommandClient(String, String)}
+     * for creating a command client and provide custom (and shorter) <em>reply-to identifier</em>
      * to be used in the reply-to address.
      *
-     * @param tenantId The tenant that the device belongs to.
-     * @param deviceId The device to send the commands to.
+     * @param tenantId The tenant of the devices to which commands shall be sent.
      * @return A future that will complete with the command and control client (if successful) or
      *         fail if the client cannot be created, e.g. because the underlying connection
      *         is not established or if a concurrent request to create a client for the same
-     *         tenant and device is already being executed.
+     *         tenant is already being executed.
      * @throws NullPointerException if the tenantId is {@code null}.
      */
-    Future<CommandClient> getOrCreateCommandClient(String tenantId, String deviceId);
+    Future<CommandClient> getOrCreateCommandClient(String tenantId);
 
     /**
-     * Gets a client for sending commands to a device.
+     * Gets a client for sending commands to devices of the given tenant.
      * <p>
      * The client returned may be either newly created or it may be an existing
-     * client for the given device.
+     * client for the given tenant and replyId.
      *
-     * @param tenantId The tenant that the device belongs to.
-     * @param deviceId The device to send the commands to.
-     * @param replyId An arbitrary string which (in conjunction with the tenant and device ID) uniquely
-     *                identifies this command client.
-     *                This identifier will only be used for creating a <em>new</em> client for the device.
-     *                If this method returns an existing client for the device then the client will use
-     *                the reply-to address determined during its initial creation. In particular, this
-     *                means that if the (existing) client has originally been created using the
-     *                {@link #getOrCreateCommandClient(String, String)} method, then the reply-to address
-     *                used by the client will most likely not contain the given identifier.
+     * @param tenantId The tenant of the devices to which commands shall be sent.
+     * @param replyId An arbitrary string which will be used to create the reply-to address to be included in
+     *                commands sent to devices of the tenant. The combination of tenant and replyId has to be
+     *                unique among all CommandClient instances to make sure command response messages can be received.
      * @return A future that will complete with the command and control client (if successful) or
      *         fail if the client cannot be created, e.g. because the underlying connection
      *         is not established or if a concurrent request to create a client for the same
-     *         tenant and device is already being executed.
+     *         tenant and replyId is already being executed.
      * @throws NullPointerException if the tenantId is {@code null}.
      */
-    Future<CommandClient> getOrCreateCommandClient(String tenantId, String deviceId, String replyId);
+    Future<CommandClient> getOrCreateCommandClient(String tenantId, String replyId);
 
     /**
      * Gets a client for sending commands to a device asynchronously, i.e. command responses get received by a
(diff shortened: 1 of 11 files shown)


client/src/main/java/org/eclipse/hono/client/ApplicationClientFactory.java
```diff
@@ -145,19 +145,18 @@ static ApplicationClientFactory create(final HonoConnection connection) {
      * adapters need to convey an encoding of the reply-to address to the device when delivering
      * the command. Consequently, the number of bytes transferred to the device depends on the
      * length of the reply-to address being used. In situations where this is a major concern it
-     * might be advisable to use {@link #getOrCreateCommandClient(String, String, String)} for
+     * might be advisable to use {@link #getOrCreateCommandClient(String, String)} for
```
sophokles73: IMHO the whole JavaDoc needs to be revised to reflect that the client is no longer scoped to a specific device anymore.

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -122,14 +118,42 @@ protected AbstractRequestResponseClient(
...
+     * @param connection The connection to the service.
+     * @param tenantId The tenant that the client should be scoped to or {@code null} if the
+     *                 client should not be scoped to a tenant.
+     * @param replyId The replyId to use in the reply-to address.
+     * @throws NullPointerException if any of context or configuration are {@code null}.
+     */
+    protected AbstractRequestResponseClient(
+            final HonoConnection connection,
+            final String tenantId,
+            final String replyId) {
```
sophokles73: I do not think that we need to provide subclasses with a way to set a specific reply-to ID. IMHO we can always simply use a UUID here.

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -323,17 +347,16 @@ protected abstract R getResult(
...
-     * Subclasses may override this method in order to use a different address for creating the sender link.
-     * For example, an empty address could be returned here to create an anonymous relay link.
+     * Subclasses may override this method in order to use a different address as default for sending messages.
      *
-     * @return The link target address.
+     * @return The message target address.
      */
-    protected String getLinkTargetAddress() {
-        return targetAddress;
+    protected String getDefaultMessageTargetAddress() {
```
sophokles73: FMPOV this should be named *getMessageTargetAddress*. By default, it returns the link target address, but that can be changed by subclasses, right? However, this method is used to determine the value to put into the *to* property of the request message, right?

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -510,13 +531,31 @@ private R getRequestResponseResult(final Message message) {
...
     private Message createMessage(final String subject, final Map<String, Object> appProperties) {
+        return createMessage(subject, getDefaultMessageTargetAddress(), appProperties);
+    }
+
+    /**
+     * Build a Proton message with a provided subject (serving as the operation that shall be invoked).
+     * The message can be extended by arbitrary application properties passed in.
+     *
+     * @param subject The subject system property of the message.
+     * @param requestTargetAddress The target address of the message.
```
sophokles73: this is just the *address* of the message (the value that goes into the *to* property), I would say.

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -510,13 +531,31 @@ private R getRequestResponseResult(final Message message) {
...
+            final Map<String, Object> appProperties) {
 
         Objects.requireNonNull(subject);
         final Message msg = ProtonHelper.message();
         final String messageId = createMessageId();
         AbstractHonoClient.setApplicationProperties(msg, appProperties);
-        if (!targetAddress.equals(getLinkTargetAddress())) {
-            msg.setAddress(targetAddress);
+        if (!requestTargetAddress.equals(linkTargetAddress)) {
+            msg.setAddress(requestTargetAddress);
```
sophokles73: is there a problem with *always* setting the message's address to the given requestTargetAddress?

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -527,6 +566,8 @@ private Message createMessage(final String subject, final Map<String, Object> ap
     /**
      * Creates a request message for a payload and sends it to the peer.
      * <p>
+     * This method uses the default target address of this client as message address.
```
sophokles73: this method uses the *getMessageTargetAddress* method to determine the value of the message's *to* property ...

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -693,6 +783,7 @@ protected final void createAndSendRequest(
      * However, the span is never finished by this method.
      * 
      * @param action The operation that the request is supposed to trigger/invoke.
+     * @param requestTargetAddress The address to send the message to.
```
sophokles73: again, simply *address* FMPOV

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
calohmn: I've put the `default` in the name because this address is not *necessarily* used when sending a message with the client.
`AbstractRequestResponseClient` has 2 `createAndSendRequest` methods now; one with and one without a `requestTargetAddress` parameter. The one without uses the `getDefaultMessageTargetAddress()` address then.

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
calohmn: That would mean changing the `ApplicationClientFactory` interface, we currently have the `getOrCreateCommandClient(tenant, replyId)` method there (along with `getOrCreateCommandClient(tenant)` that uses the generated UUID).

client/src/main/java/org/eclipse/hono/client/ApplicationClientFactory.java
calohmn: Indeed, I'll change that.

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
calohmn: The motivation for the `replyId` parameter is in the javadoc of `getOrCreateCommandClient`:

> The protocol adapters need to convey an encoding of the reply-to address to the device when delivering the command. Consequently, the number of bytes transferred to the device depends on the length of the reply-to address being used. In situations where this is a major concern it might be advisable to use `getOrCreateCommandClient(String, String)` for creating a command client and provide custom (and shorter) <em>reply-to identifier</em> to be used in the reply-to address.

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
calohmn: I don't think so. I'll remove the condition.

calohmn:
I've pushed changes according to the comments.

calohmn:
@sophokles73 I've had a look at the javadoc of the `getOrCreateCommandClient` method with the `replyId` parameter:
https://github.com/eclipse/hono/blob/f247bbc2800b539c8b4cfe7a58b21b17c9af67e0/client/src/main/java/org/eclipse/hono/client/ApplicationClientFactory.java#L172-L177
To me it would look more consistent if we make sure the `replyId` is actually used in all cases by adding the user-chosen `replyId` (not the UUID created in the `getOrCreateCommandClient(tenantId)` method) to the cache key with which CommandClient instances are cached. WDYT?

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
sophokles73: Yeah, I forgot about that. So let's keep that.

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
sophokles73: I see, you introduced a variant for sending a message using a (message) specific address for the command client which sends messages with an address that contains the device ID whereas the sender link's target address doesn't. So the default address is usually the sender link's target address, right?

calohmn: yes

calohmn:
I've amended the commit to use `replyId` as part of the cache key, if it is explicitly set.

PR #1324, opened 2019-06-26
Update documentation with management HTTP API.

jbtrystram:
follow up work of #1240 .


sophokles73:
@jbtrystram a lot of the information contained in the markdown file seems to be a duplication of what we already have in the OpenApi spec. I wonder if we would have an easier time maintaining all relevant information in the OpenApi spec file only and try to publish e.g. a Swagger UI based on it instead of publishing an HTML page that is generated from the markdown (but which is basically a duplicate of the OpenApi spec).

@dejanb @ctron any thoughts?

jbtrystram:
It is indeed a duplicate of the API apec. If we're OK with reducing a bit the consistency, then publishing a Swagger UI for the spec is better IMHO, as it is far more readable and easy to do compared to what I've done here

ctron:
I think running an instance of swagger UI on hono.eclipse.org would be awesome!

sophokles73:
> I think running an instance of swagger UI on hono.eclipse.org would be awesome!

I am glad we found some common ground again, @ctron :-) I have no clue (yet) what it takes to run a Swagger UI but I guess there is nothing that a Docker container and nginx can't do ...

So, we seem to agree that we should try to publish the documentation of the HTTP management API by means of a Swagger UI that is generated from the OpenAPI spec, right?

jbtrystram:
I think a small markdown file giving an overview with a link to the swagger API would be awesome.

ctron:
Running a Swagger UI with docker should be as easy as (changing the `URL` or course):

~~~
docker run -ti --rm -p 8080:8080 -e URL=https://raw.githubusercontent.com/ctron/hono/feature/devreg_api_1/site/static/api/device-registry.yaml swaggerapi/swagger-ui
~~~

This way, we wouldn't need to update the container when we change the spec.

However I think we should do two things in addition: 1) point out that this is a convenience 2) let the versioned documentation point the versioned spec file

This way it would be easy to browse through the "most recent" spec, using swagger UI. But there would also be a way to browse through older documentations, and have a link to older versions of the spec (without the swagger UI).

sophokles73:
@ctron I was thinking about hosting the swagger UI together with the web site at eclipse.org/hono by using the JavaScript lib they provide. We would then have all in one place and could include it in our standard docs. WDYT?

ctron:
> @ctron I was thinking about hosting the swagger UI together with the web site at eclipse.org/hono by using the JavaScript lib they provide. We would then have all in one place and could include it in our standard docs. WDYT?

That would be even better. However I am not sure if we would need some kind of CQ for the swagger UI in that case? And I also had my troubles using that deployment model, that is why I was using the  container. But sure, having a simple "file on disk" model would be much easier than running a docker container :grin:

sophokles73:
Our colleagues from the IoT Hub team are hosting it this way on Amazon S3. I hope to get some inspiration from them ;-)
I'll talk to Sharon regarding the CQ ...

sophokles73:
I have been able to integrate the Swagger UI for the Management API into the Hono documentation's API section. Looks quite nice. However, I have asked about the CQ on the incubating mailing list. It is not clear yet, FMPOV, if we really need a CQ ...

sophokles73:
I have created a CQ for the current Swagger UI ...

calohmn:
@jbtrystram The [Registering Devices](https://www.eclipse.org/hono/getting-started/#registering-devices) chapter of the `Getting started` guide still is written towards the old API. Do you want to update it as part of this PR as well, or should we create a separate issue to track this?

sophokles73:
@calohmn there already is on: #1347

sophokles73:
@jbtrystram  Having agreed to using Swagger UI for adding the Management API, there doesn't seem much reason to keep this PR open.

I will open a PR based on my prototype using Swagger UI once we have done M6.

PR #1327, opened 2019-06-28
Reduce code duplication when handling response results

calohmn:


client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -978,21 +979,43 @@ protected final boolean isCachingEnabled() {
...
         } else {
             final R result = responseCache.get(key);
+            if (currentSpan != null) {
+                TracingHelper.TAG_CACHE_HIT.set(currentSpan, result != null);
+            }
             if (result == null) {
                 return Future.failedFuture("cache miss");
             } else {
+                if (currentSpan != null) {
+                    Tags.HTTP_STATUS.set(currentSpan, result.getStatus());
```
sophokles73: IMHO this shouldn't be done here because if there is a cache miss, the client of this method needs to retrieve the result from somewhere else and then also put the status to the span. It seems natural to me that putting the status to the span is the responsibility of the client invoking this method whereas marking the span for cache hit/miss is this method's responsibility.

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -1073,4 +1096,32 @@ protected final boolean isSuccessResponse(
...
                 && MessageHelper.CONTENT_TYPE_APPLICATION_JSON.equalsIgnoreCase(contentType);
     }
+
+    /**
+     * Applies the given mapper function to the result of the given Future if it succeeded.
+     * <p>
+     * Makes sure that the given Span is finished when the given Future is completed and error information is logged
+     * to the Span if there was an error.
+     *
+     * @param resultTracker The Future supplying the <em>RequestResponseResult</em> that the mapper will be applied on.
```
sophokles73: IMHO we should simply call this *result*

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
calohmn: > this shouldn't be done here because if there is a cache miss, the client of this method needs to retrieve the result from somewhere else and then also put the status to the span.

The "getting the result from somewhere else" usually means getting it via a `createAndSendRequest()` invocation. Inside `createAndSendRequest()`, the status tag is already set on the span.

That means the place in the code where `getResponseFromCache()` and as fall-back `createAndSendRequest()` is invoked doesn't set the status tag as of now. The status tag is set deep down where the result is set in the Future. And following that logic, I think we should set it in `getResponseFromCache()` as well.

An alternative would be setting it on the level where the span is finished, i.e in `mapResultAndFinishSpan()` for example. But then I would remove the places where the status is set in `createAndSendRequest()` (or deeper down).
Maybe this is better, since we also log an error to the span at the level of `mapResultAndFinishSpan()` - not at the place where it is set in the Future result.

sophokles73: Yeah, I think that would be a more consistent approach.

calohmn: I've added a commit implementing the aforementioned alternative, setting the status tag now where also a potential error is logged to the trace span.

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
calohmn: I've changed it accordingly.

Issue #1330, opened 2019-07-01
Report payload size of messages based on the configured minimum message size in HTTP, MQTT and AMQP Adapters


kaniyan:
Its also the case that the cloud providers set _minimum size_ for incoming messages. For example, lets say a _minimum message size_ is set to 4 KB. This can have an influence on how the number of messages or data volume are calculated.  A message of size 8KB is considered as two messages and another message with 10 KB is considered as three messages. Alternatively it can have influence on how data volume is calculated: a message of size 1KB is considered as 4KB and another with 10KB is considered as 12KB.

 It would be good that if Hono also has such an option and this can be combined with the _resource limit checks_ to effectively manage the resources. Ideally the _minimum message size_ can be set as part of the _tenant configuration_ so that it gives the flexibility to configure this size for each tenant. The adapters while reporting the metrics to Prometheus can calculate the _payload size_ based on the configured _minimum message size_ per tenant and then report the calculated payload size. This behaviour can be enabled only when _minimum message size_ is configured else not.

sophokles73:
@kaniyan can this be closed?

kaniyan:
> @kaniyan can this be closed?

This can be closed after merging the documentation and device registry updates in PR #1399

sophokles73:
@kaniyan merged

kaniyan:
Closed via PRs #1333 and  #1399

PR #1333, opened 2019-07-02
[#1330] Report metrics based on minimum message size

kaniyan:
This PR is implementation for #1330.

### DIFF
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
@@ -683,42 +683,38 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
                     addMicrometerSample(commandContext, timer);
                     Tags.COMPONENT.set(commandContext.getCurrentSpan(), getTypeName());
                     final Command command = commandContext.getCommand();
+                    final Future<TenantObject> tenantTracker = getTenantConfiguration(sourceAddress.getTenantId(),
+                            commandContext.getTracingContext());
 
-                    if (!command.isValid()) {
-                        final Exception ex = new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST, "malformed command message");
-                        commandContext.reject(getErrorCondition(ex), 1);
+                    tenantTracker.compose(tenantObject -> {
+                        if (!command.isValid()) {
+                            return Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST,
+                                    "malformed command message"));
+                        }
+                        if (!sender.isOpen()) {
+                            return Future.failedFuture(new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE,
+                                    "sender link is not open"));
+                        }
+                        return checkMessageLimit(tenantObject, command.getPayloadSize());
+                    }).compose(success -> {
+                        onCommandReceived(tenantTracker.result(), sender, commandContext);
+                        return Future.succeededFuture();
+                    }).otherwise(failure -> {
+                        if (failure instanceof ClientErrorException) {
+                            // issue credit so that application(s) can send the next command
+                            commandContext.reject(getErrorCondition(failure), 1);
+                        } else {
+                            commandContext.release(1);
+                        }
                         metrics.reportCommand(
                                 command.isOneWay() ? Direction.ONE_WAY : Direction.REQUEST,
                                 sourceAddress.getTenantId(),
-                                ProcessingOutcome.UNPROCESSABLE,
+                                tenantTracker.result(),
+                                ProcessingOutcome.from(failure),
                                 command.getPayloadSize(),
                                 timer);
-                    } else if (!sender.isOpen()) {
-                        commandContext.release(1);
-                        metrics.reportCommand(
-                                command.isOneWay() ? Direction.ONE_WAY : Direction.REQUEST,
-                                sourceAddress.getTenantId(),
-                                ProcessingOutcome.UNDELIVERABLE,
-                                command.getPayloadSize(),
-                                timer);
-                    } else {
-                        getTenantConfiguration(sourceAddress.getTenantId(), commandContext.getTracingContext())
-                                .compose(tenantObject -> checkMessageLimit(tenantObject, command.getPayloadSize()))
-                                .setHandler(result -> {
-                                    if (result.succeeded()) {
-                                        onCommandReceived(sender, commandContext);
-                                    } else {
-                                        // issue credit so that application(s) can send the next command
-                                        commandContext.reject(getErrorCondition(result.cause()), 1);
-                                        metrics.reportCommand(
-                                                command.isOneWay() ? Direction.ONE_WAY : Direction.REQUEST,
-                                                sourceAddress.getTenantId(),
-                                                ProcessingOutcome.from(result.cause()),
-                                                command.getPayloadSize(),
-                                                timer);
-                                    }
-                                });
-                    }
+                        return null;
+                    });
                 }, closeHandler -> {},
                 DEFAULT_COMMAND_CONSUMER_CHECK_INTERVAL_MILLIS);
     }
@@ -730,11 +726,12 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
      * This implementation simply forwards the command to the device
      * via the given link and flows a single credit back to the application.
      * 
+     * @param tenantObject The tenant configuration object.
      * @param sender The link for sending the command to the device.
      * @param commandContext The context in which the adapter receives the command message.
      * @throws NullPointerException if any of the parameters are {@code null}.
      */
-    protected void onCommandReceived(final ProtonSender sender, final CommandContext commandContext) {
+    protected void onCommandReceived(final TenantObject tenantObject, final ProtonSender sender, final CommandContext commandContext) {
 
         Objects.requireNonNull(sender);
         Objects.requireNonNull(commandContext);
@@ -777,6 +774,7 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
             metrics.reportCommand(
                     command.isOneWay() ? Direction.ONE_WAY : Direction.REQUEST,
                     command.getTenant(),
+                    tenantObject,
                     outcome,
                     command.getPayloadSize(),
                     getMicrometerSample(commandContext));
@@ -863,14 +861,15 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
 
         final Future<JsonObject> tokenFuture = getRegistrationAssertion(resource.getTenantId(), resource.getResourceId(),
                 context.getAuthenticatedDevice(), currentSpan.context());
-        final Future<TenantObject> tenantEnabledFuture = getTenantConfiguration(resource.getTenantId(),
-                currentSpan.context())
+        final Future<TenantObject> tenantTracker = getTenantConfiguration(resource.getTenantId(),
+                currentSpan.context());
+        final Future<TenantObject> tenantValidationTracker = tenantTracker
                         .compose(tenantObject -> CompositeFuture
                                 .all(isAdapterEnabled(tenantObject),
                                         checkMessageLimit(tenantObject, context.getPayloadSize()))
                                 .map(success -> tenantObject));
 
-        return CompositeFuture.all(tenantEnabledFuture, tokenFuture, senderFuture)
+        return CompositeFuture.all(tenantValidationTracker, tokenFuture, senderFuture)
                 .compose(ok -> {
 
                     final DownstreamSender sender = senderFuture.result();
@@ -878,7 +877,7 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
                             context.getMessage(),
                             ResourceIdentifier.from(context.getEndpoint().getCanonicalName(), resource.getTenantId(), resource.getResourceId()),
                             context.getAddress().toString(),
-                            tenantEnabledFuture.result(),
+                            tenantValidationTracker.result(),
                             tokenFuture.result(),
                             null); // no TTD
 
@@ -896,11 +895,11 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
                             context.getEndpoint().getCanonicalName(),
                             resource.getTenantId(),
                             resource.getResourceId(), t);
-                    final ProcessingOutcome outcome = t instanceof ClientErrorException ? ProcessingOutcome.UNPROCESSABLE : ProcessingOutcome.UNDELIVERABLE;
                     metrics.reportTelemetry(
                             context.getEndpoint(),
                             resource.getTenantId(),
-                            outcome,
+                            tenantTracker.result(),
+                            ProcessingOutcome.from(t),
                             context.isRemotelySettled() ? QoS.AT_MOST_ONCE : QoS.AT_LEAST_ONCE,
                             context.getPayloadSize(),
                             context.getTimer());
@@ -911,6 +910,7 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
                     metrics.reportTelemetry(
                             context.getEndpoint(),
                             resource.getTenantId(),
+                            tenantTracker.result(),
                             ProcessingOutcome.FORWARDED,
                             context.isRemotelySettled() ? QoS.AT_MOST_ONCE : QoS.AT_LEAST_ONCE,
                             context.getPayloadSize(),
@@ -924,64 +924,69 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
             final ResourceIdentifier resource,
             final Span currentSpan) {
 
-        final Future<CommandResponse> response = Optional.ofNullable(CommandResponse.from(context.getMessage()))
+        final Future<CommandResponse> responseTracker = Optional.ofNullable(CommandResponse.from(context.getMessage()))
                 .map(r -> Future.succeededFuture(r))
                 .orElseGet(() -> {
-                    TracingHelper.logError(currentSpan, String.format("invalid message (correlationId: %s, address: %s, status: %s)",
-                                    context.getMessage().getCorrelationId(), context.getMessage().getAddress(), MessageHelper.getStatus(context.getMessage())));
+                    TracingHelper.logError(currentSpan,
+                            String.format("invalid message (correlationId: %s, address: %s, status: %s)",
+                                    context.getMessage().getCorrelationId(), context.getMessage().getAddress(),
+                                    MessageHelper.getStatus(context.getMessage())));
                     return Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST,
                             "malformed command response message"));
                 });
+        final Future<TenantObject> tenantTracker = getTenantConfiguration(resource.getTenantId(),
+                currentSpan.context());
 
-        return response.compose(commandResponse -> {
+        return CompositeFuture.all(tenantTracker, responseTracker)
+                .compose(ok -> {
+                    final CommandResponse commandResponse = responseTracker.result();
+                    LOG.trace("sending command response [device-id: {}, status: {}, correlation-id: {}, reply-to: {}]",
+                            resource.getResourceId(), commandResponse.getStatus(), commandResponse.getCorrelationId(),
+                            commandResponse.getReplyToId());
 
-            LOG.trace("sending command response [device-id: {}, status: {}, correlation-id: {}, reply-to: {}]",
-                    resource.getResourceId(), commandResponse.getStatus(), commandResponse.getCorrelationId(),
-                    commandResponse.getReplyToId());
+                    final Map<String, Object> items = new HashMap<>(3);
+                    items.put(Fields.EVENT, "sending command response");
+                    items.put(TracingHelper.TAG_CORRELATION_ID.getKey(), commandResponse.getCorrelationId());
+                    items.put(MessageHelper.APP_PROPERTY_STATUS, commandResponse.getStatus());
+                    currentSpan.log(items);
 
-            final Map<String, Object> items = new HashMap<>(3);
-            items.put(Fields.EVENT, "sending command response");
-            items.put(TracingHelper.TAG_CORRELATION_ID.getKey(), commandResponse.getCorrelationId());
-            items.put(MessageHelper.APP_PROPERTY_STATUS, commandResponse.getStatus());
-            currentSpan.log(items);
+                    final Future<JsonObject> tokenFuture = getRegistrationAssertion(resource.getTenantId(),
+                            resource.getResourceId(), context.getAuthenticatedDevice(), currentSpan.context());
+                    final Future<TenantObject> tenantValidationTracker = CompositeFuture
+                            .all(isAdapterEnabled(tenantTracker.result()),
+                                    checkMessageLimit(tenantTracker.result(), context.getPayloadSize()))
+                            .map(success -> tenantTracker.result());
 
-            final Future<JsonObject> tokenFuture = getRegistrationAssertion(resource.getTenantId(),
-                    resource.getResourceId(), context.getAuthenticatedDevice(), currentSpan.context());
-            final Future<TenantObject> tenantEnabledFuture = getTenantConfiguration(resource.getTenantId(),
-                    currentSpan.context())
-                            .compose(tenantObject -> CompositeFuture
-                                    .all(isAdapterEnabled(tenantObject),
-                                            checkMessageLimit(tenantObject, context.getPayloadSize()))
-                                    .map(success -> tenantObject));
+                    return CompositeFuture.all(tenantValidationTracker, tokenFuture)
+                            .compose(success -> sendCommandResponse(resource.getTenantId(), commandResponse,
+                                    currentSpan.context()));
+                }).map(delivery -> {
 
-            return CompositeFuture.all(tenantEnabledFuture, tokenFuture)
-                    .compose(ok -> sendCommandResponse(resource.getTenantId(), commandResponse, currentSpan.context()));
-        }).map(delivery -> {
+                    LOG.trace("forwarded command response from device [tenant: {}, device-id: {}]",
+                            resource.getTenantId(), resource.getResourceId());
+                    metrics.reportCommand(
+                            Direction.RESPONSE,
+                            resource.getTenantId(),
+                            tenantTracker.result(),
+                            ProcessingOutcome.FORWARDED,
+                            context.getPayloadSize(),
+                            context.getTimer());
+                    return delivery;
 
-            LOG.trace("forwarded command response from device [tenant: {}, device-id: {}]",
-                    resource.getTenantId(), resource.getResourceId());
-            metrics.reportCommand(
-                    Direction.RESPONSE,
-                    resource.getTenantId(),
-                    ProcessingOutcome.FORWARDED,
-                    context.getPayloadSize(),
-                    context.getTimer());
-            return delivery;
+                }).recover(t -> {
 
-        }).recover(t -> {
+                    LOG.debug("cannot process command response from device [tenant: {}, device-id: {}]",
+                            resource.getTenantId(), resource.getResourceId(), t);
+                    metrics.reportCommand(
+                            Direction.RESPONSE,
+                            resource.getTenantId(),
+                            tenantTracker.result(),
+                            ProcessingOutcome.from(t),
+                            context.getPayloadSize(),
+                            context.getTimer());
 
-            LOG.debug("cannot process command response from device [tenant: {}, device-id: {}]",
-                    resource.getTenantId(), resource.getResourceId(), t);
-            final ProcessingOutcome outcome = t instanceof ClientErrorException ? ProcessingOutcome.UNPROCESSABLE : ProcessingOutcome.UNDELIVERABLE;
-            metrics.reportCommand(
-                    Direction.RESPONSE,
-                    resource.getTenantId(),
-                    outcome,
-                    context.getPayloadSize(),
-                    context.getTimer());
-
-            return Future.failedFuture(t);
-        });
+                    return Future.failedFuture(t);
+                });
     }
 
     /**
--- a/adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
+++ b/adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
@@ -243,7 +243,7 @@ public class VertxBasedAmqpProtocolAdapterTest {
         when(telemetrySender.send(any(Message.class), (SpanContext) any())).thenReturn(Future.succeededFuture(mock(ProtonDelivery.class)));
 
         // which is enabled for a tenant
-        givenAConfiguredTenant(TEST_TENANT_ID, true);
+        final TenantObject tenantObject = givenAConfiguredTenant(TEST_TENANT_ID, true);
 
         // IF a device sends a 'fire and forget' telemetry message
         final ProtonDelivery delivery = mock(ProtonDelivery.class);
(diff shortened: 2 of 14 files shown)


core/src/main/java/org/eclipse/hono/util/TenantConstants.java
```diff
@@ -25,6 +25,10 @@
      */
     public static final int DEFAULT_MAX_TTD = 60; // seconds
 
+    /**
+     * The default message size is set to -1, which implies no minimum size is defined.
+     */
+    public static final int DEFAULT_MINIMUM_MESSAGE_SIZE = -1;
```
sophokles73: wouldn't 0 (zero) do the same trick but be more *intuitive* in this case?

adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -856,6 +856,50 @@ public void testMessageLimitExceededForAnEventMessage(final TestContext ctx) {
...
+        final TenantObject tenantConfig = TenantObject.from(TEST_TENANT_ID, true)
+                .setMinimumMessageSize(4 * 1024);
+        when(tenantClient.get(eq(TEST_TENANT_ID), any())).thenReturn(Future.succeededFuture(tenantConfig));
+
+        // WHEN a gateway uploads a message
+        final ProtonDelivery delivery = mock(ProtonDelivery.class);
+        when(delivery.remotelySettled()).thenReturn(true);
+        final String to = ResourceIdentifier.from(TelemetryConstants.TELEMETRY_ENDPOINT, TEST_TENANT_ID, TEST_DEVICE)
+                .toString();
+        final Buffer payload = Buffer.buffer("some payload");
```
sophokles73: using something like
```
payload = Buffer.buffer("payload much smaller than 4kb");
```
would improve readability, don't you think?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -887,7 +887,7 @@ protected void onCommandReceived(final ProtonSender sender, final CommandContext
                             resource.getTenantId(),
                             outcome,
                             context.isRemotelySettled() ? QoS.AT_MOST_ONCE : QoS.AT_LEAST_ONCE,
-                            context.getPayloadSize(),
+                            calculatePayloadSize(context.getPayloadSize(), tenantTracker.result()),
```
sophokles73: This means that we *always* charge the client at least the real payload size, right?
If we experience a `ServerErrorException` then I think that we should report a payload size of 0 in order to not charge the client at all, because in this case it is not the client's fault that we cannot process the message, is it?

core/src/main/java/org/eclipse/hono/util/TenantConstants.java
kaniyan: Zero also works. I will update it.

adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
kaniyan: :+1:

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
kaniyan: very valid point. I will update it.

kaniyan: I guess not for all `ServerErrorException`. The one with cause as `no credit available` due to availability issues of the north bound applications may need to be charged. WDYT?

kaniyan:
@sophokles73  In this current approach, the _payload size_ is calculated inside the adapters and reported for every messages. May be a downside of this approach is, it doesn't ensure that all custom protocol adapters based on Hono provided APIs also calculate the same way and report the metrics. 

 As an alternate, how about adding  overloaded methods of `reportTelemetry(...)` and `reportCommand(...)` with an additional parameter of type `TenantObject` and deprecate the existing ones. And the calculation of _payload size_ is done within these methods.  WDYT?
```
void reportTelemetry(
            MetricsTags.EndpointType type,
            String tenantId,
            TenantObject tenantObject,
            MetricsTags.ProcessingOutcome outcome,
            MetricsTags.QoS qos,
            int payloadSize,
            MetricsTags.TtdStatus ttdStatus,
            Sample timer);
```
```

File: adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java (hunk 3 above)
Comment (sophokles73) [reply adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java]: In general I would agree. However, this would require quite some special error handling. Do you have an idea how we would do that efficiently?

**Comment by sophokles73** [comment]
> May be a downside of this approach is, it doesn't ensure that all custom protocol adapters based on Hono provided APIs also calculate the same way and report the metrics.

In fact, this argument is already true for the standard adapters where we need to make sure that the payload size is calculated consistently across all adapters as well. So I agree that it would be good to do the math in a central place instead of doing it in every adapter. However, I do not think that this solves the problem of the custom adapters as we have only very little control over the way they are implemented. In particular, we cannot enforce usage of our metrics client, can we? In fact, we wouldn't even want a custom adapter to report data to *our* metrics backend at all since a custom adapter is in the responsibility of a *customer* as the name implies. I think we will need to find another solution for accounting for the payload size of messages transmitted via custom adapters ...

**Comment by kaniyan** [comment]
> > May be a downside of this approach is, it doesn't ensure that all custom protocol adapters based on Hono provided APIs also calculate the same way and report the metrics.
> 
> In fact, this argument is already true for the standard adapters where we need to make sure that the payload size is calculated consistently across all adapters as well. So I agree that it would be good to do the math in a central place instead of doing it in every adapter. However, I do not think that this solves the problem of the custom adapters as we have only very little control over the way they are implemented. In particular, we cannot enforce usage of our metrics client, can we? In fact, we wouldn't even want a custom adapter to report data to _our_ metrics backend at all since a custom adapter is in the responsibility of a _customer_ as the name implies. I think we will need to find another solution for accounting for the payload size of messages transmitted via custom adapters ...

I agree that we need to find a solution for the custom adapters. For the sake of standard adapters, I will move the calculation part to a central place as discussed above.

File: adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java (hunk 3 above)
Comment (kaniyan) [reply adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java]: > This means that we _always_ charge the client at least the real payload size, right?
> If we experience a `ServerErrorException` then I think that we should report a payload size of 0 in order to not charge the client at all, because in this case it is not the client's fault that we cannot process the message, is it?

If we experience a `ServerErrorException`, then we normally record the processing outcome as `UNDELIVERABLE` in the metrics. When we aggregate data from the metrics back-end to verify the message limit, we look for the metrics data with processing outcomes as `FORWARDED` or `UNPROCESSABLE`. So I don't think we need to make the payload as 0 in case of `ServerErrorException` as processing outcome `UNDELIVERABLE` is anyway not part of the message limit calculation.


> In general I would agree. However, this would require quite some special error handling. Do you have an idea how we would do that efficiently?

With this current setup, it seems to me that the only way is to identify based on the error message `no credit available`and I don't find it nice to make use of it. As an alternative, similar to  the `ResourceConflictException` which extends `ClientErrorException`, a new Exception namely `NoCreditAvailableException` can be sub-classed from the `ServerErrorException`. This can be used by the adapters to record the processing outcome as `UNPROCESSABLE` in case of `no credit available`. 

But I am not sure that `no credit available` issue is always due to the availability issues of _northbound applications_ or _devices_. @calohmn Whats your view on this?

Comment (calohmn) [reply adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java]: I think a `no credit available` exception when trying to send a event/telemetry message downstream from a protocol adapter will *probably* mean the northbound application isn't connected. There may be the case however that the error stems from the device registry not giving out credits (maybe due to heavy load) or the AMQP network (e.g. Qpid) not giving out credits for a connected application. Such cases probably point to issues with the Hono setup and scaling of components.

Comment (sophokles73) [reply adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java]: Then why don't we create a `NoDownstreamCreditException` instead and only use that in the downstream senders when they have no credit? We could stick to the existing pattern used in the device registry clients where we create a `ServerErrorException` with code 503 instead.

Comment (kaniyan) [reply adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java]: > Then why don't we create a `NoDownstreamCreditException` instead and only use that in the downstream senders when they have no credit? We could stick to the existing pattern used in the device registry clients where we create a `ServerErrorException` with code 503 instead.

I would suggest a separate PR for implementing the `NoDownstreamCreditException` and recording the metrics with outcome as `UNPROCESSABLE` in case of this scenario.

I would continue with this PR by moving the business message payload size calculation into a central location (inside `reportTelemetry(...)` and `reportCommand(...)` methods inside `Metrics`).

**Comment by kaniyan** [comment]
> I would continue with this PR by moving the business message payload size calculation into a central location (inside `reportTelemetry(...)` and `reportCommand(...)` methods inside `Metrics`).

@sophokles73 I have moved the calculation part to a central location and also extended it for command and control too. Would you mind taking a look?

File: service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java [hunk 4]
```diff
@@ -82,6 +83,30 @@ void reportTelemetry(
...
+     * @param payloadSize The number of bytes contained in the message's payload.
+     * @param timer The timer indicating the amount of time used
+     *              for processing the message.
+     * @throws NullPointerException if any of the parameters are {@code null}.
+     * @throws IllegalArgumentException if type is neither telemetry nor event or
+     *                    if payload size is negative.
+     */
+    void reportTelemetry(
+            MetricsTags.EndpointType type,
+            String tenantId,
```
sophokles73: IMHO we do not need this if we also include the `TenantObject`, right?

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
```diff
@@ -106,11 +131,57 @@ void reportTelemetry(
...
+     * @param ttdStatus The outcome of processing the TTD value contained in the message.
+     * @param timer The timer indicating the amount of time used
+     *              for processing the message.
+     * @throws NullPointerException if any of the parameters are {@code null}.
+     * @throws IllegalArgumentException if type is neither telemetry nor event or
+     *                    if payload size is negative.
+     */
+    void reportTelemetry(
+            MetricsTags.EndpointType type,
+            String tenantId,
```
sophokles73: obsolete

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
```diff
@@ -121,6 +192,7 @@ void reportTelemetry(
     void reportCommand(
             MetricsTags.Direction direction,
             String tenantId,
+            TenantObject tenantObject,
```
sophokles73: tenantId is obsolete, isn't it?

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
```diff
@@ -106,11 +131,57 @@ void reportTelemetry(
...
+            MetricsTags.ProcessingOutcome outcome,
+            int payloadSize,
+            Sample timer);
+
     /**
      * Reports a command &amp; control message being transferred to/from a device.
      *
      * @param direction The command message's direction.
      * @param tenantId The tenant that the device belongs to.
+     * @param tenantObject The tenant configuration object.
```
sophokles73: the JavaDoc needs to explain when and why I would want to use this variant of the method, wouldn't you agree?

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
kaniyan: There is a scenario where there can be issue retrieving `TenantObject`using `getTenantConfiguration(...)`. In this case `TenantObject` will be null and tenant Id is still required to report the metrics.

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
kaniyan: yes, it needs to be there until all the adapters use the new method. Shall I mark them as deprecated or remove them ?

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
kaniyan: I will update it.

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
sophokles73: I think that all adapters need to be changed to use the new variant of this method ...

kaniyan: How about the old variant of this method? Should we mark them as deprecated or remove them after all adapters have been changed to use the new variant?

sophokles73: I think we should remove it

sophokles73:
@kaniyan are you working on this one?

kaniyan:
> @kaniyan are you working on this one?

yes, I am preparing first a PR to equip _AMQP_, _HTTP_ and _MQTT_ adapters. For  the rest of the adapters a separate one.

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
kaniyan: @sophokles73 The obsolete variant of the method `reportCommand(...) `  in `Metrics` interface has been removed as part of the latest commits pushed. The method `reportTelemetry(..) `is still being in use in COAP Adpater. It will be removed in the PR for the issue #1330 for COAP Adapter.

kaniyan:
@sophokles Would you mind taking a look at the latest commits :)

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandHandler.java
```diff
@@ -198,4 +203,39 @@ private void cancelTimer(final Long timerId) {
         vertx.cancelTimer(timerId);
         LOG.trace("Canceled Timer [timer-id: {}}", timerId);
     }
+
+    /**
+     * A class to facilitate storing timer id, tenantObject, command subscription and commandContext in a map.
+     * 
+     */
+    private class WaitingForAcknowledgementMapValue {
```
sophokles73: How about calling this *PendingCommandRequest* instead?
The JavaDoc is not very helpful as it only spells out what you can see from the class' structure. Please describe what the purpose of the class is instead ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandHandler.java
```diff
@@ -198,4 +203,39 @@ private void cancelTimer(final Long timerId) {
...
+        /**
+         * Creates a new WaitingForAcknowledgementMapValue instance.
+         * 
+         * @param timerId The unique ID of the timer.
+         * @param tenantObject The tenant configuration object.
+         * @param subscription The device's command subscription.
+         * @param commandContext The commandContext of the command sent.
+         * @throws NullPointerException if any of the parameters are {@code null}.
+         */
+        private WaitingForAcknowledgementMapValue(final Long timerId, final TenantObject tenantObject,
```
sophokles73: maybe also consider providing a static factory method? I find those easier to use than constructors. WDYT?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -539,6 +539,27 @@ public TenantObject setResourceLimits(final JsonObject resourceLimits) {
...
+                TenantConstants.DEFAULT_MINIMUM_MESSAGE_SIZE);
+    }
+
+    /**
+     * Sets the minimum message size in bytes.
+     * 
+     * @param payloadSize The payload size of the incoming message.
+     * @return The TenantObject.
+     */
+    public TenantObject setMinimumMessageSize(final int payloadSize) {
```
sophokles73: maybe reject negative values?

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
```diff
@@ -82,6 +83,36 @@ void reportTelemetry(
...
+     * <p> 
+     *     The payload size of the message is calculated based on the configured 
+     *     minimum message size and the calculated size is reported. 
+     * <p>
+     *     The configured minimum message size is retrieved from the tenant 
+     *     configuration object.
+     * 
+     * @param type The type of message received, e.g. <em>telemetry</em> or <em>event</em>.
+     * @param tenantId The tenant that the device belongs to.
+     * @param tenantObject The tenant configuration object.
```
sophokles73: this may  be `null` right?

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
```diff
@@ -82,6 +83,36 @@ void reportTelemetry(
...
+     * @param outcome The outcome of processing the message.
+     * @param qos The delivery semantics used for sending the message downstream.
+     * @param payloadSize The number of bytes contained in the message's payload.
+     * @param timer The timer indicating the amount of time used
+     *              for processing the message.
+     * @throws NullPointerException if any of the parameters are {@code null}.
```
sophokles73: except *tenantObject*?

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
```diff
@@ -106,11 +137,50 @@ void reportTelemetry(
...
+     * <p> 
+     *     The payload size of the message is calculated based on the configured 
+     *     minimum message size and the calculated size is reported. 
+     * <p>
+     *     The configured minimum message size is retrieved from the tenant 
+     *     configuration object.
+     *     
+     * @param type The type of message received, e.g. <em>telemetry</em> or <em>event</em>.
+     * @param tenantId The tenant that the device belongs to.
+     * @param tenantObject The tenant configuration object.
```
sophokles73: may be `null`?

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
```diff
@@ -106,11 +137,50 @@ void reportTelemetry(
...
+            MetricsTags.ProcessingOutcome outcome,
+            MetricsTags.QoS qos,
+            int payloadSize,
+            MetricsTags.TtdStatus ttdStatus,
+            Sample timer);
+
     /**
      * Reports a command &amp; control message being transferred to/from a device.
+     * <p> 
+     *     The payload size of the message is calculated based on the configured
```
sophokles73: no need to indent

service-base/src/test/java/org/eclipse/hono/service/metric/MicrometerBasedMetricsTest.java
```diff
@@ -157,4 +159,27 @@ public void testReportTelemetryInvokesLegacyMetrics(final MeterRegistry registry
...
     }
+
+    /**
+     * Verifies that the calculated payload size for messages of different sizes are correct.
+     * 
+     * @param registry : the registry that the tests should be run against.
+     */
+    @ParameterizedTest
+    @MethodSource("registries")
+    public void testCalculatePayloadSize(final MeterRegistry registry) {
```
sophokles73: I'd rather have a test which verifies that the correct values are set on the MeterRegistry when reporting messages via  Metrics' interface methods ..

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -817,32 +818,36 @@ void handlePublishedMessage(final MqttContext context) {
         }
 
         final MetricsTags.QoS qos = MetricsTags.QoS.from(ctx.message().qosLevel().value());
-        return uploadMessage(
+        final Future<TenantObject> tenantTracker = getTenantConfiguration(tenant, ctx.getTracingContext());
+
+        return tenantTracker.compose(tenantObject -> uploadMessage(
                 Objects.requireNonNull(ctx),
-                Objects.requireNonNull(tenant),
+                Objects.requireNonNull(tenantObject),
```
sophokles73: this has different semantics than the check for *tenant* being `null`

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -872,32 +877,36 @@ void handlePublishedMessage(final MqttContext context) {
...
```
sophokles73: see above

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: `Objects.requireNonNull(tenant)` is already being checked while retrieving the tenant object using `getTenantConfiguration(...)` and therefore its not needed again. Just I released that there is also no need for `Objects.requireNonNull(tenantObject)` inside the _compose_ block. The compose _block_ is anyway executed, if the tenant object is successfully retrieved using `getTenantConfiguration(...)`.

service-base/src/test/java/org/eclipse/hono/service/metric/MicrometerBasedMetricsTest.java
kaniyan: @sophokles73 I have created tests accordingly and pushed the changes.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandHandler.java
kaniyan: I have created a static factory method for the above.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandHandler.java
kaniyan: Done

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
kaniyan: Javadoc has been updated.

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
kaniyan: Javadoc has been updated.

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
kaniyan: Javadoc has been updated.

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
kaniyan: Javadoc has been updated.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -683,42 +683,43 @@ private Span newSpan(final String operationName, final Device authenticatedDevic
...
+                            return Future.succeededFuture();
+                        } else {
+                            return checkMessageLimit(tenantObject, command.getPayloadSize())
+                                    .compose(ok -> {
+                                        onCommandReceived(tenantObject, sender, commandContext);
+                                        return Future.succeededFuture();
+                                    });
+                        }
+                    }).setHandler(result -> {
+                        if (result.failed()) {
```
sophokles73: it looks a little strange that there is only a *failed* block. What about moving the *onCommandReceived* invocation down here into a *succeeded* block and just leave the *checkMessageLimit* invocation in the block above?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -662,34 +663,34 @@ protected final void onUnsubscribe(
...
-                                                timer);
-                                    }
-                                });
-                    } else {
+                                        onCommandReceived(tenantTracker.result(), mqttEndpoint, sub, commandContext,
+                                                cmdHandler);
+                                        return Future.succeededFuture();
+                                    });
+                        }
+                    }).otherwise(failure -> {
```
sophokles73: this could be modified to resemble the code used in the AMQP adapter a little more, i.e. using *setHandler* here with both a *succeeded* and a *failed* branch. WDYT?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -817,32 +818,36 @@ void handlePublishedMessage(final MqttContext context) {
...
-            metrics.reportTelemetry(
-                    ctx.endpoint(),
-                    ctx.tenant(),
-                    ProcessingOutcome.from(t),
-                    qos,
-                    payload.length(),
-                    ctx.getTimer());
-            return Future.failedFuture(t);
-        });
+                ctx.endpoint()))
```
sophokles73: does it make sense to postpone the null checks to *after* the Future for getting the tenant configuration has completed? IMHO the null checks should be moved to the beginning of the method ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandHandler.java
```diff
@@ -198,4 +203,50 @@ private void cancelTimer(final Long timerId) {
         vertx.cancelTimer(timerId);
         LOG.trace("Canceled Timer [timer-id: {}}", timerId);
     }
+
+    /**
+     * A class to facilitate storing of information in connection with the pending command requests.
+     * The pending command requests are tracked using a map in the enclosing class {@link CommandHandler}
+     * and is used to handle PUBACKs from devices.
```
sophokles73: :+1:

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
kaniyan: The `else if ` block comprising `!sender.isOpen()` returns _succeeded Future_. Therefore _onCommandReceived_  cannot be directly moved to the _succeeded_ block. Personally I would prefer a `otherwise` block instead of `setHandler` here. IMHO it also reduces no. of lines of code as there is no need for _if_ and _else_ blocks for _success_ or _failure_. Personally I find the `otherwise` block as intuitive as it deals only with _failures_. I guess its also a matter of taste. If you insist on using `setHandler`, I can refactor accordingly the whole part so that the _onCommandReceived_ can be moved to the _succeeded_ block.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: I have moved it to the beginning.

sophokles73: have you?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
sophokles73: then I guess you should replace the *setHandler* with *otherwise*, right? :-)

kaniyan: yes. :-)

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: I will push the changes along with the one for the above comment.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: As above, I would keep the `otherwise` block.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: @sophokles73 I have pushed the changes. Pls have a look.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -683,42 +683,42 @@ private Span newSpan(final String operationName, final Device authenticatedDevic
...
+                        } else {
+                            return checkMessageLimit(tenantObject, command.getPayloadSize())
+                                    .compose(ok -> {
+                                        onCommandReceived(tenantObject, sender, commandContext);
+                                        return Future.succeededFuture();
+                                    });
+                        }
+                    }).otherwise(failure -> {
+                        // issue credit so that application(s) can send the next command
+                        commandContext.reject(getErrorCondition(failure), 1);
```
sophokles73: IMHO this is not yet correct. We reject here based on the assumption that we are dealing with a client induced error, e.g. a malformed command message. However, this code will also be executed if we cannot retrieve the tenant object due to lack of credit (similar to `!sender.isOpen()`). In that case, however, I believe that we would also need to release the command message ...

kaniyan: @sophokles73 The Command message is now being _rejected_ if the error is `ClientErrorException` else _released_. I have pushed the changes for _AMQP_ and _MQTT_ adapters.

sophokles73:
@kaniyan can you please create issues for adding this functionality to the other adapters so that we do not forget?

kaniyan:
> @kaniyan can you please create issues for adding this functionality to the other adapters so that we do not forget?

@sophokles73 There is already one for the COAP adapter #1381. I have just created an issue for LoRA adapter  too #1397.

PR #1335, opened 2019-07-03
[#1210] Improve reconnect interval handling

### LINKED ISSUE #1210: Improve reconnect interval handling
The handling of the *reconnect interval* concerning AMQP connections between Hono components should be improved.

Currently, when for example the connection between a protocol adapter and the device registry fails or is closed, the re-connect attempt happens only after 500ms (that is the DEFAULT_RECONNECT_INTERVAL_MILLIS = 500).

I guess this was introduced for the scenario that the device registry is indeed unavailable and for example currently being (re)started, so that an immediately-after re-connect attempt would probably fail.

But, there are other scenarios:
- The device registry could be deployed with multiple instances (Kubernetes pods) and even if one instance is currently not responding, the next connection attempt could reach another, fully available instance.
- The device registry could have actively closed the connection with a "access token expired" response, meaning there is no reason to wait for a re-connect attempt.

IMHO, these two scenarios are quite common, therefore we should introduce a way to handle reconnects where there isn't such a long delay before doing the first retry.

Suggested in the mailinglist discussion on this was to use an increasing timeout, starting with zero, up to a maximum:
````
minDelay: 0
maxDelay: 1000
increment: 100
````
On connect, reset relay to "min". On connection failure increment the delay if it is below "max".

### PR DESCRIPTION
**Pull request by calohmn**
This fixes #1210:

There is now a variable delay period between reconnect attempts, configured via 'ReconnectMinDelay', 'ReconnectMaxDelay' and 'ReconnectDelayIncrement' properties.

### CODE REVIEW

File: client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java [hunk 1]
```diff
@@ -492,14 +494,17 @@ private void reconnect(
...
             }
-            final long reconnectInterval = Optional.ofNullable(clientOptions)
-                    .map(o -> o.getReconnectInterval())
-                    .orElse(Constants.DEFAULT_RECONNECT_INTERVAL_MILLIS);
-            log.trace("scheduling new connection attempt in {}ms ...", reconnectInterval);
-            // give Vert.x some time to clean up NetClient
-            vertx.setTimer(reconnectInterval, tid -> {
+            final long reconnectInterval = Math.min(clientConfigProperties.getReconnectMaxDelay(),
+                    clientConfigProperties.getReconnectMinDelay()
+                            + reconnectAttempt * clientConfigProperties.getReconnectDelayIncrement());
```
sophokles73: it seems that you do not want to use the usual *exponential back-off* algorithm here. Is there a particular reason for that? I am not sure if incrementing the delay in steps of 100ms will actually help the peer to relief from its problems ...

calohmn: Right, I've changed the implementation, now applying an exponential backoff with jitter, the delay being a random between the minDelay and the value exponentially increasing with every attempt. The delay is capped by the maxDelay.

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -492,14 +495,21 @@ private void reconnect(
...
-            final long reconnectInterval = Optional.ofNullable(clientOptions)
-                    .map(o -> o.getReconnectInterval())
-                    .orElse(Constants.DEFAULT_RECONNECT_INTERVAL_MILLIS);
-            log.trace("scheduling new connection attempt in {}ms ...", reconnectInterval);
-            // give Vert.x some time to clean up NetClient
-            vertx.setTimer(reconnectInterval, tid -> {
+            // apply exponential backoff with jitter
+            final long currentMaxDelay = (long) Math.pow(2, reconnectAttempt - 1)
+                    * clientConfigProperties.getReconnectDelayIncrement();
+            final long reconnectInterval = currentMaxDelay > clientConfigProperties.getReconnectMinDelay()
```
sophokles73: this is very hard to read and understand, can we *untangle* this and use multiple expressions instead of a single one?

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -49,6 +49,22 @@
...
+    /**
+     * The default minimum amount of time (milliseconds) to wait before trying to re-establish an
+     * AMQP connection with the peer.
+     */
+    public static final int DEFAULT_RECONNECT_MIN_DELAY = 0; // ms
+    /**
+     * The default maximum amount of time (milliseconds) to wait before trying to re-establish an
+     * AMQP connection with the peer.
+     */
+    public static final int DEFAULT_RECONNECT_MAX_DELAY = 1000; // ms
```
sophokles73: I think that this is a little low and may not allow stressed components to catch up again. What about something in the order of 30 seconds instead?

calohmn: 30s sounds a bit much too me. I would rather use something like 7s.
That still wouldn't put much pressure on a "stressed component" but would allow for potentially faster recovery if the target component is just being restarted.

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
calohmn: Ok, I've split it up.

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -492,14 +495,27 @@ private void reconnect(
...
-            // give Vert.x some time to clean up NetClient
-            vertx.setTimer(reconnectInterval, tid -> {
+            // apply exponential backoff with jitter
+            // determine the max delay for this reconnect attempt as 2^attempt * delayIncrement
+            final long currentMaxDelay = (long) Math.pow(2, reconnectAttempt - 1)
+                    * clientConfigProperties.getReconnectDelayIncrement();
+            final long reconnectInterval;
+            if (currentMaxDelay > clientConfigProperties.getReconnectMinDelay()) {
+                // let the actual reconnect delay be a random between the minDelay and the currentMaxDelay,
+                // capped by the overall maxDelay
```
sophokles73: that is quite a spread, isn't it? it means that we wait for anything between 0 and 7000ms, right? The jitter is usually within 
95 and 105% (?) of the target value, just enough to make it not strictly deterministic ...

calohmn: It is, I was opting for trying to evenly spread out the calls to the stressed/restarting component (e.g. device registry) coming from different components (e.g. protocol adapters), taking into account that we rather have a shorter delay than a longer one. In that sense it's a kind of "full jitter" approach (see also [here](https://aws.amazon.com/de/blogs/architecture/exponential-backoff-and-jitter/)).

PR #1341, opened 2019-07-09
[#1339] Improve handling of command correlation ID.

### LINKED ISSUE #1339: Mismatch expectation towards correlation ID
Looking at the documentation, it looks like the "correlation id" **optional**:

* https://www.eclipse.org/hono/docs/latest/api/command-and-control-api/#send-a-one-way-command
* https://www.eclipse.org/hono/docs/latest/api/command-and-control-api/#send-a-request-response-command

However, looking at the actual code, this looks different:

https://github.com/eclipse/hono/blob/1574f9b3297d47a94ec465c578d470ffc5d76b97/client/src/test/java/org/eclipse/hono/client/CommandTest.java#L206-L214

https://github.com/eclipse/hono/blob/1574f9b3297d47a94ec465c578d470ffc5d76b97/client/src/main/java/org/eclipse/hono/client/Command.java#L129-L131

calohmn:
This fixes #1339.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/Command.java
+++ b/client/src/main/java/org/eclipse/hono/client/Command.java
@@ -95,7 +95,7 @@ public final class Command {
      * @param tenantId The tenant that the device belongs to.
      * @param deviceId The identifier of the device.
      * @return The command.
-     * @throws NullPointerException if any of the parameters are {@code null}.
+     * @throws NullPointerException if any of the parameters is {@code null}.
      */
     public static Command from(
             final Message message,
@@ -111,23 +111,17 @@ public final class Command {
             validationErrorJoiner.add("subject not set");
         }
 
-        final String correlationId = Optional.ofNullable(message.getCorrelationId()).map(obj -> {
-            if (obj instanceof String) {
-                return (String) obj;
+        String correlationId = null;
+        final Object correlationIdObj = MessageHelper.getCorrelationId(message);
+        if (correlationIdObj != null) {
+            if (correlationIdObj instanceof String) {
+                correlationId = (String) correlationIdObj;
             } else {
-                return null;
+                validationErrorJoiner.add("message/correlation-id is not of type string, actual type: " + correlationIdObj.getClass().getName());
             }
-        }).orElseGet(() -> {
-            final Object obj = message.getMessageId();
-            if (obj instanceof String) {
-                return (String) obj;
-            } else {
-                return null;
-            }
-        });
-
-        if (correlationId == null) {
-            validationErrorJoiner.add("message/correlation-id not set");
+        } else if (message.getReplyTo() != null) {
+            // correlation id is required if a command response is expected
+            validationErrorJoiner.add("neither message-id nor correlation-id is set");
         }
 
         String originalReplyToId = null;
@@ -255,7 +249,7 @@ public final class Command {
     /**
      * Gets the request identifier of this command.
      *
-     * @return The identifier.
+     * @return The identifier or {@code null} if not set.
      * @throws IllegalStateException if this command is invalid.
      */
     public String getRequestId() {
@@ -343,7 +337,7 @@ public final class Command {
     /**
      * Gets the ID to use for correlating a response to this command.
      *
-     * @return The identifier.
+     * @return The identifier or {@code null} if not set.
      * @throws IllegalStateException if this command is invalid.
      */
     public String getCorrelationId() {
--- a/client/src/test/java/org/eclipse/hono/client/CommandTest.java
+++ b/client/src/test/java/org/eclipse/hono/client/CommandTest.java
@@ -157,6 +157,41 @@ public class CommandTest {
     }
 
     /**
+     * Verifies that a command can be created from a valid message that has an empty reply-to property
+     * and no message-id and correlation-id properties.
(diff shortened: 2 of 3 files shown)


ctron:
Is there a reason why we limit this to "string" type? Why not simply accept all types?

calohmn:
@ctron Which location in the code or docs are you referring to?

For the command request message, all AMQP *message-id* types are supported now for the `message-id` and `correlation-id` properties. 
For the command response message, sent by the protocol adapter, the `correlation-id` is of type `string` (otherwise we would have to encode the type in the reply id somehow).

(In the docs, the *message-id* property is still marked as of type `string` - in all the Hono APIs. I would change this in a subsequent commit, having a look at the implications in the other places.)

ctron:
I am talking about the method `MessageHelper.getMessageIdStringValue(correlationIdObj)`.

It only accepts `String`, `UUID`, `UnsignedLong`, and `Binary`. Why not `Decimal128`, or `UnsignedInteger` or simply `Number`?

And then again, why not let the protocol adapter handle that conversion. I agree, on the HTTP side you need to convert that to a string, but for the AMQP protocol adapter, that may not be necessary.

But I also think it is ok to align that to a string, and use the `getMessageIdStringValue` method, but then I would expect that most of the other scalar values can be used as well.

calohmn:
> It only accepts String, UUID, UnsignedLong, and Binary. Why not Decimal128, or UnsignedInteger or simply Number?

Looking at the AMQP spec, it looks to me that only these 4 types are supported, as these are the only ones that provide the `message-id` archetype (see sections 3.2.11-14). This is also reflected in [EventBusMessage#encodeIdToJson](https://github.com/eclipse/hono/blob/aa5232e6793d603a24ed3a7b2e8072882c34b316/core/src/main/java/org/eclipse/hono/util/EventBusMessage.java#L567) for example.

> And then again, why not let the protocol adapter handle that conversion. I agree, on the HTTP side you need to convert that to a string, but for the AMQP protocol adapter, that may not be necessary.

The conversion of the correlation id property value to its string representation is done in the `Command` class already because the string value is needed for the `Command.requestId`, which will then be used by the protocol adapters in the protocol specific command message sent to the device.

ctron:
> Looking at the AMQP spec, it looks to me that only these 4 types are supported, as these are the only ones that provide the message-id archetype (see sections 3.2.11-14). This is also reflected in EventBusMessage#encodeIdToJson for example.

You are right. I wasn't aware of that.

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -835,4 +838,24 @@ public static Message getShallowCopy(final Message message) {
...
+     */
+    public static String getMessageIdStringValue(final Object messageIdValue) {
+        Objects.requireNonNull(messageIdValue);
+
+        if (messageIdValue instanceof String || messageIdValue instanceof UnsignedLong
+                || messageIdValue instanceof UUID) {
+            return messageIdValue.toString();
+        } else if (messageIdValue instanceof Binary) {
+            return Base64.getEncoder().encodeToString(((Binary) messageIdValue).getArray());
+        } else {
```
sophokles73: I do not think that this will work because it effectively changes the type of the correlation ID to `String` before sending it to the device. The device will then include this (String typed) correlation ID in its response sent to the adapter and the adapter will put it into the response AMQP message as a String typed correlation ID.
Thus, when an application sends a request/response command and uses e.g. a UUID as the correlation ID, I expect it to fail in comparing the response's correlation ID with the expected value ...

sophokles73:
> Is there a reason why we limit this to "string" type? Why not simply accept all types?

We chose to limit it because otherwise we would need to encode the original type of correlation ID into the reply ID we forward to the device. FMPOV it is not worth the effort doing so. I do not expect any client having a problem using a String typed message/correlation-ID. Do you?

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
calohmn: The idea here was that we explicitly document that the command response correlation ID will be a String:

> Response Message Format
> correlation-id | yes | properties | string
> MUST contain the correlation ID used to match the command message with the response message containing the result of execution on the device. The ID here is the string value of the correlation ID set on the command message (a binary value is base64-encoded here).

If an application chooses to use a non-String correlation ID for a command message, it will have to match the command response according to the string representation.

I see that putting such a requirement on the application is very much similar to requiring it to use a command message with a string correlation ID in the first place. 

Still I would rather prefer the above all-types-allowed solution, not rejecting command messages with a non-string correlation id. How correlation ids are to be matched is not described in the AMQP spec anyway.

sophokles73: I have to disagree. FMPOV the restriction that we currently have in place, leaves no doubt as to what a client needs to put in there and how IDs are to be matched. Receiving back the String representation of the ID that a client put into the request in the response seems very strange to me.
FMPOV we either properly support all types, i.e. we also encode the message-id type into the reply-id or we stick to just supporting string typed message-ids.

calohmn: Well, I don't have a strong preference here, requiring a string correlation id in the command message is fine with me as well. I've amended the commit accordingly.

PR #1342, opened 2019-07-09
[#1238] Split up hono.app by introducing health check config properties

mbaeuerle:
First part to implement #1238. 
This splits up hono.app and moves all health check related properties into a new config properties file.

In a second part the properties will be extended and used to make it possible to secure the health check with SSL.


core/src/main/java/org/eclipse/hono/config/HealthCheckConfigProperties.java
```diff
@@ -0,0 +1,77 @@
...
+import org.eclipse.hono.util.Constants;
+import org.eclipse.hono.util.PortConfigurationHelper;
+
+import java.util.Objects;
+
+/**
+ * Configuration of common properties that are valid for a health check server.
+ *
+ */
+public class HealthCheckConfigProperties extends AbstractConfig {
```
sophokles73: In order to not introduce too much duplication of code, I would suggest to rename this class to `ServerConfig` and then let other config classes like `org.eclipse.hono.util.config.ServiceConfigProperties` extend it. We can also pull up that class' *nativeTlsRequired*, *insecurePortEnabled*, *insecurePortBindAddress* and *insecurePort* properties into this class. WDYT?

mbaeuerle: Sounds reasonable to me.
So do we also want to use `ServerConfig` as super class of `ClilentConfigProperties`? In this case we should think about replacing `ClientConfigProperties.host` with `ServerConfig.bindAddress`. 
However I see 29 usages of `getHost()` so this would cause quite some changes. But then we also should think about the naming, as `ServerConfig` would not be proper any longer.

services/auth/src/main/java/org/eclipse/hono/service/auth/impl/ApplicationConfig.java
```diff
@@ -93,16 +94,27 @@ public ObjectFactoryCreatingFactoryBean authServerFactory() {
...
     }
 
+    /**
+     * Exposes properties for configuring the health check as a Spring bean.
+     *
+     * @return The health check configuration properties.
+     */
+    @Bean
+    @ConfigurationProperties(prefix = "hono.health-check")
+    public ServerConfig healthCheckConfigProperties() {
```
mbaeuerle: @sophokles73 I am unsure if we want to name it that way even though we create a `ServerConfig` object. WDYT?

core/src/main/java/org/eclipse/hono/config/HealthCheckConfigProperties.java
sophokles73: > So do we also want to use ServerConfig as super class of ClilentConfigProperties?

Why would we want to do that? The `ServerConfig` properties are not relevant for clients, are they?

services/auth/src/main/java/org/eclipse/hono/service/auth/impl/ApplicationConfig.java
sophokles73: That's ok FMPOV

site/documentation/content/admin-guide/amqp-adapter-config.md
```diff
@@ -17,8 +17,8 @@ The following table provides an overview of the configuration variables and corr
...
-| `HONO_APP_HEALTH_CHECK_BIND_ADDRESS`<br>`--hono.app.healthCheckBindAddress` | no | `127.0.0.1` | The IP address of the network interface that the HTTP server, which exposes the service's health check resources, should be bound to. The HTTP server will only be started if `HONO_APP_HEALTH_CHECK_BIND_ADDRESS` is set explicitly. |
+| `HONO_HEALTH_CHECK_PORT`<br>`--hono.healthCheck.port` | no | - | The port that the HTTP server, which exposes the service's health check resources, should bind to. If set, the adapter will expose a *readiness* probe at URI `/readiness` and a *liveness* probe at URI `/liveness`. |
+| `HONO_HEALTH_CHECK_BIND_ADDRESS`<br>`--hono.healthCheck.bindAddress` | no | `127.0.0.1` | The IP address of the network interface that the HTTP server, which exposes the service's health check resources, should be bound to. The HTTP server will only be started if `HONO_HEALTH_CHECK_BIND_ADDRESS` is set explicitly. |
```
sophokles73: can you move these variables to the correct place based on lexical order?

site/documentation/content/admin-guide/device-registry-config.md
```diff
@@ -26,8 +26,8 @@ The following table provides an overview of the configuration variables and corr
...
-| `HONO_APP_HEALTH_CHECK_BIND_ADDRESS`<br>`--hono.app.healthCheckBindAddress` | no | `127.0.0.1` | The IP address of the network interface that the HTTP server, which exposes the service's health check resources, should be bound to. The HTTP server will only be started if `HONO_APP_HEALTH_CHECK_BIND_ADDRESS` is set explicitly. |
+| `HONO_HEALTH_CHECK_PORT`<br>`--hono.healthCheck.port` | no | - | The port that the HTTP server, which exposes the service's health check resources, should bind to. If set, the adapter will expose a *readiness* probe at URI `/readiness` and a *liveness* probe at URI `/liveness`. |
+| `HONO_HEALTH_CHECK_BIND_ADDRESS`<br>`--hono.healthCheck.bindAddress` | no | `127.0.0.1` | The IP address of the network interface that the HTTP server, which exposes the service's health check resources, should be bound to. The HTTP server will only be started if `HONO_HEALTH_CHECK_BIND_ADDRESS` is set explicitly. |
```
sophokles73: can you move these variables to the correct place based on lexical order?

site/documentation/content/admin-guide/http-adapter-config.md
```diff
@@ -17,8 +17,8 @@ The following table provides an overview of the configuration variables and corr
...
-| `HONO_APP_HEALTH_CHECK_BIND_ADDRESS`<br>`--hono.app.healthCheckBindAddress` | no | `127.0.0.1` | The IP address of the network interface that the HTTP server, which exposes the service's health check resources, should be bound to. The HTTP server will only be started if `HONO_APP_HEALTH_CHECK_BIND_ADDRESS` is set explicitly. |
+| `HONO_HEALTH_CHECK_PORT`<br>`--hono.healthCheck.port` | no | - | The port that the HTTP server, which exposes the service's health check resources, should bind to. If set, the adapter will expose a *readiness* probe at URI `/readiness` and a *liveness* probe at URI `/liveness`. |
+| `HONO_HEALTH_CHECK_BIND_ADDRESS`<br>`--hono.healthCheck.bindAddress` | no | `127.0.0.1` | The IP address of the network interface that the HTTP server, which exposes the service's health check resources, should be bound to. The HTTP server will only be started if `HONO_HEALTH_CHECK_BIND_ADDRESS` is set explicitly. |
```
sophokles73: can you move these variables to the correct place based on lexical order?

site/documentation/content/admin-guide/kura-adapter-config.md
```diff
@@ -17,8 +17,8 @@ The following table provides an overview of the configuration variables and corr
...
-| `HONO_APP_HEALTH_CHECK_BIND_ADDRESS`<br>`--hono.app.healthCheckBindAddress` | no | `127.0.0.1` | The IP address of the network interface that the HTTP server, which exposes the service's health check resources, should be bound to. The HTTP server will only be started if `HONO_APP_HEALTH_CHECK_BIND_ADDRESS` is set explicitly. |
+| `HONO_HEALTH_CHECK_PORT`<br>`--hono.healthCheck.port` | no | - | The port that the HTTP server, which exposes the service's health check resources, should bind to. If set, the adapter will expose a *readiness* probe at URI `/readiness` and a *liveness* probe at URI `/liveness`. |
+| `HONO_HEALTH_CHECK_BIND_ADDRESS`<br>`--hono.healthCheck.bindAddress` | no | `127.0.0.1` | The IP address of the network interface that the HTTP server, which exposes the service's health check resources, should be bound to. The HTTP server will only be started if `HONO_HEALTH_CHECK_BIND_ADDRESS` is set explicitly. |
```
sophokles73: can you move these variables to the correct place based on lexical order?

site/documentation/content/admin-guide/mqtt-adapter-config.md
```diff
@@ -17,8 +17,8 @@ The following table provides an overview of the configuration variables and corr
...
-| `HONO_APP_HEALTH_CHECK_BIND_ADDRESS`<br>`--hono.app.healthCheckBindAddress` | no | `127.0.0.1` | The IP address of the network interface that the HTTP server, which exposes the service's health check resources, should be bound to. The HTTP server will only be started if `HONO_APP_HEALTH_CHECK_BIND_ADDRESS` is set explicitly. |
+| `HONO_HEALTH_CHECK_PORT`<br>`--hono.healthCheck.port` | no | - | The port that the HTTP server, which exposes the service's health check resources, should bind to. If set, the adapter will expose a *readiness* probe at URI `/readiness` and a *liveness* probe at URI `/liveness`. |
+| `HONO_HEALTH_CHECK_BIND_ADDRESS`<br>`--hono.healthCheck.bindAddress` | no | `127.0.0.1` | The IP address of the network interface that the HTTP server, which exposes the service's health check resources, should be bound to. The HTTP server will only be started if `HONO_HEALTH_CHECK_BIND_ADDRESS` is set explicitly. |
```
sophokles73: can you move these variables to the correct place based on lexical order?

core/src/main/java/org/eclipse/hono/config/HealthCheckConfigProperties.java
mbaeuerle: At least they share `port` and `bindingAddress` (as `host`). But I guess the `insecure*` properties don't make sense in the client properties.

sophokles73: As you said, these have different semantics. So we should not have `ClientConfigProperties` extend `ServerConfig`.

sophokles73:
@mbaeuerle can you also rebase the PR to current master?

core/src/main/java/org/eclipse/hono/config/ServerConfig.java
```diff
@@ -0,0 +1,209 @@
...
+ *******************************************************************************/
+
+package org.eclipse.hono.config;
+
+import java.util.Objects;
+
+import org.eclipse.hono.util.Constants;
+
+/**
+ * Configuration of common properties that are valid for a health check server.
```
sophokles73: > ... that are common for components accepting requests/connections on a network socket.

mbaeuerle: Thanks, I missed that when I changed the name.

mbaeuerle:
An integration test failed after I changed the docs, is there a way to rerun the test?

sophokles73:
> An integration test failed after I changed the docs, is there a way to rerun the test?

I just did ...

sophokles73:
@mbaeuerle  can you please also add an entry to the release notes indicating this change of config property names?

mbaeuerle:
@sophokles73 is this the right amount of detail in the [release notes](https://github.com/eclipse/hono/commit/70fd986840222dce722a156e737aa0ee07124a17#diff-65d083e37344d5560964edf6e0a55ac8R22)?

Issue #1284, opened 2019-06-11
Extend message limit check to include commands


kaniyan:
With the implementation of  #929, Hono now supports _message limit checks_ for _telemetry_ and _event_ messages. This needs to be extended to cover the _commands_ being sent from north bound applications. Also the protocol adapters should be enabled to verify the _message limit_ before accepting _commands_ from north bound applications.

sophokles73:
@kaniyan can this be closed?

kaniyan:
yes.

kaniyan:
Fixed via [3e788a4](https://github.com/eclipse/hono/commit/3e788a43d8c0e39ceda2d85859ebc81753ee4565)

PR #1343, opened 2019-07-10
[#1284] Extend message limits for command and control

kaniyan:
With this PR, the Protocol Adapters are now equipped to check the _message limits_ before sending any _commands_ and accepting any _command responses_ as described in #1284.

### DIFF
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
@@ -375,7 +375,7 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
             }
             return null;
         }).otherwise(t -> {
-            con.setCondition(AmqpContext.getErrorCondition(t));
+            con.setCondition(getErrorCondition(t));
             con.close();
             TracingHelper.logError(span, t);
             return null;
@@ -541,7 +541,7 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
             return d;
         }).recover(t -> {
             if (t instanceof ClientErrorException) {
-                final ErrorCondition condition = AmqpContext.getErrorCondition(t);
+                final ErrorCondition condition = getErrorCondition(t);
                 MessageHelper.rejected(ctx.delivery(), condition);
             } else {
                 ProtonHelper.released(ctx.delivery(), true);
@@ -686,7 +686,7 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
 
                     if (!command.isValid()) {
                         final Exception ex = new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST, "malformed command message");
-                        commandContext.reject(AmqpContext.getErrorCondition(ex), 1);
+                        commandContext.reject(getErrorCondition(ex), 1);
                         metrics.reportCommand(
                                 command.isOneWay() ? Direction.ONE_WAY : Direction.REQUEST,
                                 sourceAddress.getTenantId(),
@@ -788,7 +788,7 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
             final Throwable t,
             final Span span) {
 
-        final ErrorCondition ec = AmqpContext.getErrorCondition(t);
+        final ErrorCondition ec = getErrorCondition(t);
         LOG.debug("closing link with error condition [symbol: {}, description: {}]", ec.getCondition(), ec.getDescription());
         link.setCondition(ec);
         link.close();
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
@@ -702,7 +702,22 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
                                 command.getPayloadSize(),
                                 timer);
                     } else {
-                        onCommandReceived(sender, commandContext);
+                        getTenantConfiguration(sourceAddress.getTenantId(), commandContext.getTracingContext())
+                                .compose(tenantObject -> checkMessageLimit(tenantObject, command.getPayloadSize()))
+                                .setHandler(result -> {
+                                    if (result.succeeded()) {
+                                        onCommandReceived(sender, commandContext);
+                                    } else {
+                                        // issue credit so that application(s) can send the next command
+                                        commandContext.reject(getErrorCondition(result.cause()), 1);
+                                        metrics.reportCommand(
+                                                command.isOneWay() ? Direction.ONE_WAY : Direction.REQUEST,
+                                                sourceAddress.getTenantId(),
+                                                ProcessingOutcome.from(result.cause()),
+                                                command.getPayloadSize(),
+                                                timer);
+                                    }
+                                });
                     }
                 }, closeHandler -> {},
                 DEFAULT_COMMAND_CONSUMER_CHECK_INTERVAL_MILLIS);
@@ -933,7 +948,11 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
             final Future<JsonObject> tokenFuture = getRegistrationAssertion(resource.getTenantId(),
                     resource.getResourceId(), context.getAuthenticatedDevice(), currentSpan.context());
             final Future<TenantObject> tenantEnabledFuture = getTenantConfiguration(resource.getTenantId(),
-                    currentSpan.context()).compose(tenantObject -> isAdapterEnabled(tenantObject));
+                    currentSpan.context())
+                            .compose(tenantObject -> CompositeFuture
+                                    .all(isAdapterEnabled(tenantObject),
+                                            checkMessageLimit(tenantObject, context.getPayloadSize()))
+                                    .map(success -> tenantObject));
 
             return CompositeFuture.all(tenantEnabledFuture, tokenFuture)
                     .compose(ok -> sendCommandResponse(resource.getTenantId(), commandResponse, currentSpan.context()));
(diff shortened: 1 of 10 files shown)


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -702,7 +702,23 @@ private Span newSpan(final String operationName, final Device authenticatedDevic
...
+                                    return Future.succeededFuture();
+                                })
+                                .otherwise(t -> {
+                                    // issue credit so that application(s) can send the next command
+                                    commandContext.reject(getErrorCondition(t), 1);
+                                    metrics.reportCommand(
+                                            command.isOneWay() ? Direction.ONE_WAY : Direction.REQUEST,
+                                            sourceAddress.getTenantId(),
+                                            ProcessingOutcome.from(t),
+                                            command.getPayloadSize(),
```
sophokles73: so, if the message limit is reached, we do not process the command but bill the customer for it?

kaniyan: In the above case, if the message limit is reached, the command is not processed and the corresponding metrics for that command is reported with `ProcessingOutcome` as `UNPROCESSABLE`. 

The metrics data from the backend having `ProcessingOutcome` tagged as `FORWARDED` or `UNPROCESSABLE` are used for the message limit calculation. The same is also applicable to _telemetry/event_ messages. What do you mean by bill the customer? Can you pls explain more?

sophokles73: hmm. Let's say I have a limit of 10 MB. From my application I send several commands with a big payload which cannot be processed because I never include a subject. My understanding is that the commands will all *eat up* my message volume, i.e. I have to pay for them, regardless of whether they have reached the destination (device). So, once the limit is reached, will the payload still be accounted for, i.e. will I still need to pay for it, even if now all messages are being rejected? I think I should, because my sending the of these (feasible) commands still generates costs on our side, right? So, based on that, reporting the command's payload seems to be correct but should also be subjected to the *minimum message size* calculation, right?
The problem is that we cannot prevent the application from sending additional commands, thus causing more and more costs, because we only check the limit *after* we have accepted the messages via the AMQP Network ...

kaniyan: > So, based on that, reporting the command's payload seems to be correct but should also be subjected to the _minimum message size_ calculation, right?

yes. This PR is to address the issue #1284, which is to equip the adapters with message limit checks while processing commands. The minimum message size calculation as described in issue #1330 will be part of another [PR](https://github.com/eclipse/hono/pull/1333).

> The problem is that we cannot prevent the application from sending additional commands, thus causing more and more costs, because we only check the limit _after_ we have accepted the messages via the AMQP Network ...

yes. I think the same is also applicable for the _Telemetry/Event_ messages i.e we cannot prevent the devices from sending additional messages after the limit is exceeded. If this is the indented behaviour i.e the customer will be charged for the additional non processed messages then it should be well documented, so that customers can design their implementation accordingly.  

Thereby making it also important in Hono to inform the devices and north bound applications with this exact error reason so that they can act upon it. With HTTP and AMQP protocols I believe that they are well informed with status codes _429_ and `amqp:resource-limit-exceeded`. But the MQTT protocol doesn't provide space for such an option and we are closing the MQTT connection as recommended by the spec. IMHO it would be worth to think in a separate issue how to communicate the errors effectively in case of MQTT. May be a separate topic to which the interested devices can subscribe to know about the errors until we support MQTT 5 ;-). MQTT 5 has option to send reason codes along with control packets.

sophokles73:
> But the MQTT protocol doesn't provide space for such an option and we are closing the MQTT connection as recommended by the spec. 

Now that we disconnect on error, we could simply reject new connections from devices of the tenant if the limit has been reached ...

sophokles73:
For HTTP there doesn't seem to be any way to prevent rogue devices (or attackers) to produce costs because we can only reject a request once we are processing it and by that time, the payload of the request has already been uploaded ...

kaniyan:
> > But the MQTT protocol doesn't provide space for such an option and we are closing the MQTT connection as recommended by the spec.
> 
> Now that we disconnect on error, we could simply reject new connections from devices of the tenant if the limit has been reached ...
> We might also be able to use the same approach for devices using AMQP and other stateful protocols.

yes, we can do that for messages published by the devices. But for the north bound application this validation happens at the adapter side and by that time the connection is already established. 
I will raise a new issue for this, so that we can discuss there about the various possibilities and which can be followed by a PR.

kaniyan:
> yes, we can do that for messages published by the devices. But for the north bound application this validation happens at the adapter side and by that time the connection is already established.
> I will raise a new issue for this, so that we can discuss there about the various possibilities and which can be followed by a PR.

@sophokles73 A separate ssue #1350 has been created to address the above discussed point. Would you mind taking a look at this current PR and merge it?

adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapterTest.java
```diff
@@ -68,6 +67,7 @@
 import org.junit.runner.RunWith;
 import org.mockito.ArgumentCaptor;
 
+import io.netty.handler.codec.http.HttpResponseStatus;
```
sophokles73: I think we should use our own constant definition instead of depending on a low level library ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1211,7 +1230,7 @@ protected void onMessageSent(final MqttContext ctx) {
      * This default implementation does nothing.
      * <p>
      * Subclasses should override this method in order to e.g. update metrics counters.
-     * 
+     *
```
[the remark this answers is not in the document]
kaniyan: Done and in future it shouldn't introduce such formatting.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -98,6 +100,7 @@
      */
     protected static final String KEY_MICROMETER_SAMPLE = "micrometer.sample";
 
+    private static final int HTTP_TOO_MANY_REQUESTS = 429;
```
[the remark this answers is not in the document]
kaniyan: Done

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -928,12 +928,29 @@ protected void setNonEmptyResponsePayload(final HttpServerResponse response, fin
...
-                                addMicrometerSample(commandContext, commandSample);
-                                // put command context to routing context and notify
-                                ctx.put(CommandContext.KEY_COMMAND_CONTEXT, commandContext);
-                                cancelCommandReceptionTimer(ctx);
-                                setTtdStatus(ctx, TtdStatus.COMMAND);
-                                responseReady.tryComplete();
+                                getTenantConfiguration(tenantId, commandContext.getTracingContext())
+                                        .compose(tenantObject -> checkMessageLimit(tenantObject,
+                                                command.getPayloadSize()))
+                                        .compose(success -> {
```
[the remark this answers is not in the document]
kaniyan: Done

adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapterTest.java
kaniyan: Replaced with `HttpUtils.HTTP_TOO_MANY_REQUESTS`.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -928,12 +928,29 @@ protected void setNonEmptyResponsePayload(final HttpServerResponse response, fin
...
+                                        .otherwise(t -> {
+                                            // issue credit so that application(s) can send the next command
+                                            commandContext.reject(getErrorCondition(t), 1);
+                                            metrics.reportCommand(
+                                                    command.isOneWay() ? Direction.ONE_WAY : Direction.REQUEST,
+                                                    tenantId,
+                                                    ProcessingOutcome.from(t),
+                                                    command.getPayloadSize(),
+                                                    commandSample);
+                                            return null;
```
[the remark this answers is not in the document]
kaniyan: `responseReady.tryComplete()` has been included.

kaniyan:
@sophokles73 I have pushed the changes.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -928,12 +928,28 @@ protected void setNonEmptyResponsePayload(final HttpServerResponse response, fin
...
+                                getTenantConfiguration(tenantId, commandContext.getTracingContext())
+                                        .compose(tenantObject -> checkMessageLimit(tenantObject,
+                                                command.getPayloadSize()))
+                                        .setHandler(result -> {
+                                            if (result.succeeded()) {
+                                                addMicrometerSample(commandContext, commandSample);
+                                                // put command context to routing context and notify
+                                                ctx.put(CommandContext.KEY_COMMAND_CONTEXT, commandContext);
+                                                cancelCommandReceptionTimer(ctx);
+                                                setTtdStatus(ctx, TtdStatus.COMMAND);
```
sophokles73: IMHO we need to cancel the timer and set the status in both cases so these two lines should be moved down to where we complete the responseReady future

kaniyan: I moved them.

sophokles73:
@kaniyan can you rebase?

kaniyan:
> @kaniyan can you rebase?

@sophokles73 rebased.

PR #1344, opened 2019-07-10
Migrate DeviceConnectionService to new Service structure

calohmn:


services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/AutowiredDeviceConnectionAdapter.java
```diff
@@ -0,0 +1,48 @@
...
+ * This wires up the actual service instance with the mapping to the event bus implementation. It is intended to be used
+ * in a Spring Boot environment.
+ */
+@Component
+public final class AutowiredDeviceConnectionAdapter extends EventBusDeviceConnectionAdapter<Void> {
+
+    private DeviceConnectionService service;
+
+    @Autowired
+    @Qualifier("serviceImpl")
```
calohmn: Note: The other `Autowired*Adapter` classes have the `backend` qualifier here, but since there is no need for a unifying `DeviceConnectionBackend`, there is no separate `backend` bean used here.

service-base/src/main/java/org/eclipse/hono/service/deviceconnection/BaseDeviceConnectionService.java
```diff
@@ -18,167 +18,38 @@
...
  *
- * @param <T> The type of configuration properties this service requires.
+ * @param <T> The type of configuration class this service supports.
+ * @deprecated - Use {@link DeviceConnectionService} instead.
  */
-public abstract class BaseDeviceConnectionService<T> extends EventBusService<T> implements DeviceConnectionService {
-
-    private static final String SPAN_NAME_GET_LAST_GATEWAY = "get last known gateway";
-    private static final String SPAN_NAME_SET_LAST_GATEWAY = "set last known gateway";
+@Deprecated
```
ctron: I am not sure if we really need this class. It is deprecated, but we also didn't have this in the last release.

calohmn: The `BaseDeviceConnectionService` class is used in the `Application` class of the legacy device registry implementation.
I think it's more straightforward using the deprecated class there in the same fashion as the other deprecated service base classes.

ctron: I was assuming to use the new pattern in the legacy device registry as well. Its just a bit weird that we add a new and immediately deprecated feature.

But yes, it might be good to simply repeat the pattern.

PR #1349, opened 2019-07-11
[#1238] Introduce secure in addition to insecure health check endpoints

mbaeuerle:
This is the second part of #1238 which will make it possible to expose a secure and an insecure health check endpoint.

This first draft is intended to discuss about the implementation.

Open questions:
- Which one of the health check endpoints should be the default in Hono? (Depending on that I will also update the deployment configurations)
- Do we want the full fledged tls / native tls support like in `org.eclipse.hono.service.AbstractServiceBase#addTlsKeyCertOptions`?


mbaeuerle:
@sophokles73 WDYT about the open points and the overall direction of this PR?

sophokles73:
> Which one of the health check endpoints should be the default in Hono? (Depending on that I will also update the deployment configurations)

I do not think that there is a default. In our example deployment, I would, however, like to use the secure endpoint because we already have server certs for all components.
In the integration tests, it makes sense to stick with the insecure endpoint because the health check is only used by Docker during the start-up of the components and I do not want to add our CA cert to Docker's trust store for that purpose.

sophokles73:
> Do we want the full fledged tls / native tls support like in org.eclipse.hono.service.AbstractServiceBase#addTlsKeyCertOptions?

I do not think that this is necessary for now. It can easily be added later, though ...

sophokles73:
@mbaeuerle is this ready for review? If so, then please remove the *draft* mark ...

mbaeuerle:
@sophokles73 I am currently working on unit tests for this. As the integration tests will still use the insecure health checks it seems difficult to test it with integration tests.
As soon as I finish, I will remove the _draft_ mark!

mbaeuerle:
@sophokles73 I added a test and marked the PR as ready for review now.
We have to think about the documentation of the feature. Previously the `setPort` set the insecure port, now it sets the secure port. So I would rewrite the API changes from the previous commit like so, also in the adapter docs:
> 
> * The configuration parameters for the health check endpoint were renamed and moved from 
>   `hono.app` to `hono.healthCheck`. `hono.app.healthCheckPort` is now
>   `hono.healthCheck.insecurePort` and `hono.app.healthCheckBindAddress` is now 
>   `hono.healthCheck.insecurePortBindAdress`. See the individual 
>   [Adapter Configuration guides]({{% doclink "/admin-guide/" %}}) 
>   for more information on the new naming. 

The parameters of the new secure endpoint then also have to be added.
WDYT?

sophokles73:
> The parameters of the new secure endpoint then also have to be added.
WDYT?

sounds reasonable :+1:

service-base/src/test/java/org/eclipse/hono/service/VertxBasedHealthCheckServerTest.java
```diff
@@ -0,0 +1,180 @@
...
+        config.setInsecurePort(0);
+        config.setInsecurePortBindAddress(HOST);
+
+        final VertxBasedHealthCheckServer server = new VertxBasedHealthCheckServer(vertx, config);
+        registerHealthChecks(ctx, server, 1);
+
+        server.start().compose(result -> Future.succeededFuture(getWebClient(server, false)))
+                .compose(httpClient -> checkHealth(ctx, httpClient, "/liveness"))
+                .compose(httpClient -> checkHealth(ctx, httpClient, "/readiness"))
+                .compose(result -> server.stop())
```
sophokles73: IMHO the server should be stopped in any case, i.e. the server should be stopped in a separate method that is annotated with `@AfterEach`

service-base/src/test/java/org/eclipse/hono/service/VertxBasedHealthCheckServerTest.java
```diff
@@ -0,0 +1,180 @@
...
+import io.vertx.core.Future;
+import io.vertx.core.Vertx;
+import io.vertx.ext.healthchecks.HealthCheckHandler;
+import io.vertx.ext.web.client.WebClient;
+import io.vertx.ext.web.client.WebClientOptions;
+import io.vertx.ext.web.client.predicate.ResponsePredicate;
+import org.eclipse.hono.config.ServerConfig;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
```
sophokles73: please use JUnit 5 instead

tests/src/test/resources/amqp/application.yml
```diff
@@ -3,8 +3,9 @@ hono:
     maxInstances: 1
     startupTimeout: 120
   healthCheck:
-    bindAddress: 0.0.0.0
-    port: ${vertx.health.port}
+    insecurePortBindAddress: 0.0.0.0
+    insecurePort: ${vertx.health.port}
+    insecurePortEnabled: true
```
sophokles73: no need to set this to true if an insecure port is set explicitly

mbaeuerle:
> Have you tried to run the example deployment in minikube? I wonder if the readiness probes can be invoked successfully, given that the server certs are not signed by a public CA ...

Yes I have run it with minikube and had no issues with the certificate. When connecting directly with curl to the health check endpoints I indeed run into the issue that the certificate is not trusted.
Seems like Kubernetes does not check this: 

>  If scheme field is set to HTTPS, the kubelet sends an HTTPS request skipping the certificate verification. [Source](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes)

sophokles73:
@mbaeuerle are you making the remaining changes?

service-base/src/test/java/org/eclipse/hono/service/VertxBasedHealthCheckServerTest.java
mbaeuerle: So the suggestion is to introduce a member variable which holds the current `VertxBasedHealthCheckServer` and in the `@AfterEach` annotated method this server is stopped if it exists?

sophokles73: That should do the trick, shouldn't it?

mbaeuerle:
@sophokles73 I made the requested changes and also updated the documentation/release-notes.

I stumbled upon a question: Does the `Auth Server` expose a health check? I haven't found the configuration parameters mentioned in the [docs](https://www.eclipse.org/hono/docs/latest/admin-guide/auth-server-config/)  but the health endpoint is configured in its [deployment config](https://github.com/eclipse/hono/blob/master/deploy/src/main/config/hono-service-auth-config.yml#L4). If I manually call the health check endpoint in my minikube deployment it returns 000 as status code with empty content.
Does the Auth Server expose a Health Check server at all?

sophokles73:
> Does the Auth Server expose a health check?

yes, it does. However, no particular checks are implemented. We mainly need to start the Health Check server in order to expose the scraping endpoint for Prometheus ...

mbaeuerle:
>  I haven't found the configuration parameters mentioned in the docs.

Sorry my sentence was misleading. What I meant was that the health check parameters are missing in the Auth Server docs. So as we expose the health check endpoint in the Auth Server, IMHO it makes sense to also add the parameters to the docs.

sophokles73:
> So as we expose the health check endpoint in the Auth Server, IMHO it makes sense to also add the parameters to the docs.

Can you create an issue for that?

sophokles73:
@mbaeuerle can you also rebase?

mbaeuerle:
@sophokles73 I created the ticket about the missing docs here: https://github.com/eclipse/hono/issues/1383

PR #1351, opened 2019-07-11
Fix typos and minor issues

kaniyan:
Typos and minor issues have been addressed in this PR.


core/src/main/java/org/eclipse/hono/util/CredentialsObject.java
```diff
@@ -294,6 +294,7 @@ public void checkSecrets(final BiConsumer<String, JsonObject> secretValidator) {
                         checkHashedPassword(secret);
                         secretValidator.accept(getType(), secret);
                     });
+                    break;
```
ctron: I actually think we should add a checkstyle rule to catch things like this.

sophokles73: @ctron are you adding the rule?

Issue #896, opened 2018-11-14
Enable forced sampling of traces for a specific tenant (Http adapter)


calohmn:
The OpenTracing Specification allows setting a specific tag (`sampling.priority` - see [Span tags](https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-tags-table)), 
with which it can be enforced that trace spans are actually captured in the Tracing component.

To enable debugging of problems related to messages of a specific tenant in Hono, it would be good to be able to define such an enforced trace sampling for all spans regarding that specific tenant.

For that, the tenant JSON structure, as defined in the Tenant API, could be extended with an additional, optional attribute: `tracing-sampling-priority` (integer). 
It's semantics should follow the one given for the `sampling.priority` tag:
````
If greater than 0, a hint to the Tracer to do its best to capture the trace. 
If 0, a hint to the trace to not-capture the trace. 
If absent, the Tracer should use its default sampling mechanism.
````
For completeness, I would suggest having a value of `-1` be equivalent with the value being absent.

This `tracing-sampling-priority` setting of a tenant should then be applied (in the sense of setting the `sampling.priority` tag) to the traces associated with that tenant.

EDIT: this issue shall track the implementation of this feature for the HTTP adapter only.

calohmn:
I've done some testing with setting the `sampling.priority` tag.
It looks like Jaeger currently requires that tag to be present in all spans of a trace for the complete trace to be forcefully sampled. I.e. it is not enough to set the tag on one span (root span or a child span) and then have all the spans of that trace be automatically forcefully sampled.

That means that the information, whether spans should be forcefully sampled, has to be given to all methods where new child spans get created.
In Hono, there is currently the `SpanContext` object used as a method parameter in these methods.
As a first step (and first PR), I would suggest replacing these parameters with a new Object that holds `SpanContext` and the sampling priority.

(Thinking further about tracing across components, this all probably means that we should consider transferring the sampling priority information in a baggage item so that other components can implement the tenant-specific tracing priority as well.)

calohmn:
In case there are a lot of traces being generated for a particular tenant, it would be a good feature to have an option to only force sampling of *error traces* for that particular tenant.

This could be implemented in these ways:
- A: Either with these attributes in the tenant json:
````
tracing-sampling-priority             [0|1|-1]    => optional, default -1 (i.e. use overall sampling strategy)
tracing-sampling-priority-applies-to  [error|all] => optional, default value could be "error"
````
- B: or with these:
[...]

PR #1352, opened 2019-07-12
Refactor upload*Message() methods

calohmn:
This serves as a preparation for #896, for which the `getTenantConfiguration` invocation shall be put first in the `upload*Message()` methods of the protocol adapters.

Objective of this PR:
- reduce indentation levels
- reduce exit points of the methods, combining error handling and metric reporting

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -64,6 +64,7 @@ import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.Router;
 import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.handler.BodyHandler;
+import io.vertx.proton.ProtonDelivery;
 
 /**
  * Base class for a Vert.x based Hono protocol adapter that uses the HTTP protocol.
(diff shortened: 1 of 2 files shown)


adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -1082,28 +1083,23 @@ public final void uploadCommandResponseMessage(
...
                     tenant,
                     deviceId,
                     authenticatedDevice,
                     currentSpan.context());
             final Future<Void> tenantEnabledTracker = getTenantConfiguration(tenant, currentSpan.context())
                     .compose(tenantObject -> isAdapterEnabled(tenantObject).map(ok -> null));
-            CompositeFuture.all(deviceRegistrationTracker, tenantEnabledTracker)
-                    .compose(ok -> sendCommandResponse(tenant, commandResponse, currentSpan.context()))
+            return CompositeFuture.all(deviceRegistrationTracker, tenantEnabledTracker)
+                    .compose(ok -> sendCommandResponse(tenant, commandResponseFuture.result(), currentSpan.context()))
```
ctron: I don't understand the reason to use the future here. I get the reason for that in general, but in this location I think you could/should use `resp` instead, which should be the same value, unless I missed something.

If I understood it correctly, then technically it should't make a difference if you use `resp`, `commandResponseFuture.result()` or `commandResponse` here. However code might change over time, and this would seem more robust to me.

calohmn: Ok, I've changed it to `resp` and then renamed it to `commandResponse` (renaming the other `commandResponse`).

PR #1353, opened 2019-07-12
Add checkstyle check for missing "break" in switch

ctron:
The build of this PR will fail, as long as #1351 is not merged.


sophokles73:
@ctron it seems that the check also detects some false positives which need to be marked accordingly ...

sophokles73:
@ctron the pattern matching doesn't seem to work yet ...

sophokles73:
@ctron never mind, I will do it myself ...

PR #1356, opened 2019-07-15
[#896] Implement config option to force sampling of traces

calohmn:
Fix for #896. 

This introduces config options to enable forced sampling of traces (i.e. recording of traces in the tracing backend) for a specific tenant or device.

**Per tenant:**
In the Json structure returned by the Tenant API, a field `trace-sampling-mode` can be set with a value of "all" (meaning sampling of all traces is enforced) or "default".
````
  {
    "enabled": true,
    "trace-sampling-mode": "all"
  }
````

**Per device:**
In the Json structure returned by the Tenant API, a field `trace-sampling-mode-per-device` can be set with a map value where a field with the device id as name and a value of "all" (meaning sampling of all traces is enforced) or "default" is set.
````
  {
    "enabled": true,
    "trace-sampling-mode-per-device": {
      "4711": "all"
    }
  }
````

**Limitations:**
The sampling of traces can only be enforced by setting a "sampling-priority" tag on all spans that shall be sampled. That means that all spans created and finished *before* getting the tenant Json can not be taken into account here. For example, from all the spans created as part of uploading telemetry data in the HTTP protocol adapter:
````
hono-adapter-http: POST
  hono-adapter-http: authenticate device
    hono-adapter-http: get Credentials
      hono-service-device-registry: process request message
        hono-service-device-registry: get Credentials
    hono-adapter-http: validate Credentials
  hono-adapter-http: upload telemetry
    hono-adapter-http: get Tenant by ID
      hono-service-device-registry: process request message
        hono-service-device-registry: get Tenant
    hono-adapter-http: assert Device Registration
      hono-service-device-registry: process request message
        hono-service-device-registry: assert Device Registration
hono-adapter-http: forward Telemetry data
````
only these get forcefully sampled for now:
````
hono-adapter-http: POST
  hono-adapter-http: upload telemetry
    hono-adapter-http: assert Device Registration
      hono-service-device-registry: process request message
        hono-service-device-registry: assert Device Registration
hono-adapter-http: forward Telemetry data
````

**Scope:**
As part of this PR, the feature has been implemented for the HTTP, MQTT and AMQP adapters.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/tracing/TracingHelper.java
+++ b/core/src/main/java/org/eclipse/hono/tracing/TracingHelper.java
@@ -19,6 +19,8 @@ import java.util.Objects;
 import java.util.Optional;
 
 import org.apache.qpid.proton.message.Message;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import io.opentracing.References;
 import io.opentracing.Span;
@@ -97,6 +99,8 @@ public final class TracingHelper {
     private static final String JSON_KEY_SPAN_CONTEXT = "span-context";
 
     private static final String AMQP_ANNOTATION_NAME_TRACE_CONTEXT = "x-opt-trace-context";
+
+    private static final Logger LOG = LoggerFactory.getLogger(TracingHelper.class);
 
     private TracingHelper() {
         // prevent instantiation
@@ -305,11 +309,15 @@ public final class TracingHelper {
     }
 
     /**
-     * Creates a span builder that is initialized with the given operation name and a child reference to the given span
-     * context (if set).
+     * Creates a span builder that is initialized with the given operation name and a child-of reference to the given
+     * span context (if set).
+     * <p>
+     * If the given span context contains a "sampling.priority" baggage item, it is set as a tag in the returned span
+     * builder.
      *
      * @param tracer The Tracer to use.
-     * @param spanContext The span context to set as child reference (may be null).
+     * @param spanContext The span context that shall be the parent of the Span being built and that is used to derive
+     *            the sampling priority from (may be null).
      * @param operationName The operation name to set for the span
      * @return The span builder.
      * @throws NullPointerException if tracer or operationName is {@code null}.
@@ -322,9 +330,13 @@ public final class TracingHelper {
     /**
      * Creates a span builder that is initialized with the given operation name and a follows-from reference to the
      * given span context (if set).
+     * <p>
+     * If the given span context contains a "sampling.priority" baggage item, it is set as a tag in the returned span
+     * builder.
      *
      * @param tracer The Tracer to use.
-     * @param spanContext The span context to set as follows-from reference (may be null).
+     * @param spanContext The span context that the span being build shall have a follows-from reference to and that is
+     *            used to derive the sampling priority from (may be null).
      * @param operationName The operation name to set for the span
      * @return The span builder.
      * @throws NullPointerException if tracer or operationName is {@code null}.
@@ -337,9 +349,12 @@ public final class TracingHelper {
     /**
      * Creates a span builder that is initialized with the given operation name and a reference to the given span
      * context (if set).
+     * <p>
+     * If the given span context contains a "sampling.priority" baggage item, it is set as a tag in the returned span
+     * builder.
      *
      * @param tracer The Tracer to use.
-     * @param spanContext The span context to set as reference (may be null).
+     * @param spanContext The span context to set as reference and to derive the sampling priority from (may be null).
      * @param operationName The operation name to set for the span
      * @param referenceType The type of reference towards the span context.
      * @return The span builder.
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java [hunk 1]
```diff
@@ -1640,4 +1642,23 @@ public final ConnectionLimitManager getConnectionLimitManager() {
...
+     * <p>
+     * The sampling priority defines in how far <em>OpenTracing</em> spans created when processing
+     * messages for the tenant and device shall be recorded (sampled) by the tracing system.
+     *
+     * @param tenantObject The tenant configuration to get the sampling priority information from.
+     * @param deviceId The device id (may be null).
+     * @param span The span to apply the sampling priority to.
+     */
+    protected void applyTraceSamplingPriority(final TenantObject tenantObject, final String deviceId, final Span span) {
+        final Optional<Integer> samplingPriority = tenantObject.getTraceSamplingPriority(deviceId);
```
sophokles73: IMHO if we already return an `Optional` instead of just the int, then we should at least take advantage of the *syntactic sugar* that if offers, don't you think:
```
tenantObject.getTraceSamplingPriority(deviceId)
.ifPresent(prio -> (...));
```

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -1640,4 +1642,23 @@ public final ConnectionLimitManager getConnectionLimitManager() {
...
+     * Applies the sampling priority configured for the given tenant and device to the given span.
+     * <p>
+     * The sampling priority defines in how far <em>OpenTracing</em> spans created when processing
+     * messages for the tenant and device shall be recorded (sampled) by the tracing system.
+     *
+     * @param tenantObject The tenant configuration to get the sampling priority information from.
+     * @param deviceId The device id (may be null).
+     * @param span The span to apply the sampling priority to.
+     */
+    protected void applyTraceSamplingPriority(final TenantObject tenantObject, final String deviceId, final Span span) {
```
sophokles73: is this supposed to be overridden by subclasses?

core/src/main/java/org/eclipse/hono/tracing/TracingHelper.java
```diff
@@ -307,9 +311,13 @@ public static SpanContext extractSpanContext(final Tracer tracer, final MultiMap
     /**
      * Creates a span builder that is initialized with the given operation name and a child reference to the given span
      * context (if set).
+     * <p>
+     * If the given span context contains a "sampling.priority" baggage item, it is set as a tag in the returned span
+     * builder.
      *
      * @param tracer The Tracer to use.
-     * @param spanContext The span context to set as child reference (may be null).
+     * @param spanContext The span context to set as child reference and to derive the sampling priority from (may be
```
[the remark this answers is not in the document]
calohmn: Right, I've corrected that.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
calohmn: Yes, I've changed that accordingly.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
calohmn: No, I've marked the method as `final`.

sophokles73:
It looks ok to me. However, I am not very convinced of the overall functionality provided. The fact that some of the interactions are not traced will lead to a lot of confusion with users FMPOV as you will need to have a good understanding of how this is implemented in order to understand which interactions you can expect to show up in the traces and which not. All in all I do not think that this is the way to go as it also makes the code (even) harder to understand as well. However, if you and the other committers think that this is what we need/want, then be it ...

calohmn:
I agree that it is not very satisfactory to have not all spans of the trace be forcefully sampled here.

In order to forcefully sample *all* spans, we would have to extract tenant information early on and use this to determine the tenant object `trace-sampling-mode` setting - *before* querying the credentials service. In the case of the HTTP adapter, we would probably have to do this in the `HonoBasicAuthHandler` and the `X509AuthHandler`.
My first intuition here was that this would be more invasive and more complex than the current solution (which I think doesn't add too much complexity).

But let me have a closer look, how it would look like.

sophokles73:
> But let me have a closer look, how it would look like.

My intention was not to cause even more work on this. I can think of a more elegant way of doing this at the very beginning of any request coming in but that would require the use of SNI for tenant selection. That is a thing to consider anyway, FMPOV ...

calohmn:
@sophokles73 I've made a new implementation in #1382. There, all trace spans get forcefully sampled and the logic is encapsulated in one class `TenantTraceSamplingHandler`, registered as handler for the routes of the HTTP adapter.
Implementation for AMQP and MQTT will go into separate PRs.

Issue #1369, opened 2019-07-17
Device registry doesn't start without a health check port configured


jbtrystram:
The documentation specifies this property as not mandatory. Should we update the doc or is this a bug ?

edit : also , providing a value with an environment variable (i.e. `HONO_HEALTH_CHECK_PORT` do not work anymore)

sophokles73:
Looking at the source code, it seems to me that this property has been required for quite some time. However, we might want to define a (reasonable) default value for it so that we do not need to explicitly set it?

jbtrystram:
I don't see the value of it being a mandatory configuration value. A default value would be enough I think.
I'll have a quick look to add a default value.

sophokles73:
I guess the original idea was to not start the health check server at all, if no port has been configured explicitly. However, I do not see (m)any use cases where you would want to explicitly prevent the health check server from being started. In particular, since we are also using the health check server to expose the scraping endpoint for Prometheus ...
@dejanb @ctron WDYT?

jbtrystram:
Configuring a value as an environment variable doesn't work either for the auth service, so it is not limited to the device registry.

sophokles73:
@jbtrystram please create a new issue for your problems with the environment variables not being picked up. This issue has been created for another reason ...

PR #1373, opened 2019-07-17
Add default config values for healthCheck server.

jbtrystram:
Fix a part of #1369 . It now uses a default port of 8088 and bind to localhost.

Also, I could not get the "default" behaviour to work, as it is documented.  when no health check options are set, it should start without the health check server. It just fails to start in my tests. 

Now it works as documented.


jbtrystram:
I guess i should wait for https://github.com/eclipse/hono/pull/1349 to be merged and update accordingly :)

sophokles73:
yep

jbtrystram:
@sophokles73 this is rebased on master

service-base/src/main/java/org/eclipse/hono/service/VertxBasedHealthCheckServer.java
```diff
@@ -69,13 +70,13 @@
...
      * @throws NullPointerException if vertx is {@code null}.
      * @throws NullPointerException if config is {@code null}.
-     * @throws IllegalArgumentException if health check port is not configured.
      */
     public VertxBasedHealthCheckServer(final Vertx vertx, final ServerConfig config) {
         this.vertx = Objects.requireNonNull(vertx);
         this.config = Objects.requireNonNull(config);
         if (!isSecureEndpointConfigured() && !isInsecureEndpointConfigured()) {
-            throw new IllegalArgumentException("No health check configured");
+            config.setInsecurePort(DEFAULT_HEALTH_CHECK_PORT);
```
sophokles73: IMHO we should not overwrite the config properties but instead use `ServerConfig.getPort(String)` when starting the server to fall back to the default port.

service-base/src/main/java/org/eclipse/hono/service/VertxBasedHealthCheckServer.java
```diff
@@ -69,13 +70,13 @@
...
+            LOG.info("health checkc HTTP server will use default insecure port: {}", config.getInsecurePort());
```
sophokles73: spelling

site/documentation/content/admin-guide/amqp-adapter-config.md
```diff
@@ -36,8 +36,8 @@ The following table provides an overview of the configuration variables and corr
...
-| `HONO_HEALTH_CHECK_INSECURE_PORT`<br>`--hono.healthCheck.insecurePort` | no | - | The insecure port the HTTP server, which exposes the service's health check resources, should listen on. If set, the adapter will expose a *readiness* probe at URI `/readiness` and a *liveness* probe at URI `/liveness`. |
+| `HONO_HEALTH_CHECK_INSECURE_PORT_BIND_ADDRESS`<br>`--hono.healthCheck.insecurePortBindAddress` | no | `127.0.0.1` | The IP address of the network interface that the insecure port, which exposes the service's health check resources, should be bound to. |
+| `HONO_HEALTH_CHECK_INSECURE_PORT`<br>`--hono.healthCheck.insecurePort` | no | `8088` | The insecure port of the HTTP server, which exposes the service's health check resources, should listen on. The adapter will expose a *readiness* probe at URI `/readiness` and a *liveness* probe at URI `/liveness`. |
```
sophokles73: please apply your changes to the secure port description as well
and don't forget the Auth Server config guide

jbtrystram: So the default behaviour would be have default values for secure + insecure ports ?

sophokles73: good question. So far, we used the port property to control whether a server should be started at all. Now, with having a default port, I guess we could use the availability of key material as a discriminator for which server to start, e.g. if there is key material (and no ports being set explicitly), we start the secure server (using the default port and the key material). If no key material is available (and no ports are set explicitly), we start the insecure server. The default port should be the same for both, though, FMPOV.
WDYT?

jbtrystram: That sounds reasonable. The only issue i see with this solution is that if the default port is the same for secure and insecure we may end up with mismatches, where the client may expect a insecure connection while the server only accept a secure one.

jbtrystram:
@sophokles73  I think that should do it. Here is a current behaviour : 
- if `keyPath` and `CertPath` options are present ( and valid) the HealtCheck server will start in secure mode.
- If `keyPath` and `CertPath`  are not present or incorrect the server will be started in insecure mode (with a warning)
- If the secure options are correct and both `port` `insecurePort` settings are set, both servers will be started.

WDYT ?

site/documentation/content/admin-guide/amqp-adapter-config.md
jbtrystram: update on this : i tested to start both secure and insecure endpoint with the same port number : both works fine, so this is not an issue. Nice ! :)

sophokles73: Are you saying that the health server can be successfully accessed by a client using either HTTP or HTTPS in this case?

jbtrystram: yes

service-base/src/main/java/org/eclipse/hono/service/VertxBasedHealthCheckServer.java
```diff
@@ -128,17 +116,23 @@ public void setAdditionalResources(final List<Handler<Router>> resourceProviders
         registerAdditionalResources();
 
         final Future<Void> result = Future.future();
-        CompositeFuture.all(bindSecureHttpServer(), bindInsecureHttpServer())
-                .compose(r -> result.complete(), result);
+        bindSecureHttpServer().setHandler(res -> {
+            if (res.failed()){
+                bindInsecureHttpServer().compose(r -> result.complete(), result);
+            } else if (config.isInsecurePortEnabled() && config.isSecurePortEnabled()) {
+                bindInsecureHttpServer().compose(r -> result.complete(), result);
```
sophokles73: this binds the insecure server to the same port as the secure server, right? What is the purpose of this? Does this even work? IMHO the insecure server should only be started in this case (i.e. the secure server has been started already) if the insecure port is different from the secure one ...

jbtrystram: that what I  wanted to do in the first place, but I tried this and it seems to work, but I there was a environment variable in my setup that caused me to misunderstand the result. I'll push a fix right now

jbtrystram: @sophokles73  now it should be good.

sophokles73: I think this is still not correct. The user can indicate that the insecure server should be started by doing any or both of the following:
* set *insecurePortEnabled* to `true`
* set *insecurePort* to a value other than `Constants.PORT_UNCONFIGURED`

If only *insecurePortEnabled* is set to `true` but no specific port is set, then the default insecure port is used.
IMHO this means for the HealthCheckServer, assuming that the secure server has started successfully:
only start the insecure server
* if *insecurePortEnabled* is set to `true` and *config.getInsecurePort(DEFAULT_PORT)* is different from *server.actualPort()*, or
* if *insecurePort* is not `Constants.PORT_UNDEFINED` and *config.getInsecurePort()* is different from *server.actualPort()*

jbtrystram: Any suggestion for the insecure port default value ? What about `8043` ?

sophokles73: I think we can still use the same default port. If somebody wants to use a secure **and** insecure server, they should be required to select either a specific secure and/or insecure port ...

jbtrystram: :+1:

site/homepage/content/release-notes.md
```diff
@@ -21,6 +21,9 @@ title = "Release Notes"
   for details regarding these new variables.
 * The *message limit* checks is now extended to include command and control messages.
   Please refer to the [resource limits] ({{% doclink "/concepts/resource-limits/" %}}) for details.
+* The health check endpoint will bind to the localhost interface and use a default
```
sophokles73: hmmm, this doesn't make much sense, does it? The idea behind using a default port was that people would get away with not specifying anything, right? However, in order to get a usable health check server, you still need to set the bind address explicitly because otherwise the server is bound to the loopback device only, i.e. it can't be accessed from any other node, which kind of makes the whole thing useless ...

That is also the reason why we originally had the sentence in the docs, that you need to set the bind address in order to get the server started. So, for the health check server we would probably want to require users to explicitly configure the bind address because otherwise the health check server is useless anyway. However, we should not by default simply start the server on `0.0.0.0` because that makes it available to all other nodes and I guess we should leave it for the user to decide if that is really intended.
WDYT?

jbtrystram: Yeah, that make sense. I'm so used to develop locally I forgot that `127.0.0.1` was not reachable from any other nodes :sweat_smile: 
So if I summarize the ways to get a health Check server started by doing any of the following:
For the insecure server : 
- set InsecurePortEnabled to true AND set a InsecureBindAddress  to a value other than _127.0.0.1_ .
- set InsecurePort to a value other than Constants.PORT_UNCONFIGURED AND set a InsecureBindAddress  to a value other than _127.0.0.1_ .

For the secure server : 
- set a BindAddress  to a value other than _127.0.0.1_ AND provide valid key material.
- set Port to a value other than Constants.PORT_UNCONFIGURED AND set a BindAddress  to a value other than _127.0.0.1_  AND provide valid key material.

If that sounds correct I'll implement that

sophokles73: That sounds good :-) Make sure to keep the check for making sure that secure and insecure port are different, though ...

service-base/src/main/java/org/eclipse/hono/service/VertxBasedHealthCheckServer.java
```diff
@@ -159,18 +161,20 @@ public void setAdditionalResources(final List<Handler<Router>> resourceProviders
...
-        } else {
-            return Future.succeededFuture();
         }
+            return result;
     }
 
     private Future<Void> bindSecureHttpServer() {
-        if (config.isSecurePortEnabled()) {
+        if (config.getBindAddress() == Constants.LOOPBACK_DEVICE_ADDRESS
+        || ( config.getBindAddress() == Constants.LOOPBACK_DEVICE_ADDRESS && config.getPort() == Constants.PORT_UNCONFIGURED )) {
```
sophokles73: IMHO it should be sufficient to check if a bind address has been set, shouldn't it?

service-base/src/main/java/org/eclipse/hono/service/VertxBasedHealthCheckServer.java
```diff
@@ -128,17 +118,29 @@ public void setAdditionalResources(final List<Handler<Router>> resourceProviders
...
         return result;
     }
 
     private Future<Void> bindInsecureHttpServer() {
-        if (config.isInsecurePortEnabled()) {
-            final Future<Void> result = Future.future();
+        final Future<Void> result = Future.future();
 
+        if ((config.getInsecurePortBindAddress() == Constants.LOOPBACK_DEVICE_ADDRESS && config.isInsecurePortEnabled())
+             || ( config.getInsecurePortBindAddress() == Constants.LOOPBACK_DEVICE_ADDRESS && config.getInsecurePort() == Constants.PORT_UNCONFIGURED )) {
```
sophokles73: IMHO it should be sufficient to check if a bind address has been set here.
The `else` branch should then check if either the `insecurePortEnabled` property has been set to `true` or a port has been set explicitly ...

service-base/src/main/java/org/eclipse/hono/service/VertxBasedHealthCheckServer.java
jbtrystram: ah good catch, there is no need to check the port here

service-base/src/main/java/org/eclipse/hono/service/VertxBasedHealthCheckServer.java
jbtrystram: I did ask myself which way to go as well. Both works fine, but checking the port configuration separately allows for a more preciser log message. I'll update in that direction.

jbtrystram: I updated to match what we agreed on earlier, but I feel there is a little bit of redundancy .

service-base/src/main/java/org/eclipse/hono/service/VertxBasedHealthCheckServer.java
```diff
@@ -128,17 +118,31 @@ public void setAdditionalResources(final List<Handler<Router>> resourceProviders
...
 
     private Future<Void> bindInsecureHttpServer() {
-        if (config.isInsecurePortEnabled()) {
-            final Future<Void> result = Future.future();
+        final Future<Void> result = Future.future();
 
+        if (config.getInsecurePortBindAddress() == Constants.LOOPBACK_DEVICE_ADDRESS) {
+            LOG.warn("Won't start insecure health checks HTTP server: no bind address configured.");
+            return Future.failedFuture("NOT_ENABLED");
+        } else if (config.isInsecurePortEnabled() || config.getInsecurePort() == Constants.PORT_UNCONFIGURED ) {
```
sophokles73: it would be easier to read if we would check for the condition that actually allows to start the server:
```
else if (config.isInsecurePortEnabled() || config.getInsecurePort() != Constants.PORT_UNCONFIGURED) {
  // start server
} else {
  // log error
}
```
In any case, the current check is wrong as it doesn't start the server if the *insecurePortEnabled* flag is set to `true` ...

jbtrystram: thanks for catching that. It's probably a bad copy and paste..

service-base/src/main/java/org/eclipse/hono/service/VertxBasedHealthCheckServer.java
```diff
@@ -128,17 +118,28 @@ public void setAdditionalResources(final List<Handler<Router>> resourceProviders
...
-                .compose(r -> result.complete(), result);
+            bindSecureHttpServer().setHandler(res -> {
+                if (res.failed()){
+                    bindInsecureHttpServer().compose(r -> result.complete(), result);
+                } else if ( (config.isInsecurePortEnabled() && config.getInsecurePort(DEFAULT_PORT) != server.actualPort() )
+                        || (config.getInsecurePort() != Constants.PORT_UNCONFIGURED && config.getInsecurePort() != server.actualPort())) {
+                    bindInsecureHttpServer().compose(r -> result.complete(), result);
+                } else {
+                    result.complete();
+                }
```
calohmn: @jbtrystram The `isInsecurePortEnabled` javadoc states that `If {@code false}, it is guaranteed by the server that no opened port is insecure`.
With the above code and the conditions in `bindInsecureHttpServer()` however, an insecure port may indeed be opened even if `isInsecurePortEnabled == false`.

It's also what I would expect, that for example with a config like
````
insecurePortEnabled: false
insecurePort: 8088
````
_no_ insecure port 8088 would be opened.

sophokles73: @calohmn The health check server does not consider this property at all. See https://www.eclipse.org/hono/docs/admin-guide/monitoring-tracing-config/#health-check-server-configuration

calohmn: This makes the usage of `isInsecurePortEnabled()` here a bug then, I suppose, and all the more confusing anyway, IMHO.

And I think it would be good to add a note in the above code then, that `InsecurePortEnabled` is always implicitly `true` for the health check. Otherwise someone used to how the insecure port is handled in other places of Hono might wonder why the flag is not considered here.

sophokles73: @calohmn this code has already been changed on master. You are reviewing *old* code ...

calohmn: oh, sorry, happened when using 2 local Hono repos with one not being up-to-date.

Issue #1355, opened 2019-07-15
Error during startup of example Device Registry


calohmn:
During startup of the device registry component (with the new API), there is this error in the log file:
````
16:20:04.024 [vert.x-eventloop-thread-0] DEBUG o.e.h.d.FileBasedRegistrationService - loading devices for tenant [DEFAULT_TENANT]
16:20:04.040 [vert.x-eventloop-thread-0] DEBUG o.e.h.d.FileBasedRegistrationService - cannot load device identities from file [/var/lib/hono/device-registry/device-identities.json]: Cannot deserialize instance of `java.util.ArrayList` out of VALUE_STRING token
 at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: org.eclipse.hono.service.management.device.Device["via"])
````
This points to the `via` entry in the `device-identities.json` not supporting a string value anymore but requiring an array now.

calohmn:
I think, the fact that a `via` string value is not supported anymore should also be mentioned in the release-notes.

What about date entries like `2017-05-01T14:00:00+01:00` not working anymore - should we consider this a bug? Or just require the UTC format with the new API?

calohmn:
Regarding the date value in the credentials object:
_Previously_, the conversion from `String` to `Instant` was done manually (using `CredentialsObject#getTimestampIfPresentForField`, which uses `DateTimeFormatter.ISO_OFFSET_DATE_TIME`).

_Now_ the conversion is done via the `jsonObject.mapTo(CommonCredential.class)` invocation, which will use the `Json#InstantSerializer` (which uses `DateTimeFormatter.ISO_INSTANT`) to map the `CommonSecret.notBefore` `Instant` field.

dejanb:
@calohmn I think we can support single string converting to array using Jackson settings, so we can maintain backward compatibility in that way. I'll take a look into this.

dejanb:
> I think, the fact that a `via` string value is not supported anymore should also be mentioned in the release-notes.
> 
> What about date entries like `2017-05-01T14:00:00+01:00` not working anymore - should we consider this a bug? Or just require the UTC format with the new API?

Hm, as far as I can see this date format is used in example and test credentials.

https://github.com/eclipse/hono/blob/master/tests/src/test/resources/deviceregistry/credentials.json#L12

 Can you give me the exact value that's causing this problem?

calohmn:
[...]

PR #1374, opened 2019-07-17
[#1378] Support datetime format with timezone offset

dejanb:
This should allow encoding datetime values both with and without timezone offset. The value will be decoded without the offset.

### DIFF
--- a/service-base/pom.xml
+++ b/service-base/pom.xml
@@ -44,6 +44,11 @@
       <artifactId>hono-core</artifactId>
     </dependency>
     <dependency>
+        <groupId>com.fasterxml.jackson.datatype</groupId>
+        <artifactId>jackson-datatype-jsr310</artifactId>
+        <version>${jackson.version}</version>
+    </dependency>
+    <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
--- a/service-base/src/main/java/org/eclipse/hono/service/management/credentials/CommonSecret.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/management/credentials/CommonSecret.java
@@ -16,17 +16,10 @@ import static org.eclipse.hono.util.RegistryManagementConstants.FIELD_SECRETS_NO
 import static org.eclipse.hono.util.RegistryManagementConstants.FIELD_SECRETS_NOT_BEFORE;
 
 import java.time.Instant;
-import java.time.OffsetDateTime;
-import java.time.ZoneId;
-import java.time.format.DateTimeFormatter;
-import java.time.format.DateTimeParseException;
-import java.util.Optional;
 
-import com.fasterxml.jackson.annotation.JsonGetter;
-import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.annotation.JsonSetter;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.MoreObjects.ToStringHelper;
 
@@ -39,8 +32,12 @@ public abstract class CommonSecret {
     @JsonProperty
     private Boolean enabled;
 
-    private OffsetDateTime notBefore;
-    private OffsetDateTime notAfter;
+    @JsonProperty(FIELD_SECRETS_NOT_BEFORE)
+    @JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ss[XXX]", timezone = "UTC")
+    private Instant notBefore;
+    @JsonProperty(FIELD_SECRETS_NOT_AFTER)
+    @JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ss[XXX]", timezone = "UTC")
+    private Instant notAfter;
     @JsonProperty
     private String comment;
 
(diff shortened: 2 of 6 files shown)


service-base/pom.xml
```diff
@@ -43,6 +43,11 @@
       <groupId>org.eclipse.hono</groupId>
       <artifactId>hono-core</artifactId>
     </dependency>
+    <dependency>
+        <groupId>com.fasterxml.jackson.datatype</groupId>
+        <artifactId>jackson-datatype-jsr310</artifactId>
```
sophokles73: we'll need a CQ for this

dejanb: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20419

sophokles73:
@dejanb so, in order to be able to at least do the example deployment, we will need to change the date/time format used in the example credentials, right?

dejanb:
@sophokles73 no, this should work properly with the existing credentials. There's a test in this pr to guard that

sophokles73:
@dejanb according to #1355 the credentials cannot be retrieved from the registry, though ...

dejanb:
@sophokles73 this is the pr to fix that :)

sophokles73:
I know, but because we'll have to wait for the CQ to be approved, we cannot merge it (yet). For the time being, though, it would be nice to have a working system, don't you think?

dejanb:
Depends on how long it takes for this to be merged. I would like to have a release with the original example credentials for backward compatibility. 
I'll create a pr with the temporal change in the examples.

ctron:
A quick question, and please forgive me that I didn't follow this discussion: but why do we need a timezone? Other than to read old (legacy) JSON files?

dejanb:
@ctron we had this discussion on the chat the other day. Resolution: to make it easier for users to define proper time without the constant need to recalculate to UTC

ctron:
But that would be a feature for a UI, not the storage backend. Because now you need to always recalculate to UTC when checking. Which might be more frequent than editing.

dejanb:
This is just one more option for the user, considering that we don't have an UI :)

ctron:
Can we suggest to prefer UTC.

service-base/pom.xml
Alfusainey: any news on this yet?

service-base/src/test/java/org/eclipse/hono/service/management/credentials/SecretsTest.java
```diff
@@ -134,4 +143,30 @@ public void testDecodeGeneric() {
...
+                .put(FIELD_SECRETS_NOT_BEFORE, "2017-05-01T14:00:00+01:00")
+                .put(FIELD_SECRETS_NOT_AFTER, "2037-06-01T14:00:00Z");
+
+
+        final PasswordSecret secret = json.mapTo(PasswordSecret.class);
+        final LocalDateTime before = LocalDateTime.of(2017, 05, 01, 14, 00, 00);
+        assertEquals(before.toInstant(ZoneOffset.of("+01:00")), secret.getNotBefore());
+
+        final JsonObject decodedSecret = JsonObject.mapFrom(secret);
+        assertEquals("2017-05-01T13:00:00Z", decodedSecret.getValue("not-before"));
```
Alfusainey: @dejanb Don't you think the expected value in this case should be: `2017-05-01T14:00:00+01:00`, exactly as was specified in the JSON object and not the value in `DateTimeFormatter.ISO_INSTANT` format?

dejanb: @Alfusainey the idea is that ISO_INSTANT is preferred format and that we just supporting defining dates in different formats for better usability.

dejanb:
@sophokles73 CQ has been approved. Do you want to proceed with this or keep the manual formatting?

sophokles73:
> Do you want to proceed with this or keep the manual formatting?

At the moment I do not see much benefit in introducing the additional dependency. Do you?

service-base/src/test/java/org/eclipse/hono/service/management/credentials/SecretsTest.java
Alfusainey: thanks 👍

sophokles73:
@dejanb with #1403 around the corner it probably makes sense to rebase and merge this PR. WDYT?

service-base/pom.xml
sophokles73: the CQ is approved

service-base/src/main/java/org/eclipse/hono/service/management/credentials/CommonSecret.java
```diff
@@ -52,116 +49,20 @@ public void setEnabled(final Boolean enabled) {
...
-     * Gets the earliest instant in time that this secret may be used for authenticating a device.
-     * 
-     * @return The instant.
-     */
-    @JsonIgnore
     public Instant getNotBefore() {
-        return Optional.ofNullable(notBefore)
-                .map(nb -> nb.toInstant())
-                .orElse(null);
+        return notBefore;
```
sophokles73: can we agree to keep the JavaDoc?

service-base/src/main/java/org/eclipse/hono/service/management/credentials/CommonSecret.java
```diff
@@ -52,116 +49,20 @@ public void setEnabled(final Boolean enabled) {
...
-     * @param notBefore The instant.
-     */
-    @JsonIgnore
     public void setNotBefore(final Instant notBefore) {
-        if (notBefore == null) {
-            this.notBefore = null;
-        } else {
-            this.notBefore = OffsetDateTime.ofInstant(notBefore, ZoneId.systemDefault());;
-        }
+        this.notBefore = notBefore;
```
sophokles73: keep JavaDoc?

service-base/src/main/java/org/eclipse/hono/service/management/credentials/CommonSecret.java
```diff
@@ -52,116 +49,20 @@ public void setEnabled(final Boolean enabled) {
...
-     * @see DateTimeFormatter#ISO_OFFSET_DATE_TIME
-     */
-    @JsonSetter(FIELD_SECRETS_NOT_AFTER)
-    public void setNotAfter(final String notAfter) {
-        try {
-            this.notAfter = DateTimeFormatter.ISO_OFFSET_DATE_TIME.parse(notAfter, OffsetDateTime::from);
-        } catch (DateTimeParseException e) {
-            this.notAfter = null;
-        }
+        return notAfter;
```
sophokles73: keep JavaDoc?

service-base/src/main/java/org/eclipse/hono/service/management/credentials/CommonSecret.java
```diff
@@ -52,116 +49,20 @@ public void setEnabled(final Boolean enabled) {
...
-     * @param notAfter The instant.
-     */
-    @JsonIgnore
     public void setNotAfter(final Instant notAfter) {
-        if (notAfter == null) {
-            this.notAfter = null;
-        } else {
-        this.notAfter = OffsetDateTime.ofInstant(notAfter, ZoneId.systemDefault());
-        }
+        this.notAfter = notAfter;
```
sophokles73: keep JavaDoc?

dejanb:
@sophokles73 this should be now good to go. Can you give it another look?

PR #1375, opened 2019-07-18
Remove usage of TenantObject in code performing tenant API client requests

Alfusainey:
This patch attempts to restrict the usage of the `TenantObject` value to what it is created for: encapsulating the tenant payload retrieved through the get operation of the tenant API.

This restriction avoids the unnecessary/costly mapping from a `TenantObject` to a `JsonObject`, as we currently do in test cases performing tenant API client requests

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/management/tenant/Adapter.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/management/tenant/Adapter.java
@@ -15,7 +15,11 @@ package org.eclipse.hono.service.management.tenant;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
+
 import org.eclipse.hono.util.RegistryManagementConstants;
 
 /**
@@ -35,6 +39,18 @@ public class Adapter {
     @JsonProperty(RegistryManagementConstants.FIELD_EXT)
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
     private Map<String, Object> extensions;
+
+    /**
+     * Creates a new adapter instance for the given type.
+     *
+     * @param type The adapter type.
+     * @throws NullPointerException if the type is null.
+     */
+    public Adapter(@JsonProperty("type") final String type) {
+        Objects.requireNonNull(type);
+
+        this.type = type;
+    }
 
     /**
      * Set the enabled property.
@@ -95,4 +111,25 @@ public class Adapter {
     public Map<String, Object> getExtensions() {
         return this.extensions;
     }
+
+    /**
+     * Adds an extension property to this adapter.
+     * <p>
+     * If an extension property already exist for the specified key, the old value is replaced by the specified value.
+     *
+     * @param key The key of the entry.
+     * @param value The value of the entry.
+     * @return This instance, to allow chained invocations.
+     * @throws NullPointerException if any of the arguments is {@code null}.
+     */
+    public Adapter putExtension(final String key, final Object value) {
+        Objects.requireNonNull(key);
+        Objects.requireNonNull(value);
+        if (this.extensions == null) {
+            this.extensions = new HashMap<>();
+        }
+        this.extensions.put(key, value);
+        return this;
+    }
+
 }
--- a/service-base/src/main/java/org/eclipse/hono/service/management/tenant/ResourceLimits.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/management/tenant/ResourceLimits.java
@@ -13,7 +13,9 @@
 
 package org.eclipse.hono.service.management.tenant;
 
+import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 
 import org.eclipse.hono.util.RegistryManagementConstants;
 
@@ -48,4 +50,25 @@ public class ResourceLimits {
     public Map<String, Object> getExtensions() {
         return this.extensions;
     }
+
+    /**
+     * Adds an extension property to this resource limit.
+     * <p>
+     * If an extension property already exist for the specified key, the old value is replaced by the specified value.
+     *
+     * @param key The key of the entry.
+     * @param value The value of the entry.
+     * @return This instance, to allow chained invocations.
+     * @throws NullPointerException if any of the arguments is {@code null}.
+     */
+    public ResourceLimits putExtension(final String key, final Object value) {
+        Objects.requireNonNull(key);
+        Objects.requireNonNull(value);
+        if (this.extensions == null) {
+            this.extensions = new HashMap<>();
+        }
+        this.extensions.put(key, value);
+        return this;
+    }
+
 }
--- a/service-base/src/main/java/org/eclipse/hono/service/management/tenant/Tenant.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/management/tenant/Tenant.java
@@ -20,6 +20,8 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+
 import org.eclipse.hono.util.RegistryManagementConstants;
 
 /**
@@ -125,17 +127,41 @@ public class Tenant {
     }
 
     /**
-     * Add a new extension entry to the Tenant.
+     * Adds an extension property to this tenant.
+     * <p>
+     * If an extension property already exist for the specified key, the old value is replaced by the specified value.
      *
      * @param key The key of the entry.
      * @param value The value of the entry.
-     * @return This instance, to allowed chained invocations.
+     * @return This instance, to allow chained invocations.
+     * @throws NullPointerException if any of the arguments is {@code null}.
      */
     public Tenant putExtension(final String key, final Object value) {
+        Objects.requireNonNull(key);
+        Objects.requireNonNull(value);
         if (this.extensions == null) {
             this.extensions = new HashMap<>();
         }
         this.extensions.put(key, value);
         return this;
     }
+
+    /**
+     * Adds the specified adapter to the list of adapters for this Tenant.
+     * 
+     * @param adapter The adapter to add for the tenant.
+     * @return        This instance, to allow chained invocations.
+     * @throws        NullPointerException if the specified adapter is {@code null}.
+     */
+    public Tenant addAdapterConfig(final Adapter adapter) {
+
+        Objects.requireNonNull(adapter);
+
+        if (adapters == null) {
+            adapters = new LinkedList<>();
+        }
+        adapters.add(adapter);
+        return this;
+    }
+
 }
--- a/service-base/src/test/java/org/eclipse/hono/service/management/tenant/TenantTest.java
+++ b/service-base/src/test/java/org/eclipse/hono/service/management/tenant/TenantTest.java
@@ -23,10 +23,11 @@ import static org.junit.jupiter.api.Assertions.*;
 import io.vertx.core.json.JsonArray;
 
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
 
 import io.vertx.core.json.Json;
 import io.vertx.core.json.JsonObject;
+
+import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.RegistryManagementConstants;
 import org.hamcrest.collection.IsEmptyIterable;
 import org.junit.jupiter.api.Test;
@@ -219,27 +220,20 @@ class TenantTest {
     @Test
     public void testSerializeAdapters() {
 
-        final Adapter httpAdapter = new Adapter()
-                .setType("http")
-                .setEnabled(false)
-                .setDeviceAuthenticationRequired(true);
-        final Adapter mqttAdapter = new Adapter()
-                .setType("mqtt")
-                .setEnabled(true)
-                .setDeviceAuthenticationRequired(true);
-
-        final ArrayList<Adapter> adapters = new ArrayList<>();
-        adapters.add(httpAdapter);
-        adapters.add(mqttAdapter);
-
         final Tenant tenant = new Tenant();
         tenant.setEnabled(true);
-        tenant.setAdapters(adapters);
+        tenant
+            .addAdapterConfig(new Adapter(Constants.PROTOCOL_ADAPTER_TYPE_HTTP)
+                    .setEnabled(false)
+                    .setDeviceAuthenticationRequired(true))
+            .addAdapterConfig(new Adapter(Constants.PROTOCOL_ADAPTER_TYPE_MQTT)
+                    .setEnabled(true)
+                    .setDeviceAuthenticationRequired(true));
 
         final JsonArray result = JsonObject.mapFrom(tenant).getJsonArray(FIELD_ADAPTERS);
         assertNotNull(result);
-        assertEquals("http", result.getJsonObject(0).getString(FIELD_ADAPTERS_TYPE));
-        assertEquals("mqtt", result.getJsonObject(1).getString(FIELD_ADAPTERS_TYPE));
+        assertEquals(Constants.PROTOCOL_ADAPTER_TYPE_HTTP, result.getJsonObject(0).getString(FIELD_ADAPTERS_TYPE));
+        assertEquals(Constants.PROTOCOL_ADAPTER_TYPE_MQTT, result.getJsonObject(1).getString(FIELD_ADAPTERS_TYPE));
         assertEquals(false, result.getJsonObject(0).getBoolean(FIELD_ENABLED));
         assertEquals(true, result.getJsonObject(0).getBoolean(FIELD_ADAPTERS_DEVICE_AUTHENTICATION_REQUIRED));
     }
--- a/tests/src/test/java/org/eclipse/hono/tests/DeviceRegistryHttpClient.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/DeviceRegistryHttpClient.java
@@ -34,7 +34,6 @@ import org.eclipse.hono.service.management.credentials.X509CertificateSecret;
 import org.eclipse.hono.service.management.device.Device;
 import org.eclipse.hono.service.management.tenant.Tenant;
 import org.eclipse.hono.util.RegistryManagementConstants;
-import org.eclipse.hono.util.TenantConstants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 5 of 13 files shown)


tests/src/test/java/org/eclipse/hono/tests/registry/TenantApiTests.java
```diff
@@ -83,14 +83,14 @@ public void testGetTenant(final VertxTestContext ctx) {
...
-        final TenantObject tenant = TenantObject.from(tenantId, true)
-                .setResourceLimits(resourceLimits)
-                .setDefaults(defaults)
-                .setAdapterConfigurations(adapterConfig)
-                .setProperty("customer", "ACME Inc.");
+        final JsonObject tenantPayload = TenantUtil.newTenant(tenantId, true)
+                .put(TenantConstants.FIELD_RESOURCE_LIMITS, resourceLimits)
+                .put(TenantConstants.FIELD_PAYLOAD_DEFAULTS, defaults)
+                .put(TenantConstants.FIELD_ADAPTERS, adapterConfig)
+                .put("customer", "ACME Inc.");
```
sophokles73: this kind of code where we need to use the property names explicitly is exactly what we want to prevent by using a value object.
FMPOV we first need to create a new value object to use for interacting with the Management API (e.g. TenantManagementObject`?) and use it in all code where we use the HTTP API to e.g. register a tenant in the integration tests. `TenantUtil` doesn't really serve that purpose FMPOV.
We can then remove everything from `TenantObject` that is not needed.

jbtrystram: see `org.eclipse.hono.service.management.tenant.Tenant`

Alfusainey: yes sure 👍 I have seen that value object class

sophokles73: I don't get it. Why is that a reason not to use the `Tenant` class? We only need it for setting up the  fixture, right? We do so utilizing the `DeviceRegistryHttpClient`, so all we would need to do is have that client also support passing in a `Tenant` object instead of a `JsonObject`, right?

Alfusainey: sure.. am working on this now

Alfusainey: I updated this PR. the build passes on my local machine.. don't know why its failing on travis

jbtrystram:
@Alfusainey  there is now two different Travis builds, for each device registry. In this case the integration tests with the legacy registry are failing. You can test it locally with the following maven profiles : `run-tests,tests-legacy-device-registry`.
(note that you still need the `run-tests` profile, `tests-legacy-device-registry` is just a toggle to select the legacy implementation of the registry)

Alfusainey:
@jbtrystram thanks :+1:  i had actually thought i ran legacy profile.. the build should pass now

sophokles73:
@Alfusainey can you rebase this PR?

Alfusainey:
> @Alfusainey can you rebase this PR?

done!

service-base/src/main/java/org/eclipse/hono/service/management/tenant/ResourceLimits.java
```diff
@@ -46,4 +49,20 @@ public void setExtensions(final Map<String, Object> extensions) {
     public Map<String, Object> getExtensions() {
         return this.extensions;
     }
+
+    @Override
+    public boolean equals(final Object obj) {
```
sophokles73: is there a particular reason why you override hash and equals?

tests/src/test/java/org/eclipse/hono/tests/DeviceRegistryHttpClient.java
```diff
@@ -138,17 +122,18 @@ public DeviceRegistryHttpClient(final Vertx vertx, final String host, final int
      * Adds configuration information for a tenant.
      *
      * @param tenantId The id of the tenant to add.
-     * @param requestPayload The request payload as specified by the Tenant API.
+     * @param requestPayload The request payload as specified by the Tenant management API.
```
sophokles73: IMHO you should adapt the JavaDoc of the other methods as you did here...

tests/src/test/java/org/eclipse/hono/tests/LegacyDeviceRegistryHttpClient.java
```diff
@@ -594,19 +593,20 @@ public LegacyDeviceRegistryHttpClient(final Vertx vertx, final String host, fina
     /**
      * Creates a tenant and adds a device to it with a given password.
      * <p>
-     * This method simply invokes {@link #addDeviceForTenant(TenantObject, String, JsonObject, String)}
+     * This method simply invokes {@link #addDeviceForTenant(String, JsonObject, String, JsonObject, String)}
```
sophokles73: why do you want to replace the TenantObject here with a JsonObject? Is there a reason not to stick with the original code (which we will remove soon anyway)?

core/src/main/java/org/eclipse/hono/util/TenantConstants.java
```diff
@@ -87,7 +87,7 @@
     /**
      * The name of the property that contains the configuration options for the resource limits.
      */
-    public static final String FIELD_RESOURCE_LIMITS = "resource-limits";
+    public static final String FIELD_RESOURCE_LIMITS = "limits";
```
sophokles73: the Tenant API defines this as `resource-limits` ...

tests/src/test/java/org/eclipse/hono/tests/registry/TenantApiTests.java
```diff
@@ -65,41 +66,47 @@
...
+        Tenants.setAdapterEnabled(tenant, Constants.PROTOCOL_ADAPTER_TYPE_HTTP, true, true, httpAdapterExtensions.getMap());
 
         getHelper().registry
-        .addTenant(JsonObject.mapFrom(tenant))
+        .addTenant(tenantId, tenant)
         .compose(ok -> getAdminClient().get(tenantId))
         .setHandler(ctx.succeeding(tenantObject -> {
             ctx.verify(() -> {
-                assertThat(tenantObject.isEnabled()).isTrue();
+                final Tenant convertedTenantObject = Tenants.convertTenantObject(tenantObject);
```
sophokles73: why do you want to convert the returned object here? The object has been retrieved using the `TenantClient` which is used by the protocol adapters. IMHO it would make more sense to verify, that the data returned by the registry contains all data that has been put into it using the Management API when we set up the fixture for this test.

tests/src/test/java/org/eclipse/hono/tests/Tenants.java
```diff
@@ -31,29 +41,95 @@
     private Tenants() {
     }
 
+    /**
+     * Convert the given TenantObject to a management API Tenant object.
+     * 
+     * @param tenantObject The tenant object to convert.
+     * @return The tenant or {@code null} if the tenant object to convert is null.
+     */
+    public static Tenant convertTenantObject(final TenantObject tenantObject) {
```
sophokles73: I do not see what we need this method for

tests/src/test/java/org/eclipse/hono/tests/registry/TenantApiTests.java
Alfusainey: that make sense 👍 removing the conversion will answer the following question

> is there a particular reason why you override hash and equals?

Alfusainey:
@sophokles73 i updated this PR..

service-base/src/main/java/org/eclipse/hono/service/management/tenant/ResourceLimits.java
```diff
@@ -17,6 +17,8 @@
 
 import org.eclipse.hono.util.RegistryManagementConstants;
 
+import org.eclipse.hono.util.RegistryManagementConstants;
```
sophokles73: this looks like a duplicate

tests/src/test/java/org/eclipse/hono/tests/Tenants.java
```diff
@@ -66,14 +80,42 @@ public static Tenant createTenantForTrustAnchor(final X509Certificate cert, fina
...
+    /**
+     * Creates a new adapter section and set the given properties for the adapter.
+     * 
+     * @param tenant        The tenant to process.
+     * @param type          The adapter type to configure.
+     * @param state         The enabled state.
+     * @param authRequired  The device-authentication-required property.
+     * @param extensions    The extensions properties of the tenant.
+     */
+    public static void setAdapterEnabled(final Tenant tenant, final String type, final Boolean state, final boolean authRequired, final Map<String, Object> extensions) {
```
sophokles73: this method's name is misleading. In fact, it *adds* a new configuration for an adapter based on the given properties.
FMPOV something like *addAdapterConfiguration* would be more appropriate. But then, why isn't this a method of the `Tenant` object itself?

tests/src/test/java/org/eclipse/hono/tests/registry/TenantApiTests.java
```diff
@@ -65,41 +69,63 @@
...
-                        .put(TenantConstants.FIELD_ADAPTERS_TYPE, Constants.PROTOCOL_ADAPTER_TYPE_HTTP)
-                        .put(TenantConstants.FIELD_ENABLED, true)
-                        .put(TenantConstants.FIELD_ADAPTERS_DEVICE_AUTHENTICATION_REQUIRED, true)
-                        .put("deployment", new JsonObject()
-                                .put("maxInstances", 4)));
+                        .put("effective-since", "2019-04-27T12:00:00+00:00");
+
+        final ResourceLimits resourceLimits = new ResourceLimits();
+        resourceLimits.setMaxConnections(100000);
+        if (resourceLimits.getExtensions() == null) {
```
sophokles73: how could it be not `null`?

tests/src/test/java/org/eclipse/hono/tests/registry/TenantApiTests.java
```diff
@@ -65,41 +69,63 @@
...
+            resourceLimits.setExtensions(new HashMap<>());
+        }
+
+        resourceLimits.getExtensions().put("data-volume", dataVolume.getMap());
```
sophokles73: what about just
```
resourceLimits.setExtensions(Collections.singletonMap("data-volume", dataVolume.getMap()));
```

tests/src/test/java/org/eclipse/hono/tests/registry/TenantApiTests.java
```diff
@@ -65,41 +69,63 @@
...
-                assertThat(tenantObject.isEnabled()).isTrue();
-                assertThat(tenantObject.getTenantId()).isEqualTo(tenantId);
-                assertThat(tenantObject.getResourceLimits()).isEqualTo(resourceLimits);
-                assertThat(tenantObject.getDefaults()).isEqualTo(defaults);
-                assertThat(tenantObject.getAdapterConfigurations()).isEqualTo(adapterConfig);
-                assertThat(tenantObject.getProperty("customer", String.class)).isEqualTo("ACME Inc.");
+                assertThat(tenantObject.getResourceLimits()).isEqualTo(expectedTenantObject.getResourceLimits());
+                assertThat(tenantObject.getDefaults()).isEqualTo(expectedTenantObject.getDefaults());
+                assertThat(tenantObject.getAdapterConfigurations()).isEqualTo(expectedTenantObject.getAdapterConfigurations());
+                assertThat(tenantObject.getProperty("ext", JsonObject.class).getString("customer")).isEqualTo("ACME Inc.");
```
sophokles73: why don't you simply compare to the values of the *tenant* object that you registered? I do not see why we need to create the *expectedTenantObject*, even more so, since the type of object returned by `getAdminClient().get(String)` is `Tenant`, not `TenantObject`, isn't it?

tests/src/test/java/org/eclipse/hono/tests/registry/TenantHttpIT.java
```diff
@@ -265,35 +262,25 @@ public void testGetAddedTenant(final VertxTestContext context)  {
...
-                .put(RegistryManagementConstants.FIELD_EXT, new JsonObject()
-                        .put("options", new JsonObject().put("maxInstances", 4))));
-        final TenantObject tenantPayload = TenantObject.from(true)
-                .setProperty(RegistryManagementConstants.FIELD_EXT, new JsonObject().put("plan", "unlimited"))
-                .setAdapterConfigurations(adapterConfigs);
-        return JsonObject.mapFrom(tenantPayload);
+    private static Tenant buildTenantPayload() {
+        final Tenant tenant = new Tenant();
+        tenant.setExtensions(new JsonObject().put("plan", "unlimited").getMap());
+        Tenants.setAdapterEnabled(tenant, Constants.PROTOCOL_ADAPTER_TYPE_HTTP, true, true);
```
sophokles73: again, I think that the *setAdapterEnabled* method should be replaced by `Tenant.addAdapterConfiguration` ...

tests/src/test/java/org/eclipse/hono/tests/Tenants.java
Alfusainey: > this method's name is misleading. In fact, it adds a new configuration for an adapter based on the given properties.

Yes, i agree!

> But then, why isn't this a method of the Tenant object itself?

IMHO, we should keep the `Tenant` class as simple as possible and avoid populating it with utility methods (as we currently do in the `TenantObject` class). If anybody would read the device registry API, especially the `Tenant` component schema, one would see a direct relationship between the schema and the Java class.

IMHO, we should not add utility methods to that class, to the point that it becomes hard to see how it even relates to the schema defined in the API specification. I would suggest creating a specialized utility class, that defines only static methods, for adding such helper methods. WDYT?

tests/src/test/java/org/eclipse/hono/tests/registry/TenantApiTests.java
Alfusainey: No, `getAdminClient().get(String)` returns a `TenantObject`(specifically, it returns `Future<TenantObject>`.  We use the `Tenant` value object when sending tenant requests and `TenantObject` value object for the response

tests/src/test/java/org/eclipse/hono/tests/registry/TenantHttpIT.java
Alfusainey: please see my comment above

tests/src/test/java/org/eclipse/hono/tests/registry/TenantApiTests.java
sophokles73: Ah, right, I see.

tests/src/test/java/org/eclipse/hono/tests/Tenants.java
sophokles73: I would (partially) agree if we would be generating the value objects from the OpenAPI spec. But since we don't, I feel that this more like a case of an anemic value object. There is no problem at all to see the relation between these objects and the OpenAPI spec's schema definition. They have the same name and properties. Adding methods that make using them much easier poses no problem at all FMPOV.
The confusion with the `TenantObject` only stems from the fact that we have been using the same object for both, the Management API and the Tenant API.

@dejanb @ctron WDYT?

dejanb: @sophokles73 @Alfusainey I generally agree that we can add these methods that deal with adapters directly to the `Tenant` class. But maybe for this use case, we can just define a new constructor for the `Adapter` class and `Tenant.addAdapter()` method, so we could do things like

`tenant.addAdapter(new Adapter(...))`

Alfusainey: thanks, i updated the PR to allow creating an adapter instance for a given type

Alfusainey:
@sophokles73 patch updated..

service-base/src/main/java/org/eclipse/hono/service/management/tenant/Adapter.java
```diff
@@ -95,4 +111,20 @@ public Adapter setExtensions(final Map<String, Object> extensions) {
...
     }
+
+    /**
+     * Add a new extension entry to the Adapter.
+     *
+     * @param key The key of the entry.
+     * @param value The value of the entry.
+     * @return This instance, to allow chained invocations.
+     */
+    public Adapter putExtension(final String key, final Object value) {
```
sophokles73: null checks?

service-base/src/main/java/org/eclipse/hono/service/management/tenant/ResourceLimits.java
```diff
@@ -48,4 +49,20 @@ public void setExtensions(final Map<String, Object> extensions) {
...
     }
+
+    /**
+     * Add a new extension entry.
+     *
+     * @param key The key of the entry.
+     * @param value The value of the entry.
+     * @return This instance, to allow chained invocations.
+     */
+    public ResourceLimits putExtension(final String key, final Object value) {
```
sophokles73: null checks?

service-base/src/main/java/org/eclipse/hono/service/management/tenant/Tenant.java
```diff
@@ -129,7 +130,7 @@ public void setTrustedCertificateAuthority(final TrustedCertificateAuthority tru
      *
      * @param key The key of the entry.
      * @param value The value of the entry.
-     * @return This instance, to allowed chained invocations.
+     * @return This instance, to allow chained invocations.
```
sophokles73: null checks?

IMHO the description should be changed to
> Adds or replaces an extension property to this tenant.

to also reflect the *put* characteristics ...

service-base/src/main/java/org/eclipse/hono/service/management/tenant/Tenant.java
```diff
@@ -138,4 +139,19 @@ public Tenant putExtension(final String key, final Object value) {
...
         return this;
     }
+
+    /**
+     * Adds the specified adapter to the list of adapters for this Tenant.
+     * 
+     * @param adapter The adapter to add for the tenant.
+     * @return        This instance, to allow chained invocations.
+     */
+    public Tenant addAdapter(final Adapter adapter) {
```
sophokles73: what if adapter is `null`?

sophokles73: actually, this method does not add an adapter but an adapter *configuration*. As such, I think the method should be called accordingly, e.g. *addAdapterConfig*

service-base/src/test/java/org/eclipse/hono/service/management/tenant/TenantTest.java
```diff
@@ -219,27 +220,20 @@ public void testEncodeMinimumMessageSize() {
...
+            .addAdapter(new Adapter(Constants.PROTOCOL_ADAPTER_TYPE_MQTT)
+                    .setEnabled(true)
+                    .setDeviceAuthenticationRequired(true));
 
         final JsonArray result = JsonObject.mapFrom(tenant).getJsonArray(FIELD_ADAPTERS);
         assertNotNull(result);
-        assertEquals("http", result.getJsonObject(0).getString(FIELD_ADAPTERS_TYPE));
-        assertEquals("mqtt", result.getJsonObject(1).getString(FIELD_ADAPTERS_TYPE));
+        assertEquals(Constants.PROTOCOL_ADAPTER_TYPE_HTTP, result.getJsonObject(0).getString(FIELD_ADAPTERS_TYPE));
+        assertEquals(Constants.PROTOCOL_ADAPTER_TYPE_MQTT, result.getJsonObject(1).getString(FIELD_ADAPTERS_TYPE));
```
sophokles73: :+1:

service-base/src/main/java/org/eclipse/hono/service/management/tenant/Adapter.java
```diff
@@ -95,4 +111,20 @@ public Adapter setExtensions(final Map<String, Object> extensions) {
     public Map<String, Object> getExtensions() {
         return this.extensions;
     }
+
+    /**
+     * Add a new extension entry to the Adapter.
```
sophokles73: IMHO this should be changed to
> Adds or replaces an extension property to this adapter.

to also reflect the *put* characteristics ...

service-base/src/main/java/org/eclipse/hono/service/management/tenant/ResourceLimits.java
```diff
@@ -48,4 +49,20 @@ public void setExtensions(final Map<String, Object> extensions) {
     public Map<String, Object> getExtensions() {
         return this.extensions;
     }
+
+    /**
+     * Add a new extension entry.
```
sophokles73: IMHO this should be changed to
> Adds or replaces an extension property.

to also reflect the *put* characteristics ...

service-base/src/main/java/org/eclipse/hono/service/management/tenant/Tenant.java
Alfusainey: Ok, now how about:

> Adds an extension property to this tenant. If an extension property already exist for the specified key, the old value is replaced by the specified value?

or maybe i should add a paragraph after the first sentence?

Alfusainey: how about extracting the `putExtension` call to a parent class and all classes requiring additional extensions will extend from this parent?

sophokles73: > or maybe i should add a paragraph after the first sentence?

:+1: 

> how about extracting the putExtension call to a parent class and all classes requiring additional extensions will extend from this parent?

IMHO that is not necessary

Issue #1370, opened 2019-07-17
Creating a tenant or device with a generated ID refuses an empty body


jbtrystram:
When I want to ceate a tenant or a device using the ID generation feature from the device registry, i would expect it to work even if the request body is empty. A workaround is to provide an empty JsonObject in the request body.

sophokles73:
I agree that this should work with an empty body (and no *content-type* header being set)

PR #1376, opened 2019-07-18
Management API : allow empty body for creating tenant or device with generated ID

jbtrystram:
this fixes https://github.com/eclipse/hono/issues/1370

### DIFF
--- a/tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationHttpIT.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationHttpIT.java
@@ -179,15 +179,27 @@ public class DeviceRegistrationHttpIT {
     }
 
     /**
-     * Verifies that a device cannot be registered if the request
+     * Verifies that a device can be registered if the request
      * does not contain a body.
      * 
      * @param ctx The vert.x test context
      */
     @Test
-    public void testAddDeviceFailsForMissingBody(final TestContext ctx) {
+    public void testAddDeviceSucceedsForEmptyBody(final TestContext ctx) {
 
-        registry.registerDevice(TENANT, deviceId, null, HttpURLConnection.HTTP_BAD_REQUEST).setHandler(ctx.asyncAssertSuccess());
+        registry.registerDevice(TENANT, deviceId, null, HttpURLConnection.HTTP_CREATED).setHandler(ctx.asyncAssertSuccess());
+    }
+
+    /**
+     * Verifies that a device can be registered if the request
+     * does not contain a body nor a content type.
+     *
+     * @param ctx The vert.x test context
+     */
+    @Test
+    public void testAddDeviceSucceedsForEmptyBodyAndContentType(final TestContext ctx) {
+
+        registry.registerDevice(TENANT, deviceId, null, null, HttpURLConnection.HTTP_CREATED).setHandler(ctx.asyncAssertSuccess());
     }
 
     /**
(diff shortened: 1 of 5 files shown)


sophokles73:
does this also allow omitting the content-type in the request?

jbtrystram:
> does this also allow omitting the content-type in the request?

@sophokles73 i am not certain. 
`curl -i -X POST http://$REGISTRY_IP:28080/tenants ` this works fine, and nothing is specified. as there is no body i would expect no content type is specified but i don't master Curl enough to be certain.

jbtrystram:
@sophokles73  `curl -i -H "content-type: " -X POST http://$REGISTRY_IP:28080/tenants `
this works as well. WDYT ?

sophokles73:
@jbtrystram you should run `curl -v -X POST http://$REGISTRY_IP:28080/tenants` to see the headers of the request being sent. I assume that no *Content-Type* header will be sent ...

jbtrystram:
> @jbtrystram you should run `curl -v -X POST http://$REGISTRY_IP:28080/tenants` to see the headers of the request being sent. I assume that no _Content-Type_ header will be sent ...

@sophokles73  indeed ! 
```
[hono] curl -v -X POST http://$REGISTRY_IP:28080/tenants                                                                                          getting-started  ✱
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 28080 (#0)
> POST /tenants HTTP/1.1
> Host: localhost:28080
> User-Agent: curl/7.64.0
> Accept: */*
> 
< HTTP/1.1 201 Created
< etag: c734d8a6-0172-4835-a37b-7a5237f840c1
< location: /tenants/64cc21b5-49ff-47a3-844f-6d0cb190a9f5
< content-type: application/json; charset=utf-8
< content-length: 45
< 
* Connection #0 to host localhost left intact
{"id":"64cc21b5-49ff-47a3-844f-6d0cb190a9f5"}% 
```

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationHttpIT.java
```diff
@@ -179,15 +179,15 @@ public void testAddDeviceFailsForMissingContentType(final TestContext ctx) {
...
     /**
-     * Verifies that a device cannot be registered if the request
+     * Verifies that a device can be registered if the request
      * does not contain a body.
      * 
      * @param ctx The vert.x test context
      */
     @Test
-    public void testAddDeviceFailsForMissingBody(final TestContext ctx) {
+    public void testAddDeviceSucceedsForMissingBody(final TestContext ctx) {
```
sophokles73: I would suggest to name this *testAddDeviceSucceedsForEmptyBody* as the body is not missing in this case, is it?

jbtrystram: Indeed, in this case, the body seems to be an empty JsonObject.. Now I don't understand why the test fails, as it should work with an empty JsonObject, as demonstrated in my other PR..

jbtrystram:
@sophokles73  I added a test to verify that it succeed with no body nor content-type set.

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceRegistrationHttpIT.java
sophokles73: from the perspective of the *registerDevice* method that is correct. However, we are not testing the client here but the service implementation and from the point of view of the service implementation the body is not *missing* but simply *empty*. The same holds true for the test that you have added ...

jbtrystram: I updated the names

Issue #1378, opened 2019-07-18
Cannot retrieve example credentials from Device Registry


sophokles73:
Another issue occurs when trying to retrieve the example credentials for device 4711 via the new management API:
````
java.lang.IllegalArgumentException: Expected an ISO 8601 formatted date time
 at [Source: UNKNOWN; line: -1, column: -1] (through reference chain: org.eclipse.hono.service.management.credentials.PasswordCredential["secrets"]->java.util.ArrayList[0]->org.eclipse.hono.service.management.credentials.PasswordSecret["not-before"])
  at com.fasterxml.jackson.databind.ObjectMapper._convert(ObjectMapper.java:3750)
  at com.fasterxml.jackson.databind.ObjectMapper.convertValue(ObjectMapper.java:3668)
  at io.vertx.core.json.JsonObject.mapTo(JsonObject.java:106)
  at org.eclipse.hono.deviceregistry.FileBasedCredentialsService.get(FileBasedCredentialsService.java:662)
  at org.eclipse.hono.deviceregistry.FileBasedDeviceBackend.get(FileBasedDeviceBackend.java:203)
  at org.eclipse.hono.service.management.credentials.EventBusCredentialsManagementAdapter.processGetRequest(EventBusCredentialsManagementAdapter.java:214)
  at org.eclipse.hono.service.management.credentials.EventBusCredentialsManagementAdapter.processRequest(EventBusCredentialsManagementAdapter.java:66)
  at org.eclipse.hono.service.EventBusService.processRequestMessage(EventBusService.java:160)
  at io.vertx.core.eventbus.impl.HandlerRegistration.deliver(HandlerRegistration.java:276)
  at io.vertx.core.eventbus.impl.HandlerRegistration.handle(HandlerRegistration.java:254)
  at io.vertx.core.eventbus.impl.EventBusImpl$InboundDeliveryContext.next(EventBusImpl.java:578)
  at io.vertx.core.eventbus.impl.EventBusImpl.lambda$deliverToHandler$5(EventBusImpl.java:537)
  at io.vertx.core.impl.ContextImpl.executeTask(ContextImpl.java:320)
  at io.vertx.core.impl.EventLoopContext.lambda$executeAsync$0(EventLoopContext.java:38)
  at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163)
[...]

PR #1380, opened 2019-07-18
[#1355] Change the format of credentials temporarily

dejanb:


dejanb:
This actually fixes [#1378]

PR #1382, opened 2019-07-18
[#896] Implement trace sampling mode option in HTTP adapter

calohmn:
Fix for #896. 

This introduces config options to enable forced sampling of traces (i.e. recording of traces in the tracing backend) for a specific tenant or device.

**Per tenant:**
In the Json structure returned by the Tenant API, a field `trace-sampling-mode` can be set with a value of "all" (meaning sampling of all traces is enforced) or "default".
````
  {
    "enabled": true,
    "trace-sampling-mode": "all"
  }
````

**Per device (auth id):**
In the Json structure returned by the Tenant API, a field `trace-sampling-mode-per-auth-id` can be set with a map value where a field with a device auth id as name and a value of "all" (meaning sampling of all traces is enforced) or "default" is set.
````
  {
    "enabled": true,
    "trace-sampling-mode-per-auth-id": {
      "4711": "all"
    }
  }
````

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -36,6 +36,7 @@ import org.eclipse.hono.service.auth.DeviceUser;
 import org.eclipse.hono.service.http.ComponentMetaDataDecorator;
 import org.eclipse.hono.service.http.DefaultFailureHandler;
 import org.eclipse.hono.service.http.HttpUtils;
+import org.eclipse.hono.service.http.TenantTraceSamplingHandler;
 import org.eclipse.hono.service.metric.MetricsTags;
 import org.eclipse.hono.service.metric.MetricsTags.Direction;
 import org.eclipse.hono.service.metric.MetricsTags.EndpointType;
@@ -295,11 +296,28 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
         final Router router = Router.router(vertx);
         LOG.info("limiting size of inbound request body to {} bytes", getConfig().getMaxPayloadSize());
         router.route().handler(BodyHandler.create(DEFAULT_UPLOADS_DIRECTORY).setBodyLimit(getConfig().getMaxPayloadSize()));
+        Optional.ofNullable(getTenantTraceSamplingHandler())
+                .ifPresent(tenantTraceSamplingHandler -> router.route().handler(tenantTraceSamplingHandler)
+                        .failureHandler(tenantTraceSamplingHandler));
         addTracingHandler(router, -5);
         // add default handler for failed routes
         router.route().order(-1).failureHandler(new DefaultFailureHandler());
 
         return router;
+    }
+
+    /**
+     * Gets a handler that determines the tenant associated with a request and applies the tenant specific trace
+     * sampling configuration (if set).
+     * <p>
+     * This method returns {@code null} by default.
+     * <p>
+     * Subclasses may override this method in order to return an appropriate handler.
+     * 
+     * @return The handler or {@code null}.
+     */
+    protected TenantTraceSamplingHandler getTenantTraceSamplingHandler() {
+        return null;
     }
 
     /**
--- a/adapters/http-vertx/src/main/java/org/eclipse/hono/adapter/http/impl/VertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx/src/main/java/org/eclipse/hono/adapter/http/impl/VertxBasedHttpProtocolAdapter.java
@@ -29,8 +29,12 @@ import org.eclipse.hono.service.auth.device.UsernamePasswordCredentials;
 import org.eclipse.hono.service.auth.device.X509AuthProvider;
 import org.eclipse.hono.service.http.HonoBasicAuthHandler;
 import org.eclipse.hono.service.http.HonoChainAuthHandler;
+import org.eclipse.hono.service.http.HttpContext;
+import org.eclipse.hono.service.http.HttpContextTenantAndAuthIdProvider;
 import org.eclipse.hono.service.http.HttpUtils;
+import org.eclipse.hono.service.http.TenantTraceSamplingHandler;
 import org.eclipse.hono.service.http.X509AuthHandler;
+import org.eclipse.hono.service.tenant.ExecutionContextTenantAndAuthIdProvider;
 import org.eclipse.hono.util.CommandConstants;
 import org.eclipse.hono.util.Constants;
 
@@ -57,6 +61,7 @@ public final class VertxBasedHttpProtocolAdapter extends AbstractVertxBasedHttpP
 
     private HonoClientBasedAuthProvider<UsernamePasswordCredentials> usernamePasswordAuthProvider;
     private HonoClientBasedAuthProvider<SubjectDnCredentials> clientCertAuthProvider;
+    private ExecutionContextTenantAndAuthIdProvider<HttpContext> tenantObjectWithAuthIdProvider;
 
     /**
      * Sets the provider to use for authenticating devices based on
@@ -84,6 +89,23 @@ public final class VertxBasedHttpProtocolAdapter extends AbstractVertxBasedHttpP
      */
     public void setClientCertAuthProvider(final HonoClientBasedAuthProvider<SubjectDnCredentials> provider) {
         this.clientCertAuthProvider = Objects.requireNonNull(provider);
+    }
+
+    /**
+     * Sets the provider that determines the tenant and auth-id associated with a request.
+     *
+     * @param provider The provider to use.
+     * @throws NullPointerException if provider is {@code null}.
+     */
+    public void setTenantObjectWithAuthIdProvider(final ExecutionContextTenantAndAuthIdProvider<HttpContext> provider) {
+        this.tenantObjectWithAuthIdProvider = Objects.requireNonNull(provider);
+    }
+
+    @Override
+    protected TenantTraceSamplingHandler getTenantTraceSamplingHandler() {
+        return new TenantTraceSamplingHandler(
+                Optional.ofNullable(tenantObjectWithAuthIdProvider)
+                        .orElse(new HttpContextTenantAndAuthIdProvider(getConfig(), getTenantClientFactory(), PARAM_TENANT, PARAM_DEVICE_ID)));
     }
 
     /**
--- a/service-base/src/main/java/org/eclipse/hono/service/http/HttpContextTenantAndAuthIdProvider.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/http/HttpContextTenantAndAuthIdProvider.java
@@ -1,0 +1,114 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.service.http;
+
+import java.util.Base64;
+import java.util.Objects;
+
+import javax.net.ssl.SSLSession;
+
+import org.eclipse.hono.client.TenantClientFactory;
+import org.eclipse.hono.config.ProtocolAdapterProperties;
+import org.eclipse.hono.service.tenant.BaseExecutionContextTenantAndAuthIdProvider;
+import org.eclipse.hono.service.tenant.ExecutionContextTenantAndAuthIdProvider;
+import org.eclipse.hono.service.tenant.TenantObjectWithAuthId;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.opentracing.SpanContext;
+import io.vertx.core.Future;
+import io.vertx.core.http.HttpHeaders;
+import io.vertx.ext.web.RoutingContext;
+
+/**
+ * Provides a method to determine the tenant and auth-id of a HTTP request from the given HttpContext.
+ */
+public class HttpContextTenantAndAuthIdProvider extends BaseExecutionContextTenantAndAuthIdProvider
+        implements ExecutionContextTenantAndAuthIdProvider<HttpContext> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(HttpContextTenantAndAuthIdProvider.class);
+
+    private final String tenantIdContextParamName;
+    private final String deviceIdContextParamName;
+
+    /**
+     * Creates a new HttpContextTenantAndAuthIdProvider.
+     * 
+     * @param config The configuration.
+     * @param tenantClientFactory The factory to use for creating a Tenant service client.
+     * @param tenantIdContextParamName The name of the HttpContext parameter name that provides the tenant id in case of
+     *            an unauthenticated request.
+     * @param deviceIdContextParamName The name of the HttpContext parameter name that provides the device id in case of
+     *            an unauthenticated request. The device id will be used as auth-id for the created
+     *            TenantObjectWithAuthId objects.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     */
+    public HttpContextTenantAndAuthIdProvider(final ProtocolAdapterProperties config,
+            final TenantClientFactory tenantClientFactory,
+            final String tenantIdContextParamName,
+            final String deviceIdContextParamName) {
+        super(config, tenantClientFactory);
+        this.tenantIdContextParamName = Objects.requireNonNull(tenantIdContextParamName);
+        this.deviceIdContextParamName = Objects.requireNonNull(deviceIdContextParamName);
+    }
+
+    @Override
+    public Future<TenantObjectWithAuthId> get(final HttpContext context, final SpanContext spanContext) {
+        if (config.isAuthenticationRequired()) {
+            return getTenantViaCert(context.getRoutingContext(), spanContext)
+                    .recover(thr -> getTenantFromAuthHeader(context.getRoutingContext(), spanContext));
+        }
+        final String tenantId = context.get(tenantIdContextParamName);
+        if (tenantId != null) {
+            // unauthenticated request
+            final String deviceId = context.get(deviceIdContextParamName);
+            return tenantClientFactory.getOrCreateTenantClient()
+                    .compose(tenantClient -> tenantClient.get(tenantId, spanContext))
+                    .map(tenantObject -> new TenantObjectWithAuthId(tenantObject, deviceId));
+        }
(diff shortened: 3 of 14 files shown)

### CODE REVIEW

File: service-base/src/main/java/org/eclipse/hono/service/http/TenantTraceSamplingHandler.java [hunk 1]
```diff
@@ -0,0 +1,188 @@
...
+                    tenantIdTmp = Constants.DEFAULT_TENANT;
+                    authIdTmp = userName;
+                } else {
+                    // userName is <authId>@<tenantId>
+                    final String[] userComponents = userName.split("@", 2);
+                    if (userComponents.length == 2) {
+                        tenantIdTmp = userComponents[1];
+                        authIdTmp = userComponents[0];
+                    }
+                }
```
calohmn: Note: The logic above:
````
                if (config.isSingleTenant()) {
                    tenantIdTmp = Constants.DEFAULT_TENANT;
                    authIdTmp = userName;
                } else {
                    // userName is <authId>@<tenantId>
                    final String[] userComponents = userName.split("@", 2);
                    if (userComponents.length == 2) {
                        tenantIdTmp = userComponents[1];
                        authIdTmp = userComponents[0];
                    }
                }
````
comes from `UsernamePasswordAuthProvider#getCredentials` (and the `UsernamePasswordCredentials.create` call inside).
I've considered using the `usernamePasswordAuthProvider` instance of the `VertxBasedHttpProtocolAdapter` here, to call `getCredentials` in order to determine tenant/authId from the userName.
But, `getCredentials` belongs to the `CredentialsApiAuthProvider` class. The `VertxBasedHttpProtocolAdapter.usernamePasswordAuthProvider` instance however is of the `HonoClientBasedAuthProvider` type, which is the parent class of `CredentialsApiAuthProvider`. So, using `getCredentials` doesn't work here in all cases.

How about we require the `VertxBasedHttpProtocolAdapter` auth providers to be `CredentialsApiAuthProvider` instances (instead of the more generic type)? Aren't adapters supposed to use the Credentials API to fetch credentials anyway?

Comment (calohmn) [reply service-base/src/main/java/org/eclipse/hono/service/http/TenantTraceSamplingHandler.java]: In any case, I can put the above logic in a separate class so that a separate implementation may be used in the case that the `VertxBasedHttpProtocolAdapter` instance uses a completely different custom auth provider to extract tenant/auth-id from the `RoutingContext`.

I would do that in a subseqent commit then, along with some refactoring enabling code sharing with the upcoming trace-sampling-mode implementations for the other adapters.

File: adapters/http-vertx/src/main/java/org/eclipse/hono/adapter/http/impl/VertxBasedHttpProtocolAdapter.java [hunk 2]
```diff
@@ -110,22 +113,23 @@ protected void addRoutes(final Router router) {
                     Optional.ofNullable(usernamePasswordAuthProvider).orElse(
                             new UsernamePasswordAuthProvider(getCredentialsClientFactory(), getConfig(), tracer)),
                     getConfig().getRealm(), tracer));
-            addTelemetryApiRoutes(router, authHandler);
-            addEventApiRoutes(router, authHandler);
-            addCommandResponseRoutes(CommandConstants.COMMAND_ENDPOINT, router, authHandler);
-            addCommandResponseRoutes(CommandConstants.COMMAND_LEGACY_ENDPOINT, router, authHandler);
+            addTelemetryApiRoutes(router, tenantTraceSamplingHandler, authHandler);
```
sophokles73: wouldn't it be easier to just add it as a handler that is used for all requests? I guess the overhead is minimal as the handler never fails but just continues with the next handler if no Basic auth header is found in the request, right?

calohmn: Yes, I've amended the commit to use only one generic route handler in the base class.
I've also refactored the new `TenantTraceSamplingHandler` class so that logic can be reused in other protocol adapters, to be done in follow-up commits.

service-base/src/main/java/org/eclipse/hono/service/http/TenantTraceSamplingHandler.java
calohmn: I've amended the commit here with the above mentioned refactoring.

calohmn:
I've amended the commit with the following correction:
When publishing telemetry/event data for an authenticated Gateway, the (gateway) `auth-id` is now taken for retrieving the value from the `trace-sampling-mode-per-auth-id` map - not the "device-id" URL part.

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantObjectWithAuthId.java
```diff
@@ -0,0 +1,44 @@
...
+ *******************************************************************************/
+
+package org.eclipse.hono.service.tenant;
+
+import org.eclipse.hono.util.TenantObject;
+
+/**
+ * Class that keeps a TenantObject and an auth-id.
+ */
+public class TenantObjectWithAuthId {
```
sophokles73: final ?

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantObjectWithAuthId.java
```diff
@@ -0,0 +1,44 @@
...
+    private final String authId;
+
+    /**
+     * Creates a new TenantObjectWithAuthId.
+     *
+     * @param tenantObject The tenant object.
+     * @param authId The identity a device authenticates as. May also be the device id if no authentication is used.
+     */
+    public TenantObjectWithAuthId(final TenantObject tenantObject, final String authId) {
+        this.tenantObject = tenantObject;
```
sophokles73: null checks?

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantObjectWithAuthId.java
```diff
@@ -0,0 +1,44 @@
...
+        this.authId = authId;
+    }
+
+    public TenantObject getTenantObject() {
```
sophokles73: JavaDoc?

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantObjectWithAuthId.java
```diff
@@ -0,0 +1,44 @@
...
+        return tenantObject;
+    }
+
+    public String getAuthId() {
```
sophokles73: JavaDoc?

service-base/src/main/java/org/eclipse/hono/service/tenant/ExecutionContextTenantAndAuthIdProvider.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @param <T> The type of ExecutionContext used.
+ */
+public interface ExecutionContextTenantAndAuthIdProvider<T extends ExecutionContext> {
+
+    /**
+     * Get the tenant and auth-id from the given ExecutionContext.
+     *
+     * @param context The execution context.
+     * @param spanContext The OpenTracing context to use for tracking the operation.
```
sophokles73: is this optional?

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseExecutionContextTenantAndAuthIdProvider.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Creates a new BaseExecutionContextTenantAndAuthIdProvider for the given config and tenantClientFactory.
+     *
+     * @param config The configuration.
+     * @param tenantClientFactory The factory to use for creating a Tenant service client.
+     */
+    public BaseExecutionContextTenantAndAuthIdProvider(final ProtocolAdapterProperties config,
+            final TenantClientFactory tenantClientFactory) {
+        this.config = config;
```
sophokles73: null checks?

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseExecutionContextTenantAndAuthIdProvider.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Gets a {@link TenantObjectWithAuthId} from the X509 certificate of the given {@link SSLSession}.
+     *
+     * @param sslSession The SSL session.
+     * @param spanContext The OpenTracing context to use for tracking the operation.
+     * @return A future indicating the outcome of the operation.
+     */
+    protected Future<TenantObjectWithAuthId> getFromSslSessionCert(final SSLSession sslSession,
+            final SpanContext spanContext) {
```
sophokles73: null checks?

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseExecutionContextTenantAndAuthIdProvider.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * Gets a {@link TenantObjectWithAuthId} from the X509 certificate of the given {@link SSLSession}.
+     *
+     * @param sslSession The SSL session.
+     * @param spanContext The OpenTracing context to use for tracking the operation.
+     * @return A future indicating the outcome of the operation.
+     */
+    protected Future<TenantObjectWithAuthId> getFromSslSessionCert(final SSLSession sslSession,
```
sophokles73: this sounds quite technical. I'd say we are talking about the client cert, i.e. *getFromClientCertificate* ?

service-base/src/main/java/org/eclipse/hono/service/http/TenantTraceSamplingHandler.java
```diff
@@ -0,0 +1,62 @@
...
+public class TenantTraceSamplingHandler implements Handler<RoutingContext> {
+
+    private final ExecutionContextTenantAndAuthIdProvider<HttpContext> tenantObjectWithAuthIdProvider;
+
+    /**
+     * Creates a new Handler for the given config and tenantClientFactory.
+     *
+     * @param tenantObjectWithAuthIdProvider Provides the tenant from the HttpContext.
+     */
+    public TenantTraceSamplingHandler(final ExecutionContextTenantAndAuthIdProvider<HttpContext> tenantObjectWithAuthIdProvider) {
```
sophokles73: null check?

service-base/src/main/java/org/eclipse/hono/service/http/HttpContextTenantAndAuthIdProvider.java
```diff
@@ -0,0 +1,112 @@
...
+     * 
+     * @param config The configuration.
+     * @param tenantClientFactory The factory to use for creating a Tenant service client.
+     * @param tenantIdContextParamName The name of the HttpContext parameter name that provides the tenant id in case of
+     *            an unauthenticated request.
+     * @param deviceIdContextParamName The name of the HttpContext parameter name that provides the device id in case of
+     *            an unauthenticated request. The device id will be used as auth-id for the created
+     *            TenantObjectWithAuthId objects.
+     */
+    public HttpContextTenantAndAuthIdProvider(final ProtocolAdapterProperties config,
```
sophokles73: null checks?

service-base/src/main/java/org/eclipse/hono/service/tenant/ExecutionContextTenantAndAuthIdProvider.java
```diff
@@ -0,0 +1,37 @@
...
+     * @return A future indicating the outcome of the operation.
```
sophokles73: Are there any specific exceptions being used to indicate certain types of problems? Or is it just: if the future is failed then the we could not determine the tenant and auth-id from the request?

service-base/src/main/java/org/eclipse/hono/service/http/HttpContextTenantAndAuthIdProvider.java
```diff
@@ -0,0 +1,112 @@
...
+            return tenantClientFactory.getOrCreateTenantClient()
+                    .compose(tenantClient -> tenantClient.get(tenantId, spanContext))
+                    .map(tenantObject -> new TenantObjectWithAuthId(tenantObject, deviceId));
+        }
+        return Future.failedFuture("tenant could not be determined");
+    }
+
+    private Future<TenantObjectWithAuthId> getTenantViaCert(final RoutingContext ctx, final SpanContext spanContext) {
+        if (!ctx.request().isSSL()) {
+            return Future.failedFuture("no cert found (not SSL/TLS encrypted)");
```
sophokles73: it is quite expensive to create the stack trace associated with the failed future. Since this outcome will be very common, e.g. if all devices use Basic auth, we should probably better return `null` as the value in a succeeded future in this case. Same holds true for Basic auth ...

Another option would be to (re-)use a statically initialized exception instance for creating the failed future ...

WDYT?

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -295,13 +296,30 @@ protected Router createRouter() {
         final Router router = Router.router(vertx);
         LOG.info("limiting size of inbound request body to {} bytes", getConfig().getMaxPayloadSize());
         router.route().handler(BodyHandler.create(DEFAULT_UPLOADS_DIRECTORY).setBodyLimit(getConfig().getMaxPayloadSize()));
+        Optional.ofNullable(getTenantTraceSamplingHandler())
+                .ifPresent(tenantTraceSamplingHandler -> router.route().handler(tenantTraceSamplingHandler)
```
sophokles73: are you trying to insert this handler at a particular position? If so, IMHO you should make this explicit by setting the order ...

service-base/src/main/java/org/eclipse/hono/service/tenant/TenantTraceSamplingHelper.java
```diff
@@ -0,0 +1,54 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.opentracing.Span;
+
+/**
+ * A helper class for applying the tenant specific trace sampling
+ * configuration.
+ */
+public class TenantTraceSamplingHelper {
```
sophokles73: final?

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
calohmn: This handler should be inserted directly after the BodyHandler one line above (otherwise there is an exception when processing requests). Since there is no order set for the BodyHandler, I haven't used one for the TenantTraceSamplingHandler either.

service-base/src/main/java/org/eclipse/hono/service/http/HttpContextTenantAndAuthIdProvider.java
calohmn: `Future.failedFuture(String)` will create a `NoStackTraceThrowable` (where `Throwable.fillInStackTrace()` doesn't get invoked). Therefore I don't think using failed futures here is more expensive than using succeeded futures.

service-base/src/main/java/org/eclipse/hono/service/tenant/ExecutionContextTenantAndAuthIdProvider.java
calohmn: I've changed the javadoc:

> The future will fail if tenant and auth-id information could not be retrieved from the ExecutionContext
> or if there was an error obtaining the tenant object. In the latter case the future will be failed with a
> ServiceInvocationException.
> 
> Otherwise the future will contain the created <em>TenantObjectWithAuthId</em>.

service-base/src/main/java/org/eclipse/hono/service/tenant/ExecutionContextTenantAndAuthIdProvider.java
calohmn: Yes, I've changed the javadoc accordingly.

calohmn:
I've made the changes according to the comments.

service-base/src/main/java/org/eclipse/hono/service/http/HttpContextTenantAndAuthIdProvider.java
sophokles73: Didn't know that `Future.failedFuture(String)` doesn't create a stack trace :+1:

service-base/src/test/java/org/eclipse/hono/service/http/TenantTraceSamplingHandlerTest.java
```diff
@@ -0,0 +1,547 @@
...
+        }).when(tenantClient).get(eq(tenantId), any(SpanContext.class));
+
+        // WHEN handling a request for that tenant
+        final String authorization = "BASIC " +
+                Base64.getEncoder().encodeToString((authId + "@" + tenantId + ":password").getBytes(StandardCharsets.UTF_8));
+        final MultiMap headers = mock(MultiMap.class);
+        when(headers.get(eq(HttpHeaders.AUTHORIZATION))).thenReturn(authorization);
+        final HttpServerRequest req = mock(HttpServerRequest.class);
+        when(req.headers()).thenReturn(headers);
+        when(ctx.request()).thenReturn(req);
```
sophokles73: there seems to be a lot of redundant code with setting up the fixture among the test methods. Maybe you can factor out some of the mock creation code into a helper method?

calohmn: Right, I've extracted some methods there.
(Also made a small change in `TenantTraceSamplingHelper` in preparation for support in other adapters.)

service-base/src/test/java/org/eclipse/hono/service/http/TenantTraceSamplingHandlerTest.java
```diff
@@ -0,0 +1,538 @@
...
+        final String authId = "testAuthId";
+
+        final TenantObject testTenant = TenantObject.from(tenantId, true);
+        testTenant.setProperty(TenantConstants.FIELD_TRACE_SAMPLING_MODE, TraceSamplingMode.ALL.getFieldValue());
+        final JsonObject samplingModePerAuthId = new JsonObject();
+        samplingModePerAuthId.put(authId, TraceSamplingMode.DEFAULT.getFieldValue());
+        testTenant.setProperty(TenantConstants.FIELD_TRACE_SAMPLING_MODE_PER_AUTH_ID, samplingModePerAuthId);
+        doAnswer(invocation -> {
+            return Future.succeededFuture(testTenant);
+        }).when(tenantClient).get(eq(tenantId), any(SpanContext.class));
```
sophokles73: just a hint: this can be written like
```
when(tenantClient.get(eq(tenantId), any(SpanContext.class))).thenReturn(Future.succeededFuture(testTenant));
```

which is easier to read (and write) ;-)

sophokles73:
@calohmn what about the other adapters? can you create an issue for them so that we do not forget to add the functionality?

calohmn:
@sophokles73 I've added individual issues for implementing this in the remaining protocol adapters.

Issue #1377, opened 2019-07-18
Support 'pwd-plain' when updating credentials


calohmn:
Add support for using `pwd-plain` in a credentials secret when using the new device registry management API to update a credential.

PR #1387, opened 2019-07-19
[#1377] Support pwd-plain in the new device management API

dejanb:


service-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordSecret.java
```diff
@@ -12,13 +12,16 @@
  *******************************************************************************/
 package org.eclipse.hono.service.management.credentials;
 
+import io.vertx.core.json.JsonObject;
```
jbtrystram: import order

service-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordSecret.java
```diff
@@ -12,13 +12,16 @@
...
 import static org.eclipse.hono.util.CredentialsConstants.FIELD_SECRETS_HASH_FUNCTION;
 import static org.eclipse.hono.util.CredentialsConstants.FIELD_SECRETS_PWD_HASH;
+import static org.eclipse.hono.util.CredentialsConstants.FIELD_SECRETS_PWD_PLAIN;
```
sophokles73: all these constants should be imported from `RegistryManagementConstants` instead of `CredentialsConstants`

service-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordSecret.java
```diff
@@ -68,6 +81,9 @@ protected ToStringHelper toStringHelper() {
     @Override
     public void checkValidity() {
         super.checkValidity();
+        if (passwordPlain != null && !passwordPlain.isEmpty()) {
```
jbtrystram: You can use `! Strings.isNullOrEmpty(passwordPlain)` as well :)

sophokles73: `org.eclipse.hono.util.Strings.isNullOrEmpty()`?

service-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordSecret.java
```diff
@@ -76,4 +92,18 @@ public void checkValidity() {
         }
     }
 
+    /**
+     * Encodes the value in {@link #passwordPlain} using provided encoder.
+     * @param encoder Password encoder to use.
+     */
+    public void encode(final HonoPasswordEncoder encoder) {
+        if (passwordPlain != null && !passwordPlain.isEmpty()) {
```
sophokles73: `org.eclipse.hono.util.Strings.isNullOrEmpty()`?

jbtrystram: `! Strings.isNullOrEmpty(passwordPlain)` again ?

site/documentation/static/api/device-registry-v1.yaml
```diff
@@ -635,6 +635,10 @@ components:
                  "salt":
                     type: string
                     format: byte
+                 "pwd-plain":
```
sophokles73: what about putting this into a separate type, e.g. `PlaintextPasswordSecret`, which doesn't have all the other properties? That would make it totally clear that a client cannot choose the hash-function etc.

dejanb: @sophokles73 I think that would spiral up in complexity (I wanted to keep it simple).

In that case we would need to have either:
1. `PlainPasswordCredential` type as well
2. or come up with a common type for "password" secrets to be able to use them both in `PasswordCredential` interchangeably. 

In order to ensure that plain password is never used in `get` operations we would need to introduce the new CredentialsSet (based on solution 1.) that would distinct if `PlainPasswordCredential` is allowed or not.

Should we go with all this or add more documentation on what are the expectation from PasswordSecret? I'm OK with both approaches in general.

dejanb: Thinking more about this, I don't think it makes much sense for management API to ever hand-over any secret specifics, like hash-functions, secrets, etc. Maybe `get` should just return `CommonSecret` data? @ctron @sophokles73 WDYT?

sophokles73: Well, the use case for the *GET* operation is to retrieve the data and (potentially) edit/change it and then write it back, isn't it? For e.g. PSK I think it makes a lot of sense to get the secret in the response. You might also want to adjust/set the validity period of secrets ...

sophokles73: > Should we go with all this or add more documentation on what are the expectation from PasswordSecret?

I'll leave that for you to decide. I just wanted to mention the possibility ...

dejanb: `CommonSecret` data will allow you to edit other fields, like validity period, enabled/disabled and comment. 
Maybe for pre-shared keys it makes sense, but I don't see any value in handing over password hashes other than introducing a security risk. 
These types of APIs usually don't allow you to edit passwords, just reseting them.

I need to let this brew more over the weekend :)

site/documentation/static/api/device-registry-v1.yaml
```diff
@@ -351,6 +355,11 @@ paths:
                   application/json:
                      schema:
                         $ref: '#/components/schemas/CredentialsSet'
+                     examples:
+                        Full Credentials:
+                           $ref: '#/components/examples/HahsedPasswordExample'
```
ctron: Typo: `HahsedPasswordExample`

site/documentation/static/api/device-registry-v1.yaml
```diff
@@ -375,6 +387,11 @@ paths:
                application/json:
                   schema:
                      $ref: '#/components/schemas/CredentialsSet'
+                  examples:
+                     Hashed Password:
+                        $ref: '#/components/examples/HahsedPasswordExample'
```
ctron: Same typo.

site/documentation/static/api/device-registry-v1.yaml
```diff
@@ -629,16 +650,28 @@ components:
                  "hash-function":
                     type: string
                     example: sha-512
+                    description: The name of the hash function used to create the password hash (defined in 'pwd-hash' property).
+                                 If password is defined using 'pwd-plain' property, this value will be ignored by the device registry.
+                                 This property should be empty when returning passwords from the device registry.
```
ctron: I would prefer pointing out the two choices here, like:

This property may be empty on read results, if the device registry conceals authentication secrets. In this case the `id` field must be set instead.

site/documentation/static/api/device-registry-v1.yaml
```diff
@@ -629,12 +650,28 @@ components:
...
                     type: string
                     format: byte
+                    description: The Base64 encoding of the salt used in the password hash (defined in 'pwd-hash' property).
+                                 If password is defined using 'pwd-plain' property, this value will be ignored by the device registry.
+                                 This property should be empty when returning passwords from the device registry.
+                 "pwd-plain":
+                    type: string
+                    format: byte
+                    description: The clear text value of the password to be hashed by the device registry.
+                                 If this property is specified, the registry will ignore user-provided hash properties ('hash-function', 'pwd-hash' and 'salt')
```
ctron:
```suggestion
                                 If this property is specified, the registry will ignore user-provided hash properties ('hash-function', 'pwd-hash' and 'salt').
```

site/documentation/static/api/device-registry-v1.yaml
```diff
@@ -629,12 +650,28 @@ components:
...
+                                 This property should be never stored by the device registry
```
ctron:
```suggestion
                                 This property should be never stored by the device registry.
```

site/documentation/static/api/device-registry-v1.yaml
```diff
@@ -792,3 +829,36 @@ components:
       BasicAuth:
          type: http
          scheme: basic
+
+   examples:
+      HahsedPasswordExample:
```
ctron: Same typo :grin:

site/documentation/static/api/device-registry-v1.yaml
dejanb: Yes, I forgot to update it. Thanks!

site/documentation/static/api/device-registry-v1.yaml
dejanb: @sophokles73 I ended up leaving up the same type as we had before. I think it will work better in the get/update workflow along with the other changes introduced. OpenApi doesn't define parameters dependencies yet, but it will in the future (https://swagger.io/docs/specification/describing-parameters/), which will allow us to define this better. For now we're following their recommendations and describing these dependencies.

As for the discussion on should we return secret details or not, we introduced the possibility for implementations to support both. If registry works with full secret details everything will work as it was working so far.

If registry don't want to expose secret details, it should generate an "id" for created secrets. So, get operation will return only this id and metadata for the secret. The update will merge data, so it can delete/update existing secrets or create new ones.

Let me know how this sounds to you.

site/documentation/static/api/device-registry-v1.yaml
```diff
@@ -629,12 +650,31 @@ components:
...
+                    description: The Base64 encoding of the salt used in the password hash (defined in 'pwd-hash' property).
+                                 If password is defined using 'pwd-plain' property, this value will be ignored by the device registry.
+                                 This property should be empty when returning passwords from the device registry using only secret metadata.
+                                 In this case the id field must be set instead.
+                 "pwd-plain":
+                    type: string
+                    format: byte
+                    description: The clear text value of the password to be hashed by the device registry.
+                                 If this property is specified, the registry will ignore user-provided hash properties ('hash-function', 'pwd-hash' and 'salt').
+                                 This property should be never stored by the device registry.
```
calohmn: `should never be stored`

dejanb:
@sophokles73 from my point of view this is ready to go. Let me know if you have any other thoughts around it.

site/documentation/static/api/device-registry-v1.yaml
```diff
@@ -629,12 +650,31 @@ components:
                  "hash-function":
                     type: string
                     example: sha-512
+                    description: The name of the hash function used to create the password hash (defined in 'pwd-hash' property).
+                                 If password is defined using 'pwd-plain' property, this value will be ignored by the device registry.
```
calohmn: I think it sounds better with added articles:
````
(defined in the 'pwd-hash' property).
If the password is defined using a 'pwd-plain' property, this value will be ignored by the device registry.
````
Same in the two description blocks below.

**Comment by dejanb** [comment]
@calohmn Thanks for the edits. I fixed them and did some more minor ones. Give it another read.

File: service-base/src/main/java/org/eclipse/hono/service/management/credentials/CommonCredential.java [hunk 14]
```diff
@@ -108,6 +108,5 @@ public void checkValidity() {
         if (this.authId == null || this.authId.isEmpty()) {
             throw new IllegalStateException("missing auth ID");
         }
-        getSecrets().forEach(CommonSecret::checkValidity);
     }
```
calohmn: Javadoc needs to be updated here (removing the `This also calls {@link CommonSecret#checkValidity()} for all secrets.` part).

dejanb: fixed

PR #1390, opened 2019-07-23
[#1389] Implement trace sampling mode option in MQTT adapter

### LINKED ISSUE #1389: Enable forced sampling of traces for a specific tenant (MQTT adapter)
The new config options introduced in #896 to enable forced sampling of traces (i.e. recording of traces in the tracing backend) for a specific tenant or auth-id should also be taken into account in the MQTT adapter.

calohmn:
This fixes  #1389.

The Tenant JSON config options described in #1382 are used to set a trace sampling priority in the MQTT adapter.

(Feature is also implemented in the Kura protocol adapter, using the `AbstractVertxBasedMqttProtocolAdapter` base class.)

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
@@ -49,6 +49,8 @@ import org.eclipse.hono.service.metric.MetricsTags;
 import org.eclipse.hono.service.metric.MetricsTags.Direction;
 import org.eclipse.hono.service.metric.MetricsTags.EndpointType;
 import org.eclipse.hono.service.metric.MetricsTags.ProcessingOutcome;
+import org.eclipse.hono.service.tenant.ExecutionContextTenantAndAuthIdProvider;
+import org.eclipse.hono.service.tenant.TenantTraceSamplingHelper;
 import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.CommandConstants;
 import org.eclipse.hono.util.Constants;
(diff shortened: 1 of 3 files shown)


adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -435,8 +465,11 @@ private void handleConnectionRequestResult(final MqttEndpoint endpoint,
             final Span currentSpan) {
 
         final MqttContext context = MqttContext.fromConnectPacket(endpoint);
-        context.setTracingContext(currentSpan.context());
-        final Future<DeviceUser> authAttempt = authenticate(context, currentSpan);
+        final Future<Optional<Integer>> traceSamplingPrio = applyTenantTraceSamplingPriority(context, currentSpan);
```
ctron: Why not use `OptionalInt` instead?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -446,7 +479,7 @@ private void handleConnectionRequestResult(final MqttEndpoint endpoint,
                         checkDeviceRegistration(authenticatedDevice, currentSpan.context()))
                         .map(ok -> authenticatedDevice))
                 .compose(authenticatedDevice -> createLinks(authenticatedDevice, currentSpan))
-                .compose(authenticatedDevice -> registerHandlers(endpoint, authenticatedDevice))
+                .compose(authenticatedDevice -> registerHandlers(endpoint, authenticatedDevice, traceSamplingPrio.result()))
```
ctron: I would prefer `priority` over `prio`.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
calohmn: Good point. As this change also involves other classes, I would change that in a subsequent commit

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
calohmn: I've renamed that accordingly in the class.

PR #1391, opened 2019-07-23
Improve the selection of documentation versions.

b-abel:
Versions of the documentation are now mapped on languages for Hugo,
allowing to leverage the multilingual features of Hugo. The current
stable version is a level above the other versions, improving the
search engine ranking of this version. Version "stable" has the URL
path "/hono/docs/", nightly master has "/hono/docs/dev/" and e.g.
version 1.2.3 would be under "/hono/docs/1.2/". Having a version marked
as stable is required for this to work.
The version is now selected in the documentation itself and no longer
on the homepage.


b-abel:
scrrenshot03

sophokles73:
I like the approach a lot. WDYT @dejanb @calohmn @ctron

b-abel:
Advantages are, that you can now easily change the version while staying on the same page, it is always displayed, which version you are reading. Before there was no possibility to change the version inside the documentation, you always needed to go back to the homepage (opened in a new tab) and then select another version (which again opened in a new tab).
The whole build process is a bit easier and resources like JS files or images are only stored once and not once per version.

calohmn:
I like it as well. :+1:

PR #1403, opened 2019-07-24
[#623] Add support for multiple trust configurations for a tenant

Alfusainey:
This PR addresses #623 by adding support for multiple trust configurations for the management API. 


sophokles73:
The example device registry already supports time offset in date/time properties. No need to wait for #1374

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorFactory.java
```diff
@@ -294,7 +294,7 @@ private void verifyExternal(final Handler<AsyncResult<DeviceUser>> completer) {
                 tenantTracker
                         .compose(tenant -> {
                             try {
-                                final TrustAnchor trustAnchor = tenantTracker.result().getTrustAnchor();
+                                final TrustAnchor trustAnchor = tenantTracker.result().getTrustAnchors().get(0);
```
sophokles73: I thought the idea is to use (potentially) multiple trust anchors for validating the client certificate, isn't it?

Alfusainey: the idea was to leave out this part for another PR since this one is already too big to review. do you want me to add that in this PR?

site/documentation/content/api/Tenant-API.md
```diff
@@ -154,17 +154,37 @@ The JSON structure below contains example information for tenant `TEST_TENANT`.
...
+| *public-key*             | *yes*       | *string*     |               | The Base64 encoded binary DER encoding of the trusted root certificate's public key. |
+| *not-before*             | *no*       | *string*      |               | The property indicating that the trusted root X.509 certificate is not valid before the given date. The value MUST be an **ISO 8601 compliant** [*combined date and time representation*](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations) in extended format. |
+| *not-after*              | *no*       | *string*      |               | The property indicating that the trusted root X.509 certificate is not valid after the given date. The value MUST be an **ISO 8601 compliant** [*combined date and time representation*](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations) in extended format.|
```
sophokles73: why do you want to add these properties again? The client (adapter) should *only* ever receive currently valid trust anchors ...

site/documentation/static/api/device-registry-v1.yaml
```diff
@@ -481,12 +484,30 @@ components:
...
                type: string
                format: byte
                description: The Base64 encoded binary DER encoding of the
                             trusted root certificate. Either this property
                             or `public-key` must be set.
+
+            "not-before":
+               type: string
+               format: date-time
+               example: 2015-01-01T00:00:00+0000
```
sophokles73: you can come up with a better example, can't you?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorFactory.java
Alfusainey: yes, exactly

service-base/src/main/java/org/eclipse/hono/service/management/tenant/Tenant.java
```diff
@@ -114,12 +115,12 @@ public void setLimits(final ResourceLimits limits) {
...
-    public TrustedCertificateAuthority getTrustedCertificateAuthority() {
-        return trustedCertificateAuthority;
+    public List<TrustedCertificateAuthority> getTrustedAuthorities() {
+        return trustedAuthorities;
     }
 
-    public void setTrustedCertificateAuthority(final TrustedCertificateAuthority trustedCertificateAuthority) {
-        this.trustedCertificateAuthority = trustedCertificateAuthority;
+    public void setTrustedAuthorities(final List<TrustedCertificateAuthority> trustedAuthorities) {
+        this.trustedAuthorities = trustedAuthorities;
```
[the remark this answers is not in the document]
Alfusainey: > null check?

At the class-level, the `Include.NON_NULL` annotation is defined for all properties. so it the value is null, it won't be included

site/documentation/content/api/Tenant-API.md
Alfusainey: i forgot to remove those

Alfusainey:
@sophokles73 patch updated

Alfusainey:
@sophokles73 can you take a look at this one too?

sophokles73:
@Alfusainey I'd rather postpone this until we have either merged or closed #1374 ...

Alfusainey:
> @Alfusainey I'd rather postpone this until we have either merged or closed #1374 ...

ok

sophokles73:
@Alfusainey ok, #1374 has been merged so you can apply the same pattern to the other value objects that deal with instants in time ...

Alfusainey:
> @Alfusainey ok, #1374 has been merged so you can apply the same pattern to the other value objects that deal with instants in time ...

sure, actually the only thing i need to do is to remove the added dependency since its now added in #1374.

i updated the PR to remove the dependency

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorFactory.java
```diff
@@ -334,7 +334,7 @@ public boolean succeeded() {
                         try {
                             return certValidator.validate(
                                     Collections.singletonList(deviceCert),
-                                    tenant.getTrustAnchor()).map(ok -> tenant);
+                                    tenant.getTrustAnchors().get(0)).map(ok -> tenant);
```
sophokles73: I would rather not commit this change as it seems kind of arbitrary, which certificate will be used as the trust anchor. Instead, I think that you should first extend the `X509CertificateChainValidator` interface with a method that accepts a list of TrustAnchors instead of just a single TrustAnchor. Please put this change into a separate commit to make it easier to see which changes are related to each other. Then we should use all the trust anchors contained in the `TenantObject` when performing the validation here.

Alfusainey: ok, i will do that 👍

Alfusainey: @sophokles73 patch updated

service-base/src/main/java/org/eclipse/hono/service/auth/X509CertificateChainValidator.java
```diff
@@ -28,15 +27,16 @@
...
-     * @throws IllegalArgumentException if the chain is empty.
+     * @param trustAnchors The list of trust anchors to use for validating the chain.
+     * @return A composite future, succeeded when the certificate path is valid for any the trust anchors.
+     *         Otherwise, the composite future is failed when all the trust anchors cannot be validated by the certificate path.
+     *
+     * @throws NullPointerException if any of the paramerters are {@code null}.
+     * @throws IllegalArgumentException if the chain and/or trust anchors is empty.
      */
-    Future<Void> validate(List<X509Certificate> chain, TrustAnchor trustAnchor);
+    CompositeFuture validate(List<X509Certificate> chain, List<TrustAnchor> trustAnchors);
```
sophokles73: Well, I was actually thinking about simply adding another variant of this method that accepts a List of anchors ...

Alfusainey: since one can create a list containing a single trust anchor, don't you think that this is just a general case of the current call: e.g `validate(chain, List.of(trustAnchor))`?

sophokles73: That is true but keeping the existing method does not break any client code and also is a little more convenient. I would also rather have a `Future<Void>` as the return value instead of a `CompositeFuture`.

Alfusainey: @sophokles73 done

service-base/src/main/java/org/eclipse/hono/service/auth/X509CertificateChainValidator.java
```diff
@@ -39,4 +41,44 @@
...
+     * @param chain The certificate chain to validate. The end certificate
+     *              must be at position 0.
+     * @param trustAnchors The list of trust anchors to use for validating the chain.
+     * @return A completed future if the certificate path is valid for any of the trust anchors.
+     *         Otherwise, the future will be failed when all the trust anchors cannot be validated by the certificate path.
+     *
+     * @throws NullPointerException if any of the parameters are {@code null}.
+     * @throws IllegalArgumentException if the chain and/or trust anchors is empty.
+     */
+    default Future<Void> validate(List<X509Certificate> chain, List<TrustAnchor> trustAnchors) {
```
sophokles73: I do not think that we need to provide a default implementation here. Just defining the method will do. In fact, most implementations will probably delegate to this variant as indicated earlier by you, instead of the other way around ...

Alfusainey: > I do not think that we need to provide a default implementation here. Just defining the method will do.

ok and this will mean that the `X509CertificateChainValidator` can no longer be a functional interface -> i updated the patch

service-base/src/main/java/org/eclipse/hono/service/auth/X509CertificateChainValidator.java
```diff
@@ -39,4 +38,18 @@
...
      */
     Future<Void> validate(List<X509Certificate> chain, TrustAnchor trustAnchor);
+
+    /**
+     * Validates a certificate path based on a list of trust anchors.
+     * 
+     * @param chain The certificate chain to validate. The end certificate
+     *              must be at position 0.
+     * @param trustAnchors The list of trust anchors to use for validating the chain.
+     * @return A completed future if the certificate path is valid for any of the trust anchors.
```
sophokles73: Can we please use the same text as that of the other method variant? There is no reason why this method should behave differently, is there?

service-base/src/main/java/org/eclipse/hono/service/auth/device/DeviceCertificateValidator.java
```diff
@@ -45,14 +47,40 @@
...
+            if (validationCheck.succeeded()) {
+                result.complete();
+            } else {
+                result.fail(validationCheck.cause());
+            }
+        });
+        return result;
+    }
+
+    private Future<Void> doValidate(final List<X509Certificate> chain, final TrustAnchor trustAnchor) {
```
sophokles73: FMPOV there is no need to iteratively invoke this method for each trust anchor.
Instead, simply pass in all trust anchors to the constructor for the `PKIXParameters` ...

Alfusainey: @sophokles73 if this looks good, i will move on to adding the ITs for the tenant service

sophokles73: Yes, this looks good :+1: 
On to the next step ...

Alfusainey:
@sophokles73 Patch updated. 

I couldn't find a way to generate and signed custom certificates using the Java security API. OpenJDK has an [open issue](https://bugs.openjdk.java.net/browse/JDK-8058778) to provide such an API but this issue is still not resolved.

For this reason, I decided to rely on the third party **bouncy castle** library to generate and signed certificates programatically. The reason why I want to do this is to have control over certificate generation, signing and validity period during integration testing.

What I wanted in a nutshell ist:

1. Create certificates with custom validity (notBefore, notAfter) periods.
2. Create and sign certificates with a given CA certificate

When I use `io.vertx.core.net.SelfSignedCertificate`, I cannot achieve either of the above. With `io.netty.handler.ssl.util.SelfSignedCertificate`, I can only achieve (1) but not (2). So in essence, I don't only need a self-signed certificate but one that can also be signed and verified by another certificate.

This is the reason why I introduced the new `CustomSignedCertificate` class and added a test dependency to the Bouncy castle library (for testing purposes).

Some more ITs will follow

tests/src/test/java/org/eclipse/hono/tests/registry/TenantHttpIT.java
```diff
@@ -214,6 +240,12 @@ public void testUpdateTenantFailsForNonExistingTenant(final VertxTestContext con
     public void testRemoveTenantSucceeds(final VertxTestContext context) {
 
         final Tenant tenantPayload = buildTenantPayload();
+        final var trustedCa = new TrustedCertificateAuthority()
```
sophokles73: what is this good for?

tests/src/test/java/org/eclipse/hono/tests/registry/TenantHttpIT.java
```diff
@@ -268,8 +300,26 @@ public void testGetAddedTenant(final VertxTestContext context)  {
...
 
+    /**
+     * Verifies that the service returns a 400 status code for an add tenant request with a <em>trusted-ca</em> configuration
+     * containing a JSON object as a value.
+     * 
+     * @param context The Vert.x test context.
+     */
+    @Test
+    public void testAddTenantFailsWhenOldTrustModelIsUsed(final VertxTestContext context) {
+        final JsonObject trustConfig = new JsonObject()
```
sophokles73: I assume the test is for the registry to check if the trusted CA spec is an array of objects and not just an object. If so, then we can simply use an empty JsonObject here, can't we?

Alfusainey:
> I am not convinced that we need to introduce a dependency on Bouncy Castle here. based on the checks that we need to do, I believe that we can simply get away with the already existing certificates that we have in the demo-certs folder. Which checks do you think would require a dependency on Bouncy Castle?

I want to perform the following checks:
1. Generate two trusted CA certificates (one expired certificate and one valid certificate).
2. Generate two client certificates, one signed by the expired trusted CA and the other signed by the valid certificate.

Then I want to have two test cases that:
1. Verifies that a "protocol adapter" will fail to authenticate a device belonging to a tenant which is configured with an expired certificate. In this test case, the tenant will contain a single trusted CA config containing the expired certificate
2. Verifies that a protocol adapter will successfully authenticate a device belonging to a tenant configured with **at least** one valid trust anchor. Here, I will configure the tenant to contain both the expired and valid trust certificates

Yes, I agree that I can either use the existing certificates or generate new ones on the command-line but I was thinking of doing this in the code itself.

Do you want me to remove the dependency and generate new ones for my use-case?

tests/src/test/java/org/eclipse/hono/tests/registry/TenantHttpIT.java
Alfusainey: sure 👍

tests/src/test/java/org/eclipse/hono/tests/registry/TenantHttpIT.java
Alfusainey: nothing really, just added an additional config for the tenant payload

sophokles73:
> Verifies that a "protocol adapter" will fail to authenticate a device belonging to a tenant which is configured with an expired certificate. In this test case, the tenant will contain a single trusted CA config containing the expired certificate

For this, you do not even need a certificate because the Tenant service would not include the expired certificate's public key in its response to the get Tenant call from the protocol adapter, i.e. it is sufficient to register a Tenant with a public key whose *not after* instant is in the past ...

sophokles73:
> Verifies that a protocol adapter will successfully authenticate a device belonging to a tenant configured with at least one valid trust anchor. Here, I will configure the tenant to contain both the expired and valid trust certificates

A device's client cert can be signed by a single root CA only. It would therefore be sufficient to have the Tenant registered with the valid root CA's public key that can be used to verify the client cert's signature and then any other public key that is no longer valid. The latter one will not even be returned to the adapter by the Tenant service. Therefore, it would probably make more sense to register another public key with the tenant that is still valid but which cannot be used to verify the device's client cert's signature, because that public key will at least be returned to the adapter as part of the trust anchor list ...

All in all, it looks that we only need one client cert and the root ca that has been used for signing it. And we already have that ...

Alfusainey:
> > Verifies that a "protocol adapter" will fail to authenticate a device belonging to a tenant which is configured with an expired certificate. In this test case, the tenant will contain a single trusted CA config containing the expired certificate
> 
> For this, you do not even need a certificate because the Tenant service would not include the expired certificate's public key in its response to the get Tenant call from the protocol adapter, i.e. it is sufficient to register a Tenant with a public key whose _not after_ instant is in the past ...

Yes you're right! in fact this means that a call to `TenantObject#getTrustAnchors()` will return `null` in this case and the certificate path validation will fail during device authentication. It is this exact code path that I want to test i.e the case when `getTrustAnchors()` returns `null` because the root CA expired.

But I see your point and remove this test case.

Alfusainey:
> Therefore, it would probably make more sense to register another public key with the tenant that is still valid but which cannot be used to verify the device's client cert's signature, because that public key will at least be returned to the adapter as part of the trust anchor list ...

yes, I see your point! I was just curious to know how would the `current` implementation handle authentication for devices belonging to a tenant whose root trusted CA expired. 

> All in all, it looks that we only need one client cert and the root ca that has been used for signing it. And we already have that ...

ok 👍

Alfusainey:
@sophokles73 I updated the PR.

I think it is not strictly required to use any of the generated demo certs we have available to IT test this behaviour

> Therefore, it would probably make more sense to register another public key with the tenant that is still valid

to achieve that, I did the ff:
1. Generate a self-signed certificate for a device
2. register this certificate as trust anchor for the tenant the device belongs to
3. Generate a new `key pair` and register the public key of the pair as another trust anchor for the tenant the device belongs to. (now the tenant has two valid trust anchors)
4. authenticate to the adapter using the device's self-signed certificate

In this scenario, the device should successfully authenticate.

The other scenario is to generate two `key pairs` and register both public keys as trust anchors for tenant. Then generate a self-signed certificate for a device belonging to this tenant and authenticate the device using this self-sign certificate. Since the self-sign cert is not registered as trust anchor for the tenant, the authentication should fail

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -117,112 +131,90 @@ public TenantObject setEnabled(final boolean flag) {
...
+    public TenantObject addTrustAnchor(final PublicKey publicKey, final X500Principal subjectDn,
+            final Instant notBefore, final Instant notAfter) {
 
         Objects.requireNonNull(publicKey);
         Objects.requireNonNull(subjectDn);
 
         final JsonObject trustedCa = new JsonObject();
         trustedCa.put(TenantConstants.FIELD_PAYLOAD_SUBJECT_DN, subjectDn.getName(X500Principal.RFC2253));
         trustedCa.put(TenantConstants.FIELD_PAYLOAD_PUBLIC_KEY, publicKey.getEncoded());
+        trustedCa.put(TenantConstants.FIELD_PAYLOAD_NOT_BEFORE, notBefore.toString());
```
sophokles73: I would rather use *toEpochMillis()* here which uses less memory and also does not require special formatting.
BTW using a class' *toString()* method should not generally be expected to yield a specific defined format which you can rely on containing certain information at specific places. Instant.toString() states that it uses an ISO format, but this might easily change in the next Java version ...

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -117,112 +131,90 @@ public TenantObject setEnabled(final boolean flag) {
...
         trustedCa.put(TenantConstants.FIELD_PAYLOAD_KEY_ALGORITHM, publicKey.getAlgorithm());
-        return setProperty(TenantConstants.FIELD_PAYLOAD_TRUSTED_CA, trustedCa);
+
+        return addTrustedCA(trustedCa);
     }
 
     /**
-     * Sets the trusted certificate authority to use for authenticating
-     * devices of this tenant.
+     * Sets the trusted certificate authority to use for authenticating devices of this tenant. The certificate must
```
sophokles73: Actually, the method names indicates that the certificate will be added only ...

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -117,112 +131,90 @@ public TenantObject setEnabled(final boolean flag) {
...
-                } catch (final IllegalArgumentException e) {
-                    // Base64 decoding failed
-                    throw new CertificateParsingException("cert property does not contain valid Base64 scheme", e);
-                }
-            } else {
-                return null;
+            try {
+                final JsonObject trustedCa = new JsonObject()
+                        .put(TenantConstants.FIELD_PAYLOAD_SUBJECT_DN, subjectDn)
+                        .put(TenantConstants.FIELD_PAYLOAD_CERT, certificate.getEncoded());
```
sophokles73: IMHO we should *always* only store the key, subject DN and validity period, and *not* the certificate itself. Accepting a certificate to add to the trust store is just a convenience but the internal data structure should always be the same. This will spare us many checks when processing a TenantObject instance ...

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -240,24 +232,27 @@ public X509Certificate getTrustedCertificateAuthority() throws CertificateExcept
      * @return The trust anchor or {@code null} if no trusted certificate authority
```
sophokles73: It is usually more convenient to return an empty list instead of `null`. This makes code using this method easier to implement because there is no need for null checks on the returned result

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -55,10 +63,16 @@
...
     @JsonIgnore
     private List<Map<String, Object>> adapterConfigurations;
+
+    @JsonIgnore
+    private List<TrustAnchor> trustAnchors;
+
     @JsonIgnore
     private ResourceLimits resourceLimits;
     @JsonIgnore
     private TrustAnchor trustAnchor;
```
sophokles73: what is this good for?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -117,112 +131,90 @@ public TenantObject setEnabled(final boolean flag) {
...
 
     /**
-     * Gets the subject DN of this tenant's configured trusted
-     * certificate authority.
+     * Get the subject DN of the configured trusted CA.
      * 
      * @return The DN or {@code null} if no CA has been set.
      */
     @JsonIgnore
+    @Deprecated
```
sophokles73: FMPOV we should simply remove this method

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -240,24 +232,27 @@ public X509Certificate getTrustedCertificateAuthority() throws CertificateExcept
...
+     *              key type, malformed certificate or public key encoding etc.
      */
     @JsonIgnore
-    public TrustAnchor getTrustAnchor() throws GeneralSecurityException {
-
-        if (trustAnchor != null) {
-            return trustAnchor;
+    public List<TrustAnchor> getTrustAnchors() throws GeneralSecurityException {
+        if (trustAnchors != null) {
+            return trustAnchors;
```
sophokles73: this means that if I add another trust anchor after having invoked this method, the newly added trust anchor will not be contained in the returned result, right?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -641,4 +642,168 @@ public OptionalInt getTraceSamplingPriority(final String authId) {
...
+                subjectNames.add(new X500Principal(subjectDn));
+            });
+            return subjectNames;
+        }
+    }
+
+    /**
+     * Get the list of configured trusted CAs for this tenant.
+     * 
+     * @return The list of configured trusted CAs or {@code null} if this tenant has no configured trusted CA.
```
sophokles73: should return an empty list instead of `null` FMPOV

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -641,4 +642,168 @@ public OptionalInt getTraceSamplingPriority(final String authId) {
...
+    }
+
+    /**
+     * Add a trusted CA configuration for this tenant.
+     * 
+     * @param trustedCa The trust configuration to add.
+     * 
+     * @return This tenant so the API can be used fluently.
+     */
+    public TenantObject addTrustedCA(final JsonObject trustedCa) {
```
sophokles73: is this really supposed to be public? Who else needs this method?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: I might have mistakenly left it there after a rebase to resolve conflicts. will remove

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: sure, its also not used anyway

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -117,112 +131,90 @@ public TenantObject setEnabled(final boolean flag) {
...
      * 
-     * @param publicKey The CA's public key.
-     * @param subjectDn The CA's subject DN.
+     * @param publicKey The trusted CA's public key.
+     * @param subjectDn The trusted CA's subject DN.
+     * @param notBefore The start date in which the trustedCa becomes valid.
+     * @param notAfter The end date in which the trustedCa becomes invalid.
+     * 
      * @return This tenant for command chaining.
      * @throws NullPointerException if any of the parameters is {@code null}.
```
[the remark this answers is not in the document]
Alfusainey: I think we already discussed that the validity period is required for public keys of trusted CAs?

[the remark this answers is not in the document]
Alfusainey: I see your point :-)

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: can you please clarify? I don't understand

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: true 👍

core/src/main/java/org/eclipse/hono/util/TenantObject.java
sophokles73: the method is called *addTrustedCa* but the docs state that the method *sets* the CA ...

Alfusainey: I see 👍

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: yes.

> if I add another trust anchor after having invoked this method

but you usually will not do this right? this method is called when the tenant data stored in the registry has been retrieved via the TenantService

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: good catch

core/src/main/java/org/eclipse/hono/util/TenantObject.java
sophokles73: that is one way of using this class, yes. However, since it is part of Hono's public API, we should not make any assumptions regarding how it will be used. Can you please change it so that the result returned by this method is will consistently return the current trust anchors?

Alfusainey: > Can you please change it so that the result returned by this method is will consistently return the current trust anchors?

yes, I will

Alfusainey:
@sophokles73 pr updated!

I have reserve the whole day today to complete all work for #623 :-)

Alfusainey:
please trigger the build again. its fails saying:

```json
[ERROR]     Unknown packaging: bundle @ org.eclipse.hono:hono-core:[unknown-version], /home/travis/build/eclipse/hono/core/pom.xml, line 25, column 14
413[ERROR]   
414[ERROR]   The project org.eclipse.hono:hono-client:1.0.0-SNAPSHOT (/home/travis/build/eclipse/hono/client/pom.xml) has 2 errors
415[ERROR]     Unresolveable build extension: Plugin org.apache.felix:maven-bundle-plugin:3.5.0 or one of its dependencies could 
```

which is not related to my changes

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -641,4 +591,165 @@ public OptionalInt getTraceSamplingPriority(final String authId) {
...
+     * 
+     * @return This tenant so the API can be used fluently.
+     */
+    private TenantObject addTrustedCA(final JsonObject trustedCa) {
+
+        if (trustConfigurations == null) {
+            trustConfigurations = new HashMap<>();
+        }
+        final String subjectDn = getProperty(trustedCa, TenantConstants.FIELD_PAYLOAD_SUBJECT_DN, String.class);
+        if (Strings.isNullOrEmpty(subjectDn)) {
```
sophokles73: this check should be at the start of the method, in order to prevent unnecessary creation of the HashMap

core/src/main/java/org/eclipse/hono/util/TenantObject.java
sophokles73: shouldn't it return an empty list?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -117,173 +124,116 @@ public TenantObject setEnabled(final boolean flag) {
...
-            final X509Certificate cert = getTrustedCertificateAuthority();
-            if (cert != null) {
-                trustAnchor = new TrustAnchor(cert, null);
-                return trustAnchor;
-            } else {
-                return getTrustAnchorForPublicKey(
-                        getProperty(TenantConstants.FIELD_PAYLOAD_TRUSTED_CA, JsonObject.class));
+    public List<TrustAnchor> getTrustAnchors() throws GeneralSecurityException {
+        final List<JsonObject> configs = Optional.ofNullable(getTrustedCAs()).orElse(Collections.emptyList());
+        if (trustAnchors == null || trustAnchors.size() != configs.size()) {
```
sophokles73: this is based on the fact that you can add duplicate trusted CAs, right?
In any case, the number of trusted CAs is expected to be very low, i.e. single digits. IMHO it would be totally ok (and much more robust), to simply *always* newly create the TrustAnchor instances from the adapterConfigurations ...

Alfusainey: yes, i thought about this too.

Alfusainey:
@sophokles73 am now working on adding sanity checks in the tenant service so that tenant payloads that contains trust configs where the public keys are set also contains the validity period

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -117,173 +122,113 @@ public TenantObject setEnabled(final boolean flag) {
...
-     * @throws CertificateException if the value of <em>cert</em> property cannot be parsed into
-     *             an X.509 certificate.
-     */
-    @JsonIgnore
-    public X509Certificate getTrustedCertificateAuthority() throws CertificateException {
-
-        final JsonObject trustedCa = getProperty(TenantConstants.FIELD_PAYLOAD_TRUSTED_CA, JsonObject.class);
-        if (trustedCa == null) {
-            return null;
+        final String subjectDn = certificate.getSubjectX500Principal().getName();
```
sophokles73: why not simply
```
return addTrustAnchor(
  certificate..getPublicKey(),
  certificate..getSubjectX500Principal(),
  certificate.getNotBefore().toInstant(),
  certificate.getNotAfter().toInstant());
```

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -641,4 +586,166 @@ public OptionalInt getTraceSamplingPriority(final String authId) {
...
+        }
+    }
+
+    /**
+     * Get the set of subject distinguished names configure for this tenant.
+     * 
+     * @return The set of subject distinguished names or an empty set if no CA has been configured for this tenant.
+     */
+    @JsonIgnore
+    public Set<X500Principal> getTrustedCaSubjectDns() {
```
sophokles73: is this method actually used anywhere?

Alfusainey: yes, since no two different tenants are allowed to have the same subject name, this method is used to facilitate that check (in `FileBasedTenantService`). however, with this change, a tenant can have multiple trust configs with the same subject name but different public keys

core/src/test/java/org/eclipse/hono/util/TenantObjectTest.java
```diff
@@ -229,11 +229,12 @@ public void testGetTrustAnchorForPublicKeyWithoutValidityPeriod() {
...
-        final TenantObject obj = tenantPayload.mapTo(TenantObject.class);
+
         try {
-            assertThat(obj.getTrustAnchors().size(), is(0));
-        } catch (GeneralSecurityException e) {
-            // should never get here
+            tenantPayload.mapTo(TenantObject.class);
+            fail("should not have been able to create a TenantObject from a malformed trusted CA");
+        } catch (final IllegalArgumentException e) {
+            // expected
```
Alfusainey: @sophokles73: i will like to get you opinion on this: do you think this will cause a back incompatibility issue if existing content in a device registry database does not store the validity period of public keys? currently, Hono does not require this at all and existing data might not have such validity periods recorded

otherwise, we can enforce this check only when storing the data and not when loading it

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -748,4 +740,20 @@ private Instant getNotAfter(final JsonObject trustedCa) {
...
+
+    private boolean isValidTenantConfig(final JsonObject trustedCa) {
+
+        final String subjectDn = getProperty(trustedCa, TenantConstants.FIELD_PAYLOAD_SUBJECT_DN, String.class);
+        final String publicKey = getProperty(trustedCa, TenantConstants.FIELD_PAYLOAD_PUBLIC_KEY, String.class);
+        final Instant notBefore = getNotBefore(trustedCa);
+        final Instant notAfter = getNotAfter(trustedCa);
+
+        if (Strings.isNullOrEmpty(subjectDn) ||
+                // if the public key is present, then
```
sophokles73: is there a case where the public key does not need to be present?

Alfusainey: no, there is not. do you mean we check for its absence?

sophokles73: I would expect this method to throw an IllegalArgumentException if the passed in JSON object did not contain a public key, wouldn't you?

core/src/test/java/org/eclipse/hono/util/TenantObjectTest.java
sophokles73: I would enforce it only when storing the data for now.

Alfusainey: ok

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: sure 👍

site/documentation/content/api/tenant/index.md
```diff
@@ -170,17 +171,33 @@ The table below provides an overview of the members defined for the *tracing* JS
...
+| *public-key*             | *yes*       | *string*     |               | The Base64 encoded binary DER encoding of the trusted root certificate's public key. |
 | *algorithm*              | *no*       | *string*      | `RSA`        | The name of the public key algorithm. Supported values are `RSA` and `EC`. This property is ignored if the *cert* property is used to store a certificate. |
 
-* The *subject-dn* MUST be unique among all registered tenants.
-* Either the *cert* or the *public-key* MUST be set.
+**Examples**
+
+Below is an example for a payload of the response to a *get* request for tenant `TEST_TENANT`. The tenant is configured with two valid trusted certificate authorities having the same *public-key* and *subject-dn* but with overlapping validity periods.
```
sophokles73: they obviously do not have the same public key and it is hard to tell whether they have overlapping validity periods, but they are definitely valid right now

site/documentation/content/api/tenant/index.md
```diff
@@ -170,17 +171,33 @@ The table below provides an overview of the members defined for the *tracing* JS
...
+
+~~~json
+ {
+   "tenant-id" : "TEST_TENANT",
+   "enabled" : true,
+   "trusted-ca": [ {
+           "subject-dn": "CN=ca,OU=Hono,O=Eclipse",
+           "public-key": "NOTAPUBLICKEY==",
```
sophokles73: this should be replace with something like `PublicKey==` ...

site/documentation/content/api/tenant/index.md
```diff
@@ -232,6 +249,8 @@ The table below contains the properties which are used to configure a tenant's d
 | *effective-since*        | *yes*     | *string*      | `-`           | The point in time at which the current settings became effective, i.e. the start of the first accounting period based on these settings. The value MUST be an [ISO 8601 compliant *combined date and time representation in extended format*](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations).|
 | *max-bytes*              | *no*      | *number*      | `-1`          | The maximum number of bytes allowed for the tenant for each accounting period. MUST be an integer. A negative value indicates that no limit is set. |
 | *period*                 | *no*      | *object*      | `-`          | The mode and length of an accounting period, i.e. the data usage can limited based on the defined number of days or on a monthly basis. |
+| *period-in-days*         | *no*      | *number*      | `30`          | The length of an accounting period, i.e. the number of days over which the data usage is to be limited. MUST be a positive integer. |
```
sophokles73: this doesn't seem to be related to this PR, is it?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -117,173 +122,103 @@ public TenantObject setEnabled(final boolean flag) {
...
 
         Objects.requireNonNull(publicKey);
         Objects.requireNonNull(subjectDn);
+        Objects.requireNonNull(notBefore);
+        Objects.requireNonNull(notAfter);
 
         final JsonObject trustedCa = new JsonObject();
         trustedCa.put(TenantConstants.FIELD_PAYLOAD_SUBJECT_DN, subjectDn.getName(X500Principal.RFC2253));
         trustedCa.put(TenantConstants.FIELD_PAYLOAD_PUBLIC_KEY, publicKey.getEncoded());
+        trustedCa.put(TenantConstants.FIELD_PAYLOAD_NOT_BEFORE, notBefore.toEpochMilli());
```
sophokles73: I think we will have a problem here. The device registry implementation currently uses `TenantObject` as its data format for storing tenant information in-memory. However, it uses `Tenant` and `TenantCertificateAuthority` from the service-base module in the Management API. `Tenant` instances are therefore mapped hence and forth to `TenantObject` instances.

The `TenantObject` however, has been originally designed as the value object for the result of the *get Tenant* operation of the Tenant API. The result does not contain *not-before* and *not-after* (because the Tenant service is supposed to have returned only currently valid CAs). 

If we want to continue to use `TenantObject` within the device registry as the data store format, then the *not-before* and *not-after* property format needs to match the format used in the `TenantCertificateAuthority`, which is the ISO string format ...

However, I am not really convinced that it is a good idea to use `TenantObject` as the internal data store format in the device registry because it requires us to keep the validity period properties in `TenantObject`, which the Tenant API doesn't require ... Instead, the device registry should probably go back to keeping the data in plain `JsonObject`s as we originally did ...

Alfusainey: > then the not-before and not-after property format needs to match the format used in the TenantCertificateAuthority, which is the ISO string format ...

this was the reason i used the `toString()` representation of the instant since the management API assumes that the format is an ISO format

> Instead, the device registry should probably go back to keeping the data in plain JsonObjects as we originally did ...

i have the feeling even we use a `JsonObject` to store the registry data in-memory, it will still contain these properties (since they are stored in a database/record system). if the tenantobject does not provide any API to retrieve the `not_before`/`not-after` properties, wouldn't that be what the Tenant API requires i.e hiding these properties from clients?

site/documentation/content/api/tenant/index.md
Alfusainey: this happens when i resolved this branch to remove the conflicts introduced by this line. but yes, it is not related to this PR

site/documentation/content/api/tenant/index.md
Alfusainey: the example JSON evolved overtime but not the surrounding text :-)

site/documentation/content/api/tenant/index.md
Alfusainey: ok. since the second CA uses an `EC` key type, i will replace it with `ECKey==` to reflect the type of key it uses

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: @sophokles73 so whats the status of this PR? if you think we should definitely deprecate the `TenantObject` class and replace it with a `JsonObject` then i will suggest that we do that in a separate PR. this one is already quiet fat IMHO :-) WDYT?

sophokles73: The device registry will in any case need to keep track of the *not-before* and *not-after* properties. The registry needs these properties to determine which CAs to return to an adapter.

The `TenantObject` class has been designed to contain all information defined by the response payload format of the Tenant API's *get Tenant* operation. It will continue to serve this purpose in the Hono Client.

The question is, if the `TenantObject` is also suitable as the data container format for the Device Registry. If it should serve this purpose as well, then it obviously needs to contain the *not-before* and *not-after* properties as well, and in that case it needs to keep them in the ISO String format.

However, if the registry contains all data as plain `JsonObject`s which comply with the schema defined by the Management API, then the `TenantObject` does not need to contain the validity period, because instances of it will only contain CAs which are valid at the moment the *get Tenant* request gets processed.

So we have two options, either 
1. keep the validity period in `TenantObject` and change the response payload of the *get Tenant* operation to also (optionally) include the *not-before* and *not-after* properties, or
2. change the Device Registry implementation to use plain `JsonObject`s for keeping the data in memory and adapt the mapping to/from the Management API accordingly

Thinking a little about it, I guess option 1 is easier to implement while still leaving us the option to later change the Device Registry implementation and no longer include the validity period in the response to *get Tenant*.

WDYT?

Alfusainey: @sophokles73 yes, I agree! I will go ahead and implement option one now.. 

> keep the validity period in TenantObject and change the response payload of the get Tenant operation to also (optionally) include the not-before and not-after properties, or

so this means that we should also update the Tenant API spec to (optionally) include these properties, right?

sophokles73: exactly

Alfusainey: @sophokles73 i pushed the changes

site/documentation/content/api/tenant/index.md
```diff
@@ -176,7 +176,9 @@ If the *trusted-ca* property is provided, then it MUST contain a non-empty array
...
 | :------------------------| :--------: | :------------ | :------------ | :---------- |
 | *subject-dn*             | *yes*      | *string*      |               | The subject DN of the trusted root certificate in the format defined by [RFC 2253](https://www.ietf.org/rfc/rfc2253.txt). Trusted CAs of the same tenant MAY share the same subject DN (e.g. allowing for the definition of overlapping validity periods). However, trusted CAs of different tenants MUST NOT share the same subject DN in order to allow for the look up of a tenant by the subject DN of one of its trusted CAs. |
-| *public-key*             | *yes*       | *string*     |               | The Base64 encoded binary DER encoding of the trusted root certificate's public key. |
+| *public-key*             | *yes*       | *string*     |               | The Base64 encoded binary DER encoding of the trusted root certificate's public key.  |
+| *not-before*             | *no*       | *string*     |               | The [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations) string representation of the start date of the trusted certificate's validity period.|
```
sophokles73: I'd rather put the validity period props after the *algorithm* property so that the algorithm is *closer* the public key.
IMHO it would also be good to explain that the validity period does not need to be verified by the client because the payload will only contain CAs which are valid at the time of retrieval.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
sophokles73: I thought we agreed that the *not-before* and the *not-after* should be stored in ISO format, didn't we?

site/documentation/content/api/tenant/index.md
Alfusainey: ok, updated the pr

core/src/main/java/org/eclipse/hono/util/TenantObject.java
sophokles73: we seem to be back to the old problem ;-)
IMHO we should explicitly use a `DateTimeFormatter` to produce the intended ISO format instead of relying on *toString()*.
The same holds true for parsing the values ...

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -641,4 +578,172 @@ public OptionalInt getTraceSamplingPriority(final String authId) {
...
+        trustedCas.add(trustedCa);
+        trustConfigurations.put(subjectDn, trustedCas);
+        return this;
+    }
+
+    private Instant getNotBefore(final JsonObject trustedCa) {
+        Instant instant = null;
+        final String strValue = getProperty(trustedCa, TenantConstants.FIELD_PAYLOAD_NOT_BEFORE, String.class);
+        if (strValue != null) {
+            instant = DateTimeFormatter.ISO_INSTANT.parse(strValue, Instant::from);
```
sophokles73: IMHO this will fail to parse `2019-09-23T13:15:01+02:00` ...
I believe that you need to use `DateTimeFormatter.ISO_OFFSET_DATE_TIME` for that to work.

Maybe you can add a corresponding test case and see if it works.

Alfusainey: ok

Alfusainey: @sophokles73

> IMHO this will fail to parse 2019-09-23T13:15:01+02:00

but the device registry should not store this datetime value because it is not supported by Hono. As of  #1463, Hono now explicitly uses `DateTimeFormatter.ISO_INSTANT` for all its date time values (including the management API). the date time fields are now annotated with `@HonoTimestamp` and this annotation internally uses `DateTimeFormatter.ISO_INSTANT` as the supported date time format.

so I think the registry should reject storing such datetime values, WDYT?

sophokles73: I do not think that's true. The Management API supports clients specifying *not-before* values like `2019-09-23T13:15:01+02:00`. These are correctly parsed (see `SecretsTest`) and then *normalized* to UTC. The `TenantObject` therefore should support this too FMPOV

Alfusainey: @sophokles73 Its true that clients can specify such values, but then all values are converted to `DateTimeFormatter.ISO_INSTANT` before they are transported to the backend service

 if you take a look at`SecretsTests#testDateFormats()`  you will see that the decoded `not-before` and `not-after` properties are not exactly in the same form as given by the client. 

> The TenantObject therefore should support this too FMPOV

ok, let me think about this :)

sophokles73: > Its true that clients can specify such values, but then all values are converted to DateTimeFormatter.ISO_INSTANT before they are transported to the backend service.

What do mean with *before they are transported to the backend service*? The data is contained in a JSON object that is POSTed to the device registry's HTTP endpoint ...

sophokles73: > if you take a look atSecretsTests#testDateFormats() you will see that the decoded not-before and not-after properties are not exactly in the same form as given by the client.

they don't need to. it is only important that the registry can parse these encoded instants. The same holds true for the TenantClient. It needs to be able to parse such encoded instants. It may then (internally) always us UTC to represent them, though ...

Alfusainey: I see your point! I think the way to approach this would be to define a specialise class in `org.eclipse.hono.util` to represent the domain object for a trusted certificate authority. Then I can annotate the `not-before` and `not-after` instances of the class using `@HonoTimestamp` so that Jackson can handle the parsing of the properties properly (as done in the management API)

sophokles73: That would be an option, I guess. Based on our recent approach for sharing the low level value objects, you may even simply move the existing `TrustedCertificateAuthority` class from service-base to core and re-use/modify it accordingly. Remember that the Management API will also need to accept a list of CAs instead of just a single one ...

Alfusainey: yes, thats what am doing actually!

> Remember that the Management API will also need to accept a list of CAs instead of just a single one ...

yes, it already does that in this PR

sophokles73: sure

Alfusainey: @sophokles73 I have been struggling with the following problem since in the morning and its not really funny :-)

for some reason, deserialising a `JsonObject` to a `org.eclipse.hono.util.TrustedCertificateAuthoritory` does not correctly parse a `datetime` with `offset` to an instant and this problem only happens in the `hono-core` module

example, giving this json:

```Java
final JsonObject json = new JsonObject()
                .put("subject-dn", "cn=test")
                .put("public-key", "notpublicKey".getBytes())
                .put("not-before", "2019-04-05T13:45:07-04:00")
                .put("not-after", "2037-06-01T14:00:00Z");
```
I want to convert it to an instance of `org.eclipse.hono.util.TrustedCertificateAuthoritory`,  so I did: `final TrustedCertificateAuthority ca = json.mapTo(TrustedCertificateAuthority.class);`

however, when I read the instant value using `ca.getNotBefore()` I get `2019-04-05T13:45:07Z`, the `time` value(i.e  13:45) is wrong.

The irony is when I run the same exact code inside of package `org.eclipse.hono.service.management.tenant` :

`final org.eclipse.hono.util.TrustedCertificateAuthority ca = json.mapTo(org.eclipse.hono.util.TrustedCertificateAuthority.class);` then the result I get is `2019-04-05T17:45:07Z` (which is the correct value, IMO).

I don't understand why deserialising a JSON to a `org.eclipse.hono.util.TrustedCertificateAuthority` should yield different results in different modules. I removed the following dependencies in the `core/pom.xml`, which I was suspecting was causing the problem in the `core` module, but with no success:

```xml
    <dependency>
      <groupId>com.fasterxml.jackson.core</groupId>
      <artifactId>jackson-databind</artifactId>
      <version>${jackson.version}</version>
    </dependency>
    <dependency>
      <groupId>com.fasterxml.jackson.core</groupId>
      <artifactId>jackson-annotations</artifactId>
      <version>${jackson.version}</version>
```

have you ever had this problem with the Jackson library?

BTW: the `TrustedCertificateAuthority` is exactly the same as in the `service-base` and `core` modules.

sophokles73: Can you post the code of your `TrustedCertifcateAuthority` class in core?

Alfusainey:
```java
package org.eclipse.hono.util;

import java.time.Instant;

import org.eclipse.hono.annotation.HonoTimestamp;

import com.fasterxml.jackson.annotation.JsonInclude;
import com.fasterxml.jackson.annotation.JsonProperty;
import com.fasterxml.jackson.annotation.JsonInclude.Include;

/**
 * A trusted CA configuration.
 *
 */
@JsonInclude(value = Include.NON_NULL)
public class TrustedCertificateAuthority {

    @JsonProperty(value = TenantConstants.FIELD_PAYLOAD_SUBJECT_DN, required = true)
    private String subjectDn;

    @JsonProperty(TenantConstants.FIELD_PAYLOAD_PUBLIC_KEY)
    private byte[] publicKey;

    @JsonProperty(TenantConstants.FIELD_PAYLOAD_CERT)
    private byte[] certificate;

    @JsonProperty(TenantConstants.FIELD_PAYLOAD_NOT_BEFORE)
    @HonoTimestamp
    private Instant notBefore;

    @JsonProperty(TenantConstants.FIELD_PAYLOAD_NOT_AFTER)
    @HonoTimestamp
    private Instant notAfter;

    @JsonProperty(TenantConstants.FIELD_PAYLOAD_KEY_ALGORITHM)
    private String keyAlgorithm;

    /**
     * Sets the subject-dn configuration property.
     *
     * @param subjectDn The subject-dn property to set.
     * @return          a reference to this for fluent use.
     */
    public TrustedCertificateAuthority setSubjectDn(final String subjectDn) {
        this.subjectDn = subjectDn;
        return this;
    }

    public String getSubjectDn() {
        return subjectDn;
    }

    /**
     * Sets the public-key configuration property.
     *
     * @param publicKey The public-key property of the trusted root certificate.
     * @return          a reference to this for fluent use.
     */
    public TrustedCertificateAuthority setPublicKey(final byte[] publicKey) {
        this.publicKey = publicKey;
        return this;
    }

    public byte[] getPublicKey() {
        return publicKey;
    }

    /**
     * Sets the cert configuration property.
     *
     * @param certificate The cert property of the trusted root certificate.
     * @return            a reference to this for fluent use.
     */
    public TrustedCertificateAuthority setCertificate(final byte[] certificate) {
        this.certificate = certificate;
        return this;
    }

    public byte[] getCertificate() {
        return certificate;
    }

    /**
     * Sets the not-before property.
     * 
     * @param notBefore The not-before value to assign.
     * 
     * @return A reference to this API so it can be used fluently.
     */
    public TrustedCertificateAuthority setNotBefore(final Instant notBefore) {
        this.notBefore = notBefore;
        return this;
    }

    public Instant getNotBefore() {
        return notBefore;
    }

    /**
     * Sets the not-after property.
     * 
     * @param notAfter The not-after value to assign.
     *
     * @return A reference to this API so it can be used fluently.
     */
    public TrustedCertificateAuthority setNotAfter(final Instant notAfter) {
        this.notAfter = notAfter;
        return this;
    }

    public Instant getNotAfter() {
        return notAfter;
    }

    /**
     * Sets the algorithm property of the trusted root certificate.
     * 
     * @param keyAlgorithm  The algorithm name of the public key.
     * @return              a reference to this for fluent use.
     */
    public TrustedCertificateAuthority setKeyAlgorithm(final String keyAlgorithm) {
        this.keyAlgorithm = keyAlgorithm;
        return this;
    }

    public String getKeyAlgorithm() {
        return keyAlgorithm;
    }
}
```

sophokles73: The difference between the test case running in `service-base` and the test case running in `core` is the JVM. In `service-base` it is Java 11, in `core` it is Java 8. Run the test in `core` using Java 11, I get the same (expected) result as in `service-base`.
That said, I believe that the deserializer used in `HonoTimestamp` is wrong. It should be `DateTimeFormatter.ISO_OFFSET_DATE_TIME` instead of `DateTimeFormatter.ISO_INSTANT`.
I will test this ...

Alfusainey: you're really amazing! yes, that was exactly it. thank you 🙏

Alfusainey:
@sophokles73 patch updated

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -696,54 +668,47 @@ private TrustAnchor getTrustAnchor(final JsonObject trustedCa) throws GeneralSec
...
+        // public key or the encoded certificate must be present
+        if (Strings.isNullOrEmpty(cert) && Strings.isNullOrEmpty(publicKey)) {
             return false;
         }
         return true;
     }
+
+    /**
+     * TODO.
+     * @return TODO.
```
Alfusainey: opps, just saw this.. will change tomorrow

core/pom.xml
```diff
@@ -30,7 +30,8 @@
   <url>https://www.eclipse.org/hono</url>
 
   <properties>
-    <java.level>1.8</java.level>
+    <main.java.level>1.8</main.java.level>
+    <test.java.level>11</test.java.level>
```
sophokles73: I do not think that this is the solution to the problem. The underlying problem FMPOV is that the `HonoTimestamp` annotation is not implemented correctly and, coincidentally, works on Java 11 but doesn't on Java 8. I'd rather fix the underlying issue than addressing the symptoms only. Let my dig into this a little more ...

Alfusainey: ok

sophokles73: @Alfusainey 

I have pushed a change to master which should fix the problem with the `HonoTimestamp` annotation. After rebasing, you should be able to run your test case in core without any further modifications of the JVM being used ...

Alfusainey: nice 👍 i will undo this last commit and force push to this branch

core/src/main/java/org/eclipse/hono/util/TrustedCertificateAuthority.java
```diff
@@ -0,0 +1,143 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+
+/**
+ * A trusted CA configuration.
+ *
+ */
+@JsonInclude(value = Include.NON_NULL)
+public class TrustedCertificateAuthority {
```
sophokles73: IMHO we should declare this class final as I do not think that it is meant to be extended.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -590,24 +582,28 @@ public OptionalInt getTraceSamplingPriority(final String authId) {
...
+            final TrustedCertificateAuthority trustedCa = config.mapTo(TrustedCertificateAuthority.class);
+
             addTrustedCA(trustedCa);
         } else if (List.class.isInstance(configuration)) {
 
             final List<Map<String, Object>> configurations = (List<Map<String, Object>>) configuration;
-            configurations.forEach(json -> {
-                final JsonObject trustedCa = new JsonObject(json);
+            configurations.forEach(config -> {
+                final TrustedCertificateAuthority trustedCa = new JsonObject(config).mapTo(TrustedCertificateAuthority.class);
```
sophokles73: what if the JSON cannot be mapped to `TrustedCertificateAuthority`?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -696,54 +668,47 @@ private TrustAnchor getTrustAnchor(final JsonObject trustedCa) throws GeneralSec
...
         }
         return true;
     }
+
+    /**
+     * TODO.
+     * @return TODO.
+     */
+    @JsonIgnore
+    public List<JsonObject> getTrustedCAs() {
```
sophokles73: what is this method good for? who needs it?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -638,44 +628,26 @@ public TenantObject setTrustedCa(final Object configuration) {
...
+        final byte[] encodedKey = trustedCa.getPublicKey();
+        final String type = Optional.ofNullable(trustedCa.getKeyAlgorithm()).orElse("RSA");
 
-        final Instant notBefore = getNotBefore(trustedCa);
-        final Instant notAfter = getNotAfter(trustedCa);
+        final Instant notBefore = trustedCa.getNotBefore();
+        final Instant notAfter = trustedCa.getNotAfter();
         final Instant now = Instant.now();
 
         if (subjectDn == null || encodedKey == null || notBefore == null || notAfter == null) {
```
sophokles73: the check against *now* means that we *always* transfer *all* CAs from the registry to the protocol adapter?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: its used in the `FileBasedTenantService#convertTenantObject` when converting a `TenantObject` to a `Tenant`. I think the its better to call it `getTrustedCaConfigurations()`?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: I don't understand?

if `notBefore` is later than the current instant of time, then its early to use the CA. if now is at a later point in time than `notAfter`, then it expired. In any of these cases, a `null` trust anchor is returned. so not *all* the CAs from the registry are returned, only valid ones

core/src/main/java/org/eclipse/hono/util/TenantObject.java
sophokles73: IMHO it should call *getTrustedAuthorities*, shouldn't it? Since we are sharing the `TrustedCertificateAuthority` class there is nothing to convert, is there?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
sophokles73: who is going to invoke this method?I believe it will be the protocol adapter, right? If that is true, then the `TenantObject` instance has been deserialized from the payload received in response to the *get Tenant* operation invocation, right? So far, we have said that the registry only transfers *currently valid* CAs to the adapters, so this filtering shouldn't be necessary, right?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: > IMHO it should call getTrustedAuthorities, shouldn't it?

there is already such a method which returns a `List<TrustedCertificateAuthority>`. The `getTrustedCAs` method returns the trust configurations in JSON format.

> Since we are sharing the TrustedCertificateAuthority class there is nothing to convert, is there?

But they are of different types. its the entire tenant object thats converted, not just the `TrustedCertificateAuthority`:

```java
    static Tenant convertTenantObject(final TenantObject tenantObject) {

        if (tenantObject == null) {
            return null;
        }

        final var tenant = new Tenant();

        tenant.setEnabled(tenantObject.isEnabled());
       ...
        Optional.ofNullable(tenantObject.getTrustedCAs())
                .ifPresent(trustConfigs -> setTrustedAuthorities(tenant, trustConfigs)); 
      ...
        return tenant;
    }
```

I cannot do `tenant:: setTrustedAuthorities` even if the method returns `org.eclipse.hono.TrustedCertificateAuthority`, because the method expects a different type

core/src/main/java/org/eclipse/hono/util/TenantObject.java
Alfusainey: I see your point. I added this to test the `TenantObject` class: so that it can return only the valid anchors even if it contains invalid trust configurations. shall I remove it together with the corresponding tests in `TenantObjectTest`?

Alfusainey: I will put the check for validity inside of the `TrustedCertificateAuthority` class

core/src/main/java/org/eclipse/hono/util/TenantObject.java
sophokles73: ???
just do
```
tenant.setTrustedAuthorities(new LinkedList(tenantObject.getTrustedAuthorities()));
```

Alfusainey:
@sophokles73 the last commit contains the changes i mention in the Gitter channel. i will update the management spec if you think this change makes any sense

Alfusainey:
@sophokles73 any news on this?

Alfusainey:
@sophokles73 any news on this PR?

sophokles73:
@Alfusainey right at this moment I do not think that it will make it into 1.0.0. FMPOV we first need to discuss the necessary changes in the example registry before we can go on with this. At the moment, though, this is not the top priority ...

Alfusainey:
> @Alfusainey right at this moment I do not think that it will make it into 1.0.0. FMPOV we first need to discuss the necessary changes in the example registry before we can go on with this. At the moment, though, this is not the top priority ...

this is a bit sad 😢

sophokles73:
I am sorry for all the effort you put into this so far. However, since this is a feature that is at the core of Hono's security architecture, I do not want to rush it but instead make sure that we do the right thing here.
This also doesn't mean that we will not need to continue to work on it, it just means that it might be postponed to e.g. 1.1.0.

Alfusainey:
> I am sorry for all the effort you put into this so far. However, since this is a feature that is at the core of Hono's security architecture, I do not want to rush it but instead make sure that we do the right thing here.

yes, I agree

> This also doesn't mean that we will not need to continue to work on it, it just means that it might be postponed to e.g. 1.1.0.

sure! I was just looking forward to seeing it in 1.0 :-) will now spend sometime reading through the Hono docs. to see if I can spot spelling mistakes + other minor improvements

PR #1404, opened 2019-07-25
Make order in which HTTP handlers are added more obvious

calohmn:


sophokles73:
@calohmn are you going to merge this or does it need more work?

PR #1409, opened 2019-07-25
Add HTTP Management API to API section.

sophokles73:
Use Swagger UI for rendering the OpenAPI spec file.

You can take a look at the rendered UI by means of running
```sh
hono/site/documentation> hugo server
```
and then opening the URL output on the console. The Swagger UI is in the API section ...


calohmn:
@sophokles73 Is the used Swagger UI version documented somewhere? Such info would be beneficial in deciding whether to update the Swagger UI (e.g. for an update to 3.23.2 in order to prevent empty request "Examples" combo boxes from being shown).

sophokles73:
@calohmn no, it isn't, apart from the original CQ which contains the version.
We should probably rename the swagger related artifacts to include the version number. WDYT?

calohmn:
@sophokles73 yes, we can do that.

sophokles73:
@calohmn are you doing it?

calohmn:
@sophokles73 yes

Issue #1408, opened 2019-07-25
Version identifier should be part of resource URIs


sophokles73:
Currently, the `device-registry-v1.yaml` defines the resource URIs without the `v1` prefix. However, I believe they need to be part of them in order to e.g. clearly distinguish them from later versions.
Example:
`/tenants/${tenantId}` should be `/v1/tenants/${tenantId}` instead.

sophokles73:
@ctron @dejanb @calohmn WDYT?

dejanb:
@sophokles73 sounds right to me

calohmn:
@sophokles73 I agree.

sophokles73:
Ok, then I'll create a PR.

dejanb:
@sophokles73 on the closer look OpenAPI says that version should be defined in the 'server' component.

https://swagger.io/docs/specification/api-host-and-base-path/
https://swagger.io/specification/

When you asked the question, I was under impression that we don't have v1 defined anywhere. Why do you think it's better to have it explicitly in paths?

sophokles73:
hmm, originally, I thought that for an implementation of the API, it would be helpful (or even required), to be able to determine the version of the API from the URI, that the client invokes. However, for that it is not necessary to include the version identifier in the resource paths of the OpenAPI spec. An implementation can inspect the request URI in any case.
However, if the version identifier is not part of the resource path(s), then you can expose the API using arbitrary base URIs, e.g. `/v1`, `/bumlux` or `v3`. I am not 100% sure, how the OpenAPI spec will be used in practice. My understanding now (after reading the sections you pointed out) is, that the document we publish on our website will be *specific* to the Sandbox that we are hosting and thus will expose the endpoints on a *server* that we define in the document ourselves. That server contains the `v1` base URI and thus everything seems to be ok.
It should be fair for an implementation of this API to assume that the client in fact has sent a request for the specific version that is being implemented. It is not the responsibility of the implementing code to *route* requests to the handlers implementing the matching version. That should be done at the level of the web server (servlet engine etc).
So, all in all, I guess that I am now convinced that we shouldn't include the version identifier in the resource path :-)

sophokles73:
not an issue after all

PR #1410, opened 2019-07-26
[#1408] Include version identifier in resource URIs.

sophokles73:
adds version identifier to resource URIs in OpenAPI spec file


sophokles73:
I am no longer convinced that we should do this.

Issue #1188, opened 2019-04-24
Protocol adapters do never close "inactive" sender links


DanielMaier-BSI:
Hono protocol adapters create AMQP sender links for events and telemetry data for each tenant on the first message for a particular tenant. However, these links get never closed. Would it be an improvement to close these links after some time of inactivity, as links require some memory on both ends? From my point of view this inactivity period might be quite high per default, e.g. one day. 

Background: We execute some system tests periodically with new tenant for each test and delete these tenants afterwards. Over time we get many open sender links in protocol adapters that will never be needed again. These sender links get closed only when the connection gets closed, e.g. when we restart protocol adapter.

WDYT? Is this requirement to specific to our setup or might others also run into this issue?

sophokles73:
Sounds like a reasonable feature to me. I think that we could close links after a much shorter period of inactivity (e.g. 1h) because if the devices of the affected tenant send messages at such a low rate then it probably doesn't hurt much if the downstream link gets established for each individual message.

DanielMaier-BSI:
I am also fine with one hour inactivity period. I had only in mind (I don't know if I remember correctly and if it is still the case) that messages to protocol adapters without pre-established links often fail with no credit issues.

sophokles73:
That is indeed an issue. We might be able to address this with a well-chosen *flow latency* on the protocol adapters. However, if multiple devices of the same tenant start to send a message almost at the same time, then only the first message will be sent over the newly established downstream link whereas messages that have been uploaded afterwards, while the link is not established yet, will get the dreaded 503 ...

ctron:
Maybe that is some configuration value that should be specified on a "per-tenant" basis. As this could be lower for your "test tenants" and higher for normal tenants.

sophokles73:
I do not quite see how it is related to the tenant. The time it takes to receive the first credits depends on the AMQP Messaging Network peer that the adapter is connected to and the network latency between them. But at the time the sender link is established, there is nothing tenant specific happening FMPOV.

ctron:
No, what I meant was the initial time of inactivity before the link gets closed. Instead of using 1 hour or 1 day, this could be configured for the tenant.

As my understanding is, that tenants get created for testing only. And those tenants could actually have a pretty short inactivity timeout (1 minute). While normal tenants could actually have a day (or some other system wide default).

sophokles73:
Fixed via #1416

PR #1416, opened 2019-07-31
[#1188] Close inactive downstream sender links.

b-abel:
There are a number of options where this feature (#1188) can be implemented and this is just a draft to discuss the general direction. As I am new to the Hono client code I would love to get some feedback.

When a downstream link is opened by a protocol adapter a timer is started that closes this link after a certain time. Each time the a message is sent on this link the timer is restarted. The time duration is configured in the `ProtocolAdapterProperties` section (e.g. "hono.mqtt.inactiveLinkTimeout") and defaults to 1 hour.

I tried to accomplish two things here:

1. be specific for downstream sender links (instead of a generic solution for all sender links)
2. be specific for protocol adapters regarding the configuration (instead of adding it to the very generic `ClientConfigProperties`). 

The result is FMPOV not really elegant, which is mainly because the downstream connection is mostly configured with `ClientConfigProperties` (in "hono.messaging"). 
Another point is that even though the class `AbstractDownstreamSender` sounds like a perfect candidate for the functionality, it's sub-classes need to be changed as well because they add custom send methods.


sophokles73:
> be specific for downstream sender links (instead of a generic solution for all sender links)

I actually do not see why we shouldn't apply this to any kind of client. However, in case of a `RequestResponseClient` we would obviously want to close both sender and receiver links ...

> be specific for protocol adapters regarding the configuration (instead of adding it to the very generic ClientConfigProperties)

Based on my comment above, IMHO it would be ok to have it in the `ClientConfigProperties`. The default should probably be to *never* time out, though ...

sophokles73:
> When a downstream link is opened by a protocol adapter a timer is started that closes this link after a certain time. Each time the a message is sent on this link the timer is restarted. 

I do not think that this is necessary. Instead, IMHO it should be sufficient to only *tag* the sender link with the current timestamp whenever a message is being sent over it. When the timer fires, it can then check if ( current time - timestamp from sender ) >= configured inactivity period. If it isn't, a new timer can be started with a delay of ( timestamp from sender + configured inactivity period ).
WDYT?

b-abel:
> IMHO it should be sufficient to only _tag_ the sender link with the current timestamp 

What do you mean by _tagging_ the link?

The advantage is that I don't need to cancel timers?

sophokles73:
> What do you mean by tagging the link?

Adding the current timestamp to the link's *attachments* ...

> The advantage is that I don't need to cancel timers?

Yes, you never need to cancel and schedule a new timer for each message being sent.

b-abel:
I pushed a new proposal with the changes discussed above. Documentation (and Javadoc) is still to be added.

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -221,4 +222,49 @@ protected static final void setApplicationProperties(final Message msg, final Ma
...
+        });
+    }
+
+    // visible for testing
+    long getRemainingTimeout(final long lastSend, final long now, final long timeout) {
+        final long timeDiff = now - lastSend;
+        if (timeDiff >= timeout) {
+            return 0L;
+        } else {
+            return timeout - timeDiff;
```
sophokles73: I think we should be a little more lenient here:
```
} else {
  final long remainingTimeout = timeout - timeDiff;
  if (remainingTimeout > timeDiff) {
    return timeout;
  } else {
    return remainingTimeout;
  }
}
```
otherwise we might end up with starting a new timer (with ever decreasing delay) for each message being sent, right?

client/src/test/java/org/eclipse/hono/client/impl/AbstractHonoClientTest.java
```diff
@@ -68,4 +83,83 @@ public void testApplicationPropertiesAreSetAtTheMessage(final TestContext ctx) {
...
+    /**
+     * TODO.
+     * @throws InterruptedException if sleep is interrupted.
+     */
+    @Test
+    public void testTimerIsRestartedWhenTimeoutNotExceeded() throws InterruptedException {
+        final AbstractHonoClient client = createClient();
+        client.connection.getConfig().setInactiveLinkTimeout(100L);
+        final long now = Instant.now().toEpochMilli();
+        when(client.sender.attachments().get("last-send-time", Long.class)).thenReturn(now);
```
sophokles73: constant

client/src/test/java/org/eclipse/hono/client/impl/AbstractHonoClientTest.java
```diff
@@ -68,4 +83,83 @@ public void testApplicationPropertiesAreSetAtTheMessage(final TestContext ctx) {
...
+    }
+
+    /**
+     * TODO.
+     */
+    @Test
+    public void testTimeoutExceededAndNeverSent() {
+        final AbstractHonoClient client = createClient();
+        client.connection.getConfig().setInactiveLinkTimeout(1L);
+        when(client.sender.attachments().get("last-send-time", Long.class)).thenReturn(0L);
```
sophokles73: constant

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -40,6 +40,11 @@
      * when no frames have been received from the remote peer.
      */
     public static final int DEFAULT_IDLE_TIMEOUT = 16000; //ms
+    /**
+     * The default amount of time (milliseconds) after which a a sender link will be closed when no messages have been
+     * sent.
```
sophokles73: It would be good to know, what the default semantics are, i.e. never time out?

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
b-abel: Note: @sophokles73 and I agreed offline that this is not an issue.

b-abel:
@sophokles73 Thank you for your detailed feedback. I just added the changes (and will continue with writing documentation).

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -45,6 +46,7 @@
  */
 public abstract class AbstractHonoClient {
 
+    public static final String ATTACHMENT_LAST_SEND_TIME = "last-send-time";
```
sophokles73: this needs a JavaDoc comment

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -221,4 +223,49 @@ protected static final void setApplicationProperties(final Message msg, final Ma
             msg.setApplicationProperties(applicationProperties);
         }
     }
+
+    /**
+     * TODO.
```
sophokles73: this needs proper JavaDoc

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -221,4 +223,49 @@ protected static final void setApplicationProperties(final Message msg, final Ma
...
+        if (inactiveLinkTimeout > 0) {
+            startAutoCloseTimer(inactiveLinkTimeout);
+        }
+    }
+
+    /**
+     * Stores the current time stamp on the sender link. This is used to detect and close inactive AMQP sender links.
+     */
+    protected final void storeLastSendTime() {
+        sender.attachments().set(ATTACHMENT_LAST_SEND_TIME, Long.class, Instant.now().toEpochMilli());
```
sophokles73: I am not sure but I do not think that we need to create an `Instant` object here but can simply use good old `System.currentTimeMillis()`. WDYT?

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -221,4 +223,49 @@ protected static final void setApplicationProperties(final Message msg, final Ma
...
+    }
+
+    private void startAutoCloseTimer(final long delay) {
+        connection.getVertx().setTimer(delay, id -> {
+            final Long lastSendTime = sender.attachments().get(ATTACHMENT_LAST_SEND_TIME, Long.class);
+            long remaining = 0;
+            if (lastSendTime != null) {
+                remaining = getRemainingTimeout(lastSendTime, Instant.now().toEpochMilli(),
```
sophokles73: see above

client/src/test/java/org/eclipse/hono/client/impl/AbstractHonoClientTest.java
```diff
@@ -68,4 +82,80 @@ public void testApplicationPropertiesAreSetAtTheMessage(final TestContext ctx) {
         verify(msg).setApplicationProperties(applicationPropsCaptor.capture());
         assertThat(applicationPropsCaptor.getValue().getValue(), is(applicationProps));
     }
+
+    /**
+     * TODO.
```
sophokles73: please provide proper docs

client/src/test/java/org/eclipse/hono/client/impl/AbstractHonoClientTest.java
```diff
@@ -68,4 +82,80 @@ public void testApplicationPropertiesAreSetAtTheMessage(final TestContext ctx) {
...
+     */
+    @Test
+    public void testNoAutoCloseOnDefault() {
+        final AbstractHonoClient client = createClient();
+        client.startAutoCloseTimer();
+        verify(client.connection.getVertx(), never()).setTimer(eq(1L), anyHandler());
+    }
+
+    /**
+     * TODO.
```
sophokles73: please provide proper docs

client/src/test/java/org/eclipse/hono/client/impl/AbstractHonoClientTest.java
```diff
@@ -68,4 +82,80 @@ public void testApplicationPropertiesAreSetAtTheMessage(final TestContext ctx) {
...
+        client.connection.getConfig().setInactiveLinkTimeout(1L);
+        when(client.sender.attachments().get(AbstractHonoClient.ATTACHMENT_LAST_SEND_TIME, Long.class)).thenReturn(0L);
+
+        client.startAutoCloseTimer();
+        verify(client.connection.getVertx()).setTimer(eq(1L), anyHandler());
+        verify(client.connection).closeAndFree(eq(client.sender), anyHandler());
+    }
+
+    /**
+     * TODO.
```
sophokles73: please provide proper docs

client/src/test/java/org/eclipse/hono/client/impl/AbstractHonoClientTest.java
```diff
@@ -68,4 +82,80 @@ public void testApplicationPropertiesAreSetAtTheMessage(final TestContext ctx) {
...
+        when(connection.getConfig()).thenReturn(new ClientConfigProperties());
+
+        final AbstractHonoClient client = new AbstractHonoClient(connection) {
+        };
+        client.sender = protonSender;
+        return client;
+    }
+
+    /**
+     * TODO.
```
sophokles73: please provide proper docs

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -40,6 +40,11 @@
      * when no frames have been received from the remote peer.
      */
     public static final int DEFAULT_IDLE_TIMEOUT = 16000; //ms
+    /**
+     * The default amount of time (milliseconds) after which a a sender link will be closed when no messages have been
+     * sent. The value <code>0</code> disables automatic link timeout.
+     */
+    public static final long DEFAULT_INACTIVE_LINK_TIMEOUT = 0; // ms
```
sophokles73: FMPOV you usually use `-1` for such a purpose as `0` would theoretically be a usable value. WDYT?

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -756,4 +762,22 @@ public final void setIdleTimeout(final int idleTimeoutMillis) {
             this.idleTimeoutMillis = idleTimeoutMillis;
         }
     }
+
+    /**
+     * TODO.
```
sophokles73: please provide proper docs.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -756,4 +762,22 @@ public final void setIdleTimeout(final int idleTimeoutMillis) {
...
+     * 
+     * @return The inactiveLinkTimeout in milliseconds.
+     */
+    public long getInactiveLinkTimeout() {
```
sophokles73: final

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -756,4 +762,22 @@ public final void setIdleTimeout(final int idleTimeoutMillis) {
...
+        return inactiveLinkTimeout;
+    }
+
+    /**
+     * TODO.
```
sophokles73: please provide proper docs.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -756,4 +762,22 @@ public final void setIdleTimeout(final int idleTimeoutMillis) {
...
+     * 
+     * @param inactiveLinkTimeout The inactiveLinkTimeout in milliseconds.
+     */
+    public void setInactiveLinkTimeout(final long inactiveLinkTimeout) {
```
sophokles73: final
should we enforce a minimum value, e.g. 0 or 60 seconds? does it make sense to measure the timeout in milliseconds? Seconds should do the trick FMPOV

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
b-abel: For me the 0 here has the semantic of: a limit (timeout) of 0 is no limit and means unlimited. But I see now that you can read it as "it instantly times out". This would not makes sense for this use case (and the vert.X timer which consumes this value currently does not accept a delay of 0). So, I'm OK with changing this to `-1` for disabling the timeout.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
b-abel: > should we enforce a minimum value, e.g. 0 or 60 seconds? 

I already enforce a minimum value of 0. Even though I don't think that very small values are useful.

> does it make sense to measure the timeout in milliseconds? Seconds should do the trick FMPOV

I used milliseconds to use the same unit as the other time related properties in `ClientConfigProperties`, hoping that this would be less confusing and error prone in the configuration and it is the unit the timer uses, so i do not need to do calculations between the unit of the timeout and the unit of the implementation (Vertx timer, and timestamps). But on the other hand I expect timeouts to be rather hours than fractions of a second and values in milliseconds get really big. So I can happily change that.

sophokles73: that was my concern as well. I also expect the timeout to be rather hours than seconds. So why not use minutes?

b-abel: For production use minutes are probably a good choice. But for integration tests even one one minute is too long FMPOV.

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -46,7 +46,10 @@
  */
 public abstract class AbstractHonoClient {
 
-    public static final String ATTACHMENT_LAST_SEND_TIME = "last-send-time";
+    /**
+     * They key that the timestamp of the last message send is stored under in a {@code ProtonLink}'s attachments.
```
sophokles73: The key ...

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -225,36 +228,41 @@ protected static final void setApplicationProperties(final Message msg, final Ma
...
-        final long inactiveLinkTimeout = connection.getConfig().getInactiveLinkTimeout();
-        if (inactiveLinkTimeout > 0) {
-            startAutoCloseTimer(inactiveLinkTimeout);
+        final long inactiveLinkTimeoutMillis = getInactiveLinkTimeoutMillis();
+        if (inactiveLinkTimeoutMillis > 0) {
+            startAutoCloseTimer(inactiveLinkTimeoutMillis);
         }
     }
 
+    private long getInactiveLinkTimeoutMillis() {
```
sophokles73: this should be a method of `ClientConfigProperties`, shouldn't it?

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -225,36 +228,41 @@ protected static final void setApplicationProperties(final Message msg, final Ma
...
+        return unit.toMillis(connection.getConfig().getInactiveLinkTimeout());
+    }
+
     /**
      * Stores the current time stamp on the sender link. This is used to detect and close inactive sender links as
      * documented in {@link #startAutoCloseTimer()}.
      */
     protected final void storeLastSendTime() {
-        sender.attachments().set(ATTACHMENT_LAST_SEND_TIME, Long.class, Instant.now().toEpochMilli());
+        sender.attachments().set(KEY_LAST_SEND_TIME, Long.class, System.currentTimeMillis());
```
sophokles73: the JavaDoc refers to the private *startAutoCloseTimer* method's documentation. However, that method has no JavaDoc and even if it had, you couldn't `@link` to it ...

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -41,10 +42,18 @@
      */
     public static final int DEFAULT_IDLE_TIMEOUT = 16000; //ms
     /**
-     * The default amount of time (milliseconds) after which a a sender link will be closed when no messages have been
-     * sent. The value <code>0</code> disables automatic link timeout.
+     * The default amount of time (time unit configured with {@link #setInactiveLinkTimeoutUnit(TimeUnit)}) after which
+     * a a sender link will be closed when no messages have been sent. The default value disables automatic link
```
sophokles73: a a  ...

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -764,30 +776,59 @@ public final void setIdleTimeout(final int idleTimeoutMillis) {
...
-     * The default value of this property is <code>0</code>, which disables automatic link timeout.
+     * The default value of this property is {@link #DEFAULT_INACTIVE_LINK_TIMEOUT}, which disables automatic link
+     * timeout.
      *
-     * @return The inactiveLinkTimeout in milliseconds.
+     * @return The amount of time to wait for inactive links.
+     * @see #getInactiveLinkTimeoutUnit()
      */
-    public long getInactiveLinkTimeout() {
+    public final long getInactiveLinkTimeout() {
```
sophokles73: FMPOV it would be nice to be able to get the timeout as a `java.time.Duration` which includes both, the amount as well as the unit and provides methods for getting the duration as minutes, hours, millis etc.

The setters would be used by Spring to set the values, but the Duration could then be used much easier in client code. WDYT?

site/documentation/content/admin-guide/hono-client-configuration.md
```diff
@@ -25,7 +25,8 @@ The variable names contain `${PREFIX}` as a placeholder for the particular *comm
...
-| `${PREFIX}_INACTIVE_LINK_TIMEOUT`<br>`--${prefix}.inactiveLinkTimeout` | no | `0` | Sets the amount of time in milliseconds after which a AMQP sender link will be closed when no message has been sent on it. Links will not automatically be closed if the value is `0`. |
+| `${PREFIX}_INACTIVE_LINK_TIMEOUT`<br>`--${prefix}.inactiveLinkTimeout` | no | `-1` | Sets the amount of time after which an AMQP sender link will be closed when no message has been sent on it. Links will not be closed automatically if the value is `-1`. |
+| `${PREFIX}_INACTIVE_LINK_TIMEOUT_UNIT`<br>`--${prefix}.inactiveLinkTimeoutUnit` | no | minutes | Sets the time unit for the inactive link timeout. Valid values are the enum constants of [Java's TimeUnit](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/TimeUnit.html), e.g. "HOURS". |
```
sophokles73: we should put single back quotes around the default value as we do for the other values as well.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
sophokles73: This seems to be supported out-of-the-box already in Spring Boot: https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-conversion-duration

b-abel: I like the idea. Even though I prefer to use the method `getInactiveLinkTimeoutMillis()` you proposed above.

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
b-abel: It links to the method `protected final void startAutoCloseTimer()`, which has a JavaDoc (the method `startAutoCloseTimer(final long delay)`, which takes a parameter is private).

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
b-abel: Note: now using one property for the amount of time and one for the time unit.

b-abel:
The property `inactiveLinkTimeout` in `ClientConfigProperties` is now of type `java.time.Duration`. The helm chart and the sandbox configure protocol adapters with a timeout of one hour.

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -221,4 +226,57 @@ protected static final void setApplicationProperties(final Message msg, final Ma
...
+     * Starts a timer with the delay configured by {@link ClientConfigProperties#getInactiveLinkTimeout()}. The timer
+     * checks if a message has been sent on the sender link in this period (subclass needs to call
+     * {@link #storeLastSendTime()} when sending a message). If no message has been sent in the timout period, it closes
+     * the links, otherwise a new timer is scheduled to recheck.
+     * <p>
+     * This method needs to be called by subclasses where {@link #sender} is set.
+     * <p>
+     * If {@link ClientConfigProperties#getInactiveLinkTimeout()} is zero, this method does nothing.
+     */
+    protected final void startAutoCloseTimer() {
```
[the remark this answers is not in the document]
b-abel: What about `startLinksAutoCloseTimer`?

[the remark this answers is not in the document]
calohmn: Well, `startAutoCloseLinksTimer` still sounds better in my ears (same as I would prefer e.g. "startRemoveSubscriptionsTimer" to "startSubscriptionsRemoveTimer").

b-abel:
@calohmn I added your suggestions.

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -70,6 +70,8 @@
      */
     protected List<Symbol> offeredCapabilities = Collections.emptyList();
 
+    private long autoCloseLinksTimerId = -1;
```
calohmn: I think it would be better using a `Long` here and initializing the value with `null`. Then there can be a null check added around the cancelTimer invocation and, as first thing inside the timer handler in `startAutoCloseLinksTimer`, the `autoCloseLinksTimerId` can be reset to `null`.

b-abel: > Then there can be a null check added around the cancelTimer invocation 

The reason to use `long` instead of `Long`, that I do not _need_ the `null` check :smile: (`cancelTimer` does nothing for a negative ID). But yes, I see that `null` is more obvious.

sophokles73:
@calohmn can you give this a final look and approve/merge?

Issue #1337, opened 2019-07-03
Support limiting the data volume per tenant for the remaining amount of days of the current calender month


alinaserg:
With the implementation of #929, Hono supports message limit check of data volume per time period. This time period can be defined e.g. in days.

The logic with limit check must be extended to limit the data volume not only for the time period but also on the calender month base. It means that the data volume is monthly calculated (the customer gets monthly data volume at the beginning of the month and can use it till the end of the calender month). 

The additional functionality is necessary if the customer books the service not on the first day of the calender month but e.g. in the middle of the month, then only the data volume for the remaining days of the current month are available. In this case it will be only the half of the monthly volume.

Ex.: service is booked on the 15. June, data volume for the whole month is 100 MB, till the end of the month is left 15 days, that is why the data limits must be set accordingly to the 50 MB.
With the beginning of the new calender month the data limits must be set to the limits of the whole month, it means the customer gets 100 MB of data volume.

sophokles73:
just to be clear about this: the idea is that the amount of data is always the same for each (calendar) month while the number of days over which the limit is to be imposed varies between months, right?

alinaserg:
The amount of data is always the same for each calender month, but if the service is booked not at the beginning of the calender month, but a few days later the average amount of data per day must be calculated and of course will be differently for different calender months. It means that if the service booked on the 2nd of February and on the 2nd of May the calculated remaining data volume will be a little bit different in these two cases.

sophokles73:
@kaniyan can this be closed?

kaniyan:
@sophokles73  yes.

@alinaserg This has been closed via PRs #1420 and #1491

PR #1420, opened 2019-08-01
[#1337] Message limit checks now also supports monthly calculation.

kaniyan:
This PR implements #1337. 

With this PR,  Hono now also supports _monthly_ mode of calculation. In the _monthly_ mode, the data usage is calculated from the beginning till the end of the current calendar month and compared with the limits already set using the `max-bytes` parameter. There is an exception to this calculation in the first month on which this limit becomes effective. In this case, the `max-bytes` value is recalculated based on the remaining days in the month from the `effective-since` date. The data usage for these remaining days is compared with the above recalculated value.

Below configurations illustrate how the `mode` can be set to _monthly_ or _days_.

```
"data-volume": {
        "max-bytes": 2147483648,
        "period": {
          "mode": "monthly",
        },
        "effective-since": "2019-07-27T14:30Z"
      }
```

```
"data-volume": {
        "max-bytes": 2147483648,
        "period": {
          "mode": "days",
          "no-of-days": 30
        },
        "effective-since": "2019-07-27T14:30Z"
      }
```

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
@@ -22,6 +22,7 @@ import java.time.ZoneOffset;
 import java.time.temporal.TemporalAdjusters;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.function.Supplier;
 
 import org.eclipse.hono.cache.CacheProvider;
 import org.eclipse.hono.cache.ExpiringValueCache;
@@ -218,10 +219,12 @@ public final class PrometheusBasedResourceLimitChecks implements ResourceLimitCh
         if (maxBytes == -1 || effectiveSince == null || PeriodMode.UNKNOWN.equals(periodMode) || payloadSize <= 0) {
             return Future.succeededFuture(Boolean.FALSE);
         } else {
-            final long dataUsagePeriod = calculateDataUsagePeriod(
-                    OffsetDateTime.ofInstant(effectiveSince, ZoneOffset.UTC), periodMode, periodInDays);
-            final long allowedMaxBytes = calculateDataVolume(OffsetDateTime.ofInstant(effectiveSince, ZoneOffset.UTC),
-                    periodMode, maxBytes);
+            final long allowedMaxBytes = getAndStoreInCache(limitsCache,
+                    String.format("%s_allowed_max_bytes", tenant.getTenantId()),
+                    () -> calculateDataVolume(OffsetDateTime.ofInstant(effectiveSince, ZoneOffset.UTC), periodMode, maxBytes));
+            final long dataUsagePeriod = getAndStoreInCache(limitsCache,
+                    String.format("%s_data_usage_period", tenant.getTenantId()),
+                    () -> calculateDataUsagePeriod(OffsetDateTime.ofInstant(effectiveSince, ZoneOffset.UTC), periodMode, periodInDays));
 
             if (dataUsagePeriod <= 0) {
                 return Future.succeededFuture(Boolean.FALSE);
@@ -408,6 +411,14 @@ public final class PrometheusBasedResourceLimitChecks implements ResourceLimitCh
         default:
             return 0L;
         }
+    }
+
+    private long getAndStoreInCache(final ExpiringValueCache<Object, Object> cache, final String key,
+            final Supplier<Long> valueSupplier) {
+        return Optional.ofNullable(cache)
+                .map(success -> cache.get(key))
+                .map(cachedValue -> (long) cachedValue)
+                .orElseGet(() -> addToCache(cache, key, valueSupplier.get()));
     }
 
     private long addToCache(final ExpiringValueCache<Object, Object> cache, final String key, final long result) {
--- a/service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
@@ -16,6 +16,10 @@ import static java.time.temporal.ChronoUnit.DAYS;
 
 import java.time.Duration;
 import java.time.Instant;
+import java.time.OffsetDateTime;
+import java.time.YearMonth;
+import java.time.ZoneOffset;
+import java.time.temporal.TemporalAdjusters;
 import java.util.Objects;
 import java.util.Optional;
 
@@ -54,6 +58,57 @@ public final class PrometheusBasedResourceLimitChecks implements ResourceLimitCh
     private static final Logger log = LoggerFactory.getLogger(PrometheusBasedResourceLimitChecks.class);
     private static final String QUERY_URI = "/api/v1/query";
     private static final String LIMITS_CACHE_NAME = "resource-limits";
+
+    /**
+     * The mode of the data volume calculation.
+     *
+     */
+    private enum PeriodMode {
+        /**
+         * The mode of the data volume calculation in terms of days.
+         */
+        DAYS("days"),
+        /**
+         * The mode of the data volume calculation is monthly.
+         */
+        MONTHLY("monthly"),
+        /**
+         * The unknown mode.
+         */
+        UNKNOWN("unknown");
+
+        private final String mode;
+
+        PeriodMode(final String mode) {
+            this.mode = mode;
+        }
+
+        /**
+         * Construct a PeriodMode from a given value.
+         *
+         * @param value The value from which the PeriodMode needs to be constructed.
+         * @return The PeriodMode as enum
+         */
+        static PeriodMode from(final String value) {
+            if (value != null) {
+                for (PeriodMode mode : values()) {
+                    if (value.equalsIgnoreCase(mode.value())) {
+                        return mode;
+                    }
+                }
+            }
+            return UNKNOWN;
+        }
+
+        /**
+         * Gets the string equivalent of the mode.
+         *
+         * @return The value of the mode.
+         */
+        String value() {
+            return this.mode;
+        }
+    }
 
     private final WebClient client;
     private final PrometheusBasedResourceLimitChecksConfig config;
@@ -112,6 +167,23 @@ public final class PrometheusBasedResourceLimitChecks implements ResourceLimitCh
     /**
      * Checks if the maximum data volume for the messages configured for a tenant
      * have been reached.
+     * <p>
+     * There are two supported modes of calculation which are <em>days</em> and 
+     * <em>monthly</em> and can be set using {@link TenantConstants#FIELD_PERIOD_MODE}. 
+     * <p>
+     * In the <em>monthly</em> mode, the data usage is calculated from the 
+     * beginning till the end of the current (Gregorian) calendar month and 
+     * compared with the maxBytes already set using {@link TenantConstants#FIELD_MAX_BYTES}.
+     * There is an exception to this calculation in the first month on which this
+     * limit becomes effective. In this case, the maxBytes value is recalculated 
+     * based on the remaining days in the month from the effectiveSince date.
+     * The data usage for these remaining days is then compared with the above 
+     * recalculated value.
+     * <p>
+     * If the mode is set as <em>days</em> then the data usage is calculated 
+     * based on the number of days set using {@link TenantConstants#FIELD_PERIOD_NO_OF_DAYS} 
+     * and compared with the maxBytes.
+     * 
      * @param tenant The tenant configuration to check the limit against.
      * @param payloadSize The message payload size in bytes.
      * @return A future indicating the outcome of the check.
@@ -132,21 +204,26 @@ public final class PrometheusBasedResourceLimitChecks implements ResourceLimitCh
         final DataVolume dataVolumeConfig = tenant.getResourceLimits().getDataVolume();
         final long maxBytes = dataVolumeConfig.getMaxBytes();
         final Instant effectiveSince = dataVolumeConfig.getEffectiveSince();
+        final PeriodMode periodMode = PeriodMode.from(dataVolumeConfig.getPeriod().getMode());
         final long periodInDays = Optional.ofNullable(dataVolumeConfig.getPeriod())
                 .map(DataVolumePeriod::getNoOfDays)
                 .orElse(0);
 
-        log.trace("message limit config for tenant [{}] are [{}:{}, {}:{}, {}:{}]", tenant.getTenantId(),
+        log.trace("message limit config for tenant [{}] are [{}:{}, {}:{}, {}:{}, {}:{}]", tenant.getTenantId(),
                 TenantConstants.FIELD_MAX_BYTES, maxBytes,
                 TenantConstants.FIELD_EFFECTIVE_SINCE, effectiveSince,
+                TenantConstants.FIELD_PERIOD_MODE, periodMode,
                 TenantConstants.FIELD_PERIOD_NO_OF_DAYS, periodInDays);
 
-        if (maxBytes == -1 || effectiveSince == null || periodInDays <= 0 || payloadSize <= 0) {
+        if (maxBytes == -1 || effectiveSince == null || PeriodMode.UNKNOWN.equals(periodMode) || payloadSize <= 0) {
             return Future.succeededFuture(Boolean.FALSE);
         } else {
-            final long dataUsagePeriod = calculateDataUsagePeriod(effectiveSince, periodInDays);
+            final long dataUsagePeriod = calculateDataUsagePeriod(
+                    OffsetDateTime.ofInstant(effectiveSince, ZoneOffset.UTC), periodMode, periodInDays);
+            final long allowedMaxBytes = calculateDataVolume(OffsetDateTime.ofInstant(effectiveSince, ZoneOffset.UTC),
+                    periodMode, maxBytes);
 
-            if (dataUsagePeriod <= 0) {
+            if (dataUsagePeriod <= 0 || allowedMaxBytes <= 0) {
                 return Future.succeededFuture(Boolean.FALSE);
             }
 
@@ -172,13 +249,14 @@ public final class PrometheusBasedResourceLimitChecks implements ResourceLimitCh
                     .orElseGet(() -> executeQuery(queryParams)
                             .map(bytesConsumed -> addToCache(limitsCache, key, bytesConsumed)))
                     .map(bytesConsumed -> {
-                        if ((bytesConsumed + payloadSize) <= maxBytes) {
+                        if ((bytesConsumed + payloadSize) <= allowedMaxBytes) {
                             return Boolean.FALSE;
                         } else {
-                            log.trace("data limit exceeded for tenant. [{}] are [bytesUsed:{}, {}:{}, {}:{}, {}:{}]",
-                                    tenant.getTenantId(), bytesConsumed, TenantConstants.FIELD_MAX_BYTES,
-                                    maxBytes,
+                            log.trace(
+                                    "data limit exceeded for tenant. [{}] are [bytesUsed:{}, allowed max-bytes:{}, {}:{}, {}:{}, {}:{}]",
+                                    tenant.getTenantId(), bytesConsumed, allowedMaxBytes,
                                     TenantConstants.FIELD_EFFECTIVE_SINCE, effectiveSince,
+                                    TenantConstants.FIELD_PERIOD_MODE, periodMode,
                                     TenantConstants.FIELD_PERIOD_NO_OF_DAYS, periodInDays);
                             return Boolean.TRUE;
                         }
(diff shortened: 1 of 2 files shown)


service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -237,34 +267,55 @@ long getMaximumNumberOfBytes(final TenantObject tenant) {
...
+     * @return The number of days for which the data usage is calculated or 0 if not set.
      */
     long getPeriodInDays(final TenantObject tenant) {
         return Optional.ofNullable(tenant.getResourceLimits())
                 .map(limits -> limits.getJsonObject(FIELD_DATA_VOLUME))
-                .map(dataVolumeLimits -> dataVolumeLimits.getLong(FIELD_PERIOD_IN_DAYS, DEFAULT_PERIOD_IN_DAYS))
-                .orElse(DEFAULT_PERIOD_IN_DAYS);
+                .map(dataVolumeLimits -> dataVolumeLimits.getJsonObject(FIELD_PERIOD))
+                .map(noOfDays -> noOfDays.getLong(FIELD_PERIOD_NO_OF_DAYS))
+                .orElse(0L);
```
sophokles73: this should probably also go into `TenantObject` ...

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -237,34 +267,55 @@ long getMaximumNumberOfBytes(final TenantObject tenant) {
...
      * The date string should comply to the {@link DateTimeFormatter#ISO_OFFSET_DATE_TIME}
      *
      * @param tenant The tenant configuration object.
-     * @return The instant on which the data volume limit came into effective or {@code null} if not set.
+     * @return The offsetDateTime on which the data volume limit came into effective or {@code null} if not set.
      * @throws DateTimeParseException if the date string fails to comply with the
      *             {@link DateTimeFormatter#ISO_OFFSET_DATE_TIME} format.
      */
-    Instant getEffectiveSince(final TenantObject tenant) {
+    OffsetDateTime getEffectiveSince(final TenantObject tenant) {
```
sophokles73: and this as well ...

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -148,6 +157,23 @@ public PrometheusBasedResourceLimitChecks(final WebClient webClient,
     /**
      * Checks if the maximum data volume for the messages configured for a tenant
      * have been reached.
+     * <p>
+     * The are two supported modes of calculation which are <em>days</em> and
```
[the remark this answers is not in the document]
kaniyan: Typo I will change it.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -237,34 +267,55 @@ long getMaximumNumberOfBytes(final TenantObject tenant) {
     }
 
     /**
-     * Gets the number of days for which the data usage is calculated and compared against the
-     * {@link #getMaximumNumberOfBytes(TenantObject)}.
+     * Gets the mode of the calculation. The supported modes are <em>days</em> and <em>monthly</em>.
      *
      * @param tenant The tenant configuration object.
-     * @return The number of days for which the data usage is calculated or {@link #DEFAULT_PERIOD_IN_DAYS} if not set.
+     * @return The mode of the calculation or {@code null} if not set.
```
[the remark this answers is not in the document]
kaniyan: I remember in the past we decided to keep it away from the `TenantObject`. `TenantObject` offers methods to set and get the `resource-limits` and it doesn't worry what is inside the `resource-limits` configuration. IMHO the advantage of this approach is, that any custom implementations of the `ResourceLimitChecks` are now free to decide the configuration parameters based on their needs. May be a custom implementation doesn't need any of the configuration parameters that we have defined already.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -148,6 +157,23 @@ public PrometheusBasedResourceLimitChecks(final WebClient webClient,
     /**
      * Checks if the maximum data volume for the messages configured for a tenant
      * have been reached.
+     * <p>
+     * The are two supported modes of calculation which are <em>days</em> and 
+     * <em>monthly</em> and can be set using {@link #FIELD_PERIOD_MODE}. 
+     * <p>
+     * In the <em>monthly</em> mode, the data usage is calculated from the 
+     * beginning till the end of the current calendar month and compared with
```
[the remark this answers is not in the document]
kaniyan: Done and I have pushed the changes.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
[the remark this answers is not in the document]
sophokles73: Hmmm, that makes sense to some degree. However, the Tenant API already defines most of the particular properties that we use for enforcing message limits. If we want to follow the approach you outlined above, then IMHO we would need to remove these properties from the Tenant API spec and just mention, that the Tenant API may return additional properties that can be used by the adapters for such purposes.

However, since we have implemented message limit enforcement as a standard functionality in the adapters, I do not see why we wouldn't/shouldn't explicitly define and explain these parameters. Or am I mistaken? Otherwise, where would be the right place to document the parameters expected by the adapters?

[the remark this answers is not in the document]
kaniyan: > Hmmm, that makes sense to some degree. However, the Tenant API already defines most of the particular properties that we use for enforcing message limits. If we want to follow the approach you outlined above, then IMHO we would need to remove these properties from the Tenant API spec and just mention, that the Tenant API may return additional properties that can be used by the adapters for such purposes.

> 
> However, since we have implemented message limit enforcement as a standard functionality in the adapters, I do not see why we wouldn't/shouldn't explicitly define and explain these parameters. Or am I mistaken? Otherwise, where would be the right place to document the parameters expected by the adapters?

The limit enforcement in the adapters just  take _tenantObject_ as a parameter and doesn't worry about what is inside. What lies inside is taken care by the underlying `ResourceLimitChecks` implementation. I would personally prefer removing these properties from TenantAPI spec and delegate these to the `ResourceLimitChecks` implementation. On the other hand, I don't see any other better place other than the TenantAPI to document.

IMHO it is also acceptable to keep these parameters  in Tenant API provided the _resource-limits_ section in the API offers provision to set and return addition properties which can be used by any custom `ResourceLimitChecks` implementation.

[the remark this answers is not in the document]
sophokles73: > IMHO it is also acceptable to keep these parameters in Tenant API provided the resource-limits section in the API offers provision to set and return addition properties which can be used by any custom ResourceLimitChecks implementation.

I think so too. As a consequence, we could then indeed move the methods to `TenantObject`, as these properties are now *officially* required/supported, right?

The Management API already supports adding arbitrary additional properties by means of the *ext* field. We would then simply add that field to the Tenant API (and `TenantObject`) as well.  WDYT?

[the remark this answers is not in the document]
kaniyan: > I think so too. As a consequence, we could then indeed move the methods to `TenantObject`, as these properties are now _officially_ required/supported, right?
> 
> The Management API already supports adding arbitrary additional properties by means of the _ext_ field. We would then simply add that field to the Tenant API (and `TenantObject`) as well. WDYT?

It sounds good. I will create an issue for the support of `extensions` in the Tenant API (`TenantObject`) and suggest a separate PR for that.

To continue with this PR, I would suggest moving the above mentioned methods to the `TenantObject`.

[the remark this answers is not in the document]
kaniyan: @sophokles73 I have now added those methods to the `TenantObject`. Also updated the  Device Management API and Device Registry implementation where the `data-volume` configurations were missing.

sophokles73:
It feels very strange that we are not using the same value objects for those data structures that we actually *want* to be shared between the Management API and Tenant API. For example, I believe that we should explicitly use the `ResourceLimits` value object in both the `TenantObject` as well as the `Tenant` object, in order to make sure that we use the same names for the properties and use the same structure. In that case, we deliberately define this data structure to be shared between the APIs, right?

I know that I insisted in considering the two APIs as independent from each other (and I still do). However, if we define the *embedded* data structures to be shared, then I think that we should also share the value object classes between the API implementations.

@ctron @jbtrystram WDYT?

calohmn:
An additional value object that I would also want to be shared between Management and Tenant API is the `TenantTracingConfig`, as proposed in https://github.com/eclipse/hono/pull/1425/files#diff-54f99f84bf5f05f9bdb138880794685b.

kaniyan:
Then I will share the `ResourceLimits` value objects between the `TenantObject` and `Tenant`.

kaniyan:
> Then I will share the `ResourceLimits` value objects between the `TenantObject` and `Tenant`.

@sophokles73 I have pushed the additional changes to explicitly use the `ResourceLimits` value objects in both the `TenantObject` as well as the `Tenant` object. Would you mind taking a look?

sophokles73:
@ctron are you also ok with sharing the *low-level* value objects between the APIs?

core/src/main/java/org/eclipse/hono/util/TenantDataVolume.java
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * Sets the period for the data usage calculation.
+     *
+     * @param period The period for the data usage calculation.
+     */
+    public void setPeriod(final TenantDataVolumePeriod period) {
+        this.period = period;
+    }
+
+    private OffsetDateTime parse(final String timestamp) {
```
ctron: I would strongly suggest to **not** use offsetted DateTime. This will only cause problems unless you have a real reason, which in most cases only exists using this in UI.

What you probably want is `Instant`, which is a specific point in time.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -617,4 +637,17 @@ public OptionalInt getTraceSamplingPriority(final String authId) {
         }
         return TenantConstants.TraceSamplingMode.from(traceSamplingMode).toSamplingPriority();
     }
+
+    private OffsetDateTime parse(final String timestamp) {
```
ctron: Again, try to use `Instant`.

core/src/main/java/org/eclipse/hono/util/TenantDataVolume.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.Optional;
+
+/**
+ * Data volume definition of the tenant resource limits.
+ */
+public class TenantDataVolume {
+
+    private OffsetDateTime effectiveSince;
+
+    @JsonProperty(RegistryManagementConstants.FIELD_MAX_BYTES)
```
sophokles73: this should be used from `TenantConstants` FMPOV as we define the type in the scope of the `TenantObject` ...

core/src/main/java/org/eclipse/hono/util/TenantDataVolume.java
```diff
@@ -0,0 +1,126 @@
...
+    private long maxBytes = -1;
+
+    @JsonProperty(RegistryManagementConstants.FIELD_PERIOD)
```
sophokles73: this should be used from `TenantConstants` FMPOV as we define the type in the scope of the `TenantObject` ...

sophokles73:
@kaniyan IMHO it would be easier, if you could extract the changes necessary to add the data volume value object to the Tenant and Management APIs and put them into a separate PR addressing #1146 
Once these changes are in place, you can rebase this PR on top of it. WDYT?

kaniyan:
> @kaniyan IMHO it would be easier, if you could extract the changes necessary to add the data volume value object to the Tenant and Management APIs and put them into a separate PR addressing #1146
> Once these changes are in place, you can rebase this PR on top of it. WDYT?

@sophokles73 It sounds good. I will extract the data volume part involving Tenant and Management APIS into a  separate PR for the issue #1460

core/src/main/java/org/eclipse/hono/util/TenantDataVolume.java
kaniyan: I will change it to use `Instant`.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
kaniyan: I will change it to use `Instant`.

kaniyan:
> > @kaniyan IMHO it would be easier, if you could extract the changes necessary to add the data volume value object to the Tenant and Management APIs and put them into a separate PR addressing #1146
> > Once these changes are in place, you can rebase this PR on top of it. WDYT?
> 
> @sophokles73 It sounds good. I will extract the data volume part involving Tenant and Management APIS into a separate PR for the issue #1460

@sophokles73 I have pushed the changes pertaining to the monthly mode of message limit checks. Would you mind taking a look.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -261,19 +339,74 @@ private Long extractLongValue(final JsonObject response) {
...
+     * @param maxBytes The maximum allowed bytes defined in configuration by 
+     *                 {@link TenantConstants#FIELD_MAX_BYTES}. 
+     * @return The allowed data-volume in bytes.
      */
-    private long calculateDataUsagePeriod(final Instant effectiveSince, final long periodInDays) {
-        final long inclusiveDaysBetween = DAYS.between(effectiveSince, Instant.now()) + 1;
-        if (inclusiveDaysBetween > 0 && periodInDays > 0) {
-            final long dataUsagePeriodInDays = inclusiveDaysBetween % periodInDays;
-            return dataUsagePeriodInDays == 0 ? periodInDays : dataUsagePeriodInDays;
+    private long calculateDataVolume(final OffsetDateTime effectiveSince, final PeriodMode mode,
```
sophokles73: I think we should accept an `Instant` instead of an `OffsetDateTime` here ..

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -261,19 +339,74 @@ private Long extractLongValue(final JsonObject response) {
...
+     *
+     * @param effectiveSince The point of time on which the data volume limit came
+     *                      into effect.
+     * @param mode The mode of the period defined by
+     *                   {@link TenantConstants#FIELD_PERIOD_MODE}.
+     * @param periodInDays The number of days defined by 
+     *                     {@link TenantConstants#FIELD_PERIOD_NO_OF_DAYS}. 
+     * @return The period in days for which the data usage is to be calculated.
+     */
+    private long calculateDataUsagePeriod(final OffsetDateTime effectiveSince, final PeriodMode mode,
```
sophokles73: I think we should accept an `Instant` instead of an `OffsetDateTime` here ..

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -172,13 +249,14 @@ public PrometheusBasedResourceLimitChecks(final WebClient webClient,
...
                     .map(bytesConsumed -> {
-                        if ((bytesConsumed + payloadSize) <= maxBytes) {
+                        if ((bytesConsumed + payloadSize) <= allowedMaxBytes) {
                             return Boolean.FALSE;
                         } else {
-                            log.trace("data limit exceeded for tenant. [{}] are [bytesUsed:{}, {}:{}, {}:{}, {}:{}]",
-                                    tenant.getTenantId(), bytesConsumed, TenantConstants.FIELD_MAX_BYTES,
-                                    maxBytes,
+                            log.trace(
+                                    "data limit exceeded for tenant. [{}] are [bytesUsed:{}, allowed max-bytes:{}, {}:{}, {}:{}, {}:{}]",
```
sophokles73: this is no english sentence

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -261,19 +339,74 @@ private Long extractLongValue(final JsonObject response) {
...
-    private long calculateDataUsagePeriod(final Instant effectiveSince, final long periodInDays) {
-        final long inclusiveDaysBetween = DAYS.between(effectiveSince, Instant.now()) + 1;
-        if (inclusiveDaysBetween > 0 && periodInDays > 0) {
-            final long dataUsagePeriodInDays = inclusiveDaysBetween % periodInDays;
-            return dataUsagePeriodInDays == 0 ? periodInDays : dataUsagePeriodInDays;
+    private long calculateDataVolume(final OffsetDateTime effectiveSince, final PeriodMode mode,
+            final long maxBytes) {
+        if (PeriodMode.MONTHLY.equals(mode)
+                && maxBytes > 0
+                && OffsetDateTime.now(effectiveSince.getOffset()).compareTo(effectiveSince) >= 0
```
sophokles73: `!OffsetDateTime.now().isBefore(effectiveSince)`?

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -132,21 +204,26 @@ public PrometheusBasedResourceLimitChecks(final WebClient webClient,
...
             return Future.succeededFuture(Boolean.FALSE);
         } else {
-            final long dataUsagePeriod = calculateDataUsagePeriod(effectiveSince, periodInDays);
+            final long dataUsagePeriod = calculateDataUsagePeriod(
+                    OffsetDateTime.ofInstant(effectiveSince, ZoneOffset.UTC), periodMode, periodInDays);
+            final long allowedMaxBytes = calculateDataVolume(OffsetDateTime.ofInstant(effectiveSince, ZoneOffset.UTC),
+                    periodMode, maxBytes);
 
-            if (dataUsagePeriod <= 0) {
+            if (dataUsagePeriod <= 0 || allowedMaxBytes <= 0) {
```
sophokles73: why isn't the limit reached if allowed max bytes is 0? To me, the value suggests: no data is allowed at all. Or am I mistaken?

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -261,19 +339,74 @@ private Long extractLongValue(final JsonObject response) {
...
+     *                   {@link TenantConstants#FIELD_PERIOD_MODE}.
+     * @param periodInDays The number of days defined by 
+     *                     {@link TenantConstants#FIELD_PERIOD_NO_OF_DAYS}. 
+     * @return The period in days for which the data usage is to be calculated.
+     */
+    private long calculateDataUsagePeriod(final OffsetDateTime effectiveSince, final PeriodMode mode,
+            final long periodInDays) {
+        final long inclusiveDaysBetween = DAYS.between(effectiveSince,
+                OffsetDateTime.now(effectiveSince.getOffset())) + 1;
+        if (PeriodMode.DAYS.equals(mode)) {
```
sophokles73: use a switch statement?

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
kaniyan: That was a typing mistake, I will update it.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
kaniyan: Since we don't serialize/de-serialize here, I thought its fine to use the `OffsetDateTime`. `OffsetDateTime` comes with functions like `getDayOfMonth` which is quite handy to perform calculations.

sophokles73: `Instant.get(ChronoField.DAY_OF_MONTH)`

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
kaniyan: If it is zero then it is interpreted as no data is allowed and this check I will remove it.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
kaniyan: `Instant.now().get(ChronoField.DAY_OF_MONTH)` throws
`Exception in thread "main" java.time.temporal.UnsupportedTemporalTypeException: Unsupported field: DayOfMonth`. It also the same with YearMonth.from(Instant.now()).

It seems a zone or offset is needed for it as below:
`Instant.now().atOffset(ZoneOffset.UTC).get(ChronoField.DAY_OF_MONTH)`

kaniyan: Instead of converting `Instant` to offset during every such cases as above, I would prefer leaving it as `OffsetDateTime` instead of `Instant`. WDYT?

calohmn: IMHO, using `OffsetDateTime` looks ok here. Quoting the `OffsetDateTime` javadoc:
`OffsetDateTime adds to the instant the offset from UTC/Greenwich, which allows the local date-time to be obtained.` And by means of the local date, the day of month can be obtained (via `OffsetDateTime.get(ChronoField.DAY_OF_MONTH)` or `OffsetDateTime.getDayOfMonth()`).

kaniyan: > IMHO, using `OffsetDateTime` looks ok here. Quoting the `OffsetDateTime` javadoc:
> `OffsetDateTime adds to the instant the offset from UTC/Greenwich, which allows the local date-time to be obtained.` And by means of the local date, the day of month can be obtained (via `OffsetDateTime.get(ChronoField.DAY_OF_MONTH)` or `OffsetDateTime.getDayOfMonth()`).

:+1: I have pushed changes corresponding to the other review comments and also changes to store the calculated max bytes and data volume period in a cache instead of calculating every time.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
```diff
@@ -261,19 +342,83 @@ private Long extractLongValue(final JsonObject response) {
...
+            if (YearMonth.now().equals(YearMonth.from(effectiveSince)) && effectiveSince.getDayOfMonth() != 1) {
+                return inclusiveDaysBetween;
+            }
+            return OffsetDateTime.now(effectiveSince.getOffset()).getDayOfMonth();
+        default:
+            return 0L;
+        }
+    }
+
+    private long getAndStoreInCache(final ExpiringValueCache<Object, Object> cache, final String key,
```
calohmn: How about naming the method `getOrAddToCache` instead?

calohmn:
I think it would be good to add unit tests for the `calculateDataVolume()` and `calculateDataUsagePeriod()` methods, also to document what the special cases are there.

service-base/src/main/java/org/eclipse/hono/service/plan/PrometheusBasedResourceLimitChecks.java
kaniyan: I also had that option in mind and somehow ended up with the current one with the intention of _get/calculate_ the value and store in the cache. `getOrAddToCache` also sound good.

kaniyan:
> I think it would be good to add unit tests for the `calculateDataVolume()` and `calculateDataUsagePeriod()` methods, also to document what the special cases are there.

@calohmn I have pushed the requested changes with more unit tests. A separate PR is planned next to explain this monthly feature in the documentation.

PR #1425, opened 2019-08-02
[#1406] Add trace sampling mode properties in the Management API spec

### LINKED ISSUE #1406: Support configuration of forced tracing via Device Registry Management API
I think this needs to be added to the Management API spec and we need to make sure that the example Device Registry supports it.

calohmn:
This fixes #1406.


calohmn:
I've just seen that the `Tenant` class needs to be updated as well. Will update this PR accordingly...

calohmn:
I've amended the commit and added support for the properties in the management API `Tenant` class.
I've also put the 2 properties under a common "tracing" property:
````
  {
    "enabled": true,
    "tracing": {
      "sampling-mode": "all",
      "sampling-mode-per-auth-id": {
        "4711": "all"
      }
    }
  }
````

PR #1432, opened 2019-08-07
Refactor AmqpAdapterSaslAuthenticatorFactory

calohmn:
This is a preparation for #1396, making it easier to intercept SASL Authentication credentials once they have been parsed.

### DIFF
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorFactory.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorFactory.java
@@ -51,7 +51,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.opentracing.Span;
-import io.vertx.core.AsyncResult;
 import io.vertx.core.CompositeFuture;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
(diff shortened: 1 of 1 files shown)


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorFactory.java
```diff
@@ -222,97 +232,86 @@ public void process(final Handler<Boolean> completionHandler) {
...
 
         @Override
         public boolean succeeded() {
             return succeeded;
         }
 
-        private void verifyPlain(final byte[] saslResponse, final Handler<AsyncResult<DeviceUser>> completer) {
-
-            currentSpan.log("authenticating device using SASL PLAIN");
+        private Future<String[]> parseSaslResponse(final byte[] saslResponse) {
```
sophokles73: why not return `Future<UsernamePasswordCredentials>` instead? IMHO that would be make it easier to understand the semantics of the values returned in the future ...

calohmn: I'd rather keep all the information from the `saslResponse` in the returned object here still (including the `authzid`).

The `VertxBasedAmqpProtocolAdapter` allows for a custom `authenticatorFactory` to be set but still makes quite some assumptions on what the `authenticatorFactory`'s `ProtonSaslAuthenticator` ought to do (e.g. setting certain connection attachments). Therefore I'm thinking towards creating an AuthenticatorFactory base class that provides methods for fulfilling these assumptions while staying rather generic, hence the intention to stay low-level with the sasl response parsing here.

sophokles73: I see

PR #1433, opened 2019-08-07
Drop the unnecessary type argument to event bus classes

ctron:
This change drops inheritance on `ConfigurationSupportingVerticle` as most classes implement this now with `<Void>`.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/credentials/BaseCredentialsService.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/credentials/BaseCredentialsService.java
@@ -35,8 +35,10 @@ import io.vertx.core.json.JsonObject;
  * @deprecated - Use {@link CredentialsService} and {@link org.eclipse.hono.service.management.credentials.CredentialsManagementService} instead.
  */
 @Deprecated
-public abstract class BaseCredentialsService<T> extends EventBusCredentialsAdapter<T>
+public abstract class BaseCredentialsService<T> extends EventBusCredentialsAdapter
         implements CredentialsService {
+
+    private T config;
 
     @Override
     protected CredentialsService getService() {
@@ -94,5 +96,36 @@ public abstract class BaseCredentialsService<T> extends EventBusCredentialsAdapt
      */
     protected void handleUnimplementedOperation(final Handler<AsyncResult<CredentialsResult<JsonObject>>> resultHandler) {
         resultHandler.handle(Future.succeededFuture(CredentialsResult.from(HttpURLConnection.HTTP_NOT_IMPLEMENTED)));
+    }
+
+    /**
+     * Sets the specific object instance to use for configuring this <em>Verticle</em>.
+     * 
+     * @param props The properties.
+     */
+    protected final void setSpecificConfig(final T props) {
+        this.config = props;
+    }
+
+    /**
+     * Sets the properties to use for configuring this <em>Verticle</em>.
+     * <p>
+     * Subclasses <em>must</em> invoke {@link #setSpecificConfig(Object)} with the configuration object.
+     * <p>
+     * This method mainly exists so that subclasses can annotate its concrete implementation with Spring annotations
+     * like {@code Autowired} and/or {@code Qualifier} to get injected a particular bean instance.
+     * 
+     * @param configuration The configuration properties.
+     * @throws NullPointerException if configuration is {@code null}.
+     */
+    public abstract void setConfig(T configuration);
+
+    /**
+     * Gets the properties that this <em>Verticle</em> has been configured with.
+     * 
+     * @return The properties or {@code null} if not set.
+     */
+    public final T getConfig() {
+        return this.config;
     }
 }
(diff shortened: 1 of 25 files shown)


sophokles73:
I am not sure if I get the purpose of this, but it looks like we will have a lot more (duplicate) code than before the change ...

ctron:
> I am not sure if I get the purpose of this, but it looks like we will have a lot more (duplicate) code than before the change ...

If you look at the deprecated classes, then yes. However that is due to the fact that Java doesn't support multiple inheritance.

If you ignore the deprecated classes, then change drops the "setConfig" concept, which gets used by all of those classes only with a `Void` type argument. As none of those classes make use of this construct.

So for the moment we do have a bit of code duplication, but only in the deprecated classes. Assuming the go away pretty soon, I think it is a good compromise.

service-base/src/main/java/org/eclipse/hono/service/credentials/BaseCredentialsService.java
```diff
@@ -95,4 +97,35 @@ public abstract void getAll(String tenantId, String deviceId, Span span,
...
+        this.config = props;
+    }
+
+    /**
+     * Sets the properties to use for configuring this <em>Verticle</em>.
+     * <p>
+     * Subclasses <em>must</em> invoke {@link #setSpecificConfig(Object)} with the configuration object.
+     * <p>
+     * This method mainly exists so that subclasses can annotate its concrete implementation with Spring annotations
+     * like @c{@code Autowired} and/or {@code Qualifier} to get injected a particular bean instance.
```
sophokles73: typo

service-base/src/main/java/org/eclipse/hono/service/credentials/CompleteBaseCredentialsService.java
```diff
@@ -119,4 +121,35 @@ protected void handleUnimplementedOperation(
...
+        this.config = props;
+    }
+
+    /**
+     * Sets the properties to use for configuring this <em>Verticle</em>.
+     * <p>
+     * Subclasses <em>must</em> invoke {@link #setSpecificConfig(Object)} with the configuration object.
+     * <p>
+     * This method mainly exists so that subclasses can annotate its concrete implementation with Spring annotations
+     * like @c{@code Autowired} and/or {@code Qualifier} to get injected a particular bean instance.
```
sophokles73: typo

service-base/src/main/java/org/eclipse/hono/service/deviceconnection/BaseDeviceConnectionService.java
```diff
@@ -84,4 +86,35 @@ public void getLastKnownGatewayForDevice(final String tenantId, final String dev
...
+        this.config = props;
+    }
+
+    /**
+     * Sets the properties to use for configuring this <em>Verticle</em>.
+     * <p>
+     * Subclasses <em>must</em> invoke {@link #setSpecificConfig(Object)} with the configuration object.
+     * <p>
+     * This method mainly exists so that subclasses can annotate its concrete implementation with Spring annotations
+     * like @c{@code Autowired} and/or {@code Qualifier} to get injected a particular bean instance.
```
sophokles73: typo

service-base/src/main/java/org/eclipse/hono/service/registration/BaseRegistrationService.java
```diff
@@ -124,4 +126,35 @@ protected final JsonObject getAssertionPayload(final String tenantId, final Stri
...
+        this.config = props;
+    }
+
+    /**
+     * Sets the properties to use for configuring this <em>Verticle</em>.
+     * <p>
+     * Subclasses <em>must</em> invoke {@link #setSpecificConfig(Object)} with the configuration object.
+     * <p>
+     * This method mainly exists so that subclasses can annotate its concrete implementation with Spring annotations
+     * like @c{@code Autowired} and/or {@code Qualifier} to get injected a particular bean instance.
```
sophokles73: typo

service-base/src/main/java/org/eclipse/hono/service/tenant/BaseTenantService.java
```diff
@@ -89,4 +91,35 @@ public void get(final X500Principal subjectDn, final Span span,
...
+        this.config = props;
+    }
+
+    /**
+     * Sets the properties to use for configuring this <em>Verticle</em>.
+     * <p>
+     * Subclasses <em>must</em> invoke {@link #setSpecificConfig(Object)} with the configuration object.
+     * <p>
+     * This method mainly exists so that subclasses can annotate its concrete implementation with Spring annotations
+     * like @c{@code Autowired} and/or {@code Qualifier} to get injected a particular bean instance.
```
sophokles73: typo

service-base/src/test/java/org/eclipse/hono/service/tenant/BaseTenantServiceTest.java
```diff
@@ -147,4 +147,5 @@ public void setConfig(final ServiceConfigProperties configuration) {
             }
         };
     }
+
```
sophokles73: ?

service-base/src/main/java/org/eclipse/hono/service/credentials/BaseCredentialsService.java
ctron: That typo was already present in the original source. I fixed it there, as well all in the copied locations.

service-base/src/test/java/org/eclipse/hono/service/tenant/BaseTenantServiceTest.java
ctron: I assume you refer to the empty line, I removed it.

PR #1439, opened 2019-08-09
Data Grid based Device Connection service

sophokles73:
Adds an Infinispan Server based implementation of the Device Connection API.
The Helm chart can be configured to deploy an example data grid and the service instead of using the simple implementation of the example Device Registry.


site/documentation/content/deployment/helm-based-deployment.md
```diff
@@ -213,6 +213,35 @@ The newly built images can then be deployed using Helm:
 helm install --dep-up --name hono --namespace hono target/deploy/helm/
 ~~~
 
+## Using the Device Connection Service
+
+Hono's example Device Registry component contains a simple in-memory implementation of the [Device Connection API]({{< relref "/api/device-connection" >}}).
+This example implementation is used by default when the example registry is deployed.
+
+Hono also contains a production ready, data grid based implementation of the Device Connection API which can be deployed and used instead of
+the example implementation. The component can be deployed by means of setting the *deviceConnectionService.enable* property to `true` when
```
calohmn: *deviceConnectionService.enable*: trailing 'd' missing here.

deploy/src/main/deploy/helm/templates/hono-service-device-connection/hono-service-device-connection-deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      {{- include "hono.matchLabels" $args | nindent 6 }}
+  template:
+    metadata:
+      {{- include "hono.metadata" $args | nindent 6 }}
+    spec:
+      containers:
```
calohmn: How about adding tracing support by including
````
{{- include "hono.jaeger.agent" . | indent 6 }}
````
here and 
````
{{- include "hono.jaeger.clientConf" $args | indent 8 }}
````
below (as part of the `env`)  block. 
This also prevents the warning on startup:
````
12:35:42.192 [main] WARN  i.o.c.tracerresolver.TracerResolver - Error getting tracer using io.jaegertracing.tracerresolver.internal.JaegerTracerFactory@1eb6749b: Service name must not be null or empty
java.lang.IllegalArgumentException: Service name must not be null or empty
  at io.jaegertracing.internal.JaegerTracer$Builder.checkValidServiceName(JaegerTracer.java:653)
````

File: services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java [hunk 3]
```diff
@@ -0,0 +1,217 @@
...
+            final String deviceId,
+            final String gatewayId,
+            final Span span,
+            final Handler<AsyncResult<DeviceConnectionResult>> resultHandler) {
+
+        if (cache == null) {
+            resultHandler.handle(Future.failedFuture(new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "no connection to remote cache")));
+        } else {
+            cache
+            .putAsync(getKey(tenantId, deviceId), gatewayId)
```
ctron: I think it would make sense to set a TTL here. And allow for automatic cache cleanup.

sophokles73:
> So, I think the solution could either be to implement the liveness check or, as the better solution IMHO, to implement a retry (with exponential backoff) around the cache startup logic where the data-grid connection is made, like we do in the other hono components.
(Using the Infinispan maxRetries doesn't look appropriate to me because there doesn't seem to be a delay between connection attempts.)
WDYT?

Yes, I have experienced similar issues but hoped that 100 retries would do the trick in any case. The Infinispan client API is quite limited when it comes to hooking into the connection lifecycle, so I tend to favor the approach implementing the liveness check and let the service restart if the connection is not established/dropped. It is a little more *aggressive* but should work for the time being.

deploy/src/main/deploy/helm/templates/hono-service-device-connection/hono-service-device-connection-deployment.yaml
sophokles73: absolutely, will do

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
sophokles73: @calohmn and I discussed using an expiring cache as well but in the end thought that the data itself will be valid (and useful) until being updated. Based on that, the data does not seem to be of an ephemeral nature really. This is also one of the reasons for using a distributed/sharded cache which can scale out horizontally, i.e. without the need to keep all data on each node of the cluster (fully replicated).
However, what would be a reasonable value for a TTL from your point of view?

sophokles73:
@calohmn I have pushed some changes ...

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
ctron: I think that you need some way to clean up the cache eventually. The entry is no longer valid, when the device is no longer registered with the system. Whatever the reason for that is.

You can create some elaborate cleanup mechanism, or simply set a TTL of e.g. 1 month.

sophokles73: I agree that the entry becomes irrelevant once the device is removed from the system and I do agree that implementing a sophisticated mechanism for removing entries for deleted devices would be a little over the top for now. So let's start with a TTL of 1 month and see how we fare ...

sophokles73: I have pushed a corresponding change ...

calohmn:
> Yes, I have experienced similar issues but hoped that 100 retries would do the trick in any case. The Infinispan client API is quite limited when it comes to hooking into the connection lifecycle, so I tend to favor the approach implementing the liveness check and let the service restart if the connection is not established/dropped. It is a little more aggressive but should work for the time being.

Yes, ok.

sophokles73:
@calohmn there also seems to have been a bug in the example data grid config which required 2 instances to be up and running to form the cluster. However, the deployment resource only started 1 instance by default. so I think that this has delayed startup of the data grid as well. I have now reduced the required number of cache instances to 1. I hope that this also helps with startup time ...

calohmn:
@sophokles73 Yes, it helps. I now get the cluster fully ready with just one restart of the device-connection container, which is ok, I guess.

sophokles73:
@ctron @calohmn anything else we need to change?

calohmn:
@sophokles73 not as far as I'm concerned.

PR #1442, opened 2019-08-13
Fix webchat link.

b-abel:
Freenode had changed their web chat software. This fix adapts the link
so that the channel is again configured correctly in the login form.


calohmn:
Well spotted. I think we could add something like `?nick=guest?` to have the equivalent of the previous `randomnick=1`. WDYT?

b-abel:
> we could add something like `?nick=guest?`

Or `?nick=honofan?`? :laughing:

b-abel:
Done.

PR #1447, opened 2019-08-16
Give read-only access to the application properties.

ctron:
This is necessary in order to allow making a copy of the object.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/RequestResponseResult.java
+++ b/core/src/main/java/org/eclipse/hono/util/RequestResponseResult.java
@@ -104,6 +104,15 @@ public class RequestResponseResult<T> {
     }
 
     /**
+     * Gets read-only access to the response message's <em>application-properties</em>.
+     * 
+     * @return The unmodifiable map of the application properties. Never returns {@code null}.
+     */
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/eclipse/hono/util/RequestResponseResult.java
```diff
@@ -103,6 +103,15 @@ public final CacheDirective getCacheDirective() {
         }
     }
 
+    /**
+     * Gets access read-only access to the response message's <em>application-properties</em>.
```
sophokles73: Gets read-only access ...

core/src/main/java/org/eclipse/hono/util/RequestResponseResult.java
```diff
@@ -103,6 +103,15 @@ public final CacheDirective getCacheDirective() {
...
+     * 
+     * @return The unmodifiable map of the application properties. Never returns {@code null}.
+     */
+    public Map<String, Object> getApplicationProperties() {
```
sophokles73: FMPOV this should be final as all the other accessors as there is no obvious reason for it to be overridden, is there?

PR #1452, opened 2019-08-20
Align image tags with helm deployment and registry deployment improvements.

kaizimmerm:
I added a few improvements for AKS or k8s deployments in general.

* Aligned image tagging between build and (helm) deployment, i.e.
  * Align to set a custom registry (which was possible to deploy in with Helm but not to tag in the build).
  * Align to set a custom org (which was possible tag in build but not to deploy with Helm)
* Registry deployment improvements:
  * Add Azure AKS docs how to set storage class with retain policy.
  * Switch from k8s `deployment` to `StatefulSet` in order to avoid deadlock situations at update/redeployment time due to the `ReadWriteOnce` access.
  * Allow for registry to configure a custom storage class and disable the example storage override at container startup.
* Fix helm folder according to helm spec (i.e. needs to be identical to chart name for instance to allow `helm package`).

I apologise that my editor removed a few whitespaces.

Kai

### DIFF
--- a/deploy/pom.xml
+++ b/deploy/pom.xml
@@ -71,13 +71,13 @@
 
     <plugins>
       <plugin>
-        <!-- Copy legal documents from "legal" module to "target/classes" folder so that we make sure to include 
+        <!-- Copy legal documents from "legal" module to "target/classes" folder so that we make sure to include
           legal docs in all modules. -->
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-dependency-plugin</artifactId>
         <executions>
           <execution>
-            <!-- Execution and configuration for copying certificates from related module to "target/classes" folder 
+            <!-- Execution and configuration for copying certificates from related module to "target/classes" folder
               so that we can include them in the image. -->
             <id>copy_demo_certs</id>
             <phase>generate-resources</phase>
@@ -139,7 +139,7 @@
               <goal>copy-resources</goal>
             </goals>
             <configuration>
-              <outputDirectory>${project.build.directory}/deploy/helm</outputDirectory>
+              <outputDirectory>${project.build.directory}/deploy/helm/eclipse-hono</outputDirectory>
               <resources>
                 <resource>
                   <directory>${project.build.directory}/config/hono-demo-certs-jar</directory>
--- a/deploy/src/main/deploy/helm/templates/hono-adapter-amqp/hono-adapter-amqp-vertx-deployment.yaml
+++ b/deploy/src/main/deploy/helm/templates/hono-adapter-amqp/hono-adapter-amqp-vertx-deployment.yaml
@@ -14,7 +14,7 @@ spec:
     spec:
       containers:
       {{- include "hono.jaeger.agent" . | indent 6 }}
-      - image: {{ .Values.honoContainerRegistry }}/eclipse/hono-adapter-amqp-vertx:{{ .Values.honoImageTag }}
+      - image: {{ .Values.adapters.amqp.imageName }}
         imagePullPolicy: IfNotPresent
         name: eclipse-hono-adapter-amqp-vertx
         ports:
--- a/deploy/src/main/deploy/helm/templates/hono-service-auth/hono-service-auth-deployment.yaml
+++ b/deploy/src/main/deploy/helm/templates/hono-service-auth/hono-service-auth-deployment.yaml
@@ -13,7 +13,7 @@ spec:
       {{- include "hono.metadata" $args | nindent 6 }}
     spec:
       containers:
-      - image: {{ .Values.honoContainerRegistry }}/eclipse/hono-service-auth:{{ .Values.honoImageTag }}
+      - image: {{ .Values.authServer.imageName }}
         imagePullPolicy: IfNotPresent
         name: eclipse-hono-service-auth
         ports:
--- a/deploy/src/main/deploy/helm/templates/hono-service-device-registry/hono-service-device-registry-deployment.yaml
+++ b/deploy/src/main/deploy/helm/templates/hono-service-device-registry/hono-service-device-registry-deployment.yaml
@@ -1,5 +1,5 @@
 apiVersion: apps/v1beta1
-kind: Deployment
+kind: StatefulSet
 metadata:
   {{- $args := dict "dot" . "component" "service-device-registry" "name" "service-device-registry" }}
   {{- include "hono.metadata" $args | nindent 2 }}
@@ -12,6 +12,7 @@ spec:
     metadata:
       {{- include "hono.metadata" $args | nindent 6 }}
     spec:
+{{- if .Values.deviceRegistry.resetFiles }}
       initContainers:
       - name: copy-example-data
         image: busybox
@@ -26,9 +27,10 @@ spec:
           name: conf
         - mountPath: /var/lib/hono/device-registry
           name: registry
+{{- end }}
       containers:
       {{- include "hono.jaeger.agent" . | indent 6 }}
-      - image: {{ .Values.honoContainerRegistry }}/eclipse/hono-service-device-registry:{{ .Values.honoImageTag }}
+      - image: {{ .Values.deviceRegistry.imageName }}
         imagePullPolicy: IfNotPresent
         name: eclipse-hono-service-device-registry
         ports:
--- a/deploy/src/main/deploy/helm/values.yaml
+++ b/deploy/src/main/deploy/helm/values.yaml
@@ -17,6 +17,12 @@ artemisImageName: ${artemis.image.name}
 # deviceRegistryDeployExample indicates whether the example Device Registry
 # should be deployed and used.
 deviceRegistryDeployExample: true
+deviceRegistry:
+  # Persistent volume claim storage class
+  storageClass: default
+  # Set to false to not override registry files with example content
+  resetFiles: true
+  imageName: ${docker.repository}/hono-service-device-registry:${project.version}
 
 
 # dataGridDeployExample indicates whether the example data grid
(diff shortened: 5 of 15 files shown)


deploy/pom.xml
```diff
@@ -139,7 +139,7 @@
               <goal>copy-resources</goal>
             </goals>
             <configuration>
-              <outputDirectory>${project.build.directory}/deploy/helm</outputDirectory>
+              <outputDirectory>${project.build.directory}/deploy/helm/eclipse-hono</outputDirectory>
```
sophokles73: is there  a particular reason why you want to move this to a subfolder?

deploy/src/main/deploy/helm/templates/hono-adapter-amqp/hono-adapter-amqp-vertx-deployment.yaml
```diff
@@ -14,7 +14,7 @@ spec:
     spec:
       containers:
       {{- include "hono.jaeger.agent" . | indent 6 }}
-      - image: {{ .Values.honoContainerRegistry }}/eclipse/hono-adapter-amqp-vertx:{{ .Values.honoImageTag }}
+      - image: {{ .Values.honoContainerRegistry }}/{{ .Values.honoImageOrgName }}/hono-adapter-amqp-vertx:{{ .Values.honoImageTag }}
```
sophokles73: instead of using multiple variables to assemble the image name from, I had already thought about using a single property for the complete image name, e.g. `adapters.amqp.imageName`.
FMPOV this would be a little easier to understand and document. WDYT?

deploy/pom.xml
kaizimmerm: It is specified by helm that the chart has to be in a folder with the chart name. In fact `helm package` returns an error if the folder is not compliant. There is some [debate on this](https://github.com/helm/helm/issues/1979#issuecomment-280507379) but my understanding is that the rule still stands.

deploy/src/main/deploy/helm/templates/hono-adapter-amqp/hono-adapter-amqp-vertx-deployment.yaml
kaizimmerm: Yes, I would have actually preferred it that way. It is as well what we did in [Ditto](https://github.com/eclipse/ditto/blob/master/deployment/helm/eclipse-ditto/values.yaml). However, it would break the current helm values structure that's why I added a property instead of changing the current one.

However, I can change it if that is ok for you.

sophokles73: FMPOV we should do that. We haven't released 1.0 yet so I think we can get away with this change as most people will probably not change the name anyway ...

@dejanb WDYT?

deploy/src/main/deploy/helm/templates/hono-service-device-registry/hono-service-device-registry-deployment.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: apps/v1beta1
-kind: Deployment
+kind: StatefulSet
```
sophokles73: I see your point here and I agree that the stateful set model is a better fit for the example device registry in general. However, in practice the registry will only work with a single pod anyway, regardless of whether we use a *Deployment* or *StatefulSet*. Would you agree?

kaizimmerm: True, but when you re-deploy/update k8s will boot up a second pod and this case there is a high probability to tun into a "Multi-Attach error" if the second pod goes onto another node in the cluster which can happen with `deployment`.

sophokles73: I see, and with a *Stateful Set* k8s does what instead when re-deploying/updating?

kaizimmerm: Well, in order to maintain its _sticky identity_ for pods it does not try do the update zero downtime, i.e. destroys the old pod first and then boots the new one.

I have not found a way to get that behaviour (i.e. no zero downtime) out of _deployment_. Based on all of the discussions around Multi-Attach error it seems _ StatefulSet_ is the solution of choice.

sophokles73: Ok, I have no issues with changing to *Stateful Set* for the example registry here. We use it for demo purposes only anyway ...

deploy/src/main/deploy/helm/templates/hono-adapter-amqp/hono-adapter-amqp-vertx-deployment.yaml
dejanb: Yeah, that sounds good to me

sophokles73: @kaizimmerm I think we are in consensus here. So, please change to a single property only ...

deploy/src/main/deploy/helm/templates/hono-adapter-amqp/hono-adapter-amqp-vertx-deployment.yaml
```diff
@@ -14,7 +14,7 @@ spec:
     spec:
       containers:
       {{- include "hono.jaeger.agent" . | indent 6 }}
-      - image: {{ .Values.honoContainerRegistry }}/eclipse/hono-adapter-amqp-vertx:{{ .Values.honoImageTag }}
+      - image: {{ .Values.adapters.amqp.imageName }}:{{ .Values.honoImageTag }}
```
sophokles73: FMPOV the image name should also include the tag

deploy/src/main/deploy/helm/templates/hono-service-auth/hono-service-auth-deployment.yaml
```diff
@@ -13,7 +13,7 @@ spec:
       {{- include "hono.metadata" $args | nindent 6 }}
     spec:
       containers:
-      - image: {{ .Values.honoContainerRegistry }}/eclipse/hono-service-auth:{{ .Values.honoImageTag }}
+      - image: {{ .Values.deviceRegistry.imageName }}:{{ .Values.honoImageTag }}
```
sophokles73: this should be `authService.imageName`, I guess ...

deploy/src/main/deploy/helm/templates/hono-service-device-registry/hono-service-device-registry-deployment.yaml
```diff
@@ -26,9 +27,10 @@ spec:
           name: conf
         - mountPath: /var/lib/hono/device-registry
           name: registry
+{{- end }}
       containers:
       {{- include "hono.jaeger.agent" . | indent 6 }}
-      - image: {{ .Values.honoContainerRegistry }}/eclipse/hono-service-device-registry:{{ .Values.honoImageTag }}
+      - image: {{ .Values.deviceRegistry.imageName }}:{{ .Values.honoImageTag }}
```
sophokles73: again, the image name should contain the tag FMPOV

deploy/src/main/deploy/helm/values.yaml
```diff
@@ -17,6 +17,16 @@ artemisImageName: ${artemis.image.name}
 # deviceRegistryDeployExample indicates whether the example Device Registry
 # should be deployed and used.
 deviceRegistryDeployExample: true
+deviceRegistry:
+  # Persistent volume claim storage class
+  storageClass: default
+  # Set to false to not override registry files with example content
+  resetFiles: true
+  imageName: ${docker.repository}/hono-service-device-registry
```
sophokles73: include tag

deploy/src/main/deploy/helm/templates/hono-adapter-amqp/hono-adapter-amqp-vertx-deployment.yaml
kaizimmerm: I have updated the PR. However, I have still a few tests to run. So stay tuned.

kaizimmerm: @sophokles73 my tests are complete. IMHO you could merge.

kaizimmerm:
@sophokles73 I will look into the helm guidance as well.

kaizimmerm:
@sophokles73 I adapted the helm deploy guidance.

site/documentation/content/deployment/helm-based-deployment.md
```diff
@@ -186,7 +186,7 @@ Once the images have been pushed, the deployment can be done using Helm:
 
 ~~~sh
 # in Hono working tree directory: hono/deploy
-helm install --dep-up --name hono --namespace hono --set honoContainerRegistry=my.registry.io target/deploy/helm/
+helm install --dep-up --name hono --namespace hono target/deploy/helm/eclipse-hono/
```
sophokles73: FMPOV this will only work if not using a custom tag when pushing the images to the private container registry. Otherwise, the images will not be found when the pods are starting up because they are looked up using the Maven `${project.version}` instead of the custom tag ...

kaizimmerm: That is independent of the registry setting, right?

But true I could add a chapter on the custom tag topic.

kaizimmerm: Ah, now I get it. The push images line in docs defines "1.0-CUSTOM"

sophokles73: BTW, there are still merge conflicts ...

kaizimmerm: @sophokles73 are you sure, I merged today with master and GitHub tells me no conflicts? I updated the docs btw.

sophokles73: > This branch cannot be rebased due to conflicts
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.

sophokles73:
@kaizimmerm This looks good now. Can you squash and rebase?

kaizimmerm:
@sophokles73 done

PR #1453, opened 2019-08-20
add support for auhentication for the device management HTTP API

dejanb:


dejanb:
I figured a much simpler way to implement this, so please wait with the reviews.

dejanb:
@ctron @sophokles73 it's now ready for review

service-base/src/main/java/org/eclipse/hono/service/http/HttpServiceBase.java
```diff
@@ -185,7 +200,10 @@ protected Router createRouter() {
         LOG.info("limiting size of inbound request body to {} bytes", getConfig().getMaxPayloadSize());
         matchAllRoute.handler(BodyHandler.create().setUploadsDirectory(DEFAULT_UPLOADS_DIRECTORY)
                 .setBodyLimit(getConfig().getMaxPayloadSize()));
-
+        //4. AuthHandler
+        if (authHandler != null) {
+            matchAllRoute.handler(authHandler);
```
sophokles73: how about letting subclasses decide which routes to configure the auth handler for, e.g. by delegating to a protected method passing in the matchAllRoute and the router and providing a default implementation?

services/device-registry-legacy/src/main/java/org/eclipse/hono/deviceregistry/ApplicationConfig.java
```diff
@@ -215,6 +216,17 @@ public ServiceConfigProperties restProperties() {
...
     }
 
+    /**
+     * Creates an authentication handler used by device registry management HTTP API.
+     *
+     * @return The handler.
+     */
+    @Bean
+    @Scope("prototype")
+    public AuthHandler authHandler() {
```
sophokles73: Is this just a placeholder which you will fill in before merging this PR?
Otherwise I do not see the value of this, given that the auth handler is not a required property ...

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/ApplicationConfig.java
```diff
@@ -206,6 +207,17 @@ public DeviceManagementHttpEndpoint registrationHttpEndpoint() {
...
     }
 
+    /**
+     * Creates an authentication handler used by device registry management HTTP API.
+     *
+     * @return The handler.
+     */
+    @Bean
+    @Scope("prototype")
+    public AuthHandler authHandler() {
```
sophokles73: same here

service-base/src/main/java/org/eclipse/hono/service/http/HttpServiceBase.java
dejanb: Done

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/ApplicationConfig.java
dejanb: It was used for manual testing and I thought it'd be good to leave it as a placeholder for the future. I removed it as suggested.

services/device-registry-legacy/src/main/java/org/eclipse/hono/deviceregistry/ApplicationConfig.java
dejanb: It was used for manual testing and I thought it'd be good to leave it as a placeholder for the future. I removed it as suggested.

dejanb:
> Are you actually planning to also provide/use an auth handler for the example device registry? If so, I guess it would be worth considering to also add a session handler so that not each request needs to be re-authenticated ...

As discussed on the community call, we'll try to provide Kubernets/Openshift based auth handlers in the future. But that will be covered by separate issues/ PRs

Issue #1227, opened 2019-05-08
Remove metrics for no longer active tenants from Prometheus response


Christian-Schmid:
Hi,

we've just started to use the new metrics based on Prometheus in Eclipse Hono. Very nice work!
However, one question has come to our attention regarding the amount of responded DataPoints of protocol adapters: 
In the Prometheus response I could observe two groups of data points: 
a) "Static data points" which are always there, e.g. with information of the jvm (for example: ``jvm_buffer_memory_used_bytes``)
b) "Dynamic data points" labeled with a certain ``tenant``. E.g.: ``hono_messages_received_seconds_max{component_name="hono-http",component_type="adapter",host="hono-adapter-http-vertx-7cd64d4867-c8p9s",qos="1",status="forwarded",tenant="acme",type="event",} 0.0``

The amount of returned DataPoints in the b) case grows with the amount of tenants which where at one point interacting with the protocol adapter. 

From the observations I made it looks like when a tenant stops interacting with the protocol adapter its metrics will still be returned in future. This leads to an over time growing amount of returned data points. This is inconvenient as it results in high load on the metric infrastructure and thus to high costs. Some cloud providers charge you based on the amount of points reported per scrape.

It would be nice that metrics for no longer interacting (e.g. by a specified time frame) tenants are removed from the result. 

What are your options? 

Thanks 
Chris

ctron:
I think this goes hand in hand with issue #1188 ... maybe we do need some kind of "tenant closed" event (internally!) on which we can hook in functionality like that.

PR #1456, opened 2019-08-21
[#1227] Remove metrics for "inactive" tenants.

b-abel:
The goal in this PR is to solve #1227 by removing message related metrics when corresponding downstream link times out (#1188). 
The device connection related metrics ("hono.connections.authenticated") could be removed when the connection count for a tenant drops to 0.

This logic is the same for every type of protocol adapter, so it should be implemented in the common base class `AbstractProtocolAdapterBase`. The first step for this is to pull the metrics object up to this class from the individual protocol adapter classes. This requires the base class to be parametrized with the concrete type of metrics for a protocol adapter.

Following steps are:
* expose a handler for the link close event to the protocol adapter and
* implement the removal of the metrics

### DIFF
--- a/adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
+++ b/adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
@@ -89,6 +89,7 @@ import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
 import io.vertx.core.buffer.Buffer;
+import io.vertx.core.eventbus.EventBus;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.unit.Async;
 import io.vertx.ext.unit.TestContext;
@@ -1037,7 +1038,9 @@ public class VertxBasedAmqpProtocolAdapterTest {
         adapter.setMetrics(metrics);
         adapter.setResourceLimitChecks(resourceLimitChecks);
 
-        adapter.init(Vertx.vertx(), mock(Context.class));
+        final Vertx vertx = mock(Vertx.class);
+        when(vertx.eventBus()).thenReturn(mock(EventBus.class));
+        adapter.init(vertx, mock(Context.class));
         return adapter;
     }
 
--- a/adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
+++ b/adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
@@ -84,8 +84,10 @@ import org.mockito.ArgumentCaptor;
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
 import io.vertx.core.AsyncResult;
+import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
+import io.vertx.core.Vertx;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.unit.Async;
@@ -1034,6 +1036,8 @@ public class VertxBasedAmqpProtocolAdapterTest {
         adapter.setDeviceConnectionClientFactory(deviceConnectionClientFactory);
         adapter.setMetrics(metrics);
         adapter.setResourceLimitChecks(resourceLimitChecks);
+
+        adapter.init(Vertx.vertx(), mock(Context.class));
         return adapter;
     }
 
--- a/client/src/test/java/org/eclipse/hono/client/impl/CommandConsumerFactoryImplTest.java
+++ b/client/src/test/java/org/eclipse/hono/client/impl/CommandConsumerFactoryImplTest.java
@@ -96,6 +96,7 @@ public class CommandConsumerFactoryImplTest {
         vertx = mock(Vertx.class);
         context = HonoClientUnitTestHelper.mockContext(vertx);
         when(vertx.getOrCreateContext()).thenReturn(context);
+        when(vertx.eventBus()).thenReturn(mock(EventBus.class));
         doAnswer(invocation -> {
             final Handler<Void> handler = invocation.getArgument(1);
             handler.handle(null);
@@ -219,7 +220,6 @@ public class CommandConsumerFactoryImplTest {
         when(deviceSpecificCommandReceiver.getRemoteSource()).thenReturn(source);
         when(deviceSpecificCommandReceiver.isOpen()).thenReturn(Boolean.TRUE);
         when(vertx.setPeriodic(anyLong(), anyHandler())).thenReturn(10L);
-        when(vertx.eventBus()).thenReturn(mock(EventBus.class));
 
         // GIVEN a command consumer
         commandConsumerFactory.createCommandConsumer(tenantId, deviceId, commandHandler, null, 5000L)
@@ -266,7 +266,6 @@ public class CommandConsumerFactoryImplTest {
         when(deviceSpecificCommandReceiver.getSource()).thenReturn(source);
         when(deviceSpecificCommandReceiver.getRemoteSource()).thenReturn(source);
         when(vertx.setPeriodic(anyLong(), anyHandler())).thenReturn(10L);
-        when(vertx.eventBus()).thenReturn(mock(EventBus.class));
         doAnswer(invocation -> {
             final Handler<Void> handler = invocation.getArgument(1);
             handler.handle(null);
--- a/client/src/test/java/org/eclipse/hono/client/impl/CommandConsumerFactoryImplTest.java
+++ b/client/src/test/java/org/eclipse/hono/client/impl/CommandConsumerFactoryImplTest.java
@@ -53,6 +53,7 @@ import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
+import io.vertx.core.eventbus.EventBus;
 import io.vertx.ext.unit.TestContext;
 import io.vertx.ext.unit.junit.VertxUnitRunner;
 import io.vertx.proton.ProtonMessageHandler;
@@ -218,6 +219,7 @@ public class CommandConsumerFactoryImplTest {
         when(deviceSpecificCommandReceiver.getRemoteSource()).thenReturn(source);
         when(deviceSpecificCommandReceiver.isOpen()).thenReturn(Boolean.TRUE);
         when(vertx.setPeriodic(anyLong(), anyHandler())).thenReturn(10L);
+        when(vertx.eventBus()).thenReturn(mock(EventBus.class));
 
         // GIVEN a command consumer
         commandConsumerFactory.createCommandConsumer(tenantId, deviceId, commandHandler, null, 5000L)
@@ -264,6 +266,7 @@ public class CommandConsumerFactoryImplTest {
         when(deviceSpecificCommandReceiver.getSource()).thenReturn(source);
         when(deviceSpecificCommandReceiver.getRemoteSource()).thenReturn(source);
         when(vertx.setPeriodic(anyLong(), anyHandler())).thenReturn(10L);
+        when(vertx.eventBus()).thenReturn(mock(EventBus.class));
         doAnswer(invocation -> {
             final Handler<Void> handler = invocation.getArgument(1);
             handler.handle(null);
--- a/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
@@ -43,12 +43,15 @@ import org.eclipse.hono.client.ServerErrorException;
 import org.eclipse.hono.client.ServiceInvocationException;
 import org.eclipse.hono.client.TenantClient;
 import org.eclipse.hono.client.TenantClientFactory;
+import org.eclipse.hono.client.impl.AbstractHonoClient;
 import org.eclipse.hono.config.AbstractConfig;
 import org.eclipse.hono.config.ProtocolAdapterProperties;
 import org.eclipse.hono.service.auth.ValidityBasedTrustOptions;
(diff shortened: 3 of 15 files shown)


b-abel:
To make the review easier, the first commit contains only the refactoring of the protocol adapter classes where the metrics object is moved to the base class.

b-abel:
Rebased to resolve conflicts.

The commits up to now include the removal of metrics for telemetry (and event) messages as well as the metrics for (authenticated) connections. The former is triggered by the timeout of the tenant's sender link (#1188), the latter happens when the last (southbound) connection for a tenant is closed. 

Is this the right event or do we need another mechanism for the removal of connection metrics? WDYT?

Open: the removal of metrics for command and control. I would like to add this to this PR, but feedback on the current is more than welcome.

b-abel:
Do we want to replace usages of `AbstractHonoClient#closeLinks()` that provide a non-empty handler by using the event bus based notification?

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -120,6 +124,16 @@ public DownstreamSenderFactory getMessageSenderClient() {
 
     };
 
+    /**
+     * Creates an instance initialized with the given metrics.
+     * 
+     * @param metrics The metrics to use.
+     * @throws NullPointerException if metrics is {@code null}.
+     */
+    public AbstractProtocolAdapterBase(final U metrics) {
```
sophokles73: I think this should be protected as the abstract class cannot be instantiated anyway ...

service-base/src/test/java/org/eclipse/hono/service/AbstractProtocolAdapterBaseTest.java
```diff
@@ -589,30 +591,33 @@ public void testCheckConnectionLimitFailsIfMessageLimitIsReached(final VertxTest
...
+    private AbstractProtocolAdapterBase<ProtocolAdapterProperties, Metrics> newProtocolAdapter(
             final ProtocolAdapterProperties props,
             final String typeName,
             final Handler<Void> startupHandler,
             final Handler<Void> commandConnectionEstablishedHandler,
             final Handler<Void> commandConnectionLostHandler) {
 
-        final AbstractProtocolAdapterBase<ProtocolAdapterProperties> result = new AbstractProtocolAdapterBase<>() {
+        final AbstractProtocolAdapterBase<ProtocolAdapterProperties, Metrics> result = new AbstractProtocolAdapterBase<>(
+                mock(NoopBasedMetrics.class)) {
```
sophokles73: we usually do not mock classes but interfaces only. Is there a particular reason why you do not want to mock `Metrics` instead?

b-abel: The reason is that I don't need to mock the event bus for the NoopBasedMetrics, since it is not used in that case. But I thinking again, this reason is not good enough, so I'll change that.

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
```diff
@@ -123,6 +125,21 @@ void reportTelemetry(
             MetricsTags.TtdStatus ttdStatus,
             Sample timer);
 
+    /**
+     *
```
sophokles73: please remove this empty line

service-base/src/main/java/org/eclipse/hono/service/metric/Metrics.java
```diff
@@ -123,6 +125,21 @@ void reportTelemetry(
...
+     * Removes the metrics of telemetry or event messages for the given tenant.
+     * <p>
+     * This is used to clean the metrics for tenants that are no longer active.
```
sophokles73: what about
> This is done to prevent reporting of never changing meter values for inactive tenants.

service-base/src/test/java/org/eclipse/hono/service/metric/MicrometerBasedMetricsTest.java
```diff
@@ -218,4 +219,99 @@ public void testPayloadSizeForCommandMessages(final MeterRegistry registry) {
...
+        assertEquals(2, search.meters().size());
+        assertEquals(4, metrics.getNumberOfConnections());
+
+        // WHEN closing all connections for one tenant
+        metrics.decrementConnections(tenant1);
+        assertEquals(2, search.meters().size());
+        metrics.decrementConnections(tenant1);
+
+        // THEN the metric for this tenant is removed
+        assertEquals(1, search.meters().size());
```
sophokles73: IMHO you should make sure that the correct metric has been removed, i.e. the one for tenant1 ...

service-base/src/test/java/org/eclipse/hono/service/metric/MicrometerBasedMetricsTest.java
```diff
@@ -218,4 +219,99 @@ public void testPayloadSizeForCommandMessages(final MeterRegistry registry) {
...
+    @ParameterizedTest
+    @MethodSource("registries")
+    public void testAuthenticatedConnections(final MeterRegistry registry) {
+
+        final MicrometerBasedMetrics metrics = new MicrometerBasedMetrics(registry);
+
+        final String tenant1 = "tenant1";
+        final String tenant2 = "tenant2";
+
+        // GIVEN a metrics instance that has recorded one telemetry message and one event per tenant
```
sophokles73: IMHO this comment does not reflect what we are setting up as the fixture, does it?

service-base/src/test/java/org/eclipse/hono/service/metric/MicrometerBasedMetricsTest.java
```diff
@@ -218,4 +219,99 @@ public void testPayloadSizeForCommandMessages(final MeterRegistry registry) {
...
+                registry.find(MicrometerBasedMetrics.METER_MESSAGES_RECEIVED).meters().size());
+        assertEquals(4,
+                registry.find(MicrometerBasedMetrics.METER_MESSAGES_PAYLOAD).meters().size());
+
+        // WHEN removing the telemetry metrics for tenant1
+        metrics.removeTelemetryMetricsForTenant(EndpointType.TELEMETRY, tenant1);
+
+        // THEN only the metric for telemetry and this tenant is removed
+        assertEquals(1,
+                registry.find(MicrometerBasedMetrics.METER_MESSAGES_RECEIVED).tags(tenantTag1).meters().size());
```
sophokles73: IMHO the search needs to also include the message type tag to make sure that the right metric has been removed ...

client/src/test/java/org/eclipse/hono/client/impl/CommandConsumerFactoryImplTest.java
```diff
@@ -218,6 +219,7 @@ public void testLocalCloseRemovesCommandConsumerFromCache(final TestContext ctx)
         when(deviceSpecificCommandReceiver.getRemoteSource()).thenReturn(source);
         when(deviceSpecificCommandReceiver.isOpen()).thenReturn(Boolean.TRUE);
         when(vertx.setPeriodic(anyLong(), anyHandler())).thenReturn(10L);
+        when(vertx.eventBus()).thenReturn(mock(EventBus.class));
```
sophokles73: I think this can be set up for all tests in one place, can't it?

adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -1034,6 +1036,8 @@ private VertxBasedAmqpProtocolAdapter getAdapter(final ProtonServer server) {
         adapter.setDeviceConnectionClientFactory(deviceConnectionClientFactory);
         adapter.setMetrics(metrics);
         adapter.setResourceLimitChecks(resourceLimitChecks);
+
+        adapter.init(Vertx.vertx(), mock(Context.class));
```
sophokles73: if possible, can we use a mocked Vertx here?

sophokles73:
@b-abel looks good so far FMPOV

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -110,8 +117,14 @@ public final void incrementConnections(final String tenantId) {
     public final void decrementConnections(final String tenantId) {
 
         Objects.requireNonNull(tenantId);
-        gaugeForTenant(METER_CONNECTIONS_AUTHENTICATED, this.authenticatedConnections, tenantId, AtomicLong::new)
+        final long connections = gaugeForTenant(METER_CONNECTIONS_AUTHENTICATED, this.authenticatedConnections,
```
ctron: Are we sure that this code is thread safe?

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -220,6 +233,20 @@ public final void reportTelemetry(
...
+    public void removeTelemetryMetricsForTenant(final MetricsTags.EndpointType type, final String tenantId) {
+        Objects.requireNonNull(type);
+        Objects.requireNonNull(tenantId);
+
+        if (type != MetricsTags.EndpointType.TELEMETRY && type != MetricsTags.EndpointType.EVENT) {
+            throw new IllegalArgumentException("invalid type, must be either telemetry or event");
+        }
+
+        final Tags tags = Tags.of(MetricsTags.getTenantTag(tenantId), type.asTag());
+        registry.find(METER_MESSAGES_PAYLOAD).tags(tags).meters().forEach(registry::remove);
```
ctron: Are we sure that this code is thread safe?

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -433,6 +472,21 @@ protected final ResourceLimitChecks getResourceLimitChecks() {
         return result;
     }
 
+    private void registerForLinkCloseEvent() {
+        if (this.metrics instanceof NoopBasedMetrics) {
+            return;
+        }
+
+        vertx.eventBus().consumer(AbstractHonoClient.EVENT_BUS_ADDRESS_CLIENT_SENDER_CLOSE, message -> {
```
ctron: I don't understand the need to use the event bus here? Wouldn't it be possible to directly call this method?

b-abel: Currently the handler for link close events is not exposed by the client factories. This needs to be changed if we want to protocol adapters to be notified.
By doing so the API of the client would become more complex, because they hide the creation of the client in "getOrCreate" methods (e.g. [DownstreamSenderFactory#getTelemetrySender(String)](https://github.com/eclipse/hono/blob/ae47a9d25a13cf24695a7dfc35421087c3fd8861/client/src/main/java/org/eclipse/hono/client/DownstreamSenderFactory.java#L50)). Adding the close handler as a parameter to those methods would make them more error-prone, in that you must not provide a new handler instance for subsequent calls. 
Using the event bus decouples the client and the protocol adapter and I would rather remove the handler from the closeLinks() method. What downsides do you see in this solution?

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
b-abel: IMHO the trigger (connection count for the tenant drops to 0) is a problem and parallel connects and disconnects are quite likely to happen. The value can change quite often between two scraping intervals and can only be removed if it is 0 for a longer time. So this needs indeed to be changed. The question remains what a better trigger event could be.

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
b-abel: MeterRegistry#remove(Meter) does synchronization, but the timeout event, the cleanup and the creation of a new  entries are of course not synchronized together. I don't think that it would be a major problem if a tenant that has been inactive for e.g. one hour (default timeout) really happens to send a new message _during_ this cleanup and this single event is not reflected in the metrics (which are probably not 100% accurate anyways).

b-abel:
@ctron I had a discussion with @sophokles73 regarding the thread-safety topic and the triggers for "tenant timeout" events and we came to the conclusion that it would be better to manage the timeout centrally in the Metrics implementation. The proposal in this PR as well as the solution in #1416 for #1188 manage the "tenant close" events individually for each AMQP link, which makes things complicated when deploying multiple vertical instances, which manage their own connections and links independently from each other.
What we instead need is managing the timeout centrally in the Vert.x container for all vertical instances. The metrics implementation is such a central place and it already manages state for each tenant the container had seen, so it seems to be a natural fit for managing the timeout. This means changing the implementation from #1416, to emmiting the "tenant close" event by the metrics and consuming it by the Hono client to close the links. While doing so I would like to have one timeout event for all link types (telemetry, event, c&c) instead of having individual ones per link type.
I would like to provide a new PR for that.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
ctron: Ok, I see the problem. I agree that having this in the "getOrCreate" method is a bad idea.

The event bus has a tendency to make things: a) more complex b) error prune due to not having interfaces c) performance impacts vs a direct invocation … But I think you are right and in this case it is a good compromise.

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
ctron: I am not so worried about correct metrics, more about he facts that two thread could update the same, backing collection. Depending on the collection it may be that concurrent iterations and removing/adding can become a problem, and throw a CME. Which might even be the case in a single thread when you have multiple entries to remove.

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
ctron: I think you can only solve this properly by synchronizing the increment and decrement operation.

sophokles73:
replaced by #1493

PR #1461, opened 2019-08-26
Consider SSLException terminal for connection attempt.

### LINKED ISSUE #1424: HonoClient endlessly retries to establish connection that fails due to SSL exception
When the client tries to open a connection to a service and the TLS handshake fails, the client re-tries forever to establish the connection.
However, the handshake failure will most likely not go away without reconfiguring the client and/or server. Thus, the SSLException should be considered a *terminal error* that stops the client from re-trying.

sophokles73:
The HonoClient now stops a connection attempt if it runs into an
SSLException which indicates a permanent problem with the TLS handshake.

Fixes #1424

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
@@ -25,6 +25,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import javax.net.ssl.SSLException;
 import javax.security.sasl.AuthenticationException;
 
 import org.apache.qpid.proton.amqp.Symbol;
@@ -521,7 +522,8 @@ public class HonoConnectionImpl implements HonoConnection {
 
     private boolean isTerminalConnectionError(final Throwable connectionFailureCause) {
 
-        return connectionFailureCause instanceof AuthenticationException ||
+        return connectionFailureCause instanceof SSLException ||
+                connectionFailureCause instanceof AuthenticationException ||
                 connectionFailureCause instanceof MechanismMismatchException ||
                 (connectionFailureCause instanceof SaslSystemException && ((SaslSystemException) connectionFailureCause).isPermanent());
     }
@@ -541,6 +543,11 @@ public class HonoConnectionImpl implements HonoConnection {
         } else if (connectionFailureCause instanceof MechanismMismatchException) {
             connectionHandler.handle(Future.failedFuture(
                     new ClientErrorException(HttpURLConnection.HTTP_UNAUTHORIZED, "no suitable SASL mechanism found for authentication with server")));
+        } else if (connectionFailureCause instanceof SSLException) {
+            connectionHandler.handle(Future.failedFuture(
+                    new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST,
+                            "TLS handshake with server failed: " + connectionFailureCause.getMessage(),
+                            connectionFailureCause)));
         } else {
             connectionHandler.handle(Future.failedFuture(
                     new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "failed to connect",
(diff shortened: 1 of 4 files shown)


client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -541,6 +543,10 @@ private void failConnectionAttempt(final Throwable connectionFailureCause, final
         } else if (connectionFailureCause instanceof MechanismMismatchException) {
             connectionHandler.handle(Future.failedFuture(
                     new ClientErrorException(HttpURLConnection.HTTP_UNAUTHORIZED, "no suitable SASL mechanism found for authentication with server")));
+        } else if (connectionFailureCause instanceof SSLException) {
+            connectionHandler.handle(Future.failedFuture(
+                    new ClientErrorException(HttpURLConnection.HTTP_UNAUTHORIZED,
+                            "TLS handshake with server failed: " + connectionFailureCause.getMessage())));
```
ctron: I think it is helpful to have the root cause as well, instead of only having the top message.

sophokles73: I have pushed a corresponding change ...

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -541,6 +543,10 @@ private void failConnectionAttempt(final Throwable connectionFailureCause, final
         } else if (connectionFailureCause instanceof MechanismMismatchException) {
             connectionHandler.handle(Future.failedFuture(
                     new ClientErrorException(HttpURLConnection.HTTP_UNAUTHORIZED, "no suitable SASL mechanism found for authentication with server")));
+        } else if (connectionFailureCause instanceof SSLException) {
+            connectionHandler.handle(Future.failedFuture(
+                    new ClientErrorException(HttpURLConnection.HTTP_UNAUTHORIZED,
```
calohmn: Wouldn't a `400 Bad Request` be more appropriate here?
(Possibly `401 Unauthorized` could be tied to the `SSLPeerUnverifiedException` subclass of `SSLException`, although I wonder if that can occur here.)

tests/src/test/java/org/eclipse/hono/tests/client/HonoClientImplIT.java
```diff
@@ -53,27 +56,51 @@ public static void init() {
...
+        downstreamProps.setReconnectAttempts(-1);
+
+        clientFactory = IntegrationTestApplicationClientFactory.create(HonoConnection.newConnection(vertx, downstreamProps));
+        // WHEN the client tries to connect
+        clientFactory.connect().setHandler(ctx.failing(t -> {
+            // THEN the connection attempt fails due to lack of authorization
+            ctx.verify(() -> {
+                assertThat(ServiceInvocationException.extractStatusCode(t)).isEqualTo(HttpURLConnection.HTTP_UNAUTHORIZED);
+            });
+        ctx.completeNow();
```
calohmn: indentation

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
sophokles73: I gave that a lot of thought as well :-) You could also try to use a more specific exception for `SSLKeyException`. However, since it is used to report problems with both the client as well as the server cert, you don't exactly know if it should be a `ClientErrorException` or a `ServerErrorException`.
I had also thought about just using a 400 code but again, the problem may also lie on the server side, so it is hard to do the exact right thing here.
I am fine with using the 400 code but I think we should stick with a single exception (and code) that we use for all TLS related problems for simplicity. In the end, we are also including the root cause which might be used to better understand the underlying problem. In any case, the SSLException should be considered a terminal condition ...

sophokles73:
@ctron @calohmn anything else we need to change?

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
calohmn: Ok, I would opt for using `400 Bad Request` as the single exception then.
Otherwise an "unauthorized" exception could be confusing in the case that there is an issue with incompatible cipher suites for example.

sophokles73: done

sophokles73:
@calohmn can you take another look?

PR #1463, opened 2019-08-27
Fix an issue encoding the timestamp.

ctron:
This change sets up explicit encoding of the timestamp, and also enforces this when encoding.

Instead of expecting the user of the classes to register an additional data type module, the fields explicitly declare their serializer now.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/annotation/HonoTimestamp.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/annotation/HonoTimestamp.java
@@ -1,0 +1,63 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.hono.service.annotation;
+
+import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
+import static java.lang.annotation.ElementType.FIELD;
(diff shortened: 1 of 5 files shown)


service-base/src/main/java/org/eclipse/hono/service/annotation/HonoTimestamp.java
```diff
@@ -0,0 +1,63 @@
...
+@JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ss[XXX]", timezone = "UTC")
+@JsonSerialize(using = InstantSerializer.class)
+@JsonDeserialize(using = HonoInstantDeserializer.class)
+public @interface HonoTimestamp {
+}
+
+/**
+ * A deserializer for Hono's format of {@link Instant}s.
+ */
+class HonoInstantDeserializer extends InstantDeserializer<Instant> {
```
dejanb: This class looks identical to InstantDeserializer.INSTANT, can we use that one?

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
```diff
@@ -592,16 +580,6 @@ protected void verifyBcryptPasswordHash(final String pwdHash) {
         }
     }
 
-    private void checkHashedPassword(final PasswordSecret secret) {
```
dejanb: Was this intentional? It doesn't seems related to the rest of the changes.

service-base/src/main/java/org/eclipse/hono/service/annotation/HonoTimestamp.java
ctron: Unfortunately, we can't. Because for using this in annotations, we need to have a Class, rather than an instance of a class.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
ctron: It caused a warning, since the method was not used by anyone..

Issue #1460, opened 2019-08-26
Management API does not support Tenant's data limit


sophokles73:
The Management API does not explicitly support management of a tenant's data volume limits as required by the Tenant API. It could be specified managed as an *extension* property. However, in that case the registry implementation would need to map the *data-volume* extension property accordingly when returning tenant configuration to a Tenant API client.
I suggest to explicitly add the *data-volume* spec to the Management API's ResourceLimits object instead.

sophokles73:
@ctron @dejanb WDYT?

sophokles73:
fixed via #1465

PR #1465, opened 2019-08-27
[#1460] Extend support to data limit in Management and Tenant APIs

kaniyan:
This PR implements #1460 
- Management and Tenant API now supports data limit configuration.
- Updated default resource limit checks implementation due to changes in the Tenant API.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/DataVolume.java
+++ b/core/src/main/java/org/eclipse/hono/util/DataVolume.java
@@ -1,0 +1,109 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.hono.util;
+
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
+
+import org.eclipse.hono.annotation.HonoTimestamp;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.fasterxml.jackson.annotation.JsonProperty;
+/**
+ * Data volume definition of the tenant resource limits.
+ */
+@JsonInclude(Include.NON_DEFAULT)
+public class DataVolume {
+
+    @JsonProperty(value = TenantConstants.FIELD_EFFECTIVE_SINCE, required = true)
+    @HonoTimestamp
+    private Instant effectiveSince;
+
+    @JsonProperty(TenantConstants.FIELD_MAX_BYTES)
+    private long maxBytes = TenantConstants.DEFAULT_MAX_BYTES;
+
+    @JsonProperty(TenantConstants.FIELD_PERIOD)
+    private DataVolumePeriod period;
+
+    /**
+     * Gets the point in time on which the data volume limit came into effect.
+     *
+     * @return The instant on which the data volume limit came into effective or 
+     *         {@code null} if not set.
+     */
+    public final Instant getEffectiveSince() {
+        return effectiveSince;
+    }
+
+    /**
+     * Sets the point in time on which the data volume limit came into effect.
+     * 
+     * @param effectiveSince the point in time on which the data volume limit came into effect
+     *                       and it comply to the {@link DateTimeFormatter#ISO_OFFSET_DATE_TIME}.
+     * @return  a reference to this for fluent use.
+     */
+    public final DataVolume setEffectiveSince(final Instant effectiveSince) {
+        this.effectiveSince = effectiveSince;
+        return this;
+    }
+
+    /**
+     * Gets the maximum number of bytes to be allowed for the time period defined by the
+     * {@link TenantConstants#FIELD_PERIOD_MODE} and 
+     * {@link TenantConstants#FIELD_PERIOD_NO_OF_DAYS}.
+     *
+     * @return The maximum number of bytes or {@link TenantConstants#DEFAULT_MAX_BYTES} 
+     *         if not set.
+     */
+    public final long getMaxBytes() {
+        return maxBytes;
+    }
+
+    /**
+     * Sets the maximum number of bytes to be allowed for the time period defined by the
+     * {@link TenantConstants#FIELD_PERIOD_MODE} and 
+     * {@link TenantConstants#FIELD_PERIOD_NO_OF_DAYS}.
+     *
+     * @param maxBytes The maximum number of bytes to be allowed.
+     * @return  a reference to this for fluent use.
+     * @throws IllegalArgumentException if the maximum number of bytes is set to less than -1.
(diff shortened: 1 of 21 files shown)


core/src/main/java/org/eclipse/hono/util/DataVolume.java
```diff
@@ -0,0 +1,142 @@
...
+    private DataVolumePeriod period;
+
+    /**
+     * Gets the point in time on which the data volume limit came into effect.
+     *
+     * @return The instant as an ISO string or 
+     *         {@code null} if not set.
+     */
+    @JsonGetter(TenantConstants.FIELD_EFFECTIVE_SINCE)
+    public String getEffectiveSinceAsString() {
```
sophokles73: we should be able to use a standard Jackson serializer as introduced in #1374

core/src/main/java/org/eclipse/hono/util/DataVolumePeriod.java
```diff
@@ -0,0 +1,75 @@
...
+
+
+/**
+ * Data volume period definition of the tenant resource limits .
+ */
+@JsonInclude(Include.NON_DEFAULT)
+public class DataVolumePeriod {
+
+    @JsonProperty(value = TenantConstants.FIELD_PERIOD_MODE, required = true)
+    private String mode;
```
sophokles73: we should probably use an enum for the supported modes, don't you think?

core/src/main/java/org/eclipse/hono/util/ResourceLimits.java
```diff
@@ -41,21 +42,46 @@
...
     public int getMaxConnections() {
         return this.maxConnections;
     }
 
+    /**
+     * Gets the data volume properties which are required for the message limit verification.
+     *
+     * @return The data volume properties.
+     */
+    public DataVolume getDataVolume() {
```
sophokles73: FMPOV all of the getters and setter should be declared final as there is no obvious reason for subclasses to override them, is there?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -537,7 +541,21 @@ public JsonObject getResourceLimits() {
     @JsonIgnore
     public TenantObject setResourceLimits(final JsonObject resourceLimits) {
         Objects.requireNonNull(resourceLimits);
-        return setProperty(TenantConstants.FIELD_RESOURCE_LIMITS, resourceLimits);
+        return setResourceLimits(resourceLimits.mapTo(ResourceLimits.class));
```
sophokles73: IMHO this will throw an exception if the given JSON cannot be deserialized ...

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -575,6 +577,36 @@ components:
                      - all
                      - none
 
+      DataVolume:
+         type: object
+         additionalProperties: false
+         properties:
+            "effective-since":
```
sophokles73: this property seems to be mandatory, we should also mark it as such in the value object ...

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -575,6 +577,36 @@ components:
...
+      Period:
+         type: object
+         additionalProperties: false
+         properties:
+            "mode":
+               type: string
+               description: The mode of the data usage caluclation. The supported modes by
+                            the default resource limit checks implementation are `days`
+                            and `monthly`.
+            "no-of-days":
```
sophokles73: both properties are required, aren't they?

sophokles73:
I think the Tenant API also needs to be extended to support the different *period modes* ...

core/src/main/java/org/eclipse/hono/util/DataVolumePeriod.java
kaniyan: Its a good point and I thought about it too. The supported modes by the reference limit checks implementation are _days_ and _monthly_. If any user decides to support more modes like _yearly_ or _quarterly_ it would be easier if this is just a string and he/she can easily handle in their custom limit checks implementation. Thats why I didn't change it to enum.

core/src/main/java/org/eclipse/hono/util/ResourceLimits.java
kaniyan: `Tenant` and classes used within `Tenant` don't follow this pattern and hence I got bit influenced from it. I don't think there is any obvious reason to override. I will declare the getters and setters as _final_ in `ResourceLimits`, `DataVolume` and `DataVolumePeriod`.

core/src/main/java/org/eclipse/hono/util/DataVolumePeriod.java
sophokles73: I see, makes sense to me.

site/documentation/content/api/management/device-registry-v1.yaml
sophokles73: so, at least the *mode* is required then, right? If so, we should mark it accordingly

kaniyan: I will mark the _mode_ as required.

core/src/main/java/org/eclipse/hono/util/ResourceLimits.java
```diff
@@ -41,21 +42,46 @@
      * 
      * @param maxConnections The maximum number of connections to set.
      * @return A reference to this for fluent use.
+     * @throws IllegalArgumentException if the maximum connections is set to less than -1.
      */
     public ResourceLimits setMaxConnections(final int maxConnections) {
+        if (maxConnections < -1) {
+            throw new IllegalArgumentException("maximum Connections can be set to >= -1");
```
dejanb: Captions and wording seems a bit off here ... Maybe "Maximum connections property must be set to value >= -1"

core/src/main/java/org/eclipse/hono/util/DataVolume.java
dejanb: Also, take a look at the recent changes done in #1463. I think it should be aligned with that.

kaniyan:
@sophokles73 @dejanb  I have pushed the changes based on the comments. Would you mind taking a look.

sophokles73:
@kaniyan can you squash and rebase? I can then merge the PR ...

kaniyan:
> @kaniyan can you squash and rebase? I can then merge the PR ...

@sophokles73 Done.

sophokles73:
@kaniyan I think we also need to update the Tenant API with the new data volume limit properties as well ...

kaniyan:
> @kaniyan I think we also need to update the Tenant API with the new data volume limit properties as well ...

Currently the Tenant and Management API share the data volume value objects. `TenantObject` is already equipped to set and retrieve _resource-limits_ which internally contain data volume limit properties. I didn't get it. Would you mind elaborating a bit more on what do you mean with need to update the Tenant API.

sophokles73:
@kaniyan The Tenant AP currently lacks the additional *mode* property:

https://www.eclipse.org/hono/docs/dev/api/tenant/#data-volume-configuration-format

kaniyan:
> @kaniyan The Tenant API currently lacks the additional _mode_ property:
> 
> https://www.eclipse.org/hono/docs/dev/api/tenant/#data-volume-configuration-format
I will update the docs.

kaniyan:
@sophokles73 Would you mind taking a look at the updated Tenant API documentation.

Issue #1448, opened 2019-08-19
Change semantics of default TTL on downstream events


sophokles73:
Currently, the protocol adapters set a TTL on downstream events as follows:
1. if message contains a TTL, keep it
2.1. otherwise, if a default TTL is set at the tenant level, use it
2.2. if a default TTL is set at the device level, use it.

In production environments, however, a service provider will probably want to be able to set an upper limit for message TTL in order to ensure reasonable resource utilization.

In such a scenario, the *max TTL* would not be settable by the tenant but by the service provider only and the protocol adapters should ensure that the TTL set on a downstream event does not exceed the max TTL. It seems to make sense to be able to support both, a *default TTL* and a *max TTL* at the tenant level so that downstream events can get a default TTL value but adapters can also use the max TTL as an upper limit if a default TTL is specified at the device level by the device owner.

calohmn:
So that means that if a *max TTL* but *no* *default TTL*  is set on the tenant level and if no default TTL is set for the device in the device registry, then the effective TTL on the downstream event will be the *max TTL*, correct?

In general, we have to take the tenant *max TTL* also into account for the implementation of  #267 and #266, where setting the TTL per event message shall get supported.

sophokles73:
> So that means that if a max TTL but no default TTL is set on the tenant level and if no default TTL is set for the device in the device registry, then the effective TTL on the downstream event will be the max TTL, correct?

FMPOV the downstream message should have no effective TTL at all in this case ...

calohmn:
> FMPOV the downstream message should have no effective TTL at all in this case ...

But no effective TTL means an infinite TTL, which is greater than the max TTL...

sophokles73:
That is true :-) So I guess you're right. In that case the effective TTL should be the max TTL.

Issue #267, opened 2017-07-07
Provide means to provide a TTL value for Events published to MQTT Adapter


sophokles73:
A device should have a way to indicate a time-to-live (TTL) for an event it publishes to Hono's Event API using the MQTT Adapter.

pellmann:
If we set TTL for AMQP, we need to be sure, that this is handled as expected from all involved intermediaries. I think this is not handled by Hono Messaging and I am not sure about the Dispatch Router (?). From the AMQP spec for TTL:

> Duration in milliseconds for which the message is to be considered "live". If this is set then a message expiration time will be computed based on the time of arrival at an intermediary. Messages that live longer than their expiration time will be discarded (or dead lettered). When a message is transmitted by an intermediary that was received with a ttl, the transmitted message's header SHOULD contain a ttl that is computed as the difference between the current time and the formerly computed message expiration time, i.e., the reduced ttl, so that messages will eventually die if they end up in a delivery loop.

For my understanding TTL is more important for command and control and maybe we should think about it together with this? WDYT?

ppatierno:
First of all, I can say that the dispatch router doesn't handle the TTL property. The router doesn't support the "store and forward" mechanism as a broker so the TTL doesn't make sense, because it means to count the time before message is delivered so it means "storing" something that isn't supported by the router.
The Hono messaging doesn't handle it as well and I agree on that.
Only the broker (Artemis) handles TTL.

I think that TTL is important for command and control but even useful for events.
Because we have a queue storing events and the consumer could be offline, it's better to have a mechanism to get rid of "old" events if it doesn't make sense when the consumer comes back online (this is the same concept as what happens with command & control with a really "old" command).

ppatierno:
Regarding the MQTT adapter we know that our supported protocol version 3.1.1 doesn't have something like the TTL (they added the "publication expiry interval" in the next v5).
I think that the only way is to add such value (as any other property) as a sort of "property bag" at the end of the topic as it happens for HTTP query string : 

/telemetry/DEFAULT_TENANT?ttl=10

pellmann:
I also think the property bag option would be the best, if we want to have this feature in the (3.1.1) MQTT adapter. An other possible solution could be an option in the device registry - so you have different TTL for different devices or tenants, maybe for different endpoints.

pellmann:
@ppatierno  - my understanding of the paragraph in the AMQP spec was not, that this would be only true for an intermediary, who stores the message. But you think so, right?

[...]

Issue #266, opened 2017-07-07
Provide means to provide a TTL value for Events published to HTTP Adapter


sophokles73:
A device should have a way to indicate a *time-to-live* (TTL) for an event it publishes to Hono's Event API using the HTTP Adapter.

PR #1467, opened 2019-08-29
[#1448] Use max-ttl to limit downstream event TTL.

sophokles73:
Support definition of a maximum TTL per tenant to limit the default TTL applied to downstream events from devices.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/ResourceLimits.java
+++ b/core/src/main/java/org/eclipse/hono/util/ResourceLimits.java
@@ -28,7 +28,9 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 public class ResourceLimits {
 
     @JsonProperty(TenantConstants.FIELD_MAX_CONNECTIONS)
-    private int maxConnections = TenantConstants.DEFAULT_MAX_CONNECTIONS;
+    private int maxConnections = TenantConstants.UNLIMITED_CONNECTIONS;
+    @JsonProperty(TenantConstants.FIELD_MAX_TTL)
+    private long maxTtl = TenantConstants.UNLIMITED_TTL;
 
     @JsonProperty(TenantConstants.FIELD_DATA_VOLUME)
     private DataVolume dataVolume;
@@ -55,11 +57,33 @@ public class ResourceLimits {
     /**
      * Gets the maximum number of connected devices a tenant supports.
      * 
-     * @return The maximum number of connections or {@link TenantConstants#DEFAULT_MAX_CONNECTIONS}
+     * @return The maximum number of connections or {@link TenantConstants#UNLIMITED_CONNECTIONS}
      *         if not set.
      */
     public final int getMaxConnections() {
         return this.maxConnections;
+    }
+
+    /**
+     * Sets the maximum time-to-live to use for events published by
+     * devices of a tenant.
+     * 
+     * @param maxTtl The time-to-live in seconds.
+     * @return A reference to this for fluent use.
+     */
+    public ResourceLimits setMaxTtl(final long maxTtl) {
+        this.maxTtl = maxTtl;
+        return this;
+    }
+
+    /**
+     * Gets the maximum time-to-live to use for events published by
+     * devices of a tenant.
+     * 
+     * @return The time-to-live in seconds.
+     */
+    public long getMaxTtl() {
+        return this.maxTtl;
     }
 
     /**
--- a/core/src/main/java/org/eclipse/hono/util/ResourceLimits.java
+++ b/core/src/main/java/org/eclipse/hono/util/ResourceLimits.java
@@ -70,8 +70,12 @@ public class ResourceLimits {
      * 
      * @param maxTtl The time-to-live in seconds.
      * @return A reference to this for fluent use.
+     * @throws IllegalArgumentException if the time-to-live is set to less than -1.
      */
     public ResourceLimits setMaxTtl(final long maxTtl) {
+        if (maxTtl < -1) {
+            throw new IllegalArgumentException("Maximum time-to-live property must be set to value >= -1");
+        }
         this.maxTtl = maxTtl;
         return this;
     }
--- a/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
@@ -1222,8 +1222,8 @@ public abstract class AbstractProtocolAdapterBase<T extends ProtocolAdapterPrope
                 final boolean isEvent = target.getEndpoint().equals(EventConstants.EVENT_ENDPOINT);
                 if (isEvent && message.getTtl() == 0 && Number.class.isInstance(prop.getValue())) {
                     final long defaultTtl = ((Number) prop.getValue()).longValue();
-                    if (maxTtl > -1 && defaultTtl > maxTtl) {
-                            message.setTtl(maxTtl);
+                    if (maxTtl != TenantConstants.UNLIMITED_TTL && defaultTtl > maxTtl) {
+                        message.setTtl(maxTtl);
                     } else {
                         message.setTtl(defaultTtl);
                     }
--- a/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
@@ -1173,14 +1173,30 @@ public abstract class AbstractProtocolAdapterBase<T extends ProtocolAdapterPrope
             final TenantObject tenant,
             final JsonObject registrationInfo) {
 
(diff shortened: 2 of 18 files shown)


service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -1193,15 +1209,24 @@ private void addProperties(
...
         defaults.forEach(prop -> {
 
             switch (prop.getKey()) {
             case MessageHelper.SYS_HEADER_PROPERTY_TTL:
                 final boolean isEvent = target.getEndpoint().equals(EventConstants.EVENT_ENDPOINT);
                 if (isEvent && message.getTtl() == 0 && Number.class.isInstance(prop.getValue())) {
-                    message.setTtl(((Number) prop.getValue()).longValue());
+                    final long defaultTtl = ((Number) prop.getValue()).longValue();
+                    if (maxTtl > -1 && defaultTtl > maxTtl) {
+                            message.setTtl(maxTtl);
```
calohmn: indentation

core/src/main/java/org/eclipse/hono/util/ResourceLimits.java
```diff
@@ -55,13 +57,35 @@ public final ResourceLimits setMaxConnections(final int maxConnections) {
...
     }
 
+    /**
+     * Sets the maximum time-to-live to use for events published by
+     * devices of a tenant.
+     * 
+     * @param maxTtl The time-to-live in seconds.
+     * @return A reference to this for fluent use.
+     */
+    public ResourceLimits setMaxTtl(final long maxTtl) {
```
calohmn: It would be good to check whether `maxTtl <-1` here and throw an IllegalArgumentException if that's the case.
Otherwise there would be an exception in the proton code if `maxTtl < -1` (in Message.setTtl()).

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -526,8 +526,19 @@ components:
...
+                            from devices of this tenant.
+                            A value of -1 (the default) indicates that no limit is set.
+            "max-ttl":
+               type: integer
+               default: -1
+               description: The maximum time-to-live (in seconds) to use for events published by
+                            devices of this tenant. Any default TTL value specified
+                            at either the tenant or device level will be limited to
+                            the max value specified here.
+                            If this property is set to a value other than -1 and no
```
calohmn: "greater than"  instead of "other than" would be more concise IMHO.

site/documentation/content/api/tenant/index.md
```diff
@@ -201,13 +201,25 @@ This allows for future *well-known* additions and also allows to add further inf
...
-| *data-volume*            | *no*      | *object*      | `-`           | The maximum data volume allowed for the given tenant. Refer to  [Data Volume Configuration Format]({{< relref "#data-volume-configuration-format" >}}) for details.|
+| Name                | Mandatory | JSON Type     | Default Value | Description |
+| :------------------ | :-------: | :------------ | :------------ | :---------- |
+| *max-connections*   | *no*      | *number*      | `-1`          | The maximum number of concurrent connections allowed from devices of this tenant. The default value `-1` indicates that no limit is set. |
+| *max-ttl*           | *no*      | *number*      | `-1`          | The maximum time-to-live (in seconds) to use for events published by  devices of this tenant. Any default TTL value specified at either the tenant or device level will be limited to the max value specified here. If this property is set to a value other than `-1` and no default TTL is specified for a device, the max value will be used for events published by the device. A value of `-1` (the default) indicates that no limit is set. |
```
calohmn: "greater than"  instead of "other than" would be more concise IMHO.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
sophokles73: fixed

site/documentation/content/api/tenant/index.md
```diff
@@ -201,13 +201,25 @@ This allows for future *well-known* additions and also allows to add further inf
...
+| *data-volume*       | *no*      | *object*      | `-`           | The maximum data volume allowed for the given tenant. Refer to  [Data Volume Configuration Format]({{< relref "#data-volume-configuration-format" >}}) for details.|
 
 Protocol adapters SHOULD use the *max-connections* property to determine if a device's connection request should be accepted or rejected.
 
+Protocol adapters SHOULD use the *max-ttl* property to determine the *effective time-to-live* for events published by devices as follows:
+
+1. If a non-default *max-ttl* is set for the tenant, use that value as the *effective ttl*, otherwise set *effective ttl* to `-1`.
+1. If the event published by the device
+   1. contains a *ttl* property and *effective ttl* is not `-1` and the *ttl* value provided by the device is smaller than the
```
calohmn: Concerning the "event .. contains a ttl property" part: This might be confusing here because with #267, #266 still open, our protocol adapters don't support adopting a specific ttl for an event yet.

site/documentation/content/api/tenant/index.md
```diff
@@ -201,13 +201,25 @@ This allows for future *well-known* additions and also allows to add further inf
...
+      *effective ttl*, use the device provided *ttl* value as the new *effective ttl*.
+   1. does not contain a *ttl* property but a default *ttl* value is configured for the device (either at the tenant or the device level)
```
calohmn: I think it would be good to mention that device level has precedence here.

calohmn: Speaking of the default "ttl" on the tenant or device level: 
I think it would be helpful to have this `ttl` JSON object in the `defaults` object be explicitly documented in the Tenant Information Format and the Device Registry Docs, as well as in the device-registry-v1.yaml. 
Doesn't have to be done in this PR though.

site/documentation/content/api/tenant/index.md
sophokles73: The AMQP adapter does. The text is already complicated enough so I'd rather not make it even more complicated by adding another exception here ... I also do not think that people will have a problem figuring out which adapter supports this and which doesn't.

site/documentation/content/api/tenant/index.md
sophokles73: The default properties are a very generic construct and I am not sure, if we want to prescribe any semantics in the Tenant API. Also, the adapter user guides already define, which default properties they support ...

sophokles73:
@calohmn I have pushed some changes ...

core/src/main/java/org/eclipse/hono/util/ResourceLimits.java
calohmn: I was wondering on how to deal with a maxTtl value of 0.
The Proton MessageImpl.setTtl(0) implementation seems inconsistent here (such a header value only being set if there is already a header set).
Of course the main question is whether there is a use case for this. It could potentially be used to effectively disable the event transport (temporarily), while still allowing telemetry - provided Artemis really uses the 0 value there to discard message directly.
@sophokles73 Do you think it would be worth it to consider such a case, testing it and making the behaviour for maxTtl=0 consistent?

sophokles73: my understanding is that proton won't set a message's ttl to 0, but instead use this value to *clear* an existing value, i.e. remove the ttl property altogether. A TTL doesn't make much sense because it would mean that the receiver would discard the message immediately on reception, right?
So, no, I don't think that we need to spend much effort on it ...

calohmn: I didn't see a 0 ttl value excluded from being put into the amqp message headers on the wire, neither it being ignored in Artemis. And my guess towards a use case was explicitly based upon Artemis potentially discarding all such messages directly.
But, that's just a guess from the code and I haven't tested it.
And yes, it's probably not worth the effort.

site/documentation/content/api/tenant/index.md
calohmn: Ah, yes, didn't check the adapter user guides. It's not very prominent there though.
And the restriction on the value not being adopted if max-ttl is set is not mentioned there.
Having it all together in the device-registry-v1.yaml with cross references between ttl and max-ttl sounds more user-friendly to me. But again, this can be done in a separate PR.

sophokles73:
> The unit for the new max-ttl value is documented here as seconds.
The AMQP spec however uses milliseconds for the ttl header (which the max-ttl value is mapped to).

Thanks for spotting, @calohmn ! Do we agree, though, that we want to specify the TTL in seconds rather than millis?

calohmn:
@sophokles73 yes, using seconds looks more convenient here.
I think we should use seconds for the `ttl` device/tenant `defaults` entry as well then, document it explicitly and make clear that its unit is different from the one used in the AMQP message header (see also #1470).

sophokles73:

PR #1473, opened 2019-09-03
[#1471] Fix broken "terminal error" connection handling

ctron:


client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -376,11 +366,7 @@ private void connect(
                         conAttempt -> {
                             connecting.compareAndSet(true, false);
                             if (conAttempt.failed()) {
-                                if ((!isIgnoreTerminalCondition()) && isTerminalConnectionError(conAttempt.cause())) {
-                                    failConnectionAttempt(conAttempt.cause(), connectionHandler);
-                                } else {
-                                    reconnect(conAttempt.cause(), connectionHandler, disconnectHandler);
-                                }
+                                reconnect(conAttempt.cause(), connectionHandler, disconnectHandler);
```
sophokles73: IMHO we should still log out specific error messages for the former terminal types of connection errors as we do in the *failConnectionAttempt* method (which is now never invoked with the default reconnect attempts set to -1). Otherwise it might be hard for operators to fix the problem ...

ctron: But that wouldn't be any "terminal errors" but more something like "noteworthy errors".

ctron: I added the logging, based on the previous classification.

PR #1475, opened 2019-09-03
[#1188]Use metrics to monitor tenant timeout.

b-abel:
This PR addresses #1188 and provides a better solution for "tenant timeouts" than the one which has been implemented in PR #1416. 
Instead of having individual timeouts for each AMQP link, the timeout is now centrally monitored by the metrics implementation, which publishes an event when a tenant times out. It is intended to be used for #1227 as well.
Please see https://github.com/eclipse/hono/pull/1456#issuecomment-526076614 for the reasoning.


b-abel:
@calohmn Maybe you can do a review as well?

calohmn:
@b-abel Can you resolve the merge conflict here?

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -30,9 +32,13 @@
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.Timer;
 import io.micrometer.core.instrument.Timer.Sample;
+import io.vertx.core.Vertx;
 
 /**
- * Micrometer based metrics implementation.
+ * Micrometer based metrics implementation. Checks if tenant {@link ProtocolAdapterProperties#getIdleTenantTimeout()} is
+ * exceeded and publishes the tenantId on the event bus address {@link Constants#EVENT_BUS_ADDRESS_TENANT_TIMED_OUT}
+ * once a tenant times out.
```
calohmn: Speaking of a "tenant time out" might be confusing.
How about "once the tenant idle timeout has been reached"?
And maybe use `<p>Also checks if [...]` above, making clear the idle timeout check is not the main purpose of the class.

client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClient.java
```diff
@@ -221,4 +223,28 @@ protected static final void setApplicationProperties(final Message msg, final Ma
...
+     * <p>
+     * Subclasses need to call this method after the sender has been set.
+     */
+    protected final void registerForTenantTimeoutEvents() {
+        if (this.sender != null) {
+            final String address = this.sender.getTarget().getAddress();
+            if (address != null) {
+                final String tenantId = ResourceIdentifier.fromString(address).getTenantId();
+                if (tenantId != null) {
+                    connection.getVertx().eventBus().consumer(Constants.EVENT_BUS_ADDRESS_TENANT_TIMED_OUT, msg -> {
```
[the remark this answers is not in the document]
b-abel: You are right. The event bus consumer needs to be in another class.

b-abel:
The event bus consumer is now registered in the (caching) client factories. This is IMHO the right places for it with the current design. They manage the creation of clients, their caching and their removal. Now they close the links when they receive a tenant timeout event and remove the client from the cache.

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
```diff
@@ -13,17 +13,27 @@
...
+     * disables automatic tenant timeout.
+     */
+    public static final Duration DEFAULT_IDLE_TENANT_TIMEOUT = Duration.ZERO;
+
     private boolean authenticationRequired = true;
     private boolean jmsVendorPropsEnabled = false;
     private boolean defaultsEnabled = true;
     private long eventLoopBlockedCheckTimeout = 5000L;
     private int maxConnections = 0;
+    private Duration idleTenantTimeout = DEFAULT_IDLE_TENANT_TIMEOUT;
```
calohmn: `idleTenantTimeout` sounds a bit odd to me, how about using `tenantIdleTimeout` instead?

client/src/main/java/org/eclipse/hono/client/impl/DeviceConnectionClientFactoryImpl.java
```diff
@@ -74,4 +78,12 @@ protected void onDisconnect() {
     private void removeDeviceConnectionClient(final String tenantId) {
         deviceConnectionClientFactory.removeClient(DeviceConnectionClientImpl.getTargetAddress(tenantId));
     }
+
+    private void handleTenantTimeout(final Message<String> msg) {
```
calohmn: There are 4 places with a similar `handleTenantTimeout` method now. I had a look here and could imagine getting rid of the code duplication with a method like:
````
    private static <T extends RequestResponseClient> void closeAndRemoveClient(
            final CachingClientFactory<T> cachingClientFactory, Supplier<String> clientKeySupplier) {
        final String key = clientKeySupplier.get();
        final T client = cachingClientFactory.getClient(key);
        if (client != null) {
            client.close(v -> cachingClientFactory.removeClient(key));
        }
    }
````
invoking it like this:
````
        connection.getVertx().eventBus().consumer(Constants.EVENT_BUS_ADDRESS_TENANT_TIMED_OUT,
                msg -> {
                    final String tenantId = msg.body().toString();
                    closeAndRemoveClient(deviceConnectionClientFactory,
                            () -> DeviceConnectionClientImpl.getTargetAddress(tenantId));
                });
````
But this would still look like an intermediate step to me, as I could imagine having a `CachingClientFactory.closeAndRemoveClient(Supplier<String> clientKeySupplier)` in the end. For this, the `CachingClientFactory<T>` class has to be changed so that the `T` extends an interface with a `close` method though. So refactoring the `handleTenantTimeout` methods could be postponed until after the `CachingClientFactory` has been changed. WDYT?

File: client/src/test/java/org/eclipse/hono/client/impl/AbstractHonoClientFactoryTest.java [hunk 5]
```diff
@@ -0,0 +1,123 @@
...
+ * 
+ * @param <T> The type of the client that the factory builds.
+ */
+@RunWith(VertxUnitRunner.class)
+public abstract class AbstractHonoClientFactoryTest<T> {
+
+    private static long TIMEOUT = 1000L;
+
+    /**
+     * Gets a future with a client of type `T` from it's factory.
```
calohmn: typo: its

client/src/main/java/org/eclipse/hono/client/impl/DeviceConnectionClientFactoryImpl.java
b-abel: Yes, I would welcome a refactoring that minimizes code duplication and agree that this should be done later on and not within this PR.

b-abel:
The requested changes are included. I would like to squash, rebase and add release notes before merging.

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -61,28 +68,37 @@
...
+    private long tenantIdleTimeout = DEFAULT_TENANT_IDLE_TIMEOUT;
 
     private LegacyMetrics legacyMetrics;
 
     /**
      * Creates a new metrics instance.
      * 
      * @param registry The meter registry to use.
+     * @param vertx The Vert.x instance to use.
      * @throws NullPointerException if registry is {@code null}.
```
calohmn: registry or vertx

site/documentation/content/admin-guide/amqp-adapter-config.md
```diff
@@ -20,6 +20,7 @@ The following table provides an overview of the configuration variables and corr
...
 | `HONO_AMQP_DEFAULTS_ENABLED`<br>`--hono.amqp.defaultsEnabled` | no | `true` | If set to `true` the protocol adapter uses *default values* registered for a device to augment messages published by the device with missing information like a content type. In particular, the protocol adapter adds default values registered for the device as (application) properties with the same name to the AMQP 1.0 messages it sends downstream to the AMQP Messaging Network. |
+| `HONO_AMQP_TENANT_IDLE_TIMEOUT`<br>`--hono.amqp.tenantIdleTimeout` | no | `0ms` | The duration after which the protocol adapter removes local state of the tenant (e.g. open AMQP links) with an amount and an unit, e.g. `2h` for 2 hours. See the [Spring Boot documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-conversion-duration) for an explanation of the format. The value `0ms` disables the timeout. |
```
calohmn: typo: "an unit" -> "a unit" (also in other places)

b-abel:
Done.

PR #1481, opened 2019-09-10
[#1480] MQTT-Adapter: fix wrong topic for commands sent to gateway; add documentation

### LINKED ISSUE #1480: MQTT-Adapter: fix wrong topic for commands sent to gateway; add documentation
With #930 support has been added for command messages to be redirected to the current gateway that the target device is connected to.

In the case of the MQTT adapter, this means:
- an authenticated gateway will subscribe for commands (not for a particular device but for all devices using that gateway),
- a northbound application will send a command to a particular device (not knowing its gateway id),
- the protocol adapter will determine the last known gateway for that device and publish the command to it.

What is missing here is that the topic used when publishing the command to the gateway doesn't contain the original device id. The topic currently looks just like this: `command///req/xyz/light`. Instead, the device id should be set in there (e.g. `command//4712/req/xyz/light`) so that the gateway can properly handle the command.

Also, documenation for that scenario is missing still.

calohmn:
This fixes #1480.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/Command.java
+++ b/client/src/main/java/org/eclipse/hono/client/Command.java
@@ -234,6 +234,10 @@ public final class Command {
 
     /**
      * Gets the device's identifier.
+     * <p>
+     * In the case that the command got redirected to a gateway,
+     * the id returned here is a gateway id. See {@link #getOriginalDeviceId()}
+     * for the original device id in that case.
      *
      * @return The identifier.
      * @throws IllegalStateException if this command is invalid.
@@ -241,6 +245,48 @@ public final class Command {
     public String getDeviceId() {
         if (isValid()) {
             return deviceId;
+        } else {
+            throw new IllegalStateException("command is invalid");
+        }
+    }
+
+    /**
+     * Checks whether the command is targeted at a gateway.
+     * <p>
+     * This is the case when the commands got redirected and hence
+     * the device id (ie. the gateway id in that case) is different
+     * from the original device id.
+     *
+     * @return {@code true} if the device id is a gateway id.
+     * @throws IllegalStateException if this command is invalid.
+     */
+    public boolean isTargetedAtGateway() {
+        if (isValid()) {
+            final String originalDeviceId = getOriginalDeviceId();
+            return originalDeviceId != null && !originalDeviceId.equals(getDeviceId());
+        } else {
+            throw new IllegalStateException("command is invalid");
+        }
+    }
+
+    /**
+     * Gets the device identifier used in the original command. It is extracted from the
+     * <em>to</em> property of the command AMQP message.
+     * <p>
+     * This id differs from {@link #getDeviceId()} if the command got redirected to a gateway
+     * ({@link #getDeviceId()} returns the gateway id in that case).
+     *
+     * @return The identifier.
+     * @throws IllegalStateException if this command is invalid.
+     */
+    public String getOriginalDeviceId() {
+        if (isValid()) {
+            // commands directed at the legacy control endpoint didn't have to have the message address set
+            // (and it's content didn't get checked), that's why the 'deviceId' field is used as fallback here
+            final String addressDeviceId = message.getAddress() != null
+                    ? ResourceIdentifier.fromString(message.getAddress()).getResourceId()
+                    : null;
+            return addressDeviceId != null ? addressDeviceId : deviceId;
         } else {
             throw new IllegalStateException("command is invalid");
         }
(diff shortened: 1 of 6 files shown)


client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -246,6 +250,47 @@ public String getDeviceId() {
...
+        if (isValid()) {
+            final String originalDeviceId = getOriginalDeviceId();
+            return originalDeviceId != null && !originalDeviceId.equals(getDeviceId());
+        } else {
+            throw new IllegalStateException("command is invalid");
+        }
+    }
+
+    /**
+     * Gets the device identifier used in the original command.
```
sophokles73: IMHO the comment should explicitly explain that the value returned is extracted from the AMQP message's *to* property.

client/src/main/java/org/eclipse/hono/client/impl/DelegatedCommandSenderImpl.java
```diff
@@ -201,14 +197,10 @@ static String getTargetAddress(final String tenantId, final String deviceId) {
...
 
-    private static Message createDelegatedCommandMessage(final Message originalMessage, final String targetAddress,
-            final String replyToAddress) {
-        Objects.requireNonNull(targetAddress);
+    private static Message createDelegatedCommandMessage(final Message originalMessage, final String replyToAddress) {
         Objects.requireNonNull(originalMessage);
         // copy original message
         final Message msg = MessageHelper.getShallowCopy(originalMessage);
-        // set target address
-        msg.setAddress(targetAddress);
```
sophokles73: why is this no longer necessary?

site/documentation/content/concepts/command-and-control.md
```diff
@@ -59,11 +59,21 @@ and *one-way* commands on the topic:
...
-while unauthenticated devices have to fully specify their `${tenant}` and `${device-id}` during the subscription.
+while unauthenticated devices have to fully specify their `${tenant}` and `${device-id}` in the subscription topic:
+
+* `command/${tenant-id}/${device-id}/req/#`
+
+Authenticated gateways subscribe to
+
+* `command/+/+/req/#`
+
+and will receive commands on a topic where the target device id is set.
```
sophokles73: and will receive commands on a topic that contains the target device id.

site/documentation/content/user-guide/mqtt-adapter.md
```diff
@@ -259,6 +259,27 @@ and *one-way* commands to the topic:
...
+
+If a device is configured in such a way that there can be *one* gateway, acting on behalf of the device, a command sent to this device will by default be directed to that gateway.
+
+If a device is configured to be used with *multiple* gateways, the particular gateway that last acted on behalf of the device will be the target that commands for that device will be routed to. The mapping of device and gateway last used by the device is updated whenever a device sends a telemetry, event or command response message via the gateway. This means that for a device configured to be used via multiple gateways to receive commands, the device first has to send at least one telemetry or event message to establish which gateway to use for receiving commands for that device.
+
+An authenticated gateway MUST use the following topic filter to subscribe to commands:
```
sophokles73: ... to commands for all the devices it acts on behalf of:

site/documentation/content/user-guide/mqtt-adapter.md
```diff
@@ -259,6 +259,27 @@ and *one-way* commands to the topic:
...
+* `command/+/+/req/#`
+
+**Example**
+
+    mosquitto_sub -v -u 'gw@DEFAULT_TENANT' -P gw-secret -t command/+/+/req/#
+
+The adapter will then publish commands for devices, that the gateway has acted on behalf of, to topic:
+
+* for *Request/Response* commands: `command//${device-id}/req/${req-id}/${command}[/*][/property-bag]`
+* for *one-way* commands: `command//${device-id}/req//${command}[/*][/property-bag]`
```
sophokles73: the topic we send commands to cannot contain a property bag, can it?

client/src/main/java/org/eclipse/hono/client/impl/DelegatedCommandSenderImpl.java
calohmn: It was only necessary because we used an anonymous relay link here. But with #1474 this has been changed. We use a non-empty link address now and therefore we don't need to explicitly set the message address.

The address of the original command message is supposed to be preserved here. This is relevant for the case that the message is to be delegated to a gateway.

site/documentation/content/user-guide/mqtt-adapter.md
calohmn: Right, I've removed the `[/*][/property-bag]` part also in the other places on the page.

calohmn:
@sophokles73 I've made the corresponding changes.

client/src/main/java/org/eclipse/hono/client/impl/DelegatedCommandSenderImpl.java
sophokles73: > The address of the original command message is supposed to be preserved here. 

it will be copied by the *getShallowCopy* method?

calohmn: yes (along with the other message properties).

PR #1490, opened 2019-09-13
Extract AddressAuthzHelper out of SimpleAuthenticationServer

calohmn:
This extracts the code for supporting the Qpid Dispatch Router `ADDRESS-AUTHZ` capability out of the `SimpleAuthenticationServer` into its own class in the service-base module.

This makes it easier to write own Hono Auth server implementations that also support the Qpid authorization feature.

### DIFF
--- a/services/auth/src/main/java/org/eclipse/hono/service/auth/impl/SimpleAuthenticationServer.java
+++ b/services/auth/src/main/java/org/eclipse/hono/service/auth/impl/SimpleAuthenticationServer.java
@@ -13,24 +13,13 @@
 
 package org.eclipse.hono.service.auth.impl;
 
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
-import java.util.stream.Collectors;
-
-import org.apache.qpid.proton.amqp.Symbol;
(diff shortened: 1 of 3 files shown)


services/auth/src/main/java/org/eclipse/hono/service/auth/impl/SimpleAuthenticationServer.java
```diff
@@ -110,17 +81,16 @@ protected void setRemoteConnectionOpenHandler(final ProtonConnection connection)
...
-     * 
+     *
      * @param connection The connection opened by the peer.
      */
     @Override
     protected void processRemoteOpen(final ProtonConnection connection) {
-        final boolean isAddressAuthz = Arrays.stream(connection.getRemoteDesiredCapabilities())
-                .anyMatch(symbol -> symbol.equals(CAPABILITY_ADDRESS_AUTHZ));
-        if (isAddressAuthz) {
+        final AddressAuthzProcessor addressAuthzProcessor = new AddressAuthzProcessor();
```
dejanb: Wondering as this class doesn't hold any state if this all could be just static calls?

calohmn: Right, I've changed the methods accordingly and renamed the class to `AddressAuthzHelper`.

PR #1494, opened 2019-09-16
[#1488] Add 'device_id' property for gateway commands; update docs

### LINKED ISSUE #1488: AMQP-Adapter: include target device id for commands sent to gateway; add documentation
Similar to #1480:
With #930 support has been added for command messages to be redirected to the current gateway that the target device is connected to.

In the case of the AMQP adapter, this means:
- an authenticated gateway will open a receiver link for commands (not for a particular device but for all devices using that gateway),
- a northbound application will send a command to a particular device (not knowing its gateway id),
- the protocol adapter will determine the last known gateway for that device and publish the command to it.

What is missing here is that the command AMQP message sent to the gateway doesn't contain the original device id. The original "To" address in the message sent by the application could be preserved. In addition, an extra property with the device id could be added.

Also, documenation for that scenario is missing still.

calohmn:
This fixes #1488:
When sending commands to gateway devices, a "device_id" application property is added to the command AMQP message, containing the device id of the original command message.

The AMQP adapter documentation has been updated to describe the scenario that an authenticated gateway wants to receive commands.

### DIFF
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
@@ -842,6 +842,9 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
             // fall back to message ID
             msg.setCorrelationId(msg.getMessageId());
         }
+        if (command.isTargetedAtGateway()) {
+            MessageHelper.addDeviceId(msg, command.getOriginalDeviceId());
+        }
 
         // TODO time out waiting for disposition update
         sender.send(msg, delivery -> {
--- a/site/documentation/content/user-guide/amqp-adapter.md
+++ b/site/documentation/content/user-guide/amqp-adapter.md
@@ -166,7 +166,7 @@ A device that publishes data on behalf of another device is called a gateway dev
 
 **Examples**
 
-A Gateway connecting to the adapter using `gw@DEFAULT_TENANT` as username and `gw-secret` as password and then publishing some JSON data for device `4711`:
+A gateway connecting to the adapter using `gw@DEFAULT_TENANT` as username and `gw-secret` as password and then publishing some JSON data for device `4711`:
 
     # in directory: hono/cli/target/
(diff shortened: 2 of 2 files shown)


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -842,6 +842,9 @@ protected void onCommandReceived(final TenantObject tenantObject, final ProtonSe
             // fall back to message ID
             msg.setCorrelationId(msg.getMessageId());
         }
+        if (command.isTargetedAtGateway()) {
+            MessageHelper.addProperty(msg, MessageHelper.APP_PROPERTY_DEVICE_ID, command.getOriginalDeviceId());
```
sophokles73: MessageHelper.addDeviceId()

site/documentation/content/user-guide/amqp-adapter.md
```diff
@@ -258,6 +258,7 @@ A device MUST use the following source address in its *attach* frame to open a l
 
 * `command` (authenticated device)
 * `command/${tenant}/${device-id}` (unauthenticated device)
+* `command` (authenticated gateway)
```
sophokles73: what about an unauthenticated gateway? Why not simply amend the existing lines with *device/gateway*?

calohmn: As far as I see it, the concept of an unauthenticated gateway is generally not supported in Hono, or is it?
I've amended the first line by adding "/gateway".

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
calohmn: Changed accordingly.

site/documentation/content/user-guide/amqp-adapter.md
sophokles73: hmm, I guess you are right. Without authentication we are not able to determine the gateway's device ID. Have we explicitly stated that somewhere in the docs?

calohmn: As far as I see, it isn't *explicitly* mentioned in the docs.
We could add a corresponding note in the protocol adapter user guides in the `Device authentication` chapters. I can have a look at that.

PR #1499, opened 2019-09-17
Improved Azure guidance: Azure Service Bus, ARM templates and service endpoints

kaizimmerm:
Primarily improvements for Azure deployments:

- [Azure Resource Manager (ARM)](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-overview) templates for an automated infrastructure deployment.
- [Azure Service Bus](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-messaging-overview) for the _Eclipse Hono™ AMQP 1.0 Messaging Network_ instead of a self hosted ActiveMQ Artemis.
- [Virtual Network (VNet) service endpoints](https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-service-endpoints-overview) ensure protected communication between AKS and Azure Service Bus.
- Static IPs and DNS names for Azure load balancers provisioned.
- Created a dedicated page in documentation for Azure as this now beyond k8s deployment.

I had to change a few non Azure related aspects as well to make it happen:

- Helm values extended to define annotations for services as well as load balancer IPs. The default did not change though.
- Upgraded dispatch router to 1.8 as only this one works with Azure Service Bus.
- Adapted the HTTP load-test script. It seemed it was based on a different/deprecated registry API?
- I configured in the router config an address for event as well to have a balanced event distribution over multiple clients. Maybe I got the event concept wrong but I would expect this behaviour not only for telemetry.

Let me know what you think,

Kai


bom/pom.xml
```diff
@@ -31,7 +31,7 @@
     <artemis.image.name>enmasseproject/activemq-artemis:2.2.0-4</artemis.image.name>
     <assertj-core.version>3.11.1</assertj-core.version>
     <californium.version>2.0.0-M16</californium.version>
-    <dispatch-router.image.name>quay.io/enmasse/qdrouterd-base:1.7.0</dispatch-router.image.name>
+    <dispatch-router.image.name>quay.io/enmasse/qdrouterd-base:1.8.0</dispatch-router.image.name>
```
sophokles73: sadly, we cannot simply update the image to a new minor version as Dispatch Router is a pre-req for Hono, i.e. we need to file a CQ for this.

deploy/src/main/deploy/azure/managed-premium-retain.yaml
```diff
@@ -0,0 +1,9 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: managed-premium-retain
+provisioner: kubernetes.io/azure-disk
+reclaimPolicy: Retain
+parameters:
+  storageaccounttype: Premium_LRS
+  kind: Managed
```
sophokles73: EOL?

kaizimmerm:
Hi @sophokles73 ,

thanks for your feedback. Your observations are correct. 

However, I would like keep the PR together. I made no statement that the Service Bus option changes the overall scalability/reliability as it is still a single Qpid and not a full network. I added in fact (similar to the OpenShift guidance) a warning on top of the guide that we are still taking about a demo setup.

But I do believe that removing Artemis is one less thing to worry about even for a demo or for development purposes.

What I can offer is to make the Azure guidance more explicit that SB is simply an optional alternative?

Kai

bom/pom.xml
kaizimmerm: Ok, I will remove it and simply state in the guide that when taking the Service Bus option you have to deploy a 1.8

deploy/src/main/deploy/helm/templates/dispatch-router/config/qdrouterd-with-broker.yaml
```diff
@@ -66,23 +66,38 @@
   ["connector", {
     "sslProfile": "external",
     "name": "broker",
-    "host": "{{ .Release.Name }}-artemis",
     "port": 5671,
     "role": "route-container",
+{{- if .Values.azure.serviceBus.enabled }}
```
dejanb: It might make more sense to create and use completely new configuration (qrouterd-with-azure) than to template this one. It might get complicated to see what it actually resolves to in different scenarios.

deploy/src/main/deploy/helm/values.yaml
```diff
@@ -3,16 +3,28 @@
 # Declare variables to be passed into your templates.
 
 # amqpMessagingNetworkDeployExample indicates whether the example AMQP Messaging Network
-# consisting of a single Dispatch Router and Artemis broker should be
+# consisting of a single Dispatch Router and Artemis broker (or Azure ServiceBus) should be
```
dejanb: I don't want to play Azure role by any means, but based on your discussions, maybe we can make this general external AMQP broker option? So, you can either use internally deployed Artemis broker, or configure parameters to the external one?

deploy/src/main/deploy/helm/templates/dispatch-router/config/qdrouterd-with-broker.yaml
kaizimmerm: Good idea, I will change that. 👍

deploy/src/main/deploy/helm/values.yaml
kaizimmerm: I like the idea, so _amqpMessagingNetworkDeployExample_  would refer to the router only and then there is an internal and external option on the broker?.

I will create a proposal.

kaizimmerm: @dejanb , @sophokles73 thoughts?

```yaml
# amqpMessagingNetworkExample indicates whether the example AMQP Messaging Network
# consisting of a single Dispatch Router and Broker should be
# deployed and used. By default an internal Broker is deployed.
# As alternative an external Broker can be configured as well.
amqpMessagingNetworkExample:
  enabled: true
  # dispatchRouter.imageName contains the name (including tag)
  # of the container image to use for the example AQMP Messaging Network Router.
  dispatchRouter:
    imageName: ${dispatch-router.image.name}
    svc:
      annotations: {}
      loadBalancerIP:
  # internalBroker.imageName contains the name (including tag) of the container
  # image to use for the example AMQP Messaging Network Broker
  internalBroker:
    enabled: true
    imageName: ${artemis.image.name}
  # alternative external Broker. A qrouterd configuration file of 'type' has to be available.
  externalBroker:
    enabled: false
    type: servicebus
    saslUsername:
    saslPassword:
    host:
```

Based on the comment above I would then introduce a broker config "qrouterd-with-servicebus.yaml"

dejanb: @kaizimmerm This looks good to me. Thanks

kaizimmerm: @dejanb after testing the varaiants I managed to get it even simpler. Plus the the _amqpMessagingNetworkExample.broker.type_ value drives the decision over the router's config file.

```yaml
# amqpMessagingNetworkExample indicates whether the example AMQP Messaging Network
# consisting of a single Dispatch Router and Broker should be
# deployed and used. By default an internal Broker is deployed.
# As alternative an external Broker can be configured as well.
amqpMessagingNetworkExample:
  enabled: true
  # dispatchRouter.imageName contains the name (including tag)
  # of the container image to use for the example AQMP Messaging Network Router.
  dispatchRouter:
    imageName: ${dispatch-router.image.name}
    svc:
      annotations: {}
      loadBalancerIP:
  # AMQP Messaging Network Broker configuration.
  broker:
    type: artemis
    artemis:
    # imageName contains the name (including tag) of the container
    # image to use for the example AMQP Messaging Network Broker
      imageName: ${artemis.image.name}
   # Azure Service Bus as alternative to self-hosted Artemis ActiveMQ for event persistence
    servicebus:
      saslUsername:
      saslPassword:
      host:
```

dejanb: @kaizimmerm That indeed looks clean 👍

dejanb:
@kaizimmerm can you rebase this PR? Seems like there's a conflict with master

kaizimmerm:
@dejanb I clean it up. Single commit now as well.

sophokles73:
@kaizimmerm I think this needs to be rebased once more ...

kaizimmerm:
@sophokles73 done. 

Note: I had to do one more change when I merged the new lora adapter in. _adapters.lora.enabled_ helm value did exist but was not taken into account in the templates. I added that.

kaizimmerm:
@sophokles73 ...and again, seems you fixed the lora enabled in the meantime as well.

deploy/src/main/deploy/helm/values.yaml
```diff
@@ -109,6 +126,9 @@ adapters:
   #deviceConnectionSpec:
   #deviceRegistrationSpec:
   #tenantSpec:
+    svc:
```
sophokles73: I believe this belongs to the AMQP adapter config, right? If so, then it needs to go *before* `#amqpMessagingNetworkSpec` ...

jmeter/src/jmeter/http_messaging_throughput_test.jmx
```diff
@@ -105,21 +105,14 @@
...
-          <boolProp name="HTTPSampler.postBodyRaw">true</boolProp>
-          <elementProp name="HTTPsampler.Arguments" elementType="Arguments">
-            <collectionProp name="Arguments.arguments">
-              <elementProp name="" elementType="HTTPArgument">
-                <boolProp name="HTTPArgument.always_encode">false</boolProp>
-                <stringProp name="Argument.value">{&quot;device-id&quot;: &quot;device_${__threadNum}&quot;}</stringProp>
-                <stringProp name="Argument.metadata">=</stringProp>
-              </elementProp>
-            </collectionProp>
+          <elementProp name="HTTPsampler.Arguments" elementType="Arguments" guiclass="HTTPArgumentsPanel" testclass="Arguments" testname="User Defined Variables" enabled="true">
```
sophokles73: this seems to be completely unrelated to adding support for ServiceBus, isn't it? If so, would you mind putting this into a separate PR?

site/documentation/content/deployment/azure.md
```diff
@@ -0,0 +1,207 @@
++++
+title = "Microsoft Azure"
+weight = 475
+aliases = [
+    "/deployment/azure"
+]
++++
+
+This guide describes how Eclipse Hono™ can be deployed on Microsoft Azure. It includes:
```
sophokles73: I do not quite see the necessity for having a separate document for deployment on Azure. In fact, we have put a lot of effort in always using Helm to deploy Hono, regardless of where the Kubernetes cluster is running.

FMPOV it would therefore be preferable to include Azure specific steps to create a Kubernetes cluster either in the existing `create-kubernetes-cluster.md` document or create a `create-kubernetes-cluster-on-azure.md` document and link to it from the existing document. In any case, the instructions for deploying Hono using Helm should go into the existing `helm-based-deployment.md` document. You are welcome to add a section there which deals with the specifics of specifying the loadbalancer IPs and how to use ServiceBus instead of Artemis.
Otherwise we will end up with a separate deployment document for each specific combination of options ...

site/documentation/content/deployment/create-kubernetes-cluster.md
```diff
@@ -19,214 +21,22 @@ Please refer to Kubernetes' [Minikube setup guide](https://kubernetes.io/docs/se
...
+- **memory**: 8192 (MB)
 
 The command to start the VM will look something like this:
+
 ```sh
 minikube start --cpus 2 --memory 8192
 ```
 
-After the Minikube VM has started successfully, the `minikube tunnel` command should be run in order to support Hono's services being deployed using the *LoadBalancer* type. Please refer to the [Minikube Networking docs](https://github.com/kubernetes/minikube/blob/master/docs/networking.md#access-to-loadbalancer-services-using-minikube-tunnel) for details.
+After the Minikube VM has started successfully, the `minikube tunnel` command should be run in order to support Hono's services being deployed using the _LoadBalancer_ type. Please refer to the [Minikube Networking docs](https://github.com/kubernetes/minikube/blob/master/docs/networking.md#access-to-loadbalancer-services-using-minikube-tunnel) for details.
```
sophokles73: I do not see the point of this change. Can you either explain or revert?

site/documentation/content/deployment/create-kubernetes-cluster.md
```diff
@@ -19,214 +21,22 @@ Please refer to Kubernetes' [Minikube setup guide](https://kubernetes.io/docs/se
...
 
 {{% note title="Supported Kubernetes Versions" %}}
-Minikube will use the most recent Kubernetes version that was available when it has been compiled by default. Hono *should* run on any version of Kubernetes starting with 1.13.6. However, it has been tested with several specific versions only so if you experience any issues with running Hono on a more recent version, please try to deploy to 1.13.6 before
+Minikube will use the most recent Kubernetes version that was available when it has been compiled by default. Hono _should_ run on any version of Kubernetes starting with 1.13.6. However, it has been tested with several specific versions only so if you experience any issues with running Hono on a more recent version, please try to deploy to 1.13.6 before
```
sophokles73: I do not see the point of this change. Can you either explain or revert?

jmeter/src/jmeter/http_messaging_throughput_test.jmx
kaizimmerm: Done. https://github.com/eclipse/hono/pull/1522

site/documentation/content/deployment/azure.md
kaizimmerm: @sophokles73 understood. Thank you for the feedback. I will try to find a way and come back to you.

site/documentation/content/deployment/create-kubernetes-cluster.md
kaizimmerm: Yeah, its my editor. Sry about that. I have to disable _pretty print_ :)

site/documentation/content/deployment/azure.md
kaizimmerm: @sophokles73 done, I split the guidance into the two existing pages.

tests/pom.xml
```diff
@@ -476,7 +476,7 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
                       <exec>
                         <arg>/sbin/qdrouterd</arg>
                         <arg>-c</arg>
-                        <arg>/etc/hono/qdrouterd-with-broker.json</arg>
+                        <arg>/etc/hono/qdrouterd-with-artemis.json</arg>
```
sophokles73: this refers to the config file that has been copied into the container from location `src/test/resources/qpid/qdrouter-with-broker.json`. So unless you have changed the name of that file (which I think you haven't), this should remain the way it was ... We do not want to run the integration tests against ServiceBus either, do we?
IMHO this is also the reason for the failing Travis build ...

kaizimmerm: @sophokles73 ahhhh, that makes a lot of sense. Thank you!

deploy/src/main/deploy/helm/templates/dispatch-router/config/qrouterd-with-servicebus.yaml
```diff
@@ -0,0 +1,145 @@
...
+    "host": "0.0.0.0",
+    "port": 5673,
+    "maxSessionFrames": 500,
+    "linkCapacity": 100,
+    "authenticatePeer": true,
+    "saslMechanisms": "EXTERNAL"
+  }],
+
+  ["connector", {
+    "sslProfile": "external",
```
sophokles73: I am not sure if this will work out-of-the-box with ServiceBus using a certificate that has been signed by a public CA, because the *external* SSL profile is configured to use *trusted-certs.pem* as the trust store, which includes only the CA certs used by the Hono example. Have you actually tried this?

deploy/src/main/deploy/helm/templates/dispatch-router/dispatch-router-secret.yaml
```diff
@@ -1,13 +1,14 @@
-{{- if .Values.amqpMessagingNetworkDeployExample -}}
+{{- if .Values.amqpMessagingNetworkExample.enabled -}}
+{{-  $qrouterdConfig := printf "%s%s%s" "hono.qdrouterd-with-" .Values.amqpMessagingNetworkExample.broker.type ".json" -}}
```
sophokles73: what about
```
$qrouterdConfig := printf "hono.qdrouterd-with-%s.json" .Values.amqpMessagingNetworkExample.broker.type
```
?

deploy/src/main/deploy/helm/templates/dispatch-router/config/qrouterd-with-servicebus.yaml
kaizimmerm: I does work, for sure. Let me see if I can figure out why.

deploy/src/main/deploy/helm/templates/dispatch-router/dispatch-router-secret.yaml
kaizimmerm: Definitely better

deploy/src/main/deploy/helm/templates/dispatch-router/config/qrouterd-with-servicebus.yaml
kaizimmerm: I could not figure it out exactly. I guess there are two options, either the router is not trying to TLS client authenticate as SASL is configured as well or SB is plain ignoring it just taking the SASL plain auth.

Anyhow, I created a new profile only defining the TLS version which is sufficient. Makes it clearer that way.

PR #1501, opened 2019-09-18
Added banner, log config, keys and certs for lora adapter

BobClaerhout:


adapters/lora-vertx/src/main/resources/logback-spring.xml
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
```
sophokles73: this should be just 2019 as the file has been created newly

PR #1502, opened 2019-09-18
Add lora adapter to helm deploy

BobClaerhout:


sophokles73:
@BobClaerhout I think this needs to be rebased so that it doesn't contain all the cert changes again ...

BobClaerhout:
@sophokles73, That's one possibility. Or, if we merge the other PR first, this PR will be updated and only have the second commit.

BobClaerhout:
Rebased

deploy/src/main/deploy/lora-adapter.credentials
```diff
@@ -0,0 +1,2 @@
+username=lora-adapter@HONO
+password=lora-secret
```
sophokles73: the *lora-adapter@HONO* user also needs to be added to `deploy/example-permissions.json` in order for other services being able to authenticate the LoRa adapter using the Auth Server.

deploy/src/main/deploy/helm/templates/prometheus/prometheus-config.yaml
```diff
@@ -24,6 +24,9 @@ data:
     {{- end }}
     {{- include "hono.prometheus.scrapeJob" (dict "dot" . "serviceName" "adapter-amqp-vertx") | nindent 4 }}
     {{- include "hono.prometheus.scrapeJob" (dict "dot" . "serviceName" "adapter-http-vertx") | nindent 4 }}
+    {{- if .Values.adapters.lora.enabled }}
+      {{- include "hono.prometheus.scrapeJob" (dict "dot" . "serviceName" "adapter-lora") | nindent 4 }}
```
sophokles73: the service names seems to be `adapter-lora-vertx`, doesn't it?

BobClaerhout: yes, you're right. Changed it

deploy/src/main/deploy/lora-adapter.credentials
BobClaerhout: Added

PR #1504, opened 2019-09-19
Improve Helm chart

sophokles73:
The chart now supports configuration of custom service implementations (AMQP Network and registry).
In particular, the chart supports mounting additional secrets into the protocol adapter containers which can be used to provide custom keys, certs, credentials.

This PR only shows the necessary changes by example of the AMQP adapter. If we agree that this is the way to go, I will make corresponding changes for the other adapters as well.

### DIFF
--- a/deploy/src/main/deploy/helm/values.yaml
+++ b/deploy/src/main/deploy/helm/values.yaml
@@ -5,17 +5,31 @@
 # amqpMessagingNetworkDeployExample indicates whether the example AMQP Messaging Network
 # consisting of a single Dispatch Router and Artemis broker should be
 # deployed and used.
+# If this property is set to false, then the protocol adapters need to be
+# explicitly configured for connecting to an existing AMQP Messaging Network
+# using properties
+# - "adapters.amqpMessagingNetworkSpec"
+# - "adapters.commandAndControlSpec"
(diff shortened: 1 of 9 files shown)


deploy/src/main/deploy/helm/values.yaml
```diff
@@ -5,17 +5,31 @@
...
 # deviceRegistryDeployExample indicates whether the example Device Registry
 # should be deployed and used.
+# If this property is set to false, then the protocol adapters need to be
+# explicitly configured for connecting to the Tenant, Device Registration,
+# Credentials and Device Connection services using properties
+# - "adapters.tenantSpec"
+# - "adapters.deviceRegistrationSpec"
+# - "adapters.credentialsSpec"
+# - "adapters.deviceConnectionSpec"
 deviceRegistryDeployExample: true
```
dejanb: I might be missing it, but I don't see this value being use to actually decide wether to deploy the example registry or not?

sophokles73: We use it to determine if the adapters should be configured to connect to the example registry but we are not using it (yet) to prevent the example registry from being deployed in the first place. I wonder how I could miss that ... thanks for pointing it out :+1:

sophokles73:
already merged

PR #1505, opened 2019-09-19
Increase font size of homepage to look more like the documentation.

b-abel:
The font sizes of the homepage and the documentation part differ greatly, as both use different Hugo themes. To better align them, this PR increases the font size (and line spacing) of the homepage.

### DIFF
--- a/site/homepage/static/css/custom.css
+++ b/site/homepage/static/css/custom.css
@@ -70,11 +70,11 @@ div#landing-vp {
 
 /* Change font size to Bootstrap 4 defaults, which are similar to the values in theme "learn" (used for the documentation). */
 html {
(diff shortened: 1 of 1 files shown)


site/homepage/static/css/custom.css
```diff
@@ -67,3 +67,50 @@ div#landing-vp {
 #navbar .logo {
   height: 46px;
 }
+
+/* Change font size to Bootstrap 4 defaults, which are similar to the values in theme "learn" (used for the documentation). */
+html {
+  font-size: 16px; /* defines the rem */
```
sophokles73: why isn't this defined in *rem*?

b-abel: Changed that.

PR #1516, opened 2019-09-20
[#1506] Skip update of last known gateway for HTTP adapter

### LINKED ISSUE #1506: HTTP-Adapter: Enable gateways to receive commands for a specific device
The scenario here is about an **authenticated gateway** sending an event with a `hono-ttd` parameter (indicating it wants to receive a command).

Following the documentation (chapter `Publish an Event (authenticated Gateway)`), one might think the URL to use here would be `/event/${tenantId}/${deviceId}` (containing the device id, not the gateway id).

**However, the gateway will receive no command in that case** (provided the command is directed at the new `command` address).

**Background:**
The HTTP protocol adapter will create a mapping using the `device connection` API between the device id given in the URL and the gateway id (provided via the authentication data).
The HTTP protocol adapter will also create a command consumer with the "device id" in its link address.

Upon receiving a command, the GatewayMapper logic inside the protocol adapter will map the device id (given in the command address) to the gateway id. The problem is, that there is no command consumer listening on the gateway address.

**Proposed solution:**
Looking at the MQTT and AMQP adapters, we only support authenticated gateways that subscribe for commands using an address that doesn't contain a specific device id. 
In the MQTT case, the topic `command/+/+/req/#` is to be used, for AMQP it is the `command` link address.

Adopting a corresponding behaviour for the HTTP adapter, it would make sense to only support authenticated gateways sending events with the `hono-ttd` parameter using the `/event` URL (without device id). That would cause the gateway to receive a command for _any_ device connected to the gateway. The documentation should be adapted accordingly in that case then.

The **downside** here is, that the northbound application will receive the notification about the command consumer being ready associated with the _gateway id_ - although involvement of a gateway should probably be kept hidden to the application.
(In the MQTT and AMQP adapter cases, the notification event is also associated with the gateway id - but here the overall scenario could be that applications expect long-running command consumers, not waiting for notifications, not like in the HTTP case.)

**_EDIT_**
Solution shall be disabling the update of the last-known-gateway for the HTTP adapter instead - see below.

calohmn:
This fixes #1506.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
@@ -940,6 +940,43 @@ public abstract class AbstractProtocolAdapterBase<T extends ProtocolAdapterPrope
             final String deviceId,
             final Device authenticatedDevice,
             final SpanContext context) {
+        return getRegistrationAssertion(tenantId, deviceId, authenticatedDevice, context, true);
+    }
+
+    /**
+     * Gets an assertion for a device's registration status.
+     * <p>
+     * The returned JSON object may include <em>default</em>
+     * values for properties to set on messages published by the device under
+     * property {@link RegistrationConstants#FIELD_PAYLOAD_DEFAULTS}.
+     * <p>
+     * Note that this method will also update the last gateway associated with the given device if
+     * <em>updateLastGateway</em> is {@code true}.
+     *
+     * @param tenantId The tenant that the device belongs to.
+     * @param deviceId The device to get the assertion for.
+     * @param authenticatedDevice The device that has authenticated to this protocol adapter.
+     *            <p>
+     *            If not {@code null} then the authenticated device is compared to the given tenant and device ID. If
+     *            they differ in the device identifier, then the authenticated device is considered to be a gateway
+     *            acting on behalf of the device.
+     * @param context The currently active OpenTracing span that is used to
+     *                trace the retrieval of the assertion.
+     * @param updateLastGateway If {@code true}, the last gateway associated with the given device will be updated.
+     *                          Setting this value to {@code false} and thereby skipping the update makes sense for
+     *                          protocol adapters where gateways subscribe to commands for a specific device (not
+     *                          for all devices connected to the gateway). This is usually the case for adapters
+     *                          that use short <em>ttd</em> (time-till-disconnect) values, possibly just expecting
+     *                          one command message for a specific device before closing the command consumer again.
+     * @return The assertion.
+     * @throws NullPointerException if any of tenant or device ID are {@code null}.
+     */
+    protected final Future<JsonObject> getRegistrationAssertion(
+            final String tenantId,
+            final String deviceId,
+            final Device authenticatedDevice,
+            final SpanContext context,
+            final boolean updateLastGateway) {
 
         Objects.requireNonNull(tenantId);
         Objects.requireNonNull(deviceId);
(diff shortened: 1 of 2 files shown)


service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -940,6 +940,43 @@ protected void onCommandConnectionEstablished(final HonoConnection commandConnec
...
+     *            <p>
+     *            If not {@code null} then the authenticated device is compared to the given tenant and device ID. If
+     *            they differ in the device identifier, then the authenticated device is considered to be a gateway
+     *            acting on behalf of the device.
+     * @param context The currently active OpenTracing span that is used to
+     *                trace the retrieval of the assertion.
+     * @param updateLastGateway If {@code true}, the last gateway associated with the given device will be updated.
+     *                          Setting this value to {@code false} and thereby skipping the update makes sense for
+     *                          protocol adapters where gateways subscribe to commands for a specific device (not
+     *                          for all devices connected to the gateway). This is usually the case for adapters
```
sophokles73: This is usually the case for adapters that use a stateless protocol like HTTP for communication with devices. In such a case, the adapter will usually delay the response to a device's request for the time indicated by the *ttd* request property in order to prevent buffering of commands in the adapter.

calohmn: > the adapter will usually delay the response to a device's request for the time indicated by the ttd request property

Is that really the case? To me it looks like a received command at the HTTP adapter will directly be sent to the device as part of the HTTP response - without further delay.
And preventing further commands to be received and buffered is done by not giving back any further credit to the messaging network.

In any case, what the "This is usually.." sentence should do IMHO, is to make clear why a adapter would want to only support commands for *specific devices*, instead of *gateways*. In your version this doesn't become so clear to me.

sophokles73: > To me it looks like a received command at the HTTP adapter will directly be sent to the device as part of the HTTP response - without further delay.

The important thing here is that the adapter needs to associate a command for a device with a gateway's HTTP request on behalf of that specific device. The way we have implemented the command forwarding in the HTTP response assumes that a command contained in the HTTP response is targeted at the device that the request was referring to. Thus, there is no way for the adapter to send a command to a device outside of the scope of an HTTP request from (or on behalf of) that device.

Alternatively, the adapter could be modified to support *long polling* of gateways, i.e. a gateway could consecutively POST empty notifications to `/event`, thus indicating that the gateway itself is ready to receive commands. The adapter could then indeed just open a single command receiver for the gateway and forward commands for devices in the HTTP responses to the gateway. However, this would require us to include the target device ID in the response (a header) and also this would probably not work too well because the gateway might send each of its *long polling* requests to a different adapter instance. In any case, the adapter would then also need to update a device's via information on every message ...

So, I do not think that it is about the amount of time to wait but more about the fact that the HTTP adapter (currently) is not able to include commands for arbitrary devices in HTTP responses for a gateway. BTW we have the same problem with the MQTT adapter, I guess, because there we also (currently) cannot include the target device ID in a command message being published to a command receiver topic (which therefore includes the device either explicitly or implicitly).

sophokles73: @calohmn is this still the way to go in context of #1552 ?

calohmn: @sophokles73 No, this PR becomes obsolete with #1552.

PR #1524, opened 2019-09-25
[#267] Equip MQTT Adapter to support setting TTL for events published by devices. 

kaniyan:
The devices can now specify *time-to-live* for _event_ messages it publishes to Hono using the MQTT adapter. This can be done by setting the *hono-ttl* property in the *property-bag* explicitly. This  PR address the issue #267.

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
@@ -14,6 +14,7 @@
 package org.eclipse.hono.adapter.mqtt;
 
 import java.net.HttpURLConnection;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
@@ -1130,7 +1131,8 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends MqttProtoc
                     payload,
                     tenantObject,
                     tokenTracker.result(),
-                    null);
+                    null,
+                    EndpointType.EVENT.equals(endpoint) ? getTimeToLive(ctx.propertyBag()) : null);
 
             addRetainAnnotation(ctx, downstreamMessage, currentSpan);
             customizeDownstreamMessage(downstreamMessage, ctx);
@@ -1405,6 +1407,30 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends MqttProtoc
                 commandContext.release(1);
             }
         });
+    }
+
+    /**
+     * Gets the <em>time-to-live</em> duration from the given property bag object.
+     *
+     * @param propertyBag The property bag object.
+     * @return The <em>time-to-live</em> duration or {@code null} if
+     * <ul>
+     *     <li>the given property bag object is {@code null.}</li>
+     *     <li>no property with id {@link org.eclipse.hono.util.Constants#HEADER_TIME_TO_LIVE} exists.</li>
+     *     <li>the contained value cannot be parsed as a Long.</li>
+     *     <li>the contained value is negative.</li>
+     * </ul>
+     */
+    protected final Duration getTimeToLive(final PropertyBag propertyBag) {
+        try {
+            return Optional.ofNullable(propertyBag)
+                    .map(propBag -> propBag.getProperty(Constants.HEADER_TIME_TO_LIVE))
+                    .map(Long::parseLong)
+                    .map(ttl -> ttl < 0 ? null : Duration.ofSeconds(ttl))
+                    .orElse(null);
+        } catch (final NumberFormatException e) {
+            return null;
+        }
     }
 
     private void afterCommandPublished(
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttContext.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttContext.java
@@ -14,6 +14,7 @@
 package org.eclipse.hono.adapter.mqtt;
 
 import java.util.Objects;
+import java.util.Optional;
 
 import org.eclipse.hono.auth.Device;
 import org.eclipse.hono.service.metric.MetricsTags;
@@ -39,6 +40,7 @@ public final class MqttContext extends MapBasedExecutionContext {
     private String contentType;
     private Sample timer;
     private MetricsTags.EndpointType endpoint;
+    private PropertyBag propertyBag;
 
     private MqttContext() {
     }
@@ -83,7 +85,11 @@ public final class MqttContext extends MapBasedExecutionContext {
         result.authenticatedDevice = authenticatedDevice;
         if (publishedMessage.topicName() != null) {
             try {
-                result.topic = ResourceIdentifier.fromString(publishedMessage.topicName());
+                Optional.ofNullable(PropertyBag.fromTopic(publishedMessage.topicName()))
+                        .ifPresentOrElse(propertyBag -> {
+                            result.topic = ResourceIdentifier.fromString(propertyBag.topicWithoutPropertyBag());
+                            result.propertyBag = propertyBag;
+                        }, () -> result.topic = ResourceIdentifier.fromString(publishedMessage.topicName()));
                 result.endpoint = MetricsTags.EndpointType.fromString(result.topic.getEndpoint());
             } catch (final IllegalArgumentException e) {
                 // malformed topic
@@ -180,6 +186,17 @@ public final class MqttContext extends MapBasedExecutionContext {
         } else {
             return null;
         }
+    }
+
+    /**
+     * Gets the property bag object from the <em>property-bag</em>
+     * set in the message's topic.
+     * 
+     * @return The property bag object or {@code null} if
+     *         there is no property bag set in the topic.
+     */
+    public PropertyBag propertyBag(){
+        return this.propertyBag;
     }
 
     /**
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/PropertyBag.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/PropertyBag.java
@@ -1,0 +1,79 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.hono.adapter.mqtt;
+
+import io.netty.handler.codec.http.QueryStringDecoder;
+
+import java.util.List;
+import java.util.Map;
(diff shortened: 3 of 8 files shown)


kaniyan:
> Having the property bag related code as part of the `MqttContext` doesn't feel right to me. I'd rather have it encapsulated in a separate class `PropertyBag` which hides all the internals and provides methods for retrieving the property values. WDYT?

I also thought of a separate class. But then I decided against it as the `PropertyBag` is just being used for setting the `hono-ttl`property and not many uses cases.  Also `MQTTContext` provides relevant information for processing a MQTT message and hence I placed it there. I think its a good idea to move all those methods to a separate class which adds to more separation of concerns.

kaniyan:
@sophokles73 I have created a separate class `PropertyBag` containing relevant functions and pushed the changes. Would you mind taking a look?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/PropertyBag.java
```diff
@@ -0,0 +1,99 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
```
sophokles73: this is just 2019

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/PropertyBag.java
```diff
@@ -0,0 +1,99 @@
...
+import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * A collection of methods for processing <em>property-bag</em> set at the end of a topic.
+ *
+ */
+public final class PropertyBag {
```
sophokles73: I had expected to find something like
```
static PropertyBag fromTopic(final String topic) {
}
```
which would create a `PropertyBag` instance encapsulating all the properties and providing methods for retrieving property values ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/PropertyBag.java
kaniyan: Done

kaniyan:
@sophokles73 I have pushed the requested changes. Would you mind taking a look?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/PropertyBag.java
```diff
@@ -0,0 +1,111 @@
...
+
+    private final String topic;
+
+    private PropertyBag(final String topic) {
+        this.topic = topic;
```
sophokles73: `PropertyBag` seems to be an immutable. Wouldn't it therefore make sense to parse the properties contained in the topic once here, instead of parsing the properties every time you want to retrieve a value?

kaniyan: @sophokles73 I have pushed the corresponding changes. Please have a look.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttContext.java
```diff
@@ -83,7 +87,9 @@ public static MqttContext fromPublishPacket(
         result.authenticatedDevice = authenticatedDevice;
         if (publishedMessage.topicName() != null) {
             try {
-                result.topic = ResourceIdentifier.fromString(publishedMessage.topicName());
+                final PropertyBag propertyBag = PropertyBag.fromTopic(publishedMessage.topicName());
+                result.timeToLive = extractTimeToLive(propertyBag);
```
sophokles73: IMHO the context should provide access to the PropertyBag itself but it should have no knowledge about the semantics of particular properties contained in it, don't you think?

kaniyan: @sophokles73 IMHO `PropertyBag` class should not know the semantics of the properties. It should be straight forward knowing how to extract the properties and retrieve a particular property without any context.  With that in my mind, the next suitable place for the `time-to-live` is `MQTTContext` as it already holds information required to process a message such as topic, message, endpoint etc.

sophokles73: No, PropertyBag should not know about *time-to-live* but MqttContext shouldn't either. My point is that the MQTT adapter should do something like
```
context.getPropertyBag().getProperty("ttl");
```
Otherwise, I would wonder why the MqttContext knows about ttl but not about the other properties contained in the PropertyBag? How does the MQTT adapter get access to these?

kaniyan: I also tempted to the add the `propertyBag` as an attribute in `MQTTContext`.  Right now we use the `property-bag` only for `time-to-live` and there are no other use cases. Hence I exposed only `time-to-live` in `MQTTContext` keeping in mind that in future if any use case pops up, it would be still easy to expose it as it is already available in `MQTTContext`. It is also fine for me to provide already access to the `propertyBag` and do the computation of `time-to-live` in the _MQTTAdapter_ base class and also making it reusable by the custom MQTT adapters.

sophokles73: that would be great :+1:

kaniyan: @sophokles73 I have pushed the changes. Please have a look.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/PropertyBag.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Retrieves the <em>property-bag</em> from the given topic and 
+     * extracts the key/value pairs into a map.
+     * 
+     * @return The properties from the <em>property-bag</em> or {@code null}
+     *         if no <em>property-bag</em> is present or the topic is {@code null}.
+     */
+    private Map<String, List<String>> getProperties() {
+        final int index = topic.lastIndexOf("?");
```
sophokles73: IMHO this is not precise enough. A topic name may as well contain segments that contain `?`.
I think that we need to check for the last segment of the topic name to actually begin with the `?` character, right?

kaniyan: @sophokles73 Ah I see the point now regarding additional topic segments. A topic such as `event/tenant/device/seg1/seg2?param=1/seg3/seg4` which has no `property-bag ` will be considered as the one with the `property-bag`. I think `topic.lastIndexOf("/?")` instead of `topic.lastIndexOf("?")` should do the trick. The last index of the separator (/) will be used to identify the last segment and the marker (?) to identify the beginning of the `property-bag`. WDYT?

sophokles73: I think so, yes. Consequently, I think that the *topic without the property-bag* should not include the trailing `/` ...

kaniyan: :+1:

kaniyan: @sophokles73 I have replaced it with `topic.lastIndexOf("/?")` with an additional test. Would you mind taking a look.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1407,6 +1409,30 @@ protected final void onCommandReceived(
...
+     * @param propertyBag The property bag object.
+     * @return The <em>time-to-live</em> duration or {@code null} if
+     * <ul>
+     *     <li>the given property bag object is {@code null.}</li>
+     *     <li>no property with id {@link org.eclipse.hono.util.Constants#HEADER_TIME_TO_LIVE} exists.</li>
+     *     <li>the contained value cannot be parsed as a Long.</li>
+     *     <li>the contained value is negative.</li>
+     * </ul>
+     */
+    protected Duration getTimeToLive(final PropertyBag propertyBag) {
```
sophokles73: final ?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttContext.java
```diff
@@ -182,6 +185,17 @@ public String tenant() {
...
     }
 
+    /**
+     * Gets the property bag object from the <em>property-bag</em>
+     * set in the message's topic.
+     * 
+     * @return The property bag object or {@code null} if
+     *         there is no property bag set in the topic.
+     */
+    public PropertyBag propertyBag(){
```
sophokles73: final?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/PropertyBag.java
```diff
@@ -0,0 +1,99 @@
...
+        return topicWithoutPropertyBag;
+    }
+
+    /**
+     * Trims the <em>property-bag</em> from the topic and returns the rest.
+     *
+     * @return The topic without the <em>property-bag</em>.
+     */
+    private String extractTopicWithoutPropertyBag() {
+        final int index = topic.lastIndexOf("/?");
```
sophokles73: IMHO we should do this and *getProperties* in the same method/step in order to prevent searching for the index twice ...

kaniyan: @sophokles73 I have pushed the changes, pls have a look again.

site/documentation/content/user-guide/mqtt-adapter.md
```diff
@@ -127,6 +127,12 @@ This requires that
 
 The protocol adapter checks the configured [message limit] ({{< relref "/concepts/resource-limits.md" >}}) before accepting any event messages. If the message limit is exceeded or the incoming event message cannot be processed, the connection to the client is closed.
 
+The devices can optionally indicate a *time-to-live* duration for event messages by setting the *hono-ttl* property explicitly in the `property-bag`. The `property-bag` is an optional collection of properties intended for the receiver of the message. A property bag is only allowed at the very end of a topic. It always starts with a `?` character, followed by pairs of URL encoded property names and values that are separated by `&`. The following example shows a property bag that contains two properties *seqNo* and *importance*:
```
sophokles73: ... it always starts with `/?`, followed by ...

site/documentation/content/user-guide/mqtt-adapter.md
```diff
@@ -127,6 +127,12 @@ This requires that
...
+
+    /topic/name/?seqNo=10034&importance="high"
```
sophokles73: do we actually support quotes? If so, what meaning do they have?
My understanding is that every value in the property bag is considered a string, isn't it?

kaniyan: Every value is considered as a string and the above quotes doesn't have a special purpose. I will remove it.

PR #1526, opened 2019-09-26
Adapt Azure guidance to helm changes.

kaizimmerm:
Updated Azure guide after https://github.com/eclipse/hono/commit/114c5a1c8782bf2da007973dc07cad58ead71934


kaizimmerm:
@sophokles73 

I thought about this several times actually. However, considering that all these values are dynamic by means of coming from Azure I figured that the alternatives do not look any more elegant.

I might be wrong here though. If you have a good idea let me know.

sophokles73:
> However, considering that all these values are dynamic by means of coming from Azure I figured that the alternatives do not look any more elegant.

That's true. I didn't think about that ...

PR #1528, opened 2019-09-30
Prepare configuration of customers/clients/adopters section on homepage.

b-abel:
Add the configuration of the "carousel" component in the "adopters" section. Reduced maximum images per page to 4.


site/homepage/static/js/front.js
```diff
@@ -0,0 +1,390 @@
...
+
+  if (windowWidth !== newWindowWidth) {
+    setTimeout(function () {
+      $(this).alignElementsSameHeight()
+      fullScreenContainer()
+      pictureZoom()
+    }, 205)
+    windowWidth = newWindowWidth
+  }
+})
```
sophokles73: where does all this code come from?

b-abel: The file is from the theme (Hugo theme "Universal"). It is included to change the configuration of the carousel "Customers". If you want to see what has changed, you unfortunately need to compare the file with the one in the theme. 

The changed part looked before  like the following: 

~~~js
    $('.customers').owlCarousel({
      items: 6,
      itemsDesktopSmall: [990, 4],
      itemsTablet: [768, 2],
      itemsMobile: [480, 1]
    })
~~~

and is now:

~~~js
    $('.customers').owlCarousel({
      items: 4,
      itemsDesktopSmall: [990, 3],
      itemsTablet: [768, 2],
      itemsMobile: [480, 1],
      slideSpeed: 300,
      paginationSpeed: 400,
      autoPlay: true,
      stopOnHover: true,
      singleItem: false
    })
~~~

It configures the [Owl Carousel plugin](http://www.landmarkmlp.com/js-plugin/owl.carousel/index.html#customizing) used by the theme.

b-abel:
This PR does not yet enable the section. To display the section the property `params.clients.enable` must be set to `true`. 
At the moment I wouldn't enable the feature because there are two bugs on the side of the Eclipse Foundation which should be resolved first: https://github.com/EclipseFdn/iot.eclipse.org/issues/229 and https://github.com/EclipseFdn/iot.eclipse.org/issues/228.

PR #1532, opened 2019-10-01
[#1138] Add CORS headers to device registry management API

dejanb:


service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -360,4 +364,32 @@ protected void extractIfMatchVersionParam(final RoutingContext ctx) {
...
         ctx.next();
     }
+
+    /**
+     * Creates default CORS handler that allows 'POST', 'GET', 'PUT' and 'DELETE' methods for the specified origin.
+     *
+     * @param allowedOrigin The allowed origin pattern.
+     * @return The handler.
+     */
+    protected CorsHandler createDefaultCorsHandler(final String allowedOrigin) {
```
sophokles73: final?

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -360,4 +364,32 @@ protected void extractIfMatchVersionParam(final RoutingContext ctx) {
...
+    }
+
+    /**
+     * Creates CORS Handler that allows HTTP methods for the specified origin.
+     *
+     * @param allowedOrigin The allowed origin pattern.
+     * @param methods Set of allowed HTTP methods
+     * @return The handler.
+     */
+    protected CorsHandler createCorsHandler(final String allowedOrigin, final Set<HttpMethod> methods) {
```
sophokles73: final?

core/src/main/java/org/eclipse/hono/config/ServiceConfigProperties.java
```diff
@@ -163,4 +165,33 @@ public final void setReceiverLinkCredit(final int receiverLinkCredit) {
...
+     * Gets the allowed origin pattern for CORS handler.
+     * <p>
+     * The allowed origin pattern for CORS is returned to clients via the <em>Access-Control-Allow-Origin</em> header.
+     * It can be used by Web Applications to make sure that requests go only to trusted backend entities.
+     * <p>
+     * The default value is '*'.
+     *
+     * @return The allowed origin pattern for CORS handler.
+     */
+    public String getCorsAllowedOrigin() {
```
sophokles73: final?

core/src/main/java/org/eclipse/hono/config/ServiceConfigProperties.java
```diff
@@ -163,4 +165,33 @@ public final void setReceiverLinkCredit(final int receiverLinkCredit) {
...
+     * <p>
+     * The allowed origin pattern for CORS is returned to clients via the <em>Access-Control-Allow-Origin</em> header.
+     * It can be used by Web Applications to make sure that requests go only to trusted backend entities.
+     * <p>
+     * The default value is '*'.
+     *
+     * @param corsAllowedOrigin The allowed origin pattern for CORS handler.
+     * @throws NullPointerException if the allowed origin pattern is {@code null}.
+     */
+    public void setCorsAllowedOrigin(final String corsAllowedOrigin) {
```
sophokles73: final?

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -360,4 +364,32 @@ protected void extractIfMatchVersionParam(final RoutingContext ctx) {
...
+        return CorsHandler.create(allowedOrigin)
+                .allowedMethods(methods)
+                .exposedHeader(HttpHeaders.CONTENT_TYPE.toString());
```
sophokles73: what about the *Authorization* header used for authentication of clients?

dejanb: Good catch. Added.

service-base/src/main/java/org/eclipse/hono/service/management/credentials/CredentialsManagementHttpEndpoint.java
```diff
@@ -71,6 +73,10 @@ public void addRoutes(final Router router) {
         final String pathWithTenantAndDeviceId = String.format("/%s/:%s/:%s",
                 getName(), PARAM_TENANT_ID, PARAM_DEVICE_ID);
 
+
+        // Add CORS handler
+        router.route(pathWithTenantAndDeviceId).handler(createCorsHandler(config.getCorsAllowedOrigin(), EnumSet.of(HttpMethod.PUT)));
```
ctron: Shouldn't this be PUT and GET?

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -360,4 +364,33 @@ protected void extractIfMatchVersionParam(final RoutingContext ctx) {
...
+     *
+     * @param allowedOrigin The allowed origin pattern.
+     * @param methods Set of allowed HTTP methods
+     * @return The handler.
+     */
+    protected final CorsHandler createCorsHandler(final String allowedOrigin, final Set<HttpMethod> methods) {
+        return CorsHandler.create(allowedOrigin)
+                .allowedMethods(methods)
+                .allowedHeader(HttpHeaders.CONTENT_TYPE.toString())
+                .allowedHeader(HttpHeaders.AUTHORIZATION.toString());
```
sophokles73: we will also need to expose the *ETag* header and allow the *If-Match* header ...

service-base/src/main/java/org/eclipse/hono/service/management/credentials/CredentialsManagementHttpEndpoint.java
sophokles73: I think so too. The GET requests supported by the Management API are never *simple* because they always include the *ETag* header and will usually have a content type that is not among the *simple* ones ...

dejanb:
Pushed updated commit with the following changes:

* Added If-Match and ETag headers
* Added GET method where it was missing
* All existing device registry API tests are now sending and checking cors headers
* Added preflight check tests

@sophokles73 @ctron can you take a new look into it.

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -360,4 +364,35 @@ protected void extractIfMatchVersionParam(final RoutingContext ctx) {
...
+     * @param methods Set of allowed HTTP methods
+     * @return The handler.
+     */
+    protected final CorsHandler createCorsHandler(final String allowedOrigin, final Set<HttpMethod> methods) {
+        return CorsHandler.create(allowedOrigin)
+                .allowedMethods(methods)
+                .allowedHeader(HttpHeaders.CONTENT_TYPE.toString())
+                .allowedHeader(HttpHeaders.AUTHORIZATION.toString())
+                .allowedHeader(HttpHeaders.IF_MATCH.toString())
+                .allowedHeader(HttpHeaders.ETAG.toString());
```
sophokles73: IMHO this needs to be `.exposeHeader()` because we want the client to be able to read this header from the response ...

tests/src/test/java/org/eclipse/hono/tests/registry/CorsIT.java
```diff
@@ -0,0 +1,224 @@
...
+import static org.eclipse.hono.tests.DeviceRegistryHttpClient.TEMPLATE_URI_CREDENTIALS_BY_DEVICE;
+import static org.eclipse.hono.tests.DeviceRegistryHttpClient.TEMPLATE_URI_REGISTRATION_INSTANCE;
+import static org.eclipse.hono.tests.DeviceRegistryHttpClient.TEMPLATE_URI_TENANT_INSTANCE;
+import static org.eclipse.hono.util.Constants.DEFAULT_TENANT;
+
+import org.eclipse.hono.tests.CrudHttpClient;
+import org.eclipse.hono.tests.IntegrationTestSupport;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
```
sophokles73: IMHO we should use JUnit5 for new tests ...

tests/src/test/java/org/eclipse/hono/tests/registry/CorsIT.java
```diff
@@ -0,0 +1,224 @@
...
+ *******************************************************************************/
+package org.eclipse.hono.tests.registry;
+
+import io.vertx.core.MultiMap;
+import io.vertx.core.Vertx;
+import io.vertx.core.http.HttpHeaders;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.ext.unit.TestContext;
+import io.vertx.ext.unit.junit.VertxUnitRunner;
+import static org.eclipse.hono.tests.DeviceRegistryHttpClient.TEMPLATE_URI_CREDENTIALS_BY_DEVICE;
```
sophokles73: the imports are out of order and
IMHO we should not statically import constants but only import the defining class and the then refer to the constants in a qualified way. Otherwise, it is very hard to see where the constant comes from when reading the code ...

tests/src/test/java/org/eclipse/hono/tests/registry/CorsIT.java
```diff
@@ -0,0 +1,224 @@
...
+
+
+    private static void assertAccessControlHeaders(final TestContext ctx, final MultiMap headers, final HttpMethod expectedAllowedMethod) {
+
+        ctx.assertTrue(headers.get(HttpHeaders.ACCESS_CONTROL_ALLOW_METHODS).contains(expectedAllowedMethod.name()));
+        ctx.assertEquals("*", headers.get(HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN));
+        ctx.assertTrue(headers.get(HttpHeaders.ACCESS_CONTROL_ALLOW_HEADERS).contains(HttpHeaders.AUTHORIZATION));
+        ctx.assertTrue(headers.get(HttpHeaders.ACCESS_CONTROL_ALLOW_HEADERS).contains(HttpHeaders.CONTENT_TYPE));
+        ctx.assertTrue(headers.get(HttpHeaders.ACCESS_CONTROL_ALLOW_HEADERS).contains(HttpHeaders.IF_MATCH));
+        ctx.assertTrue(headers.get(HttpHeaders.ACCESS_CONTROL_ALLOW_HEADERS).contains(HttpHeaders.ETAG));
```
sophokles73: this is not supposed to show up in the allowed headers but in the exposed headers instead

dejanb: Reading the spec, it says that server returns allow headers as a part of the preflight check. So this seems OK to me. Exposed headers are (as far as I can see) returned as the part of the regular calls. So this looks OK to me. The question is whether we should add these headers to the list of exposed headers. Looking at the HTTP adapter, it does the same thing as here.

tests/src/test/java/org/eclipse/hono/tests/registry/CorsIT.java
dejanb: Changed it. BTW. this is not done consistently throughout codebase, so maybe we should add checkstyle rule if that's what we want to do?

tests/src/test/java/org/eclipse/hono/tests/registry/CorsIT.java
sophokles73: I agree, in the preflight check, the server should return the list of headers in the *allowed headers* that the client may include in its request, right? Since the *ETag* header is only included in responses from the server, it shouldn't be included here FMPOV. The pre-flight request shouldn't include it in its *allowed headers* either.
This, of course, needs to be changed in conjunction with canging the definition of the CorsHandler in the AbstractHttpEndpoint as noted in an earlier comment: *ETag* should be added as an *exposed header*.

tests/src/test/java/org/eclipse/hono/tests/registry/CorsIT.java
sophokles73: This is at least what I would like to do. Not sure if there is a corresponding Checkstyle check, though ...

tests/src/test/java/org/eclipse/hono/tests/registry/CorsIT.java
dejanb: OK, now that made sense :) Fixed!

PR #1536, opened 2019-10-02
Enable Integration Tests to use already running docker containers

calohmn:
This adds the possibility to run Integration Tests on Docker containers left running from a previous test run.

With this, running the Integration Tests multiple times while developing new tests will be much faster, since docker containers don't need to be built, started and stopped each time.

Note that there is a new dependenc added here:
`properties-maven-plugin`

A possibility to set the logging level in the Docker containers has been added as well.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -410,6 +410,11 @@
             </signature>
           </configuration>
         </plugin>
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>properties-maven-plugin</artifactId>
+          <version>1.0.0</version>
+        </plugin>
       </plugins>
     </pluginManagement>
 
--- a/tests/pom.xml
+++ b/tests/pom.xml
@@ -243,6 +243,7 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch 
                       <port>+auth.ip:auth.amqp.port:5672</port>
                       <port>+auth.ip:auth.health.port:${vertx.health.port}</port>
                     </ports>
+                    <portPropertyFile>${project.build.directory}/docker/auth.port.properties</portPropertyFile>
                     <network>
                       <mode>custom</mode>
                       <name>hono</name>
@@ -319,6 +320,7 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch 
                       <port>+deviceregistry.ip:deviceregistry.http.port:8080</port>
                       <port>+deviceregistry.ip:deviceregistry.health.port:${vertx.health.port}</port>
                     </ports>
+                    <portPropertyFile>${project.build.directory}/docker/deviceregistry.port.properties</portPropertyFile>
                     <network>
                       <mode>custom</mode>
                       <name>hono</name>
@@ -389,6 +391,7 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch 
                     <ports>
                       <port>+broker.ip:broker.amqp.port:5671</port>
                     </ports>
+                    <portPropertyFile>${project.build.directory}/docker/broker.port.properties</portPropertyFile>
                     <network>
                       <mode>custom</mode>
                       <name>hono</name>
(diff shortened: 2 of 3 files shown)


tests/pom.xml
```diff
@@ -981,5 +992,55 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
...
+                </goals>
+                <configuration>
+                  <files>
+                    <file>${project.build.directory}/docker/auth.port.properties</file>
+                    <file>${project.build.directory}/docker/deviceregistry.port.properties</file>
+                    <file>${project.build.directory}/docker/broker.port.properties</file>
+                    <file>${project.build.directory}/docker/qpid.port.properties</file>
+                    <file>${project.build.directory}/docker/http.port.properties</file>
+                    <file>${project.build.directory}/docker/mqtt.port.properties</file>
+                    <file>${project.build.directory}/docker/adapter.amqp.port.properties</file>
```
sophokles73: why is this file name prefixed with `adapter.` while the others aren't?

tests/readme.md
```diff
@@ -26,3 +26,27 @@ This starts the following Docker containers and runs the test cases against them
...
+
+### Running the Tests without starting/stopping the containers
+
+When running the tests with the `docker.keepRunning` property, the Docker containers will not be stopped and removed once the tests are complete:
+
+    $ mvn verify -Prun-tests -Ddocker.keepRunning
+
+Subsequent test runs can use the running containers and will thereby finish much faster by adding the `useRunningContainers` profile to the maven command:
+
+    $ mvn verify -Prun-tests,useRunningContainers
```
sophokles73: I assume I can also combine this with `-Dit.test=`, right?

pom.xml
```diff
@@ -409,6 +409,11 @@
             </signature>
           </configuration>
         </plugin>
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>properties-maven-plugin</artifactId>
```
sophokles73: we'll need a works-with CQ for this dependency ...

tests/pom.xml
calohmn: That is in analogy to the naming of the maven properties (`mqtt.port`, `http.port, `adapter.amqp.port`).
I can have a look to make this more consistent overall.

tests/readme.md
calohmn: Yes.

tests/pom.xml
sophokles73: Does this mean that the file name represents the *base* name of the properties it contains? If so, then that is ok for me. I just wanted to understand ...

calohmn: > Does this mean that the file name represents the *base* name of the properties it contains? 

Yes, that was the idea.

sophokles73: Taking a quick look at the maven-properties-plugin docs it doesn't look like this convention is required/enforced, is it? If not then I would propose that we use the same pattern for the endpoint property files across all adapters/components. WDYT?

sophokles73:
@calohmn have you created a CQ for the newly introduced plugin?

pom.xml
sophokles73: @calohmn  did you create the CQ for this already?

tests/pom.xml
```diff
@@ -56,6 +56,8 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
     <http.java.options></http.java.options>
     <mqtt.java.options></mqtt.java.options>
     <deviceregistry.image>hono-service-device-registry</deviceregistry.image>
+    <!-- logging profile: either prod, dev or trace -->
+    <logging.profile>prod</logging.profile>
```
sophokles73: this enhancement doesn't seem related to the original scope of this PR. Would you mind putting it into a separate commit (no need for a PR FMPOV)?

pom.xml
calohmn: Yes, I've created it.

tests/pom.xml
calohmn: I've adapted the properties file names for the adapters now, having them all start with the "adapter." prefix. (The conformity to the maven property names is not that important here, I think.)

tests/pom.xml
calohmn: Yes, done.

calohmn:
Doing another test run here, there is now an error stopping the containers when using `docker:stop` in the maven command (`Failed to execute goal io.fabric8:docker-maven-plugin:0.32.0:stop (default-cli) on project hono-tests: At least one exception thrown during container removal.`). I'll have a closer look...

calohmn:
(rebased)

calohmn:
Fixed the issue with stopping and removing the containers.
This should now be triggered by running `mvn verify -PstopContainers`.

PR #1538, opened 2019-10-02
Improve log output of parameterized tests

calohmn:
Improve log output of parameterized tests; fix typos.

Before, the log output for parameterized tests was just:
````
o.e.h.t.mqtt.CommandAndControlMqttIT - running [6] gateway device: true, southbound endpoint: control, northbound endpoint: command, topic filter: control/+/+/req/#
````
With this commit, it will be:
````
o.e.h.t.mqtt.CommandAndControlMqttIT - running testSendCommandFailsForMalformedMessage(MqttCommandEndpointConfiguration, VertxTestContext) [6]; parameters: gateway device: true, southbound endpoint: control, northbound endpoint: command, topic filter: control/+/+/req/#
````

### DIFF
--- a/tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpUploadTestBase.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpUploadTestBase.java
@@ -199,7 +199,7 @@ public abstract class AmqpUploadTestBase extends AmqpAdapterTestBase {
      * @param ctx The Vert.x test context.
      * @throws InterruptedException if test is interrupted while running.
      */
-    @ParameterizedTest
+    @ParameterizedTest(name = IntegrationTestSupport.PARAMETERIZED_TEST_NAME_PATTERN)
     @MethodSource("senderQoSTypes")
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpUploadTestBase.java [hunk 1]
```diff
@@ -13,6 +13,7 @@
 package org.eclipse.hono.tests.amqp;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.eclipse.hono.tests.IntegrationTestSupport.PARAMETERIZED_TEST_NAME_PATTERN;
```
sophokles73: Can we please not statically import constants? This makes it very hard to know where it comes from when reading code that makes use of them ...

calohmn: I've changed that. I agree with Dejan (other PR comment) that we should create a Checkstyle rule for that then.

Issue #1521, opened 2019-09-24
Topic structure for subcribing to commands seems wrong


sophokles73:
The MQTT adapter requires authenticated devices to subcribe to topic
```
command/+/+/req/#
```
in order to receive commands.

The `+` wildcard characters used at the tenant and device ID positions might be misinterpreted to indicate that the device will receive commands for all devices of all tenants.
The name of the topic that the MQTT adapter publishes command messages to (`command///req/${req-id}/${command}` does not include any tenant nor device ID, indicating that the device's tenant and device ID are to be assumed.

FMPOV the topic that authenticated devices must use for subscribing to commands should be changed to
```
command///req/#
```
This would use the same logic as in the topic name for publishing commands, i.e. omitting the tenant and device ID implies that the tenant and device ID of the authenticated device is to be assumed.

For backwards compatibility we should declare the current topic name as deprecated but still accept it for some time.

calohmn:
Makes sense to me.

PR #1539, opened 2019-10-02
Fix topic names for subscribing to commands.

sophokles73:
addresses #1521


Alfusainey:
I think the command & control `concepts` docs is also affected by this change

PR #1540, opened 2019-10-04
Fix static imports.

sophokles73:
Adds a checkstyle rule which prevents static import of members.
Exceptions are added to allow the usual static import of assertions etc in test classes.

### DIFF
--- a/adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
+++ b/adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
@@ -12,13 +12,13 @@
  *******************************************************************************/
 package org.eclipse.hono.adapter.amqp.impl;
 
-import static org.hamcrest.core.IsEqual.equalTo;
 import static org.junit.Assert.assertThat;
 import static org.mockito.ArgumentMatchers.any;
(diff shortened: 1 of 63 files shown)


ctron:
> Adds a checkstyle rule which prevents static import of members.
> Exceptions are added to allow the usual static import of assertions etc in test classes.

Is this really an improvement? If we allow it for asserts, then why not in the other locations as well?

sophokles73:
> Is this really an improvement? 

FMPOV it is since it makes it easier to see where a constant/member has been defined when reading code.

> If we allow it for asserts, then why not in the other locations as well?

In test cases most (if not all) of the static imports are *well known*, i.e. everybody knows that *assertThat* has been imported from an assertion framework. In contrast to that, I am not able to recall where all the constants and helper functions that we have defined all over Hono's modules actually come from. So, for me it is very helpful to see the (defining) class name before the member name.

ctron:
I do understand that. However, to me the information where this comes from is only a "mouseover" or <kbd>F3</kbd> away.

On the other side, not having the defining class in the code makes it easier to read IMHO. As, for example, I am more interested in the information that the result code is `HTTP_OK`, and I don't really care where this was defined.

So if you all think this is right style, I am not opposing it. If you think it is, then go ahead and merge it.

kaniyan:
I have the same opinion as of @ctron regarding the static imports. For me the code with static imports are much more readable and looks more elegant compared to the other. Also for me it is a matter of personal taste and I can live with the other option.

PR #1542, opened 2019-10-07
[#1541] Allow null link address for anonymous relay in createSender()

### LINKED ISSUE #1541: Enable HonoConnection to create a sender link on the anonymous relay
Currently, `HonoConnection.createSender()` doesn't allow the link target address to be `null`.

That is however a valid use case since a `null` address is used to establish a sender link to the anonymous relay (see [here](http://docs.oasis-open.org/amqp/anonterm/v1.0/anonterm-v1.0.html) or the [ProtonConnection](https://github.com/vert-x3/vertx-proton/blob/c02a1a4bce17daf8626c9eb11afd30857ffdbf02/src/main/java/io/vertx/proton/ProtonConnection.java#L77) class).

And that is especially useful for creating a telemetry/event sender link to the Hono AMQP adapter, which requires such a link on the anonymous relay.

calohmn:
This fixes #1541.

Note that using an empty string as target address is still supported by means of a corresponding check in `isLinkEstablished`.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
+++ b/core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
@@ -144,10 +144,7 @@ public final class HonoProtonHelper {
      */
     public static boolean isLinkEstablished(final ProtonLink<?> link) {
         if (link instanceof ProtonSender) {
-            // check for non-null remote target address - or if it is null, for an anonymous sender link (with empty local target address)
-            return link.getRemoteTarget() != null
-                    && (link.getRemoteTarget().getAddress() != null
-                            || (link.getTarget() != null && link.getTarget().getAddress() != null && link.getTarget().getAddress().isEmpty()));
+            return link.getRemoteTarget() != null;
         } else if (link instanceof ProtonReceiver) {
             return link.getRemoteSource() != null && link.getRemoteSource().getAddress() != null;
         } else {
--- a/client/src/main/java/org/eclipse/hono/client/HonoConnection.java
+++ b/client/src/main/java/org/eclipse/hono/client/HonoConnection.java
@@ -245,13 +245,15 @@ public interface HonoConnection extends ConnectionLifecycle<HonoConnection> {
 
     /**
      * Creates a sender link.
-     * 
-     * @param targetAddress The target address of the link.
+     *
+     * @param targetAddress The target address of the link. If the address is {@code null}, the
+     *                      sender link will be established to the 'anonymous relay' and each
+     *                      message must specify its destination address.
      * @param qos The quality of service to use for the link.
      * @param remoteCloseHook The handler to invoke when the link is closed by the peer (may be {@code null}).
      * @return A future for the created link. The future will be completed once the link is open.
      *         The future will fail with a {@link ServiceInvocationException} if the link cannot be opened.
-     * @throws NullPointerException if any of the arguments other than close hook is {@code null}.
+     * @throws NullPointerException if qos is {@code null}.
      */
     Future<ProtonSender> createSender(
             String targetAddress,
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
```diff
@@ -144,10 +144,10 @@ public static void setDefaultCloseHandler(final ProtonSession session) {
      */
     public static boolean isLinkEstablished(final ProtonLink<?> link) {
         if (link instanceof ProtonSender) {
-            // check for non-null remote target address - or if it is null, for an anonymous sender link (with empty local target address)
+            // check for non-null remote target address - or if it is null, for an anonymous sender link (with null local target address) or an empty target address
             return link.getRemoteTarget() != null
                     && (link.getRemoteTarget().getAddress() != null
-                            || (link.getTarget() != null && link.getTarget().getAddress() != null && link.getTarget().getAddress().isEmpty()));
+                            || (link.getTarget() != null && Strings.isNullOrEmpty(link.getTarget().getAddress())));
```
sophokles73: My understanding of the Anonymous Relay spec is that the corresponding target address is `null` and not the empty string. If we agree on that, then I would rather only check for a `null` target address here and instead fix the places in Hono where we establish an anonymous sender link with the dispatch router accordingly. AFAIK we do not explicitly state anywhere in Hono's API that we accept an empty string instead of `null` as the target address to indicate an anonymous sender, do we? In fact, IMHO we should use this issue to amend the AMQP User Guide with appropriate wording making clear that we expect a client to use `null` as the target address instead of an emtpy string ...

calohmn: > My understanding of the Anonymous Relay spec is that the corresponding target address is `null` and not the empty string. If we agree on that, then I would rather only check for a `null` target address here

So you mean not supporting the empty target address any more, correct?
We should reject this case in `createSender()` then already ("unsupported address"), IMHO.

> instead fix the places in Hono where we establish an anonymous sender link with the dispatch router accordingly

There are no such places any more AFAIK. The `DelegatedCommandSenderImpl` used this but was changed to use an explicit target address.

> In fact, IMHO we should use this issue to amend the AMQP User Guide [...]

Yes, I'll update that.

sophokles73: > So you mean not supporting the empty target address any more, correct?

Yes, we should only consider the `null` target address as the one targeting the anonymous relay.

> We should reject this case in createSender() then already ("unsupported address"), IMHO.

Not sure if I understand correctly. Do you mean to not accept `""` as the target address at all? Again, I would argue that the empty string is a valid address and should therefore be supported (though no Hono component uses it currently).

> There are no such places any more AFAIK. The DelegatedCommandSenderImpl used this but was changed to use an explicit target address.

Even better :-)

calohmn: > Not sure if I understand correctly. Do you mean to not accept `""` as the target address at all? Again, I would argue that the empty string is a valid address and should therefore be supported (though no Hono component uses it currently).

Well, if we don't not allow `link.getTarget().getAddress()` to be the empty string in the above code snippet, there will always be an exception in invoking `createSender()` with an empty string address. Which would mean we don't support it at all. 
That's why I deduced you wanted to remove support for it.

> Again, I would argue that the empty string is a valid address and should therefore be supported

Yes, it also would feel a bit strange rejecting a valid address (AMQP spec and Proton wise) in the generic HonoConnectionImpl class.

So, we keep it as suggested in this PR?

sophokles73: No, but I think the *isLinkEstablished* check becomes much simpler for a sender:

```
return link.getRemoteTarget() != null;
```

or am I mistaken?

sophokles73: @calohmn any additional thoughts on this?

calohmn: > No, but I think the isLinkEstablished check becomes much simpler for a sender:

Right, I've changed it to that.

PR #1550, opened 2019-10-07
Add cors support to release notes and remove proxy from the spec

dejanb:


site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -61,12 +61,10 @@ tags:
         url: https://www.eclipse.org/hono/docs/concepts/device-identity/
 
 servers:
-   - url: '{corsProxy}{server}/v1'
+   - url: '{server}/v1'
```
sophokles73: not sure if we want to deploy this immediately because the Sandbox is still running 1.0.0-M7 ...
on the other hand, this will be deployed as part of the *dev* docs whereas the 1.0.0-M7 docs will still contain the CORS proxy. WDYT?

dejanb: I haven't looked closely into sandbox lately, but I would assume we would need to upgrade it manually. So nothing would break unexpectedly?

sophokles73: Well, if we expect the Sandbox to run the latest snapshot (nightly) then yes, we need to update it manually. However, if we are ok with the Sandbox running the latest milestone, then we do not need to do anything. Personally, I tend to the latter ...

dejanb: Hm, yeah. I don't think we can always guarantee that dev docs will work with sandbox. And I think we should do this before the release anyways.

sophokles73: > And I think we should do this before the release anyways.

Do *what*?

dejanb: Merge this change in the spec.

sophokles73: Yes, IMHO you should merge this PR *now* and we should **not** update the Sandbox with a 1.0.0-SNAPSHOT but instead simply wait for the 1.0.0 release and then update the Sandbox. Do you agree?

dejanb: Yeah, that sounds good

PR #1557, opened 2019-10-09
Update concept documentation

b-abel:
This PR contains various updates and improvements to the concept pages in the documentation. To make the review easier it is split into separate commits.

### DIFF
--- a/site/documentation/content/concepts/device-identity.md
+++ b/site/documentation/content/concepts/device-identity.md
@@ -12,7 +12,7 @@ Hono does not make any assumptions about the format of a device identifier (or *
 
 ## Tenant
 
-Hono supports the logical partitioning of devices into groups called *tenants*. Each tenant has a unique identifier, a string called the *tenant-id*, and can be used to provide a logical grouping of devices belonging e.g. to the same application scope or organizational unit. Each device can thus be uniquely identified by the tuple (*tenant-id*, *device-id*). This tuple is broadly used throughout Hono's APIs when addressing a particular device.
+Hono supports the logical partitioning of devices into groups called [*tenants*]({{< relref "/concepts/tenancy" >}}). Each tenant has a unique identifier, a string called the *tenant-id*, and can be used to provide a logical grouping of devices belonging e.g. to the same application scope or organizational unit. Each device can thus be uniquely identified by the tuple (*tenant-id*, *device-id*). This tuple is broadly used throughout Hono's APIs when addressing a particular device.
 
 ## Device Registration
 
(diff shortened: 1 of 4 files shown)


site/documentation/content/concepts/tenancy.md
```diff
@@ -49,7 +49,7 @@ only a subset of Hono's functionality. This feature is foreseen to be especially
 
 
 Please refer to the [Tenant API]({{< relref "/api/tenant" >}}) to find out which protocol adapter properties are
-configurable on a tenant level. The number of such properties may grow during further development of Hono.
+configurable on a tenant level.
```
sophokles73: what about 
> ... which protocol adapter properties can be configured at the tenant level.

?

site/documentation/content/concepts/tenancy.md
```diff
@@ -72,9 +72,7 @@ is addressing the tenants themselves.
 All devices of Hono belong to exactly one tenant, so any data is sent in the context of the
 tenant a device belongs to.
 
-It is important here to focus on Hono's view of a device: it is on a logical level, meaning a device is created in the 
-context of a tenant purely by using Hono's Credentials and Device Registration API. 
-
+"Device" means here a logical entity consisting of a set of registry data.
```
sophokles73: IMHO we can completely omit this sentence and rephrase the following one as
> A physical device will usually be represented in Hono as an entity in the device registry, having a unique identity and belonging to exactly one tenant.

WDYT?

b-abel: That sounds great, but there is an overlap with the previous sentence:  

> All devices of Hono belong to exactly one tenant, so any data is sent in the context of the
tenant a device belongs to. A physical device will usually be represented in Hono as an entity in the device registry, having a unique identity and belonging to exactly one tenant.

We could skip the first sentence and put it like this:

> A physical device will usually be represented in Hono as an entity in the device registry, having a unique identity and belonging to exactly one tenant. All data sent from a device, as well as from the application to the device, is therefore treated as belonging to the corresponding tenant.

WDYT?

site/documentation/content/concepts/device-identity.md
```diff
@@ -12,7 +12,7 @@ Hono does not make any assumptions about the format of a device identifier (or *
 
 ## Tenant
 
-Hono supports the logical partitioning of devices into groups called *tenants*. Each tenant has a unique identifier, a string called the *tenant-id*, and can be used to provide a logical grouping of devices belonging e.g. to the same application scope or organizational unit. Each device can thus be uniquely identified by the tuple (*tenant-id*, *device-id*). This tuple is broadly used throughout Hono's APIs when addressing a particular device.
+Hono supports the logical partitioning of devices into groups called [*tenants*]({{< relref "/concepts/tenancy" >}}). Each tenant has a unique identifier, a string called the *tenant-id*, and can be used to provide a logical grouping of devices belonging e.g. to the same application scope or organizational unit. Each device can thus be uniquely identified by the tuple (*tenant-id*, *device-id*). This tuple is broadly used throughout Hono's APIs when addressing a particular device.
 
 ## Device Registration
```
b-abel: This section sounds as if Hono always requires authentication of devices. In fact, this can currently be turned off for most protocol adapters.

b-abel:
The concept page about [Device Authentication](https://www.eclipse.org/hono/docs/dev/concepts/device-identity/#device-authentication) sounds as if Hono always requires authentication of devices. In fact, this can currently be turned off for most protocol adapters. Should we add a sentence here (or remove the unauthenticated mode altogether ;) )?

b-abel:
I think we should squash this before merging it.

sophokles73:
> In fact, this can currently be turned off for most protocol adapters. Should we add a sentence here (or remove the unauthenticated mode altogether ;) )?

That is true, but by default, devices are required to authenticate so I do not see a problem here ...

PR #1560, opened 2019-10-10
Prevent that anchors are positioned behind menu bar in documentation.

b-abel:
This adds CSS to prevent that anchors are positioned behind the (sticky) menu bar in the documentation (similar to the solution for the homepage).

### DIFF
--- a/site/documentation/static/css/hono.css
+++ b/site/documentation/static/css/hono.css
@@ -89,3 +89,11 @@
 .swagger-ui table, .swagger-ui thead, .swagger-ui td {
 	border: none;
 }
+
+/* prevents that the anchor is positioned behind the menu bar */
+:target:before {
+  content: "";
+  display: block;
+  height: 60px;
+  margin: -60px 0 0;
+}


sophokles73:
@b-abel I have no clue what this is about. Maybe a screenshot would help to illustrate the issue?

b-abel:
If you click on a link in the documentation whose target is within the documentation (see the address bar in the screenshots), the browser scrolls to that place (usually a heading). But this is hidden by the sticky menu bar. The CSS makes sure that you don't scroll that far and that you can still see the headline. 

Before this PR:

before

After:

after

Is it understandable now?

site/documentation/static/css/hono.css
```diff
@@ -89,3 +89,11 @@
 .swagger-ui table, .swagger-ui thead, .swagger-ui td {
 	border: none;
 }
+
+/* prevents that the anchor is positioned behind the menu bar */
+:target:before {
+  content: "";
+  display: block;
+  height: 60px;
```
sophokles73: is this number of pixels determined from the currently used heading font?

b-abel: No, this does not depend on the size of headings, but on that of the menu bar. The height of the menu bar is nowhere absolutely defined, so this is an approximation. The menu bar is usually 48px high (but this is calculated dynamically in the JS code for the sticky menu bar and could change if you put large elements into it).
TBH my solution is a bit hacky. I spent a lot of time to solve this for the theme used for the homepage part and did not find a better solution. So I am glad that this could easily be adapted for the documentation part.

sophokles73: I see. Fair enough.

PR #1561, opened 2019-10-10
Minor cosmetic changes

Alfusainey:
Change Summary:

* Use the specialized IntPredicate instead of Predicate<Integer> for HTTP status codes
* Replace occurrences of non-constant "LOG" variable to "log"

plus other minor fixes

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
@@ -101,7 +101,7 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends MqttProtoc
 
     private static final int IANA_MQTT_PORT = 1883;
     private static final int IANA_SECURE_MQTT_PORT = 8883;
-    private static final String KEY_TOPIC_NAME_FILTER = "filter";
+    private static final String KEY_TOPIC_FILTER = "filter";
 
     private MqttAdapterMetrics metrics = MqttAdapterMetrics.NOOP;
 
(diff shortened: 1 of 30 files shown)


adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -101,6 +101,7 @@
 
     private static final int IANA_MQTT_PORT = 1883;
     private static final int IANA_SECURE_MQTT_PORT = 8883;
+    private static final String KEY_TOPIC_NAME_FILTER = "filter";
```
sophokles73: IMHO that should be `KEY_TOPIC_FILTER`

client/src/main/java/org/eclipse/hono/client/CommandContext.java
```diff
@@ -46,6 +46,8 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(CommandContext.class);
 
+    private static final String ERROR_MSG_INVALID_CREDIT_VALUE = "";
```
sophokles73: IMHO the value should be the message, right?

client/src/main/java/org/eclipse/hono/client/CommandResponse.java
```diff
@@ -34,8 +34,8 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(CommandResponse.class);
 
-    private static final Predicate<Integer> INVALID_STATUS_CODE = code ->
-        code == null || code < 200 || (code >= 300 && code < 400) || code >= 600;
+    private static final IntPredicate INVALID_STATUS_CODE = code ->
```
sophokles73: this change seems to require *more* code being written ..
please revert this change.

sophokles73: No, the idea is to not do this at all because it doesn't make things easier.

sophokles73:
Thanks for contributing :+1:

Issue #1555, opened 2019-10-09
Inconsistent validation of request payload in tenant registration operations.


kaniyan:
There has been some inconsistency in the validation of the request payload during tenant registration operations in the _file based device registry_. In some cases, it's been able to _add_ or _update_ a tenant even with attributes containing _invalid_ values.

For example, it's been **able** to _add_ or _update_ a tenant with invalid _minimum message size_ or  _sampling mode_. After that the _get_ tenant operation returns `400 Bad Request Error`.
 
```
{
    "enabled": true,
    "minimum-message-size": -100,
    "tracing":{
    	"sampling-mode": "xyz"
    }
}
```

On the other hand, it is **not able** to `add` or `update` a tenant with the below invalid request payload containing negative value for `max-connections`. In this case `400 Bad Request Error` is returned as expected.

```
{
    "enabled": true,
    "resource-limits": {
        "max-connections": -100
    }
}
```
**Findings:** 

In the _file based device registry_, while performing _add_ and _update_ operations, the request payload is validated and mapped to `TenantObject` and then stored in a file. And they are not validated against the Tenant Management API. A solution shall be, that before accepting the payload, it should also be validated against the `org.eclipse.hono.service.management.tenant.Tenant`.

sophokles73:
@kaniyan can this be closed?

kaniyan:
This has been closed through PR #1562.

PR #1562, opened 2019-10-10
[#1555] Validate tenant management payload before adding or updating a tenant.

kaniyan:
This PR fixes the issue #1555. A solution to issue #1555 shall be validating the _tenant management payload_ before _adding_ or _updating_ a tenant in the `file based device registry`.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/management/tenant/EventBusTenantManagementAdapter.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/management/tenant/EventBusTenantManagementAdapter.java
@@ -173,7 +173,7 @@ public abstract class EventBusTenantManagementAdapter extends EventBusService {
         }
     }
 
-    Future<EventBusMessage> processGetRequest(final EventBusMessage request) {
+    private Future<EventBusMessage> processGetRequest(final EventBusMessage request) {
 
         final String tenantId = request.getTenant();
         final SpanContext spanContext = request.getSpanContext();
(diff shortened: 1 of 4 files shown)


services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedTenantService.java
```diff
@@ -565,4 +573,15 @@ private String generateTenantId() {
...
+    private boolean isValidTenantPayload(final JsonObject jsonObject, final Span span) {
+        try {
+            jsonObject.mapTo(Tenant.class);
+            return true;
+        } catch (final IllegalArgumentException e) {
+            log.debug("Error parsing payload of tenant request", e);
+            TracingHelper.logError(span, e);
+            return false;
+        }
+    }
```
sophokles73: The request payload is already being checked in https://github.com/eclipse/hono/blob/0999336ef6aaad9c629878046619cc8a78ee376d/service-base/src/main/java/org/eclipse/hono/service/management/tenant/EventBusTenantManagementAdapter.java#L204

I guess you should add the additional check there ...

kaniyan: @sophokles73 Thanks for pointing out and I will update here. But it seems to me that if the Tenant API attributes get changed then above validations should be adapted manually. 

IMHO `Tenant.class` acts as a blueprint and the validation can be automated by trying to map the payload with this class as in this PR. For example `hasValidAdapterSpec()` method is not needed if the payload is already validated with `Tenant.class`. Any validations that are not specified in the `Tenant.class` can be added additionally above. WDYT?

sophokles73: absolutely :+1:

kaniyan: @sophokles73 I have pushed the changes. Would you mind having a look?

service-base/src/main/java/org/eclipse/hono/service/management/tenant/EventBusTenantManagementAdapter.java
```diff
@@ -201,38 +201,19 @@ protected final String getEventBusAddress() {
...
-                return false;
-            } else {
-                final boolean containsInvalidAdapter = adapters.stream()
-                        .anyMatch(obj -> !(obj instanceof JsonObject) ||
-                                !((JsonObject) obj).containsKey(RegistryManagementConstants.FIELD_ADAPTERS_TYPE));
-                if (containsInvalidAdapter) {
-                    return false;
-                }
+        try {
+            payload.mapTo(Tenant.class);
```
sophokles73: this is not equivalent to the original checks because the *setAdapters* method doesn't prevent setting an empty list (yet).

service-base/src/main/java/org/eclipse/hono/service/management/tenant/EventBusTenantManagementAdapter.java
```diff
@@ -201,38 +201,19 @@ protected final String getEventBusAddress() {
...
+            if (hasValidTrustedCaSpec(payload)) {
```
sophokles73: IMHO this should be
```
return hasValidTrustedCaSpec(payload);
```

service-base/src/main/java/org/eclipse/hono/service/management/tenant/EventBusTenantManagementAdapter.java
kaniyan: @sophokles73 I have added a separate method for the above validation and pushed the changes. Pls take a look.

service-base/src/main/java/org/eclipse/hono/service/management/tenant/EventBusTenantManagementAdapter.java
```diff
@@ -315,6 +294,24 @@ private boolean hasValidTrustedCaSpec(final JsonObject payload) {
         }
     }
 
+    /**
+     * Checks if a payload contains no empty adapters list.
+     *
+     * @param payload The payload to check.
+     * @return boolean {@code true} if there is no empty adapters list.
+     */
+    private boolean hasNoEmptyAdapterSpec(final JsonObject payload) {
```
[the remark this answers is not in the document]
kaniyan: I have moved the check to the `Tenant.setAdapters()`. Please take a look again.

service-base/src/main/java/org/eclipse/hono/service/management/tenant/Tenant.java
```diff
@@ -166,12 +166,16 @@ public final Tenant setDefaults(final Map<String, Object> defaults) {
...
      * @param adapters The configuration properties.
      * @return This instance, to allow chained invocations.
+     * @throws IllegalArgumentException if the adapters list is empty or {@code null}
      */
     public final Tenant setAdapters(final List<Adapter> adapters) {
-        this.adapters.clear();
-        if (adapters != null) {
-            this.adapters.addAll(adapters);
+
+        if (adapters == null || adapters.isEmpty()) {
```
sophokles73: I do not quite see why we need to prevent `null`here. IMHO setting to `null` should simply clear the list (as before).

kaniyan: @sophokles73 Done.

Issue #1558, opened 2019-10-10
Device Registry should not accept already hashed passwords


jbtrystram:
Hi, 

It was discussed [1] during the management API refactor that we should not accept passwords already hashed by the client, as the client could pick a weak hashing algorithm. This decision has not been implemented yet, as far as I know :) 

[1] : https://github.com/eclipse/hono/pull/1130#issuecomment-479829977

PR #1565, opened 2019-10-11
[#1558] refuse already hashed password and add fields to secrets

jbtrystram:
Follow up to #1377 : 
- HTTP Management API now refuses PUT credentials request containing hashed passwords. (400 Bad request)
- An `id` field is attached to each secret.
- Reponses to GET on credentials from the managment API do not contain hashed-password details anymore.
- updated integration tests to now use plain text passwords and verify each secret contains an ID.

The second part of this work is to filter `PUT` requests to the device registry credentials resources to allow the update of a single password using the `id` field.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordSecret.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordSecret.java
@@ -138,6 +138,9 @@ public class PasswordSecret extends CommonSecret {
         if (!Strings.isNullOrEmpty(passwordHash)) {
             throw new IllegalStateException(String.format("'%s' must be empty", RegistryManagementConstants.FIELD_SECRETS_PWD_HASH));
         }
+        if (!Strings.isNullOrEmpty(salt)) {
+            throw new IllegalStateException(String.format("'%s' must be empty", RegistryManagementConstants.FIELD_SECRETS_SALT));
+        }
     }
 
     /**
--- a/service-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordSecret.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordSecret.java
@@ -124,6 +124,23 @@ public class PasswordSecret extends CommonSecret {
     }
 
     /**
+     * Verifies that only the "password-plain" field is set in the Password Secret, i.e. the password is not hashed.
+     *
+     * @throws IllegalStateException if this is not the case.
+     */
+    public void checkIsPlainText() {
+        if (Strings.isNullOrEmpty(passwordPlain)) {
+            throw new IllegalStateException(String.format("'%s' must be provided", RegistryManagementConstants.FIELD_SECRETS_PWD_PLAIN));
+        }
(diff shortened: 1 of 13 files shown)


b-abel:
Would it make sense to remove the examples of hashed passwords from the documentation (https://www.eclipse.org/hono/docs/dev/user-guide/device-registry/#update-credentials-for-a-device) in this PR?

dejanb:
> Would it make sense to remove the examples of hashed passwords from the documentation (https://www.eclipse.org/hono/docs/dev/user-guide/device-registry/#update-credentials-for-a-device) in this PR?

And actually change the spec to reflect these changes.

service-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordSecret.java
```diff
@@ -123,6 +123,23 @@ public void checkValidity() {
...
+    public void checkIsPlainText() {
+        if (Strings.isNullOrEmpty(passwordPlain)) {
+            throw new IllegalStateException(String.format("'%s' must be provided", RegistryManagementConstants.FIELD_SECRETS_PWD_PLAIN));
+        }
+        if (!Strings.isNullOrEmpty(hashFunction)) {
+            throw new IllegalStateException(String.format("'%s' must be empty", RegistryManagementConstants.FIELD_SECRETS_HASH_FUNCTION));
+        }
+        if (!Strings.isNullOrEmpty(passwordHash)) {
+            throw new IllegalStateException(String.format("'%s' must be empty", RegistryManagementConstants.FIELD_SECRETS_PWD_HASH));
+        }
```
sophokles73: what about the *salt*? IMHO we should also not accept any salt provided by a client, right?

jbtrystram: Updated !

b-abel:
Do the examples in https://github.com/eclipse/hono/blob/master/site/documentation/content/api/credentials/index.md need an update as well?

jbtrystram:
@sophokles73 @dejanb In order to clarify this I split that PR in two, one for each feature.

I opened https://github.com/eclipse/hono/pull/1711 to address the point of the device registry returning the hashed-password secrets details. 

On the topic of refusing the hashed-passwords for the `PUT` operation we would like to revisit this, as there are some legitimate use cases for this feature.

sophokles73:
> On the topic of refusing the hashed-passwords for the PUT operation we would like to revisit this, as there are some legitimate use cases for this feature.

@jbtrystram when you say *we would like to revisit*, who do you mean with *we*? Is that a discussion that needs to/will happen outside of the Hono project?

sophokles73:
@jbtrystram does this also mean that this PR will not be part of 1.1.0 as it needs more time discussing?

jbtrystram:
@sophokles73 I mean we as the Red Hat people, would like to share our thoughts resulting internal discussions with the Hono project about #1574 that was merged earlier. 

As @ctron pointed out there are some valid use cases where having the device registry ingesting already hashed passwords make sense.

No I think this PR should not be in 1.1.0, but #1710 should.

sophokles73:
ok, then I'll unschedule it ...

PR #1574, opened 2019-10-15
Deprecate defining password secrets using password hash, function and…

dejanb:
… salt. Additionally fix credentials examples

### DIFF
--- a/site/documentation/content/api/management/device-registry-v1.yaml
+++ b/site/documentation/content/api/management/device-registry-v1.yaml
@@ -841,6 +841,10 @@ components:
             - $ref: '#/components/schemas/CommonSecret'
 
       PasswordSecret:
+         description: |
+                      Password based secret definition <br>
+                      __NOTE__: Defining password secrets with user provided password hash, function and salt is deprecated and will be removed
+                      in the upcoming versions. You should use `pwd-plain` property only going forward.
          additionalProperties: false
          allOf:
             - $ref: '#/components/schemas/CommonSecret'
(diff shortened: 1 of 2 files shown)


site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -1036,32 +1043,39 @@ components:
...
-               "pwd-hash": "AQIDBAUGBwg=",
-               "salt": "Mq7wFw==",
-               "hash-function": "sha-512"
-            }]
+            [{
+                auth-id: sensor1,
+                type: hashed-password,
+                secrets: [{
+                             "not-after": "2027-12-24T19:00:00Z",
+                             "pwd-hash": "AQIDBAUGBwg=",
```
sophokles73: now that we have deprecated these fields, I do not think that we should use them in example payloads, wouldn't you agree?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -1036,32 +1043,39 @@ components:
...
-               "pwd-hash": "",
-               "salt": "",
-               "hash-function": ""
+           [{
+               auth-id: sensor1,
+               type: hashed-password,
+               secrets: [{
+                  "id": "349556ea-4902-47c7-beb0-1009ab693fb4",
+                  "not-after": "2027-12-24T19:00:00Z",
+                  "pwd-plain": "",
```
sophokles73: see above

site/documentation/content/api/management/device-registry-v1.yaml
dejanb: I was thinking about that, but decided to leave it in for the time being as we have a lot of examples everywhere with this. I'd revisit all docs and examples as part of the #1565.

sophokles73: I see

sophokles73:
@dejanb Oh, before we forget, you should document this in the release notes as well ... under *Deprecations*

dejanb:
> @dejanb Oh, before we forget, you should document this in the release notes as well ... under _Deprecations_

It's already in there. Unless there's some other place I missed?

Issue #1523, opened 2019-09-24
Enable gateways to subscribe for commands of a specific device


calohmn:
With the implementation of #930, we have added support for gateways creating command subscriptions without having to use individual device ids. 

The gateway will for example just create a MQTT subscription or AMQP receiver link on the `command` topic/link address. Northbound applications will send commands with the specific device id in the message address and the gateway mapping logic (based on the `lastKnownGateway` information for the device) will forward such commands to the corresponding gateway.

*What is not currently possible* in such a scenario is that the applications will get notified of when a particular device is ready to receive commands.

That however is a cornerstone of the Hono command & control feature in how it is defined for non-gateway devices.

In order to support this principle also for gateways, the command handling logic in the protocol adapters should be adapted to enable gateways to alternatively subscribe for commands of a specific device.
Hono currently doesn't support this because there would be *device*-specific AMQP consumers created in the protocol adapter, whereas the gateway mapping logic would always redirect the command to a *gateway*-specific consumer.

When receiving such a command subscription with a device id from a gateway, the protocol adapter should forward a corresponding notification with the device id to the northbound application.

To sum it up, the goal here would be to have gateways be able to
- either create a single subscription for commands regarding *all devices* that are currently mapped to the gateway (there will be no individual device-ready-for-commands notification here)
- or create individual command subscriptions for specific devices - with corresponding downstream notifications.

---
As to the implementation of this:
Even though the device id is given with the subscription request, the protocol adapter shall create/use a gateway-specific command consumer and map the command received on that consumer to the original device of that command.

sophokles73:
@calohmn my understanding is that this is the overall issue for dding this functionality to *all* adapters, i.e. there will be subsequent PRs for adding the functionality to HTTP, MQTT, AMQP, right?

calohmn:
@sophokles73 Right, I'm about to create a PR for that.

calohmn:
@sophokles73 I've added the relevant PRs for
- AMQP/MQTT/HTTP adapter changes
- Integration tests
- HTTP adapter change to include the target device id in the HTTP headers of the response with the command
- Documentation / release notes

calohmn:
All corresponding PRs have been merged now.

PR #1577, opened 2019-10-15
[#1523] MQTT adapter: Let gateways subscribe for commands of a specific device

calohmn:
This is for #1523.

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
@@ -70,6 +70,7 @@ import io.opentracing.tag.Tags;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
+import io.vertx.core.Handler;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.json.JsonObject;
 import io.vertx.mqtt.MqttConnectionException;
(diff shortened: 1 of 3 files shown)


calohmn:
Note: When a gateway has subscribed for commands _of a specific device_ (as introduced as a feature here), the topic, that the gateway will receive the commands on, will still contain the device id, although that is in a way superfluous.

Nonetheless, I think it's better to use a common topic structure along with the case that the gateway has subscribed for commands of _all_ his connected devices.

Relevant code fragment:
https://github.com/eclipse/hono/blob/90a391f100be39b2e500ade589e73d071d306ac1/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java#L1356-L1364

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -730,47 +731,51 @@ protected final void onUnsubscribe(
...
+                        sub.getTenant(),
+                        tenantTracker.result(),
+                        ProcessingOutcome.from(failure),
+                        command.getPayloadSize(),
+                        timer);
+                return null;
+            });
+        };
+        final Handler<Void> remoteCloseHandler = remoteClose -> {};
+        if (sub.getAuthenticatedDeviceId() != null && !sub.getAuthenticatedDeviceId().equals(sub.getDeviceId())) {
```
sophokles73: this check looks like it should be implemented in a method (e.g. *isGatewaySubscription*) in `Subscription`, don't you think?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandSubscription.java
```diff
@@ -101,14 +102,14 @@ private CommandSubscription(final String topic, final Device authenticatedDevice
             }
         } else {
             isAuthenticated = true;
-            if ((tenant != null && !authenticatedDevice.getTenantId().equals(tenant)) ||
-                    (deviceId != null && !authenticatedDevice.getDeviceId().equals(deviceId))) {
+            if (tenant != null && !authenticatedDevice.getTenantId().equals(tenant)) {
```
sophokles73: I think the message in the thrown exception needs to be adapted as well

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
calohmn: Good point, I've refactored it accordingly.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandSubscription.java
calohmn: Yes, I've adapted it.

PR #1579, opened 2019-10-15
[#1523] HTTP adapter: Let gateways subscribe for commands of a specific device

calohmn:
This is for #1523.

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -57,6 +57,7 @@ import io.opentracing.contrib.vertx.ext.web.WebSpanDecorator;
 import io.opentracing.tag.Tags;
 import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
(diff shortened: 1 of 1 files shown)


adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -940,93 +947,110 @@ protected void setNonEmptyResponsePayload(final HttpServerResponse response, fin
...
+                log.debug("command message is invalid: {}", command);
+                commandContext.reject(new ErrorCondition(Constants.AMQP_BAD_REQUEST, "malformed command message"));
+            }
+            // we do not issue any new credit because the
+            // consumer is supposed to deliver a single command
+            // only per HTTP request
+        };
+
+        final Handler<Void> remoteCloseHandler = remoteDetach -> {
+            log.debug("peer closed command receiver link [tenant-id: {}, device-id: {}]", tenantObject.getTenantId(), deviceId);
```
sophokles73: what about including the gateway ID here as well?

calohmn: Yes, I've added that.

PR #1580, opened 2019-10-15
[#1523] Add integration tests for gateways subscribing for device commands

calohmn:
This is for #1523.

Note that the tests will fail until #1578, #1577 and #1579 are merged. I will rebase this PR then.
EDIT: branch is rebased now.

### DIFF
--- a/tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
@@ -38,6 +38,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
 
 import org.eclipse.hono.client.HonoConnection;
 import org.eclipse.hono.client.ServiceInvocationException;
@@ -682,26 +683,22 @@ public final class IntegrationTestSupport {
     }
 
     /**
-     * Registers a new device for a tenant that is connected via a gateway.
+     * Registers a new device for a tenant that is connected via the given gateway.
      * 
      * @param tenantId The tenant that the gateway and device belong to.
-     * @param deviceId The device identifier.
-     * @param isGatewayDevice {@code true} if the given device is a gateway.
+     * @param gatewayId The gateway identifier.
      * @param timeoutSeconds The number of seconds to wait for the setup to succeed.
-     * @return The device identifier to use for sending commands.
-     *         The identifier will be that of the given device if it is not a gateway,
-     *         otherwise it will be the identifier of the newly registered device.
+     * @return The device identifier of the newly registered device.
      * @throws IllegalStateException if setup failed.
      */
     public String setupGatewayDeviceBlocking(
             final String tenantId,
-            final String deviceId,
-            final boolean isGatewayDevice,
+            final String gatewayId,
             final int timeoutSeconds) {
 
         final CompletableFuture<String> result = new CompletableFuture<>();
 
-        setupGatewayDevice(tenantId, deviceId, isGatewayDevice)
+        setupGatewayDevice(tenantId, gatewayId)
         .setHandler(attempt -> {
             if (attempt.succeeded()) {
                 result.complete(attempt.result());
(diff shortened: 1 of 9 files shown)


tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
```diff
@@ -1059,6 +1046,23 @@ public String setupGatewayDeviceBlocking(
...
+     *
+     * @param forceCommandRerouting Supplies the value for the "force-command-rerouting" property. A {@code true}
+     *                              value of this property causes the command message to be rerouted to the
+     *                              AMQP messaging network, mimicking the behaviour when the command message
+     *                              has reached a protocol adapter instance that the command target device is
+     *                              not connected to, so that the message needs to be delegated to the correct
+     *                              protocol adapter instance. See the <em>CommandConsumerFactoryImpl</em> class.
+     * @return The properties map.
+     */
+    public static Map<String, Object> getCommandMessageProperties(final Supplier<Boolean> forceCommandRerouting) {
```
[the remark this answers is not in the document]
calohmn: Yes, I've changed that.

calohmn:
@sophokles73 There have been recurrent failures in the HTTP tests (`testUploadMessagesWithTtdThatReplyWithOneWayCommand`) in connection with the `GATEWAY_FOR_SINGLE_DEVICE` mode. 
It occurs in case there is a new `ttd` request coming in at the HTTP adapter, while the command consumer for the last request hasn't been closed yet. This doesn't cause the issue directly (it also happened before the changes for #1523) but for some reason there are now some responses sent back from the HTTP adapter that don't contain commands. The issue seems to occur among the test runs that explicitly trigger the commands to be rerouted after gateway-mapping via Qpid. 

For now I've disabled the HTTP adapter test `GATEWAY_FOR_SINGLE_DEVICE` parameter combinations along with a corresponding `TODO`.

tests/src/test/java/org/eclipse/hono/tests/http/HttpTestBase.java
```diff
@@ -1103,6 +1114,18 @@ public void testUploadMessagesWithTtdThatReplyWithOneWayCommand(
                 });
     }
 
+    private Future<MultiMap> sendHttpRequestForGatewayOrDevice(final Buffer payload, final MultiMap requestHeaders,
+            final HttpCommandEndpointConfiguration endpointConfig, final String requestDeviceId) {
+        if (endpointConfig.isSubscribeAsGatewayForSingleDevice()) {
+            final String uri = getEndpointUri() + "/" + tenantId + "/" + requestDeviceId;
+            // GW uses PUT when acting on behalf of a device
+            return httpClient.update(uri, payload, requestHeaders, status -> status == HttpURLConnection.HTTP_OK);
```
sophokles73: looks like you found the issue, too ;-) I had been investigating the problems with the test as well and also stumbled across this line ...

However,  I do not quite understand, why this fixes the issue. More specifically, I wonder, why during the test runs, the HTTP adapter actually seems to return a 202 to some of the requests. Do you know why that is?

calohmn:
Found the issue to be a rather trivial one in the HTTP integration test class (wrong status code check).
I've removed the `TODO` and reinstated the `GATEWAY_FOR_SINGLE_DEVICE` tests.

tests/src/test/java/org/eclipse/hono/tests/http/HttpTestBase.java
calohmn: In the test runs, where 202 is returned, the creation of a command consumer fails because there is still a consumer from a previous test run which has not been fully closed yet. This causes a `ResourceConflictException` to be thrown and higher up caught in the HTTP adapter `createCommandConsumer` method:
https://github.com/eclipse/hono/blob/a170706410e9988a8a8471d6ef56ed5face2dcea/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java#L1040-L1050

So, the `let the other request handle the command (if any)` is taken, the `responseReady` Future is triggered (with `null` as value which causes no `ttd` notification to be sent downstream!) and the response processing continues in the main `doUploadMessage` method, where this method
https://github.com/eclipse/hono/blob/a170706410e9988a8a8471d6ef56ed5face2dcea/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java#L717
will set a 202 response because no command was received.

sophokles73: When running the test where we expect one-way commands, we sometimes see re-tries of the request because the initial request yielded a 412. I erroneously assumed that the 412 was actually indicating exactly this case. However, the 412 (precondition failed) just indicates that the response contained an unexpected status code, in this case *not 200 but 202* ...
Ok, all seems to be in good shape now. Great work, @calohmn !!

calohmn: Yes, the fact that the original response status code is hidden and only a 412 is shown has caused a great deal of confusion with me as well.
https://github.com/eclipse/hono/blob/99b39c4c9fa4bc9bf99062e2b924c7e7df42d7af/tests/src/test/java/org/eclipse/hono/tests/CrudHttpClient.java#L598-L600
We should change this so that the original status code is somehow taken over (maybe just in the exception text). Then we can change the log output so that there isn't just the 412 here:
`re-trying request [43] which failed with unexpected status code 412`

sophokles73: I agree, are you going to fix this?

calohmn: yes

PR #1581, opened 2019-10-16
[#1523] HTTP adapter: Add http header with command target device

calohmn:
This is for #1523.
When the HTTP adapter sends a command to a gateway, acting on behalf of a device, a HTTP header will be set with the target device id.

(An integration test for this can be added once #1580 has been merged.)

### DIFF
--- a/site/documentation/content/user-guide/http-adapter.md
+++ b/site/documentation/content/user-guide/http-adapter.md
@@ -46,6 +46,7 @@ Before accepting any telemetry or event or command messages, the HTTP adapter ve
   * (optional) `content-type`: A media type describing the semantics and format of payload contained in the response body. This header will only be present if the response contains a command to be executed by the device which requires input data.
   * (optional) `hono-command`: The name of the command to execute. This header will only be present if the response contains a command to be executed by the device.
   * (optional) `hono-cmd-req-id`: An identifier that the device must include in its response to a command. This header will only be present if the response contains a command to be executed by the device.
+  * (optional) `hono-cmd-target-device`: The id of the device that shall execute the command. This header will only be present if the response contains a command to be executed by the device and if the response goes to a gateway that acts on behalf of the target device.
 * Response Body:
   * (optional) Arbitrary data serving as input to a command to be executed by the device, if status code is 200.
   * (optional) Error details, if status code is >= 400.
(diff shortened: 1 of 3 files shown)


site/documentation/content/user-guide/http-adapter.md
```diff
@@ -46,6 +46,7 @@ Before accepting any telemetry or event or command messages, the HTTP adapter ve
   * (optional) `content-type`: A media type describing the semantics and format of payload contained in the response body. This header will only be present if the response contains a command to be executed by the device which requires input data.
   * (optional) `hono-command`: The name of the command to execute. This header will only be present if the response contains a command to be executed by the device.
   * (optional) `hono-cmd-req-id`: An identifier that the device must include in its response to a command. This header will only be present if the response contains a command to be executed by the device.
+  * (optional) `hono-cmd-target-device`: The id of the device that shall execute the command. This header will only be present if the response contains a command to be executed by the device and if the response goes to a gateway that acts on behalf of the target device.
```
sophokles73: This means that the authenticated device is indeed a gateway and has published a message on its own behalf including a TTD, right? Is it really possible that the response then contains a command for one of the devices that this gateway is allowed to act on behalf of?

calohmn: > This means that the authenticated device is indeed a gateway and has published a message on its own behalf including a TTD, right?

Yes

> Is it really possible that the response then contains a command for one of the devices that this gateway is allowed to act on behalf of?

Yes. The requirements for this are:
- (1) the northbound application has sent the command on the "command/tenantId" endpoint, not using the legacy control endpoint (so that the gateway-mapping is applied),
- (2) the device that the command is targeted at has last sent a telemetry/event message via the same gateway that has subscribed here,
- (3) the gateway hasn't issued a device-specific command subscription for the device the command is targeted at.

Because of (2), the gateway-mapping logic targets the incoming command to the mapped gateway.

Admittedly, this whole scenario of a gateway, wanting to get one command via HTTP for any of the devices it acts on behalf of, could be seen as rather exotic, because the northbound application would have to know about the gateway, so that it uses the ttd-notification message with that gateway id to send a command.

But then again the behaviour here is consistent to the one in the AMQP and MQTT adapters, where such a subscription makes more sense, with the long-lived nature of the subscriptions there.

sophokles73: Ok, I understand. If you ask me, I think that we should add a *note* to the HTTP user guide here, explaining these rare circumstances. Otherwise, we will have forgotten about them in four weeks time already ... ;-)

PR #1583, opened 2019-10-16
Add namespace to the templates

ctron:
I have no real experience with Helm, but to me it looks like we are missing the namespace in the metadata section. Otherwise using `helm template --namespace foo` will not work.

### DIFF
--- a/deploy/src/main/deploy/helm/templates/_helpers.tpl
+++ b/deploy/src/main/deploy/helm/templates/_helpers.tpl
@@ -51,6 +51,7 @@ The scope passed in is expected to be a dict with keys
 */}}
 {{- define "hono.metadata" -}}
 name: {{ .dot.Release.Name }}-{{ .name }}
+namespace: {{ .dot.Release.Namespace }}
 labels:
   app.kubernetes.io/name: {{ template "hono.name" .dot }}
   helm.sh/chart: {{ template "hono.chart" .dot }}


sophokles73:
I do not see why we need this. You indicate the namespace that you want to deploy to on the command line, either using `helm install ... --namespace hono` or `kubectl apply -f ./eclipse-hono -R -n hono`.

There is no need to encode the namespace into the resource files.

sophokles73:
@dejanb @ctron I do not think that we need this and therefore do not think that we should merge this PR.

dejanb:
@ctron can you give a bit more details on what issues you're having with namespaces?

ctron:
Helm provides the switch `--namespace`:

~~~
      --namespace string         namespace to install the release into
~~~

However, that has no effect if the namespace field is not present.

So when you follow the documentation that we have around helm (https://www.eclipse.org/hono/docs/deployment/helm-based-deployment/), then this will simply not work.

I agree that you can using `kubectl apply` with `-n` to achieve the same effect, but if helm offers a `--namespace` argument, then I think we should support it.

sophokles73:
> However, that has no effect if the namespace field is not present.

I do that all the time and all the resources are deployed to the intended namespace (`hono` in my case).

sophokles73:
@ctron what makes you believe that the `--namespace` param has no effect when deploying?

ctron:
> @ctron what makes you believe that the `--namespace` param has no effect when deploying?

You can easily try it:

~~~
[jreimann@jreimann example]$ cat templates/configmap.yaml 
kind: ConfigMap
apiVersion: v1
metadata:
  name: foo
[jreimann@jreimann example]$ helm template .
---
# Source: example/templates/configmap.yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: foo

[jreimann@jreimann example]$ helm template . --namespace bar
---
# Source: example/templates/configmap.yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: foo

~~~

ctron:
With the change you have the correct behavior:

~~~
[jreimann@jreimann example]$ cat templates/configmap.yaml 
kind: ConfigMap
apiVersion: v1
metadata:
  name: foo
  namespace: {{ .Release.Namespace }}
[jreimann@jreimann example]$ helm template .
---
# Source: example/templates/configmap.yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: foo
  namespace: package-zero

[jreimann@jreimann example]$ helm template . --namespace bar
---
# Source: example/templates/configmap.yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: foo
  namespace: bar

~~~

sophokles73:
I do not see why you want/need the generated resource files to contain the namespace property. If you want to deploy the resource files to a specific namespace, use `kubectl -n` ...
You seem to assume that Helm's `--namespace` argument could/should be used to add the namespace property to the resource files. I have no idea if that is what the Helm creators had intended it to do. But since we clearly describe how Hono can be deployed to a specific namespace either using `helm install` or `kubectl apply`, I do not see why we need to add this feature based on that (unproven) assumption.

ctron:
I am not sure what you mean by unproven assumption. You can see the output of the behavior in the previous comments. You can also see the output of the `--help` command from helm. Which is also in the Helm documentation.

You can also have a look at how others do it: https://github.com/helm/charts/blob/master/stable/prometheus-operator/templates/prometheus/ingress.yaml#L14

sophokles73:
I was referring to the fact that I do not know if the Helm authors have intended the behavior you seem to assume with the `--namespace` option used in conjunction with `helm template`.
But in the end, I do not care anymore because the outcome will be the same anyway. If you merge this PR and somebody uses it the way you assume it should be used, then the resource files will contain the namespace property (preventing them from being deployed to another namespace?), but otherwise, if you just use `helm template` without specifying `--namespace`, then the templates will come out as they do now and you can deploy them to any namespace you like using `kubectl -n`.
I would, however then propose to change the deployment instructions to omit the `--namespace` argument from the `helm template` command (as it has no meaning anyway, as you pointed out).

deploy/src/main/deploy/helm/templates/_helpers.tpl
```diff
@@ -51,6 +51,7 @@ The scope passed in is expected to be a dict with keys
 */}}
 {{- define "hono.metadata" -}}
 name: {{ .dot.Release.Name }}-{{ .name }}
+namespace: {{ .dot.Release.Namespace }}
```
sophokles73: not sure if `.Release.Namespace` will always have a value, i.e. even if you do not specify the `--namespace` argument.
It might be safer to wrap the whole property in that case, so that it is omitted if there is no namespace value:
```
{{- with .dot.Release.Namespace }}
namespace: {{ . }}
{{. end }}
```
WDYT?

ctron: Yes, it always has a value: https://github.com/helm/helm/blob/master/cmd/helm/template.go#L136

sophokles73:
> I would, however then propose to change the deployment instructions to omit the --namespace argument from the helm template command (as it has no meaning anyway, as you pointed out).

I tried this out and it seems that the Prometheus and Grafana charts also write the namespace property to the resource files. So we need to keep the `--namespace` argument in the `helm template` invocation.

@ctron are you going to merge this for 1.0.0?

sophokles73:
@ctron last chance to get this into 1.0.0 :-)

PR #1596, opened 2019-10-22
Azure: Fix ARM template and provide option to define k8s version for AKS

kaizimmerm:
- Fixed an issue where the templates where not valid.
- Added option to define k8s version of AKS (especially necessary when the default in the template is no longer supported by Azure).
- Small fix on shell command as provided in Azure deployment guidance.

Kai

### DIFF
--- a/deploy/src/main/deploy/azure/arm/honoInfrastructureDeployment.json
+++ b/deploy/src/main/deploy/azure/arm/honoInfrastructureDeployment.json
@@ -16,6 +16,13 @@
                 "description": "Optional AKS cluster name. Otherwise calculated from uniqueSolutionPrefix."
             }
         },
+        "kubernetesVersion": {
+            "type": "string",
+            "defaultValue": "1.14.7",
+            "metadata": {
+                "description": "Kubernetes version."
+            }
+        },
         "vnetName": {
             "type": "string",
             "defaultValue": "",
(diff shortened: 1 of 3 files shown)


deploy/src/main/deploy/azure/arm/honoInfrastructureDeployment.json
```diff
@@ -318,51 +330,51 @@
         },
         "mqttPublicIPAddress": {
             "type": "string",
-            "value": "[reference('honoKubernetesDeployment').outputs.mqttIpAddress.value]"
+            "value": "[reference('honoKubernetesDeployment', '2019-06-01').outputs.mqttIpAddress.value]"
```
sophokles73: are you sure that the date needs to be in here?

kaizimmerm: It is not a date, but API version. If missing the resource manager will throw an error like 

`"Azure Error: InvalidTemplate
Message: Deployment template validation failed: 'The template output reference to 'Microsoft.Resources/deployments/honoServiceBusDeployment' requires an API version. Please see https://aka.ms/arm-template for usage details.'."
`

sophokles73: I see :-)

PR #1597, opened 2019-10-22
Share the same value object for adapter configurations between the Management and Tenant APIs

kaniyan:
In this PR, `TenantObject` is refactored to use `Adapter` as a data model instead of `List<Map<String, Object>>` to store adapter configurations. This also enables to share the same value object for Adapter configurations between the Management API and Tenant API.

In this PR,
- The `Adapter` class has been moved to the `core` module from the `service-base`.
- The below new methods are added to directly deal with objects of type `Adapter`.
```
List<Adapter> getAdapters();
TenantObject setAdapters(final List<Adapter> adapters) ;
Adapter getAdapter(final String type);
```
- The below methods are deprecated in favor of the above.

```
List<Map<String, Object>> getAdapterConfigurationsAsMaps()
JsonArray getAdapterConfigurations()
JsonObject getAdapterConfiguration(final String type)
TenantObject setAdapterConfigurations(final List<Map<String, Object>> configurations)
TenantObject setAdapterConfigurations(final JsonArray configurations)
```

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/TenantObject.java
+++ b/core/src/main/java/org/eclipse/hono/util/TenantObject.java
@@ -48,7 +48,7 @@ import io.vertx.core.json.JsonObject;
 public final class TenantObject extends JsonBackedValueObject {
 
     @JsonIgnore
-    private List<Map<String, Object>> adapterConfigurations;
+    private List<Adapter> adapters;
     @JsonIgnore
     private ResourceLimits resourceLimits;
     @JsonIgnore
@@ -251,20 +251,78 @@ public final class TenantObject extends JsonBackedValueObject {
     }
 
     /**
+     * Gets the list of configured adapters for the tenant.
+     *
+     * @return The list of configured adapters or {@code null} if not set.
+     */
+    @JsonIgnore
+    public List<Adapter> getAdapters() {
+        return adapters;
+    }
+
+    /**
+     * Sets the given list of adapters for the tenant.
+     *
+     * @param adapters The list of adapters or {@code null} if not set.
+     * @return This tenant for command chaining.
+     * @throws IllegalArgumentException if more than one of the adapters have the same <em>type</em>.
+     */
+    @JsonIgnore
+    public TenantObject setAdapters(final List<Adapter> adapters) {
+        this.adapters = validateAdapterTypes(adapters);
+        return this;
+    }
+
+    @JsonIgnore
+    private List<Adapter> validateAdapterTypes(final List<Adapter> adapters) {
+        if (adapters != null) {
+            final Set<String> uniqueAdapterTypes = adapters.stream()
+                    .map(Adapter::getType)
+                    .collect(Collectors.toSet());
+            if (adapters.size() != uniqueAdapterTypes.size()) {
+                throw new IllegalArgumentException("Each adapter must have a unique type");
+            }
+        }
+        return adapters;
+    }
+
+    /**
+     * Gets the adapter configuration for the given type.
+     *
+     * @param type The adapter's type.
+     * @return The adapter configuration or {@code null} if not set.
+     * @throws NullPointerException if type is {@code null}.
+     */
+    @JsonIgnore
+    public Adapter getAdapter(final String type) {
+
+        Objects.requireNonNull(type);
+
+        return Optional.ofNullable(adapters)
+                .flatMap(ok -> adapters.stream()
+                        .filter(adapter -> type.equals(adapter.getType()))
+                        .findFirst())
+                .orElse(null);
+    }
+
+    /**
      * Gets the configuration information for this tenant's
      * configured adapters.
      * 
      * @return An unmodifiable list of configuration properties or
      *         {@code null} if no specific configuration has been
      *         set for any protocol adapter.
+     * @deprecated Use {@link #getAdapters()} instead.
      */
+    @Deprecated
     @JsonProperty(TenantConstants.FIELD_ADAPTERS)
     public List<Map<String, Object>> getAdapterConfigurationsAsMaps() {
-        if (adapterConfigurations == null) {
-            return null;
-        } else {
-            return adapterConfigurations;
-        }
+        return Optional.ofNullable(adapters)
+                .map(adapters -> adapters.stream()
+                        .map(JsonObject::mapFrom)
+                        .map(JsonObject::getMap)
+                        .collect(Collectors.toList()))
+                .orElse(null);
     }
 
     /**
@@ -274,16 +332,17 @@ public final class TenantObject extends JsonBackedValueObject {
      * @return The configuration properties for this tenant's
      *         configured adapters or {@code null} if no specific
      *         configuration has been set for any protocol adapter.
+     * @deprecated Use {@link #getAdapters()} instead.
      */
+    @Deprecated
     @JsonIgnore
     public JsonArray getAdapterConfigurations() {
-        if (adapterConfigurations == null) {
-            return null;
-        } else {
-            final JsonArray result = new JsonArray();
-            adapterConfigurations.stream().forEach(config -> result.add(new JsonObject(config)));
-            return result;
-        }
+        return Optional.ofNullable(adapters)
+                .map(adapters -> adapters.stream()
+                        .map(JsonObject::mapFrom)
+                        .collect(Collectors.toList()))
+                .map(JsonArray::new)
+                .orElse(null);
     }
 
     /**
@@ -293,19 +352,16 @@ public final class TenantObject extends JsonBackedValueObject {
      * @return The configuration properties or {@code null} if no specific
      *         properties have been set.
      * @throws NullPointerException if type is {@code null}.
+     * @deprecated Use {@link #getAdapter(String type)} instead.
      */
+    @Deprecated
     public JsonObject getAdapterConfiguration(final String type) {
 
         Objects.requireNonNull(type);
-        if (adapterConfigurations == null) {
-            return null;
-        } else {
-            return adapterConfigurations.stream()
-                    .filter(entry -> type.equalsIgnoreCase((String) entry.get(TenantConstants.FIELD_ADAPTERS_TYPE)))
-                    .findFirst()
-                    .map(entry -> new JsonObject(entry))
-                    .orElse(null);
-        }
+
+        return Optional.ofNullable(getAdapter(type))
+                .map(JsonObject::mapFrom)
+                .orElse(null);
     }
 
     /**
@@ -317,14 +373,21 @@ public final class TenantObject extends JsonBackedValueObject {
      *                              copied into a new list in order to prevent modification
      *                              of the list after this method has been invoked.
      * @return This tenant for command chaining.
+     * @throws IllegalArgumentException if the given configurations does not contain a <em>type</em> or
+     *                                  if more than one of the adapter configurations have the same <em>type</em> or
+     *                                  if error parsing the given configuration.
+     * @deprecated Use {@link #setAdapters(List adapters)} instead.
      */
+    @Deprecated
     @JsonProperty(TenantConstants.FIELD_ADAPTERS)
     public TenantObject setAdapterConfigurations(final List<Map<String, Object>> configurations) {
-        if (configurations == null) {
-            this.adapterConfigurations = null;
-        } else {
-            this.adapterConfigurations = new LinkedList<>(configurations);
-        }
+        final List<Adapter> adapters = Optional.ofNullable(configurations)
+                .map(ok -> configurations.stream()
+                        .map(JsonObject::mapFrom)
+                        .map(config -> config.mapTo(Adapter.class))
+                        .collect(Collectors.toList()))
+                .orElse(null);
+        setAdapters(adapters);
         return this;
     }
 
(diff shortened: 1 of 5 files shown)


sophokles73:
Moving the `Adapter` class to the core module constitutes an API breaking change FMPOV. We could do that in Hono 2.0.0. Do you think that there is some value in adding a *copy* of the `Adapter` class to core, but maybe naming `ProtocolAdapterConfig`, so that we can at least take advantage of it in the `TenantObject`? We could then remove the original `Adapter` class in Hono 2.0.0 in favor of the `ProtocolAdapterConfig` class.
WDYT? @dejanb @ctron

kaniyan:
@sophokles73 FMPOV we can have a copy of the `Adapter` under the _core_ module and use this copy in the `TenantObject`. And _deprecate_ the other one in the _service-base_ module. So that the users are informed about the deprecation and will have enough time to switch to the other one. We could remove the original `Adapter` in _service-base_ and it's usages in 2.0.0. FMPOV the name `Adapter` suits well following the same naming convention as other attributes such as as `ResourceLimits`, `Defaults` and  `Extensions`. WDYT?

sophokles73:
@kaniyan any plans to continue work on this one?

kaniyan:
> @kaniyan any plans to continue work on this one?

Based on the last discussion, 
- a copy of the `Adapter` has been added to the _core_. The Tenant API (`TenantObject`) now offers more methods that make use of this new Adapter class and some existing methods have been deprecated.
- The `org.eclipse.hono.service.management.tenant.Adapter` has been marked as deprecated. The info that will be removed in Hono 2.0.0 is documented as part of the javadoc.
- The methods dealing with adapter configurations in Tenant Mangement API (`Tenant`) that use `org.eclipse.hono.service.management.tenant.Adapter` have been marked as deprecated. Additionally it is mentioned in the javadoc that from Hono 2.0.0  `org.eclipse.hono.service.management.tenant.Adapter` will be used by those methods.

Would you mind taking a look?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -293,16 +339,17 @@ public TenantObject setAdapterConfigurations(final List<Map<String, Object>> con
      *                       configured adapters or {@code null} in order to
      *                       remove any existing configuration.
      * @return This tenant for command chaining.
+     * @deprecated Use {@link #setAdapters(List adapters)} instead.
```
sophokles73: IMHO this method now also may throw an `IllegalArgumentException` ...

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -311,24 +358,32 @@ public TenantObject setAdapterConfigurations(final JsonArray configurations) {
...
+                    throw new IllegalArgumentException(
+                            String.format("Already an adapter of the type [%s] exists", adapter.getType()));
+                }
             }
-            adapterConfigurations.add(config.getMap());
-        } else {
-            throw new IllegalArgumentException("adapter configuration must contain type field");
+            adapters.add(adapter);
+        } catch (final IllegalArgumentException e) {
+            throw new IllegalArgumentException("Error adding the adapter configuration", e);
```
sophokles73: does this really have any value?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -207,21 +207,65 @@ private TrustAnchor getTrustAnchorForPublicKey(final JsonObject keyProps) {
...
      */
+    @Deprecated
     @JsonProperty(TenantConstants.FIELD_ADAPTERS)
     public List<Map<String, Object>> getAdapterConfigurationsAsMaps() {
-        if (adapterConfigurations == null) {
-            return null;
-        } else {
-            return adapterConfigurations;
-        }
+        return Optional.ofNullable(adapters)
```
sophokles73: IMHO the `@JsonProperty` annotation should be moved to either the *getAdapters* method or the *adapters* field ...

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -274,13 +316,17 @@ public JsonObject getAdapterConfiguration(final String type) {
      *                              copied into a new list in order to prevent modification
      *                              of the list after this method has been invoked.
      * @return This tenant for command chaining.
+     * @deprecated Use {@link #setAdapters(List adapters)} instead.
      */
+    @Deprecated
```
sophokles73: IMHO the `@JsonProperty` annotation should be moved to either the *setAdapters* method or the *adapters* field ...

core/src/main/java/org/eclipse/hono/util/TenantObject.java
kaniyan: At first I did add Json annotation to the _setAdapters_ and _getAdapters_ methods and then I reverted back. As I was not sure if it is allowed now or only changes as part of release 2.0.0.

sophokles73: Hmm, the only thing that might break is a custom annotation processor that someone has built for this annotation in this class.

@dejanb @ctron any thoughts on this?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
kaniyan: `config.mapTo(Adapter.class)` in this method also throws an `IllegalArgumentException` when the object is not mappable. I wanted to give more info by saying _"Error adding the adapter configuration"_ plus the reason given by `JsonObject.mapto(...)`. I think in this case, the name of this method `addAdapterConfiguration(...)` is already self explanatory. I will remove this part.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
kaniyan: I have added info about this `IllegalArgumentException` to the javadoc of this method.

sophokles73:
@kaniyan did you notice that the build fails?

kaniyan:
> @kaniyan did you notice that the build fails?
Thanks for the info. I took a look and fixed it.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -354,25 +404,29 @@ public TenantObject setAdapterConfigurations(final JsonArray configurations) {
...
-                adapterConfigurations = new LinkedList<>();
-            }
-            adapterConfigurations.add(config.getMap());
+        final Adapter adapter = config.mapTo(Adapter.class);
+
+        if (adapters == null) {
+            adapters = new LinkedList<>();
         } else {
-            throw new IllegalArgumentException("adapter configuration must contain type field");
+            final boolean hasAdapterOfSameType = adapters.stream()
```
sophokles73: isn't there a method for this already, e.g. *getAdapter(String)*?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -317,13 +359,20 @@ public JsonObject getAdapterConfiguration(final String type) {
...
     @JsonProperty(TenantConstants.FIELD_ADAPTERS)
     public TenantObject setAdapterConfigurations(final List<Map<String, Object>> configurations) {
         if (configurations == null) {
-            this.adapterConfigurations = null;
+            this.adapters = null;
         } else {
-            this.adapterConfigurations = new LinkedList<>(configurations);
+            configurations.stream()
+                    .map(JsonObject::mapFrom)
+                    .forEach(this::addAdapterConfiguration);
```
sophokles73: this does not replace the existing adapter configs with the given ones, does it?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
kaniyan: Yes, i have replaced it now with _getAdapter(String)_.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
kaniyan: Thanks for pointing out. This has been fixed now.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -250,21 +250,65 @@ public boolean isAutoProvisioningEnabled(final String subjectDn) {
...
+    public TenantObject setAdapters(final List<Adapter> adapters) {
+        this.adapters = adapters;
+        return this;
+    }
+
+    /**
+     * Gets the adapter configuration for the given type.
+     *
+     * @param type The adapter's type.
+     * @return The adapter configuration or {@code null} if not set.
```
sophokles73: `@throws NullpointerException ...`

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -293,19 +339,15 @@ public JsonArray getAdapterConfigurations() {
      * @return The configuration properties or {@code null} if no specific
      *         properties have been set.
      * @throws NullPointerException if type is {@code null}.
+     * @throws IllegalArgumentException if the conversion of the adapter object fails.
```
sophokles73: when/why would that happen? We have explicitly taken care that the `Adapter` class can be mapped to JSON, haven't we?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -336,43 +384,57 @@ public TenantObject setAdapterConfigurations(final List<Map<String, Object>> con
...
-        } else {
-            this.adapterConfigurations = new LinkedList<>();
-            configurations.stream().filter(obj -> JsonObject.class.isInstance(obj)).forEach(config -> {
-                addAdapterConfiguration((JsonObject) config);
-            });
+        this.adapters = null;
+        if (configurations != null) {
+            configurations.stream()
+                    .filter(JsonObject.class::isInstance)
+                    .forEach(config -> addAdapterConfiguration((JsonObject) config));
```
sophokles73: this means that we end up with a *half-populated* adapters list if one of the configurations from the JsonArray doesn't contain a *type*, right?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
kaniyan: Added.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
kaniyan: It should not happen as we already take care with the `Adapter` class. I will remove this from the _javadoc_.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
kaniyan: yes and to resolve this now I have added a validation to check if the given adapter configurations are valid (i.e being able to map to `Adapter.class` and ensure if the _adapter type_ in the list is unique). Only on successful validation, the given configurations are added.

kaniyan: Also new tests have been added to ensure the above.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -250,21 +250,79 @@ public boolean isAutoProvisioningEnabled(final String subjectDn) {
...
+    private List<Adapter> validateAdapterTypes(final List<Adapter> adapters) {
+        if (adapters != null) {
+            final Set<String> uniqueAdapterTypes = adapters.stream()
+                    .map(Adapter::getType)
+                    .collect(Collectors.toSet());
+            if (adapters.size() != uniqueAdapterTypes.size()) {
+                throw new IllegalArgumentException("Each adapter must have a unique type");
+            }
+        }
+        return adapters;
```
sophokles73: :+1:

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -336,43 +399,63 @@ public TenantObject setAdapterConfigurations(final List<Map<String, Object>> con
...
-        if (configurations == null) {
-            this.adapterConfigurations = null;
-        } else {
-            this.adapterConfigurations = new LinkedList<>();
-            configurations.stream().filter(obj -> JsonObject.class.isInstance(obj)).forEach(config -> {
-                addAdapterConfiguration((JsonObject) config);
-            });
+        final List<Adapter> adapters = Optional.ofNullable(configurations)
+                .map(ok -> configurations.stream()
+                        .map(JsonObject::mapFrom)
```
sophokles73: IMHO at this place we first need to filter for JsonObject instances in order to get the same behavior as in the original implementation, right?

kaniyan: I have removed it as it doesn't makes sense to filter out only instances of the `JsonObject` but in the other method `setAdapterConfigurations(final List<Map<String, Object>> configurations)` we don't do it. I agree with you that the behavior should remain the same as in the original implementation. I have again added this filter for _JsonObject_.

Issue #1509, opened 2019-09-19
How to re-connect to AMQP adapter after network failure to continue to receive commands?


selobosi:
We (finally) use Hono 1.0-M7 and during development we switch networks every once in a while or turn them off and on again. We have an app that runs on a device and connects to the Hono AMQP adapter. We use command and control and the app receives commands via Hono's AMQP adapter.

The app reconnects to the AMQP adapter when the network is gone. We see in the logs that when the app reconnects, the logs of Hono's AMQP adapter contain the message `cannot create concurrent command consumer` and subsequently, the device will not receive commands meant for it. The message `cannot create concurrent command consumer` does not appear in the logs of Hono's AMQP adapter after the app has been shut down and restarted. I attached an [annotated log file](https://github.com/eclipse/hono/files/3632404/Annotated-AMQP-Adapter-Log-masked.txt) - searching for `consumer` reveals relevant parts.

Have others experienced such behaviour? How could the device address the issue?

sophokles73:
This sounds like the AMQP adapter does not tidy up a command consumer for a device if the connection to the device is lost unexpectedly instead of being closed gracefully.
@calohmn WDYT?

selobosi:
Just to update the relevance of this issue: We are now also affected by this issue when demonstrating our work to stakeholders. E.g. when we turn off the device without shutting down the app, this has the same effect as switching networks: when connecting to Hono again, the device won't receive commands anymore. We can still work around the issue, but the work-around causes overhead and is prone to error.

sophokles73:
@calohmn any opinion on this?

calohmn:
@selobosi Looking at the log file, the `cannot create concurrent command consumer` error comes *before* the AMQP adapter takes notice of the connection not being active anymore (`lost connection to device`). The period between the `STOPPING NETWORK` mark and the time that the adapter notices that is more than a minute. 
Have you tried setting the idle timeout when opening the connection to the AMQP adapter to a value lower than that?

The idleTimeout check that the AMQP adapter is doing on the connection takes both the idleTimeout advertised by the client as well as the local idleTimeout of the AMQP server (currently not set) into account. I think we should make the idleTimeout / heartbeat of the AMQP server of the AMQP adapter configurable to better handle situations like the above. 
Until then, you could try setting the idle timeout in your client.
As long as the idleTimeout check notices the inactive connection _before_ the device opens a new command consumer, just from looking at the code and the log output above, I think the above error shouldn't occur.

[...]

PR #1603, opened 2019-10-28
[#1509] Use AMQP idle timeout to detect stale connections.

sophokles73:
The AMQP adapter now uses the idle timeout mechanism to detect and close
connections to devices. The default value of 60 seconds can be
overridden using an environment variable.


sophokles73:
@calohmn I have pushed a fix.

Issue #1585, opened 2019-10-17
Document concept behind C&C handling for gateways


calohmn:
The concept behind how commands are handled in scenarios involving gateways should be documented. This involves the AMQP message flow as well as the implementation regarding the different command consumer classes.

Related issues: #1523, #930.
To be kept in mind: the concept might change again for Hono 1.1.0 - see #1272.

sophokles73:
@calohmn are you planning to add this to 1.0.0? Or will this be added after the release?

calohmn:
@sophokles73 after the 1.0.0 release.

sophokles73:
@calohmn is this done, i.e. can we close this?

calohmn:
yes, closing it.

PR #1613, opened 2019-11-05
[#1585] Adapt command & control concept page

calohmn:
This is for #1585.

The command & control concepts page has been rewritten to focus on the way command & control is handled by the protocol adapters in general, also describing the special support for commands directed at gateways.

The contents of the old concepts page shall get re-used in a subsequent commit, possibly integrating the respective chapters in the protocol adapter user guides.


calohmn:
Also adopted to `1.0.0-docfix` branch.

PR #1614, opened 2019-11-06
Migration to vert.x 3.8.3 step1

sophokles73:
This is the first step in a series of PRs which will migrate the existing code base to vert.x 3.8.3.
The focus is on removing all usage of API that has been deprecated in vert.x 3.8.0 around Futures.
This PR lays the foundation by focusing on the core and client modules, subsequent PRs will focus on services and adapters.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/HonoConnection.java
+++ b/client/src/main/java/org/eclipse/hono/client/HonoConnection.java
@@ -21,6 +21,7 @@ import io.opentracing.Tracer;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
+import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
 import io.vertx.proton.ProtonClientOptions;
 import io.vertx.proton.ProtonLink;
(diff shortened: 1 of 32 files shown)


client/src/main/java/org/eclipse/hono/client/impl/CommandConsumerFactoryImpl.java
```diff
@@ -233,66 +234,118 @@ private String getDeviceKey(final String tenantId, final String deviceId) {
...
+                                        gatewayOrDeviceId,
+                                        () -> destinationCommandConsumer.getCommandHandlers(),
+                                        checkInterval);
+                            }
+                            return deviceSpecificConsumerFuture.result();
+                        });
+            })
+            .setHandler(result);
+
+//            if (!result.isComplete()) {
```
dejanb: Remove the old code?

client/src/main/java/org/eclipse/hono/client/HonoConnection.java
```diff
@@ -240,9 +240,25 @@ static HonoConnection newConnection(final Vertx vertx, final ClientConfigPropert
...
+     * 
+     * @param <T> The type of the result that the code produces.
+     * @param codeToRun The code to execute. The code is required to either complete or
+     *                  fail the promise that is passed into the handler.
+     * @return The future containing the result of the promise passed in to the handler
+     *         for executing the code. The future thus indicates the outcome of executing
+     *         the code. The future will always be failed with a {@link ServerErrorException}
+     *         if this connection's <em>context</em> property is {@code null}.
+     */
+    <T> Future<T> executeOnContext(Handler<Promise<T>> codeToRun);
```
dejanb: Wondering if we can get away without this and just use HonoProtonHelper method directly?

sophokles73: This method is mainly here in order to encapsulate the vert.x Context used by the connection. If we remove this method then client code needs to extract the connection's context from the connection before invoking HonoProtonHelper.executeOnContext().

sophokles73:
@dejanb I have removed the obsolete code ...

Issue #1462, opened 2019-08-27
Upgrade to vertx 3.8.x


kaniyan:
Upgrade to vertx 3.8.x which already use netty 4.1.39.Final and  proton-j 0.33.2.
For more info refer[ release notes](https://github.com/vert-x3/wiki/wiki/3.8.1-Release-Notes).

calohmn:
See also [3.8.0 Deprecations and breaking changes](https://github.com/vert-x3/wiki/wiki/3.8.0-Deprecations-and-breaking-changes).

sophokles73:
we have (manually) upgraded to Netty 4.1.39.Final already (#1450). I think we should defer this upgrade to post 1.0.0 as it will also require several new CQs ...

sophokles73:
closed via #1634

PR #1619, opened 2019-11-08
[#1462] Fix usage of deprecated API in service-base module

sophokles73:
This step of migration to vert.x 3.8.3 tries to remove usage of deprecated API from the service-base module.
Not that in several cases I was not able to simply change a method's signature to accept a `Promise` instead of a `Future` because we forgot to declare the method (or class) `final` :-( Otherwise we would break the API if somebody had extended the class and overridden the method ...

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/AbstractApplication.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/AbstractApplication.java
@@ -24,6 +24,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 
 import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
+import io.vertx.core.Promise;
 
 /**
  * A base class for implementing Spring Boot applications.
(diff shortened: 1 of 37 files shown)


sophokles73:
@ctron any thought regarding how we should handle the non-final methods?

ctron:
Taking a look at the OSGi semantic versioning whitepaper:

> A package can contain an API. There are two types of clients for these API packages: API consumers and API implementation providers. A change in the second (minor) part of the version signals that the change is backward compatible with consumers of the API package but not with the providers of that API. That is, when the API package goes from version 1.5 to 1.6 it is no longer compatible with a provider of that API but consumers of that API are backward compatible with that API package.

So I would say, create new methods. Ensure that the API is compatible for consumers of the API, and name it `1.1.0`.

dejanb:
> So I would say, create new methods. Ensure that the API is compatible for consumers of the API, and name it `1.1.0`.

This sounds reasonable approach to me

Issue #1440, opened 2019-08-12
Support tracing of resource limit checks


kaniyan:
The message limits check and the concurrent connection limit check should be traced using OpenTracing.

sophokles73:
closed via #1622

PR #1622, opened 2019-11-13
[#1440] Track resource limit checks using OpenTracing.

kaniyan:
This is a PR for  #1440. 

_Summary of changes:_
* The _ResourceLimitChecks_ interface has been extended here with methods to include an argument of type `SpanContext` to support tracing and the old methods are deprecated.
* Now in the default Prometheus based _resource-limits_ implementation several events are being traced.
* The existing methods `checkConnectionLimit(...)` and `checkMessageLimit(..)` in the base adapter class are deprecated and new methods that take an extra argument of type `SpanContext`are added.
* The protocol adapters now support the tracing of resource limit checks by making use of the above methods with an extra `SpanContext` argument.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
@@ -23,18 +23,25 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Supplier;
 
+import io.opentracing.noop.NoopTracerFactory;
 import org.eclipse.hono.cache.CacheProvider;
 import org.eclipse.hono.cache.ExpiringValueCache;
 import org.eclipse.hono.client.ServiceInvocationException;
 import org.eclipse.hono.service.metric.MetricsTags;
 import org.eclipse.hono.service.metric.MicrometerBasedMetrics;
+import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.DataVolume;
 import org.eclipse.hono.util.DataVolumePeriod;
+import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.TenantConstants;
 import org.eclipse.hono.util.TenantObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import io.opentracing.Span;
+import io.opentracing.SpanContext;
+import io.opentracing.Tracer;
+import io.opentracing.tag.Tags;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.core.json.JsonArray;
@@ -59,6 +66,7 @@ public final class PrometheusBasedResourceLimitChecks implements ResourceLimitCh
     private static final Logger log = LoggerFactory.getLogger(PrometheusBasedResourceLimitChecks.class);
     private static final String QUERY_URI = "/api/v1/query";
     private static final String LIMITS_CACHE_NAME = "resource-limits";
+    private final Tracer tracer;
 
     /**
      * The mode of the data volume calculation.
@@ -126,17 +134,45 @@ public final class PrometheusBasedResourceLimitChecks implements ResourceLimitCh
      */
     public PrometheusBasedResourceLimitChecks(final WebClient webClient,
             final PrometheusBasedResourceLimitChecksConfig config, final CacheProvider cacheProvider) {
+        this(webClient, config, cacheProvider, NoopTracerFactory.create());
+    }
+
+    /**
+     * Creates new checks.
+     *
+     * @param webClient The client to use for querying the Prometheus server.
+     * @param config The PrometheusBasedResourceLimitChecks configuration object.
+     * @param cacheProvider The cache provider to use for creating caches for metrics data retrieved
+     *                      from the prometheus backend or {@code null} if they should not be cached.
+     * @param tracer The tracer instance.
+     * @throws NullPointerException if any of the parameters except cacheProvider are {@code null}.
+     */
+    public PrometheusBasedResourceLimitChecks(final WebClient webClient,
+            final PrometheusBasedResourceLimitChecksConfig config, final CacheProvider cacheProvider, final Tracer tracer) {
         this.client = Objects.requireNonNull(webClient);
         this.config = Objects.requireNonNull(config);
         this.limitsCache = Optional.ofNullable(cacheProvider)
                 .map(provider -> provider.getCache(LIMITS_CACHE_NAME))
                 .orElse(null);
+        this.tracer = Objects.requireNonNull(tracer);
     }
 
     @Override
     public Future<Boolean> isConnectionLimitReached(final TenantObject tenant) {
+        return isConnectionLimitReached(tenant, null);
+    }
+
+    @Override
+    public Future<Boolean> isConnectionLimitReached(final TenantObject tenant, final SpanContext spanContext) {
 
         Objects.requireNonNull(tenant);
+
+        final Span span = TracingHelper.buildChildSpan(tracer, spanContext, "verify connection limit")
+                .ignoreActiveSpan()
+                .withTag(Tags.COMPONENT.getKey(), getClass().getSimpleName())
+                .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_CLIENT)
+                .withTag(MessageHelper.APP_PROPERTY_TENANT_ID, tenant.getTenantId())
+                .start();
         if (tenant.getResourceLimits() == null) {
             log.trace("No connection limits configured for the tenant [{}]", tenant.getTenantId());
             return Future.succeededFuture(Boolean.FALSE);
@@ -151,18 +187,35 @@ public final class PrometheusBasedResourceLimitChecks implements ResourceLimitCh
         } else {
             final String queryParams = String.format("sum(%s{tenant=\"%s\"})", CONNECTIONS_METRIC_NAME,
                     tenant.getTenantId());
-            return executeQuery(queryParams)
+            return executeQuery(queryParams, span)
                     .map(currentConnections -> {
                         if (currentConnections < maxConnections) {
+                            final String logMessage = String.format(
+                                    "Connection limit not exceeded [tenant: %s, current connections: %s, max-connections: %s]",
+                                    tenant.getTenantId(), currentConnections, maxConnections);
+                            span.log(logMessage);
+                            log.trace(logMessage);
                             return Boolean.FALSE;
                         } else {
-                            log.trace(
-                                    "connection limit exceeded [tenant: {}, current connections: {}, max-connections: {}]",
+                            final String logMessage = String.format(
+                                    "Connection limit exceeded [tenant: %s, current connections: %s, max-connections: %s]",
                                     tenant.getTenantId(), currentConnections, maxConnections);
+                            TracingHelper.logError(span, logMessage);
+                            log.trace(logMessage);
                             return Boolean.TRUE;
                         }
-                    }).otherwise(Boolean.FALSE);
+                    })
+                    .otherwise(failure -> Boolean.FALSE)
+                    .map(result -> {
+                        span.finish();
+                        return result;
+                    });
         }
+    }
+
+    @Override
+    public Future<Boolean> isMessageLimitReached(final TenantObject tenant, final long payloadSize) {
+        return isMessageLimitReached(tenant, payloadSize, null);
     }
 
     /**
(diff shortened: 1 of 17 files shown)


service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -250,12 +273,18 @@ public PrometheusBasedResourceLimitChecks(final WebClient webClient,
...
                     .map(cachedValue -> Future.succeededFuture((long) cachedValue))
-                    .orElseGet(() -> executeQuery(queryParams)
+                    .orElseGet(() -> executeQuery(queryParams, span)
                             .map(bytesConsumed -> addToCache(limitsCache, key, bytesConsumed)))
                     .map(bytesConsumed -> {
                         if ((bytesConsumed + payloadSize) <= allowedMaxBytes) {
                             return Boolean.FALSE;
                         } else {
+                            TracingHelper.logError(span, String.format(
+                                    "Data limit exceeded for the tenant [%s]. [consumed bytes:{}, allowed max-bytes:%s, %s:%s, %s:%s, %s:%s]",
```
calohmn: Wrong placeholder used here: `consumed bytes:{}`

kaniyan: Done.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -280,9 +309,11 @@ public PrometheusBasedResourceLimitChecks(final WebClient webClient,
         .send(sendAttempt -> {
             if (sendAttempt.succeeded()) {
                 final HttpResponse<JsonObject> response = sendAttempt.result();
-                result.complete(extractLongValue(response.body()));
+                result.complete(extractLongValue(response.body(), span));
             } else {
-                log.debug("error fetching result from Prometheus: {}", sendAttempt.cause().getMessage());
+                TracingHelper.logError(span,
+                        String.format("Error fetching result from Prometheus: %s", sendAttempt.cause().getMessage()));
+                log.debug("Error fetching result from Prometheus: {}", sendAttempt.cause().getMessage());
```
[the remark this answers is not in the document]
kaniyan: @calohmn Done. would you mind taking a look again?

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/ResourceLimitChecks.java
```diff
@@ -31,9 +32,26 @@
...
+     * @param tenantObject The tenant configuration to check the limit against.
+     * @param span The active OpenTracing span for this operation. It is not to be closed in this method!
+     *             An implementation should log (error) events on this span, and it may set tags and use 
+     *             this span as the parent for any spans created in this method.
+     * @return A future indicating the outcome of the check.
+     *         <p>
+     *         The future will be failed with a {@link ServiceInvocationException}
+     *         if the check could not be performed.
+     */
+    Future<Boolean> isConnectionLimitReached(TenantObject tenantObject, Span span);
```
sophokles73: FMPOV we should pass in a `SpanContext` instead of a `Span`. It should be up to the implementation whether it wants to trace the execution of the check using a new (child) span or not.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/ResourceLimitChecks.java
```diff
@@ -44,6 +62,24 @@
...
+     * @param payloadSize The message payload size in bytes.
+     * @param span The active OpenTracing span for this operation. It is not to be closed in this method!
+     *             An implementation should log (error) events on this span, and it may set tags and use 
+     *             this span as the parent for any spans created in this method.
+     * @return A future indicating the outcome of the check.
+     *         <p>
+     *         The future will be failed with a {@link ServiceInvocationException}
+     *         if the check could not be performed.
+     */
+    Future<Boolean> isMessageLimitReached(TenantObject tenantObject, long payloadSize, Span span);
```
sophokles73: same here

sophokles73:
@kaniyan any news on this?

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/ResourceLimitChecks.java
kaniyan: I created a child span in the abstract base class and passed it to the resource limit checks. I intended to make it as a child span as it looks more elegant and easier to track the flow in this case. In Hono currently I see two approaches. One is passing the `SpanContext`and the other is passing the `Span` as in the `TenantManagementService`. I chose the `Span` in the above implementation. IMHO passing the `Span` shouldn't prevent the implementation to create any new child spans if needed, as `Span.context()` returns the `SpanContext`.

sophokles73: That is true but it also allows the method to finish the span. FMPOV it is more about the question who is in control of and responsible for the life-cycle of a span. In this case, my feeling is that the responsibility should lie with the `ResourceLimitChecks` implementation. If the `AbstractProtocolAdapterBase` class creates a Span, then it should either be in control of the span's life-cycle itself (meaning that it only passes a SpanContext to the invoked methods of `ResourceLimitChecks`) or the `ResourceLimitChecks` interface JavaDoc needs to be very specific about the expectation, how implementations should handle the passed in Span. Personally, I believe that passing in a SpanContext only is much safer and provides for better encapsulation of the Span life-cycle.

kaniyan: > In this case, my feeling is that the responsibility should lie with the `ResourceLimitChecks` implementation.

I think so too that the responsibility should lie with the `ResourceLimitChecks` and I will pass the `SpanContext` instead of `Span`.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/ResourceLimitChecks.java
```diff
@@ -31,9 +32,26 @@
...
+     * @param tenantObject The tenant configuration to check the limit against.
+     * @param spanContext The currently active OpenTracing span context that is used to
+     *                    trace the limits verification or {@code null}
+     *                    if no span is currently active.
+     * @return A future indicating the outcome of the check.
+     *         <p>
+     *         The future will be failed with a {@link ServiceInvocationException}
+     *         if the check could not be performed.
+     */
+    Future<Boolean> isConnectionLimitReached(TenantObject tenantObject, SpanContext spanContext);
```
sophokles73: what if tenant or span context are `null`?

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/ResourceLimitChecks.java
```diff
@@ -44,6 +62,24 @@
...
+     * @param payloadSize The message payload size in bytes.
+     * @param spanContext The currently active OpenTracing span context that is used to
+     *                    trace the limits verification or {@code null}
+     *                    if no span is currently active.
+     * @return A future indicating the outcome of the check.
+     *         <p>
+     *         The future will be failed with a {@link ServiceInvocationException}
+     *         if the check could not be performed.
+     */
+    Future<Boolean> isMessageLimitReached(TenantObject tenantObject, long payloadSize, SpanContext spanContext);
```
sophokles73: what if tenant or span context are `null`?

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/ResourceLimitChecks.java
kaniyan: javadoc has been appended with ` @throws NullPointerException if the tenant object is null.`

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/ResourceLimitChecks.java
kaniyan: javadoc has been appended with ` @throws NullPointerException if the tenant object is null.`

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -122,21 +129,35 @@ String value() {
      * @param config The PrometheusBasedResourceLimitChecks configuration object.
      * @param cacheProvider The cache provider to use for creating caches for metrics data retrieved
      *                      from the prometheus backend or {@code null} if they should not be cached.
+     * @param tracer The tracer instance.
      * @throws NullPointerException if any of the parameters except cacheProvider are {@code null}.
      */
     public PrometheusBasedResourceLimitChecks(final WebClient webClient,
-            final PrometheusBasedResourceLimitChecksConfig config, final CacheProvider cacheProvider) {
+            final PrometheusBasedResourceLimitChecksConfig config, final CacheProvider cacheProvider, final Tracer tracer) {
```
sophokles73: this is a non backward compatible change to public API. IMHO we need to overload the constructor instead ...

kaniyan: yes, I have added an overloaded constructor and pushed the changes.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -123,20 +131,56 @@ String value() {
      * @param cacheProvider The cache provider to use for creating caches for metrics data retrieved
      *                      from the prometheus backend or {@code null} if they should not be cached.
      * @throws NullPointerException if any of the parameters except cacheProvider are {@code null}.
+     * @deprecated Use {@link #PrometheusBasedResourceLimitChecks(WebClient, PrometheusBasedResourceLimitChecksConfig, 
+     *                  CacheProvider, Tracer)} instead.
      */
+    @Deprecated
```
sophokles73: I do not really see why this constructor should be deprecated. We do not have any problem if anybody still uses it and doesn't have any need to trace execution, do we?
Also, I think the implementation could be
```
this(webClient, config, cacheProvider, NoopTracerFactory.create());
```
instead of duplicating the code. WDYT?

kaniyan: Sounds good. I have pushed the changes, pls have a look.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -151,20 +187,33 @@ public PrometheusBasedResourceLimitChecks(final WebClient webClient,
         } else {
             final String queryParams = String.format("sum(%s{tenant=\"%s\"})", CONNECTIONS_METRIC_NAME,
                     tenant.getTenantId());
-            return executeQuery(queryParams)
+            return executeQuery(queryParams, span)
                     .map(currentConnections -> {
                         if (currentConnections < maxConnections) {
+                            span.finish();
```
sophokles73: how about logging the number of current connections vs the max connections here as well?

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -151,20 +187,33 @@ public PrometheusBasedResourceLimitChecks(final WebClient webClient,
...
                                     "connection limit exceeded [tenant: {}, current connections: {}, max-connections: {}]",
                                     tenant.getTenantId(), currentConnections, maxConnections);
+                            span.finish();
                             return Boolean.TRUE;
                         }
-                    }).otherwise(Boolean.FALSE);
+                    }).otherwise(failure -> {
+                        span.finish();
+                        return Boolean.FALSE;
+                    });
```
sophokles73: IMHO you can simply finish the span here:
```
.otherwise(failure -> {
  return Boolean.FALSE;
})
.map(result -> {
  span.finish();
  return result;
});
```
instead of finishing the span in three places ...

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -251,25 +314,36 @@ public PrometheusBasedResourceLimitChecks(final WebClient webClient,
...
                                     TenantConstants.FIELD_PERIOD_MODE, periodMode,
                                     TenantConstants.FIELD_PERIOD_NO_OF_DAYS, periodInDays);
+                            span.finish();
                             return Boolean.TRUE;
                         }
-                    }).otherwise(Boolean.FALSE);
+                    }).otherwise(failed -> {
+                        span.finish();
+                        return Boolean.FALSE;
+                    });
```
sophokles73: see above

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -251,25 +314,36 @@ public PrometheusBasedResourceLimitChecks(final WebClient webClient,
             return Optional.ofNullable(limitsCache)
                     .map(success -> limitsCache.get(key))
                     .map(cachedValue -> Future.succeededFuture((long) cachedValue))
-                    .orElseGet(() -> executeQuery(queryParams)
+                    .orElseGet(() -> executeQuery(queryParams, span)
                             .map(bytesConsumed -> addToCache(limitsCache, key, bytesConsumed)))
                     .map(bytesConsumed -> {
                         if ((bytesConsumed + payloadSize) <= allowedMaxBytes) {
+                            span.finish();
```
sophokles73: see above

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
kaniyan: The logging as mentioned above has been added.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
kaniyan: Now it has been replaced with map to finish the span.

service-base/src/test/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecksTest.java
```diff
@@ -90,11 +97,20 @@ public void setup() {
         cacheProvider = mock(CacheProvider.class);
         when(cacheProvider.getCache(any())).thenReturn(limitsCache);
 
+        spanContext = mock(SpanContext.class);
+
+        span = mock(Span.class);
+        when(span.context()).thenReturn(spanContext);
+        final Tracer.SpanBuilder spanBuilder = mockSpanBuilder(span);
+
+        tracer = mock(Tracer.class);
```
sophokles73: unless you also want to verify that Spans are created and finished, you will probably be better off with simply using the `NoopTracer` here ...

kaniyan: Now `mock(NoopTracer.class)` is used instead of `mock(Tracer.class)`.

Issue #1296, opened 2019-06-18
Feature request: Just-in-Time device provisioning


alinaserg:
We have got a new requirement to register devices when they attempt to connect to the IoT platform the first time. The provisioning setting should be done based on X.509 client certificate.
What do you think regarding this feature?

sophokles73:
Not sure what you mean by *IoT platform*. I assume you are talking about Eclipse Hono and the device connecting to e.g. the MQTT protocol adapter, presenting an X.509 client certificate for authentication, right?

alinaserg:
Yes, you are right.

alinaserg:
Under just-in-time provisioning we understand the following: not provisioned in Hono devices attempt to establish the first connection (in our case devices connect via MQTT protocol Adapter) using a certificate. As the devices are not provisioned, the automatically registration of the devices should be started (Device Registry API should be called), if the devices have valid certificates. No manual provisioning of the devices is not necessary any more. After that devices are provisioned and registered in Hono and can establish connections.

We propose to take device credentials from the  trusted CA certificate, which must be deposited in Hono before the device tries to establish the connection. It should also guarantee that only the devices with valid certificates can be registered.

sophokles73:
So, in any case, the corresponding tenant(s) that the devices belong to would need to have been created upfront, along with the trusted CAs that should be used for authenticating the (unregistered) devices.

That said, trying to use the Device Registration API for the purpose of adding an unknown device would not work anymore because we decided to remove the optional operations from all of Hono's device registry related AMQP based APIs in favor of the upcoming HTTP based Device Registry API.

So, based on that new API, the protocol adapter would theoretically be able to add a device for a tenant (and credentials) based on information contained in the client's certificate. However, I am not sure if that is what we really want to do because it would introduce a new dependency on the device registry's HTTP API to the protocol adapters (which already have several dependencies).

Another option might be to allow an implementation of the Device Registration and Credentials APIs to support *auto provisioning* based on the information that the adapters provide when invoking the existing operations. However, some of them would probably need to be extended with some optional information.

alinaserg:
The second option sounds interesting for us. Do you need some more information from us or it is already enough for you to accept it as a request for a new feature?

b-abel:
I would like to work on this topic.

sophokles73:
@b-abel any suggestions yet as to how you would like to approach this?

b-abel:
I currently see 2 main options: 
[...]

PR #1631, opened 2019-11-22
[#1296] Add feature auto provisioning

b-abel:
This adds minimal support for automatic device provisioning (#1296) when authenticating with X509 client certificates. Support is added for a flag in the client configuration that allows automatic provisioning to be enabled or disabled for each trusted CA configured. When a device connects to a protocol adapter, the adapter checks whether the flag for the trust anchor used is set to true. If so, it places the serialized device certificate in the payload of the GET request to the Credentials API. This way, the device registry has all the information available to perform automatic provisioning if necessary.


b-abel:
I will add documentation and tests when we have agreed on the implementation.

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -192,21 +219,41 @@ private TrustAnchor getTrustAnchorForPublicKey(final JsonObject keyProps) {
...
                     // Base64 decoding failed
                     return null;
                 }
+
+                final Boolean autoProvisioningEnabled = getProperty(keyProps,
+                        TenantConstants.FIELD_AUTO_PROVISIONING_ENABLED, Boolean.class, false);
+                if (autoProvisioningEnabled) {
+                    autoProvisioningEnabledIssuerDNs.add(subjectDn);
+                }
+                return trustAnchor;
```
sophokles73: I do not really like the interlocking of the *trustAnchors* and the *autoProvisioningEnabledIssuerDNs* fields created by your changes. IMHO it makes the code hard(er) to understand and maintain.
Given that we are dealing with a few (< 10) trust anchors anyway, I think that it would be perfectly fine to iterate over the JSON objects in order to determine if a trust anchor may be used for auto-provisioning. WDYT?

service-base/src/main/java/org/eclipse/hono/service/auth/device/SubjectDnCredentials.java
```diff
@@ -30,6 +30,8 @@
  */
 public class SubjectDnCredentials extends AbstractDeviceCredentials {
 
+    private byte[] clientCertificate = null;
```
sophokles73: I would rather have a generic way to add custom properties to any type of credentials. This is actually what we promised with the Credentials API but never stood up to. We originally introduced that feature in teh API to allow the MQTT adapter to include the MQTT client-id in the get Credentials request but never implemented it.
This seems to be a good opportunity to introduce this in a generic way. I can imagine that the `AbstractDeviceCredenials` class supports a `JsonObject` typed property that can be set in the constructor. WDYT?

service-base/src/main/java/org/eclipse/hono/service/auth/device/CredentialsApiAuthProvider.java
```diff
@@ -90,7 +90,9 @@ public CredentialsApiAuthProvider(final CredentialsClientFactory credentialsClie
      * @return A future containing the credentials on record as retrieved from
      *         Hono's <em>Credentials</em> API.
      * @throws NullPointerException if device credentials is {@code null}.
+     * @deprecated Use {@link #getCredentialsForDevice(AbstractDeviceCredentials, SpanContext)} instead.
      */
+    @Deprecated
```
b-abel: My gut feeling is that creating a new version of this method is overkill since the method is final as well as the only place in Hono's codebase where it is used (which is in the same class). So the only case in which changing the signature might cause problems is when someone uses it for something completely different. WDYT?

sophokles73: My understanding of generics is limited (at best) but IMHO you could simply replace the `DeviceCredentials` parameter type with `T` without breaking any client code because `DeviceCredentials` in fact `extends AbstractDeviceCredentials` ...

b-abel: If I change the type of the parameter to T you could no longer pass in an implementation of the interface `DeviceCredentials` that does not inherit from `AbstractDeviceCredentials`.

sophokles73: Ah, I thought `DeviceCredentials` was a class, my bad ...

service-base/src/main/java/org/eclipse/hono/service/auth/device/SubjectDnCredentials.java
b-abel: I like the idea. If you want to have the custom properties in any type of credentials, shouldn't it be added to the interface `DeviceCredentials`?

sophokles73: I guess so. Just remember to introduce it in a backwards compatible way ...

core/src/main/java/org/eclipse/hono/util/TenantObject.java
b-abel: I understand that you want to separate the _trustAnchors_ from the flag. But it still makes sense to build up such a map (separately from the parsing of trust anchors), doesn't it? As I understand the TenantObject will be cached, so the same instance can be used for many authentication requests (potentially of different devices of the same tenant) and there is no need to parse the same json again and again. WDYT?

sophokles73: I do not believe that there is much *parsing* involved as a `JsonObject` is basically a Map of Maps ... I doubt that there would be much of a performance difference ...

client/src/main/java/org/eclipse/hono/client/impl/AbstractRequestResponseClient.java
```diff
@@ -1001,7 +1001,7 @@ protected final boolean isCachingEnabled() {
      * <li>if the response does not contain any cache directive and the response's status code is
      * one of the codes defined by <a href="https://tools.ietf.org/html/rfc2616#section-13.4">
      * RFC 2616, Section 13.4 Response Cacheability</a>, the response is put to the cache using
-     * the default timeout returned by {@link #getResponseCacheDefaultTimeout()}<li>
+     * the default timeout returned by {@link #getResponseCacheDefaultTimeout()}</li>
```
sophokles73: I wonder why this is not detected by checkstyle

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -120,19 +120,44 @@ public TenantObject setEnabled(final boolean flag) {
      * @param subjectDn The CA's subject DN.
      * @return This tenant for command chaining.
      * @throws NullPointerException if any of the parameters is {@code null}.
+     * @deprecated Use {@link #addTrustAnchor(PublicKey, X500Principal, Boolean)} instead.
```
sophokles73: I do not see why this should be deprecated. Its original purpose is still valid, isn't it? We should document, though, that it clears the set of trust anchors and adds the given one but doesn't allow it to be used for auto provisioning.

service-base/src/main/java/org/eclipse/hono/service/auth/device/DeviceCredentials.java
```diff
@@ -39,4 +44,17 @@
...
+     * Gets additional properties that can be used to get credentials from the Credentials API.
+     * <p>
+     * This default implementation returns an empty JSON object.
+     *
+     * @return An empty JsonObject. An implementation can return a JSON object containing arbitrary properties. It is
+     *         intended to be directly usable to pass it into
+     *         {@link org.eclipse.hono.client.CredentialsClient#get(String, String, JsonObject, SpanContext)}.
+     */
+    default JsonObject getClientContext() {
+        return emptyClientContext;
```
sophokles73: IMHO we should return a new instance here instead of a shared instance in order to prevent surprises when one client starts to add properties to the returned instance ...

service-base/src/main/java/org/eclipse/hono/service/auth/device/TenantServiceBasedX509Authentication.java
```diff
@@ -222,12 +227,26 @@ public TenantServiceBasedX509Authentication(
...
                 .put(CredentialsConstants.FIELD_PAYLOAD_SUBJECT_DN, subjectDn)
-                .put(CredentialsConstants.FIELD_PAYLOAD_TENANT_ID, tenant.getTenantId()));
+                .put(CredentialsConstants.FIELD_PAYLOAD_TENANT_ID, tenant.getTenantId());
+
+        final String issuerDn = deviceCert.getIssuerX500Principal().getName(X500Principal.RFC2253);
+        if (tenant.isAutoProvisioningEnabled(issuerDn)) {
+            try {
+                authInfo.put(CredentialsConstants.FIELD_CLIENT_CERT, deviceCert.getEncoded());
+            } catch (CertificateEncodingException e) {
+                log.error("Encoding of device certificate [" + deviceCert.getSubjectX500Principal().getName()
```
sophokles73: please use `{}` in order to inject the principal name

service-base/src/main/java/org/eclipse/hono/service/auth/device/X509AuthProvider.java
```diff
@@ -59,7 +59,8 @@ public X509AuthProvider(final CredentialsClientFactory credentialsClientFactory,
      * The JSON object passed in is required to contain a <em>subject-dn</em> property.
      * If the <em>singleTenant</em> service config property is {@code false}, then the
      * object is also required to contain a <em>tenant-id</em> property, otherwise
-     * the default tenant is used.
+     * the default tenant is used. If the object contains the property
```
sophokles73: > If the authentication information contains the <em>client-certificate</em> property ...

service-base/src/main/java/org/eclipse/hono/service/auth/device/X509AuthProvider.java
```diff
@@ -83,7 +84,12 @@ protected SubjectDnCredentials getCredentials(final JsonObject authInfo) {
             if (tenantId == null || subjectDn == null) {
                 return null;
             } else {
-                return SubjectDnCredentials.create(tenantId, subjectDn);
+                final JsonObject clientContext = new JsonObject();
+                final byte[] clientCert = authInfo.getBinary(CredentialsConstants.FIELD_CLIENT_CERT);
```
sophokles73: IMHO this can throw a runtime exception if the property value is not a valid Base64 string ...

core/src/main/java/org/eclipse/hono/util/TenantObject.java
b-abel: FMPOV the add operation is more useful since Hono supports multiple trust anchors. I do not see anything special for auto-provisioning. Both methods are only used in tests. But on the other hand, it does not hurt to keep it.

service-base/src/main/java/org/eclipse/hono/service/auth/device/DeviceCredentials.java
b-abel: I was more worried about the surprise if it always returns a new instance. But you are right, this is not better. So, I'll change it.

b-abel:
I applied the rested changes and added unit tests.

b-abel:
I will add documentation. 

I'm thinking about adding support for this feature to the file-based device registry if this is not too much effort. This could serve as an implementation example and it would make it relatively easy to provide integration tests for this feature. 
WDYT?

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -207,6 +233,23 @@ private TrustAnchor getTrustAnchorForPublicKey(final JsonObject keyProps) {
         }
     }
 
+    /**
+     * Checks whether auto-provisioning is enabled for the given CA issuer DN.
```
sophokles73: IMHO this is a little misleading as it is unclear which property will be checked. What about
> Checks whether auto-provisioning is enabled for a CA.
>
> @param subjectDn The subject DN of the CA to check.

service-base/src/main/java/org/eclipse/hono/service/auth/device/DeviceCredentials.java
```diff
@@ -39,4 +42,17 @@
      * @return The tenant.
      */
     String getTenantId();
+
+    /**
+     * Gets additional properties that can be used to get credentials from the Credentials API.
+     * <p>
+     * This default implementation returns an empty JSON object.
+     *
+     * @return A new JsonObject. An implementation can return a JSON object containing arbitrary properties. It is
```
sophokles73: FMPOV this parameter declaration should be applicable to implementing classes as well. Thus, IMHO it should not describe the default implementation strategy but instead the abstract semantics of the object returned, e.g.
> @return The properties representing the client context.

The describing sentence *An implementation ...* should be moved to a paragraph of the method description.

service-base/src/main/java/org/eclipse/hono/service/auth/device/DeviceCredentials.java
```diff
@@ -39,4 +42,17 @@
      * @return The tenant.
      */
     String getTenantId();
+
+    /**
+     * Gets additional properties that can be used to get credentials from the Credentials API.
```
sophokles73: IMHO this should describe the purpose of the method in a more abstract way, not by means of one of its potential use cases.

service-base/src/main/java/org/eclipse/hono/service/auth/device/TenantServiceBasedX509Authentication.java
```diff
@@ -128,6 +129,9 @@ public TenantServiceBasedX509Authentication(
      *           "tenant-id": [identifier of the tenant that the device belongs to]
      *         }
      *         </pre>
+     *
+     *         Optional it may additionally have the property <em>client-certificate</em> as a Base64 encoded byte
```
sophokles73: What about

> If auto-provisioning is enabled for the trust anchor being used, the JSON object may optionally contain the DER encoding of the (validated) client certificate as a Base64 encoded byte array in the <em>client-certificate</em> property.

service-base/src/main/java/org/eclipse/hono/service/auth/device/X509AuthProvider.java
```diff
@@ -59,11 +59,21 @@ public X509AuthProvider(final CredentialsClientFactory credentialsClientFactory,
...
-     * the default tenant is used.
+     * the default tenant is used. If the authentication information contains a 
+     * <em>client-certificate</em> property, it is put into the returned credentials.
+     * <p>
+     * The values of the named properties are required to have the following types:
+     * <ul>
+     *     <li>tenant-id: {@link String}</li>
+     *     <li>subject-dn: {@link String}</li>
+     *     <li>client-certificate: Base64 encoded byte array as returned by {@link JsonObject#put(String, byte[])}</li>
+     * </ul>
```
sophokles73: FMPOV there is no reason to restrict the additional properties to just *client-certificate* here.
We can allow any other properties in addition to *tenant-id* and *subject-dn* and simply copy all of them to the client context. WDYT?

sophokles73:
@b-abel do you want to add documentation to this PR or create another one for it?

b-abel:
I am currently working on the documentation and would rather add it in this PR.

b-abel:
I added a small enhancement to the "Credentials API": the "Get" request can now return the status code 201. The documentation of the Credentials API states: "The response message’s status property may contain the following codes". So, I would like to explicitly add code 201 here. Which is FMPOV perfectly fine for a new minor version.
The `CredentialsClientImpl` now treats it the same way as a 200, so this makes no difference from the protocol adapters POV, but it allows to extend device registries and allow for cleaner APIs without breaking the client implementation. 
WDYT?

sophokles73:
@b-abel I cannot find any change to the Credentials API regarding the new status code ...

b-abel:
@sophokles73 I only committed the change in the client. But it might be easier to follow with documentation, so I will add it as well.

sophokles73:
@b-abel IMHO it should be the other way around: first we specify the intended behavior by means of the API, then we implement the spec and create clients for it :-)

b-abel:
> @b-abel IMHO it should be the other way around: first we specify the intended behavior by means of the API, then we implement the spec and create clients for it :-)

D'accord. Currently (Hono 1.0.x) the Credentials API does not restrict the status codes in response to the "get" operation, but the client (`CredentialsClient` / `CredentialsClientImpl`) treat the codes 201-299 as errors. I think we could at least support 201 on the client-side in Hono 1.1.

b-abel:
Regarding the documentation: Is this enough? Or do we additionally want to document it in the user guide for each protocol adapter? Or do we even need a central place to describe the whole idea (e.g. in concepts)?

site/documentation/content/api/credentials/index.md
```diff
@@ -50,6 +50,11 @@ The body of the request MUST consist of a single *Data* section containing a UTF
 | :--------------- | :-------: | :--------- | :---------- |
 | *type*           | *yes*     | *string*   | The type of credentials to look up. Potential values include (but are not limited to) `psk`, `x509-cert`, `hashed-password` etc. |
 | *auth-id*        | *yes*     | *string*   | The authentication identifier to look up credentials for. |
+| *client-certificate* | *no*  | *string*   | The client certificate the device authenticated with. If present, it MUST be the DER encoding of the (validated) X.509 client certificate as a Base64 encoded byte array. |
+
+The *client-certificate* property MAY be used by the service implementation for auto-provisioning of devices. 
+To do so, the device registry needs to create credentials (and registration data) for the device if they do not already exist. 
+It SHOULD then return the newly created credentials in the response payload.
```
sophokles73: Actually, it MUST return the credentials as specified further down the page. However, IMHO we should simply omit this sentence here because, as I already mentioned, the response payload requirements are stated further down anyway ...

site/documentation/content/api/credentials/index.md
```diff
@@ -62,6 +67,16 @@ The following request payload may be used to look up the hashed password for a d
 }
 ~~~
 
+The following request payload may be used to look up *or create* `x509-cert` credentials for a device with the authentication identifier `sensor1`:
```
sophokles73: that is not the authentication identifier ...

What happens if the auth-id differs from the subject DN in the certificate?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -581,10 +581,20 @@ components:
                   a certificate but then extract all relevant data and store
                   it in the properties described above.
                   Either this property or `public-key` must be set.
+            "auto-provisioning-enabled":
+               type: boolean
+               default: false
+               description: |
+                  Indicates whether this trusted certificate may be used for auto-provisioning of devices.
```
sophokles73: ... for automatically provisioning devices.

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -581,10 +581,20 @@ components:
                   a certificate but then extract all relevant data and store
                   it in the properties described above.
                   Either this property or `public-key` must be set.
+            "auto-provisioning-enabled":
```
sophokles73: IMHO the management API should return an error if a client tries to register a trust anchor with this property set to `true` but the registry implementation doesn't support auto-provisioning.

site/documentation/content/api/tenant/index.md
```diff
@@ -177,6 +177,7 @@ The table below provides an overview of the members of a JSON object representin
 | *subject-dn*             | *yes*      | *string*      | `-`           | The subject DN of the trusted root certificate in the format defined by [RFC 2253](https://www.ietf.org/rfc/rfc2253.txt). |
 | *public-key*             | *yes*      | *string*      | `-`           | The Base64 encoded binary DER encoding of the trusted root certificate's public key. |
 | *algorithm*              | *no*       | *string*      | `RSA`         | The name of the public key algorithm. Supported values are `RSA` and `EC`. |
+| *auto-provisioning-enabled* | *no*    | *boolean*     | `false`       | If set to `true`, protocol adapters MAY try to request auto-provisioning of devices that authenticate with a client certificate issued by this CA. Otherwise, protocol adapters MUST NOT request auto-provisioning. |
```
sophokles73: FMPOV a protocol adapter should assume that the registry supports auto-provisioning if this property is included and set to `true`, i.e. the adapter then MAY request the registry to automatically provision the device based on its client certificate (and use the status code to determine the outcome). My point is that an adapter MAY not only *try* to do this but it should assume that the registry actually supports doing so ...

site/documentation/content/api/credentials/index.md
b-abel: > What happens if the auth-id differs from the subject DN in the certificate?

We could add here that the device registry must check that they match and return a 400 if they don't. 
Another option would be to remove the auth-id property if a client certificate is present in the request, with the downside that the property is not always mandatory. WDYT?

sophokles73: IMHO returning 400 is the better option as it does not require any changes to the default mechanism of requesting credentials.

site/homepage/content/release-notes.md
```diff
@@ -11,6 +11,14 @@ title = "Release Notes"
 * The (experimental) CoAP adapter now supports tracking of request processing using
   OpenTracing.
 * A lorawan provider has been added for the loriot network provider.
+* Hono now supports "auto-provisioning" of devices that authenticate with X.509 client certificates.
```
sophokles73: please use emphasis instead of quotes

site/homepage/content/release-notes.md
```diff
@@ -11,6 +11,14 @@ title = "Release Notes"
 * The (experimental) CoAP adapter now supports tracking of request processing using
   OpenTracing.
 * A lorawan provider has been added for the loriot network provider.
+* Hono now supports "auto-provisioning" of devices that authenticate with X.509 client certificates. 
+  To enable it, the *auto-provisioning-enabled* property on the trust anchor 
+  must be set to `true` in the tenant configuration. The protocol adapters add the client certificate to the "get" request to the
```
sophokles73: we definitely need a more detailed description of the *auto provisioning* concept which provides information about the overall purpose, constraints, pre-conditions etc.

b-abel: Do you mean in the release notes? Or in the documentation, e.g. user guides or a concept page?

sophokles73: I mean in a concept page and maybe also in the user guides ...

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -107,6 +107,17 @@ paths:
                      example:
                         error: "Root Certificate Authority is already used by other tenant"
                         subject-dn: "CN=devices,OU=iot,O=ACME"
+            501:
```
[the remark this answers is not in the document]
b-abel: If we use a more generic code like 400, I would not document it in the swagger file. So I would remove it from the yaml if we go for 400. WDYT?

b-abel:
I added a concept page for this feature. If you approve it, I would not add it to the user guides of the protocol adapters. I do not see what should go there as the idea is that the device does not need to know if it is registered.

b-abel:
@sophokles73 IMHO the page Device Identity and the user guide of the device registry (together with the management API specification) currently describe device provisioning. Therefore it is a little difficult for me not to produce too much overlap. Please find my proposal in the last commit.

site/homepage/content/release-notes.md
```diff
@@ -11,6 +11,15 @@ title = "Release Notes"
 * The (experimental) CoAP adapter now supports tracking of request processing using
   OpenTracing.
 * A lorawan provider has been added for the loriot network provider.
+* Hono now supports *auto-provisioning* of devices that authenticate with X.509 client certificates. 
+  To enable it, the *auto-provisioning-enabled* property on the trust anchor 
+  must be set to `true` in the tenant configuration. The protocol adapters add the client certificate to the "get" request to the 
+  Credentials API, if this is enabled for the trust anchor used, so that the device registry can create 
+  credentials (and device registration data) "on-the-fly".
```
sophokles73: please do not use quotes ("...") but use emphasis (*...*) instead

site/documentation/content/api/management/device-registry-v1.yaml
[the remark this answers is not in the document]
sophokles73: Works for me.

site/documentation/content/api/credentials/index.md
```diff
@@ -74,14 +88,15 @@ A response to a *get credentials* request contains the following properties:
...
-The response message payload MUST contain credential information as defined in [Credentials Format]({{< relref "#credentials-format" >}}) if the *status* is `200`.
+The response message payload MUST contain credential information as defined in [Credentials Format]({{< relref "#credentials-format" >}}) if the *status* is `200` or `201`.
 
 The response message's *status* property may contain the following codes:
 
 | Code  | Description |
 | :---- | :---------- |
 | *200* | OK, the payload contains the credentials for the authentication identifier. |
-| *400* | Bad Request, the request message did not contain all mandatory properties.
+| *201* | Created, the payload contains the just created credentials for the authentication identifier. |
```
sophokles73: .. contains the newly created credentials ...

client/src/main/java/org/eclipse/hono/client/CredentialsClient.java
```diff
@@ -38,7 +38,7 @@
      * @param authId The authentication identifier used in the credentials to retrieve.
      * @return A future indicating the result of the operation.
      *         <p>
-     *         The future will succeed if a response with status 200 has been received from the
+     *         The future will succeed if a response with status 200 or 201 has been received from the
```
sophokles73: .. with status 2xx has been received ... ?

site/documentation/content/api/management/device-registry-v1.yaml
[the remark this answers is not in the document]
sophokles73: Are you going to remove the 501 then?

site/homepage/content/release-notes.md
```diff
@@ -11,6 +11,15 @@ title = "Release Notes"
 * The (experimental) CoAP adapter now supports tracking of request processing using
   OpenTracing.
 * A lorawan provider has been added for the loriot network provider.
+* Hono now supports *auto-provisioning* of devices that authenticate with X.509 client certificates. 
+  To enable it, the *auto-provisioning-enabled* property on the trust anchor 
+  must be set to `true` in the tenant configuration. The protocol adapters add the client certificate to the *get* request to the 
+  Credentials API, if this is enabled for the trust anchor used, so that the device registry can create 
+  credentials (and device registration data) "on-the-fly".
```
sophokles73: IMHO we can remove these sentences as the details are covered by the information referred to below ...

client/src/main/java/org/eclipse/hono/client/CredentialsClient.java
b-abel: Sounds good, but would require to change the implementation. Then we would have to deal with 204 where we can not return a payload...

b-abel:
Since there have been no further comments by others, I would remove the error codes 501 from the management API. This means that we leave the behavior of Device Registries that do not support auto-provisioning completely unspecified, right? FMPOV this is OK.

client/src/main/java/org/eclipse/hono/client/CredentialsClient.java
sophokles73: Why is that? The JavaDoc explicitly refers to the status codes that are defined by the Credentials API. These do not include 204, do they?

b-abel: Yes, that makes sense.

b-abel:

sophokles73:
@b-abel  ok, can you squash and rebase?

Issue #1617, opened 2019-11-07
Use SNI to select server certificates out of set.


boaks:
For central deployments with multiple applications it may be required, that the same protocol adapter endpoints are reached by different DNS names. For simple server certificate validation that would require either "long list of subject-alternative-names" or a mechanism to select a matching certificate out of a set of certificates by that DNS name. 
Therefore I would like to be able to use [SNI](https://tools.ietf.org/html/rfc6066#section-3) to select a matching certificate out of a set.
To do so, the configuration should be extended to multiple certificates and related private keys.
The configuration should contain the DNS hostnames as provided in the SNI to lookup the specific certificate-private-key pair.

sophokles73:
Sounds reasonable. The vert.x framework used by Hono already supports using SNI by means of putting multiple key/cert pairs into a key store and use that for the server side identity. The only thing currently missing in Hono for this to work is actually enabling SNI in the NetServer.

bordeuax:
Hi @sophokles73 , I want to clarify (and yes of course correct me if i am wrong) that currently Eclipse Hono can be configured with two types of Key/certificate format JKS and PEM . We will have only in case of JKS full SNI enabled endpoints. And the requirements for SNI will be configuration key "sni: true" and JKS format for key/certificate.

sophokles73:
> currently Eclipse Hono can be configured with two types of Key/certificate format JKS and PEM

It's actually three types: JKS, p12 and plain PEM. The former two are key store formats and may contain multiple keys/certs whereas the last one may only contain a single key/cert.

You can use any of the file formats for SNI, though. See [vert.x docs for more info](https://vertx.io/docs/vertx-core/java/#ssl) ...

bordeuax:
Thanks for precising regarding JKS, P12 and PEM formats. My point here is to have full enable SNI endpoints that not depending of the certificate format. From technical point of view we can have multiple PEM files because [PemKeyCertOptions]( https://vertx.io/docs/apidocs/io/vertx/core/net/PemKeyCertOptions.html) support also a list of files.

PR #1639, opened 2019-11-26
[#1617] Add SNI configuration support

bordeuax:
Add support for SNI configuration :
*  Into server  side are `sni: true` option that enable SNI support 
*  Client side we have` sniServerName` that override host name into TLS SNI negotiation 
*  Documentation for protocols adapters and client 

For more info https://github.com/eclipse/hono/issues/1617


core/src/main/java/org/eclipse/hono/config/AbstractConfig.java
```diff
@@ -54,6 +54,7 @@
     private FileFormat trustStoreFormat;
     private FileFormat keyFormat;
     private List<String> secureProtocols = Collections.singletonList("TLSv1.2");
+    private boolean sni;
```
sophokles73: IMHO this should go into `ServerConfig` as it is only relevant/used for a server.

core/src/main/java/org/eclipse/hono/config/AbstractConfig.java
```diff
@@ -432,6 +433,22 @@ public final void setSecureProtocols(final List<String> enabledProtocols) {
         this.secureProtocols = enabledProtocols;
     }
 
+    /**
+     * Set whether the server supports Server Name Indication.
```
sophokles73: Please use active present tense, i.e. 
> Sets whether the server should support Server Name Indication for TLS connections.

core/src/main/java/org/eclipse/hono/config/AbstractConfig.java
```diff
@@ -432,6 +433,22 @@ public final void setSecureProtocols(final List<String> enabledProtocols) {
...
+     *
+     * @param sni Server Name Indication status.
```
sophokles73: for boolean parameters we usually use something like
> `true` if the server should use SNI.

core/src/main/java/org/eclipse/hono/config/AbstractConfig.java
```diff
@@ -432,6 +433,22 @@ public final void setSecureProtocols(final List<String> enabledProtocols) {
...
+     */
+    public final void setSni(final boolean sni) {
+        this.sni = sni;
+    }
+
+    /**
+     * @return whether the server supports Server Name Indication.
```
sophokles73: Description is missing

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -555,6 +557,30 @@ public final void setTlsEnabled(final boolean enabled) {
         this.tlsEnabled = enabled;
     }
 
+    /**
+     * Get the host override for TLS SNI Server Name set by {@link #setSniServerName(String)}.
```
sophokles73: active present tense, no need to refer to the setter (this should be clear by means of the property name)

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -555,6 +557,30 @@ public final void setTlsEnabled(final boolean enabled) {
...
+     *
+     * @return  the sniServerName
```
sophokles73: Please start the description with an upper case letter and finish with a period (`.`)

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -555,6 +557,30 @@ public final void setTlsEnabled(final boolean enabled) {
...
+     */
+    public final String getSniServerName() {
+        return sniServerName;
+    }
+
+    /**
+     * Explicitly override the host to use for the TLS SNI server name.
```
sophokles73: IMHO this should be simply
> Sets the host name to indicate in the Server Name Indication TLS extension.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -555,6 +557,30 @@ public final void setTlsEnabled(final boolean enabled) {
...
+     *
```
sophokles73: Please insert a `<p>` between the paragraphs, otherwise the JavaDoc will contain consecutive sentences in a single paragraph.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -555,6 +557,30 @@ public final void setTlsEnabled(final boolean enabled) {
...
+     * If neither the {@link ClientConfigProperties#setHost(String) host} or SNI server name is explicitly
+     * overridden, the host specified in {@link ClientConfigProperties#getHost()} will be used, with SNI performed implicitly
```
sophokles73: The *host* is used if the *SniServerName* hasn't been set explicitly, isn't it?

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -555,6 +557,30 @@ public final void setTlsEnabled(final boolean enabled) {
...
+     * where a FQDN was specified.
+     *
+     * This method should typically only be needed to set different values for the host and SNI server name.
+     *
+     * @param sniServerName to set as SNI server name
```
sophokles73: Please start the description with an upper case letter and finish with a period, e.g. 
> `@param` sniServerName The host name to use in the SNI extension.

core/src/main/java/org/eclipse/hono/connection/impl/ConnectionFactoryImpl.java
```diff
@@ -324,6 +324,10 @@ private void addTlsTrustOptions(final ProtonClientOptions clientOptions) {
                 logger.debug("enabling secure protocol [{}]", protocol);
                 clientOptions.addEnabledSecureTransportProtocol(protocol);
             });
+
+            if (config.getSniServerName() != null && !config.getSniServerName().trim().isEmpty()){
```
sophokles73: org.eclipse.hono.utils.Strings.isNullOrEmpty

service-base/src/main/java/org/eclipse/hono/service/AbstractServiceBase.java
```diff
@@ -432,6 +432,9 @@ protected final void addTlsKeyCertOptions(final NetServerOptions serverOptions)
                 log.info("enabling secure protocol [{}]", protocol);
                 serverOptions.addEnabledSecureTransportProtocol(protocol);
             });
+
+            serverOptions.setSni(getConfig().isSni());
+            log.info("enabling TLS SNI [{}]", getConfig().isSni());
```
sophokles73: how about logging this line *only* if we are actually enabling SNI? Otherwise, maybe change the log statement to
> Service supports TLS ServerNameIndication: {}

site/documentation/content/admin-guide/amqp-adapter-config.md
```diff
@@ -27,6 +27,7 @@ The following table provides an overview of the configuration variables and corr
...
 | `HONO_AMQP_KEY_STORE_PASSWORD`<br>`--hono.amqp.keyStorePassword` | no | - | The password required to read the contents of the key store. |
 | `HONO_AMQP_KEY_STORE_PATH`<br>`--hono.amqp.keyStorePath` | no | - | The absolute path to the Java key store containing the private key and certificate that the protocol adapter should use for authenticating to clients. Either this option or the `HONO_AMQP_KEY_PATH` and `HONO_AMQP_CERT_PATH` options need to be set in order to enable TLS secured connections with clients. The key store format can be either `JKS` or `PKCS12` indicated by a `.jks` or `.p12` file suffix respectively. |
+| `HONO_AMQP_SNI`<br>`--hono.amqp.sni` | no | `false` | Set whether the server supports Server Name Indication. By default, the server will not support SNI and the option is `false`. However, if set to `true` then the key store format, `HONO_AMQP_KEY_STORE_PATH`,  should be either `JKS` or `PKCS12` indicated by a `.jks` or `.p12` file suffix respectively. |
```
sophokles73: IMHO it would be helpful to add a section to `secure_communication.md` describing how SNI can be used with the protocol adapters and what the server side key stores need to contain in order to make it work. This can also be a reference to the vert.x documentation in order to avoid duplication. But since SNI it very commonly known, I think we should explain a little. Would you agree?

bordeuax: Yes totally make sense, I will add.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
bordeuax: How I know this is the default behavior of ` io.vertx.proton.ProtonClient` :  if SNI Server name is not specified `host `name is used.

sophokles73: https://vertx.io/docs/vertx-core/java/#ssl

site/documentation/content/admin-guide/amqp-adapter-config.md
bordeuax: I added the documentation , please take a look and welcome if are any remarks !

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -93,6 +93,7 @@
     private long requestTimeoutMillis = DEFAULT_REQUEST_TIMEOUT;
     private long sendMessageTimeoutMillis = DEFAULT_SEND_MESSAGE_TIMEOUT;
     private boolean tlsEnabled = false;
+    private String sniServerName;
```
sophokles73: having taken a look at the JavaDoc of `io.vertx.proton.ProtonClientOptions`, I believe that we should set the *virtualHost* property instead of the *sniServerName* property. My understanding is that this will set both the server name in the SNI extension at the TLS level as well as the *hostname* property in the AMQP open frame. The latter is much easier to access from application level code (e.g. the protocol adapters) in order to e.g. determine the tenant from the hostname in the future ...

bordeuax: In TLS SNI communication we have 2 scenarios : 

1.  Client establish TLS SNI connection and  `virtualHost ` are equal with  `sniServerName `
In this case ,  we need only to set  `virtualHost ` that automatically set  `sniServerName `

2. Client establish TLS SNI connection and  `virtualHost `are different from   `sniServerName ` 
In this case , we need to set `virtualHost ` that set also  `sniServerName ` and after override `sniServerName `  by setting this via `setSniServerName(...)`

sophokles73: The second case is not a supported scenario in AMQP. In fact, its semantics is not specified.
Fro the AMQP 1.0 spec, section 2.7.1 (*open* frame):
> hostname: the name of the target host
The name of the host (either fully qualified or relative) to which the sending peer is connecting. It is
not mandatory to provide the hostname. If no hostname is provided the receiving peer SHOULD
select a default based on its own configuration. This field can be used by AMQP proxies to
determine the correct back-end service to connect the client to.
This field MAY already have been specified by the sasl-init frame, if a SASL layer is used, or,
the server name indication extension as described in RFC-4366, if a TLS layer is used, in which
case this field SHOULD be null or contain the same value. It is undefined what a different value
to that already specified means.

I therefore do not think that we should allow clients to set an SNI server name that is different from the virtual host. Would you agree?

bordeuax: Yes, for AMQP 1.0 totally make sense.

bordeuax: In case that TLS SNI will be mainly use by protocols adapter maybe make sens to move the configuration from `ServerConfig 
`  to `ProtocolAdapterProperties` ?

sophokles73: But that would also limit it to be used by protocol adapters as well. FMPOV SNI is a concept of general interest on the server side. Thus, I would rather keep it in `ServerConfig`.

site/documentation/content/admin-guide/secure_communication.md
```diff
@@ -127,6 +127,43 @@ In order to address this problem, the Netty networking library that is used in H
...
+* Public Key Infrastructure(PKI) for your IoT devices can have many trust stores.
+* Specific tenant TLS requirements.           
+
+Assuming that you have one of use cases that require SNI, the following steps are necessary to enable on Hono protocols adapters TLS SNI:
+
+* Certificate generation 
+
+Specific Hono protocol adapter should be configured to use key store, for example for MQTT adapter we have `--hono.mqtt.keyStorePath` for more info please refer to the specific protocol adapter configuration. 
+The key store can be only JKS or P12 format. Please note, that PEM format is not a key store and in this case PEM format will not work.
+The Common Name (CN), also known as the Fully Qualified Domain Name (FQDN), of yours certificate that are stored in key store should correspond to SNI name requested by the IoT device.
```
sophokles73: My understanding is that using the common name for the FQDN is discouraged these days in favor of adding a *subject alternative name* (SAN) property to the certificate. vert.x also considers this property when matching the SNI server name with a certificate. Based on that, I think we should rather recommend to use a corresponding SAN entry. WDYT?

bordeuax: I also can see some ambiguity between what i wrote and vert.x documentation. I will update it.

PR #1640, opened 2019-11-26
[#1635] Support overriding supported SASL mechanisms

### LINKED ISSUE #1635: Only supported SASL mechanisms should be advertised to the client
Currently a Hono component including an `AuthenticationService` implementation for handling the SASL exchange (e.g. the `FileBasedAuthenticationService` in the Hono Auth component) always indicates both `PLAIN` and `EXTERNAL` as supported SASL mechanisms to clients.

For `AuthenticationService` implementations that only support one or the other, the advertised mechanisms should be configurable.

calohmn:
This fixes #1635.


service-base/src/main/java/org/eclipse/hono/service/auth/HonoSaslAuthenticatorFactory.java
```diff
@@ -40,11 +40,16 @@
...
      * @param vertx the Vertx environment to run the factory in.
      * @param validator The object to use for validating auth tokens.
+     * @param actualAuthenticationService The authentication service that will handle the authentication requests sent
+     *            on the Vert.x event bus. Used for determining the supported SASL mechanisms.
      * @throws NullPointerException if any of the parameters is {@code null}.
      */
     @Autowired
-    public HonoSaslAuthenticatorFactory(final Vertx vertx, @Qualifier(AuthenticationConstants.QUALIFIER_AUTHENTICATION) final AuthTokenHelper validator) {
-        this(new EventBusAuthenticationService(vertx, validator));
+    public HonoSaslAuthenticatorFactory(final Vertx vertx,
```
sophokles73: this is public API so we cannot simply change it in a non backward compatible way. Apart from that, IMHO the restriction of the mechanisms is not so much a matter of the capabilities of the underlying `AuthenticationService` but instead is a matter of configuration. The `FileBasedAuthenticationService` can very well authenticate a client based on the certificate's subject DN. However, in certain scenarios you do not want the `SaslAuthenticator` sitting in front of it to offer the *EXTERNAL* mechanism because it doesn't make sense in the given system setup. IMHO we should therefore simply add constructors to this class which also accept a list of supported mechanisms and which is then used to configure the `SaslAuthenticator`.

sophokles73:
@calohmn any news on this?

service-base/src/main/java/org/eclipse/hono/service/auth/HonoSaslAuthenticatorFactory.java
calohmn: I agree that it would be better to have the `supportedSaslMechanisms` be configurable via a config property. I'll have a look at that and update this PR.

calohmn: > IMHO we should therefore simply add constructors to this class which also accept a list of supported mechanisms and which is then used to configure the `SaslAuthenticator`.

I would rather have the supported mechanisms be exposed via an `AuthenticationService#getSupportedSaslMechanisms()` method, because I see the corresponding config property as belonging to the config properties of the authentication service as well. See the 2nd commit I just added for the added config properties. (I first wanted to put these in a common base class but this doesn't work because of the existing inheritance hierarchy in `AuthenticationServerClientConfigProperties`).

I've amended the first commit to add a constructor to `HonoSaslAuthenticatorFactory` matching the current one to keep backwards compatibility. I've also added some checks, skipping stuff if one or the other mechanism isn't configured to be supported.

calohmn:
@sophokles73 See comment above.

service-base/src/main/java/org/eclipse/hono/service/auth/HonoSaslAuthenticatorFactory.java
```diff
@@ -37,13 +37,33 @@
...
+     * <p>
+     * Verifies credentials by means of sending authentication requests to address
+     * {@link AuthenticationConstants#EVENT_BUS_ADDRESS_AUTHENTICATION_IN} on the Vert.x
+     * Event Bus.
+     *
+     * @param vertx the Vertx environment to run the factory in.
+     * @param validator The object to use for validating auth tokens.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     */
+    public HonoSaslAuthenticatorFactory(final Vertx vertx, final AuthTokenHelper validator) {
```
sophokles73: FMPOV this should go *before* the new constructor variant. Is there a particular reason why the qualifier has been removed from the *validator* parameter?

calohmn: > Is there a particular reason why the qualifier has been removed from the validator parameter?

FMPOV, the qualifier was included in connection with the `Autowired` annotation. Since the newly introduced constructor has the `Autowired` annotation now, I've removed the `Autowired` annotation here along with the qualifier for the _validator_ parameter.

sophokles73: Ah, I see. So this constructor will not be subject to auto wiring at all.

calohmn:
@sophokles73 I've made the suggested changes, also adding extra validation in `setSupportedSaslMechanism`.

calohmn:
(Rearranged changes into 2 commits.)

sophokles73:
@calohmn how about an entry in the release notes regarding this new feature?

calohmn:
@sophokles73 I've already done so (see https://github.com/eclipse/hono/commit/3de45e3de00f3ddaef5132ebcd967d5927d28e22).

sophokles73:
@calohmn IMHO we should also configure the AuthServer and example registry in our Helm chart accordingly.

PR #1641, opened 2019-11-29
[Idea] Share test helper classes among multiple maven modules

kaniyan:
While I was writing tests for #1440, I found out that I could reuse a method from a test helper class `HonoClientUnitTestHelper.mockSpanBuilder(...)` in tests across several modules. But it is not currently possible as test classes are not packaged by Maven. In order to share test helper classes among various modules, how about using _maven-jar_ plugin to build a jar with those _test-classes_ and then then add it as a dependency wherever needed. 

For example, a new package `org.eclipse.hono.test` under _tests_ folder in _core_ module could contain test helper classes that could be used by tests in other modules. The classes in this package could be used to create a _test-jar_ as below.

```
     <plugin>
        <artifactId>maven-jar-plugin</artifactId>
        <configuration>
          <includes>
            <include>**/test/*</include>
          </includes>
        </configuration>
        <executions>
          <execution>
            <goals>
              <goal>test-jar</goal>
            </goals>
          </execution>
        </executions>
      </plugin>
```

Then the above artifact containing test helper classes could be added as a dependency with type _test-jar_ as below.

```
    <dependency>
      <artifactId>hono-core</artifactId>
      <type>test-jar</type>
      <scope>test</scope>
    </dependency>
```
WDYT?

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -53,6 +53,13 @@
     </dependency>
     <dependency>
       <groupId>org.eclipse.hono</groupId>
+      <artifactId>hono-core</artifactId>
+      <type>test-jar</type>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.hono</groupId>
       <artifactId>hono-legal</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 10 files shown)


sophokles73:
@kaniyan that's a nice idea, we are already using this approach for the test classes of the service-base module. I agree that there are several helper classes which would be valuable for implementing other tests as well.

kaniyan:
This has been closed in favour of #2249.

PR #1643, opened 2019-11-29
enable set maxInstances independent from processor count

rpahli:
We run hono on kubernetes and the cpu usage is currently not very hight, and we want to increase the vertx instances but this is not possible because of the processor count check.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/config/ApplicationConfigProperties.java
+++ b/core/src/main/java/org/eclipse/hono/config/ApplicationConfigProperties.java
@@ -51,14 +51,14 @@ public class ApplicationConfigProperties {
      * <p>
      * The number is calculated as follows:
      * <ol>
-     * <li>if 0 &lt; <em>maxInstances</em> &lt; #processors, then return <em>maxInstances</em></li>
+     * <li>if 0 &lt; <em>maxInstances</em>, then return <em>maxInstances</em></li>
      * <li>else return {@code Runtime.getRuntime().availableProcessors()}</li>
      * </ol>
      *
      * @return the number of verticles to deploy.
      */
     public final int getMaxInstances() {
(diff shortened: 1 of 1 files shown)


sophokles73:
Rico,

can you elaborate a little on what you mean *CPU usage is not very high*, i.e. how many CPU shares does your POD get (requests, limits) and how much of that is actually used? I am asking because I find it rather difficult to identify the number of verticle instances being the limiting factor here. My experiments on real hardware seem to indicate that vert.x is quite good at utilizing CPU if there is enough workload.

Kai

sophokles73:
@rpahli any news on this?

rpahli:
Sry I was busy,

 I try to start a hono adapter without any limitations on kubernetes but the adapter used only one cpu. That's the reason I want to override the cpu by myself but the CPU check limits me.

sophokles73:
> I try to start a hono adapter without any limitations on kubernetes but the adapter used only one cpu.

How many cores does the Kubernetes node have that the adapter pod got scheduled on?
Did you set *maxInstances* to 0?

rpahli:
> > I try to start a hono adapter without any limitations on kubernetes but the adapter used only one cpu.
> 
> How many cores does the Kubernetes node have that the adapter pod got scheduled on?
> Did you set _maxInstances_ to 0?

Yes I tryed it with 0 and the node has 2 cores.

sophokles73:
> Yes I tryed it with 0 and the node has 2 cores.

Ok, and what did you expect to happen then when you started publishing messages to the adapter?

rpahli:
@sophokles73 sry for the late replay. We have now a solution which fit's for us. I think we can close the PR.

sophokles73:
@rpahli may I ask how your solution looks like? Maybe others would be interested to know as well ...
You can, of course, close this issue whenever you want as you opened it in the first place.

PR #1647, opened 2019-12-02
Upgrade to Artemis 2.10.1 and Dispatch Router 1.8.0

sophokles73:
Proposal for new approach of using more recent, standard Artemis image


sophokles73:
replaced by #1651

PR #1649, opened 2019-12-06
Fix tracing of received commands

calohmn:
This fixes the `delegate Command request` tracing span containing `null` values for tenant and device id.
It also sets a `gateway_id` span tag if applicable.


sophokles73:
@calohmn are you going to fix the unit tests as well?

calohmn:
@sophokles73 test fixed now - travis is green.

PR #1651, opened 2019-12-06
Update artemis and router

dejanb:


dejanb:
We should be able to control it using `JAVA_OPTS` env variable. The launch.sh script should use that when creating a broker instance. I'll give it a thorough test to be sure.

dejanb:
Added options for VM memory management that should have the same effect as the old configuration. It's just for the reference as the default ones should be good enough.

We should move the artemis image to jdk-11 profile next. I opened Jira for that https://github.com/EnMasseProject/artemis-docker-base/issues/12

deploy/src/main/deploy/helm/templates/artemis/artemis-deployment.yaml
```diff
@@ -27,12 +27,17 @@ spec:
...
-        - name: ARTEMIS_CONFIGURATION
-          value: /etc/hono
+        - name: AMQ_NAME
+          value: custom
+        - name: HOME
+          value: /var/run/artemis/
+        - name: JAVA_INITIAL_MEM_RATIO
+          value: "30"
+        - name: JAVA_MAX_MEM_RATIO
+          value: "70"
```
sophokles73: Does this mean that we expect Artemis to successfully start up with 30% of 256MB?

dejanb: It's a bit confusing, as it is 30% of max value (360M) which is roughly what it was configured before. BTW. Defaults are 256M max and 64M init, which works fine in system tests.

When we switch to openjdk-11 profile we can switch to -XX:InitialRAMPercentage / -XX:MaxRAMPercentage config combo.

sophokles73: Ok, so currently, the Artemis image uses Java < 9 and manually computes the Xms and Xmx values from the environment variables `JAVA_INITIAL_MEM_RATIO` and `JAVA_MAX_MEM_RATIO`?

dejanb: It uses JDK 11, but launch scripts use "legacy" mode which compute manually all  the parameters. I'll need to investigate further to see what does it take to upgrade this image to use proper mode.

sophokles73: I just wanted to understand. If you think that, for the time being, using it the way you did is the right way, then I am fine with that ...

dejanb: Sure thing. Yeah, I think this is good to go and we should continue to work on getting artemis image to the latest features. Do you want me to squash these commits?

sophokles73: IMHO this small change doesn't require more than one commit, does it?

sophokles73:
@dejanb we also need CQs for new versions of Artemis and its dependencies if the versions have changed  at minor version level or higher ... Note that the IP process at the Eclipse Foundation has changed and only requires license compatibility checks now, i.e. all CQs should be Type A.

sophokles73:
@dejanb it looks like Artemis is logging quite verbosely at INFO level. The log for the integration tests are flooded with messages from Artemis which haven't been there with the previous version. IMHO we should either find out why Artemis logs these messages and whether this is due to some wrong usage on Hono's side or we should reduce the log level to e.g. ERROR. WDYT?

dejanb:
@sophokles73 #1659 should fix the logging. CQs are raised.

PR #1654, opened 2019-12-09
[#1653] Prevent the commandConsumer from not getting closed

### LINKED ISSUE #1653: HTTP adapter: "message consumer already in use" error for TTD requests
The HTTP adapter may get into a state where each new telemetry/event request with a `ttd` param for a specific device will fail to create a command consumer. Internally, a `message consumer already in use` is thrown because there already is a command consumer from a previous request that didn't get closed. 
This only gets resolved when the AMQP messaging network closes the existing command consumer link or if the adapter gets restarted.

Looking at a heap dump of an adapter instance that got into this state revealed that the command consumer didn't get closed at a particular HTTP request when there was an error creating the AMQP message sender (`not connected`).

calohmn:
This fixes #1653.

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -675,7 +675,9 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
                         responseReady,
                         currentSpan));
 
-        CompositeFuture.all(senderTracker, commandConsumerTracker)
+        // "join" used here to make sure that the "commandConsumerTracker" is completed when handling
+        // a failed "senderTracker" below, allowing a contained command consumer to be closed there.
+        CompositeFuture.join(senderTracker, commandConsumerTracker)
         .compose(ok -> {
 
             final DownstreamSender sender = senderTracker.result();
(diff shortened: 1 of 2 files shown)


calohmn:
https://github.com/eclipse/hono/blob/cedfd283b0f1f0696cf2186ad9d99fd7050a6f2d/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java#L790-L801

Issue was, that the `commandConsumerTracker` was not completed here yet.
Now it is ensured that it is.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -675,7 +675,7 @@ private void doUploadMessage(
                         responseReady,
                         currentSpan));
 
-        CompositeFuture.all(senderTracker, commandConsumerTracker)
+        CompositeFuture.join(senderTracker, commandConsumerTracker)
```
ctron: I think I understood what happens here. But I think it would be worth spending a few words, in a source code comment, of why "join" is needed here.

Aside from that I would wish we could split of those nested monster methods and make them easier to understand.

sophokles73: I agree. IMHO adding the comment shouldn't be a big thing. Splitting up the *monster* might deserve a separate issue. Would that work for everybody?

calohmn: I agree about the comment as well and I've amended the commit accordingly.

And yes, I also would like to see the method refactored. IMHO, this should be done in a separate PR though.

ctron: > Would that work for everybody?

Absolutely! It was just a general remark, after trying to understand what is going on.

Issue #1644, opened 2019-12-02
Document limitations of Gateways subscribing for single device commands (MQTT/AMQP)


calohmn:
Scenario:
Hono is deployed in such a way that the AMQP and MQTT adapter endpoints are represented each by a loadbalancer with _multiple_ protocol adapter instances behind it.

Implications for Command&Control:
In such a setup, gateways initiating multiple subscriptions (via AMQP/MQTT) for commands of *specific* devices must do so over the same AMQP/MQTT connection.
This should be made clear in the documentation.

Background:
For now it has to be made sure that command subscriptions related to one specific gateway are handled by the same protocol adapter instance - otherwise there will be 2 concurrent consumer links on the gateway-specific address and commands for a specific device might end up at the wrong instance. Using the same AMQP/MQTT connection ensures that the same adapter instance is used.
With #1632, using a different connection and arriving at a different adapter instance will lead to the other subscriptions being closed, thereby preventing lost command messages.
With #1272 (Hono 1.1), the overall link handling regarding command & control shall be changed. With that, the above limitation doesn't exist anymore.

PR #1657, opened 2019-12-09
[#1645] Add tenant option disabling gateway mapping (HTTP adapter)

### LINKED ISSUE #1645: HTTP: Concurrency issue with gateways subscribing for single device commands
Scenario:
- Hono is deployed in such a way that the HTTP adapter endpoint is represented by a loadbalancer with multiple protocol adapter instances behind it.
- a gateway sends multiple, potentially concurrent, event/telemetry messages with a _ttd_ parameter (ie. waiting for a command message) on behalf of a specific device.

Potential issue:
If multiple concurrent requests related to the same gateway end up on different protocol adapter instances, multiple command consumer links on the gateway-specific address will be opened on the different instances. This may lead to commands reaching the wrong adapter instance and getting released. 

Solution:
For Hono 1.0.x, the solution is to effectively prevent the creation of the gateway-specific command consumer links when a gateway issues a HTTP `ttd` request on behalf of a specific device.
That means that the above HTTP requests shall lead to command consumer links being created for the specific device, not the gateway. Furthermore, the HTTP adapter shall always set the last-known-gateway flag to the device itself, not the gateway. This effectively disables the gateway-mapping for commands targeted at the HTTP adapter.
Note that with this, there is the limitation that gateways connected via HTTP can't request commands for _any_ device connected to the gateway. The gateway must always send the `ttd` requests on behalf of a specific device. This should be noted in the documentation.
With Hono 1.1 and #1272 having been implemented, this issue doesn't arise and the above limitation can be removed.

(See #1644 for the AMQP/MQTT adapter scenario.)

calohmn:
This fixes #1645.
This prevents issues with commands not reaching the correct adapter instance in case of concurrent HTTP requests. 
The solution here is to effectively disable the command gateway mapping for HTTP adapters by:
- always creating a device specific command consumer link (even if a gateway is involved)
- always mapping the `last-known-gateway` to the device id itself, not the gateway.

This changed behaviour is only enabled for tenants having the new `support-concurrent-gateway-device-command-requests` option set to `true` in the `adapters/ext` field.

This change can probably be reverted again once #1272 is implemented and adapter-specific command consumer links are introduced.

### DIFF
--- a/adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapterTest.java
+++ b/adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapterTest.java
@@ -32,12 +32,14 @@ import static org.mockito.Mockito.when;
 import java.net.HttpURLConnection;
 
 import org.apache.qpid.proton.message.Message;
+import org.eclipse.hono.auth.Device;
 import org.eclipse.hono.client.ClientErrorException;
 import org.eclipse.hono.client.CommandConsumerFactory;
 import org.eclipse.hono.client.CommandContext;
 import org.eclipse.hono.client.CommandResponse;
 import org.eclipse.hono.client.CommandResponseSender;
 import org.eclipse.hono.client.CredentialsClientFactory;
+import org.eclipse.hono.client.DeviceConnectionClient;
 import org.eclipse.hono.client.DeviceConnectionClientFactory;
 import org.eclipse.hono.client.DownstreamSender;
 import org.eclipse.hono.client.DownstreamSenderFactory;
@@ -61,6 +63,7 @@ import org.eclipse.hono.service.resourcelimits.ResourceLimitChecks;
 import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.EventConstants;
 import org.eclipse.hono.util.MessageHelper;
+import org.eclipse.hono.util.RegistrationConstants;
 import org.eclipse.hono.util.TenantConstants;
 import org.eclipse.hono.util.TenantObject;
 import org.junit.Before;
@@ -69,6 +72,7 @@ import org.junit.Test;
 import org.junit.rules.Timeout;
 import org.junit.runner.RunWith;
 import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
 
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
@@ -83,6 +87,7 @@ import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.HttpServer;
 import io.vertx.core.http.HttpServerRequest;
 import io.vertx.core.http.HttpServerResponse;
+import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.unit.Async;
 import io.vertx.ext.unit.TestContext;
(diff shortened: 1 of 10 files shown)


service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -1021,7 +1021,9 @@ protected void onCommandConnectionEstablished(final HonoConnection commandConnec
             return Future.succeededFuture(registrationAssertion);
         }
 
-        final Future<String> gatewayId = getGatewayId(tenantId, deviceId, authenticatedDevice);
+        final Future<String> gatewayId = isGatewayMappingEnabled()
+                ? getGatewayId(tenantId, deviceId, authenticatedDevice)
+                : Future.succeededFuture(deviceId);
```
sophokles73: my understanding is that this is an optimization in order to prevent the expensive invocation of the (remote) Device Connection service which would return the device ID anyway because that is what the HTTP adapter has mapped the (origin) device ID to in the first place. Is this correct?

calohmn:
I've also added a corresponding note in the release notes.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
calohmn: The entry in the Device Connection service needs to be set here because an incoming command message (on the `command/{tenant_id}` address) that is ultimately directed at a HTTP gateway could have first reached an MQTT adapter for example. The MQTT adapter would do the gateway mapping then.
That is why we can't just disable the gateway mapping logic in the HTTP adapter but instead have to make sure that the mapping logic does the job of mapping the command to the target device id, not the gateway in this case.

site/homepage/content/release-notes.md
```diff
@@ -22,6 +22,13 @@ title = "Release Notes"
 * The Hono chart can now also be deployed using the recently released Helm 3. In fact,
   Helm 3 is now the recommended way for deploying Hono as it doesn't require installation
   of any Helm specific components to the Kubernetes cluster.
+* A potential issue has been resolved where some command messages might not have been sent
+  to the corresponding gateway. The scenario here involves the gateway sending event/telemetry
+  messages via HTTP with a `hono-ttd` header in order to receive commands, and doing so
+  with multiple concurrent requests for *different* devices. The solution makes it
+  necessary that gateways can no longer send a event/telemetry message via HTTP with a 
+  `hono-ttd` header in order to receive commands for *any* device that has last sent a
+  telemetry or event message via this gateway.
```
sophokles73: IMHO this constitutes a breaking API change, in this case the HTTP adapter's public API. I therefore do not think that we can include such a change in a service release. Another option would be to make this behavior change configurable per tenant with a default of the *existing, faulty* behavior.

calohmn: Right. I think introducing an option in the tenant data looks doable here. I'll explore this and update this PR.

calohmn: Ok, so we would introduce a new boolean flag:
- value `false` (default) means the behaviour stays the same as today: 
  - a gateway can still send *one* HTTP request to request a command from *any* device that is connected to the gateway
  - concurrent `ttd` requests from the gateway for different devices have the described issue that commands could get lost
- value `true` means
  - the concurrency issue doesn't exist (gateway-mapping is disabled)
  - but the gateway cannot send *one* HTTP request to request a command from *any* device

I would name the flag `support-concurrent-gateway-device-command-requests` (or `support-concurrent-gateway-device-command-subscriptions` or something like that).

The question is now where to put it in the tenant data.
I think we could put it in the `adapters` field of the tenant data.
If we define the option in the `Adapter` schema type, we would either need to specify in the documentation that this only applies to the HTTP adapter or we implement it for the other adapters as well (although for them it wasn't seen as that necessary - see #1644). (Alternative would be putting the option in the `adapters/ext` section, although this option is not really external here so I think it wouldn't fit.)
Or we use this as motivation to define a new generic `adapters/flags` (or something like that) container field, under which any kind of adapter-type specific flags shall go.
I would tend towards the last option. WDYT?

sophokles73: > Alternative would be putting the option in the adapters/ext section, although this option is not really external here so I think it wouldn't fit.

`ext` here stands for *extensions* ... FMPOV it the new property would be a good fit for this category as it only applies to the HTTP adapter and we intend to remove the flag anyway once we have moved on to single command receiver per protocol adapter instance.

We could simply document this extension property in the HTTP adapter admin guide.

calohmn: (I thought `ext` was reserved for extensions outside of the Hono API, ie. external to it. But I guess it's not that strict.)
Ok, yes, that's also the least invasive approach.

calohmn:
I've amended the commit so that the changed behaviour is only enabled for tenants having the new `support-concurrent-gateway-device-command-requests` option set to `true` in the `adapters/ext` field.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -1085,6 +1087,26 @@ protected void onCommandConnectionEstablished(final HonoConnection commandConnec
...
+     * <p>
+     * May be overridden by sub-classes to disable the gateway mapping for this adapter.
+     *
+     * @param tenantId The tenant that the device belongs to.
+     * @param deviceId The device to update the last known gateway for.
+     * @param authenticatedDevice The device that has authenticated to this protocol adapter.
+     * @return A succeeded future indicating the outcome of the operation. Its value will be {@code true} if gateway
+     *         mapping is enabled for this adapter.
+     */
+    protected Future<Boolean> isGatewayMappingEnabled(final String tenantId, final String deviceId,
```
sophokles73: how about declaring this *abtract* and adding final implementations to the standard adapters in order to prevent users from overriding it, which would make it harder to remove the method again once we have #1272 in place?

calohmn: Making it _abstract_ would impose the requirement to override the method for non-standard adapters. I would rather not have that (makes it also harder removing the method again).

As an alternative, we could therefore keep the method but still override it with final methods in the standard adapters (in accordance to the protocol adapter methods where the command consumer gets created usually being final as well).

calohmn: I've amended the commit accordingly (omitting `VertxBasedAmqpProtocolAdapter` since that class is final anyway).

tests/src/test/java/org/eclipse/hono/tests/http/HttpTestBase.java
```diff
@@ -920,6 +922,39 @@ public void testUploadMessagesWithTtdThatReplyWithCommand(
...
+    @MethodSource("commandAndControlVariants")
+    public void testUploadMessagesWithTtdThatReplyWithCommandWithGatewayMappingDisabled(
+            final HttpCommandEndpointConfiguration endpointConfig,
+            final VertxTestContext ctx) throws InterruptedException {
+
+        if (endpointConfig.isSubscribeAsGatewayForAllDevices()) {
+            logger.info("test skipped (GATEWAY_FOR_ALL_DEVICES not supported if gateway mapping disabled)");
+            ctx.completeNow();
+            return;
+        }
```
sophokles73: This looks like a hack to me ;-) IMHO we should simply have another method source for the parameters that this test supports ...

calohmn: Right, I've committed a cleaner solution.

tests/src/test/java/org/eclipse/hono/tests/http/HttpTestBase.java
```diff
@@ -230,6 +232,16 @@ public static void disconnect(final VertxTestContext ctx) {
...
 
+    /**
+     * Creates the endpoint configuration variants for Command &amp; Control scenarios where gateway mapping
+     * shall be disabled (the 'GATEWAY_FOR_ALL_DEVICES' case is not supported there).
+     *
+     * @return The configurations.
+     */
+    static Stream<HttpCommandEndpointConfiguration> commandAndControlVariantsForTestsWithGatewayMappingDisabled() {
+        return commandAndControlVariants().filter(endpointConfig -> !endpointConfig.isSubscribeAsGatewayForAllDevices());
+    }
```
sophokles73: :+1:

adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapterTest.java
```diff
@@ -787,6 +792,114 @@ public void testUploadTelemetryWithTtdClosesCommandConsumerIfSenderCreationFaile
...
+        adapter.uploadTelemetryMessage(ctx1, tenantId, deviceId);
+
+        // THEN the device receives a 202 response immediately
+        verify(response1).setStatusCode(202);
+        verify(response1).end();
+        // and the last known gateway has been set to the gateway
+        verify(devConClient).setLastKnownGatewayForDevice(eq(deviceId), eq(gatewayId), any());
+
+        // and the command consumer has been created with the gateway id
+        verify(commandConsumerFactory).createCommandConsumer(eq(tenantId), eq(deviceId), eq(gatewayId), any(Handler.class), any(Handler.class));
```
sophokles73: since this is all about making sure that one variant of the method is invoked and not the other, I would suggest to also check that the other variant has **not** been invoked:
```
verify(commandConsumerFactory, never()).createCommandConsumer(anyString(), anyString(), any(Handler.class), any(Handler.class));
```
and do the same below for the other variant.
WDYT?

calohmn: Yes, that makes it more explicit. I've changed the method accordingly.

PR #1666, opened 2019-12-12
[#1661, #504] Include MQTT client ID in Credentials.get request

### LINKED ISSUE #1661: MQTT adapter does not include MQTT client ID when retrieving credentials
The MQTT adapter is supposed to include a connecting device's MQTT client identifier in its request to invoke the Credentials API's *get* operation. This mechanism had been specified as part of making the MQTT adapter compatible with Eclipse Kapua's way of authenticating Eclipse Kura gateways.

The Credentials API provides means to include arbitrary *client context* information in the request to the *get* operation. However, the MQTT adapter does not include the MQTT client identifier yet.

sophokles73:
The MQTT adapter now always includes the device's MQTT client identifier
in the *client context* which is sent to the Credentials service in the
get request.

In order for this to work, the example Credentials service
implementation had to be changed so that it no longer asserts that a
device's credentials on record also contain the properties submitted in
the client context. This check doesn't seem to make much sense in the
context of the use case at hand anyway. The overall idea of including
the client context was not to have additional matching criteria for
credentials but to allow a Credentials service implementation to e.g.
use information from the client context as the device ID or
authentication identifier (as in the Kapua case)

Fixes #1661

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/auth/device/ExecutionContextAuthHandler.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/auth/device/ExecutionContextAuthHandler.java
@@ -72,24 +72,23 @@ public abstract class ExecutionContextAuthHandler<T extends ExecutionContext> im
 
         final Promise<DeviceUser> result = Promise.promise();
         parseCredentials(context)
-        .compose(authInfo -> {
-            final Promise<User> authResult = Promise.promise();
-            getAuthProvider(context).authenticate(authInfo, authResult);
-            return authResult.future();
-        })
-        .setHandler(authAttempt -> {
-            if (authAttempt.succeeded()) {
-                if (authAttempt.result() instanceof DeviceUser) {
-                    result.complete((DeviceUser) authAttempt.result());
-                } else {
-                    log.warn("configured AuthProvider does not return DeviceUser instances [type returned: {}",
-                            authAttempt.result().getClass().getName());
-                    result.fail(new ClientErrorException(HttpURLConnection.HTTP_UNAUTHORIZED));
-                }
-            } else {
-                result.fail(authAttempt.cause());
-            }
-        });
+                .compose(authInfo -> {
+                    final Promise<User> authResult = Promise.promise();
+                    getAuthProvider(context).authenticate(authInfo, authResult);
+                    return authResult.future();
+                }).setHandler(authAttempt -> {
+                    if (authAttempt.succeeded()) {
+                        if (authAttempt.result() instanceof DeviceUser) {
+                            result.complete((DeviceUser) authAttempt.result());
+                        } else {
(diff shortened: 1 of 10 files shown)


service-base/src/main/java/org/eclipse/hono/service/auth/device/ExecutionContextAuthHandler.java
```diff
@@ -61,25 +68,28 @@ public final AuthProvider getAuthProvider() {
     @Override
     public Future<DeviceUser> authenticateDevice(final T context) {
 
+        Objects.requireNonNull(context);
+
         final Promise<DeviceUser> result = Promise.promise();
         parseCredentials(context)
-                .compose(authInfo -> {
```
ctron: It is hard for me to spot any change here. To me the code formatting looks off to what we had in the past. If we want to change the formatting, then I think we should agree on that, and make this a dedicated changed. Rather than making this as part of an actual PR. Or at least confine it to a dedicated commit.

dejanb:
The original check is not that the credential on record also contain the property, but "if it contains. it's the same". So, it shouldn't affect this change. I agree it's not completely necessary from the functioning perspective of this implementation, but it is useful for testing (e.g. whether context is properly propagated and can be used in base classes)

sophokles73:
> The original check is not "whether the credential on record also contain the property", but "if it contains it, it's the same".

With the existing check, no connection will be established with any MQTT device if you do not register the MQTT device's client-id with the device credentials. FMPOV that doesn't make much sense because for normal operation in Hono, the MQTT client id is irrelevant.

If at all, the checking logic should be the other way around FMPOV: if extra properties have been registered with the device credentials, then we verify that the client context provided by the adapter includes these properties. Any additional properties in the client context are then ignored ...

That would allow the establishment of MQTT connections even without registering the MQTT client id for each device in the registry.

service-base/src/main/java/org/eclipse/hono/service/auth/device/ExecutionContextAuthHandler.java
sophokles73: That is true, our current code formatting rules seem to indeed enforce the original indentation.
I will revert the change. In general, do you think it is a good idea to automatically format the code according to our formatter rules on each save in the IDE?

dejanb:
I see now that check return false if property does not exists. I think the intent here was to have credential rejected if it contains the property that is not equal with one provided (so just changing that last case to true). 

But I agree that reversing the check also makes sense.

I don't have a strong opinion on that. Just think it might be useful to test the use case when we provide the context somehow.

sophokles73:
@dejanb Ok, then I will change the check according to your suggestion and add an integration test case for it.

sophokles73:
@dejanb I have pushed my changes, would you mind taking another look?

PR #1668, opened 2019-12-13
[#1515] Use configured initial credits in DestinationCommandConsumer

### LINKED ISSUE #1515: Command & Control: Give more than one credit for receiving commands for gateways
Currently, the AMQP and MQTT adapters will only return _one_ credit on the command consumer link once they have processed a command.

I think this may cause delays when applications are trying to send several commands at once, possibly directed at a gateway that consumes commands for several devices.

calohmn:
This fixes #1515.


calohmn:
Added a comment at `CommandContext.flow()` as that method has now become obsolete.

PR #1670, opened 2019-12-13
adds property 'memberOf' for handling gateway group memberships 

eriksven:
This PR adds a new 'memberOf' property to the devices in the device registry. The property indicates the membership with a gateway group that can be referenced in the 'via' property of another device. Further, this PR adds that the assert endpoint in the device registration endpoint resolves groups to its members in the 'via' property of the response message. Hence, this device registration endpoint does not return the registered groups.


bs-jokri:
The following test fails

```
[ERROR]   DeviceRegistrationAmqpIT.testAssertRegistrationSucceedsForDeviceViaGateway java.lang.AssertionError: 

Expecting:

  <["8f107d1d-355f-435b-bf2d-74b124b55f6c"]>

to contain exactly (and in same order):

  <["8f107d1d-355f-435b-bf2d-74b124b55f6c", "another-gateway"]>

but could not find the following elements:

  <["another-gateway"]>

[ERROR]   DeviceRegistrationJmsIT.testAssertRegistrationSucceedsForDeviceViaGateway java.lang.AssertionError: 

Expecting:

  <["be80d7d5-9f2f-4061-ad78-4cbae95815fd"]>

to contain exactly (and in same order):

  <["be80d7d5-9f2f-4061-ad78-4cbae95815fd", "another-gateway"]>

but could not find the following elements:

  <["another-gateway"]>
```

eriksven:
I took a quick look in what the caused the test to fail. The issue is the 'via' property in the response from the device registration endpoint (https://www.eclipse.org/hono/docs/api/device-registration/). 
For the 'via' property it says:

> The IDs (JSON strings) of gateways which may act on behalf of the device. (...)

For me, only registered gateways can (or may) act on behalf of the device. Because of that my implementation filters all ids from the 'via' list of the device that do not belong to a registered device. 
This filter also makes it easier to distinguish between ids for groups and for device in the 'via' property. 

The test in question lists the device 'another-gateway' in the 'via' property of the device and expects this 'another-gateway' in the response as well. Since in the test no device with the id 'another-gateway' is registered (and thus may act on behalf of the device), the implementation from this PR does not include 'another-gateway' in the response causing the test to fail.

WDYT? Is my interpretation and implementation to restrictive or can I adapt the test?

site/documentation/content/api/device-registration/index.md
```diff
@@ -60,7 +60,7 @@ In case of a successful invocation of the operation, the body of the response me
...
 | *device-id*      | *yes*     | *string*      | The ID of the device that is subject of the assertion. |
-| *via*            | *no*      | *array*       | The IDs (JSON strings) of gateways which may act on behalf of the device. This property MUST be set if any gateways are registered for the device. If the assertion request contained a *gateway_id* property and the response's *status* property has value `200` (indicating a successful assertion) then the array MUST at least contain the gateway ID from the request. |
+| *via*            | *no*      | *array*       | The IDs (JSON strings) of gateways which may act on behalf of the device. This property MUST be set if any gateways are registered for the device. This property does not include gateway groups that may be registered to act on behalf of the device. Instead, this property may contain devices that are member of one of the groups that is registered to act for the device. If the assertion request contained a *gateway_id* property and the response's *status* property has value `200` (indicating a successful assertion) then the array MUST at least contain the gateway ID from the request.
```
sophokles73: FMPOV this change is not necessary as the protocol adapters should not need to bother with gateway groups or any such thing. The whole concept of GW groups should be encapsulated in the device registry implementation FMPOV.
The adapters have been expecting to receive GW IDs and that is what they continue to do. If possible, I would like to confine the definition/mentioning of GW groups to the Device Registry Management API which is the place where I think we should start this discussion as it needs to be adapted to support the *memberOf* property etc, isn't it?

eriksven:
Yes the main point of mentioning of the gateway groups should be in the Device Registry Management API.  My hope was that this change would automatically be reflected in the Swagger UI of that API. 

My intention here was to emphasize that other services like the protocol adapters outside the device registry should not expect any gateway groups in the response from the Device Registration API  endpoint and that the gateway groups are resolved in the response. 

I see your point that this may cause more confusion than clarity when one does not deal with the group concept and thus will remove the remark.

sophokles73:
> Yes the main point of mentioning of the gateway groups should be in the Device Registry Management API. My hope was that this change would automatically be reflected in the Swagger UI of that API.

The swagger UI is created from `site/documentation/content/api/management/device-registry-v1.yaml`. FMPOV this is the place where you should start adding the new property ...

eriksven:
I made an error while updating the source branch for this PR. Because of that I re-opened another PR for this feature in #1683

PR #1671, opened 2019-12-16
Add Hotrod based DeviceConnectionClient

sophokles73:
The client implementation connects directly to an Infinispan data grid
to access device connection information. This client is an optimized
version of the existing client, skipping the Device Connection service
altogether. Note that this also means that the client can access all
data for all tenants and should therefore not be used with custom
protocol adapters.

### DIFF
--- a/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/DeviceConnectionInfoCache.java
+++ b/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/DeviceConnectionInfoCache.java
@@ -1,0 +1,67 @@
+/**
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.hono.deviceconnection.infinispan.client;
+
+import io.opentracing.SpanContext;
+import io.vertx.core.Future;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * A repository for keeping connection information about devices.
+ *
+ */
+public interface DeviceConnectionInfoCache {
+
+    /**
+     * Sets the gateway that last acted on behalf of a device.
+     * <p>
+     * If a device connects directly instead of through a gateway, the device identifier itself is to be used as value
+     * for the <em>gatewayId</em> parameter.
+     *
+     * @param tenant The tenant that the device belongs to.
+     * @param deviceId The device identifier.
+     * @param gatewayId The gateway identifier. This may be the same as the device identifier if the device is
+     *                  (currently) not connected via a gateway but directly to a protocol adapter.
+     * @param context The currently active OpenTracing span or {@code null} if no span is currently active.
+     *            Implementing classes should use this as the parent for any span they create for tracing
+     *            the execution of this operation.
+     * @return A future indicating the outcome of the operation.
+     *         <p>
+     *         The future will be succeeded if the device connection information has been updated.
+     *         Otherwise the future will be failed with a {@link org.eclipse.hono.client.ServiceInvocationException}.
+     * @throws NullPointerException if tenant, device id or gateway id are {@code null}.
+     */
+    Future<Void> setLastKnownGatewayForDevice(String tenant, String deviceId, String gatewayId, SpanContext context);
+
+    /**
+     * Gets the gateway that last acted on behalf of a device.
+     * <p>
+     * If no last known gateway has been set for the given device yet, a failed future with status
+     * <em>404</em> is returned.
+     *
+     * @param tenant The tenant that the device belongs to.
+     * @param deviceId The device identifier.
+     * @param context The currently active OpenTracing span or {@code null} if no span is currently active.
+     *            Implementing classes should use this as the parent for any span they create for tracing
+     *            the execution of this operation.
+     * @return A future indicating the outcome of the operation.
+     *         <p>
+     *         The future will be succeeded with a JSON object containing the currently mapped gateway ID
+     *         in the <em>gateway-id</em> property, if device connection information has been found for
+     *         the given device.
+     *         Otherwise the future will be failed with a {@link org.eclipse.hono.client.ServiceInvocationException}.
+     * @throws NullPointerException if tenant or device id are {@code null}.
+     */
+    Future<JsonObject> getLastKnownGatewayForDevice(String tenant, String deviceId, SpanContext context);
+}
--- a/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodBasedDeviceConnectionClient.java
+++ b/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodBasedDeviceConnectionClient.java
@@ -14,11 +14,9 @@
 
 package org.eclipse.hono.deviceconnection.infinispan.client;
 
-import java.net.HttpURLConnection;
 import java.util.Objects;
 
 import org.eclipse.hono.client.DeviceConnectionClient;
-import org.eclipse.hono.client.ServerErrorException;
 
 import io.opentracing.SpanContext;
 import io.vertx.core.AsyncResult;
@@ -65,7 +63,7 @@ public final class HotrodBasedDeviceConnectionClient implements DeviceConnection
      */
     @Override
     public boolean isOpen() {
-        return cache != null;
+        return true;
     }
 
     /**
@@ -106,11 +104,7 @@ public final class HotrodBasedDeviceConnectionClient implements DeviceConnection
     @Override
     public Future<Void> setLastKnownGatewayForDevice(final String deviceId, final String gatewayId, final SpanContext context) {
 
-        if (isOpen()) {
-            return cache.setLastKnownGatewayForDevice(tenantId, deviceId, gatewayId, context);
-        } else {
-            return Future.failedFuture(new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "no connection to remote cache"));
-        }
+        return cache.setLastKnownGatewayForDevice(tenantId, deviceId, gatewayId, context);
     }
 
     /**
@@ -119,10 +113,6 @@ public final class HotrodBasedDeviceConnectionClient implements DeviceConnection
     @Override
     public Future<JsonObject> getLastKnownGatewayForDevice(final String deviceId, final SpanContext context) {
 
-        if (isOpen()) {
-            return cache.getLastKnownGatewayForDevice(tenantId, deviceId, context);
-        } else {
-            return Future.failedFuture(new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "no connection to remote cache"));
-        }
+        return cache.getLastKnownGatewayForDevice(tenantId, deviceId, context);
     }
 }
--- a/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodBasedDeviceConnectionClient.java
+++ b/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodBasedDeviceConnectionClient.java
@@ -1,0 +1,128 @@
+/**
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+
+package org.eclipse.hono.deviceconnection.infinispan.client;
+
+import java.net.HttpURLConnection;
+import java.util.Objects;
+
+import org.eclipse.hono.client.DeviceConnectionClient;
+import org.eclipse.hono.client.ServerErrorException;
+
+import io.opentracing.SpanContext;
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.json.JsonObject;
+
+
+/**
+ * A client for accessing device connection information in a data grid using the
+ * Hotrod protocol.
+ *
+ */
+public final class HotrodBasedDeviceConnectionClient implements DeviceConnectionClient {
+
+    final String tenantId;
+    final DeviceConnectionInfoCache cache;
+
+    /**
+     * Creates a client for accessing device connection information.
+     * 
+     * @param tenantId The tenant that this client is scoped to.
+     * @param cache The remote cache that contains the data.
+     */
+    public HotrodBasedDeviceConnectionClient(final String tenantId, final DeviceConnectionInfoCache cache) {
+        this.tenantId = Objects.requireNonNull(tenantId);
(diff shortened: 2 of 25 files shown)


service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -855,33 +880,36 @@ protected void onCommandConnectionEstablished(final HonoConnection commandConnec
...
-                .orElse(Future.failedFuture(new ServerErrorException(
-                        HttpURLConnection.HTTP_UNAVAILABLE, "Device Connection client factory is not set")));
-        return CompositeFuture.all(tenantCheck, registrationCheck, credentialsCheck, messagingCheck, commandCheck,
-                deviceConnectionCheck).mapEmpty();
+                        HttpURLConnection.HTTP_UNAVAILABLE, "Command & Control client factory is not set"))));
+        if (deviceConnectionClientFactory instanceof DeviceConnectionClientFactory) {
+            connections.add(Optional.ofNullable(deviceConnectionClientFactory)
+                    .map(client -> ((DeviceConnectionClientFactory) client).isConnected())
+                    .orElse(Future.failedFuture(new ServerErrorException(
+                            HttpURLConnection.HTTP_UNAVAILABLE, "Device Connection client factory is not set"))));
```
calohmn: Block can be simplified since `deviceConnectionClientFactory` can't be null here.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/DeviceConnectionInfoCache.java
```diff
@@ -0,0 +1,67 @@
...
+     *            the execution of this operation.
+     * @return A future indicating the outcome of the operation.
+     *         <p>
+     *         The future will be succeeded with a JSON object containing the currently mapped gateway ID
+     *         in the <em>gateway-id</em> property, if device connection information has been found for
+     *         the given device.
+     *         Otherwise the future will be failed with a {@link org.eclipse.hono.client.ServiceInvocationException}.
+     * @throws NullPointerException if tenant or device id are {@code null}.
+     */
+    Future<JsonObject> getLastKnownGatewayForDevice(String tenant, String deviceId, SpanContext context);
```
ctron: Wouldn't it make more sense to return `Optional<JsonObject>`, and indicate a missing entry by returning an empty optional? Creating an exception always has the overhead of filling in the stack trace. And "not finding" anything in cache isn't necessarily a failure either.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodBasedDeviceConnectionClient.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     * 
+     * @return {@code true} if this client is connected to the data grid.
+     */
+    @Override
+    public boolean isOpen() {
+        return cache != null;
```
ctron: I fail to see how `cache` could ever be `null`.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodBasedDeviceConnectionClientFactory.java
```diff
@@ -0,0 +1,107 @@
...
+        cache.start();
+        LOG.info("successfully connected to remote cache");
+    }
+
+    /**
+     * Shuts down the factory and releases all resources.
+     */
+    @PreDestroy
+    public void stop() {
+        clients.invalidateAll();
```
ctron: Should `close` be called on the clients?

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodBasedDeviceConnectionInfoCache.java
```diff
@@ -0,0 +1,123 @@
...
+     * {@inheritDoc}
+     * 
+     * If this method is invoked from a vert.x Context, then the returned future will be completed on that context.
+     */
+    @Override
+    public Future<Void> setLastKnownGatewayForDevice(final String tenantId, final String deviceId, final String gatewayId, final SpanContext context) {
+
+        final Promise<Void> result = Promise.promise();
+
+        if (cache == null) {
```
ctron: I fail to see how this could ever be `null`.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodBasedDeviceConnectionInfoCache.java
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Future<JsonObject> getLastKnownGatewayForDevice(final String tenantId, final String deviceId, final SpanContext context) {
+
+        final Promise<JsonObject> result = Promise.promise();
+
+        if (cache == null) {
```
ctron: I fail to see how this could ever be `null`.

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
```diff
@@ -89,19 +87,22 @@ protected void doStart(final Future<Void> startFuture) {
...
-                cache = cacheManager.getCache("device-connection");
-                cache.start();
+                final BasicCache<String, String> remoteCache = cacheManager.getCache("device-connection");
+                remoteCache.start();
+                cache = new HotrodBasedDeviceConnectionInfoCache(remoteCache);
                 r.complete(cacheManager);
             }, attempt -> {
                 if (attempt.succeeded()) {
-                    LOG.info("successfully connected to remote cache");
+                    log.info("successfully connected to remote cache");
```
ctron: I don't understand why the logger is renamed from `LOG` to `log` here, where in other files of the PR a new logger named `LOG`. Is being created.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -855,33 +880,36 @@ protected void onCommandConnectionEstablished(final HonoConnection commandConnec
...
-                        HttpURLConnection.HTTP_UNAVAILABLE, "Command & Control client factory is not set")));
-        final Future<Void> deviceConnectionCheck = Optional.ofNullable(deviceConnectionClientFactory)
-                .map(client -> client.isConnected())
-                .orElse(Future.failedFuture(new ServerErrorException(
-                        HttpURLConnection.HTTP_UNAVAILABLE, "Device Connection client factory is not set")));
-        return CompositeFuture.all(tenantCheck, registrationCheck, credentialsCheck, messagingCheck, commandCheck,
-                deviceConnectionCheck).mapEmpty();
+                        HttpURLConnection.HTTP_UNAVAILABLE, "Command & Control client factory is not set"))));
+        if (deviceConnectionClientFactory instanceof DeviceConnectionClientFactory) {
+            connections.add(Optional.ofNullable(deviceConnectionClientFactory)
```
ctron: I don't think `deviceConnectionClientFactory` can be `null` here.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/DeviceConnectionInfoCache.java
sophokles73: Then we would not be able to pass on any (potential) exception that occurred while accessing the (remote) cache. On the other hand, there is not much that client code of this method could do with this exception anyway, right? We already pass in a SpanContext so an implementation can already properly log the problem. That would probably also account for the *setLastKnownGateway* method as well, and we should change its return value to `void` accordingly, shouldn't we?

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
sophokles73: It is not renamed, *log* is simply the inherited logger from the super class. That logger is not static final, thus its name is lower case ...

sophokles73:
> One major item which seems to be missing is that there is not health check if the hotrod connection is still alive. This can be done with e.g. a call to org.infinispan.client.hotrod.RemoteCache.serverStatistics(), which would fail in that case.

I actually had thought about this but didn't find out how to do it. Thanks for the hint @ctron :+1:

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodBasedDeviceConnectionClientFactory.java
sophokles73: To be honest: i have no clue. I haven't been able to find any documentation regarding if and when to invoke this method on a cache. Have you found any such resources and can point me into the right direction?

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodBasedDeviceConnectionInfoCache.java
```diff
@@ -0,0 +1,123 @@
...
+                    result.complete();
+                } else {
+                    LOG.debug("failed to set last known gateway [tenant: {}, device-id: {}, gateway: {}]",
+                            tenantId, deviceId, gatewayId, error);
+                    result.fail(new ServerErrorException(HttpURLConnection.HTTP_INTERNAL_ERROR, error));
+                }
+            });
+        }
+
+        return HonoProtonHelper.handleOnContext(result, Vertx.currentContext());
```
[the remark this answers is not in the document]
sophokles73: I agree. Any proposal for an alternative location for that helper method?

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
```diff
@@ -161,18 +162,9 @@ public void setLastKnownGatewayForDevice(
         if (cache == null) {
             resultHandler.handle(Future.failedFuture(new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "no connection to remote cache")));
```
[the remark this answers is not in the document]
sophokles73: this will be logged to the span by the super class already if I am not mistaken

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
ctron: Still, it is weird that some classes use uppercase loggers, were this one now uses lowercase.

sophokles73: Well, I guess we're just following field naming rules here. Constants are all upper case, instance fields are lower case, right?

ctron:
> According to the JavaDocs of `RemoteCache.serverStatistics()`, this method never seems to fail ...

I wouldn't bet on that.

sophokles73:
> I wouldn't bet on that.

Me neither :-) That is just my interpretation of the lack of any exceptions being declared in the JavaDoc. However, given the overall *quality* of the JavaDocs I guess you cannot assume anything beyond the specific code in implementing classes. So, what's your take? Which exceptions should we check for in order to determine if the connection is alive?

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
```diff
@@ -161,18 +162,9 @@ public void setLastKnownGatewayForDevice(
...
-                    resultHandler.handle(Future.succeededFuture(DeviceConnectionResult.from(HttpURLConnection.HTTP_NO_CONTENT)));
-                } else {
-                    log.debug("failed to set last known gateway [tenant: {}, device-id: {}, gateway: {}]",
-                            tenantId, deviceId, gatewayId, error);
-                    resultHandler.handle(Future.failedFuture(new ServerErrorException(HttpURLConnection.HTTP_INTERNAL_ERROR, error)));
-                }
-            });
+            cache.setLastKnownGatewayForDevice(tenantId, deviceId, gatewayId, span.context())
+            .map(ok -> DeviceConnectionResult.from(HttpURLConnection.HTTP_NO_CONTENT))
+            .setHandler(resultHandler);
```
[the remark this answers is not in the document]
sophokles73: same here

sophokles73:
@ctron @calohmn I have pushed some changes. Would you mind taking another look?

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodBasedDeviceConnectionClientFactory.java
sophokles73: Ah, I got confused here. You are referring to the `HotrodBasedDeviceConnectionClient` instances in the Caffeine cache. Since the *close* method of that class doesn't do anything anyway, I guess we can skip it.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodBasedDeviceConnectionInfoCache.java
```diff
@@ -0,0 +1,123 @@
...
+
+    final BasicCache<String, String> cache;
+
+    /**
+     * Creates a client for accessing device connection information.
+     * 
+     * @param cache The remote cache that contains the data.
+     */
+    public HotrodBasedDeviceConnectionInfoCache(final BasicCache<String, String> cache) {
+        this.cache = Objects.requireNonNull(cache);
```
kaniyan: IMHO this constructor throws `NullPointerException`, if the _cache_ is _null_ and this info could be added in the _javadoc_.

ctron:
> So, what's your take?

I simply catch all exceptions and threat them as a KO on the health check. The same for `null` results:

https://github.com/EnMasseProject/enmasse/blob/f05d81d3a43cb6a48cf7a664570c89640cbe996d/iot/iot-device-registry-infinispan/src/main/java/io/enmasse/iot/registry/infinispan/health/AbstractCacheHealthCheck.java#L30-L39

For now, this worked.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -197,27 +200,39 @@ public final TenantClientFactory getTenantClientFactory() {
...
      * Gets a client for interacting with the Device Connection service.
      *
      * @param tenantId The tenant that the client is scoped to.
      * @return The client.
+     * @throws IllegalStateException if no client factory is set.
      */
     protected final Future<DeviceConnectionClient> getDeviceConnectionClient(final String tenantId) {
-        return getDeviceConnectionClientFactory().getOrCreateDeviceConnectionClient(tenantId);
+
+        if (deviceConnectionClientFactory == null) {
```
b-abel: When could this be `null`? Before `startInternal()` is called?

sophokles73: yes. I do not know of any other formal way of expressing that setting this property is required for this object functioning properly ...

sophokles73:
closed in favor of #1727

Issue #1346, opened 2019-07-10
Do not use vert.x EventBus in Device Registry anymore


ctron:
As discussed in today's community call, we want to think about dropping the use of the vertx event bus, between the communication endpoints and the service interfaces.

The proposal is to directly wire up the communication endpoint (like the HTTP device management endpoint, or the AMQP registration endpoint) with the service that implements the functionality.

This would reduce the complexity and the code we need to maintain. Also should this improve the performance, as we no longer need to channel everything through the even bus.

jbtrystram:
I can't self-assign this i as I am not (yet?) a commiter, but I am working on this :)

sophokles73:
@dejanb @ctron @calohmn I agree that this would save some code. I wonder, though, if it also removes some of the flexibility we get by decoupling the service implementation from the endpoints via the event bus.
In the current setup we are able to scale up the endpoints independently from the service implementation, i.e. we can have four Verticle instances of the HTTP endpoint while keeping just one instance of the service implementation. In case of the file based registry, this is important because we cannot have multiple instances of it. Otherwise we would need to use the same service implementation instance in all of the HTTP endpoint Verticle instances and start synchronizing method invocations.

I do not really think that the event bus really is a bottleneck (at least for now). Giving up the (deployment) flexibility it provides thus should be considered carefully FMPOV.

I just wanted to bring this up so that we can weigh the pros against the cons and make an informed decision.

dejanb:
@sophokles73 Do you have any particular numbers/experience in scaling file-based registry using event bus? I'm not totally sure it can improve things dramatically in that scenario. Also, how focused should we be on scalability of this use case, on behalf of optimizing the code for production-ready implementations?
IMHO, we should keep the service implementations straight-forward and leave scaling out to deployment (k8s) layer.

sophokles73:
@dejanb I do not think that we have any issue at all with the scalability of the file based example registry, nor does the example registry profit from multiple HTTP endpoint verticle instances.

> IMHO, we should keep the service implementations straight-forward and leave scaling out to deployment (k8s) layer.

Do I understand you correctly in that we should not care about vertical scaling (i.e. using multiple verticle instances) but instead focus on horizontal scaling (using multiple pods) only?

dejanb:
Yeah, that gave us the best results so far in our tests. I don't think event bus adds much benefit in the architecture we are using. But adds some overhead in code and runtime execution path.

sophokles73:
Fair enough

[...]

PR #1672, opened 2019-12-16
[#1346] Remove event bus from device management endpoint

dejanb:
This is the first PR aiming to remove event bus from endpoint implementations. The idea is to do it step by step, so that we can review changes more easily in series of smaller PRs. More code improvements (and removal) will be done once all HTTP endpoints are refactored.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
@@ -23,6 +23,7 @@ import java.util.function.IntPredicate;
 
 import org.eclipse.hono.client.ClientErrorException;
 import org.eclipse.hono.service.AbstractEndpoint;
+import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.EventBusMessage;
 import org.eclipse.hono.util.MessageHelper;
@@ -31,6 +32,8 @@ import org.eclipse.hono.util.Strings;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 
+import io.opentracing.Span;
+import io.opentracing.tag.Tags;
 import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
 import io.vertx.core.eventbus.DeliveryOptions;
@@ -346,6 +349,47 @@ public abstract class AbstractHttpEndpoint<T> extends AbstractEndpoint implement
     protected final String getDeviceIdParam(final RoutingContext ctx) {
         return ctx.request().getParam(PARAM_DEVICE_ID);
     }
+
+    /**
+     * Get request parameter value and check if it has been set. If it's not set, fail the request.
+     *
+     * @param paramName The name of the parameter to get.
+     * @param ctx The routing context of the request.
+     * @param span The active OpenTracing span for this operation. In case of the missing mandatory parameter, the error is logged and the span is finished.
+     *             Otherwise, the parameter is set as a span tag.
+     * @return The value of the parameter if it's set or {@code null} otherwise.
+     * @throws NullPointerException If ctx or paramName are {@code null}.
+     */
+    protected final String getMandatoryRequestParam(final String paramName, final RoutingContext ctx, final Span span) {
+        return getRequestParam(paramName, ctx, span, false);
+    }
+
+    /**
+     * Get request parameter value. Optionally, if parameter has not been set, fail the request.
+     *
+     * @param paramName The name of the parameter to get.
+     * @param ctx The routing context of the request.
+     * @param span The active OpenTracing span for this operation. In case of the missing mandatory parameter, the error is logged and the span is finished.
+     *             Otherwise, the parameter is set as a span tag.
+     * @param optional Whether to check if parameter has been set or not.
+     * @return The value of the parameter if it's set or {@code null} otherwise.
+     * @throws NullPointerException If ctx or paramName are {@code null}.
+     */
+    protected final String getRequestParam(final String paramName, final RoutingContext ctx, final Span span, final boolean optional) {
+        final String value = ctx.request().getParam(paramName);
+        if (!optional && value == null) {
+            final String msg = String.format("Missing request parameter: %s", paramName);
+            TracingHelper.logError(span, msg);
+            HttpUtils.badRequest(ctx, msg);
+            Tags.HTTP_STATUS.set(span, HttpURLConnection.HTTP_BAD_REQUEST);
+            span.finish();
+            return null;
+        } else {
+            span.setTag(paramName, value);
+            return value;
+        }
+    }
+
 
     /**
      * Check the ETags values from the HTTP if-Match header if they exist,
--- a/service-base/src/main/java/org/eclipse/hono/service/management/Util.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/management/Util.java
@@ -13,8 +13,10 @@
 
 package org.eclipse.hono.service.management;
 
+import java.net.HttpURLConnection;
 import java.util.Objects;
 
+import org.eclipse.hono.service.http.HttpUtils;
 import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.MessageHelper;
 
@@ -22,6 +24,11 @@ import io.opentracing.Span;
 import io.opentracing.SpanContext;
 import io.opentracing.Tracer;
 import io.opentracing.tag.Tags;
+import io.vertx.core.Handler;
+import io.vertx.core.http.HttpHeaders;
+import io.vertx.core.http.HttpServerResponse;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.RoutingContext;
 
 /**
  * Utility class for the management HTTP API.
@@ -32,6 +39,22 @@ public class Util {
      * Prevent instantiation.
      */
     private Util(){}
+
+    /**
+     * Creates a new <em>OpenTracing</em> span for tracing the execution of a service operation.
+     *
+     * @param operationName The operation name that the span should be created for.
+     * @param spanContext Existing span context.
+     * @param tracer the Tracer instance.
+     * @param className The class name to insert in the Span.
+     * @return The new {@code Span}.
+     * @throws NullPointerException if operationName is {@code null}.
+     */
+    public static final Span newChildSpan(final String operationName, final SpanContext spanContext,
+            final Tracer tracer, final String className) {
+        return newChildSpan(operationName, spanContext, tracer, null, null, className);
+
+    }
 
     /**
      * Creates a new <em>OpenTracing</em> span for tracing the execution of a service operation.
@@ -82,6 +105,57 @@ public class Util {
             final Tracer tracer,
             final String tenantId, final String className) {
         return newChildSpan(operationName, spanContext, tracer, tenantId, null, className);
+    }
+
+    /**
+     * Creates a response based on generic result.
+     * <p>
+     * The behavior is as follows:
+     * <ol>
+     * <li>Set the status code on the response.</li>
+     * <li>If the status code represents an error condition (i.e. the code is &gt;= 400),
+     * then the JSON object passed in to the returned handler is written to the response body.</li>
+     * <li>Otherwise, the JSON object is written to the response body and the given custom handler is
+     * invoked (if not {@code null}).</li>
+     * </ol>
+     *
+     * @param ctx The routing context of the request.
+     * @param result The generic result of the operation.
+     * @param customHandler An (optional) handler for post processing the HTTP response, e.g. to set any additional HTTP
+     *                      headers. The handler <em>must not</em> write to response body. May be {@code null}.
+     * @param span The active OpenTracing span for this operation. The status of the response is logged and span is finished.
+     */
+    public static final void writeResponse(final RoutingContext ctx, final Result<?> result, final Handler<HttpServerResponse> customHandler, final Span span) {
+        final int status = result.getStatus();
+        final HttpServerResponse response = ctx.response();
+        response.setStatusCode(status);
+        if (status >= 400) {
+            HttpUtils.setResponseBody(response, JsonObject.mapFrom(result.getPayload()));
+        } else if (status == HttpURLConnection.HTTP_CREATED) {
+            HttpUtils.setResponseBody(response, JsonObject.mapFrom(result.getPayload()));
+            if (customHandler != null) {
+                customHandler.handle(response);
+            }
+        }
+        Tags.HTTP_STATUS.set(span, status);
+        span.finish();
+        response.end();
+    }
+
+    /**
+     * Creates a response based on operation result (including resource version).
+     * <p>
+     * Sets ETAG header and then calls {@link #writeResponse}
+     *
+     * @param ctx The routing context of the request.
+     * @param result The operation result of the operation.
+     * @param customHandler An (optional) handler for post processing the HTTP response, e.g. to set any additional HTTP
+     *                      headers. The handler <em>must not</em> write to response body. May be {@code null}.
+     * @param span The active OpenTracing span for this operation. The status of the response is logged and span is finished.
+     */
+    public static final void writeOperationResponse(final RoutingContext ctx, final OperationResult<?> result, final Handler<HttpServerResponse> customHandler, final Span span) {
+        result.getResourceVersion().ifPresent(v -> ctx.response().putHeader(HttpHeaders.ETAG, v));
+        writeResponse(ctx, result, customHandler, span);
     }
 
 }
--- a/service-base/src/main/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementHttpEndpoint.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementHttpEndpoint.java
@@ -1,0 +1,251 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.service.management.device;
+
+import java.net.HttpURLConnection;
+import java.util.EnumSet;
+import java.util.Optional;
+import org.eclipse.hono.client.ClientErrorException;
+import org.eclipse.hono.config.ServiceConfigProperties;
+import org.eclipse.hono.service.http.AbstractHttpEndpoint;
+import org.eclipse.hono.service.http.HttpUtils;
+import org.eclipse.hono.service.http.TracingHandler;
+import org.eclipse.hono.service.management.Id;
+import org.eclipse.hono.service.management.OperationResult;
+import org.eclipse.hono.service.management.Result;
+import org.eclipse.hono.service.management.Util;
+import org.eclipse.hono.tracing.TracingHelper;
+import org.eclipse.hono.util.RegistryManagementConstants;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import io.opentracing.Span;
+import io.opentracing.tag.Tags;
+import io.vertx.core.Promise;
+import io.vertx.core.Vertx;
+import io.vertx.core.http.HttpHeaders;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.core.http.HttpServerResponse;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.Router;
+import io.vertx.ext.web.RoutingContext;
+
+/**
+ * An {@code HttpEndpoint} for managing device registration information.
+ * <p>
+ * This endpoint implements Hono's <a href="https://www.eclipse.org/hono/docs/api/device-registration/">Device
+ * Registration API</a>. It receives HTTP requests representing operation invocations and sends them to an address on
+ * the vertx event bus for processing. The outcome is then returned to the peer in the HTTP response.
+ */
+public abstract class AbstractDeviceManagementHttpEndpoint extends AbstractHttpEndpoint<ServiceConfigProperties> {
+
+    private static final String SPAN_NAME_CREATE_DEVICE = "create Device from management API";
+    private static final String SPAN_NAME_GET_DEVICE = "get Device from management API";
+    private static final String SPAN_NAME_UPDATE_DEVICE = "update Device from management API";
+    private static final String SPAN_NAME_REMOVE_DEVICE = "remove Device from management API";
+
+    private static final String DEVICE_MANAGEMENT_ENDPOINT_NAME = String.format("%s/%s",
+                    RegistryManagementConstants.API_VERSION,
+                    RegistryManagementConstants.DEVICES_HTTP_ENDPOINT);
+
+    /**
+     * Creates an endpoint for a Vertx instance.
+     *
+     * @param vertx The Vertx instance to use.
+     * @throws NullPointerException if vertx is {@code null};
+     */
+    @Autowired
+    public AbstractDeviceManagementHttpEndpoint(final Vertx vertx) {
+        super(vertx);
+    }
+
+    /**
+     * Returns an empty String as this implementation does not use event bus.
+     */
+    @Override
+    protected String getEventBusAddress() {
+        return "";
+    }
+
+    @Override
+    public String getName() {
+        return DEVICE_MANAGEMENT_ENDPOINT_NAME;
+    }
+
+    @Override
+    public void addRoutes(final Router router) {
+
+        final String pathWithTenant = String.format("/%s/:%s", getName(), PARAM_TENANT_ID);
+        final String pathWithTenantAndDeviceId = String.format("/%s/:%s/:%s", getName(), PARAM_TENANT_ID,
+                PARAM_DEVICE_ID);
+
+        // Add CORS handler
+        router.route(pathWithTenant).handler(createCorsHandler(config.getCorsAllowedOrigin(), EnumSet.of(HttpMethod.POST)));
+        router.route(pathWithTenantAndDeviceId).handler(createDefaultCorsHandler(config.getCorsAllowedOrigin()));
+
+
+        // CREATE device with auto-generated deviceID
+        router.post(pathWithTenant)
+                .handler(this::extractOptionalJsonPayload)
+                .handler(this::doCreateDevice);
+
+        // CREATE device
+        router.post(pathWithTenantAndDeviceId)
+                .handler(this::extractOptionalJsonPayload)
+                .handler(this::doCreateDevice);
+
+        // GET device
+        router.get(pathWithTenantAndDeviceId)
+                .handler(this::doGetDevice);
+
+        // UPDATE existing device
+        router.put(pathWithTenantAndDeviceId)
+                .handler(this::extractRequiredJsonPayload)
+                .handler(this::extractIfMatchVersionParam)
+                .handler(this::doUpdateDevice);
+
+        // DELETE device
+        router.delete(pathWithTenantAndDeviceId)
+                .handler(this::extractIfMatchVersionParam)
+                .handler(this::doDeleteDevice);
+    }
+
+    /**
+     * The service to forward requests to.
+     *
+     * @return The service to bind to, must never return {@code null}.
+     */
+    protected abstract DeviceManagementService getService();
+
+    private void doGetDevice(final RoutingContext ctx) {
+
+        final Span span = Util.newChildSpan(SPAN_NAME_GET_DEVICE, TracingHandler.serverSpanContext(ctx), tracer, getClass().getSimpleName());
+
+        final String deviceId = getMandatoryRequestParam(PARAM_DEVICE_ID, ctx, span);
+        final String tenantId = getMandatoryRequestParam(PARAM_TENANT_ID, ctx, span);
+
+        final HttpServerResponse response = ctx.response();
+
+        logger.debug("retrieving device [{}] of tenant [{}]", deviceId, tenantId);
+        final Promise<OperationResult<Device>> result = Promise.promise();
+        result.future().setHandler(handler -> {
+            final OperationResult<Device> operationResult = handler.result();
+            final int status = operationResult.getStatus();
+            response.setStatusCode(status);
+            switch (status) {
+                case HttpURLConnection.HTTP_OK:
+                    operationResult.getResourceVersion().ifPresent(v -> response.putHeader(HttpHeaders.ETAG, v));
+                    HttpUtils.setResponseBody(response, JsonObject.mapFrom(operationResult.getPayload()));
+                    // falls through intentionally
+                default:
+                    Tags.HTTP_STATUS.set(span, status);
+                    span.finish();
+                    response.end();
+            }
+        });
+
+        getService().readDevice(tenantId, deviceId, span, result);
+    }
+
+    private void doCreateDevice(final RoutingContext ctx) {
+
+        final Span span = Util.newChildSpan(SPAN_NAME_CREATE_DEVICE, TracingHandler.serverSpanContext(ctx), tracer, getClass().getSimpleName());
+
+        final String tenantId = getMandatoryRequestParam(PARAM_TENANT_ID, ctx, span);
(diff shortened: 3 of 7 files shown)


sophokles73:
@dejanb The example registry deploys one verticle for the AMQP 1.0 based endpoints (facing the adapters) and one for the HTTP based endpoints (for device management). Both interact with the *same* service implementation Verticle(s). If we remove the event bus between these components and inject the same service implementation instance(s) into both, the AMQP endpoint Verticle and the HTTP endpoint Verticle, then we would need to start synchronizing access to the service implementation methods. Is that what you have in mind? Otherwise we would need to expose both the  AMQP endpoints as well as the HTTP endpoints in the same single Verticle, right?

dejanb:
@sophokles73 I'm not sure this is a valid concern. I might be missing something, but I don't think event bus consumer handlers are synchronized, so the same would be the problem in the existing code? Also, two endpoints use different interfaces, so they are not competing with each other. As I said, I might have a blind spot, so please let me know if you see that we're introducing the problem with this refactor.

sophokles73:
> I might be missing something, but I don't think event bus consumer handlers are synchronized, so the same would be the problem in the existing code?

The messages will be received over the event bus one-by-one in the implementation class and are guaranteed to always be processed on the same event loop thread. Thus, we do not need to synchronize our code in this case. So, yes, the event bus (implicitly) makes sure that the consumer code is not invoked concurrently.

> Also, two endpoints use different interfaces, so they are not competing with each other.

The `DeviceRegistryAmqpServer` and the `DeviceRegistryRestServer` are both Verticles. When instances of them are deployed to vert.x, they get assigned an event loop thread on which they will process all requests. In most cases, these two event loop threads will be different. Thus, if the server Verticle instances both invoke methods on the same service implementation instance (e.g. FileBasedRegistrationService), then they will potentially do so concurrently from different threads ...

dejanb:
> The messages will be received over the event bus one-by-one in the implementation class and are guaranteed to always be processed on the same event loop thread. Thus, we do not need to synchronize our code in this case. So, yes, the event bus (implicitly) makes sure that the consumer code is not invoked concurrently.

I see your point. That's assuming we're having only a single consumer, which is the case here as we'll always have a single vertical of device registry service implementation.


> > Also, two endpoints use different interfaces, so they are not competing with each other.
> 
> The `DeviceRegistryAmqpServer` and the `DeviceRegistryRestServer` are both Verticles. When instances of them are deployed to vert.x, they get assigned an event loop thread on which they will process all requests. In most cases, these two event loop threads will be different. Thus, if the server Verticle instances both invoke methods on the same service implementation instance (e.g. FileBasedRegistrationService), then they will potentially do so concurrently from different threads ...

What I meant here is that AMQP server will only ever invoke methods of `RegistrationService` interface, while REST server will exclusively use `DeviceManagementService` methods.

This doesn't solve the problem of multiple REST server verticles doing invoking methods concurrently. We said that we don't want to have that, but currently there's nothing in the code that prevents it.

sophokles73:
> What I meant here is that AMQP server will only ever invoke methods of RegistrationService interface, while REST server will exclusively use DeviceManagementService methods.

Unless you are also planning to split up the service implementation class, they actually will be invoking methods on the same instance. `FileBasedRegistrationService` implements both interfaces and there is no guarantee that private helper methods used by the methods defined by these interfaces are not invoked concurrently ...

https://github.com/eclipse/hono/blob/c63ddeb614111e201d36a6020605df5351412cb1/services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedRegistrationService.java#L62

dejanb:
+1 

I'll check to see what are potential concurrent access points.

Would you have something against doing sync on this (service) level vs. event bus level as it is today?

sophokles73:
> Would you have something against doing sync on this (service) level vs. event bus level as it is today?

As long as the synchronization only happens inside of the file based service implementation classes, I guess it's ok. Otherwise I would start to wonder if the gains are worth the effort ...

dejanb:
Makes sense. I'll revisit it with this perspective and let you know. Cheers

dejanb:
@sophokles73 I believe that use of concurrent maps should be enough to tackle all potential concurrency issues. Take a look and let me know what you think.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/AutowiredDeviceManagementHttpEndpoint.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * A default http endpoint implementation of the {@link DeviceManagementHttpEndpoint}.
+ * <p>
+ * This wires up the actual service instance with the mapping to the http endopint implementation. It is intended to be used
+ * in a Spring Boot environment.
+ */
+@Component
+@ConditionalOnBean(DeviceManagementService.class)
+public class AutowiredDeviceManagementHttpEndpoint extends DeviceManagementHttpEndpoint {
```
sophokles73: final ?

service-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementHttpEndpoint.java
```diff
@@ -100,29 +113,42 @@ public void addRoutes(final Router router) {
...
         final HttpServerResponse response = ctx.response();
 
-        final JsonObject requestMsg = EventBusMessage.forOperation(RegistryManagementConstants.ACTION_GET)
-                .setTenant(tenantId)
-                .setDeviceId(deviceId)
-                .toJson();
+        final Span span = Util.newChildSpan(SPAN_NAME_GET_DEVICE, TracingHandler.serverSpanContext(ctx), tracer, tenantId, deviceId, getClass().getSimpleName());
+
+        if (tenantId == null || deviceId == null) {
+            HttpUtils.badRequest(ctx, "Missing tenant or device id");
```
sophokles73: shouldn't the processing of the request end here?

service-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementHttpEndpoint.java
```diff
@@ -166,15 +200,26 @@ private void doUpdateDevice(final RoutingContext ctx) {
             payload.remove(RegistryManagementConstants.FIELD_PAYLOAD_DEVICE_ID);
         }
         final String tenantId = getTenantParam(ctx);
+
+        if (tenantId == null || deviceId == null) {
+            HttpUtils.badRequest(ctx, "Missing tenant or device id");
```
sophokles73: shouldn't request processing end here?

service-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementHttpEndpoint.java
```diff
@@ -184,14 +229,27 @@ private void doDeleteDevice(final RoutingContext ctx) {
...
-                .setTenant(tenantId)
-                .setDeviceId(deviceId)
-                .setResourceVersion(ctx.get(KEY_RESOURCE_VERSION))
-                .toJson();
+        final Optional<String> resourceVersion = Optional.ofNullable(ctx.get(KEY_RESOURCE_VERSION));
+
+        final Span span = Util.newChildSpan(SPAN_NAME_REMOVE_DEVICE, TracingHandler.serverSpanContext(ctx), tracer, tenantId, deviceId, getClass().getSimpleName());
+
+        if (tenantId == null || deviceId == null) {
+            HttpUtils.badRequest(ctx, "Missing tenant or device id");
```
sophokles73: end request processing?

service-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementHttpEndpoint.java
dejanb: Yes. Good catch that was missing in the old code as well. The request ends by badReqeust() call, but we should return immediatelly.

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -395,4 +397,52 @@ protected final CorsHandler createCorsHandler(final String allowedOrigin, final
                 .allowedHeader(HttpHeaders.IF_MATCH.toString())
                 .exposedHeader(HttpHeaders.ETAG.toString());
     }
+
+    /**
+     * Creates a response based on generic result.
+     * <p>
+     * The default behavior is as follows:
```
sophokles73: this sounds like the *default* behavior might be changed somehow, but since this method is final, I guess that this is *the behavior*, isn't it?

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -395,4 +397,52 @@ protected final CorsHandler createCorsHandler(final String allowedOrigin, final
...
+     * <li>Otherwise, the JSON object is written to the response body and the given custom handler is
+     * invoked (if not {@code null}).</li>
+     * </ol>
+     *
+     * @param ctx The routing context of the request.
+     * @param result The generic result of the operation.
+     * @param customHandler An (optional) handler for post processing the HTTP response, e.g. to set any additional HTTP
+     *                      headers. The handler <em>must not</em> write to response body. May be {@code null}.
+     */
+    protected final void createResponse(final RoutingContext ctx, final Result<?> result, final Handler<HttpServerResponse> customHandler) {
```
sophokles73: *createResponse* to me sounds like an object would be created and returned. This is obviously not the case so what about renaming this to e.g. *writeResponse* ?

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -395,4 +397,52 @@ protected final CorsHandler createCorsHandler(final String allowedOrigin, final
...
+     * Creates a response based on operation result (including resource version).
+     * <p>
+     * Sets ETAG header and then calls {@link #createResponse}
+     *
+     * @param ctx The routing context of the request.
+     * @param result The operation result of the operation.
+     * @param customHandler An (optional) handler for post processing the HTTP response, e.g. to set any additional HTTP
+     *                      headers. The handler <em>must not</em> write to response body. May be {@code null}.
+     */
+    protected final void createOperationResponse(final RoutingContext ctx, final OperationResult<?> result, final Handler<HttpServerResponse> customHandler) {
```
sophokles73: see above

service-base/src/main/java/org/eclipse/hono/service/management/device/EventBusDeviceManagementAdapter.java
```diff
@@ -1,213 +0,0 @@
...
-import io.vertx.core.Verticle;
-import io.vertx.core.json.JsonObject;
-
-/**
- * An adapter, hooking up the {@link DeviceManagementService} with the event bus.
- * <p>
- * In particular, this base class provides support for receiving service invocation request messages via vert.x' event
- * bus and route them to specific methods corresponding to the operation indicated in the message.
- */
-public abstract class EventBusDeviceManagementAdapter extends EventBusService
```
sophokles73: FMPOV the service-base module contains public API, doesn't it? If so, then we should first deprecate this class, indicating that it is no longer used/necessary and that it will be removed in Hono 2.0.0. WDYT?

service-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementHttpEndpoint.java
```diff
@@ -37,7 +45,12 @@
  * Registration API</a>. It receives HTTP requests representing operation invocations and sends them to an address on
  * the vertx event bus for processing. The outcome is then returned to the peer in the HTTP response.
  */
-public final class DeviceManagementHttpEndpoint extends AbstractHttpEndpoint<ServiceConfigProperties> {
+public abstract class DeviceManagementHttpEndpoint extends AbstractHttpEndpoint<ServiceConfigProperties> {
```
sophokles73: FMPOV the service-base module contains public API. As such, I think that we should not change this class but instead deprecate it do all modifications in AutowiredDeviceManagementHttpEndpoint only. WDYT?

service-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementHttpEndpoint.java
```diff
@@ -100,29 +113,43 @@ public void addRoutes(final Router router) {
...
 
         final String deviceId = getDeviceIdParam(ctx);
         final String tenantId = getTenantParam(ctx);
         final HttpServerResponse response = ctx.response();
 
-        final JsonObject requestMsg = EventBusMessage.forOperation(RegistryManagementConstants.ACTION_GET)
-                .setTenant(tenantId)
-                .setDeviceId(deviceId)
-                .toJson();
+        final Span span = Util.newChildSpan(SPAN_NAME_GET_DEVICE, TracingHandler.serverSpanContext(ctx), tracer, tenantId, deviceId, getClass().getSimpleName());
```
calohmn: This span, and the other spans in this class, doesn't get finished. This is an issue in the `EventBus[*]ManagementAdapter` classes as well. I've created #1702 to fix that in the event bus classes (which we still might want to keep for Hono 1.1, see comment below). You could have a look there.

service-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementHttpEndpoint.java
dejanb: I reverted it and create a new AbstractDeviceManagementHttpEndpoint class instead

service-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementHttpEndpoint.java
dejanb: Thanks for the suggestions. I tried to add it to this implementation. Can you take another look and make sure it's all there?

calohmn: I've had a look at `AbstractDeviceManagementHttpEndpoint` and the spans still need to get finished there. The places where to insert `span.finish()` would be the premature method exits, e.g. before the `return` here:
````
            TracingHelper.logError(span, "missing body");
            HttpUtils.badRequest(ctx, "missing body");
            return;
````
and also in the result handlers of the service invocations (e.g. after `writeOperationResponse`).

File: service-base/src/main/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementHttpEndpoint.java [hunk 11]
```diff
@@ -0,0 +1,271 @@
...
+
+        final JsonObject payload = ctx.get(KEY_REQUEST_BODY);
+        if (payload == null) {
+            TracingHelper.logError(span, "missing body");
+            HttpUtils.badRequest(ctx, "missing body");
+            return;
+        }
+
+        if (deviceId != null) {
+            if (!(deviceId instanceof String)) {
```
calohmn: This condition is always `false`.

service-base/src/main/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementHttpEndpoint.java
```diff
@@ -0,0 +1,271 @@
...
+     */
+    @Autowired
+    public AbstractDeviceManagementHttpEndpoint(final Vertx vertx) {
+        super(vertx);
+    }
+
+    @Override
+    protected String getEventBusAddress() {
+        return RegistryManagementConstants.EVENT_BUS_ADDRESS_DEVICE_MANAGEMENT_IN;
+    }
```
jbtrystram: This is no longer needed

service-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementHttpEndpoint.java
dejanb: Good catch. I think the param checks have to be done before span is created. At least that's consistent with the current implementation. I also cleaned it up a bit. Can you give it another look?

service-base/src/main/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementHttpEndpoint.java
dejanb: We can't remove it until all services are refactored. I removed the actual value for now to reduce the confusion.

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -347,6 +349,37 @@ protected final String getDeviceIdParam(final RoutingContext ctx) {
...
 
+    /**
+     * Get request parameter value and check if it has been set. If it's not set, fail the request.
+     *
+     * @param paramName The name of the parameter to get.
+     * @param ctx The routing context of the request.
+     * @return The value of the parameter if it's set or {@code null} otherwise.
+     * @throws NullPointerException If ctx is {@code null}.
+     */
+    protected final String getRequestParam(final String paramName, final RoutingContext ctx) {
```
sophokles73: maybe this should then be named *getMandatoryRequestParameter* in order to more easily distinguish from the variant accepting the boolean param?

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -394,4 +427,52 @@ protected final CorsHandler createCorsHandler(final String allowedOrigin, final
...
+     *
+     * @param ctx The routing context of the request.
+     * @param result The operation result of the operation.
+     * @param customHandler An (optional) handler for post processing the HTTP response, e.g. to set any additional HTTP
+     *                      headers. The handler <em>must not</em> write to response body. May be {@code null}.
+     */
+    protected final void writeOperationResponse(final RoutingContext ctx, final OperationResult<?> result, final Handler<HttpServerResponse> customHandler) {
+        result.getResourceVersion().ifPresent(v -> ctx.response().putHeader(HttpHeaders.ETAG, v));
+        writeResponse(ctx, result, customHandler);
+    }
```
sophokles73: I have to admit that I do not really like the dependency on the management package that is being introduced here. Up until now, this class (and package) has been providing generic functionality for implementing HTTP endpoints. These imports now tie it to the management API which I don't think is desirable nor necessary. How about moving these two methods to an abstract endpoint class in the `org.eclipse.hono.service.management` package?

service-base/src/main/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementHttpEndpoint.java
```diff
@@ -0,0 +1,242 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
```
sophokles73: this should be 2020 now, I guess ;-)

service-base/src/main/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementHttpEndpoint.java
```diff
@@ -0,0 +1,242 @@
...
+     * @throws NullPointerException if vertx is {@code null};
+     */
+    @Autowired
+    public AbstractDeviceManagementHttpEndpoint(final Vertx vertx) {
+        super(vertx);
+    }
+
+    @Override
+    protected String getEventBusAddress() {
+        return "";
```
sophokles73: maybe you should add a JavaDoc comment, why this returns the empty string

service-base/src/main/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementHttpEndpoint.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    @Override
+    public String getName() {
+        return String.format("%s/%s",
```
sophokles73: IMHO this should be defined as a constant. No need to recreate it on each invocation, is there?

service-base/src/main/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementHttpEndpoint.java
```diff
@@ -0,0 +1,242 @@
...
+    protected abstract DeviceManagementService getService();
+
+    private void doGetDevice(final RoutingContext ctx) {
+
+        final String deviceId = getRequestParam(PARAM_DEVICE_ID, ctx);
+        final String tenantId = getRequestParam(PARAM_TENANT_ID, ctx);
+
+        final HttpServerResponse response = ctx.response();
+
+        final Span span = Util.newChildSpan(SPAN_NAME_GET_DEVICE, TracingHandler.serverSpanContext(ctx), tracer, tenantId, deviceId, getClass().getSimpleName());
```
sophokles73: FMPOV it would be desirable to start the span *before* extracting the mandatory parameters and then log any missing parameters to the span. That way, it is easier for support people to determine why a request failed.

service-base/src/main/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementHttpEndpoint.java
```diff
@@ -0,0 +1,242 @@
...
+            final OperationResult<Device> operationResult = handler.result();
+            final int status = operationResult.getStatus();
+            response.setStatusCode(status);
+            switch (status) {
+                case HttpURLConnection.HTTP_OK:
+                    operationResult.getResourceVersion().ifPresent(v -> response.putHeader(HttpHeaders.ETAG, v));
+                    HttpUtils.setResponseBody(response, JsonObject.mapFrom(operationResult.getPayload()));
+                    // falls through intentionally
+                default:
+                    span.finish();
```
sophokles73: FMPOV we should record the outcome (status code) of the request in the span

service-base/src/main/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementHttpEndpoint.java
```diff
@@ -0,0 +1,242 @@
...
+        final String tenantId = getRequestParam(PARAM_TENANT_ID, ctx);
+        final String deviceId = getRequestParam(PARAM_DEVICE_ID, ctx, true);
+
+        final JsonObject payload = ctx.get(KEY_REQUEST_BODY);
+        if (payload == null) {
+            HttpUtils.badRequest(ctx, "Missing request body");
+            return;
+        }
+
+        final Span span = Util.newChildSpan(SPAN_NAME_CREATE_DEVICE, TracingHandler.serverSpanContext(ctx), tracer, tenantId, deviceId, getClass().getSimpleName());
```
sophokles73: same here

service-base/src/main/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementHttpEndpoint.java
```diff
@@ -0,0 +1,242 @@
...
+    private void doUpdateDevice(final RoutingContext ctx) {
+
+        final String deviceId = getRequestParam(PARAM_DEVICE_ID, ctx);
+        final JsonObject payload = ctx.get(KEY_REQUEST_BODY);
+        if (payload != null) {
+            payload.remove(RegistryManagementConstants.FIELD_PAYLOAD_DEVICE_ID);
+        }
+        final String tenantId = getRequestParam(PARAM_TENANT_ID, ctx);
+
+        final Span span = Util.newChildSpan(SPAN_NAME_UPDATE_DEVICE, TracingHandler.serverSpanContext(ctx), tracer, tenantId, deviceId, getClass().getSimpleName());
```
sophokles73: same here

service-base/src/main/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementHttpEndpoint.java
```diff
@@ -0,0 +1,242 @@
...
+    private void doDeleteDevice(final RoutingContext ctx) {
+
+        final String deviceId = getRequestParam(PARAM_DEVICE_ID, ctx);
+        final String tenantId = getRequestParam(PARAM_TENANT_ID, ctx);
+
+        logger.debug("removing device [tenant: {}, device: {}]", tenantId, deviceId);
+
+        final Optional<String> resourceVersion = Optional.ofNullable(ctx.get(KEY_RESOURCE_VERSION));
+
+        final Span span = Util.newChildSpan(SPAN_NAME_REMOVE_DEVICE, TracingHandler.serverSpanContext(ctx), tracer, tenantId, deviceId, getClass().getSimpleName());
```
sophokles73: same here

service-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementHttpEndpoint.java
```diff
@@ -37,6 +37,7 @@
  * Registration API</a>. It receives HTTP requests representing operation invocations and sends them to an address on
  * the vertx event bus for processing. The outcome is then returned to the peer in the HTTP response.
  */
+@Deprecated(forRemoval = true)
```
sophokles73: maybe you can also add a `@deprecated` tag to the JavaDoc explaining why this class will be removed?

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
dejanb: These are now moved to the Util class.

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -347,6 +349,46 @@ protected final String getDeviceIdParam(final RoutingContext ctx) {
...
+    /**
+     * Get request parameter value. Optionally, if parameter has not been set, fail the request.
+     *
+     * @param paramName The name of the parameter to get.
+     * @param ctx The routing context of the request.
+     * @param span The active OpenTracing span for this operation. In case of the missing mandatory parameter, the error is logged and the span is finished.
+     *             Otherwise, the parameter is set as a span tag.
+     * @param optional Whether to check if parameter has been set or not.
+     * @return The value of the parameter if it's set or {@code null} otherwise.
+     * @throws NullPointerException If ctx is {@code null}.
```
sophokles73: I am pretty sure that you will also get a `NullPointException` if the paramName is `null`.

service-base/src/main/java/org/eclipse/hono/service/management/Util.java
```diff
@@ -84,4 +107,55 @@ public static final Span newChildSpan(final String operationName, final SpanCont
...
+        response.setStatusCode(status);
+        if (status >= 400) {
+            HttpUtils.setResponseBody(response, JsonObject.mapFrom(result.getPayload()));
+        } else if (status == HttpURLConnection.HTTP_CREATED) {
+            HttpUtils.setResponseBody(response, JsonObject.mapFrom(result.getPayload()));
+            if (customHandler != null) {
+                customHandler.handle(response);
+            }
+        }
+        span.log(String.format("request finished with status: %s", status));
```
sophokles73: I actually had in mind to set the `Tags.HTTP_STATUS` tag on the span. But maybe that is already done somewhere else? Who is actually creating the Span that is passed in here? Is this the span that is created by the `TracingHandler` that is registered by the `DeviceRegistryRestServer`?

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/ApplicationConfig.java
```diff
@@ -198,8 +198,8 @@ public ServiceConfigProperties restProperties() {
     @Bean
     @Scope("prototype")
     @ConditionalOnBean(DeviceManagementService.class)
-    public DeviceManagementHttpEndpoint registrationHttpEndpoint() {
-        return new DeviceManagementHttpEndpoint(vertx());
+    public AbstractDeviceManagementHttpEndpoint registrationHttpEndpoint() {
```
sophokles73: is it necessary to return an `AbstractDeviceManagementHttpEndpoint` instead of just a `DeviceManagementHttpEndpoint` ?

service-base/src/main/java/org/eclipse/hono/service/management/Util.java
dejanb: Yeah, that's the one and I don't see it being set anywhere else in this case. Fixed now

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/ApplicationConfig.java
dejanb: It is. These are not related by any means, the 'AbstractDeviceManagementHttpEndpoint' is completely new, so that 'DeviceManagementHttpEndpoint' could be deprecated.

service-base/src/main/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementHttpEndpoint.java
```diff
@@ -0,0 +1,250 @@
...
+            final OperationResult<Device> operationResult = handler.result();
+            final int status = operationResult.getStatus();
+            response.setStatusCode(status);
+            switch (status) {
+                case HttpURLConnection.HTTP_OK:
+                    operationResult.getResourceVersion().ifPresent(v -> response.putHeader(HttpHeaders.ETAG, v));
+                    HttpUtils.setResponseBody(response, JsonObject.mapFrom(operationResult.getPayload()));
+                    // falls through intentionally
+                default:
+                    span.log(String.format("request finished with status: %s", status));
```
sophokles73: this should set the HTTP_STATUS tag instead, right?

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -347,6 +350,47 @@ protected final String getDeviceIdParam(final RoutingContext ctx) {
...
+     * @throws NullPointerException If ctx or paramName are {@code null}.
+     */
+    protected final String getRequestParam(final String paramName, final RoutingContext ctx, final Span span, final boolean optional) {
+        final String value = ctx.request().getParam(paramName);
+        if (!optional && value == null) {
+            final String msg = String.format("Missing request parameter: %s", paramName);
+            TracingHelper.logError(span, msg);
+            HttpUtils.badRequest(ctx, msg);
+            Tags.HTTP_STATUS.set(span, HttpURLConnection.HTTP_BAD_REQUEST);
+            span.finish();
```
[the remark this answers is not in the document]
dejanb: This is just a helper function to deal with the common code we need to get device and tenant ids in various places ... We can make it more general in the future if there is need for that

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -347,6 +350,47 @@ protected final String getDeviceIdParam(final RoutingContext ctx) {
...
+            return null;
+        } else {
+            span.setTag(paramName, value);
```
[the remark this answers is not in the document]
dejanb: We set device and tenant ids as tags for these spans (at least that was the behaviour before the refactoring)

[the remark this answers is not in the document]
dejanb: That was the starting point, but there were too much repetitive code for both parameters in all methods. Hence the idea to make it more reusable. I'll see how to improve it in the future.

sophokles73:
@dejanb IMHO you will also need to remove `org.eclipse.hono.deviceregistry.AutowiredDeviceManagementAdapter` from the example device registry, as it is no longer used but will still get instantiated by means of Spring Boot's component scan.

BTW as this is now part of 1.1.0, I hope that you are making similar changes for the other types (Tenants, Credentials) as well, right?

dejanb:
@sophokles73 
> @dejanb IMHO you will also need to remove `org.eclipse.hono.deviceregistry.AutowiredDeviceManagementAdapter` from the example device registry, as it is no longer used but will still get instantiated by means of Spring Boot's component scan.
> 

We can also deprecate the class and prevent spring boot from instantiating it. There's #1730 for this. Let me know if that makes sense to you? 

> BTW as this is now part of 1.1.0, I hope that you are making similar changes for the other types (Tenants, Credentials) as well, right?

That's the plan.

sophokles73:
> We can also deprecate the class and prevent spring boot from instantiating it. There's #1730 for this. Let me know if that makes sense to you?

FMPOV we should remove it completely. Since the example registry doesn't need it, why would we need to keep it?

dejanb:
The only reason I can see is to be consistent as it comes in the combination with `DeviceManagementHttpEndpoint` which is also not used anymore, but deprecated at the moment. I would remove or deprecate both of them.

PR #1676, opened 2019-12-17
Include connection to grid in readiness check

sophokles73:
Also improved initial connect behavior by means of repeatedly trying to
establish the connection.

### DIFF
--- a/services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
+++ b/services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
@@ -16,6 +16,7 @@ package org.eclipse.hono.deviceconnection.infinispan;
 
 import java.net.HttpURLConnection;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.eclipse.hono.client.ServerErrorException;
 import org.eclipse.hono.service.HealthCheckProvider;
@@ -48,6 +49,7 @@ public class RemoteCacheBasedDeviceConnectionService extends EventBusDeviceConne
 
     private RemoteCacheContainer cacheManager;
     private BasicCache<String, String> cache;
+    private AtomicBoolean connecting = new AtomicBoolean(false);
 
     /**
      * Sets the cache manager to use for retrieving a cache.
@@ -96,32 +98,37 @@ public class RemoteCacheBasedDeviceConnectionService extends EventBusDeviceConne
 
     private void connectToGrid() {
 
-        context.executeBlocking(r -> {
-            try {
-                if (!cacheManager.isStarted()) {
-                    log.debug("trying to start cache manager");
-                    cacheManager.start();
-                    log.info("started cache manager, now connecting to remote cache");
+        if (connecting.compareAndSet(false, true)) {
+
+            context.executeBlocking(r -> {
+                try {
+                    if (!cacheManager.isStarted()) {
+                        log.debug("trying to start cache manager");
+                        cacheManager.start();
+                        log.info("started cache manager, now connecting to remote cache");
+                    }
+                    log.debug("trying to connect to remote cache");
+                    cache = cacheManager.getCache(CACHE_NAME);
+                    if (cache == null) {
+                        r.fail(new IllegalStateException("remote cache [" + CACHE_NAME + "] does not exist"));
+                    } else {
+                        cache.start();
+                        r.complete(cacheManager);
+                    }
+                } catch (final Throwable t) {
+                    r.fail(t);
                 }
-                log.debug("trying to connect to remote cache");
-                cache = cacheManager.getCache(CACHE_NAME);
-                if (cache == null) {
-                    r.fail(new IllegalStateException("remote cache [" + CACHE_NAME + "] does not exist"));
+            }, attempt -> {
+                if (attempt.succeeded()) {
+                    log.info("successfully connected to remote cache");
                 } else {
-                    cache.start();
-                    r.complete(cacheManager);
+                    log.debug("failed to connect to remote cache: {}", attempt.cause().getMessage());
                 }
-            } catch (final Throwable t) {
-                r.fail(t);
-            }
-        }, attempt -> {
-            if (attempt.succeeded()) {
-                log.info("successfully connected to remote cache");
-            } else {
-                log.debug("failed to connect to remote cache: {}", attempt.cause().getMessage());
-                vertx.setTimer(5000, timerId -> connectToGrid());
-            }
-        });
+                connecting.set(false);
+            });
+        } else {
+            log.info("already trying to establish connection to data grid");
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
```diff
@@ -86,28 +84,36 @@ protected final void doStart(final Promise<Void> startPromise) {
...
+    private void connectToGrid() {
+
+        context.executeBlocking(r -> {
+            if (!cacheManager.isStarted()) {
+                log.debug("trying to start cache manager");
+                cacheManager.start();
+                log.info("started cache manager, now connecting to remote cache");
+            }
+            log.debug("trying to connect to remote cache");
+            cache = cacheManager.getCache("device-connection");
```
ctron: If this returns `null`, then the next call will fail with a NPE. I think a more helpful exception would be appropriate here.

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
```diff
@@ -86,28 +84,36 @@ protected final void doStart(final Promise<Void> startPromise) {
...
+            cache.start();
+            r.complete(cacheManager);
+        }, attempt -> {
+            if (attempt.succeeded()) {
+                log.info("successfully connected to remote cache");
+            } else {
+                log.debug("failed to connect to remote cache: {}", attempt.cause().getMessage());
```
ctron: A failed connection to a crucial components should IMHO: a) log to at least "INFO", b) log the full stack trace, not only the message.

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
```diff
@@ -45,7 +44,6 @@
  */
 public class RemoteCacheBasedDeviceConnectionService extends EventBusDeviceConnectionAdapter implements DeviceConnectionService, HealthCheckProvider {
 
-    private static final Logger LOG = LoggerFactory.getLogger(RemoteCacheBasedDeviceConnectionService.class);
```
ctron: I think it is helpful to know that problems originate in this class.

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
```diff
@@ -86,28 +84,36 @@ protected final void doStart(final Promise<Void> startPromise) {
...
-            });
-            result.complete();
+            connectToGrid();
+            startFuture.complete();
         }
     }
 
+    private void connectToGrid() {
+
+        context.executeBlocking(r -> {
```
ctron: I think the block misses the error handling:

> A Future instance is passed into blockingCodeHandler. When the blocking code successfully completes, the handler should call the Promise.complete(T) or Promise.complete(Object) method, or the Promise.fail(java.lang.Throwable) method if it failed.

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
sophokles73: IMHO the *log* instance does provide for this as it is created using the current class name:
```
    /**
     * A logger to be shared by subclasses.
     */
```

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
sophokles73: `ContextImpl.executeBlocking` already takes care of it by failing the promise if the blocking code throws a `Throwable`. However, this behavior is not specified by `Context.executeBlocking()`. Given that the infinispan methods in question do not document any exceptions being thrown, we probably should explicitly catch all `Throwable`s here for transparency ...

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
sophokles73: I didn't know that this method can return `null`. Again, the JavaDoc doesn't seem to allow for it. However, haven taken a look at `RemoteCacheManager.getCache()` I see that it can :-1: 
Thanks for pointing out. It's a real shame that the JavaDocs of Inifispan seem to be completely useless ... I hope we are doing a much better job with Hono :-)

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
sophokles73: In general, I agree. To me it is still unclear how the client establishes the connection to the grid and in particular, how it behaves if the connection is lost some time after it has been established initially. The current code explicitly takes care of initial connection but currently does not help with re-establishing a lost connection. Do you have any insight into this? Will the client try to re-connect on itself? Will it endlessly try to re-establish the connection?

sophokles73:
@ctron would you mind taking another look?

sophokles73:
@ctron would you mind taking another look?

sophokles73:
@ctron if you find the time, I'd really like to you to take another look at the changes I have pushed.

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
ctron: Ok, what I meant was: that I think it is helpful to know that a log/problem originates in this, or parent/sub-classes.

Right now, also messages from e.g. `EventBusDeviceConnectionAdapter` will have the name of `RemoteCacheBasedDeviceConnectionService`. So if you see a message in the log, you always need to walk through the class hierarchy, in order to find the right source.

So should there be sub-classes of this class, the would all log messages from this class, under a different name.

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
ctron: > Will the client try to re-connect on itself? Will it endlessly try to re-establish the connection?

While not having any knowledge about the code, I learned this from discussions. So yes, once this is started, it should continue to re-connect.

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
ctron: We do :)

Issue #1287, opened 2019-06-13
Add metrics to keep track of the total time the devices are connected to Hono


kaniyan:
Add metrics to keep track of the total time the devices are connected to Hono per tenant. 

Currently there are metrics in Hono to track the number of simultaneous device connections. In addition to it, it is also valuable to track how long the devices are connected to Hono per tenant.   This can also be used to effectively manage the resources by setting the _resource-limits_ and also shown in the dashboard.

kaniyan:
The MQTT and AMQP protocol adapters which maintain connection state already provide metrics namely `hono.connections` to track the number of connected devices per tenant.

Now we would also like to find out the total time the devices are connected per tenant. After taking a closer look at the existing metrics, I think the metrics `hono.connections` could well also be used to compute the total connection time per tenant. Currently this metric is being recorded using type _Gauge_, scrapped at regular intervals and stored as time series in the Prometheus server. By finding the sum of the number of device connections reported at every scrape for the given time period and multiplying it with the scrape interval should give the total device connection time per tenant. 

A sample PromQL query to get the total connection time of a tenant for the last one hour is given below:

```
sum(sum_over_time(hono_connections_authenticated {tenant="DEFAULT_TENANT"} [1h])) * scraping interval

```

The applications like resource limit checks that would like to calculate the total connection time should also be aware of the scrapping interval. 

In this approach, the level of accuracy depends on the scrapping interval. For example if the scrapping interval is 5 seconds, then a connection established and disconnected within 5 seconds may not be recorded by the metrics `hono.connections` and so the connection time.

sophokles73:
I am a little concerned with the fact that the usefulness of this approach is so closely tied to the definition of an appropriate scraping interval. When using a short interval, the accurateness of the measures values might be sufficient. However, I am not sure if 5 seconds is a reasonable interval length in all cases.
Using a longer interval might in fact lead to a connection time of 0. For example when using a 20 second scrape interval, if all devices connect, send data and disconnect every 20 seconds, we might end up in a situation where the devices always connect in between the scrapes, right? FMPOV this example isn't very far fetched. Even if only, say, half of the devices connect in between the scrapes, the resulting accuracy of the measurement won't be sufficient, would it?

kaniyan:
yes as the scraping interval gets longer and quicker the devices connect and disconnect, then less accurate the computed connection time using the metrics `hono.connections`. I think then it makes sense to make use of a new metrics to record the connection time.

[...]

PR #1678, opened 2019-12-18
[#1287] Add metrics to track the device connection time at tenant level.

kaniyan:
This PR employs new metric to track the device connection time of tenants. Some optimizations, javadoc, docs and tests will be done incrementally once the core concept is reviewed. 

A solution to track the device connection time could be to use a metrics of type _Timer_ by start tracking the time once a device _connects_ and stop tracking once that device _disconnects_ and report the measured value to the _Prometheus_. After giving some thought, I think this way is not quite useful as the data will be stored in Prometheus only after a device disconnects. In case of MQTT or AMQP, it is possible that the devices stay connected for a long time like weeks, months or even longer. If resource-limit checks are to make use of this data to set a limit based on the _device connection time_, then this data won't be accurate and might not be quite useful. To resolve this recording of the device connection at regular intervals is needed.

In this PR, a new class `DeviceConnectionTimeTracker` is added which tracks and records the time in Prometheus once a device connects/disconnects. Additionally, it also records the device connection time at regular intervals with the help of _vertx_ timer.


sophokles73:
Hmm, what about this scenario:
A tenant has a single device. Every minute the device wakes up, connects, sends data and disconnects after 5 secs and goes to sleep again. 

To me it looks like we would never record any connection time for this tenant, would we?

kaniyan:
> Hmm, what about this scenario:
> A tenant has a single device. Every minute the device wakes up, connects, sends data and disconnects after 5 secs and goes to sleep again.
> 
> To me it looks like we would never record any connection time for this tenant, would we?

This implementation covers this scenario too. In this case, when a devices connects, immediately a  `DeviceConnectionTimeTracker` is started and once the device disconnects, this tracker is stopped and the tracked connection time is recorded using the metric type _Timer_.

sophokles73:
Ah, I missed the reporting being done in the *finish* method. This looks like a very clever approach to the issue :+1: 
Some thoughts that come to my mind:
in cases where a large number of devices re-connect almost at once, e.g. after an adapter instance pod has been shut down/crashed, there will be many timers being stopped and started again etc. I wonder if this poses an issue or if we might be better of with a single timer per tenant which gets updated with the current number of connected devices whenever it changes. WDYT?

kaniyan:
> Ah, I missed the reporting being done in the _finish_ method. This looks like a very clever approach to the issue 
> Some thoughts that come to my mind:
> in cases where a large number of devices re-connect almost at once, e.g. after an adapter instance pod has been shut down/crashed, there will be many timers being stopped and started again etc. I wonder if this poses an issue or if we might be better of with a single timer per tenant which gets updated with the current number of connected devices whenever it changes. WDYT?


Currently in this approach, there is only one `DeviceConnectionTimeTracker` per _tenant_ and it means only one timer per _tenant_. The trackers are stored in a `ConcurrentHashMap` named as _connectionTimeTrackers_ with _tenantId_ as the key. Every time a device connects or disconnects, it is checked for any existing tracker in the map for that _tenant_, if exists then the tracker is set to _finish_ (which stops the timer and reports the calculated value) . Then a new tracker with the current number of device connections replaces the existing one in the map for that _tenant_. Apart from that, the tracker reports the time using the given metric type at regular intervals.

sophokles73:
@kaniyan I know all that from reading the code :-) The point I tried to make is that we probably do not need to create a new timer for the tenant each time a device connects or disconnects. Instead, we could start a timer for a tenant when the tenant's first device connects and then keep that timer running until the last device of the tenant disconnects. In between, we could then simply update the number of currently connected devices when a device connects/disconnects. Does that make sense?

kaniyan:
> @kaniyan I know all that from reading the code :-) The point I tried to make is that we probably do not need to create a new timer for the tenant each time a device connects or disconnects. Instead, we could start a timer for a tenant when the tenant's first device connects and then keep that timer running until the last device of the tenant disconnects. In between, we could then simply update the number of currently connected devices when a device connects/disconnects. Does that make sense?

I see your point to avoid creating new timers every time a device connect or disconnects. I have changed the implementation now in such a way that the number of currently connected devices could be updated. During the update, it need to be ensured that the timer is also not reporting at the same time. So that I use now `vertx.periodicStream( ...)` which offers support to _pause_ and _resume_ timers. Would you mind taking a look?

service-base/src/main/java/org/eclipse/hono/service/metric/DeviceConnectionTimeTracker.java
```diff
@@ -0,0 +1,167 @@
...
+        if (noOfDeviceConnections == 0) {
+            this.timer.cancel();
+            log.trace("Stopped the device connection time tracker for the tenant [{}].", tenantId);
+            return null;
+        }
+        this.timer.resume();
+        return this;
+    }
+
+    private void report(final long noOfDeviceConnections) {
```
sophokles73: FMPOV this should not need any argument as the current value to report can simply be read from the *noOfDeviceConnections* field, can't it?

service-base/src/main/java/org/eclipse/hono/service/metric/DeviceConnectionTimeTracker.java
```diff
@@ -0,0 +1,167 @@
...
+    private DeviceConnectionTimeTracker(
+            final String tenantId, final Vertx vertx,
+            final long noOfDeviceConnections,
+            final Consumer<Long> recorder) {
+        this.tenantId = tenantId;
+        this.noOfDeviceConnections = new AtomicLong(noOfDeviceConnections);
+        this.recorder = recorder;
+        this.startInstant.set(Instant.now());
+        this.timer = vertx
+                .periodicStream(DEVICE_CONNECTION_TIME_RECORDING_INTERVAL_IN_MS)
```
sophokles73: I do not quite see why you want to use a periodicStream here. Can you explain?

service-base/src/main/java/org/eclipse/hono/service/metric/DeviceConnectionTimeTracker.java
kaniyan: yes, I will update it.

service-base/src/main/java/org/eclipse/hono/service/metric/DeviceConnectionTimeTracker.java
kaniyan: Here I use the periodicStream mainly to take advantage of its methods namely `pause()` and `resume()`. Let us assume a new device is connected or disconnected, then the `updateNoOfDeviceConnections(...)` is called. This method reports the connection time before updating the number of device connections with the actual value. While it reports the connection time, it is also possible that the timer reporting at regular intervals also reports at the same time. In order to avoid that, the timer is paused, the value is reported and then finally the timer is resumed.

sophokles73: IMHO that cannot happen because both handlers will be run on the same event loop, i.e. while the handler executing *updateNoOfDeviceConnections* is running, vert.x will make sure that the handler triggered by the timer is not run concurrently. In fact, vert.x guarantees that only exactly one handler will run at any given time on an event loop. But may be you should verify by checking if the timer task is run on the same thread as the handler updating the number of devices. But I am pretty sure that it is ...

kaniyan: As I checked, the timer task is running on the same thread as you mentioned above. Then in this case, I will simply use `vertx.setPeriodic(...)` instead of a _periodicStream_.

sophokles73: I think that will make it much easier to read/understand

kaniyan: @sophokles73 I have pushed the corresponding changes. Would you mind taking a look?

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -395,4 +403,32 @@ private void handleTenantTimeout(final String tenantId) {
...
+                        .map(tracker -> tracker.updateNoOfDeviceConnections(deviceConnectionsCount))
+                        .orElseGet(() -> {
+                            if (deviceConnectionsCount > 0) {
+                                return DeviceConnectionTimeTracker.Builder
+                                        .createInstance()
+                                        .ForTenant(tenant)
+                                        .withVertx(vertx)
+                                        .withNumberOfDeviceConnections(deviceConnectionsCount)
+                                        .recordUsing(connectionTime -> registry
+                                                .timer(METER_CONNECTIONS_AUTHENTICATED_TIME,
```
sophokles73: I wonder if a Timer is the right meter type to use here. Since we are not really timing any particular interaction, I think that simply using a Gauge would work here as well, preventing the higher memory footprint a Timer involves. WDYT?

service-base/src/main/java/org/eclipse/hono/service/metric/DeviceConnectionTimeTracker.java
```diff
@@ -0,0 +1,164 @@
...
+        private Builder() {
+            // prevent instantiation
+        }
+
+        /**
+         * Creates an instance of the Builder.
+         * 
+         * @return a reference to this for fluent use.
+         */
+        public static Builder createInstance() {
```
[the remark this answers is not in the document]
kaniyan: Apart from creating an instance of the Builder, it has no other use. This can be removed as now `forTenant(...)` is a static factory method and returns an instance of the Builder.

service-base/src/main/java/org/eclipse/hono/service/metric/DeviceConnectionTimeTracker.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        /**
+         * Sets the tenant id.
+         *
+         * @param tenantId The tenant id that the devices belongs to.
+         * @return a reference to this for fluent use.
+         * @throws NullPointerException if the tenantId is {@code null}.
+         */
+        public Builder ForTenant(final String tenantId) {
```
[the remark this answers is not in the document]
kaniyan: I will change it as a static factory method.

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -49,6 +51,10 @@
      * The name of the meter for authenticated connections.
      */
     public static final String METER_CONNECTIONS_AUTHENTICATED = "hono.connections.authenticated";
+    /**
+     * The name of the meter for authenticated connections time.
```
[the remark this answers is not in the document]
kaniyan: I followed the same description pattern as the already existing constants in that class. I agree that adding the purpose of the meter is more helpful.

service-base/src/main/java/org/eclipse/hono/service/metric/DeviceConnectionTimeTracker.java
```diff
@@ -0,0 +1,164 @@
...
+     */
+    public DeviceConnectionTimeTracker updateNoOfDeviceConnections(final long noOfDeviceConnections) {
+        report();
+        this.noOfDeviceConnections.set(noOfDeviceConnections);
+        log.trace("Updated number of device connections for the tenant [{}] to [{}]", tenantId,
+                noOfDeviceConnections);
+        if (noOfDeviceConnections == 0) {
+            vertx.cancelTimer(timerId);
+            log.trace("Stopped the device connection time tracker for the tenant [{}].", tenantId);
+            return null;
```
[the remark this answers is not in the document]
kaniyan: When there are no devices connected then there is no need to track the connection time. Hence the timer is stopped and _null_ is returned. On _null_ value, the `connectionTimeTrackers.compute(...)` in `MicrometerBasedMetrics` handles this by removing the tenant entry from the map. I would update the java doc regarding the _null_ value.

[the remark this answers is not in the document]
sophokles73: makes sense

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
kaniyan: I replaced now with a metric type of Gauge and pushed the changes. Please take a look.

service-base/src/main/java/org/eclipse/hono/service/metric/DeviceConnectionTimeTracker.java
```diff
@@ -0,0 +1,153 @@
...
+        public static Builder forTenant(final String tenantId) {
+            final Builder builder = new Builder();
+            builder.tenantId = Objects.requireNonNull(tenantId);
+            return builder;
+        }
+
+        /**
+         * Sets the number of device connections.
+         * 
+         * @param noOfDeviceConnections The number of device connections.
```
sophokles73: can this be 0 or negative?

service-base/src/main/java/org/eclipse/hono/service/metric/DeviceConnectionTimeTracker.java
```diff
@@ -0,0 +1,153 @@
...
+            final String tenantId, final Vertx vertx,
+            final long noOfDeviceConnections,
+            final Consumer<Long> recorder) {
+        this.tenantId = tenantId;
+        this.noOfDeviceConnections = new AtomicLong(noOfDeviceConnections);
+        this.recorder = recorder;
+        this.startInstant.set(Instant.now());
+        this.vertx = vertx;
+        this.timerId = this.vertx.setPeriodic(DEVICE_CONNECTION_TIME_RECORDING_INTERVAL_IN_MS, id -> report());
+        log.trace("Stared a device connection time tracker for the tenant [{}].", tenantId);
```
sophokles73: Started ...

service-base/src/main/java/org/eclipse/hono/service/metric/DeviceConnectionTimeTracker.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /**
+     * Reports the device connection time and then updates the number of device connections with the given value.
+     * 
+     * @param noOfDeviceConnections The number of device connections.
```
sophokles73: can this be 0 or negative?

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -49,6 +50,10 @@
      * The name of the meter for authenticated connections.
      */
     public static final String METER_CONNECTIONS_AUTHENTICATED = "hono.connections.authenticated";
+    /**
+     * This metric is used to keep track of the amount of time that the authenticated devices are connected to Hono.
+     */
+    public static final String METER_CONNECTIONS_AUTHENTICATED_TIME = "hono.connections.authenticated.time";
```
sophokles73: IMHO *time* should be replaced with something more specific to the purpose, e.g. *duration*?

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -79,6 +84,8 @@
     protected final MeterRegistry registry;
 
     private final Map<String, AtomicLong> authenticatedConnections = new ConcurrentHashMap<>();
+    private final Map<String, AtomicLong> authenticatedConnectionsTime = new ConcurrentHashMap<>();
```
sophokles73: why do we need two maps here? The DeviceConnectionTimeTracker already contains the number of connections, doesn't it?

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -127,7 +138,7 @@ public final void incrementConnections(final String tenantId) {
         Objects.requireNonNull(tenantId);
         gaugeForTenant(METER_CONNECTIONS_AUTHENTICATED, this.authenticatedConnections, tenantId, AtomicLong::new)
                 .incrementAndGet();
-        this.totalCurrentConnections.incrementAndGet();
+        trackDeviceConnectionTime(tenantId, this.totalCurrentConnections.incrementAndGet());
```
sophokles73: tare you sure that you want to pass in the overall number of connected devices across all tenants? I think you'd want to pass in the value returned by the previous statement, wouldn't you?

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -138,7 +149,7 @@ public final void decrementConnections(final String tenantId) {
         Objects.requireNonNull(tenantId);
         gaugeForTenant(METER_CONNECTIONS_AUTHENTICATED, this.authenticatedConnections, tenantId, AtomicLong::new)
                 .decrementAndGet();
-        this.totalCurrentConnections.decrementAndGet();
+        trackDeviceConnectionTime(tenantId, this.totalCurrentConnections.decrementAndGet());
```
sophokles73: see above

service-base/src/main/java/org/eclipse/hono/service/metric/DeviceConnectionTimeTracker.java
kaniyan: This can be 0 but not negative. I have added now a validation to check if it is negative. If it is negative, then it throws an `IllegalArgumentException`.

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
kaniyan: I will update it.

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
kaniyan: My bad, I will correct it.

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
kaniyan: `DeviceConnectionTimeTracker` has already number of connections. The map above is to keep the reference to the gauge state object that records the connection time. 

Micrometer documentation says that "It is your responsibility to hold a strong reference to the state object that you are measuring with a Gauge.  Once the object being gauged is de-referenced and is garbage collected, Micrometer will start reporting a NaN or nothing for a gauge". This is the situation I face, if I try to maintain the reference within the `DeviceConnectionTimer`. When the number of connections are zero, null is returned and this leads to garbage collection of the gauge state object reference and resulting in NaN and henceforth the values are not recorded in Prometheus. 

To overcome this, I am maintaining the state object references in a map here in paar with maps used to store the references of the number of connections.

sophokles73: The more I think about it the less convinced I am that using a Gauge makes much sense. I think your original approach using a Timer is the right choice as it better reflects the nature of what we're doing here of reporting distinct amounts of (connection) time whereas a Gauge is more like a counter which, in our case, would need to continuously be increased by the latest amount of time. However, this way the value would increase forever until it would eventually exceed the MAX or we would need to reset it periodically. All of that doesn't really feel right so IMHO going back to the Timer is the right thing to do. Sorry for the inconvenience this has caused but I was hoping to get away with a less resource hungry alternative ...

kaniyan: Considering the lower memory footprint of _Gauge_, it was worth a try. Now I will revert to the original approach using a _Timer_.

kaniyan: I have reverted back to the _Timer_ approach and pushed those changes. Would you mind taking a look again? Meanwhile I will also update the documentation.

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -49,6 +51,10 @@
      * The name of the meter for authenticated connections.
      */
     public static final String METER_CONNECTIONS_AUTHENTICATED = "hono.connections.authenticated";
+    /**
+     * This metric is used to keep track of the connection duration of the authenticated devices to Hono.
+     */
+    public static final String METER_CONNECTIONS_DURATION_AUTHENTICATED = "hono.connections.duration.authenticated";
```
sophokles73: For aesthetic reasons I'd prefer `hono.connections.authenticated.duration` so that the names form something like a hierarchy. WDYT?

kaniyan: For me, it sounded easier to understand when `duration` is next to the `connections` namely `hono.connections.duration.authenticated`. I can revert to the old one if you prefer.

kaniyan:
> did you remove the unit tests by intention?

No, previously the existing tests failed as I used the `PeriodicStream` and it's methods. Hence I have mocked that. Now its no more needed and hence I reverted those changes and the existing tests run without error. Currently I am looking to add some unit tests for this new feature.

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
sophokles73: That would be nice.

kaniyan: I have reverted it to `hono.connections.authenticated.duration`. Please take a look again.

site/documentation/content/api/Metrics.md
```diff
@@ -74,6 +74,7 @@ Metrics provided by the protocol adapters are:
...
 | *hono.connections.unauthenticated* | Gauge               | *host*, *component-type*, *component-name*                                                   | Current number of connected, unauthenticated devices. <br/> **NB** This metric is only supported by protocol adapters that maintain *connection state* with authenticated devices. In particular, the HTTP adapter does not support this metric. |
+| *hono.connections.authenticated.duration* | Timer        | *host*, *component-type*, *component-name*, *tenant*                                         | The length of time for which the authenticated devices stay connected to Hono. <br/> **NB** This metric is only supported by protocol adapters that maintain *connection state* with authenticated devices. In particular, the HTTP adapter does not support this metric. |
```
sophokles73: how about this for description:
The overall amount of time that authenticated devices have been connected to protocol adapters. **NB** ...

kaniyan: It sounds even better. I have changed it and pushed the commit.

sophokles73: Are you also adding unit tests to this PR?

kaniyan: yes, I am working on it now.

kaniyan: I have added tests. Would you mind taking a look?

service-base/src/test/java/org/eclipse/hono/service/metric/MicrometerBasedMetricsTest.java
```diff
@@ -205,6 +206,24 @@ public void testPayloadSizeForCommandMessages(final MeterRegistry registry) {
...
+    /**
+     * Verifies that the connection time duration is recorded for the given tenant when devices get connected and
+     * disconnected.
+     *
+     * @param registry The registry that the tests should be run against.
+     */
+    @ParameterizedTest
+    @MethodSource("registries")
+    public void testConnectionTimeDuration(final MeterRegistry registry) {
+        final MeterRegistry meterRegistry = Mockito.spy(registry);
```
sophokles73: The whole point of a parameterized test is that you use the passed in object(s), isn't it? FMPOV you can verify if the meters have been updated correctly in the same way as in the existing test cases ...

service-base/src/test/java/org/eclipse/hono/service/metric/DeviceConnectionDurationTrackerTest.java
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * Sets up the fixture.
+     */
+    @BeforeEach
+    public void setup() {
+        vertx = mock(Vertx.class);
+    }
+
+    /**
+     * verifies that the recording of the connection duration is scheduled with the specified interval.
```
sophokles73: Please start with a capital letter

service-base/src/test/java/org/eclipse/hono/service/metric/DeviceConnectionDurationTrackerTest.java
```diff
@@ -0,0 +1,104 @@
...
+                .withRecordingInterval(10L)
+                .recordUsing(duration -> {
+                })
+                .start();
+
+        verify(vertx).setPeriodic(eq(10L), any(Handler.class));
+    }
+
+    /**
+     * verifies that the scheduled recording of the connection duration is cancelled
```
sophokles73: Please start with a capital letter

service-base/src/test/java/org/eclipse/hono/service/metric/MicrometerBasedMetricsTest.java
kaniyan: Ofcourse the passed in object is already being used ;-) 
As per my understanding, only a spy or a mock object can be used within the `verify` method.
Hence here a spy of the passed in object namely `meterRegistry` is created and it is used in `verify(meterRegistry, times(1)).timer(...` to verify if the `timer(..)` is being invoked.

sophokles73: Oh, my bad. I didn't notice. In any case, you nee to [be careful with spying on real objects](https://javadoc.io/doc/org.mockito/mockito-core/2.28.2/org/mockito/Mockito.html#13). I this case it should be ok though, as the *timer* method is not final.

service-base/src/test/java/org/eclipse/hono/service/metric/DeviceConnectionDurationTrackerTest.java
kaniyan: I have changed and pushed the changes regarding capital letter. Please have a look again.

sophokles73:
@kaniyan this looks good to me :+1: 
Have you also tested this with e.g. a real Prometheus server and a corresponding query for e.g. retrieving the overall connection time of all devices of tenant `DEFAULT_TENANT` during the last x hours/days?

kaniyan:
> @kaniyan this looks good to me 
> Have you also tested this with e.g. a real Prometheus server and a corresponding query for e.g. retrieving the overall connection time of all devices of tenant `DEFAULT_TENANT` during the last x hours/days?

@sophokles73 I have already tested with a real Prometheus server deployed in my local _minikube_ cluster. I could see the recording of the connection time as in the below screen shot, while a device is connected over MQTT.

image


The below query could be used to retrieve the overall connection time of all devices for the example tenant namely `DEFAULT_TENANT`during the last 7 days.

`floor( sum( increase( hono_connections_authenticated_duration_seconds_sum{ tenant="DEFAULT_TENANT"} [7d])))`

sophokles73:
@kaniyan :+1: 
just out of curiosity: why do you need the `floor( sum( increase` in front of the query?

kaniyan:
> @kaniyan 
> just out of curiosity: why do you need the `floor( sum( increase` in front of the query?

@sophokles73 To round the value down to the nearest integer. For example, the above query  without floor in my local Prometheus server is _7101.113580598749_.

sophokles73:
yes, that explains the `floor` but why do you need the `sum(increase(`? Doesn't `hono_connections_authenticated_duration_seconds_sum{ tenant="DEFAULT_TENANT"} [7d]` already contain the sum over the last 7 days?

kaniyan:
> yes, that explains the `floor` but why do you need the `sum(increase(`? Doesn't `hono_connections_authenticated_duration_seconds_sum{ tenant="DEFAULT_TENANT"} [7d]` already contain the sum over the last 7 days?

`hono_connections_authenticated_duration_seconds_sum{ tenant="DEFAULT_TENANT"} [7d]`  will return all the values that we have recorded within the last 7 days for all time series. This alone doesn't provide the total connection time for the last 7 days.

PR #1680, opened 2019-12-18
[#1673] Let checkConnected() wait on concurrent connect attempts

### LINKED ISSUE #1673: HonoConnection: Make sender/receiver creation wait on reconnect attempt
Currently, when a `HonoConnection` gets disconnected and subsequently tries to reconnect itself, all sender/receiver link creation attempts fail immediately while the reconnect is still in progress.
This is done in the `checkConnected` method, invoked when a link gets created.

<details>
<summary>See stack trace of a "not connected" error there.</summary>

~~~
07:05:57.803 [vert.x-eventloop-thread-0] DEBUG o.e.h.c.impl.CachingClientFactory - failed to create new client for [registration/mytenant]
org.eclipse.hono.client.ServerErrorException: not connected
  at org.eclipse.hono.client.impl.HonoConnectionImpl.checkConnected(HonoConnectionImpl.java:256)
  at org.eclipse.hono.client.impl.HonoConnectionImpl.checkConnected(HonoConnectionImpl.java:248)
  at org.eclipse.hono.client.impl.HonoConnectionImpl.lambda$createReceiver$21(HonoConnectionImpl.java:737)
  at org.eclipse.hono.util.HonoProtonHelper.executeOrRunOnContext(HonoProtonHelper.java:180)
  at org.eclipse.hono.client.impl.HonoConnectionImpl.executeOrRunOnContext(HonoConnectionImpl.java:229)
  at org.eclipse.hono.client.impl.HonoConnectionImpl.createReceiver(HonoConnectionImpl.java:736)
  at org.eclipse.hono.client.impl.HonoConnectionImpl.createReceiver(HonoConnectionImpl.java:717)
  at org.eclipse.hono.client.impl.HonoConnectionImpl.createReceiver(HonoConnectionImpl.java:707)
  at org.eclipse.hono.client.impl.AbstractRequestResponseClient.createReceiver(AbstractRequestResponseClient.java:409)
  at org.eclipse.hono.client.impl.AbstractRequestResponseClient.createLinks(AbstractRequestResponseClient.java:391)
  at org.eclipse.hono.client.impl.RegistrationClientImpl.create(RegistrationClientImpl.java:151)
  at org.eclipse.hono.client.impl.RegistrationClientFactoryImpl.lambda$getOrCreateRegistrationClient$1(RegistrationClientFactoryImpl.java:74)
  at org.eclipse.hono.client.impl.CachingClientFactory.getOrCreateClient(CachingClientFactory.java:174)
  at org.eclipse.hono.client.impl.CachingClientFactory.getOrCreateClient(CachingClientFactory.java:146)
  at org.eclipse.hono.client.impl.RegistrationClientFactoryImpl.lambda$getOrCreateRegistrationClient$2(RegistrationClientFactoryImpl.java:72)
  at org.eclipse.hono.util.HonoProtonHelper.executeOrRunOnContext(HonoProtonHelper.java:180)
  at org.eclipse.hono.client.impl.HonoConnectionImpl.executeOrRunOnContext(HonoConnectionImpl.java:229)
  at org.eclipse.hono.client.impl.RegistrationClientFactoryImpl.getOrCreateRegistrationClient(RegistrationClientFactoryImpl.java:71)
  at org.eclipse.hono.service.AbstractProtocolAdapterBase.getRegistrationClient(AbstractProtocolAdapterBase.java:914)
  at org.eclipse.hono.service.AbstractProtocolAdapterBase.lambda$getRegistrationAssertion$28(AbstractProtocolAdapterBase.java:950)
  at io.vertx.core.Future.lambda$compose$1(Future.java:265)
  at io.vertx.core.impl.FutureImpl.setHandler(FutureImpl.java:79)
  at io.vertx.core.Future.compose(Future.java:261)
  at org.eclipse.hono.service.AbstractProtocolAdapterBase.getRegistrationAssertion(AbstractProtocolAdapterBase.java:950)
  at org.eclipse.hono.adapter.http.AbstractVertxBasedHttpProtocolAdapter.doUploadMessage(AbstractVertxBasedHttpProtocolAdapter.java:637)
  at org.eclipse.hono.adapter.http.AbstractVertxBasedHttpProtocolAdapter.uploadTelemetryMessage(AbstractVertxBasedHttpProtocolAdapter.java:541)
  at org.eclipse.hono.adapter.http.AbstractVertxBasedHttpProtocolAdapter.uploadTelemetryMessage(AbstractVertxBasedHttpProtocolAdapter.java:515)
  at org.eclipse.hono.adapter.http.impl.VertxBasedHttpProtocolAdapter.handlePostTelemetry(VertxBasedHttpProtocolAdapter.java:306)
~~~

</details>

To prevent these failures (possibly leading to event/telemetry messages not getting delivered), we could let all `checkConnected` invocations wait on the current reconnect attempt (timing it out after the connect-timeout period).

calohmn:
This fixes #1673.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
@@ -24,6 +24,7 @@ import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
 import javax.net.ssl.SSLException;
 import javax.security.sasl.AuthenticationException;
@@ -109,6 +110,12 @@ public class HonoConnectionImpl implements HonoConnection {
     private final AtomicBoolean disconnecting = new AtomicBoolean(false);
     private final ConnectionFactory connectionFactory;
     private final Object connectionLock = new Object();
+    /**
+     * Reference to a list of promises created in {@link #checkConnected(Handler)}, waiting to be completed
+     * in a concurrent {@link #connect()} invocation. When {@link #connect()} is finished, the reference
+     * value will be set back to {@code null}.
+     */
+    private final AtomicReference<List<Promise<Void>>> checkConnectedPromises = new AtomicReference<>();
 
     private ProtonClientOptions clientOptions;
     private AtomicInteger connectAttempts;
@@ -258,6 +265,9 @@ public class HonoConnectionImpl implements HonoConnection {
 
     /**
      * {@inheritDoc}
+     * <p>
+     * If a connection attempt is currently in progress, the returned future is completed
+     * with the outcome of the connection attempt.
      */
     @Override
     public final Future<Void> isConnected() {
@@ -266,6 +276,9 @@ public class HonoConnectionImpl implements HonoConnection {
 
     /**
      * Checks if this client is currently connected to the server.
+     * <p>
+     * If a connection attempt is currently in progress, the returned future is completed
+     * with the outcome of the connection attempt.
      *
      * @return A succeeded future if this client is connected.
      */
@@ -278,9 +291,44 @@ public class HonoConnectionImpl implements HonoConnection {
     private void checkConnected(final Handler<AsyncResult<Void>> resultHandler) {
         if (isConnectedInternal()) {
             resultHandler.handle(Future.succeededFuture());
+        } else if (connecting.get()) {
+            // connect attempt in progress - let its completion complete the resultHandler here
+            log.debug("connection attempt to server [{}:{}] in progress, connection check will be completed with its result",
+                    connectionFactory.getHost(), connectionFactory.getPort());
+            final Promise<Void> promiseToAdd = Promise.promise();
+            promiseToAdd.future().setHandler(resultHandler);
+            // atomically add the promise to checkConnectedPromises - but only if checkConnectedPromises hasn't been cleared already
+            final List<Promise<Void>> newCheckConnectedPromises = checkConnectedPromises
+                    .accumulateAndGet(Collections.singletonList(promiseToAdd), (existing, toAdd) -> {
+                        // no modification of the existing list done here, keeping the accumulatorFunction function side-effect free as required
+                        if (existing == null) {
+                            return null;
+                        }
+                        final List<Promise<Void>> promises = new ArrayList<>(existing.size() + 1);
+                        promises.addAll(existing);
+                        promises.add(toAdd.get(0));
+                        return promises;
+                    });
+            if (newCheckConnectedPromises == null) {
+                // checkConnectedPromises wasn't updated with our promiseToAdd because the checkConnectedPromises were cleared in between
+                resultHandler.handle(Future.failedFuture(
+                        new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "not connected")));
+            }
         } else {
             resultHandler.handle(Future.failedFuture(
                     new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "not connected")));
+        }
+    }
+
+    private void removeAndCompleteCheckConnectedPromises(final AsyncResult<HonoConnection> connectionResult) {
+        final List<Promise<Void>> promises = checkConnectedPromises.getAndSet(null);
+        if (promises != null) {
+            final Context ctx = vertx.getOrCreateContext();
+            if (connectionResult.succeeded()) {
+                promises.forEach(promise -> ctx.runOnContext(v -> promise.complete()));
+            } else {
+                promises.forEach(promise -> ctx.runOnContext(v -> promise.fail(connectionResult.cause())));
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -279,8 +292,47 @@ private void checkConnected(final Handler<AsyncResult<Void>> resultHandler) {
...
+                resultHandler.handle(Future.failedFuture(
+                        new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "not connected")));
+            }
+        }
+    }
+
+    private void removeAndSucceedCheckConnectedPromises() {
+        final List<Promise<Void>> promises = checkConnectedPromises.getAndSet(null);
+        if (promises != null) {
+            promises.forEach(Promise::complete);
```
sophokles73: I am a little concerned about invoking all of these handlers here from the same (current) handler on the event loop (Don't block the event loop!). This is most likely not a problem for just a few handlers but since we do not know what the handlers of the promises actually do, it might be better to schedule each completion separately, i.e.:

```
promises.forEach(promise -> context.runOnContext((promise.complete()));
```
WDYT?

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -370,6 +422,9 @@ private void connect(
             final Handler<AsyncResult<HonoConnection>> connectionHandler,
             final Handler<ProtonConnection> disconnectHandler) {
 
+        // Note: in all places that the connectionHandler is completed here, the removeAndSucceedCheckConnectedPromises()
+        // or removeAndFailCheckConnectedPromises() method has to be invoked as well (completing concurrently created checkConnected() futures)!
+
```
sophokles73: what about wrapping the given connectionHandler into a handler that does both. That way we can simply continue to complete one handler only ...

calohmn: Good point. I initially thought of doing that as well, but refrained from doing so because `connect()` is invoked recursively and I didn't want to decorate/wrap the handler multiple times. But yes, an explicit handler class and an `instanceof` check before wrapping the handler can help.

sophokles73: > ... because connect() is invoked recursively ...

is it? I think that we invoke it repeatedly by means of scheduling a new handler on the event loop. But AFAIK we do not invoke it recursively. At least I hope that we are not doing that ...

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
calohmn: Yes, it's probably better. To be taken into account here is that `context` will usually be `null` in case of a failed connection attempt.

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
calohmn: I meant recursively in the sense of the logical call hierarchy (ignoring Vert.x handlers in between), concentrating on the fact that the `connectionHandler` param given to the `connect()` method (and possibly being decorated/wrapped there), possibly ends up in another `connect()` invocation.

calohmn:
@sophokles73 I've made changes corresponding to the comments above.

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -279,8 +292,52 @@ private void checkConnected(final Handler<AsyncResult<Void>> resultHandler) {
         if (isConnectedInternal()) {
             resultHandler.handle(Future.succeededFuture());
         } else {
-            resultHandler.handle(Future.failedFuture(
-                    new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "not connected")));
+            if (connecting.get()) {
```
sophokles73: what about 
```
} else if (connecting.get()) {
```

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -279,8 +292,52 @@ private void checkConnected(final Handler<AsyncResult<Void>> resultHandler) {
...
+            if (connectionResult.succeeded()) {
+                promises.forEach(promise -> runOnContextOrDirectly(v -> promise.complete()));
+            } else {
+                promises.forEach(promise -> runOnContextOrDirectly(v -> promise.fail(connectionResult.cause())));
+            }
+        }
+    }
+
+    private void runOnContextOrDirectly(final Handler<Void> action) {
+        if (context != null) {
```
sophokles73: FMPOV we can simply always do
```
vertx.getOrCreateContext().runOnContext(v -> action.handle(null));
```
I do not think that it is important, which event loop thread it is that the handler will be invoked on. If the connection has been established successfully, then any subsequent attempts to use that connection will always be executed on the connection's context anyway ...

calohmn: Ok, yes. I've amended the commit accordingly.

PR #1683, opened 2019-12-19
Pr/1072 gateway groups

eriksven:
Added a "memberOf" property to the devices endpoint in the Device Registration API. This property defines the membership of a device to a gateway group that can be used in the 'via' property of another device instead of listing the members of the group in that 'via' list.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
@@ -14,7 +14,6 @@
 package org.eclipse.hono.service.registration;
 
 import java.net.HttpURLConnection;
-import java.util.Collections;
 import java.util.Objects;
 
 import org.eclipse.hono.tracing.TracingHelper;
@@ -76,6 +75,27 @@ public abstract class AbstractRegistrationService implements RegistrationService
             Span span,
             Handler<AsyncResult<RegistrationResult>> resultHandler);
 
+
+    /**
+     * Takes the 'viaGroups' list of a device and resolve all group ids with the ids of the devices that are
+     * a member of those groups.
+     *
+     * @param tenantId The tenant the device belongs to.
+     * @param viaGroups The viaGroups list of a device. This list contains the ids of groups.
+     * @param span The active OpenTracing span for this operation. It is not to be closed in this method! An
+     *            implementation should log (error) events on this span and it may set tags and use this span as the
+     *            parent for any spans created in this method.
+     * @param resultHandler The handler to invoke with the result of the operation.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     *
+     */
+    protected abstract void resolveGroupMembers(
+            String tenantId,
+            JsonArray viaGroups,
+            Span span,
+            Handler<AsyncResult<JsonArray>> resultHandler);
+
+
     @Override
     public final void assertRegistration(
             final String tenantId,
@@ -111,7 +131,7 @@ public abstract class AbstractRegistrationService implements RegistrationService
         .compose(result -> {
             if (isDeviceEnabled(result)) {
                 final JsonObject deviceData = result.getPayload().getJsonObject(RegistrationConstants.FIELD_DATA);
-                return Future.succeededFuture(createSuccessfulRegistrationResult(tenantId, deviceId, deviceData));
+                return createSuccessfulRegistrationResult(tenantId, deviceId, deviceData, span);
             } else {
                 TracingHelper.logError(span, "device not enabled");
                 return Future.succeededFuture(RegistrationResult.from(HttpURLConnection.HTTP_NOT_FOUND));
@@ -182,7 +202,7 @@ public abstract class AbstractRegistrationService implements RegistrationService
                 }
 
                 if (isGatewayAuthorized(gatewayId, gatewayData, deviceId, deviceData)) {
-                    return Future.succeededFuture(createSuccessfulRegistrationResult(tenantId, deviceId, deviceData));
+                    return createSuccessfulRegistrationResult(tenantId, deviceId, deviceData, span);
                 } else {
                     log.debug("gateway not authorized");
                     TracingHelper.logError(span, "gateway not authorized");
@@ -210,38 +230,57 @@ public abstract class AbstractRegistrationService implements RegistrationService
      */
     protected boolean isGatewayAuthorized(final String gatewayId, final JsonObject gatewayData,
             final String deviceId, final JsonObject deviceData) {
-
         Objects.requireNonNull(gatewayId);
         Objects.requireNonNull(gatewayData);
         Objects.requireNonNull(deviceId);
         Objects.requireNonNull(deviceData);
 
-        final Object obj = deviceData.getValue(RegistrationConstants.FIELD_VIA);
-        if (obj instanceof JsonArray) {
-            // get the first matching entry
-            return ((JsonArray) obj)
-                    .stream()
-                    .filter(String.class::isInstance)
-                    .anyMatch(gatewayId::equals);
-        } else if (obj instanceof String) {
-            // compare the string directly
-            return gatewayId.equals(obj);
-        } else {
-            // wrong type -> not authorized
-            return false;
-        }
+        final JsonArray via = convertObjectToJsonArray(deviceData.getValue(RegistrationConstants.FIELD_VIA));
+        final JsonArray viaGroups = convertObjectToJsonArray(deviceData.getValue(RegistrationConstants.FIELD_VIA_GROUPS));
+        final JsonArray gatewayDataMemberOf = convertObjectToJsonArray(gatewayData.getValue(RegistrationConstants.FIELD_MEMBER_OF));
+
+        return isGatewayInVia(gatewayId, via) || anyGatewayGroupInViaGroups(gatewayDataMemberOf, viaGroups);
     }
 
-    private RegistrationResult createSuccessfulRegistrationResult(
+    private boolean isGatewayInVia(final String gatewayId, final JsonArray via) {
+        return via
+                .stream()
+                .filter(String.class::isInstance)
+                .anyMatch(gatewayId::equals);
+    }
+
+    private boolean anyGatewayGroupInViaGroups(final JsonArray gatewayDataMemberOf, final JsonArray viaGroups) {
+        return viaGroups
+                .stream()
+                .filter(String.class::isInstance)
+                .anyMatch(gatewayDataMemberOf::contains);
+    }
+
+    private JsonArray convertObjectToJsonArray(final Object object) {
+        final JsonArray array;
+        if (object instanceof JsonArray) {
+            array = (JsonArray) object;
+        } else {
+            array = new JsonArray();
+            if (object instanceof String) {
+                array.add((String) object);
+            }
+        }
+        return array;
+    }
+
+    private Future<RegistrationResult> createSuccessfulRegistrationResult(
             final String tenantId,
             final String deviceId,
-            final JsonObject deviceData) {
-
-        return RegistrationResult.from(
+            final JsonObject deviceData,
+            final Span span) {
+        return getAssertionPayload(tenantId, deviceId, deviceData, span)
+                .compose(payload -> Future.succeededFuture(RegistrationResult.from(
                 HttpURLConnection.HTTP_OK,
-                getAssertionPayload(tenantId, deviceId, deviceData),
-                getRegistrationAssertionCacheDirective(deviceId, tenantId));
+                payload,
+                getRegistrationAssertionCacheDirective(deviceId, tenantId))));
     }
+
 
     /**
      * Creates the payload of the assert Registration response message.
@@ -252,28 +291,62 @@ public abstract class AbstractRegistrationService implements RegistrationService
      * 
      * @param tenantId The tenant the device belongs to.
      * @param deviceId The device to create the assertion token for.
-     * @param registrationInfo The device's registration information.
-     * @return The payload.
+     * @param registrationInfo The device registration information.
+     * @return A Future to the payload.
      */
-    protected final JsonObject getAssertionPayload(final String tenantId, final String deviceId,
+    protected final Future<JsonObject> getAssertionPayload(final String tenantId, final String deviceId,
             final JsonObject registrationInfo) {
-
-        final JsonObject result = new JsonObject()
-                .put(RegistrationConstants.FIELD_PAYLOAD_DEVICE_ID, deviceId);
-        final JsonArray viaObj = getSupportedGatewaysForDevice(tenantId, deviceId, registrationInfo);
-        if (!viaObj.isEmpty()) {
-            result.put(RegistrationConstants.FIELD_VIA, viaObj);
-        }
-        final JsonObject defaults = registrationInfo.getJsonObject(RegistrationConstants.FIELD_PAYLOAD_DEFAULTS);
-        if (defaults != null) {
-            result.put(RegistrationConstants.FIELD_PAYLOAD_DEFAULTS, defaults);
-        }
-        return result;
+        return getSupportedGatewaysForDevice(tenantId, deviceId, registrationInfo, NoopSpan.INSTANCE)
+                .compose(via -> {
+            final JsonObject result = new JsonObject()
+                    .put(RegistrationConstants.FIELD_PAYLOAD_DEVICE_ID, deviceId);
+            if (!via.isEmpty()) {
+                result.put(RegistrationConstants.FIELD_VIA, via);
+            }
+            final JsonObject defaults = registrationInfo.getJsonObject(RegistrationConstants.FIELD_PAYLOAD_DEFAULTS);
+            if (defaults != null) {
+                result.put(RegistrationConstants.FIELD_PAYLOAD_DEFAULTS, defaults);
+            }
+            return Future.succeededFuture(result);
+        });
     }
 
     /**
-     * Checks if a gateway may act on behalf of the given device. This is determined by checking whether the
-     * {@link RegistrationConstants#FIELD_VIA} property of the device registration data has one or more entries.
+     * Creates the payload of the assert Registration response message.
+     * <p>
+     * The returned JSON object may contain the {@link RegistrationConstants#FIELD_VIA} property of the device's
+     * registration information and may also contain <em>default</em> values registered for the device under key
+     * {@link RegistrationConstants#FIELD_PAYLOAD_DEFAULTS}.
+     *
+     * @param tenantId The tenant the device belongs to.
+     * @param deviceId The device to create the assertion token for.
+     * @param registrationInfo The device registration information.
+     * @param span The active OpenTracing span for this operation. It is not to be closed in this method! An
+     *            implementation should log (error) events on this span and it may set tags and use this span as the
+     *            parent for any spans created in this method.
+     * @return A Future to the payload.
+     */
+    protected final Future<JsonObject> getAssertionPayload(final String tenantId, final String deviceId,
+                                                           final JsonObject registrationInfo, final Span span) {
+        return getSupportedGatewaysForDevice(tenantId, deviceId, registrationInfo, span)
+                .compose(via -> {
+                    final JsonObject result = new JsonObject()
+                            .put(RegistrationConstants.FIELD_PAYLOAD_DEVICE_ID, deviceId);
+                    if (!via.isEmpty()) {
+                        result.put(RegistrationConstants.FIELD_VIA, via);
+                    }
+                    final JsonObject defaults = registrationInfo.getJsonObject(RegistrationConstants.FIELD_PAYLOAD_DEFAULTS);
+                    if (defaults != null) {
+                        result.put(RegistrationConstants.FIELD_PAYLOAD_DEFAULTS, defaults);
+                    }
+                    return Future.succeededFuture(result);
+                });
+    }
+
+    /**
+     * Checks if a given device allows a gateway to possibly act on behalf of the given device. This is determined by checking whether the
+     * {@link RegistrationConstants#FIELD_VIA} property or the {@link RegistrationConstants#FIELD_VIA_GROUPS} property
+     * of the registration data of the given device have one or more entries.
      * <p>
      * Subclasses may override this method to provide a different means to determine gateway support.
      *
@@ -285,8 +358,12 @@ public abstract class AbstractRegistrationService implements RegistrationService
     protected boolean isGatewaySupportedForDevice(final String tenantId, final String deviceId,
             final JsonObject registrationInfo) {
         final Object viaObj = registrationInfo.getValue(RegistrationConstants.FIELD_VIA);
+        final Object viaGroups = registrationInfo.getValue(RegistrationConstants.FIELD_VIA_GROUPS);
+
         return (viaObj instanceof String && !((String) viaObj).isEmpty())
-                || (viaObj instanceof JsonArray && !((JsonArray) viaObj).isEmpty());
+                || (viaObj instanceof JsonArray && !((JsonArray) viaObj).isEmpty())
+                || (viaGroups instanceof String && !((String) viaGroups).isEmpty())
+                || (viaGroups instanceof JsonArray && !((JsonArray) viaGroups).isEmpty());
     }
 
     private boolean isDeviceEnabled(final RegistrationResult registrationResult) {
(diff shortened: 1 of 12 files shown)


site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -746,7 +746,12 @@ components:
                type: array
                items:
                   type: string
-               description: The device IDs of the gateways this device is assigned to.
+               description: The device IDs of the gateways and the IDs of the gateway groups that are registered to act on behalf of this device. Note that "via" and "memberOf" must not be set at the same time.
```
sophokles73: FMPOV the ambiguous usage of this field to contain both device IDs as well as group IDs will make the process of resolving the list of gateways IDs unnecessarily complicated. IMHO we should use a dedicated property for gateway groups instead, e.g. *viaGroups*. This would also make it easier to take advantage of referential integrity checks if the persistence technology supports it. I also have a hard time imagining how a UI should process this mixed list of gateway and group IDs which need to resolved differently ...

@ctron @dejanb @jbtrystram @calohmn any opinion on this?

calohmn: Yes, I agree, a separate _viaGroups_ property would make things easier and more explicit here.

service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
```diff
@@ -216,22 +237,49 @@ protected boolean isGatewayAuthorized(final String gatewayId, final JsonObject g
...
+        } else if (via instanceof String && memberOf instanceof JsonArray) {
+            return gatewayId.equals(via) || ((JsonArray) memberOf).contains(via);
+        } else if (via instanceof JsonArray && memberOf instanceof String) {
+            return isGatewayInVia(gatewayId, (JsonArray) via) || isGatewayGroupInVia((String) memberOf, (JsonArray) via);
+        } else if (via instanceof String && memberOf instanceof String) {
+            return gatewayId.equals(via) || memberOf.equals(via);
+        } else if (via instanceof JsonArray && memberOf == null) {
+            return isGatewayInVia(gatewayId, (JsonArray) via);
+        } else if (via instanceof String && memberOf == null) {
+            return gatewayId.equals(via);
```
calohmn: I think working only with "via" and "memberOf" (Json-) arrays (creating a single element array in case of a String value or an empty array in case of an empty "memberOf") would simplify the above block quite a bit.

service-base/src/test/java/org/eclipse/hono/service/registration/BaseRegistrationServiceTest.java
```diff
@@ -348,8 +498,34 @@ protected static final JsonObject getResultPayload(final String deviceId, final
...
+                    new JsonObject()
+                            .put(RegistrationConstants.FIELD_ENABLED, true)
+                            .put(RegistrationConstants.FIELD_MEMBER_OF, "group-1"));
+            return Future.succeededFuture(RegistrationResult.from(HttpURLConnection.HTTP_OK, responsePayload));
+        } else if ("gw-6".equals(deviceId)) {
+            final JsonObject responsePayload = getResultPayload(
+                    "gw-6",
+                    new JsonObject()
+                            .put(RegistrationConstants.FIELD_ENABLED, true)
+                            .put(RegistrationConstants.FIELD_MEMBER_OF, new JsonArray().add("group-1").add(("group-2"))));
```
calohmn: Superfluous parentheses (`("group-2")`).

service-base/src/test/java/org/eclipse/hono/service/registration/BaseRegistrationServiceTest.java
```diff
@@ -348,8 +498,34 @@ protected static final JsonObject getResultPayload(final String deviceId, final
...
+    private Future<JsonArray> resolveGatewayGroup(final JsonArray via) {
+        if (new JsonArray().add("group-1").equals(via)) {
+            return Future.succeededFuture(new JsonArray().add("gw-5").add("gw-6"));
+        } else {
+            return Future.succeededFuture(via);
+        }
+
+
+
+
```
calohmn: Superfluous empty lines.

site/documentation/content/api/management/device-registry-v1.yaml
eriksven: I agree too and will add the _viaGroups_ property.

sophokles73:
@eriksven your changes seem to have broken the integration tests. Can you take a look?

eriksven:
@sophokles73 Thank you for the reminder. I will have a further look into the integration tests. My first try for a fix unfortunately caused a new error.

calohmn:
@eriksven Can you resolve the conflict here?

service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
```diff
@@ -210,40 +235,63 @@ public void assertRegistration(
...
+    }
+
+    private boolean isGatewayInVia(final String gatewayId, final JsonArray via) {
+        return via
+                .stream()
+                .filter(String.class::isInstance)
+                .anyMatch(gatewayId::equals);
+    }
+
+    private boolean isGatewayInViaGroups(final JsonArray memberOf, final JsonArray viaGroups) {
```
calohmn: Some renaming would make this more understandable, IMHO. (At first sight, I thought the gatewayId is searched for in the viaGroups here). E.g.
isGatewayInViaGroups -> isGatewayGroupInViaGroups
memberOf -> gatewayDataMemberOf or gatewayGroups or gatewayDataGroups (also in calling method)

service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
```diff
@@ -210,40 +235,63 @@ public void assertRegistration(
...
+            array = new JsonArray();
+            if (object instanceof String) {
+                array.add((String) object);
+            }
         }
+        return array;
     }
 
-    private RegistrationResult createSuccessfulRegistrationResult(
+    private void createSuccessfulRegistrationResult(
```
calohmn: How about using a `Future<RegistrationResult>` return value instead of the `Handler` parameter.
That would make the code easier.

service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
```diff
@@ -271,6 +320,37 @@ protected final JsonObject getAssertionPayload(final String tenantId, final Stri
...
+     * The returned JSON object may contain the {@link RegistrationConstants#FIELD_VIA} property of the device's
+     * registration information and may also contain <em>default</em> values registered for the device under key
+     * {@link RegistrationConstants#FIELD_PAYLOAD_DEFAULTS}.
+     *
+     * @param tenantId The tenant the device belongs to.
+     * @param deviceId The device to create the assertion token for.
+     * @param registrationInfo The device registration information.
+     * @param resultHandler The handler to invoke with the result of this operation.
+     */
+    protected final void getAssertionPayload(final String tenantId, final String deviceId,
```
calohmn: It would be preferable to use a `Future<JsonObject>` return value here, instead of using the `Handler` param.
We just had a community call where we decided that we only declare the documented APIs on the Hono website as public (mailinglist announcement to follow). In that sense we can remove the old `getAssertionPayload` method above and use the `Future` return value here.

service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
```diff
@@ -335,4 +422,38 @@ protected JsonArray getSupportedGatewaysForDevice(final String tenantId, final S
...
+     * returned.
+     * <p>
+     * Subclasses may override this method to provide a different means to determine the supported gateways.
+     *
+     * @param tenantId The tenant id.
+     * @param deviceId The device id.
+     * @param registrationInfo The device's registration information.
+     * @param resultHandler The handler to invoke with the result of this operation.
+     */
+    protected void getSupportedGatewaysForDevice(final String tenantId, final String deviceId, final JsonObject registrationInfo,
```
calohmn: Same as above: It is preferable to have the `Future<JsonArray>` return value here instead of the `Handler` parameter.
The other `getSupportedGatewaysForDevice` method can be removed.

eriksven:
Thank you very much for the review. I am currently traveling for a conference and will unfortunately not be able to work on this PR before next Tuesday.

service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
```diff
@@ -210,40 +235,63 @@ public void assertRegistration(
...
                 HttpURLConnection.HTTP_OK,
-                getAssertionPayload(tenantId, deviceId, deviceData),
-                getRegistrationAssertionCacheDirective(deviceId, tenantId));
+                payload,
+                getRegistrationAssertionCacheDirective(deviceId, tenantId)))).setHandler(registrationResultInfoTracker);
     }
 
+
     /**
+     * @deprecated As of release 1.2.0 replaced by {@link #getAssertionPayload(String, String, JsonObject, Handler)}
```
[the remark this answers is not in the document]
eriksven: Removing this method with Hono 2.0.0 would be fine with me as well. But I would prefer to already set a release version in which this removal is planned.

service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
eriksven: The reason for choosing the approach with a `Handler` parameter was that this was done similar in other protected or public methods of this class such as `getDevice` or `resolveGroupMember`. Thus my intention was not to mix the approaches. But considering your description of the decision from the community call, I will change it here to  `Future<JsonObject>' which I prefer as well.

service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
```diff
@@ -314,18 +398,21 @@ protected CacheDirective getRegistrationAssertionCacheDirective(final String dev
     }
 
     /**
+     * @deprecated As of release 1.2.0, replaced by {@link #getSupportedGatewaysForDevice(String, String, JsonObject, Handler)}
```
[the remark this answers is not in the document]
eriksven: Considering the above mentioned community call decision, could we remove this deprecated method too since it is not part of the publicly documented API? To me it looks this is the case but I might forget something.

service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
```diff
@@ -335,4 +422,38 @@ protected JsonArray getSupportedGatewaysForDevice(final String tenantId, final S
         return viaObj instanceof JsonArray ? (JsonArray) viaObj : new JsonArray(Collections.emptyList());
     }
 
+    /**
+     * Gets the list of gateways that may act on behalf of the given device.
+     * <p>
+     *
+     * This default implementation gets the list of gateways from the value of the
+     * {@link RegistrationConstants#FIELD_VIA} property in the device's registration information.
+     * If a gatewayId is already known (e.g. through an assert request) this device is also included in the return value.
```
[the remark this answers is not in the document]
eriksven: You are right that is done here. So I will remove the sentence.

service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
```diff
@@ -335,4 +422,38 @@ protected JsonArray getSupportedGatewaysForDevice(final String tenantId, final S
...
+     * If no acting gateway is known and the {@link RegistrationConstants#FIELD_VIA} is empty the groups in the
+     * {@link RegistrationConstants#FIELD_VIA_GROUPS} property in the device's registration information are resolved and
+     * returned.
```
eriksven: This sentence is also not reflected in the current implementation of this PR. My proposal is to stick to the current implementation and to always resolve the groupMembers of the groups mentioned in the `memberOf´ property of the device. This mitigates possible side effects where the content of the `via` value in the response of the `assert` endpoint of the Device Registration API may change drastically depending on the presence of a value in the `via` property of the device in question. 
However, always resolving the group introduces additional overhead. In many situations this overhead may even be unnecessary if the caller only needs to identify one possible gateway and is not interested in all available devices. At the same time, a part of this overhead could be mitigated by the usage of caching mechanisms.
WDYT?

service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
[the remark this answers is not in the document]
eriksven: I did not see your comment below earlier, where already mentioned as well that the old version of `getSupportedGatewaysForDevice` can be removed. So I will remove it.

service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
calohmn: > My proposal is to stick to the current implementation and to always resolve the groupMembers of the groups mentioned in the `memberOf` property of the device.

I agree. The result of the `assertRegistration` request is cached anyway in the protocol adapters, so the additional overhead is probably not very noticable anyway, IMHO.

service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
```diff
@@ -316,23 +360,32 @@ protected CacheDirective getRegistrationAssertionCacheDirective(final String dev
...
+     * {@link RegistrationConstants#FIELD_VIA} property in the device's registration information and resolves
+     * the members of the groups that are in the {@link RegistrationConstants#FIELD_VIA_GROUPS} property of the device.
      * <p>
      * Subclasses may override this method to provide a different means to determine the supported gateways.
      *
+     *
      * @param tenantId The tenant id.
      * @param deviceId The device id.
      * @param registrationInfo The device's registration information.
      * @return The list of gateways as a JSON array of Strings (never {@code null}).
```
calohmn: This should be adapted to mention that a Future is returned here, which _may_ also fail, I guess (depending on the resolveGroupMembers implementation).

service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
```diff
@@ -252,28 +290,30 @@ private RegistrationResult createSuccessfulRegistrationResult(
      * 
      * @param tenantId The tenant the device belongs to.
      * @param deviceId The device to create the assertion token for.
-     * @param registrationInfo The device's registration information.
+     * @param registrationInfo The device registration information.
      * @return The payload.
```
calohmn: This should be adapted to mention that a Future is returned here.

service-base/src/main/java/org/eclipse/hono/service/registration/AbstractRegistrationService.java
```diff
@@ -316,23 +360,32 @@ protected CacheDirective getRegistrationAssertionCacheDirective(final String dev
...
      * Subclasses may override this method to provide a different means to determine the supported gateways.
      *
+     *
      * @param tenantId The tenant id.
      * @param deviceId The device id.
      * @param registrationInfo The device's registration information.
      * @return The list of gateways as a JSON array of Strings (never {@code null}).
      */
-    protected JsonArray getSupportedGatewaysForDevice(final String tenantId, final String deviceId,
+    protected Future<JsonArray> getSupportedGatewaysForDevice(final String tenantId, final String deviceId,
```
calohmn: How about adding a `span` parameter here (also in an additional variant of `getAssertionPayload`)? This could be used for the `resolveGroupMembers` invocation.

site/documentation/content/concepts/gateways.md
```diff
@@ -0,0 +1,24 @@
++++
+title = "Gateways"
+weight = 182
++++
+
+This page describes how gateways and groups of gateways are represented and identified throughout Hono and its APIs.
+<!--more-->
+ ---
+
+In some circumstances, a device may not directly communicate with Hono. An example is a device that does not have the capabilities to directly communicate with one of the protocol adapters.  Instead, such devices can connect to Hono through a gateway which are devices or software systems that act on behalf of other devices. From the perspective of Hono a gateway is another device following the description in [Device Identity]({{< relref "device-identity" >}}).
```
calohmn: grammar: `through a gateway which are devices or software systems`

site/documentation/content/concepts/gateways.md
```diff
@@ -0,0 +1,24 @@
...
+
+Part of the functionality provided by the  [Device Registration API]({{< relref "/api/device-registration" >}}) is to assert that a device is actually registered to act for a given device as a gateway. Thus, for instance, the protocol adapters use this API to verify that the involved devices and the combination of device and gateway are valid.
+
+In general, it possible to register multiple gateways with a device. Sometimes, for instance, while sending commands to the device, Hono needs to decide which of the possible gateways should be used to transmit that message to the device. One solution is to choose the last known gateway over which the device has communicated in the past. To store and retrieve this information the [Device Connection API]({{< relref "/api/device-connection" >}}) offers the opportunity to set and get the last known gateway for each device.
```
calohmn: "In general, it is possible"

site/documentation/content/concepts/gateways.md
```diff
@@ -0,0 +1,24 @@
...
+
+
+## Gateway Groups
+For more complex or larger scenarios like with a larger number of gateways per device or where many devices use the same gateways, it can become cumbersome to list all possible gateways in each 'via' property of the affected devices. This becomes even more complex when the gateways change frequently. To overcome such situations it is possible to define gateway groups in the Device Registry of Hono which describe a set of gateways.  
+
+To add a device to a gateway group, one can add the id of the gateway group to the list in the 'memberOf' property of that device. In the representation of the device that shall communicate of the members of a gateway group one can add the id of the group to the list in 'viaGroups' property of the device representation.
```
calohmn: > `In the representation of the device that shall communicate of the members of a gateway group one can add`

Did you mean:
`In the representation of the device that shall communicate via one of the members of a gateway group, one can add`
?

site/documentation/content/concepts/gateways.md
```diff
@@ -0,0 +1,24 @@
...
+
+To add a device to a gateway group, one can add the id of the gateway group to the list in the 'memberOf' property of that device. In the representation of the device that shall communicate of the members of a gateway group one can add the id of the group to the list in 'viaGroups' property of the device representation.       
+
+It is important to note that the concept of gateway groups is only known in the scope of the Device Registry. This means that the ids of the groups are not known or used by other services within Hono. As a consequence, the "Assert Device Registration" endpoint in the [Device Registration API]({{< relref "/api/device-registration" >}}) does not include the ids of gateway groups when it includes the registered gateways for a device in a response message. Instead, the endpoint resolves the members of the gateway groups adds them to the 'via' property of the response message. Note that the endpoint will not resolve chains of gateways of groups of groups. Thus, every potential gateways needs to be either mentioned in the 'via' property of the device in question or a direct member of a group mentioned in the 'viaGroups' property of the device in question.
```
calohmn: > `the endpoint resolves the members of the gateway groups adds them to the 'via' property of the response message`

`the endpoint resolves the members of the gateway groups *and* adds them to the 'via' property of the response message`

calohmn: > `will not resolve chains of gateways of groups of groups`

That sounds strange. How about "Will not resolve gateway groups recursively"?

calohmn: > `Thus, every potential gateways`

`Thus, every potential gateway`

site/documentation/content/concepts/gateways.md
eriksven: Yes

service-base/src/test/java/org/eclipse/hono/service/registration/AbstractRegistrationServiceTest.java
```diff
@@ -208,7 +210,10 @@ public void testGetSucceedsForRegisteredDeviceWithData(final VertxTestContext ct
 
                     assertNotNull(s.getPayload());
                     assertEquals(vias, s.getPayload().getVia());
+                    assertEquals(viaGroups, s.getPayload().getViaGroups());
 
+                    ctx.completeNow();
+                    /*
                     getRegistrationService().assertRegistration(TENANT, deviceId, ctx.succeeding(s2 -> {
```
calohmn: I guess the "getRegistrationService().assertRegistration" block can be activated again (and the above `ctx.completeNow();` removed) ?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -765,7 +765,17 @@ components:
...
+            "viaGroups":
+               type: array
+               items:
+                  type: string
+               description: The IDs of the gateway groups that are registered to act on behalf of this device. Note that "viaGroups" and "memberOf" must not be set at the same time.
+            "memberOf":
+               type: array
+               items:
+                  type: string
+               description: The IDs of the gateway groups in which this device is a member. Note that neither "via" and "memberOf" nor "viaGroups" and "memberOf" must not be set at the same time.
```
sophokles73: this looks like a new feature that a registry implementing this spec needs to support. FMPOV we should therefore bump the spec's minor version. WDYT @ctron @dejanb @calohmn

calohmn: @sophokles73 I agree. So that would be here:
https://github.com/eclipse/hono/blob/master/site/documentation/content/api/management/device-registry-v1.yaml#L39
(rebase needed first)

calohmn: Another thing regarding the description above: the double negation in the 2nd sentence sounds wrong. I'd suggest:
`Note that both "via" and "memberOf" as well as both "viaGroups" and "memberOf" must not be set at the same time.`

eriksven: @calohmn good point. I changed the formulation to two sentences.

site/documentation/content/api/device-registration/index.md
```diff
@@ -60,7 +60,7 @@ In case of a successful invocation of the operation, the body of the response me
 | Name             | Mandatory | JSON Type     | Description |
 | :--------------- | :-------: | :------------ | :---------- |
 | *device-id*      | *yes*     | *string*      | The ID of the device that is subject of the assertion. |
-| *via*            | *no*      | *array*       | The IDs (JSON strings) of gateways which may act on behalf of the device. This property MUST be set if any gateways are registered for the device. If the assertion request contained a *gateway_id* property and the response's *status* property has value `200` (indicating a successful assertion) then the array MUST at least contain the gateway ID from the request. |
+| *via*            | *no*      | *array*       | The IDs (JSON strings) of gateways which may act on behalf of the device. This property MUST be set if any gateways are registered for the device. If the assertion request contained a *gateway_id* property and the response's *status* property has value `200` (indicating a successful assertion) then the array MUST at least contain the gateway ID from the request.
```
sophokles73: the trailing `|` is missing ...

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -765,7 +765,17 @@ components:
...
+            "viaGroups":
+               type: array
+               items:
+                  type: string
+               description: The IDs of the gateway groups that are registered to act on behalf of this device. Note that "viaGroups" and "memberOf" must not be set at the same time.
+            "memberOf":
+               type: array
+               items:
+                  type: string
+               description: The IDs of the gateway groups in which this device is a member. Note that "via" and "memberOf" must not be set at the same time. The same applies for "viaGroups" and "memberOf" which must not be set at the same time too. The reason is that Eclipse Hono does not support groups of gateway groups.
```
sophokles73: can you please bump the version of the spec to `1.1.0`

eriksven: Done, I hope I understood you correctly and adapted the right version parameter.

sophokles73: you did

site/documentation/content/concepts/gateways.md
```diff
@@ -0,0 +1,29 @@
...
+
+
+## Gateway Groups
+For more complex or larger scenarios like with a larger number of gateways per device or where many devices use the same gateways, it can become cumbersome to list all possible gateways in each 'via' property of the affected devices. This becomes even more complex when the gateways change frequently. To overcome such situations it is possible to define gateway groups in the Device Registry of Hono which describe a set of gateways.  
+
+To add a device to a gateway group, one can add the id of the gateway group to the list in the 'memberOf' property of that device. In the representation of the device that shall communicate via one of the members of a gateway group one can add the id of the group to the list in 'viaGroups' property of the device representation.
```
BobClaerhout: This first sentence is a bit confusing to me. Shouldn't it be about the gateway instead of the device? It seems more logic to me to add a gateway to a gateway group but that might be a personal opinion.

sophokles73: @BobClaerhout what about something like

> To add a (gateway) device to a gateway group, one can add the id of the gateway group to the list  in the 'memberOf' property of that (gateway) device.

BobClaerhout: yes, that should do it

eriksven: Just to understand. You are both in favor of removing the first paragraph? The intention was to give a motivation of why gateway groups exist. I will reorder this part a bit and create a new PR.

BobClaerhout: No, I'm not in favor of removing it. The first sentence actually is about gateways while the documentation is talking about devices. This was a bit confusing to me.

eriksven: Alright, it looks like @sophokles73 was faster and already included the proposed change during his reordering.

PR #1688, opened 2020-01-03
[#1687] Close root span if connection got closed prematurely

### LINKED ISSUE #1687: HTTP adapter: Tracing root span may not get closed
If a HTTP request to the HTTP protocol adapter is closed by the client before it has received a response, the tracing root span created by the adapter doesn't get closed.

In the Jaeger UI, there are `<trace-without-root-span>` entries shown in such cases.

This can easily be reproduced with a longer-running request containing the `ttd` parameter when this request gets cancelled by the client before the `ttd` period has elapsed.

calohmn:
This fixes #1687.

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -292,6 +292,7 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
      * <ol>
      * <li>a handler to add a Micrometer {@code Timer.Sample} to the routing context,</li>
      * <li>a handler and failure handler that creates tracing data for all server requests,</li>
+     * <li>a handler to log when the connection is closed prematurely,</li>
      * <li>a default failure handler,</li>
      * <li>a handler limiting the body size of requests to the maximum payload size set in the <em>config</em>
      * properties,</li>
@@ -315,16 +316,24 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
         final TracingHandler tracingHandler = createTracingHandler();
         matchAllRoute.handler(tracingHandler).failureHandler(tracingHandler);
 
-        // 3. default handler for failed routes
+        // 3. handler to log when the connection is closed prematurely
+        matchAllRoute.handler(ctx -> {
+            if (!ctx.response().closed() && !ctx.response().ended()) {
+                ctx.response().closeHandler(v -> logResponseGettingClosedPrematurely(ctx));
+            }
+            ctx.next();
+        });
+
+        // 4. default handler for failed routes
         matchAllRoute.failureHandler(new DefaultFailureHandler());
 
-        // 4. BodyHandler with request size limit
+        // 5. BodyHandler with request size limit
         log.info("limiting size of inbound request body to {} bytes", getConfig().getMaxPayloadSize());
         final BodyHandler bodyHandler = BodyHandler.create(DEFAULT_UPLOADS_DIRECTORY)
                 .setBodyLimit(getConfig().getMaxPayloadSize());
         matchAllRoute.handler(bodyHandler);
 
-        // 5. handler to set the trace sampling priority
+        // 6. handler to set the trace sampling priority
         Optional.ofNullable(getTenantTraceSamplingHandler())
                 .ifPresent(tenantTraceSamplingHandler -> matchAllRoute.handler(tenantTraceSamplingHandler));
         return router;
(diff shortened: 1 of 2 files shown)


calohmn:
The root span created by the `io.opentracing.contrib.vertx.ext.web.TracingHandler` is usually finished via a `bodyEndHandler` that gets added to the `RoutingContext` there. However, if a connection gets closed by the client _before_ the HTTP response got sent, the `bodyEndHandler` doesn't get invoked. Therefore, in such a case the span needs to be finished manually.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -860,18 +879,21 @@ protected Integer getTimeUntilDisconnectFromRequest(final RoutingContext ctx) {
...
             final String deviceId,
             final Span currentSpan) {
 
-        if (messageConsumer != null && !ctx.response().closed()) {
+        if (messageConsumer != null && !ctx.response().closed() && !ctx.response().ended()) {
             ctx.response().closeHandler(v -> {
                 log.debug("device [tenant: {}, device-id: {}] closed connection before response could be sent",
                         tenantId, deviceId);
                 currentSpan.log("device closed connection");
+                TracingHelper.logError(currentSpan, "cannot send HTTP response to device: response already closed");
```
sophokles73: can we merge this with the previous message being logged to the current span?

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -860,18 +879,21 @@ protected Integer getTimeUntilDisconnectFromRequest(final RoutingContext ctx) {
      * @param deviceId The identifier of the device.
      * @param currentSpan The <em>OpenTracing</em> Span used for tracking the processing of the request.
      */
-    private void addConnectionCloseHandler(
+    private void setTtdRequestConnectionCloseHandler(
```
sophokles73: this handler seems to be useful/required for any request, not just TTD requests, doesn't it?

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -315,16 +316,24 @@ protected Router createRouter() {
         final TracingHandler tracingHandler = createTracingHandler();
         matchAllRoute.handler(tracingHandler).failureHandler(tracingHandler);
 
-        // 3. default handler for failed routes
+        // 3. handler to ensure tracing root span is finished if the request gets closed prematurely
+        matchAllRoute.handler(ctx -> {
+            if (!ctx.response().closed() && !ctx.response().ended()) {
+                ctx.response().closeHandler(v -> finishRootSpanOnResponseGettingClosedPrematurely(ctx));
```
sophokles73: the close handler that is set here will be overwritten later by the *doUploadMessage* method, right?
Why don't we set a failure Handler here instead which finishes the root span and modify the *doUploadMessage* method to fail the RoutingContext e.g. with an IllegalStateException if the request is being closed before we can send the response or if sending the response fails?

calohmn: > the close handler that is set here will be overwritten later by the doUploadMessage method, right?

Yes, that's intended.

> Why don't we set a failure Handler here instead which finishes the root span and modify the doUploadMessage method to fail the RoutingContext e.g. with an IllegalStateException if the request is being closed before we can send the response or if sending the response fails?

Hmm, I'll think about that. I'm currently doing some more testing.

calohmn:
@sophokles73 I've found another solution here. The problem actually was, that in the cases where the root span didn't get finished, the `response.end()` method hadn't been called (or in another case `context.fail()` didn't get called because of an unhandled exception). Even if a response was closed prematurely, calling `end()` or `context.fail()` will call the bodyEndHandler or failureHandler set by the `TracingHandler`.
So, making sure `end()` or `fail()` gets called is enough here.

service-base/src/main/java/org/eclipse/hono/service/http/HttpUtils.java
```diff
@@ -181,7 +181,7 @@ public static void failWithHeaders(
         if (ctx.failed() || ctx.response().ended()) {
             // nothing to do
         } else {
-            if (headers != null) {
+            if (headers != null && !ctx.response().closed()) {
```
sophokles73: shouldn't this better be part of the condition in the first if statement?

calohmn: No, if the response is closed but not failed/ended yet, the `fail` method should get called here (which then also causes the `TracingHandler` failureHandler to get called).

The extra `closed` check above is to ensure that the `putHeader` invocation doesn't throw an exception. It would do so for a closed response using Vert.x 3.7.1 (Hono 1.0). In Vert.x 3.8.4 (Hono master) this check isn't strictly needed any more, but I think the extra condition doesn't hurt here.

sophokles73: I see, then adding a corresponding comment would improve readability a lot here FMPOV.

calohmn: Yes, I've added comments.

PR #1690, opened 2020-01-03
Add HonoConnection#getRemoteContainer, use it in client span

calohmn:
This adds a tag to tracing spans created by `AbstractHonoClient` implementations, containing the remote container id of the AMQP connection (as advertised in the `open` frame by the peer).

For that, a `HoneConnection#getRemoteContainer` method is added.

The remote container id information in traces can be useful in cases where the AMQP messaging network is represented by multiple containers behind a load balancer. For debugging error scenarios, it can be helpful to know to which container exactly a connection has been established.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/HonoConnection.java
+++ b/client/src/main/java/org/eclipse/hono/client/HonoConnection.java
@@ -219,6 +219,17 @@ public interface HonoConnection extends ConnectionLifecycle<HonoConnection> {
     boolean isShutdown();
 
     /**
+     * Gets the remote container id as advertised by the peer.
+     * <p>
+     * This default implementation simply returns {@code N/A}.
+     *
+     * @return The remote container id or {@code null}.
+     */
+    default String getRemoteContainer() {
+        return "N/A";
+    }
+
+    /**
      * Checks if this client supports a certain capability.
      * <p>
      * The result of this method should only be considered reliable if this client is connected to the server.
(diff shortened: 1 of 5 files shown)


client/src/main/java/org/eclipse/hono/client/HonoConnection.java
```diff
@@ -218,6 +218,15 @@ static HonoConnection newConnection(final Vertx vertx, final ClientConfigPropert
...
     boolean isShutdown();
 
+    /**
+     * Gets the remote container id as advertised by the peer.
+     * <p>
+     * Returns {@code null} if the client is currently not connected.
+     *
+     * @return The remote container id or {@code null}.
+     */
+    String getRemoteContainer();
```
sophokles73: IMHO this should be a default implementation returning e.g. `N/A` in order to remain backwards compatible ...

client/src/main/java/org/eclipse/hono/client/HonoConnection.java
```diff
@@ -218,6 +218,15 @@ static HonoConnection newConnection(final Vertx vertx, final ClientConfigPropert
      */
     boolean isShutdown();
 
+    /**
+     * Gets the remote container id as advertised by the peer.
+     * <p>
+     * Returns {@code null} if the client is currently not connected.
```
sophokles73: I do not think that we should put that much semantics into the returned value in order to not encourage users to actually test connectivity using this property ...

calohmn:
@sophokles73 I've amended the commit accordingly.

PR #1693, opened 2020-01-07
Add Auto-Provisioning to Device Registry + Integration Tests

b-abel:
This PR completes the auto-provisioning (#1296) feature. It

1. adds support to the AMQP adapter
2. adds an implementation of the feature to the example device registry
3. adds integration tests for this feature for the AMQP, MQTT and HTTP adapters.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/AbstractBaseApplication.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/AbstractBaseApplication.java
@@ -14,6 +14,7 @@
 package org.eclipse.hono.service;
 
 import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
@@ -277,6 +278,9 @@ public abstract class AbstractBaseApplication implements ApplicationRunner {
      * @param provider The provider of the health checks.
      */
     protected final void registerHealthchecks(final HealthCheckProvider provider) {
-        healthCheckServer.registerHealthCheckResources(provider);
+        Optional.ofNullable(provider).ifPresent(p -> {
+            log.debug("registering health checks [provider: {}]", p.getClass().getName());
+            healthCheckServer.registerHealthCheckResources(p);
+        });
     }
 }
--- a/services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/Application.java
+++ b/services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/Application.java
@@ -63,10 +63,15 @@ public class Application extends AbstractApplication {
      * 
      * @param service The service.
      * @throws NullPointerException if service is {@code null}.
+     * @throws IllegalArgumentException if the given service is not a {@code Verticle}.
      */
     @Autowired
     public void setAuthenticationService(final AuthenticationService service) {
-        this.authService = Objects.requireNonNull(service);
+        Objects.requireNonNull(service);
+        if (!(service instanceof Verticle)) {
+            throw new IllegalArgumentException("authentication service must be a vert.x Verticle");
+        }
+        this.authService = service;
     }
 
     /**
@@ -85,7 +90,7 @@ public class Application extends AbstractApplication {
             log.info("deploying component [{}]", component.getClass().getName());
             getVertx().deployVerticle((Verticle) component, result);
             return result.future().map(id -> {
-                registerHealthCheckProvider(serviceImplementation);
+                registerHealthCheckProvider(component);
                 return id;
             });
         } else {
@@ -106,7 +111,6 @@ public class Application extends AbstractApplication {
 
     private void registerHealthCheckProvider(final Object obj) {
         if (obj instanceof HealthCheckProvider) {
-            log.debug("registering health check provider [{}]", obj.getClass().getName());
             registerHealthchecks((HealthCheckProvider) obj);
         }
     }


services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedDeviceBackend.java
```diff
@@ -186,7 +194,66 @@ public final void get(final String tenantId, final String type, final String aut
     public void get(final String tenantId, final String type, final String authId, final JsonObject clientContext,
             final Span span,
             final Handler<AsyncResult<CredentialsResult<JsonObject>>> resultHandler) {
-        credentialsService.get(tenantId, type, authId, clientContext, span, resultHandler);
+        credentialsService.get(tenantId, type, authId, clientContext, span, ar -> {
+            if (ar.result().getStatus() == HttpURLConnection.HTTP_NOT_FOUND
+                    && isAutoProvisioningEnabled(type, clientContext)) {
+                provisionDevice(tenantId, authId, clientContext, span, resultHandler);
+            } else {
+                resultHandler.handle(Future.succeededFuture(ar.result()));
```
sophokles73: `resultHandler.handle(ar)` should do the trick, shouldn't it?

service-base/src/main/java/org/eclipse/hono/service/management/device/AutoProvisioningEnabledDeviceBackend.java
```diff
@@ -0,0 +1,90 @@
...
+
+                    final String deviceId = r.getPayload().getId();
+
+                    final Promise<OperationResult<Void>> credPromise = Promise.promise();
+                    set(tenantId, deviceId, Optional.empty(), List.of(certCredential), span, credPromise);
+                    return credPromise.future()
+                            .map(v -> {
+                                if (v.isError()) {
+                                    return OperationResult.ok(v.getStatus(),
+                                            "Auto-provisioning failed: credentials could not be set for device ["
```
sophokles73: shouldn't the creation of the device identity be reverted in this case? Otherwise, we would end up with a *half-provisioned* device, wouldn't we?

service-base/src/main/java/org/eclipse/hono/service/management/device/AutoProvisioningService.java
```diff
@@ -0,0 +1,50 @@
...
+import io.opentracing.Span;
+import io.vertx.core.Future;
+
+/**
+ * A service for automatically provisioning devices.
+ *
+ * @see <a href="https://www.eclipse.org/hono/docs/dev/concepts/device-provisioning/#automatic-device-provisioning">
+ *      Automatic Device Provisioning</a>
+ */
+public interface AutoProvisioningService {
```
sophokles73: this interface doesn't seem to be used anywhere except for the `AutoProvisioningEnabledDeviceBackend` ... 
Is it needed for anything else?

service-base/src/main/java/org/eclipse/hono/service/auth/device/SubjectDnCredentials.java
```diff
@@ -43,6 +43,7 @@ private SubjectDnCredentials(final String tenantId, final String authId, final J
      * @param subjectDn The subject DN of the device's client certificate.
      * @return The credentials.
      * @throws NullPointerException if any of the parameters are {@code null}.
+     * @deprecated Use {@link #create(String, String, JsonObject)} instead.
```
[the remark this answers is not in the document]
b-abel: Yes, an empty client context is still valid. Since I have added a new version with a client context parameter for each *create* method, I think the old methods should be removed: a) the new methods are more explicit, b) the old ones are no longer used by Hono, c) if you do use them, it is harder to tell if it is intended not to provide the client context (in this case the AMQP adapter actually used one of them and therefore inadvertently does not supported auto-provisioning).

My general thoughts on this topic: It is not yet decided which part of the code should be seen as API. By adding new versions of the methods that expose the client context, the old versions have become convenient methods. I prefer not to have many convenient methods in an API, as they may not age well and add unnecessary complexity. If we just add new code, we will be faced with a massive increase in complexity without, as in this case, an increased set of features.
If it is not part of the API, the unused methods could be removed immediately.

service-base/src/main/java/org/eclipse/hono/service/management/device/AutoProvisioningService.java
b-abel: I think an `AutoProvisioningService` that is not a `DeviceBackend` could be possible. But tbh, it just follows the very modular structure of interfaces within the Device Registry. Which to my understanding should provide great flexibility for *other* implementations, while their usage cannot be fully derived from the implementation of the example device registry.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedDeviceBackend.java
```diff
@@ -186,7 +194,66 @@ public final void get(final String tenantId, final String type, final String aut
     public void get(final String tenantId, final String type, final String authId, final JsonObject clientContext,
             final Span span,
             final Handler<AsyncResult<CredentialsResult<JsonObject>>> resultHandler) {
-        credentialsService.get(tenantId, type, authId, clientContext, span, resultHandler);
+        credentialsService.get(tenantId, type, authId, clientContext, span, ar -> {
+            if (ar.result().getStatus() == HttpURLConnection.HTTP_NOT_FOUND
```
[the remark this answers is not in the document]
b-abel: Yes, but the code in this section only uses successful futures, while errors are encoded in the payload (e.g. 'CredentialsResult'). But anyway, I can add the check to secure this.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedDeviceBackend.java
b-abel: In the current implementation of the `CredentialsService', the AsyncResult will always be successful, but if we assume that this could be changed (as discussed above), then your proposal would change the behavior, wouldn't it? So IMHO we should either leave this as it is *or* add the check if the AsyncResult was successful. WDYT?

service-base/src/main/java/org/eclipse/hono/service/auth/device/SubjectDnCredentials.java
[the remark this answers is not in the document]
sophokles73: I see your point but in this case I also do not see any risk of overloading the class with convenience factory methods. On the contrary, I do not see why I should need to pass an empty JsonObject if there is no client context ...

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedDeviceBackend.java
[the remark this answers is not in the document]
sophokles73: Since the `CredentialsService` interface does not explicitly state that the handler will never be invoked with a failed `AsyncResult`, I think you should. This will make the code more robust against erroneous implementations anyway.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedDeviceBackend.java
sophokles73: I do not think that we are changing any behavior here because the `AsyncResult` can already fail FMPOV.

service-base/src/main/java/org/eclipse/hono/service/management/device/AutoProvisioningService.java
sophokles73: In that case, why would anybody want to implement just `AutoProvisioningService` if this interface is not used anywhere?

b-abel: I removed the interface.

service-base/src/test/java/org/eclipse/hono/service/management/device/AutoProvisioningEnabledDeviceBackendTest.java
```diff
@@ -0,0 +1,95 @@
...
+import io.vertx.core.Promise;
+import io.vertx.core.net.SelfSignedCertificate;
+import io.vertx.junit5.VertxExtension;
+import io.vertx.junit5.VertxTestContext;
+
+/**
+ * Tests verifying behavior of {@link AutoProvisioningEnabledDeviceBackend}.
+ */
+@ExtendWith(VertxExtension.class)
+class AutoProvisioningEnabledDeviceBackendTest {
```
sophokles73: can you please make this public as all our other test classes?

service-base/src/test/java/org/eclipse/hono/service/management/device/AutoProvisioningEnabledDeviceBackendTest.java
```diff
@@ -0,0 +1,95 @@
...
+        final SelfSignedCertificate ssc = SelfSignedCertificate.create("eclipse.org");
+        final CertificateFactory factory = CertificateFactory.getInstance("X.509");
+        cert = (X509Certificate) factory.generateCertificate(new FileInputStream(ssc.certificatePath()));
+    }
+
+    /**
+     * Verifies that a device is created and credentials are set for it when auto-provisioning a device.
+     */
+    @Test
+    void provisionDevice(final VertxTestContext ctx) {
```
sophokles73: public
we usually use a naming pattern like
```
testProvisionDeviceDoesSomething()
```

service-base/src/test/java/org/eclipse/hono/service/management/device/AutoProvisioningEnabledDeviceBackendTest.java
```diff
@@ -0,0 +1,95 @@
...
+        final Future<OperationResult<String>> result = underTest.provisionDevice(TENANT_ID, cert, NoopSpan.INSTANCE);
+
+        // THEN the device is created and credentials are set
+        result.setHandler(ctx.succeeding(ok -> {
+            verify(underTest).createDevice(eq(TENANT_ID), any(), any(), any(), any(Promise.class));
+            verify(underTest).set(eq(TENANT_ID), eq(DEVICE_ID), any(), any(), any(), any(Promise.class));
+            ctx.completeNow();
+        }));
+    }
+
```
sophokles73: what about a test for verifying that the device identity is removed if creating the credentials fails?

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/Application.java
```diff
@@ -39,6 +39,7 @@
 @ComponentScan("org.eclipse.hono.service.auth")
 @ComponentScan("org.eclipse.hono.service.metric")
 @ComponentScan("org.eclipse.hono.deviceregistry")
+@ComponentScan("org.eclipse.hono.service.management.device")
```
sophokles73: why do we need this additional annotation?

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpConnectionIT.java
```diff
@@ -87,6 +87,26 @@ public void testConnectSucceedsForRegisteredDevice(final VertxTestContext ctx) {
...
+        final SelfSignedCertificate deviceCert = SelfSignedCertificate.create(UUID.randomUUID().toString());
+
+        helper.getCertificate(deviceCert.certificatePath())
+                .compose(cert -> {
+                    final var tenant = Tenants.createTenantForTrustAnchor(cert);
+                    tenant.getTrustedCertificateAuthorities().get(0).setAutoProvisioningEnabled(true);
+                    return helper.registry.addTenant(tenantId, tenant);
+                })
+                .compose(ok -> connectToAdapter(deviceCert))
+                .setHandler(ctx.completing());
```
sophokles73: there seem to be assertions missing here:
* device identity has been created
* credentials have been created

service-base/src/test/java/org/eclipse/hono/service/management/device/AutoProvisioningEnabledDeviceBackendTest.java
b-abel: Sure. I was not aware of this convention. Since public classes and methods are no longer necessary with JUnit 5 and I found a couple of tests in Hono that are not public, I assumed we would move in that direction.

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpConnectionIT.java
b-abel: Wouldn't that belong to an integration test of the device registry? Here I would rather add a test that the connect *does not* succeed if auto-provisioning is disabled for the trust nachor.

sophokles73: That  would make sense. Do you have a corresponding IT for the registry?

service-base/src/test/java/org/eclipse/hono/service/management/device/AutoProvisioningEnabledDeviceBackendTest.java
sophokles73: Do we? I was not aware of this. In any case, the checkstyle rules for JavaDoc do not kick in for package private classes and methods but IMHO we should enforce proper JavaDocs for test cases as well ...

b-abel: Maybe we could change the checkstyle rules to enforce JavaDocs for package private classes and methods in the tests like this https://stackoverflow.com/a/25898007/7740194 
I like the possibility to have less boilerplate and visual clutter and there is no real need for a higher visibility level. WDYT?

sophokles73: I think that this is a lot of effort for you to be able to save typing *public* ;-)
When I create a test case, the (public) class is automatically generated by Eclipse IDE and I only need to add the test methods ...

b-abel: > When I create a test case, the (public) class is automatically generated by Eclipse IDE and I only need to add the test methods ...

I'm sorry for you, when I create a (JUnit 5) test case, the package private class is automatically generated by the IntelliJ IDE along with the package private test methods and I just have to add the logic ;-) 
Saving typing is not really worth it. Having a cleaner code might be worth it.

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpConnectionIT.java
b-abel: I'm working on it

service-base/src/test/java/org/eclipse/hono/service/management/device/AutoProvisioningEnabledDeviceBackendTest.java
sophokles73: I do not really see why the code would become *cleaner* if all classes and/or methods are package private instead of public. But maybe the real problem is that you cannot configure IntelliJ to create them as public in the first place? ;-)

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/Application.java
b-abel: My bad. We do not need this with the current solution.

b-abel:
I pushed an update with the requested changes.

service-base/src/test/java/org/eclipse/hono/service/management/device/AutoProvisioningEnabledDeviceBackendTest.java
```diff
@@ -92,4 +101,45 @@ void provisionDevice(final VertxTestContext ctx) {
...
+            promise.complete(OperationResult.empty(403)); // creation of credentials fails
+            return null;
+        }).when(underTest).set(any(), any(), any(), any(), any(), any(Promise.class));
+
+        // WHEN provisioning a device from a certificate
+        final Future<OperationResult<String>> result = underTest.provisionDevice(TENANT_ID, cert, NoopSpan.INSTANCE);
+
+        // THEN the device is deleted
+        result.setHandler(ctx.succeeding(ok -> {
+            verify(underTest).deleteDevice(eq(TENANT_ID), eq(DEVICE_ID), any(), any(), any(Promise.class));
```
sophokles73: this needs to go into a
```
ctx.verify(() -> {
  ...
});
```
block in order for the ctx to fail in case of an assertion error.

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpConnectionIT.java
```diff
@@ -107,6 +107,35 @@ public void testAutoProvisioning(final VertxTestContext ctx) {
...
     }
 
+    /**
+     * Verifies that the adapter rejects connection attempts from an unknown device for which auto-provisioning is
+     * disabled.
+     *
+     * @param ctx The test context
+     */
+    @Test
+    public void testConnectFailsIfAutoProvisioningIsDisabled(final VertxTestContext ctx) {
```
sophokles73: not sure but don't we already have a test for an unknown device's connect attempt? This is basically the same thing, isn't it?

b-abel: Yeah, it _should_ be the same. The protocol adapter checks whether automatic provisioning is enabled. This checks that automatic provisioning is not requested from the device registry by the protocol adapter if it is disabled for the trust anchor used.

sophokles73:
@dejanb @calohmn @ctron would you also like to take a look? Otherwise I'd like to merge this PR ...

b-abel:
I don't know what the problem with the ECA check is. It's very strange. Revalidation did not help and my Eclipse account looks OK. Every commit has had the same footer for years. The summary above says "1 error, 2 failures." So I'm assuming the check didn't fail but caused an error. 

Let's see if the problem goes away when I squash the commits.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/AmqpAdapterSaslAuthenticatorFactory.java
```diff
@@ -384,9 +386,20 @@ public boolean succeeded() {
                         }
                     })
                     .compose(tenant -> {
+                        final JsonObject clientContext = new JsonObject();
+                        final String issuerDn = deviceCert.getIssuerX500Principal().getName(X500Principal.RFC2253);
+                        if (tenant.isAutoProvisioningEnabled(issuerDn)) {
+                            try {
+                                clientContext.put(CredentialsConstants.FIELD_CLIENT_CERT, deviceCert.getEncoded());
+                            } catch (CertificateEncodingException e) {
```
calohmn: final

sophokles73:
@b-abel can you squash and rebase?

b-abel:
I am currently fixing this.

b-abel:
Replaced by #1706.

PR #1699, opened 2020-01-10
[#1691] Add isConnected(timeout); use it in client factories

### LINKED ISSUE #1691: HonoConnection: add timeout for connection check
Followup of #1673:

As implemented for #1673, the connection check in the HonoConnectionImpl class (`checkConnected` method) now waits for an ongoing connection attempt to finish.
It is implemented in such a way that this wait period spans across multiple reconnect attempts.

If the configured number of reconnect attempts is not limited (which is the default), there is no timeout mechanism currently in place for the `checkConnected` method.

In practice, this means that in case that all device registry instances become unavailable, the first telemetry/event request will _not_ time out. (Subsequent requests will time out, internally producing an "already trying to create a client for .." error.)

To solve this, I suggest letting `checkConnected` wait at most the configured `connect timeout` period (configured in ClientConfigProperties).
(An alternative would be to let `checkConnected` only wait for _one_ reconnect attempt. But in practise, upon shutting down of one device registry instance, there could be some quickly coming "connection refused" errors for the reconnect attempts before another device registry instance is targeted, therefore doing multiple reconnect attempts would be better.)

calohmn:
This fixes #1691.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClientFactory.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClientFactory.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2019, 2020 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -21,6 +21,7 @@ import org.eclipse.hono.client.DisconnectListener;
 import org.eclipse.hono.client.HonoConnection;
 import org.eclipse.hono.client.ReconnectListener;
 import org.eclipse.hono.client.ServerErrorException;
+import org.eclipse.hono.config.ClientConfigProperties;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -93,6 +94,37 @@ abstract class AbstractHonoClientFactory implements ConnectionLifecycle<HonoConn
     @Override
     public final Future<Void> isConnected() {
         return connection.isConnected();
+    }
+
+    /**
+     * Checks whether this client is connected to the service.
+     * <p>
+     * If a connection attempt is currently in progress, the returned future is completed
+     * with the outcome of the connection attempt. If the connection attempt (including
+     * potential reconnect attempts) isn't finished after the given timeout, the returned
+     * future is failed.
+     * <p>
+     * Simply delegates to {@link HonoConnection#isConnected(long)}.
+     * 
+     * @param waitForCurrentConnectAttemptTimeout The maximum number of milliseconds to wait for
+     *                                            an ongoing connection attempt to finish.
+     * @return A succeeded future if this factory is connected.
+     *         Otherwise, the future will be failed with a {@link ServerErrorException}.
+     */
+    @Override
+    public final Future<Void> isConnected(final long waitForCurrentConnectAttemptTimeout) {
+        return connection.isConnected(waitForCurrentConnectAttemptTimeout);
+    }
+
+    /**
+     * Gets the default timeout used when checking whether this client is connected to the service.
+     * <p>
+     * The value returned here is the {@link ClientConfigProperties#getLinkEstablishmentTimeout()}.
+     *
+     * @return The timeout value in milliseconds.
+     */
+    public final long getDefaultConnectionCheckTimeout() {
+        return connection.getConfig().getLinkEstablishmentTimeout();
     }
 
     /**
--- a/client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientFactoryImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientFactoryImpl.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2019, 2020 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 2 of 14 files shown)


client/src/main/java/org/eclipse/hono/client/impl/AbstractHonoClientFactory.java
```diff
@@ -95,6 +95,26 @@ public void addReconnectListener(final ReconnectListener<HonoConnection> listene
...
+     * Simply delegates to {@link HonoConnection#isConnected(long)}.
+     * 
+     * @param waitForCurrentConnectAttemptTimeout The maximum number of milliseconds to wait for
+     *                                            on ongoing connection attempt to finish.
+     * @return A succeeded future if this factory is connected.
+     *         Otherwise, the future will be failed with a {@link ServerErrorException}.
+     */
+    @Override
+    public final Future<Void> isConnected(final long waitForCurrentConnectAttemptTimeout) {
+        return connection.isConnected();
```
sophokles73: shouldn't this be `connection.isConnected(waitForCurrentConnectionAttemptTimeout)`?

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -291,44 +286,82 @@ public final void addReconnectListener(final ReconnectListener<HonoConnection> l
...
-                // checkConnectedPromises wasn't updated with our promiseToAdd because the checkConnectedPromises were cleared in between
-                resultHandler.handle(Future.failedFuture(
-                        new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "not connected")));
+            final CheckConnectedPromise promiseToAdd = new CheckConnectedPromise(resultHandler);
+            if (!addToCheckConnectedPromises(promiseToAdd)) {
+                // checkConnectedPromises has been cleared in between
+                checkConnected(resultHandler);
+            } else {
+                // promise added to list, now use a timer to ensure that we wait no more than the given timeout
+                promiseToAdd.setTimerId(vertx.setTimer(waitForCurrentConnectAttemptTimeout, id -> {
```
sophokles73: shouldn't this better be encapsulated in `CheckConnectPromise` (which might be renamed to `ExpiringPromise` according to its purpose)?

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -291,44 +286,82 @@ public final void addReconnectListener(final ReconnectListener<HonoConnection> l
...
+        final List<CheckConnectedPromise> promises = checkConnectedPromises.getAndSet(null);
+        if (promises != null && !promises.isEmpty()) {
+            log.trace("completing {} accumulated connection checks", promises.size());
             final Context ctx = vertx.getOrCreateContext();
-            if (connectionResult.succeeded()) {
-                promises.forEach(promise -> ctx.runOnContext(v -> promise.complete()));
-            } else {
-                promises.forEach(promise -> ctx.runOnContext(v -> promise.fail(connectionResult.cause())));
-            }
+            promises.forEach(promise -> ctx.runOnContext(v -> promise.tryCompleteAndCancelTimer(connectionResult)));
```
sophokles73: how about moving all of this into a separate class?

client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientFactoryImpl.java
```diff
@@ -67,17 +67,18 @@ protected void onDisconnect() {
...
-                    CredentialsClientImpl.getTargetAddress(tenantId),
-                    () -> CredentialsClientImpl.create(
-                            cacheProvider,
-                            connection,
-                            tenantId,
-                            this::removeCredentialsClient,
-                            this::removeCredentialsClient),
-                    result);
-        });
+        return connection.isConnected(connection.getConfig().getRequestTimeout())
```
sophokles73: in all the other places we use the request timeout for limiting the time to wait for the processing of a request over an already established connection, don't we? So, in order to improve chances that the request, that his method invocation is part of, succeeds, we might want to slightly reduce the time to wait a little so that there is some time left to establish the link and send the request message?

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
calohmn: I'll have a look...

client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientFactoryImpl.java
calohmn: That would only be needed if we had another timeout mechanism around the combind "getOrCreateClient+sendRequest" invocations, with a timeout value of requestTimeout+X.

But I don't see such a timeout mechanism being in place and therefore the number of milliseconds to subtract here would just be arbitrary. And then I would prefer to keep using the full request timeout value.

And thinking of the default value of 200ms here, I wouldn't want to decrease that value further (I think a somewhat higher value would even be better in many cases).

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
calohmn: Yes, I've refactored it accordingly.

client/src/main/java/org/eclipse/hono/client/impl/CredentialsClientFactoryImpl.java
sophokles73: You are right in pointing out that the request timeout only covers the time it takes to send the request and receive the response over an already established link. In fact, when a client invokes this method and the connection is established but the link for the given tenant is not established, then the client will wait for at most `ClientConfigProperties.linkEstablishmentTimeout` milis for the link to be established before timing out.

So, if the connection is *not* established but an attempt to connect is already going on, I guess that if we allow for roughly the same amount of time to establish the connection, the client should get a fair chance to get its request fulfilled but should not be required to wait exceptionally long for a failed result. WDYT?

calohmn: Yes, I think that would be a good choice. I've adapted the commit accordingly.

client/src/main/java/org/eclipse/hono/client/impl/DeferredConnectionCheckHandler.java
```diff
@@ -0,0 +1,205 @@
...
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.Promise;
+import io.vertx.core.Vertx;
+
+/**
+ * Handles connection checks that should be completed only when a concurrent connection attempt has finished
+ * (or when a timeout period has elapsed).
+ */
+public class DeferredConnectionCheckHandler {
```
sophokles73: final ?

PR #1700, opened 2020-01-10
[#1336] Extract span context from incoming MQTT message

### LINKED ISSUE #1336: Support providing a trace context along with MQTT messages
In order to have a complete overview of a telemetry/event message lifecycle in a tracing solution, it would be good to have the message related trace spans created in a protocol adapter be combined with trace spans created in the originating device/gateway. For that, the message reaching the protocol adapter must have trace context information and the protocol adapter must use these.

The HTTP adapter already supports having a OpenTracing span context parameter provided via the HTTP headers of an incoming message.

A corresponding feature should be added to the MQTT adapter, providing support for extracting trace context information from a "property bag" at the end of the MQTT topic name.

(This relates to #267 in that there we also need support for "property bags".)

calohmn:
This fixes #1336.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/tracing/TracingHelper.java
+++ b/core/src/main/java/org/eclipse/hono/tracing/TracingHelper.java
@@ -14,9 +14,12 @@
 package org.eclipse.hono.tracing;
 
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.function.BiConsumer;
+import java.util.function.Supplier;
 
 import org.apache.qpid.proton.message.Message;
 
@@ -27,6 +30,7 @@ import io.opentracing.Tracer;
 import io.opentracing.log.Fields;
 import io.opentracing.noop.NoopSpanContext;
 import io.opentracing.propagation.Format;
+import io.opentracing.propagation.TextMap;
 import io.opentracing.tag.BooleanTag;
 import io.opentracing.tag.IntTag;
 import io.opentracing.tag.StringTag;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/eclipse/hono/tracing/TracingHelper.java
```diff
@@ -308,6 +312,71 @@ public static SpanContext extractSpanContext(final Tracer tracer, final MultiMap
...
+     * This provides a generic way to serialize a span context in any kind of textual data.
+     * See {@link #extractSpanContext(Tracer, Supplier)} for the corresponding method to deserialize the
+     * context from that data.
+     *
+     * @param tracer The Tracer to use for injecting the context.
+     * @param spanContext The context to inject.
+     * @param keyValueConsumer The operation that will receive the key-value pairs representing the context.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     */
+    public static void injectSpanContext(final Tracer tracer, final SpanContext spanContext,
```
sophokles73: where is this used?

calohmn: Currently nowhere. But that method makes sense as the counterpart to the `extractSpanContext(tracer, supplier)` method and may be useful in external code using `TracingHelper`.

sophokles73: ok, just wanted to make sure I'm not missing something ;-)

sophokles73:
@calohmn do you want to document this new feature in the release notes?

calohmn:
@sophokles73 As noted in #1336, the use cases for this feature are somewhat special. Therefore I don't think we need to mention it in the release notes.

PR #1702, opened 2020-01-10
[#1701] Finish spans in management API event bus adapters

### LINKED ISSUE #1701: Trace spans not getting finished in management API
A lot of trace spans created in the management API classes are not getting finished.
This is in the `EventBus[*]ManagementAdapter` classes.

Regarding Hono 1.1:
These classes might become obsolete (see #1346), but for 1.1 they still might be kept, in which case the issue should still get fixed there.

calohmn:
This fixes #1701.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/management/credentials/EventBusCredentialsManagementAdapter.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/management/credentials/EventBusCredentialsManagementAdapter.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2019, 2020 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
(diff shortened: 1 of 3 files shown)


service-base/src/main/java/org/eclipse/hono/service/management/credentials/EventBusCredentialsManagementAdapter.java
```diff
@@ -94,31 +95,27 @@ protected String getEventBusAddress() {
...
-            final Promise<OperationResult<Void>> result = Promise.promise();
-
-            return secretsFuture.compose(secrets -> {
-                getService().set(tenantId, deviceId, resourceVersion, secrets, span, result);
-                        return result.future().map(res -> {
-                            return res.createResponse(request, id -> null).setDeviceId(deviceId);
-                        });
-                    }
-            );
-        } catch (final IllegalStateException e) {
```
calohmn: Note that this `IllegalStateException` (which should rather be an `IllegalArgumentException`, see added TODO) is now caught in `credentialsFromPayload()`.

Issue #1681, opened 2019-12-19
Upgrade to Opentracing 0.33


sophokles73:
There have been some minor API changes. Upgrading to Opentracing 0.33 is a pre-requisite for using a more recent version of the Jaeger Java client.

PR #1705, opened 2020-01-15
[#1681] Upgrade to Opentracing 0.33

sophokles73:
The TracingHandler classes have been copied to the service-base module
so that we can more easily adapt them to our needs. The upstream project
doesn't seem to be very active and it is just a few classes.

Also upgraded to Jaeger java-client 0.35.2 and Jaeger backend 1.16.

### DIFF
--- a/adapters/http-vertx/src/test/java/org/eclipse/hono/adapter/http/impl/VertxBasedHttpProtocolAdapterTest.java
+++ b/adapters/http-vertx/src/test/java/org/eclipse/hono/adapter/http/impl/VertxBasedHttpProtocolAdapterTest.java
@@ -69,7 +69,6 @@ import io.vertx.core.MultiMap;
 import io.vertx.core.Vertx;
 import io.vertx.core.http.HttpHeaders;
 import io.vertx.core.json.JsonObject;
-import io.vertx.core.logging.SLF4JLogDelegateFactory;
 import io.vertx.ext.auth.User;
 import io.vertx.ext.unit.TestContext;
 import io.vertx.ext.unit.junit.VertxUnitRunner;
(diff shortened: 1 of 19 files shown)


adapters/http-vertx/src/test/java/org/eclipse/hono/adapter/http/impl/VertxBasedHttpProtocolAdapterTest.java
```diff
@@ -129,6 +130,8 @@
     @SuppressWarnings("unchecked")
     @BeforeClass
     public static void deployAdapter(final TestContext ctx) {
+
+        System.setProperty("vertx.logger-delegate-factory-class-name", SLF4JLogDelegateFactory.class.getName());
```
ctron: If this is not required by the test, then I think it would make sense to put this into the maven build instead.

service-base/src/main/java/org/eclipse/hono/service/http/TracingHandler.java
```diff
@@ -0,0 +1,163 @@
...
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
+import io.vertx.ext.web.RoutingContext;
+
+/**
+ * Handler which creates tracing data for all server requests. It should be added to
+ * {@link io.vertx.ext.web.Route#handler(Handler)} and {@link io.vertx.ext.web.Route#failureHandler(Handler)} as the
+ * first in the chain.
+ * <p>
+ * This class has been copied from the <a href="https://github.com/opentracing-contrib/java-vertx-web">
```
ctron: I do think that we still need a CQ upfront when we take over code into our repository.

sophokles73: Yes, we already have a CQ for this: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16576

adapters/http-vertx/src/test/java/org/eclipse/hono/adapter/http/impl/VertxBasedHttpProtocolAdapterTest.java
sophokles73: Oh, that was a test only. I will remove that altogether ...

sophokles73:
closing in favor of #1707 in an attempt to fix the ECA check ...

PR #1706, opened 2020-01-15
Add Auto-Provisioning to Device Registry + Integration Tests (2nd try)

b-abel:
This replaces PR #1693 which constantly fails due to an erroneous ECA check. I hope that creating a new PR solves the problem.

This PR completes the auto-provisioning (#1296) feature. It

 * adds support to the AMQP adapter
 * adds an implementation of the feature to the example device registry
 * adds integration tests for this feature for the AMQP, MQTT and HTTP adapters.


b-abel:
ECA check still fails for an unknown reason, stating that the sign-off-footer would be missing, which is not the case. I tried a lot of different configurations, all with the same result. I'm afraid that the check is broken.

b-abel:
> ECA check still fails for an unknown reason, stating that the sign-off-footer would be missing, which is not the case. I tried a lot of different configurations, all with the same result. I'm afraid that the check is broken.

Changing the email to lowercase solved the problems with ECA.

PR #1707, opened 2020-01-16
[#1681] Upgrade Opentracing and Jaeger

sophokles73:
The TracingHandler classes have been copied to the service-base module
so that we can more easily adapt them to our needs. The upstream project
doesn't seem to be very active and it is just a few classes.

Also upgraded to Jaeger java-client 0.35.2 and Jaeger backend 1.16.


sophokles73:
@ctron would you mind taking another look? BTW we already have a [CQ](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=16576) for the component that the code has been copied from ...

sophokles73:

Issue #1710, opened 2020-01-17
Device registry should not handout hashed-passwords secrets


jbtrystram:
But instead only gives the metadata of the secret, like expiration dates etc and assign an ID to the secret.

sophokles73:
@jbtrystram @dejanb can this be closed?

dejanb:
Fixed with #1711

PR #1711, opened 2020-01-17
[#1710] stop registry to return hashed-passwords details

jbtrystram:
The device registry now only returns metadata and a secret ID for each secret.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
+++ b/core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
@@ -105,6 +105,10 @@ public final class RegistryManagementConstants extends RequestResponseApiConstan
      * The name of the field that contains the number of credentials contained in a message.
      */
     public static final String FIELD_CREDENTIALS_TOTAL           = "total";
+    /**
+     * The name of the field that contains the id of the entity (e.g. secret id).
+     */
+    public static final String FIELD_ID = "id";
 
     /* secrets fields */
     /**
--- a/service-base/src/main/java/org/eclipse/hono/service/management/credentials/CommonSecret.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/management/credentials/CommonSecret.java
@@ -31,6 +31,9 @@ public abstract class CommonSecret {
     @JsonProperty
     private Boolean enabled;
 
+    @JsonProperty
+    private String id;
+
     @JsonProperty(RegistryManagementConstants.FIELD_SECRETS_NOT_BEFORE)
     @HonoTimestamp
     private Instant notBefore;
@@ -52,6 +55,22 @@ public abstract class CommonSecret {
      */
     public CommonSecret setEnabled(final Boolean enabled) {
         this.enabled = enabled;
+        return this;
+    }
+
+    public String getId() {
+        return id;
+    }
+
+    /**
+     * Sets the ID of the secret. This id may be assigned by the device registry.
+     * The id must be unique within the credentials set containing it.
+     *
+     * @param id The string to set as the id.
+     * @return   a reference to this for fluent use.
+     */
+    public CommonSecret setId(final String id) {
+        this.id = id;
         return this;
     }
 
--- a/service-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordSecret.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordSecret.java
@@ -112,6 +112,9 @@ public class PasswordSecret extends CommonSecret {
     @Override
     public void checkValidity() {
         super.checkValidity();
+        if (containsOnlySecretId()) {
+            return;
+        }
         if (!Strings.isNullOrEmpty(passwordPlain)) {
             throw new IllegalStateException(String.format("'%s' must be empty", RegistryManagementConstants.FIELD_SECRETS_PWD_PLAIN));
         }
@@ -121,6 +124,18 @@ public class PasswordSecret extends CommonSecret {
         if (Strings.isNullOrEmpty(passwordHash)) {
             throw new IllegalStateException(String.format("'%s' must not be empty", RegistryManagementConstants.FIELD_SECRETS_PWD_HASH));
         }
+    }
+
+    /**
+     * Asserts if the secret do not contain secret details and only an id.
+     *
+     * @return        true if an ID is present and no new secret detail is set.
+     */
+    public boolean containsOnlySecretId() {
+        return (!Strings.isNullOrEmpty(getId())
+                && Strings.isNullOrEmpty(passwordPlain)
+                && Strings.isNullOrEmpty(hashFunction)
+                && Strings.isNullOrEmpty(passwordHash));
     }
 
     /**
--- a/services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
+++ b/services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
@@ -43,6 +43,7 @@ import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.CredentialsConstants;
 import org.eclipse.hono.util.CredentialsResult;
 import org.eclipse.hono.util.RegistryManagementConstants;
+import org.eclipse.hono.util.Strings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -479,16 +480,18 @@ public final class FileBasedCredentialsService extends AbstractVerticle
         final var newVersion = UUID.randomUUID().toString();
         setResourceVersion(tenantId, deviceId, newVersion);
 
-        // clean out all existing credentials for this device
-        try {
-            removeAllForDevice(tenantId, deviceId, span);
-        } catch (final ClientErrorException e) {
-            TracingHelper.logError(span, e);
-            return OperationResult.empty(e.getErrorCode());
-        }
 
         // authId->credentials[]
         final Map<String, JsonArray> credentialsForTenant = createOrGetCredentialsForTenant(tenantId);
+
+        if (!credentialsForTenant.isEmpty()) {
+            try {
+                verifyOverwriteEnabled(span);
+            } catch (ClientErrorException e) {
+                TracingHelper.logError(span, e);
+                return OperationResult.empty(e.getErrorCode());
+            }
+        }
 
         // now add the new ones
         for (final CommonCredential credential : credentials) {
@@ -504,10 +507,10 @@ public final class FileBasedCredentialsService extends AbstractVerticle
             final String authId = credential.getAuthId();
             final JsonObject credentialObject = JsonObject.mapFrom(credential);
             final String type = credentialObject.getString(CredentialsConstants.FIELD_TYPE);
-            final var json = createOrGetAuthIdCredentials(authId, credentialsForTenant);
+            final JsonArray json = createOrGetAuthIdCredentials(authId, credentialsForTenant);
 
             // find credentials - matching by type
-            var credentialsJson = json.stream()
+            JsonObject credentialsJson = json.stream()
                     .filter(JsonObject.class::isInstance).map(JsonObject.class::cast)
                     .filter(j -> type.equals(j.getString(CredentialsConstants.FIELD_TYPE)))
                     .findAny().orElse(null);
@@ -529,16 +532,68 @@ public final class FileBasedCredentialsService extends AbstractVerticle
                 return OperationResult.empty(HttpURLConnection.HTTP_CONFLICT);
             }
 
-            // start updating
-            dirty = true;
-
-            var secretsJson = credentialsJson.getJsonArray(CredentialsConstants.FIELD_SECRETS);
+            JsonArray secretsJson = credentialsJson.getJsonArray(CredentialsConstants.FIELD_SECRETS);
             if (secretsJson == null) {
                 // secrets field was missing, assign
                 secretsJson = new JsonArray();
-                credentialsJson.put(CredentialsConstants.FIELD_SECRETS, secretsJson);
             }
-            secretsJson.addAll(credentialObject.getJsonArray(CredentialsConstants.FIELD_SECRETS));
+
+            final JsonArray inputSecrets = credentialObject.getJsonArray(CredentialsConstants.FIELD_SECRETS);
+            final JsonArray definitiveInputSecret = new JsonArray();
+            for (Object inputSecret : inputSecrets) {
+                final JsonObject secret = (JsonObject) inputSecret;
+                final String secretId = secret.getString(RegistryManagementConstants.FIELD_ID);
+
+                // No secret ID specified : create a new secret
+                if (Strings.isNullOrEmpty(secretId)) {
+                    secret.put(RegistryManagementConstants.FIELD_ID, UUID.randomUUID().toString());
+                    definitiveInputSecret.add(secret);
+                // secret ID specified
+                } else {
+                    // Find the corresponding secret with the given ID.
+                    boolean found = false;
+                    for (Object st : secretsJson) {
+                        final JsonObject existingSecret = (JsonObject) st;
+                        if (secretId.equals(existingSecret.getString(RegistryManagementConstants.FIELD_ID))) {
+                            found = true;
+                            final JsonObject newSecret = new JsonObject();
+                            copySecretFields(secret, newSecret);
+
+                            // update the secret : remove the old values.
+                            if (secret.containsKey(CredentialsConstants.FIELD_SECRETS_PWD_PLAIN)
+                                    || secret.containsKey(CredentialsConstants.FIELD_SECRETS_KEY)
+                                    || secret.containsKey(CredentialsConstants.FIELD_SECRETS_PWD_HASH)
+                                    || secret.containsKey(CredentialsConstants.FIELD_SECRETS_SALT)) {
+
+                                removePasswordDetailsFromSecret(newSecret);
+                            }
+                            // then copy the new details.
+                            for (String field : secret.fieldNames()) {
+                                newSecret.put(field, secret.getValue(field));
+                            }
+                            definitiveInputSecret.add(newSecret);
+                        }
+                    }
+
+                    // check if the secretID given was found
+                    if (!found) {
+                        TracingHelper.logError(span, "secret ID given does not exist for this auth-id and type.");
+                        return OperationResult.empty(HttpURLConnection.HTTP_BAD_REQUEST);
+                    }
+                }
+            }
+
+            dirty = true;
+
+            // Now we can remove all the secrets
+            secretsJson.clear();
+
+            // Write the new secrets
+            secretsJson.addAll(definitiveInputSecret);
+
+            // Commit the update
+            credentialsJson.put(CredentialsConstants.FIELD_SECRETS, secretsJson);
+
             credentialsForTenant.put(authId, json);
         }
 
@@ -557,13 +612,16 @@ public final class FileBasedCredentialsService extends AbstractVerticle
             for (final PasswordSecret passwordSecret : ((PasswordCredential) credential).getSecrets()) {
                 passwordSecret.encode(passwordEncoder);
                 passwordSecret.checkValidity();
-                switch (passwordSecret.getHashFunction()) {
+                if (!passwordSecret.containsOnlySecretId()) {
+                    switch (passwordSecret.getHashFunction()) {
                     case RegistryManagementConstants.HASH_FUNCTION_BCRYPT:
                         final String pwdHash = passwordSecret.getPasswordHash();
                         verifyBcryptPasswordHash(pwdHash);
                         break;
                     default:
                         // pass
+                    }
+                    // pass
                 }
                 // pass
             }
@@ -589,13 +647,11 @@ public final class FileBasedCredentialsService extends AbstractVerticle
 
     /**
      * Remove all credentials that point to a device.
-     * 
+     *
      * @param tenantId The tenant to process.
      * @param deviceId The device id to look for.
      */
     private void removeAllForDevice(final String tenantId, final String deviceId, final Span span) {
-
-        final boolean canModify = getConfig().isModificationEnabled();
 
         final Map<String, JsonArray> credentialsForTenant = createOrGetCredentialsForTenant(tenantId);
 
@@ -615,12 +671,7 @@ public final class FileBasedCredentialsService extends AbstractVerticle
                     continue;
                 }
 
-                // check if we may overwrite
-
-                if (!canModify) {
-                    TracingHelper.logError(span, "Modification is disabled for the Credentials service.");
-                    throw new ClientErrorException(HttpURLConnection.HTTP_FORBIDDEN);
-                }
+                verifyOverwriteEnabled(span);
 
                 // remove device from credentials set
                 i.remove();
@@ -663,6 +714,7 @@ public final class FileBasedCredentialsService extends AbstractVerticle
         for (final Object credential : matchingCredentials) {
             final JsonObject credentialsObject = (JsonObject) credential;
             credentialsObject.remove(CredentialsConstants.FIELD_PAYLOAD_DEVICE_ID);
+            removePasswordDetailsFromCredential(credentialsObject);
             final CommonCredential cred = credentialsObject.mapTo(CommonCredential.class);
             credentials.add(cred);
         }
@@ -760,7 +812,7 @@ public final class FileBasedCredentialsService extends AbstractVerticle
 
     /**
      * Create or get credentials map for a single tenant.
-     * 
+     *
      * @param tenantId The tenant to get
      * @return The map, never returns {@code null}.
      */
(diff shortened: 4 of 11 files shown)


sophokles73:
@jbtrystram some integration tests seem to fail due to the changes you have made. Can you check?

jbtrystram:
@sophokles73 I am on this today :)

site/documentation/content/api/credentials/index.md
```diff
@@ -155,6 +155,7 @@ Below is an example for a payload containing [a hashed password]({{< relref "#ha
   "auth-id": "sensor1",
   "enabled": true,
   "secrets": [{
+    "id": "349556ea-4902-47c7-beb0-1009ab693fb4",
```
sophokles73: IMHO there is no need to include the id in the response payload. What should a protocol adapter do with it?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -1077,15 +1075,10 @@ components:
...
-                  "pwd-hash": "",
-                  "salt": "",
-                  "hash-function": ""
-               }]
+            auth-id: sensor1
+            type: hashed-password
+            secrets: [{
+               "id": "349556ea-4902-47c7-beb0-1009ab693fb4",
+               "not-after": "2027-12-24T19:00:00Z",
+               "pwd-plain": ""
```
sophokles73: what is this property good for in this context? Isn't this the response to a *get credentials* request?

site/documentation/content/user-guide/device-registry.md
```diff
@@ -299,9 +299,9 @@ HTTP/1.1 204 No Content
 ETag: becc93d7-ab0f-48ec-ad26-debdf339cbf4x
 ~~~
     
-This uses a convenient option which lets the Device Registry do the hashing of the password. The following command retrieves the credentials that are stored by the Device Registry as a result of the command above: 
-    
-~~~sh    
+This lets the Device Registry do the hashing of the password. The following command retrieves some metadata credentials that are stored by the Device Registry as a result of the command above:
```
sophokles73: I would rather keep the original wording and instead explain explicitly, why the response payload does not contain the password itself ...

site/homepage/content/release-notes.md
```diff
@@ -25,6 +25,11 @@ title = "Release Notes"
   mechanisms advertised to a client connecting to these components. This can be used to
   restrict the support to only one of the SASL PLAIN and EXTERNAL mechanisms instead of both. 
 
+### API Changes
+
+ * The device registry credentials endpoint will no longer handout details for hashed-password secrets.
```
sophokles73: please be more explicit about which information is no longer returned

site/documentation/content/api/management/device-registry-v1.yaml
jbtrystram: the ID property of the secret is provided here so the user can keep track of the different secrets without handing over the actual values. I also thought the `id` property could be provided in the payload to a `PUT` request on the management API if the user need to update this specific secret. (without creating another one, or submitting all of the other credentials, but that is not implemented yet) .

site/documentation/content/user-guide/device-registry.md
jbtrystram: I updated the wording. I am not really happy with it though, I am open to suggestions :)

site/documentation/content/api/credentials/index.md
jbtrystram: Indeed, the adapters don't have any use of this. I made a mistake by editing that file ! I reverted the changes

site/homepage/content/release-notes.md
jbtrystram: I updated. Let me know what you think.

site/documentation/content/user-guide/device-registry.md
```diff
@@ -319,6 +319,7 @@ ETag: becc93d7-ab0f-48ec-ad26-debdf339cbf4x
   ]
 }]
 ~~~
+The above result does not contains `pwd_hash` and `hash_function` of the secret, as these value are confidential.
```
sophokles73: ... not contain ... as these values ...

site/homepage/content/release-notes.md
```diff
@@ -27,8 +27,9 @@ title = "Release Notes"
 
 ### API Changes
 
- * The device registry credentials endpoint will no longer handout details for hashed-password secrets.
-   Each password is given an ID which is now returned by the device registry.
+ * The device registry credentials endpoint will no longer handout sensitive details for hashed-password secrets.
+   `pwd-hash`, `Salt` and `hash-function` are stored by the device registry but not returned to the user.
```
sophokles73: `salt`

jbtrystram: Updated, thanks for catching this !

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -107,6 +107,10 @@
     public static final String FIELD_CREDENTIALS_TOTAL           = "total";
 
     /* secrets fields */
+    /**
+     * The name of the field that contains the secret id.
+     */
+    public static final String FIELD_SECRETS_ID                 = "id";
```
sophokles73: How about renaming this to the more general `FIELD_ID` since it is not really secrets specific anyway and might be used for other entities as well. WDYT?

service-base/src/main/java/org/eclipse/hono/service/management/credentials/CommonSecret.java
```diff
@@ -55,6 +58,21 @@ public CommonSecret setEnabled(final Boolean enabled) {
         return this;
     }
 
+    public String getId() {
+        return id;
+    }
+
+    /**
+     * Sets the ID of the secret. This id may be assigned by the device registry.
```
sophokles73: what is the uniqueness scope of this ID? I assume it is just the containing Credentials instance, right? However, IMHO we should document this here.

site/documentation/content/api/management/device-registry-v1.yaml
sophokles73: I actually meant the empty *pwd-plain* property ...
So, if I get it right, then the *id* property currently has no purpose at all because it is not considered when adding/updating credentials, right?

site/documentation/content/user-guide/device-registry.md
```diff
@@ -314,23 +314,21 @@ ETag: becc93d7-ab0f-48ec-ad26-debdf339cbf4x
...
-      "hash-function": "bcrypt"
+      "id": "349556ea-4902-47c7-beb0-1009ab693fb4"
     }
   ]
 }]
 ~~~
-    
+The above result does not contain `pwd_hash` and `hash_function` of the secret, as these values are confidential.
+
 The following commands add some `hashed-password` credentials for device `4720` using authentication identifier `sensor20`:
```
sophokles73: does this actually *add* credentials or does it *replace* all existing credentials? IMHO it is the latter ...

site/documentation/content/user-guide/device-registry.md
```diff
@@ -403,7 +398,7 @@ ETag: d383ba4d-1853-4d03-b322-b7ff5af05ae2
 
 **Example**
 
-The following command retrieves credentials for device `4720`:
+The following command retrieves credentials metadata for device `4720`:
```
sophokles73: I guess the example payload should at least contain two secrets with e.g. adjoining validity periods so that you can see, what the *id* would be useful for (e.g. for changing one of the two passwords) ...

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsHttpIT.java
```diff
@@ -403,9 +403,11 @@ public void testGetAddedCredentials(final VertxTestContext context) {
                 HttpURLConnection.HTTP_NO_CONTENT)
                 .compose(ar -> registry.getCredentials(TENANT, deviceId))
                 .setHandler(context.succeeding(b -> {
-                    context.verify(() -> assertThat(b.toJsonArray())
-                            .isEqualTo(new JsonArray().add(JsonObject.mapFrom(hashedPasswordCredential))));
-                    context.completeNow();
+                    final PasswordCredential cred = b.toJsonArray().getJsonObject(0).mapTo(PasswordCredential.class);
+                    cred.getSecrets().forEach(secret -> {
+                        context.verify(() -> assertThat(secret.getId()).isNotNull());
+                        context.completeNow();;
```
sophokles73: you shouldn't complete the context for each secret ...

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsHttpIT.java
```diff
@@ -541,8 +543,33 @@ public void testGetAddedCredentialsButWithWrongAuthId(final VertxTestContext con
...
+                        ((JsonObject) secret).remove(RegistryManagementConstants.FIELD_SECRETS_ID);
+                    });
+        });
+
+        // The returned secrets won't contains the hashed password details fields, strip them from the expected values.
         final JsonArray expectedArray = new JsonArray();
-        expected.stream().forEach(credential -> expectedArray.add(JsonObject.mapFrom(credential)));
+        expected.stream().forEach(credential -> {
+            final JsonObject jsonCredential = JsonObject.mapFrom(credential);
+            expectedArray.add(stripHashAndSaltFromPasswordSecret(jsonCredential));
```
sophokles73: there is not much left to compare, is there?

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
jbtrystram: Why not. I updated the description so it doesn't lower the code readability.

service-base/src/main/java/org/eclipse/hono/service/management/credentials/CommonSecret.java
jbtrystram: I updated the description.

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsHttpIT.java
jbtrystram: the credential type, auth-id and all the credential-level fields are still there

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsHttpIT.java
jbtrystram: Good catch, thanks

site/documentation/content/user-guide/device-registry.md
jbtrystram: At this point of the guide there are no credentials registered for device `4720`, so it actually `add` some. I changed the wording to be more clear though.

jbtrystram:
@sophokles73 for some reason I can't reply to one of your comments in the review.
> I actually meant the empty pwd-plain property ...
> So, if I get it right, then the id property currently has no purpose at all because it is not considered when adding/updating credentials, right?

the `id` field still have a use IMHO. When the user set some credentials, he can "verify" they are in the registry with receiving an `id` for each password. I think if I don't know hono very well I would be confused if I received an empty secrets array after registering a hashed password. With the `id` field I know the password is there.

site/documentation/content/user-guide/device-registry.md
jbtrystram: Good idea ! I added another secrets with `not-after` and `not-before` overlapping with only one day.

site/documentation/content/api/management/device-registry-v1.yaml
jbtrystram: Indeed, the `pwd-plain`  does not have any sense here. removed.
the id field still have a use IMHO. When the user set some credentials, he can "verify" they are in the registry with receiving an id for each password. I think if I don't know hono very well I would be confused if I received an empty secrets array after registering a hashed password. With the id field I know the password is there.

service-base/src/main/java/org/eclipse/hono/service/management/credentials/CommonSecret.java
sophokles73: Can these IDs be assigned by a client, i.e. when registering the credentials? If so, do we check that the IDs are unique within the credentials?

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
```diff
@@ -539,6 +539,10 @@ public void set(final String tenantId, final String deviceId, final Optional<Str
                 credentialsJson.put(CredentialsConstants.FIELD_SECRETS, secretsJson);
             }
             secretsJson.addAll(credentialObject.getJsonArray(CredentialsConstants.FIELD_SECRETS));
+            // add an id for each secret.
+            secretsJson.forEach(secret -> {
```
sophokles73: what if they already have an ID, e.g. provided by the client?

jbtrystram:
@sophokles73 @ctron  I implemented the "id" feature for secrets in the file based registry.

It should works as follow : 

- When a user create a secret, the secret is assigned an ID that is returned on `GET` calls.
- A secret can be updated if the corresponding ID is put in the payload on a `PUT` call. A new ID will be assigned.
- If the user provide an Id that doesn't exist it will be ignored.
- On a `PUT` call, all the already existing secrets for which the IDs are not specified in the payload will be deleted.
- - On a `PUT` call, if only the ID is specified in the secret payload, the secret will be kept as-is.

I am going to write a few tests to make sure all works as expected.

ctron:
This looks good to me in general. My only comment would be, that I think we should have integration tests for use cases we had in mind. Like 1) Create a new credential set 2) read it back 3) change a single secret 4) verify the update by authenticating a device

Aside from that, but that is not something this PR has touched: I think we should remove the "deprecation" note on the hashed passwords when setting secrets.

jbtrystram:
I found a couple of corner cases so I had to rework this . Here are the features now : 

- When a user create a secret, the secret is assigned an ID that is returned on GET calls. A secret can be updated if the corresponding ID is put in the payload on a PUT call. A new ID will be assigned. All the fields that existed before for this ID will be carried over.
- If the user provide an ID that doesn't exist, the secret will be ignored.
- The `PUT` call can also be used to update/add meta-data fields to an existing secret without changing the actual secret values. (e.g. the password). The ID won't be changed as long as the one of the following fields is not changed : `pwd-plain`, `key`, `pwd-hash`, `salt`.
- On a PUT call, if only the ID is specified in the secret payload, the secret will be kept as-is.

It is not possible to delete only one field. The whole secret must be deleted. This can be done with a `PUT` call not containing the ID you want to delete.

jbtrystram:
@ctron @sophokles73 I added unit tests to validate the following scenarios : 
- Verify that created secrets contains an ID.
- Verify secrets details are not returned by registry
- Verify one secret can be deleted and the other stays there
- verify that an new ID is assigned when a  secret is changed
- verify the ID doesn't change when metadata only is updated + the other metada that already existed are still there

I think this is ready to be reviewed :) 
I'll squash the commits as soon as it's ready to merge.

sophokles73:
@jbtrystram 

> When a user create a secret, the secret is assigned an ID that is returned on GET calls. A secret can be updated if the corresponding ID is put in the payload on a PUT call. A new ID will be assigned. All the fields that existed before for this ID will be carried over.

I do not quite understand why we would want to assign a new ID with every update. Is there a particular reason for that? The term identifier to suggests that I will be able to refer to the identified object by means of the id for the whole lifespan of the object. So, FMPOV if I change an existing  password, then I would expect the ID to remain the same.

> If the user provide an ID that doesn't exist, the secret will be ignored.

I assume we are still talking about a PUT request. FMPOV it would be more transparent to return a 400 BAD REQUEST in this case in order to not let the client believe that he has actually updated the credentials when in fact we have silently ignored the request ...

jbtrystram:
@sophokles73 
 
> I do not quite understand why we would want to assign a new ID with every update. Is there a particular reason for that? The term identifier to suggests that I will be able to refer to the identified object by means of the id for the whole lifespan of the object. So, FMPOV if I change an existing password, then I would expect the ID to remain the same.

My idea was that a new ID would reflect that the change has actually been processed. But I guess that the return code does the same job. I removed this, the ID now stays the same.

> > If the user provide an ID that doesn't exist, the secret will be ignored.
> 
> I assume we are still talking about a PUT request. FMPOV it would be more transparent to return a 400 BAD REQUEST in this case in order to not let the client believe that he has actually updated the credentials when in fact we have silently ignored the request ...

Good point. I updated to return `400 Bad request` and to log the error in that case.

kaniyan:
@jbtrystram IMHO also the example credentials (_example-credentials.json_) in the file based device-registry should also be updated in accordance with the changes in this PR. When get credentials operation is invoked for example for device id _4711_ or _gw-1_, it still hands out values such as _pwd-hash_, _hash-function_ etc.

For me it seems that the existing secrets without ids should be migrated by the users so that the credentials endpoint no longer hands out sensitive details. FMPOV it is better to add this info in the release notes too. WDYT?

jbtrystram:
> @jbtrystram IMHO also the example credentials (_example-credentials.json_) in the file based device-registry should also be updated in accordance with the changes in this PR. When get credentials operation is invoked for example for device id _4711_ or _gw-1_, it still hands out values such as _pwd-hash_, _hash-function_ etc.

Good catch! That happened only one loading credentials from a file, which i did not tested. Thanks, I fixed it.

> For me it seems that the existing secrets without ids should be migrated by the users so that the credentials endpoint no longer hands out sensitive details. FMPOV it is better to add this info in the release notes too. WDYT?

The endpoint will no longer hand out sensitive details even if the secrets don't contain IDs. That said it would indeed be good to advise users to delete and recreate the credentials entries so the secrets are assigned ids

jbtrystram:
@ctron @sophokles73 

Here is what the feature set looks like now after your feedback :
- When a user create a secret, the secret is assigned an ID that is returned on GET calls. A secret can be updated if the corresponding ID is put in the payload on a PUT call. All existing metadata fields will be deleted if not re-supplied in the payload (keeping the get-> put flow).
- If the user provide an ID that doesn't exist, the registry will respond with `400 Bad Request`
- The PUT call can also be used to update/add meta-data fields to an existing secret without changing the actual secret values. (e.g. the password). The following values will be caried over from the existing secret : pwd-plain, key, pwd-hash, salt.

ctron:
Sounds right to me. With the exception that the `pwd-plain` should **never** be stored.

jbtrystram:
> Sounds right to me. With the exception that the `pwd-plain` should **never** be stored.

It is not ! Sorry my statement was confusing, It's just that this is the values the code check for, as a double check :)

sophokles73:
@jbtrystram  Just for my understanding: I assume that I can add credentials to a set of existing credentials by means of PUTting the payload I got back using  a GET and adding new credentials without an ID to the credentials array, right?

jbtrystram:
> @jbtrystram Just for my understanding: I assume that I can add credentials to a set of existing credentials by means of PUTting the payload I got back using a GET and adding new credentials without an ID to the credentials array, right?

of course

service-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordSecret.java
```diff
@@ -112,6 +112,9 @@ protected ToStringHelper toStringHelper() {
     @Override
     public void checkValidity() {
         super.checkValidity();
+        if (containsOnlySecretId()){
```
dejanb: Can we keep formatting consistent. There are a couple of places where there's no space after bracket and at some places there's a space in negation statement. We can also consider adding it to the check style.

service-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordSecret.java
```diff
@@ -123,6 +126,18 @@ public void checkValidity() {
         }
     }
 
+    /**
+     * Asserts if the secret do not contain secret details and only an id.
+     *
+     * @return        true if an ID is present and no new secret detail is set.
+     */
+    public boolean containsOnlySecretId(){
```
dejanb: Inconsistent formatting.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
```diff
@@ -479,17 +480,19 @@ public void set(final String tenantId, final String deviceId, final Optional<Str
...
-            removeAllForDevice(tenantId, deviceId, span);
-        } catch (final ClientErrorException e) {
-            TracingHelper.logError(span, e);
-            return OperationResult.empty(e.getErrorCode());
-        }
 
         // authId->credentials[]
         final Map<String, JsonArray> credentialsForTenant = createOrGetCredentialsForTenant(tenantId);
 
+        if (! credentialsForTenant.isEmpty()){
```
dejanb: Inconsistent formatting.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
```diff
@@ -479,17 +480,19 @@ public void set(final String tenantId, final String deviceId, final Optional<Str
...
+            try {
+                verifyOverwriteEnabled(span);
+            } catch (ClientErrorException e){
```
dejanb: Inconsistent formatting.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
```diff
@@ -804,4 +856,41 @@ public String toString() {
     protected int getMaxBcryptIterations() {
         return getConfig().getMaxBcryptIterations();
     }
+
+    /**
+     * Strips the hashed-password details from the jsonObject if needed.
+     */
+    private static void removePasswordDetailsFromCredential(final JsonObject credential){
```
dejanb: Inconsistent formatting.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
```diff
@@ -804,4 +856,41 @@ public String toString() {
...
+        if (credential.getString(CredentialsConstants.FIELD_TYPE).equals(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD)){
```
dejanb: Inconsistent formatting.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
```diff
@@ -804,4 +856,41 @@ public String toString() {
...
+
+            credential.getJsonArray(CredentialsConstants.FIELD_SECRETS).forEach(secret ->
+                    removePasswordDetailsFromSecret((JsonObject) secret));
+        }
+    }
+
+    private static void removePasswordDetailsFromSecret(final JsonObject secret){
```
dejanb: Inconsistent formatting.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
```diff
@@ -804,4 +856,41 @@ public String toString() {
...
+
+        secret.remove(CredentialsConstants.FIELD_SECRETS_HASH_FUNCTION);
+        secret.remove(CredentialsConstants.FIELD_SECRETS_PWD_HASH);
+        secret.remove(CredentialsConstants.FIELD_SECRETS_SALT);
+        secret.remove(CredentialsConstants.FIELD_SECRETS_PWD_PLAIN);
+        secret.remove(CredentialsConstants.FIELD_SECRETS_KEY);
+    }
+
+    private static void copySecretFields(final JsonObject in, final JsonObject out){
```
dejanb: Inconsistent formatting.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
```diff
@@ -804,4 +856,41 @@ public String toString() {
...
+        out.put(CredentialsConstants.FIELD_SECRETS_HASH_FUNCTION, in.getString(CredentialsConstants.FIELD_SECRETS_HASH_FUNCTION));
+        out.put(CredentialsConstants.FIELD_SECRETS_PWD_HASH, in.getString(CredentialsConstants.FIELD_SECRETS_PWD_HASH));
+        out.put(CredentialsConstants.FIELD_SECRETS_SALT, in.getString(CredentialsConstants.FIELD_SECRETS_SALT));
+        out.put(CredentialsConstants.FIELD_SECRETS_PWD_PLAIN, in.getString(CredentialsConstants.FIELD_SECRETS_PWD_PLAIN));
+    }
+
+    private void verifyOverwriteEnabled(final Span span) {
+
+        // check if we may overwrite
+        if (! config.isModificationEnabled()) {
```
dejanb: Inconsistent formatting.

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsHttpIT.java
```diff
@@ -566,4 +593,20 @@ private static PskCredential newPskCredentials(final String authId) {
     private JsonObject extractFirstCredential(final JsonObject json) {
         return json.getJsonArray(CredentialsConstants.CREDENTIALS_ENDPOINT).getJsonObject(0);
     }
+
+    private static JsonObject stripHashAndSaltFromPasswordSecret(final JsonObject credential){
```
dejanb: Inconsistent formatting.

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsHttpIT.java
```diff
@@ -541,8 +543,33 @@ public void testGetAddedCredentialsButWithWrongAuthId(final VertxTestContext con
...
+            JsonObject.mapFrom(credential).getJsonArray(CredentialsConstants.FIELD_SECRETS)
+                    .forEach(secret -> {
+                        // each secret should contain an ID.
+                        assertThat(JsonObject.mapFrom(secret)
+                                .getString(RegistryManagementConstants.FIELD_ID)).isNotNull();
+                    });
+        });
+
+        // secrets id were added by registry, strip it so we can compare other fields.
+        responseBody.forEach(credential ->{
```
dejanb: Inconsistent formatting.

services/device-registry/src/main/java/org/eclipse/hono/deviceregistry/FileBasedCredentialsService.java
```diff
@@ -529,16 +532,68 @@ public void set(final String tenantId, final String deviceId, final Optional<Str
...
+                            // then copy the new details.
+                            for (String field : secret.fieldNames()) {
+                                newSecret.put(field, secret.getValue(field));
+                            }
+                            definitiveInputSecret.add(newSecret);
+                        }
+                    }
+
+                    // check if the secretID given was found
+                    if (! found) {
```
dejanb: Inconsistent formatting.

site/homepage/content/release-notes.md
```diff
@@ -58,6 +58,12 @@ title = "Release Notes"
 
 ## 1.0.3
 
+### API Changes
```
dejanb: This should probably go as 1.2.0 release API changes?

jbtrystram:
Thanks for your remarks @dejanb . I created a new section for 1.2.0 in the release notes and moved the changes.

PR #1713, opened 2020-01-17
Use Always image pull policy for JRE base image

sophokles73:
Hono's container images use a base image with a jre-11 tag only, i.e.
there is no specific version being set. In order to always build using
the most recent Java 11 JRE, the image pull policy has been changed to
Always for the service components.

For running the integration tests, the image pull policy is still
IfNotPresent because the snapshot images being tested usually won't be
available from the Docker registry.


sophokles73:
@ctron would you mind taking a look?

Issue #1716, opened 2020-01-22
Support limiting the connections based on the device connection time


kaniyan:
Protocol adapters should support limiting the number of connections based on the total connection time of devices per tenant over a defined time frame. The limit must be configurable at the tenant level. The connection time tracked by the metrics defined in #1287 could be used to implement this.

sophokles73:
@kaniyan looks like we can close this one now, right?

kaniyan:
yes. This has been closed via PRs #1722, #1731 and #1734.

PR #1722, opened 2020-01-24
[#1716] Add new resource-limits based on the device connection duration

kaniyan:
This PR is for the issue #1716. This PR extends the Tenant API to support a new additional _resource-limit_ configuration for _device connection duration_. Also the `ResourceLimitChecks` API has been extended to validate if the max connection duration of a tenant is reached or not.

NB: This PR has been raised even though the implementation is not complete, so that the concept can be reviewed at an early stage.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/ConnectionDuration.java
+++ b/core/src/main/java/org/eclipse/hono/util/ConnectionDuration.java
@@ -1,0 +1,113 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.hono.util;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.eclipse.hono.annotation.HonoTimestamp;
+
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
+import java.util.Objects;
+
+/**
+ * The resource limits definition corresponding to the connection duration.
+ */
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+public class ConnectionDuration {
+
+    @JsonProperty(value = TenantConstants.FIELD_EFFECTIVE_SINCE, required = true)
+    @HonoTimestamp
+    private Instant effectiveSince;
+
+    @JsonProperty(TenantConstants.FIELD_MAX_MINUTES)
+    private long maxMinutes = TenantConstants.UNLIMITED_MINUTES;
+
+    @JsonProperty(value = TenantConstants.FIELD_PERIOD, required = true)
+    private ResourceLimitsPeriod period;
+
+    /**
+     * Gets the point in time on which the connection duration limit came into effect.
+     *
+     * @return The instant on which the connection duration limit came into effective or 
+     *         {@code null} if not set.
+     */
+    public final Instant getEffectiveSince() {
+        return effectiveSince;
+    }
+
+    /**
+     * Sets the point in time on which the connection duration limit came into effect.
+     *
+     * @param effectiveSince the point in time on which the connection duration limit came into effect
+     *                       and it comply to the {@link DateTimeFormatter#ISO_OFFSET_DATE_TIME}.
+     * @return  a reference to this for fluent use.
+     * @throws NullPointerException if effectiveSince is {@code null}.
+     */
+    public final ConnectionDuration setEffectiveSince(final Instant effectiveSince) {
+        this.effectiveSince = Objects.requireNonNull(effectiveSince);
+        return this;
+    }
+
+    /**
+     * Gets the maximum device connection duration in minutes to be allowed for the time period 
+     * defined by the {@link TenantConstants#FIELD_PERIOD_MODE} and 
+     * {@link TenantConstants#FIELD_PERIOD_NO_OF_DAYS}.
+     *
+     * @return The maximum connection duration in minutes or {@link TenantConstants#UNLIMITED_MINUTES}
+     *         if not set.
+     */
+    public final long getMaxMinutes(){
+        return maxMinutes;
+    }
+
+    /**
+     * Sets the maximum device connection duration in minutes to be allowed for the time period
+     * defined by the {@link TenantConstants#FIELD_PERIOD_MODE} and 
+     * {@link TenantConstants#FIELD_PERIOD_NO_OF_DAYS}.
+     *
+     * @param maxMinutes The maximum connection duration in minutes to be allowed.
+     * @return  a reference to this for fluent use.
+     * @throws IllegalArgumentException if the maximum number of minutes is set to less than 
+     *                                  {@link TenantConstants#UNLIMITED_MINUTES}.
+     */
+    public final ConnectionDuration setMaxDuration(final long maxMinutes) {
+        if (maxMinutes < TenantConstants.UNLIMITED_MINUTES) {
+            throw new IllegalArgumentException(
+                    String.format("Maximum minutes must be set to value >= %s", TenantConstants.UNLIMITED_MINUTES));
+        }
+        this.maxMinutes = maxMinutes;
+        return this;
+    }
+
+    /**
+     * Gets the period for the connection duration calculation.
+     *
+     * @return The period for the connection duration calculation.
+     */
+    public final ResourceLimitsPeriod getPeriod() {
+        return period;
+    }
+
+    /**
+     * Sets the period for the connection duration calculation.
+     *
+     * @param period The period for the connection duration calculation.
+     * @return a reference to this for fluent use.
+     * @throws NullPointerException if period is {@code null}.
+     */
+    public final ConnectionDuration setPeriod(final ResourceLimitsPeriod period) {
+        this.period = Objects.requireNonNull(period);
+        return this;
+    }
+}
--- a/core/src/main/java/org/eclipse/hono/util/ResourceLimitsPeriod.java
+++ b/core/src/main/java/org/eclipse/hono/util/ResourceLimitsPeriod.java
@@ -1,0 +1,76 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.hono.util;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+/**
+ * The period definition corresponding to a resource limit for a tenant.
+ */
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+public class ResourceLimitsPeriod {
+
+    @JsonProperty(value = TenantConstants.FIELD_PERIOD_MODE, required = true)
+    private String mode;
+
+    @JsonProperty(value = TenantConstants.FIELD_PERIOD_NO_OF_DAYS)
+    private int noOfDays;
+
+    /**
+     * Gets the mode of period for resource limit calculation.
+     *
+     * @return The mode of period for resource limit calculation.
+     */
+    public final String getMode() {
+        return mode;
+    }
+
+    /**
+     * Sets the mode of period for resource limit calculation.
+     *
+     * @param mode The mode of period for resource limit calculation.
+     * @return  a reference to this for fluent use.
+     * @throws NullPointerException if mode is {@code null}.
+     */
+    public final ResourceLimitsPeriod setMode(final String mode) {
+        this.mode = Objects.requireNonNull(mode);
+        return this;
+    }
+
+    /**
+     * Gets the number of days for which resource usage is calculated.
+     *
+     * @return The number of days for a resource limit calculation.
+     */
+    public final int getNoOfDays() {
+        return noOfDays;
+    }
+
+    /**
+     * Sets the number of days for which resource usage is calculated.
+     *
+     * @param noOfDays The number of days for which resource usage is calculated.
+     * @return  a reference to this for fluent use.
+     * @throws IllegalArgumentException if the number of days is negative.
+     */
+    public final ResourceLimitsPeriod setNoOfDays(final int noOfDays) {
+        if (noOfDays < 0) {
+            throw new IllegalArgumentException("Number of days property must be  set to value >= 0");
+        }
+        this.noOfDays = noOfDays;
+        return this;
+    }
+}
--- a/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
@@ -32,9 +32,11 @@ import org.eclipse.hono.client.ServiceInvocationException;
 import org.eclipse.hono.service.metric.MetricsTags;
 import org.eclipse.hono.service.metric.MicrometerBasedMetrics;
 import org.eclipse.hono.tracing.TracingHelper;
+import org.eclipse.hono.util.ConnectionDuration;
 import org.eclipse.hono.util.DataVolume;
 import org.eclipse.hono.util.DataVolumePeriod;
 import org.eclipse.hono.util.MessageHelper;
+import org.eclipse.hono.util.ResourceLimitsPeriod;
 import org.eclipse.hono.util.TenantConstants;
 import org.eclipse.hono.util.TenantObject;
 import org.slf4j.Logger;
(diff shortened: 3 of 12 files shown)


service-base/src/main/java/org/eclipse/hono/service/resourcelimits/ResourceLimitChecks.java
```diff
@@ -86,4 +86,20 @@
...
+     * @param spanContext The currently active OpenTracing span context that is used to
+     *                    trace the limits verification or {@code null}
+     *                    if no span is currently active.
+     * @return A future indicating the outcome of the check.
+     *         <p>
+     *         The future will be failed with a {@link ServiceInvocationException}
+     *         if the check could not be performed.
+     * @throws NullPointerException if the tenant object is null.
+     */
+    Future<Boolean> isConnectionDurationLimitReached(TenantObject tenantObject, SpanContext spanContext);
```
sophokles73: this should also have a default implementation in order to not break backward compatibility ...

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -715,6 +717,25 @@ components:
...
+            - effective-since
+         properties:
+            "effective-since":
+               type: string
+               format: date-time
+               description: The date-time on which the connection duration limit came into effect.
+            "max-seconds":
+               type: integer
+               default: -1
+               description: The maximum number of seconds to be allowed for a tenant for the
```
sophokles73: I wonder if *seconds* is an appropriate unit here. What will be typical values?

core/src/main/java/org/eclipse/hono/util/ResourceLimitsPeriod.java
```diff
@@ -0,0 +1,73 @@
...
+package org.eclipse.hono.util;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * The period definition corresponding to a resource limit for a tenant.
+ */
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+public class ResourceLimitsPeriod {
```
sophokles73: this looks very similar to `DataVolumePeriod`, doesn't it?

kaniyan: Yes. My idea is to deprecate `DataVolumePeriod` and use this in future for the `DataVolume`.

sophokles73: sounds reasonable, the deprecation should then also be part of this PR, I guess

kaniyan: yes, I will add it.

site/documentation/content/api/management/device-registry-v1.yaml
kaniyan: Instead of _seconds_, I think _minutes_ might be more appropriate unit here. For example AWS  describes connection time in terms of _minutes_ (2,250,000 minutes of connection for Free tier) & also in Hub service plans it is also defined in _minutes_.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/ResourceLimitChecks.java
kaniyan: I will add it.

site/documentation/content/api/management/device-registry-v1.yaml
sophokles73: So, can we think of a reasonable use case where you would actually want to limit the overall connection time of all devices of a Tenant to something less than a minute? In the current Prometheus based implementation of the limits, we actually cache the data for more than a minute, don't we? So we usually wouldn't even be able to enforce such a limit properly. Personally, I therefore also believe that minutes should be more appropriate here ...

core/src/main/java/org/eclipse/hono/util/ConnectionDuration.java
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * Sets the maximum device connection duration in minutes to be allowed for the time period
+     * defined by the {@link TenantConstants#FIELD_PERIOD_MODE} and 
+     * {@link TenantConstants#FIELD_PERIOD_NO_OF_DAYS}.
+     *
+     * @param maxMinutes The maximum connection duration in minutes to be allowed.
+     * @return  a reference to this for fluent use.
+     * @throws IllegalArgumentException if the maximum number of minutes is set to less than -1.
+     */
+    public final ConnectionDuration setMaxMinutes(final long maxMinutes) {
```
sophokles73: I would rather have the name indicate the purpose than the unit of the argument, e.g. *setMaxDuration*. The parameter name and JavaDoc already indicate the unit, don't they?

kaniyan: I set it to _setMaxMinutes_ to keep in alignment with the existing naming convention in _resource-limits_ namely`DataVolume.setMaxBytes(...)`. _setMaxDuration_ makes more sense and we could also rename later `DataVolume.setMaxBytes(...)` to `DataVolume.setMaxVolume(...)`.

core/src/main/java/org/eclipse/hono/util/ConnectionDuration.java
```diff
@@ -0,0 +1,88 @@
...
+        if (maxMinutes < -1) {
```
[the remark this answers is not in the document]
kaniyan: IMHO 0 is a reasonable value. It can be set to 0 to ensure that no more connections are allowed. Also we allow data volume to set to 0.

core/src/main/java/org/eclipse/hono/util/ConnectionDuration.java
sophokles73: Hm, now that you mention the `DataVolume` spec ... The whole class is already named `ConnectionDuration` so it probably is obvious that we are setting the max duration. Let's stick with *maxMinutes* and *maxBytes*. At least, this way it is consistent and we do not need to rename existing properties.

core/src/main/java/org/eclipse/hono/util/ConnectionDuration.java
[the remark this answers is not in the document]
sophokles73: Ok, then we should allow it here as well.

core/src/main/java/org/eclipse/hono/util/ConnectionDuration.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @JsonProperty(value = TenantConstants.FIELD_EFFECTIVE_SINCE, required = true)
+    @HonoTimestamp
+    private Instant effectiveSince;
+
+    @JsonProperty(TenantConstants.FIELD_MAX_MINUTES)
+    private long maxMinutes = TenantConstants.UNLIMITED_MINUTES;
+
+    @JsonProperty(TenantConstants.FIELD_PERIOD)
+    private ResourceLimitsPeriod period;
```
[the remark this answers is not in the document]
kaniyan: My bad, I have added it now.

kaniyan:
@sophokles73 I have also added the reference implementation for _isConnectionDurationLimitReached_ and tests. Would you mind taking a look?

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -470,69 +576,61 @@ private Long extractLongValue(final JsonObject response, final Span span) {
...
-     *                 {@link TenantConstants#FIELD_MAX_BYTES}. 
-     * @return The allowed data-volume in bytes.
+     * @param configuredLimit The configured limit. 
+     * @return The effective resource limit that has been calculated.
      */
-    long calculateDataVolume(
+    long calculateEffectiveLimit(
             final OffsetDateTime effectiveSince,
-            final OffsetDateTime targetDateTime,
+            final OffsetDateTime currentDateTime,
```
sophokles73: if this is *always* the current instant then I do not think that you need to pass it in as a parameter, do you?

kaniyan: It need not be passed. I will update it.

kaniyan: As I checked, there is an existing test namely _verifyEffectiveResourceLimitCalculation_ uses this method. It verifies the calculated limit values under various scenarios. If the current date is always the actual date time then the expected result also keeps changing and unable to verify. It is better if we have this parameter, so that a fixed date time can be passed and the calculated value can be verified.  WDYT?

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -292,6 +296,108 @@ private Span createSpan(final String name, final SpanContext parent, final Tenan
...
+                    return b;
+                });
+    }
+
+    private void checkConnectionDurationLimit(final TenantObject tenant, final Map<String, Object> items,
+            final Span span, final Promise<Boolean> result) {
+        final ConnectionDuration connectionDurationConfig = tenant.getResourceLimits().getConnectionDuration();
+        final long maxConnectionDurationInMinutes = connectionDurationConfig.getMaxMinutes();
+        final Instant effectiveSince = connectionDurationConfig.getEffectiveSince();
+        final PeriodMode periodMode = PeriodMode.from(connectionDurationConfig.getPeriod().getMode());
```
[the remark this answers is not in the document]
kaniyan: IMHO the property _period_ should not be optional. The property _connection-duration_, doesn't make sense without _period_. I would update it.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
sophokles73: That is also ok and indeed helpful for testing. However, the parameter then should not be named *currentDateTime* but keep its original name FMPOV

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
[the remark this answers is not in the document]
sophokles73: sounds reasonable

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
kaniyan: yes, I will keep the original name.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -470,69 +576,61 @@ private Long extractLongValue(final JsonObject response, final Span span) {
...
-     *                     {@link TenantConstants#FIELD_PERIOD_NO_OF_DAYS}. 
-     * @return The period in days for which the data usage is to be calculated.
+     * @param effectiveSince The point of time on which the resource limit came into effect.
+     * @param currentDateTime The current date and time used for the resource usage period calculation.
+     * @param mode The mode of the period defined by {@link TenantConstants#FIELD_PERIOD_MODE}.
+     * @param periodInDays The number of days defined by {@link TenantConstants#FIELD_PERIOD_NO_OF_DAYS}. 
+     * @return The period in days for which the resource usage is to be calculated.
      */
-    long calculateDataUsagePeriod(
+    long calculateResourceUsagePeriod(
```
sophokles73: could this method be moved to `ResourceLimitPeriod`? It doesn't seem to be tied to a Prometheus specific implementation of checks but provides a rather generic helper, doesn't it?

sophokles73:
@kaniyan it seems like you have changed the email address of your Eclipse Foundation account. That's why the ECA check fails. Please re-create the commits using exactly the same email address that you used at Eclipse for the author, committer and signed-off-by footer.

kaniyan:
> @kaniyan it seems like you have changed the email address of your Eclipse Foundation account. That's why the ECA check fails. Please re-create the commits using exactly the same email address that you used at Eclipse for the author, committer and signed-off-by footer.

@sophokles73 Thank you for the info. I already updated it. Currently I am checking on moving _calculateResourceUsagePeriod_ method to _ResourceLimitPeriod_ class. After that I will push the commits signed with new gpg key.

sophokles73:
> After that I will push the commits signed with new gpg key.

There is no need to sign your commits with a gpg key. You only need to make sure that you use the same email address in the commits as you registered with the Eclipse Foundation.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
kaniyan: I have moved the `calculateResourceUsagePeriod` to `ResourceLimitPeriod`. I have pushed the changes. Would you mind taking a look?

core/src/main/java/org/eclipse/hono/util/ConnectionDuration.java
```diff
@@ -0,0 +1,110 @@
...
+        return period;
+    }
+
+    /**
+     * Sets the period for the connection duration calculation.
+     *
+     * @param period The period for the connection duration calculation.
+     * @return a reference to this for fluent use.
+     */
+    public final ConnectionDuration setPeriod(final ResourceLimitsPeriod period) {
```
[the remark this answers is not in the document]
kaniyan: Done.

core/src/main/java/org/eclipse/hono/util/ResourceLimitsPeriod.java
```diff
@@ -0,0 +1,73 @@
...
+        return mode;
+    }
+
+    /**
+     * Sets the mode of period for resource limit calculation.
+     *
+     * @param mode The mode of period for resource limit calculation.
+     * @return  a reference to this for fluent use.
+     */
+    public final ResourceLimitsPeriod setMode(final String mode) {
```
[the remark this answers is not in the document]
kaniyan: Done.

kaniyan:
> > After that I will push the commits signed with new gpg key.
> 
> There is no need to sign your commits with a gpg key. You only need to make sure that you use the same email address in the commits as you registered with the Eclipse Foundation.

I have changed the email id in commit, but still the eca validation fails. I will take a look at it.

sophokles73:
the author in your commits still contains the old email address. You need to create new commits in order to change this, you cannot change the author of an existing commit.

core/src/main/java/org/eclipse/hono/util/ResourceLimitsPeriod.java
```diff
@@ -0,0 +1,172 @@
...
+     * {@link TenantConstants#FIELD_PERIOD_MODE}.
+     *
+     * @param effectiveSince The point of time on which the resource limit came into effect.
+     * @param targetDateTime The target date and time used for the resource usage period calculation.
+     * @param mode The mode of the period defined by {@link TenantConstants#FIELD_PERIOD_MODE}.
+     * @param periodInDays The number of days defined by {@link TenantConstants#FIELD_PERIOD_NO_OF_DAYS}. 
+     * @return The period in days for which the resource usage is to be calculated.
+     * @throws NullPointerException if effectiveSince, targetDateTime or mode is {@code null}.
+     */
+    public static long calculateResourceUsagePeriod(
```
sophokles73: I would have expected this method to not be final but instead operate on the instance's fields instead of passing in the mode and periodInDays or doesn't that work?

kaniyan: The `DataVolume` is still using `DataVolumePeriod` not  `ResourceLimitsPeriod`. Until then the mode and periodInDays have to be passed.

sophokles73: We could create a common abstract base class for `DataVolumePeriod` and `ResourceLimitsPeriod` which defines the shared functionality. WDYT?

kaniyan: Yes we can create one base class. But I think then it needs to be marked as _deprecated_ for removal in 2.0.0. Once the _DataVolumePeriod_ has been removed, then this new base class should be merged with `ResourceLimitsPeriod`. Considering the new base class and the related changes, I am wondering if it is worthwhile to move now the `calculateResourceUsagePeriod` from the `PrometheusBasedResourceLimitChecks`. I think while removing the `DataVolumePeriod`, this activity can also be carried out. WDYT?

sophokles73: I agree, it's not worth the effort to move the code at all for the time being ...

kaniyan: For the time being I have reverted this method back to the `PrometheusBasedResourceLimitChecks`. Please take a look.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -292,6 +296,108 @@ private Span createSpan(final String name, final SpanContext parent, final Tenan
...
+                        connectionDurationUsagePeriod);
+                final String key = String.format("%s_minutes_consumed", tenant.getTenantId());
+
+                Optional.ofNullable(limitsCache)
+                        .map(ok -> limitsCache.get(key))
+                        .map(cachedValue -> Future.succeededFuture((long) cachedValue))
+                        .orElseGet(() -> executeQuery(queryParams, span)
+                                .map(minutesConnected -> addToCache(limitsCache, key, minutesConnected)))
+                        .map(minutesConnected -> {
+                            items.put("current period connection duration limit in minutes", minutesConnected);
```
sophokles73: this is not a *limit* but simply the current period's connection duration, isn't it?

kaniyan: yes, I have updated it and pushed the change.

sophokles73:
@kaniyan my understanding is that this PR is supposed to add the *possibility* to check the connection duration but not to add the actual checking to the protocol adapters, or is it?

kaniyan:
> @kaniyan my understanding is that this PR is supposed to add the _possibility_ to check the connection duration but not to add the actual checking to the protocol adapters, or is it?

yes,absolutely. I will raise another PR to add the actual checking to the protocol adapters.

Issue #1642, opened 2019-11-29
Add DeviceConnectionClient that connects to data grid directly


sophokles73:
The current `DeviceConnectionClientImpl` that is part of Hono's client module connects to a Device Connection service implementation using AMQP 1.0.
The Device Connection service implementation that ships with Hono uses an Infinispan data gri for storing the device connection information. Hono's standard protocol adapters could also connect directly to the data grid in order to read/write the connection info, thus skipping the Device Connection service and saving a network hop and protocol translation.

PR #1727, opened 2020-01-29
[#1642] Add Hotrod based DeviceConnectionClient

sophokles73:
The client implementation connects directly to an Infinispan data grid
to access device connection information. This client is an optimized
version of the existing client, skipping the Device Connection service
altogether. Note that this also means that the client can access all
data for all tenants and should therefore not be used with custom
protocol adapters.

It might also be possible to split up the PR into two parts:
A first one introducing the Infinispan based DeviceConnectionClient implementation and a second one which refactors the existing Device Connection service implementation to use the new client.


client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCache.java
```diff
@@ -0,0 +1,279 @@
...
+import io.vertx.core.Vertx;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * A HotrodCache.
+ *
+ * @param <K> The type of keys used by the cache.
+ * @param <V> The type of values stored in the cache.
+ */
+public final class HotrodCache<K, V> implements ConnectionLifecycle<HotrodCache<K, V>> {
```
dejanb: I find having another cache abstraction on top of the existing one a bit of over-complication. We have an abstraction in enmasse of CacheProvider, which adds necessary cache handling but then services use infinispan interfaces directly. 

https://github.com/EnMasseProject/enmasse/blob/master/iot/iot-infinispan-base/src/main/java/io/enmasse/iot/infinispan/cache/AbstractCacheProvider.java

Maybe at some point we can align two implementations more.

client-device-connection-infinispan/pom.xml
```diff
@@ -0,0 +1,58 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
dejanb: I'm not sure this module should be the "root" one of the whole Hono project. I'm not sure yet what would be the best place for it. Maybe put it in the "client" one, or as a part of the service implementation. Or even refactor the client one to have submodules like adapters and services.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCache.java
```diff
@@ -0,0 +1,279 @@
...
+        disconnect(r -> {});
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void disconnect(final Handler<AsyncResult<Void>> completionHandler) {
+
+        vertx.executeBlocking(r -> {
```
dejanb: Why not use RemoteCache or AdvanceCache interfaces and then use "async" variant of the calls instead of handling this manually?

dejanb:
@sophokles73 Did you considered implementing this as a "distributed cache" solution instead of the client-server model. It might be a better fit in my mind.

sophokles73:
>Did you considered implementing this as a "distributed cache" solution instead of the client-server model. It might be a better fit in my mind.

You mean with the data residing in caches at the protocol adapters? Does this allow the data to be partitioned among the adapter instances? If so, I wonder about the impact of adding/removing protocol adapter instances to this setup. I guess the data partitions would need to be rebalanced. When you scale out/in this would probably lead to a lot of traffic just for rebalancing. If this is not supported then we are talking about fully replicated caches, right? I wonder if that is a good approach when our overall goal is to keep the memory footprint low (and the footprint of our adapters already is too high FMPOV). WDYT?

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCache.java
sophokles73: The HotrodCache is adding the connection checking capability that we need for the readiness and liveness checks. Apart from that, it doesn't have much value. But we do need these checks in multiple places so I thought it would be a good idea to have them implemented in a single place.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCache.java
sophokles73: I originally had that but the code is not much different because you still need to adapt the standard Java `CompletableFuture` to vert.x `Promise` ...

client-device-connection-infinispan/pom.xml
sophokles73: We definitely need to refactor the client module. For the time being, I did not want to put it in there because it has a dependency on `service-base`, which `client` doesn't have so far. But I agree, this is far from optimal and we might be able to come up with a better layout if we move some of the more generic interfaces from service-base to core. Maybe we can talk about this next week at BCX.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCache.java
dejanb: Yeah, we created some separate components for infinispan registry that does the same. The only thing here is that this abstraction now "narrows" the original cache api, which the introduce one more thing to maintain going forward.
Maybe another topic for F2F at BCX to see if we can get two implementations more inlined. And maintained at one place.

client-device-connection-infinispan/pom.xml
dejanb: Sure. The only concern is that once we release with the module having it's maven coordinates defined. Later changes would be braking.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCache.java
dejanb: OK, makes sense.

dejanb:
> You mean with the data residing in caches at the protocol adapters? Does this allow the data to be partitioned among the adapter instances? If so, I wonder about the impact of adding/removing protocol adapter instances to this setup. I guess the data partitions would need to be rebalanced. When you scale out/in this would probably lead to a lot of traffic just for rebalancing. If this is not supported then we are talking about fully replicated caches, right? I wonder if that is a good approach when our overall goal is to keep the memory footprint low (and the footprint of our adapters already is too high FMPOV). WDYT?

I never investigated this in more details. Just wanted to reason out loud if it's something we can take a look in the future. 
In my mind "distributed cache" keeps data partitioned, so every adapter keeps what it knows. As these data are temporal, I don't think we need to rebalance data. But then again the question is how querying work with potential large number of instances.

client-device-connection-infinispan/pom.xml
sophokles73: This module is only used by the protocol adapters anyway so I do not see a big issue there.

sophokles73:
> As these data are temporal, I don't think we need to rebalance data. But then again the question is how querying work with potential large number of instances.

This is not really true. The data needs to be consistently shared among all protocol adapter instances. So either the client side caches need to replicate the data amongst them or they need to access the data from a server side cache (which we are currently doing). The latter approach has the advantage that we can more easily distribute the data to multiple nodes without too much overhead when adding/removing a node (which should happen very seldomly compared to adding/removing an adapter instance).

calohmn:
> So either the client side caches need to replicate the data amongst them or they need to access the data from a server side cache (which we are currently doing).

Choosing just between one or the other, I would also opt for the latter (server side cache).
In general I think a question is also what the read/write ratio of the contained data is. For data that is only rarely updated but gets read often, the client side cache performance might be worth the extra replication overhead.
As for our use-case, the last-known-gateway data is updated rather often, so, the server side cache is probably better.

I also thought about the extra cache to be added for #1272, storing a mapping between device id and id of the protocol adapter instance that has a command consumer for that device. For long-running command subscriptions via AMQP/MQTT (i.e. not many write operations for updating the mapping) with lots of commands getting handled (i.e. many read operations), a client side cache might be an option for that. However, a scenario of having short-term command subscriptions, with only one or a few commands getting received per new subscription also might be common (for HTTP it is always the case).
So, I don't see that much benefit necessarily in introducing client side caches.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/DeviceConnectionInfoCache.java
```diff
@@ -0,0 +1,67 @@
...
+
+import io.opentracing.SpanContext;
+import io.vertx.core.Future;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * A repository for keeping connection information about devices.
+ *
+ */
+public interface DeviceConnectionInfoCache {
```
calohmn: This name very much implies there being only one cache for the device connection data. However, for #1272 there will be other data added (mapping between device id and id of the protocol adapter instance that has a command consumer for that device). I would rather put this data in a separate Infinispan cache. So then there would be 2 Infinispan caches used for device connection data.

So, for now the name is Ok, for #1272 I think it will better be renamed to something like `LastKnownGatewayInfoCache`.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -719,8 +744,9 @@ protected void doStop(final Promise<Void> stopPromise) {
      *         connection cannot be established.
      * @throws NullPointerException if serviceName is {@code null}.
      * @throws IllegalArgumentException if factory is {@code null}.
+     * @param <T> The type of connection that the factory uses.
      */
-    protected final Future<HonoConnection> connectToService(final ConnectionLifecycle<HonoConnection> factory, final String serviceName) {
+    protected final <T> Future<T> connectToService(final ConnectionLifecycle<T> factory, final String serviceName) {
```
calohmn: I think the type parameter `T` should better be renamed so that it doesn't hide the `T` parameter of the enclosing class.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -738,12 +764,13 @@ protected void doStop(final Promise<Void> stopPromise) {
      *         connection cannot be established.
      * @throws NullPointerException if serviceName is {@code null}.
      * @throws IllegalArgumentException if factory is {@code null}.
+     * @param <T> The type of connection that the factory uses.
      */
-    protected final Future<HonoConnection> connectToService(
-            final ConnectionLifecycle<HonoConnection> factory,
+    protected final <T> Future<T> connectToService(
```
calohmn: I think the type parameter T should better be renamed so that it doesn't hide the T parameter of the enclosing class.

services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionService.java
```diff
@@ -46,100 +37,16 @@
  */
 public class RemoteCacheBasedDeviceConnectionService extends EventBusDeviceConnectionAdapter implements DeviceConnectionService, HealthCheckProvider {
 
-    private static final String CACHE_NAME = "device-connection";
-
-    private RemoteCacheContainer cacheManager;
-    private BasicCache<String, String> cache;
-    private AtomicBoolean connecting = new AtomicBoolean(false);
+    private DeviceConnectionInfoCache cache;
```
calohmn: final?

adapters/amqp-vertx/src/main/resources/logback-spring.xml
```diff
@@ -41,6 +42,7 @@
     <logger name="org.eclipse.hono.client" level="DEBUG"/>
     <logger name="org.eclipse.hono.client.impl.AbstractRequestResponseClient" level="DEBUG"/>
     <logger name="org.eclipse.hono.connection" level="DEBUG"/>
+    <logger name="org.eclipse.hono.deviceconnection.infinispan.client" level="DEBUG"/>
```
calohmn: How about adding the added logger definitions also to the other adapters?

dejanb:
@sophokles73 @calohmn Sounds good. Let's stick with the client-server model then

PR #1730, opened 2020-01-30
[#1346] Prevent instantiating deprecated classes

dejanb:


dejanb:
I'll merge this as we discussed we want to keep both implementation for the time being (we'll decide which one is used based on the future developments). And also, not instantiating deprecated components seems like a common sense to me.

PR #1731, opened 2020-01-30
[#1716] Extend protocol adapters to check for connection duration limit.

kaniyan:
In this PR, the adapter base class has been extended to check the connection duration limit before accepting any connection request from devices.


kaniyan:
I will update the tests, so that the travis build succeeds.

PR #1733, opened 2020-01-30
[#1346] Remove event bus from tenant management endpoint

dejanb:
First stab at converting Tenant API to direct implementation. Note that this is merely conversion of existing code. I think we can do more work in aligning and improving implementation across services, but I would leave that for the separate PR considering the time we have for the release

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/management/Util.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/management/Util.java
@@ -15,10 +15,12 @@ package org.eclipse.hono.service.management;
 
 import java.net.HttpURLConnection;
 import java.util.Objects;
+import java.util.Optional;
 
 import org.eclipse.hono.service.http.HttpUtils;
 import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.MessageHelper;
+import org.eclipse.hono.util.RegistryManagementConstants;
 
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
@@ -108,20 +110,22 @@ public class Util {
     }
 
     /**
-     * Creates a response based on generic result.
+     * Writes a response based on generic result.
      * <p>
      * The behavior is as follows:
      * <ol>
      * <li>Set the status code on the response.</li>
      * <li>If the status code represents an error condition (i.e. the code is &gt;= 400),
-     * then the JSON object passed in to the returned handler is written to the response body.</li>
-     * <li>Otherwise, the JSON object is written to the response body and the given custom handler is
+     * then the JSON object passed in the result is written to the response body.</li>
+     * <li>If the result is created (the code is = 201), the JSON object is written to the response body and the given custom handler is
      * invoked (if not {@code null}).</li>
+     * <li>Sets the status of the tracing span and finishes it.</li>
+     * <li>Ends a response.</li>
      * </ol>
      *
      * @param ctx The routing context of the request.
      * @param result The generic result of the operation.
-     * @param customHandler An (optional) handler for post processing the HTTP response, e.g. to set any additional HTTP
+     * @param customHandler An (optional) handler for post processing successful HTTP response, e.g. to set any additional HTTP
      *                      headers. The handler <em>must not</em> write to response body. May be {@code null}.
      * @param span The active OpenTracing span for this operation. The status of the response is logged and span is finished.
      */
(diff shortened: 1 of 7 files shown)


sophokles73:
@dejanb one of the integration tests is failing. Can you check if this is related to your changes?

dejanb:
> I have to admit that I have a very bad gut feeling regarding this change and the ones that need to follow for Credentials. This is partly due to the fact that you indicated that you won't have any time to work on it next week before the 1.1.0 release and partly because it looks like the HTTP Management spec seems to require some more review in some places. I would therefore rather revert the previous changes you made to the Device management endpoint and move the whole issue to 1.2.0. WDYT?

I'm not a big fan of trying to do it all in one big PR that will take weeks to resolve. I would rather do it incrementally and improve things as we go. Having said that, as there is a time pressure I can only do Credential API refactoring on this level until the release. So if you wish to revert the changes and revisit this later on, please do so.

sophokles73:
> I'm not a big fan of trying to do it all in one big PR that will take weeks to resolve. I would rather do it incrementally and improve things as we go.

Neither am I. However, you seem to have changed the behavior of the endpoint with this (very small) PR and given the circumstances, I wonder if we will be able to figure out the minimum changes required in time.

> Having said that, as there is a time pressure I can only do Credential API refactoring on this level until the release.

I would be ok with just doing the minimal changes necessary. However, even that might not be as easy as we have imagined and might therefore require a little more time to do it properly. Do you think you will find the time to do this until Feb 6th? I am willing to do the reviewing, but if something is not consistent with the existing behavior then we will need to discuss and adapt, wouldn't you agree? My only concern is that you will probably not find the time to do that. But that is a question you need to answer. If you can then let's do it. If not, then let's revert. WDYT?

dejanb:
> I would be ok with just doing the minimal changes necessary. However, even that might not be as easy as we have imagined and might therefore require a little more time to do it properly. Do you think you will find the time to do this until Feb 6th? I am willing to do the reviewing, but if something is not consistent with the existing behavior then we will need to discuss and adapt, wouldn't you agree? My only concern is that you will probably not find the time to do that. But that is a question you need to answer. If you can then let's do it. If not, then let's revert. WDYT?

Why not just revet to instantiating the original implementation in application config? It will be much less messy than reverting the whole code and it will allow use to fix bugs in both versions (see #1735) and make a switch when we're confident in the new implementation.

service-base/src/main/java/org/eclipse/hono/service/management/tenant/AbstractTenantManagementHttpEndpoint.java
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    private void createTenant(final RoutingContext ctx) {
+
+        final Span span = Util.newChildSpan(SPAN_NAME_CREATE_TENANT, TracingHandler.serverSpanContext(ctx), tracer, getClass().getSimpleName());
+
+        final String tenantId = getRequestParam(PARAM_TENANT_ID, ctx, span, true);
+
+        final JsonObject payload = ctx.get(KEY_REQUEST_BODY);
+        if (payload == null) {
```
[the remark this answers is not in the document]
dejanb: This is a valid concern that I think we need fix in a separate issue (#1736). I aligned new implementation to the current one and we should fix both of them, once we decide on the approach we want to take.

dejanb:
> @dejanb one of the integration tests is failing. Can you check if this is related to your changes?

This is fixed with #1735

service-base/src/main/java/org/eclipse/hono/service/management/tenant/AbstractTenantManagementHttpEndpoint.java
```diff
@@ -0,0 +1,337 @@
...
+
+    /**
+     * Check that the tenantId value is not blank then
+     * update the payload (that was put to the RoutingContext ctx with the
+     * key {@link #KEY_REQUEST_BODY}) with the tenant value retrieved from the RoutingContext.
+     * The tenantId value is associated with the key {@link RegistryManagementConstants#FIELD_PAYLOAD_TENANT_ID}.
+     *
+     * @param ctx The routing context to retrieve the JSON request body from.
+     */
+    protected void updatePayloadWithTenantId(final RoutingContext ctx) {
```
jbtrystram: Maybe this can moved to `AbstractHttpEndpoint` as the other endpoints do the same check ?

dejanb: I was putting those in `Util` class so far, but as @sophokles73 suggested we can introduce a new abstract class with all these methods, to keep `AbstractHttpEndpoint` free of device registry abstractions.

service-base/src/main/java/org/eclipse/hono/service/management/tenant/AbstractTenantManagementHttpEndpoint.java
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    /**
+     * Checks the request payload for validity.
+     *
+     * @param payload The payload to check.
+     * @return boolean The result of the check : {@link Boolean#TRUE} if the payload is valid, {@link Boolean#FALSE} otherwise.
+     * @throws NullPointerException If the payload is {@code null}.
+     */
+    boolean isValidRequestPayload(final JsonObject payload) {
```
sophokles73: is this supposed to be overridden?

service-base/src/main/java/org/eclipse/hono/service/management/tenant/AbstractTenantManagementHttpEndpoint.java
```diff
@@ -0,0 +1,340 @@
...
+        if (isValidRequestPayload(payload)) {
+            logger.debug("creating tenant [{}]", tenantId);
+
+            addNotPresentFieldsWithDefaultValuesForTenant(payload);
+
+            final Promise<OperationResult<Id>> result = Promise.promise();
+            result.future().setHandler(handler -> {
+                final OperationResult<Id> operationResult = handler.result();
+
+                final String createdTenantId = Optional.ofNullable(operationResult.getPayload()).map(Id::getId).orElse(null);
```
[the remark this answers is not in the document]
dejanb: It is `null` when operation fails ... It's used like that across different services and implementations ... Let's revisit this after we refactor all services in an effort to clean/improve this code.

service-base/src/main/java/org/eclipse/hono/service/management/Util.java
```diff
@@ -121,7 +123,7 @@ public static final Span newChildSpan(final String operationName, final SpanCont
      *
      * @param ctx The routing context of the request.
      * @param result The generic result of the operation.
-     * @param customHandler An (optional) handler for post processing the HTTP response, e.g. to set any additional HTTP
+     * @param customHandler An (optional) handler for post processing successful HTTP response, e.g. to set any additional HTTP
```
sophokles73: can you please review the method's description in the JavaDoc? I don't think that it reflects the behavior of this method. It refers to a *returned handler* but the method returns `void`. It also doesn't mention that the body is only written and the handler is only invoked if the status code is 201 Created. IMHO the status code to check for should also be passed in as a parameter in order to make this method more useful. WDYT?

dejanb: Updated comments a bit. I wouldn't change the signature unless we actually have a case where we would reuse it.

dejanb:
> extending `AbstractHttpEndpoint` in this way doesn't feel good to me because we try to _suppress_ its concept of forwarding requests via the event bus and thus makes it extremely hard to understand what is really going on during runtime. Maybe we can first extract the event bus related concepts of `AbstractHttpEndpoint` into a new subclass `AbstractEventBusBasedHttpAdapter`? The `AbstractDeviceManagementHttpAdapter` and `AbstractTenantManagementHttpEndpoint` classes would then be much cleaner and easier to understand. WDYT?

We discussed this already above. I agree, and I would at least have `AbstractManagementHttpEndpoint` which would contain methods we now keep in `Util` and see if we need another class or we can simply deprecate methods related to event bus. 
But, let's let's keep this first step scoped and make these kind of improvements in subsequent pr(s).

service-base/src/main/java/org/eclipse/hono/service/management/tenant/AbstractTenantManagementHttpEndpoint.java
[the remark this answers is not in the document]
sophokles73: ok

dejanb:
@sophokles73 I raised a new issue to track abstract classes improvement (#1760)

PR #1737, opened 2020-01-31
[#1328] Remove support for legacy "control" endpoints

### LINKED ISSUE #1328: Remove deprecated "control" endpoints
This relates to #930 and #1073.

The old northbound and southbound "control" endpoints should be removed.
This also entails removing the bit flag encoded in the device-facing reply-to address (see `org.eclipse.hono.client.Command#FLAG_*`) and removing the "*LegacyEndpoint*" test classes (having `useLegacyCommandEndpoint` return true).

Note: The `CommandConsumerFactoryImpl` still needs to work with two kinds of receiver links. The receiver link with the "control" address needs its address renamed to something like "cmd_internal". `DelegatedCommandSenderImpl` needs to target that endpoint then.

It has still got to be decided, at which point these things should be done.

calohmn:
This fixes #1328. 

Also adds command message address validation as the "to" property of a command message is now mandatory (see 2nd commit).


sophokles73:
@calohmn you should add a corresponding note to the release notes

calohmn:
@sophokles73 Yes, added here: 58c9434829d0a4caef189486feb0b1af80068eb3.

Issue #1441, opened 2019-08-12
Hardcoded event bus send timeout in AbstractEndpoint leads to timeouts for long lasting operations


DanielMaier-BSI:
https://github.com/eclipse/hono/blob/c21831c27fe11b21d8b09f4d39592b5ed2ad4293/service-base/src/main/java/org/eclipse/hono/service/AbstractEndpoint.java#L146-L151 sets send timeout of event bus to 3 seconds. This timeout gets also used by RequestResponseEndpoint. This leads to issues during creation of tenants in our tenant manager implementation as creating tenant in dependent systems, e.g. enmasse, may last longer. Would it make sense from your perspective to make this configurable somehow?

sophokles73:
It does and we would be happy to receive a corresponding PR :-)

PR #1743, opened 2020-02-04
[#1441] Make hard-coded event bus send timeout configurable.

kaniyan:
This PR is intended for the issue #1441. 
The hard-coded send time out value can now be configured.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/config/ServiceConfigProperties.java
+++ b/core/src/main/java/org/eclipse/hono/config/ServiceConfigProperties.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2020 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -27,7 +27,19 @@ public class ServiceConfigProperties extends ServerConfig {
      */
     public static final int DEFAULT_RECEIVER_LINK_CREDITS = 100;
 
+    /**
+     * The default send timeout value in milliseconds, which is 
+     * to be used when sending a message on the vert.x event bus.
+     */
+    public  static final long DEFAULT_SEND_TIMEOUT_IN_MS = 3000;
+
     private static final int MIN_PAYLOAD_SIZE  = 128; // bytes
+
+    /**
+     * The minimum send timeout value in milliseconds is 500 ms
+     * and any value less than this minimum value is not accepted.
+     */
+    private static final long MIN_SEND_TIMEOUT_IN_MS = 500;
 
     private boolean singleTenant = false;
     private boolean networkDebugLogging = false;
@@ -35,6 +47,7 @@ public class ServiceConfigProperties extends ServerConfig {
     private int maxPayloadSize = 2048;
     private int receiverLinkCredit = DEFAULT_RECEIVER_LINK_CREDITS;
     private String corsAllowedOrigin = "*";
+    private long sendTimeOutInMs = DEFAULT_SEND_TIMEOUT_IN_MS;
 
     /**
      * Sets the maximum size of a message payload this server accepts from clients.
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/eclipse/hono/config/ServiceConfigProperties.java
```diff
@@ -205,4 +212,32 @@ public final String getCorsAllowedOrigin() {
...
+     * message on the vert.x event bus.
+     * <p>
+     * The default value of this property is {@link #DEFAULT_SEND_TIMEOUT_IN_MS}.
+     * 
+     * @param sendTimeOutInMs The send timeout value in milliseconds.
+     * @throws IllegalArgumentException if the timeout is negative.
+     */
+    public final void setSendTimeOut(final long sendTimeOutInMs) {
+        if (sendTimeOutInMs < 0) {
+            throw new IllegalArgumentException("send time out value must not be negative");
```
sophokles73: does it make sense to set this to 0?

kaniyan: Just I checked the implementation of `DeliveryOptions.setSendTimeout(long timeout)`. They throw `IllegalArgumentException`, if the value is _< 1_ but its not documented in that method's _javadoc_. Anyway it cannot be set to 0. I will update it.

sophokles73: My point is actually that IMHO it doesn't make sense to accept a very small value here, i.e. we should enforce a minimum value that makes some sense, e.g. 500ms?

kaniyan: I see now what do you mean. But how do we conclude for such a minimum value?  Anyway we set a default value of 3000ms. `DeliveryOptions.setSendTimeout(...)` allows to set any timeout value greater than 0ms. IMHO why should we forbid this flexibility by setting a minimum value?

sophokles73: Well, as I pointed out: because the system will not work if you set it to 1ms, will it?

kaniyan: > Well, as I pointed out: because the system will not work if you set it to 1ms, will it?

Ok I see the point. The systems commonly we use are not yet so fast enough to set the value to 1 ms ;-). I will update it to a minimum value of 500ms.

kaniyan: I have updated it with a minimum value of _500ms_ and pushed the changes. Would you mind taking a look?

service-base/src/main/java/org/eclipse/hono/service/AbstractEndpoint.java
```diff
@@ -171,10 +172,23 @@ public void registerReadinessChecks(final HealthCheckHandler handler) {
      *  
      * @param spanContext The {@code SpanContext} (may be {@code null}).
      * @return The {@code DeliveryOptions}.
+     * @deprecated Use {@link #createEventBusMessageDeliveryOptions(long, SpanContext)} instead.
```
sophokles73: no need to deprecate this anymore, since it is private API now ...

kaniyan: ok, I will remove it.

kaniyan:
> would you mind documenting this new feature in the release notes as well?

@sophokles73 sure. WDYT of the below one?
* Hono supports now setting a time out value when sending a message on the vert.x event bus. This is
  configured using the property `sendTimeOutInMs` in `org.eclipse.hono.config.ServiceConfigProperties`.

kaniyan:
> > would you mind documenting this new feature in the release notes as well?
> 
> @sophokles73 sure. WDYT of the below one?
> 
> * Hono supports now setting a time out value when sending a message on the vert.x event bus. This is  configured using the property `sendTimeOutInMs` in `org.eclipse.hono.config.ServiceConfigProperties`.

I have added the above to the release notes.

PR #1744, opened 2020-02-04
[#1726] Fix: accept command responses without payload in AMQP adapter.

### LINKED ISSUE #1726: AMQP adapter does not accept command response without payload
The Command & Control API defines that command responses MAY contain a payload (https://www.eclipse.org/hono/docs/dev/api/command-and-control/#send-a-request-response-command). The AMQP adapter rejects command responses without payload with the error message "empty notifications must not contain payload".

b-abel:
This fixes #1726.

### DIFF
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
@@ -978,20 +978,20 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
                     "empty notifications must not contain payload"));
         }
 
-        return contentTypeCheck.future()
-                .compose(ok -> {
-                    switch (context.getEndpoint()) {
-                    case TELEMETRY:
-                        return doUploadMessage(context, resource, getTelemetrySender(resource.getTenantId()), currentSpan);
-                    case EVENT:
-                        return doUploadMessage(context, resource, getEventSender(resource.getTenantId()), currentSpan);
-                    case COMMAND_RESPONSE:
-                        return doUploadCommandResponseMessage(context, resource, currentSpan);
-                    default:
-                        return Future
-                                .failedFuture(new ClientErrorException(HttpURLConnection.HTTP_NOT_FOUND, "unknown endpoint"));
-                    }
-                });
+        switch (context.getEndpoint()) {
+        case TELEMETRY:
+            return contentTypeCheck.future()
+                    .compose(ok -> doUploadMessage(context, resource, getTelemetrySender(resource.getTenantId()),
+                            currentSpan));
+        case EVENT:
+            return contentTypeCheck.future()
+                    .compose(ok -> doUploadMessage(context, resource, getEventSender(resource.getTenantId()),
+                            currentSpan));
+        case COMMAND_RESPONSE:
+            return doUploadCommandResponseMessage(context, resource, currentSpan);
+        default:
+            return Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_NOT_FOUND, "unknown endpoint"));
+        }
     }
 
     private Future<ProtonDelivery> doUploadMessage(
--- a/adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapterTest.java
+++ b/adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapterTest.java
@@ -544,7 +544,7 @@ public class AbstractVertxBasedHttpProtocolAdapterTest {
         final AbstractVertxBasedHttpProtocolAdapter<HttpProtocolAdapterProperties> adapter = getAdapter(server, null);
 
         // WHEN a device publishes a command response with an empty payload
-        final Buffer payload = Buffer.buffer();
+        final Buffer payload = null;
         final HttpServerResponse response = mock(HttpServerResponse.class);
         final RoutingContext ctx = newRoutingContext(payload, "application/text", mock(HttpServerRequest.class), response);
         when(ctx.addBodyEndHandler(any(Handler.class))).thenAnswer(invocation -> {
@@ -563,7 +563,7 @@ public class AbstractVertxBasedHttpProtocolAdapterTest {
                 eq("tenant"),
                 any(),
                 eq(ProcessingOutcome.FORWARDED),
-                eq(payload.length()),
+                eq(0),
                 any());
     }
 
--- a/adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapterTest.java
+++ b/adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapterTest.java
@@ -528,6 +528,46 @@ public class AbstractVertxBasedHttpProtocolAdapterTest {
     }
 
     /**
+     * Verifies that the adapter accepts a command response message with an empty body.
+     */
+    @SuppressWarnings("unchecked")
+    @Test
+    public void testUploadEmptyCommandResponseSucceeds() {
+
+        // GIVEN an adapter with a downstream application attached
+        final CommandResponseSender sender = mock(CommandResponseSender.class);
+        when(sender.sendCommandResponse(any(CommandResponse.class), (SpanContext) any())).thenReturn(Future.succeededFuture(mock(ProtonDelivery.class)));
+
+        when(commandConsumerFactory.getCommandResponseSender(anyString(), anyString())).thenReturn(Future.succeededFuture(sender));
+
+        final HttpServer server = getHttpServer(false);
+        final AbstractVertxBasedHttpProtocolAdapter<HttpProtocolAdapterProperties> adapter = getAdapter(server, null);
+
+        // WHEN a device publishes a command response with an empty payload
+        final Buffer payload = Buffer.buffer();
+        final HttpServerResponse response = mock(HttpServerResponse.class);
+        final RoutingContext ctx = newRoutingContext(payload, "application/text", mock(HttpServerRequest.class), response);
+        when(ctx.addBodyEndHandler(any(Handler.class))).thenAnswer(invocation -> {
+            final Handler<Void> handler = invocation.getArgument(0);
+            handler.handle(null);
+            return 0;
+        });
+
+        adapter.uploadCommandResponseMessage(ctx, "tenant", "device", CMD_REQ_ID, 200);
+
+        // then it is forwarded successfully
+        verify(response).setStatusCode(202);
+        verify(response).end();
+        verify(metrics).reportCommand(
+                eq(Direction.RESPONSE),
+                eq("tenant"),
+                any(),
+                eq(ProcessingOutcome.FORWARDED),
+                eq(payload.length()),
+                any());
+    }
+
+    /**
      * Verifies that the adapter fails the upload of a command response with a 403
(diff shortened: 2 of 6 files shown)


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -978,20 +978,20 @@ protected void onCommandReceived(final TenantObject tenantObject, final ProtonSe
...
+                            currentSpan));
+        case EVENT:
+            return contentTypeCheck.future()
+                    .compose(ok -> doUploadMessage(context, resource, getEventSender(resource.getTenantId()),
+                            currentSpan));
+        case COMMAND_RESPONSE:
+            return doUploadCommandResponseMessage(context, resource, currentSpan);
+        default:
+            return Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_NOT_FOUND, "unknown endpoint"));
+        }
```
sophokles73: FMPOV the only thing that needs to be changed/fixed is the *isPayloadOfIndicatedType* method which should verify that the payload is indeed empty if the content type is `EventConstants.CONTENT_TYPE_EMPTY_NOTIFICATION` but otherwise return `true`. WDYT?

b-abel: That makes sense. The problem I see here is that the method does not do what its name promises, and the method is called in all protocol adapters for every telemetry and event message. My suggestion would be to change it to the way you described it. Additionally, we rename it, for example to `isEmtpyNotificationReallyEmpty`. 
Do we also want to change the invocations so that it is only executed for events?

sophokles73: > My suggestion would be to change it to the way you described it. Additionally, we rename it, for example to isEmtpyNotificationReallyEmpty.

+1

> Do we also want to change the invocations so that it is only executed for events?

No, let's keep it the way it is. IIRC we allow devices to send empty notifications as both telemetry as well as event messages ...

b-abel: Just to be sure: we are changing the behavior of the protocol adapters by that in that they now generally accept empty messages, aren't we? At least one unit test suggest that it was intended to block empty messages (except the empty notification): https://github.com/eclipse/hono/blob/0840f2e6c173d3d47fba6cbdfecf5ceb363e0147/service-base/src/test/java/org/eclipse/hono/service/AbstractProtocolAdapterBaseTest.java#L483
This test fails with the change discussed here. I've seen no failing IT for this.

sophokles73: Yeah, we originally thought it would have value to prevent devices from sending empty messages. However, I can't recall what that value would actually be. On the other hand, I can imagine a device sending an empty message with an application specific content-type which can be used to trigger some behavior in a downstream application. Requiring the device to include a byte in the payload seems useless, doesn't it?
The Telemetry API currently defines that the payload of a message MUST consist of a Data section. The same definition is used by the Event API which doesn't seem to make sense since we already allow the payload to be empty for empty notification (events). The C&C API explicitly allows the payload of a command response to be empty.
So, FMPOV we should
* proceed as discussed with this PR
* remove the conflicting unit test
* adapt the Telemetry API wording to make the payload optional

b-abel: The user guides of the protocol adapters state as well that the payload is required. Should we change the documentation there as well? FMPOV this needs to be tested before changing the user guide. If we remove the explicit check, empty messages might or might not work in the adapters.

sophokles73: You could add corresponding unit tests, I guess

sophokles73: And yes, I think we should adapt the user guides accordingly as well.

b-abel: The MQTT adapter does not allow an empty payload and explicitly checks that (removing this check involves adding null checks further down...). The HTTP adapter, on the other hand, takes messages with empty payload, but then does not set the content type (`MessageHelper.setPayload()`). 
Of course, this could be changed, but I'm not sure if we want to change the behavior of the protocol adapters in a fix for one specific adapter on the last day before the release. 
Anyway, we should have a separate issue for this.FMPOV allowing empty messages with custom content types is a great idea but a separate feature and not part of this bug fix. WDYT?

sophokles73: That is probably true. Do the other adapters allow empty command responses?

b-abel: COAP: no, Lora: no, Sigfox: yes.

sophokles73: Well at least the CoAP and Sigfox adapters do not support C&C at all.
The Lora adapter should be fixed as well, I guess. What about the MQTT and HTTP adapters?

b-abel: I'm sorry @sophokles73, I misread your question. My answer referred to which adapters explicitly reject empty payload for messages in general (i.e. especially telemetry and events). The bug in this ticket affects definitively only the AMQP adapter. All other adapters seem to have no problem with empty command responses.

sophokles73: ok, then let's fix the AMQP adapter and add unit tests that make sure that all adapters which support C&C accept empty command responses.

adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapterTest.java
```diff
@@ -527,6 +527,46 @@ public void testUploadCommandResponseWaitsForAcceptedOutcome() {
...
+        final CommandResponseSender sender = mock(CommandResponseSender.class);
+        when(sender.sendCommandResponse(any(CommandResponse.class), (SpanContext) any())).thenReturn(Future.succeededFuture(mock(ProtonDelivery.class)));
+
+        when(commandConsumerFactory.getCommandResponseSender(anyString(), anyString())).thenReturn(Future.succeededFuture(sender));
+
+        final HttpServer server = getHttpServer(false);
+        final AbstractVertxBasedHttpProtocolAdapter<HttpProtocolAdapterProperties> adapter = getAdapter(server, null);
+
+        // WHEN a device publishes a command response with an empty payload
+        final Buffer payload = Buffer.buffer();
```
sophokles73: I am not sure if this is the same as *no payload*. Can you change this to `null`?

b-abel: From what I understand Netty (and thus Vert.x MQTT) does not support this: 
https://github.com/netty/netty/blob/56055f4404fbebeb6403165d352d62330a1bf81d/codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttPublishMessage.java#L41-L52
The [MQTT 3.1.1 standard](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718040) says: " It is valid for a PUBLISH Packet to contain a zero length payload." Which of course does not tell how this is implemented... 
Hono's MQTT adapter will throw null pointer exceptions.

b-abel: @sophokles73 If you prefer I can still change the test so that the payload is `null` and fix the (few) places where this would cause NPEs in the HTTP and MQTT adapters.

sophokles73: no

b-abel:
I removed the laces where a payload that is null would cause NPEs for the HTTP adapter where the request body can be null. To be on the safe side and to keep the code similar to the HTTP case, I added those checks to MQTT adapter as well.

adapters/mqtt-vertx-base/src/test/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapterTest.java
```diff
@@ -679,6 +681,49 @@ public void testUploadTelemetryMessageSendsPubAckOnSuccess(final VertxTestContex
...
+        when(messageFromDevice.qosLevel()).thenReturn(MqttQoS.AT_MOST_ONCE);
+        when(messageFromDevice.messageId()).thenReturn(5555555);
+        when(messageFromDevice.topicName()).thenReturn("command///res/1010f8ab0b53-bd96-4d99-9d9c-56b868474a6a/200");
+
+        // ... with an empty payload
+        when(messageFromDevice.payload()).thenReturn(null);
+        final ResourceIdentifier address = ResourceIdentifier
+                .fromString("command/my-tenant/4712/res/1010f8ab0b53-bd96-4d99-9d9c-56b868474a6a/200");
+        adapter.uploadCommandResponseMessage(newMqttContext(messageFromDevice, endpoint), address)
+                .setHandler(ctx.completing());
```
sophokles73: this  should be `ctx.succeeding(Handler<T>)` and the assertion should be put into the handler. Otherwise the test context will most likely be (successfully) completed before the assertion is executed.
BTW IMHO you should also verify that the command sender has been invoked ...

b-abel: I pushed a fix.

adapters/mqtt-vertx-base/src/test/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapterTest.java
```diff
@@ -712,16 +719,19 @@ public void testUploadEmptyCommandResponseSucceeds(final VertxTestContext ctx) {
         final ResourceIdentifier address = ResourceIdentifier
                 .fromString("command/my-tenant/4712/res/1010f8ab0b53-bd96-4d99-9d9c-56b868474a6a/200");
         adapter.uploadCommandResponseMessage(newMqttContext(messageFromDevice, endpoint), address)
-                .setHandler(ctx.completing());
+                .setHandler(ctx.succeeding(result -> {
+                    verify(sender).sendCommandResponse(any(), any());
```
sophokles73: this needs to be wrapped in `ctx.verify(Handler<Void>)`

b-abel: Right. Sorry. Done that.

PR #1747, opened 2020-02-05
Update documentation for removal of the obsolete DataVolumePeriod

kaniyan:


kaniyan:
> In order to improve readability I would suggest to move the Resource Limits Period Configuration Format section _below_ the Data Volume Configuration Format.

I have moved the Resource Limits Period Configuration Format section as mentioned above and pushed the changes.

site/documentation/content/api/tenant/index.md
```diff
@@ -279,18 +270,18 @@ The table below contains the properties which are used to configure a tenant's d
...
 | *effective-since*        | *yes*     | *string*      | `-`           | The point in time at which the current settings became effective, i.e. the start of the first accounting period based on these settings. The value MUST be an [ISO 8601 compliant *combined date and time representation in extended format*](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations).|
 | *max-bytes*              | *no*      | *number*      | `-1`          | The maximum number of bytes allowed for the tenant for each accounting period. MUST be an integer. A negative value indicates that no limit is set. |
-| *period*                 | *no*      | *object*      | `-`          | The mode and length of an accounting period, i.e. the data usage can limited based on the defined number of days or on a monthly basis. |
+| *period*                 | *yes*      | *object*     | `-`           | The mode and length of an accounting period, i.e. the data usage can limited based on the defined number of days or on a monthly basis. For more information, please refer to the [resource limits period]({{< relref "#resource-limits-period-configuration-format" >}}).|
```
sophokles73: This actually constitutes a breaking change of the (public) Tenant API. I don't think that we can do this in 1.1.0. However, IIRC the Prometheus based resource limits check currently expects to always find this information, doesn't it? If so, we will need to change this and fall back to a default value (e.g. `monthly`) if no resource limits period is specified.
The HTTP management API also does not require this to be specified when defining Data Volume limits for a tenant.

site/documentation/content/api/tenant/index.md
```diff
@@ -258,19 +258,10 @@ The table below contains the properties which are used to configure a tenant's c
...
 | *effective-since*        | *yes*     | *string*      | `-`           | The point in time at which the current settings became effective, i.e. the start of the first accounting period based on these settings. The value MUST be an [ISO 8601 compliant *combined date and time representation in extended format*](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations).|
 | *max-minutes*            | *no*      | *number*      | `-1`        | The maximum connection duration in minutes allowed for the tenant for each accounting period. MUST be an integer. Minus one indicates that no limit is set. |
-| *period*                 | *no*      | *object*      | `-`          | The mode and length of an accounting period, i.e. the resource usage is calculated based on the defined number of days or on a monthly basis. |
+| *period*                 | *yes*     | *object*      | `-`         | The mode and length of an accounting period, i.e. the resource usage is calculated based on the defined number of days or on a monthly basis.  For more information, please refer to the [resource limits period]({{< relref "#resource-limits-period-configuration-format" >}}).|
```
sophokles73: Since this is a new (optional) type of limits we are introducing, it is ok to define it as mandatory from a formal point of view. However, IMHO we should keep this aligned with the Data Volume limit and thus better keep it optional, also defining a fall-back default if not set explicitly. The HTTP Management API also needs to be adapted accordingly

kaniyan: It makes sense to keep it optional and in sync with that of the _DataVolume_. I would update the documentation here and also the HTTP Management API to set it as optional. I would propose a separate PR for setting the default value to _monthly_.

site/documentation/content/api/tenant/index.md
kaniyan: I would update the documentation here to set is as _optional_ and a separate PR for setting the default value to _monthly_. WDYT?

sophokles73: If we change this to *optional* then the Prometheus based implementation breaks. So, I'd rather have that case handled as part of this PR.

kaniyan: As part of this PR, I would change the _period_ as _optional_ and set the default value to _monthly_ for both _DataVolume_ and the _ConnectionDuration_. Just to make sure before making any changes.

sophokles73: This PR should not change the period from being optional (as it had been before).
We should probably document, that if the period is not set explicitly, *monthly* is assumed as the default value. The Prometheus based implementation needs to be adapted accordingly.

kaniyan: > This actually constitutes a breaking change of the (public) Tenant API. I don't think that we can do this in 1.1.0.

@sophokles73 Please correct me if I am wrong. Removing `DataVolumePeriod` type from Tenant API FMPOV is also a breaking change of the (public) Tenant API in 1.1.0. May be we should remove that  in 2.0.0. Unless the users replace it with the new `ResourceLimitsPeriod` type the build fails.

sophokles73: Not sure what you mean by *the build fails*. Which build is that?

PR #1762, opened 2020-02-11
Add command and control to coap.

boaks:
Use query-parameter for "hono-ttd" instead of option and replace
content-type "application/vnd.eclipse-hono-empty-notification" by
query-parameter "empty". Both, custom option and content-type, would
require a code-point reservation at IANA. Though the use-case is very
hono specific, I think using query-parameters is more appropriate.

### DIFF
--- a/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
+++ b/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
@@ -307,13 +307,13 @@ public abstract class AbstractVertxBasedCoapAdapter<T extends CoapAdapterPropert
     private Future<CoapServer> bindSecureEndpoint(final CoapServer startingServer, final NetworkConfig config) {
 
         final ApplicationLevelInfoSupplier deviceResolver = Optional.ofNullable(honoDeviceResolver)
-                .orElse(new DefaultDeviceResolver(context, getConfig(), getCredentialsClientFactory()));
+                .orElse(new DefaultDeviceResolver(context, tracer, getTypeName(), getConfig(), getCredentialsClientFactory()));
         final PskStore store = Optional.ofNullable(pskStore)
                 .orElseGet(() -> {
                     if (deviceResolver instanceof PskStore) {
                         return (PskStore) deviceResolver;
                     } else {
-                        return new DefaultDeviceResolver(context, getConfig(), getCredentialsClientFactory());
+                        return new DefaultDeviceResolver(context, tracer, getTypeName(), getConfig(), getCredentialsClientFactory());
                     }
                 });
 
--- a/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
+++ b/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
@@ -49,6 +49,7 @@ import org.eclipse.hono.auth.Device;
 import org.eclipse.hono.client.ClientErrorException;
 import org.eclipse.hono.client.Command;
 import org.eclipse.hono.client.CommandContext;
+import org.eclipse.hono.client.CommandResponse;
 import org.eclipse.hono.client.DownstreamSender;
 import org.eclipse.hono.client.MessageConsumer;
 import org.eclipse.hono.client.ResourceConflictException;
@@ -1115,4 +1116,94 @@ public abstract class AbstractVertxBasedCoapAdapter<T extends CoapAdapterPropert
 
         return Future.succeededFuture(result);
     }
+
+    /**
+     * Uploads a command response message to Hono.
+     *
+     * @param context The context representing the request to be processed.
+     * @param authenticatedDevice authenticated device
+     * @param device message's origin device
+     * @return A succeeded future containing the CoAP status code that has been returned to the device.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     */
+    public final Future<ResponseCode> uploadCommandResponseMessage(final CoapContext context, final Device authenticatedDevice,
+                final Device device) {
+
+        Objects.requireNonNull(context);
+        Objects.requireNonNull(authenticatedDevice);
+        Objects.requireNonNull(device);
+
+        final Buffer payload = context.getPayload();
+        final String contentType = MediaTypeRegistry.toString(context.getExchange().getRequestOptions().getContentFormat());
+        final String commandRequestId = context.getCommandRequestId();
+        final Integer responseStatus = context.getCommandResponseStatus();
+        log.debug("processing response to command [tenantId: {}, deviceId: {}, cmd-req-id: {}, status code: {}]",
+                device.getTenantId(), device.getDeviceId(), commandRequestId, responseStatus);
+
+        final Span currentSpan = TracingHelper
+                .buildChildSpan(tracer, context.getTracingContext(), "upload Command response")
+                .ignoreActiveSpan()
+                .withTag(Tags.COMPONENT.getKey(), getTypeName())
+                .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_CLIENT)
+                .withTag(MessageHelper.APP_PROPERTY_TENANT_ID, device.getTenantId())
+                .withTag(MessageHelper.APP_PROPERTY_DEVICE_ID, device.getDeviceId())
+                .withTag(Constants.HEADER_COMMAND_RESPONSE_STATUS, responseStatus)
+                .withTag(Constants.HEADER_COMMAND_REQUEST_ID, commandRequestId)
+                .withTag(TracingHelper.TAG_AUTHENTICATED.getKey(), authenticatedDevice != null)
+                .start();
+
+        final CommandResponse cmdResponseOrNull = CommandResponse.from(commandRequestId, device.getTenantId(), device.getDeviceId(), payload,
+                contentType, responseStatus);
+        final Future<TenantObject> tenantTracker = getTenantConfiguration(device.getTenantId(), currentSpan.context());
+        final Future<CommandResponse> commandResponseTracker = cmdResponseOrNull != null
+                ? Future.succeededFuture(cmdResponseOrNull)
+                : Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST,
+                        String.format("command-request-id [%s] or status code [%s] is missing/invalid",
+                                commandRequestId, responseStatus)));
+
+        return CompositeFuture.all(tenantTracker, commandResponseTracker)
+                .compose(commandResponse -> {
+                    final Future<JsonObject> deviceRegistrationTracker = getRegistrationAssertion(
+                            device.getTenantId(),
+                            device.getDeviceId(),
+                            authenticatedDevice,
+                            currentSpan.context());
+                    final Future<Void> tenantValidationTracker = CompositeFuture
+                            .all(isAdapterEnabled(tenantTracker.result()),
+                                    checkMessageLimit(tenantTracker.result(), payload.length(), currentSpan.context()))
+                            .map(ok -> null);
+
+                    return CompositeFuture.all(tenantValidationTracker, deviceRegistrationTracker)
+                            .compose(ok -> sendCommandResponse(device.getTenantId(), commandResponseTracker.result(), currentSpan.context()))
+                            .map(delivery -> {
+                                log.trace("delivered command response [command-request-id: {}] to application",
+                                        commandRequestId);
+                                currentSpan.log("delivered command response to application");
+                                currentSpan.finish();
+                                metrics.reportCommand(
+                                        Direction.RESPONSE,
+                                        device.getTenantId(),
+                                        tenantTracker.result(),
+                                        ProcessingOutcome.FORWARDED,
+                                        payload.length(),
+                                        context.getTimer());
+                                context.respondWithCode(ResponseCode.CHANGED);
+                                return ResponseCode.CHANGED;
+                            });
+                }).otherwise(t -> {
+                    log.debug("could not send command response [command-request-id: {}] to application",
+                            commandRequestId, t);
+                    TracingHelper.logError(currentSpan, t);
+                    currentSpan.finish();
+                    metrics.reportCommand(
+                            Direction.RESPONSE,
+                            device.getTenantId(),
+                            tenantTracker.result(),
+                            ProcessingOutcome.from(t),
+                            payload.length(),
+                            context.getTimer());
+                    return CoapErrorResponse.respond(context.getExchange(), t);
+                });
+    }
+
 }
--- a/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
+++ b/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
@@ -24,13 +24,17 @@ import java.util.HashSet;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
+import org.apache.qpid.proton.amqp.transport.ErrorCondition;
 import org.apache.qpid.proton.message.Message;
 import org.eclipse.californium.core.CoapServer;
 import org.eclipse.californium.core.coap.CoAP;
 import org.eclipse.californium.core.coap.CoAP.ResponseCode;
 import org.eclipse.californium.core.coap.MediaTypeRegistry;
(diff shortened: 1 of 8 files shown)


boaks:
It would be great, if I could get some review feedback while I continue to work on some integration tests for this feature.

sophokles73:
@boaks not sure if I find the time this week as I am very busy preparing for BCX

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
```diff
@@ -645,34 +666,91 @@ protected void preShutdown() {
...
+                    .compose(ok -> {
+                        final Integer ttdParam = getTimeUntilDisconnectFromRequest(context);
+                        return getTimeUntilDisconnect(tenantTracker.result(), ttdParam).map(effectiveTtd -> {
+                            if (effectiveTtd != null) {
+                                currentSpan.setTag(MessageHelper.APP_PROPERTY_DEVICE_TTD, effectiveTtd);
+                            }
+                            return effectiveTtd;
+                        });
+                    });
+            final Future<MessageConsumer> commandConsumerTracker = ttdTracker
```
calohmn: It looks like the `commandConsumerTracker` is not used below in the sense that it is `compose`d with other Futures, to handle command consumer creation failures.

If I understand the intention here correctly, the command and control handling in `doUploadMessage` is similar to the one in the HTTP adapter, meaning that when the returned Future of  `doUploadMessage`  is finished, the created command consumer is to be closed again already, right?
The line below:
`Optional.ofNullable(commandConsumerTracker.result()).ifPresent(consumer -> consumer.close(null));`
means: If the command consumer creation was successfully finished, close the consumer.
But what if at that point in time, the consumer creation wasn't finished **yet**. When the creation succeeds at a later point, the consumer will never get closed.

Therefore the code here has to be adapted to make sure that a successfully created command consumer is closed under all circumstances. That means, making sure that some logic is done *after* the `commandConsumerTracker` is completed (using `commandConsumerTracker.compose` or `CompositeFuture.join` as done [here](https://github.com/eclipse/hono/blob/bbacfb9d8797e9e3ba95dcc75166ce1b7647fd23/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java#L697)).

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
```diff
@@ -715,4 +796,348 @@ protected void preShutdown() {
...
+     *
+     * @param ctx The device's currently executing HTTP request.
+     * @param responseReady The future to complete when the time has expired.
+     * @param delaySecs The number of seconds to wait for a command.
+     */
+    private void addCommandReceptionTimer(
+            final CoapContext context,
+            final AtomicBoolean requestProcessed,
+            final Handler<AsyncResult<Void>> responseReady,
+            final long delaySecs) {
```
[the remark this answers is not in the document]
boaks: > missing requestProcessed param javadoc

Uups, that's a copy & paste error, it's missing in the http adapter as well.

Fixed, both.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
```diff
@@ -595,6 +611,7 @@ protected void preShutdown() {
      * follows:
      * <ul>
      * <li>2.04 (Changed) - if the message has been forwarded downstream.</li>
+     * <li>2.05 (Content) - if the message has been answer with a command.</li>
```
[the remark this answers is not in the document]
boaks: fixed

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
boaks: Hopefully fixed :-).

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
```diff
@@ -633,46 +650,105 @@ protected void preShutdown() {
...
+                    && !device.getDeviceId().equals(authenticatedDevice.getDeviceId())
+                            ? authenticatedDevice.getDeviceId()
+                            : null;
 
             final Span currentSpan = TracingHelper
                     .buildChildSpan(tracer, context.getTracingContext(),
                             "upload " + endpoint.getCanonicalName())
                     .ignoreActiveSpan()
-                    .withTag(Tags.COMPONENT.getKey(), getTypeName())
-                    .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_CLIENT)
```
sophokles73: is there a particular reason why you want to remove these tags?

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
```diff
@@ -715,4 +794,348 @@ protected void preShutdown() {
...
+     * @param context The CoAP context containing the request.
+     * @return The time till disconnect or {@code null} if
+     *         <ul>
+     *         <li>the request doesn't contain a {@link org.eclipse.hono.util.Constants#HEADER_TIME_TILL_DISCONNECT}
+     *         query parameter.</li>
+     *         <li>the contained value cannot be parsed as an Integer</li>
+     *         </ul>
+     * @throws NullPointerException if context is {@code null}.
+     */
+    protected Integer getTimeUntilDisconnectFromRequest(final CoapContext context) {
```
sophokles73: how about moving this method to `CoapContext`?

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
boaks: They are already included in the parent spawn. 
I recognized, that the "kind" changes from "server" to "client". And coap uses "http-uri".
So my impression was, that the tags are not consequent maintained. Therefore I adjusted them in the was, I thought it is useful.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
boaks: Possible.

boaks: Done.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
sophokles73: The *kind* changes from `server` to `client` because this new span is used to track the protocol adapter's (client) request to the AMQP Messaging Network to forward the message to a downstream consumer. The parent span has been created in response to a device's request to upload data (hence `server`).
The tags are not inherited, thus it makes sense to set them on each new span so that they can easily be found when searching for tag values. FMPOV you should therefore revert this change ...

boaks: done.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapContext.java
```diff
@@ -94,4 +108,109 @@ public Sample getTimer() {
...
+                return pathList.get(pathList.size() - 1);
+            }
+        }
+        return null;
+    }
+
+    /**
+     * Get command response status of response for command.
+     * 
+     * @return status, or {@link null}, if not available.
```
calohmn: `{@code null}` instead of `link`

boaks: Uups .. you're right!

boaks:
Beside of the missing integrations test (I start now working on that), I got a mystic message from the cli-client, profile command:

> org.eclipse.hono.cli.app.Commander - Error sending command: status code 201 does not represent an error

Is that usual? Or does it show, that something is not proper?

Testing a oneway command, I see two irritating entries under the 
"hono-adapter-coap-vertx.delegate and send command": 

"hono-service-device-registry.process request message"
and
"hono-service-device-registry.assert Device Registration"

Is that intended?Or does this indicate something unusual?

calohmn:
> Testing a oneway command, I see two irritating entries in Jaeger under the
"hono-adapter-coap-vertx.delegate and send command":
"hono-service-device-registry.process request message"
and
"hono-service-device-registry.assert Device Registration"
Is that intended?Or does this indicate something unusual?

Yes, that is normal. When processing an incoming request, the device data is checked as part of the logic to map the command to a gateway if needed.

boaks:
I got a first impression about the command and control integration test of http. 
That test implementation requires much more understanding of hono/vert.x internals. That will take quite long ...

sophokles73:
@boaks I think we should add the ITs in another PR. The scope of this PR is big enough FMPOV.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
```diff
@@ -1112,4 +1096,95 @@ private TtdStatus getTtdStatus(final CoapContext context) {
...
+                    log.debug("could not send command response [command-request-id: {}] to application",
+                            commandRequestId, t);
+                    TracingHelper.logError(currentSpan, t);
+                    currentSpan.finish();
+                    metrics.reportCommand(
+                            Direction.RESPONSE,
+                            device.getTenantId(),
+                            tenantTracker.result(),
+                            ProcessingOutcome.from(t),
+                            payload.length(),
```
sophokles73: payloadSize?

boaks:
> I think we should add the ITs in another PR. The scope of this PR is big enough FMPOV.

If you're OK with that approach, I will go for an additional PR.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
boaks: Here it seems to be cosmetic, payload will be always provided but may be `null`.
Anyway, if will remove the `payloadSize` and use always the `payload.length()`, though it's very straight forward, when payload is never `null`.
Just to mention: the "payload.length()" is copied form http, where payload may be `null`. so it may cause troubles there.

boaks:
I adapted the coap-response with the command to use the "location" to pass the path for a optional further coap-request as response to the command. Makes it easier for clients, because they just need to send the second request using that location as path.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DefaultDeviceResolver.java
```diff
@@ -103,6 +116,13 @@ private static SecretKey getKey(final JsonObject candidateSecret) {
         }
     }
 
+    private Span newSpan(final String operation) {
+        return tracer.buildSpan(operation)
+                .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_SERVER)
```
sophokles73: if these spans are suppose to be used for tracking the retrieval of the PSK from the device registry, then FMPOV the kind is `SPAN_KIND_CLIENT`.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DefaultDeviceResolver.java
```diff
@@ -111,12 +131,15 @@ public AdditionalInfo getInfo(final Principal clientIdentity) {
         final Map<String, Object> result = new HashMap<>();
 
         if (clientIdentity instanceof PreSharedKeyIdentity) {
+            final Span span = newSpan("PSK-INFO");
```
sophokles73: IMHO the operation should be a little more expressive, e.g. *get device identity info* or something like that?

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DefaultDeviceResolver.java
```diff
@@ -142,16 +167,20 @@ public AdditionalInfo getInfo(final Principal clientIdentity) {
 
     @Override
     public SecretKey getKey(final PskPublicInformation identity) {
+        final Span span = newSpan("PSK-KEY");
```
sophokles73: more expressive name?

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DefaultDeviceResolver.java
```diff
@@ -160,13 +189,18 @@ public SecretKey getKey(final PskPublicInformation identity) {
                 }
             });
         });
+        SecretKey key;
         try {
             // credentials client will wait limited time only
-            return secret.join();
+            key = secret.join();
+            span.setTag("secret", true);
```
sophokles73: are you sure that this should be set as a tag instead of just being logged to the span?
if this indeed needs to be a tag then I'd rather have a constant for the Tag or at least its name ...

boaks: I have no glue, what the difference of a tag or logged should be :-).

sophokles73: A tag on a Span is a searchable *label*, if you will. You can use it to find Spans that have a particular tag value. A Span's log is not searchable per se and is usually only shown, once a Span of interest has been identified. From what I see in the code, the outcome of retrieving the secret should either be logged to the Span (`span.log()`) or should be reflected by means of the standard `http.status_code` tag, e.g. setting it to 200 or 404 ...

boaks: OK, I will adapt it to "log".

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DefaultDeviceResolver.java
boaks: FMPOV, two spans are used for that. The outer one describes the incoming dtls handshake message (dtls server), the inner, the "getCredentials" call (amqp client). So, aligned to the POST stuff, I chose "server" for the outer and "client" for the inner.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DefaultDeviceResolver.java
boaks: Adapted to "PSK-getDeviceIdentityInfo"

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DefaultDeviceResolver.java
boaks: Adapted to "PSK-getSecretKey"

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DefaultDeviceResolver.java
sophokles73: Can you point me to the place where the outer Span for tracing the DTLS handshake is created? I am afraid that I couldn't find it.

boaks: FMPOV that are the callbacks 

```
@Override
public SecretKey getKey(final PskPublicInformation identity)
 
@Override
public AdditionalInfo getInfo(final Principal clientIdentity)
```

boaks:
I adapted the coap-response and cmd&ctrl consumer closing according the http implementation.
With that, the also newly introduced IT for coap cmd&ctrl (PR #1795) works reliable.

boaks:
Now with updated dependency to Californium 2.1.0

Issue #1761, opened 2020-02-11
Client for the AMQP adapter


b-abel:
I think it would be beneficial to provide a ( southbound) client implementation for the AMQP protocol adapter. Similar to the `ApplicationClientFactory` it should provide a lean API that can be easily used by _users_ (in the sense of the User Guide) of Hono without having to bother with types that are mainly useful for internal use of Hono. This client could be used for easy interaction with the AMQP adapter and to implement protocol gateways (#1646).

sophokles73:
sounds like a great idea :+1:

b-abel:
I can provide a PR for this.

b-abel:
I added a PR based on existing code from the `client` module for this. I think it's good to integrate this now to have something usable for the time being.
Perspectively, it would be good to have a simple client interface that does not expose how the AMQP links are used (e.g. by separate clients to be retrieved from the factory, by disclosing the caching of the links and by creating clients per tenant). I would like to provide this later, and this could then replace the solution in this first PR.

sophokles73:
@b-abel what is the status of this PR? Are you going to add docs and tests or is this supposed to be merged as-is?

PR #1765, opened 2020-02-12
[#1761] Add AMQP adapter client.

b-abel:
This is the first solution for #1761, which on the one hand uses existing code from the client module and on the other hand aims to expose as few classes as possible to the outside world that are designed for internal use in the protocol adapters. 

The entry point is the factory interface `AmqpAdapterClientFactory`, which can be used to create clients.
I had to put some classes into the existing `client.impl` package because of the visibility of the code used, the rest I tried to separate a little bit.
Tests (and documentation) will be added once we have agreed on the overall direction.

It is a quick and easy solution, in a next step it would be good to have a simpler client interface that does not expose the link handling and scopes the connection to the tenant, but provides simple *send* methods. I hope to do another PR on this soon. For that, I would make a separate Maven module with minimal dependencies.


b-abel:
Added tracing and inherit from classes used inside the protocol adapters.

client/src/main/java/org/eclipse/hono/client/device/amqp/AmqpAdapterClientFactory.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    Future<EventSender> getOrCreateEventSender();
+
+    /**
+     * Creates a client for consuming commands from Hono's AMQP protocol adapter for a specific device.
+     * <p>
+     * When implementing a (protocol) gateway, this can be used to receive commands for a specific device.
+     * <p>
+     * The command passed in to the command consumer will be settled automatically if the consumer does not throw an
+     * exception.
```
calohmn: I guess `if the messageHandler does not throw an exception` is meant here. However, in that case the command message would also be settled (see https://github.com/eclipse/hono/blob/525b3aa4ccabbcfbc1c97eef53684efb5d7934da/client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java#L823). In general, the message handler is not really supposed to throw an exception, though.
Therefore I think the `if` part of the sentenced can be removed here.

client/src/main/java/org/eclipse/hono/client/device/amqp/AmqpAdapterClientFactory.java
```diff
@@ -0,0 +1,124 @@
...
+     * @see "https://www.eclipse.org/hono/docs/dev/user-guide/amqp-adapter/"
+     */
+    Future<MessageConsumer> createDeviceSpecificCommandConsumer(String deviceId, Consumer<Message> messageHandler);
+
+    /**
+     * Creates a client for consuming commands from Hono's AMQP protocol adapter for an authenticated device or on a
+     * (protocol) gateway for all devices on whose behalf it acts.
+     * <p>
+     * The command passed in to the command consumer will be settled automatically if the consumer does not throw an
+     * exception.
```
calohmn: IMHO, the `if` part of the sentenced can be removed here (see above).

client/src/main/java/org/eclipse/hono/client/device/amqp/CommandResponder.java
```diff
@@ -0,0 +1,54 @@
...
+import org.eclipse.hono.client.ServiceInvocationException;
+import org.eclipse.hono.client.device.amqp.internal.AmqpSenderLink;
+
+import io.vertx.core.Future;
+import io.vertx.proton.ProtonDelivery;
+
+/**
+ * A sender to send back the response message of a command to Hono's AMQP adapter.
+ */
+public interface CommandResponder extends AmqpSenderLink {
```
calohmn: IMHO, `CommandResponseSender` would make it more obvious at first glance, that there is a `sender` link behind this class.

client/src/main/java/org/eclipse/hono/client/device/amqp/TraceableCommandResponder.java
```diff
@@ -0,0 +1,55 @@
...
+import org.eclipse.hono.client.ServiceInvocationException;
+
+import io.opentracing.SpanContext;
+import io.vertx.core.Future;
+import io.vertx.proton.ProtonDelivery;
+
+/**
+ * A sender to send back the response message of a command to Hono's AMQP adapter.
+ */
+public interface TraceableCommandResponder extends CommandResponder {
```
calohmn: I would suggest `TraceableCommandResponseSender` as name here instead (see above).

client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientCommandResponseSender.java
```diff
@@ -0,0 +1,131 @@
...
+     * 
+     * @param commandResponse not used.
+     * @param context not used.
+     * @return nothing.
+     * @throws UnsupportedOperationException always.
+     */
+    @Override
+    public Future<ProtonDelivery> sendCommandResponse(final CommandResponse commandResponse,
+            final SpanContext context) {
+        throw new UnsupportedOperationException("This method must not be used outside a protocol adapter");
```
calohmn: I think the exception text and the method javadoc is confusing here - if the class and method is used inside a protocol adapter, there is still this exception.
I guess the javadoc could state that the _overridden method_ is only meant for use by a protocol adapter to send the command response downstream, and that this use case isn't supported for this class.

client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientCommandResponseSender.java
```diff
@@ -0,0 +1,131 @@
...
+     * @param closeHook not used.
+     * @return nothing.
+     * @throws UnsupportedOperationException always.
+     */
+    public static Future<CommandResponseSender> create(
+            final HonoConnection con,
+            final String tenantId,
+            final String replyId,
+            final Handler<String> closeHook) {
+        throw new UnsupportedOperationException("This method must not be used outside a protocol adapter");
```
calohmn: IMHO, shadowing the (unsupported) parent static factory method with what is now in `createWithAnonymousLinkAddress` would be a better idea. The unused `replyId` parameter could be ignored and another `create` method could be added without the `replyId` parameter.

client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientEventSenderImpl.java
```diff
@@ -0,0 +1,103 @@
...
+     * @param tenantId not used.
+     * @param remoteCloseHook not used.
+     * @return nothing.
+     * @throws UnsupportedOperationException always.
+     */
+    public static Future<DownstreamSender> create(
+            final HonoConnection con,
+            final String tenantId,
+            final Handler<String> remoteCloseHook) {
+        throw new UnsupportedOperationException("This method must not be used outside a protocol adapter");
```
calohmn: IMHO, shadowing the (unsupported) parent static factory method with what is now in `createWithAnonymousLinkAddress` would be a better idea (i.e. renaming `createWithAnonymousLinkAddress` to just `create`). WDYT?

client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientTelemetrySenderImpl.java
```diff
@@ -0,0 +1,123 @@
...
+     * @param tenantId not used.
+     * @param remoteCloseHook not used.
+     * @return nothing.
+     * @throws UnsupportedOperationException always.
+     */
+    public static Future<DownstreamSender> create(
+            final HonoConnection con,
+            final String tenantId,
+            final Handler<String> remoteCloseHook) {
+        throw new UnsupportedOperationException("This method must not be used outside a protocol adapter");
```
calohmn: IMHO, shadowing the (unsupported) parent static factory method with what is now in `createWithAnonymousLinkAddress` would be a better idea (i.e. renaming `createWithAnonymousLinkAddress` to just `create`). WDYT?

client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientCommandResponseSender.java
```diff
@@ -0,0 +1,131 @@
...
+    @Override
+    public Future<ProtonDelivery> sendCommandResponse(final String deviceId, final String targetAddress,
+            final String correlationId, final int status, final byte[] payload,
+            final String contentType, final Map<String, ?> properties, final SpanContext context) {
+
+        Objects.requireNonNull(deviceId);
+        Objects.requireNonNull(targetAddress);
+        Objects.requireNonNull(correlationId);
+
+        final Message message = ProtonHelper.message();
```
calohmn: I think we should also set a message id here.

client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientEventSenderImpl.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    public Future<ProtonDelivery> send(final String deviceId, final byte[] payload, final String contentType,
+            final Map<String, ?> properties, final SpanContext context) {
+        Objects.requireNonNull(deviceId);
+        Objects.requireNonNull(payload);
+        Objects.requireNonNull(contentType);
+
+        final Message msg = ProtonHelper.message();
```
calohmn: I think we should also set a message id here.

client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientTelemetrySenderImpl.java
```diff
@@ -0,0 +1,123 @@
...
+        return sendAndWaitForOutcome(msg, context);
+    }
+
+    private Message createMessage(final String deviceId, final byte[] payload, final String contentType,
+            final Map<String, ?> properties) {
+        Objects.requireNonNull(deviceId);
+        Objects.requireNonNull(payload);
+        Objects.requireNonNull(contentType);
+
+        final Message msg = ProtonHelper.message();
```
calohmn: I think we should also set a message id here.

client/src/main/java/org/eclipse/hono/client/device/amqp/internal/AmqpAdapterClientCommandConsumer.java
```diff
@@ -0,0 +1,177 @@
...
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.client.device.amqp.internal;
```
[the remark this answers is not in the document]
b-abel: That's right. As mentioned above, I tried to pack the new classes into better-named packages (also to prepare for splitting the clients later). Unfortunately, I could not move some classes because of the visibility of the code used there. If we give higher priority to putting them together in the same package, then we will probably have to put everything back into the `client` package.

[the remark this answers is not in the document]
b-abel: Do you think it should be moved to `client.impl`?

client/src/main/java/org/eclipse/hono/client/device/amqp/CommandResponder.java
b-abel: Yes, that is more obvious. This already exists in the client package. That's why I hesitated to use the same name because it might be too difficult to distinguish between them just by the package.

client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientCommandResponseSender.java
b-abel: That does not work because of incompatible return types.

client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientCommandResponseSender.java
b-abel: This is done further down in `AbstractSender#sendMessageAndWaitForOutcome(Message, Span)`.

client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientTelemetrySenderImpl.java
b-abel: This is done further down in `AbstractSender#sendMessageAndWaitForOutcome(Message, Span)`. And for QoS 0 in `TelemetrySenderImpl#sendMessage(Message, Span)`.

client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientEventSenderImpl.java
b-abel: This is done further down in `AbstractSender#sendMessageAndWaitForOutcome(Message, Span)`.

client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientCommandResponseSender.java
calohmn: Ok, right, I've overlooked that.

client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientCommandResponseSender.java
calohmn: Ok, I see.

b-abel:
@calohmn What is the status of this issue?

client/src/main/java/org/eclipse/hono/client/device/amqp/CommandResponder.java
calohmn: Ok, I see.

client/src/main/java/org/eclipse/hono/client/device/amqp/internal/AmqpAdapterClientCommandConsumer.java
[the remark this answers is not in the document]
calohmn: I think the current structure is OK. Probably packages will later be reorganized anyway when splitting up the hono client module.

calohmn:
@b-abel Can you squash and rebase (also to check with the new checkstyle rules introduced since) ?

b-abel:
Done.

client/src/main/java/org/eclipse/hono/client/device/amqp/AmqpAdapterClientFactory.java
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * A Vert.x based factory for creating clients for Hono's AMQP adapter.
+ */
+public interface AmqpAdapterClientFactory extends ConnectionLifecycle<HonoConnection> {
+
+    /**
+     * Creates a new factory for an existing connection.
+     *
+     * @param connection The connection to use.
+     * @param tenantId The ID of the tenant for which the connection is authenticated.
```
sophokles73: I am not sure if I understand the purpose of this parameter. My understanding is that the factory will be scoped to the tenant that the auth-id resolves to during authentication of the underlying AMQP connection, or am I mistaken?

client/src/main/java/org/eclipse/hono/client/device/amqp/TraceableCommandResponder.java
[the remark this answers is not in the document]
b-abel: I'm not quite sure what the question is aiming at. Hence the long answer. 
Why interfaces? 
The client provides an API that Hono users interact with directly and must therefore remain stable. In our efforts to make changes after the 1.0 release without breaking the API, I have found that this was very difficult, where we had no interfaces and worked directly with the implementations (not really surprising). So I made sure that the user only works with interfaces as much as possible.
Why is tracing not included in the interfaces?
First of all, I designed the interfaces for the AMQP adapter client without tracing, because it should be as simple and generally understandable as possible. In the typical use cases, no distributed tracing is needed because it only makes sense if the client and the Hono installation are operated by the same party. In general, I think that overloaded methods in APIs of libraries should be an exception.  However, Carsten asked me to provide tracing support anyway, because it can be helpful in some cases (e.g. for end-to-end testing of a Hono installation). I have hidden these a bit and would document them accordingly (I still have to do that. In Readme or developer guide).

client/src/main/java/org/eclipse/hono/client/device/amqp/AmqpAdapterClientFactory.java
b-abel: Unfortunately, the parameter is expected by the parent classes (the constructor in `AbstractSender` requires the tenantId to be not `null`). Of course, this is ugly, because the tenant must match the one used for authentication.
Do you have any ideas on how to get rid of it? Set a dummy or empty string? (would probably only have negative effects for tracing). Read the Auth-Id from the connection and parse it? (wouldn't work for certificate-based authentication). Change the base classes? ...

sophokles73: I see

client/src/main/java/org/eclipse/hono/client/device/amqp/TraceableCommandResponder.java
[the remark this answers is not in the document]
sophokles73: I see

sophokles73:
@b-abel are you going to integrate this new client into the protocol gateway example code?

b-abel:
> @b-abel are you going to integrate this new client into the protocol gateway example code?

Yes.

b-abel:
I would add tests and documentation to this PR. Or do you prefer to open a new one for that?

sophokles73:
@b-abel any way you want it ... as long as you **do** create docs and tests ;-)

sophokles73:
@b-abel so, what's it gonna be? This PR or a separate one?

b-abel:
I will add them in this PR.

sophokles73:
@b-abel what is the status of this PR? Are you still working on docs and tests?

b-abel:
> @b-abel what is the status of this PR? Are you still working on docs and tests?

Yes.

b-abel:
> @b-abel what is the status of this PR? Are you still working on docs and tests?

Done.

client/src/test/java/org/eclipse/hono/client/device/amqp/TelemetrySenderTest.java
```diff
@@ -0,0 +1,80 @@
...
+    @Test
+    public void testSendAndWaitForOutcomeProducesValidMessage() {
+
+        // GIVEN a TelemetrySender instance
+        final TelemetrySender telemetrySender = AmqpAdapterClientTelemetrySenderImpl
+                .createWithAnonymousLinkAddress(connection, tenantId, s -> {
+                }).result();
+
+        // WHEN sending a message using the API
+        final Future<ProtonDelivery> deliveryFuture = telemetrySender.sendAndWaitForOutcome(deviceId, payload,
```
sophokles73: FMPOV this test should at least also verify that the TelemetrySender waits for the disposition update from the peer ...

client/src/test/java/org/eclipse/hono/client/device/amqp/EventSenderTest.java
```diff
@@ -0,0 +1,53 @@
...
+    @Test
+    public void testMessageIsValid() {
+
+        // GIVEN a EventSender instance
+        final EventSender eventSender = AmqpAdapterClientEventSenderImpl
+                .createWithAnonymousLinkAddress(connection, tenantId, s -> {
+                }).result();
+
+        // WHEN sending a message using the API
+        final Future<ProtonDelivery> deliveryFuture = eventSender.send(deviceId, payload, contentType,
```
sophokles73: FMPOV this test should at also verify that the EventSender waits for the disposition update from the peer ...

client/src/test/java/org/eclipse/hono/client/device/amqp/TraceableCommandResponderTest.java
```diff
@@ -0,0 +1,56 @@
...
+        // GIVEN a CommandResponder instance
+        final TraceableCommandResponder commandResponder = ((TraceableCommandResponder) AmqpAdapterClientCommandResponseSender
+                .createWithAnonymousLinkAddress(connection, tenantId, s -> {
+                }).result());
+
+        // WHEN sending a message using the API
+        final String targetAddress = "command_response/test-tenant/test-device/123";
+        final String correlationId = "0";
+        final int status = 200;
+        final Future<ProtonDelivery> deliveryFuture = commandResponder.sendCommandResponse(deviceId, targetAddress,
```
sophokles73: FMPOV this test should also verify that the CommandResponder waits for the disposition update from the peer ...

client/src/test/java/org/eclipse/hono/client/device/amqp/TraceableCommandResponderTest.java
```diff
@@ -0,0 +1,56 @@
...
+                correlationId, status, payload, contentType,
+                applicationProperties, null);
```
sophokles73: If this is the test for the *traceable* version then I would expect to find some assertion(s) regarding the tracing actually happening ...

site/documentation/content/dev-guide/amqp_adapter_client.md
```diff
@@ -0,0 +1,113 @@
...
+The senders manage the underlying AMQP sender links, which are restored after a temporary interruption of the connection. 
+('HonoConnection' handles automatic reconnection). To achieve this, a caching mechanism is used, so that 
+ defective links are replaced by new ones. It is important that the sender is always retrieved from the factory before sending.
+ Do not keep links to sender objects and reuse them for subsequent send operations!
+{{% /note %}}
+
+
+## Create a Client Instance 
+
+```
```
sophokles73: please do not put code into the docs. We will never be able to keep this in sync with changes that we make to the code itself. Please refer to example classes or test cases in the code base instead.

b-abel:

client/src/test/java/org/eclipse/hono/client/AbstractAmqpAdapterClientDownstreamSenderTestBase.java
```diff
@@ -38,50 +46,82 @@
...
-        assertThat(MessageHelper.getPayloadAsString(message)).isEqualTo(new String(payload));
-        assertThat(message.getContentType()).isEqualTo(contentType);
+        assertThat(MessageHelper.getPayloadAsString(message)).isEqualTo(new String(PAYLOAD));
+        assertThat(message.getContentType()).isEqualTo(CONTENT_TYPE);
+
+        assertThat(message.getApplicationProperties().getValue().get(TEST_PROPERTY_KEY)).isEqualTo(TEST_PROPERTY_VALUE);
 
-        final Map<String, Object> applicationProperties = message.getApplicationProperties().getValue();
-        assertThat(applicationProperties.get(testPropertyKey)).isEqualTo(testPropertyValue);
+        assertThat(message.getApplicationProperties().getValue().get("device_id")).isEqualTo(DEVICE_ID);
```
sophokles73: `MessageHelper.getDeviceId(Message)`

client/src/test/java/org/eclipse/hono/client/AbstractAmqpAdapterClientDownstreamSenderTestBase.java
```diff
@@ -38,50 +46,82 @@
...
         final Message message = messageArgumentCaptor.getValue();
 
         assertThat(message.getAddress()).isEqualTo(expectedAddress);
 
-        assertThat(MessageHelper.getPayloadAsString(message)).isEqualTo(new String(payload));
-        assertThat(message.getContentType()).isEqualTo(contentType);
+        assertThat(MessageHelper.getPayloadAsString(message)).isEqualTo(new String(PAYLOAD));
+        assertThat(message.getContentType()).isEqualTo(CONTENT_TYPE);
+
+        assertThat(message.getApplicationProperties().getValue().get(TEST_PROPERTY_KEY)).isEqualTo(TEST_PROPERTY_VALUE);
```
sophokles73: `MessageHelper.getApplicationProperty()`?

client/src/test/java/org/eclipse/hono/client/device/amqp/CommandResponderTest.java
```diff
@@ -14,43 +14,120 @@
...
 import io.vertx.proton.ProtonDelivery;
 
 /**
  * Verifies behavior of {@link CommandResponder}.
  *
  */
+@ExtendWith(VertxExtension.class)
 public class CommandResponderTest extends AbstractAmqpAdapterClientDownstreamSenderTestBase {
 
+    private static final String ADDRESS = "command_response/" + TENANT_ID + "/" + DEVICE_ID + "/123";
```
sophokles73: `CommandConstants.COMMAND_RESPONSE_ENDPOINT`

client/src/test/java/org/eclipse/hono/client/device/amqp/CommandResponderTest.java
```diff
@@ -14,43 +14,120 @@
...
+        final Future<ProtonDelivery> deliveryFuture = commandResponder.sendCommandResponse(DEVICE_ID, ADDRESS,
+                CORRELATION_ID, STATUS, PAYLOAD, CONTENT_TYPE, APPLICATION_PROPERTIES);
 
-        // THEN the AMQP message produces by the client conforms to the expectations of the AMQP protocol adapter
-        assertThat(deliveryFuture.succeeded());
+        deliveryFuture.setHandler(ctx.completing());
 
-        assertMessageConformsAmqpAdapterSpec(targetAddress);
+        // THEN the future waits for the disposition to be updated by the peer
+        Thread.sleep(100L);
```
sophokles73: ???

b-abel:
I just pushed an update.

client/src/test/java/org/eclipse/hono/client/device/amqp/TelemetrySenderTest.java
```diff
@@ -0,0 +1,167 @@
...
+        final TelemetrySender telemetrySender = createTelemetrySender();
+
+        // WHEN sending a message using the API
+        final Future<ProtonDelivery> deliveryFuture = telemetrySender.sendAndWaitForOutcome(DEVICE_ID, PAYLOAD,
+                CONTENT_TYPE, APPLICATION_PROPERTIES);
+
+        deliveryFuture.setHandler(ctx.completing());
+
+        // THEN the future waits for the disposition to be updated by the peer
+        Thread.sleep(100L);
```
sophokles73: can you please make sure to remove all occurrences of this?

b-abel: Sorry, I missed them.

Issue #1736, opened 2020-01-31
Align Tenant management API implementation with the spec


dejanb:
At the moment Tenant management API specification defines that payload (without any mandatory parameters) is required during tenant creation. The implementation at the moment accepts empty body and adds 'enabled' field in that case. Empty body is also used in getting started examples.

We need to align specification with the implementation and examples.

sophokles73:
FMPOV there is no need to supply and data in a request payload if I want to use the defaults. SO I would say that the spec should be *relaxed* accordingly.

sophokles73:
@calohmn @ctron WDYT?

calohmn:
I also think that we can change the spec to make the request body optional.

dejanb:
I agree as well.

One question that I have is that with this change and changes in the #1711, should we bump the spec version?

sophokles73:
> One question that I have is that with this change and changes in the #1711, should we bump the spec version?

I think we should. Remember that this is public API and we need to apply semantic versioning.
The change introduced in #1711 seems to only be concerning the wording being changed from *should not be included* to *must not be included* and the change in this PR also looks backward compatible to me. So I would suggest to bump to `1.0.1`. WDYT? @dejanb @ctron @calohmn

sophokles73:
@dejanb can this be closed?

sophokles73:
@dejanb can this be closed?

PR #1769, opened 2020-02-12
[#1736] Make the body optional in tenant creation request

jbtrystram:

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/management/tenant/AbstractTenantManagementHttpEndpoint.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/management/tenant/AbstractTenantManagementHttpEndpoint.java
@@ -167,7 +167,7 @@ public abstract class AbstractTenantManagementHttpEndpoint extends AbstractHttpE
         final JsonObject payload = Util.getRequestPayload(ctx.get(KEY_REQUEST_BODY));
 
         if (isValidRequestPayload(payload)) {
-            logger.debug("creating tenant [{}]", Optional.of(tenantId).orElse("<auto>"));
+            logger.debug("creating tenant [{}]", Optional.ofNullable(tenantId).orElse("<auto>"));
 
             addNotPresentFieldsWithDefaultValuesForTenant(payload);
 
@@ -310,11 +310,6 @@ public abstract class AbstractTenantManagementHttpEndpoint extends AbstractHttpE
      *       or the JsonArray contains elements that are not of type {@link JsonObject}.
      */
     protected final void addNotPresentFieldsWithDefaultValuesForTenant(final JsonObject checkedPayload) {
-        if (!checkedPayload.containsKey(RegistryManagementConstants.FIELD_ENABLED)) {
-            logger.trace("adding 'enabled' key to payload");
-            checkedPayload.put(RegistryManagementConstants.FIELD_ENABLED, Boolean.TRUE);
-        }
-
         final JsonArray adapters = checkedPayload.getJsonArray(RegistryManagementConstants.FIELD_ADAPTERS);
         if (adapters != null) {
             adapters.forEach(elem -> addNotPresentFieldsWithDefaultValuesForAdapter((JsonObject) elem));
--- a/site/documentation/content/api/management/device-registry-v1.yaml
+++ b/site/documentation/content/api/management/device-registry-v1.yaml
@@ -36,7 +36,7 @@ info:
    license:
       name: EPL-2.0
       url: https://www.eclipse.org/legal/epl-2.0/
-   version: 1.0.0
+   version: 1.0.1
 
 externalDocs:
    description: Eclipse Hono™ web page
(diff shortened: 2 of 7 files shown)


jbtrystram:
@sophokles73 I updated the spec version. 

After manual testing I fixed the code and added a couple of tests to the systemtests to cover some uncovered use cases. 

You can no add a tenant with : 
`curl -X POST http://localhost:28080/v1/tenants/`

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -86,7 +86,7 @@ paths:
                application/json:
                   schema:
                      $ref: '#/components/schemas/Tenant'
-            required: true
+            required: false
```
dejanb: I think this should be false for the case where tenant id is supplied as well

dejanb: Also, we have the same case for the `devices` endpoint. I would make it consistent between `tenants` and `devices`, but keep it required for the `credentials`. @sophokles73  wdyt?

sophokles73: +1 for keeping it consistent, i.e. do not require a body for devices as well. However, that could go into a separate PR.

sophokles73:
@jbtrystram can you also please check the Device Registry User Guide? IMHO the description there still defines the body for creating a tenant as mandatory ...

site/documentation/content/api/management/device-registry-v1.yaml
dejanb: Sounds good. Let's wrap tenants in this one and tackle devices in the separate one

jbtrystram: @dejanb I have aligned this for when an ID is supplied. I've also updated the `devices` section to match. 
Creating a device is now as simple as : 
`curl -i -X POST http://localhost:28080/v1/devices/DEFAULT_TENANT`

jbtrystram:
@sophokles73 I updated the user guide and added examples for creating tenants and devices using the ID generation feature.

service-base/src/main/java/org/eclipse/hono/service/management/tenant/AbstractTenantManagementHttpEndpoint.java
```diff
@@ -310,11 +310,6 @@ protected final boolean isValidRequestPayload(final JsonObject payload) {
      *       or the JsonArray contains elements that are not of type {@link JsonObject}.
      */
     protected final void addNotPresentFieldsWithDefaultValuesForTenant(final JsonObject checkedPayload) {
-        if (!checkedPayload.containsKey(RegistryManagementConstants.FIELD_ENABLED)) {
-            logger.trace("adding 'enabled' key to payload");
-            checkedPayload.put(RegistryManagementConstants.FIELD_ENABLED, Boolean.TRUE);
-        }
-
```
sophokles73: is there a particular reason why you want to remove this?

tests/src/test/java/org/eclipse/hono/tests/registry/TenantHttpIT.java
```diff
@@ -164,6 +165,38 @@ public void testAddTenantSucceedsForEmptyBody(final VertxTestContext context) {
...
+
+        registry.addTenant("")
+                .compose(res -> {
+                    // compare the changed field only
+                    context.verify(() -> {
+                        assertNotNull(res.get("location"));
+                        // update the global tenantId value for cleanup
+                        tenantId = res.get("location");
+                    });
+        return Future.succeededFuture();
```
sophokles73: this needs to be indented ...

service-base/src/main/java/org/eclipse/hono/service/management/tenant/AbstractTenantManagementHttpEndpoint.java
jbtrystram: This step is already done in `updatePayloadWithTenantId` in the same class.

jbtrystram: Adding the enabled field is already done a few line before, when calling `Util.getRequestPayload(JsonObject)` which already add the field.

dejanb:
@jbtrystram can you squash commits?

tests/src/test/java/org/eclipse/hono/tests/registry/TenantHttpIT.java
```diff
@@ -164,6 +165,38 @@ public void testAddTenantSucceedsForEmptyBody(final VertxTestContext context) {
...
+                .setHandler(context.completing());
+    }
+
+    /**
+     * Verifies that the service successfully create a tenant with a generated tenant ID.
+     *
+     * @param context The vert.x test context.
+     */
+    @Test
+    public void testAddTenantSucceedsWithAGeneratedId(final VertxTestContext context) {
```
sophokles73: since you also allow to create a device without providing a body, shouldn't we also add a corresponding test case to the suite for the `/devices` resources?

jbtrystram: I added a test

PR #1773, opened 2020-02-18
Fix http/lora/sigfox metrics

BobClaerhout:

### DIFF
--- a/adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/impl/Config.java
+++ b/adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/impl/Config.java
@@ -15,9 +15,7 @@ package org.eclipse.hono.adapter.lora.impl;
 
 import javax.annotation.PostConstruct;
 
-import org.eclipse.hono.adapter.http.HttpAdapterMetrics;
 import org.eclipse.hono.adapter.http.HttpProtocolAdapterProperties;
-import org.eclipse.hono.adapter.http.MicrometerBasedHttpAdapterMetrics;
 import org.eclipse.hono.adapter.lora.LoraProtocolAdapterProperties;
 import org.eclipse.hono.client.RequestResponseClientConfigProperties;
 import org.eclipse.hono.config.ClientConfigProperties;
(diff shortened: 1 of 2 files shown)


BobClaerhout:
If a 1.1.1 release will be created to fix this bug. Is it possible to include my other change in the lora adapter? Or is this not allowed?

b-abel:
I suggest removing the `@Autowired` annotation from `MicrometerBasedHttpAdapterMetrics` together with `@Component`. It's no longer needed.

sophokles73:
@BobClaerhout not sure if I understand correctly, but is there an issue with the way it was before?

BobClaerhout:
I had following exception while starting up the lora adapter:
```
Caused by: org.springframework.beans.factory.NoUniqueBeanDefinitionException: No qualifying bean of type 'org.eclipse.hono.adapter.http.HttpAdapterMetrics' available: expected single matching bean but found 2: micrometerBasedHttpAdapterMetrics,adapterMetrics
    at org.springframework.beans.factory.config.DependencyDescriptor.resolveNotUnique(DependencyDescriptor.java:221)
    at org.springframework.beans.factory.support.DefaultListableBeanFactory.doResolveDependency(DefaultListableBeanFactory.java:1233)
    at org.springframework.beans.factory.support.DefaultListableBeanFactory.resolveDependency(DefaultListableBeanFactory.java:1175)
    at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredMethodElement.inject(AutowiredAnnotationBeanPostProcessor.java:670)
    ... 31 common frames omitted
```
So both the lora adapter and the sigfox adapter wheren't able to start.  
If you are talking about the mic, that's a different case. This is not really a bug but would be nice for me to be in the 1.1.1 so I can just deploy the 1.1.1 and do not have to backport that change myself.

sophokles73:
The problem here is that the lora adapter's config class manually instantiates the `HttpAdapterMetrics` class while also scanning `org.eclipse.hono.adapter.http` for components.

Other adapters deriving from the HTTP adapter might still rely on the metrics class to be instantiated by a component scan so FMPOV the solution to this problem is not to remove the `@Component` annotation from `HttpAdapterMetrics` but to do only one of scanning the package for components or manually instantiating the metrics class.

@BobClaerhout I suggest you simply remove the *adapterMetrics* method from the lora adapter's `Config` class instead.

BobClaerhout:
Ok, that's fine by me as well. I followed the suggestion from @b-abel in this case. @b-abel do you have any arguments on this?

sophokles73:
@BobClaerhout are you going to make the changes?

BobClaerhout:
@sophokles73, I sure can. Was waiting for @b-abel response but I can do them anyway.

sophokles73:
@BobClaerhout just go ahead

BobClaerhout:
@sophokles73 updated

adapters/http-vertx/src/main/java/org/eclipse/hono/adapter/http/impl/Config.java
```diff
@@ -122,4 +122,5 @@ public ObjectFactoryCreatingFactoryBean serviceFactory() {
         factory.setTargetBeanName(BEAN_NAME_VERTX_BASED_HTTP_PROTOCOL_ADAPTER);
         return factory;
     }
+
```
sophokles73: I don't think that we need this, do we?

BobClaerhout: Removed

Issue #1679, opened 2019-12-18
MongoDB based device registry


mhemmeter:
As of today Eclipse Hono provides file based implementations of the following APIs: Device Registry Management API, Device Registration API, Credentials API and Tenant API. These implementations are not meant for productive use. Therefore it is required that adopters implement these APIs or use an existing system in their concrete deployment scenario. The latter one is however not always existing.

To simplify usage of Eclipse Hono and increase adoption I think it makes sense to provide a production ready implementation of these APIs. 

I see the following requirements for a production ready implementation:
- The implementation is built on top of a scalable persistence layer. In our deployment of Eclipse Hono a MongoDB is used and it scales very well. Also Eclipse Ditto provides a persistence layer built on top of MongoDB, so using that technology simplifies the combined usage of different Eclipse IoT projects.
- All implementations are secured with an authentication mechanism

sophokles73:
I sympathize with the idea of having such a Mongo DB based implementation :+1:

mhemmeter:
I would propose to reduce the scope of this issue and focus on building the implementation on top of MongoDB as the persistence layer. We can secure the API in a follow-up issue I will create.

sophokles73:
@kaniyan anything left to do for this issue?

kaniyan:
@sophokles73 The implementation is over. The documentation needs to be done.

sophokles73:
@kaniyan ok, are you already working on that?

kaniyan:
@sophokles73 Currently I am working on helm charts for this mongo db based device registry, thereby to make a PR in the IoT packages repo. After that I can start with the documentation. We could also close this issue and I can open a new one to track the documentation separately.

mhemmeter:
From my perspective documentation should be part of this issue.

sophokles73:
@kaniyan Can you please first add the documentation for the Mongo DB registry and only after that is finished continue with the PR for the Hono Helm chart?

kaniyan:
@sophokles73 sure, I can proceed with the documentation first and thereby we can close this issue and then proceed with the Helm chart.

kaniyan:
@mhemmeter @sophokles73 The documentation for the _MongoDB based device registry_ is now over with the merge of PRs #2035 and #2027. We can now close this issue.

mhemmeter:
Thanks, @kaniyan. I'm closing the issue.

PR #1777, opened 2020-02-19
[#1679] Split and refactor device registry module.

kaniyan:
Currently I am working on the issue #1679, which is implementing a _Mongodb_ based device-registry. IMHO the existing device-registry module could be split into various sub-modules and also the classes could be reorganised under different packages. I think this enables to understand the code better and also to easier to maintain. This PR is a proposal how it could be refactored.

 In the existing device-registry module, although there is a possibility to add more implementations (eg. _Mongodb_, _Infinispan_ based etc.) by enabling the spring _services/repositories_ based on the conditional property, this approach bundles all the the implementation classes and the dependencies into a single jar/image. With the propsal in this PR, this can be avoided and each device-registry implementations could be created into a separate jar/image. 

The helm charts could also be adapted in such a way to specify which implementation ie. a _File_ or _Mongodb_  or _Infinispan_ based device-registry should be deployed.


calohmn:
With these changes there is now a
`services/device-registry` module with
`device-registry-base` and
`device-registry-file`
as submodules of that.

My intuition here would be to rather have the extra `device-registry` layer removed, so that we end up with
`services/device-registry-base` and
`services/device-registry-file`.
WDYT?

kaniyan:
My idea was to group the _base_ module and other device registry implementations under the `device-registry` module. Its also fine for me to move them under `services` and remove the `device-registry` module.

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/deviceConnection/AutowiredDeviceConnectionAdapter.java
```diff
@@ -11,7 +11,7 @@
  * SPDX-License-Identifier: EPL-2.0
  *******************************************************************************/
 
-package org.eclipse.hono.deviceregistry;
+package org.eclipse.hono.deviceregistry.service.deviceConnection;
```
calohmn: Package name should be all lowercase, so `deviceconnection`.

services/device-registry-base/src/test/java/org/eclipse/hono/deviceregistry/service/deviceConnection/MapBasedDeviceConnectionServiceTest.java
```diff
@@ -11,7 +11,7 @@
  * SPDX-License-Identifier: EPL-2.0
  *******************************************************************************/
 
-package org.eclipse.hono.deviceregistry;
+package org.eclipse.hono.deviceregistry.service.deviceConnection;
```
calohmn: Package name all lowercase (`deviceconnection`) also in `test` directory.

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/deviceConnection/AutowiredDeviceConnectionAdapter.java
kaniyan: My bad. I have changed it now.

PR #1786, opened 2020-02-24
[#1785] Fix gateway id getting set in reply-to address

### LINKED ISSUE #1785: AMQP adapter: command response from gateway has wrong device id property
When a northbound application sends a command to a device that is connected via a gateway, the device gets the command with a `reply-to` address that contains the gateway id instead of the device id.
When processing a command response message with that address, the AMQP adapter will then set the `device_id` property to the gateway id, not the device id.
This leads to the northbound application getting a command response message with the wrong `device_id` property value in such a scenario.

calohmn:
This fixes #1785.
The target device id instead of the gateway id has to be set in the reply-to address of a command message. Otherwise command response messages will get the wrong `device_id` property value.
AMQP integration tests now check for the correct property value.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/Command.java
+++ b/client/src/main/java/org/eclipse/hono/client/Command.java
@@ -54,7 +54,8 @@ public final class Command {
             final String tenantId,
             final String deviceId,
             final String correlationId,
-            final String replyToId) {
+            final String replyToId,
+            final String requestId) {
 
         this.validationError = validationError;
         this.message = message;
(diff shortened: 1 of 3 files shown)


client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -85,7 +87,8 @@ private Command(
      *
      * @param message The message containing the command.
      * @param tenantId The tenant that the device belongs to.
-     * @param deviceId The identifier of the device.
+     * @param deviceId The identifier of the device that the command will be sent to. This could be a mapped gateway,
+     *                 the original command target device is given in the message address in that case.
```
sophokles73: I do not quite get the meaning of the second sentence. Can you try to rephrase it?

calohmn: Done.

Issue #1760, opened 2020-02-10
Refactor AbstractHttpEndpoint concepts


dejanb:
Currently we have `AbstractHttpEndpoint` base class that is used both in event bus and direct implementation of the management APIs. As an effort to improve this part of the code we should split concepts used as a base of these implementations in appropriate base classes

jbtrystram:
@dejanb can we close this ?

PR #1789, opened 2020-02-24
[#1760] Remove redundant methods in abstractHttpEndpoint

jbtrystram:
Also deprecate methods tied to the old event bus implementation. See #1760 


sophokles73:
@jbtrystram can you rebase? I have moved the old event bus related functionality to their own (deprecated) subclass ...

dejanb:
@jbtrystram  Looks good so far. I would also move static methods from `Util` class we added recently, to the `AbstractHttpEndpoint`

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -322,4 +294,146 @@ protected final CorsHandler createCorsHandler(final String allowedOrigin, final
...
+     * Creates a new <em>OpenTracing</em> span for tracing the execution of a service operation.
+     *
+     * @param operationName The operation name that the span should be created for.
+     * @param spanContext Existing span context.
+     * @param tracer the Tracer instance.
+     * @param className The class name to insert in the Span.
+     * @return The new {@code Span}.
+     * @throws NullPointerException if operationName is {@code null}.
+     */
+    public static final Span newChildSpan(final String operationName, final SpanContext spanContext,
```
dejanb: I would leave span related methods in the separate class, as they are used outside of these endpoints.

service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -322,4 +294,146 @@ protected final CorsHandler createCorsHandler(final String allowedOrigin, final
...
+     * <li>Ends a response.</li>
+     * </ol>
+     *
+     * @param ctx The routing context of the request.
+     * @param result The generic result of the operation.
+     * @param customHandler An (optional) handler for post processing successful HTTP response, e.g. to set any additional HTTP
+     *                      headers. The handler <em>must not</em> write to response body. May be {@code null}.
+     * @param span The active OpenTracing span for this operation. The status of the response is logged and span is finished.
+     */
+    public static final void writeResponse(final RoutingContext ctx, final Result<?> result, final Handler<HttpServerResponse> customHandler, final Span span) {
```
dejanb: These methods should stay here, but as regular protected methods.

PR #1793, opened 2020-02-25
Provide abstract base classes for implementing device registries

ctron:
This PR provides the abstract base classes for implementing a device registry, that we currently have in EnMasse. As discussed last week during the Eclipse IoT F2F with @kaniyan, we create a PR for this early, in order to discuss.

Current open topics are:

* The abstract classes use Java futures, instead of vertx futures. This was just a better fit in the context of Infinispan. However using handlers seemed much more error prone than using futures.
* It currently misses the actual JDBC or Infinispan implementation. So it might seem a bit strange, but it focuses on the refactorings.
* The `TenantInformationService` misses an implementation. We use Kubernetes underneath, however adding this implementation will introduce additional dependencies, also on EnMasse internal structures. My proposal would be to create a `TenantInformationService` based on the AMQP client of Hono.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/MoreFutures.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/MoreFutures.java
@@ -1,0 +1,131 @@
+/*******************************************************************************
+ * Copyright (c) 2019, 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.service;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.function.Supplier;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+
+/**
+ * Helper class to work with futures.
+ */
+public final class MoreFutures {
+
+    private static final Logger log = LoggerFactory.getLogger(MoreFutures.class);
+
+    private MoreFutures() {}
+
+    /**
+     * Complete the provided handler with the provided future supplier.
+     * <p>
+     * The method will call the supplier for a {@link CompletableFuture}. If that calls
+     * fails, then the handler will be completed directly. Otherwise the future will
+     * be chained to complete the provided handler.
+     *
+     * @param <T> The type of the result.
+     * @param supplier The supplier of a {@link CompletableFuture}.
+     * @param handler The handler to complete.
+     */
+    public static <T> void completeHandler(final Supplier<CompletableFuture<T>> supplier, final Handler<AsyncResult<T>> handler) {
+        if (supplier == null) {
+            handler.handle(Future.failedFuture(new NullPointerException("'future' to handle must not be 'null'")));
+        }
+
+        final CompletableFuture<T> future;
+        try {
+            future = supplier.get();
+        } catch (final Exception e) {
+            log.info("Failed to prepare future", e);
+            handler.handle(Future.failedFuture(e));
+            return;
+        }
+
+        future.whenComplete((result, error) -> {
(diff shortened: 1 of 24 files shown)


kaniyan:
@ctron Thank you for creating this PR as we discussed. I find it good that that the refactored structure offers better readability than the existing code and offers some improvements. Please find my remarks as below.
- In this PR, I do not see abstract implementations for `TenantManagementService` (HTTP) and `TenantService` (AMQP). Have those been omitted intentionally? Our existing device registry supports these APIs. IMHO, I think we should also continue offering the same.
- I understood that the `TenantInformationService` is a new API that offers internal service implementation having method to check if a tenant exists or not. It is good to have such checks before creating or updating a device.
- I find it good that the handlers are now been replaced with futures. IMHO we should use _vertx_ futures instead of _java_ futures.
- I see that the auto provisioning and event bus related changes are yet to be integrated in this approach.

It seems to me that the amount of changes in this PR is quite big. Wouldn’t it be easier, if we incorporate changes incrementally instead of a single shot? Meanwhile in the PR #1777, I have refactored and tried to keep the changes minimal. What I did is mainly splitting the device registry module into sub modules, reorganized the classes into appropriate packages and extracted some methods to a utility class for re-usability. I believe that it has improved the readability a lot compared to the old structure. At the same time it is also easier now to stack up with varied persistence layer implementations. Would you mind also taking a look at it? IMHO, this code can be further developed incrementally by adding the changes such as replacing handlers with Futures, _tenantExists_ checks and so on. WDYT?

calohmn:
> The abstract classes use Java futures, instead of vertx futures. This was just a better fit in the context of Infinispan. However using handlers seemed much more error prone than using futures.

I think we should not use Java futures here. All our other code uses vertx futures and introducing it just for the one Infinispan device registry implementation based on the classes here isn't reason enough, IMHO. Regarding the handler params: I think in general we should better replace these with corresponding Future return values in most places - also in the service interfaces (e.g. RegistrationService).


> It currently misses the actual JDBC or Infinispan implementation. So it might seem a bit strange, but it focuses on the refactorings.

I think as part of the refactoring and for Hono 1.2, the file based device registry implementation should be adapted as well.
All in all, I think starting out with PR #1777 would offer a more incremental approach. Maybe changes in this PR can be put on top of and/or reconciled with that.

ctron:
I completely agree with using Vertx Futures over Java Futures. I think we discusses this briefly during the face-to-face. This is just the current state from the EnMasse repository. This is a base for discussion, and I would like try to avoid wasting time in refactoring this PR before we came to an agreement.

The TenantService is currently not implemented in the same way. There is an alternative implementation based on Kubernetes in EnMasse. But I think I could simply add the implementation, in the same pattern as this PR has for the other services. Again, I would like to avoid wasting time, so let's come to a conclusion first.

Yes, the code is based on Hono 1.0.x, so everything after that has to be added as well.

> It seems to me that the amount of changes in this PR is quite big. Wouldn’t it be easier, if we incorporate changes incrementally instead of a single shot? 

I already tried to shrink the PR to most essential changes. This is simple the changes we have, and which I think make sense to add. Of course, if you prefer to move on with #1777, but if you plan to add changes from this PR to #1777, I don't see how this would result in a smaller PR. This PR only adds new content, and doesn't modify anything yet.

kaniyan:
> I already tried to shrink the PR to most essential changes. This is simple the changes we have, and which I think make sense to add. Of course, if you prefer to move on with #1777, but if you plan to add changes from this PR to #1777, I don't see how this would result in a smaller PR. This PR only adds new content, and doesn't modify anything yet.

I suggested adding some features from this PR to PR #1777 on an iterative basis and not in a single shot. In that case, IMHO it is still small. At the end of each iteration, we will be having working device registry implementations and tests.

What I infer from the above discussions is, that this PR needs quite some time to be merged considering the work further to be done and the adaptions to be made in the demo device registry/tests which are not yet there. Containing many changes, it is natural that many of them need further discussions. For example, @calohmn suggested in his last comment to return the Future also in the service interfaces. Hence I assume that we cannot start with the issue #1679, which is implementing a _mongodb_ based device registry until this PR (plus related tests/demo implementation) is merged. 

The PR #1777 contains already the tests and file based implementation already adapted. Plus the changes are not big and this could be merged in a short span of time.  So next, we could start working on the _mongodb_ based implementation. In the next iterations or in parallel, we could focus on the other changes like checking if a tenant exist during device registration operations, returning of vertx features instead of handlers etc. IMHO, I do not see that these features are crucial that they should be included in the first iteration. FMPOV it is also easier to discuss, review and merge, if the amount of changes in a PR is small.

sophokles73:
@ctron @dejanb do we still need this PR?

dejanb:
@sophokles73 I made a series of PRs to replace this one. I'm wrapping up the last one now and will close this, once we have all that merged.

Issue #1796, opened 2020-02-26
Add received tenant id to Command validation error message?


boaks:
```
Command Command.from(
            final Message message,
            final String tenantId,
            final String deviceId);
```

checks for the matching tenant

```
final ResourceIdentifier addressIdentifier = ResourceIdentifier.fromString(message.getAddress());
if (!tenantId.equals(addressIdentifier.getTenantId())) {
    validationErrorJoiner.add("address contains wrong tenant");
}
```

is it possible to add that wrong tenant to the error message?

sophokles73:
I don't see why not. Are you creating a PR?

boaks:
> I don't see why not. 

sounds good.

> Are you creating a PR?

I will do so.

PR #1798, opened 2020-02-27
Add wrong tenant to validation error message.

boaks:
Issue #1796

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/Command.java
+++ b/client/src/main/java/org/eclipse/hono/client/Command.java
@@ -109,7 +109,7 @@ public final class Command {
         } else {
             final ResourceIdentifier addressIdentifier = ResourceIdentifier.fromString(message.getAddress());
             if (!tenantId.equals(addressIdentifier.getTenantId())) {
-                validationErrorJoiner.add("address contains wrong tenant");
+                validationErrorJoiner.add("address contains wrong tenant '" + addressIdentifier.getTenantId() + "'");
             }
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -109,7 +109,7 @@ public static Command from(
         } else {
             final ResourceIdentifier addressIdentifier = ResourceIdentifier.fromString(message.getAddress());
             if (!tenantId.equals(addressIdentifier.getTenantId())) {
-                validationErrorJoiner.add("address contains wrong tenant");
+                validationErrorJoiner.add("address contains wrong tenant '" + addressIdentifier.getTenantId() + "'");
```
sophokles73: well, wouldn't it make sense to include both the expected tenant as well as the found (wrong) one?

boaks: In my case the `Command.toString()` was used, which already contained the expected tenant. That was mixing me up. Because you get a error message, telling the wrong tenant is used, but displays the expected tenant. 
So, I'm not sure, if adding the expected tenant makes sense for other cases. For my case, it's not required.

calohmn: I think adding the wrong tenant is enough. `Command.toString()` output would then be something like:
`Invalid Command [tenant-id: myTestTenant, device-id: myTestDevice. error: address contains wrong tenant: 'tenantFromAddress']`

PR #1805, opened 2020-03-03
Update pipeline scripts in accordance with the refactored device registry

kaniyan:
The jenkins pipeline scripts have been updated to use the  artifact _hono-service-device-registry-file_ instead of _hono-service-device-registry_, which exists no more after the device registry got refactored. Once the _mongodb_ based device registry is available then in addition to the _hono-service-device-registry-file_ it (_hono-service-device-registry-mongodb_) could also be added.

### DIFF
--- a/deploy/src/main/sandbox/sandbox_deploy.sh
+++ b/deploy/src/main/sandbox/sandbox_deploy.sh
@@ -170,7 +170,7 @@ docker service create $CREATE_OPTIONS --name ${hono.registration.service} -p 256
   --env SPRING_PROFILES_ACTIVE=dev,prometheus \
   --mount type=volume,source=device-registry,target=/var/lib/hono/device-registry \
   --mount type=volume,source=hono-extensions,target=/opt/hono/extensions,readonly \
-  ${docker.image.org-name}/hono-service-device-registry:$HONO_VERSION
+  ${docker.image.org-name}/hono-service-device-registry-file:$HONO_VERSION
 echo ... done
 
 echo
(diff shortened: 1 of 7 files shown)


calohmn:
It looks like there are other places as well where `hono-service-device-registry` would need to be updated to `hono-service-device-registry-file` (e.g. `push_hono_images.sh`, `sandbox_deploy.sh`).

As there is no other device registry implementation in Hono yet, I think it's better to remove the `-file` suffix for now instead.
This will also ensure that the docker image name stays the same and that the IoT packages Helm charts don't need to be adapted to deploy a Hono master snapshot.

kaniyan:
@calohmn This has been closed for now, as it has been decided to revert the name to `hono-service-device-registry` instead of `hono-service-device-registry-file`. A separate PR with all the changes required to name it as `hono-service-device-registry-file` will be made later.

calohmn:
@kaniyan I've given it another thought and think we should go ahead with this PR, making the mentioned additions. Regarding the renamed docker image and the IoT packages charts, I now think that is not really a problem - to use a Hono master snapshot both the image name + version have to be given there anyway and therefore the new device registry image name just has to be used.

kaniyan:
> 
> 
> @kaniyan I've given it another thought and think we should go ahead with this PR, making the mentioned additions. Regarding the renamed docker image and the IoT packages charts, I now think that is not really a problem - to use a Hono master snapshot both the image name + version have to be given there anyway and therefore the new device registry image name just has to be used.

@calohmn In addition to the jenkins scripts, now I have updated the sandbox and push images scripts to use `hono-service-device-registry-file` and also the documentation.

PR #1803, opened 2020-03-03
[#1802] Refactor device management service to return future

kaniyan:
The idea is to incrementally address the issue #1802 in a series of PRs, so that it is easier to review. As part of this PR the `DeviceManagementService` is refactored to return _vertx_ futures and also the related implementations.

Issue #1802, opened 2020-03-03
Refactor device registry services to return vertx future


kaniyan:
As discussed in the last Hono meeting, the following device registry services are to be refactored to return _vertx_ futures instead of taking an _handler_ as an argument. The futures are better to deal with than the async handlers, easier to compose and to make the code more readable.

- `DeviceManagementService`
- `CredentialsManagementService`
- `TenantManagementService`
- `TenantService`
- `CredentialsService`
- `RegistrationService`
- `DeviceConnectionService`

PR #1810, opened 2020-03-04
[#1802] Refactor device registration service

kaniyan:
As part of this PR, the `RegistrationService` is refactored to return _vertx_ futures and also the related implementations. This should to be merged after PR #1803.


service-base/src/main/java/org/eclipse/hono/service/registration/RegistrationService.java
```diff
@@ -13,11 +13,10 @@
 
 package org.eclipse.hono.service.registration;
 
+import io.vertx.core.Future;
 import org.eclipse.hono.util.RegistrationResult;
 
 import io.opentracing.Span;
```
calohmn: import order

kaniyan: fixed

PR #1797, opened 2020-02-26
#1486 Improve usability of CLI

johnMinelli:
I tried to follow the suggestions detailed in the issue, to let the CLI be more attractive and self explanatory. I also tried to maintain for back-compatibility the spring configuration at startup. Executing the client with all parameters as in the guide those become the default (so you don't need to specify every time them when executing commands) otherwise the parameters in the application.yml were read and stored as default.

Signed-off-by: Giovanni Minelli <giovanni.minelli3@gmail.com>

Issue #1486, opened 2019-09-12
Improve usability of CLI


b-abel:
The CLI sample application is currently heavily based on the use of Spring Boot configuration. This is passed on to the user and is not what I expect from a CLI application. 
To customize the configuration, the user must understand Spring Boot and analyze the existing profiles. We should not expect this from every user. Especially since the use of the CLI is very prominent, because without it you can't even perform the simplest action in the "Getting started" guide (i.e. sending telemetry message).
I suggest to provide a simple command line usage, be it with arguments or interactive.

ctron:
I had the same issue, and created a simple command line tool for that: https://github.com/ctron/hot

I would be happy to contribute that to the Hono project.

johnMinelli:
I'm working on it. I committed the cli structure in my fork.

b-abel:
The CLI currently contains logic for both the northbound applications (in the cloud, connecting against the messaging endpoint) and the (southbound) device side (connection against the AMQP protocol adapter). This dual role makes it confusing to use the CLI. Since both sides are independent of each other and only "by coincidence" use both AMQP. It would probably be better to provide two separate applications instead of starting the same one with two different profiles. 
The device side could build on the `AmqpAdapterClientFactory' from #1765. This is a client for the AMQP adapter (there is no implementation in Hono for this yet), which can be easily used for different use cases.  
The other application could probably be based on `ApplicationClientFactory' in the same way (maybe it could also replace the example application in `example/hono-client-examples/`).

johnMinelli:
Hi, i got a deep sight in your PR code in the meantime to understand what do you mean. Do you suggest me to fork your code and finish the cli with your class or to setup and refine all with the actual AmqpCliClient class and add your factory in a second time when merged with master?

b-abel:
Hi, @johnMinelli! Sorry for the late reply, I've been ill for a while. I was hoping the PR was ready. I'm currently clarifying the status. I'll probably know more tomorrow.

johnMinelli:
'Ill' is a buzzword nowdays ahah. I've done this #1811 and commented some line to insert the factory you have made

b-abel:
@johnMinelli FYI: My PR #1765 has been merged today.

sophokles73:
This has been done with the new Picocli based CLI from #2914

PR #1811, opened 2020-03-04
#1486 CLI Improve usability

johnMinelli:
I tried to follow the suggestions detailed in the issue, to let the CLI be more attractive and self explanatory. I also tried to maintain for back-compatibility the spring configuration at startup. Executing the client with all parameters as in the guide those become the default (so you don't need to specify every time them when executing commands) otherwise the parameters in the application.yml were read and stored as default.(It's a duplicate of #1797)


calohmn:
@johnMinelli The Travis build is failing here because of errors during javadoc creation:
````
[ERROR] -> [Help 1]

org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.1.0:jar (attach-javadocs) on project hono-cli: MavenReportException: Error while generating Javadoc: 

Exit code: 1 - /home/travis/build/eclipse/hono/cli/src/main/java/org/eclipse/hono/cli/shell/InputReader.java:108: error: self-closing element not allowed

     * <br/>

       ^

/home/travis/build/eclipse/hono/cli/src/main/java/org/eclipse/hono/cli/shell/InputReader.java:109: error: self-closing element not allowed

     * Passing null for defaultValue signifies that there is no default value.<br/>

                                                                              ^
````
If you run `mvn clean install -DcreateJavadoc=true` locally, you can reproduce the errors.

**Comment by johnMinelli** [comment]
Enabling javadoc i've an error in Hotrod Device Connection client 
>     [ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.1.0:jar (attach-javadocs) on project client-device-connection-infinispan: MavenReportException: Error while generating Javadoc: 
>     [ERROR] Exit code: 1 - C:\Users\Gio\Desktop\git\hono\client-device-connection-infinispan\src\main\java\org\eclipse\hono\deviceconnection\infinispan\client\HotrodBasedDeviceConnectionInfo.java:22: error: cannot access HealthCheckProvider
>     [ERROR] import org.eclipse.hono.service.HealthCheckProvider;
>     [ERROR]                                ^
>     [ERROR]   bad class file: C:\Users\Gio\.m2\repository\org\eclipse\hono\hono-service-base\1.2.0-SNAPSHOT\hono-service-base-1.2.0-SNAPSHOT.jar(org/eclipse/hono/service/HealthCheckProvider.class)
>     [ERROR]     class file has wrong version 55.0, should be 52.0
>     [ERROR]     Please remove or make sure it appears in the correct subdirectory of the classpath.

**Comment by calohmn** [comment]
@johnMinelli Which JDK are you using? Somehow the javadoc command seems to use a different source version (1.8 instead of 1.11). I cannot reproduce this with OpenJDK 11 on Linux.
An idea: You could edit `client-device-connection-infinispan/pom.xml` and put 
`<properties><java.level>1.11</java.level></properties>` 
below the `<description>A Hotrod based client ...` line and then build again. (Just for testing, should not be committed in this PR).

In any case, to resolve the aforementioned javadoc error on Travis, you could simply replace the 2 `<br/>` occurrences in the javadoc of the `InputReader#promptWithOptions` method with `<br>`. Just tested it locally - there are no further javadoc errors.

File: cli/src/main/java/org/eclipse/hono/cli/config/SpringAppConfig.java [hunk 1]
```diff
@@ -11,24 +11,23 @@
  * SPDX-License-Identifier: EPL-2.0
  *******************************************************************************/
 
-package org.eclipse.hono.cli;
+package org.eclipse.hono.cli.config;
 
+import io.vertx.core.Vertx;
+import io.vertx.core.VertxOptions;
```
sophokles73: import order

cli/src/main/java/org/eclipse/hono/cli/config/SpringShellConfig.java
```diff
@@ -0,0 +1,79 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
```
sophokles73: just the current year for new classes

cli/src/main/java/org/eclipse/hono/cli/shell/ClihonoPromptProvider.java
```diff
@@ -0,0 +1,34 @@
...
+import org.jline.utils.AttributedString;
+import org.jline.utils.AttributedStyle;
+import org.springframework.shell.jline.PromptProvider;
+import org.springframework.stereotype.Component;
+
+/**
+ * Shell class.
+ */
+@Component
+public class ClihonoPromptProvider implements PromptProvider {
```
sophokles73: CliHonoPromptProvider

cli/src/main/java/org/eclipse/hono/cli/shell/ClihonoPromptProvider.java
```diff
@@ -0,0 +1,34 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
```
sophokles73: just the current year

cli/src/main/java/org/eclipse/hono/cli/shell/InputReader.java
```diff
@@ -0,0 +1,204 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
```
sophokles73: current year only

cli/src/main/java/org/eclipse/hono/cli/shell/InputReader.java
```diff
@@ -0,0 +1,204 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.cli.shell;
+
+import org.jline.reader.LineReader;
+import org.springframework.util.StringUtils;
+
+import java.util.ArrayList;
```
sophokles73: import order

cli/src/main/java/org/eclipse/hono/cli/shell/PromptColor.java
```diff
@@ -0,0 +1,46 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
```
sophokles73: current year only

cli/src/main/java/org/eclipse/hono/cli/shell/ShellHelper.java
```diff
@@ -0,0 +1,162 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
```
sophokles73: current year only

johnMinelli:
A dumb question: the "resolve conversation" is just to take track of the changes? Do i need to fix the code and recommit, right?

sophokles73:
> A dumb question: the "resolve conversation" is just to take track of the changes? Do i need to fix the code and recommit, right?

None of your questions are dumb at all. yes and yes :-)

cli/src/main/java/org/eclipse/hono/cli/adapter/TelemetryAndEvent.java
```diff
@@ -0,0 +1,124 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
```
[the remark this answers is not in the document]
johnMinelli: Do i need to update the header as "Copyright (c) 2019, 2020 Contributors to the Eclipse Foundation" for every file?

cli/pom.xml
```diff
@@ -80,6 +66,56 @@
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-engine</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jline</groupId>
+            <artifactId>jline</artifactId>
+            <version>3.4.0</version>
```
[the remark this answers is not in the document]
johnMinelli: There is a particular rule to follow for the order of deps in the `bom/pom.xml`?

cli/src/main/java/org/eclipse/hono/cli/adapter/AmqpCliClient.java
```diff
@@ -33,29 +35,37 @@
 /**
  * Abstract base CLI connection for interacting with Hono's AMQP adapter.
  */
-public abstract class AmqpCliClient extends AbstractCliClient {
+public class AmqpCliClient extends AbstractCliClient {
 
+    /**
+     * The connection to the AMQP org.eclipse.hono.cli.app.adapter.
+     */
+    public ProtonConnection adapterConnection;
```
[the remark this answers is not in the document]
johnMinelli: It's used in `org\eclipse\hono\example\protocolgateway\adapter\CommandAndControlReceiver.java` and `...\TelemetryAndEventSender.java`. Before these classes extended the class with spring and connection, now the spring part and the connection are in two different file, so these classes extends the spring and create an instance of AmqpCliClient for the connection.

cli/src/main/java/org/eclipse/hono/cli/adapter/AmqpCliClient.java
```diff
@@ -67,7 +77,7 @@ public void setClientConfig(final ClientConfigProperties props) {
      * @throws IllegalStateException if the connection to the org.eclipse.hono.cli.app.adapter is
      *         not established.
      */
-    protected Future<ProtonSender> createSender() {
+    public Future<ProtonSender> createSender() {
```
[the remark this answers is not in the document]
johnMinelli: Same as the other method

cli/src/main/java/org/eclipse/hono/cli/adapter/AmqpCliClient.java
```diff
@@ -85,8 +95,8 @@ public void setClientConfig(final ClientConfigProperties props) {
      * @return A future containing the established connection. The future will
      *         be succeeded once the connection is open.
      */
-    protected Future<ProtonConnection> connectToAdapter() {
-
+    public Future<ProtonConnection> connectToAdapter() {
```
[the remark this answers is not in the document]
johnMinelli: Same as the other method

cli/pom.xml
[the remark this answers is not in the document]
sophokles73: please add the version property in alphabetical order and add the dependencies to the end of the existing non-test scope deps

cli/src/main/java/org/eclipse/hono/cli/adapter/AmqpCliClient.java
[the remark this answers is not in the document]
sophokles73: then please create an accessor and declare the field private for the time being ...

sophokles73:
@johnMinelli can you also please rebase?

cli/src/main/java/org/eclipse/hono/cli/AbstractCliClient.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2018, 2020 Contributors to the Eclipse Foundation
```
sophokles73: This should specify the start and end of the interval during which this file was created and modified. Sorry for not having been specific enough about it. So, this should read
```
 * Copyright (c) 2016, 2020 Contributors to the Eclipse Foundation
```

cli/src/main/java/org/eclipse/hono/cli/Application.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2019, 2020 Contributors to the Eclipse Foundation
```
sophokles73: same here

cli/src/main/java/org/eclipse/hono/cli/adapter/AmqpCliClient.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2019, 2020 Contributors to the Eclipse Foundation
```
sophokles73: same here

cli/src/main/java/org/eclipse/hono/cli/application/Commander.java
```diff
@@ -1,59 +1,84 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
- * 
+ * Copyright (c) 2016, 2019, 2020 Contributors to the Eclipse Foundation
```
sophokles73: same here

cli/src/main/java/org/eclipse/hono/cli/application/Receiver.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2019, 2020 Contributors to the Eclipse Foundation
```
sophokles73: same here

cli/src/main/java/org/eclipse/hono/cli/config/SpringAppConfig.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2019, 2020 Contributors to the Eclipse Foundation
```
sophokles73: same here

cli/src/main/java/org/eclipse/hono/cli/shell/ClihonoPromptProvider.java
sophokles73: Can you please change the class name?

cli/src/test/java/org/eclipse/hono/cli/client/ReceiverTest.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2019, 2020 Contributors to the Eclipse Foundation
```
sophokles73: same here

cli/src/main/java/org/eclipse/hono/cli/client/ApplicationClient.java
```diff
@@ -0,0 +1,351 @@
...
+import org.springframework.shell.standard.ShellCommandGroup;
+import org.springframework.shell.standard.ShellComponent;
+import org.springframework.shell.standard.ShellMethod;
+import org.springframework.shell.standard.ShellOption;
+import org.springframework.stereotype.Component;
+import org.eclipse.hono.cli.client.ClientConfig.MessageTypeProvider;
+
+import io.vertx.core.Vertx;
+import sun.misc.Signal;
+import sun.misc.SignalHandler;
```
sophokles73: are you sure that these are needed?

johnMinelli: Yep, It's to catch the interrupt signal from console

johnMinelli:
What's the problem with the name class CliHonoPromptProvider? I've no problem renaming it, what name do you suggest @sophokles73 ?

sophokles73:
> What's the problem with the name class CliHonoPromptProvider? I've no problem renaming it, what name do you suggest @sophokles73 ?

None at all, the current name, however, is `ClihonoPromptProvider`, i.e. no camel case ...

cli/src/main/java/org/eclipse/hono/cli/client/ApplicationClient.java
sophokles73: my point is more that it is usually considered bad practice to depend on classes from the JDK implementation, in this case from `sun.*`. Is there an alternative way of achieving what you are trying to do without depending on these classes?

johnMinelli: Actually i tried and this way doesn't catch the sigkill from terminal. I tried also with `addShutdownHook` but doesn't work. I'm not sure how to handle it.
Actually i removed that and the side effect is that with ctrl+c you end the whole execution. Any suggestion? Also to run it from command line do i need to use the "*-exec.jar" right? How i can build it? Running maven i compile only the "*.jar" and the "*-source.jar"

sophokles73:
@johnMinelli Given that I am not familiar with Spring Shell, I have a hard time understanding how your code is actually supposed to work. Which class do I need to run in order to ... do what exactly? Is the idea that I get a *Hono client shell* where I can enter commands to execute towards Hono? Can you provide some instructions please regarding how to get started?

bom/pom.xml
```diff
@@ -200,6 +202,51 @@
...
+        <version>${spring-shell.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.springframework.shell</groupId>
+        <artifactId>spring-shell-standard</artifactId>
+        <version>${spring-shell.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.springframework.shell</groupId>
+        <artifactId>spring-shell-starter</artifactId>
```
sophokles73: my understanding is that it should be sufficient to define this as a dependency in order to pull in all the other deps, right?

cli/src/main/java/org/eclipse/hono/cli/client/ApplicationClient.java
```diff
@@ -0,0 +1,351 @@
...
+                    return;
+                }
+                //create the connection
+                try {
+                    clientFactory = ApplicationClientFactory.create(HonoConnection.newConnection(vertx, fork.honoClientConfig));
+                    latch = new CountDownLatch(1);
+                    final Commander commanderInstance = new Commander(clientFactory, vertx, fork, inputReader);
+                    commanderInstance.start(latch);
+                    //to handle forced interruption
+                    Signal.handle(new Signal("INT"), new SignalHandler() {
```
sophokles73: Can you explain what you are trying to achieve here?

bom/pom.xml
johnMinelli: maybe... i let the IDE handling the imports. I'll try to remove the dependencies

cli/src/main/java/org/eclipse/hono/cli/client/ApplicationClient.java
johnMinelli: For insight in the shell, here you can find all: https://projects.spring.io/spring-shell/
However the basic knowledge is that the methods annotated as a `@Command` will be executed as a command in the shell.

> what you are trying to achieve here?

If you run the command "receiver" for example and then want to stop it, pressing CTRL+C, you'll exit the entire shell but the expected behaviour is to end just the execution of the command. (and then if you want you can exit the shell)

sophokles73: FMPOV that is not important. Let's go with the default behavior in a first step, i.e. when pressing CTRL + C, the shell terminates. In practice, you will need to run two shells anyway, one for receiving messages and another one for interactive tasks ...

johnMinelli:
i pulled all from hono/master in my master rebased my dev-branch in my master and than put my changes. That didn't resolve the conflicts?  (However i checked and are just some spaces)

bom/pom.xml
johnMinelli: you were right i removed useless reference both for jline and spring-shell

sophokles73:
you will need to resolve the conflicts in your branch, otherwise we cannot build you code ...

johnMinelli:
> I pulled all from hono/master in my master, rebased my dev-branch in my master and than
Actually i rebased in my remote/master and not in the local updated master...

But now my situation is this: 
image

That are all the hono commit from the moment of my branch creation. If i push i got a PR of 100k+ LOC ahah

sophokles73:
> That are all the hono commit from the moment of my branch creation.

That's what you'd expect when you rebase on HEAD of master, isn't it? Your branch should now be one or two commits ahead of master when you do a `git status` in your workspace ...

johnMinelli:
Why Travis build is still pending?

cli/src/test/java/org/eclipse/hono/cli/client/ReceiverTest.java
sophokles73: The format for the period is LOWER BOUNDARY, UPPER BOUNDARY.
In this case, it should be `2016, 2020` ...

cli/src/main/java/org/eclipse/hono/cli/client/ClientConfig.java
```diff
@@ -0,0 +1,95 @@
...
+
+    public ClientConfigProperties honoClientConfig;
+
+    public String tenantId;
+    public String deviceId;
+    public String messageType;
+    public int connectionRetryInterval;
+    public int requestTimeoutInSecs;
+    public String messageAddress;
+    public String payload;
```
sophokles73: please make these fields private and add accessors, if necessary, in order to increase encapsulation. Otherwise it is very hard to control the state of this class.

cli/src/main/java/org/eclipse/hono/cli/client/ClientConfig.java
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Constructor.
+     *
+     * @param tenantId Parameter for command execution
+     * @param deviceId Parameter for command execution
+     * @param messageType Parameter for command execution
+     * @param connectionRetryInterval Parameter for command execution
+     * @param requestTimeoutInSecs Parameter for command execution
+     * @param messageAddress Parameter for command execution
+     * @param payload Parameter for command execution
```
sophokles73: please provide meaningful descriptions

cli/src/main/java/org/eclipse/hono/cli/client/ClientConfig.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public ClientConfig(final String tenantId, final String deviceId, final String messageType, final int connectionRetryInterval, final int requestTimeoutInSecs, final String messageAddress, final String payload) {
```
sophokles73: it is kind of hard to understand why message payload would be part of client configuration. On the other hand, the host and port to connect to and the credentials seem to be missing. Or am I mistaken?

cli/src/main/java/org/eclipse/hono/cli/client/ApplicationClient.java
```diff
@@ -0,0 +1,321 @@
...
+        this.vertx = Objects.requireNonNull(vertx);
+        this.ctx = vertx.getOrCreateContext();
+    }
+    /**
+     * Sets the default client configuration with the parameters got at startup.
+     *
+     * @param honoClientConfig The Client Configuration class instance.
+     */
+    @Autowired
+    public final void setClientConfigProperties(final ClientConfigProperties honoClientConfig) {
```
sophokles73: what do we need this for? Aren't all necessary parameters being provided when entering a command in the shell?

cli/src/main/java/org/eclipse/hono/cli/client/ApplicationClient.java
```diff
@@ -0,0 +1,321 @@
...
+        clientConfig = new ClientConfig(tenantId, deviceId, messageType, connectionRetryInterval, commandTimeoutInSeconds, messageAddress, messagePayload);
+        clientConfig.honoClientConfig = Objects.requireNonNull(honoClientConfig);
+    }
+    /**
+     * Sets the Client Factory.
+     *
+     * @param factory The ApplicationClientFactory instance.
+     */
+    @Autowired
+    public final void setApplicationClientFactory(final ApplicationClientFactory factory) {
```
sophokles73: same here, my understanding is that we create a new factory instance for every command entered on the shell anyway ... so why do we need to inject one here?

cli/src/main/java/org/eclipse/hono/cli/client/ApplicationClient.java
```diff
@@ -0,0 +1,321 @@
...
+//    }
+
+
+    /**
+     * A base class for implementing command line client's methods that access Hono's
+     * northbound APIs.
+     *
+     */
+    @ShellComponent
+    @ShellCommandGroup("Northbounds Commands (Application)")
```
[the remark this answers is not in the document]
johnMinelli: because the shell annotation group apply only on a class and i think that this subdivision it's very straightforward for a user

cli/src/main/java/org/eclipse/hono/cli/client/ApplicationClient.java
```diff
@@ -0,0 +1,321 @@
...
+import org.springframework.shell.standard.ShellCommandGroup;
+import org.springframework.shell.standard.ShellComponent;
+import org.springframework.shell.standard.ShellMethod;
+import org.springframework.shell.standard.ShellOption;
+import org.springframework.stereotype.Component;
+
+import io.vertx.core.Vertx;
+
+/**
+ * Being a Component will be injected by spring.
```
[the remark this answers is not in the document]
johnMinelli: yes it is, you are right

cli/src/main/java/org/eclipse/hono/cli/shell/CliHonoPromptProvider.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+@Component
+public class CliHonoPromptProvider implements PromptProvider {
+    /**
+     * Override the shell config to set the custom prefix.
+     * @return The attribute custom string.
+     */
+    @Override
+    public AttributedString getPrompt() {
+        return new AttributedString("HONO:>", AttributedStyle.DEFAULT.foreground(AttributedStyle.BLUE));
```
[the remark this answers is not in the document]
johnMinelli: Jline offers: BLACK, RED, GREEN, YELLOW, BLUE, MAGENTA, CYAN, WHITE, BRIGHT;
Do you prefer a Hono theme `(#F08712)` ?

cli/src/main/java/org/eclipse/hono/cli/client/ClientConfig.java
johnMinelli: Actually both the constructor and the name class are pretty misleading...
I'll make you a question before fix it:
I wanted to maintain the spring parameters compatibility at startup and those are the parameters in the ApplicationClient annotated with the `@Value` tag.
Those parameters constitute the basic env to execute the commands while the other parameters like host, port etc can be inserted at the moment of executing a single command 
(e.g. receive --host --port --username --password --tenant --message)
Said that, i'm not an expert of UX (maybe the opposite) but i made a thought on a possible use case:
As a user i want to send many event with the same looong payload

than a great behaviour will be:

> java -jar hono-cli-*-exec.jar --message.payload="that's a long one" --tenant.id=$MY_TENANT --device.id=$MY_DEVICE

>HONO:> send --hostname $AMQP_NETWORK_IP --port 15672 --username me --password secret
>HONO:> send --hostname $AMQP_NETWORK_IP --port 15672 --username me --password secret
>HONO:> send --hostname $AMQP_NETWORK_IP --port 15672 --username me --password secret --message "this time the message it's short"
>HONO:> send --hostname $AMQP_NETWORK_IP --port 15672 --username me --password secret

In this way you don't have to insert the credential directly in the shell but you can have the advantage of the sping env. The question is: what are the parameters that is better to acquire at startup and what in the command? And also the parameters can be acquired in a more straightforward way (but more rambling):

>java -jar hono-cli-*-exec.jar

>HONO:> send --hostname $AMQP_NETWORK_IP --port 15672 
>What message do you want to send?
>HONO:> "Here the body message"
>What is the message type? ["event","telemetry"]
>HONO:> "event"
>Message sent

Also i wanted to underline the concept that the history of the commands executed (like up and down arrow on a regular shell) it's not present.

cli/src/main/java/org/eclipse/hono/cli/shell/CliHonoPromptProvider.java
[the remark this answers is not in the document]
sophokles73: what about just `WHITE`?

cli/src/main/java/org/eclipse/hono/cli/client/ClientConfig.java
sophokles73: My understanding is that the shell will be used to do two different things
1. interact with Hono's north bound API, i.e. act as an application, and
2. interact with Hono's AMQP protocol adapter, i.e. act as a device

FMPOV the commands for doing so should therefore be put into one class each which also maintains the connection state to the corresponding endpoint, i.e. either the dispatch router or the AMQP adapter. These classes can then be annotated with `@ShellComponent` and their (command) methods with `@ShellCommand`. I also think that these classes should provide means to establish a connection and to disconnect and that other commands, e.g. for sending telemetry or events, should be available only once the connection has been established.

That said, it might make sense to be able to provide *default* values for e.g. the host and port to connect to or the username and password so that when running the *connect* command, you do not need to specify them again. This would also alleviate the user from having to specify any of the connection related properties when using e.g. *sendTelemetry* or *sendEvent*.

Separating the connection establishment from the command execution itself makes the current code very hard to read and understand. Maybe you can re-factor the classes based on my comments above. WDYT?

johnMinelli: I guess that give the user a client that abstract the structure of Hono and let him make less choice as possible is the objective. If i got it right you are suggesting something like:

>java -jar hono-cli.jar
>HONO:>help
>Available commands:
>`        `connect AMQP-adapter [-H --host] [-P --port] [-u --username] [-p --password]
>`        `connect messegging-network [-H --host] [-P --port] [-u --username] [-p --password]
>HONO:> connect AMQP-adapter --hostname $AMQP_NETWORK_IP --port 15672 --username me --password secret
>Connected!
>HONO:>help
>... the two commands ... or disconnect or exit

or maybe you prefer a
>java -jar hono-cli.jar --adapter --hostname $AMQP_NETWORK_IP --port 15672 --username me --password secret
>HONO:>help
>... the two commands ... or disconnect=exit

The second is almost the approach used now but with parameters more human-usable<

However i totally agree to separate the acquisition of parameters by context as you proposed.

Also i've another question: the comments and the messages along with the commands must refer to the behaviour or to the connection aspects? i.e.:
"Act as a device" or "Connect to the AMQP-adapter". The user want to use the client to test how works a device or want to test the adapter?
I speak as a newcome to Hono and if i can say my two cents:
If I don't know how all the structure works but I want to make some test,, having many terms (that yes refers to the same thing) it confuse me. At least that was my personal first approach with Hono.

sophokles73: > f i got it right you are suggesting something like

exactly :-)

> Also i've another question: the comments and the messages along with the commands must refer to the behaviour or to the connection aspects? i.e.:
"Act as a device" or "Connect to the AMQP-adapter". The user want to use the client to test how works a device or want to test the adapter?

In the end, the CLI can be used to do both. New users will most likely use it to *act as a device* as part of the getting started guide, whereas the developers of adapters will probably want to use the CLI to test the adapters.

> I speak as a newcome to Hono and if i can say my two cents:
If I don't know how all the structure works but I want to make some test,, having many terms (that yes refers to the same thing) it confuse me. At least that was my personal first approach with Hono.

Totally valid point. Do you have a suggestion as to how we could improve the terminology and make it less confusing?

johnMinelli:
what's the problem?

cli/src/main/java/org/eclipse/hono/cli/client/Client.java
```diff
@@ -0,0 +1,80 @@
...
+        cfp.setHost(host);
+        cfp.setPort(port);
+        cfp.setUsername(username);
+        cfp.setPassword(password);
+        //create the factory with connection info
+        // The #1765 will provide the factory for the AMQP adapter
+        try {
+            if (profile == ClientConfig.Profile.APPLICATION) {
+                clientConfig.createFactory(vertx, cfp, profile);
+            }
```
johnMinelli: Actually i added this before commit because device commands not require the creation of the factory, but since the method to check the availability `connectionDevAvailability` check anyway the factory value the device methods can't be executed... my bad.
The integration with the AMQP factory PR will fit.
However i'll wait to re-commit the change, because i haven't figured out what's wrong with the build.

sophokles73: The failing integration test on Travis does not seem to be related to your changes. We'll need to look into that separately. I have restarted the Travis build for now, expecting it to succeed this time ...

cli/src/main/java/org/eclipse/hono/cli/application/ApplicationProfile.java
```diff
@@ -0,0 +1,110 @@
...
+import org.springframework.shell.standard.ShellOption;
+import org.springframework.shell.standard.ValueProviderSupport;
+import org.springframework.stereotype.Component;
+
+/**
+ * This class contains the methods that the user can execute acting as an application consumer.
+ * To execute these commands the ClientConfig.profile must be valued as APPLICATION and the connection must be created
+ * with the sub command connect of the shell.
+ */
+    @ShellComponent
```
sophokles73: why is all this indented?

cli/src/main/java/org/eclipse/hono/cli/AbstractCliClient.java
```diff
@@ -44,32 +45,79 @@
...
+        this.clientConfig = config;
+    }
+
+    /**
+     * Check the availability of the factory and of the profile active.
+     * It allows the execution of sub commands of the application profile.
+     *
+     * @return Availability status.
+     */
+    protected Availability subCommandsAppAvailability() {
```
sophokles73: FMPOV this belongs into the class implementing all the application related commands ...

cli/src/main/java/org/eclipse/hono/cli/AbstractCliClient.java
```diff
@@ -44,32 +45,79 @@
...
+                : Availability.unavailable("You are not connected to the Messaging Network");
+    }
+
+    /**
+     * Check the availability of the factory and of the profile active.
+     * It allows the execution of sub commands of the device profile.
+     *
+     * @return Availability status.
+     */
+    protected Availability subCommandsDevAvailability() {
```
sophokles73: FMPOV this belongs into the class implementing the AMQP adapter related commands ...

cli/src/main/java/org/eclipse/hono/cli/AbstractSubCommand.java
```diff
@@ -0,0 +1,47 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.vertx.core.Context;
+import io.vertx.core.Vertx;
+
+/**
+ * Abstract base class for the class containing the sub commands execution.
+ */
+public abstract class AbstractSubCommand {
```
sophokles73: what is a *SubCommand*? In Spring Shell parlais, these are all just commands, right?
So this might be an *AbstractCommand*?

cli/src/main/java/org/eclipse/hono/cli/client/Client.java
```diff
@@ -0,0 +1,80 @@
...
+ * This class contains the methods that the user can execute to create or destroy the factory given connection info.
+ * The vertx instance wil be prepared and the factory correspondent to the profile chosen will be created.
+ */
+@ShellComponent
+@Component
+public class Client extends AbstractCliClient {
+
+    @ShellMethod("Create the connection to a given endpoint\n" +
+            "\t\t\t[-H --host] [-P --port] [-u --username] [-p --password]")
+    void connect(@ShellOption(value = {"-H", "--host"},  defaultValue = "", help = "The address of the endpoint which you want to establish a connection to") final String host,
```
sophokles73: this belongs to the *ApplicationProfile* and the *DeviceProfile* FMPOV

johnMinelli: I disagree on that:
The user asks to connect submitting parameters [-H --host] [-P --port] [-u --username] [-p --password]
Then he will be prompted to insert the profile:

> Profile:
> [APPLICATION]: Connect to the Messaging Network
> [DEVICE]: Connect to the AMQP Adapter
> Please enter one of the [] values

(*)*There the factory corresponding to the profile will be created with the connection instance, but still not connected*

And return a positive feedback to the user that can now use the methods of the profile choosen. 

I thought that you asked for something like that. But say me if i'm wrong.

johnMinelli: If this is all right i have a small issue with the AMQP factory. @b-abel (Sorry to throw you in the middle of the conversation i'll try to give you a bit of context)
If the behaviour in the previusly message is right, at the moment (*), when i create the factory (that can be a factory with the application methods or a factory with device methods to interact with the AMQP adapter) i don't have the tenantId that i need to create the AMQP factory.

Do i need to ask the user for the tenantId previusly?
Or maybe it's my code that is bad structured?
Or i need to specify the tenant later adding a method in the factoryImpl to set it?

sophokles73: > I thought that you asked for something like that. But say me if i'm wrong.


I think that we want to achieve the same thing. My proposal is simply to have specific methods for connecting to the endpoints
* *connectApplication*
* *connectAmqp*

Later we will probably also want to add *connectMqtt* in order to connect to the MQTT adapter ...
Using specific methods allows us to increase cohesion within the **Profile* classes FMPOV.

johnMinelli: ok i thought that a single connect command would be nicer but with future implementation maybe it's less practical.

So in the connectAmqp do i need to acquire also the tenantId?

sophokles73: the tenant ID is usually part of the username: `{auth-id}@{tenant-id}` so no need to have an extra param for that FMPOV

cli/src/main/java/org/eclipse/hono/cli/client/ClientConfig.java
```diff
@@ -0,0 +1,86 @@
...
+
+import io.vertx.core.Vertx;
+
+/**
+ * Configuration class. Holds all the parameters to execute a command.
+ * <p>
+ * A default instance will be created at startup and it will be cloned and modified with custom
+ * parameter for the execution of a command.
+ */
+public class ClientConfig {
```
sophokles73: there doesn't seem to be a real need for this class, is there? The default properties are passed in to the **Profile* classes via *ClientConfigProperties* and for the *connect** commands, you only need to create a copy of it and add the properties provided to the command as params (which you already do). From these updated *ClientConfigProperties* you can then create the client factory and the connection, right?

johnMinelli: Exactly, the commented description is an oversight.
I'm also thinking to refactor the name of the class. Since it holds only the profile and the factory for the active profile it's not a real `ClientConfig`... what do you say?

sophokles73: My point is that I believe that we can safely remove this class as it seems to have no real purpose. In fact, FMPOV it makes things more complicated than they need to be ...

johnMinelli: Ok i see, since i do't have anymore a generalized connection method i don't need it. You are right, I have to pay more attention in the refactoring of the code.

cli/src/main/java/org/eclipse/hono/cli/AbstractCliClient.java
```diff
@@ -44,32 +59,87 @@
...
+        return StringUtils.isEmpty(profile) ? null : clientFactory;
+    }
+    /**
+     * Accessor to the active profile.
+     *
+     * @return The active profile.
+     */
+    protected Profile getProfile() {
+        return profile;
+    }
```
sophokles73, line 142: none of these methods seem to be used in `AbstractCliClient` but their implementation seems to be specific to the concrete subclasses. So FMPOV these methods should be removed here and should be implemented in the concrete subclasses instead. There is no need to duplicate the fact that we have multiple subclasses by introducing the `Profile` enum whose sole purpose seems to be to indicate, in which subclass we are currently running code ...

johnMinelli: Can i delete the disconnect command?(And the class Client with it)

Removing the profile flag and moving down the factory, that method needs something like a boolean flag in the `AbstractCliClient` class so i can interrupt the availability of the single profile when the disconnect is called, but i don't like the flag thing, and also from the User POV calling disconnect or calling another `connect-something` doesn't change anything.

However (just to underline it) moving the factory down of a level and removing the disconnect, if he user call a

> connect-application

and then a 

> connect-device

the memory remains allocated for both factories. Not a big deal, but the only solution to this is to make two disconnect methods that FMPOV are evil for a user.

sophokles73: Keeping the factory instances is no problem at all. I don't quite get why having two disconnect methods would be a bad thing. What other way would there be for a user to indicate whether he wants to close the connection to the north bound API or the protocol adapter? There are also two different methods for establishing the connection aren't there?

johnMinelli: Actually the `disconnect` and `connect` not disconnect or connect just create the assumptions to execute the commands relative to the profile choose. Only at the execution of the command the connection is instantiated.
So why bother the user to choose what he want to disconnect? and furthermore what's the difference for him if he disconnect or not? (I reasoned in the same way for the connect but you said that for future implementation like MQTT adapter was better to make it specific). 
(I said 'evil' for the user not 'bad' 😊)
However if you say two disconnect, i'll do so.


(update)
I can't disconnect (i.e. `clientFactory = null`) from another class, the variable it's not in common between the class Client and the others 2 of the profiles, and that's why i used previously the bean class ClientConfig. So i guess i'll go for 2 disconnect beside what i have said.

cli/src/main/java/org/eclipse/hono/cli/application/ApplicationProfile.java
```diff
@@ -0,0 +1,147 @@
...
+            shellHelper.printInfo("The previously connection will be destroyed");
+        }
+        final ClientConfigProperties cfp = new ClientConfigProperties(this.honoClientConfig);
+        cfp.setHost(host);
+        cfp.setPort(port);
+        cfp.setUsername(username);
+        cfp.setPassword(password);
+        //create the factory with connection info
+        try {
+            clientFactory = ApplicationClientFactory.create(HonoConnection.newConnection(vertx, cfp));
```
sophokles73, line 75: this should not only create the connection factory but should actually establish the connection. The connection can then be held in a field and can be used to determine the availability state for the commands (*HonoConnection.isConnected()*). FMPOV there is no reason for connecting/disconnecting for every command as there also is a certain overhead involved ... Plus, the methods start actually doing what their names suggest ;-)

johnMinelli: FMPOV is the best approach since for slow/instable connection you can stop the execution of commands (without closing the client) still having stale messages in/outgoing that can make noise in the next execution. So the overhead is justified

sophokles73: The execution of a command can be stopped anyway, regardless of whether the connection is already established or is being established as part of the command.
*Execution of a command* means sending an AMQP frame and (often) receiving an AMQP frame in response. I can always stop waiting for the response to arrive. There is no need to establish a new connection for every command ...

cli/src/main/java/org/eclipse/hono/cli/AbstractCliClient.java
```diff
@@ -44,32 +52,81 @@
...
+
+    /**
+     * Check the availability of the factory.
+     * If not present it doesn't allow the execution of commands.
+     *
+     * @return Availability status.
+     */
+    protected Availability connectionAvailability() {
+        return clientFactory != null
```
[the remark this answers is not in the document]
johnMinelli: isConnected() return a Future<void> how can i use it rightly for a check?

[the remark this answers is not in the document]
sophokles73: you'll need to use a `CompletableFuture` and link its completion to the `Future<Void>` returned by *isConnectect()*
```
final CompletableFuture<Void> result = new CompletableFuture<>();
isConnected()
  .onSuccess(ok -> result.complete(null))
  .onFailure(t -> result.completeExceptionally(t));
return result.join();
```

cli/src/main/java/org/eclipse/hono/cli/AbstractCliClient.java
```diff
@@ -44,32 +54,91 @@
...
+     * Spring Shell output helper.
+     * <p>
+     * it can be passed to the method class instead to use the log (it's more stylish).
+     */
+    @Autowired
+    protected ShellHelper shellHelper;
+    /**
+     * The client factory methods.
+     */
+    protected ConnectionLifecycle<HonoConnection> clientFactory;
```
sophokles73, line 83: IMHO we should parameterize the `AbstractCliClient` class with the type of client factory that it supports.
That way we would also get rid of the ugly cast that is done in the subclasses when a command is being executed.

cli/src/main/java/org/eclipse/hono/cli/application/Receiver.java
```diff
@@ -0,0 +1,131 @@
...
+            if (connectionStatus.succeeded()) {
+                clientFactory.addReconnectListener(this::createConsumer);
+                createConsumer(con);
+            } else {
+                close(connectionStatus.cause());
+            }
+        });
+    }
+
+    private CompositeFuture createConsumer(final HonoConnection connection) {
```
sophokles73, line 84: what is the connection parameter good for?

cli/src/main/java/org/eclipse/hono/cli/application/Commander.java
```diff
@@ -1,66 +1,95 @@
...
+    public void start(final CountDownLatch latch, final HonoConnection con) {
+        this.latch = latch;
         workerExecutor = vertx.createSharedWorkerExecutor("user-input-pool", 3, TimeUnit.HOURS.toNanos(1));
-        clientFactory.connect().setHandler(connectAttempt -> {
-            if (connectAttempt.succeeded()) {
+        clientFactory.isConnected().setHandler(connectionStatus -> {
+            if (connectionStatus.succeeded()) {
                 clientFactory.addReconnectListener(this::startCommandClient);
-                startCommandClient(connectAttempt.result());
+                startCommandClient(con);
```
sophokles73, line 90: what is the connection parameter good for?

calohmn:
@johnMinelli Can you rebase? 

The conflict in the `Commander` class can be resolved by adopting your changes and then just replacing `.setHandler(` with `.onComplete(`.

Regarding the `Receiver[*]` classes: 
grafik
It looks to me you have moved the `Receiver` class to a different package and changed it. I think you should adopt the changes from the commit https://github.com/eclipse/hono/commit/c7bb1e579b0f26126b53edc1e413a4752d6ee282 to your `Receiver` class and also adopt the `ReceiverStatistics` class into your branch.
Can you check whether also to adopt the `ReceiverTest` class? I see it isn't there anymore in your branch.

cli/src/main/java/org/eclipse/hono/cli/application/Receiver.java
johnMinelli: since the factory has a paramenter

` void addReconnectListener(ReconnectListener<T> listener);`

johnMinelli:
> The conflict in the `Commander` class can be resolved by adopting your changes and then just replacing `.setHandler(` with `.onComplete(`.

When the receiver is called the connection is already enstablished so i can't just replace.

However i'll add the statistics profile

johnMinelli:
Also @calohmn how the `ReceiverStatistics` constructor is supposed to work? I haven't tried but @Autowired without a Receiver @Bean how it get the instance?

calohmn:
>> The conflict in the `Commander` class can be resolved by adopting your changes and then just replacing `.setHandler(` with `.onComplete(`.

> When the commander/receiver is called the connection is already enstablished so i can't just replace.

This replacement doesn't involve a behaviour change. `io.vertx.core.Future#setHandler` does the same thing as `io.vertx.core.Future#onComplete`. It's just that the `setHandler` is marked deprecated in the next vert.x version. Therefore this replacement was done.

> Also @calohmn how the `ReceiverStatistics` constructor is supposed to work? I haven't tried but `@Autowired` without a Receiver `@Bean` how it get the instance?

In the master branch, the `Receiver` class has a `@Component` annotation, so there is a bean created from it. This is then autowired as the `ReceiverStatistics` constructor param.

cli/pom.xml
```diff
@@ -11,125 +11,122 @@
    
     SPDX-License-Identifier: EPL-2.0
  -->
-<project xmlns="http://maven.apache.org/POM/4.0.0"
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-  <modelVersion>4.0.0</modelVersion>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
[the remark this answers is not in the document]
sophokles73: I was talking about the whole document ... :-)

cli/src/main/java/org/eclipse/hono/cli/AbstractCliClient.java
sophokles73: Any reason why you do not want to parameterize the `AbstractCliClient` class with the client factory type?

cli/src/main/java/org/eclipse/hono/cli/AbstractCliClient.java
```diff
@@ -44,32 +54,91 @@
...
+        }
+    }
+
+    /**
+     * Check the availability of the factory.
+     * If not present it doesn't allow the execution of commands.
+     *
+     * @return Availability status.
+     */
+    protected Availability connectionAvailability() {
```
sophokles73, line 127: If I am not mistaken, then this method will operate on the *clientFactory* field defined by this abstract class. Thus, invoking it from any of its subclasses will not yield the desired effect of checking the status of the factory kept in the *clientFactory* fields defined in those subclasses.

PR #1812, opened 2020-03-05
Add tracer instance in HotrodBasedDeviceConnectionInfo

calohmn:
A Tracer instance is now available in `HotrodBasedDeviceConnectionInfo`, making it possible to create extra spans in the `DeviceConnectionInfo` method implementations. 
This will be extra useful for the additional Device Connection API methods to be created for #1272.

### DIFF
--- a/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCacheConfig.java
+++ b/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCacheConfig.java
@@ -13,6 +13,8 @@
 
 package org.eclipse.hono.deviceconnection.infinispan.client;
 
+import java.util.Optional;
+
 import org.eclipse.hono.client.BasicDeviceConnectionClientFactory;
 import org.eclipse.hono.util.DeviceConnectionConstants;
 import org.infinispan.client.hotrod.RemoteCacheManager;
@@ -22,6 +24,8 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
+import io.opentracing.Tracer;
+import io.opentracing.noop.NoopTracerFactory;
 import io.vertx.core.Vertx;
 
 /**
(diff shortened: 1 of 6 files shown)


client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCacheConfig.java
```diff
@@ -79,12 +83,15 @@ public RemoteCacheManager remoteCacheManager() {
...
      * @return The factory.
      */
     @Bean
     @Qualifier(DeviceConnectionConstants.DEVICE_CONNECTION_ENDPOINT)
     @ConditionalOnProperty(prefix = "hono.device-connection", name = "server-list")
-    public BasicDeviceConnectionClientFactory hotrodBasedDeviceConnectionClientFactory(final HotrodCache<String, String> cache) {
-        return new HotrodBasedDeviceConnectionClientFactory(cache);
+    public BasicDeviceConnectionClientFactory hotrodBasedDeviceConnectionClientFactory(
+            final HotrodCache<String, String> cache, final Optional<Tracer> tracer) {
+        return new HotrodBasedDeviceConnectionClientFactory(cache, tracer.orElse(NoopTracerFactory.create()));
```
sophokles73: this will throw a `NullPointerException` if *tracer* is `null`, right? Is there actually a factory method producing an `Optional<Tracer>`?

calohmn: The method is supposed to be called by Spring.
Using `Optional` here is a way to tell Spring that such a `Tracer` bean is optional. In that sense it is an equivalent to using `@Autowired(required = false)` on a field or method (using Autowired on a method parameter is mostly ignored, as written [here](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/annotation/Autowired.html). Another alternative would be to use a `@Nullable` annotation, which I also see as a good solution, which would require an additional dependency though (`com.google.code.findbugs:jsr305` I guess we haven't discussed usage of that yet).

sophokles73: Ah, I learn something new every day :-) In any case, IMHO you should document the `NullPointerException` ...

PR #1809, opened 2020-03-04
Refactor Tenant AMQP endpoint not to use event bus

dejanb:
This is a first stub at creating a base and tenant AMQP endpoint implementation that doesn't use event bus. It's mostly based on the existing implementation. There are things I would improve, but I think it's a solid base we can continue working on (implement other services and improve things on the go).

PR #1813, opened 2020-03-05
Refactor credentials AMQP endpoint not to use event bus

jbtrystram:
This PR builds on @dejanb PR #1809, following the same model but with the credentials endpoint.


dejanb:
Looks good in general. Let's revisit when #1809 is merged as there are a lot of overlap.

dejanb:
@jbtrystram can you rebase please?

dejanb:
This has been merged manually due to the change in import order.

PR #1818, opened 2020-03-06
[#1272] Add docs for new Device Connection API methods

calohmn:
This is for #1272.


site/documentation/content/api/device-connection/index.md
```diff
@@ -107,6 +107,176 @@ Implementors of this API may return a *404* status code in order to indicate tha
...
+**Response Message Format**
+
+A response to a *set command-handling adapter instance for device* request contains the [Standard Response Properties]({{< relref "#standard-response-properties" >}}).
+
+The response message's *status* property may contain the following codes:
+
+| Code  | Description |
+| :---- | :---------- |
+| *204* | OK, the adapter instance mapping information for the device has been removed. |
+| *400* | Bad Request, the command-handling adapter adapter instance information has not been updated due to there being no such mapping entry or due to missing data in the request. |
```
sophokles73: I do not quite understand what
> has not been updated due to there being no such mapping entry 

means. The next paragraph describes that a `412` could be returned if the *device_id* is currently not mapped to the *adapter_instance_id* from the request. Is this the same case?

site/documentation/content/api/device-connection/index.md
```diff
@@ -107,6 +107,176 @@ Implementors of this API may return a *404* status code in order to indicate tha
...
+
+Implementors of this API may return a *404* status code in order to indicate that no device with the given identifier exists for the given tenant or that no adapter instance is associated with the device. Furthermore, in case the currently mapped adapter instance doesn't match the given adapter instance id, a *412* status code may be returned. However, performing any of these checks is optional.
+
+For status codes indicating an error (codes in the `400 - 499` range) the message body MAY contain a detailed description of the error that occurred.
+
+## Get command-handling protocol adapter instances for device
+
+Clients use this command to *get* information about the adapter instances that can handle command & control messages for the given device, possibly using one of the given gateways.
```
sophokles73: IMHO we should omit
> , possibly using one of the given gateways.

here because it is unclear where the *given gateways* come from and it is not important to mention them here in order for the reader to understand the purpose of this operation.

site/documentation/content/api/device-connection/index.md
```diff
@@ -107,6 +107,176 @@ Implementors of this API may return a *404* status code in order to indicate tha
...
+
+As part of this request, the client is supposed to provide the list of gateways that may act on behalf of the given device. This API doesn't mandate checks on the validity of the given device and the gateway list in order not to introduce a dependency on the *Device Registration API*.
```
sophokles73: As part of *the* request message ...?

IMHO a hint would be useful as to where the client might get the *list of gateways that may act on behalf of the given device* ...

site/documentation/content/api/device-connection/index.md
```diff
@@ -107,6 +107,176 @@ Implementors of this API may return a *404* status code in order to indicate tha
...
+
+The command implementation MUST determine the adapter instances by applying the following rules (in the given order):
+- If an adapter instance is associated with the given device, this adapter instance is returned as the *single* returned list entry.
```
sophokles73: maybe use numbering if order is important?

site/documentation/content/api/device-connection/index.md
```diff
@@ -107,6 +107,176 @@ Implementors of this API may return a *404* status code in order to indicate tha
...
+{{< figure src="get_cmd_handling_adapter_instances.svg" title="Client retrieving the list of command-handling adapter instances for a device" alt="A client sends a request message for retrieving the list of command-handling adapter instances for a device and receives a response containing the information" >}}
+
+**Request Message Format**
+
+The following table provides an overview of the properties a client needs to set on a message to retrieve the command-handling adapter instances for a device in addition to the [Standard Request Properties]({{< relref "#standard-request-properties" >}}).
+
+| Name           | Mandatory | Location                 | AMQP Type | Description |
+| :------------- | :-------: | :----------------------- | :-------- | :---------- |
+| *subject*      | yes       | *properties*             | *string*  | MUST be set to `get-cmd-handling-adapter-instances`. |
+| *gateway_ids*  | yes       | *properties*             | *string*  | A JSON array containing the IDs of the gateways that may act on behalf of the given device. |
```
sophokles73: FMPOV we should put a JSON array containing the ids into the body of the request message ...
Otherwise, the location needs to be *application-properties* ...

site/documentation/content/api/device-connection/index.md
```diff
@@ -107,6 +107,176 @@ Implementors of this API may return a *404* status code in order to indicate tha
...
+| Name                | Mandatory | JSON Type     | Description |
+| :------------------ | :-------: | :------------ | :---------- |
+| *adapter-instances* | *yes*     | *array*       | A non-empty array of JSON objects that represent the command-handling adapter instances along with the device or gateway id. |
+
+Each entry in the *adapter-instances* array has the following properties:
+
+| Name                  | Mandatory | JSON Type     | Description |
+| :-------------------- | :-------: | :------------ | :---------- |
+| *adapter-instance-id* | *yes*     | *string*      | The ID of the protocol adapter instance handling command & control messages for the device given in the *device-id* property. |
+| *device-id*           | *yes*     | *string*      | The ID of the gateway or device that the protocol adapter instance given by the *adapter-instance-id* is handling command & control messages for. |
```
sophokles73: Based on the examples given below, this doesn't seem to necessarily be the *device_id* that has been passed in with the request message. A hint would be nice that this is indeed the case and why the client would actually care about this ID.

site/documentation/content/api/device-connection/index.md
```diff
@@ -107,6 +107,176 @@ Implementors of this API may return a *404* status code in order to indicate tha
...
+  ]
+}
+~~~
+
+
+The response message's *status* property may contain the following codes:
+
+| Code | Description |
+| :--- | :---------- |
+| *200* | OK, the payload contains the gateway ID. |
```
sophokles73: it contains the adapter instances ...

site/documentation/content/api/device-connection/index.md
```diff
@@ -107,6 +107,176 @@ Implementors of this API may return a *404* status code in order to indicate tha
...
+| *400* | Bad Request, the request message does not contain all required information/properties. |
+| *404* | Not Found, there is no command-handling adapter instance assigned to the device. |
+
+Implementors of this API may return a *404* status code in order to indicate that no device with the given identifier exists for the given tenant. However, performing such a check is optional.
```
sophokles73: the table above seems to indicate that this check is actually mandatory for an implementation to perform ...

site/documentation/content/api/device-connection/index.md
calohmn: > Otherwise, the location needs to be *application-properties* ...

Right, that was a copy/paste error here.

Regarding using the request message instead: I had this in my initial commit here, but I think putting the operation parameters in the (application-) properties of the request message is more commonly used in our APIs.

site/documentation/content/api/device-connection/index.md
calohmn: Yes, I have rephrased the sentence above.

site/documentation/content/api/device-connection/index.md
calohmn: The check mentioned in the table is a different one, actually.
The entry in the table makes it mandatory to return a 404 if not command-handling instance is found.
The extra hint below states that the implementation may first check if the given tenant actually has the given device. That would make for a more insightful error message if there was the wrong device id being used. However, this is probably a somewhat theoretical case since a client would usually invoke the "assert device registration" method shortly before and would get that error there already.

site/documentation/content/api/device-connection/index.md
```diff
@@ -107,6 +107,178 @@ Implementors of this API may return a *404* status code in order to indicate tha
...
+**Response Message Format**
+
+A response to a *set command-handling adapter instance for device* request contains the [Standard Response Properties]({{< relref "#standard-response-properties" >}}).
+
+The response message's *status* property may contain the following codes:
+
+| Code  | Description |
+| :---- | :---------- |
+| *204* | OK, the adapter instance for the device has been updated. |
+| *400* | Bad Request, the command-handling adapter adapter instance has not been updated due to invalid or missing data in the request. |
```
sophokles73: adapter adapter

site/documentation/content/api/device-connection/index.md
```diff
@@ -107,6 +107,178 @@ Implementors of this API may return a *404* status code in order to indicate tha
...
+**Response Message Format**
+
+A response to a *set command-handling adapter instance for device* request contains the [Standard Response Properties]({{< relref "#standard-response-properties" >}}).
+
+The response message's *status* property may contain the following codes:
+
+| Code  | Description |
+| :---- | :---------- |
+| *204* | OK, the adapter instance mapping information for the device has been removed. |
+| *400* | Bad Request, the command-handling adapter adapter instance information has not been updated due to missing data in the request or because there is no mapping associating the given tenant and device with the given adapter instance. |
```
sophokles73: > or because there is no mapping associating the given tenant and device with the given adapter instance

Isn't this exactly the cause that would lead to a *412* being returned as described in the next paragraph?

site/documentation/content/api/device-connection/index.md
```diff
@@ -107,6 +107,178 @@ Implementors of this API may return a *404* status code in order to indicate tha
...
+{{< figure src="get_cmd_handling_adapter_instances.svg" title="Client retrieving the list of command-handling adapter instances for a device" alt="A client sends a request message for retrieving the list of command-handling adapter instances for a device and receives a response containing the information" >}}
+
+**Request Message Format**
+
+The following table provides an overview of the properties a client needs to set on a message to retrieve the command-handling adapter instances for a device in addition to the [Standard Request Properties]({{< relref "#standard-request-properties" >}}).
+
+| Name           | Mandatory | Location                 | AMQP Type | Description |
+| :------------- | :-------: | :----------------------- | :-------- | :---------- |
+| *subject*      | yes       | *properties*             | *string*  | MUST be set to `get-cmd-handling-adapter-instances`. |
+| *gateway_ids*  | yes       | *application-properties* | *string*  | A JSON array containing the IDs of the gateways that may act on behalf of the given device. |
```
sophokles73: I still believe that we should rather put complex context into the body instead ...
We do this in other cases as well, e.g. https://www.eclipse.org/hono/docs/api/credentials/#get-credentials

site/documentation/content/api/device-connection/index.md
```diff
@@ -107,6 +107,178 @@ Implementors of this API may return a *404* status code in order to indicate tha
...
+| :--------------- | :-------: | :----------------------- | :----------- | :---------- |
+| *content-type*   | no        | *properties*             | *string*     | MUST be set to `application/json` if the invocation of the operation was successful and the body of the response message contains payload as described below. |
+
+The result of a successful invocation is carried in a single Data section of the response message as a UTF-8 encoded string representation of a single JSON object. It is an error to include payload that is not of this type.
+
+The response message JSON object has the following properties:
+
+| Name                | Mandatory | JSON Type     | Description |
+| :------------------ | :-------: | :------------ | :---------- |
+| *adapter-instances* | *yes*     | *array*       | A non-empty array of JSON objects that represent the command-handling adapter instances along with the device or gateway id. |
```
sophokles73: wouldn't it be sufficient to just put an array into the response body, i.e. omit the *adapter-instances* property?

site/documentation/content/api/device-connection/index.md
sophokles73: I see, then what about starting like
> Implementors of this API may *also* return a *404* status ...

site/documentation/content/api/device-connection/index.md
calohmn: From a purely API standpoint, yes.
Introducing an JsonArray result would mean that there are different result types for the Device Connection API operations. From an implementation perspective, this means there cannot be just one `DeviceConnectionResult` object used for all methods (extending `RequestResponseResult<JsonObject>`), but there would have to be another `DeviceConnectionArrayResult` (extending `RequestResponseResult<JsonArray>`). 

But, thinking more about it, it is indeed a bit of a crutch, this extra "adapter-instances" field and there probably won't be any other fields, getting added on the same level (like some `last-updated` field for example; wouldn't make sense there).
So, yes, I can change it.

sophokles73: I was just wondering. One good reason to keep the JSON obect based result would indeed be to keep it extensible. Maybe we want to add additional properties sometime later? I do see the drawback of having to create another result class so I'll leave it up to you whether you want to change it or not :-)

site/documentation/content/api/device-connection/index.md
calohmn: Well, I don't have a strong opinion here, I can change it back to using the request body.

site/documentation/content/api/device-connection/index.md
calohmn: Yes, I'll rephrase that, putting 404 in the table and adding a hint below that an implementor *may* also return 412 in the above case instead of 404.
I don't want to make that distinction mandatory, because in an implementation, underlying method calls like `Map.remove(key, value)` may not provide that information.

site/documentation/content/api/device-connection/index.md
calohmn: Ok, extensibility and a common `DeviceConnectionResult` - I'll keep it as a JSON object.

site/documentation/content/api/device-connection/index.md
```diff
@@ -107,6 +107,192 @@ Implementors of this API may return a *404* status code in order to indicate tha
...
+| Name                  | Mandatory | Location                 | AMQP Type | Description |
+| :-------------------- | :-------: | :----------------------- | :-------- | :---------- |
+| *subject*             | yes       | *properties*             | *string*  | MUST be set to `remove-cmd-handling-adapter-instance`. |
+| *adapter_instance_id* | yes       | *application-properties* | *string*  | The identifier of the protocol adapter instance to remove the mapping information for. Only if this adapter instance is currently associated with the device or gateway identified by the *device_id* property, the mapping entry will be removed. |
+
+The body of the message SHOULD be empty and will be ignored if it is not.
+
+**Response Message Format**
+
+A response to a *set command-handling adapter instance for device* request contains the [Standard Response Properties]({{< relref "#standard-response-properties" >}}).
```
sophokles73: response to a *remove command-handling adapter* ...

calohmn: Fixed.

PR #1826, opened 2020-03-10
[#1802] Refactor device connection service

kaniyan:
In this PR, the `DeviceConnectionService` is refactored to return _vertx_ future and also the related classes have been adapted. Also the release notes has been updated for #1802.

### DIFF
--- a/site/homepage/content/release-notes.md
+++ b/site/homepage/content/release-notes.md
@@ -13,11 +13,14 @@ title = "Release Notes"
 
 ### API Changes
 
- * The device registry credentials endpoint will no longer handout sensitive details for hashed-password secrets.
-   `pwd-hash`, `salt` and `hash-function` are stored by the device registry but not returned to the user.
-   Each secret is given an ID which is now returned, along with the other metadata (time validity and optional fields).
+* The device registry credentials endpoint will no longer handout sensitive details for hashed-password secrets.
+  `pwd-hash`, `salt` and `hash-function` are stored by the device registry but not returned to the user.
+  Each secret is given an ID which is now returned, along with the other metadata (time validity and optional fields).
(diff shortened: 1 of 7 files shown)


site/homepage/content/release-notes.md
```diff
@@ -11,6 +11,13 @@ title = "Release Notes"
   This has the advantage of saving the network hop to the Device Connection service.
   Refer to the protocol adapter Admin Guides for details.
 
+### Fixes & Enhancements
+
+* The service base classes for implementing device registry namely `CredentialsManagementService`,
+  `CredentialsService`, `DeviceConnectionService`, `DeviceManagementService`, `RegistrationService`,
+  `TenantManagementService`and `TenantService` have been refactored now to return a Vertx future instead of 
+   taking an asynchronous handler as an argument.
```
calohmn: I think that fits better in the "API Changes" section.

I would omit the "for implementing device registry namely" part and would use 
"refactored to now return a Vert.x `Future` instead of taking a `Handler` as an argument"
in the last part.
And I would prepend "The methods of [the service base classes]".

kaniyan: @calohmn I have modified it and could you take a look below before I push the changes?

The methods of the service base classes `CredentialsManagementService`, `CredentialsService`, `DeviceConnectionService`, `DeviceManagementService`, `RegistrationService`, `TenantManagementService` and  `TenantService` are refactored now to return a Vert.x Future instead of taking a Handler as an argument.

calohmn: How about: "have been refactored to now return"?

kaniyan: Sounds good. I have modified and pushed the changes.

PR #1828, opened 2020-03-11
[#1272] Implement new Device Connection API methods

calohmn:
This is for #1272:
Adds an implementation for the new Device Connection API methods to the Infinispan client.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -546,6 +546,12 @@
         <version>${infinispan.version}</version>
       </dependency>
       <dependency>
+        <groupId>org.infinispan</groupId>
+        <artifactId>infinispan-query-dsl</artifactId>
+        <version>${infinispan.version}</version>
+        <scope>test</scope>
+      </dependency>
+      <dependency>
         <groupId>com.fasterxml.jackson.datatype</groupId>
         <artifactId>jackson-datatype-jsr310</artifactId>
         <version>${jackson.version}</version>
--- a/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/Versioned.java
+++ b/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/Versioned.java
@@ -1,0 +1,45 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.deviceconnection.infinispan.client;
+
+/**
+ * A versioned entry.
+ * 
+ * @param <T> The payload type.
+ */
+public final class Versioned<T> {
+
+    private final long version;
+    private final T value;
+
+    /**
+     * Created a new versioned entry.
+     * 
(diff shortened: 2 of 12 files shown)


client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/Versioned.java
```diff
@@ -0,0 +1,45 @@
...
+ *******************************************************************************/
+
+package org.eclipse.hono.deviceconnection.infinispan.client;
+
+/**
+ * A versioned entry.
+ * 
+ * @param <T> The payload type.
+ */
+public class Versioned<T> {
```
sophokles73: final ?

bom/pom.xml
```diff
@@ -545,6 +545,11 @@
         <artifactId>infinispan-client-hotrod</artifactId>
         <version>${infinispan.version}</version>
       </dependency>
+      <dependency>
+        <groupId>org.infinispan</groupId>
+        <artifactId>infinispan-query-dsl</artifactId>
```
sophokles73: what do we need this for?

client-device-connection-infinispan/src/test/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCacheTest.java
```diff
@@ -152,10 +161,149 @@ void testPutFails(final VertxTestContext ctx) {
...
+                    ctx.completeNow();
+                }));
+    }
+
+    private org.infinispan.client.hotrod.RemoteCache<Object, Object> givenAConnectedCache() {
+        final Configuration configuration = mock(Configuration.class);
         @SuppressWarnings("unchecked")
-        final BasicCache<Object, Object> result = mock(BasicCache.class);
-        when(remoteCacheManager.getCache(anyString())).thenReturn(result);
+        final org.infinispan.client.hotrod.RemoteCache<Object, Object> result = mock(org.infinispan.client.hotrod.RemoteCache.class);
```
calohmn: Since the `HotrodCache` class uses a `org.infinispan.client.hotrod.RemoteCache` field now (instead of `org.infinispan.commons.api.BasicCache`), this mock had to be changed.
Unfortunately, the `org.infinispan.client.hotrod.RemoteCache` has a dependency on the `infinispan-query-dsl` module (for the `retrieveEntriesByQuery` method) and without having that module integrated, creating the mock here will fail.
Therefore I've added `infinispan-query-dsl` as a test-only dependency (CQ still needed for that). As alternative option, I only see removing this test class.

bom/pom.xml
calohmn: see my other comment below

sophokles73: IMHO we should define the scope here as `test` already to make it clear that we use this for testing only

calohmn: Ok, I've added the scope here.

services/device-connection/src/test/java/org/eclipse/hono/deviceconnection/infinispan/RemoteCacheBasedDeviceConnectionServiceTest.java
```diff
@@ -139,4 +139,28 @@ public void testGetLastKnownGatewayForDeviceNotFound(final VertxTestContext ctx)
             ctx.completeNow();
         }));
     }
+
+    /**
+     * Verifies that the last known gateway id can be set via the <em>setCommandHandlingAdapterInstance</em> operation.
```
sophokles73: the last known gateway id ?

calohmn: Fixed.

PR #1830, opened 2020-03-11
Extend tracing with provsion to accept message and error as arguments

kaniyan:
During the implementation of _mongodb_ based device registry, I came across the below tracing helper methods that takes either a _message_ or an _error_ as an argument. 
```
logError(final Span span, final String message)
logError(final Span span, final Throwable error)
```
In some cases, I prefer to log both the custom message and also the thrown exception. Hence as part of this PR, I have added the below method that takes both _message_ and _error_ as arguments and then logs them.
 
`logError(final Span span, final String message, final Throwable error)`

### DIFF
--- a/core/src/main/java/org/eclipse/hono/tracing/TracingHelper.java
+++ b/core/src/main/java/org/eclipse/hono/tracing/TracingHelper.java
@@ -173,6 +173,31 @@ public final class TracingHelper {
     }
 
     /**
+     * Marks an <em>OpenTracing</em> span as erroneous, logs a message and an error.
+     * <p>
+     * This method does <em>not</em> finish the span.
+     *
+     * @param span The span to mark.
+     * @param message The message to log on the span.
+     * @param error The error to log on the span.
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/eclipse/hono/tracing/TracingHelper.java
```diff
@@ -159,6 +159,28 @@ public static void logError(final Span span, final String message) {
...
 
+    /**
+     * Marks an <em>OpenTracing</em> span as erroneous, logs a message and an error.
+     * <p>
+     * This method does <em>not</em> finish the span.
+     *
+     * @param span The span to mark.
+     * @param message The message to log on the span.
+     * @param error The error to log on the span.
+     * @throws NullPointerException if message or error is {@code null}.
```
sophokles73: wouldn't it make more sense to allow any one of them to be `null` and then just log the non-`null` argument?

kaniyan: :+1:  sounds good.

PR #1831, opened 2020-03-12
Refactor Device Connection AMQP endpoint not to use event bus

dejanb:


service-base/src/main/java/org/eclipse/hono/service/deviceconnection/AbstractDeviceConnectionAmqpEndpoint.java
```diff
@@ -0,0 +1,331 @@
...
+                            HttpURLConnection.HTTP_BAD_REQUEST,
+                            "request message body contains malformed JSON"));
+        }
+
+        final Span span = TracingHelper.buildServerChildSpan(
+                tracer,
+                TracingHelper.extractSpanContext(tracer, request),
+                SPAN_NAME_GET_CMD_HANDLING_ADAPTER_INSTANCES,
+                getClass().getSimpleName()
+        ).start();
```
calohmn: How about creating the span before parsing the payload - then that error can be logged in the trace span.
See also https://github.com/eclipse/hono/commit/8f2958f6ba175f605ea13e431ba1efba8bed5f7f.

dejanb: @calohmn  Good catch. Fixed!

PR #1832, opened 2020-03-12
[#1679] Extract utility methods for reuse in device registry implementations

kaniyan:
While working on #1679, I noticed that these methods in this PR could be reused among various device registry implementations and hence extracted them to a utility class in the device registry base module.


kaniyan:
> Looks good in general. Just thinking as all these methods are related to the `Tenant` object, we could add them (with a bit of modification) there instead of using the separate util class. WDYT?

@Dejan It sounds good. Except the method `getCacheDirective(...)` , the others we could move it to the `org.eclipse.hono.service.management.tenant.Tenant` class. With some modification I imagine something like `public JsonObject convertTenant(final String tenantId)` in `Tenant` class. What I find it bit strange here is that we have to pass the _tenantId_ as _Tenant_ class doesn't have _tenantId_. WDYT?  IMHO since we focus now on bringing out the device registry implementation asap, we could also address it in a separate PR later.

dejanb:
@kaniyan absolutely. Something to think about for the future.

PR #1834, opened 2020-03-13
Remove unused constants.

kaniyan:
The unused constants in `FileBasedTenantBackend` have been removed.


dejanb:
@kaniyan I think these should live in `RegistryManagementConstants` as `FIELD_TENANT` and `FIELD_DEVICES`. These two particularly are not used, but there are the same ones in underlying classes (e.g. `FileBasedRegistrationService`), so it'd be good to clean that all up with this PR.

kaniyan:
> 
> 
> @kaniyan I think these should live in `RegistryManagementConstants` as `FIELD_TENANT` and `FIELD_DEVICES`. These two particularly are not used, but there are the same ones in underlying classes (e.g. `FileBasedRegistrationService`), so it'd be good to clean that all up with this PR.

@dejanb Sure, I will take a look and  make changes.

PR #1835, opened 2020-03-13
Provide classes that can be used as unique keys for device registry s…

dejanb:
…torage implementations

### DIFF
--- a/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/CredentialKey.java
+++ b/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/CredentialKey.java
@@ -1,0 +1,130 @@
+/*******************************************************************************
+ * Copyright (c) 2019, 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.deviceregistry.service.credentials;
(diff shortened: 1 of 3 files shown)


dejanb:
@ctron these are distilled from your original PR. I tried to align them and make them look like the similar classes already used in Hono. Can you check if I didn't miss something. Note that `TenantHandle` is now `TenantKey`

@kaniyan Do you have a need for this concept in your implementation. And can you use these classes if so.

I'll continue working on the abstract storage layer and tenant validation service.

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/CredentialKey.java
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Creates a credential key from tenant identifier, authentication identifier and type.
+     *
+     * @param tenantId The tenant identifier.
+     * @param authId The authentication identifier.
+     * @param type The credential type.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     * @return The credential key.
+     */
+    public static CredentialKey from(final String tenantId, final String authId, final String type) {
```
ctron: I would recommend to name the method `credentialKey` instead of `from`. The reason for that is, when using static imports, you will get a name clash with the other `from` methods.

dejanb: Static imports are treated as errors by checkstyle plugin, so we can use them. And we have pattern of using `from()` methods in these kind of value objects throughout the codebase. Also, there was also `of()` in `TenantHandle`, so it should be consistent now.

kaniyan:
> @kaniyan Do you have a need for this concept in your implementation. And can you use these classes if so.

@dejanb Its good to have these classes. Currently I have dto classes that hold required keys in addition to the content. In the next iterations, I will check how we could make use of these classes.

dejanb:
@kaniyan great. Let me know if you see anything missing.

PR #1837, opened 2020-03-13
Add coap user-guide.

boaks:
Part 1, telemetry.

### DIFF
--- a/site/documentation/content/user-guide/coap-adapter.md
+++ b/site/documentation/content/user-guide/coap-adapter.md
@@ -1,0 +1,230 @@
++++
+title = "CoAP Adapter"
+weight = 225
++++
+
+The CoAP protocol adapter exposes [CoAP](https://tools.ietf.org/html/rfc7252) based endpoints for Eclipse Hono&trade;'s south bound Telemetry, Event and Command & Control APIs.
+<!--more-->
+
+## Device Authentication
+
+The CoAP adapter by default requires clients (devices or gateway components) to authenticate during connection establishment. The adapter supports currently only [PSK](https://tools.ietf.org/html/rfc4279) as part of a DTLS handshake for that purpose. The other variants mentioned in [Securing CoAP](https://tools.ietf.org/html/rfc7252#section-9) may follow in the future.
+
+The adapter tries to authenticate the device using this mechanisms
+
+### PSK
+
+The *identity* provided in the ClientKeyExchange must have the form *auth-id@tenant*, e.g. `sensor1@DEFAULT_TENANT`. The adapter performs the handshake using the credentials the `configured Credentials service - (documentation pending)` has one record for the client. The adapter uses the Credentials API's *get* operation to retrieve the credentials on record with the *tenant* and *auth-id* provided by the device in the *identity* and `psk` as the *type* of secret as query parameters.
+
+The examples below refer to devices `4711` and `gw-1` of tenant `DEFAULT_TENANT` using *auth-ids* `sensor1` and `gw1` and corresponding passwords. The example deployment as described in the [Deployment Guides]({{< relref "deployment" >}}) comes pre-configured with the corresponding entities in its device registry component.
+Please refer to the [Credentials API]({{< relref "/api/credentials#standard-credential-types" >}}) for details regarding the different types of secrets.
+
+**NB** There is a subtle difference between the *device identifier* (*device-id*) and the *auth-id* a device uses for authentication. See [Device Identity]({{< relref "/concepts/device-identity.md" >}}) for a discussion of the concepts.
+
+## Message Limits
+
+Before accepting any telemetry or event or command messages, the CoAP adapter verifies that the configured [message limit]({{< relref "/concepts/resource-limits#messages-limit" >}}) is not exceeded.
+For CoAP currently only the messages limit is verified, the connections limit is not available. If the limit is exceeded then the incoming message is discarded with the status code `429 Too Many Requests`. 
+
+## CoAP Content Formats
+
+CoAP doesn't use a textual identifier for content types. Instead numbers are used, which are maintained by the [IANA](https://www.iana.org/).
+For the currently assigned numbers, please check [IANA - CoAP Content Formats](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#content-formats).
+Content types that do not map to one of the registered CoAP content formats are ignored.
+
+## Publish Telemetry Data (authenticated Device)
+
+The device is authenticated using PSK.
+
+* URI: `/telemetry`
+* Method: `POST`
+* Type:
+  * `CON`: *at least once* (`1`) QoS levels
+  * `NON`: *at most once* (`0`) QoS levels
+* Request Options:
+  * (optional) `content-format`: The type of payload contained in the request body. Required, if request contains payload.
+  * (optional) `URI-query: hono-ttd`: The number of seconds the device will wait for the response.
+  * (optional) `URI-query: empty`: If the request doesn't contain payload.
+* Request Body:
+  * (optional) Arbitrary payload encoded according to the given content type. Maybe empty, if `URI-query: empty` is provided.
+* Response Options:
+  * (optional) `content-format`: A media type describing the semantics and format of payload contained in the response body. This option will only be present if the response contains a command to be executed by the device which requires input data. If the command's content type do not map to one of the registered CoAP content formats, the option will not be present.
+  * (optional) `location-query: hono-command`: The name of the command to execute. This option will only be present if the response contains a command to be executed by the device.
+  * (optional) `location-path: ('command'|'command_response/'<command-request-id>)`: The location path contains `'command'` for one-way-commands, or `'command_response' / <command-request-id>`  for commands expecting  a response. In case of a expected response, the location-path contains exactly the URI-path expected for the response.
+* Response Body:
+  * (optional) Arbitrary data serving as input to a command to be executed by the device, if status code is 2.05 (Content).
+  * (optional) Error details, if status code is >= 4.00.
+* Status Codes:
+  * 2.04 (Changed): The telemetry data has been accepted for processing. Note that if the message type is `NON` (*at most once* semantics),
+    this status code does **not** mean that the message has been delivered to any potential consumer. However, if the message type is `CON` (*at least once* semantics), then the adapter waits for the message to be delivered and accepted by a downstream consumer before responding with this status code.
+  * 2.05 (Content): The telemetry data has been accepted for processing. The response contains a command for the device to execute.
+  * 4.03 (Forbidden): The request cannot be processed because the device's registration status cannot be asserted. Possible reasons for this include:
+        * The given tenant is not allowed to use this protocol adapter.
+  * 4.04 (Not Found): The request cannot be processed because the device is disabled or does not exist.
+  * 4.06 (Not Acceptable): The request body is empty without `URI-query: empty`.
+  * 4.29 (Too Many Requests): The request cannot be processed because the tenant's message limit for the current period is exceeded.
+  * 5.03 (Service Unavailable): The request cannot be processed because there is no consumer of telemetry data for the given tenant connected to Hono.
+
+This is the preferred way for devices to publish telemetry data. It is available only if the protocol adapter is configured to require devices to authenticate (which is the default).
+
+If the `hono-ttd` URI-query option is set in order to receive a command and if the authenticated device is actually a gateway, the returned command will be the first command that the northbound application has sent to either the gateway itself or to *any* device that has last sent a telemetry or event message via this gateway.
+
+**Examples**
+
+The coap-client contained in the linux package `libcoap-1-0-bin` unfortunately doesn't support DTLS 1.2. To use the coap-client with DTLS 1.2 therefore requires a own [local build](https://libcoap.net/install.html).
+
+Publish some JSON data for device `4711` (using default message type `CON` *at least once* QoS):
+
+~~~sh
+./coap-client -u sensor1@DEFAULT_TENANT -k hono-secret -m POST coaps://hono.eclipseprojects.io/telemetry -t application/json -e '{"temp": 5}'
+~~~
+
+{{% note %}}
+`./coap-client` reports only error response-codes, so the expected 2.04 will not be report!
+{{% /note %}}
+
+Publish some JSON data for device `4711` using message type `NON` *at most once* QoS:
+
+~~~sh
+./coap-client -u sensor1@DEFAULT_TENANT -k hono-secret -N -m POST coaps://hono.eclipseprojects.io/telemetry -t application/json -e '{"temp": 5}'
+~~~
+
+Publish some JSON data for device `4711`, indicating that the device will wait for 10 seconds to receive the response:
+
+~~~sh
+./coap-client -u sensor1@DEFAULT_TENANT -k hono-secret -m POST coaps://hono.eclipseprojects.io/telemetry?hono-ttd=10 -t application/json -e '{"temp": 5}'
+
+{
+  "brightness": 87
+}
+~~~
+
+## Publish Telemetry Data (unauthenticated Device)
+
+* URI: `/telemetry/${tenantId}/${deviceId}`
+* Method: `PUT`
+* Type:
+  * `CON`: *at least once* (`1`) QoS levels
+  * `NON`: *at most once* (`0`) QoS levels
+* Request Options:
+  * (optional) `content-format`: The type of payload contained in the request body. Required, if request contains payload.
+  * (optional) `URI-query: hono-ttd`: The number of seconds the device will wait for the response.
+  * (optional) `URI-query: empty`: If the request doesn't contain payload.
+* Request Body:
+  * (optional) Arbitrary payload encoded according to the given content type. Maybe empty, if `URI-query: empty` is provided.
+* Response Options:
+  * (optional) `content-format`: A media type describing the semantics and format of payload contained in the response body. This header will only be present if the response contains a command to be executed by the device which requires input data. If the command's content type do not map to one of the registered CoAP content formats, the option will not be present.
+  * (optional) `location-query: hono-command`: The name of the command to execute. This option will only be present if the response contains a command to be executed by the device.
+  * (optional) `location-path: ('command'|'command_response/'<command-request-id>)`: The location path contains `'command'` for one-way-commands, or `'command_response' / <command-request-id>`  for commands expecting  a response. In case of a expected response, the location-path contains exactly the URI-path expected for the response.
+* Response Body:
+  * (optional) Arbitrary data serving as input to a command to be executed by the device, if status code is 2.05 (Content).
+  * (optional) Error details, if status code is >= 4.00.
+* Status Codes:
+  * 2.04 (Changed): The telemetry data has been accepted for processing. Note that if the message type is `NON` (*at most once* semantics),
+    this status code does **not** mean that the message has been delivered to any potential consumer. However, if the message type is `CON` (*at least once* semantics), then the adapter waits for the message to be delivered and accepted by a downstream consumer before responding with this status code.
+  * 2.05 (Content): The telemetry data has been accepted for processing. The response contains a command for the device to execute.
+  * 4.03 (Forbidden): The request cannot be processed because the device's registration status cannot be asserted. Possible reasons for this include:
+        * The given tenant is not allowed to use this protocol adapter.
+        * The given device does not belong to the given tenant.
+  * 4.04 (Not Found): The request cannot be processed because the device is disabled or does not exist.
+  * 4.06 (Not Acceptable): The request body is empty without `URI-query: empty`.
+  * 4.29 (Too Many Requests): The request cannot be processed because the tenant's message limit for the current period is exceeded.
+  * 5.03 (Service Unavailable): The request cannot be processed because there is no consumer of telemetry data for the given tenant connected to Hono.
+
+This resource MUST be used by devices that have not authenticated to the protocol adapter. Note that this requires the `HONO_COAP_AUTHENTICATION_REQUIRED` configuration property to be explicitly set to `false`.
+
+**Examples**
+
+Publish some JSON data for device `4711` (using default message type `CON` *at least once* QoS):
+
+~~~sh
+./coap-client -m PUT coap://hono.eclipseprojects.io/telemetry/DEFAULT_TENANT/4711 -t application/json -e '{"temp": 5}'
+~~~
+
+Publish some JSON data for device `4711` using message type `NON` *at most once* QoS:
+
+~~~sh
+./coap-client -N -m PUT coap://hono.eclipseprojects.io/telemetry/DEFAULT_TENANT/4711 -t application/json -e '{"temp": 5}'
+~~~
+
+Publish some JSON data for device `4711`, indicating that the device will wait for 10 seconds to receive the response:
+
+~~~sh
+./coap-client -m PUT coap://hono.eclipseprojects.io/telemetry/DEFAULT_TENANT/4711?hono-ttd=10 -t application/json -e '{"temp": 5}'
+
+{
+  "brightness": 87
+}
+~~~
+
+## Publish Telemetry Data (authenticated Gateway)
+
+The gateway is authenticated using PSK.
+
+* URI: `/telemetry/${tenantId}/${deviceId}`
+* Method: `PUT`
+* Type:
+  * `CON`: *at least once* (`1`) QoS levels
+  * `NON`: *at most once* (`0`) QoS levels
+* Request Options:
+  * (optional) `content-format`: The type of payload contained in the request body. Required, if request contains payload.
+  * (optional) `URI-query: hono-ttd`: The number of seconds the device will wait for the response.
+  * (optional) `URI-query: empty`: If the request doesn't contain payload.
+* Request Body:
+  * (optional) Arbitrary payload encoded according to the given content type. Maybe empty, if `URI-query: empty` is provided.
+* Response Options:
+  * (optional) `content-foramt`: A media type describing the semantics and format of payload contained in the response body. This option will only be present if the response contains a command to be executed by the device which requires input data. If the command's content type do not map to one of the registered CoAP content formats, the option will not be present.
+  * (optional) `location-query: hono-command`: The name of the command to execute. This option will only be present if the response contains a command to be executed by the device.
+  * (optional) `location-path: ('command'|'command_response')/${tenantId}/${deviceId}/<command-request-id>)`: The location path contains `command/${tenantId}/${deviceId}` for one-way-commands, or `command_response/${tenantId}/${deviceId}/<command-request-id>`  for commands expecting  a response. In case of a expected response, the location-path contains exactly the URI-path expected for the response. The `${tenantId}/${deviceId}/` only be present if the response contains a command to be executed by the device.
+* Response Body:
+  * (optional) Arbitrary data serving as input to a command to be executed by the device, if status code is 2.05 (Content).
+  * (optional) Error details, if status code is >= 4.00.
+* Status Codes:
+  * 2.04 (Changed): The telemetry data has been accepted for processing.  Note that if the message type is `NON` (*at most once* semantics),
+    this status code does **not** mean that the message has been delivered to any potential consumer. However, if the message type is `CON` (*at least once* semantics), then the adapter waits for the message to be delivered and accepted by a downstream consumer before responding with this status code.
+  * 2.05 (Content): The telemetry data has been accepted for processing. The response contains a command for the device to execute.
+  * 4.03 (Forbidden): The request cannot be processed because the device's registration status cannot be asserted. Possible reasons for this include:
+        * The tenant that the gateway belongs to is not allowed to use this protocol adapter.
+        * The device belongs to another tenant than the gateway.
+        * The gateway is not authorized to act *on behalf of* the device.
+        * The gateway associated with the device is not registered or disabled.
+  * 4.04 (Not Found): The request cannot be processed because the device is disabled or does not exist.
+  * 4.06 (Not Acceptable): The request body is empty without `URI-query: empty`.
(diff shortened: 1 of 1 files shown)


boaks:
Mainly derived from http

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,231 @@
...
+
+The examples below refer to devices `4711` and `gw-1` of tenant `DEFAULT_TENANT` using *auth-ids* `sensor1` and `gw1` and corresponding passwords. The example deployment as described in the [Deployment Guides]({{< relref "deployment" >}}) comes pre-configured with the corresponding entities in its device registry component.
+Please refer to the [Credentials API]({{< relref "/api/credentials#standard-credential-types" >}}) for details regarding the different types of secrets.
+
+**NB** There is a subtle difference between the *device identifier* (*device-id*) and the *auth-id* a device uses for authentication. See [Device Identity]({{< relref "/concepts/device-identity.md" >}}) for a discussion of the concepts.
+
+## Message Limits
+
+Before accepting any telemetry or event or command messages, the CoAP adapter verifies that the configured [message limit] ({{< relref "/concepts/resource-limits.md" >}}) is not exceeded. If the limit is exceeded then the incoming message is discarded with the status code `429 Too Many Requests`.
```
sophokles73: is this actually the case?

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Device Authentication
+
+The CoAP adapter by default requires clients (devices or gateway components) to authenticate during connection establishment. The adapter supports currently only [PSK](https://tools.ietf.org/html/rfc4279) as part of a DTLS handshake for that purpose. The other variants mentioned in [Securing CoAP](https://tools.ietf.org/html/rfc7252#section-9) may follow in the future.
+
+The adapter tries to authenticate the device using this mechanisms
+
+### PSK
+
+The *identity* provided in the ClientKeyExchange must have the form *auth-id@tenant*, e.g. `sensor1@DEFAULT_TENANT`. The adapter performs the handshake using the credentials `configured Credentials service - (documentation pending)` has one record for the client. The adapter uses the Credentials API's *get* operation to retrieve the credentials on record with the *tenant* and *auth-id* provided by the device in the *identity* and `psk` as the *type* of secret as query parameters.
```
sophokles73: the second sentence seems to be garbled ...

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,231 @@
...
+  * `CON`: *at least once* (`1`) QoS levels
+  * `NON`: *at most once* (`0`) QoS levels
+* Request Options:
+  * (required) `content-type`: The type of payload contained in the request body.
+  * (optional) `URI-query: hono-ttd`: The number of seconds the device will wait for the response.
+  * (optional) `URI-query: empty`: If the request doesn't contain payload.
+* Request Body:
+  * (required) Arbitrary payload encoded according to the given content type.
+* Response Options:
+  * (optional) `content-type`: A media type describing the semantics and format of payload contained in the response body. This header will only be present if the response contains a command to be executed by the device which requires input data.
```
sophokles73: I wonder how this should work since there (currently?) is no way to map arbitrary content type strings provided by a downstream application to a CoAP content type option value, or is there?

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,231 @@
...
+* Method: `POST`
+* Type:
+  * `CON`: *at least once* (`1`) QoS levels
+  * `NON`: *at most once* (`0`) QoS levels
+* Request Options:
+  * (required) `content-type`: The type of payload contained in the request body.
+  * (optional) `URI-query: hono-ttd`: The number of seconds the device will wait for the response.
+  * (optional) `URI-query: empty`: If the request doesn't contain payload.
+* Request Body:
+  * (required) Arbitrary payload encoded according to the given content type.
```
sophokles73: seems to be optional if there is a way to indicate that the request doesn't contain payload ...

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,231 @@
...
+* Response Options:
+  * (optional) `content-type`: A media type describing the semantics and format of payload contained in the response body. This header will only be present if the response contains a command to be executed by the device which requires input data.
+  * (optional) `location-query: hono-command`: The name of the command to execute. This header will only be present if the response contains a command to be executed by the device.
+  * (optional) `location-path: ('command'|'command_response/'<command-request-id>)`: The location path contains `'command'` for one-way-commands, or `'command_response' / <command-request-id>`  for commands expecting  a response. In case of a expected response, the location-path contains exactly the URI-path expected for the response.
+* Response Body:
+  * (optional) Arbitrary data serving as input to a command to be executed by the device, if status code is 200.
```
sophokles73: IMHO this needs to be 205 in the CoAP case?

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,231 @@
...
+  * (optional) Error details, if status code is >= 400.
+* Status Codes:
+  * 205 (CONTENT): The telemetry data has been accepted for processing. The response contains a command for the device to execute.
```
sophokles73: please move this after 204 ... the notes regarding the difference between *NON* and *CON* should be included here as well, I guess.

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,231 @@
...
+  * 204 (CHANGED): The telemetry data has been accepted for processing. Note that if the message type is `NON` (*at most once* semantics),
+    this status code does **not** mean that the message has been delivered to any potential consumer. However, if the message type is `CON` (*at least once* semantics), then the adapter waits for the message to be delivered and accepted by a downstream consumer before responding with
+    this status code.
+  * 400 (Bad Request): The request cannot be processed. Possible reasons for this include:
+        * The content type header is missing.
+        * The request body is empty.
+        * The QoS header value is invalid.
```
sophokles73: is there a QoS header?

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,231 @@
...
+* Response Body:
+  * (optional) Arbitrary data serving as input to a command to be executed by the device, if status code is 200.
+  * (optional) Error details, if status code is >= 400.
+* Status Codes:
+  * 205 (CONTENT): The telemetry data has been accepted for processing. The response contains a command for the device to execute.
+  * 204 (CHANGED): The telemetry data has been accepted for processing. Note that if the message type is `NON` (*at most once* semantics),
+    this status code does **not** mean that the message has been delivered to any potential consumer. However, if the message type is `CON` (*at least once* semantics), then the adapter waits for the message to be delivered and accepted by a downstream consumer before responding with
+    this status code.
+  * 400 (Bad Request): The request cannot be processed. Possible reasons for this include:
+        * The content type header is missing.
```
sophokles73: content type option?

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,231 @@
...
+        * The request body is empty.
+        * The QoS header value is invalid.
+  * 401 (Unauthorized): The request cannot be processed because the request does not contain valid credentials.
```
sophokles73: the request itself never contains credentials, does it?

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,231 @@
...
+  * 403 (Forbidden): The request cannot be processed because the device's registration status cannot be asserted. Possible reasons for this include:
+        * The given tenant is not allowed to use this protocol adapter.
+  * 404 (Not Found): The request cannot be processed because the device is disabled or does not exist.
+  * 429 (Too Many Requests): The request cannot be processed because the tenant's message limit for the current period is exceeded.
+  * 503 (Service Unavailable): The request cannot be processed because there is no consumer of telemetry data for the given tenant connected to Hono.
+
+This is the preferred way for devices to publish telemetry data. It is available only if the protocol adapter is configured to require devices to authenticate (which is the default).
+
+If the `hono-ttd` URI-query option is set in order to receive a command and if the authenticated device is actually a gateway, the returned command will be the first command that the northbound application has sent to either the gateway itself or (if not using the legacy control endpoint) to *any* device that has last sent a telemetry or event message via this gateway.
```
sophokles73: there is no legacy control endpoint anymore ...

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,231 @@
...
+
+The coap-client contained in the linux package `libcoap-1-0-bin` unfortunately doesn't support DTLS 1.2. To use the coap-client with DTLS 1.2 therefore requires a own [local build](https://libcoap.net/install.html).
+
+Publish some JSON data for device `4711` (using default message type `CON` *at least once* QoS):
+
+~~~sh
+./coap-client -u sensor1@DEFAULT_TENANT -k hono-secret -m POST coaps://hono.eclipseprojects.io/telemetry -t application/json -e '{"temp": 5}'
+~~~
+
+(`./coap-client` reports only error response-codes, so the expected 2.04 will not be report!)
```
sophokles73: please change this into a *note*

site/documentation/content/user-guide/coap-adapter.md
boaks: `checkMessageLimit` is called, but I didn't check, if it really works.

@kaniyan 
Anything else, which must be called?

site/documentation/content/user-guide/coap-adapter.md
boaks: That simply requires a IANA registration for the content type :-).

"arbitrary content type strings" as the ""application/vnd.eclipse-hono-empty-notification" don't work. Maybe adding a supported content-type table at the bottom helps here.

site/documentation/content/user-guide/coap-adapter.md
boaks: The same is true for http, so it maybe corrected there as well.

site/documentation/content/user-guide/coap-adapter.md
boaks: OK

site/documentation/content/user-guide/coap-adapter.md
boaks: OK

site/documentation/content/user-guide/coap-adapter.md
boaks: If the DTLS handshakes fails, no response is sent at all. So the 401 will only be sent back, if some hono-internal component reports that. I can't see, how that source could be verified efficiently, if it reports that error or not and what will be that cause for 401.

site/documentation/content/user-guide/coap-adapter.md
boaks: Please propose the right section (possibly update http-guide?).

site/documentation/content/user-guide/coap-adapter.md
kaniyan: @boaks The adapters use only `checkMessageLimit` from the base class to verify message limit.

sophokles73: However, my understanding is that we (currently) do not report established (DTLS) connections so we are probably only checking for overall message data volume. If so, we should document this fact here ...

site/documentation/content/user-guide/coap-adapter.md
sophokles73: That would be a start. We will also need to document what we do with *unsupported* content-types, right?

site/documentation/content/user-guide/coap-adapter.md
sophokles73: Be my guest :-)

site/documentation/content/user-guide/coap-adapter.md
sophokles73: ???

site/documentation/content/user-guide/coap-adapter.md
boaks: You mean, add that there is currently no forced "connection limit"?

site/documentation/content/user-guide/coap-adapter.md
boaks: I have no glue, what this part was good for. And I don't know, what should be written instead. Hono has quite a lot implemented in the base-classes, and I don't know, what is changed there  or not.

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,231 @@
++++
+title = "CoAP Adapter"
+weight = 210
```
sophokles73: IMHO we should set this to 225 so that it will appear right after the MQTT adapter

site/documentation/content/user-guide/coap-adapter.md
sophokles73: yes, I would propose to do so.

site/documentation/content/user-guide/coap-adapter.md
sophokles73: just remove the
> (if not using the legacy control endpoint)

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,231 @@
...
+* Method: `POST`
+* Type:
+  * `CON`: *at least once* (`1`) QoS levels
+  * `NON`: *at most once* (`0`) QoS levels
+* Request Options:
+  * (required) `content-type`: The type of payload contained in the request body.
+  * (optional) `URI-query: hono-ttd`: The number of seconds the device will wait for the response.
+  * (optional) `URI-query: empty`: If the request doesn't contain payload.
+* Request Body:
+  * (optional) Arbitrary payload encoded according to the given content type. Maybe empty, if `URI-query: empty` is provided.
```
sophokles73: how about using 65000 as the content format option value in this case?

site/documentation/content/user-guide/coap-adapter.md
boaks: Ok

site/documentation/content/user-guide/coap-adapter.md
boaks: Please verify, that the new version is not that garbled

site/documentation/content/user-guide/coap-adapter.md
boaks: "unsupported content-types" from "north-bound"? They are ignored.
Are there other possibilities? Commands? One-Way commands?

site/documentation/content/user-guide/coap-adapter.md
boaks: Removed

site/documentation/content/user-guide/coap-adapter.md
boaks: OK

site/documentation/content/user-guide/coap-adapter.md
boaks: That would require a definition how to translate them.
FMPOV, that doesn't pay off.

boaks:
I Hope I addressed to most issues. Please "recheck", if something important is missing.

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,238 @@
...
+## Message Limits
+
+Before accepting any telemetry or event or command messages, the CoAP adapter verifies that the configured [message limit]({{< relref "/concepts/resource-limits#messages-limit" >}}) is not exceeded.
+For CoAP currently only the messages limit is verified, the connections limit is not available. If the limit is exceeded then the incoming message is discarded with the status code `429 Too Many Requests`. 
+
+## CoAP Content Formats
+
+CoAP doesn't use a textual identifier for content types. Instead numbers are used, which are maintained by the [IANA](https://www.iana.org/).
+For the currenlty assigned numbers, please check [IANA - CoAP Content Formats](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#content-formats).
+The use of unsupported content types in messages sent to the device are ignored!
```
sophokles73: What about
> Content types that do not map to one of the registered CoAP content formats are ignored.

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,238 @@
...
+  * (optional) Arbitrary data serving as input to a command to be executed by the device, if status code is 2.05 (CONTENT).
+  * (optional) Error details, if status code is >= 4.00.
+* Status Codes:
+  * 2.04 (CHANGED): The telemetry data has been accepted for processing. Note that if the message type is `NON` (*at most once* semantics),
+    this status code does **not** mean that the message has been delivered to any potential consumer. However, if the message type is `CON` (*at least once* semantics), then the adapter waits for the message to be delivered and accepted by a downstream consumer before responding with this status code.
+  * 2.05 (CONTENT): The telemetry data has been accepted for processing. The response contains a command for the device to execute.
+  * 4.00 (Bad Request): The request cannot be processed. Possible reasons for this include:
+        * The content format option is missing.
+        * The request body is empty without `URI-query: empty`.
+  * 4.01 (Unauthorized): The request cannot be processed because invalid credentials. If PSK authentication fails, no response is receieved!
```
sophokles73: FMPOV this will never be returned and we should therefore remove it

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,238 @@
...
+* Response Body:
+  * (optional) Arbitrary data serving as input to a command to be executed by the device, if status code is 2.05 (CONTENT).
+  * (optional) Error details, if status code is >= 4.00.
+* Status Codes:
+  * 2.04 (CHANGED): The telemetry data has been accepted for processing. Note that if the message type is `NON` (*at most once* semantics),
+    this status code does **not** mean that the message has been delivered to any potential consumer. However, if the message type is `CON` (*at least once* semantics), then the adapter waits for the message to be delivered and accepted by a downstream consumer before responding with this status code.
+  * 2.05 (CONTENT): The telemetry data has been accepted for processing. The response contains a command for the device to execute.
+  * 4.00 (Bad Request): The request cannot be processed. Possible reasons for this include:
+        * The content format option is missing.
```
sophokles73: ... but the request body is not empty. However, the code suggests that in this case a 4.06 (Not Acceptable) is being returned ...

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,238 @@
...
+  * (optional) `URI-query: empty`: If the request doesn't contain payload.
+* Request Body:
+  * (optional) Arbitrary payload encoded according to the given content type. Maybe empty, if `URI-query: empty` is provided.
+* Response Options:
+  * (optional) `content-format`: A media type describing the semantics and format of payload contained in the response body. This option will only be present if the response contains a command to be executed by the device which requires input data.
+  * (optional) `location-query: hono-command`: The name of the command to execute. This option will only be present if the response contains a command to be executed by the device.
+  * (optional) `location-path: ('command'|'command_response/'<command-request-id>)`: The location path contains `'command'` for one-way-commands, or `'command_response' / <command-request-id>`  for commands expecting  a response. In case of a expected response, the location-path contains exactly the URI-path expected for the response.
+* Response Body:
+  * (optional) Arbitrary data serving as input to a command to be executed by the device, if status code is 2.05 (CONTENT).
```
sophokles73: ... 2.05 (Content)

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,238 @@
...
+  * (optional) Error details, if status code is >= 4.00.
+* Status Codes:
+  * 2.04 (CHANGED): The telemetry data has been accepted for processing. Note that if the message type is `NON` (*at most once* semantics),
```
sophokles73: 2.04 (Changed):

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,238 @@
...
+    this status code does **not** mean that the message has been delivered to any potential consumer. However, if the message type is `CON` (*at least once* semantics), then the adapter waits for the message to be delivered and accepted by a downstream consumer before responding with this status code.
+  * 2.05 (CONTENT): The telemetry data has been accepted for processing. The response contains a command for the device to execute.
```
sophokles73: 2.05 (Content):

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Message Limits
+
+Before accepting any telemetry or event or command messages, the CoAP adapter verifies that the configured [message limit]({{< relref "/concepts/resource-limits#messages-limit" >}}) is not exceeded.
+For CoAP currently only the messages limit is verified, the connections limit is not available. If the limit is exceeded then the incoming message is discarded with the status code `429 Too Many Requests`. 
+
+## CoAP Content Formats
+
+CoAP doesn't use a textual identifier for content types. Instead numbers are used, which are maintained by the [IANA](https://www.iana.org/).
+For the currenlty assigned numbers, please check [IANA - CoAP Content Formats](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#content-formats).
```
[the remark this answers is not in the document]
boaks: Done

site/documentation/content/user-guide/coap-adapter.md
boaks: Done

site/documentation/content/user-guide/coap-adapter.md
boaks: Done

site/documentation/content/user-guide/coap-adapter.md
boaks: Done

site/documentation/content/user-guide/coap-adapter.md
boaks: Done

site/documentation/content/user-guide/coap-adapter.md
boaks: Done

site/documentation/content/user-guide/coap-adapter.md
boaks: Done

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,230 @@
...
+  * `CON`: *at least once* (`1`) QoS levels
+  * `NON`: *at most once* (`0`) QoS levels
+* Request Options:
+  * (optional) `content-format`: The type of payload contained in the request body. Required, if request contains payload.
+  * (optional) `URI-query: hono-ttd`: The number of seconds the device will wait for the response.
+  * (optional) `URI-query: empty`: If the request doesn't contain payload.
+* Request Body:
+  * (optional) Arbitrary payload encoded according to the given content type. Maybe empty, if `URI-query: empty` is provided.
+* Response Options:
+  * (optional) `content-format`: A media type describing the semantics and format of payload contained in the response body. This option will only be present if the response contains a command to be executed by the device which requires input data.
```
sophokles73: IMHO we should make it more explicit that even if the response contains a command, the content format might still be empty, i.e. if the command sent by the application contained an unsupported (unmappable) content type.

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,230 @@
...
+  * (optional) Arbitrary data serving as input to a command to be executed by the device, if status code is 2.05 (Content).
+  * (optional) Error details, if status code is >= 4.00.
+* Status Codes:
+  * 2.04 (Changed): The telemetry data has been accepted for processing. Note that if the message type is `NON` (*at most once* semantics),
+    this status code does **not** mean that the message has been delivered to any potential consumer. However, if the message type is `CON` (*at least once* semantics), then the adapter waits for the message to be delivered and accepted by a downstream consumer before responding with this status code.
+  * 2.05 (Content): The telemetry data has been accepted for processing. The response contains a command for the device to execute.
+  * 4.03 (Forbidden): The request cannot be processed because the device's registration status cannot be asserted. Possible reasons for this include:
+        * The given tenant is not allowed to use this protocol adapter.
+  * 4.04 (Not Found): The request cannot be processed because the device is disabled or does not exist.
+  * 4.06 (Not Acceptable): The request body is empty without `URI-query: empty`.
```
sophokles73: Based on the definition at https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html, 4.06 doesn't seem to be the right status code for this. Instead, we should use 4.00 as we do in the HTTP adapter. But this can be done in a separate PR.

site/documentation/content/user-guide/coap-adapter.md
boaks: Statement added.

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,230 @@
...
+  * (optional) `URI-query: empty`: If the request doesn't contain payload.
+* Request Body:
+  * (optional) Arbitrary payload encoded according to the given content type. Maybe empty, if `URI-query: empty` is provided.
+* Response Options:
+  * (optional) `content-format`: A media type describing the semantics and format of payload contained in the response body. This option will only be present if the response contains a command to be executed by the device which requires input data. If the command's content type do not map to one of the registered CoAP content formats, the option will not be present.
+  * (optional) `location-query: hono-command`: The name of the command to execute. This option will only be present if the response contains a command to be executed by the device.
+  * (optional) `location-path: ('command'|'command_response/'<command-request-id>)`: The location path contains `'command'` for one-way-commands, or `'command_response' / <command-request-id>`  for commands expecting  a response. In case of a expected response, the location-path contains exactly the URI-path expected for the response.
```
sophokles73: wouldn't it be sufficient to simply omit this option in case of a one-way command?

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -0,0 +1,230 @@
...
+  * `NON`: *at most once* (`0`) QoS levels
+* Request Options:
+  * (optional) `content-format`: The type of payload contained in the request body. Required, if request contains payload.
+  * (optional) `URI-query: hono-ttd`: The number of seconds the device will wait for the response.
+  * (optional) `URI-query: empty`: If the request doesn't contain payload.
+* Request Body:
+  * (optional) Arbitrary payload encoded according to the given content type. Maybe empty, if `URI-query: empty` is provided.
+* Response Options:
+  * (optional) `content-format`: A media type describing the semantics and format of payload contained in the response body. This option will only be present if the response contains a command to be executed by the device which requires input data. If the command's content type do not map to one of the registered CoAP content formats, the option will not be present.
+  * (optional) `location-query: hono-command`: The name of the command to execute. This option will only be present if the response contains a command to be executed by the device.
```
sophokles73: I have to admit that I am not really fond of abusing existing option codes for conveying information that is not at all related to the option's original semantics. This brings me back to introducing our own content format and option codes in the > 65000 range. For the options, this merely requires documenting them here. For the content-formats, we could create a wrapper around Cf's `MediaTypeRegistry` class which contains these codes as well ...
WDYT?

site/documentation/content/user-guide/coap-adapter.md
boaks: At least for the gateway-variant below, it's required.
Using it enables later "homogeneous" extensions, otherwise "none" must be a extra case, which defaults then to "command".

site/documentation/content/user-guide/coap-adapter.md
boaks: Do you mean a "custom option" instead of the location?
Or a "custom content format", which structures the response's payload in "response-URI" and "command-payload"?

sophokles73: Both, using a custom option instead of `location-query` and `location-path` as well as using custom content-format codes for e.g. `application/vnd.eclipse-hono-empty-notification`.

boaks: Both requires, that a implementation is able to process such "custom" values. I just don't know, if that is the case.

sophokles73: When you say *implementation*, what are you referring to? A device? A backend application?

boaks: A embedded device. At least, this should be checked for "libcoap" and "erbium". And for the content-format, californium must be adapted.
I would prefer to raise a question in the ietf core-mailing list about the usage of the "location". For me this case looks pretty close to the original intention of the "location".

sophokles73: > And for the content-format, californium must be adapted.And for the content-format, californium must be adapted.

IMHO we can simply wrap the `MediaTypeRegistry` and use our wrapper instead in the CoAP adapter code. Don't you think?

sophokles73: > At least, this should be checked for "libcoap" and "erbium"

Can you do that?

boaks: I have only 1 day left before my vacation and I'm sorry, but this will not be possible before that.

I want to release Californium 2.2.0-RC1 tomorrow, and so I'm currently just too busy.

sophokles73: We don't need to do that as part of this PR. This is just about documenting the current status quo, i.e. the way it is implemented at the moment. However, I wanted to start a discussion around whether it should stay that way.

boaks: OK, but as I wrote, I would prefer to ask the ietf core list about their opinion using the location option. I hope, I can write to the list tomorrow.

boaks: Reading [2.01 Created](https://tools.ietf.org/html/rfc7252#section-5.9.1.1) my impression is, that the "location" is used to control a cache. Assuming no cache, it doesn't harm.

PR #1838, opened 2020-03-16
Fix test issues with commands sent concurrently

calohmn:
Running the command & control integration tests, there is a potential issue when one command gets sent while another is still not acknowledged (no delivery update received yet).
This is caused by the same command client instance getting used via the cache. The command client instance might get closed in one `sendCommand` invocation while the corresponding sender link is still used in another concurrent `sendCommand` invocation.
This PR fixes that.


sophokles73:
@calohmn can you point out an IT where this supposedly happens?

calohmn:
@sophokles73 This has occurred in a local test environment with the (yet uncommitted) changes for  #1272. 
The relevant test method was `HttpTestBase#testUploadMessagesWithTtdThatReplyWithOneWayCommand`, while running `TelemetryHttpIT` (with `forceCommandRerouting` set).
It looks like the timing is different with the new C&C implementation (2 new links not having to be created when processing a command).

sophokles73:
@calohmn wouldn't it make more sense to simply *not* close the command sender link after each command message being sent? It looks correct to me that we use the same command sender for all commands to be sent within the same IT using the same tenant. That is what you would do in practice as well if you have a single downstream application, isn't it?

calohmn:
@sophokles73 Yes, right, I was already wondering why that was done there.
I've amended the commit accordingly.

Issue #1821, opened 2020-03-09
TLS support and basic HTTP authentication for Prometheus resource limits check


bordeuax:
By default Prometheus don't support  TLS and HTTP authentication. But, in the majority of production cases or SaaSs, Prometheus is accessible via TLS plus basic authentication.

sophokles73:
Just for my understanding, you need Hono's `PrometheusBasedResourceLimitsCheck` implementation to support using TLS for connecting to Prometheus query API and authenticate to the Prometheus server using HTTP Basic Auth, right?

bordeuax:
Hi @sophokles73 , yes your understanding is right. It's will be nice to have at configuration level `-- hono.resourceLimits.prometheusBased.*`  following options: 
* HTTP plain communication 
* HTTP basic authentication or something more generic to support also token based authentication.  This will be independently if i have HTTP or HTTPS 
* TLS configuration for HTTPS

PR #1839, opened 2020-03-16
[#1821] Add config properties for TLS and/or client auth

sophokles73:
The Prometheus based resource limits checker can now be configured to
use TLS and provide credentials when connecting to Prometheus' query
API.


sophokles73:

PR #1840, opened 2020-03-16
Add TenantInformationService functionality

dejanb:


services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/tenant/AutowiredTenantInformationService.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * A default implementation of {@link TenantInformationService} that uses embedded {@link TenantService} to verify if tenant exists.
+ */
+@Component
+public class AutowiredTenantInformationService implements TenantInformationService {
+
+    TenantService service;
+
+    @Override
+    public Future<TenantKey> tenantExists(final String tenantName, final Span span) {
```
kaniyan: For the other APIs, we use _tenantId_ as the identifier. Is it intentionally here set to _tenantName_?

dejanb: @kaniyan make sense. I'll make a change.

PR #1842, opened 2020-03-17
[#1679] Add mongodb based implementation of device registration and management services

kaniyan:
This PR is for the issue #1679, which is to provide a _mongodb_ based device registry. As a first step towards it, in this PR `DeviceManagementService `and `RegistrationService` have been implemented. In spite of some work left, I have raised this PR to collect some feedback. The tests will be added incrementally. Also some optimisations like pulling common configuration properties to the base module is in progress.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
+++ b/core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
@@ -214,10 +214,6 @@ public final class RegistryManagementConstants extends RequestResponseApiConstan
      */
     public static final String FIELD_ADAPTERS_DEVICE_AUTHENTICATION_REQUIRED = "device-authentication-required";
     /**
-     * The name of the JSON property containing the device identifier.
-     */
-    public static final String FIELD_DEVICE = "device";
-    /**
      * The name of the JSON array containing device registration information for a tenant.
      */
     public static final String FIELD_DEVICES = "devices";
@@ -274,16 +270,6 @@ public final class RegistryManagementConstants extends RequestResponseApiConstan
      * a {@link TracingSamplingMode} value.
      */
     public static final String FIELD_TRACING_SAMPLING_MODE_PER_AUTH_ID = "sampling-mode-per-auth-id";
-
-    /**
-     * The name of the JSON property containing the last modification date and time.
-     */
-    public static final String FIELD_UPDATED_ON = "updatedOn";
-
-    /**
-     * The name of the JSON property containing the version of the tenant or device or credentials information.
-     */
-    public static final String FIELD_VERSION = "version";
 
     private static final Set<String> ACTIONS = new HashSet<>(
             Arrays.asList(ACTION_CREATE, ACTION_GET, ACTION_DELETE, ACTION_UPDATE));
--- a/core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
+++ b/core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
@@ -214,6 +214,10 @@ public final class RegistryManagementConstants extends RequestResponseApiConstan
      */
     public static final String FIELD_ADAPTERS_DEVICE_AUTHENTICATION_REQUIRED = "device-authentication-required";
     /**
+     * The name of the JSON property containing the device identifier.
+     */
+    public static final String FIELD_DEVICE = "device";
+    /**
      * The name of the JSON array containing device registration information for a tenant.
      */
     public static final String FIELD_DEVICES = "devices";
@@ -270,6 +274,16 @@ public final class RegistryManagementConstants extends RequestResponseApiConstan
      * a {@link TracingSamplingMode} value.
      */
     public static final String FIELD_TRACING_SAMPLING_MODE_PER_AUTH_ID = "sampling-mode-per-auth-id";
+
+    /**
+     * The name of the JSON property containing the last modification date and time.
+     */
+    public static final String FIELD_UPDATED_ON = "updatedOn";
+
+    /**
+     * The name of the JSON property containing the version of the tenant or device or credentials information.
+     */
+    public static final String FIELD_VERSION = "version";
 
     private static final Set<String> ACTIONS = new HashSet<>(
             Arrays.asList(ACTION_CREATE, ACTION_GET, ACTION_DELETE, ACTION_UPDATE));
--- a/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
+++ b/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
@@ -15,6 +15,7 @@ package org.eclipse.hono.deviceregistry.util;
 import java.time.Instant;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.UUID;
 import java.util.stream.Collectors;
 
 import org.eclipse.hono.service.management.tenant.Tenant;
@@ -123,5 +124,14 @@ public final class DeviceRegistryUtils {
         } else {
             return CacheDirective.noCacheDirective();
         }
+    }
+
+    /**
+     * Gets a unique identifier generated using {@link UUID#randomUUID()}.
+     * 
+     * @return The generated unique identifier.
+     */
+    public static String getUniqueIdentifier() {
+        return UUID.randomUUID().toString();
     }
 }
--- a/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
+++ b/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
@@ -12,18 +12,30 @@
  *******************************************************************************/
 package org.eclipse.hono.deviceregistry.util;
 
+import java.io.ByteArrayInputStream;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
 import java.time.Instant;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
+import javax.security.auth.x500.X500Principal;
+
 import org.eclipse.hono.service.management.tenant.Tenant;
+import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.CacheDirective;
+import org.eclipse.hono.util.CredentialsConstants;
 import org.eclipse.hono.util.RegistryManagementConstants;
 import org.eclipse.hono.util.TenantConstants;
 import org.eclipse.hono.util.TenantObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import io.opentracing.Span;
+import io.vertx.core.Future;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 
@@ -32,6 +44,8 @@ import io.vertx.core.json.JsonObject;
  *
  */
 public final class DeviceRegistryUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DeviceRegistryUtils.class);
 
     private DeviceRegistryUtils() {
         // prevent instantiation
@@ -133,5 +147,53 @@ public final class DeviceRegistryUtils {
      */
     public static String getUniqueIdentifier() {
         return UUID.randomUUID().toString();
+    }
+
+    /**
+     * Gets the certificate of the device to be provisioned from the client context.
+     *
+     * @param tenantId The tenant to which the device belongs.
+     * @param authId The authentication identifier.
+     * @param clientContext The client context that can be used to get the X.509 certificate 
+     *                      of the device to be provisioned.
+     * @param span The active OpenTracing span for this operation. It is not to be closed in this method! An
+     *             implementation should log (error) events on this span and it may set tags and use this span 
+     *             as the parent for any spans created in this method.
+     * @return A future indicating the outcome of the operation. If the operation succeeds, the
+     *         retrieved certificate is returned. Else {@link Optional#empty()} is returned.
+     * @throws NullPointerException if any of the parameters except span is {@code null}.
+     */
+    public static Future<Optional<X509Certificate>> getCertificateFromClientContext(
+            final String tenantId,
+            final String authId,
+            final JsonObject clientContext,
+            final Span span) {
+
+        Objects.requireNonNull(tenantId);
+        Objects.requireNonNull(authId);
+        Objects.requireNonNull(clientContext);
+
+        try {
+            final byte[] bytes = clientContext.getBinary(CredentialsConstants.FIELD_CLIENT_CERT);
+            if (bytes == null) {
+                throw new IllegalArgumentException(
+                        String.format("The client context doesn't contain a certificate for tenant [%s]", tenantId));
+            }
+            final CertificateFactory factory = CertificateFactory.getInstance("X.509");
+            final X509Certificate cert = (X509Certificate) factory.generateCertificate(new ByteArrayInputStream(bytes));
+
+            if (!cert.getSubjectX500Principal().getName(X500Principal.RFC2253).equals(authId)) {
+                throw new IllegalArgumentException(
+                        String.format("Subject DN of the client certificate does not match authId [%s] for tenant [%s]",
+                                authId, tenantId));
+            }
+            return Future.succeededFuture(Optional.of(cert));
+        } catch (final CertificateException | ClassCastException | IllegalArgumentException error) {
+            final String errorMessage = String.format(
+                    "Error getting certificate from client context with authId [%s] for tenant [%s]", authId, tenantId);
+            LOG.error(errorMessage, error);
+            TracingHelper.logError(span, errorMessage, error);
+            return Future.succeededFuture(Optional.empty());
+        }
     }
 }
--- a/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
+++ b/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
@@ -13,6 +13,7 @@
 package org.eclipse.hono.deviceregistry.util;
 
 import java.io.ByteArrayInputStream;
+import java.net.HttpURLConnection;
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
@@ -24,6 +25,7 @@ import java.util.stream.Collectors;
 
 import javax.security.auth.x500.X500Principal;
 
+import org.eclipse.hono.client.ClientErrorException;
 import org.eclipse.hono.service.management.tenant.Tenant;
 import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.CacheDirective;
@@ -176,8 +178,7 @@ public final class DeviceRegistryUtils {
         try {
             final byte[] bytes = clientContext.getBinary(CredentialsConstants.FIELD_CLIENT_CERT);
             if (bytes == null) {
-                throw new IllegalArgumentException(
-                        String.format("The client context doesn't contain a certificate for tenant [%s]", tenantId));
+                return Future.succeededFuture(Optional.empty());
             }
             final CertificateFactory factory = CertificateFactory.getInstance("X.509");
             final X509Certificate cert = (X509Certificate) factory.generateCertificate(new ByteArrayInputStream(bytes));
@@ -193,7 +194,8 @@ public final class DeviceRegistryUtils {
                     "Error getting certificate from client context with authId [%s] for tenant [%s]", authId, tenantId);
             LOG.error(errorMessage, error);
             TracingHelper.logError(span, errorMessage, error);
-            return Future.succeededFuture(Optional.empty());
+            return Future
+                    .failedFuture(new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST, errorMessage, error));
         }
     }
 }
--- a/services/device-registry-mongodb/pom.xml
+++ b/services/device-registry-mongodb/pom.xml
@@ -1,0 +1,152 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright (c) 2020 Contributors to the Eclipse Foundation
+   
+    See the NOTICE file(s) distributed with this work for additional
+    information regarding copyright ownership.
+   
+    This program and the accompanying materials are made available under the
+    terms of the Eclipse Public License 2.0 which is available at
+    http://www.eclipse.org/legal/epl-2.0
+   
+    SPDX-License-Identifier: EPL-2.0
+ -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>org.eclipse.hono</groupId>
+        <artifactId>hono-services</artifactId>
+        <version>1.2.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>hono-service-device-registry-mongodb</artifactId>
+    <name>Hono Mongodb Based Device Registry</name>
+    <description>A Mongodb based device registry for Hono implementing the Device Registration and Credentials APIs.
+    </description>
+    <url>https://www.eclipse.org/hono</url>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.eclipse.hono</groupId>
+            <artifactId>hono-service-device-registry-base</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.vertx</groupId>
+            <artifactId>vertx-mongo-client</artifactId>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.springframework.boot</groupId>
+                <artifactId>spring-boot-maven-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+            </plugin>
+        </plugins>
+    </build>
+
+    <profiles>
+        <profile>
+            <id>build-docker-image</id>
+            <activation>
+                <property>
+                    <name>docker.host</name>
+                </property>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>io.fabric8</groupId>
+                        <artifactId>docker-maven-plugin</artifactId>
+                        <configuration>
+                            <images>
+                                <image>
+                                    <build>
+                                        <from>${java-base-image.name}</from>
+                                        <labels>
+                                            <implemented.api.1>Tenant</implemented.api.1>
+                                            <implemented.api.2>Device Registration</implemented.api.2>
+                                            <implemented.api.3>Credentials</implemented.api.3>
+                                        </labels>
+                                        <ports>
+                                            <port>5671</port>
+                                            <port>5672</port>
+                                            <port>8080</port>
+                                            <port>8443</port>
+                                            <port>${vertx.health.port}</port>
+                                        </ports>
+                                        <cmd>
+                                            <exec>
+                                                <arg>java</arg>
+                                                <arg>--illegal-access=permit</arg>
+                                                <arg>-Dvertx.cacheDirBase=/tmp</arg>
+                                                <arg>-Dloader.home=/opt/hono</arg>
+                                                <arg>-Dloader.path=extensions</arg>
+                                                <arg>-cp</arg>
+                                                <arg>/opt/hono/${project.artifactId}-${project.version}-${classifier.spring.boot.artifact}.jar</arg>
+                                                <arg>org.springframework.boot.loader.PropertiesLauncher</arg>
+                                            </exec>
+                                        </cmd>
+                                        <assembly>
+                                            <mode>dir</mode>
+                                            <basedir>/</basedir>
+                                            <inline>
+                                                <fileSets>
+                                                    <fileSet>
+                                                        <directory>${project.build.directory}</directory>
+                                                        <outputDirectory>opt/hono</outputDirectory>
+                                                        <includes>
+                                                            <include>${project.artifactId}-${project.version}-${classifier.spring.boot.artifact}.jar</include>
+                                                        </includes>
+                                                    </fileSet>
+                                                </fileSets>
+                                            </inline>
+                                        </assembly>
+                                    </build>
+                                </image>
+                            </images>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
+        <profile>
+            <id>jaeger</id>
+            <dependencies>
+                <dependency>
+                    <groupId>io.jaegertracing</groupId>
+                    <artifactId>jaeger-client</artifactId>
+                </dependency>
+            </dependencies>
+        </profile>
+        <profile>
+            <id>docker-push-image</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>io.fabric8</groupId>
+                        <artifactId>docker-maven-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>docker-push-image</id>
+                                <phase>install</phase>
+                                <goals>
+                                    <goal>push</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+
+</project>
--- a/services/device-registry-mongodb/pom.xml
+++ b/services/device-registry-mongodb/pom.xml
@@ -31,7 +31,6 @@
         <dependency>
             <groupId>org.eclipse.hono</groupId>
             <artifactId>hono-service-device-registry-base</artifactId>
-            <version>${project.version}</version>
         </dependency>
         <dependency>
             <groupId>io.vertx</groupId>
--- a/services/device-registry-mongodb/pom.xml
+++ b/services/device-registry-mongodb/pom.xml
@@ -17,7 +17,7 @@
     <parent>
         <groupId>org.eclipse.hono</groupId>
         <artifactId>hono-services</artifactId>
-        <version>1.2.0-SNAPSHOT</version>
+        <version>1.3.0-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/ApplicationConfig.java
+++ b/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/ApplicationConfig.java
@@ -1,0 +1,294 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.deviceregistry;
+
+import java.util.Optional;
+
+import org.eclipse.hono.config.ApplicationConfigProperties;
+import org.eclipse.hono.config.ServerConfig;
+import org.eclipse.hono.config.ServiceConfigProperties;
+import org.eclipse.hono.config.VertxProperties;
+import org.eclipse.hono.deviceregistry.mongodb.config.MongoDbBasedRegistrationConfigProperties;
+import org.eclipse.hono.deviceregistry.mongodb.config.MongoDbConfigProperties;
+import org.eclipse.hono.deviceregistry.mongodb.service.MongoDbBasedRegistrationService;
+import org.eclipse.hono.deviceregistry.mongodb.utils.MongoDbCallExecutor;
+import org.eclipse.hono.deviceregistry.service.credentials.AutowiredCredentialsAmqpEndpoint;
+import org.eclipse.hono.deviceregistry.service.credentials.AutowiredCredentialsManagementHttpEndpoint;
+import org.eclipse.hono.deviceregistry.service.device.AutowiredDeviceManagementHttpEndpoint;
+import org.eclipse.hono.deviceregistry.service.device.AutowiredRegistrationAmqpEndpoint;
+import org.eclipse.hono.deviceregistry.service.deviceconnection.MapBasedDeviceConnectionsConfigProperties;
+import org.eclipse.hono.deviceregistry.service.tenant.AutowiredTenantAmqpEndpoint;
+import org.eclipse.hono.deviceregistry.service.tenant.AutowiredTenantManagementHttpEndpoint;
+import org.eclipse.hono.service.HealthCheckServer;
+import org.eclipse.hono.service.VertxBasedHealthCheckServer;
+import org.eclipse.hono.service.amqp.AmqpEndpoint;
+import org.eclipse.hono.service.deviceconnection.DeviceConnectionAmqpEndpoint;
+import org.eclipse.hono.service.http.HttpEndpoint;
+import org.eclipse.hono.service.metric.MetricsTags;
+import org.eclipse.hono.util.Constants;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.boot.actuate.autoconfigure.metrics.MeterRegistryCustomizer;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Scope;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.opentracing.Tracer;
+import io.opentracing.contrib.tracerresolver.TracerResolver;
+import io.opentracing.noop.NoopTracerFactory;
+import io.vertx.core.Vertx;
+import io.vertx.core.VertxOptions;
+
+/**
+ * Spring Boot configuration for the mongodb based device registry application.
+ */
+@Configuration
+public class ApplicationConfig {
+
+    /**
+     * Exposes a Vert.x instance as a Spring bean.
+     * <p>
+     * This method creates new Vert.x default options and invokes
+     * {@link VertxProperties#configureVertx(VertxOptions)} on the object returned
+     * by {@link #vertxProperties()}.
+     *
+     * @return The Vert.x instance.
+     */
+    @Bean
+    public Vertx vertx() {
+        return Vertx.vertx(vertxProperties().configureVertx(new VertxOptions()));
+    }
+
+    /**
+     * Exposes configuration properties for Vert.x.
+     *
+     * @return The properties.
+     */
+    @ConfigurationProperties("hono.vertx")
+    @Bean
+    public VertxProperties vertxProperties() {
+        return new VertxProperties();
+    }
+
+    /**
+     * Exposes an OpenTracing {@code Tracer} as a Spring Bean.
+     * <p>
+     * The Tracer will be resolved by means of a Java service lookup.
+     * If no tracer can be resolved this way, the {@code NoopTracer} is
+     * returned.
+     *
+     * @return The tracer.
+     */
+    @Bean
+    public Tracer getTracer() {
+
+        return Optional.ofNullable(TracerResolver.resolveTracer())
+                .orElse(NoopTracerFactory.create());
+    }
+
+    /**
+     * Gets general properties for configuring the Device Registry Spring Boot application.
+     *
+     * @return The properties.
+     */
+    @Bean
+    @ConfigurationProperties(prefix = "hono.app")
+    public ApplicationConfigProperties applicationConfigProperties() {
+        return new ApplicationConfigProperties();
+    }
+
+    /**
+     * Exposes properties for configuring the health check as a Spring bean.
+     *
+     * @return The health check configuration properties.
+     */
+    @Bean
+    @ConfigurationProperties(prefix = "hono.health-check")
+    public ServerConfig healthCheckConfigProperties() {
+        return new ServerConfig();
+    }
+
+    /**
+     * Gets properties for configuring the Device Registry's AMQP 1.0 endpoint.
+     *
+     * @return The properties.
+     */
+    @Qualifier(Constants.QUALIFIER_AMQP)
+    @Bean
+    @ConfigurationProperties(prefix = "hono.registry.amqp")
+    public ServiceConfigProperties amqpServerProperties() {
+        final ServiceConfigProperties props = new ServiceConfigProperties();
+        return props;
+    }
+
+    /**
+     * Creates a new instance of an AMQP 1.0 protocol handler for Hono's <em>Device Registration</em> API.
+     *
+     * @return The handler.
+     */
+    @Bean
+    @Scope("prototype")
+    public AmqpEndpoint registrationAmqpEndpoint() {
+        return new AutowiredRegistrationAmqpEndpoint(vertx());
+    }
+
+    /**
+     * Creates a new instance of an AMQP 1.0 protocol handler for Hono's <em>Credentials</em> API.
+     * 
+     * @return The handler.
+     */
+    @Bean
+    @Scope("prototype")
+    public AmqpEndpoint credentialsAmqpEndpoint() {
+        return new AutowiredCredentialsAmqpEndpoint(vertx());
+    }
+
+    /**
+     * Creates a new instance of an AMQP 1.0 protocol handler for Hono's <em>Tenant</em> API.
+     *
+     * @return The handler.
+     */
+    @Bean
+    @Scope("prototype")
+    public AmqpEndpoint tenantAmqpEndpoint() {
+        return new AutowiredTenantAmqpEndpoint(vertx());
+    }
+
+    /**
+     * Creates a new instance of an AMQP 1.0 protocol handler for Hono's <em>Device Connection</em> API.
+     *
+     * @return The handler.
+     */
+    @Bean
+    @Scope("prototype")
+    public AmqpEndpoint deviceConnectionAmqpEndpoint() {
+        return new DeviceConnectionAmqpEndpoint(vertx());
+    }
+
+    /**
+     * Gets properties for configuring the HTTP based Device Registry Management endpoint.
+     *
+     * @return The properties.
+     */
+    @Qualifier(Constants.QUALIFIER_REST)
+    @Bean
+    @ConfigurationProperties(prefix = "hono.registry.http")
+    public ServiceConfigProperties httpServerProperties() {
+        return new ServiceConfigProperties();
+    }
+
+    /**
+     * Creates a new instance of an HTTP protocol handler for the <em>devices</em> resources
+     * of Hono's Device Registry Management API's.
+     *
+     * @return The handler.
+     */
+    @Bean
+    @Scope("prototype")
+    public HttpEndpoint deviceHttpEndpoint() {
+        return new AutowiredDeviceManagementHttpEndpoint(vertx());
+    }
+
+    /**
+     * Creates a new instance of an HTTP protocol handler for the <em>credentials</em> resources
+     * of Hono's Device Registry Management API's.
+     *
+     * @return The handler.
+     */
+    @Bean
+    @Scope("prototype")
+    public HttpEndpoint credentialsHttpEndpoint() {
+        return new AutowiredCredentialsManagementHttpEndpoint(vertx());
+    }
+
+    /**
+     * Creates a new instance of an HTTP protocol handler for the <em>tenants</em> resources
+     * of Hono's Device Registry Management API's.
+     *
+     * @return The handler.
+     */
+    @Bean
+    @Scope("prototype")
+    public HttpEndpoint tenantHttpEndpoint() {
+        return new AutowiredTenantManagementHttpEndpoint(vertx());
+    }
+
+    /**
+     * Gets properties for configuring {@code InMemoryDeviceConnectionService} which implements
+     * the <em>Device Connection</em> API.
+     *
+     * @return The properties.
+     */
+    @Bean
+    @ConfigurationProperties(prefix = "hono.device-connection.svc")
+    public MapBasedDeviceConnectionsConfigProperties deviceConnectionsProperties() {
+        return new MapBasedDeviceConnectionsConfigProperties();
+    }
+
+    /**
+     * Gets the mongodb config properties.
+     *
+     * @return The mongodb config properties.
+     */
+    @Bean
+    @ConfigurationProperties(prefix = "hono.mongodb")
+    public MongoDbConfigProperties mongoDbConfigProperties() {
+        return new MongoDbConfigProperties();
+    }
+
+    /**
+     * Gets properties for configuring {@link MongoDbBasedRegistrationService} which implements the <em>Device
+     * Registration</em> API.
+     *
+     * @return The properties.
+     */
+    @Bean
+    @ConfigurationProperties(prefix = "hono.registry.svc")
+    public MongoDbBasedRegistrationConfigProperties serviceProperties() {
+        return new MongoDbBasedRegistrationConfigProperties();
+    }
+
+    /**
+     * Customizer for meter registry.
+     *
+     * @return The new meter registry customizer.
+     */
+    @Bean
+    public MeterRegistryCustomizer<MeterRegistry> commonTags() {
+
+        return r -> r.config().commonTags(MetricsTags.forService(Constants.SERVICE_NAME_DEVICE_REGISTRY));
+
+    }
+
+    /**
+     * Exposes the health check server as a Spring bean.
+     *
+     * @return The health check server.
+     */
+    @Bean
+    public HealthCheckServer healthCheckServer() {
+        return new VertxBasedHealthCheckServer(vertx(), healthCheckConfigProperties());
+    }
+
+    /**
+     * Gets a {@link MongoDbCallExecutor} instance containing helper methods for mongodb interaction.
+     *
+     * @return An instance of the helper class {@link MongoDbCallExecutor}.
+     */
+    @Bean
+    public MongoDbCallExecutor mongoDBCallExecutor() {
+        return new MongoDbCallExecutor(vertx(), mongoDbConfigProperties());
+    }
+}
--- a/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/ApplicationConfig.java
+++ b/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/ApplicationConfig.java
@@ -33,7 +33,7 @@ import org.eclipse.hono.deviceregistry.service.tenant.AutowiredTenantManagementH
 import org.eclipse.hono.service.HealthCheckServer;
 import org.eclipse.hono.service.VertxBasedHealthCheckServer;
 import org.eclipse.hono.service.amqp.AmqpEndpoint;
-import org.eclipse.hono.service.deviceconnection.DeviceConnectionAmqpEndpoint;
+import org.eclipse.hono.service.deviceconnection.AutowiredDeviceConnectionAmqpEndpoint;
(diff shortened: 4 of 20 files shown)


services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedCredentialsService.java
```diff
@@ -0,0 +1,90 @@
...
+        //TODO
+        return null;
+    }
+
+    /**
+     * Remove all the credentials for the given device ID.
+     * @param tenantId the Id of the tenant which the device belongs to.
+     * @param deviceId the id of the device that is deleted.
+     * @param span The active OpenTracing span for this operation.
+     * @return A future indicating the outcome of the operation.
```
sophokles73: `@throws NullPointerException ...`?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedDeviceBackend.java
```diff
@@ -0,0 +1,270 @@
...
+import io.vertx.core.Future;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * A device backend that leverages and unifies {@link MongoDbBasedRegistrationService} and
+ * {@link MongoDbBasedCredentialsService}.
+ */
+@Repository
+@Qualifier("backend")
+@ConditionalOnProperty(name = "hono.app.type", havingValue = "file", matchIfMissing = true)
```
sophokles73: This annotation should be removed FMPOV

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedDeviceBackend.java
```diff
@@ -0,0 +1,270 @@
...
+            cert = (X509Certificate) factory.generateCertificate(new ByteArrayInputStream(bytes));
+
+            if (!cert.getSubjectX500Principal().getName(X500Principal.RFC2253).equals(authId)) {
+                throw new IllegalArgumentException("Subject DN of the client certificate does not match authId");
+            }
+        } catch (final CertificateException | ClassCastException | IllegalArgumentException e) {
+            TracingHelper.logError(span, e);
+            final int status = HttpURLConnection.HTTP_BAD_REQUEST;
+            return Future.succeededFuture(createErrorCredentialsResult(status, e.getMessage()));
+        }
```
sophokles73: this seems to be the same code as in the file bsaed registry, isn't it?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -0,0 +1,369 @@
...
+/**
+ * This is an implementation of the device registration service and the device management service where data 
+ * is stored in a mongodb database.
+ *
+ * @see <a href="https://www.eclipse.org/hono/docs/api/device-registration/">Device Registration API</a>
+ * @see <a href="https://www.eclipse.org/hono/docs/api/management/">Device Registry Management API</a>
+ */
+@Component
+@Qualifier("serviceImpl")
+@ConditionalOnProperty(name = "hono.app.type", havingValue = "mongodb", matchIfMissing = true)
```
sophokles73: What purpose does this serve? FMPOV this can be removed ..

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -0,0 +1,369 @@
...
+    private MongoDbCallExecutor mongoDbCallExecutor;
+
+    /**
+     * Registration service, based on {@link AbstractRegistrationService}.
+     * <p>
+     * This helps work around Java's inability to inherit from multiple base classes. We create a new Registration
+     * service, overriding the implementation of {@link AbstractRegistrationService} with the implementation of our
+     * {@link MongoDbBasedRegistrationService#getDevice(String, String, Span)}.
+     */
+    private final AbstractRegistrationService registrationService = new AbstractRegistrationService() {
```
sophokles73: this will be very hard to test. why don't you simply extend *AbstractRegistrationService* and also implement `Verticle`?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -0,0 +1,369 @@
...
+        @Override
+        public Future<JsonArray> resolveGroupMembers(final String tenantId, final JsonArray viaGroups, final Span span) {
+            return MongoDbBasedRegistrationService.this.resolveGroupMembers(tenantId, viaGroups);
+        }
+    };
+
+    /**
+     * Creates an instance of the {@link MongoDbCallExecutor}.
+     *
+     * @param mongoDbCallExecutor An instance of the mongoDbCallExecutor.
```
sophokles73: can this be `null`?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -0,0 +1,369 @@
...
+     */
+    @Autowired
+    public void setExecutor(final MongoDbCallExecutor mongoDbCallExecutor) {
+        this.mongoDbCallExecutor = mongoDbCallExecutor;
+        this.mongoClient = this.mongoDbCallExecutor.getMongoClient();
+    }
+
+    @Autowired
+    public void setConfig(final MongoDbBasedRegistrationConfigProperties config) {
```
sophokles73: javadoc?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -0,0 +1,369 @@
...
+        this.config = config;
+    }
+
+    public MongoDbBasedRegistrationConfigProperties getConfig() {
```
sophokles73: is this going to be used?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -0,0 +1,369 @@
...
+ * This is an implementation of the device registration service and the device management service where data 
+ * is stored in a mongodb database.
+ *
+ * @see <a href="https://www.eclipse.org/hono/docs/api/device-registration/">Device Registration API</a>
+ * @see <a href="https://www.eclipse.org/hono/docs/api/management/">Device Registry Management API</a>
+ */
+@Component
+@Qualifier("serviceImpl")
+@ConditionalOnProperty(name = "hono.app.type", havingValue = "mongodb", matchIfMissing = true)
+public class MongoDbBasedRegistrationService extends AbstractVerticle
```
sophokles73: final?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    public MongoDbBasedRegistrationConfigProperties getConfig() {
+        return config;
+    }
+
+    @Override
+    public void start(final Promise<Void> startPromise) {
+
+        mongoDbCallExecutor.createCollectionIndex(getConfig().getCollectionName(),
```
sophokles73: if this is what I think it is ((re-)creating indices), then I believe it should be run periodically and independently from the start-up of this verticle

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    @Override
+    public Future<OperationResult<Id>> createDevice(final String tenantId, final Optional<String> deviceId,
+            final Device device, final Span span) {
+
+        Objects.requireNonNull(tenantId);
+        Objects.requireNonNull(deviceId);
+
+        final String deviceIdValue = deviceId.orElse(UUID.randomUUID().toString());
```
sophokles73: how about encapsulating the generating of (unique) identifiers in a utility method so that the strategy can be implemented/modified in a single place?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -0,0 +1,369 @@
...
+                                        Optional.ofNullable(
+                                                DeviceRegistryUtils.getCacheDirective(getConfig().getCacheMaxAge())),
+                                        Optional.ofNullable(deviceDto.getVersion()))))
+                        .orElseGet(() -> {
+                            TracingHelper.logError(span, String.format("Device [%s] not found.", deviceId));
+                            return Future.succeededFuture(OperationResult.empty(HttpURLConnection.HTTP_NOT_FOUND));
+                        }));
+    }
+
+    private Future<OperationResult<Id>> ProcessUpdateDevice(final String tenantId, final String deviceId,
```
sophokles73: *processUpdateDevice*

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -0,0 +1,369 @@
...
+
+    private Future<RegistrationResult> getDevice(final String tenantId, final String deviceId, final Span span) {
+
+        return processReadDevice(tenantId, deviceId, span)
+                .compose(result -> Future.succeededFuture(RegistrationResult.from(result.getStatus(),
+                        convertDevice(deviceId, result.getPayload()), result.getCacheDirective().orElse(null))));
+
+    }
+
+    private boolean ifDuplicateKeyError(final Throwable throwable) {
```
sophokles73: *isDuplicateKeyError*

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/config/MongoDbConfigProperties.java
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    /**
+     * Returns the properties of this instance in a JsonObject suitable for
+     * initializing a Vertx MongoClient object. Note: if the connectionString is
+     * set, it will override all other connection settings.
+     *
+     * @return MongoDB client config object
+     */
+    public JsonObject asMongoClientConfigJson() {
```
sophokles73: the format returned by this method seems to be specific to the Mongo DB driver being actually used while the properties of this class seem to be generic. I would therefore propose to move this method to the `MongoDbExecutor` service which is already tied to the specific MongoDB driver ...

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbCallExecutor.java
```diff
@@ -0,0 +1,125 @@
...
+        return mongoClient;
+    }
+
+    /**
+     * Creates mongodb collection index. Wrapper of {@link #createIndex(String, JsonObject, IndexOptions, Handler)}
+     *
+     * @param collectionName The name of the collection of documents.
+     * @param keys           The keys to be indexed.
+     * @param options        The options used to configure index, which is optional.
+     * @return A succeeded Future if the tenant creation is successful,
```
sophokles73: I don't think that there is a tenant being created, is there?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDocumentBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+package org.eclipse.hono.deviceregistry.mongodb.utils;
+
+import org.eclipse.hono.util.RegistrationConstants;
+
+import io.vertx.core.json.JsonObject;
+
+/**
+ * Utility class for building Json documents for mongodb.
+ */
+public class MongoDbDocumentBuilder {
```
sophokles73: final?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDocumentBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+        document.put(RegistrationConstants.FIELD_PAYLOAD_DEVICE_ID, deviceId);
+        return this;
+    }
+
+    /**
+     * Returns the json object.
+     * 
+     * @return the json document.
+     */
+    public JsonObject create() {
```
sophokles73: *document()*? There is nothing being created here ..

services/device-registry-mongodb/src/main/resources/banner.txt
```diff
@@ -0,0 +1,15 @@
...
+  ______     _ _                  _    _                   
+ |  ____|   | (_)                | |  | |                  
+ | |__   ___| |_ _ __  ___  ___  | |__| | ___  _ __   ___  
+ |  __| / __| | | '_ \/ __|/ _ \ |  __  |/ _ \| '_ \ / _ \ 
+ | |___| (__| | | |_) \__ \  __/ | |  | | (_) | | | | (_) |
+ |______\___|_|_| .__/|___/\___| |_|  |_|\___/|_| |_|\___/ 
+                | |                                        
+                |_|                                        
+
+Eclipse Hono Device Registry Server ${application.formatted-version}
```
sophokles73: This should include hint that this is the Mongo DB based registry, shouldn't it?

services/device-registry-mongodb/src/main/resources/logback-spring.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright (c) 2016, 2020 Contributors to the Eclipse Foundation
```
sophokles73: just 2020

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -213,6 +213,10 @@
      * all devices to authenticate.
      */
     public static final String FIELD_ADAPTERS_DEVICE_AUTHENTICATION_REQUIRED = "device-authentication-required";
+    /**
+     * The name of the JSON property containing the device identifier.
+     */
+    public static final String FIELD_DEVICE = "device";
```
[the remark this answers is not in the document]
kaniyan: The value corresponds to FIELD_DEVICE is "device" and not "device-id" and  I don't find any such field under constants class and hence I added one.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/ApplicationConfig.java
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    /**
+     * Gets properties for configuring the HTTP based Device Registry Management endpoint.
+     *
+     * @return The properties.
+     */
+    @Qualifier(Constants.QUALIFIER_REST)
+    @Bean
+    @ConfigurationProperties(prefix = "hono.registry.rest")
```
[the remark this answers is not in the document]
kaniyan: I tired to keep it in align with the existing file based device registry that we have and there it is declared as `hono.registry.rest`. I can change it here to `hono.registry.http`.

services/device-registry-mongodb/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+    <artifactId>hono-service-device-registry-mongodb</artifactId>
+    <name>Hono Mongodb Based Device Registry</name>
+    <description>A Mongodb based device registry for Hono implementing the Device Registration and Credentials APIs.
+    </description>
+    <url>https://www.eclipse.org/hono</url>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.eclipse.hono</groupId>
+            <artifactId>hono-service-device-registry-base</artifactId>
```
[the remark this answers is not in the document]
kaniyan: Done.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/config/MongoDbBasedRegistrationConfigProperties.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+public final class MongoDbBasedRegistrationConfigProperties extends AbstractMongoDbBasedRegistryConfigProperties {
+
+    /**
+     * The default number of devices that can be registered for each tenant.
+     */
+    public static final int DEFAULT_MAX_DEVICES_PER_TENANT = 100;
+    private static final String DEFAULT_DEVICE_COLLECTION_NAME = "devices";
+
+    private int maxDevicesPerTenant = DEFAULT_MAX_DEVICES_PER_TENANT;
```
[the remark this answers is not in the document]
kaniyan: In our [documentation](https://www.eclipse.org/hono/docs/admin-guide/device-registry-config/) we have specified already that the default value is _100_ and so our file based device registry. At the first  iteration, I tried to keep it in sync with that and then add more optimised values. It makes sense to make it unlimited now for _mongodb_ based in comparison with the file based.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedDeviceBackend.java
kaniyan: Yes. My plan is to pull up commonly used properties and such reusable codes to the base module.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: This method `getConfig()`is already being used instead of directly using the instance variable `config`.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/ApplicationConfig.java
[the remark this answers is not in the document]
kaniyan: I would suggest raising a separate PR fixing this also in the _file_ and _mongodb_ based device registry, docs and also the corresponding helm charts for device registry in _IoT packages_. WDYT?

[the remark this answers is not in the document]
sophokles73: I would rather not change the prefix for the existing example registry in order to keep existing config yaml files working. However, this is a newly introduced registry so we can do what we want here...

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
sophokles73: But that is not necessary, is it?
The question is: will any other class be required to get access to this service's configuration using this method? If not (which I assume) we should remove the method, shouldn't we?

kaniyan: It makes sense and I will remove it.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/ApplicationConfig.java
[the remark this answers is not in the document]
kaniyan: :+1: Done.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedDeviceBackend.java
kaniyan: @sophokles73 I could imagine pulling up this method to the `AutoProvisioningEnabledDeviceBackend` as given below. WDYT?

```
    /**
     * Registers a device together with a set of credentials for the given client certificate.
     *
     * @param tenantId The tenant to which the device belongs.
     * @param authId The authentication identifier.
     * @param clientContext The client context that can be used to get the X.509 certificate 
     *                      of the device to be provisioned.
     * @param span The active OpenTracing span for this operation. It is not to be closed in this method! An
     *            implementation should log (error) events on this span and it may set tags and use this span as the
     *            parent for any spans created in this method.
     * @return A (succeeded) future containing the result of the operation. The <em>status</em> will be
     *         <ul>
     *         <li><em>201 CREATED</em> if the device has successfully been provisioned.</li>
     *         <li><em>4XX</em> if the provisioning failed. The payload may contain an error description.</li>
     *         </ul>
     */
    default Future<OperationResult<String>> provisionDevice(
            final String tenantId,
            final String authId,
            final JsonObject clientContext,
            final Span span) {
        // TODO.
        return null;

    }
```

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: @sophokles73 I understood from the above code, that they try creating indices during the startup of the component. If it doesn't succeed even after retries, then the component also fails to start up. 

> independently from the start-up of this verticle

Do you mean that even though the index creation fails, the component should start up?

>  then I believe it should be run periodically

Do you mean it should periodically try to create indexes until it is successfully created?

sophokles73: I am not familiar with how indices are being created in maintained in Mongo DB. However, in general with DBMS, indices need to be created as part of the logical data model, e.g. when creating a table in a relational system. I wonder if this is a one time activity only or if Mongo requires the indices to be *refreshed* periodically. If so, I think this should be done e.g. from a side-car container cron job ...

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedDeviceBackend.java
sophokles73: I was more thinking about pulling up a utility method for parsing the client context ...

kaniyan:
@sophokles73 I have incorporated the requested changes except _mongodb_ indices and pulling up a utility method for parsing the client context. These two are in progress. Once they are done I will push a separate commit for that. Would you mind taking a look again?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -114,6 +95,30 @@ public void setConfig(final MongoDbBasedRegistrationConfigProperties config) {
...
     }
 
+    /**
+     * Sets the Vert.x instance to deploy the service to.
+     *
+     * @param vertx The vertx instance.
+     * @throws NullPointerException if vertx is {@code null}.
+     */
+    @Autowired
+    public void setVertx(final Vertx vertx) {
```
sophokles73: this method is not necessary, vert.x will invoke *init* with the vert.x instance and the event-loop context during deployment of the verticle. So, instead, you should implement the *init* method accordingly ...

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -285,15 +290,14 @@ private boolean isDuplicateKeyError(final Throwable throwable) {
                         Optional.of(device.getVersion())))
                 .recover(error -> {
                     if (isDuplicateKeyError(error)) {
-                        log.debug("Device [{}] already exists for the tenant [{}]", device.getDeviceId(),
-                                device.getTenantId(), error);
+                        log.debug("Device [{}] already exists for the tenant [{}]", device.getDeviceId());
```
sophokles73: there seems to be a parameter missing

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -285,15 +290,14 @@ private boolean isDuplicateKeyError(final Throwable throwable) {
...
                         TracingHelper.logError(span, String.format("Device [%s] already exists for the tenant [%s]",
                                 device.getDeviceId(), device.getTenantId()));
                         return Future.succeededFuture(
                                 OperationResult.empty(HttpURLConnection.HTTP_CONFLICT));
                     } else {
                         log.error("Error adding device [{}] for the tenant [{}]", device.getDeviceId(),
-                                device.getTenantId(), error);
+                                device.getTenantId());
```
sophokles73: why don't you want to log the exception anymore?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: @sophokles73 As I understood from the _MongoDB_ documentation and what I have tested locally,  the indices once created remains there. Also _MongoDB_ does not recreate an index, if the index already exists for the same keys. I couldn't think of any place where we could run the index creation only once in our setup. Hence the start-up section of the component seems to be the next appropriate place for me. 

Also what I understood from the _MongoDB_ docs is that the index creation process holds an exclusive lock on the specified collection at the start and end of the index build. All subsequent operations on the collection must wait until the exclusive lock is released. But it allows interleaving read and write operations during the _majority_ of the index build. This tells me that its slightly better to run it during the start-up of the component rather independently from the start-up.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: My bad, I will update it.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: In the logs, the entire stack traces are getting logged. Since it is also being logged in span, I removed it here. If you think its better to have it there, I could add it again.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: > So, instead, you should implement the _init_ method accordingly ...

Do you mean something like below.
```
    @Override
    public void init(final Vertx vertx, final Context context) {
        this.vertx = vertx;
    }

    @Override
    public Vertx getVertx() {
        return vertx;
    }
```

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: The tracing need not be always enabled. For better debugging, it is better to have the error there. I will add it again.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedDeviceBackend.java
kaniyan: I have now pulled up a utility method for getting the certificate from the client context. Once this is merged, I will raise a separate PR for refactoring the file based device registry with this new utility method.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: @sophokles73 I have pushed the changes except the above one and regarding indices. Would you mind taking a look again?

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
```diff
@@ -134,4 +149,48 @@ public static CacheDirective getCacheDirective(final int cacheMaxAge) {
...
+            final String authId,
+            final JsonObject clientContext,
+            final Span span) throws ClientErrorException {
+
+        Objects.requireNonNull(tenantId);
+        Objects.requireNonNull(authId);
+        Objects.requireNonNull(clientContext);
+
+        try {
+            final byte[] bytes = clientContext.getBinary(CredentialsConstants.FIELD_CLIENT_CERT);
```
sophokles73: what if the client context doesn't contain a certificate?

kaniyan: @sophokles73 I just pulled up the already existing code to a utility class and didn't go through the details inside the method assuming that it has been already checked. How about adding a null check and throwing `IllegalArgumentException` saying that the certificate is not available?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: @sophokles7 Would you mind taking a look at the above comment? WDYT about it?

sophokles73: Then let's keep it like it is ...

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/config/MongoDbConfigProperties.java
```diff
@@ -0,0 +1,256 @@
...
+     * @return The connection string.
+     */
+    public String getConnectionString() {
+        return connectionString;
+    }
+
+    /**
+     * Sets the connection string for the mongodb client. If set, the connection string
+     * overrides the other connection settings. Format:
+     * mongodb://[username:password@]host1[:port1][,host2[:port2],...[,hostN[:portN]]][/[database][?options]]
```
sophokles73: IMHO a link to the resource where the connection string format is defined would be helpful ...

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/config/MongoDbConfigProperties.java
```diff
@@ -0,0 +1,256 @@
...
+     */
+    public MongoDbConfigProperties setConnectionString(final String connectionString) {
+        this.connectionString = Objects.requireNonNull(connectionString);
+        return this;
+    }
+
+    /**
+     * Gets the time in milliseconds that the mongo driver will wait to select a
+     * server for an operation before raising an error.
+     *
```
sophokles73: plaess document the default value as well

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/config/MongoDbConfigProperties.java
```diff
@@ -0,0 +1,256 @@
...
+        if (serverSelectionTimeoutInMs <= 0) {
+            throw new IllegalArgumentException("server selection timeout must be greater than zero");
+        }
+        this.serverSelectionTimeoutInMs = serverSelectionTimeoutInMs;
+        return this;
+    }
+
+    /**
+     * Gets the timeout in milliseconds to attempt a connection before timing out.
+     *
```
sophokles73: please document the default value and its semantics as well

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/BaseDto.java
```diff
@@ -0,0 +1,82 @@
...
+    public BaseDto() {
+        // Explicit default constructor.
+    }
+
+    /**
+     * Gets the version of the document.
+     * 
+     * @return The version of the document or {@code null} if not set.
+     */
+    public final String getVersion() {
```
sophokles73: do we need an explicit version if we already have the *last updated* timestamp?

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
sophokles73: that would only make sense if the client has a chance of preventing it to be thrown. For that, it would need to check for the presence of the JSON property, right? That, however, is what we are trying to encapsulate here, isn't it?

How about returning a `Future<Optional<X509Certificate>>` and let the client only process the result if it is not empty?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbCallExecutor.java
```diff
@@ -0,0 +1,152 @@
...
+                            } else {
+                                log.error("Error creating index", res2.cause());
+                                indexCreationPromise.fail(res2.cause());
+                            }
+                        }));
+            } else {
+                log.error("Error creating index", res.cause());
+                indexCreationPromise.fail(res.cause());
+            }
+        });
```
sophokles73: I believe this can be done more elegantly using recursion, can't it?

kaniyan: This is the contribution code from the hub and the next step is making optimisations as you suggested.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/BaseDto.java
kaniyan: The version is mapped to _Etags_ and also used to verify them.

sophokles73: I know, that's why I am asking. We can simply use the *last updated* timestamp as the version ...

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbCallExecutor.java
sophokles73: I assume you have made a note to your journal so that you don't forget it ;-)

kaniyan: sure :-)

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: @sophokles When you said implement the _init_ method accordingly, is it as below?
 ```
@Override
    public void init(final Vertx vertx, final Context context) {
    }
```
Or should I create a vertx instance with the given vertx properties inside the `init` method? 
 ```
@Override
    public void init(final Vertx vertx, final Context context) {
      this.vertx = Vertx.vertx(vertxProperties().configureVertx(new VertxOptions()));
    }
```
Would you mind giving some more insights for me?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/BaseDto.java
kaniyan: I guess that is as unique as the UUID. In that case, we can use it.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
sophokles73: The former

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/config/MongoDbConfigProperties.java
kaniyan: I have mentioned that in the _setter_ method and now I will add that to the _getter_ method too.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/config/MongoDbConfigProperties.java
kaniyan: Done.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/BaseDto.java
sophokles73: Neither the version nor the timestamp are nor need to be unique. Their only purpose is to serve as a reference for implementing optimistic locking for specific DTO instances.

kaniyan: May be I should have added more info instead of just mentioning it to be unique. It need not be unique always and it could be reused. My point was in order to have a strong validator, the RFC spec suggests that it should be _unique across all versions of all representations associated with a particular resource over time_. My concern was, what if a resource has been updated more than one time within a very short span of time lets say a millisecond (eg. Automated scripts to update resources). I think this could be effectively handled by also making use of the nanoseconds as below.
`final String version = String.format("%d-%d", lastUpdatedOn.toEpochMilli(), lastUpdatedOn.getNano())`

sophokles73: good point :-) no matter how fine the resolution of the timestamp used will be, we might still run into the problem of having two updates occurring at (almost) the same time ... based on that it is probably better to keep the strictly increasing sequence number as the version. Thanks for pointing that out :+1:

kaniyan: @sophokles73 I have used now an increasing sequence number as the version and also implemented other comments. I have rebased with the master and pushed the changes now. Last 3 commits are the new ones for review. Would you mind taking a look again?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedDeviceBackend.java
```diff
@@ -161,23 +159,23 @@ public MongoDbBasedDeviceBackend(
...
+                                                        return Future.succeededFuture(
+                                                                createErrorCredentialsResult(r.getStatus(),
+                                                                        r.getPayload()));
+                                                    } else {
+                                                        return getNewCredentials(tenantId, authId, span);
+                                                    }
+                                                }))
+                                        .orElse(Future.succeededFuture(
+                                                createErrorCredentialsResult(HttpURLConnection.HTTP_BAD_REQUEST,
+                                                        "Not able to get the certificate from the client context"))));
```
sophokles73: why is it a bad request if the client context does not contain a certificate? Maybe the adapter doesn't want to support auto provisioning?

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
```diff
@@ -160,22 +159,26 @@ public static String getUniqueIdentifier() {
...
 
         Objects.requireNonNull(tenantId);
         Objects.requireNonNull(authId);
         Objects.requireNonNull(clientContext);
 
         try {
             final byte[] bytes = clientContext.getBinary(CredentialsConstants.FIELD_CLIENT_CERT);
+            if (bytes == null) {
+                throw new IllegalArgumentException(
+                        String.format("The client context doesn't contain a certificate for tenant [%s]", tenantId));
```
sophokles73: this just means that the adapter doesn't support auto provisioning, right? So, FMPOV this should result in
```
return Optional.empty();
```

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
```diff
@@ -184,13 +187,13 @@ public static X509Certificate getCertificateFromClientContext(
...
             }
-            return cert;
+            return Future.succeededFuture(Optional.of(cert));
         } catch (final CertificateException | ClassCastException | IllegalArgumentException error) {
             final String errorMessage = String.format(
                     "Error getting certificate from client context with authId [%s] for tenant [%s]", authId, tenantId);
             LOG.error(errorMessage, error);
             TracingHelper.logError(span, errorMessage, error);
-            throw new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST, error.getMessage());
+            return Future.succeededFuture(Optional.empty());
```
sophokles73: this means that the adapter supports auto provisioning but the data is malformed. Thus, this should result in a failed future with a `ClientErrorException(400)`, right?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/BaseDto.java
```diff
@@ -43,22 +45,56 @@ public BaseDto() {
...
+
+    /**
+     * Gets the version of the document.
+     *
      * @return The version of the document or {@code null} if not set.
      */
+    @JsonIgnore
     public final String getVersion() {
-        return version;
+        return String.format("%d", versionNo);
```
sophokles73: you really want to compute this every time this method is invoked?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/BaseDto.java
```diff
@@ -14,23 +14,25 @@
...
- * TODO.
+ * The base class for implementing a DTO (Data Transfer Object) to store data in mongodb.
  */
 @JsonIgnoreProperties(ignoreUnknown = true)
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public abstract class BaseDto {
 
     @JsonProperty(value = RegistryManagementConstants.FIELD_VERSION, required = true)
-    protected String version;
+    protected Long versionNo;
```
sophokles73: is it helpful to use a Long here instead of a String, i.e. does the Mongo DB driver update this field automatically or do you need to set this value manually whenever you save a modified object instance?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -364,15 +317,32 @@ private boolean isDuplicateKeyError(final Throwable throwable) {
...
+                                Id.of(deviceDto.getDeviceId()),
                                 Optional.empty(),
                                 Optional.of(deviceDto.getVersion()));
                     }
                 });
     }
+
+    private <T> Future<T> isMaxDevicesLimitReached(final String tenantId) {
+        final Promise<Long> findExistingNoOfDevicesPromise = Promise.promise();
+        mongoClient.count(config.getCollectionName(), new JsonObject(), findExistingNoOfDevicesPromise);
```
sophokles73: this doesn't seem to account for the tenant given but seems to count the number of elements in the collection for all tenants, doesn't it?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -179,28 +171,11 @@ public void stop(final Promise<Void> stopPromise) {
...
-                        TracingHelper.logError(span, "Resource Version mismatch");
-                        return Future.succeededFuture(OperationResult.empty(HttpURLConnection.HTTP_PRECON_FAILED));
-                    }
-                    return processUpdateDevice(tenantId, deviceId, deviceDto, span);
-                });
+        return MongoDbServiceUtils.isAllowedToModify(config, tenantId)
+                .compose(ok -> findDevice(tenantId, deviceId))
+                .compose(deviceDto -> MongoDbServiceUtils.checkResourceVersion(deviceDto, resourceVersion))
+                .compose(deviceDto -> processUpdateDevice(tenantId, deviceDto, span))
+                .recover(error -> Future.succeededFuture(MongoDbServiceUtils.mapErrorToResult(error, span)));
```
sophokles73: my understanding of this code is that we first load the entity, then check if the resource version is correct, then modify the entity and then write it back to the database. I am no expert in Mongo but I would assume that you can simply invoke an update operation providing filter criteria in order to do that in one go, can't you?
This approach here also seems to have the problem that after you retrieved the device and verified that the version matches, another update could take place which changes the version *before* you are writing back the object, or does the Mongo DB driver prevent this somehow?

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
kaniyan: I am not deep into this auto-provisioning topic. As per my understanding, we are checking first if  auto-provisioning is enabled using `isAutoProvisioningEnabled(...)`. If it is enabled, then  `getCertificateFromClientContext(...)` is being invoked. Since it is already enabled, I thought  that the client context should return a certificate and hence I made it so.
I will replace it with ` return Optional.empty();`

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedDeviceBackend.java
kaniyan: I see the point. If the client context does not contain a certificate, instead of `BAD_REQUEST`, I will pass the original result which is `HTTP_NOT_FOUND` as below.
`orElse(Future.succeededFuture(result)));`

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
kaniyan: I have fixed now this utility method for getting a certificate from the client context according to your comments and pushed the commit.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/BaseDto.java
kaniyan: Since it is a sequence number, I thought its safe and better to use the `Long` instead of storing as a `String`. Mongo DB driver update this field automatically based on the value store in the database. But the version number is being incremented manually before the database update.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: My bad. I fixed it now.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/BaseDto.java
kaniyan: As I thought through it, FMPOV normally the read operations in device registry are performed more frequently than the write operations. Every time type casting or formatting during read operation is not so efficient. So, I would replace the data type of the `version` to `String`. The typecasting could be performed during add/update operations. WDYT?

sophokles73: That is probably true.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/config/MongoDbConfigProperties.java
sophokles73: and what is the default value of the mongodb client?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/BaseDto.java
sophokles73: I agree and I wonder why we have a *getVersion* method that returns a String. Do we need such a method?
Does the `null` value have any special semantics for this property?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
sophokles73: did you?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/BaseDto.java
kaniyan: In many device registration operations, `Future<OperationResult>` is being returned. When an operation is successful `OperationResult.ok(...)` is being used for that and it expects the version to be String. 

```
OperationResult<T> ok(final int status, final T payload,
            final Optional<CacheDirective> directive, final Optional<String> resourceVersion)
```

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: I fixed it locally. I didn't push it yet. I will include in the next commit and push it.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/BaseDto.java
kaniyan: In many device registration operations, `Future<OperationResult>` is being returned. When an operation is successful `OperationResult.ok(...)` is being used for that and it expects the version to be String. 

```
OperationResult<T> ok(final int status, final T payload,
            final Optional<CacheDirective> directive, final Optional<String> resourceVersion)
```
For that purpose, `getVersion()` is needed and it returns a `String`.

kaniyan: > Does the `null` value have any special semantics for this property?

Actually not, since we always store the version eventhough the incoming requests do not contain any version.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/BaseDto.java
sophokles73: I see. So let's go with a String valued version field then and do the serialization as part of the setter, ok?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: I see the point regarding another update between the retrieval of the device and updating the retrieved device. I don't see any explicit locking mechanisms with the library that we use to lock more than one transactions. In that case, we can also include  _version_ in the filter criteria for the update. The update will be performed only when the _version_ also matches or no _version_ is available in the request. The current code for retrieval of the device and the version check before performing update operation could be removed.

In this approach, when no resource(_device_) is updated, there are atleast two possible reasons. Once is "device not found" and the other is "resource version mismatch". In order to report the response code accordingly, a find query need to be executed to retrieve the device followed by version check.


> I see. So let's go with a String valued version field then and do the serialization as part of the setter, ok?

Previously we discussed to make the _version_ in the `BaseDto` class as `String`. This approach somehow forces us to declare the _version_ as `long`. When a request to update a device arrives, it may not contain a version and in that case, we simply update with the given device information and increment the version value by one. In this approach, during the update operation, we don't know what is the version of the resource and hence we cannot manually increment to one unless queried from the database. For that purpose, we could use "$inc" operator from MongoDB. Using this, we could instruct the MongoDB to increment the _version_ field by one during update operation. One prerequisite is, that this field should be a number.

sophokles73: I guess that is the reason why the example registry uses UUID for the version in the first place :-) A new UUID can be computed independently from the existing version ...
So where does this leave us? We stick with the UUID or do you want to use the *auto increment* feature of MongoDB?

kaniyan: As  I understood from the gitter chat, we would like to stick to passing the `version` using `String` data type. IMHO allowing the _version_ as _string_ and then enforcing that the string should be a valid number doesn't look so nice and also we could avoid the typecasting concerns. Hence I would prefer using UUID.

kaniyan: Now I use UUID and pushed the corresponding changes. Pls take a look :-)

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/DeviceDto.java
```diff
@@ -44,18 +44,16 @@ public DeviceDto() {
...
-        this.deviceId = Objects.requireNonNull(deviceId);
-        this.device = device;
-        this.versionNo = version;
-        this.updatedOn = Objects.requireNonNull(updatedOn);
+    public DeviceDto(final String tenantId, final String deviceId, final Device device, final String version) {
+        setTenantId(tenantId);
+        setDeviceId(deviceId);
+        setDevice(device);
+        setVersion(version);
+        setUpdatedOn(Instant.now());
```
sophokles73, line 56: are the params allowed to be `null` or not (as indicated by the JavaDoc)?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -304,39 +323,67 @@ private boolean isDuplicateKeyError(final Throwable throwable) {
...
+                            // that the device is not found or version mismatch.
+                            if (resourceVersion.isPresent()) {
+                                return findDevice(tenantId, deviceId)
+                                        .compose(foundDevice -> checkForVersionMismatchAndFail(deviceId,
+                                                foundDevice.getVersion(), resourceVersion.get()));
+                            } else {
+                                return Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_NOT_FOUND,
+                                        String.format("Device [%s] not found.", deviceId)));
+                            }
+                        }));
```
sophokles73, line 361: looks like this is the same code as for the *delete* case, isn't it?

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
sophokles73, line 219: is this field used by the Device Registry Management API types? If not, then I believe it belongs into the MongoDB module as it is implementation specific, isn't it?

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -271,6 +275,16 @@
...
 
+    /**
+     * The name of the JSON property containing the last modification date and time.
+     */
+    public static final String FIELD_UPDATED_ON = "updatedOn";
+
+    /**
+     * The name of the JSON property containing the version of the tenant or device or credentials information.
+     */
+    public static final String FIELD_VERSION = "version";
```
sophokles73, line 286: are these fields used by the Device Registry Management API types? If not, then I believe they belong into the MongoDB module as they are implementation specific, aren't they?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/DeviceDto.java
kaniyan: As indicated by the _JavaDoc_, the _params_ are not allowed to be `null` except the `device`. The checks are being done in the respective _setters_. Checking for `null` again in this constructor seemed to me repetitive and hence I have removed it and then retained the `JavaDoc`.

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
kaniyan: I have moved this to the utility class inside the MongoDB module.

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
kaniyan: I have moved this to the utility class inside the MongoDB module.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: I have extracted this code to a method and use it.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbCallExecutor.java
kaniyan: @sophokles73 As we discussed above, I have now refactored the index creation code using recursion and pushed the changes.

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
sophokles73: have you?

kaniyan: I am sorry. I have moved the fields `updatedOn` and the `version` and somehow missed this. I have changed now and pushed a commit for it.

kaniyan:
@sophokles73 I have prepared it for merging by squashing the commits and rebasing with master.

PR #1843, opened 2020-03-18
Switch from using sync to async Infinispan methods

ctron:


client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCache.java
```diff
@@ -162,44 +162,45 @@ public void addReconnectListener(final ReconnectListener<HotrodCache<K, V>> list
...
-            vertx.executeBlocking(outcome -> {
-                try {
-                    final V replacedValue = cache.put(key, value);
-                    outcome.complete(replacedValue);
-                } catch (final Throwable e) {
-                    outcome.fail(e);
-                }
-            }, result);
+
+            return MoreFutures.create(() -> {
```
calohmn: I find the naming of `MoreFutures.create()` somewhat unintuitive (are more/multiple Futures created here?).
Looking at the Javadoc of the class and the method, I think something like `FuturesHelper.mapCompletionStageResult()` would make it more obvious what is being done here. WDYT?

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCache.java
```diff
@@ -162,44 +162,45 @@ public void addReconnectListener(final ReconnectListener<HotrodCache<K, V>> list
...
 
     @Override
     public Future<Boolean> removeWithVersion(final K key, final long version) {
 
-        final Promise<Boolean> result = Promise.promise();
         if (cache == null) {
-            result.fail(new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "no connection to data grid"));
+
+            return Future.failedFuture(new ServerErrorException(
+                    HttpURLConnection.HTTP_UNAVAILABLE, "no connection to data grid"));
```
calohmn: Replace with `return noConnectionFailure();`?

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCache.java
```diff
@@ -235,38 +235,46 @@ public void addReconnectListener(final ReconnectListener<HotrodCache<K, V>> list
     @Override
     public Future<Versioned<V>> getWithVersion(final K key) {
 
-        final Promise<Versioned<V>> result = Promise.promise();
         if (cache == null) {
-            result.fail(new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "no connection to data grid"));
+
+            return Future.failedFuture(new ServerErrorException(
+                    HttpURLConnection.HTTP_UNAVAILABLE, "no connection to data grid"));
```
calohmn: Replace with `return noConnectionFailure();`?

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCache.java
```diff
@@ -235,38 +235,46 @@ public void addReconnectListener(final ReconnectListener<HotrodCache<K, V>> list
...
 
     @Override
     public Future<Map<K, V>> getAll(final Set<? extends K> keys) {
-        final Promise<Map<K, V>> result = Promise.promise();
+
         if (cache == null) {
-            result.fail(new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "no connection to data grid"));
+
+            return Future.failedFuture(new ServerErrorException(
+                    HttpURLConnection.HTTP_UNAVAILABLE, "no connection to data grid"));
```
calohmn: Replace with `return noConnectionFailure();`?

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCache.java
ctron: I followed this pattern: https://guava.dev/releases/23.0/api/docs/com/google/common/base/MoreObjects.html

But I am open to making changes to the name.

calohmn: I think with `MoreObjects` that pattern works, because it complements the `Objects` helper class. Since there is no `Futures` helper class for Vert.x or Java futures, this pattern doesn't work so well here, IMHO.

sophokles73: Then why not simply call the class `Futures`?

ctron: Because we already have a class called `Futures` in our classpath: `com.google.common.util.concurrent.Futures`.

ctron: Assuming you still prefer `Futures`, I updated the pull request to that name. If you re-consider, I can still revert back to `MoreFutures`.

PR #1844, opened 2020-03-19
Provide abstract base classes for implementing device registries

dejanb:


services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsManagementService.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    protected abstract Future<OperationResult<List<CommonCredential>>> processReadCredentials(DeviceKey key, Span span);
+
+    @Override
+    public Future<OperationResult<Void>> updateCredentials(final String tenantId, final String deviceId, final List<CommonCredential> credentials, final Optional<String> resourceVersion, final Span span) {
+        Objects.requireNonNull(tenantId);
+        Objects.requireNonNull(deviceId);
+        Objects.requireNonNull(resourceVersion);
+
+        //TODO add credentials verification and encoding logic here to be shared between implementations
```
ctron: Is that intentional?

dejanb: Just to scope the work. I started looking into it and to do it properly it needed some more effort.

PR #1849, opened 2020-03-20
Move Futures class to the core module

dejanb:


dejanb:
This class is useful beyond the device connection implementation (we have more stuff in enmasse that I intend to transfer to Hono), so let's move it to the core.

PR #1851, opened 2020-03-23
[#1272] Add SpanContext parameter to createCommandConsumer()

calohmn:
This is for #1272.

### DIFF
--- a/adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/impl/LoraProtocolAdapter.java
+++ b/adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/impl/LoraProtocolAdapter.java
@@ -334,7 +334,8 @@ public final class LoraProtocolAdapter extends AbstractVertxBasedHttpProtocolAda
         return getCommandConsumerFactory().createCommandConsumer(
                 tenantId,
                 LORA_COMMAND_CONSUMER_DEVICE_ID,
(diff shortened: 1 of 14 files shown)


adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/impl/LoraProtocolAdapter.java
```diff
@@ -334,7 +334,8 @@ private void scheduleStartLoraCommandConsumer(final String tenantId) {
         return getCommandConsumerFactory().createCommandConsumer(
                 tenantId,
                 LORA_COMMAND_CONSUMER_DEVICE_ID,
-                receivedCommandContext -> commandConsumer(tenantId, receivedCommandContext));
+                receivedCommandContext -> commandConsumer(tenantId, receivedCommandContext),
+                null);
```
ctron: As the method later on calls `span.context()` I would expect the method to fail. Why not pass `NoopSpan.INSTANCE`?

calohmn: The `null` stands for the `SpanContext` parameter here, and that may be `null`. So I don't think it is a problem.

ctron: Indeed, it looked to me like this was the `Span` parameter.

PR #1853, opened 2020-03-23
TenantInformation service should return result instead of the direct …

dejanb:
…value

### DIFF
--- a/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/device/AbstractRegistrationService.java
+++ b/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/device/AbstractRegistrationService.java
@@ -18,6 +18,7 @@ import java.util.Objects;
 
 import org.eclipse.hono.client.ServiceInvocationException;
 import org.eclipse.hono.deviceregistry.service.tenant.TenantInformationService;
+import org.eclipse.hono.deviceregistry.service.tenant.TenantKey;
 import org.eclipse.hono.service.registration.RegistrationService;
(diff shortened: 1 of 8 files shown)


dejanb:
As this internally, this is to align `TenantInformation` service with other services. For the future versions, I would suggest we stop using `Result` objects. I think as we don't have internal message passing using event bus, they are not needed and we can use standard failed future mechanism. WDYT?

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/device/AbstractRegistrationService.java
```diff
@@ -110,15 +111,17 @@ public void setTenantInformationService(final TenantInformationService tenantInf
...
+    public Future<RegistrationResult> assertRegistration(final String tenantId, final String deviceId, final Span span) {
 
         Objects.requireNonNull(tenantId);
         Objects.requireNonNull(deviceId);
         Objects.requireNonNull(span);
 
         return this.tenantInformationService
                 .tenantExists(tenantId, span)
-                .flatMap(tenantKey -> processAssertRegistration(DeviceKey.from(tenantKey, deviceId), span)
+                .flatMap(tenantKeyResult -> tenantKeyResult.isError()
```
sophokles73: is there a particular reason why you are using *flatMap* in some places and *compose* in other places?

dejanb: Changed everything to `compose` to make it uniform across the codebase.

PR #1854, opened 2020-03-24
Add credential checking and encoding to the abstract service

dejanb:


services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsManagementService.java
```diff
@@ -44,6 +65,11 @@ public void setTenantInformationService(final TenantInformationService tenantInf
         this.tenantInformationService = tenantInformationService;
     }
 
+    @Autowired
+    public void setPasswordEncoder(final HonoPasswordEncoder passwordEncoder) {
+        this.passwordEncoder = passwordEncoder;
```
sophokles73: does it make sense to allow setting this to `null`?

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsManagementService.java
```diff
@@ -91,4 +122,51 @@ public void setTenantInformationService(final TenantInformationService tenantInf
                         : processReadCredentials(DeviceKey.from(result.getPayload(), deviceId), span));
     }
 
+    private Future<List<CommonCredential>> verifyAndEncodePasswords(final List<CommonCredential> credentials) {
+        // Check if we need to encode passwords
+        if (!needToEncode(credentials)) {
```
sophokles73: personally, I find it easier to read (and understand) the *positive* condition, i.e.
```
if (needToEncode(credentials)) {
} else {
}
```

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsManagementService.java
```diff
@@ -91,4 +122,51 @@ public void setTenantInformationService(final TenantInformationService tenantInf
...
+     * @return {@code true} is the list contains at least one password which needs to be encoded on the
+     *         server side.
+     */
+    private static boolean needToEncode(final List<CommonCredential> credentials) {
+        return credentials
+                .stream()
+                .filter(PasswordCredential.class::isInstance)
+                .map(PasswordCredential.class::cast)
+                .flatMap(c -> c.getSecrets().stream())
+                .anyMatch(s -> s.getPasswordPlain() != null && !s.getPasswordPlain().isEmpty());
```
sophokles73: `!Strings.isNullOrEmpty(s.getPasswordPlain())`?

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsManagementService.java
dejanb: I added `requireNonNull` check for tenant information service as well. We can introduce Noop version for folks that don't want to use autowired one.

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsManagementService.java
```diff
@@ -91,4 +131,51 @@ public void setTenantInformationService(final TenantInformationService tenantInf
                         : processReadCredentials(DeviceKey.from(result.getPayload(), deviceId), span));
     }
 
+    private Future<List<CommonCredential>> verifyAndEncodePasswords(final List<CommonCredential> credentials) {
+        // Check if we need to encode passwords
+        if (!needToEncode(credentials)) {
```
sophokles73: now you also need to remove the `!` ;-)

dejanb: Gah! thanks :)

Issue #1860, opened 2020-03-27
Command & Control errors if adapter configured with maxInstances > 1


calohmn:
If a protocol adapter is configured with a `hono.app.maxInstances` value greater 1, there are errors receiving Command & Control messages.

Can be reproduced with the integration tests, adapting the configuration accordingly.

calohmn:
With `hono.app.maxInstances` > 1 there are multiple `command_internal/<adapterInstanceId>` consumer links being created, currently all with the same id. 

But each consumer needs to get its own id so that commands get received on the consumer instance where the corresponding handler got registered.

That means that for `hono.app.maxInstances` > 1, the notion of `adapterInstanceId` doesn't fit very well since it is not supposed to be *one* id representing the whole protocol adapter instance jvm process. 
While the term `adapterInstanceId` gives a good enough idea of the principle that commands may need to get forwarded to another adapter instance (where the matching consumer was registered), I think we should nonetheless consider renaming it to e.g. `commandConsumerId` (or `targetCommandConsumerId`) in the next Hono release.

calohmn:
The fix should also get adopted for the next 1.2 patch version.

PR #1861, opened 2020-03-29
[#1860] Use unique adapter instance id per consumer factory

calohmn:
Fix for #1860.

Makes sure that each `ProtocolAdapterCommandConsumerFactory` instance gets assigned its own unique `adapterInstanceId` (which then kind of represents the `AbstractProtocolAdapterBase` instance that the command consumer factory instance is used in).

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/HonoConnection.java
+++ b/client/src/main/java/org/eclipse/hono/client/HonoConnection.java
@@ -225,14 +225,19 @@ public interface HonoConnection extends ConnectionLifecycle<HonoConnection> {
 
     /**
      * Gets the remote container id as advertised by the peer.
-     * <p>
-     * This default implementation simply returns {@code N/A}.
      *
      * @return The remote container id or {@code null}.
      */
-    default String getRemoteContainer() {
-        return "N/A";
-    }
+    String getRemoteContainerId();
+
+    /**
+     * Gets the container id that is advertised to the peer.
+     * <p>
+     * The identifier is supposed to be unique for this HonoConnection.
+     *
+     * @return The container id.
+     */
+    String getContainerId();
 
     /**
      * Checks if this client supports a certain capability.
--- a/client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
@@ -19,6 +19,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.UUID;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
@@ -111,6 +112,7 @@ public class HonoConnectionImpl implements HonoConnection {
     private final ConnectionFactory connectionFactory;
     private final Object connectionLock = new Object();
 
+    private final String containerId;
     private final DeferredConnectionCheckHandler deferredConnectionCheckHandler;
 
     private ProtonClientOptions clientOptions;
@@ -156,6 +158,8 @@ public class HonoConnectionImpl implements HonoConnection {
         } else {
             this.connectionFactory = ConnectionFactory.newConnectionFactory(this.vertx, clientConfigProperties);
         }
+        this.containerId = ConnectionFactory.createContainerId(clientConfigProperties.getName(),
+                clientConfigProperties.getServerRole(), UUID.randomUUID());
         this.clientConfigProperties = clientConfigProperties;
         this.connectAttempts = new AtomicInteger(0);
     }
@@ -403,6 +407,9 @@ public class HonoConnectionImpl implements HonoConnection {
                 clientOptions = options;
                 connectionFactory.connect(
                         clientOptions,
+                        null,
+                        null,
+                        containerId,
                         remoteClose -> onRemoteClose(remoteClose, disconnectHandler),
                         failedConnection -> onRemoteDisconnect(failedConnection, disconnectHandler),
                         conAttempt -> {
@@ -1043,11 +1050,16 @@ public class HonoConnectionImpl implements HonoConnection {
      * @return The remote container id or {@code null}.
      */
     @Override
-    public String getRemoteContainer() {
+    public String getRemoteContainerId() {
         if (!isConnectedInternal()) {
             return null;
         }
         return connection.getRemoteContainer();
+    }
+
+    @Override
+    public String getContainerId() {
+        return containerId;
     }
 
     //-----------------------------------< private methods >---
--- a/core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
+++ b/core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
@@ -13,8 +13,6 @@
 
 package org.eclipse.hono.config;
 
-import java.util.UUID;
-
 import org.eclipse.hono.util.Constants;
 
 /**
@@ -70,7 +68,6 @@ public class ClientConfigProperties extends AuthenticatingClientConfigProperties
      */
     public static final long DEFAULT_SEND_MESSAGE_TIMEOUT = 1000L; // ms
 
-    private final String containerIdUuidPart = UUID.randomUUID().toString();
     private String amqpHostname;
     private int connectTimeoutMillis = DEFAULT_CONNECT_TIMEOUT;
     private long flowLatency = DEFAULT_FLOW_LATENCY;
@@ -119,7 +116,6 @@ public class ClientConfigProperties extends AuthenticatingClientConfigProperties
      * Gets the name being indicated as part of the <em>container-id</em> in the client's AMQP <em>Open</em> frame.
      * 
      * @return The name or {@code null} if no name has been set.
-     * @see #getContainerId()
      */
     public final String getName() {
         return name;
@@ -129,7 +125,6 @@ public class ClientConfigProperties extends AuthenticatingClientConfigProperties
      * Sets the name to indicate as part of the <em>container-id</em> in the client's AMQP <em>Open</em> frame.
      * 
      * @param name The name to set.
-     * @see #getContainerId()
      */
     public final void setName(final String name) {
         this.name = name;
@@ -151,17 +146,6 @@ public class ClientConfigProperties extends AuthenticatingClientConfigProperties
      */
     public final void setAmqpHostname(final String amqpHostname) {
         this.amqpHostname = amqpHostname;
-    }
-
-    /**
-     * Gets the identifier being indicated as the <em>container-id</em> in the client's AMQP <em>Open</em> frame.
-     * <p>
-     * The identifier consists of the name returned by {@link #getName()} plus a UUID.
-     *
-     * @return The identifier.
-     */
-    public final String getContainerId() {
-        return String.format("%s-%s", getName(), containerIdUuidPart);
     }
 
     /**
--- a/core/src/main/java/org/eclipse/hono/connection/ConnectionFactory.java
+++ b/core/src/main/java/org/eclipse/hono/connection/ConnectionFactory.java
@@ -13,6 +13,8 @@
 
 package org.eclipse.hono.connection;
 
+import java.util.UUID;
+
 import org.eclipse.hono.config.ClientConfigProperties;
 import org.eclipse.hono.connection.impl.ConnectionFactoryImpl;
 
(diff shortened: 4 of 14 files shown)


client/src/main/java/org/eclipse/hono/client/HonoConnection.java
```diff
@@ -82,6 +83,24 @@ static HonoConnection newConnection(final Vertx vertx, final ClientConfigPropert
...
+     * @param clientConfigProperties The client properties to use.
+     * @return The newly created connection. Note that the underlying AMQP connection will not be established
+     *         until one of its <em>connect</em> methods is invoked.
+     * @throws NullPointerException if vertx or clientConfigProperties is {@code null}.
+     */
+    static HonoConnection newConnection(final Vertx vertx, final ConnectionFactory connectionFactory,
+            final ClientConfigProperties clientConfigProperties) {
+        return new HonoConnectionImpl(vertx, connectionFactory, clientConfigProperties);
+    }
+
```
sophokles73, line 103: I do not really see why we need this additional constructor. In fact, I believe we once had one like this one but removed it because it was hard to make sure that the config props passed in are the same as the ones used for creating the connection factory. Also, `HonoConnectionImpl(Vertx, ClientConfigProperties)` seems to be doing the same thing as what you are doing in `AbstractAdapterConfig` or am I mistaken?

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -159,7 +159,11 @@ public final void setAmqpHostname(final String amqpHostname) {
      * The identifier consists of the name returned by {@link #getName()} plus a UUID.
      *
      * @return The identifier.
+     * @deprecated The preferred way to obtain the <em>container-id</em> for the AMQP connection is now via
+     *             the connection factory, so that not necessarily all connections configured with this properties
+     *             object get the same id, but rather only all connections of the same connection factory instance do.
      */
+    @Deprecated
```
sophokles73, line 166: Is there a particular reason why you do not want to simply remove this method?

calohmn: This is mainly for merging this commit as-is in 1.2.2. For 1.3, this method can then be removed, yes.

client/src/main/java/org/eclipse/hono/client/HonoConnection.java
calohmn: I liked the idea of having the `adapterInstanceId` also represented in the AMQP connection `container-id` advertised in the OPEN frame. So I wanted to keep that.

Before this PR, the `adapterInstanceId` / `container-id` UUID was initialized in the `ClientConfigProperties`. So, in order to have a unique `adapterInstanceId` per `ProtocolAdapterCommandConsumerFactory`, a new `ClientConfigProperties` instance was needed. This is somethat counterintuitive, therefore I opted for handling the `container-id` in the `ConnectionFactory` instead, using a new `ConnectionFactory` instance for each `ProtocolAdapterCommandConsumerFactory` instance.

This leads to the need for accessing the `ConnectionFactory` instance here:
https://github.com/eclipse/hono/blob/f2b20a9f8d21b25205410a3cd26574bc389d1d89/service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java#L452-L459
An alternative would be adding a `HonoConnection.getConnectionFactory()` method, but I preferred the extra HonoConnection factory method.

But yes, I see your point about the config props passed in that factory method.
As a solution, I guess we could either
- remove the `ClientConfigProperties` param from `HonoConnectionImpl(Vertx, ConnectionFactory, ClientConfigProperties)` and the factory method (and adding `ConnectionFactory.getConfig`).
- introduce `HonoConnection.getConnectionFactory()`.
- or, deem all this not worth the effort and handle `adapterInstanceId` independent of the used `container-id`, keeping `adapterInstanceId` internal to the `ProtocolAdapterCommandConsumerFactory` instance.

I think I would prefer the first option here - but would also be OK with the others. WDYT?

sophokles73: I am not sure if I can follow. My understanding is that we need to make sure that a call to `ProtocolAdapterCommandConsumerFactory.createCommandConsumer()` creates a receiver link with a unique source address of the form `command_internal/{unique_id}`. Now we would like to use the container-id of the underlying connection as the *unique_id*, right? Can't we then simply keep the container-id in a field of `HonoConnectionImpl` and make it available via e.g. `HonoConnection.getContainerId()` in order to use it when establishing the receiver link?

calohmn: Yes, that is also an option. I would let `HonoConnection.getContainerId()` call `connectionFactory.getContainerId()` then.
I've amended the commit here accordingly.

As I don't see a sensible default implementation of the new `HonoConnection.getContainerId()` method, I've set up the changes to be for 1.3 only (also removing unused methods).
For 1.2.2, I'll create a separate PR then, using the approach where the `ClientConfigProperties` get copied to get a separate `adapterInstanceId`.

client/src/main/java/org/eclipse/hono/client/HonoConnection.java
```diff
@@ -230,10 +230,19 @@ static HonoConnection newConnection(final Vertx vertx, final ClientConfigPropert
...
      */
-    default String getRemoteContainer() {
+    default String getRemoteContainerId() {
         return "N/A";
     }
 
+    /**
+     * Gets the container id as advertised to the peer.
+     * <p>
+     * The identifier consists of the name set in the configuration plus a UUID.
```
sophokles73, line 240: Is this necessarily the case? I would rather document the requirement regarding the ID's uniqueness instead of describing (and thus defining) how it needs to be created. WDYT?

client/src/main/java/org/eclipse/hono/client/HonoConnection.java
```diff
@@ -230,10 +230,19 @@ static HonoConnection newConnection(final Vertx vertx, final ClientConfigPropert
      *
      * @return The remote container id or {@code null}.
      */
-    default String getRemoteContainer() {
+    default String getRemoteContainerId() {
```
sophokles73, line 233: just wondering: shouldn't this better return `null` by default? Otherwise, what purpose does it have to allow for `null` to be returned in the first place?

core/src/main/java/org/eclipse/hono/connection/impl/ConnectionFactoryImpl.java
```diff
@@ -44,6 +45,7 @@
 public final class ConnectionFactoryImpl implements ConnectionFactory {
 
     private static final Logger logger = LoggerFactory.getLogger(ConnectionFactoryImpl.class);
+    private final String containerIdUuidPart = UUID.randomUUID().toString();
```
sophokles73, line 48: this means that all connections produced by this factory share the same UUID, right? Doesn't it make more sense that each connection has a distinct UUID?

core/src/main/java/org/eclipse/hono/connection/impl/ConnectionFactoryImpl.java
```diff
@@ -82,6 +84,11 @@ public String getName() {
         return config.getName();
     }
 
+    @Override
+    public String getContainerId() {
+        return String.format("%s-%s", getName(), containerIdUuidPart);
+    }
+
```
sophokles73, line 91: FMPOV each connection created by this factory should have its own unique container ID, e.g.
```
private String getContainerId() {
  return String.format("%s-%s-%s", config.getName(), config.getServerRole(), UUID.randomUUID())
}
```

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -1043,13 +1043,18 @@ public final void disconnect(final Handler<AsyncResult<Void>> completionHandler)
...
+    public String getRemoteContainerId() {
         if (!isConnectedInternal()) {
             return null;
         }
         return connection.getRemoteContainer();
     }
 
+    @Override
+    public String getContainerId() {
+        return connectionFactory.getContainerId();
```
sophokles73, line 1055: FMPOV this should return the (local) container ID of the ProtonConnection instead ...

core/src/main/java/org/eclipse/hono/connection/ConnectionFactory.java
```diff
@@ -28,12 +30,23 @@
...
+     * Gets the identifier being indicated as the <em>container-id</em> in the client's AMQP <em>Open</em> frame.
+     * <p>
+     * The identifier consists of the name returned by {@link #getName()} plus a UUID.
+     *
+     * @return The identifier.
+     */
+    default String getContainerId() {
+        return String.format("%s-%s", getName(), UUID.randomUUID());
+    }
+
```
sophokles73, line 49: this is not needed FMPOV

core/src/main/java/org/eclipse/hono/connection/impl/ConnectionFactoryImpl.java
calohmn: Letting `ConnectionFactoryImpl.connect()` assign a new random container ID to each newly created `ProtonConnection` will prevent us from using it as the `adapterInstanceId` though. Otherwise a reconnect on the `HonoConnection` used by the `ProtocolAdapterCommandConsumerFactory` would cause the `adapterInstanceId` to be different from the container id.

Sticking to the plan of keeping `adapterInstanceId` and container id the same would mean having the container id stored in the `HonoConnectionImpl` instance and setting that same id on every `ProtonConnection` instance that gets created via the `HonoConnectionImpl` instance.
I think that is a valid approach - the `HonoConnection` only has one open `ProtonConnection` at a time, so there wouldn't be 2 open `ProtonConnection`s with the same container id.

To do this, I would add a `ConnectionFactory.connect` method with an extra container id parameter.

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
calohmn: Presuming that we want to use the same container ID for all `ProtonConnection`s created by this `HonoConnectionImpl` (see discussion below):
Using `connection.getContainer()` would complicate things for the scenario here as that would only work if the connection has already been established. I would rather define `HonoConnection.getContainerId()` as the container id that _has been_ or that _will be_ advertised to the remote peer. The implementation would simply return the value of a new `HonoConnectionImpl.containerId` field here.

client/src/main/java/org/eclipse/hono/client/HonoConnection.java
calohmn: I think we can remove the default implementation here - it was only added for compatibility reasons.

client/src/main/java/org/eclipse/hono/client/HonoConnection.java
calohmn: Right, I've adapted the javadoc accordingly.

core/src/main/java/org/eclipse/hono/connection/ConnectionFactory.java
calohmn: I've removed it.

core/src/main/java/org/eclipse/hono/connection/impl/ConnectionFactoryImpl.java
calohmn: I've committed the aforementioned changes. I've also removed `ConnectionFactory.getName()`  - its javadoc stated that the name is used in the container id - that is not _necessarily_ the case anymore because the container id is now provided as an extra param to `ConnectionFactory.connect()`.

core/src/main/java/org/eclipse/hono/connection/impl/ConnectionFactoryImpl.java
calohmn: In practise I guess we mostly use one factory instance only for one `HonoConnection` instance, that would mean a `HonoConnection` instance gets a distinct UUID. But you are right, it would make more sense to enforce this by keeping the UUID in the `HonoConnection`.
I don't think we need a distinct UUID per `ProtonConnection` - see my other comment.

calohmn:
(Travis run is green actually - I guess it wasn't updated here properly because of the GitHub downtime yesterday.)

sophokles73:
@calohmn can you resolve the conflicts?

site/homepage/content/release-notes.md
```diff
@@ -8,6 +8,10 @@ title = "Release Notes"
 
 * An admin guide for the CoAP adapter has been added.
 
+### API Changes
+
+* The `getRemoteContainer` method in `org.eclipse.hono.client.HonoConnection` has been renamed to `getRemoteContainerId`.
```
sophokles73, line 13: maybe we also want to mention the changes to `ConnectionFactory`?

calohmn: Yes, I've added that.

Issue #1790, opened 2020-02-24
Optional custom mapper


BobClaerhout:
After the discussion in the F2F last week, I hereby present you a brief overview of the discussed issue. Hopefully, I haven't got anything wrong or misinterpreted. Feel free to comment and complete where necessary.
# Problem
Currenlty, Hono is payload agnostic. It does not care what payload is sent and just transparantly passes it to the business application.  
However, sometimes the payload contains information which might be useful in some cases. When the payload contains the deviceId, it is useful to do a (partial) mapping in Hono of this payload in order to authorize the gateway-device combination. If this is possible, you can take advantage of the via property in Hono and authorize the gateway-device combination.  
If (partial) mapping of the payload becomes possible, it might make sense to simplify the lora and sigfox adapters again (for simplification I'll only write down lora adapter but in all cases, this should relate to the sigfox adapter as well). Right now, the only difference between the lora-adapter and the http-adapter is the lora specific mapping of the metadata. If an optional hook would be introduced in hono, the lora adapter could be removed and merged with the http adapter. The optional hook for mapping metadata and payload can be specified for lora, sigfox or a custom implementation.

# Proposal
![Proposal](https://user-images.githubusercontent.com/3315222/75165096-23d6b680-5722-11ea-9b1f-eb575fcd50fa.png)
* Configure the hono/tenant/gateway/device (the more finegrained configuration overwrites the parent one) with a mapper and a version. This could be in the defaults section or a new attribute (my preference goes to a new attribute to keep the defaults attribute customizable).
* Configure mapper in adapter (via config map in all adapters) to include:
  * Mapper name
  * Mapper endpoint (HTTP or AMQP)
* Add lora mapper
* Add sigfox mapper
* Every mapper should return deviceId, metadata, payload (might be changed in the mapper)
* Optional: maybe we can derive the mapper from the used url to which the telemtry/event has been published.

BobClaerhout:
@sophokles73, @ctron WDYT?
Any remarks on the design? Is this enough for you and can we start implementing it?

ctron:
I am not sure about the configuration, mostly because I don't understand what you have in mind.

>*  Configure mapper in adapter (via config map in all adapters) to include:
>     * Mapper name
>     * Mapper endpoint (HTTP or AMQP)

I would expect the tenant/gateway/device to have some kind of "mapper" endpoint configured. If it does have that, all traffic would be run through that mapper before forwarding it to the downstream message endpoint.

[...]

PR #1862, opened 2020-03-30
custom mapper

BobClaerhout:
WIP on the custom mapper: https://github.com/eclipse/hono/issues/1790
Remarks welcome to get this done correctly instead of one big commit which requires a lot of refactoring/rework.


BobClaerhout:
Issue: https://github.com/eclipse/hono/issues/1790

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
```diff
@@ -229,4 +231,15 @@ public Duration getTenantIdleTimeout() {
...
     }
+
+    /**
+     * Gets the configured mappers for this adapter
+     * <p>
+     * The default value of this property is an empty hashMap. Ths results in not having custom mapping at all.
+     *
+     * @return HashMap of mapper endpoints
+     */
+    public HashMap<String, MapperEndpoint> getMapperEndpoints() {
```
sophokles73: how do you set these?

core/src/main/java/org/eclipse/hono/util/RegistrationConstants.java
```diff
@@ -71,6 +71,14 @@
      */
     public static final String FIELD_MEMBER_OF = "memberOf";
 
+    /**
+     * The name of the field in a device registration information that contains the identifier of the names
+     * of the groups in which the device is member of (either as string value or inside a JSON array).
+     * Note, that it is only possible to either set the 'via' or the 'memberOf' property since groups of groups
+     * are currently not supported.
```
sophokles73: This seems to be the description of the *memberOf* field ...

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
```diff
@@ -14,27 +14,42 @@
...
+import io.vertx.mqtt.messages.MqttPublishMessage;
 
 /**
  * A Vert.x based Hono protocol adapter for publishing messages to Hono's Telemetry and Event APIs using MQTT.
  */
 public final class VertxBasedMqttProtocolAdapter extends AbstractVertxBasedMqttProtocolAdapter<MqttProtocolAdapterProperties> {
 
+    private static final String HEXES = "0123456789ABCDEF";
+
+    static String getHex(final byte[] raw) {
```
sophokles73: better use *com.google.common.io.BaseEncoding.base16().encode(byte[])*

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
```diff
@@ -96,4 +112,38 @@ protected String getTypeName() {
...
     }
+
+    private Future<ResourceIdentifier> mapMessage(final MqttContext ctx, final ResourceIdentifier targetAddress, final MqttPublishMessage message) {
+        return getRegistrationAssertion(targetAddress.getTenantId(), targetAddress.getResourceId(),
+                ctx.authenticatedDevice(), ctx.getTracingContext()).compose(registratonInfo -> {
+            final Promise<ResourceIdentifier> result = Promise.promise();
+            if (!registratonInfo.containsKey(RegistrationConstants.FIELD_MAPPER)) {
+                result.complete(targetAddress);
+                return result.future();
+            }
```
sophokles73: The response to *get Registration Assertion* only contains the *default* properties to be included in all downstream messages. I don't think that this field should be configured as a *default* property, though ...

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
```diff
@@ -96,4 +112,38 @@ protected String getTypeName() {
...
+            final HashMap<String, MapperEndpoint> mapperEndpoints = getConfig().getMapperEndpoints();
+            if (!mapperEndpoints.containsKey(mapper)) {
+                result.complete(targetAddress);
+                return result.future();
+            }
+            final MapperEndpoint mapperEndpoint = mapperEndpoints.get(mapper);
+            final WebClient webClient = WebClient.create(vertx);
+            webClient.get(mapperEndpoint.getPort(), mapperEndpoint.getHost(), mapperEndpoint.getUri())
+                    .addQueryParam("payload", getHex(message.payload().getBytes()))
```
sophokles73: I would rather do a POST and include the payload as a plain byte array in the request body so that there is no need to encode the data to a HEX string.

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
```diff
@@ -96,4 +112,38 @@ protected String getTypeName() {
...
+            if (!registratonInfo.containsKey(RegistrationConstants.FIELD_MAPPER)) {
+                result.complete(targetAddress);
+                return result.future();
+            }
+            final String mapper = registratonInfo.getString(RegistrationConstants.FIELD_MAPPER);
+            if (mapper.isBlank()) {
+                result.complete(targetAddress);
+                return result.future();
+            }
+            final HashMap<String, MapperEndpoint> mapperEndpoints = getConfig().getMapperEndpoints();
```
sophokles73: Do we really need to configure the mapping endpoints *per device*? Will they be different for all devices or can't we simply configure the (limited set of) mapper endpoints at the tenant level instead?

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
BobClaerhout: Did not run it yet but was hoping for the spring-boot to provide this via the properties file

core/src/main/java/org/eclipse/hono/util/RegistrationConstants.java
BobClaerhout: right, forgot to change it.

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
BobClaerhout: Ok, this is valuable information to me. I was indeed not trying to get the _defaults_ property but trying to get the (newly created) mapper field as you see below. How do I get these properties?

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
BobClaerhout: I agree, better this way

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
BobClaerhout: In our case, we might have different mapping endpoint per device in the same tenant.
Also, I thought this configuration is on the protocol adapter (which supports multiple tenants) so a list of mapperEndpoints should be injected here or am I wrong?

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
BobClaerhout: Removed this code as we are now sending the buffer. No hex needed

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
sophokles73: Let me try to understand what you want to do first. The device that has been authenticated at this point is a gateway, right? You now want to retrieve the configuration data that has been registered for the gateway in order to determine the URL to invoke for mapping payload data, right?

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
sophokles73: you better try this out :-)

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
BobClaerhout: yes, that's right.

sophokles73: Ok, then you would probably want to access data that has been registered under `ext` for the gateway. However, the Device Registration API doesn't support this for now. We can, of course, change it to either support a dedicated *get Registration* operation or we also include the `ext` data in the response to the *get Registration Assertion* operation. We would need to make that optional, however, for the time being. Otherwise, we would introduce an implementation breaking change to the Device Registration API. This means that the whole mapping functionality that you want to add would only work with a device registry that supports returning the `ext` properties ...

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
BobClaerhout: I tried it out and it works :)
Following configuration results in 2 mapperEndpoints being configured in the (mqtt) adapter:
```
  mqtt:
    .....
    mapperEndpoints:
      test:
        host: host
        port: 123
        uri: uri
      test2:
        host: host2
        port: 321
        uri: uri2
```

sophokles73: Now you will only need to find a way to document this :-)

site/documentation/content/admin-guide/mqtt-adapter-config.md
```diff
@@ -40,6 +40,9 @@ The following table provides an overview of the configuration variables and corr
...
 | `HONO_MQTT_SECURE_PROTOCOLS`<br>`--hono.mqtt.secureProtocols` | no | `TLSv1.2` | A (comma separated) list of secure protocols that are supported when negotiating TLS sessions. Please refer to the [vert.x documentation](https://vertx.io/docs/vertx-core/java/#ssl) for a list of supported protocol names. |
 | `HONO_MQTT_TENANT_IDLE_TIMEOUT`<br>`--hono.mqtt.tenantIdleTimeout` | no | `0ms` | The duration after which the protocol adapter removes local state of the tenant (e.g. open AMQP links) with an amount and a unit, e.g. `2h` for 2 hours. See the [Spring Boot documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-conversion-duration) for an explanation of the format. The value `0ms` disables the timeout. |
+| `hono.mqtt.mapperEndpoints.<mapperName>.host`<br>`--hono.mqtt.mapperEndpoints.<mapperName>.host` | no | `mapper-hono.svc` | The host to which the adapter should connect to call the custom mapper. Replace `<mapperName>` with the name of the mapper to configure. |
```
sophokles73: Environment variable names are upper case and use an underscore instead of `.`

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
BobClaerhout: new day, new insights. Shouldn't have added the underscore in `MAPPER_ENDPOINTS` but left it as is: `MAPPERENDPOINTS`. Now it's wrorking. Documentation also updated.

site/documentation/content/admin-guide/mqtt-adapter-config.md
BobClaerhout: Fixed.

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
BobClaerhout: Why can't we get it the same as the VIA property retrieved in the following line:
https://github.com/eclipse/hono/blob/9a37b826a2dff6ed800c73966539ebb8cfb8eb1b/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java#L1108

BobClaerhout: I've changed the device registry GetAssertionPayload to provide it. Question remains: is this allowed?

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
sophokles73: can you also add a setter so that this can be used in unit tests without requiring reflection?

sophokles73: please also make the getter and setter final

core/src/main/java/org/eclipse/hono/config/MapperEndpoint.java
```diff
@@ -0,0 +1,82 @@
...
+ *******************************************************************************/
+
+package org.eclipse.hono.config;
+
+import java.util.Objects;
+
+/**
+ * Configuration class for custom mappers.
+ */
+public class MapperEndpoint {
```
sophokles73: final?

core/src/main/java/org/eclipse/hono/config/MapperEndpoint.java
```diff
@@ -0,0 +1,82 @@
...
+        return host;
+    }
+
+    /**
+     * Sets the host name or IP address of this mapper.
+     *
+     * @param host The host name or IP address.
+     * @throws NullPointerException if host is {@code null}.
+     */
+    public void setHost(final String host) {
```
sophokles73: how about making this class immutable and provide a static factory method that accepts all three properties?

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MappedMessage.java
```diff
@@ -0,0 +1,17 @@
+package org.eclipse.hono.adapter.mqtt.impl;
+
+import io.vertx.mqtt.messages.MqttPublishMessage;
+import org.eclipse.hono.adapter.mqtt.MqttContext;
+import org.eclipse.hono.util.ResourceIdentifier;
+
+public class MappedMessage {
```
sophokles73: This seems to be a class that is only used within this package so I propose to declare it and its properties package private in order to prevent other code to start using an depending on it ...

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
sophokles73: I see. However, instead of exposing the Map to this class here, what about adding a *getMapperEndpoint(String)* to the `ProtocolAdapterProperties` class instead?

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
```diff
@@ -96,4 +111,48 @@ protected String getTypeName() {
...
+                            String deviceId = targetAddress.getResourceId();
+                            if (headers.contains(APP_PROPERTY_DEVICE_ID)){
+                                deviceId = headers.get(APP_PROPERTY_DEVICE_ID);
+                            }
+                            result.complete(
+                                    new MappedMessage(ctx, ResourceIdentifier.from(targetAddress,
+                                            targetAddress.getTenantId(), deviceId), mqttPublishMessage)
+                            );
+                        }
+                    });
```
sophokles73: this should probably go into its own method ...

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
```diff
@@ -254,4 +243,17 @@ public void setTenantIdleTimeout(final Duration tenantIdleTimeout) {
...
+    /**
+     * Gets the configured mapper for the given key.
+     *
+     * @param key The key to identify the mapper
+     * @return the mapperEndpoint. If not found, will return {@code null}.
+     */
+    public MapperEndpoint getMapperEndpoint(final String key) {
+        if (!mapperEndpoints.containsKey(key)) {
+            return null;
+        }
```
sophokles73: I would assume that this can be safely removed, can't it?

core/src/main/java/org/eclipse/hono/config/MapperEndpoint.java
BobClaerhout: Doesn't work with the spring boot properties binding.

core/src/main/java/org/eclipse/hono/config/MapperEndpoint.java
```diff
@@ -79,4 +79,20 @@ public String getUri() {
...
+     * @param uri The uri to use when requesting a mapping
+     * @return the mapperEndpoint configuration
+     */
+    public static MapperEndpoint from(final String host, final int port, final String uri) {
+        final MapperEndpoint ep = new MapperEndpoint();
+        ep.host = host;
+        ep.port = port;
+        ep.uri = uri;
+        return ep;
+    }
```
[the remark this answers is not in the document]
BobClaerhout: Needed for spring boot properties binding.

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
BobClaerhout: If somebody configures a mapper on a device which doesn't exist we will get a keyNotFoundException if we ommit this check? Apart from the fact that you can catch that exception I rather not have an exception at all.

sophokles73: If I am not mistaken then [Map.get()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html#get(java.lang.Object)) returns `null` already if the map doesn't contain the given key ...

BobClaerhout: You're right. Habit from another language :)

adapters/mqtt-vertx/src/test/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapterTest.java
```diff
@@ -257,6 +257,47 @@ public void testMapTopicSupportsShortAndLongTopicNames(final VertxTestContext ct
...
+        context = newContext(message, null);
+        adapter.mapTopic(context).setHandler(ctx.succeeding(address -> {
+            ctx.verify(() -> assertThat(MetricsTags.EndpointType.fromString(address.getEndpoint())).isEqualTo(MetricsTags.EndpointType.TELEMETRY));
+        }));
+
+        message = newMessage(MqttQoS.AT_LEAST_ONCE, "unknown");
+        context = newContext(message, null);
+        adapter.mapTopic(context).setHandler(ctx.failing());
+        ctx.completeNow();
+
```
sophokles73, line 298: I'd rather have a single combination of qos and endpoint but have a real test instead of all combinations without any assertions.
For that, I guess the `WebClient` needs to be a field of the adapter so that we can easily use a mock in order to verify the adapter's interactions with it ...

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/device/AbstractRegistrationService.java
```diff
@@ -343,6 +343,10 @@ private JsonArray convertObjectToJsonArray(final Object object) {
                     if (!via.isEmpty()) {
                         result.put(RegistrationConstants.FIELD_VIA, via);
                     }
+                    if (registrationInfo.containsKey(RegistrationConstants.FIELD_MAPPER)) {
+                        result.put(RegistrationConstants.FIELD_MAPPER,
+                                registrationInfo.getString(RegistrationConstants.FIELD_MAPPER));
+                    }
```
sophokles73, line 349: please do not look up the key twice ...

adapters/mqtt-vertx/src/test/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapterTest.java
BobClaerhout: I agree. Test was still todo though. Will do that today and will create a field for the web client indeed.

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
```diff
@@ -49,6 +52,23 @@ protected String getTypeName() {
...
+        }
+        return webClient;
+    }
+
+    /**
+     * Sets the vertx webClient to use when mapping payload.
+     *
+     * @param webClient the webClient to set
+     */
+    public void setWebClient(final WebClient webClient) {
```
sophokles73, line 67: this should be autowired and injected via Spring FMPOV

sophokles73: In fact, I believe that it makes a lot of sense to create a dedicated interface for message mapping that hides away the specifics of how the mapper is invoked (HTTP in this case) and which also encapsulates the configuration of the supported mapper endpoints. WDYT?

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
```diff
@@ -147,28 +148,33 @@ public void setWebClient(final WebClient webClient) {
                 return result.future();
             }
 
-            return mapMessageRequest(ctx, targetAddress, message, mapperEndpoint);
+            final VertxHttpHeaders headers = new VertxHttpHeaders();
+            registratonInfo.iterator().forEachRemaining(stringObjectEntry -> headers.add(stringObjectEntry.getKey(),
+                    stringObjectEntry.getValue()));
+
```
sophokles73, line 154: what is this good for?

BobClaerhout: It allows for some configuration in the defaults to be passed on to the mapper. I think this can be useful if you're writing your own mapping.

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
BobClaerhout: I agree. I've created another class for this. However, I can't seem to autowire the `WebClient` in it. Can you point me to where I should configure this?

adapters/mqtt-vertx/src/test/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapterTest.java
BobClaerhout: I added a more effective test now.

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
BobClaerhout: Done

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
sophokles73: The *defaults* have well defined semantics and I would rather not encourage people to abuse them for something else. If we need to be able to forward properties that have been registered for the device then these should probably come from the *ext* properties ...

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,111 @@
...
+public class MessageMapping {
+
+    private WebClient webClient;
+
+    private WebClient getWebClient() {
+        if (webClient == null) {
+            webClient = WebClient.create(null);
+        }
+        return webClient;
+    }
```
sophokles73, line 49: we don't need this FMPOV

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * Sets the vertx webClient to use when mapping payload.
+     *
+     * @param webClient the webClient to set
+     */
+    @Autowired
+    public void setWebClient(final WebClient webClient) {
+        this.webClient = webClient;
+    }
```
sophokles73, line 59: IMHO we should inject this via the class' constructor so we can be sure that it will always be set.

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,111 @@
...
+import io.vertx.ext.web.client.HttpResponse;
+import io.vertx.ext.web.client.WebClient;
+import io.vertx.mqtt.messages.MqttPublishMessage;
+import io.vertx.mqtt.messages.impl.MqttPublishMessageImpl;
+
+/**
+ * A Vert.x based message mapping component.
+ */
+@Component
+public class MessageMapping {
```
sophokles73, line 40: final?

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,111 @@
...
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.client.HttpResponse;
+import io.vertx.ext.web.client.WebClient;
+import io.vertx.mqtt.messages.MqttPublishMessage;
+import io.vertx.mqtt.messages.impl.MqttPublishMessageImpl;
+
+/**
+ * A Vert.x based message mapping component.
+ */
+@Component
```
sophokles73, line 39: I'd rather not annotate this but instead create an instance of it in the `Config` class

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,111 @@
...
+     * Sets the vertx webClient to use when mapping payload.
+     *
+     * @param webClient the webClient to set
+     */
+    @Autowired
+    public void setWebClient(final WebClient webClient) {
+        this.webClient = webClient;
+    }
+
+    Future<MappedMessage> mapMessage(final MqttContext ctx, final ResourceIdentifier targetAddress,
```
sophokles73, line 61: public ?
could also use a proper JavaDoc comment

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        final VertxHttpHeaders headers = new VertxHttpHeaders();
+        registratonInfo.iterator().forEachRemaining(stringObjectEntry -> headers.add(stringObjectEntry.getKey(),
+                stringObjectEntry.getValue()));
+
+        return mapMessageRequest(ctx, targetAddress, message, mapperEndpoint, headers);
+    }
+
+    Future<MappedMessage> mapMessageRequest(final MqttContext ctx, final ResourceIdentifier targetAddress,
```
sophokles73, line 87: private?

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
```diff
@@ -19,29 +19,21 @@
...
 import io.vertx.mqtt.messages.MqttPublishMessage;
-import io.vertx.mqtt.messages.impl.MqttPublishMessageImpl;
+
 /**
  * A Vert.x based Hono protocol adapter for publishing messages to Hono's Telemetry and Event APIs using MQTT.
  */
 public final class VertxBasedMqttProtocolAdapter extends AbstractVertxBasedMqttProtocolAdapter<MqttProtocolAdapterProperties> {
-
-    private WebClient webClient;
+    MessageMapping messageMapping;
```
sophokles73, line 36: private

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
```diff
@@ -88,6 +73,12 @@ public void setWebClient(final WebClient webClient) {
                 });
     }
 
+    private Future<MappedMessage> mapMessage(final MqttContext ctx, final ResourceIdentifier targetAddress,
+                                 final MqttPublishMessage message) {
+        return getRegistrationAssertion(targetAddress.getTenantId(), targetAddress.getResourceId(),
+                ctx.authenticatedDevice(), ctx.getTracingContext()).compose(registratonInfo -> messageMapping.mapMessage(ctx, targetAddress, message, registratonInfo, getConfig()));
```
sophokles73, line 79: what if no *MessageMapping* instance has been set?

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
BobClaerhout: Ok. We are currently using those `defaults` to configure our custom mapping after the adapters because the `ext` properties are not passed on.
We are coming back to the discussion on Gitter... Also the reason why I asked you on Gitter. Can we just pass on all `ext` properties from now on?

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
BobClaerhout: package-private for testing

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
BobClaerhout: Done

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
BobClaerhout: Yes, removed

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
BobClaerhout: Ok, changed.

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
BobClaerhout: Fixed that

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
BobClaerhout: done

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
BobClaerhout: added a check

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
BobClaerhout: Done

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -87,7 +86,7 @@ public void setWebClient(final WebClient webClient) {
     Future<MappedMessage> mapMessageRequest(final MqttContext ctx, final ResourceIdentifier targetAddress,
                                             final MqttPublishMessage message, final MapperEndpoint mapperEndpoint, final VertxHttpHeaders headers) {
         final Promise<MappedMessage> result = Promise.promise();
-        getWebClient().post(mapperEndpoint.getPort(), mapperEndpoint.getHost(), mapperEndpoint.getUri())
+        webClient.post(mapperEndpoint.getPort(), mapperEndpoint.getHost(), mapperEndpoint.getUri())
```
sophokles73, line 89: it looks like the result will never be completed if the HTTP request fails ...

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
```diff
@@ -75,6 +75,11 @@ public void setMessageMapping(final MessageMapping messageMapping) {
 
     private Future<MappedMessage> mapMessage(final MqttContext ctx, final ResourceIdentifier targetAddress,
                                  final MqttPublishMessage message) {
+        if (messageMapping == null) {
```
sophokles73, line 78: is the *MessageMapping* bean required or not? In the Confiugration you create the bean so it will get injected. The *setMessageMapping* method's `@Autowired` annotation also requires a mapping to be injected (otherwise it should be `@Autowired(required = false)`. That setter should also throw a NullPointerException if the given mapping is `null` ...

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
BobClaerhout: good catch. It should now

BobClaerhout:
I have created a service which we will use which is listening for these request so I can properly test now

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -103,6 +103,8 @@ public MessageMapping(final WebClient webClient) {
                                 new MappedMessage(ctx, ResourceIdentifier.from(targetAddress,
                                         targetAddress.getTenantId(), deviceId), mqttPublishMessage, responseHeaders)
                         );
+                    } else {
+                        result.complete(new MappedMessage(ctx, targetAddress, message));
```
sophokles73, line 107: this probably deserves a (debug) log statement?

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MappedMessage.java
```diff
@@ -0,0 +1,106 @@
...
+        this.resource = resource;
+        this.message = message;
+        this.mapperResponseHeaders = mapperResponseHeaders;
+        this.mappedExternally = mappedExternally;
+    }
+
+    /**
+     * Creates a new mappedMessage which is actually not mapped by an external service.
+     *  @param ctx The context used for this message
+     * @param resource The resourceIdentifier used for this message
```
sophokles73, line 44: formatting

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MappedMessage.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param message The actual message with its payload
+     * @param mapperResponseHeaders The headers returned from the mapper service
+     */
+    MappedMessage(final MqttContext ctx, final ResourceIdentifier resource, final MqttPublishMessage message, final MultiMap mapperResponseHeaders) {
+        this(ctx, resource, message, mapperResponseHeaders, true);
+    }
+
+    /**
+     * Creates a new mappedMessage.
+     *  @param ctx The context used for this message
```
sophokles73, line 54: formatting

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MappedMessage.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    MappedMessage(final MqttContext ctx, final ResourceIdentifier resource, final MqttPublishMessage message) {
+        this(ctx, resource, message, null, false);
+    }
+        /**
+         * Gets the context.
+         *
+         * @return the context
+         */
+    MqttContext getCtx() {
```
sophokles73, line 66: formatting

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MappedMessage.java
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * Gets the headers returned by the mapper.
+     *
+     * @return headers returned from the mapper service
+     */
+
+    public MultiMap getMapperResponseHeaders() {
+        return mapperResponseHeaders;
+    }
```
sophokles73, line 96: formatting

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,115 @@
...
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.impl.headers.VertxHttpHeaders;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.client.HttpResponse;
+import io.vertx.ext.web.client.WebClient;
+import io.vertx.mqtt.messages.MqttPublishMessage;
+import io.vertx.mqtt.messages.impl.MqttPublishMessageImpl;
+
+/**
+ * A Vert.x based message mapping component.
```
sophokles73, line 35: Please provide a more meaningful description which indicates the purpose of this class. Nobody knows what a *mapping component* is ...

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,115 @@
...
+     * @return Mapped message
+     */
+    public Future<MappedMessage> mapMessage(final MqttContext ctx, final ResourceIdentifier targetAddress,
+                                     final MqttPublishMessage message, final JsonObject registrationInfo, final MqttProtocolAdapterProperties config) {
+        final Promise<MappedMessage> result = Promise.promise();
+        if (!registrationInfo.containsKey(RegistrationConstants.FIELD_MAPPER)) {
+            result.complete(new MappedMessage(ctx, targetAddress, message));
+            return result.future();
+        }
+        final String mapper = registrationInfo.getString(RegistrationConstants.FIELD_MAPPER);
```
sophokles73, line 67: this will throw a ClassCastException if the property is not a string. IMHO you should use a more defensive approach here ...

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,115 @@
...
+                        final HttpResponse<Buffer> httpResponse = httpResponseAsyncResult.result();
+                        if (httpResponse.statusCode() == 200) {
+                            final MqttPublishMessageImpl mqttPublishMessage = new MqttPublishMessageImpl(ctx.message().messageId(), ctx.message().qosLevel(),
+                                    ctx.message().isDup(), ctx.message().isRetain(), ctx.message().topicName(),
+                                    httpResponse.bodyAsBuffer().getByteBuf());
+                            final MultiMap responseHeaders = httpResponse.headers();
+                            String deviceId = targetAddress.getResourceId();
+                            if (headers.contains(MessageHelper.APP_PROPERTY_DEVICE_ID)) {
+                                deviceId = headers.get(MessageHelper.APP_PROPERTY_DEVICE_ID);
+                            }
```
sophokles73, line 103: If I am not mistaken this should try to retrieve the (mapped) device ID from the response headers, shouldn't it?

BobClaerhout:
@sophokles73, @ctron 
This is ready for actual review now. It is deployed in our test cluster and working properly.  
I'm very interested in any remaining remarks and/or thoughts. Shoot

calohmn:
@BobClaerhout Can you rebase? 
The `Device`, `DeviceTest` and `RegistrationServiceTests` classes have been moved (see https://github.com/eclipse/hono/commit/bd1d5c019bc5c16a81e63c9d7390e9e709387721).
Further changes in `RegistrationServiceTests` include all occurrences of `.setHandler(` having been replaced with `.onComplete(` (you can adopt your changes and do a search & replace).

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/Config.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
```
sophokles73, line 2: ????

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/Config.java
```diff
@@ -122,4 +123,15 @@ public ObjectFactoryCreatingFactoryBean serviceFactory() {
         factory.setTargetBeanName(BEAN_NAME_VERTX_BASED_MQTT_PROTOCOL_ADAPTER);
         return factory;
     }
+
+    /**
+     * Constructs messageMapping.
+     *
+     * @return Returns constructed messageMapping
```
sophokles73, line 130: I think you can do better than this ...

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MappedMessage.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.eclipse.hono.adapter.mqtt.MqttContext;
+import org.eclipse.hono.util.ResourceIdentifier;
+
+import io.vertx.mqtt.messages.MqttPublishMessage;
+
+/**
+ * Structure to provide a mapped message.
+ */
+class MappedMessage {
```
sophokles73, line 24: final ?

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MappedMessage.java
```diff
@@ -0,0 +1,67 @@
...
+    private final MqttContext ctx;
+    private final ResourceIdentifier resource;
+    private final MqttPublishMessage message;
+
+    /**
+     * Creates a new mappedMessage.
+     * @param ctx The context used for this message
+     * @param resource The resourceIdentifier used for this message
+     * @param message The actual message with its payload
```
sophokles73, line 33: I think you can do better than this.
Also, please insert an empty line between description and params.
Please finish the param description with a `.`

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,141 @@
...
+import io.vertx.core.json.Json;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.client.HttpResponse;
+import io.vertx.ext.web.client.WebClient;
+import io.vertx.mqtt.messages.MqttPublishMessage;
+import io.vertx.mqtt.messages.impl.MqttPublishMessageImpl;
+
+/**
+ * A Vert.x based message mapping component. This component requests mapping from another server over HTTP(S) if this
+ * is configured properly. The headers are overwritten with the result of the mapper (which includes the resourceId).
```
sophokles73, line 40: It would be helpful to understand what is being mapped and why, e.g. describe a common use case.

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,141 @@
...
+ */
+public final class MessageMapping {
+
+    private final Logger log = LoggerFactory.getLogger(getClass());
```
sophokles73, line 44: make this static

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,141 @@
...
+
+    private WebClient webClient;
```
sophokles73, line 46: final

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * Constructs the messageMapping client used to call external/custom messageMapping.
+     *
+     * @param webClient Vert.x webclient to use in the messageMapping
```
sophokles73, line 51: what if webclient is `null`?

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,141 @@
...
+     * Fetches the mapper if configured and calls the external mapping service.
+     *
+     * @param ctx The mqtt context
+     * @param targetAddress The resourceIdentifier with the current targetAddress
+     * @param message Received message
+     * @param registrationInfo information retrieved from the device registry
+     * @param config Mqtt protocol adapter properties
+     * @return Mapped message
+     */
+    public Future<MappedMessage> mapMessage(final MqttContext ctx, final ResourceIdentifier targetAddress,
```
sophokles73, line 67: are the params allowed to be `null`?

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,141 @@
...
+import io.vertx.ext.web.client.HttpResponse;
+import io.vertx.ext.web.client.WebClient;
+import io.vertx.mqtt.messages.MqttPublishMessage;
+import io.vertx.mqtt.messages.impl.MqttPublishMessageImpl;
+
+/**
+ * A Vert.x based message mapping component. This component requests mapping from another server over HTTP(S) if this
+ * is configured properly. The headers are overwritten with the result of the mapper (which includes the resourceId).
+ */
+public final class MessageMapping {
```
sophokles73, line 42: it would be much easier to test the interaction of the protocol adapter with the mapping service if this were an interface which can then be mocked in unit tests ...

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,141 @@
...
+     *
+     * @param ctx The mqtt context
+     * @param targetAddress The resourceIdentifier with the current targetAddress
+     * @param message Received message
+     * @param registrationInfo information retrieved from the device registry
+     * @param config Mqtt protocol adapter properties
+     * @return Mapped message
+     */
+    public Future<MappedMessage> mapMessage(final MqttContext ctx, final ResourceIdentifier targetAddress,
+                                     final MqttPublishMessage message, final JsonObject registrationInfo, final MqttProtocolAdapterProperties config) {
```
sophokles73, line 68: why do we want to pass in the static adapter config properties to each method invocation instead of keeping them in a field populated by the constructor?

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
```diff
@@ -0,0 +1,141 @@
...
+            return result.future();
+        }
+
+        final MapperEndpoint mapperEndpoint = config.getMapperEndpoint(mapper);
+        if (mapperEndpoint == null) {
+            result.complete(new MappedMessage(ctx, targetAddress, message));
+            return result.future();
+        }
+
+        final VertxHttpHeaders headers = new VertxHttpHeaders();
```
sophokles73, line 93: use `io.vertx.core.MultiMap` instead ...

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
```diff
@@ -96,4 +144,8 @@ protected String getTypeName() {
         }
         return result.future();
     }
+
+
+
+
```
sophokles73, line 150: ??

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
```diff
@@ -229,4 +231,26 @@ public Duration getTenantIdleTimeout() {
...
+
+    /**
+     * Sets the configured mappers for this adapter
+     * <p>
+     * Setting this property to an empty hashMap will disable mapping for this adapter.
+     *
+     * @param mapperEndpoints The new hashMap with mapper endpoints
+     * @throws NullPointerException if parameter is {@code null}.
+     */
+    public final void setMapperEndpoints(final HashMap<String, MapperEndpoint> mapperEndpoints) {
```
sophokles73, line 243: use `Map` as parameter type

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
```diff
@@ -34,6 +35,7 @@
     private long eventLoopBlockedCheckTimeout = 5000L;
     private int maxConnections = 0;
     private Duration tenantIdleTimeout = DEFAULT_TENANT_IDLE_TIMEOUT;
+    private HashMap<String, MapperEndpoint> mapperEndpoints = new HashMap<>();
```
sophokles73, line 38: use `Map` as type

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
```diff
@@ -229,4 +231,26 @@ public Duration getTenantIdleTimeout() {
...
+    /**
+     * Sets the configured mappers for this adapter
+     * <p>
+     * Setting this property to an empty hashMap will disable mapping for this adapter.
+     *
+     * @param mapperEndpoints The new hashMap with mapper endpoints
+     * @throws NullPointerException if parameter is {@code null}.
+     */
+    public final void setMapperEndpoints(final HashMap<String, MapperEndpoint> mapperEndpoints) {
+        this.mapperEndpoints = Objects.requireNonNull(mapperEndpoints);
```
sophokles73, line 244: you should better copy all elements, i.e.
```
this.mapperEndpoints.clear();
this.mapperEndpoints.addAll(mapperEndpoints);
```
so that future modifications of the passed in map have no impact on this config class

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Device.java
```diff
@@ -54,6 +54,9 @@
     @JsonFormat(with = JsonFormat.Feature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
     private List<String> memberOf = new LinkedList<>();
 
+    @JsonInclude(value = Include.NON_EMPTY)
+    private String mapper = "";
```
sophokles73, line 58: does this really need to be configured at the device instead of the tenant level?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Device.java
```diff
@@ -54,6 +54,9 @@
     @JsonFormat(with = JsonFormat.Feature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
     private List<String> memberOf = new LinkedList<>();
 
+    @JsonInclude(value = Include.NON_EMPTY)
```
sophokles73, line 57: IMHO this should be `Include.NON_NULL` and the field should be `null` by default.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Device.java
```diff
@@ -235,4 +239,24 @@ public Device setMemberOf(final List<String> memberOf) {
         return this;
     }
 
+    /**
+     * Sets the mapper property for this device.
+     *
+     * @param mapper The mapper property to set.
+     * @return        a reference to this for fluent use.
+     */
+    public Device setmapper(final String mapper) {
```
sophokles73, line 248: setMapper
final?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Device.java
```diff
@@ -235,4 +239,24 @@ public Device setMemberOf(final List<String> memberOf) {
         return this;
     }
 
+    /**
+     * Sets the mapper property for this device.
+     *
+     * @param mapper The mapper property to set.
+     * @return        a reference to this for fluent use.
```
sophokles73, line 246: what is the purpose of this property?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Device.java
```diff
@@ -235,4 +239,24 @@ public Device setMemberOf(final List<String> memberOf) {
...
+        this.mapper = mapper;
+        return this;
+    }
+
+    /**
+     * Gets the mapper for this devices.
+     *
+     * @return mapper for this device.
+     */
+    public String getMapper() {
```
sophokles73, line 258: final?

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/Config.java
BobClaerhout: must have been from the rebase

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
BobClaerhout: udpated

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/IMessageMapping.java
```diff
@@ -0,0 +1,42 @@
...
+import io.vertx.core.json.JsonObject;
+import io.vertx.mqtt.messages.MqttPublishMessage;
+
+/**
+ * A message mapping interface. This component requests mapping from another server is configured properly. The
+ * headers are overwritten with the result of the mapper (which includes the resourceId).
+ * E.g.: when the deviceId is in the payload of the message, the deviceId can be deducted in the custom mapper and
+ * the payload can be changed accordingly to the payload originally received by the gateway.
+ */
+public interface IMessageMapping {
```
sophokles73, line 29: We do not use this naming pattern to indicate an interface. This is simply the *MessageMapping* interface and your implementation might be the *HttpBasedMessageMapping* or something indicating its particular strategy for mapping the messages.

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/IMessageMapping.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.eclipse.hono.adapter.mqtt.MqttContext;
+import org.eclipse.hono.util.ResourceIdentifier;
+
+import io.vertx.core.Future;
+import io.vertx.core.json.JsonObject;
+import io.vertx.mqtt.messages.MqttPublishMessage;
+
+/**
+ * A message mapping interface. This component requests mapping from another server is configured properly. The
```
sophokles73, line 24: The first sentence is supposed to tell the reader a little about the nature of this class/interface. IMHO you should simply remove the first sentence.
Here is a nice [introduction to how to write good JavaDoc comments](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html) ...

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/MessageMapping.java
sophokles73: try `LoggerFactory.getLogger(MessageMapping.class)`

adapters/mqtt-vertx/src/main/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapter.java
```diff
@@ -14,48 +14,96 @@
...
     }
 
+    /**
+     * Sets the messageMapping for this adapter.
+     *
+     * @param IMessageMapping the messageMapping
+     * @throws NullPointerException if messageMapping is {@code null}.
+     */
+    @Autowired
+    public void setMessageMapping(final IMessageMapping IMessageMapping) {
```
sophokles73, line 63: parameter names always start with a lower case letter ..

adapters/mqtt-vertx/src/test/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapterTest.java
```diff
@@ -257,37 +306,58 @@ public void testMapTopicSupportsShortAndLongTopicNames(final VertxTestContext ct
...
+        headers.add(MessageHelper.APP_PROPERTY_DEVICE_ID, newDeviceId);
+        when(httpResponse.headers()).thenReturn(headers);
+        final Buffer changedBuffer = Buffer.buffer("changed");
+        when(httpResponse.bodyAsBuffer()).thenReturn(changedBuffer);
+
+        final MqttPublishMessage message = newMessage(MqttQoS.AT_LEAST_ONCE, EventConstants.EVENT_ENDPOINT);
+        final MqttContext context = newContext(message, null);
+        final MapperEndpoint mapperEndpoint = MapperEndpoint.from("host", 1234, "/uri", false);
+        messageMapping.mapMessageRequest(context, targetAddress, message, mapperEndpoint, headers).onComplete(ctx.succeeding(mappedMessage -> {
+        }));
```
sophokles73, line 339: This actually seems to be a test for the `MessageMapping` class but not for the protocol adapter, doesn't it?
Also, it doesn't seem to test anything as there are no assertions, or am I mistaken?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Device.java
```diff
@@ -54,6 +54,9 @@
     @JsonFormat(with = JsonFormat.Feature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
     private List<String> memberOf = new LinkedList<>();
 
+    @JsonInclude(value = Include.NON_NULL)
+    private String mapper = null;
```
sophokles73, line 58: just for reference, in Java you do not need to explicitly initialize with `null` as it is the default. It doesn't harm to make it explicit, though, of course ...

adapters/mqtt-vertx/src/test/java/org/eclipse/hono/adapter/mqtt/impl/VertxBasedMqttProtocolAdapterTest.java
```diff
@@ -257,37 +306,60 @@ public void testMapTopicSupportsShortAndLongTopicNames(final VertxTestContext ct
...
     }
 
-    private void givenAnAdapter() {
+    /**
+     * Verifies that the adapter calls the mapper if configured. Also the updated payload and deviceId is checked.
+     *
+     * @param ctx The helper to use for running tests on vert.x.
+     */
+    @Test
+    public void testMapperShouldUpdatePayloadAndDeviceId(final VertxTestContext ctx) {
```
sophokles73, line 315: this doesn't seem to be testing any functionality of the MQTT adapter, i.e. this should be a test case of `HttpBasedMessageMappingTest`, right?

site/documentation/content/admin-guide/mqtt-adapter-config.md
```diff
@@ -41,6 +41,9 @@ The following table provides an overview of the configuration variables and corr
...
+| `HONO_MQTT_MAPPERENDPOINTS_<mapperName>_HOST`<br>`--hono.mqtt.mapperEndpoints.<mapperName>.host` | no | - | The host to which the adapter should connect to call the custom mapper. Replace `<mapperName>` with the name of the mapper to configure. |
+| `HONO_MQTT_MAPPERENDPOINTS_<mapperName>_PORT`<br>`--hono.mqtt.mapperEndpoints.<mapperName>.port` | no | - | The port to which the adapter should connect to call the custom mapper. Replace `<mapperName>` with the name of the mapper to configure. |
+| `HONO_MQTT_MAPPERENDPOINTS_<mapperName>_URI`<br>`--hono.mqtt.mapperEndpoints.<mapperName>.uri` | no | - | The used uri to append to the host to call the mapping method on the mapping server. Replace `<mapperName>` with the name of the mapper to configure. |
```
sophokles73, line 46: can you move these to the correct position based on alphabetical order?

BobClaerhout:
Any other remarks?

PR #1879, opened 2020-04-06
[#1679] Update JavaDoc of the Device Management APIs for status codes

kaniyan:
The _JavaDoc_ of the _device management_ interface has been fixed now with the missing response codes to make it sync with that of the API definition in https://www.eclipse.org/hono/docs/api/management/.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementService.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementService.java
@@ -39,11 +39,10 @@ public interface DeviceManagementService {
      * @param span The active OpenTracing span for this operation. It is not to be closed in this method!
      *          An implementation should log (error) events on this span and it may set tags and use this span as the
      *          parent for any spans created in this method.
-     * @return A future indicating the outcome of the operation. The <em>status</em> will be
-     *            <ul>
-     *            <li><em>201 Created</em> if the device has been registered successfully.</li>
-     *            <li><em>409 Conflict</em> if a device with the given identifier already exists for the tenant.</li>
-     *            </ul>
+     * @return A future indicating the outcome of the operation.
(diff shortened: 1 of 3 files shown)


service-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementService.java
```diff
@@ -42,6 +42,9 @@
      * @return A future indicating the outcome of the operation. The <em>status</em> will be
      *            <ul>
      *            <li><em>201 Created</em> if the device has been registered successfully.</li>
+     *            <li><em>400 Bad Request</em> if the request to register the device is malformed.</li>
+     *            <li><em>401 Unauthorized</em> if the authentication credentials are required, but missing.</li>
+     *            <li><em>403 Forbidden</em> if the operation to register the device is not allowed.</li>
```
sophokles73, line 47: IMHO we shouldn't list the possible status codes here at all but instead simply refer to the Device Registry Management API docs ...
@ctron @dejanb WDYT?

kaniyan: :+1: FMPOV it is more effort to maintain the same information in 2 places and it could easily be not in sync at some point. Now I am also convinced that its better to simply refer to the Device Registry Management API docs and to remove the redundant information from the _JavaDoc_.

kaniyan: @sophokles73 The link to the API docs are documented in the _JavaDoc_ using `@see`. In this case I would simply remove the status code part from the _JavaDoc_. WDYT?

sophokles73: I would rathe radd a link explicitly to the `@return` description so that the correlation is more explicit

PR #1880, opened 2020-04-06
Add base work for the JDBC device registry implementation

ctron:
This PR contains as base module for the JDBC based device registry. By itself this module isn't really useful, but splits up the whole JDBC device registry in smaller PRs to be easier to review.

The reason for a base module, aside from more modularity, is that it is being used by:
* the device registry
* the device connection service
* additional independent tasks, like a tenant cleanup process

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -155,6 +155,11 @@
       </dependency>
       <dependency>
         <groupId>io.vertx</groupId>
+        <artifactId>vertx-jdbc-client</artifactId>
+        <version>${vertx.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.vertx</groupId>
         <artifactId>vertx-mongo-client</artifactId>
         <version>${vertx.version}</version>
       </dependency>
@@ -207,6 +212,16 @@
         <artifactId>snakeyaml</artifactId>
         <version>${snakeyaml.version}</version>
         <scope>runtime</scope>
+      </dependency>
+      <dependency>
+        <groupId>com.fasterxml.jackson.core</groupId>
+        <artifactId>jackson-databind</artifactId>
+        <version>${jackson.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>com.fasterxml.jackson.core</groupId>
+        <artifactId>jackson-core</artifactId>
+        <version>${jackson.version}</version>
       </dependency>
       <dependency>
         <groupId>org.apache.qpid</groupId>
@@ -568,6 +583,11 @@
       <dependency>
         <groupId>org.eclipse.hono</groupId>
         <artifactId>client-device-connection-infinispan</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.eclipse.hono</groupId>
+        <artifactId>service-base-jdbc</artifactId>
         <version>${project.version}</version>
       </dependency>
     </dependencies>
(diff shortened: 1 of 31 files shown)


bom/pom.xml
```diff
@@ -153,6 +153,11 @@
         <artifactId>vertx-micrometer-metrics</artifactId>
         <version>${vertx.version}</version>
       </dependency>
+      <dependency>
```
dejanb, line 156: We'll need to sort out all CQs for this PR

service-base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/config/JdbcDeviceProperties.java
```diff
@@ -0,0 +1,50 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.service.base.jdbc.config;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+
+@JsonInclude(value = Include.NON_NULL)
+public class JdbcDeviceProperties {
```
dejanb, line 20: We should add JavaDoc comments for these classes. And also enable checkstyle plugin for the module. Although I'd be in favor of relaxing some of the rules regarding the JavaDoc comments.

service-base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/device/DeviceReadResult.java
```diff
@@ -0,0 +1,46 @@
...
+
+package org.eclipse.hono.service.base.jdbc.store.device;
+
+import java.util.Optional;
+
+import org.eclipse.hono.service.management.device.Device;
+
+import com.google.common.base.MoreObjects;
+
+public class DeviceReadResult {
```
dejanb, line 22: Could we use `OperationResult` class here?

ctron:
> What about moving this module to the `services` instead of making it top level?

It is not a service by itself. So I followed the current pattern of the "client" modules:

~~~
hono/
  client/
  client-device-connection-infinispan/
  service-base/
  service-base-jdbc/
~~~

IIRC we discussed in the past, to split up the "service-base" module, so that we can independently evolve it. Maybe doing some restructuring here would be a good idea. In any case, I am open to changing the location of this module.

bom/pom.xml
ctron: No more CQs :D as long as the license is ok.

service-base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/device/DeviceReadResult.java
ctron: I would like to avoid that. First of all, that would introduce a dependency on the higher level stuff, to this lower level module. Also would the other fields (status, cache directive) have no meaning on this level.

dejanb:
> > What about moving this module to the `services` instead of making it top level?
> 
> It is not a service by itself. So I followed the current pattern of the "client" modules:
> 
> ```
> hono/
>   client/
>   client-device-connection-infinispan/
>   service-base/
>   service-base-jdbc/
> ```
> 
> IIRC we discussed in the past, to split up the "service-base" module, so that we can independently evolve it. Maybe doing some restructuring here would be a good idea. In any case, I am open to changing the location of this module.

There's a `services/device-registry-base` which is also not a service implementation, but have a similar function as this module.

The problem with client was that IIRC is that it itself doesn't have submodules like services and adapter where the infinispan stuff could live.

ctron:
> There's a `services/device-registry-base` which is also not a service implementation, but have a similar function as this module.

True, I will relocate the module.

ctron:
Ok, relocated the module, add checkstyle, and added all javadoc comments.

Issue #1865, opened 2020-03-31
Hono resource limits :  timeout for Prometheus queries  


bordeuax:
Playing around with _Hono resource limits_ I come to the conclusion that for medium loaded cluster(K8S cluster with Hono deployment) we can have network issue, also if Prometheus is heavy loaded,  PomQL queries can take unpredictable time for execution . Why not to have  _resource limits configuration - > timeout for Prometheus queries_  as a tune mechanism. Also not link this timeout only to _vert.x WebClient timeout_ but also with [Prometheus Instant queries](https://prometheus.io/docs/prometheus/latest/querying/api/#instant-queries)

sophokles73:
Sounds reasonable to have a configurable timeout for the query. Do you want to prepare a PR? Not sure about the Instant Query, though. @kaniyan WDYT?

kaniyan:
> Sounds reasonable to have a configurable timeout for the query. Do you want to prepare a PR? Not sure about the Instant Query, though. @kaniyan WDYT?

It sounds good :+1:.  Optimisations such as this, preventing concurrent query executions to retrieve the resource usage for a tenant and caching optimisations are thought through for the next iteration.

bordeuax:
Is need a PR on this topic or you plan to have a large round of discussion and collect all issues for next iteration ?

kaniyan:
> Is needed a PR on this topic or you plan to have a large round of discussion and collect all issues for next iteration ?

A PR is always welcome. If you are interested, go ahead.

sophokles73:
@bordeuax is #1866 related to your proposal to use an Instant Query for determining the number of currently connected devices?

bordeuax:
@sophokles73, when we executing a PomQL instant query via HTTP API, we have the possibility to specify as URL parameter a timeout, `timeout=<duration>` that will indicate how much time Prometheus should take to execute the query, for more info please see [Instant queries](https://prometheus.io/docs/prometheus/latest/querying/api/#expression-queries). 
I found useful, if we have a timeout on adapter side for PomQL also put this timeout as URL parameter when we executing a query.

sophokles73:
I agree, that makes sense. A PR for adding the (configurable) timeout parameter would be very welcome :-)

sophokles73:
closed via #1882

PR #1882, opened 2020-04-06
[#1865] Hono resource limits : add timeout for Prometheus queries

bordeuax:
add timeout for Prometheus queries

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
@@ -515,6 +515,11 @@ public final class PrometheusBasedResourceLimitChecks implements ResourceLimitCh
                 .expect(ResponsePredicate.SC_OK)
                 .as(BodyCodec.jsonObject());
 
+        if (config.getQueryTimeout() > 0) {
+            // enables timeout for Prometheus queries via HTTP API
+            request.timeout(config.getQueryTimeout());
+        }
+
         if (!Strings.isNullOrEmpty(config.getUsername()) && !Strings.isNullOrEmpty(config.getPassword())) {
             request.basicAuthentication(config.getUsername(), config.getPassword());
         }
--- a/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecksConfig.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecksConfig.java
@@ -32,9 +32,15 @@ public final class PrometheusBasedResourceLimitChecksConfig extends Authenticati
      */
     public static final long DEFAULT_CACHE_TIMEOUT = 600L;
 
+    /**
+     * The default timeout in milliseconds, after the client close the request to a remote server.
+     */
+    public static final long DEFAULT_QUERY_TIMEOUT = 500L;
+
     private int cacheMinSize = DEFAULT_CACHE_MIN_SIZE;
     private long cacheMaxSize = DEFAULT_CACHE_MAX_SIZE;
     private long cacheTimeout = DEFAULT_CACHE_TIMEOUT;
+    private long queryTimeout = DEFAULT_QUERY_TIMEOUT;
 
     /**
      * Creates new properties with default values.
@@ -136,6 +142,30 @@ public final class PrometheusBasedResourceLimitChecksConfig extends Authenticati
             throw new IllegalArgumentException("default cache timeout must be greater than zero");
         }
         this.cacheTimeout = timeout;
+    }
+
+    /**
+     * Gets the period of time after which a request to a Prometheus server are closed.
+     * <p>
+     * The default value of this property is {@link #DEFAULT_QUERY_TIMEOUT}.
+     *
+     * @return The timeout for the request to a remote server in milliseconds, zero or negative value is for disabled timeout.
+     */
+    public long getQueryTimeout() {
+        return queryTimeout;
+    }
+
+    /**
+     * Sets the period of time after which a request to a Prometheus server are closed.
+     * <p>
+     * Setting zero or a negative {@code timeout} disables the timeout.
+     * <p>
+     * The default value of this property is {@link #DEFAULT_QUERY_TIMEOUT}.
+     *
+     * @param timeout The timeout in milliseconds.
+     */
+    public void setQueryTimeout(final long timeout) {
+        this.queryTimeout = timeout;
     }
 
 }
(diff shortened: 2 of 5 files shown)


service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -488,7 +488,9 @@ private void checkMessageLimit(
 
         final Promise<Long> result = Promise.promise();
         LOG.trace("running Prometheus query [URL: {}, query: {}]", url, query);
-        newQueryRequest(query).send(sendAttempt -> {
+        newQueryRequest(query)
+                .timeout(config.getQueryTimeout())
```
sophokles73, line 492: IMHO this should also be done in the *newQueryRequest* method, shouldn't it? It should also only be set if the config value is > 0, right?

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -515,6 +517,10 @@ private void checkMessageLimit(
                 .expect(ResponsePredicate.SC_OK)
                 .as(BodyCodec.jsonObject());
 
+        if (config.getQueryTimeout() > 0) {
+           request.addQueryParam("timeout", String.format("%dms", config.getQueryTimeout()));
```
sophokles73, line 521: My understanding is that the query will run on Prometheus for hte given amount of time. However, since the query will be started after the request has been sent, I assume that we can run into situations where the WebClient stops waiting for the response while the query is still running. I therefore think that the timeout set on the WebClient should serve as a backstop only, i.e. it should wait for some additional time in order to allow for the query to finish in any case. WDYT?

sophokles73:
@bordeuax can you also please document this new feature in the release notes?

sophokles73:
@bordeuax please add a *Signed-off-by* footer to your commit message

sophokles73:
@bordeuax are you still working on this PR?

bordeuax:
Hi @sophokles73 , yes i am on, in the next hours i will make an update.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
bordeuax: Vertx web-client have already the logic for timeout when is zero or negative the timeout is disabled. But anyway,  I followed your suggestion and   I putted the `timeout`  in the `newQueryRequest `.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
bordeuax: You are totally right, such edge case will definitely happen. Here I see several options : 
-  add _some additional time_ (for example can be 20 or 50 percentage more than _web client_ timeout ) to `timeout` query parameter ;  In my understating this will reduce probability of such edge case but still this will happen, because  this _additional time_  will certainly depend on network and maybe some Prometheus configuration.
- do not set `timeout` query parameter; Prometheus already have a global query timeout
- have additional configuration section like query `HONO_RESOURCELIMITS_PROMETHEUSBASED_QUERYPARAMETERS_TIMEOUT`, that will give for user directly access and configuration for this value.
- keep `timeout` query parameter with the value as for web client and put additional query parameter `time `  that indicate when the execution have been started (at least this is my understating regarding `time `parameter );  for more info see https://gitlab.cncf.ci/prometheus/prometheus/blob/3bce14db85998b48dde4066d9490b4f3eb4223e1/docs/querying/api.md
@sophokles73, WDYT?

sophokles73: IMHO we should rely on Prometheus' global query timeout and use the vert.x WebClient's *timeout* property as a backstop. The backstop should be configurable using e.g. the `HONO_RESOURCELIMITS_PROMETHEUSBASED_QUERYTIMEOUT` variable.

bordeuax: OK. Fixed , followed your suggestions.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecksConfig.java
```diff
@@ -138,4 +144,28 @@ public void setCacheTimeout(final long timeout) {
         this.cacheTimeout = timeout;
     }
 
+    /**
+     * Gets the period of time after which the request to a Prometheus server are closed.
+     * <p>
```
Alfusainey, line 149: how about using **a request** instead of **the request** since you are not referring to a particular request? e.g

> Gets the period of time after which a request to a Prometheus server is closed

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecksConfig.java
```diff
@@ -138,4 +144,28 @@ public void setCacheTimeout(final long timeout) {
...
+     * The default value of this property is {@link #DEFAULT_QUERY_TIMEOUT}.
+     *
+     * @return The timeout for the request to a remote server in milliseconds, zero or negative value is for disabled timeout.
+     */
+    public long getQueryTimeout() {
+        return queryTimeout;
+    }
+
+    /**
+     * Sets the period of time after which the request to a Prometheus server are closed.
```
Alfusainey, line 159: same as above

service-base/src/test/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecksTest.java
```diff
@@ -519,6 +545,35 @@ public void testConnectionDurationLimitExceeded(final VertxTestContext ctx) {
                 }));
     }
 
+    /**
+     * Verifies that the message limit check returns {@code false} if a timeout is occurred.
```
Alfusainey, line 549: **has** occured

site/documentation/content/admin-guide/common-config.md
```diff
@@ -136,6 +136,7 @@ The following table provides an overview of the configuration variables and corr
 | `HONO_RESOURCELIMITS_PROMETHEUSBASED_CACHE_MIN_SIZE`<br>`--hono.resourceLimits.prometheusBased.cacheMinSize` | no | `20` | The minimum size of the cache to store the metrics data retrieved from the Prometheus server. The cache is used for storing the current amount of data exchanged with devices of tenants. |
 | `HONO_RESOURCELIMITS_PROMETHEUSBASED_CACHE_MAX_SIZE`<br>`--hono.resourceLimits.prometheusBased.cacheMaxSize` | no | `1000` | The maximum size of the cache to store the metrics data retrieved from the Prometheus server. |
 | `HONO_RESOURCELIMITS_PROMETHEUSBASED_CACHE_TIMEOUT`<br>`--hono.resourceLimits.prometheusBased.cacheTimeout` | no | `600` | The number of seconds after which the cached metrics data should be considered invalid. |
+| `HONO_RESOURCELIMITS_PROMETHEUSBASED_QUERY_TIMEOUT`<br>`--hono.resourceLimits.prometheusBased.queryTimeout` | no | `500` | The number of milliseconds after which the request to a Prometheus server are closed. Setting zero or a negative value disables the timeout.|
```
Alfusainey, line 139: **a**  request to  a Prometheus server **is** closed

PR #1883, opened 2020-04-07
Provide an embedded cache version for the device connection service

ctron:
This refactors the existing implementation of the Infinispan based device connection service, to use the BasicCache instead of the RemoteCache, so that most of the functionality can be re-used between the remote and embedded cache versions.

It then adds an embedded cache version, which by default runs as simple, local cache and which can be fully configured to form an embedded cluster.


client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/CacheBasedDeviceConnectionInfo.java
```diff
@@ -167,39 +164,26 @@ public HotrodBasedDeviceConnectionInfo(final RemoteCache<String, String> cache,
         Objects.requireNonNull(adapterInstanceId);
 
         final String key = getAdapterInstanceEntryKey(tenantId, deviceId);
-        return cache.getWithVersion(key)
-               .recover(t -> {
+
+        return cache
+                .remove(key, adapterInstanceId)
+                .recover(t -> {
                     LOG.debug("failed to get cache entry when trying to remove command handling adapter instance [tenant: {}, device-id: {}, adapter-instance: {}]",
```
calohmn, line 171: The content of the log message doesn't fit here any more.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/Cache.java
```diff
@@ -67,14 +57,14 @@
...
-     * Gets a value from the cache.
+     * Remove a key/value mapping from the cache.
      *
      * @param key The key.
-     * @return A succeeded future containing the value or {@code null} if the
-     *         cache didn't contain the key yet.
-     *         A failed future if the value could not be read from the cache.
+     * @param value The value.
+     * @return {@code true} if the key was mapped to the value, {@code false}
+     *         otherwise.
```
calohmn, line 65: I would prefix the `@return` sentence with "A succeeded future containing".
I think it would be good to also mention the possibility that a failed future is returned here.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/EmbeddedCache.java
```diff
@@ -0,0 +1,113 @@
...
+                    final var status = cacheManager.getStatus();
+                    if (status != ComponentStatus.RUNNING) {
+                        LOG.debug("trying to start cache manager, current state: {}", status);
+                        cacheManager.start();
+                        LOG.info("started cache manager, now connecting to remote cache");
+                    }
+                    LOG.debug("trying to get cache");
+                    setCache(cacheManager.getCache(cacheName));
+                    if (getCache() == null) {
+                        r.fail(new IllegalStateException("cache [" + cacheName + "] is not configured exist"));
```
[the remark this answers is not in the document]
ctron: Fixed.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/CacheBasedDeviceConnectionInfo.java
ctron: Fixed.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/Cache.java
ctron: Changed.

client-device-connection-infinispan/pom.xml
```diff
@@ -22,6 +22,10 @@
       <groupId>org.infinispan</groupId>
       <artifactId>infinispan-client-hotrod</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.infinispan</groupId>
+      <artifactId>infinispan-core</artifactId>
+    </dependency>
```
calohmn, line 28: For actually using the embedded cache mode in a kubernetes cluster, we would also need the `jgroups-kubernetes` dependency, wouldn't we? Or do you want to add that later, along with maybe an example config?

ctron: I would do that outside of the scope of this PR.

PR #1884, opened 2020-04-08
Manage life cycle of device registry beans from endpoints

sophokles73:
The beans implementing the example device registry have been deployed as
Verticles in the past. This has the advantage of having their lifecycle
being managed by the vert.x container (i.e. invoking their start/stop
methods).

However, it also means that we have no control over the event loop
context (thread) they will run on. For the example registry this is not
a big issue because the beans will be accessed from multiple threads
anyway (from the AMQP and the HTTP server verticles) and we already took
care of allowing for this by using ConcurrentMaps.

For registry implementations that keep the data in a back end store like
a DB, this approach is not desirable, because it would require us to
give up the "single thread only" programming model of vert.x.

Instead, the beans implementing the data access layer in the registry
now implement the Lifecycle interface and have their start/stop methods
being invoked as part of the AMQP/HTTP server verticle's life cycle.

For registries using an external persistent store, this requires
implementing the different APIs in separate beans which can then be
injected into just the AMQP server endpoints or the HTTP server
endpoints, thus making sure that they are always only accessed from a
single thread, even if the server verticles get deployed multiple times.


ctron:
I was wondering if the "start/stop" methods couldn't be pulled in some kind of super-interface. It looks rather repetitive, and as those methods got a little bit more complex now, I think it might be worth spending a thought on moving them to a single location.

dejanb:
> I was wondering if the "start/stop" methods couldn't be pulled in some kind of super-interface. It looks rather repetitive, and as those methods got a little bit more complex now, I think it might be worth spending a thought on moving them to a single location.

Yeah, thinking up the same ... Introducing `AbstractLifecycle` might be a good idea.

sophokles73:
@ctron @dejanb are you talking about the start/stop methods in the endpoints or in the service implementation classes?

dejanb:
> @ctron @dejanb are you talking about the start/stop methods in the endpoints or in the service implementation classes?

I was thinking about endpoint code that starts/stops services. That could maybe go to the `AbstractServiceEndpoint` class?

sophokles73:
> I was thinking about endpoint code that starts/stops services. That could maybe go to the AbstractServiceEndpoint class?

I guess then we would also need to pull up the getter/setter for the service to delegate to ...

sophokles73:
@dejanb @ctron I have pulled up the (common) methods within the *\*Endpoint* class hierarchy. Turns out that we don't need the *Autowiring\*Endpoint* classes in the *device-registry-base* module anymore at all :-)

PR #1886, opened 2020-04-08
[#1679] Add implementation to resolve group members in MongoDB based device registry

kaniyan:
In this PR, an implementation of the `AbstractRegistrationService.resolveGroupMembers(...)` has been added to the _MongoDB_ based device registry.

### DIFF
--- a/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
+++ b/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
@@ -13,8 +13,10 @@
 package org.eclipse.hono.deviceregistry.mongodb.service;
 
 import java.net.HttpURLConnection;
+import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import org.eclipse.hono.client.ClientErrorException;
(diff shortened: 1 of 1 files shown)


services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -314,6 +319,35 @@ private boolean isDuplicateKeyError(final Throwable throwable) {
...
+                resolveGroupMembersPromise);
+
+        return resolveGroupMembersPromise.future()
+                .map(deviceIdsList -> {
+                    final JsonArray deviceIds = Optional.ofNullable(deviceIdsList)
+                            .map(ok -> deviceIdsList.stream()
+                                    .map(json -> json.getString(RegistrationConstants.FIELD_PAYLOAD_DEVICE_ID))
+                                    .collect(Collectors.collectingAndThen(Collectors.toList(), JsonArray::new)))
+                            .orElse(new JsonArray());
+                    span.log(String.format("successfully resolved group members for the tenant [%s]", tenantId));
```
sophokles73, line 346: The span should already be tagged with the tenant ID.

kaniyan: Since the tenant ID is already being tagged, now I have removed logging the tenant ID and pushed the changes.

PR #1888, opened 2020-04-08
Add simple receiver statistic to cli client.

boaks:
No big thing, nor brilliant, but from time to time it helps to see changes in the performance easier.
 
### DIFF
--- a/cli/src/main/java/org/eclipse/hono/cli/app/Receiver.java
+++ b/cli/src/main/java/org/eclipse/hono/cli/app/Receiver.java
@@ -14,6 +14,8 @@ package org.eclipse.hono.cli.app;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
+import java.util.function.BiConsumer;
 
 import javax.annotation.PostConstruct;
 
@@ -53,6 +55,21 @@ public class Receiver extends AbstractApplicationClient {
     protected String messageType;
 
     /**
+     * Bi consumer to handle messages based on endpoint.
+     */
+    private BiConsumer<String, Message> messageHandler = (endpoint, msg) -> handleMessage(endpoint, msg);
+
+    /**
+     * Set message handler for processing adaption.
+     * 
+     * @param messageHandler message handler.
+     * @throws NullPointerException if message handlerr is {@code null}.
+     */
+    void setMessageHandler(final BiConsumer<String, Message> messageHandler) {
+        this.messageHandler = Objects.requireNonNull(messageHandler);
+    }
+
+    /**
      * Starts this component.
      * <p>
      * 
@@ -82,12 +99,16 @@ public class Receiver extends AbstractApplicationClient {
         final List<Future> consumerFutures = new ArrayList<>();
         if (messageType.equals(TYPE_EVENT) || messageType.equals(TYPE_ALL)) {
             consumerFutures.add(
-                    clientFactory.createEventConsumer(tenantId, msg -> handleMessage(TYPE_EVENT, msg), closeHandler));
+                    clientFactory.createEventConsumer(tenantId, msg -> {
+                        messageHandler.accept(TYPE_EVENT, msg);
+                    }, closeHandler));
         }
 
         if (messageType.equals(TYPE_TELEMETRY) || messageType.equals(TYPE_ALL)) {
             consumerFutures.add(
-                    clientFactory.createTelemetryConsumer(tenantId, msg -> handleMessage(TYPE_TELEMETRY, msg), closeHandler));
+                    clientFactory.createTelemetryConsumer(tenantId, msg -> {
+                        messageHandler.accept(TYPE_TELEMETRY, msg);
+                    }, closeHandler));
         }
 
         if (consumerFutures.isEmpty()) {
@@ -99,6 +120,14 @@ public class Receiver extends AbstractApplicationClient {
         return CompositeFuture.all(consumerFutures);
     }
 
+    /**
+     * Handle received message.
+     * 
+     * Write log messages to stdout.
+     * 
+     * @param endpoint receiving endpoint, "telemetry" or "event".
+     * @param msg received message
+     */
     private void handleMessage(final String endpoint, final Message msg) {
         final String deviceId = MessageHelper.getDeviceId(msg);
 
--- a/cli/src/main/java/org/eclipse/hono/cli/app/ReceiverStatistics.java
+++ b/cli/src/main/java/org/eclipse/hono/cli/app/ReceiverStatistics.java
@@ -1,0 +1,280 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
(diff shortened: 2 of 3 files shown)


cli/src/main/java/org/eclipse/hono/cli/app/ReceiverStatistics.java
```diff
@@ -0,0 +1,218 @@
...
+import io.vertx.core.buffer.Buffer;
+
+/**
+ * A command line client for statistic of received messages from via Hono's north bound Telemetry and/or Event API and
+ * <p>
+ * Messages are output to stdout.
+ * <p>
+ */
+@Component
+@Profile("statistic")
```
calohmn, line 39: Concerning the spring profile usage:
I would have expected to use `--spring.profiles.active=receiver,statistic` here, having the `statistic` profile as kind of a modifier. Having to run `--spring.profiles.active=statistic` (without `receiver`) feels unintuitive. And having to define extra properties for the `statistic` profile (same way as for the `receiver` profile) in the application.yml also doesn't look like a good solution.

How about having the `ReceiverStatistics` class not extend `Receiver`. Instead, the `Receiver` class could offer a `addMessageHandler` method, with all the added handlers getting invoked in the `handleMessage` method.
Then `ReceiverStatistics` could just be a bean that adds a handler via that `addMessageHandler` method, and the counting logic could be in that handler.

cli/src/main/java/org/eclipse/hono/cli/app/ReceiverStatistics.java
```diff
@@ -0,0 +1,218 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2019 Contributors to the Eclipse Foundation
```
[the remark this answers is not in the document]
boaks: :-) right, I kept it private quite a while :-)

cli/src/main/java/org/eclipse/hono/cli/app/ReceiverStatistics.java
```diff
@@ -0,0 +1,218 @@
...
+    private Statistic current;
+
+    /**
+     * Starts this component.
+     * <p>
+     * 
+     * @return A future indicating the outcome of the startup process.
+     */
+    @PostConstruct
+    Future<CompositeFuture> start() {
```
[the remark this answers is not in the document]
boaks: Ok

cli/src/main/java/org/eclipse/hono/cli/app/ReceiverStatistics.java
boaks: Though I'm not to common with spring (or what ever technique is used for that), that seems to be out of my possibilities ;-).

boaks: Just as additional idea:
Create a common abstract base class, derive the receiver with logging and the statistic from that. Would that be also OK?

boaks:
Please, re-check the redesign :-).

cli/src/main/java/org/eclipse/hono/cli/app/ReceiverStatistics.java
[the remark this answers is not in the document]
calohmn: But not for 4 years, is it? ;)

cli/src/main/java/org/eclipse/hono/cli/app/ReceiverStatistics.java
```diff
@@ -0,0 +1,271 @@
...
+
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Future;
+import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
+
+/**
+ * A command line client for statistic of received messages from via Hono's north bound Telemetry and/or Event API and
+ * <p>
+ * Messages are output to stdout.
```
calohmn, line 39: Can you adapt the javadoc here (class is not a receiver any more)?

cli/src/main/java/org/eclipse/hono/cli/app/Receiver.java
```diff
@@ -99,6 +120,14 @@ private CompositeFuture createConsumer(final HonoConnection connection) {
...
     }
 
+    /**
+     * Handle received messge.
+     * 
+     * Write log messages to stdout.
+     * 
+     * @param endpoint receiving endpont, "telemetry" or "event".
+     * @param msg recevied message
+     */
```
calohmn, line 130: Typos (messge, endpont, recevied). Or javadoc can be removed again (method is again private now).

cli/src/main/java/org/eclipse/hono/cli/app/Receiver.java
```diff
@@ -52,6 +54,21 @@
     @Value(value = "${message.type}")
     protected String messageType;
 
+    /**
+     * Bi consumer to handle messages based one endpoint.
```
calohmn, line 58: based on

cli/src/main/java/org/eclipse/hono/cli/app/ReceiverStatistics.java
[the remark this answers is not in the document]
boaks: I will set it to 2020 :-)

boaks:
One moment, please. The statistic doesn't reset proper, if there is no traffic.

boaks:
OK, now it clears/resets the statistics.

cli/src/main/java/org/eclipse/hono/cli/app/ReceiverStatistics.java
```diff
@@ -0,0 +1,268 @@
...
+            if (current == null || current.finished(now)) {
+                if (current == null) {
+                    LOG_STATISTIC.info("statistic: total {}", total);
+                } else {
+                    LOG_STATISTIC.info("statistic: total {}, last {}", total, current);
+                }
+                synchronized (this) {
+                    if (this.total == total && this.current == current) {
+                        if (current == null) {
+                            this.total = null;
```
calohmn, line 158: This means that the `total` statistic is reset after having received no messages in a period of 2x interval, right?
I think this might be confusing when the user just looks at the last log line after a long test run, not seeing that the total number might not be the actual grand total because it got reset a few times.

boaks: That may be more a question of your usage. I start usually that client and then I focus on my own test via the protocol adapters. Sometimes I adapt my test and want to see the changes. Without the reset I sometimes (sometimes very frequently :-) ), I forgot the restart the statistic client and as result, I had to repeat the test again with restart statistic client. though the test are more a couple of minutes, I found the auto-reset the better idea.

So, should I add a config flag for autoReset in order to be able to select the behaviour?

calohmn: Yes, I think that would be good.

boaks: I added the config and disabled it by default.

PR #1890, opened 2020-04-08
Update developers organization details.

kaniyan:
@sophokles73 There is also one more place in `NOTICE.md` to be updated. Can `Copyright 2016-2019 Bosch Software Innovations GmbH` be simply replaced with `Copyright 2016-2020 Bosch.IO GmbH` in that file?


sophokles73:
> Can Copyright 2016-2019 Bosch Software Innovations GmbH be simply replaced with Copyright 2016-2020 Bosch.IO GmbH in that file?

No, but we can add a new line
> 2020 Bosch.IO GmbH

PR #1893, opened 2020-04-09
Do not use Spring Boot component scan for creating beans

sophokles73:
The file based registry, the Mongo DB registry and the Device Connection
service now create service bean instances explicitly in a central place
which makes it easier to understand what's actually
happening.


services/device-connection/src/main/java/org/eclipse/hono/deviceconnection/infinispan/Application.java
```diff
@@ -36,10 +36,9 @@
  * The application implements Hono's <a href="https://www.eclipse.org/hono/docs/api/device-connection/">
  * Device Connection API</a>.
  */
-@ComponentScan("org.eclipse.hono.deviceconnection.infinispan")
 @ComponentScan("org.eclipse.hono.service.auth")
 @ComponentScan("org.eclipse.hono.service.metric")
-@Configuration
+@Import(value = { ApplicationConfig.class, EmbeddedCacheConfig.class, RemoteCacheConfig.class })
```
ctron, line 41: You can leave out the `value = ` part, as this is the default name and can be omitted.

PR #1900, opened 2020-04-14
[#1679] Avoid conversion from Device to JsonObject in MongoDB based assertRegistration

kaniyan:
In the existing  `processAssertRegistration(...)` implementation of the device registry, the retrieved _device_ is first mapped to a `JsonObject`  before creating a `RegistrationResult`. In this PR, this device mapping to `JsonObject` has been avoided considering the potential performance setback. @sophokles73 Would you mind taking a look?

### DIFF
--- a/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
+++ b/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
@@ -177,9 +177,11 @@ public final class MongoDbBasedRegistrationService extends AbstractRegistrationS
         Objects.requireNonNull(deviceKey);
         Objects.requireNonNull(span);
 
-        return processReadDevice(deviceKey.getTenantId(), deviceKey.getDeviceId())
-                .compose(result -> Future.succeededFuture(RegistrationResult.from(result.getStatus(),
-                        convertDevice(deviceKey.getDeviceId(), result.getPayload()), result.getCacheDirective().orElse(null))));
+        return findDeviceDocument(deviceKey.getTenantId(), deviceKey.getDeviceId())
+                .map(result -> Optional.ofNullable(result)
+                        .map(ok -> getRegistrationResult(deviceKey.getDeviceId(),
+                                result.getJsonObject(MongoDbDeviceRegistryUtils.FIELD_DEVICE)))
+                        .orElse(RegistrationResult.from(HttpURLConnection.HTTP_NOT_FOUND)));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -196,32 +198,32 @@ public void setConfig(final MongoDbBasedRegistrationConfigProperties config) {
...
                 .document();
         final Promise<JsonObject> readDevicePromise = Promise.promise();
         mongoClient.findOne(config.getCollectionName(), findDeviceQuery, null, readDevicePromise);
-        return readDevicePromise.future()
-                .compose(result -> Optional.ofNullable(result)
-                        .map(ok -> result.mapTo(DeviceDto.class))
-                        .map(Future::succeededFuture)
-                        .orElseGet(() -> Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_NOT_FOUND,
-                                String.format("Device [%s] not found.", deviceId)))));
+        return readDevicePromise.future();
```
sophokles73, line 217: why not fail the promise with a `ClientErrorException(404)` here instead of checking for a `null` result everywhere else?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDeviceRegistryUtils.java
```diff
@@ -35,6 +35,14 @@
...
 public final class MongoDbDeviceRegistryUtils {
 
+    /**
+     * The name of the JSON property containing the device identifier.
+     */
+    public static final String FIELD_DEVICE = "device";
+    /**
+     * The name of the JSON property containing the data.
+     */
+    public static final String FIELD_DATA = "data";
```
sophokles73, line 45: IMHO we don't need this as it seems to be a re-declaration of `RegistrationConstants.FIELD_DATA`, isn't it?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -196,32 +198,32 @@ public void setConfig(final MongoDbBasedRegistrationConfigProperties config) {
...
+    }
+
+    private RegistrationResult getRegistrationResult(final String deviceId, final JsonObject devicePayload) {
+        return RegistrationResult.from(HttpURLConnection.HTTP_OK,
+                Optional.ofNullable(devicePayload)
+                        .map(ok -> new JsonObject()
+                                .put(RegistryManagementConstants.FIELD_PAYLOAD_DEVICE_ID, deviceId)
+                                .put(MongoDbDeviceRegistryUtils.FIELD_DATA, devicePayload))
```
sophokles73, line 225: IMHO this should be `RegistrationConstants.FIELD_DATA`

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDeviceRegistryUtils.java
```diff
@@ -35,6 +35,14 @@
  */
 public final class MongoDbDeviceRegistryUtils {
 
+    /**
+     * The name of the JSON property containing the device identifier.
```
sophokles73, line 39: is this field really used for the device identifier? Or is it used for the device data itself?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -196,32 +198,32 @@ public void setConfig(final MongoDbBasedRegistrationConfigProperties config) {
...
-                        .orElseGet(() -> Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_NOT_FOUND,
-                                String.format("Device [%s] not found.", deviceId)))));
+        return readDevicePromise.future();
+    }
+
+    private RegistrationResult getRegistrationResult(final String deviceId, final JsonObject devicePayload) {
+        return RegistrationResult.from(HttpURLConnection.HTTP_OK,
+                Optional.ofNullable(devicePayload)
+                        .map(ok -> new JsonObject()
+                                .put(RegistryManagementConstants.FIELD_PAYLOAD_DEVICE_ID, deviceId)
```
sophokles73, line 224: IMHO this should be `RegistrationConstants.FIELD_PAYLOAD_DEVICE_ID` as the result being prepared here is defined by the Device Registration API

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: Initially I implemented as in the comment (as below), then I changed it. In this case a `recover` block is needed and there it need to be checked for the _ClientErrorException_ and create a `RegistrationResult` based on the status code. I thought in the other approach, by checking for _null_ we could directly create a `RegistrationResult` with `HTTP_NOT_FOUND` and this extra check for _ClientErrorException_ could be avoided. That's why I chose the other one. We discuss here about the error case and supposedly it's not the often occurring case, I am fine with both the approaches. I will replace it with the below code.


```
return findDeviceDocument(deviceKey.getTenantId(), deviceKey.getDeviceId())
		.map(result -> getRegistrationResult(deviceKey.getDeviceId(),
				result.getJsonObject(MongoDbDeviceRegistryUtils.FIELD_DEVICE)))
		.recover(error -> {
			log.error(error.getMessage(), error);
			TracingHelper.logError(span, error.getMessage(), error);
			if (error instanceof ClientErrorException) {
				return Future.succeededFuture(
						RegistrationResult.from(((ClientErrorException) error).getErrorCode()));
			}
			return Future.succeededFuture(RegistrationResult.from(HttpURLConnection.HTTP_INTERNAL_ERROR));
		});
```

sophokles73: No, please keep your code. You seem to have a good reason for doing it the way you do :-)

kaniyan:
@sophokles73  All the suggested changes have been incorporated and I have pushed those changes. Would you mind taking a look again?

PR #1907, opened 2020-04-16
[#1679] Pull up utility methods from file based credentials service for reuse

kaniyan:


kaniyan:
The integration tests failed due to a bug in the file based credentials service. 
This has been fixed by the PR #1909. This to be merged after PR #1909.

dejanb:
I don't have anything against moving this to the utility class. However, I was expecting that in the future all service implementations will be based on the recently introduced abstract classes (https://github.com/eclipse/hono/blob/master/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsManagementService.java), which have these methods in them.
Would you consider doing it that way?

kaniyan:
> I don't have anything against moving this to the utility class. However, I was expecting that in the future all service implementations will be based on the recently introduced abstract classes (https://github.com/eclipse/hono/blob/master/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsManagementService.java), which have these methods in them.
> Would you consider doing it that way?

IMHO this abstract classes approach is under development and some building blocks are not yet there (e.g abstract classes for the _tenant service_ and _tenant management_, fitting _auto-provisioning_ etc). 
Since we are in need of a productive device registry, we would like to bring this MongoDB based registry asap. Also we have already implemented the _tenant_ and _device (management)_ services using the already existing `DeviceBackend` approach. Implementing credential services using these abstract classes, IMHO currently doesn't fit into this approach without any tweaks.

IMHO the abstract classes approach will be in its final shape once we have a reference implementation based on it (either the JDBC based implementation or having the file based registry written using this approach). Then we could still adapt the MongoDB based device registry to that approach.

dejanb:
Sounds good.

Issue #1858, opened 2020-03-26
Prevent stale command subscription entries in Device Connection Service


calohmn:
Scenario:
A device, that is normally only interacting via a gateway, has done a `ttd` HTTP request to the Hono HTTP protocol adapter in order to receive a command message.
Having received a command message, there is some error when invoking the `removeCommandHandlingAdapterInstance` method in the Device Connection Service, as part of the protocol adapter removing the command subscription. This leads to a stale `commandHandlingAdapterInstance` entry concerning the device in the Device Connection Service.

Now, for all subsequent command subscriptions only done by the gateway of the device, there is the issue that the stale `commandHandlingAdapterInstance` entry will effectively prevent commands from getting delivered to the gateway. This is due to the device-specific `commandHandlingAdapterInstance` entry getting precedence.

Only a subsequent created and removed command subscription for the specific device (not the gateway), where the `removeCommandHandlingAdapterInstance` method succeeds, will mitigate such a situation.

---
To prevent or automatically resolve such a situation, this looks like a straightforward solution:
- Implement a retry mechanism for the `removeCommandHandlingAdapterInstance` invocation for as long as it fails. 

That is easy to implement _but_ doesn't necessarily prevent the above problem if the adapter is restarted while the `commandHandlingAdapterInstance` entry hasn't been successfully removed yet.

Therefore it looks like we need to implement the following instead (or as well):

- Let `commandHandlingAdapterInstance` entries expire after a certain time and implement periodic refresh requests from the protocol adapter while a subscription is still active.

ctron:
I think the TTD defines the timeout of such a mapping. Why not hand over the responsibility to the device connection service, to remove this latest this point in time (now+ttd).

Infinispan does have a TTL mechanism, and I guess for every SQL based backend you could easily implement a cleanup job like this.

calohmn:
@ctron yes, letting Infinispan do the work of actually deleting the entries would be the idea, and the TTD would define the lifespan of the entries - if there is a TTD.

In the case of MQTT where there is no TTD, we could use a predefined expiry value for the device connection service entries, defined so that the automatic re-creation interval for entries of still valid command subscriptions is less than that.

calohmn:
PR #1916 has been created for using the ttd as the lifespan of the mapping entry in the device connection service.

For the cases where no ttd is given and a periodic refresh of the mapping entry shall be done, a new Device Connection API method is needed for the refresh operation, performing a _conditional update_. 
[...]

PR #1908, opened 2020-04-16
[#1858] Add lifespan to setCommandHandlingAdapterInstance

calohmn:
This is for #1858:
This makes it possible to restrict the lifespan of entries set via the `setCommandHandlingAdapterInstance` method of the Device Connection API.

Actual usage of the new parameter will be added in a subsequent PR.


site/documentation/content/api/device-connection/index.md
```diff
@@ -126,6 +126,7 @@ The following table provides an overview of the properties a client needs to set
 | :-------------------- | :-------: | :----------------------- | :-------- | :---------- |
 | *subject*             | yes       | *properties*             | *string*  | MUST be set to `set-cmd-handling-adapter-instance`. |
 | *adapter_instance_id* | yes       | *application-properties* | *string*  | The identifier of the protocol adapter instance that currently handles commands for the device or gateway identified by the *device_id* property. |
+| *lifespan*            | yes       | *application-properties* | *int*     | The lifetime of the mapping entry in seconds. After that period, the mapping entry shall be treated as non-existent by the *Device Registration API* methods. A negative value is interpreted as an unlimited lifespan.|
```
sophokles73, line 129: FMPOV this is an API breaking change. How about making it optional until 2.0.0?

calohmn: Yes, good point. I've changed it accordingly.

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/deviceconnection/MapBasedDeviceConnectionService.java
```diff
@@ -43,15 +47,15 @@
...
+public class MapBasedDeviceConnectionService implements DeviceConnectionService {
 
     private static final Logger log = LoggerFactory.getLogger(MapBasedDeviceConnectionService.class);
 
     // <tenantId, <deviceId, lastKnownGatewayJson>>
     private final Map<String, Map<String, JsonObject>> lastKnownGatewaysMap = new HashMap<>();
 
     // <tenantId, <deviceId, adapterInstanceIdJson>>
-    private final Map<String, Map<String, JsonObject>> commandHandlingAdapterInstancesMap = new HashMap<>();
+    private final Map<String, Map<String, ExpiringValue<JsonObject>>> commandHandlingAdapterInstancesMap = new HashMap<>();
```
sophokles73, line 58: it might be easier to use a Caffeine *Cache* instead which already supports limiting size and defining a time-based eviction strategy ...

calohmn: I've committed a change so that a Caffeine Cache is now used, letting it do the eviction.

I've haven't used the size limited feature though (`maximumSize` on cache builder) because that would change the behaviour. When the max size is (about to get) reached on a Caffeine Cache, some existing entry gets evicted. What we want here is to keep the existing entries and prevent addition of new entries if the max size has been reached.

site/documentation/content/api/device-connection/index.md
```diff
@@ -126,6 +126,7 @@ The following table provides an overview of the properties a client needs to set
 | :-------------------- | :-------: | :----------------------- | :-------- | :---------- |
 | *subject*             | yes       | *properties*             | *string*  | MUST be set to `set-cmd-handling-adapter-instance`. |
 | *adapter_instance_id* | yes       | *application-properties* | *string*  | The identifier of the protocol adapter instance that currently handles commands for the device or gateway identified by the *device_id* property. |
+| *lifespan*            | no        | *application-properties* | *int*     | The lifetime of the mapping entry in seconds. After that period, the mapping entry shall be treated as non-existent by the *Device Registration API* methods. A negative value, as well as an omitted property, is interpreted as an unlimited lifespan.|
```
sophokles73, line 129: what if the client provides the value but the implementation doesn't support it?

calohmn: I've adapted the description as follows:
````
The desired lifespan of the mapping entry in seconds. If the *Device Registration API* implementation supports this property, the mapping entry shall be treated as non-existent after that period has elapsed. A negative value, as well as an omitted property, is interpreted as an unlimited lifespan. A *Device Registration API* implementation may not support this property, meaning that a client can't rely on the given lifespan to be applied.
````

calohmn: @sophokles73 An afterthought on this:
Since we target this change for the next minor version, meaning we can introduce incompatible changes to _providers_ of the Device Connection API, I think we _can_ indeed _require_ implementations of the API to support the `lifespan` property. And I think we should make it a requirement then (otherwise we have to maintain extra client logic for the case it isn't supported by an implementation). WDYT?

sophokles73: How could a client know if an implementation supports it or not? The implementation would simply ignore the given value and keep the mapping until it evicts it, wouldn't it?

calohmn: My point is to make it _mandatory_ for an implementation to support the `lifespan` property, so that a client can be sure it is supported.

sophokles73: I understand that. My note was in reply to your remark
> otherwise we have to maintain extra client logic for the case it isn't supported by an implementation

We have no way of making the registry supporting the property because the adapter has no way of detecting this and thus cannot e.g. issue a warning or shut down or use additional logic to compensate for it. I agree, though, that we should document the fact that, if a registry does not support the additional lifespan property, then the adapter might fail to correctly route commands.

calohmn: And as a general note concerning the `Mandatory` column in the `Request Message Format` properties table: 
I understand that in such a way that it is or isn't mandatory for the _client_ to use the property in its requests. If not stated otherwise in the description, I would think that API _implementations_ _have_ to support the property regardless, right?

calohmn: > We have no way of making the registry supporting the property because the adapter has no way of detecting this and thus cannot e.g. issue a warning or shut down or use additional logic to compensate for it.

Well, we could use the existing logic of always explicitly removing the mapping entry (with the notable exception of edge cases like when the adapter is killed before being able to do that), noticing in the return value, whether the (expired) entry was actually removed before by the API implementation.
But I would rather omit this and make it part of the contract that the API implementation just has to support the `lifespan` property.

> I agree, though, that we should document the fact that, if a registry does not support the additional lifespan property, then the adapter might fail to correctly route commands.

Yes, we can insert this as the reason why it has to be implemented in the API implementation.

sophokles73: > Since we target this change for the next minor version, meaning we can introduce incompatible changes to providers of the Device Connection API, I think we can indeed require implementations of the API to support the lifespan property.

I am not sure about this, because the (public) Device Connection API describes a dependency on an external system as opposed to APIs where Hono is the actual implementation (provider) of the API, e.g. the API defined by its protocol adapters towards the devices.

@dejanb @ctron WDYT?

calohmn: Created PR #1917 for adapting the documentation.

calohmn: > I am not sure about this, because the (public) Device Connection API describes a dependency on an external system as opposed to APIs where Hono is the actual implementation (provider) of the API, e.g. the API defined by its protocol adapters towards the devices.

I was sticking to the OSGi semantic versioning notions of _providers_ and _consumers_ (with the emphasis of keeping only _consumers_ stable between minor versions) here and seeing the device registry as a _provider_ of the different Hono APIs.

But, yes, in the sense that the device registry is external, it could be seen as even more important to keep things compatible for it between Hono minor versions than for the protocol adapter _consumers_, which are Hono internal (in a way).

But I think, if we go that way of keeping device registry implementations compatible between Hono minor versions, we have to increase Hono major versions in more cases. For example, the Command & Control changes in 1.2 (with its additional, non-optional Device Connection API methods) would have required a 2.0.
And now, the addition of another additional API flag, mentioned in https://github.com/eclipse/hono/issues/1858#issuecomment-617630849, where the support in the implementation should better be not optional, would strictly speaking require us to go for 2.0 instead of 1.3 as well.
In any case, we should decide on which approach to take.

sophokles73: The registry APIs defined as part of Hono all define APIs that Hono *consumes*, i.e. they define what Hono expects to find implemented by some third party entity. The fact that we also provide an example implementation of these APIs doesn't mean that we can change these APIs in a non-backward compatible way as long as we implement these changes in the example registry. So, indeed, I agree that we already would have needed to increase the version to 2.0.0 instead of 1.2.0 because of the newly introduced, mandatory operation in the Device Connection API.

@ctron @dejanb WDYT?

calohmn: On the community call last Thursday it was decided that we want to stick to the OSGi semantic versioning approach.
An implementation of the device registry APIs is seen as a _provider_. Hence there can be incompatible changes between Hono _minor_ versions for the _providers_, requiring changes to the implementation of the device registry APIs. But, at the same time, these API changes must keep the _consumers_, i.e. the protocol adapter implementations, compatible with the previous minor version.
That means that a device registry implementation of a newer minor version still is compatible with protocol adapters of an older minor version of the Hono APIs.
This was seen as the key point here.
It is the obligation of the device registry implementation, to keep up with additive changes in newer minor versions of the APIs then.

PR #1913, opened 2020-04-20
Add option to run integration tests with Jaeger

calohmn:
This adds an option to run the integration tests in such a way that the created traces can later on be viewed in the Jaeger UI.

### DIFF
--- a/tests/pom.xml
+++ b/tests/pom.xml
@@ -58,6 +58,9 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch 
     <!-- logging profile: either prod, dev or trace -->
     <logging.profile>prod</logging.profile>
     <trace.frames>0</trace.frames>
+    <jaeger.disabled>true</jaeger.disabled>
+    <jaeger.host>hono-jaeger.hono</jaeger.host>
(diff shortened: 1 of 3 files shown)


tests/pom.xml
```diff
@@ -58,6 +58,9 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
     <!-- logging profile: either prod, dev or trace -->
     <logging.profile>prod</logging.profile>
     <trace.frames>0</trace.frames>
+    <jaeger.disabled>true</jaeger.disabled>
```
sophokles73, line 61: can we rename this to *jaeger.enabled*? I find it much easier to understand boolean properties that are expressed in their original positive semantics ...

calohmn: That was my initial approach. However that would require quite some gymnastics to invert the value for the `<skip>` field in the Jaeger docker image run configuration. I think that is not worth the effort, especially since the `jaeger.disabled` field is not supposed to be set on the command line. Instead the `jaeger` maven profile should be used.

sophokles73: I see, makes sense

PR #1915, opened 2020-04-21
Fix JavaDoc of ProtocolAdapterProperties

kaniyan:


core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
```diff
@@ -59,7 +59,8 @@ public final boolean isAuthenticationRequired() {
      * <p>
      * The default value of this property is {@code true}.
      *
-     * @param authenticationRequired {@code true} if the server should wait for downstream connections to be established during startup.
+     * @param authenticationRequired {@code true} if the protocol adapter demands the authentication of devices to allow
```
sophokles73, line 62: how about just
> `true` if the protocol adapter should require clients to authenticate

The description already puts it into the right context, FMPOV

kaniyan: I have changed it now.

PR #1916, opened 2020-04-22
[#1858] Use ttd as command consumer lifespan

calohmn:
This is for #1858.

### DIFF
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
@@ -864,10 +864,10 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
         if (authenticatedDevice != null && !authenticatedDevice.getDeviceId().equals(sourceAddress.getResourceId())) {
             // gateway scenario
             return getCommandConsumerFactory().createCommandConsumer(sourceAddress.getTenantId(),
-                    sourceAddress.getResourceId(), authenticatedDevice.getDeviceId(), commandHandler, span.context());
+                    sourceAddress.getResourceId(), authenticatedDevice.getDeviceId(), commandHandler, null, span.context());
         } else {
             return getCommandConsumerFactory().createCommandConsumer(sourceAddress.getTenantId(),
-                    sourceAddress.getResourceId(), commandHandler, span.context());
+                    sourceAddress.getResourceId(), commandHandler, null, span.context());
         }
     }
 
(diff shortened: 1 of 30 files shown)


client/src/main/java/org/eclipse/hono/client/ProtocolAdapterCommandConsumerFactory.java
```diff
@@ -83,6 +85,7 @@ void initialize(CommandTargetMapper commandTargetMapper,
             String tenantId,
             String deviceId,
             Handler<CommandContext> commandHandler,
+            int lifespanSeconds,
```
ctron, line 88: I would prefer the use of `Duration` rather then `int`, which prevents any confusion around the unit.

client/src/main/java/org/eclipse/hono/client/ProtocolAdapterCommandConsumerFactory.java
```diff
@@ -121,6 +126,7 @@ void initialize(CommandTargetMapper commandTargetMapper,
             String deviceId,
             String gatewayId,
             Handler<CommandContext> commandHandler,
+            int lifespanSeconds,
```
ctron, line 129: Same here.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -864,10 +864,10 @@ private Span newSpan(final String operationName, final Device authenticatedDevic
         if (authenticatedDevice != null && !authenticatedDevice.getDeviceId().equals(sourceAddress.getResourceId())) {
             // gateway scenario
             return getCommandConsumerFactory().createCommandConsumer(sourceAddress.getTenantId(),
-                    sourceAddress.getResourceId(), authenticatedDevice.getDeviceId(), commandHandler, span.context());
+                    sourceAddress.getResourceId(), authenticatedDevice.getDeviceId(), commandHandler, -1, span.context());
```
ctron, line 867: If the parameter is optional, then maybe consider using `Optional`. Or maybe `null` when using `Duration` (see other comments on "int").

client/src/main/java/org/eclipse/hono/client/ProtocolAdapterCommandConsumerFactory.java
calohmn: I've amended the commit, changing the type to `Duration` (also in the `DeviceConnectionInfo`, `DeviceConnectionClient` and `DeviceConnectionService` interfaces committed earlier).
I've opted against the usage of `Optional` (it produces quite some clutter here, IMHO). The way to indicate an infinite lifespan is now using either `null` or a negative `Duration`.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
calohmn: See the other comment.

Issue #1732, opened 2020-01-30
AMQP Adapter: Use timeout when sending command to device


calohmn:
When the AMQP adapter sends a command to a device, the adapter waits for the delivery update sent from the device and then itself sends a delivery update to the northbound application.

There is currently no timeout being used when waiting for the delivery update sent from the device.
This still needs to be implemented.

sathishreddynsr:
can someone help to understand..how can we  integrate AMQP adapters to other gateway servers... i mean these AMQP adapters  are already configured to send get,pull requests.

Issue #1632, opened 2019-11-25
Devices stop receiving commands after re-connect


sophokles73:
1. A device that has subscribed to receiving commands on the AMQP adapter.
2. For some reason the AMQP connection fails.
3. The devices opens a new connection to another AMQP adapter instance and again subscribes to receive commans.

In some cases the device is not receiving any commands on its new connection.

One such case could be:
When the device considers its initial connection to the adapter to have failed, the adapter still believes that its connection to the device is alive. Thus, it does not close the command receiver link that it had opened on behalf of the device. Depending on what adapter instance the device re-connects to, the following happens:
1. If the device re-connects to the same instance, then the device's attempt to re-subscribe to commands will fail because the adapter still has such a link open for the device which the adapter still considers working. Any commands sent to the device from an application will be routed to the adapter where delivery of the command will fail because the connection to the device is broken.
2. If the device re-connects to another instance, that instance also opens a command receiver link for the device and now two such links exist and any command sent to the device by an application will either be routed to the initial adapter (where it gets released) or to the new adapter (where delivery will succeed).

There are two issues to be addressed FMPOV:
1) Adapters should close a command receiver link they have opened for a device if a command for that device cannot be delivered to the device due to a stale/broken connection. I believe that the adapter can determine this condition when it tries to forward the command to the device.
2) Adapters should have some means to announce to other adapter instances that they have opened a receiver link for a device. This way, other instances of the adapter get a chance to close their (stale) receiver links for the same device.

dejanb:
> When the device considers its initial connection to the adapter to have failed, the adapter still believes that its connection to the device is alive. Thus, it does not close the command receiver link that it had opened on behalf of the device.

I think we should be able to handle these issues on AMQP level with heart beats and idle timeouts. Is there a case where these tools are not enough?

Is there other examples of these issues other than an example described above?

sophokles73:
> I think we should be able to handle these issues on AMQP level with heart beats and idle timeouts. Is there a case where these tools are not enough?

For the AMQP adapter you are right. In MQTT the heartbeat needs to be offered by the client, i.e. the client may also choose not to provide a heart beat at all. We could, of course, change the MQTT adapter's behavior to require the client to use a heart beat of, say, 30s max. We could then update the user guides accordingly and reject any connection requests that do not comply. Is that what you had in mind?
Still, we could end up with up to 30s where a device does not get any commands after re-connect ...

sophokles73:
[...]

PR #1919, opened 2020-04-22
[#1632] Close command consumer on closed device link

calohmn:
This is for #1632 (and relates to #1732):
A timeout, while waiting for a delivery update for a command sent to a device, now not only leads to the command AMQP link with the device to be detached but also causes the command consumer to be closed.

### DIFF
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
@@ -20,6 +20,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.OptionalInt;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.qpid.proton.amqp.Symbol;
 import org.apache.qpid.proton.amqp.UnsignedLong;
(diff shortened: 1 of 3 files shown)


calohmn:
(looking into the test failure...)

sophokles73:
@calohmn can you rebase this PR

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -880,12 +888,16 @@ private Span newSpan(final String operationName, final Device authenticatedDevic
...
      * @param tenantObject The tenant configuration object.
      * @param sender The link for sending the command to the device.
+     * @param commandConsumer The command consumer.
      * @param commandContext The context in which the adapter receives the command message.
-     * @throws NullPointerException if any of the parameters are {@code null}.
+     * @throws NullPointerException if any of the parameters is {@code null}.
      */
-    protected void onCommandReceived(final TenantObject tenantObject, final ProtonSender sender, final CommandContext commandContext) {
+    protected void onCommandReceived(final TenantObject tenantObject, final ProtonSender sender,
+            final MessageConsumer commandConsumer, final CommandContext commandContext) {
```
sophokles73, line 896: given that we already handle the disposition updates via the CommandContext, I wonder if it wouldn't be more consistent (and elegant) to provide a *closeConsumer()* method in *CommandContext* instead of passing in the *MessageConsumer* as an extra argument. WDYT?

calohmn: Well, the new `CommandContext.consumer` field for that couldn't be final (the CommandContext is not necessarily tied to a consumer as of now) and would need to be set later on. In that sense I don't see the `consumer` object naturally belonging in the `CommandContext`.
But we could use a new CommandContext sub class with that field (or use composition).
I need to think about this. In any case, I would rather do this in a separate PR.

Issue #1848, opened 2020-03-20
Getting 'Failed to authenticate' error when Connection Limit exceeded. 


meshkatul-anwer-bosch:
For a Free-plan tenant, when the number of connected devices is exceeded the limit, further connection attempts were getting rejected with a cause 'Failed to authenticate'. This could be confusing the tenant is using the correct credentials for the device. 

Question is: is it even possible to provide better cause e.g. ' max connections reached'?

kaniyan:
Is it the case for `MQTT` or `AMQP` or both?

meshkatul-anwer-bosch:
This is for AMQP.

kaniyan:
In case of the _AMQP_ adapter, the adapter and tenant level connection limits are being verified inside the `AmqpAdapterSaslAuthenticatorFactory`. The result of the verification is communicated using the `SaslOutcome` and there is no standard error codes that suits our purpose here. Need to check if a wrapper could help. Or simply moving the limit checks out of the `AmqpAdapterSaslAuthenticatorFactory` and performing these checks after authentication should work fine.

sophokles73:
@kaniyan sounds like the right way to go FMPOV.

sathishreddynsr:
can someone help to understand..how can we  integrate AMQP adapters to other gateway servers... i mean these AMQP adapters  are already configured to send get,pull requests.

sophokles73:
closed via #1920

PR #1920, opened 2020-04-23
[#1848] Fix limit verification outcomes in AMQP adapter.

kaniyan:
In AMQP adapter, the checks (such as connection limit, adapter enabled or not ) are performed as part of the `AmqpAdapterSaslAuthenticatorFactory`. If the connection limit check has failed or the adapter is disabled, this approach always returned authentication failure instead of propagating the actual reason. In this PR, these checks are moved out of the `AmqpAdapterSaslAuthenticatorFactory` and there by fixing the issue #1848.


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -357,18 +357,26 @@ private void processRemoteOpen(final ProtonConnection con) {
...
                 // the SASL handshake will already have authenticated the device
-                // and will have verified that the adapter is enabled for the tenant
-                // we still need to check if the device/gateway exists and is enabled
-                checkDeviceRegistration(authenticatedDevice, span.context())
-                .map(ok -> {
-                    log.debug("{} is registered and enabled", authenticatedDevice);
-                    span.log("device is registered and enabled");
-                    return ok;
-                }).setHandler(connectAuthorizationCheck);
+                // and verify if the adapter is enabled for the tenant
```
sophokles73, line 365: what about
```
// we still need to verify that
// the adapter is enabled for the tenant,
// the device/gateway exists and is enabled and
// that the connection limit for the tenant and the adapter are not exceeded.
```

adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -13,6 +13,7 @@
 package org.eclipse.hono.adapter.amqp.impl;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
sophokles73, line 16: can you please refactor the tests to use assertj instead?

kaniyan:
```
 assertThat(errorConditionCaptor.getValue().getCondition().toString())
                .isEqualTo(AmqpError.UNAUTHORIZED_ACCESS.toString());
```

```
 assertEquals(AmqpError.UNAUTHORIZED_ACCESS, 
                errorConditionCaptor.getValue().getCondition());
```
IMHO in the above context, `asserEquals(...)` suited better as there is no need to explicitly convert to string and then compare it as in the case of `asserThat(...)` as it doesn't allow `Symbol ` as an argument. Also we widely use `assertEquals(...)` in other tests too and hence I used it here too. I could also refactor it to use it to `assertThat(...)`. WDYT?

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
kaniyan: Done.

kaniyan:
> Can you also please update the AMQP adapter user guide with information regarding its behavior wrt to exceeded limits?

@sophokles73 I have now updated the user guide and also the release notes.

adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
sophokles73: I see. Let's keep it like this, then ...

PR #1923, opened 2020-04-23
[#1714] Set TTL for connection event messages.

### LINKED ISSUE #1714: Set TTL for Connection Events
There's a specific event message type "Connection events" to inform northbound application about the connection status of the device.

Right now it is not possible to set TTL for such type of events, it actually may cause storage problem, when there's no northbound application to which these connection events must be forwarded or when a northbound application listening only to telemetry messages. In such cases connection events won't be ever delivered to northbound application and won't be automatically deleted.

To prevent such cases TTL for Connection Events may be set and applied.

Alfusainey:
fixes #1714

-  I placed the TTL config parameter in the `ClientConfigProperties` since protocol adapters connect to the messaging network to forward the `connection event` downstream. 
- I  currently set a default TTL value of 5 secs. Users can change the default  value  to any non-negative value by changing the corresponding config property.

@sophokles73 I was thinking of allowing  the  option  to set the value to `-1` if  a connection event message should be  keep in the broker (forever) until  it is consumed. WDYT? maybe this value would make a more sensible default?

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/DownstreamSender.java
+++ b/client/src/main/java/org/eclipse/hono/client/DownstreamSender.java
@@ -133,6 +133,10 @@ public interface DownstreamSender extends MessageSender {
     /**
      * Sends a message for a given device to the endpoint configured for this client.
      *
+     * @param maxTtl   The maximum TTL value in seconds indicating the duration for which the message is to be considered live.
+     *                 <p>
+     *                 The value of this parameter will be contained in the message's <em>ttl</em> header.
+     *
      * @param deviceId The id of the device.
      *                 <p>
      *                 This parameter will be used as the value for the message's application property <em>device_id</em>.
@@ -162,6 +166,7 @@ public interface DownstreamSender extends MessageSender {
      * @throws IllegalArgumentException if the content type specifies an unsupported character set.
      */
     Future<ProtonDelivery> send(
+            Long maxTtl,
             String deviceId,
             Map<String, ?> properties,
             byte[] payload,
--- a/client/src/main/java/org/eclipse/hono/client/DownstreamSender.java
+++ b/client/src/main/java/org/eclipse/hono/client/DownstreamSender.java
@@ -133,10 +133,6 @@ public interface DownstreamSender extends MessageSender {
     /**
      * Sends a message for a given device to the endpoint configured for this client.
      *
-     * @param maxTtl   The maximum TTL value in seconds indicating the duration for which the message is to be considered live.
-     *                 <p>
-     *                 The value of this parameter will be contained in the message's <em>ttl</em> header.
-     *
      * @param deviceId The id of the device.
      *                 <p>
      *                 This parameter will be used as the value for the message's application property <em>device_id</em>.
@@ -166,7 +162,6 @@ public interface DownstreamSender extends MessageSender {
      * @throws IllegalArgumentException if the content type specifies an unsupported character set.
      */
     Future<ProtonDelivery> send(
-            Long maxTtl,
             String deviceId,
             Map<String, ?> properties,
             byte[] payload,
--- a/service-base/src/main/java/org/eclipse/hono/service/monitoring/AbstractMessageSenderConnectionEventProducer.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/monitoring/AbstractMessageSenderConnectionEventProducer.java
@@ -82,23 +82,29 @@ public abstract class AbstractMessageSenderConnectionEventProducer implements Co
             return Future.succeededFuture();
         }
 
-        return getOrCreateSender(context.getMessageSenderClient(), authenticatedDevice.getTenantId())
-                .compose(sender -> {
+        return context.getTenantClient().getOrCreateTenantClient()
+                .map(tenantClient -> tenantClient.get(authenticatedDevice.getTenantId()))
+                .map(tenantObject -> {
+                    return getOrCreateSender(context.getMessageSenderClient(), authenticatedDevice.getTenantId())
+                            .compose(sender -> {
 
-                    final JsonObject payload = new JsonObject();
-                    payload.put("cause", cause);
-                    payload.put("remote-id", remoteId);
-                    payload.put("source", protocolAdapter);
+                                final JsonObject payload = new JsonObject();
+                                payload.put("cause", cause);
+                                payload.put("remote-id", remoteId);
+                                payload.put("source", protocolAdapter);
 
-                    if (data != null) {
-                        payload.put("data", data);
-                    }
+                                if (data != null) {
+                                    payload.put("data", data);
+                                }
 
-                    return sender.send(
-                            authenticatedDevice.getDeviceId(),
-                            payload.encode().getBytes(StandardCharsets.UTF_8),
-                            EventConstants.EVENT_CONNECTION_NOTIFICATION_CONTENT_TYPE
-                            );
+                                final Long maxTtl = tenantObject.result().getResourceLimits().getMaxTtl();
+                                return sender.send(
+                                        maxTtl,
+                                        authenticatedDevice.getDeviceId(),
+                                        null,
+                                        payload.encode().getBytes(StandardCharsets.UTF_8),
+                                        EventConstants.EVENT_CONNECTION_NOTIFICATION_CONTENT_TYPE);
+                            });
                 });
     }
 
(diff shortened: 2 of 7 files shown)


sophokles73:
@Alfusainey thanks for the PR. We already have a mechanism for configuring the (default and max) TTL to use for normal events being published by devices. The TTL in that case can be configured at the tenant and/or device level.
I do not see why the connection event TTL should be protocol adapter specific, i.e. there is no need to make this configurable per adapter. Instead, it is the northbound applications of a tenant which determine how long the events should be kept in the queue. I therefore believe that the TTL to use for connection events should be configurable at the tenant level with a reasonable default value. It would probably make sense to use the already existing [max-ttl](https://www.eclipse.org/hono/docs/dev/api/tenant/#resource-limits-configuration-format) for that purpose as well ...

Alfusainey:
thank kai 👍 I didnt know such a mechanism already exits in hono. i will update the patch

client/src/main/java/org/eclipse/hono/client/DownstreamSender.java
```diff
@@ -162,6 +166,7 @@
      * @throws IllegalArgumentException if the content type specifies an unsupported character set.
      */
     Future<ProtonDelivery> send(
+            Long maxTtl,
```
sophokles73, line 169: I'd rather not introduce a new parameter here but instead simply use the `MessageSender.send(Message)` method for sending the event ...

service-base/src/main/java/org/eclipse/hono/service/monitoring/AbstractMessageSenderConnectionEventProducer.java
```diff
@@ -82,23 +82,29 @@ protected AbstractMessageSenderConnectionEventProducer(
...
+                                payload.put("cause", cause);
+                                payload.put("remote-id", remoteId);
+                                payload.put("source", protocolAdapter);
+
+                                if (data != null) {
+                                    payload.put("data", data);
+                                }
+
+                                final Long maxTtl = tenantObject.result().getResourceLimits().getMaxTtl();
+                                return sender.send(
```
sophokles73, line 101: we can use `MessageHelper.newMessage()` here to create the message and send it using `DownstreamSender.send(Message)` ...

service-base/src/main/java/org/eclipse/hono/service/monitoring/ConnectionEventProducer.java
```diff
@@ -52,6 +53,14 @@
          *         use. This client has to be initialized and started.
          */
         DownstreamSenderFactory getMessageSenderClient();
+        /**
+         * Provides the tenant client which the {@link ConnectionEventProducer} should use to lookup the tenant
+         * that the device connecting to a protocol adapter belongs to.
+         * 
+         * @return The tenant client instance. This client has to be initialized and started.
+         */
+        TenantClientFactory getTenantClient();
```
sophokles73, line 62: IMHO this should be called *getTenantClientFactory* instead

client/src/main/java/org/eclipse/hono/client/DownstreamSender.java
Alfusainey: I was thinking about this too, but then the code that sets some of properties of the message (body, content-type, to, ...) will be duplicated in the two places. will fix

service-base/src/main/java/org/eclipse/hono/service/monitoring/AbstractMessageSenderConnectionEventProducer.java
Alfusainey: `MessageHelper.newMessage()` sets the message's `orig_adapter` application property to the name of the adapter forwarding the connection event. however, the name of the adapter is already included in the message's payload. i added a `todo` comment in the source code to get some feedback

Alfusainey: @sophokles73 if you have time, can you take a look at rev. 4913146 ?

service-base/src/main/java/org/eclipse/hono/service/monitoring/AbstractMessageSenderConnectionEventProducer.java
```diff
@@ -82,27 +85,45 @@ protected AbstractMessageSenderConnectionEventProducer(
...
+        return context.getTenantClientFactory().getOrCreateTenantClient()
+                .map(tenantClient -> tenantClient.get(authenticatedDevice.getTenantId()))
+                .map(tenantObjectFuture -> {
+                    return getOrCreateSender(context.getMessageSenderClient(), authenticatedDevice.getTenantId())
+                            .compose(sender -> {
+
+                                final JsonObject payload = new JsonObject();
+                                payload.put("cause", cause);
+                                payload.put("remote-id", remoteId);
+                                // TODO: can this be removed in favour of 'orig_dapter'?
```
sophokles73, line 97: I don't think so because the Event API defines this as a mandatory property of the message body.

sophokles73:
@Alfusainey can you please add a test case verifying the behavior?

Alfusainey:
> @Alfusainey can you please add a test case verifying the behavior?

sure, i will do

sophokles73:
@Alfusainey there's a Checkstyle violation in one of the integration tests which causes the Travis build to fail ...

service-base/src/test/java/org/eclipse/hono/service/AbstractProtocolAdapterBaseTest.java
```diff
@@ -686,6 +696,44 @@ public void verifyConnectionDurationLimitIsReached(final VertxTestContext ctx) {
                 }));
     }
 
+    /**
+     * Verifies that the (default) ConnectionEvent API configured for a protocol adapter
+     * sets the connection event message's TTL header value before forwarding the message
+     * to downstream aoplications.
```
sophokles73, line 702: applications

service-base/src/test/java/org/eclipse/hono/service/AbstractProtocolAdapterBaseTest.java
```diff
@@ -686,6 +696,44 @@ public void verifyConnectionDurationLimitIsReached(final VertxTestContext ctx) {
...
+     * 
+     * @param ctx The vert.x test context.
+     */
+    @Test
+    public void tesForwardedConnectionEventMessageHasTtlHeaderSet(final VertxTestContext ctx) {
```
sophokles73, line 707: testForwarded ...

service-base/src/test/java/org/eclipse/hono/service/AbstractProtocolAdapterBaseTest.java
```diff
@@ -686,6 +696,44 @@ public void verifyConnectionDurationLimitIsReached(final VertxTestContext ctx) {
...
+        final DownstreamSender connectionEventSender = mock(DownstreamSender.class);
+        when(connectionEventSender.send(any(Message.class))).thenReturn(Future.succeededFuture());
+        when(downstreamSenderFactory.getOrCreateEventSender(Constants.DEFAULT_TENANT)).thenReturn(Future.succeededFuture(connectionEventSender));
+
+        // WHEN a device, belonging to a tenant for which a max TTL is configured, connects to such an adapter
+        final Device authenticatedDevice = new Device(Constants.DEFAULT_TENANT, "4711");
+        final TenantClient tenantClient = mock(TenantClient.class);
+        when(tenantClientFactory.getOrCreateTenantClient()).thenReturn(Future.succeededFuture(tenantClient));
+        final TenantObject tenantObject = TenantObject.from(Constants.DEFAULT_TENANT, true);
+        final ResourceLimits tenantLimits = mock(ResourceLimits.class);
```
sophokles73, line 721: there is no need to mock this class, is there?

Alfusainey: no, not really. will be fixed

Alfusainey:
> @Alfusainey there's a Checkstyle violation in one of the integration tests which causes the Travis build to fail ... this is not due to your PR and should be resolved once you rebase ...

the unused import  travis is complaining about in the `AmqpUploadTestBase` class is actually used in the class. let me rebase  with master and see

Alfusainey:
> @Alfusainey there's a Checkstyle violation in one of the integration tests which causes the Travis build to fail ... this is not due to your PR and should be resolved once you rebase ...

I fixed the changes and rebased this pr with master (since I do not see the unused import travis complaining about). hopefully this will fix the travis build

PR #1931, opened 2020-04-29
Remove Spring Boot component scan annotations

sophokles73:
Based on our agreement to no longer use Spring Boot's component scan for instantiating beans in device registry implementations, I have removed the remaining annotations.


sophokles73:
> Unrelated to this, I was thinking for a while to rename "Dummy" to "Noop" to be more inline with what others use for those cases. WDYT?

Shall I include that change in this PR in a separate commit?

dejanb:
However you prefer.

sophokles73:
ok, will do ...

PR #1932, opened 2020-04-29
Use new interface in createCommandConsumer() return value.

calohmn:
`ProtocolAdapterCommandConsumerFactory.createCommandConsumer` now returns a Future with a value of the new `ProtocolAdapterCommandConsumer` interface type.
The previously used `MessageConsumer` interface with its `flow()` and `getRemainingCredit()` methods wasn't a good fit anymore because both these methods do nothing in the implementation used for the `ProtocolAdapterCommandConsumerFactory`.

Furthermore, a `SpanContext` parameter has been added to the `close()` method of the new interface. This allows the trace spans created in the `commandConsumer.close()` operation to be properly assigned to the corresponding parent span (in an `UNSUBSCRIBE` operation for example).

Special care in this PR has been taken to ensure that the parent span to the one, used for the `commandConsumer.close()` invocation, is only finished _after_ the `close()` method result Future is completed.


adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1196,12 +1201,12 @@ protected final void close(final MqttEndpoint endpoint, final Device authenticat
...
-            final Span closeHandlerSpan = newSpan("Send Disconnected Event", endpoint, authenticatedDevice,
-                    traceSamplingPriority);
-            sendDisconnectedTtdEvent(tenant, device, authenticatedDevice, span.context())
-                    .setHandler(sendAttempt -> closeHandlerSpan.finish());
-        });
+        final CompositeFuture removalDoneFuture = cmdHandler.removeAllSubscriptions((tenant, device) -> {
+            final Span sendEventSpan = newChildSpan(span.context(), "Send Disconnected Event", endpoint,
+                    authenticatedDevice);
+            return sendDisconnectedTtdEvent(tenant, device, authenticatedDevice, sendEventSpan.context())
+                    .onComplete(r -> sendEventSpan.finish()).mapEmpty();
```
sophokles73, line 1208: If I am not mistaken, then this code sends a disconnect event *for each* subscription being canceled. However, FMPOV we only need to send the disconnect *once after all* subscriptions have been canceled. WDYT?

calohmn: IMHO, the only scenario, where an MQTT device would want to create multiple subscriptions over the same MQTT connection, would be for a gateway device to add multiple subscriptions for commands targeted at different, specific devices. In that case, the `sendDisconnectedTtdEvent` method with its `deviceId` parameter should actually be called for _all_ these devices separately.
So, I think, the code makes sense, the way it is.

PR #1933, opened 2020-04-30
Externalize opentracing by default

ctron:
This removes the embedded OpenTracing API dependencies and adds a new profile named `embed-opentracing` which brings back the original behavior. However, this profile is not enabled by default.

IIRC this was the plan I discussed with @sophokles73 during BCX, the make it possible not-include the opentracing API, but also keep the current behavior as an option.

Additionally this PR fixes the OSGi ExecutionEnvironment configuration, which was still using the `java.level` property, which was removed a while back.

### DIFF
--- a/site/homepage/content/release-notes.md
+++ b/site/homepage/content/release-notes.md
@@ -45,6 +45,11 @@ title = "Release Notes"
 * The *set command-handling protocol adapter instance* operation of the *Device Connection* API
   has been extended to support an additional parameter which can be used to indicate the
   maximum amount of time that the given information is to be considered valid.
+* The `hono-core` module no longer embeds any external classes by default.
+  This includes the *OpenTracing API* and *Spring Crypto* classes. You can still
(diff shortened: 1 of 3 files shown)


ctron:
@sophokles73 I added a note to the release notes, explained what has changed and how to get back the old behavior. Do you want some additional documentation elsewhere?

site/homepage/content/release-notes.md
```diff
@@ -45,6 +45,10 @@ title = "Release Notes"
 * The *set command-handling protocol adapter instance* operation of the *Device Connection* API
   has been extended to support an additional parameter which can be used to indicate the
   maximum amount of time that the given information is to be considered valid.
+* The `hono-core` module no longer embeds the *OpenTracing API* classes by default.
```
sophokles73, line 48: nor the Spring Crypto classes, for that matter

ctron: True.

PR #1934, opened 2020-04-30
Add properties to the TenantKey class to make it more reusable withou…

dejanb:
…t inheritance


sophokles73:
@dejanb is this still relevant or can we close it?

PR #1936, opened 2020-05-04
[#1858] Add 'updateOnly' param to setCommandHandlingAdapterInstance

calohmn:
This is for #1858.

### DIFF
--- a/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/Cache.java
+++ b/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/Cache.java
@@ -45,6 +45,7 @@ public interface Cache<K, V> {
      * @return A succeeded future containing the previous value or {@code null} if the
      *         cache didn't contain the key yet.
      *         A failed future if the value could not be stored in the cache.
+     * @throws NullPointerException if any of the parameters is {@code null}.
      */
     Future<V> put(K key, V value);
 
@@ -58,8 +59,23 @@ public interface Cache<K, V> {
      * @return A succeeded future containing the previous value or {@code null} if the
      *         cache didn't contain the key yet.
      *         A failed future if the value could not be stored in the cache.
+     * @throws NullPointerException if any of the parameters is {@code null}.
      */
     Future<V> put(K key, V value, long lifespan, TimeUnit lifespanUnit);
+
+    /**
+     * Replaces the entry for a key only if currently mapped to a given value.
+     *
+     * @param key The key.
+     * @param oldValue The value to overwrite.
+     * @param newValue The value to store.
+     * @param lifespan The lifespan of the entry. A negative value is interpreted as an unlimited lifespan.
+     * @param lifespanUnit The time unit for the lifespan.
+     * @return A succeeded future containing a boolean, indicating whether the value was replaced or not.
+     *         A failed future if the value could not be stored in the cache.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     */
+    Future<Boolean> replace(K key, V oldValue, V newValue, long lifespan, TimeUnit lifespanUnit);
 
     /**
      * Gets a value from the cache.
@@ -68,6 +84,7 @@ public interface Cache<K, V> {
      * @return A succeeded future containing the value or {@code null} if the
      *         cache didn't contain the key yet.
      *         A failed future if the value could not be read from the cache.
+     * @throws NullPointerException if key is {@code null}.
      */
     Future<V> get(K key);
 
@@ -79,6 +96,7 @@ public interface Cache<K, V> {
      * @return A succeeded future containing {@code true} if the key was
      *         mapped to the value, {@code false} otherwise.
      *         A failed future if the value could not be remove from the cache.
+     * @throws NullPointerException if any of the parameters is {@code null}.
      */
     Future<Boolean> remove(K key, V value);
 
@@ -87,6 +105,7 @@ public interface Cache<K, V> {
      *
      * @param keys The keys.
      * @return A succeeded future containing a map with key/value pairs.
+     * @throws NullPointerException if keys is {@code null}.
      */
     Future<Map<K, V>> getAll(Set<? extends K> keys);
 }
--- a/client/src/test/java/org/eclipse/hono/client/impl/ProtocolAdapterCommandConsumerFactoryImplTest.java
+++ b/client/src/test/java/org/eclipse/hono/client/impl/ProtocolAdapterCommandConsumerFactoryImplTest.java
@@ -135,7 +135,7 @@ public class ProtocolAdapterCommandConsumerFactoryImplTest {
         when(deviceConnectionClientFactory.connect()).thenReturn(Future.succeededFuture(mock(HonoConnection.class)));
         when(deviceConnectionClientFactory.getOrCreateDeviceConnectionClient(anyString()))
                 .thenReturn(Future.succeededFuture(devConClient));
-        when(devConClient.setCommandHandlingAdapterInstance(anyString(), anyString(), any(), any()))
+        when(devConClient.setCommandHandlingAdapterInstance(anyString(), anyString(), any(), anyBoolean(), any()))
                 .thenReturn(Future.succeededFuture());
         when(devConClient.removeCommandHandlingAdapterInstance(anyString(), anyString(), any()))
                 .thenReturn(Future.succeededFuture(Boolean.TRUE));
@@ -187,7 +187,7 @@ public class ProtocolAdapterCommandConsumerFactoryImplTest {
                 ctx.verify(() -> {
                     verify(connection).createReceiver(eq(tenantCommandAddress), eq(ProtonQoS.AT_LEAST_ONCE), any(), anyInt(),
                             eq(false), any());
-                    verify(devConClient).setCommandHandlingAdapterInstance(eq(deviceId), anyString(), any(), any());
+                    verify(devConClient).setCommandHandlingAdapterInstance(eq(deviceId), anyString(), any(), eq(false), any());
                 });
                 ctx.completeNow();
             }));
@@ -211,7 +211,7 @@ public class ProtocolAdapterCommandConsumerFactoryImplTest {
             ctx.verify(() -> {
                 verify(connection).createReceiver(eq(tenantCommandAddress), eq(ProtonQoS.AT_LEAST_ONCE), any(), anyInt(),
                         eq(false), any());
-                verify(devConClient).setCommandHandlingAdapterInstance(eq(deviceId), anyString(), any(), any());
+                verify(devConClient).setCommandHandlingAdapterInstance(eq(deviceId), anyString(), any(), eq(false), any());
                 // verify closing the consumer is successful
                 consumer.close(ctx.succeeding(v -> {
                     ctx.verify(() -> {
--- a/service-base/src/main/java/org/eclipse/hono/service/deviceconnection/DelegatingDeviceConnectionAmqpEndpoint.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/deviceconnection/DelegatingDeviceConnectionAmqpEndpoint.java
@@ -16,6 +16,7 @@ import java.net.HttpURLConnection;
 import java.time.Duration;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 
 import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.client.ClientErrorException;
@@ -241,6 +242,9 @@ public class DelegatingDeviceConnectionAmqpEndpoint<S extends DeviceConnectionSe
         final String deviceId = MessageHelper.getDeviceId(request);
         final String adapterInstanceId = MessageHelper.getApplicationProperty(request.getApplicationProperties(), MessageHelper.APP_PROPERTY_ADAPTER_INSTANCE_ID, String.class);
         final Integer lifespanSecondsOrNull = MessageHelper.getApplicationProperty(request.getApplicationProperties(), MessageHelper.APP_PROPERTY_LIFESPAN, Integer.class);
+        final boolean updateOnly = Optional
+                .ofNullable(MessageHelper.getApplicationProperty(request.getApplicationProperties(), MessageHelper.APP_PROPERTY_UPDATE_ONLY, Boolean.class))
+                .orElse(false);
 
         final Span span = TracingHelper.buildServerChildSpan(
                 tracer,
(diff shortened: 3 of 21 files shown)


client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/Cache.java
```diff
@@ -61,6 +61,19 @@
...
+     *
+     * @param key The key.
+     * @param oldValue The value to overwrite.
+     * @param newValue The value to store.
+     * @param lifespan The lifespan of the entry. A negative value is interpreted as an unlimited lifespan.
+     * @param lifespanUnit The time unit for the lifespan.
+     * @return A succeeded future containing a boolean, indicating whether the value was replaced or not.
+     *         A failed future if the value could not be stored in the cache.
+     */
+    Future<Boolean> replace(K key, V oldValue, V newValue, long lifespan, TimeUnit lifespanUnit);
```
sophokles73, line 75: what if any of the params are `null`?

site/documentation/content/api/device-connection/index.md
```diff
@@ -129,6 +129,7 @@ The following table provides an overview of the properties a client needs to set
 | *subject*             | yes       | *properties*             | *string*  | MUST be set to `set-cmd-handling-adapter-instance`. |
 | *adapter_instance_id* | yes       | *application-properties* | *string*  | The identifier of the protocol adapter instance that currently handles commands for the device or gateway identified by the *device_id* property. |
 | *lifespan*            | no        | *application-properties* | *int*     | The lifespan of the mapping entry in seconds. After that period, the mapping entry shall be treated as non-existent by the *Device Registration API* methods. A negative value, as well as an omitted property, is interpreted as an unlimited lifespan. |
+| *only_update*         | no        | *application-properties* | *boolean* | If set to `true`, the command will only update an existing mapping entry with the given lifespan, provided such an entry with the given device id and adapter instance id exists. If the property is omitted or set to `false`, the mapping entry will be set regardless whether an entry with the given adapter instance id exists or not. |
```
sophokles73, line 132: can we please rename this parameter to *update_only*?

site/documentation/content/api/device-connection/index.md
```diff
@@ -141,9 +142,11 @@ The response message's *status* property may contain the following codes:
 | Code  | Description |
 | :---- | :---------- |
 | *204* | OK, the adapter instance for the device has been updated. |
-| *400* | Bad Request, the command-handling adapter instance has not been updated due to invalid or missing data in the request. |
+| *400* | Bad Request, the command-handling adapter instance has not been set or updated due to invalid or missing data in the request. |
+| *404* | Not Found, entry to be updated was not found. This applies to requests with *only_update* set to `true`. |
```
sophokles73, line 146: FMPOV this should be returned if no entry has been found for the given device ID, regardless of the value of *update_only*.
We should then expicitly list *412* as the value to be returned for *update_only* = `true` and a non-matching orig-value. I do not see why the registry implementation should return 404 or 412 in that case, given that we expect the registry to explicitly support this property anyway ...

calohmn: > FMPOV this should be returned if no entry has been found for the given device ID, regardless of the value of update_only.

But, whether an entry for the given device ID is found or not, only matters if `update_only` is `true` (if `update_only` is false, a new mapping entry with the device ID will just be created).

> We should then expicitly list *412* as the value to be returned for *update_only* = `true` and a non-matching orig-value. I do not see why the registry implementation should return 404 or 412 in that case, given that we expect the registry to explicitly support this property anyway ...

Two things made me go for choosing *404* as the default for both the key-not-found and adapter-instance-not-matching cases here:
- the `removeCommandHandlingAdapterInstance` command also requires the adapterInstanceId to match. For that command we have defined 404 as the default if no matching entry was found, and 412 as an optional alternative to 404.
- a straightforward implementation might use a `replace(key, oldValue, newValue)` method here. With that, there is no distinction, whether an entry for the key wasn't found or whether the `oldValue` didn't match. In fact for the cache based implementation, I've used the corresponding `cache.replaceAsync()` method. (That could of course be changed to use `cache.computeIfAbsentAsync` where a distinction between return values can be made then.)

Therefore, for the sake of keeping in sync with the other API methods and for making it easier for implementations (not requiring the key-not-found vs adapter-instance-not-matching distinction, which also isn't needed in the code using the API), I would rather keep the API definition as is.

site/documentation/content/api/device-connection/index.md
calohmn: yes, I've changed that (also renamed the parameter).

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/Cache.java
calohmn: I've added checks and adapted the javadoc (also for the other methods).

site/documentation/content/api/device-connection/index.md
sophokles73: > But, whether an entry for the given device ID is found or not, only matters if update_only is true (if update_only is false, a new mapping entry with the device ID will just be created).

true, based on that it would probably make sense to never return 404 but to return 412 if *update_only* = `true` and either no entry for the given device ID exists at all or the expected value doesn't match.

> the removeCommandHandlingAdapterInstance command also requires the adapterInstanceId to match. For that command we have defined 404 as the default if no matching entry was found, and 412 as an optional alternative to 404.

maybe we should change that as well. I don't really like the fact that a client needs to support both status codes but still doesn't really know what the problem is ...

calohmn: I would be fine with always returning 412 (and not having the distinction between the two cases).

To be consistent, I agree that `removeCommandHandlingAdapterInstance` should then be changed as well. I think, regarding compatibility, we can do this in 1.3, since "412" is already defined there as a possible return value.

Another thing would be changing the 404 status code for the optional tenant check in the set/remove operations. I would like to change that to 412 as well. But since `setLastKnownGateway` doesn't define 412 as possible return code as of now, I think we can only do this in Hono 2.0.
Therefore I would keep the 404 as the common status code for the tenant check in all methods for now.

calohmn: I've updated the documentation accordingly.
(Changes concerning `removeCommandHandlingAdapterInstance` will go in a separate PR then.)

site/documentation/content/api/device-connection/index.md
```diff
@@ -140,8 +141,9 @@ The response message's *status* property may contain the following codes:
 
 | Code  | Description |
 | :---- | :---------- |
-| *204* | OK, the adapter instance for the device has been updated. |
-| *400* | Bad Request, the command-handling adapter instance has not been updated due to invalid or missing data in the request. |
+| *204* | OK, the command-handling adapter instance for the device has been updated. |
+| *400* | Bad Request, the adapter instance for the device has not been set or updated due to invalid or missing data in the request. |
+| *412* | Precondition failed, the to be updated adapter instance entry for the device doesn't exist, or its adapter instance value doesn't match. This applies to requests with *update_only* set to `true`. |
```
sophokles73, line 146: how about
> Precondition failed, the adapter instance for the device has not been set or updated. This status is returned if the *update_only* property is set to `true` but no matching entry for the given device ID and adapter instance value exists.

client/src/test/java/org/eclipse/hono/client/impl/ProtocolAdapterCommandConsumerFactoryImplTest.java
```diff
@@ -187,7 +187,7 @@ public void testCreateCommandConsumerSucceeds(final VertxTestContext ctx) {
                 ctx.verify(() -> {
                     verify(connection).createReceiver(eq(tenantCommandAddress), eq(ProtonQoS.AT_LEAST_ONCE), any(), anyInt(),
                             eq(false), any());
-                    verify(devConClient).setCommandHandlingAdapterInstance(eq(deviceId), anyString(), any(), any());
+                    verify(devConClient).setCommandHandlingAdapterInstance(eq(deviceId), anyString(), any(), anyBoolean(), any());
```
sophokles73, line 190: don't we expect the *updateOnly* param to be `false` here?

client/src/test/java/org/eclipse/hono/client/impl/ProtocolAdapterCommandConsumerFactoryImplTest.java
```diff
@@ -211,7 +211,7 @@ public void testCreateTimeLimitedCommandConsumerSucceeds(final VertxTestContext
             ctx.verify(() -> {
                 verify(connection).createReceiver(eq(tenantCommandAddress), eq(ProtonQoS.AT_LEAST_ONCE), any(), anyInt(),
                         eq(false), any());
-                verify(devConClient).setCommandHandlingAdapterInstance(eq(deviceId), anyString(), any(), any());
+                verify(devConClient).setCommandHandlingAdapterInstance(eq(deviceId), anyString(), any(), anyBoolean(), any());
```
sophokles73, line 214: don't we expect the *updateOnly* param to be `false` here?

service-base/src/main/java/org/eclipse/hono/service/deviceconnection/DelegatingDeviceConnectionAmqpEndpoint.java
```diff
@@ -241,6 +241,7 @@ public DelegatingDeviceConnectionAmqpEndpoint(final Vertx vertx, final S service
         final String deviceId = MessageHelper.getDeviceId(request);
         final String adapterInstanceId = MessageHelper.getApplicationProperty(request.getApplicationProperties(), MessageHelper.APP_PROPERTY_ADAPTER_INSTANCE_ID, String.class);
         final Integer lifespanSecondsOrNull = MessageHelper.getApplicationProperty(request.getApplicationProperties(), MessageHelper.APP_PROPERTY_LIFESPAN, Integer.class);
+        final Boolean updateOnlyOrNull = MessageHelper.getApplicationProperty(request.getApplicationProperties(), MessageHelper.APP_PROPERTY_UPDATE_ONLY, Boolean.class);
```
sophokles73, line 244: how about
```
final boolean updateOnly = Optional.ofNullable(MessageHelper.getApplicationProperty(request.getApplicationProperties(), MessageHelper.APP_PROPERTY_UPDATE_ONLY, Boolean.class))
  .orElse(false);
```

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/deviceconnection/MapBasedDeviceConnectionService.java
```diff
@@ -128,9 +129,26 @@ public MapBasedDeviceConnectionsConfigProperties getConfig() {
...
+                                return newValue;
+                            } else {
+                                return oldValue;
+                            }
+                        });
+                if (resultingValue != null && protocolAdapterInstanceId.equals(getAdapterInstanceIdFromJson(resultingValue.getValue()))) {
+                    result = DeviceConnectionResult.from(HttpURLConnection.HTTP_NO_CONTENT);
+                } else {
+                    result = DeviceConnectionResult.from(HttpURLConnection.HTTP_PRECON_FAILED);
+                }
```
sophokles73, line 147: I think we need to also remove an expired existing value if the adapter instance doesn't match ...

sophokles73: @calohmn WDYT?

calohmn: Removing an expired entry here is something I don't quite understand - the removal of an expired entry is done under the hood by the Caffeine cache.

However, I've found two other issues here: one (updated lifespan wasn't taken into account) is already fixed (along with an added test), the other (a failed update influences lifespan of existing entry) is something I'm currently working on.

sophokles73: But we are using a simple `Map` here for keeping the data, not a Caffeine Cache, or am I mistaken?

calohmn: The `adapterInstancesForTenantMap` is a Caffeine Cache under the hood (using `cache.asMap()`).

sophokles73: I see

calohmn: 2nd issue is fixed now.

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/deviceconnection/MapBasedDeviceConnectionService.java
```diff
@@ -159,13 +171,13 @@ public long expireAfterCreate(final String key, final ExpiringValue<JsonObject>
...
                             final long currentTime, final long currentDuration) {
-                        return Long.MAX_VALUE;
+                        return value.getLifespanNanos();
                     }
 
                     @Override
                     public long expireAfterRead(final String key, final ExpiringValue<JsonObject> value,
                             final long currentTime, final long currentDuration) {
-                        return Long.MAX_VALUE;
+                        return currentDuration;
```
sophokles73, line 180: My understanding is that we do not want to change an existing value's lifespan on reading, right? Does *currentDuration* contain the object's *remaining* lifetime based on the lifetime determined at creation/update time?

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/deviceconnection/MapBasedDeviceConnectionService.java
```diff
@@ -300,6 +312,22 @@ private String getAdapterInstanceIdFromJson(final JsonObject adapterInstanceIdJs
         return expiringValue != null ? expiringValue.getValue() : null;
     }
 
+    /**
+     * Replaces the entry for a key only if the matching function applied to the current value returns {@code true}.
+     */
+    private static <V, K> boolean replaceIfMatching(final ConcurrentMap<K, V> map, final K key, final V newValue,
+            final Function<? super V, Boolean> matchingFunction) {
```
sophokles73, line 319: java.util.function.Predicate ?

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/deviceconnection/MapBasedDeviceConnectionService.java
calohmn: Yes. As the javadoc states: `The {@code currentDuration} may be returned to not modify the expiration time.`. So using `Long.MAX_VALUE` was wrong here.

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/deviceconnection/MapBasedDeviceConnectionService.java
calohmn: Yes, that's better.

PR #1941, opened 2020-05-06
[#1632] Close MQTT connection when ack timeout has occurred

calohmn:
This is for #1632:
When the MQTT adapter sends a command (with Qos 1) to the device and doesn't receive an acknowledgement in time, the MQTT connection is now closed, a disconnect notification is sent downstream and the command consumer is closed.

I've done some refactoring here regarding the place where the `ackTimeout` and also `handlePubAck()` logic is located:
Previously, some parts were in `CommandHandler`, others in `AbstractVertxBasedMqttProtocolAdapter`, with the needed objects for the logic being passed between the two. The `ackTimeout` was handled in `CommandHandler.startTimer()`.
I found that somewhat confusing. 

Now, the `ackTimeout` and `handlePubAck()` logic is defined in two Handlers in `AbstractVertxBasedMqttProtocolAdapter.afterCommandPublished`, with these Handlers passed along to the `CommandSubscriptionsManager` (renamed from `CommandHandler`) which will invoke the Handlers when needed.
(Along the way, trace handling in `handlePubAck()` was also fixed (`span.log()` had been called after  the span was already finished).)

The main new behaviour is, that in the `ackTimeout` handling, the MQTT connection with the device is disconnected now.
The needed followup-actions (removal of subscription and command consumer) are done in the `close` method of the adapter, triggered by the lost connection.

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
@@ -23,8 +23,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.OptionalInt;
-import java.util.function.BiConsumer;
-import java.util.function.Function;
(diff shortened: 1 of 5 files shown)


calohmn:
Based on the discussion in #1632, closing the connection doesn't seem like a good option. So, closing this PR.

PR #1945, opened 2020-05-07
Refactor MQTT "afterCommandPublished" handling, fix tracing, add test

calohmn:
This takes over the refactorings and tracing fixes done in https://github.com/eclipse/hono/pull/1941, but keeps the behaviour in case of a command ack timeout (i.e. the MQTT connection is kept open).


sophokles73:
@calohmn does this change the behavior of the MQTT adapter towards devices? Is it worth describing this behavior in the MQTT adapter user guide?

calohmn:
@sophokles73 no, there is no such behaviour change here.

PR #1946, opened 2020-05-07
Verify tenant from command target address

sophokles73:
When mapping and delegating a command message, the tenant is now being
extracted from the message's address instead of using the (static)
tenant used for creating the tenant scoped command receiver link.

Signed-off-by: Kai Hudalla <kai.hudalla@bosch.io>

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/MappingAndDelegatingCommandHandler.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/MappingAndDelegatingCommandHandler.java
@@ -18,6 +18,7 @@ import java.util.Objects;
 
 import org.apache.qpid.proton.amqp.messaging.ApplicationProperties;
 import org.apache.qpid.proton.amqp.messaging.Rejected;
+import org.apache.qpid.proton.amqp.transport.AmqpError;
 import org.apache.qpid.proton.amqp.transport.ErrorCondition;
 import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.client.Command;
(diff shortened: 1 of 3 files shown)


client/src/main/java/org/eclipse/hono/client/impl/MappingAndDelegatingCommandHandler.java
```diff
@@ -96,34 +96,45 @@ public MappingAndDelegatingCommandHandler(final HonoConnection connection,
...
+        // this is the place where a command message on the "command/${tenant}" address arrives *first*
+        if (message.getAddress() == null) {
+            LOG.debug("command message has no address");
+            final Rejected rejected = new Rejected();
+            rejected.setError(new ErrorCondition(Constants.AMQP_BAD_REQUEST, "missing command target address"));
+            originalMessageDelivery.disposition(rejected, true);
+            return;
+        }
+        final ResourceIdentifier targetAddress = ResourceIdentifier.fromString(message.getAddress());
+        final String tenantId = targetAddress.getTenantId();
```
calohmn, line 123: What about a downstream application sending a command on a `command/TenantA` link and using `command/TenantB/someDevice` in the command message `to` address?
Can we be sure that the AMQP messaging network rejects such a message if the sender has only been given rights to send on the `command/TenantA` address? I'm not sure if the qdrouter for example checks the `to` address here as well.

sophokles73: Good point. This, however, means that we would not be able to receive all commands for all tenants using a link source address of `tenant/#`.
BTW the current implementation also does not check if the tenant ID from the command target address matches the tenant from the link source address used to receive the command. Instead, we simply use the tenant from the link source address with the device ID from the command target address. This effectively prevents applications from sending a command to a device of another tenant but does not detect programming errors if a device with the same ID exists in multiple tenants ...

calohmn: > This, however, means that we would not be able to receive all commands for all tenants using a link source address of `tenant/#`.

(You mean `command/#`, I guess.)
I've done a test here and the Qdrouter seems to allow sending messages with a `to` address, to which the sender hasn't been given rights for, as long as there are rights for the link address. (I guess the more expensive `to` address auth check in the Qdrouter for every received message is only done in case of a sender on the anonymous link address.)
So that is something that speaks against using the `command/#` receiver link instead of the tenant based links.
Therefore I would suggest keeping the tenant parameter here. The missing comparison with the tenant from the `to` address is indeed missing and should be added.

sophokles73: @calohmn I have added the check for the tenant from the command message.
Would you mind taking another look?

PR #1948, opened 2020-05-07
[#1632] Keep link open on cmd disposition update timeout

calohmn:
Based on the discussion in #1632, the command link to the device is now kept open when the device hasn't sent a disposition update in time.

### DIFF
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
@@ -20,7 +20,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.OptionalInt;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.qpid.proton.amqp.Symbol;
 import org.apache.qpid.proton.amqp.UnsignedLong;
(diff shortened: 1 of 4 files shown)


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -762,18 +749,29 @@ protected void handleRemoteSenderOpenForCommands(
...
+    private void onCommandConnectionClose(final String tenantId, final String deviceId, final Device authenticatedDevice,
+            final ProtocolAdapterCommandConsumer consumer, final OptionalInt traceSamplingPriority) {
+
+        final Span span = newSpan("handle closing of command connection", authenticatedDevice, traceSamplingPriority);
+        final Future<ProtonDelivery> sendEventFuture = sendDisconnectedTtdEvent(tenantId, deviceId, authenticatedDevice,
+                span.context());
+        final Future<Void> closeConsumerFuture = consumer.close(span.context());
+
+        CompositeFuture.join(sendEventFuture, closeConsumerFuture).recover(thr -> {
+            Tags.ERROR.set(span, true);
```
sophokles73, line 761: why not log the error to the span?

calohmn: The error(s) will be logged at the respective child spans. Therefore I think it's not needed here, also because both of the child operations could have failed, meaning there are 2 errors.

PR #1955, opened 2020-05-12
Fix abstract credential services tests

kaniyan:
In the `AbstractCredentialsServiceTest` test class, there are few tests for creating credentials of various types for the devices. The correct way to test is first to create devices and then update those credentials. This has been fixed in this PR. Now `createDevice(...)` has been invoked before updating the credentials using `updateCredentials(...)`.


kaniyan:
This has been closed in favor of #1972.

PR #1965, opened 2020-05-18
[#1951] Close all command consumers on connection close

### LINKED ISSUE #1951: AMQP adapter: gateways with >1 command consumers over same connection not fully supported
The scenario here involves an authenticated gateway opening multiple command consumer links for specific devices (i.e. multiple links of the form `command/${tenant}/${device-id}`) over the same AMQP connection.
In that case, closing that connection means that only the last created consumer link gets its associated resources properly closed with the corresponding notification getting sent downstream.

In the code this becomes apparent by the fact that there is only one `connectionLossHandler` per connection, handling removal of the resources for _one_ command consumer.

calohmn:
This fixes #1951.

### DIFF
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
@@ -13,13 +13,17 @@
 package org.eclipse.hono.adapter.amqp.impl;
 
 import java.net.HttpURLConnection;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.OptionalInt;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import org.apache.qpid.proton.amqp.Symbol;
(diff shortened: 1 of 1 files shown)


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -1250,23 +1255,37 @@ private void reportSentCommand(final TenantObject tenantObject, final CommandCon
...
+    private static boolean removeConnectionLossHandler(final ProtonConnection con, final String key) {
+        @SuppressWarnings("unchecked")
+        final Map<String, Function<Span, Future<Void>>> handlers = con.attachments().get(KEY_CONNECTION_LOSS_HANDLERS, Map.class);
+        return handlers != null && handlers.remove(key) != null;
     }
 
     private static Device getAuthenticatedDevice(final ProtonConnection con) {
-        return Optional.ofNullable(con.attachments())
-                .map(attachments -> attachments.get(AmqpAdapterConstants.KEY_CLIENT_DEVICE, Device.class))
+        return Optional.ofNullable(con.attachments().get(AmqpAdapterConstants.KEY_CLIENT_DEVICE, Device.class))
```
sophokles73, line 1280: does *ProtonConnection.attachments()* always return a non-null value?

calohmn: With the current Proton implementation classes, yes. And besides, it would be inconsistent having the null check here, but not in the many other places where the attachments are used.

PR #1972, opened 2020-05-20
[#1679] Part 3: Add tests for mongo db based credentials services

kaniyan:
This PR contains unit tests for the MongoDB based implementation of credentials service and management APIs. 

It also contains the below fixes in the `AbstractCredentialsServiceTest` class which are required for the for the above unit tests.
1. Now the test devices are created first before adding credentials to those devices.
2. The values being set during the _Future_ completion are being used even before that Future completes. This has been fixed.


sophokles73:
@kaniyan can you please rebase and resolve conflicts?

Issue #1956, opened 2020-05-12
Define a set of authorised characters for device Id, auth-id and tenant id


jbtrystram:
Right now there is now defined set of authorised characters for those identifiers in the documentation, except that semicolon is forbidden for auth-id and a note that advise to use `[a-zA-Z0-9_-]`. 

To create a UI it would be good to have a defined regexp so the input can be validated when creating a new device for example. 

I propose to set `[a-zA-Z0-9_-]` (maybe with a dot `.` in addition) as the accepted characters in the documentation for device ID, auth-id and tenant ID.

calohmn:
@jbtrystram @dejanb @sophokles73 
As brought up in the Hono gitter channel, I think a release notes entry is missing here about a Device Registry API change, also informing how to configure the tenant/device pattern, so that e.g. systems using Ditto can still create devices (with a colon in the id).

Regarding the changes in the `device-registry-v1.yaml`, defining the schema pattern for the device and tenant identifier:
Isn't this an inconsistency that the patterns are explicitly defined in the API, while they can still be adapted using the configuration properties? I mean, a device registry that has overridden the patterns using the properties doesn't conform to the API anymore, does is?
Maybe we should therefore remove the device/tenant patterns from the API spec?
I think it would at least be good to allow the colon in the device id, to conform with Eclipse Ditto.
WDYT?

sophokles73:
> Regarding the changes in the device-registry-v1.yaml, defining the schema pattern for the device and tenant identifier:
Isn't this an inconsistency that the patterns are explicitly defined in the API, while they can still be adapted using the configuration properties? I mean, a device registry that has overridden the patterns using the properties doesn't conform to the API anymore, does is?

I agree. This seems rather inappropriate.

> Maybe we should therefore remove the device/tenant patterns from the API spec?

Yes, I agree.

> I think it would at least be good to allow the colon in the device id, to conform with Eclipse Ditto.
WDYT?

The default pattern used in the registry should do so, yes.

PR #1978, opened 2020-05-25
Add regexp validations for tenant ID, device ID, credentials types and auth-ID.

jbtrystram:
This is a proposal to fix https://github.com/eclipse/hono/issues/1956


sophokles73:
@jbtrystram can you fix the integration tests?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DelegatingDeviceManagementHttpEndpoint.java
```diff
@@ -93,21 +93,25 @@ public void addRoutes(final Router router) {
 
         // CREATE device
         router.post(pathWithTenantAndDeviceId)
+                .handler(this::verifyDeviceId)
```
dejanb, line 96: I think it would be better to do this check within span, so that we have a proper tracing/logging for the errors.

core/src/main/java/org/eclipse/hono/util/CredentialsConstants.java
```diff
@@ -116,6 +116,14 @@
      * The vert.x event bus address to which inbound credentials messages are published.
      */
     public static final String EVENT_BUS_ADDRESS_CREDENTIALS_IN = "credentials.in";
+    /**
+     * The regular expression to validate that the auth-id field supplied in credentials is legal.
+     */
+    public static final String AUTH_ID_VALUE_REGEX = "^[a-z0-9-=]+$";
```
dejanb, line 122: Maybe you can define default reges at one place, like `DEFAULT_ID_REGEX` and reuse that value.

jbtrystram: I moved the regexp for tenant id and device id to `RegistryManagementConstants.DEFAULT_ID_REGEX`

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -810,8 +810,10 @@ components:
          properties:
             "type":
                type: string
+               pattern: [a-z0-9-]
```
dejanb, line 813: We should update a default patterns for different identifiers and add some docs on the configuration.

jbtrystram: done :)

PR #1984, opened 2020-05-27
[Not to merge] Make use of already downloaded flapdoodle mongo db artifacts

kaniyan:
An Embedded Mongo DB(_de.flapdoodle.embed.mongo_) instance is being used for the _mongo db based device registry_ unit tests. This gets downloaded during every build. The flapdoodle documentation talks about a static approach which caches the extracted executables and library files. But the problem with this approach is, that it doesn't work with _travis_ and the _travis_ build fails. 

The tests in this approach work fine in a local environment. But the travis build fails. This PR is to investigate that.


kaniyan:
[ERROR] org.eclipse.hono.deviceregistry.mongodb.service.MongoDbBasedRegistrationServiceTest  Time elapsed: 0.375 s  <<< ERROR!
java.lang.IllegalArgumentException: Could NOT create Directory /home/travis/.embedmongo

It tries to create a directory inside the home directory (_/home/travis_) and it couldn't create.

kaniyan:
This has been closed in favour of #2007.

PR #1989, opened 2020-05-28
Upgrade Infinispan to the most recent stable version

ctron:


bom/pom.xml
```diff
@@ -36,7 +36,7 @@
     <guava.version>28.0-jre</guava.version>
     <caffeine.version>2.8.0</caffeine.version>
     <hamcrest-core.version>2.2</hamcrest-core.version>
-    <infinispan.version>9.4.17.Final</infinispan.version>
+    <infinispan.version>10.1.8.Final</infinispan.version>
```
sophokles73, line 39: is a 10.1.x client compatible with a 9.4.x server?

ctron: It should be: https://access.redhat.com/documentation/en-us/red_hat_data_grid/7.0/html/developer_guide/compatibility_between_server_and_hot_rod_client_versions

You can also configure the protocol level explicitly, if needed: https://access.redhat.com/webassets/avalon/d/red-hat-data-grid/8.0/api/org/infinispan/client/hotrod/configuration/package-summary.html

According to the Hotrod protocol documentation of Infinispan 10, Hotrod 2.9 (Infinispan 9.4) should still be possible with Hotrod 3.0 (Infinispan 10): https://infinispan.org/docs/stable/titles/hotrod_protocol/hotrod_protocol.html#hot_rod_protocol_2_9

sophokles73: fingers crossed ;-)

PR #1993, opened 2020-05-28
[#1679] Integration tests for the mongo db based device registry

kaniyan:


tests/pom.xml
```diff
@@ -223,6 +239,54 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
               <autoCreateCustomNetworks>true</autoCreateCustomNetworks>
               <startParallel>false</startParallel>
               <images>
+                <!-- ##### Mongo DB instance for the device registry ##### -->
+                <image>
+                  <name>${docker.image.org-name}/hono-mongodb-test</name>
+                  <alias>mongo</alias>
+                  <build>
+                    <imagePullPolicy>IfNotPresent</imagePullPolicy>
```
sophokles73, line 247: IMHO we should also skip building the image if the MongoDB profile is disabled

kaniyan: It makes sense and I have now skipped building the image if the MongoDB profile is disabled. Would you mind taking a look again?

tests/pom.xml
```diff
@@ -228,6 +244,55 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
...
+                    <network>
+                      <mode>custom</mode>
+                      <name>hono</name>
+                      <alias>${hono.mongodb.host}</alias>
+                    </network>
+                    <memorySwap>314572800</memorySwap>
+                    <memory>314572800</memory>
+                    <log>
+                      <prefix>MONGODB</prefix>
+                      <color>green</color>
```
sophokles73, line 288: please use `${log.color.hono-device-registry}` instead

kaniyan: I have updated it.

Issue #1914, opened 2020-04-21
coap-adapter, 4.01 - DTLS session does not contain authenticated Device


boaks:
In some rare cases I get responses with

4.01 - DTLS session does not contain authenticated Device

FMPOV, that is caused by a timeout in the second request of the `DefaultDeviceResolver`. If the first would fail, then the handshake would not succeed and therefore no response would be received.

Is it possible to enable the "amqp client caching" used there with a short cache timeout (e.g. 30s) in order to prevent the coap-adapter from failing with that 4.01 ?

sophokles73:
Can you please describe what you are observing and why you believe that it is a problematic/erroneous behavior?

boaks:
> coap-adapter

> I get responses

> 4.01 - DTLS session does not contain authenticated Device

I'm not sure, what should be missing here? 

> FMPOV, that is caused by a timeout in the second request of the DefaultDeviceResolver. If the first would fail, then the handshake would not succeed and therefore no response would be received.

That's my analysis. Let me try to fillin some more words. The `DefaultDeviceResolver` sends two requests for the same authid/tenant (agreed?).
I consider the first  to be in time, therefore the handshake finished successfully and the request and response could be exchanged (agreed?). 
If the first would have failed, that exchange would fail, and no response would have been received  (agreed?).
The error response then points to the second request of the `DefaultDeviceResolver`, which, according the comment there

```
// client will only wait a limited period of time,
// so no need to use get(Long, TimeUnit) here
```

may timeout. So my assumption is, the authid/tenant works (therefore the response), the second request times out, therefore the 4.01, and enable a caching with a short cache-timeout may help, if the analyis is right.

boaks:
> believe that it is a problematic/erroneous behavior?

Why should someone be irritated by a 4.01 (pointing to an authentication issue), if this is the result of timeout?

boaks:
> rare

I have seen this twice in a couple of weeks. Therefore detailed logs are not available.
It seems for me to be much easier, to try to configure the cache, and see if the problem still exists, than to explicitly try to reproduce it.

sophokles73:
I am sorry, but I still don't understand what you are talking about. Is this about a single request from a device and the DTLS handshake that is involved? Or are we talking about multiple requests (and handshakes) from the same device?

boaks:
[...]

PR #1994, opened 2020-05-28
Cache device-id for dtls to add it later to the principal.

boaks:
Issue #1914


boaks:
A good proposal to replace the californium `LeastRecentlyUsedCache`will be welcome :-).

boaks:
In the meantime I also tried to pass the device id as "custom argument" within the `PskSecretResult` back and pass that to an "extended" `AdvancedApplicationLevelInfoSupplier.getInfo(Principal peerIdentity, Object customArgument)`. If that `customArgument` is `null`, it's requested as currently implemented before this PR. That obsoletes this cache, but would slightly wire the `ApplicationLevelInfoSupplier` to the `AdvancedPskStore`.

boaks:
Californium is extended by that custom argument. Therefore this PR is obsolete.

PR #1995, opened 2020-05-29
Add searchDevicesForTenant operation in openAPI

jbtrystram:

### DIFF
--- a/site/documentation/content/api/management/device-registry-v1.yaml
+++ b/site/documentation/content/api/management/device-registry-v1.yaml
@@ -36,7 +36,7 @@ info:
    license:
       name: EPL-2.0
       url: https://www.eclipse.org/legal/epl-2.0/
-   version: 1.3.0
+   version: 1.4.0
 
 externalDocs:
    description: Eclipse Hono™ web page
@@ -324,6 +324,32 @@ paths:
             403:
                $ref: '#/components/responses/NotAllowed'
 
+      get:
+         tags:
+            - devices
+         summary: Search devices for a tenant with optional filters, paging and sorting options.
+         operationId: searchDevicesForTenant
+         requestBody:
+            description: Filters, paging and sorting options.
+            content:
+               application/json:
+                  schema:
+                     $ref: '#/components/schemas/PagingSortingFilteringOptions'
+            required: false
+         responses:
+            200:
+               description: operation successful
+               content:
+                  application/json:
+                     schema:
+                        $ref: '#/components/schemas/SearchDevicesResult'
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            404:
+               $ref: '#/components/responses/NotFound'
+
    /devices/{tenantId}/{deviceId}:
 
       parameters:
@@ -500,7 +526,7 @@ paths:
             details of its secrets like password hash and/or hash algortihm used.
             This mode of operation is called *replace mode*. However, this behavior is
             discouraged because it might disclose confidential information in plain text.
-            
+
             Instead, implementors should remove all confidential information from the secrets
             and include a (unique) identifier with each secret. Clients can then refer to
             this identifier in order to update particular secrets. This mode of operation
@@ -625,7 +651,95 @@ components:
          description: Allows arbitrary properties as extension to the ones
                       specified by the Hono API.
 
-# Tenant schema
+
+      PagingSortingFilteringOptions:
+         type: object
+         additionalProperties: false
+         properties:
+            "page":
+               $ref: '#/components/schemas/PagingOptions'
+            "filters":
+               type: array
+               description: A list of filter expressions. Objects must match all filters in order to be included in a result.
+               items:
+                  $ref: '#/components/schemas/FilterOption'
+            "sort":
+               type: array
+               description: A list of sorting options. If multiples options are provided, the results will be sorted following the order of the options in the list.
+               items:
+                  $ref: '#/components/schemas/SortingOption'
+
+      PagingOptions:
+         type: object
+         additionalProperties: false
+         description: options to page the results of bulk queries.
+         properties:
+            "limit":
+               type: integer
+               default: 30
+               minimum: 0
+               maximum: 200
+               description: The maximum number of objects to include in a response.
+            "offset":
+               type: integer
+               default: 0
+               minimum: 0
+               description: |
+                  The offset to use as a starting point to return results. This allows to go through the query results in
+                  a paged manner when the query returns more results than the limit.
+
+      FilterOption:
+            type: object
+            additionalProperties: false
+            description: Filters to be applied to bulk queries.
+            required:
+               - field
+               - value
+               - op
+            properties:
+               "field":
+                  type: string
+                  description: |
+                     A Json Pointer identifying the field to use for filtering.
+                     Json pointer is defined in RFC 6901.
+               "value":
+                  type: string
+                  description: |
+                     The value to filter the field with.
+                     Wildcard characters are supported : `*` will match zero or any number of characters
+                     and `?` will match exactly one character.
+               "op":
+                  type: string
+                  description: The operator to use when applying the filter.
+                  enum:
+                     - eq
+            examples:
+               Filters options:
+                  $ref: '#/components/examples/FilterOptionsExample'
+
+      SortingOption:
+         type: object
+         additionalProperties: false
+         description: Sorting option to be applied to bulk queries results.
+         required:
+            - field
+         properties:
+            "field":
+               type: string
+               description: |
+                  The Json Pointer identifying the field to use for sorting.
+                  Json pointer is defined in RFC 6901.
+            "direction":
+               type: string
+               default: ascending
+               enum:
+                  - ascending
+                  - descending
+         examples:
+            Sorting options:
+               $ref: '#/components/examples/SortingOptionsExample'
+
+      # Tenant schema
 
       Tenant:
          type: object
@@ -979,6 +1093,23 @@ components:
             "last-user":
                type: string
                description: The user who made the last edit on this device.
+
+      SearchDevicesResult:
+         type: object
+         description: The result of a search request.
+         additionalProperties: false
+         properties:
+            "total":
+               type: String
+               description: The total number of results returned by the query, regardless of the *limit* set in the paging options.
+            "results":
+               $ref: '#/components/schemas/DevicesList'
+
+      DevicesList:
+         type: array
+         description: A list of devices.
+         items:
+            $ref: '#/components/schemas/Device'
 
 # Credentials
 
@@ -1371,3 +1502,22 @@ components:
                   "not-after": "2027-12-24T19:00:00Z",
                }]
             }]
+      FilterOptionsExample:
+         value:
+           [
+            {
+                field: "/enabled",
+                value: true,
+                op: "eq"
+            },
+            {
+                field: "/ext/brand",
+                value: "eclipse*",
+                op: "eq"
+            }
+           ]
+      SortingOptionsExample:
+         value:
+            field: "/status/created"
+            direction: "ascending"
+


site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -461,7 +492,43 @@ components:
...
+      FilterOptions:
+            type: object
+            additionalProperties: true
+            description: Allows arbitrary filters to be applied to bulk querries.
+                         Fields to filter MUST be designated with their JSON Path representation.
+
+      SortingOptions:
+         type: object
+         additionalProperties: true
+         description: Sorting options to be applied to bulk querries results.
```
dejanb, line 528: Typo in "querries"

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -461,7 +492,43 @@ components:
...
+               type: integer
+               default: 30
+            "offset":
+               type: integer
+               default: 0
+
+      FilterOptions:
+            type: object
+            additionalProperties: true
+            description: Allows arbitrary filters to be applied to bulk querries.
```
dejanb, line 522: Typo in "querries"

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -257,6 +257,37 @@ paths:
...
+               headers:
+                  ETag:
+                     description: Version of the resource
+                     schema:
+                        type: string
+            400:
+               $ref: '#/components/responses/MalformedRequest'
+            401:
+               $ref: '#/components/responses/Unauthorized'
+            404:
```
ctron, line 288: I am not sure what "not found" would mean in this case. When searching, I would expect an empty result would still be "200 OK".

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -257,6 +257,37 @@ paths:
             403:
                $ref: '#/components/responses/NotAllowed'
 
+      get:
+         tags:
+            - devices
+         summary: Retrieve all devices for a tenant
+         operationId: getAllDevicesForTenant
```
ctron, line 264: I think the naming of this is misleading. It is not returning "all devices". I think "search" would make more sense. Same for the "summary".

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -461,7 +492,43 @@ components:
...
+               $ref: '#/components/schemas/PagingOptions'
+            "filters":
+               $ref: '#/components/schemas/FilterOptions'
+            "sort":
+               $ref: '#/components/schemas/SortingOptions'
+
+      PagingOptions:
+         type: object
+         additionalProperties: false
+         description: options to page the results of bulk querries.
```
ctron, line 510:
```suggestion
         description: options to page the results of bulk queries.
```

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -461,7 +492,43 @@ components:
...
+         properties:
+            "limit":
+               type: integer
+               default: 30
+            "offset":
+               type: integer
+               default: 0
+
+      FilterOptions:
```
ctron, line 519: I am not sure how this would look like. Maybe adding an example helps.

site/documentation/content/api/management/device-registry-v1.yaml
jbtrystram: and what if the tenant doesn't exist ?

site/documentation/content/api/management/device-registry-v1.yaml
jbtrystram: I updated the name to `searchDevicesForTenant`

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -257,6 +257,37 @@ paths:
...
+            required: false
+         responses:
+            200:
+               description: operation successful
+               content:
+                  application/json:
+                     schema:
+                        $ref: '#/components/schemas/Device'
+               headers:
+                  ETag:
```
ctron, line 280: I don't think you can provide that here using headers, as you return multiple entries, and they may have different versions.

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -257,6 +257,32 @@ paths:
...
+                  schema:
+                     $ref: '#/components/schemas/PagingSortingFilteringOptions'
+            required: false
+         responses:
+            200:
+               description: operation successful
+               content:
+                  application/json:
+                     schema:
+                        $ref: '#/components/schemas/Device'
```
sophokles73, line 278: this will most likely be an array of devices instead of just a single device, won't it?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -461,7 +487,49 @@ components:
...
+         type: object
+         additionalProperties: false
+         description: options to page the results of bulk queries.
+         properties:
+            "limit":
+               type: integer
+               default: 30
+            "offset":
+               type: integer
+               default: 0
```
sophokles73, line 512: how about some proper description of these parameters?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -1111,3 +1179,11 @@ components:
                "id": "349556ea-4902-47c7-beb0-1009ab693fb4",
                "not-after": "2027-12-24T19:00:00Z",
             }]
+      FilterOptionsExample:
+         value:
+            enabled: true
+            ext.brand: "eclipse"
+      SortingOptionsExample:
+         value:
+            status.created: "ascending"
```
sophokles73, line 1188: I would expect the possible values for the sort direction to be defined somewhere, e.g. in an enum?

jbtrystram:
@sophokles73 thanks your feedback, those are good points. I updated the descriptions and added a DevicesSet schema.

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -847,6 +857,12 @@ components:
             "ext":
                $ref: '#/components/schemas/Extensions'
 
+      DevicesSet:
+         type: array
+         description: A set of devices.
```
kaniyan, line 862: A _set_ is an unordered collection and in this case IMHO it is misleading as the type is _array_ (which is ordered). How about _A list of devices_? Also how about _DeviceList_ instead of _DeviceSet_.

jbtrystram: I updated accordingly

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -461,7 +487,59 @@ components:
...
+         type: object
+         additionalProperties: false
+         description: options to page the results of bulk queries.
+         properties:
+            "limit":
+               type: integer
+               default: 30
+               description: |
+                  The threshold on the number of results returned in a single API call. If not set, the registry will return up to 30 results.
+                  The upper bound for this value is 200 and it cannot be less than 0.
```
sophokles73, line 512: we should be able to formally define the min and max using the `minimum` and `maximum` properties ...

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -461,7 +487,59 @@ components:
...
+            "offset":
+               type: integer
+               default: 0
+               description: |
+                  The offset to use as a starting point to return results. This allows to go through the query results in
+                  a paged manner when the query returns more results than the limit.
```
sophokles73, line 518: `minimum`: 0 ?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -461,7 +487,59 @@ components:
...
+         additionalProperties: true
+         description: Sorting options to be applied to bulk queries results.
+                     The field to be used for sorting MUST be defined with its JSON Path representation.
+         default: ascending
+         enum:
+            - ascending
+            - descending
+         examples:
+            Sorting options:
+               $ref: '#/components/examples/SortingOptionsExample'
```
sophokles73, line 540: If I understand the examples correctly, then the property names of this object are supposed to be JSON paths like `status.created` and the property values need to be either `ascending` or `descending`, right?
However, my understanding of this definition is that any instance of SortingOptions will be either the literal String `ascending` (default) or `descending`. The enum property does **not** define the possible values of the *additionalProperties* but the value of the *SortingOptions* object itself. Or am I mistaken? I do not think that it would even be possible to express the enum constraint based on your approach of variable property names because AFAIK you can only define constraints on particular properties or objects. So, FMPOV the SortingOptions would need to be changed to something that looks more like this:
```json
[
  {
      "property": "status.created",
      "direction": "descending"
  }, {
       "property": "last-user",
       "direction": "ascending"
  }
]
```

jbtrystram: I think you are right. I am still not mastering the openAPI specification :)
Your suggestion fits nicely so I updated the spec with that. Thanks !

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -528,13 +530,17 @@ components:
...
-         enum:
-            - ascending
-            - descending
+         properties:
+            "field":
+               type: string
+               description: The JSON Path representation of the field to use for sorting.
+            "direction":
+               default: ascending
+                  enum:
```
sophokles73, line 541: I think the indentation is wrong here ...

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -461,7 +487,65 @@ components:
...
+      PagingSortingFilteringOptions:
+         type: object
+         additionalProperties: false
+         properties:
+            "page":
+               $ref: '#/components/schemas/PagingOptions'
+            "filters":
+               $ref: '#/components/schemas/FilterOptions'
+            "sort":
+               $ref: '#/components/schemas/SortingOptions'
```
sophokles73, line 500: can I only filter and sort by one property each?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -495,7 +495,7 @@ components:
             "page":
                $ref: '#/components/schemas/PagingOptions'
             "filters":
-               $ref: '#/components/schemas/FilterOptions'
+               $ref: '#/components/schemas/FilterOptionsSet'
```
sophokles73, line 498: are these ANDed together? In any case, we should document the semantics explicitly ...

jbtrystram: I updated the description. Let me know what you think

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -521,7 +521,8 @@ components:
 
       FilterOptionsSet:
          type: array
-         description: A set of filter options.
+         description: A set of filter options. If multiples filters are provided, they will be applied together
+         in a cumulative way.
```
kaniyan, line 525: The second line of the description has wrong indentation. Please take a look.

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -461,7 +487,72 @@ components:
...
+         items:
+            $ref: '#/components/schemas/FilterOption'
+
+      FilterOption:
+            type: object
+            additionalProperties: true
+            description: Allows arbitrary filters to be applied to bulk queries.
+                         Fields to filter MUST be designated with their JSON Path representation.
```
sophokles73, line 533: I think it would be better to use the same approach as we do for the sorting options, i.e. explicitly define two properties one containing the *path* expression and the other containing the value to match. That way we would be able to define constraints on the properties (if we so choose) and the list of filters actually makes some sense because otherwise there is no need for an array of FilterOption objects because a single FilterObject could already contain arbitrary filter expressions ...

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -461,7 +487,72 @@ components:
...
+               default: 0
+               minimum: 0
+               description: |
+                  The offset to use as a starting point to return results. This allows to go through the query results in
+                  a paged manner when the query returns more results than the limit.
+
+      FilterOptionsSet:
+         type: array
+         description: A set of filter options. If multiples filters are provided, they will be applied together
+         in a cumulative way.
```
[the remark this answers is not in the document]
jbtrystram: Thanks ! I struggled quite a bit to make this understandable, your proposal is better I think

site/documentation/content/api/management/device-registry-v1.yaml
jbtrystram: Yes it's a good idea to have sorting options and filters to be designed the same way :)

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -952,6 +1047,23 @@ components:
             "ext":
                $ref: '#/components/schemas/Extensions'
 
+      SearchDevicesResult:
+         type: object
+         description: The result of a search request.
+         additionalProperties: false
+            properties:
```
kaniyan, line 1054: The `properties` entry has incorrect indentation.

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -324,6 +324,32 @@ paths:
             403:
                $ref: '#/components/responses/NotAllowed'
 
+      get:
+         tags:
+            - devices
+         summary: Search devices for a tenant, with optional filters and paging options.
```
kaniyan, line 330: IMHO it is also better to include `sorting` in the summary. How about `Search devices for a tenant with optional filters, paging and sorting options`?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -952,6 +1047,23 @@ components:
...
+      SearchDevicesResult:
+         type: object
+         description: The result of a search request.
+         additionalProperties: false
+            properties:
+               "total":
+                  type: String
+                  description: The total number of results returned by the query, regardless of the "limit" set in the paging options.
+               "results":
+                  $ref: '#/components/schemas/Device'
```
kaniyan, line 1059: The search operation should return `#/components/schemas/DevicesList` (a list of devices) and not `#/components/schemas/Device`.

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -620,7 +646,76 @@ components:
...
+            type: object
+            additionalProperties: false
+            description: Filters to be applied to bulk queries.
+            properties:
+               "field":
+                  type: string
+                  description: The JSON Path representation of the field to use for filtering.
+               "value":
+                  type: string
+                  description: The value to filter the field with.
```
sophokles73, line 696: will this be matched exactly or can we use a regex here or wildcards?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -620,7 +646,76 @@ components:
...
+            examples:
+               Filters options:
+                  $ref: '#/components/examples/FilterOptionsExample'
+
+      SortingOptions:
```
sophokles73, line 701: this object defines a single sort option only, doesn't it?
If you want to allow for multiple sort options to be given, then the `PagingSortingFilteringOptions.sort` property needs to be an array, right?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -952,6 +1047,23 @@ components:
...
                $ref: '#/components/schemas/Extensions'
 
+      SearchDevicesResult:
+         type: object
+         description: The result of a search request.
+         additionalProperties: false
+         properties:
+            "total":
+               type: String
+               description: The total number of results returned by the query, regardless of the "limit" set in the paging options.
```
sophokles73, line 1057: how about replacing the quotes with `*limit*`?

site/documentation/content/api/management/device-registry-v1.yaml
jbtrystram: IMHO having multiples sort options at the same time is a bit weird (from a UI/UX perspective) but i guess it doesn't hurt to support it in the back end.

sophokles73: I am in no way arguing that we *must* have multiple sorting options. We could as well have just one. I only want to make sure that we know what we want/get :-) So, supporting only one sort option for the time being is fine with me. In any case, the object should be named `SortingOption` as it only defines a single property and sort direction, right? If we want to support multiple options later, then we would allow specifying an array of `SortingOption` instances in `PagingSortingFilteringOptions`. In that case (multiple options) my understanding would be that the results are ordered following the order of the `SortingOption` instances in the array ...

jbtrystram: I am fine with both as well. Sorry if my comment came across as arguing, I wasn't either. 
I am thinking that having an array to support multiples options is the most flexible. The underlying implementation can choose to only support one and apply the first element of the array.

site/documentation/content/api/management/device-registry-v1.yaml
jbtrystram: I don't know what is the better choice honnestly. I suppose it depends on the underlying database, what does it supports. 
@dejanb @ctron any opinions on this ?

sophokles73: Again, I am not arguing that this *must* support wildcards or regex. In any case, we must document the expected behavior ...

site/documentation/content/api/management/device-registry-v1.yaml
jbtrystram: I pushed a change using multiples filters and improved the documentation.

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -620,7 +646,82 @@ components:
...
-# Tenant schema
+
+      PagingSortingFilteringOptions:
+         type: object
+         additionalProperties: false
+         properties:
+            "page":
+               $ref: '#/components/schemas/PagingOptions'
+            "filters":
+               $ref: '#/components/schemas/FilterOptionsList'
```
sophokles73, line 657: how about using
```yaml
"filters":
  type: array
  items:
    $ref: '#/components/schemas/FilterOption'
```

in order to not having to create a separate object definition for each list of objects?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -620,7 +646,82 @@ components:
...
+            "sort":
+               $ref: '#/components/schemas/SortingOptions'
```
sophokles73, line 659: this refers to a non-existing type. Should probably also be
```yaml
"sort":
  type: array
  items:
    $ref: '#/components/schemas/SortingOption'
```
and then we can remove `SortingOptionsList`

sophokles73:
@jbtrystram any news on this one? It would be good if we could get it into 1.3.0 FMPOV ..

jbtrystram:
> @jbtrystram any news on this one? It would be good if we could get it into 1.3.0 FMPOV ..

@sophokles73 thanks for your suggestions of skipping the middle man and put an array of filters and sorting options, it makes more sense indeed. I updated in that direction, but kept the descriptions for the arrays, as it contained some useful info. 

I think the last detail to iron out is what kind of filter should be supported, as you pointed out, if wildcards or regexps are allowed..

ctron:
I think it would make sense to have a wildcard, but I wouldn't go for regular expressions … something like `foo like '%bar%baz'` can easily be translated into most query languages … while a full blow regular expression can be very hard.

sophokles73:
In any case, specifying and implementing this will take some time, I guess. For the time being (and for 1.3.0) I would be ok with only supporting `eq` as an operator. Bosch's biggest interest currently is being able to page through all devices. Filtering based on a sophisticated query language is (currently) not that important to us.
@jbtrystram @ctron @calohmn WDYT? Could we start with just the `eq` operator and later extend with new capabilities?

jbtrystram:
@sophokles73 @calohmn @ctron @dejanb I updated the spec in that direction, documenting that `%` and `_` wildcards are supported, and supporting only the `eq` operator. It's an enum field so we can add more operators later on

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -620,7 +646,83 @@ components:
...
+            properties:
+               "field":
+                  type: string
+                  description: The JSON Path representation of the field to use for filtering.
+               "value":
+                  type: string
+                  description: |
+                     The value to filter the field with.
+                     SQL wildcard characters are supported : `%` will match zero or any number of characters
+                     and `_` will match exactly one character.
```
sophokles73, line 699: FMPOV standard wildcards like `*` (zero or more characters) and `?` one character would be more appropriate as they are more widely known.
Just out of curiosity: if *eq* allows for using wildcards, what are be semantics of the *like* operation? I sthat about phonetic similarity or similar spelling? 
Based on the fact that we are allowing wildcards, the operation that we are talking about would rather be *like* and not *eq*, right?

jbtrystram: I went with `%` and `_` because they are standard SQL wildcards, where `*` and `?` are more used in shell environment. I have no preference there. 

And yeah, I agree, I would intuitively say that `eq *word*` is similar to `like word`, but it probably changes with actual database implementations

jbtrystram: Let's do a poll . 
@dejanb  @ctron @sophokles73 @calohmn  please vote on that comment :
:+1: for `*` and `?`
:-1: for `%` and `_`

dejanb: I see that REST APIs use mostly */? combination, so let's go with that.

jbtrystram: I updated the PR with */? as wildcards.

sophokles73: did you?

jbtrystram: ahem, sorry.. This should be fixed :grin:

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -1343,3 +1462,22 @@ components:
                   "not-after": "2027-12-24T19:00:00Z",
                }]
             }]
+      FilterOptionsExample:
+         value:
+           [
+            {
+                field: "enabled",
+                value: true
```
kaniyan, line 1470: Missing comma at the end of the field `value`.

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -1343,3 +1462,22 @@ components:
...
+                op: "eq"
+            },
+            {
+                field: "ext.brand",
+                value: "eclipse*"
```
kaniyan, line 1475: Missing comma at the end of the filed `value`.

jbtrystram:
@kaniyan thanks for spotting those issues.
@sophokles73 sorry for the delay, bumped to 1.3.0

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -36,7 +36,7 @@ info:
    license:
       name: EPL-2.0
       url: https://www.eclipse.org/legal/epl-2.0/
-   version: 1.2.0
+   version: 1.3.0
```
sophokles73, line 39: The API is already at 1.3.0 on master ...

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -620,7 +646,83 @@ components:
...
+            examples:
+               Filters options:
+                  $ref: '#/components/examples/FilterOptionsExample'
+
+      SortingOption:
+         type: object
+         additionalProperties: false
+         description: Sorting option to be applied to bulk queries results.
+         properties:
+            "field":
```
sophokles73, line 713: FMPOV this property is required, right?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -620,7 +646,83 @@ components:
...
+               "value":
+                  type: string
+                  description: |
+                     The value to filter the field with.
+                     Wildcard characters are supported : `*` will match zero or any number of characters
+                     and `?` will match exactly one character.
+               "op":
+                  description: The operator to use when applying the filter.
+                  enum:
+                     - eq
```
sophokles73, line 703: all of these properties are required, aren't they?

site/documentation/content/api/management/device-registry-v1.yaml
jbtrystram: oops, fixed

site/documentation/content/api/management/device-registry-v1.yaml
jbtrystram: fixed.

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -625,7 +651,90 @@ components:
...
+            additionalProperties: false
+            description: Filters to be applied to bulk queries.
+            required:
+               - field
+               - value
+               - op
+            properties:
+               "field":
+                  type: string
+                  required: true
```
kaniyan, line 702: `required` is an object-level attribute and not a property attribute. This doesn't belong here. This can be removed.

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -625,7 +651,90 @@ components:
...
+                  description: The JSON Path representation of the field to use for filtering.
+               "value":
+                  type: string
+                  description: |
+                     The value to filter the field with.
+                     Wildcard characters are supported : `*` will match zero or any number of characters
+                     and `?` will match exactly one character.
+               "op":
```
kaniyan, line 710: `type: string` is missing for this enum object `op`.

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -625,7 +651,90 @@ components:
...
+         type: object
+         additionalProperties: false
+         description: Sorting option to be applied to bulk queries results.
+         required:
+            - field
+         properties:
+            "field":
+               type: string
+               description: The JSON Path representation of the field to use for sorting.
+            "direction":
```
kaniyan, line 728: `type: string` is missing for this enum object `direction`.

jbtrystram:
Thanks @kaniyan for the comments, you know all the details of the openAPI syntax :)

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -625,7 +651,91 @@ components:
...
+            additionalProperties: false
+            description: Filters to be applied to bulk queries.
+            required:
+               - field
+               - value
+               - op
+            properties:
+               "field":
+                  type: string
+                  description: The JSON Path representation of the field to use for filtering.
```
sophokles73, line 702: FMPOV we need to specify explicitly which parts of JSONPath we support here. In particular we seem to only use the *dot* notation and omit the `$` indicating the root element, right? We do seem to support array indexes but do not support predicates. Instead we use a separate property for predicates (*op*), right? Without a detailed specification of what we support here, this feature will be very hard to implement consistently ...

jbtrystram:
Updated to use Json Pointer

PR #1999, opened 2020-06-02
[#1998] Set lifespan for lastKnownGateway cache entries

### LINKED ISSUE #1998: Set explicit expiration for Device Connection cache entries (prevent timeout exceptions with Infinispan 9.4))
"Last-known-gateway" entries in the cache used for the Device Connection API implementation should expire after a certain time. To achieve this, the "device-connection" cache definition for the Infinispan server was set up accordingly, using a `<expiration max-idle="2419200000" />` definition (see the corresponding file in the iot-packages repo: https://github.com/eclipse/packages/blob/ca4113c01064e16c0a9f8fc078004f83e4b0aae7/charts/hono/templates/example-data-grid/configmap.yaml#L274).

There are 2 issues with that:
- as described in the Infinspan HotRod protocol implementation javadoc (https://docs.jboss.org/infinispan/10.1/apidocs/org/infinispan/client/hotrod/RemoteCache.html), the `maxIdle` definition in the server will be ignored when using the HotRod client. Workaround is to set the desired maxIdle in each remote cache operation.
- the `RemoteCache#removeWithVersionAsync` operation will result in a timeout in an Infinispan 9.4 server having been configured with `<expiration max-idle="2419200000" />` (there is no timeout without that configuration).

Therefore we should set a `maxIdle` value on each cache operation, instead of relying on the server configuration to set the expiration.

calohmn:
This fixes #1998.

Applies a fixed lifespan for the `lastKnownGateway` cache entries. This is a minimalistic sotution for #1998. We should probably make this lifespan configurable. This can be done in a subsequent PR (possibly along with renaming `CommonCacheConfig` for that).

### DIFF
--- a/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/CacheBasedDeviceConnectionInfo.java
+++ b/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/CacheBasedDeviceConnectionInfo.java
@@ -51,6 +51,11 @@ import io.vertx.ext.healthchecks.Status;
 public final class CacheBasedDeviceConnectionInfo implements DeviceConnectionInfo, HealthCheckProvider {
 
     /**
+     * Lifespan for last-known-gateway cache entries.
+     */
+    static final Duration LAST_KNOWN_GATEWAY_CACHE_ENTRY_LIFESPAN = Duration.ofDays(28);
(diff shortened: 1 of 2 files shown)


sophokles73:
@calohmn don't forget to merge this into master **and** 1.2.x. Also, please add a corresponding entry to the release notes ...

calohmn:
@sophokles73 done.

PR #2002, opened 2020-06-03
1.2.x: Ensure Tracer gets autowired into HonoConnection

calohmn:
This fixes the issue that some spans created during Command & Control handling are missing in the overall traces.

Only needed for 1.2.x.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
@@ -41,10 +41,12 @@ import org.eclipse.hono.util.DeviceConnectionConstants;
 import org.eclipse.hono.util.RegistrationConstants;
 import org.eclipse.hono.util.TenantConstants;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.config.AutowireCapableBeanFactory;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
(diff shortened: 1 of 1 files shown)


service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
```diff
@@ -444,15 +446,19 @@ public HonoConnection commandConsumerConnection() {
...
     @Bean
     @Scope("prototype")
-    public ProtocolAdapterCommandConsumerFactory commandConsumerFactory() {
+    public ProtocolAdapterCommandConsumerFactory commandConsumerFactory(final ApplicationContext appContext) {
         // be sure to use a new ClientConfigProperties instance so that its unique container-id
         // is only used for the specific consumer factory instance created here (as adapterInstanceId, which has to be unique)
         final ClientConfigProperties clientConfigProperties = new ClientConfigProperties(commandConsumerFactoryConfig());
         final HonoConnection honoConnection = HonoConnection.newConnection(vertx(), clientConfigProperties);
+        // make sure the tracer instance gets autowired into the honoConnection
+        appContext.getAutowireCapableBeanFactory()
```
sophokles73, line 460: this should work but wouldn't it be more straight-forward, easier to understand and less Spring dependent to simply add the Tracer as a parameter to `HonoConnectionImpl`'s constructor?

calohmn: yes, but I wanted to keep the change for 1.2.x here as small as possible. (In master, this change isn't needed.) 

In master we can then get rid of autowiring the tracer instance.

Issue #1958, opened 2020-05-13
Management API : get all devices for tenant


jbtrystram:
The management API is missing a method to retrieve a list of all the devices for a given tenant. 
The request requires pagination, filters and sorting.

Here are a few ideas on how the API can look like : 
Operation : GET on /v1/devices/{tenantId}
The best practices suggest to pass the parameters in the URL, here are a few examples : 
Filtering : `/v1/devices/{tenantId}?enabled=true,via=4711`
Paging : `/v1/devices/{tenantId}?limit=30,offset=50`
Sorting : `/v1/devices/{tenantId}?status.created=ascending`

It is also possible to define JSON payload and pass the parameters in it, but it seems to be an uncommon practice: 
```
{
	"filters" : {
	    "enabled" : true,
	    "via" : 4711"
	 },
	"page" : {
	    "Limit" : 30,
	    "Offset" : 50
	},
	"sort": {
	    "status.created"="ascending"
	}
}
```

Screenshot of displaying the result :
https://marvelapp.com/5gbji22/screen/68233009 

Default paging values and limits must be set, to avoid dumping large chunks of the database in a single response.

sophokles73:
closed via #2109

bialad:
I'm running Hono v1.4.0, but get 501 Not Implemented when calling this endpoint with:

`http://{{REGISTRY_IP}}:{{REGISTRY_PORT_HTTP}}/v1/devices/{{TENANT}}?enabled=false`

Am I doing something wrong?

bialad:
I'm running Hono v1.4.0, but get 501 Not Implemented when calling this endpoint with:

`http://{{REGISTRY_IP}}:{{REGISTRY_PORT_HTTP}}/v1/devices/{{TENANT}}?enabled=false`

Am I doing something wrong?

sophokles73:
@bialad can you please post your question on stackoverflow as described in our [community guide](https://www.eclipse.org/hono/community/get-in-touch/)? We can then answer your question in a way that others can also benefit from.
Thanks

PR #2009, opened 2020-06-03
[#1958]  implementation of the search devices operation for the file-based registry

jbtrystram:
Follow up of issue #1995. This implementation on the search devices operations only support paging ( and it may not be consistent).
Filtering and sorting should be implemented using databases features.

This is a draft PR to make sure we agree on the shape of things while I write a few integration tests :smiley: 

### DIFF
--- a/services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedDeviceBackend.java
+++ b/services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedDeviceBackend.java
@@ -31,6 +31,9 @@ import org.eclipse.hono.service.management.Result;
 import org.eclipse.hono.service.management.credentials.CommonCredential;
 import org.eclipse.hono.service.management.device.AutoProvisioningEnabledDeviceBackend;
 import org.eclipse.hono.service.management.device.Device;
+import org.eclipse.hono.service.management.device.Filter;
+import org.eclipse.hono.service.management.device.SearchDevicesResult;
+import org.eclipse.hono.service.management.device.Sort;
 import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.CredentialsConstants;
 import org.eclipse.hono.util.CredentialsResult;
@@ -170,6 +173,11 @@ public class FileBasedDeviceBackend implements AutoProvisioningEnabledDeviceBack
     public Future<OperationResult<Id>> updateDevice(final String tenantId, final String deviceId, final Device device,
             final Optional<String> resourceVersion, final Span span) {
         return registrationService.updateDevice(tenantId, deviceId, device, resourceVersion, span);
+    }
+
+    @Override
+    public Future<OperationResult<SearchDevicesResult>> searchDevices(final String tenantId, final int pageSize, final int pageOffset, final List<Filter> filters, final List<Sort> sortOptions, final Span span) {
+    return registrationService.searchDevices(tenantId, pageSize, pageOffset, filters, sortOptions, span);
     }
 
     // CREDENTIALS
--- a/services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
+++ b/services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
@@ -13,6 +13,9 @@
 package org.eclipse.hono.deviceregistry.file;
 
 import java.net.HttpURLConnection;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
@@ -23,6 +26,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import org.eclipse.hono.deviceregistry.service.device.AbstractRegistrationService;
@@ -35,6 +39,10 @@ import org.eclipse.hono.service.management.OperationResult;
 import org.eclipse.hono.service.management.Result;
 import org.eclipse.hono.service.management.device.Device;
 import org.eclipse.hono.service.management.device.DeviceManagementService;
+import org.eclipse.hono.service.management.device.DeviceWithId;
+import org.eclipse.hono.service.management.device.Filter;
+import org.eclipse.hono.service.management.device.SearchDevicesResult;
+import org.eclipse.hono.service.management.device.Sort;
 import org.eclipse.hono.service.registration.RegistrationService;
 import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.RegistrationConstants;
@@ -425,6 +433,10 @@ public class FileBasedRegistrationService extends AbstractRegistrationService
 
         Objects.requireNonNull(tenantId);
         Objects.requireNonNull(deviceId);
+        // If enabled is not set (null) .isEnabled() method defaults to true.
+        // But that does not when filtering using json pointer.
+        // To workaround it we actually sets the property.
+        device.setEnabled(device.isEnabled());
 
         return Future.succeededFuture(processCreateDevice(tenantId, deviceId, device, span));
     }
@@ -517,6 +529,127 @@ public class FileBasedRegistrationService extends AbstractRegistrationService
 
     private ConcurrentMap<String, Versioned<Device>> getDevicesForTenant(final String tenantId) {
         return identities.computeIfAbsent(tenantId, id -> new ConcurrentHashMap<>());
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * The order of devices cannot be guaranteed by the underlying in-memory storage
+     * so the paging may be inconsistent.
+     */
+    @Override
+    public Future<OperationResult<SearchDevicesResult>> searchDevices(final String tenantId, final int pageSize, final int pageOffset,
+                                                                       final List<Filter> filters, final List<Sort> sortOptions, final Span span) {
+
+        final Set<String> allDevicesIds = getDevicesForTenant(tenantId).keySet();
+        List<DeviceWithId> devicesList = new ArrayList<>();
+
+        for (int i = 0; i < allDevicesIds.size(); i++) {
+            final String id = (String) allDevicesIds.toArray()[i];
+            final Device device = getDevicesForTenant(tenantId).get(id).getValue();
+            final DeviceWithId withId = DeviceWithId.from(id, device);
+
+            devicesList.add(withId);
+        }
+
+        final List<Predicate<DeviceWithId>> predicates = new ArrayList<>();
+        for (Filter f : filters) {
+            predicates.add(buildJsonBasedPredicate(f));
+        }
+
+        //Regroup predicates in a single one with AND.
+        final Predicate<DeviceWithId> predicate = predicates.stream()
+                .reduce(x -> true, Predicate::and);
+
+        // 1. apply the filter options to the list
+        devicesList = devicesList
+                .stream()
+                .filter(predicate)
+                .collect(Collectors.toList());
+
+        // 2. sort the selected elements
+        for (Sort s : sortOptions) {
+            devicesList = sortDeviceList(devicesList, s);
+        }
+
+        // 3. limit the results according to the page settings.
+        final List<DeviceWithId> returnDevicesList = new ArrayList<>();
+
+        for (int i = pageOffset * pageSize;
+             i < devicesList.size() && returnDevicesList.size() < pageSize;
+             i++) {
+            returnDevicesList.add(devicesList.get(i));
+        }
+
+        if (returnDevicesList.size() == 0) {
+            return Future.succeededFuture(
+                    OperationResult.empty(HttpURLConnection.HTTP_NOT_FOUND));
+        } else {
+            return Future.succeededFuture(
+                    OperationResult.ok(
+                            HttpURLConnection.HTTP_OK,
+                            new SearchDevicesResult(devicesList.size(), returnDevicesList),
+                            Optional.ofNullable(DeviceRegistryUtils.getCacheDirective(config.getCacheMaxAge())),
+                            Optional.empty()));
+        }
+    }
+
+    private Predicate<DeviceWithId> buildJsonBasedPredicate(final Filter filter) {
+        return deviceWithId -> {
+            final Object field = filter.getField().queryJson(JsonObject.mapFrom(deviceWithId));
+
+            if (field != null) {
+                switch (filter.getOperator()) {
+                    case eq:
+                        return field == filter.getValue();
+                    default:
+                        return true;
+                }
+            } else {
+                return true;
+            }
+        };
+    }
+
+    private List<DeviceWithId> sortDeviceList(final List<DeviceWithId> list, final Sort sort) {
+
+        return list
+                .stream()
+                .sorted(buildJsonBasedComparator(sort))
+                .collect(Collectors.toList());
+    }
+
+    private Comparator<DeviceWithId> buildJsonBasedComparator(final Sort sort) {
+
+        return new Comparator<>() {
+            @Override
+            public int compare(final DeviceWithId d1, final DeviceWithId d2) {
+                final JsonObject device1 = JsonObject.mapFrom(d1);
+                final JsonObject device2 = JsonObject.mapFrom(d2);
+
+                final Object s1 = sort.getField().queryJson(device1);
+                final Object s2 = sort.getField().queryJson(device2);
+
+                if (s1 instanceof String) {
+                    switch (sort.getDirection()) {
+                        case asc:
+                            return ((String) s1).compareTo((String) s2);
+                        default:
+                            return ((String) s2).compareTo((String) s1);
+                    }
+                } else if (s1 instanceof Integer) {
+                    switch (sort.getDirection()) {
+                        case asc:
+                            return ((Integer) s1).compareTo((Integer) s2);
+                        default:
+                            return ((Integer) s2).compareTo((Integer) s1);
+                    }
+                } else {
+                    // cannot compare, consider equals.
+                    return 0;
+                }
+            }
+        };
     }
 
     /**
--- a/services/device-registry-file/src/test/java/org/eclipse/hono/deviceregistry/file/FileBasedDeviceManagementSearchDevicesTest.java
+++ b/services/device-registry-file/src/test/java/org/eclipse/hono/deviceregistry/file/FileBasedDeviceManagementSearchDevicesTest.java
@@ -1,0 +1,132 @@
+/*****************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.hono.deviceregistry.file;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import java.net.HttpURLConnection;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import org.eclipse.hono.deviceregistry.util.DeviceRegistryUtils;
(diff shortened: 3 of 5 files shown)


services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementService.java
```diff
@@ -106,4 +108,28 @@
...
+     * @return A future indicating the outcome of the operation.
+     *         The <em>status code</em> is set as specified in the
+     *         <a href="https://www.eclipse.org/hono/docs/api/management/#/devices/searchDevicesForTenant">
+     *         Device Registry Management API - Search devices for a tenant </a>
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     * @see <a href="https://www.eclipse.org/hono/docs/api/management/#/devices/searchDevicesForTenant">
+     *      Device Registry Management API - Search devices for a tenant</a>
+     */
+    Future<OperationResult<List<DeviceWithId>>> searchDevices(String tenantId, Optional<Integer> limit, Optional<Integer> offset,
+                                                              Optional<Map> sorting, Optional<Map> filters, Span span);
```
kaniyan, line 134: IMHO it is better if the type is also specified for the _Map_ that holds the _filters_ elements(e.g `Optional<Map<String, String>> filters`). Also please check if it is also applicable for _sorting_ .

jbtrystram: I originally wanted to use `Map.Entry` for the sorting parameter but ran into issues at compilation. I'll look into it a bit more

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
```diff
@@ -520,6 +526,102 @@ private JsonObject convertDevice(final String deviceId, final Device payload) {
...
+                devicesList,
+                Optional.ofNullable(DeviceRegistryUtils.getCacheDirective(config.getCacheMaxAge())),
+                Optional.empty());
+    }
+
+    private List<DeviceWithId> filterDeviceList(final List<DeviceWithId> list, final Filter filter) {
+
+        return list
+                .stream()
+                .filter(buildJsonBasedPredicate(filter))
```
dejanb, line 578: I think it would be better to combine multiple filters into a single predicate and apply it only once, rather than iterating collection multiple times. This might not be the topic for this PR, but right now we're supporting only applying all filters. By combining predicated we could support OR functionality, but this might be left for the future.

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
```diff
@@ -520,6 +526,102 @@ private JsonObject convertDevice(final String deviceId, final Device payload) {
...
+                    return true;
+            }
+        };
+    }
+
+    private List<DeviceWithId> sortDeviceList(final List<DeviceWithId> list, final SortingOption sort) {
+
+        return list
+                .stream()
+                .sorted(buildJsonBasedComparator(sort))
```
dejanb, line 601: Same here, if we could build one comparator and then use one stream() to filter and sort that would be great.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementService.java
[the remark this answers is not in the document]
jbtrystram: I did, I created value objects for Sorting Option and filters. You commented an outdated commit, or maybe I missed something ? what do you have in mind ?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementService.java
```diff
@@ -106,4 +108,28 @@
...
+     *          parent for any spans created in this method.
+     * @return A future indicating the outcome of the operation.
+     *         The <em>status code</em> is set as specified in the
+     *         <a href="https://www.eclipse.org/hono/docs/api/management/#/devices/searchDevicesForTenant">
+     *         Device Registry Management API - Search devices for a tenant </a>
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     * @see <a href="https://www.eclipse.org/hono/docs/api/management/#/devices/searchDevicesForTenant">
+     *      Device Registry Management API - Search devices for a tenant</a>
+     */
+    Future<OperationResult<List<DeviceWithId>>> searchDevices(String tenantId, Optional<Integer> limit, Optional<Integer> offset,
```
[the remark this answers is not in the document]
jbtrystram: Very good point !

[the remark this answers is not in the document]
jbtrystram: @kaniyan added in 775da138cfcfb7fc62c245f1adb90fd247d75428

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementService.java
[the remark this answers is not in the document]
kaniyan: My bad! I looked into an outdated commit. Pls ignore the above comment.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementService.java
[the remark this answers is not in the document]
kaniyan: :+1:  IMHO it would be also useful to add some info about it and why the status code 501 is being returned to the JavaDoc.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementService.java
```diff
@@ -129,6 +131,11 @@
      * @see <a href="https://www.eclipse.org/hono/docs/api/management/#/devices/searchDevicesForTenant">
      *      Device Registry Management API - Search devices for a tenant</a>
      */
-    Future<OperationResult<List<DeviceWithId>>> searchDevices(String tenantId, Optional<Integer> limit, Optional<Integer> offset,
-                                                              Optional<List<SortingOption>> sorting, Optional<List<Filter>> filters, Span span);
+    default Future<OperationResult<List<DeviceWithId>>> searchDevices(String tenantId, Optional<Integer> limit, Optional<Integer> offset,
+                                                              Optional<List<SortingOption>> sorting, Optional<List<Filter>> filters, Span span) {
+
```
kaniyan, line 136: How about _null_ check of the arguments as described in the _JavaDoc_ of the `searchDevices` method? 
`@throws NullPointerException if any of the parameters is {@code null}.`

jbtrystram: ah good catch, thanks.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementService.java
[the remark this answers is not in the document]
jbtrystram: Done.

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
jbtrystram: I am not sure how that one is possible.

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
jbtrystram: Done :)

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DelegatingDeviceManagementHttpEndpoint.java
```diff
@@ -101,6 +105,11 @@ public void addRoutes(final Router router) {
         router.get(pathWithTenantAndDeviceId)
                 .handler(this::doGetDevice);
 
+        //GET devices for tenant
+        router.get(pathWithTenant)
```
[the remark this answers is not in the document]
dejanb: @sophokles73 I looked up a little bit on this topic. It seems like it's generally not considered a good practice but there are exceptions like Elastic supporting it (but they also have POST alternative for environments that are challenged by this).

I could see that specifying JSON query as a GET param can be a bit messy? But it still might be a better option than introducing the whole new endpoint for search.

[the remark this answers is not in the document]
sophokles73: I agree that using GET with a query string will be better.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceWithId.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Set the device Id for this device.
+     *
+     * @param deviceId The resource ID for this device.
+     * @return A reference to this for fluent use.
+     */
+    public DeviceWithId setDeviceId(final String deviceId) {
+        this.deviceId = deviceId;
+        return this;
+    }
```
sophokles73, line 65: I do not think that you want to be able to change the identifier after the instance has been created, do you?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceWithId.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Get the device Id for this device.
+     *
+     * @return A string containing the deviceID.
+     */
+    @JsonIgnore
+    public String getDeviceId() {
```
sophokles73, line 73: final?

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
```diff
@@ -520,6 +527,108 @@ private JsonObject convertDevice(final String deviceId, final Device payload) {
...
+            } else {
+              return l;
+            }
+        }).orElse(config.getDefaultPageLimit());
+        final int actualOffset = offset.orElse(0);
+
+        return Future.succeededFuture(doSearchDevices(tenantId, actualLimit, actualOffset, sorting.orElse(Collections.EMPTY_LIST), filters.orElse(Collections.EMPTY_LIST)));
+    }
+
+    private OperationResult<List<DeviceWithId>> doSearchDevices(final String tenantId, final int limit, final int offset,
```
sophokles73, line 552: If I am not mistaken then you are returning `DeviceWithId` instances to the client which include the device identifier. However, the OpenAPI spec defines the returned result as an array of Device objects which do not contain an ID.

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
```diff
@@ -520,6 +527,108 @@ private JsonObject convertDevice(final String deviceId, final Device payload) {
         return identities.computeIfAbsent(tenantId, id -> new ConcurrentHashMap<>());
     }
 
+    /**
+     * Return all devices for a tenant.
+     *
+     * This implementation is for demo purpose, it does not support filtering and sorting.
+     * Moreover, the order of devices cannot be guaranteed by the underlying in-memory storage
+     * so the paging may be inconsistent.
```
sophokles73, line 535: Is this still true?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DelegatingDeviceManagementHttpEndpoint.java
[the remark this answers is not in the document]
jbtrystram: @sophokles73 this was taken care of by @kaniyan in #2113.

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
jbtrystram: @sophokles73 this was taken care of by @kaniyan in #2113.

jbtrystram:
@kaniyan did you bake a way to have the paging default options configurable ?

kaniyan:
> @kaniyan did you bake a way to have the paging default options configurable ?

There is no need to make default page size and offset configurable as the default values are defined in the OpenAPI spec. For more information please follow this discussion [thread](https://github.com/eclipse/hono/pull/2113#discussion_r471262085.)

jbtrystram:
> > @kaniyan did you bake a way to have the paging default options configurable ?
> 
> There is no need to make default page size and offset configurable as the default values are defined in the OpenAPI spec. For more information please follow this discussion [thread](https://github.com/eclipse/hono/pull/2113#discussion_r471262085.)

Ok, thanks !

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedDeviceBackend.java
```diff
@@ -175,9 +175,8 @@ public FileBasedDeviceBackend(
     }
 
     @Override
-    public Future<OperationResult<List<DeviceWithId>>> searchDevices(final String tenantId, final Optional<Integer> limit,
-                                                                     final Optional<Integer> offset, final Optional<List<SortingOption>> sorting, final Optional<List<Filter>> filters, final Span span) {
-        return registrationService.searchDevices(tenantId, limit, offset, sorting, filters, span);
+    public Future<OperationResult<List<DeviceWithId>>> searchDevices(final String tenantId, final int pageSize, final int pageOffset, final List<Filter> filters, final List<Sort> sortOptions, final Span span) {
```
kaniyan, line 178: Could you please rebase again with the master? `searchDevices(...)` doesn't return `List<DeviceWithId>` anymore rather `SearchDevicesResult`.

kaniyan (changes requested):

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedDeviceBackend.java
jbtrystram: Done.

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
```diff
@@ -520,6 +527,120 @@ private JsonObject convertDevice(final String deviceId, final Device payload) {
         return identities.computeIfAbsent(tenantId, id -> new ConcurrentHashMap<>());
     }
 
+    /**
+     * Return all devices for a tenant.
```
kaniyan, line 531: IMHO it doesn't return all devices for a tenant. The JavaDoc from the parent class could be well reused by using {@inheritDoc} and then adding your additional comment regarding in-memory storage and paging inconsistency.

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
```diff
@@ -520,6 +527,120 @@ private JsonObject convertDevice(final String deviceId, final Device payload) {
...
+     *
+     * The order of devices cannot be guaranteed by the underlying in-memory storage
+     * so the paging may be inconsistent.
+     */
+    @Override
+    public Future<OperationResult<SearchDevicesResult>> searchDevices(final String tenantId, final int pageSize, final int pageOffset,
+                                                                       final List<Filter> filters, final List<Sort> sortOptions, final Span span) {
+
+        final Set allDevicesIds = getDevicesForTenant(tenantId).keySet();
```
kaniyan, line 540: How about `Set<String>`?

kaniyan (changes requested):

kaniyan:
@jbtrystram How about enabling unit tests by extending `AbstractDeviceManagementSearchDevicesTest`? And also integration tests by setting property `deviceregistry.supportsSearchDevices` to _true_ for file-based device registry in pom.xml of _tests_ module?

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
jbtrystram: Good catch !

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
```diff
@@ -520,6 +527,120 @@ private JsonObject convertDevice(final String deviceId, final Device payload) {
...
+
+        // 2. sort the selected elements
+        for (Sort s : sortOptions) {
+            devicesList = sortDeviceList(devicesList, s);
+        }
+
+        // 3. limit the results according to the page settings.
+        final List<DeviceWithId> returnDevicesList = new ArrayList<>();
+
+        for (int i = pageOffset; i < devicesList.size() && returnDevicesList.size() < pageSize; i++) {
```
kaniyan, line 574: pageOffset corresponds to pages and not the place holder of the result in resultset. I think here it should be int `i = pageOffset*PageSize+1`. Eg. When the pageSize is set to 10 and pageOffset to 2, the result should start with 21st device in the resultset and not 2nd.

kaniyan (changes requested):

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
```diff
@@ -520,6 +527,120 @@ private JsonObject convertDevice(final String deviceId, final Device payload) {
...
+            returnDevicesList.add(devicesList.get(i));
+        }
+
+        return Future.succeededFuture(
```
kaniyan, line 578: As per the spec, if none of the devices are matched, then 404 should be returned.

kaniyan (changes requested):

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
jbtrystram: IMO the offset should not be tied to the page size, they are different things. Otherwise If I want to ignore the first 5 results for whatever reason and have a page size of 10, how would I do ? 
Writing this makes me realise that I should not have followed your variables names, as `offset` was more suited for this variable.

sophokles73: Yes, the pageOffset contains the number of the page to retrieve. in conjunction with the pageSize (number of entries per page) you can page through the whole result set. If you want to ignore the first entries on a page, then simply retrieve the corresponding page and skip the first entries :-) This is how all APIs that I have stumbled across so far work as well ...

jbtrystram: Ok I aligned to that.

services/device-registry-file/src/test/java/org/eclipse/hono/deviceregistry/file/FileBasedDeviceManagementSearchDevicesTest.java
```diff
@@ -0,0 +1,146 @@
...
+    private FileBasedRegistrationConfigProperties registrationConfig;
+    private FileSystem fileSystem;
+
+    /**
+     * Sets up static fixture.
+     *
+     * @param testContext The test context to use for running asynchronous tests.
+     */
+    @BeforeAll
+    public void setup(final VertxTestContext testContext) {
```
kaniyan, line 71: `testContext` is not being set to complete properly. It is one of the reasons for the tests failure.
This can be removed as it is not being used insde the `setup(...)` method.

kaniyan (changes requested):

services/device-registry-file/src/test/java/org/eclipse/hono/deviceregistry/file/FileBasedDeviceManagementSearchDevicesTest.java
jbtrystram: Thanks for figuring it out !

kaniyan:
@jbtrystram The tests failed. Could you please take a look at the tests?

jbtrystram:
@kaniyan fixed !

kaniyan:
@dejan @sophokles73 The review comments are closed now. Would you mind taking a look and merge it?

jbtrystram:
> @dejan @sophokles73 The review comments are closed now. Would you mind taking a look and merge it?

it's @dejanb :)

sophokles73 (approved):
Thanks @jbtrystram for your patience with this one :+1:

Issue #1960, opened 2020-05-13
Management API : add a status section in device registration data 


jbtrystram:
Proposal : add a status section in the device payload returned by the device registry. 
This section is optional and written only by the device registry implementation. It can not be updated by the user. Any data inserted in the status section by the user in a  PUT request will be ignored.

This section would be a json object at the root level off the device registration data : 
```
{
  "enabled": true,
  "defaults": { … },
  "via": [ … ],
  "viaGroups": [ … ],
  "memberOf": [ … ],
  "ext":  { … },
  "status" : {
          created: $date-time
          last-update: $date-time
          last-user: $username
   }
}

```

It allows to have a overview of recent management activity for the devices. Screenshot : https://marvelapp.com/5gbji22/screen/68233009

kaniyan:
FMPOV this _status_ should not be only added to the _device registration data_ and it should also be available to other device registry resources such as _tenant_ and _credentials_.

ctron:
> FMPOV this _status_ should not be only added to the _device registration data_ and it should also be available to other device registry resources such as _tenant_ and _credentials_.

As the credentials are part of the device, that doesn't make much sense me. For the tenant, I agree.

jbtrystram:
> > FMPOV this _status_ should not be only added to the _device registration data_ and it should also be available to other device registry resources such as _tenant_ and _credentials_.
> 
> As the credentials are part of the device, that doesn't make much sense me. For the tenant, I agree.

Still, you don't get the status section when you retrieve credentials.

fkaltner:
Hi @ctron & @jbtrystram !

I am a bit confused about this issue: the model classes in Hono, i.e. the model for a device (https://github.com/eclipse/hono/blob/master/services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Device.java) already contain that status property. This is not yet implemented as far as I can see which is why I took a look at it, since it would be the right place for two system-internal flags required for #2053 which must not be editable by a user.

So it seems like this issue has been implemented to some extent. What was the reason to introduce that status property without a complete implementation?

Does it makes sense to complete the implementation based on what is currently on master FYPOV?

kaniyan:
> > FMPOV this _status_ should not be only added to the _device registration data_ and it should also be available to other device registry resources such as _tenant_ and _credentials_.
> 
> As the credentials are part of the device, that doesn't make much sense me. For the tenant, I agree.

[...]

PR #2016, opened 2020-06-04
[#1960] add status section for devices in API spec

jbtrystram:

### DIFF
--- a/services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Device.java
+++ b/services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Device.java
@@ -61,6 +61,10 @@ public class Device {
     @JsonFormat(with = JsonFormat.Feature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
     private List<String> memberOf = new LinkedList<>();
 
+    @JsonProperty(RegistryManagementConstants.FIELD_STATUS)
+    @JsonInclude(value = Include.NON_EMPTY)
+    private Status status = new Status();
+
     @JsonProperty(RegistryManagementConstants.FIELD_MAPPER)
     private String mapper;
 
--- a/services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Status.java
+++ b/services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Status.java
@@ -1,0 +1,78 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.service.management.device;
+
+import java.time.Instant;
+
+import org.eclipse.hono.annotation.HonoTimestamp;
+import org.eclipse.hono.util.RegistryManagementConstants;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * A status used in Device value object.
+ */
+public final class Status {
+
+    @JsonProperty(RegistryManagementConstants.FIELD_STATUS_CREATION_DATE)
+    @JsonInclude(value = JsonInclude.Include.NON_EMPTY)
+    @HonoTimestamp
+    private final Instant creationTime;
+
+    @JsonProperty(RegistryManagementConstants.FIELD_STATUS_LAST_UPDATE)
+    @JsonInclude(value = JsonInclude.Include.NON_EMPTY)
+    @HonoTimestamp
+    private Instant lastUpdate;
+
+    @JsonProperty(RegistryManagementConstants.FIELD_STATUS_LAST_USER)
+    @JsonInclude(value = JsonInclude.Include.NON_EMPTY)
+    private String lastUser;
+
+    /**
+     * Create a new Status initialised at the creation time.
+     */
+    public Status() {
+
+        this.creationTime = Instant.now();
+        this.lastUpdate = creationTime;
+    }
+
+    public Instant getCreationTime() {
+        return creationTime;
+    }
+
+    public Instant getLastUpdate() {
+        return lastUpdate;
+    }
+
+
+    /**
+     * Updates the lastUpdate and lastUser fields with the current time and user.
+     *
+     * @param lastUser : the user Id to update with.
+     * @return A reference to this for fluent use.
+     */
+    public Status update(final String lastUser) {
+        this.lastUpdate = Instant.now();
+        this.lastUser = lastUser;
+
+        return this;
+    }
+
+    public String getLastUser() {
+        return lastUser;
+    }
+
+}
--- a/site/documentation/content/api/management/device-registry-v1.yaml
+++ b/site/documentation/content/api/management/device-registry-v1.yaml
@@ -36,7 +36,7 @@ info:
    license:
       name: EPL-2.0
       url: https://www.eclipse.org/legal/epl-2.0/
-   version: 1.2.0
+   version: 1.3.0
 
 externalDocs:
    description: Eclipse Hono™ web page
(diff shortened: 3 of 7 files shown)


site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -778,6 +778,25 @@ components:
...
+
+      Status:
+         type: object
+         additionalProperties: false
+         description: Optional and written only by the device registry implementation.
+            It can not be updated by the user.
+            Any data inserted in the status section by the user in a PUT request will be ignored.
+         properties:
+            "created":
+               type: date-time
```
ctron, line 792: I would recommend to put an explicit expectation of the format here. And put a requirement on the timezone. I would propose to always enforce UTC.

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -778,6 +778,30 @@ components:
                description: The IDs of the gateway groups that this device is a member of. Note that "via" and "memberOf" must not be set at the same time. The same applies for "viaGroups" and "memberOf" which must be set at the same time too. The reason is that Eclipse Hono does not support groups of gateway groups.
             "ext":
                $ref: '#/components/schemas/Extensions'
+            "status":
+               $ref: '#/components/schemas/Status'
```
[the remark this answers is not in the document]
jbtrystram: There are several changes to the API queued at the moment, so I propose to do a version bump when all of the changes are agreed ? there are a couple of other operations I will add in the next few days :) 
I'll do another PR with the version change and we'll merge all all of those at the same time. WDYT ?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -778,6 +778,30 @@ components:
...
+                  The format must comply with RFC 3339, section 5.6. The representation must be UTC (no timezone offset).
+            "updated":
+               type: string
+               format: date-time
+               description: The date and time when the device was last updated.
+                  The format must comply with RFC 3339, section 5.6. The representation must be UTC (no timezone offset).
+            "last-user":
+               type: string
+               description: The user who made the last edit on this device.
+                  The format must comply with RFC 3339, section 5.6. The representation must be UTC (no timezone offset).
```
sophokles73, line 804: I do not think that the user id will match the date/time pattern very oftenly ;-)

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Device.java
```diff
@@ -61,6 +61,10 @@
     @JsonFormat(with = JsonFormat.Feature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
     private List<String> memberOf = new LinkedList<>();
 
+    @JsonProperty(RegistryManagementConstants.FIELD_STATUS)
+    @JsonInclude(value = Include.NON_EMPTY)
+    private Map<String, Object> status = new HashMap<>();
+
```
sophokles73, line 67: My understanding is that the *status* is a generic concept which will most likely also be added to other entities (Tenant, Credentials), right? If so, I wonder if it deserves to be represented by a dedicated class instead of a map. In particular, since the type is defined to allow no additional properties in the API spec ...

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Status.java
```diff
@@ -0,0 +1,75 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * A status used in Device value object.
+ */
+public final class Status {
+
+    @JsonProperty(RegistryManagementConstants.FIELD_STATUS_CREATION_DATE)
+    @JsonInclude(value = JsonInclude.Include.NON_EMPTY)
```
sophokles73, line 30: I guess this should be `JsonInclude.Include.NOT_NULL`. AFAIK `JsonInclude.Include.NON_EMPTY` is used for collections only, isn't it?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Status.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public Instant getLastUpdate() {
+        return lastUpdate;
+    }
+
+
+    /**
+     * Update the lastUpdate field with the current time.
+     */
+    public void update() {
```
sophokles73, line 64: how about passing in the user who performs the update as a parameter and remove the *setLastUser* method as well?
The constructor should then also accept the user parameter ...
WDYT?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Status.java
jbtrystram: Non empty is not limited to collections, it's simply that when serializing the object, if the value is empty, the field will be omitted. With `NOT_NULL` the field won't be omitted as long as it's not null.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Status.java
jbtrystram: that makes a lot of sense ! I updated with your suggestion.

sophokles73:
@jbtrystram any news on this one?

jbtrystram:
@sophokles73 Well I did the update after your last comment but never typed `git push` :sweat_smile: 
If this one seems OK to you we can let it sit there until all the other changes are accepted, then I'll bump the API version so we merge everything at once. WDYT ?

sophokles73:
@jbtrystram can you fix the failing unit tests?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Status.java
```diff
@@ -0,0 +1,78 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
```
sophokles73, line 2: 2020?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Status.java
```diff
@@ -0,0 +1,78 @@
...
+        return creationTime;
+    }
+
+    public Instant getLastUpdate() {
+        return lastUpdate;
+    }
+
+
+    /**
+     * Update the lastUpdate and lastUser fields with the current time and user.
```
sophokles73, line 62: > Updates ...

please add an empty line between the description and the parameters and remove the empty line between the parameters and the return value

site/documentation/content/api/management/device-registry-v1.yaml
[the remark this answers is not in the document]
sophokles73: I'd rather increase the version number with each PR so that we do not run at risk of losing track of it. That way, the PRs are more self contained as well and we can decide on each one of them what type of version change it constitutes ...

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -951,6 +951,29 @@ components:
...
+               description: The date and time when the device was created.
+                  The format must comply with RFC 3339, section 5.6. The representation must be UTC (no timezone offset).
+            "updated":
+               type: string
+               format: date-time
+               description: The date and time when the device was last updated.
+                  The format must comply with RFC 3339, section 5.6. The representation must be UTC (no timezone offset).
+            "last-user":
+               type: string
+               description: The user who made the last edit on this device.
```
sophokles73, line 976: please add the status to the example payloads for the device management endpoint as well ...

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceManagementIT.java
```diff
@@ -468,6 +470,11 @@ private static void assertRegistrationInformation(
             final String expectedDeviceId,
             final Device expectedData) {
 
-        assertThat(response).usingRecursiveComparison().isEqualTo(expectedData);
+        final Comparator<Instant> close = (Instant d1, Instant d2) -> d1.compareTo(d2) < 1000 ? 0 : 1;
+
+        assertThat(response)
+                .usingRecursiveComparison()
+                .withComparatorForFields(close, "status.creationTime", "status.lastUpdate")
+                .isEqualTo(expectedData);
```
sophokles73, line 478: nice :+1:

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -391,6 +391,11 @@ paths:
                                  "via": [
                                     "gw-1", "gw-4"
                                  ],
+                                 "status": {
+                                    "created": "2019-10-03T13:45:16+02:00"
+                                    "updated": "2020-05-04T10:11:12+02:00"
```
sophokles73, line 396: are you sure that these strings follow the definition of the properties? IIRC the times must not contain a time zone offset but be in UTC, right?

jbtrystram: You are correct, i made a quick copypasta there. I've put the Z for UTC.

PR #2005, opened 2020-06-03
decode chirpstack payload to hex

BobClaerhout:
Signed-off-by: Bob Claerhout <claerhout.bob@gmail.com>

PR #2023, opened 2020-06-09
Improve LoraProvider implementations

sophokles73:
Added type checks when accessing message properties.
Changed *getPayload* method to return Buffer of unencoded bytes so that
downstream consumers can more easily process the raw data sent by
devices.

### DIFF
--- a/adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraMessage.java
+++ b/adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraMessage.java
@@ -1,0 +1,60 @@
+/**
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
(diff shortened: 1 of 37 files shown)


sophokles73:
this also renders #2005 obsolete

ctron:
I am not sure I like the type argument and passing back of and forth of the JsonObject.

I think that the `getRequestMessage` method of the lora provider interface should simply return the complete result. Without requiring the `handleProviderRoute` to call into the provider again and again.

This could be done by something like:

~~~java
LoraMessage message = provider.getRequestMessage(ctx.getBody());
if (message instanceof LoraUplinkMessage) {
  LoraUplinkMessage uplinkMessage = (LoraUplinkMessage)message;
  Optional
    .ofNullable(uplinkMessage.extractNormalizedData())
    .ifPresent(data -> ctx.put(LoraConstants.NORMALIZED_PROPERTIES, data));
}
~~~

I think this would put data where it belongs, and not scatter the logic and calls as they are now.

sophokles73:
Good point, @ctron :-) Let me think about that a little ...

PR #2027, opened 2020-06-10
[#1679] Update admin guide for the MongoDB based device registry

kaniyan:


kaniyan:
> 
> 
> @kaniyan FMPOV we should have separate documents for the example registry and the Mongo DB based one so that we do not need to distinguish between them in each and every paragraph. WDYT?

@sophokles73 sounds good. I will create a separate admin guide for the MongoDB based device registry.

kaniyan:
> > @kaniyan FMPOV we should have separate documents for the example registry and the Mongo DB based one so that we do not need to distinguish between them in each and every paragraph. WDYT?
> 
> @sophokles73 sounds good. I will create a separate admin guide for the MongoDB based device registry.

@sophokles73 I have created a separate document for the MongoDB based device registry configuration. Would you mind taking a look again?

site/documentation/content/admin-guide/mongodb-device-registry-config.md
```diff
@@ -0,0 +1,90 @@
...
+| `HONO_CREDENTIALS_SVC_MAX_BCRYPT_ITERATIONS`<br>`--hono.credentials.svc.maxBcryptIterations` | no | `10` | The maximum number of iterations that are supported in password hashes using the BCrypt hash function. This limit is enforced by the device registry when adding or updating corresponding credentials. Increasing this number allows for potentially more secure password hashes to be used. However, the time required to compute the hash increases exponentially with the number of iterations. |
+| `HONO_CREDENTIALS_SVC_MODIFICATION_ENABLED`<br>`--hono.credentials.svc.modificationEnabled` | no | `true` | When set to `false` the credentials contained in the registry cannot be updated nor removed. |
+| `HONO_CREDENTIALS_SVC_RECEIVER_LINK_CREDIT`<br>`--hono.credentials.svc.receiverLinkCredit` | no | `100` | The number of credits to flow to a client connecting to the Credentials endpoint. |
+| `HONO_DEVICE_CONNECTION_SVC_MAX_DEVICES_PER_TENANT`<br>`--hono.deviceConnection.svc.maxDevicesPerTenant` | no | `100` | The number of devices per tenant for which connection related data is stored. It is an error to set this property to a value <= 0. |
```
sophokles73, line 23: this is not a supported property, is it?

site/documentation/content/admin-guide/mongodb-device-registry-config.md
```diff
@@ -0,0 +1,90 @@
...
+| `HONO_MONGODB_CONNECTION_TIMEOUT_IN_MS`<br>`--hono.mongodb.connectionTimeoutInMs` | no | `10000` | The time in milliseconds to attempt a connection before timing out.|
+| `HONO_MONGODB_DB_NAME`<br>`--hono.mongodb.dbName` | no | - | The name of the MongoDB database that should be used by the Device Registry application. |
+| `HONO_MONGODB_HOST`<br>`--hono.mongodb.host` | no | `localhost` | The host name or IP address of the MongoDB instance.|
+| `HONO_MONGODB_PORT`<br>`--hono.mongodb.port` | no | `27017` | The port that the MongoDB instance is listening on.|
+| `HONO_MONGODB_PASSWORD`<br>`--hono.mongodb.password` | no | - | The password to use for authenticating to the MongoDB instance.|
+| `HONO_MONGODB_SERVER_SELECTION_TIMEOUT_IN_MS`<br>`--hono.mongodb.serverSelectionTimeoutInMs` | no | `300000` | The time in milliseconds that the mongo driver will wait to select a server for an operation before raising an error.|
```
sophokles73, line 30: What implication does the default value have? Does this mean that an operation will not fail before 300s have passed? If so, I do not think that this is a reasonable default value.

site/documentation/content/admin-guide/mongodb-device-registry-config.md
```diff
@@ -0,0 +1,90 @@
...
+| `HONO_REGISTRY_SVC_SIGNING_SHARED_SECRET`<br>`--hono.registry.svc.signing.sharedSecret` | no  | - | A string to derive a symmetric key from that is used for signing tokens asserting a device's registration status. The key is derived from the string by using the bytes of the String's UTF8 encoding. When setting the signing key using this variable, other services that need to validate the tokens issued by this service need to be configured with the same key. Alternatively, an asymmetric key pair can be used for signing (and validating) by setting the `HONO_REGISTRY_SVC_SIGNING_KEY_PATH` variable. If none of these variables is set, startup of the server fails. |
+| `HONO_REGISTRY_SVC_SIGNING_TOKEN_EXPIRATION`<br>`--hono.registry.svc.signing.tokenExpiration` | no | `10` | The expiration period to use for the tokens asserting the registration status of devices. |
```
sophokles73, line 62: Are we using any of these parameters? IMHO we are not including any signed token in the get registration assertion response, do we?

site/documentation/content/admin-guide/secure_communication.md
```diff
@@ -49,7 +49,7 @@ The `demo-certs/certs` folder includes the following demo keys and certificates
 ### Device Registry
 
 The Device Registry supports the use of TLS for connections to protocol adapters and the Auth Server.
-Please refer to the [Device Registry admin guide]({{< relref "device-registry-config.md" >}}) for details regarding the required configuration steps.
+Please refer to the [Device Registry admin guide]({{< relref "file-based-device-registry-config.md" >}}) for details regarding the required configuration steps.
```
sophokles73, line 52: please provide a separate section for both the *file based* and the *MongoDB based* registry

site/documentation/content/admin-guide/mongodb-device-registry-config.md
kaniyan: We have this property but the default value is set to -1 which is unlimited. I will update the default value above.

site/documentation/content/admin-guide/mongodb-device-registry-config.md
kaniyan: If it is a stand-alone server, there is normally only one server available and it gets selected. In a shared clusters or replica set,  MongoDB drivers use a server selection algorithm to choose which one to connect to and for that the server selection time out is being used.

My bad, it is a typo in the JavaDoc and for the documentation I referred the JavaDoc. The default value is 30 _seconds_. I am updating it.

site/documentation/content/admin-guide/mongodb-device-registry-config.md
kaniyan: As per my understanding, we are including any signed token. I will remove these parameters.

site/documentation/content/admin-guide/secure_communication.md
kaniyan: @sophokles73 I have now included separate section for both the _file based_ and the _MongoDB based_ registry. Please take a look again.

site/documentation/content/admin-guide/mongodb-device-registry-config.md
```diff
@@ -27,7 +27,7 @@ The following table provides an overview of the configuration variables and corr
 | `HONO_MONGODB_HOST`<br>`--hono.mongodb.host` | no | `localhost` | The host name or IP address of the MongoDB instance.|
 | `HONO_MONGODB_PORT`<br>`--hono.mongodb.port` | no | `27017` | The port that the MongoDB instance is listening on.|
 | `HONO_MONGODB_PASSWORD`<br>`--hono.mongodb.password` | no | - | The password to use for authenticating to the MongoDB instance.|
-| `HONO_MONGODB_SERVER_SELECTION_TIMEOUT_IN_MS`<br>`--hono.mongodb.serverSelectionTimeoutInMs` | no | `300000` | The time in milliseconds that the mongo driver will wait to select a server for an operation before raising an error.|
+| `HONO_MONGODB_SERVER_SELECTION_TIMEOUT_IN_MS`<br>`--hono.mongodb.serverSelectionTimeoutInMs` | no | `30000` | The time in milliseconds that the mongo driver will wait to select a server for an operation before raising an error.|
```
sophokles73, line 30: Do you think that 30 seconds is a reasonable default value? The client's timeout is usually somewhere around 200 - 1000ms so I do not think that it makes a lot of sense to wait for 30 seconds in this case or am I mistaken?

site/documentation/content/admin-guide/mongodb-device-registry-config.md
sophokles73: What is this property used for in the MongoDB based implementation? My understanding is that it doesn't implement the Device Connection API, does it?

site/documentation/content/admin-guide/mongodb-device-registry-config.md
kaniyan: During the implementation, we left it to the Vert.x default values.I also think 30 seconds is quite high. How about setting it to 1000 ms?

site/documentation/content/admin-guide/mongodb-device-registry-config.md
kaniyan: I am mistaken with `HONO_REGISTRY_SVC_MAX_DEVICES_PER_TENANT`. `HONO_DEVICE_CONNECTION_SVC_MAX_DEVICES_PER_TENANT` is not being used and I will remove it.

site/documentation/content/admin-guide/mongodb-device-registry-config.md
sophokles73: Sounds reasonable to me.

kaniyan: I have set it to 1000 ms and pushed the changes.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/config/MongoDbConfigProperties.java
```diff
@@ -182,7 +183,7 @@ public MongoDbConfigProperties setConnectionString(final String connectionString
      * Gets the time in milliseconds that the mongo driver will wait to select a
      * server for an operation before raising an error.
      * <p>
-     * When this property is not set, the Vert.x mongodb client uses a default value of 30000 ms.
+     * When this property is not set, the Vert.x mongodb client uses a default value of 1000 ms.
```
sophokles73, line 186: you can use the `{@value }` tag in order to refer to a constant's value. That way you do not need to explicitly spell it out ...

kaniyan: @sophokles73 Sounds good. The corresponding changes have been made.

PR #2033, opened 2020-06-13
[#2032] Fix error in reconnect delay calculation

### LINKED ISSUE #2032: Delay between HonoConnection reconnect attempts drops to 0
The exponential backoff mechanism concerning `HonoConnection` reconnect attempts has a bug so that after the 58th attempt, the delay between reconnect attempts drops to 0 (that is with the default `reconnectDelayIncrement` of `100`).

With a higher `reconnectDelayIncrement` the issue occurs sooner, with `reconnectDelayIncrement` set `1` the issue doesn't occur.

calohmn:
This fixes #2032.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
@@ -570,26 +570,39 @@ public class HonoConnectionImpl implements HonoConnection {
                 logConnectionError(connectionFailureCause);
             }
             // apply exponential backoff with jitter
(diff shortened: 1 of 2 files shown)


sophokles73:
@calohmn should this also be backported to 1.2.x?

calohmn:
@sophokles73 Yes, I've adopted the commit in 1.2.x (https://github.com/eclipse/hono/commit/2453730e794865eede2b3f48e943f6653a1e7f47, https://github.com/eclipse/hono/commit/66c09d013382c11c742b3e96901e16a7a2ef0451).

PR #2034, opened 2020-06-16
Allow to use empty payloads with the TTN provider

ctron:
The TTN provider allows to send an empty payload message. This is indicated by an explicit `null` value in the `payload_raw` field.

This PR adds support for properly parsing and testing the TTN provider for this type of payload and also corrects the content type for that case for all providers.

### DIFF
--- a/adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/impl/LoraProtocolAdapter.java
+++ b/adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/impl/LoraProtocolAdapter.java
@@ -42,6 +42,7 @@ import org.eclipse.hono.service.http.TracingHandler;
 import org.eclipse.hono.service.http.X509AuthHandler;
 import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.Constants;
+import org.eclipse.hono.util.EventConstants;
 import org.eclipse.hono.util.MessageHelper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -202,16 +203,21 @@ public final class LoraProtocolAdapter extends AbstractVertxBasedHttpProtocolAda
                     final Buffer payload = uplinkMessage.getPayload();
 
                     Optional.ofNullable(uplinkMessage.getNormalizedData())
-                        .ifPresent(data -> ctx.put(LoraConstants.NORMALIZED_PROPERTIES, data));
+                            .ifPresent(data -> ctx.put(LoraConstants.NORMALIZED_PROPERTIES, data));
 
                     Optional.ofNullable(uplinkMessage.getAdditionalData())
-                        .ifPresent(data -> ctx.put(LoraConstants.ADDITIONAL_DATA, data));
+                            .ifPresent(data -> ctx.put(LoraConstants.ADDITIONAL_DATA, data));
 
-                    final String contentType = String.format(
-                            "%s%s%s",
-                            LoraConstants.CONTENT_TYPE_LORA_BASE,
-                            provider.getProviderName(),
-                            LoraConstants.CONTENT_TYPE_LORA_POST_FIX);
+                    final String contentType;
+                    if (payload.length() > 0) {
+                        contentType = String.format(
+                                "%s%s%s",
+                                LoraConstants.CONTENT_TYPE_LORA_BASE,
+                                provider.getProviderName(),
+                                LoraConstants.CONTENT_TYPE_LORA_POST_FIX);
+                    } else {
+                        contentType = EventConstants.CONTENT_TYPE_EMPTY_NOTIFICATION;
+                    }
 
                     uploadTelemetryMessage(ctx, gatewayDevice.getTenantId(), deviceId, payload, contentType);
                     break;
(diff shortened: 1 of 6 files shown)


sophokles73:
@ctron Just out of curiosity: what purpose would such an empty message serve? Is this supposed to be used for a *heart beat*? There is no ttd included so I don't think it is about receiving commands, which works differently in Lora anyway ...

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/impl/LoraProtocolAdapter.java
```diff
@@ -202,16 +203,21 @@ void handleProviderRoute(final RoutingContext ctx, final LoraProvider provider)
...
+
+                    final String contentType;
+                    if (payload.length() > 0) {
+                        contentType = String.format(
+                                "%s%s%s",
+                                LoraConstants.CONTENT_TYPE_LORA_BASE,
+                                provider.getProviderName(),
+                                LoraConstants.CONTENT_TYPE_LORA_POST_FIX);
+                    } else {
+                        contentType = EventConstants.CONTENT_TYPE_EMPTY_NOTIFICATION;
```
sophokles73, line 219: so far we have defined this content type for *events* only. However, the adapter sends all downstream messages using the Telemetry API ...

adapters/lora-vertx/src/test/java/org/eclipse/hono/adapter/lora/providers/LoraTestUtil.java
```diff
@@ -38,14 +41,21 @@ private LoraTestUtil() {
...
-    public static Buffer loadTestFile(final String providerName, final LoraMessageType type) throws IOException, URISyntaxException {
+    public static Buffer loadTestFile(final String providerName, final LoraMessageType type, final String... classifiers) throws IOException, URISyntaxException {
         Objects.requireNonNull(providerName);
         Objects.requireNonNull(type);
-        final URL location = LoraTestUtil.class.getResource(String.format("/payload/%s.%s.json", providerName, type.name().toLowerCase()));
+        final String name = Stream
+                .<String>concat(
+                        Stream.of(providerName, type.name().toLowerCase()),
+                        Arrays.stream(classifiers))
+                .collect(Collectors.joining("."));
```
sophokles73, line 56: nice :+1:

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/impl/LoraProtocolAdapter.java
ctron: True. However, currently there is way for the LoRaWAN adapter to send events. And, given the fact that LoRaWAN is very conservative what is being sent, I would say that transmissions might be more important, and maybe more events rather then telemetry. In any case, at the moment, a LoRaWAN device doesn't have the ability to send an empty payload. As it is a valid operation on the TTN provider side, I think it should be a valid operation in Hono as well.

sophokles73: > However, currently there is way for the LoRaWAN adapter to send events. And, given the fact that LoRaWAN is very conservative what is being sent, I would say that transmissions might be more important, and maybe more events rather then telemetry.

Not sure if I am following. Are you suggesting to let the Lora adapter send events instead of telemetry messages downstream? I am open to the idea. We might also be able to use the ACK flag of uplink messages to determine whether data should be considered an event or telemetry.

> In any case, at the moment, a LoRaWAN device doesn't have the ability to send an empty payload.

You mean because the Lora adapter doesn't support this?

> As it is a valid operation on the TTN provider side, I think it should be a valid operation in Hono as well.

If that is a valid and reasonable use case for Lora devices, I agree.

ctron: > > However, currently there is way for the LoRaWAN adapter to send events. And, given the fact that LoRaWAN is very conservative what is being sent, I would say that transmissions might be more important, and maybe more events rather then telemetry.
> 
> Not sure if I am following. Are you suggesting to let the Lora adapter send events instead of telemetry messages downstream? I am open to the idea. We might also be able to use the ACK flag of uplink messages to determine whether data should be considered an event or telemetry.

I simply think that at them moment, there is no way of choosing Event over Telemetry. Although Event sound like a reasonable feature in the context of LoRaWAN.

> 
> > In any case, at the moment, a LoRaWAN device doesn't have the ability to send an empty payload.
> 
> You mean because the Lora adapter doesn't support this?

Exactly. I wouldn't simply switch, but as you mentioned above, find a nice way to let the user choose. Either through the URL, or a query parameter of the URL, through the `LoraProvider` interface, the device configuration, a part of the LoraMessage, or a mix of these options.

> 
> > As it is a valid operation on the TTN provider side, I think it should be a valid operation in Hono as well.
> 
> If that is a valid and reasonable use case for Lora devices, I agree.

So I think a reasonable approach would be the finish this PR, and then start a new one, adding "Events" into the picture.

ctron: I added #2036 to track the Event discussion.

Issue #1367, opened 2019-07-16
Improve Device Registry API spec with examples, supported values


calohmn:
Looking at the Open API spec for the Device Registry API
https://github.com/eclipse/hono/blob/master/site/documentation/static/api/device-registry-v1.yaml
I think we can improve its usability by adding more information about what kind of values are expected for the various fields.

Example:
For a PUT request on the `/credentials/` resource, the Swagger UI creates such an example value for the request Json:
````
[
  {
    "type": "string",
    "auth-id": "string",
    "enabled": true,
    "ext": {
      "additionalProp1": {}
    },
    "secrets": [
      {
        "enabled": true,
        "not-before": "2019-07-16T10:17:22.500Z",
        "not-after": "2019-07-16T10:17:22.500Z",
        "comment": "string",
        "hash-function": "sha-512",
        "pwd-hash": "string",
        "salt": "string"
      }
    ]
  },
[...]
````
It would be useful here to note that the "type" field in the above Json excerpt must contain `hashed-password` as value and that the `hash-function` field supports the values "sha-256", "sha-512" and "bcrypt".

All in all we could integrate example values and descriptive texts for all fields that require certain values.

This relates to #1324.

sophokles73:
@calohmn can we close this with #2048 ?

calohmn:
@sophokles73 Yes, it looks good now :+1: .

PR #2035, opened 2020-06-16
[#1679] Add user guide and release notes

kaniyan:
This is to be merged after PR #2027

### DIFF
--- a/site/documentation/content/admin-guide/file-based-device-registry-config.md
+++ b/site/documentation/content/admin-guide/file-based-device-registry-config.md
@@ -11,7 +11,7 @@ Protocol adapters use these APIs to determine a device's registration status, e.
 
 There is no particular technical reason to implement these three APIs in one component, so for production scenarios there might be up to three different components each implementing one of the API's.
 
-The Device Registry component also exposes [HTTP based resources]({{< relref "/user-guide/device-registry.md" >}}) for managing tenants and the registration information and credentials of devices.
+The Device Registry component also exposes [HTTP based resources]({{< relref "/api/management" >}}) for managing tenants and the registration information and credentials of devices.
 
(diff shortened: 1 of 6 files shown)


sophokles73:
@kaniyan can you rebase this PR?

kaniyan:
> @kaniyan can you rebase this PR?



> @kaniyan can you rebase this PR?

@sophokles73 I have rebased.

kaniyan:
> I find it rather confusing to have separate admin guides but a single user guide, wouldn't you agree?

@sophokles73 Initially I also thought of two separate user guides. Then I found most part of the two guides (Managing tenants, devices and credentials) remains same. Hence I took this approach. I can also create a separate user guide for the MongoDB device registry and provide link to the file based user guide to refer to the part for managing tenants, devices and credentials. WDYT?

kaniyan:
> @kaniyan Both registry variants implement the Registry Management API, so we might be able to simply refer to the API and amend the API with the example payloads that we are currently providing in the user guide(s). WDYT?

@sophokles73 Sections  namely _Managing tenants, devices and credentials_ in Device Registry user guide are repetitive with the API docs. It is a good idea to refer to the Registry Management API. As you suggested I would create a separate user guide for the MongoDB registry and refer the Management API there in this PR. 

For amending the API with examples we have an issue #1367 open. I would suggest a separate PR for amending the API with more examples.

sophokles73:
@kaniyan sounds good :+1:

kaniyan:
@sophokles73 As we discussed above, I have now created a separate user guide for the MongoDB based device registry. Also I have referred to the Device Registry Management API page for the sections on how to manage tenants, devices and credentials. Please take a look.

PR #2039, opened 2020-06-18
Implement JSON Patch for multiples devices

jbtrystram:
allows to update parts of multiples devices registration data, using the json patch format. 

It allows the device registry implementation to override it, as some more features might be needed (such as transaction operations on database, atomic changes etc...).

I had to add a dependency on eclipse Glassfish, as Vert.x does not support Json Patch.

Here is how it works, assuming devices 4711 exist (but not 4713) : 
```
 curl --insecure -X PATCH -i -H 'Content-Type: application/json-patch' http://localhost:28080/v1/devices/DEFAULT_TENANT -d '{             
    "id" : ["4711", "4713"],
    "patch": [
    {
        "op": "add",
        "path": "/enabled",
        "value": true
    },
    {
        "op": "add",
        "path": "/ext",
        "value": {
            "key": "value"
        }
    }
]
}'
```
And the result would be : 
```
HTTP/1.1 201 Created
content-type: application/json; charset=utf-8
content-length: 234

{"4711":{"status":204,"resource-version":"1d8fe6ab-311e-4f9d-95b3-832bc90e35fa"},"4713":{"status":404,"error-message":"device '4713' cannot be retrieved"}}              
```

This is a draft as I still need to update the doc and add some unit tests. 

Some resources : 
http://jsonpatch.com/
https://json-patch-builder-online.github.io/

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -35,6 +35,7 @@
     <flapdoodle.version>2.2.0</flapdoodle.version>
     <guava.version>28.0-jre</guava.version>
     <caffeine.version>2.8.0</caffeine.version>
+    <glassfish.version>1.1.4</glassfish.version>
     <hamcrest-core.version>2.2</hamcrest-core.version>
     <infinispan.version>10.1.8.Final</infinispan.version>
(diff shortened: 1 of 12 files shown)


dejanb:
Looks good to me so far

PR #2040, opened 2020-06-19
[#2026] Renaming config property for ACK timeout of MQTT adapter.

### LINKED ISSUE #2026: Replace HONO_MQTT_COMMAND_ACK_TIMEOUT
As discussed in https://github.com/eclipse/hono/issues/1632#issuecomment-621053539:
The MQTT adapter property marking the timeout used when sending a command to a device (`HONO_MQTT_COMMAND_ACK_TIMEOUT`) should be brought in line with the corresponding property used for the AMQP adapter (`HONO_AMQP_SEND_MESSAGE_TO_DEVICE_TIMEOUT`).

That means marking `HONO_MQTT_COMMAND_ACK_TIMEOUT` as deprecated and introducing `HONO_MQTT_SEND_MESSAGE_TO_DEVICE_TIMEOUT` to replace it in the future.

For the direction "adapter to qpid" we would have the `SEND_MESSAGE_TIMEOUT` property, and for the other side, "adapter to device", we would commonly have `SEND_MESSAGE_TO_DEVICE_TIMEOUT`. And, as this is somewhat implied by the names, I think we should change the default for `SEND_MESSAGE_TO_DEVICE_TIMEOUT` to be the `SEND_MESSAGE_TIMEOUT` property value (or its default, 1s).

fkaltner:
The new property is called "HONO_MQTT_SEND_MESSAGE_TO_DEVICE_TIMEOUT" which will replace the current "HONO_MQTT_COMMAND_ACK_TIMEOUT" property. This is done in order to provide an abstraction for the configuration of protocols using a bidirectional connection (i.e. MQTT and AMQP).

This fixes #2026


sophokles73:
@calohmn would you mind reviewing this PR?

sophokles73:
You should put the
> This fixes #2026

line *before* the Signed-off-by footer

fkaltner:
@sophokles73 
The line "This fixes #2026" is only in the comment for the PR to make GitHub display the link to the PR in the issue (seems like GitHub ist not able to parse the issue number since it is wrapped in "[...]".).  It is not in the commit message.

Is putting the issue in square brackets something which is a convention in the Hono project? I saw this in other commits which is why I did this. In the Eclipse Guide (https://www.eclipse.org/projects/handbook/#resources-commit) they use this without square brackets, maybe this would make editing the PR comment obsolete.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttProtocolAdapterProperties.java
```diff
@@ -49,13 +66,49 @@ public final int getCommandAckTimeout() {
...
+     * This time out is used by the MQTT adapter for commands published with QoS 1. If there is no acknowledgement
+     * within this time limit, then the command is settled with the <em>released</em> outcome.
+     * <p>
+     * The default value of this property is {@link ClientConfigProperties#getSendMessageTimeout()}.
+     *
+     * @return The time out in milliseconds.
+     */
+    public long getSendMessageToDeviceTimeout() {
+        return sendMessageToDeviceTimeout;
+    }
```
calohmn, line 94: This method doesn't seem to be used. I would have expected it to be used in the place that up til now `getCommandAckTimeout()` was used in (i.e. in `CommandSubscriptionsManager.startTimer`).
For backwards compatibility, the `commandAckTimeout` value should still be used if `sendMessageToDeviceTimeout` isn't set.

But then there is a conflict with using `ClientConfigProperties#getSendMessageTimeout()` as default for `sendMessageToDeviceTimeout`. That means that assigning the default value of `sendMessageToDeviceTimeout` in the constructor would have to be changed (in favor of applying the order of precedence of the 3 possible values in the `getSendMessageToDeviceTimeout` method above).

But, I now think that using the `sendMessageTimeout` value of the _messaging_ configuration properties as default value might be too confusing here after all. It's also the first time that we have properties of different endpoints influence one another. And also from the constructor signature it doesn't become clear, which actual `ClientConfigProperties` are to be used here.

I think, for the time being (until we have an extra set of some default properties), we should just use the `1 second` actual default value of `sendMessageTimeout` as default for `sendMessageToDeviceTimeout`, not using the _messaging_ client properties here. WDYT?

fkaltner: Well, actually I found it confusing when I implemented it but I blamed it on me working on my first issue on Hono.

Just to make sure I got you right: we would remove the reference to `ClientConfigProperties` completely then from the constructor arguments (meaning not only in the MQTT adapter but as well in the AMQP adapter)?

site/documentation/content/admin-guide/mqtt-adapter-config.md
```diff
@@ -27,7 +27,8 @@ The following table provides an overview of the configuration variables and corr
...
 | `HONO_MQTT_CERT_PATH`<br>`--hono.mqtt.certPath` | no | - | The absolute path to the PEM file containing the certificate that the protocol adapter should use for authenticating to clients. This option must be used in conjunction with `HONO_MQTT_KEY_PATH`.<br>Alternatively, the `HONO_MQTT_KEY_STORE_PATH` option can be used to configure a key store containing both the key as well as the certificate. |
-| `HONO_MQTT_COMMAND_ACK_TIMEOUT`<br>`--hono.mqtt.commandAckTimeout` | no | `100` | The amount of time (milliseconds) after which the sending of a command to a device using QoS 1 is considered to be failed. The value of this variable should be increased in cases where devices are connected over a network with high latency. |
+| `HONO_MQTT_COMMAND_ACK_TIMEOUT`<br>`--hono.mqtt.commandAckTimeout` | no | `100` | DEPRECATED! Use HONO_MQTT_SEND_MESSAGE_TO_DEVICE_TIMEOUT instead. The amount of time (milliseconds) after which the sending of a command to a device using QoS 1 is considered to be failed. The value of this variable should be increased in cases where devices are connected over a network with high latency. |
```
calohmn, line 30: `HONO_MQTT_SEND_MESSAGE_TO_DEVICE_TIMEOUT` should be put in backticks. FMPOV, `DEPRECATED!` looks a bit dramatic, `Deprecated, use ...` would do as well.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttProtocolAdapterProperties.java
calohmn: Yes, and in the `KuraAdapterProperties`.

site/documentation/content/admin-guide/kura-adapter-config.md
```diff
@@ -44,6 +44,7 @@ The following table provides an overview of the configuration variables and corr
...
 | `HONO_KURA_TENANT_IDLE_TIMEOUT`<br>`--hono.kura.tenantIdleTimeout` | no | `0ms` | The duration after which the protocol adapter removes local state of the tenant (e.g. open AMQP links) with an amount and a unit, e.g. `2h` for 2 hours. See the [Spring Boot documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-conversion-duration) for an explanation of the format. The value `0ms` disables the timeout. |
+| `HONO_KURA_SEND_MESSAGE_TO_DEVICE_TIMEOUT`<br>`--hono.kura.sendMessageToDeviceTimeout` | no | `100` | The amount of time (milliseconds) after which the sending of a command to a device using QoS 1 is considered to be failed. The value of this variable should be increased in cases where devices are connected over a network with high latency. The default value is set to the value of the `HONO_MESSAGING_SEND_MESSAGE_TIMEOUT` option. |
```
calohmn, line 47: Adopting the value from the `DEFAULT_SEND_MESSAGE_TIMEOUT` constant, the `Default Value` column value would be `1000` then.

site/documentation/content/admin-guide/mqtt-adapter-config.md
calohmn, line 30: Adopting the value from the `DEFAULT_SEND_MESSAGE_TIMEOUT` constant, the `Default Value` column value would be `1000` then.

calohmn:
It would be good if you could also add a release-notes entry (in a `Deprecations` section).

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttProtocolAdapterProperties.java
fkaltner: Okay, I reverted my changes.

site/documentation/content/admin-guide/mqtt-adapter-config.md
fkaltner: Done. Less dramatic now :)

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttProtocolAdapterProperties.java
```diff
@@ -49,13 +58,67 @@ public final int getCommandAckTimeout() {
...
+     * The default value of this property is {@link #DEFAULT_SEND_MESSAGE_TO_DEVICE_TIMEOUT}.
+     *
+     * @return The time out in milliseconds.
+     */
+    public long getSendMessageToDeviceTimeout() {
+        return sendMessageToDeviceTimeout;
+    }
+
+    /**
+     * Sets the waiting for acknowledgement time out in milliseconds for commands published with QoS 1.
```
calohmn, line 89: Use `timeout` as one word instead? Otherwise it looks a bit strange (also in the other places in the javadocs of this class).

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttProtocolAdapterProperties.java
```diff
@@ -49,13 +58,67 @@ public final int getCommandAckTimeout() {
...
+     * This time out is used by the MQTT adapter for commands published with QoS 1. If there is no acknowledgement
+     * within this time limit, then the command is settled with the <em>released</em> outcome.
+     * <p>
+     * The default value of this property is {@link #DEFAULT_SEND_MESSAGE_TO_DEVICE_TIMEOUT}.
+     *
+     * @param sendMessageToDeviceTimeout The time out in milliseconds.
+     * @throws IllegalArgumentException if the timeout is negative.
+     */
+    public void setSendMessageToDeviceTimeout(final long sendMessageToDeviceTimeout) {
+        if (sendMessageToDeviceTimeout < 0) {
```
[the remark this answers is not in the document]
fkaltner: I took a look at `VertxBasedAmqpProtocolAdapter`. Since we do support a value of 0 there and intend to have an abstraction over protocols I think it makes sense to implement the same behavior for MQTT as you suggested 👍

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttProtocolAdapterProperties.java
fkaltner: Yep, I also stumbled upon this.

site/homepage/content/release-notes.md
```diff
@@ -69,6 +69,10 @@ title = "Release Notes"
   because the Device Registration API does not define a corresponding operation.
   Consequently, the C&C functionality of the Kerlink Lora provider which relied on the *get*
   method has been removed.
+ 
+ ### Deprecations
+ 
+ * The configuration property `HONO_MQTT_COMMAND_ACK_TIMEOUT` of the MQTT adapter is now deprecated and planned to be removed in a future release. Use `HONO_MQTT_SEND_MESSAGE_TO_DEVICE_TIMEOUT` instead.
```
calohmn, line 75: Can you remove trailing whitespaces and add a linebreak so that it is aligned with the other lines?

fkaltner: Done.

PR #2043, opened 2020-06-22
Improve description of credentials resources

sophokles73:
I have added more thorough descriptive text to the PUT credentials operation, specifying the behavior to be implemented in a registry.

Also improved code for merging credentials and secrets and added unit
tests.

### DIFF
--- a/services/device-registry-base/src/test/java/org/eclipse/hono/service/management/credentials/CredentialsTest.java
+++ b/services/device-registry-base/src/test/java/org/eclipse/hono/service/management/credentials/CredentialsTest.java
@@ -13,11 +13,15 @@
 package org.eclipse.hono.service.management.credentials;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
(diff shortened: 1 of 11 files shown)


services/device-registry-base/src/test/java/org/eclipse/hono/service/management/credentials/CredentialsTest.java
```diff
@@ -83,12 +86,14 @@ public void testEncodePasswordCredential() {
     @Test
     public void testEncodePskCredential() {
         final PskCredential credential = new PskCredential();
+                credential.setAuthId("psk-device");
+                credential.setSecrets(List.of(new PskSecret().setKey(new byte[] { 0x00, 0x01 })));
 
         final JsonObject json = JsonObject.mapFrom(credential);
+        System.out.println(json);
```
kaniyan, line 93: Is `System.out.println` here by intention?

sophokles73: Nice catch :+1: It was, but I forgot to remove it ...

PR #2044, opened 2020-06-22
Introduce value objects for meta data

sophokles73:
Added value objects to represent common meta data instead of using a
generic map.


adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/Location.java
```diff
@@ -0,0 +1,79 @@
...
+
+package org.eclipse.hono.adapter.lora;
+
+import java.util.Objects;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * A 3D geo-location that consists of longitude, latitude and (optional) altitude.
```
ctron, line 23: I think when it comes to coordinates, there should be a very explicit definition of what the coordinate system is.

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/LoraMetaData.java
```diff
@@ -0,0 +1,243 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * A container for meta information contained in Lora
+ * messages.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public final class LoraMetaData {
```
ctron, line 31: Same here, for every metric there should be a clear and explicit definition of the unit of measurement.

adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/providers/ActilityProvider.java
```diff
@@ -87,145 +86,98 @@ protected Buffer extractPayload(final JsonObject loraMessage) {
...
+     * Gets the frequency corresponding to the channel ID used by Actility/ThingWork
+     * as described in section 2.4 of the
+     * <a href="https://partners.thingpark.com/sites/default/files/2017-11/AdvancedThingParkDeveloperGuide_V4.pdf">
+     * Advanced Developer Guide</a>.
+     *
+     * @param logicalChannelId The channel ID.
+     * @return The frequency in MHz or {@code null} if the identifier is unknown.
+     */
+    private Double getFrequency(final String logicalChannelId) {
+        switch (logicalChannelId) {
```
ctron, line 159: I would say, things like this belong into an enum.

sophokles73: In general, I agree. However, I haven't found a (formal) definition of these channel IDs and I suspect that they are specific to each provider (if used at all). We could define a private enum but is that worth the effort?

ctron: Ok, in that case leaving it as is makes sense.

sophokles73:
@ctron I have added corresponding JavaDocs including units of measure. Would you mind having another look?

PR #2045, opened 2020-06-22
[#1023] Adding metric for rejected connections

### LINKED ISSUE #1023: Add metric for rejected connections
The connection limit added in #931 allows to reject connection requests that exceed a configured limit. Finding the right value for a specific system will probably require some experimenting and tuning. For this purpose as well as for constant monitoring of a productive Hono installation a metric of rejected connection attempts would be helpful.

fkaltner:
In order to be able to fine-tune connection limits it is useful to know how many connections were actually rejected. This is now enabled by introducing the metric "rejectedConnectionsTotal"

This fixes #1023


site/documentation/content/api/Metrics.md
```diff
@@ -75,6 +75,7 @@ Metrics provided by the protocol adapters are:
...
 | *hono.connections.authenticated.duration* | Timer        | *host*, *component-type*, *component-name*, *tenant*                                         | The overall amount of time that authenticated devices have been connected to protocol adapters. <br/> **NB** This metric is only supported by protocol adapters that maintain *connection state* with authenticated devices. In particular, the HTTP adapter does not support this metric. |
+| *hono.connections.rejected.total*  | Counter             | *host*, *component-type*, *component-name*, *tenant*                                         | The number of attempted connections to protocol adapters which were rejected because a configured connection limit was hit. <br/> **NB** This metric is only supported by protocol adapters that maintain *connection state* with authenticated devices. In particular, the HTTP adapter does not support this metric. |
```
sophokles73, line 78: I would rather replace this with a metric for connection attempts in general. We could then use tags to indicate the outcome like *succeeded*, *unauthenticated*, *rejected* (due to resource limitations). In any case, this metric would only make sense as a connection attempt rate, i.e. how many attempts per time-period instead of absolute numbers.

ctron:
I think it is fine as is. But maybe we could add an additional label (not sure that was the right term) to that metric, indicating *why* it was rejected.

site/documentation/content/api/Metrics.md
fkaltner: Since the number of rejected connections is a number which can only increase but never decrease I deemed a Counter to be the appropriate type as the alternatives listed by https://prometheus.io/docs/concepts/metric_types/ didn't really fit.

Isn't the `range` function of Prometheus (https://prometheus.io/docs/prometheus/latest/querying/functions/#rate) which converts this to a rate, providing the metric as you want it to be?

fkaltner: I forgot: i think having this as a more generic metric with a label is a good idea! Is this something you see as part of this issue or should I create a follow-up issue?

fkaltner: I reworked the metric using labels, see commit 3d3ce96.

This doesn't cover all outcomes like you mentioned above, but it's a start and we can continue to add more labels as needed. 

@ctron Would this also cover your comment?

site/documentation/content/api/Metrics.md
```diff
@@ -75,6 +75,7 @@ Metrics provided by the protocol adapters are:
...
 | *hono.connections.authenticated.duration* | Timer        | *host*, *component-type*, *component-name*, *tenant*                                         | The overall amount of time that authenticated devices have been connected to protocol adapters. <br/> **NB** This metric is only supported by protocol adapters that maintain *connection state* with authenticated devices. In particular, the HTTP adapter does not support this metric. |
+| *hono.connections.attempts.total*  | Counter             | *host*, *component-type*, *component-name*, *tenant*                                         | The number of connection attempts to protocol adapters. Can be distinguished by outcome by evaluating the tag "outcome" (and its corresponding "reason"). <br/> **NB** This metric is only supported by protocol adapters that maintain *connection state* with authenticated devices. In particular, the HTTP adapter does not support this metric. |
```
sophokles73, line 78: FMPOV it should also have a tag *status* or *outcome* which indicated whether the attempt succeeded or, if not, why. Possible values of the tag could therefore be *succeeded*, *rejected* and *unauthorized* FMPOV ...

fkaltner: I'm confused :) Possibly my phrasing is misleading, but it states: "Can be distinguished by outcome by evaluating the tag "outcome" (and its corresponding "reason")".

So there's one tag ("outcome") indicating the result (succeeded, rejected, etc.) and another one ("reason") which can be used for detailing the reason for the outcome (e.g. by the value "adapter-resource-limit-exceeded").

sophokles73: Each tag adds another dimension when evaluating meters in the backend. If there is no compelling reason to use two separate tags for outcome and reason I would suggest to stick with a single tag instead. For a successful outcome (connection established) there will never be a reason anyway, so filtering for successful/unsuccessful outcomes should be easy enough, right?

fkaltner: Good point. Since the Prometheus doc themselves warns about it

>CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

(https://prometheus.io/docs/practices/naming/)

I guess it's not worth having two labels for clarity. Changed with 25ce759.

site/documentation/content/api/Metrics.md
```diff
@@ -75,6 +75,7 @@ Metrics provided by the protocol adapters are:
...
 | *hono.connections.authenticated.duration* | Timer        | *host*, *component-type*, *component-name*, *tenant*                                         | The overall amount of time that authenticated devices have been connected to protocol adapters. <br/> **NB** This metric is only supported by protocol adapters that maintain *connection state* with authenticated devices. In particular, the HTTP adapter does not support this metric. |
+| *hono.connections.attempts.total*  | Counter             | *host*, *component-type*, *component-name*, *outcome*                              | The number of connection attempts to protocol adapters. The outcome can be observed by looking the tag "outcome" which contains the result and the reason. <br/> **NB** This metric is only supported by protocol adapters that maintain *connection state* with authenticated devices. In particular, the HTTP adapter does not support this metric. |
```
sophokles73, line 78: can you please replace the quotes around `"outcome"` with asterisks (`*outcome*`)?
It would also be helpful to define the possible values of the outcome tag in the corresponding table above.
FMPOV we just need `accepted`, `unauthorized`, `data-volume-exceeded`, `connection-count-exceeded` and `connection-duration-exceeded` for now ...

fkaltner: Sure, no problem. 

Regarding the possible values: at the moment I only implemented one (see our discussion following your first comment) called "rejected-adapter-connection-limit-exceeded". I can rename this if you like but as far as I know there's also the possibility that a connection may be rejected because of a tenant specific connection limit, so I guess it makes sense to make this clear in the name.

sophokles73: We could start with a single possible failure outcome for the time being. I see your point regarding the different types of limits. So maybe we can start with `adapter-connection-limit-exceeded`. WDYT? 
In any case, we should document the possible value(s) in the table ...

fkaltner: Agreed and adapted.

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -116,6 +126,10 @@ protected MicrometerBasedMetrics(final MeterRegistry registry, final Vertx vertx
             }
         });
         this.unauthenticatedConnections = registry.gauge(METER_CONNECTIONS_UNAUTHENTICATED, new AtomicLong());
+
+        this.rejectedConnections = Counter.builder(METER_CONNECTIONS_ATTEMPTS)
+            .tag(METER_CONNECTION_ATTEMPTS_TAG_KEY_OUTCOME, "adapter-connection-limit-exceeded")
+            .register(registry);
```
sophokles73, line 132: this is only capable of recording a rejected connection attempt. What about the successful attempts?

fkaltner: Right. Since we do not record successful attempts as of now, I went for the most simple solution.

But I took the chance to look at how this was done for the other metrics and adapted the implementation to match the approach of those. It's a bit more lines to read and understand but should be consistent now, see: 
df1620b

site/documentation/content/api/Metrics.md
```diff
@@ -75,6 +75,7 @@ Metrics provided by the protocol adapters are:
...
 | *hono.connections.authenticated.duration* | Timer        | *host*, *component-type*, *component-name*, *tenant*                                         | The overall amount of time that authenticated devices have been connected to protocol adapters. <br/> **NB** This metric is only supported by protocol adapters that maintain *connection state* with authenticated devices. In particular, the HTTP adapter does not support this metric. |
+| *hono.connections.attempts.total*  | Counter             | *host*, *component-type*, *component-name*, *outcome*                                        | The number of connection attempts to protocol adapters. The outcome can be observed by looking the tag *outcome* which contains the result and the reason. At the moment there's only one possible value: `adapter-connection-limit-exceeded`<br/> **NB** This metric is only supported by protocol adapters that maintain *connection state* with authenticated devices. In particular, the HTTP adapter does not support this metric. |
```
sophokles73, line 78: according to the constant definition in `MicrometerBasedMetrics`, the name is `hono.connections.attempts` only (which I find preferable).

fkaltner: Well, actually Prometheus adds the "_total" suffix (see https://prometheus.io/docs/practices/naming/#metric-names).

sophokles73: That might be true, but this spec is about the metrics that we report using the Micrometer framework in the Hono components. Prometheus is one of multiple options for a metrics backend where the meter readings are being collected. We have little (that is: no) influence on what these systems do with the data. However, this specification should define the names that we use in our components only. Does that make sense?

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -60,6 +61,14 @@
...
      */
     public static final String METER_CONNECTIONS_UNAUTHENTICATED = "hono.connections.unauthenticated";
+    /**
+     * The name of the meter for connection attempts. The outcome is signaled by the accordingly named tag.
+     */
+    public static final String METER_CONNECTIONS_ATTEMPTS = "hono.connections.attempts";
+    /**
+     * The key for the meter connection attempt tag holding the outcome of the connection attempt.
+     */
+    public static final String METER_CONNECTION_ATTEMPTS_TAG_KEY_OUTCOME = "outcome";
```
sophokles73, line 71: I believe that this is no longer necessary, is it?

Issue #2020, opened 2020-06-05
Device Registry does not handle malformed payload


sophokles73:
The device management endpoint of the example registry returns a `500` status code when trying to register a device that contains additional (unknown) properties like
```json
{
  "unexpected": "property"
}
```
However, it should return a `400` (Bad Request) in this case.

The tenant management endpoint of the example registry returns a `201` status code when trying to register a tenant that contains additional (unknown) properties like
```json
{
  "unexpected": "property"
}
```
However, it should return a `400` (Bad Request) in this case.

sophokles73:
@dejanb @ctron @calohmn The Device Registry Management API actually defines Tenant, Device and CommonCredentials to not allow additional properties.
The question is, if we want to be *lenient* when reading requests and simply ignore any additional unexpected properties or if this should result in a 400.
I think that we should be (consistently) lenient and simply ignore them. WDYT?

ctron:
I am against that. We explicitly defined "extension" sections in order to store data we do not care about.

Simply allowing "out of spec" data would invalidate the idea of defining types in the first place. We would end up with fields and expectations over time, that don't align anymore.

So if we need a new field, then we must add it. If we don't care, then it belongs into the "ext" section.

sophokles73:
I am not saying that we should accept and store the unknown properties. I am proposing to simply ignore them. Not sure if you were aware of that. But if you were and you are still in favor of rejecting requests that contain unknown properties with a 400 then that is of course also a valid option.

dejanb:
I also think it should be better to return proper error to the user (400), as accepting and ignoring properties could lead them to the false confidence that the payload format is alright and lead to other issues later on (if they want to use that ignored properties). Fail fast approach is better IMHO.

sophokles73:
@ctron @dejanb That's fine with me as well. Then let's not be lenient and return a 400 in this case. I will prepare a CQ ...

PR #2046, opened 2020-06-22
[#2020] Reject unknown properties in Tenant Management HTTP endpoint

sophokles73:
Added explicit error handling in case of failure to process requests.
Removed obsolete request payload parsing/modification.
Improved integration tests.

This is the first of 3 PRs (one for each endpoint).


service-base/src/main/java/org/eclipse/hono/service/http/AbstractHttpEndpoint.java
```diff
@@ -201,6 +205,58 @@ protected final String getDeviceIdParam(final RoutingContext ctx) {
...
+     *         <li>failed with a {@link ClientErrorException} with status 400 if a predicate has been
+     *         given and the predicate evaluates to {@code false}, or</li>
+     *         <li>otherwise be completed with the parameter value.</li>
+     *         </ul>
+     * @throws NullPointerException If ctx or paramName are {@code null}.
+     */
+    protected final Future<Optional<String>> getRequestParameter(
+            final RoutingContext ctx,
+            final String paramName,
+            final Pattern validator,
```
dejanb, line 233: What if we make validator a function? Then you can pass any logic to verify your parameter value, like pattern matching, optional, etc. It would make 'optional' field obsolete as well.

sophokles73: That might work if the return type is a `Future<String>` so that we can handle the cause for failures properly. Let me think about it ...

sophokles73:
@dejanb I have pushed a commit which goes in the direction we discussed. Would you mind taking another look?

dejanb:
@sophokles73 Looks good

PR #2051, opened 2020-06-24
#1856 negative authenticated connection count

### LINKED ISSUE #1856: AMQP metrics report negative number of authenticated connections
Periodically(I would say rather sporadically !) AMQP adapter report a negative metric number for authenticated connections. Unfortunately, for a while I tried to find the cause and reproduce this but without success. But anyway, when I took a look at  [MicrometerBasedMetrics.java](https://github.com/eclipse/hono/blob/master/service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java), I would like to propose to have more _safe_ incremental and detrimental operations for connections counters, something like this: 

```
private void tryToDecrement(AtomicLong longValue) {
   longValue.getAndUpdate(oldValue -> oldValue > 0 ? oldValue - 1 : 0);
}
```

```
private void tryToIncrement(AtomicLong longValue)
 {
  longValue.getAndUpdate(oldValue -> {
   if (oldValue == Long.MAX_VALUE) {
    throw new IllegalStateException("Long value overflow!");
   }
   return oldValue + 1L;
  });
 }
}
```

fkaltner:
This fixes #1856

### DIFF
--- a/adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
+++ b/adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
@@ -21,6 +21,7 @@ import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.argThat;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.inOrder;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
@@ -88,6 +89,7 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.ArgumentCaptor;
+import org.mockito.InOrder;
 
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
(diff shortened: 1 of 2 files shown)


adapters/amqp-vertx/src/test/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapterTest.java
```diff
@@ -1189,10 +1191,17 @@ public void testConnectionFailsIfAdapterLevelConnectionLimitIsExceeded() {
...
+        // THEN the adapter closes the connection right after it opened it
+        final ArgumentCaptor<Handler<AsyncResult<ProtonConnection>>> closeHandler = ArgumentCaptor.forClass(Handler.class);
+        verify(deviceConnection).closeHandler(closeHandler.capture());
+        closeHandler.getValue().handle(Future.succeededFuture());
         final ArgumentCaptor<ErrorCondition> errorConditionCaptor = ArgumentCaptor.forClass(ErrorCondition.class);
         verify(deviceConnection).setCondition(errorConditionCaptor.capture());
         assertEquals(AmqpError.UNAUTHORIZED_ACCESS, errorConditionCaptor.getValue().getCondition());
+        // AND increments and decrements the connection count accordingly
+        final InOrder metricsInOrderVerifier = inOrder(metrics);
+        metricsInOrderVerifier.verify(metrics).incrementConnections(TEST_TENANT_ID);
```
sophokles73, line 1203: FMPOV this check should be done immediately after the openHandler has been invoked

fkaltner: Okay for me. By this we'll make sure that the increment is actually made in the open handler and not accidentally called in the closed handler together with the decrement (would be very weird - but you never now... )

sophokles73: exactly

PR #2057, opened 2020-06-25
[#2030] Handle closed MQTT connection during CONNECT processing

### LINKED ISSUE #2030: IllegalStateException: MQTT endpoint is closed
We saw the following exception in an MQTT adapter log:
````
ERROR io.vertx.core.impl.ContextImpl Unhandled exception
java.lang.IllegalStateException: MQTT endpoint is closed
	at io.vertx.mqtt.impl.MqttEndpointImpl.checkClosed(MqttEndpointImpl.java:724)
	at io.vertx.mqtt.impl.MqttEndpointImpl.close(MqttEndpointImpl.java:674)
	at io.vertx.mqtt.impl.MqttEndpointImpl.connack(MqttEndpointImpl.java:311)
	at io.vertx.mqtt.impl.MqttEndpointImpl.reject(MqttEndpointImpl.java:343)
	at io.vertx.mqtt.impl.MqttEndpointImpl.reject(MqttEndpointImpl.java:50)
	at org.eclipse.hono.adapter.mqtt.AbstractVertxBasedMqttProtocolAdapter.handleConnectionRequestResult(AbstractVertxBasedMqttProtocolAdapter.java:454)
	at org.eclipse.hono.adapter.mqtt.AbstractVertxBasedMqttProtocolAdapter.lambda$handleEndpointConnection$12(AbstractVertxBasedMqttProtocolAdapter.java:395)
	at io.vertx.core.impl.FutureImpl.tryFail(FutureImpl.java:197)
	at io.vertx.core.impl.FutureImpl.fail(FutureImpl.java:127)
[...]
	at io.vertx.core.impl.FutureImpl.tryComplete(FutureImpl.java:152)
	at io.vertx.core.impl.FutureImpl.complete(FutureImpl.java:113)
	at org.eclipse.hono.service.resourcelimits.PrometheusBasedResourceLimitChecks.lambda$executeQuery$24(PrometheusBasedResourceLimitChecks.java:494)
	at io.vertx.ext.web.client.impl.HttpContext.handleDispatchResponse(HttpContext.java:308)
	at io.vertx.ext.web.client.impl.HttpContext.execute(HttpContext.java:295)
	at io.vertx.ext.web.client.impl.HttpContext.next(HttpContext.java:270)
	at io.vertx.ext.web.client.impl.predicate.PredicateInterceptor.handle(PredicateInterceptor.java:69)
	at io.vertx.ext.web.client.impl.predicate.PredicateInterceptor.handle(PredicateInterceptor.java:32)
	at io.vertx.ext.web.client.impl.HttpContext.next(HttpContext.java:267)
	at io.vertx.ext.web.client.impl.HttpContext.fire(HttpContext.java:277)
	at io.vertx.ext.web.client.impl.HttpContext.dispatchResponse(HttpContext.java:238)
	at io.vertx.ext.web.client.impl.HttpContext.lambda$null$2(HttpContext.java:367)
	at io.vertx.core.impl.ContextImpl.executeTask(ContextImpl.java:369)
[...]
````

There is no `CONNECT` tracing span in the OpenTracing component for the corresponding MQTT connection attempt, i.e. the exception causes the span not to get finished.

calohmn:
This fixes #2030.

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.OptionalInt;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
 import org.apache.qpid.proton.message.Message;
@@ -386,10 +387,17 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends MqttProtoc
         if (endpoint.will() != null) {
             span.log("ignoring client's last will");
         }
+        final AtomicBoolean connectionClosedPrematurely = new AtomicBoolean(false);
+        // set a preliminary close handler, will be overwritten at the end of handleConnectionRequest()
+        endpoint.closeHandler(v -> {
+            log.debug("client closed connection before CONNACK got sent to client [client-id: {}]", endpoint.clientIdentifier());
+            TracingHelper.logError(span, "client closed connection");
+            connectionClosedPrematurely.set(true);
+        });
 
         isConnected()
                 .compose(v -> handleConnectionRequest(endpoint, span))
-                .onComplete(result -> handleConnectionRequestResult(endpoint, span, result));
+                .onComplete(result -> handleConnectionRequestResult(endpoint, span, result, connectionClosedPrematurely));
 
     }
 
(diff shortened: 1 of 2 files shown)


sophokles73:
wouldn't it be enough to simply check `MqttEndpoint.isConnected()` before sending the CONNACK?

ctron:
> wouldn't it be enough to simply check `MqttEndpoint.isConnected()` before sending the CONNACK?

I just wanted to ask the same question. Taking a look at the code the internal state goes to "closed" before calling into the close handler.

calohmn:
> wouldn't it be enough to simply check `MqttEndpoint.isConnected()` before sending the CONNACK?

In `MqttEndpointImpl`, the `isConnected()` method uses the `isConnected` field which is only set to `true` in the `connack()` method. That means calling `isConnected()` on a not-yet-accepted connection will always return `false`, hence it can't be used in this place where the connection is just about to get accepted/rejected.

> Taking a look at the code the internal state goes to "closed" before calling into the close handler.

Yes, but I don't see how that is relevant here or what you are aiming at. Can you elaborate?

ctron:
> > wouldn't it be enough to simply check `MqttEndpoint.isConnected()` before sending the CONNACK?
> 
> In `MqttEndpointImpl`, the `isConnected()` method uses the `isConnected` field which is only set to `true` in the `connack()` method. That means calling `isConnected()` on a not-yet-accepted connection will always return `false`, hence it can't be used in this place where the connection is just about to get accepted/rejected.

That is unfortunate. But taking a closer look at the way you track the state, I think there is a race condition: The close handler can be called from the context of the MQTT endpoint. However, the `isConnected()` method from the protocol adapter returns a future that may (or may not) be part of the MQTT context. It probably isn't. So it is running in parallel to the other code.

So after you check with `connectionClosedPrematurely.get()` the connection might still close. And you have the situation as if the boolean variable wouldn't exist.

I think that a better approach would be to actually run into the exception, and then only log it, if the boolean variable is set. Of course you need to do this inside the vertx context of the MQTT endpoint. Otherwise the fact that the internal state is set before the close handler is called (see below) will create another race condition.

> 
> > Taking a look at the code the internal state goes to "closed" before calling into the close handler.
> 
> Yes, but I don't see how that is relevant here or what you are aiming at. Can you elaborate?

The internal state will be "closed" before you have chance to react to that in your close handler.

calohmn:
> So after you check with `connectionClosedPrematurely.get()` the connection might still close. And you have the situation as if the boolean variable wouldn't exist.

Yes, that is a case I've already taken into account (it could happen e.g. if `sendConnectedEvent` takes quite long). The `rejectConnectionRequest()` method with the `try-catch` block takes care of that.

> I think that a better approach would be to actually run into the exception, and then only log it, if the boolean variable is set. 

The main intention of using the `connectionClosedPrematurely.get()` check first, was to skip calling `sendConnectedEvent` if the client has already disconnected.

> Of course you need to do this inside the vertx context of the MQTT endpoint. Otherwise the fact that the internal state is set before the close handler is called (see below) will create another race condition.
> [..]
> The internal state will be "closed" before you have chance to react to that in your close handler.

I don't really see a race condition issue here, since all that is done in the preliminary close handler is creating log output (apart from setting the `connectionClosedPrematurely`value, which is non-essential (see above)). 
Since it all involves the same MQTT connection/endpoint, it looks to me that this is done in the same vertx event loop thread anyway.

ctron:
> > Of course you need to do this inside the vertx context of the MQTT endpoint. Otherwise the fact that the internal state is set before the close handler is called (see below) will create another race condition.
> > [..]
> > The internal state will be "closed" before you have chance to react to that in your close handler.
> 
> I don't really see a race condition issue here, since all that is done in the preliminary close handler is creating log output (apart from setting the `connectionClosedPrematurely`value, which is non-essential (see above)).
> Since it all involves the same MQTT connection/endpoint, it looks to me that this is done in the same vertx event loop thread anyway.

The close handler yes, but I don't think that `handleConnectionRequestResult` is being executed in the context of the MqttEndpoint. So checking the value of the boolean will occur in parallel to updating the boolean.

calohmn:
> The close handler yes, but I don't think that `handleConnectionRequestResult` is being executed in the context of the MqttEndpoint. So checking the value of the boolean will occur in parallel to updating the boolean.

I've done some testing which showed that `handleConnectionRequestResult` is indeed executed in the same vertx event loop thread in which also the close handler gets run. And it makes sense: 
The `handleEndpointConnection` processing stays in the context of the same verticle (i.e. the `AbstractVertxBasedMqttProtocolAdapter` instance) that handles the requests for this MqttEndpoint, even when this involves handling a response from e.g. tenant (or registration) service. The used tenant/registration client instance is also associated with that verticle instance.

And even _if_ close handler and `handleConnectionRequestResult` would be processed concurrently by 2 threads and `handleConnectionRequestResult` would be run _after_ `MqttEndpointImpl.closed` got set to `true`but _before_ the close handler was called (and the atomic boolean got updated), I still see no real problem. For such a scenario there would be the try-catch block in the `rejectConnectionRequest()` method.

ctron:
> > The close handler yes, but I don't think that `handleConnectionRequestResult` is being executed in the context of the MqttEndpoint. So checking the value of the boolean will occur in parallel to updating the boolean.
> 
> I've done some testing which showed that `handleConnectionRequestResult` is indeed executed in the same vertx event loop thread in which also the close handler gets run. And it makes sense:
> The `handleEndpointConnection` processing stays in the context of the same verticle (i.e. the `AbstractVertxBasedMqttProtocolAdapter` instance) that handles the requests for this MqttEndpoint, even when this involves handling a response from e.g. tenant (or registration) service. The used tenant/registration client instance is also associated with that verticle instance.

You are hoping that this behavior doesn't change in the future. As soon as any of these chained calls steps out of the vertx context, that no longer works.

> 
> And even _if_ close handler and `handleConnectionRequestResult` would be processed concurrently by 2 threads and `handleConnectionRequestResult` would be run _after_ `MqttEndpointImpl.closed` got set to `true`but _before_ the close handler was called (and the atomic boolean got updated), I still see no real problem. For such a scenario there would be the try-catch block in the `rejectConnectionRequest()` method.

I think the original problem you wanted to fix is the exception in the log. For that case, there still is a possible race condition. However, you could double check, when catching the exception, if you are disconnected, and then suppress logging of the exception. So you would a fast path, not hitting the exception, and when you hit the exception, you could still suppress it.

calohmn:
@ctron I've now added a check in the `catch (final IllegalStateException ex)` block to improve log output.

ctron:
> @ctron I've now added a check in the `catch (final IllegalStateException ex)` block to improve log output.

I am not sure I understand the benefit of checking the exception message string, rather then checking again for the boolean variable?

calohmn:
The `connectionClosedPrematurely` value won't have been set in that case because it is only set in the _preliminary_ close handler, set in `handleEndpointConnection`. As the last step in `handleConnectionRequest` this close handler gets replaced with the proper close handler (see comment in line 391).

Of course checking the exception message string is not pretty here, but the alternative of using reflection to check the `MqttEndpointImpl.isClosed` field value isn't either. And storing our own additional `isClosed` flag along with the endpoint (setting that flag inside the _proper_ close handler) would also require quite some effort.
As this distinction here is only used for log/tracing output and will occur rather rarely (in contrast to the scenario handled via the  `connectionClosedPrematurely` value), I think the current solution is a fair tradeoff.

Issue #2049, opened 2020-06-24
Setting device request qos-level header to messaging endpoint AMQP messages property header


DerSchwilk:
Taking the qos-level header of a devices request, supported for telemetry messages, and setting it to the forwarded AMQP message as AMQP property header, would allow QoS-handling for, on messaging endpoint, connected systems.

Currently this is benetfitial for the Eclipse Hono <- Eclipse Ditto communication. A Ditto feature is currently in development, which has the aim to allow end-to-end acknowledgements from device sending a message via Hono to the users solution connected to Ditto:
[https://github.com/eclipse/ditto/issues/661](https://github.com/eclipse/ditto/issues/661)

Since Ditto uses the Qpid JMS client which has no functionality to differentiate between presettled and not-settled messages via AMQP1.0 protocol specification, an AMQP property header for that differentiation is necessary.
We would propose naming this property header the same as for incoming telemetry messages "qos-level".

sophokles73:
Just for clarification: even if the JMS client would allow you to determine if an incoming message was (pre-)settled or unsettled, that would not help at all because Hono sends **all** messages unsettled, regardless of the QoS level used by the device. We do this because it is much easier for the Dispatch Router to properly handle flow control for unsettled messages.
When a device uses QoS 0 for a telemetry message, the protocol adapter sends an unsettled AMQP 1.0 message downstream but simply does not wait for the disposition from the downstream consumer. The protocol adapter simply acknowledges the uploaded data to the device but doesn't care if the downstream consumer accepts or rejects/releases the message.
So, in order to allow a downstream consumer to determine the QoS level used by a device for uploading telemetry data, we *have to* include such a property explicitly.

sophokles73:
@fkaltner this might be a suitable issue for you to work on as well. Otherwise, I can do it as well. WDYT?

fkaltner:
Yes, I actually thought also about taking over, so feel free to assign this to me.

DerSchwilk:
Setting this property also to event messages (QoS 1 at all times, since QoS=0 not available for events), would allow us to implement both channels in one connection-source, while defaulting no qos-level property header to QoS=0 in our header mapping.

Will this header be included on both channels?/ Can you include it on both channels?

fkaltner:
[...]

PR #2059, opened 2020-06-29
[#2049] Adding AMQP application property "qos" for northbound consumers

fkaltner:
I started to implement this and it became a bigger change than I initially thought. Hence I am opening this PR to ask for feedback / guidance 😃 :

1. In my opinion the right place for the QoS property are the subclasses of `ExecutionContext`. However the HTTP adapter didn't have one so far (although I found such a class in the code base). Hence I decided to adapt the HTTP adapter to make the adapter implementations more consistent. This is the main reason why the change came out a lot bigger than I thought.

2. Currently there are 2 QoS enums. I am tempted to refactor this as well, but the change will be bigger then. Hence I thought I will add this in a separate commit (if we agree on it) to make this a bit easier to digest.

3. The QoS enum I introduced also has a `NOT_APPLICABLE` (used for TTD and Notification events) and a `UNKNOWN` level (if a client sends a QoS level we do not know). Do you think we need those cases? What should we do in those cases? Omit the property? Also the validation of the property is not consistent (if I didn't miss something obvious): in the HTTP adapter a request with an unsupported QoS level will be rejected, in the MQTT adapter not as far as I can see.

### DIFF
--- a/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapContext.java
+++ b/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapContext.java
@@ -306,5 +306,4 @@ public final class CoapContext extends MapBasedExecutionContext {
                 })
                 .orElse(null);
     }
-
 }
--- a/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapContext.java
+++ b/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapContext.java
@@ -25,6 +25,7 @@ import org.eclipse.californium.core.server.resources.CoapExchange;
 import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.EventConstants;
 import org.eclipse.hono.util.MapBasedExecutionContext;
+import org.eclipse.hono.util.QoS;
 
 import io.micrometer.core.instrument.Timer.Sample;
 import io.vertx.core.Vertx;
@@ -306,4 +307,10 @@ public final class CoapContext extends MapBasedExecutionContext {
                 })
                 .orElse(null);
     }
+
+    @Override
+    public QoS getRequestedQos() {
+        return isConfirmable() ? QoS.AT_LEAST_ONCE : QoS.AT_MOST_ONCE;
+    }
+
 }
--- a/core/src/main/java/org/eclipse/hono/util/QoS.java
+++ b/core/src/main/java/org/eclipse/hono/util/QoS.java
@@ -1,0 +1,27 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.util;
+
+/**
+ * Denotes the QoS level (in Hono's terms) with which a message was sent by the device.
+ */
+public enum QoS {
+
+    AT_MOST_ONCE,
+    AT_LEAST_ONCE,
+    /**
+     * Indicates that the device set a QoS level which is not known or supported.
+     */
+    UNKNOWN;
+}
--- a/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
@@ -60,6 +60,7 @@ import org.eclipse.hono.util.CredentialsConstants;
 import org.eclipse.hono.util.DeviceConnectionConstants;
 import org.eclipse.hono.util.EventConstants;
 import org.eclipse.hono.util.MessageHelper;
+import org.eclipse.hono.util.QoS;
 import org.eclipse.hono.util.RegistrationConstants;
 import org.eclipse.hono.util.ResourceIdentifier;
 import org.eclipse.hono.util.Strings;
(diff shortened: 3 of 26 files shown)


adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapContext.java
```diff
@@ -307,4 +308,13 @@ private Integer getIntegerQueryParameter(final String parameterName) {
                 .orElse(null);
     }
 
+    /**
+     * Get the QoS level as set in the request by the device.
+     *
+     * @return The QoS level requested by the device.
+     */
+    public QoS getRequestedQos() {
```
sophokles73, line 316: IMHO we should add this method to `ExecutionContext` because it is a general requirement for all adapters/protocols.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapContext.java
```diff
@@ -307,4 +308,13 @@ private Integer getIntegerQueryParameter(final String parameterName) {
                 .orElse(null);
     }
 
+    /**
+     * Get the QoS level as set in the request by the device.
```
sophokles73, line 312: Can you please always use present tense active form? In this case
> Gets the quality of service level corresponding to the delivery semantics specified by the device in its request to upload data.

This applies to most (all) of your JavaDoc comments ...

fkaltner:
> I very much like the idea to consistently use an `ExecutionContext` in all protocol adapters when processing device requests. It would be nice if you could create a separate commit for the corresponding changes and then use another commit for adding the QoS level property to the `ExecutionContext` and using it in the downstream message.

I reworked my changes and force pushed only the changes related to introducing a HTTP context, see  32c09f2

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapContext.java
fkaltner: Will do.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapContext.java
sophokles73: Are you working on this? I can't find a corresponding change in your latest push?

fkaltner: Yes, I do! I will push a separate commit containing the QoS changes soon.

fkaltner: There you go, see: 9bb5266

Let me know what you think about the two QoS classes and the `NOT_APPLICABLE` and the `UNKNOWN` levels

core/src/main/java/org/eclipse/hono/util/QoS.java
```diff
@@ -0,0 +1,51 @@
...
+     * Indicates the QoS is not settable by the device.
+     */
+    NOT_APPLICABLE,
+    /**
+     * Indicates that the device set a QoS level which is not known or supported.
+     */
+    UNKNOWN;
+
+    /**
+     * Derive the QoS level from the given number.
```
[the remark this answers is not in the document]
fkaltner: Makes sense, yes this is used for MQTT only.

core/src/main/java/org/eclipse/hono/util/QoS.java
```diff
@@ -0,0 +1,51 @@
...
+     * @return The QoS level derived from the given number.
+     */
+    public static QoS from(final int number) {
+        switch (number) {
+            case 0:
+                return AT_MOST_ONCE;
+            case 1:
+                return AT_LEAST_ONCE;
+            case 2:
+                return NOT_APPLICABLE;
```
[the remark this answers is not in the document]
fkaltner: Yes, from the ordinal of the enum that's right. 

Let's ignore the ordinal for now: the reason why I introduced this was that I stumbled upon `sendTtdEvent()` in `AbstractProtocolAdapterBase` and `sendNotificationEvent` in `AbstractMessageSenderConnectionEventProducer`.

Are TTD and connection events something which will arrive at the northbound application? Should the `qos` application property set for those events as well?

service-base/src/main/java/org/eclipse/hono/service/http/HttpContext.java
```diff
@@ -73,4 +89,137 @@ public void setTracingContext(final SpanContext spanContext) {
     public SpanContext getTracingContext() {
         return spanContext;
     }
+
+    @Override
+    public QoS getRequestedQos() {
```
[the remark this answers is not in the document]
fkaltner: Based on today's Community call: @ctron mentioned that this will cause problem with Quarkus. I see your point (I guess: "Design for inheritance, or prohibit it") but wasn't sure anymore after the discussion.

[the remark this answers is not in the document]
sophokles73: Let's wait and see what happens with the Quarkus story ... in the meantime I suggest to follow good design practice here :-)

core/src/main/java/org/eclipse/hono/util/QoS.java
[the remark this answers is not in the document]
sophokles73: Yes, both are supposed to be consumed by downstream applications. AFAIC they should thus also contain the qos property.

core/src/main/java/org/eclipse/hono/util/QoS.java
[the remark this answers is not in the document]
fkaltner: After a closer look it turned out that this method was in fact used in the HTTP adapter. The HTTP adapter accidentally (I guess) shares its QoS level definition with MQTT.

I think it makes more sense to have this in the corresponding Context implementations, since it is protocol specific.

core/src/main/java/org/eclipse/hono/util/QoS.java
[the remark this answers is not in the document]
fkaltner: Alright, set their QoS to AT_LEAST_ONCE, since they use `EventSenderImpl` which send AMQP messages with `ProtonQoS.AT_LEAST_ONCE`.

service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java
```diff
@@ -1171,9 +1172,10 @@ private boolean isGatewaySupportedForDevice(final JsonObject registrationAsserti
      * Subclasses are encouraged to use this method for creating {@code Message} instances to be sent downstream in
      * order to have required Hono specific properties being set on the message automatically.
      * <p>
-     * This method simply delegates to {@link #newMessage(ResourceIdentifier, String, String, Buffer, TenantObject,
+     * This method simply delegates to {@link #newMessage(QoS, ResourceIdentifier, String, String, Buffer, TenantObject,
      * JsonObject, Integer, Duration)}.
      *
+     * @param qos The QoS level with which the device sent the message to the protocol adapter.
```
sophokles73, line 1178: same here, document `null` value behavior

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapContext.java
fkaltner: Since this is the only conversation left unresolved: is there something I missed or are you still reviewing?

sophokles73:
@fkaltner can you please squash the last three commits into one so that we end up with two commits in the PR?

PR #2061, opened 2020-07-01
Do not enforce values to be returned from remote cache

sophokles73:
The Hotrod client does no longer enforce (previous) values being
returned from the remote cache in the put and removeWithVersion methods
as in the former case the returned value isn't used anyway and in the
latter case the flag should not have any impact at all.

### DIFF
--- a/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCache.java
+++ b/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCache.java
@@ -28,7 +28,6 @@ import org.eclipse.hono.client.DisconnectListener;
 import org.eclipse.hono.client.ReconnectListener;
 import org.eclipse.hono.client.ServerErrorException;
 import org.eclipse.hono.util.Futures;
-import org.infinispan.client.hotrod.Flag;
 import org.infinispan.client.hotrod.RemoteCacheContainer;
 import org.infinispan.commons.api.BasicCache;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


ctron:
If we do not use the values, then we should change the method signatures to indicate that we do not return the value.

sophokles73:
@ctron I agree. However, this PR is for 1.2.x where I would rather keep the method signatures as they are. On master, however, we should change the signatures but that should go into a separate PR, right?

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCache.java
```diff
@@ -227,11 +218,7 @@ public void addReconnectListener(final ReconnectListener<HotrodCache<K, V>> list
 
         } else {
 
-            return Futures.create(() -> {
-                return cache
-                    .withFlags(Flag.FORCE_RETURN_VALUE)
-                    .removeWithVersionAsync(key, version);
-            });
+            return Futures.create(() -> cache.removeWithVersionAsync(key, version));
```
calohmn, line 221: When using that method in `removeCommandHandlingAdapterInstance()`, the return value is actually needed.

That means when removing the `FORCE_RETURN_VALUE` flag here, the `removeCommandHandlingAdapterInstance()` method also needs to be changed. I think we can do that here in 1.2.x - the information, whether removal succeeded or not is only logged here anyway.

sophokles73: You are suggesting that the `FORCE_RETURN_VALUE` flag determines whether we always get back a boolean as the result of *removeWithVersionAsync*, right? The only documentation I have found on this is https://infinispan.org/docs/9.4.x/user_guide/user_guide.html#return_values_2 which doesn't mention the *removeWithVersionAsync* method so I assume that the flag has no impact on that method. Or have you conducted any particular tests that would suggest otherwise?

calohmn: I mixed it up with the `removeAsync` method here. So, you're right, I think the flag can be removed here without it affecting the return value.

PR #2065, opened 2020-07-03
Fix content type of the LoRaWAN adapter

ctron:
This changes the content type of the telemetry messages sent by the LoRaWAN adapter to drop the `+json` suffix.

The payload wasn't JSON before the refactoring, but also currently it is not JSON. So that is a bug IMHO. And it causes problems with JMS clients.

In the past this wasn't a big deal, because the payload contained an actual string (base 64 encoded payload). So the message could be decoded as a JMS text message.

Now the payload is a BLOB/byte array and decoding as a String will not always work. If the decoding as a string doesn't work, JMS libraries fail to decode the message into a JMS text message, and you cannot process them.

Dropping the `+json` suffix seems to be the right cause of action IMHO, as the payload never did contain a JSON structure.

### DIFF
--- a/adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/impl/LoraProtocolAdapter.java
+++ b/adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/impl/LoraProtocolAdapter.java
@@ -213,10 +213,9 @@ public final class LoraProtocolAdapter extends AbstractVertxBasedHttpProtocolAda
                     final String contentType;
                     if (payload.length() > 0) {
                         contentType = String.format(
-                                "%s%s%s",
+                                "%s%s",
                                 LoraConstants.CONTENT_TYPE_LORA_BASE,
-                                provider.getProviderName(),
-                                LoraConstants.CONTENT_TYPE_LORA_POST_FIX);
+                                provider.getProviderName());
                     } else {
                         contentType = EventConstants.CONTENT_TYPE_EMPTY_NOTIFICATION;
                     }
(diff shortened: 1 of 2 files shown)


adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/impl/LoraProtocolAdapter.java
```diff
@@ -213,10 +213,9 @@ void handleProviderRoute(final RoutingContext ctx, final LoraProvider provider)
                     final String contentType;
                     if (payload.length() > 0) {
                         contentType = String.format(
-                                "%s%s%s",
+                                "%s%s",
                                 LoraConstants.CONTENT_TYPE_LORA_BASE,
-                                provider.getProviderName(),
-                                LoraConstants.CONTENT_TYPE_LORA_POST_FIX);
+                                provider.getProviderName());
```
sophokles73, line 218: does this content-type make a JMS client to interpret the message as a ByteMessage?

ctron: Yes.

ctron: https://github.com/apache/qpid-jms/blob/f9ab180136c790a53b775c91b24e9339aca8deec/qpid-jms-client/src/main/java/org/apache/qpid/jms/util/ContentTypeSupport.java#L101-L107

PR #2066, opened 2020-07-03
Fix return codes of Device Connection service implementations

sophokles73:
The status codes returned by the "Remove command-handling protocol
adapter instance for device" operation did not match the ones specified
by the API.

The API defines 404 to be the status code to be returned in case of
non-existing/non-matching entries and 412 being allowed in case of
non-matching entries only.


calohmn:
@sophokles73 Revisiting this now:
This change doesn't seem to reflect the most recent API definition as introduced in #1940 (one month before this PR).

To me it looks like this PR was made based on the [Hono 1.2 API definition](https://www.eclipse.org/hono/docs/1.2/api/device-connection/#remove-command-handling-protocol-adapter-instance-for-device). Hono 1.2 was the stable Hono version at the time this PR was created and therefore was the default version displayed on the website.
However, the change of the API definition towards returning 412 (#1940) was only displayed when selecting the "dev" version on the Hono website. My guess is, that when creating this PR (and when reviewing it from my part) not the most recent API docs were taken into account.

I have therefore created #2154 to revert this change here.

Issue #2070, opened 2020-07-06
Support Hono deployment in multi-tenant messaging environments


dejanb:
In multi-tenant messaging environments there could be different connection endpoints used by infrastructure and end users. For end users in this case, the addresses are usually translated internally in order to enable consistent namespaces.

For example, dispatch router enables this behaviour through the (poorly) documented `multiTenant` property

```
  ["listener", {
    "host": "0.0.0.0",
    "port": 5672,
    "linkCapacity": 100,
    "authenticatePeer": true,
    "saslMechanisms": "PLAIN",
    "saslPlugin": "Hono Auth",
    "multiTenant": true
  }]
```

When clients connect through a listener like this, all addresses in the attach frames will be prefixed with the virtual host prefix. e.g. `telemetry/DEFAULT_TENANT` would become `test-vhost/telemetry/DEFAULT_TENANT` internally. End clients are be unaware of this. There are more work being done in the router to make this more configurable in the future (mosly to support Kubernetes namespaces), but the basic principles will remain the same.

In order to be able to work in such environments, infrastructure components connecting through the non multi-tenant listeners and are responsible of doing address translation on thier own. The proposal is to add an option for protocol adapters to rewrite addresses when opening links to the AMQP network.
The idea is to provide a configuration option similar to popular web server URL rewrite rules.

The rule configuration would be in the `"$PATTERN $SUBSTITUTION"` format. The `$PATTERN` should match the original address. It uses the regular Java pattern syntax. There are two helper constants introduced:
	* `ENDPOINT` - that will match the current endpoint being used</li>
    * `ID` - a pattern that will match any identity (using default format) in the system

Example configuration could look like this:

```
  messaging:
    name: Hono adapter-http-vertx
    amqpHostname: hono-internal
    host: eclipse-hono-dispatch-router
    port: 5673
    keyPath: /etc/hono/key.pem
    certPath: /etc/hono/cert.pem
    trustStorePath: /etc/hono/trusted-certs.pem
    hostnameVerificationRequired: false
    addressRewriteRule: (${ENDPOINT})/(${ID}) test-vhost/$1/$2
  command:
    name: Hono adapter-http-vertx
    amqpHostname: hono-internal
    host: eclipse-hono-dispatch-router
    port: 5673
    keyPath: /etc/hono/key.pem
    certPath: /etc/hono/cert.pem
    trustStorePath: /etc/hono/trusted-certs.pem
    hostnameVerificationRequired: false
[...]

PR #2071, opened 2020-07-06
[#2070] Implement downstream address rewrite feature

dejanb:
Implements support for #2070

There's still documentation missing, but I'd withhold that until we have agreed upon the implementation.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientFactoryImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientFactoryImpl.java
@@ -28,7 +28,10 @@ import org.eclipse.hono.client.device.amqp.internal.AmqpAdapterClientCommandCons
 import org.eclipse.hono.client.device.amqp.internal.AmqpAdapterClientCommandResponseSender;
 import org.eclipse.hono.client.device.amqp.internal.AmqpAdapterClientEventSenderImpl;
 import org.eclipse.hono.client.device.amqp.internal.AmqpAdapterClientTelemetrySenderImpl;
+import org.eclipse.hono.util.AddressHelper;
 import org.eclipse.hono.util.CommandConstants;
+import org.eclipse.hono.util.EventConstants;
+import org.eclipse.hono.util.TelemetryConstants;
 
 import io.vertx.core.Future;
 
@@ -80,7 +83,7 @@ public final class AmqpAdapterClientFactoryImpl extends AbstractHonoClientFactor
     @Override
     public Future<TelemetrySender> getOrCreateTelemetrySender() {
 
-        final String cacheKey = TelemetrySenderImpl.getTargetAddress(tenantId, null);
+        final String cacheKey = AddressHelper.getTargetAddress(TelemetryConstants.TELEMETRY_ENDPOINT, tenantId, null, null);
         return connection.isConnected(getDefaultConnectionCheckTimeout())
                 .compose(v -> connection.executeOnContext(result -> {
                     telemetrySenderClientFactory.getOrCreateClient(
@@ -98,7 +101,7 @@ public final class AmqpAdapterClientFactoryImpl extends AbstractHonoClientFactor
     @Override
     public Future<EventSender> getOrCreateEventSender() {
 
-        final String cacheKey = EventSenderImpl.getTargetAddress(tenantId, null);
+        final String cacheKey = AddressHelper.getTargetAddress(EventConstants.EVENT_ENDPOINT, tenantId, null, null);
         return connection.isConnected(getDefaultConnectionCheckTimeout())
                 .compose(v -> connection.executeOnContext(result -> {
                     eventSenderClientFactory.getOrCreateClient(
--- a/client/src/main/java/org/eclipse/hono/client/impl/TelemetrySenderImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/TelemetrySenderImpl.java
@@ -25,6 +25,7 @@ import org.eclipse.hono.client.ServerErrorException;
 import org.eclipse.hono.client.ServiceInvocationException;
 import org.eclipse.hono.config.ClientConfigProperties;
 import org.eclipse.hono.tracing.TracingHelper;
+import org.eclipse.hono.util.AddressHelper;
 import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.TelemetryConstants;
 
@@ -59,24 +60,6 @@ public class TelemetrySenderImpl extends AbstractDownstreamSender {
         super(con, sender, tenantId, targetAddress);
     }
 
-    /**
-     * Gets the AMQP <em>target</em> address to use for uploading data to Hono's telemetry endpoint.
-     *
-     * @param tenantId The tenant to upload data for.
-     * @param deviceId The device to upload data for. If {@code null}, the target address can be used
-     *                 to upload data for arbitrary devices belonging to the tenant.
-     * @return The target address.
-     * @throws NullPointerException if tenant is {@code null}.
-     */
-    public static String getTargetAddress(final String tenantId, final String deviceId) {
-        final StringBuilder targetAddress = new StringBuilder(TelemetryConstants.TELEMETRY_ENDPOINT)
-                .append("/").append(Objects.requireNonNull(tenantId));
-        if (deviceId != null && deviceId.length() > 0) {
-            targetAddress.append("/").append(deviceId);
-        }
-        return targetAddress.toString();
-    }
-
     @Override
     public String getEndpoint() {
         return TelemetryConstants.TELEMETRY_ENDPOINT;
@@ -84,7 +67,7 @@ public class TelemetrySenderImpl extends AbstractDownstreamSender {
 
     @Override
     protected String getTo(final String deviceId) {
-        return getTargetAddress(tenantId, deviceId);
+        return AddressHelper.getTargetAddress(TelemetryConstants.TELEMETRY_ENDPOINT, tenantId, deviceId, null);
     }
 
     /**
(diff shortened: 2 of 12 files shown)


core/src/main/java/org/eclipse/hono/util/AddressHelper.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.util.Objects;
+import java.util.regex.PatternSyntaxException;
+
+/**
+ * Utility class for handling AMQP addresses.
+ */
+public class AddressHelper {
+
+    static final String ID_REGEX = "[a-zA-Z0-9-_\\.]+";
```
sophokles73, line 23: is this supposed to match the regex that is configured in the registry for tenant and device IDs? If so, I believe that we will need to find a way to keep them in sync, right?

core/src/test/java/org/eclipse/hono/util/AddressHelperTest.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Tests address rewrite rules.
+ */
+public class AddressHelperTest {
+
+    /**
+     * Tests address rewrite rules.
+     */
+    @Test
+    public void testAddressRewrite() {
```
sophokles73, line 28: maybe split this up into distinct test cases?

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -518,4 +519,24 @@ public final void setIdleTimeout(final int idleTimeoutMillis) {
...
+        return addressRewriteRule;
+    }
+
+    /**
+     * Sets the rewrite rule for downstream addresses.
+     * See {@link org.eclipse.hono.util.AddressHelper#rewrite(String, String, String)} for more information about syntax and behavior of this property.
+     *
+     * @param addressRewriteRule The rewrite rule to be applied to the address.
+     */
+    public void setAddressRewriteRule(final String addressRewriteRule) {
```
sophokles73, line 539: final ?

client/src/main/java/org/eclipse/hono/client/impl/TelemetrySenderImpl.java
```diff
@@ -59,32 +60,14 @@ protected TelemetrySenderImpl(
...
-
     @Override
     public String getEndpoint() {
         return TelemetryConstants.TELEMETRY_ENDPOINT;
     }
 
     @Override
     protected String getTo(final String deviceId) {
-        return getTargetAddress(tenantId, deviceId);
+        return AddressHelper.getTargetAddress(TelemetryConstants.TELEMETRY_ENDPOINT, tenantId, deviceId, connection.getConfig().getAddressRewriteRule());
```
sophokles73, line 70: how about keeping the *getTargetAddress* method and just change its implementation to delegate to `AddressHelper`. This would mean much fewer changes in the other code, wouldn't it?

client/src/main/java/org/eclipse/hono/client/impl/AmqpAdapterClientFactoryImpl.java
```diff
@@ -80,7 +83,7 @@ protected void onDisconnect() {
     @Override
     public Future<TelemetrySender> getOrCreateTelemetrySender() {
 
-        final String cacheKey = TelemetrySenderImpl.getTargetAddress(tenantId, null);
+        final String cacheKey = AddressHelper.getTargetAddress(TelemetryConstants.TELEMETRY_ENDPOINT, tenantId, null, connection.getConfig().getAddressRewriteRule());
```
sophokles73, line 86: the `AmqpAdapterClientFactory` is supposed to be used by devices and gateways to connect to the AMQP adapter. IMHO we do not need to rewrite addresses in that case, do we?

core/src/main/java/org/eclipse/hono/util/AddressHelper.java
ctron, line 23: I think it makes sense in this case to pre-compile the regular expression using `Pattern`.

client/src/main/java/org/eclipse/hono/client/impl/TelemetrySenderImpl.java
dejanb: @sophokles73 It would still require changes everywhere, as we need to add/pass additional argument to this method. I was having it like that in the beginning but noted that the implementation was not consistent among classes, so I moved the logic to the helper method call. At that point I didn't see any value in these intermediary methods, so I removed them.

I can reintroduce them back if you want, I don't have a strong opinion on that.

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -285,15 +285,20 @@
      */
     public static final String FIELD_TRACING_SAMPLING_MODE_PER_AUTH_ID = "sampling-mode-per-auth-id";
 
+    /**
+     * The default regular expression validating IDs contain only legal characters.
+     */
+    public static final String DEFAULT_ID_REGEX = "[a-zA-Z0-9-_\\.]+";
```
dejanb, line 291: @sophokles73 it is indeed the same regex, just without end delimiters. I moved it here and based other ones on it. Let me know if you have any other ideas on how improve on this.

core/src/main/java/org/eclipse/hono/util/AddressHelper.java
dejanb: @ctron It's not a pattern at this point, but a substitute that can be used when writing rules, so you can write `telemetry/${ID}` instead of `telemetry/[a-zA-Z0-9-_\\.]+`. It's also moved now to the other place per Kai's suggestion.

core/src/main/java/org/eclipse/hono/util/AddressHelper.java
```diff
@@ -0,0 +1,96 @@
...
+            return address.replaceAll(regexSubstitues(endpoint, regex), replacement);
+        } catch (PatternSyntaxException pe) {
+            return address;
+        }
+    }
+
+    static String regexSubstitues(final String endpoint, final String regex) {
+        return regex
+                .replaceAll("\\$\\{ENDPOINT\\}", endpoint)
+                .replaceAll("\\$\\{ID\\}", RegistryManagementConstants.DEFAULT_ID_REGEX);
```
sophokles73, line 92: what if the tenant ID and/or device ID regex are being configured explicitly, i.e. do not have their default value?

client/src/main/java/org/eclipse/hono/client/impl/TelemetrySenderImpl.java
sophokles73: No worries, I see your point ...

core/src/main/java/org/eclipse/hono/util/AddressHelper.java
dejanb: It won't work. But it's just a simple helper, so it's not something that you must use to make it work. I just thought some these would make configuration look simpler for common use-cases.

We can do more in this area and do substitutions based on configuration, environment variables, etc. But I didn't want to go there until we have the basics sorted.

If this is confusing we can remove substitutions for the time being and rethink them later on.

sophokles73: I guess it will be very hard to trace and understand the source of the problem in this case. So FMPOV we should either omit the substitution logic for the time being or integrate it properly with the configurable tenant and device ID patterns ...

dejanb: OK, I'll remove it completely for now.

dejanb:
@sophokles73 great. Yeah, I'll do that next.

core/src/main/java/org/eclipse/hono/util/AddressHelper.java
```diff
@@ -0,0 +1,82 @@
...
+            return address;
+        }
+
+        final String[] elements = rewriteRule.split(" ", 2);
+        if (elements.length != 2) {
+            return address;
+        }
+
+        try {
+            return address.replaceAll(elements[0], elements[1]);
```
ctron, line 76: I still think we should use pre-compiled patterns for this.

dejanb: It makes sense if we initialize AddressHelper and reuse that Pattern instead of using it statically. I'll take a look at that.

dejanb:
@sophokles73 @ctron Can you take another look?

* I moved pattern compilation to the config class, so that it will be compiled only once.
* Hence the rewrite method now have the whole config as a parameter, which I think will further help if we want to implement variable substitutions in the future. 
* I removed rewriting from a couple of places to make sure `message.setAddress()` uses the original address.
* Added docs

Let me know if you see anything else missing.

sophokles73:
@dejanb The integration tests fail due to JavaDoc issues ...

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -518,4 +525,54 @@ public final void setIdleTimeout(final int idleTimeoutMillis) {
...
+            }
+        }
+    }
+
+    /**
+     * Gets precompiled address rewrite pattern.
+     *
+     * @return The precompiled address rewrite pattern.
+     */
+    public Pattern getAddressRewritePattern() {
```
sophokles73, line 566: final?

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -518,4 +525,54 @@ public final void setIdleTimeout(final int idleTimeoutMillis) {
...
+        return addressRewritePattern;
+    }
+
+    /**
+     * Gets address rewrite replacement.
+     *
+     * @return The address rewrite replacement.
+     */
+    public String getAddressRewriteReplacement() {
```
sophokles73, line 575: final?

dejanb:
@sophokles73 not sure how it passed locally. Fixed now, thanks for spotting it.

sophokles73:
@dejanb sorry, but thes still fail due to JavaDoc issues ...

Issue #2075, opened 2020-07-08
AMQP adapter should check adapter connection limit for unauthenticated devices


sophokles73:
The AMQP adapter currently only checks the adapter instance's maximum number of connections for authenticated devices. However, connections from unauthenticated devices should be considered as well.

PR #2076, opened 2020-07-08
[#2075] Check adapter connection limit for unauthenticated devices

sophokles73:
The AMQP adapter now always checks the adapter's connection limit
regardless of whether the device is authenticated or not.

### DIFF
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
@@ -377,6 +377,36 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
             TracingHelper.setDeviceTags(span, authenticatedDevice.getTenantId(), authenticatedDevice.getDeviceId());
         }
 
+        checkConnectionLimitForAdapter()
+            .onFailure(ex -> {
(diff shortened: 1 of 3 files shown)


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -377,6 +377,36 @@ private void processRemoteOpen(final ProtonConnection con) {
             TracingHelper.setDeviceTags(span, authenticatedDevice.getTenantId(), authenticatedDevice.getDeviceId());
         }
 
+        checkConnectionLimitForAdapter()
```
calohmn, line 380: My first intuition here would have been to keep the order of the checks and do the (usually less expensive) tenant and device checks first. But, thinking more about it, I agree that doing the connection limit check first seems more intuitive and logical. In any case, that is also how it is done in the MQTT adapter.

sophokles73: actually, the adapter connection limit check is the cheapest one of all as it doesn't require any remote service invocation at all but simply checks an integer that is managed within the Metrics implementation class ...

calohmn: right, I confused it with the message limit check.

PR #2077, opened 2020-07-08
Remove obsolete ComponentScan

calohmn:
The package `org.eclipse.hono.service.credentials` can't be resolved here because it got moved from `service-base` to `device-registry-base` (which isn't used as a dependency here). Furthermore, the corresponding `Application` class from the MQTT adapter (as an adapter using the same base classes) also doesn't have this ComponentScan.


sophokles73:
It is unclear to me why this had been in there in the first place. Which component did we hope to inject by scanning that package? Any idea?

calohmn:
No idea. The corresponding line in the MQTT adapter application has been removed more than 1,5 years ago (https://github.com/eclipse/hono/commit/d5b88dc20b6f60fc8bc023cd4af0fa3875aeb08a).

PR #2078, opened 2020-07-08
[#1824] Improving message limit accuracy

### LINKED ISSUE #1824: Update data volume resource limits cache with each message
Current data volume (in bytes) sent by a tenant is queried from Prometheus and stored locally till cache time out expiration. But, why we are not updating this locally with each  incoming message(i mean payload size calculated based on _minimum message size_ ) after when the cache will be expired this value will be overridden with data from Prometheus. 
In case when this data will be dynamically updated the tolerance rate of messages that exceeded data limits will be very low.

fkaltner:
Hono's default implementation for resource limits, which is based on Prometheus,caches the queried metrics until they hit a configurable expiry time. The limit is only exceeded when the Prometheus metrics reports so, which may take until the cache finally expires.

This is now improved by taking "local" values into account, i.e. the cache is updated without touching expiry of its entries. When it expires it will be synchronized with Prometheus again.

This fixes #1824

### DIFF
--- a/core/src/main/java/org/eclipse/hono/cache/ExpiringValueCache.java
+++ b/core/src/main/java/org/eclipse/hono/cache/ExpiringValueCache.java
@@ -60,9 +60,11 @@ public interface ExpiringValueCache<K, V> {
      * @param key The key under which the value is stored.
      * @param value The value to store.
      *
+     * @return true, if the value was updated, false if not (in that case it was expired)
+     *
      * @throws NullPointerException if any of the parameters is {@code null}.
      */
-    void put(K key, V value);
+    boolean update(K key, V value);
 
     /**
      * Gets a value from the cache.
--- a/core/src/main/java/org/eclipse/hono/cache/ExpiringValueCache.java
+++ b/core/src/main/java/org/eclipse/hono/cache/ExpiringValueCache.java
@@ -53,20 +53,6 @@ public interface ExpiringValueCache<K, V> {
     void put(K key, V value, Duration maxAge);
 
     /**
-     * Puts a value to the cache but does not change the expiration time of the present value.
-     * <p>
-     * Any previous value for the key will be replaced with the new one.
-     *
-     * @param key The key under which the value is stored.
-     * @param value The value to store.
-     *
-     * @return true, if the value was updated, false if not (in that case it was expired)
-     *
-     * @throws NullPointerException if any of the parameters is {@code null}.
-     */
-    boolean update(K key, V value);
-
-    /**
      * Gets a value from the cache.
      *
      * @param key The key to get the value for.
--- a/core/src/main/java/org/eclipse/hono/cache/ExpiringValueCache.java
+++ b/core/src/main/java/org/eclipse/hono/cache/ExpiringValueCache.java
@@ -53,6 +53,18 @@ public interface ExpiringValueCache<K, V> {
     void put(K key, V value, Duration maxAge);
 
     /**
+     * Puts a value to the cache but does not change the expiration time of the present value.
+     * <p>
+     * Any previous value for the key will be replaced with the new one.
+     *
+     * @param key The key under which the value is stored.
+     * @param value The value to store.
+     *
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     */
+    void put(K key, V value);
+
+    /**
      * Gets a value from the cache.
      *
      * @param key The key to get the value for.
--- a/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
@@ -25,6 +25,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 
+import org.eclipse.hono.service.metric.MetricsCache;
 import org.eclipse.hono.service.metric.MetricsTags;
 import org.eclipse.hono.service.metric.MicrometerBasedMetrics;
 import org.eclipse.hono.tracing.TracingHelper;
@@ -59,7 +60,7 @@ import io.vertx.ext.web.codec.BodyCodec;
  * Resource limit checks which compare configured limits with live metrics retrieved
  * from a <em>Prometheus</em> server.
  */
-public final class PrometheusBasedResourceLimitChecks implements ResourceLimitChecks {
+public final class PrometheusBasedResourceLimitChecks implements ResourceLimitChecks, MetricsCache {
 
     private static final Logger LOG = LoggerFactory.getLogger(PrometheusBasedResourceLimitChecks.class);
 
--- a/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
@@ -330,6 +330,29 @@ public final class PrometheusBasedResourceLimitChecks implements ResourceLimitCh
                 });
     }
 
+    @Override
+    public Future<Void> addMessageBytes(final String tenant, final long payloadSize, final MetricsTags.ProcessingOutcome outcome) {
+        final Promise<Void> result = Promise.promise();
+        if (outcome == MetricsTags.ProcessingOutcome.UNDELIVERABLE || payloadSize <= 0) {
+            result.complete();
+        } else {
+            Optional.ofNullable(dataVolumeCache.getIfPresent(tenant))
+                //No data volume resource limit set
+                .orElse(CompletableFuture.completedFuture(null))
+                .whenComplete((value, error) -> {
+                    if (value != null) {
+                        dataVolumeCache.put(tenant, CompletableFuture.completedFuture(
+                                new LimitedResource<>(value.getCurrentLimit(), value.getCurrentValue() + payloadSize))
+                        );
+                    }
+
+                    result.complete();
+                });
+        }
+
+        return result.future();
+    }
+
     private void checkMessageLimit(
             final TenantObject tenant,
             final long payloadSize,
--- a/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/ResourceLimitChecks.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/ResourceLimitChecks.java
@@ -13,6 +13,7 @@
 package org.eclipse.hono.service.resourcelimits;
 
 import org.eclipse.hono.client.ServiceInvocationException;
+import org.eclipse.hono.service.metric.MetricsCache;
 import org.eclipse.hono.util.TenantObject;
 
 import io.opentracing.SpanContext;
@@ -21,7 +22,7 @@ import io.vertx.core.Future;
 /**
  * Interface to check if further connections or messages are allowed based on the configured limits.
  */
-public interface ResourceLimitChecks {
+public interface ResourceLimitChecks extends MetricsCache {
 
     /**
      * Checks if the maximum number of connections configured for a tenant
--- a/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/ResourceLimitChecks.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/resourcelimits/ResourceLimitChecks.java
@@ -13,7 +13,6 @@
 package org.eclipse.hono.service.resourcelimits;
 
 import org.eclipse.hono.client.ServiceInvocationException;
-import org.eclipse.hono.service.metric.MetricsCache;
 import org.eclipse.hono.util.TenantObject;
 
 import io.opentracing.SpanContext;
@@ -22,7 +21,7 @@ import io.vertx.core.Future;
 /**
  * Interface to check if further connections or messages are allowed based on the configured limits.
  */
-public interface ResourceLimitChecks extends MetricsCache {
+public interface ResourceLimitChecks {
 
     /**
      * Checks if the maximum number of connections configured for a tenant
--- a/service-base/src/test/java/org/eclipse/hono/service/cache/SpringBasedExpiringValueCacheTest.java
+++ b/service-base/src/test/java/org/eclipse/hono/service/cache/SpringBasedExpiringValueCacheTest.java
@@ -14,14 +14,19 @@
 package org.eclipse.hono.service.cache;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.mock;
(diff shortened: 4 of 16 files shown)


service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -467,8 +467,11 @@ private void checkMessageLimit(
                         .map(cachedValue -> Future.succeededFuture((long) cachedValue))
                         .orElseGet(() -> executeQuery(queryParams, span).map(bytesConsumed -> addToCache(limitsCache, key, bytesConsumed)))
                         .map(bytesConsumed -> {
+                            final long totalBytesConsumed = bytesConsumed + payloadSize;
+                            limitsCache.put(key, bytesConsumed + payloadSize);
```
sophokles73, line 471: shouldn't this only be put to the cache if the limit is not exceeded?

core/src/main/java/org/eclipse/hono/cache/ExpiringValueCache.java
```diff
@@ -52,6 +52,18 @@
      */
     void put(K key, V value, Duration maxAge);
 
+    /**
+     * Puts a value to the cache but does not change the expiration time of the present value.
```
sophokles73, line 56: how about calling this method *update* instead?
how about providing some feedback if the operation was successful, e.g. by means of returning a boolean?

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
fkaltner: I thought about this too. But since I observed the Prometheus metrics are also increasing even in case the message limit was hit, I though it is more consistent this way.

sophokles73: If I am not mistaken then the metrics are not increased in case of a server side error, e.g. registry not available. We only count the message bytes in if we are able to process the message or if the reason for a failure to process lies with the client (indicated by a ClientErrorException). Maybe we can follow the same logic here?

core/src/main/java/org/eclipse/hono/cache/ExpiringValueCache.java
fkaltner: Naming it _update_ makes sense!

Personally I don't think a return value is required, since it is currently not used. Anyway I added one - it doesn't add to much complexity and maybe it will be useful for someone in future...

sophokles73: How about also updating the description of the operation to reflect its new name (update)?

service-base/src/test/java/org/eclipse/hono/service/cache/SpringBasedExpiringValueCacheTest.java
```diff
@@ -86,4 +91,53 @@ public void testGetResponseFromCacheEvictsExpiredValue() {
         verify(springCache).evict("key");
     }
 
+    /**
+     * Verifies that a cache entry is not expired if the corresponding put method without an expiry is called.
+     */
+    @Test
+    public void testPutWithoutExpiryDoesNotChangeExpiry() {
```
sophokles73, line 98: testUpdate....?

service-base/src/test/java/org/eclipse/hono/service/cache/SpringBasedExpiringValueCacheTest.java
```diff
@@ -86,4 +91,53 @@ public void testGetResponseFromCacheEvictsExpiredValue() {
...
+        assertThat(result).isTrue();
+        verify(springCache, never()).evict("key");
+    }
+
+    /**
+     * Verifies that an evicted entry is not overwritten if a new value without an expiry being set is put into the
+     * cache.
+     */
+    @Test
+    public void testPutWithoutExpiryDoesNotReplaceAnEvictedEntry() {
```
sophokles73, line 129: testUpdate...?

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
fkaltner: As I said, I observed this locally on my machine. 

Let me double-check, maybe there was something else going on which lead to the increasing byte count in Prometheus.

fkaltner: I still do observe this on my machine. Here's what I did:

1. Started a local Hono instance by using the packages project
1. Created a new tenant and device
1. Configured a message limit (1024) for the created tenant
1. Send messages via HTTP adapter until the message limit was hit
1. Waited until the cache timed out (set locally to 15 seconds)
1. Send one more message
1. Observed that the `bytesConsumed` increased in `PrometheusBasedResourceLimitChecks`

Looking at the code I think this works as programmed, since `metrics.reportTelemetry(..)` is called in `.recover(...)` meaning also in an error case (see https://github.com/eclipse/hono/blob/master/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java#L794).

Also the values queried from Prometheus do include `FORWARDED` and `UNPROCESSABLE` (according my understanding this value is set in case of the returned 429) - hence the bytes reported for a message above the limit will add up.

Hopefully I did not miss something obvious. If you can confirm my observations and findings in the code I suggest to create a bug and only update the cache if the message limit is not hit.

sophokles73: It makes sense to report every request using *Metrics.reportTelemetry* in order to collect data on all requests. Another question is, which payload the customer needs to pay for. The overall number of bytes uploaded by a customer may indeed by larger than the number determined by the ResourceLimitsChecker, simply because we do not charge the customer for data that has not been processed due to a server error. That is why the query only includes `FORWARDED` and `UNPROCESSABLE` (client error).
This means that the data volume check should not add data to its local counter that cannot be processed due to a server error. Otherwise the calculation of the limit would be inconsistent with when we use consumption data retrieved freshly from Prometheus, right?

fkaltner: Yes, that is exactly my point: this would be inconsistent.

I fully agree on updating the cache only when the limit is hit.

But, if am not completely off track 😄 it's just that the Prometheus-based limit calculation does not work like this:

In case the message limit is hit, actually a `ClientErrorException(HttpResponseStatus.TOO_MANY_REQUESTS.code()))` is thrown (see: https://github.com/eclipse/hono/blob/master/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java#L662), which is then "catched" in the `.recover(...)` callback I mentioned in my previous comment, consequently being reported as a metric with a `UNPROCESSABLE` tag.

When the next message comes in Prometheus is queried by both tags `FORWARDED` and `UNPROCESSABLE`. Which makes me think that the actually rejected message adds to the payload - since we put the label "client error" on it.

Does that make any sense? 😅

sophokles73: > When the next message comes in Prometheus is queried by both tags FORWARDED and UNPROCESSABLE. Which makes me think that the actually rejected message adds to the payload - since we put the label "client error" on it.

Yes, that is expected and intended behavior. We notify the device/client about the fact that the data volume limit has been exceeded. However, we cannot prevent the device/client from sending additional messages with more payload. Once we perform the check if the limit is reached/exceeded, the data has already been transferred to our servers and thus produced costs. The only thing we can do is to charge the customer for this erratic behavior.

> I fully agree on updating the cache only when the limit is hit.

This is not the problem. The problem is that when an adapter invokes the *checkMessageLimit* method, we do not know if the data will need to be accounted for (FORWARDE or UNPROCESSABLE) or not (UNDELIVERABLE). In order to be consistent with the *current bytes consumed* as retrieved form Prometheus, we should only add the message payload length to the local counter in the former case (FORWARDED or UNPROCESSABLE).

fkaltner: > Yes, that is expected and intended behavior. 

Okay, that was what I was confused about.

I'm clear about the rest now. Thank you for the explanation.

fkaltner: As discussed I reworked my changes so that the update of the cache and the check for the limit is separated as it is done in Hono's code up to now. By this it is also assured that the byte counter in the cache is only increased for FORWARDED and UNPROCESSABLE outcomes.

fkaltner: I just took a deeper look on the changes done with https://github.com/eclipse/hono/commit/385c5a45a15a09ceacc675d892d3bf86d4a40941.

I will have to do a rework of my rework before being able to rebase - so reviewing my latest commit is rather pointless until this is done.

fkaltner: Done! Review can continue 😃

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/ResourceLimitChecks.java
```diff
@@ -56,6 +57,22 @@
...
+     *
+     * @param tenant The tenant for which the given payload should be added to the cache.
+     * @param payloadSize The amount of bytes to be added to the cache.
+     * @param outcome The outcome of the operation used for uploading the given payload. The cache will not be updated
+     *                for {@link org.eclipse.hono.service.metric.MetricsTags.ProcessingOutcome#UNDELIVERABLE} outcomes
+     *                as this is not the client's fault and shouldn't add to the number of consumed bytes.
+     *
+     * @return A future indicating when the update was done.
+     */
+    Future<Void> addToMessageBytesCache(String tenant, long payloadSize, MetricsTags.ProcessingOutcome outcome);
```
sophokles73, line 74: how about putting this method into its own interface and then inject an implementation into `MicrometerBasedMetrics` so that you can update the local cache as part of the telemetry reporting. We would then also not need to change every protocol adapter and make sure to not only report metrics via Micrometer but also invoke the ResourceLimitChecks ...

fkaltner: Sound reasonable. But I think we also need to include it as part of `reportCommand`, since the data volume resource limits also apply for commands, right?

sophokles73: yes

fkaltner: Done. Let me know what you think.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/ResourceLimitChecks.java
```diff
@@ -21,7 +22,7 @@
 /**
  * Interface to check if further connections or messages are allowed based on the configured limits.
  */
-public interface ResourceLimitChecks {
+public interface ResourceLimitChecks extends MetricsCache {
```
sophokles73, line 25: Is there any need to tie these two together?
Couldn't `PrometheusBasedResourceLImitChecks` simply *also* implement `MetricsCache`?

fkaltner: No, there's no need need to tie those together, especially because they do not form a "is a" relationship. I adapted the code to your proposal.

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -219,11 +220,11 @@ private Span createSpan(final String name, final SpanContext parent, final Tenan
 
         if (tenant.getResourceLimits() == null) {
             traceItems.put(Fields.EVENT, "no resource limits configured");
-            LOG.trace("no resource limits configured for tenant [{}]", tenant.getTenantId());
+            LOG.warn("no resource limits configured for tenant [{}]", tenant.getTenantId());
```
sophokles73, line 223: this is an expected situation which we explicitly handle. If at all, the level might be increased to DEBUG but I do not see any reason why we should flood the log with this information.

fkaltner: Sorry, just noticed this as well. Shouldn't have gone into the commit.

kaniyan:
@fkaltner @sophokles73 It is a good attempt to increase the message limit accuracy. Considering also the local values may lead to an inconsistent behaviour from the end user point of view. In a Hono setup, it is normal to have more than one protocol adapter instances. In such setup, with this solution, some protocol adapters reject messages from devices stating the reason as _"Message limit exceeded"_ and at the same time other adapter instances may still accept messages from devices. For a customer, this may appear to be an inconsistent behaviour claiming that sometimes messages go through and sometimes not.

fkaltner:
> @fkaltner @sophokles73 It is a good attempt to increase the message limit accuracy. Considering also the local values may lead to an inconsistent behaviour from the end user point of view. In a Hono setup, it is normal to have more than one protocol adapter instances. In such setup, with this solution, some protocol adapters reject messages from devices stating the reason as _"Message limit exceeded"_ and at the same time other adapter instances may still accept messages from devices. For a customer, this may appear to be an inconsistent behaviour claiming that sometimes messages go through and sometimes not.

That is a good point 👍 I think however, since the only other option I see is to have a shared state between all protocol adapter instances (i.e. store the current value of the consumed message bytes in Infinispan), that this is hard to fix. In my eyes the shared state introduces a lot of complexity and has potentially negative aspects on performance, so I would only resort to that if the current accuracy is not enough.

@bordeuax Since you opened the issue - on a second thought - WDYT?

sophokles73:
> I think however, since the only other option I see is to have a shared state between all protocol adapter instances (i.e. store the current value of the consumed message bytes in Infinispan), that this is hard to fix. In my eyes the shared state introduces a lot of complexity and has potentially negative aspects on performance, so I would only resort to that if the current accuracy is not enough.

Actually, we already have the *shared state* by means of Prometheus ...
@kaniyan You raised a very good point and I am very interested in some feedback from @mhemmeter as to whether this would be acceptable?

fkaltner:
Closing this PR (see https://github.com/eclipse/hono/issues/1824?_pjax=%23js-repo-pjax-container#issuecomment-855645893)

Issue #2074, opened 2020-07-08
IllegalStateException in AbstractRequestResponseClient


calohmn:
During some recent failed integration test runs in Travis, there were such kind of exceptions in the log:
````
13:23:04.633 [vert.x-eventloop-thread-2] ERROR io.vertx.core.impl.ContextImpl - Unhandled exception
java.lang.IllegalStateException: Result is already complete: failed
	at io.vertx.core.impl.FutureImpl.complete(FutureImpl.java:112)
	at io.vertx.core.impl.FutureImpl.handle(FutureImpl.java:176)
	at io.vertx.core.impl.FutureImpl.handle(FutureImpl.java:21)
	at org.eclipse.hono.client.impl.AbstractRequestResponseClient.lambda$sendRequest$4(AbstractRequestResponseClient.java:881)
	at io.vertx.proton.impl.ProtonDeliveryImpl.fireUpdate(ProtonDeliveryImpl.java:158)
	at io.vertx.proton.impl.ProtonTransport.handleSocketBuffer(ProtonTransport.java:161)
	at io.vertx.core.net.impl.NetSocketImpl$DataMessageHandler.handle(NetSocketImpl.java:379)
	at io.vertx.core.net.impl.NetSocketImpl.lambda$new$2(NetSocketImpl.java:101)
	at io.vertx.core.streams.impl.InboundBuffer.handleEvent(InboundBuffer.java:237)
	at io.vertx.core.streams.impl.InboundBuffer.write(InboundBuffer.java:127)
	at io.vertx.core.net.impl.NetSocketImpl.handleMessage(NetSocketImpl.java:357)
	at io.vertx.core.impl.ContextImpl.executeTask(ContextImpl.java:366)
	at io.vertx.core.impl.EventLoopContext.execute(EventLoopContext.java:43)
	at io.vertx.core.impl.ContextImpl.executeFromIO(ContextImpl.java:229)
	at io.vertx.core.net.impl.VertxHandler.channelRead(VertxHandler.java:173)
[...]
13:23:07.134 [vert.x-eventloop-thread-2] ERROR o.e.h.t.amqp.CommandAndControlAmqpIT - failed to establish connection to AMQP adapter [host: localhost, port: 32785]
org.eclipse.hono.client.ServerErrorException: connection to server lost
	at org.eclipse.hono.client.impl.ClientFactory.lambda$createClient$0(ClientFactory.java:91)
[...]

PR #2080, opened 2020-07-09
[#2074] Prevent IllegalStateException if request is already cancelled

calohmn:
This is for #2074.


sophokles73:
@calohmn how about adding this to 1.2.4 as well?

calohmn:
@sophokles73 yes, will do.

PR #2081, opened 2020-07-09
[#2068] Used cached connected check result in isConnected()

### LINKED ISSUE #2068: Make HotrodCache connection check less expensive
The `HotrodCache` `isConnected` check is rather expensive because it performs a cache PUT operation each time. That check is invoked in the protocol adapter readiness check and is also performed each time an MQTT connection is made to the MQTT adapter.

When there are many MQTT devices trying to connect to the MQTT adapter in parallel or in quick succession, a lot of unnecessary cache traffic is generated.

A possible improvement would be to only perform the actual cache PUT operation as part of the periodic readiness check. Any intermittent `isConnected` invocations could use a cached readiness check result value.

_
Side note: the `AbstractProtocolAdapterBase.isConnected()` check which invokes `HotrodCache.isConnected` is not used when a device opens an AMQP connection to the AMQP adapter.

calohmn:
This fixes #2068.

### DIFF
--- a/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/BasicCache.java
+++ b/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/BasicCache.java
@@ -49,10 +49,20 @@ public abstract class BasicCache<K, V> implements Cache<K, V>, ConnectionLifecyc
 
     private static final Logger LOG = LoggerFactory.getLogger(BasicCache.class);
 
+    /**
+     * Maximum age for a cached connection check result to be used in {@link #isConnected()}.
+     * Value is chosen to be a bit longer than the default Kubernetes readiness probe interval, so that
+     * the periodic invocations of {@link #checkForCacheAvailability()} as part of the readiness check
+     * keep the cached connection check result up to date.
+     */
+    private static final long CACHED_CONNECTION_CHECK_RESULT_MAX_AGE_MILLIS = TimeUnit.SECONDS.toMillis(11);
+
     protected final Vertx vertx;
     private final BasicCacheContainer cacheManager;
     private final K connectionCheckKey;
     private final V connectionCheckValue;
+
+    private ConnectionCheckResult lastConnectionCheckResult;
 
     private org.infinispan.commons.api.BasicCache<K, V> cache;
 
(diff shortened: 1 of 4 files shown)


client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/BasicCache.java
```diff
@@ -318,11 +334,53 @@ protected void setCache(final org.infinispan.commons.api.BasicCache<K, V> cache)
...
+        }
+
+        /**
+         * Checks if the result is older than the given time span, determined from the current point in time.
+         *
+         * @param millis The time span in milliseconds.
+         * @return {@code true} if the result is older.
+         */
+        public boolean isOlderThan(final long millis) {
+            return timestamp < (System.currentTimeMillis() - millis);
```
ctron, line 370: Rather than using "long" … I would suggest the new Java Instant type.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/BasicCache.java
```diff
@@ -49,11 +49,21 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(BasicCache.class);
 
+    /**
+     * Maximum age for a cached connection check result to be used in {@link #isConnected()}.
+     * Value is chosen to be a bit longer than the default Kubernetes readiness probe interval, so that
+     * the periodic invocations of {@link #checkForCacheAvailability()} as part of the readiness check
+     * keep the cached connection check result up to date.
+     */
+    private static final long CACHED_CONNECTION_CHECK_RESULT_MAX_AGE_MILLIS = TimeUnit.SECONDS.toMillis(11);
```
ctron, line 58: I would suggest to use Duration, rather than a plain long.

sophokles73: This effectively sets a lower boundary to the k8s readiness/liveness check frequency, doesn't it? I wonder if we could simply continuously update the connection status with each invocation of any of the cache's methods, i.e. whenever a method successfully interacts with the cache, we set the flag to true and set it to false if the invocation fails. I do not think that this would be any worse than doing the periodic put triggered by the readiness probe which doesn't guarantee that the connection works for subsequent method invocations anyway. On startup of the cache we invoke the put method once to initialize the connection status flag. WDYT?

calohmn: Yes, I think that would be a better solution still and would also simplify things in the `BasicCache` class.
Checking again, if with the newer Hotrod client there is maybe a way to use `remoteCache.getRemoteCacheManager().getConnectionCount()` or something like that, in addition to the cache operation status update, would also be good (haven't tested that yet).

I think that would better be done in a new PR, closing this here then.

PR #2085, opened 2020-07-17
Upgrade caffeine cache library

dejanb:


sophokles73:
@dejanb can we please use 2.8.4 instead? There is no clearlydefined data on 2.8.5 available yet ...

dejanb:
@sophokles73 Sure thing. I was looking the same and wanted to use 2.8.4, but I was sure you'll ask "why not 2.8.5" 🙂. I'll make a change and merge this.

PR #2086, opened 2020-07-21
linux profiles MUST NOT contain mac os dependencies

ctron:


ctron:
We also need a backport of this to 1.3.x

PR #2089, opened 2020-07-22
CoAP adapter: Skip 'PSK-getDeviceIdentityInfo' span if no lookup is done

calohmn:
The top level `PSK-getDeviceIdentityInfo` trace span (created as part of the TLS handshake handling) will be a 0ms operation with no child spans if the device id is taken from a previous lookup. Therefore the creation of such a span should better be skipped in that case, so that the span is only created if an actual device id lookup is done (in case of a session resumption).

With this PR, instead of the two `PSK-getSecretKey` and `PSK-getDeviceIdentityInfo` spans, there will only be one `PSK-getDeviceSecretKeyAndIdentityInfo` span.

### DIFF
--- a/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DefaultDeviceResolver.java
+++ b/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DefaultDeviceResolver.java
@@ -141,18 +141,17 @@ public class DefaultDeviceResolver implements ApplicationLevelInfoSupplier, Adva
         final Map<String, Object> result = new HashMap<>();
 
         if (clientIdentity instanceof PreSharedKeyIdentity) {
-            final Span span = newSpan("PSK-getDeviceIdentityInfo");
-            final PreSharedKeyDeviceIdentity deviceIdentity = getHandshakeIdentity(span, clientIdentity.getName());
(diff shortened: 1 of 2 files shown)


adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DefaultDeviceResolver.java
```diff
@@ -193,9 +192,9 @@ private void loadCredentialsForDevice(final ConnectionId cid, final PskPublicInf
         final String publicInfo = identity.getPublicInfoAsString();
         LOG.debug("getting PSK secret for identity [{}]", publicInfo);
 
-        final Span span = newSpan("PSK-getSecretKey");
+        final Span span = newSpan("PSK-getDeviceSecretKeyAndIdentityInfo");
```
boaks, line 195: I would prefer: "PSK-getDeviceCredentials"

boaks:
Beside the name "PSK-getDeviceSecretKeyAndIdentityInfo" it looks well.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DefaultDeviceResolver.java
calohmn: Good, I've changed it accordingly.

PR #2090, opened 2020-07-23
Improve configuration classes

dejanb:
While working on Quarkus PoC I found some improvements we need to in our configuration classes. The idea is to bring that in as a separate PR (as some other improvements not related to Quarkus directly).

The main changes are:
* Defining default values for every property
* Adapting proper property names with getter/setter methods
* Using better suited classes for properties

### DIFF
--- a/core/src/main/java/org/eclipse/hono/config/AbstractConfig.java
+++ b/core/src/main/java/org/eclipse/hono/config/AbstractConfig.java
@@ -22,6 +22,7 @@ import java.util.Objects;
 
 import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.PortConfigurationHelper;
+import org.eclipse.hono.util.Strings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -43,23 +44,23 @@ public abstract class AbstractConfig {
      */
     protected final Logger LOG = LoggerFactory.getLogger(getClass());
 
-    private String trustStorePath;
-    private char[] trustStorePassword;
-    private TrustOptions trustOptions;
+    private TrustOptions trustOptions = null;
+    private String trustStorePath = null;
+    private String trustStorePassword = null;
     private String pathSeparator = Constants.DEFAULT_PATH_SEPARATOR;
-    private String keyStorePath;
-    private char[] keyStorePassword;
-    private String certPath;
-    private String keyPath;
-    private KeyCertOptions keyCertOptions;
-    private FileFormat trustStoreFormat;
-    private FileFormat keyFormat;
+    private String keyStorePath = null;
+    private String keyStorePassword = null;
+    private String certPath = null;
+    private String keyPath = null;
+    private KeyCertOptions keyCertOptions = null;
+    private FileFormat trustStoreFormat = null;
+    private FileFormat keyFormat = null;
     private List<String> secureProtocols = Collections.singletonList("TLSv1.2");
 
     /**
      * Creates a new empty instance.
      */
-    protected AbstractConfig() {
+    public AbstractConfig() {
     }
 
     /**
@@ -68,9 +69,8 @@ public abstract class AbstractConfig {
      * @param other The other instance. All of the other instance's properties
      *              are copied to the newly created instance.
      */
-    protected AbstractConfig(final AbstractConfig other) {
+    public AbstractConfig(final AbstractConfig other) {
         this.certPath = other.certPath;
-        this.keyCertOptions = other.keyCertOptions;
         this.keyFormat = other.keyFormat;
         this.keyPath = other.keyPath;
         this.keyStorePassword = other.keyStorePassword;
@@ -138,7 +138,7 @@ public abstract class AbstractConfig {
      * @see #getTrustStorePath()
      */
     public final String getTrustStorePassword() {
-        return fromChars(trustStorePassword);
+        return trustStorePassword;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/eclipse/hono/config/AbstractConfig.java
```diff
@@ -41,25 +41,23 @@
...
      *  A logger to be shared with subclasses.
      */
-    protected final Logger LOG = LoggerFactory.getLogger(getClass());
+    public static final Logger LOG = LoggerFactory.getLogger(AbstractConfig.class);
 
-    private String trustStorePath;
-    private char[] trustStorePassword;
-    private TrustOptions trustOptions;
+    private String trustStorePath = "";
+    private String trustStorePassword = "";
```
calohmn, line 47: I think it's common practice to rather use `char[]` for password fields.

core/src/main/java/org/eclipse/hono/config/AbstractConfig.java
```diff
@@ -41,25 +41,23 @@
     /**
      *  A logger to be shared with subclasses.
      */
-    protected final Logger LOG = LoggerFactory.getLogger(getClass());
+    public static final Logger LOG = LoggerFactory.getLogger(AbstractConfig.class);
```
calohmn, line 44: Can you explain the motivation here? With the old definition, there is the added info in the log from which actual (sub)class the log statements are coming.

dejanb: That was left from the PoC phase as there were some issues with native image. Now everything works fine, so I reverted it back.

core/src/main/java/org/eclipse/hono/config/AbstractConfig.java
dejanb: As all other libraries we are passing these passwords to are keeping them as Strings in memory, I don't think we're getting any benefits from keeping them as char arrays. 
And if we still chose to do so, I think getter/setter methods should follow that as well. 
Anyway, in my opinion we're just keeping a lot of transformation code that gives us no real benefits.

calohmn: Ok.

core/src/main/java/org/eclipse/hono/config/AuthenticatingClientConfigProperties.java
```diff
@@ -29,14 +30,15 @@
...
     public static final String SERVER_ROLE_UNKNOWN = "unknown";
 
-    private String credentialsPath;
+    private String credentialsPath = "";
     private String host = "localhost";
     private boolean hostnameVerificationRequired = true;
-    private char[] password;
-    private int port;
+    private String password = "";
+    private int port = Constants.PORT_AMQPS;
```
sophokles73, line 37: what's the reason for having to use an empty string as the initial value?

dejanb: @sophokles73 At the moment with way we use Quarkus, it is complaining when properties don't have default values. So this is a workaround. I haven't yet raised this with the Quarkus devs, but I can do that and we can revert it in the future (I'm not too fond of it myself either).

sophokles73: yes, please do so. Honestly, I hate the guts out of it ;-)

core/src/main/java/org/eclipse/hono/config/AbstractConfig.java
```diff
@@ -186,25 +175,21 @@ public final TrustOptions getTrustOptions() {
...
             LOG.debug("using certificates from PKCS12 key store [{}] as trust anchor", this.trustStorePath);
-            trustOptions = new PfxOptions()
+            return new PfxOptions()
                     .setPath(getTrustStorePath())
                     .setPassword(getTrustStorePassword());
-            break;
         case JKS:
             LOG.debug("using certificates from JKS key store [{}] as trust anchor", this.trustStorePath);
-            trustOptions = new JksOptions()
+            return new JksOptions()
```
sophokles73, line 186: I do see value in creating the options once only. However, we then should also return the already created instance in subsequent invocations of this method. WDYT?

core/src/main/java/org/eclipse/hono/config/KeyLoader.java
```diff
@@ -162,10 +162,10 @@ public static KeyLoader fromFiles(final Vertx vertx, final String keyPath, final
         PrivateKey privateKey = null;
         List<Certificate> certChain = null;
 
-        if (keyPath != null) {
+        if (keyPath != null && !keyPath.isEmpty()) {
```
sophokles73, line 165: StringUtils.isNullOrEmpty()?

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
```diff
@@ -36,7 +36,7 @@
     private long eventLoopBlockedCheckTimeout = 5000L;
     private int maxConnections = 0;
     private Duration tenantIdleTimeout = DEFAULT_TENANT_IDLE_TIMEOUT;
-    private Map<String, MapperEndpoint> mapperEndpoints = new HashMap<>();
+    private HashMap<String, MapperEndpoint> mapperEndpoints = new HashMap<>();
```
sophokles73, line 39: We usually use an abstract type for field declarations, don't we? Why can't we keep the `Map` here?

core/src/main/java/org/eclipse/hono/config/AuthenticatingClientConfigProperties.java
dejanb: Setting them explicitly to null also does the trick and it looks a bit nicer. So let's go with that for the moment and push for the fix.

core/src/main/java/org/eclipse/hono/config/ProtocolAdapterProperties.java
dejanb: We're basically hitting this with using the Map: https://github.com/quarkusio/quarkus/issues/8017. And that's not considered as the issue at the moment. We can create our own converter perhaps and I'll ask if there are a better way to do all this, but to me this is a quick/easy workaround that we can use for now.

PR #2091, opened 2020-07-24
Refactor protocol adapters to support non spring based implementations

dejanb:
This PR refactors a bit adapters maven modules to better support non Spring based implementations. The actual changes are:
* Remove Spring dependencies from the parent module and introduce them only to the actual adapter implementation
* Move Vert.x implementation classes to the base modules, so that they can be reused


adapters/base-spring/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+        <dependency>
+          <groupId>org.springframework.boot</groupId>
+          <artifactId>spring-boot-autoconfigure</artifactId>
+        </dependency>
+        <dependency>
+          <groupId>org.springframework.boot</groupId>
+          <artifactId>spring-boot-starter-logging</artifactId>
+        </dependency>
+    </dependencies>
+
```
ctron, line 31: Double empty line :grin:

dejanb: Fixed

adapters/base-spring/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+</project>
```
sophokles73, line 32: missing EOL

dejanb:
> why do you want to move the CoAP adapter implementation class?

@sophokles73 I moved all vert.x based implementation to the base modules (where they exists), so that they can be reused between Spring and Quarkus deployments. That's how HTTP adapter works for now, but I can imagine we do the same for MQTT, CoAP, etc.

PR #2092, opened 2020-07-28
CoAP adapter: fix tracing info and add tests

calohmn:
This fixes the CoAP adapter always setting an `authenticated: true` tag on the "upload telemetry/event" trace span - even for an unauthenticated request. This is done by using a `null` value for `authenticatedDevice` variables in case of an unauthenticated request, like it is also used in the methods of the protocol adapter base classes.

Also integration tests regarding Command & Control with an unauthenticated device have been added.

The refactorings done here serve as a preparation for the fix for #1395.


tests/src/test/java/org/eclipse/hono/tests/coap/CoapTestBase.java
```diff
@@ -1053,6 +1059,53 @@ private URI getRequestUri(final String scheme, final String resource) {
...
+            final CoapCommandEndpointConfiguration endpointConfig,
+            final String requestDeviceId,
+            final int msgNo) {
+
+        if (endpointConfig.isSubscribeAsGatewayForSingleDevice()) {
+            return createCoapsRequest(Code.PUT, getMessageType(), getPutResource(tenantId, requestDeviceId), msgNo);
+        } else if (endpointConfig.isSubscribeAsUnauthenticatedDevice()) {
+            return createCoapRequest(Code.PUT, getMessageType(), getPutResource(tenantId, requestDeviceId), msgNo);
+        } else {
+            return createCoapsRequest(Code.POST, getMessageType(), getPostResource(), msgNo);
```
boaks, line 1081: I'm not sure, if a device, which doesn't use coaps, could be considered to be authenticated?

calohmn: The not-using-coaps request handling is done in the `UnauthenticatedDevice` block.

PR #2098, opened 2020-08-03
add a Json Pointer parser

jbtrystram:
It's needed to properly finish https://github.com/eclipse/hono/pull/2009/ :)

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/JsonHelper.java
+++ b/core/src/main/java/org/eclipse/hono/util/JsonHelper.java
@@ -74,6 +74,7 @@ public final class JsonHelper {
     public static <T> T getValue(final JsonArray jsonArray, final int pos, final Class<T> clazz,
                                  final T defaultValue) {
         Objects.requireNonNull(pos);
+        Objects.requireNonNull(jsonArray);
 
         if (pos > jsonArray.size()) {
             LOG.debug("index {} larger than array. Actual size is {}", pos, jsonArray.size());
@@ -141,7 +142,11 @@ public final class JsonHelper {
 
         //exit condition
         if (path.size() == 1) {
-            return getValue(jsonObject.getJsonArray(key), index, clazz, defaultValue);
+            try {
+                return getValue(jsonObject.getJsonArray(key), index, clazz, defaultValue);
+            } catch (NullPointerException e) {
+                return defaultValue;
+            }
         }
 
         return getValueFromJsonPath(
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/eclipse/hono/util/JsonHelper.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2019-2020 Contributors to the Eclipse Foundation
```
sophokles73, line 2: 2019, 2020

core/src/main/java/org/eclipse/hono/util/JsonHelper.java
```diff
@@ -55,4 +58,100 @@ private JsonHelper() {
...
+                    clazz.getSimpleName());
+            return defaultValue;
+        }
+    }
+
+    /**
+     * Return the value designated from a JsonPath in the given JsonObject.
+     *
+     * @param jsonObject The JSON object to get the value from.
+     * @param path The JsonPath representation of the value to extract.
```
sophokles73, line 101: FMPOV we need a definition of the path syntax. Is there an existing definition somewhere online which we can refer to?

dejanb:
Looks good to me in general. I agree with Kai, that it'd be good to have a link to the "spec" (https://goessner.net/articles/JsonPath/)

Another question is, why we don't use the existing library for it? https://github.com/json-path/JsonPath

sophokles73:
> Another question is, why we don't use the existing library for it?

@dejanb I had the same feeling but that doesn't help much because we are not trying to extract data from an existing JSON document but instead simply want to use the notation in order to specify a query on the underlying persistence technology.
BTW MongoDB doesn't support queries expressed in JSONPath even though its data format is already JSON :-(
Another alternative to using a subset of JSONPath might be to use [JSON Pointer](https://tools.ietf.org/html/rfc6901) instead which is much more lightweight and only specifies how to address a particular value within a document ...

jbtrystram:
@sophokles73 I was actually just looking at the Json Pointer spec and thinking it may be a better fit for that use case.

dejanb:
@sophokles73 @jbtrystram I see. I'm OK with exploring JSON Pointer solution as well.

jbtrystram:
@dejanb @sophokles73 I updated this PR with a Json Pointer implementation. I added a bunch of tests for the special use cases and as far as I know it's compliant to the spec. 
Let me know what you think.

sophokles73:
@jbtrystram what about `io.vertx.core.json.pointer.JsonPointer`? Doesn't that work?

jbtrystram:
:( damn, this was added in vertx after this [issue](https://github.com/eclipse-vertx/vert.x/issues/1661) stated they won't support it.
This can be closed I guess

PR #2100, opened 2020-08-05
[#1023] Add connection attempt outcomes for failure scenarios

sophokles73:
Connections attempts of devices with protocol adapters may fail due to
arbitrary reasons. The "outcome" tag of the "hono.connections.attempts"
meter has been extended with additional values to reflect the concrete
error that caused the failure to establish the connection.


sophokles73:
@fkaltner maybe you also want to take a look?

adapters/mqtt-vertx-base/src/test/resources/logback-test.xml
```diff
@@ -27,7 +27,7 @@
     <appender-ref ref="STDOUT" />
   </root>
 
-  <logger name="org.eclipse.hono.adapter" level="INFO"/>
+  <logger name="org.eclipse.hono.adapter" level="DEBUG"/>
```
dejanb, line 30: Do we want to leave this to DEBUG or is it just overlooked?

sophokles73: No, this should be INFO ... will fix. Thanks for spotting :+1:

PR #2106, opened 2020-08-10
actility enterprise provider added + ability to configure multiple URIs per loraProvider

BobClaerhout:


adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/providers/ActilityBaseProvider.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @Override
+    protected LoraMetaData getMetaData(final JsonObject loraMessage) {
+
+        Objects.requireNonNull(loraMessage);
+
+        return getRootObject(loraMessage)
+            .map(this::extractMetaData)
+            .orElse(null);
+    }
```
sophokles73, line 112: Are these methods supposed to be overridden by subclasses? If not then I would suggest to declare them final in order to prevent hard to find problems when somebody does ...

BobClaerhout: no they are not, changed it according to your proposal

sophokles73:
@BobClaerhout can you please rebase and resolve the conflicts?

PR #2108, opened 2020-08-12
[#1395] Implement trace sampling mode option in CoAP adapter

### LINKED ISSUE #1395: Enable forced sampling of traces for a specific tenant (CoAP adapter)
The new config options introduced in #896 to enable forced sampling of traces (i.e. recording of traces in the tracing backend) for a specific tenant or auth-id should also be taken into account in the CoAP adapter.

This relies on tracing support being added first (#1392).

calohmn:
This fixes #1395.

### DIFF
--- a/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
+++ b/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
@@ -360,14 +360,18 @@ public abstract class AbstractVertxBasedCoapAdapter<T extends CoapAdapterPropert
 
     private Future<Endpoint> createSecureEndpoint(final NetworkConfig config) {
 
+        final CoapContextTenantAndAuthIdProvider tenantObjectWithAuthIdProvider = new CoapContextTenantAndAuthIdProvider(
+                getConfig(), getTenantClientFactory());
         final ApplicationLevelInfoSupplier deviceResolver = Optional.ofNullable(honoDeviceResolver)
-                .orElse(new DefaultDeviceResolver(context, tracer, getTypeName(), getConfig(), getCredentialsClientFactory()));
+                .orElse(new DefaultDeviceResolver(context, tracer, getTypeName(), getConfig(),
+                        getCredentialsClientFactory(), tenantObjectWithAuthIdProvider));
         final AdvancedPskStore store = Optional.ofNullable(pskStore)
                 .orElseGet(() -> {
                     if (deviceResolver instanceof AdvancedPskStore) {
                         return (AdvancedPskStore) deviceResolver;
                     } else {
-                        return new DefaultDeviceResolver(context, tracer, getTypeName(), getConfig(), getCredentialsClientFactory());
+                        return new DefaultDeviceResolver(context, tracer, getTypeName(), getConfig(),
+                                getCredentialsClientFactory(), tenantObjectWithAuthIdProvider);
                     }
                 });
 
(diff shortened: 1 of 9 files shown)


adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
```diff
@@ -1085,18 +1102,14 @@ private TtdStatus getTtdStatus(final CoapContext context) {
...
      * @return A succeeded future containing the CoAP status code that has been returned to the device.
-     * @throws NullPointerException if context or device are {@code null}.
+     * @throws NullPointerException if context is {@code null}.
      */
-    public final Future<ResponseCode> uploadCommandResponseMessage(
-            final CoapContext context,
-            final Device device,
-            final Device authenticatedDevice) {
-
+    public final Future<ResponseCode> uploadCommandResponseMessage(final CoapContext context) {
```
sophokles73, line 1108: nice catch :+1:

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapContext.java
```diff
@@ -45,14 +45,20 @@
 
     private final CoapExchange exchange;
     private final Device originDevice;
+    private final String authId;
```
sophokles73, line 48: isn't this useful for all other adapters as well, i.e. doesn't it make sense to pull it up into MapBasedTelemetryExecutionContext?

calohmn:
> I wonder if we can get rid of the *TenantObjectWithIdProvider classes by simply putting the TenantObject into the ExecutionContext at the beginning of processing each message from a device. We need to retrieve the TenantObject anyway for doing checks ...

I have started looking into doing that.
I'm not sure if that ends up to be the better solution (I would want to avoid having to put the *TenantObjectWithIdProvider methods in the *ProtocolAdapter classes), but let's see. In any case, I would put the changes into another PR then.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapContext.java
calohmn: Currently that doesn't look useful. But it might be when doing the refactorings concerning the *TenantObjectWithIdProvider classes.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/RequestDeviceAndAuth.java
```diff
@@ -53,6 +59,17 @@ public Device getOriginDevice() {
...
     }
 
+    /**
+     * Gets the authentication identifier of the request.
+     * <p>
+     * Will be {@code null} for an unauthenticated request.
+     *
+     * @return The authentication identifier or {@code null}.
+     */
+    public String getAuthId() {
```
sophokles73, line 69: final ?

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DefaultDeviceResolver.java
```diff
@@ -146,6 +153,7 @@ public AdditionalInfo getInfo(final Principal clientIdentity, final Object custo
                 final String deviceId = (String) customArgument;
                 final PreSharedKeyDeviceIdentity deviceIdentity = getHandshakeIdentity(clientIdentity.getName(), null);
                 result.put("hono-device", new Device(deviceIdentity.getTenantId(), deviceId));
+                result.put("hono-auth-id", deviceIdentity.getAuthId());
```
sophokles73, line 156: I guess it would be good to define constants for these keys as they are used across different classes

Issue #2112, opened 2020-08-14
Extend MongoDB device registry to support search devices operation


kaniyan:
With the completion of the issue #1958,  the _Open API_ spec for the _Device Registry Management API_ contains also  specifications for the new _search devices_ operation. Based on this spec, the _search devices_ operation is to be implemented in the _MongoDB based device registry_.

sophokles73:
@kaniyan can this be closed?

kaniyan:
> @kaniyan can this be closed?

yes. This has been closed via #2116, #2122, #2138 and #2150.

PR #2113, opened 2020-08-14
[#2112] Part 1: Extend device registry base classes for search devices operation

kaniyan:
This PR focuses on extending the device registry base classes to support the implementation of the search devices operation in the _MongoDB_ based device registry.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
+++ b/core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
@@ -104,6 +104,46 @@ public final class RegistryManagementConstants extends RequestResponseApiConstan
      */
     public static final String FIELD_MEMBER_OF = "memberOf";
 
+    /**
+     * The name of the field that contains the JSON pointer corresponding to the field used for filtering devices.
+     */
+    public static final String FIELD_FILTER_FIELD = "field";
+
+    /**
+     * The name of the field that contains the filter JSON object for search devices operation.
+     */
+    public static final String FIELD_FILTER_JSON = "filterJson";
+
+    /**
+     * The name of the field that contains the operator used for filtering devices.
+     */
+    public static final String FIELD_FILTER_OPERATOR = "op";
+
+    /**
+     * The name of the field that contains the value used for filtering devices.
+     */
+    public static final String FIELD_FILTER_VALUE = "value";
+
+    /**
+     * The name of the field that contains the page offset for search devices operation.
+     */
+    public static final String FIELD_PAGE_OFFSET = "pageOffset";
+
+    /**
+     * The name of the field that contains the page size for search devices operation.
+     */
+    public static final String FIELD_PAGE_SIZE = "pageSize";
+
+    /**
+     * The name of the field that contains sort direction used by search devices operation to sort the result set.
+     */
+    public static final String FIELD_SORT_DIRECTION = "direction";
+
+    /**
+     * The name of the field that contains the sort JSON object used by search devices operation to sort the result set.
+     */
+    public static final String FIELD_SORT_JSON = "sortJson";
+
     // CREDENTIALS
 
     /**
--- a/core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
+++ b/core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
@@ -110,9 +110,9 @@ public final class RegistryManagementConstants extends RequestResponseApiConstan
     public static final String FIELD_FILTER_FIELD = "field";
 
     /**
-     * The name of the field that contains the filter JSON object for search devices operation.
+     * The name of the query parameter that contains the filter JSON object for search devices operation.
      */
-    public static final String FIELD_FILTER_JSON = "filterJson";
+    public static final String PARAM_FILTER_JSON = "filterJson";
 
     /**
      * The name of the field that contains the operator used for filtering devices.
@@ -125,14 +125,14 @@ public final class RegistryManagementConstants extends RequestResponseApiConstan
     public static final String FIELD_FILTER_VALUE = "value";
 
     /**
-     * The name of the field that contains the page offset for search devices operation.
+     * The name of the query parameter that contains the page offset for search devices operation.
      */
-    public static final String FIELD_PAGE_OFFSET = "pageOffset";
+    public static final String PARAM_PAGE_OFFSET = "pageOffset";
 
     /**
-     * The name of the field that contains the page size for search devices operation.
+     * The name of the query parameter that contains the page size for search devices operation.
      */
-    public static final String FIELD_PAGE_SIZE = "pageSize";
+    public static final String PARAM_PAGE_SIZE = "pageSize";
 
     /**
      * The name of the field that contains sort direction used by search devices operation to sort the result set.
@@ -140,9 +140,10 @@ public final class RegistryManagementConstants extends RequestResponseApiConstan
     public static final String FIELD_SORT_DIRECTION = "direction";
 
     /**
-     * The name of the field that contains the sort JSON object used by search devices operation to sort the result set.
+     * The name of the query parameter that contains the sort JSON object used by search devices operation to sort the
+     * result set.
      */
-    public static final String FIELD_SORT_JSON = "sortJson";
+    public static final String PARAM_SORT_JSON = "sortJson";
 
     // CREDENTIALS
 
--- a/services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DelegatingDeviceManagementHttpEndpoint.java
+++ b/services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DelegatingDeviceManagementHttpEndpoint.java
@@ -53,6 +53,11 @@ import io.vertx.ext.web.RoutingContext;
  */
 public class DelegatingDeviceManagementHttpEndpoint<S extends DeviceManagementService> extends AbstractDelegatingRegistryHttpEndpoint<S, ServiceConfigProperties> {
 
+    private static final int DEFAULT_PAGE_OFFSET = 0;
+    private static final int DEFAULT_PAGE_SIZE = 30;
+    private static final int MAX_PAGE_SIZE = 200;
+    private static final int MIN_PAGE_OFFSET = 0;
+    private static final int MIN_PAGE_SIZE = 0;
     private static final String SPAN_NAME_CREATE_DEVICE = "create Device from management API";
     private static final String SPAN_NAME_GET_DEVICE = "get Device from management API";
     private static final String SPAN_NAME_SEARCH_DEVICES = "search Devices from management API";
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -104,6 +104,46 @@
...
+
+    /**
+     * The name of the field that contains the page offset for search devices operation.
+     */
+    public static final String FIELD_PAGE_OFFSET = "pageOffset";
+
+    /**
+     * The name of the field that contains the page size for search devices operation.
+     */
+    public static final String FIELD_PAGE_SIZE = "pageSize";
```
sophokles73, line 135: these are query params, thus I guess we should name them starting with `PARAM_` ...

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -104,6 +104,46 @@
...
 
+    /**
+     * The name of the field that contains the JSON pointer corresponding to the field used for filtering devices.
+     */
+    public static final String FIELD_FILTER_FIELD = "field";
+
+    /**
+     * The name of the field that contains the filter JSON object for search devices operation.
+     */
+    public static final String FIELD_FILTER_JSON = "filterJson";
```
sophokles73, line 115: this is a query param, thus I guess we should name it starting with PARAM_ .

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -104,6 +104,46 @@
...
+
+    /**
+     * The name of the field that contains sort direction used by search devices operation to sort the result set.
+     */
+    public static final String FIELD_SORT_DIRECTION = "direction";
+
+    /**
+     * The name of the field that contains the sort JSON object used by search devices operation to sort the result set.
+     */
+    public static final String FIELD_SORT_JSON = "sortJson";
```
sophokles73, line 145: this is a query param, thus I guess we should name it starting with PARAM_ .

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DelegatingDeviceManagementHttpEndpoint.java
```diff
@@ -135,6 +143,39 @@ private void doGetDevice(final RoutingContext ctx) {
...
 
+    private void doSearchDevices(final RoutingContext ctx) {
+        final Span span = TracingHelper.buildServerChildSpan(
+                tracer,
+                TracingHandler.serverSpanContext(ctx),
+                SPAN_NAME_SEARCH_DEVICES,
+                getClass().getSimpleName()).start();
+
+        final Future<String> tenantId = getRequestParameter(ctx, PARAM_TENANT_ID,
+                getPredicate(config.getTenantIdPattern(), false));
```
sophokles73, line 154: I do not think that we need to check the tenant ID here as we are merely using it as a query parameter, right? We do not need to prevent creation of a tenant with an illegal ID here,

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DelegatingDeviceManagementHttpEndpoint.java
```diff
@@ -135,6 +143,39 @@ private void doGetDevice(final RoutingContext ctx) {
...
+        final Optional<Integer> pageSize = Optional
+                .ofNullable(ctx.request().getParam(RegistryManagementConstants.FIELD_PAGE_SIZE))
+                .map(Integer::parseInt);
```
sophokles73, line 157: what if it's a float?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DelegatingDeviceManagementHttpEndpoint.java
```diff
@@ -135,6 +143,39 @@ private void doGetDevice(final RoutingContext ctx) {
...
+        final Optional<Integer> pageOffset = Optional
+                .ofNullable(ctx.request().getParam(RegistryManagementConstants.FIELD_PAGE_OFFSET))
+                .map(Integer::parseInt);
```
sophokles73, line 160: what if it's a float?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementService.java
```diff
@@ -67,6 +71,51 @@
      */
     Future<OperationResult<Device>> readDevice(String tenantId, String deviceId, Span span);
 
+    /**
+     * Search devices belonging to the given tenant with optional filters, paging and sorting options.
```
sophokles73, line 75: Finds devices ...

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceManagementService.java
```diff
@@ -67,6 +71,51 @@
...
+        Objects.requireNonNull(pageSize);
+        Objects.requireNonNull(pageOffset);
+        Objects.requireNonNull(filters);
+        Objects.requireNonNull(sortOptions);
+        Objects.requireNonNull(span);
+
+        final Promise<OperationResult<List<DeviceWithId>>> defaultSearchDevicesResult = Promise.promise();
+        defaultSearchDevicesResult.complete(OperationResult.empty(HttpURLConnection.HTTP_NOT_IMPLEMENTED));
+
+        return defaultSearchDevicesResult.future();
```
sophokles73, line 116: how about
```
return Future.succeededFuture(OperationResult.empty(HttpURLConnection.HTTP_NOT_IMPLEMENTED));
```

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DelegatingDeviceManagementHttpEndpoint.java
```diff
@@ -244,4 +285,23 @@ private void doDeleteDevice(final RoutingContext ctx) {
...
+        final Promise<Optional<List<T>>> result = Promise.promise();
+        final Optional<List<T>> values = Optional.ofNullable(ctx.request().params()
+                .getAll(paramKey))
+                .map(jsons -> jsons
+                        .stream()
+                        .map(json -> Json.decodeValue(json, clazz))
+                        .collect(Collectors.toList()));
+        result.complete(values);
+
+        return result.future();
```
sophokles73, line 305: how about failing the result with a ClientErrorExecption in case the JSON cannot be parsed?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Filter.java
```diff
@@ -0,0 +1,109 @@
...
+    @JsonProperty(RegistryManagementConstants.FIELD_FILTER_VALUE)
+    private T value;
+
+    @JsonProperty(RegistryManagementConstants.FIELD_FILTER_OPERATOR)
+    private OPERATOR operator = OPERATOR.eq;
+
+    /**
+     * An enum defining supported filter operators.
+     */
+    public enum OPERATOR {
```
sophokles73, line 45: Operator

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Sort.java
```diff
@@ -0,0 +1,83 @@
...
+    @JsonProperty(RegistryManagementConstants.FIELD_FILTER_FIELD)
+    private JsonPointer field;
+
+    @JsonProperty(RegistryManagementConstants.FIELD_SORT_DIRECTION)
+    private DIRECTION direction = DIRECTION.asc;
+
+    /**
+     * An enum defining the sort directions.
+     */
+    public enum DIRECTION {
```
sophokles73, line 38: Direction

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Sort.java
```diff
@@ -0,0 +1,83 @@
...
+        asc,
+        desc
+    }
+
```
sophokles73, line 42: can we add a constructor that accepts the mandatory field property and remove the *setField* method and make field final?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Filter.java
```diff
@@ -0,0 +1,109 @@
...
+        eq
+    }
+
```
sophokles73, line 48: can we add a constructor accepting the mandatory properties and remove *setField* and *setValue*?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/Filter.java
kaniyan: My bad. I will change it.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DelegatingDeviceManagementHttpEndpoint.java
kaniyan: In the spec, the required _data type_ for the `pageSize` is mentioned as _integer_. If it is _float_, then `NumberFormatException` will be thrown. I didn't get the intention of the comment. Is it to handle this exception using `ClientErrorExecption`?

sophokles73: When I run a curl command on my machine against your implementation:
```
curl http://registry/v1/devices/my-tenant?pageSize=23.5
```
what will happen?

kaniyan: It will return _500 Internal Server Error_. I will handle those exceptions so that  _HTTP_BAD_REQUEST_ is returned.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DelegatingDeviceManagementHttpEndpoint.java
kaniyan: Ok. Since tenant ID here is a _path_ parameter, I will get it directly using `ctx.request().getParam(PARAM_TENANT_ID)`

kaniyan:
@sophokles73 I have made changes as per the comments and pushed those changes in a separate commit.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DelegatingDeviceManagementHttpEndpoint.java
```diff
@@ -244,4 +282,46 @@ private void doDeleteDevice(final RoutingContext ctx) {
...
+    private Future<Optional<Integer>> getRequestParameterIntegerValue(final RoutingContext ctx,
+            final String paramKey) {
+        Objects.requireNonNull(ctx);
+        Objects.requireNonNull(paramKey);
+
+        final Promise<Optional<Integer>> result = Promise.promise();
+        try {
+            final Optional<Integer> value = Optional.ofNullable(ctx.request().params().get(paramKey))
+                    .map(Integer::parseInt);
+            result.complete(value);
```
sophokles73, line 319: what about checking the min/max values and returning a default value?

kaniyan: The properties such as `tenantId` and `deviceId` patterns used for validations are made configurable by adding them to the `ServiceConfigProperties`. Following the same approach, _min_, _max_ and _default_ values for page size and offset could be made configurable. FMPOV these properties are specific to the device registry and doesn't suit well in that class. Hence I have created a new class `DeviceRegistryManagementHttpServiceConfigProperties` which extends `HttpServiceConfigProperties` to accommodate these _min_, _max_ and _default_ config properties. I have pushed a separate commit with those changes. If there is another alternate better way to achieve the above, I am open for it :)

sophokles73: FMPOV there is no need to make these configurable as we have defined all those values in the OpenAPI spec ...

PR #2117, opened 2020-08-18
Fix documentation of hono.coap.timeoutToAck property

calohmn:
The documentation had the meaning of the special values for the `hono.coap.timeoutToAck` property mixed up. This is fixed here.

The description in the javadoc has been improved a bit.


site/documentation/content/admin-guide/coap-adapter-config.md
```diff
@@ -46,7 +46,7 @@ The following table provides an overview of the configuration variables and corr
...
-| `HONO_COAP_TIMEOUT_TO_ACK`<br>`--hono.coap.timeoutToAck` | no | 500 | Timeout to send a ACK and separate response. CoAP offers a very efficient piggybacked response, if that could be generated in time. This timeout waits for responding. If it expires without result, a ACK is sent and the response later. If the  result is available in time, it is sent as piggybacked response. Special values: 0, always use piggybacked-response, -1, never use piggybacked-response.|
+| `HONO_COAP_TIMEOUT_TO_ACK`<br>`--hono.coap.timeoutToAck` | no | 500 | Timeout in milliseconds to send an ACK for a CoAP CON request. If the response is available before that timeout, a more efficient piggybacked response is used. If the timeout is reached without having received a response, an ACK is sent and the response is sent later when is has become available. Special values: `-1` means to always use a piggybacked response and never send a separate ACK; `0` means to always send a separate ACK and a separate response. |
```
sophokles73, line 49: > If the timeout is reached without having received a response, an empty ACK is sent back to the client and the response is sent in a separate CON once it becomes available.

> `-1` means to always piggyback the response in an ACK and never send a separate CON; `0` means to always send an ACK immediately and include the response in a separate CON.

calohmn: I've amended the commit accordingly.

PR #2118, opened 2020-08-18
Add JDBC device registry

ctron:
Add an implementation of the device registry using JDBC.


ctron:
I don't think we need any new CQs for this, we have both H2 and PostgreSQL in IPZilla already:

https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21245 (H2 1.4.199)
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22044 (PostgreSQL JDBC 42.2.8)

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/Statement.java
```diff
@@ -202,9 +209,9 @@ public static Statement statement(final String sql, final Object... values) {
         private final Object[] parameters;
 
         private final Tracer tracer;
-        private final Span span;
+        private final SpanContext span;
```
calohmn, line 212: Would be more consistent if "span" got renamed to "spanContext".

core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -149,10 +149,33 @@ public TenantObject addTrustAnchor(final PublicKey publicKey, final X500Principa
...
+
+    /**
+     * Adds a trusted certificate authority to use for authenticating devices of this tenant.
+     *
+     * @param publicKey The CA's public key in encoded form.
+     * @param publicKeyAlgorithm The algorithm of the public key.
+     * @param subjectDn The CA's subject DN.
+     * @param autoProvisioningEnabled A flag indicating whether this CA may be used for automatic provisioning.
+     * @return This tenant for command chaining.
+     * @throws NullPointerException if the public key or subjectDN parameters is {@code null}.
```
sophokles73, line 164: what about the algorithm? seems that we need that too, don't we? At least there seems to be no default value defined ...

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/device/AbstractRegistrationService.java
```diff
@@ -77,19 +78,19 @@ public final void setTenantInformationService(final TenantInformationService ten
...
+     * <ul>
+     * <li><em>200 OK</em>, if a device with the given ID is registered for the tenant.<br>
+     * The <em>payload</em> will contain a JSON object with the following properties:
+     *   <ul>
+     *   <li><em>device-id</em> - the device identifier</li>
+     *   <li><em>data</em> - the information registered for the device</li>
+     *   </ul>
+     * </li>
+     * <li><em>404 Not Found</em>, if no device with the given identifier is registered for the tenant.</li>
+     * </ul>
```
calohmn, line 93: When I perform a "Reformat code" in my IDE using the Eclipse Code Formatter with the "hono-code-style.xml" configuration, there is an indentation here - both for "@params" and "@return". So, IMHO, we should keep the indentation. This also applies to other javadoc changes in this PR.

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/config/JdbcDeviceProperties.java
```diff
@@ -17,14 +17,13 @@
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 
 /**
- *
+ * JDBC properties for the device store.
```
sophokles73, line 20: FMPOV it  would be less misleading if we could rename this class to *JdbcDeviceStoreProperties* ...

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/config/JdbcTenantProperties.java
```diff
@@ -0,0 +1,44 @@
...
+package org.eclipse.hono.service.base.jdbc.config;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+
+/**
+ * JDBC properties for the tenant store.
+ */
+@JsonInclude(value = Include.NON_NULL)
+public class JdbcTenantProperties {
```
sophokles73, line 23: *JdbcTenantStoreProperties*?

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/SQL.java
```diff
@@ -161,7 +161,7 @@ public static Span startSqlSpan(final Tracer tracer, final SpanContext context,
      * @param future The future of the SQL operation.
      * @param span The span to log to.
      * @param extractor An optional way to extract more information to the span. All properties added to
-     *        the provided map will be logged in case of a successful operation.
+     * the provided map will be logged in case of a successful operation.
```
sophokles73, line 164: indentation?

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/SQL.java
```diff
@@ -224,8 +224,8 @@ public static String getDatabaseDialect(final String url) {
      * @param e The throwable to search.
      * @param clazz The class to look for.
      * @return An {@link Optional} with the cause of type {@code <T>}, should that exist in the cause
-     *         chain. Otherwise, or of the throwable provided is {@code null}, {@link Optional#empty()}
-     *         will be returned.
+     * chain. Otherwise, or of the throwable provided is {@code null}, {@link Optional#empty()}
```
sophokles73, line 227: indentation?

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/SQL.java
```diff
@@ -245,9 +245,66 @@ public static String getDatabaseDialect(final String url) {
      * @param e The throwable to search.
      * @param clazz The class to look for.
      * @return {@code true} if the cause chain contains the requested exception, {@code false}
-     *         otherwise. If the throwable to check is {@code null}, false will be returned as well.
+     * otherwise. If the throwable to check is {@code null}, false will be returned as well.
```
sophokles73, line 248: indentation?

services/device-registry-jdbc/src/main/java/org/eclipse/hono/deviceregistry/jdbc/impl/CredentialsManagementServiceImpl.java
```diff
@@ -0,0 +1,106 @@
...
+import org.eclipse.hono.service.management.OperationResult;
+import org.eclipse.hono.service.management.credentials.CommonCredential;
+import org.eclipse.hono.util.CacheDirective;
+
+import io.opentracing.Span;
+import io.vertx.core.Future;
+import io.vertx.core.Vertx;
+
+/**
+ * Implement the <em>credentials management</em> service.
```
calohmn, line 36: "Implementation of the"

sophokles73:
@ctron I have started to take a look but I currently do not find  the time to do this thoroughly for all affected files. It would be very helpful if you could split this up into several (much smaller) PRs ...

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/device/TableAdapterStore.java
```diff
@@ -114,4 +195,52 @@ public TableAdapterStore(final SQLClient client, final Tracer tracer, final Stat
...
+    /**
+     * Resolve a list of group members.
+     *
+     * @param tenantId The tenant the device belongs to.
+     * @param viaGroups The viaGroups list of a device. This list contains the ids of groups.
+     * @param spanContext The span to contribute to.
+     *
+     * @return A future tracking the outcome of the operation.
+     *
+     * @see org.eclipse.hono.deviceregistry.service.device.AbstractRegistrationService#resolveGroupMembers(String, JsonArray, Span)
```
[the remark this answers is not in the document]
ctron: So you want me to remove the javadoc reference? Or make the resolveGroupMembers method public?

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/device/TableAdapterStore.java
```diff
@@ -114,4 +195,52 @@ public TableAdapterStore(final SQLClient client, final Tracer tracer, final Stat
...
+     */
+    public Future<Set<String>> resolveGroupMembers(final String tenantId, final Set<String> viaGroups, final SpanContext spanContext) {
+
+        final Span span = TracingHelper.buildChildSpan(this.tracer, spanContext, "find credentials", getClass().getSimpleName())
+                .withTag("tenant_instance_id", tenantId)
```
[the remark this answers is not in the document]
ctron: The tenant ID gets translated by the `tenantExists` method to the `TenantKey`, which may translate the tenant id, into some internal ID. So it may (or may not) be the original tenant id. Thus the addition of "instance".

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/device/TableAdapterStore.java
```diff
@@ -61,9 +67,82 @@ public TableAdapterStore(final SQLClient client, final Tracer tracer, final Stat
...
+     *
+     * @param key The key of the device to read.
+     * @param spanContext The span to contribute to.
+     *
+     * @return A future, tracking the outcome of the operation.
+     */
+    public Future<Optional<DeviceReadResult>> readDevice(final DeviceKey key, final SpanContext spanContext) {
+
+        final Span span = TracingHelper.buildChildSpan(this.tracer, spanContext, "read device", getClass().getSimpleName())
+                .withTag("tenant_instance_id", key.getTenantId())
```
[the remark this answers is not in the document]
ctron: See above

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/device/TableManagementStore.java
```diff
@@ -122,67 +183,414 @@ public TableManagementStore(final SQLClient client, final Tracer tracer, final S
...
+     * @param spanContext The span to contribute to.
+     * @return A future, tracking the outcome of the operation.
+     */
+    public Future<Versioned<Void>> createDevice(final DeviceKey key, final Device device, final SpanContext spanContext) {
+
+        final String json = Json.encode(device);
+
+        final Span span = TracingHelper.buildChildSpan(this.tracer, spanContext, "create device", getClass().getSimpleName())
+                .withTag("tenant_instance_id", key.getTenantId())
+                .withTag("device_id", key.getDeviceId())
```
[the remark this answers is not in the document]
ctron: See above

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/device/TableManagementStore.java
```diff
@@ -122,67 +183,414 @@ public TableManagementStore(final SQLClient client, final Tracer tracer, final S
...
+     * Delete all devices belonging to the provided tenant.
+     *
+     * @param tenantId The tenant to clean up.
+     * @param spanContext The span to contribute to.
+     * @return A future tracking the outcome of the operation.
+     */
+    public Future<UpdateResult> dropTenant(final String tenantId, final SpanContext spanContext) {
+
+        final Span span = TracingHelper.buildChildSpan(this.tracer, spanContext, "drop tenant", getClass().getSimpleName())
+                .withTag("tenant_instance_id", tenantId)
```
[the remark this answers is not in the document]
ctron: See above

ctron:
> @ctron I have started to take a look but I currently do not find the time to do this thoroughly for all affected files. It would be very helpful if you could split this up into several (much smaller) PRs ...

I do understand. On the other side this makes the discussion of "why the change is needed" more difficult.

What kind of split would you prefer?

ctron:
So I split this up into:

* #2148 – some basics
* #2149 – refactoring of system tests

And left out the other changes, as they build upon those two PRs

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/device/TableAdapterStore.java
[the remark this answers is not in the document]
calohmn: Had a closer look: Since this doesn't actually lead to an error when building Hono with the `javadoc` profile, I think we can keep it that way.

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/device/TableAdapterStore.java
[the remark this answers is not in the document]
calohmn: Having a look at the call hierarchy to this method here, the main path is leading down from `AbstractRegistrationService#assertRegistration`, via the `createSuccessfulRegistrationResult` method:
https://github.com/eclipse/hono/blob/27f06c09b9de76b56e30f48cb5c82949d228c005/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/device/AbstractRegistrationService.java#L125-L146
In that excerpt it looks like the tenant id that is passed on as `createSuccessfulRegistrationResult` parameter and that ends up as `resolveGroupMembers` parameter above is the _original_ (outward-facing) tenantId from _before_ the `tenantExists` invocation. But seeing this id being used in the SQL statement means it probably should really be the internal id?

In any case, if we have methods inside the data store implementation that deal with an _internal_ tenant id, I think we should make that clear by using another variable name like `tenantInternalId` or something like that. `TenantKey.tenantId` and the getter should then be renamed as well IMHO (the same in `DeviceKey` and `CredentialKey`). Otherwise this will lead to confusion like with the above method.

ctron:
Split up even further:

* #2157 – Refactor base classes
* #2158 – Refactor SQL store, add tenant store
* #2159 – Remainder of this PR

ctron:
Superseded by #2159

Issue #2115, opened 2020-08-17
Determine tenant early on in telemetry/event processing


calohmn:
Coming from the discussion in https://github.com/eclipse/hono/pull/2108#issuecomment-672876675:

When processing an incoming telemetry/event message in a protocol adapter, the associated tenant should be determined at the beginning of the message processing and the `TenantObject` should be put in the `ExecutionContext` (e.g. `MqttContext`) for that request . 

This allows to decide early on, which trace sampling mode to use for that request, and for example also allows to check the concurrent connections for that tenant before doing more expensive checks later on in request processing.

While the tenant retrieval is already done early on right now, the `TenantObject` isn't made available via the `ExecutionContext` yet.

PR #2119, opened 2020-08-19
[#2115] Determine tenant early on in request processing (MQTT)

calohmn:
This is for #2115.

When processing MQTT telemetry/event/command-response messages, the tenant is now determined early on, before the authentication check. The `TenantObject` is then made available in the `MqttContext`.


calohmn:
`MqttConnectContext` has been extracted out of the `MqttContext` class here because the context extracted from a CONNECT packet only provided the `MqttEndpoint` value, not all the other `MqttContext` fields, which was misleading.

adapters/mqtt-vertx-base/src/test/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapterTest.java
```diff
@@ -599,7 +651,7 @@ public void testUploadTelemetryMessageFailsForUnknownDevice() {
         final MqttPublishMessage msg = mock(MqttPublishMessage.class);
         when(msg.topicName()).thenReturn("t/my-tenant/unknown");
         when(msg.qosLevel()).thenReturn(MqttQoS.AT_MOST_ONCE);
-
+        when(msg.payload()).thenReturn(Buffer.buffer("hello"));
```
calohmn, line 654: Added here to prevent NPEs thrown (and caught) during the test run.

adapters/mqtt-vertx-base/src/test/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapterTest.java
```diff
@@ -1363,6 +1425,8 @@ public void verifyEventMessageUsesTtlValueGivenInPropertyBag(final VertxTestCont
         forceClientMocksToConnected();
 
         final DownstreamSender sender = mock(DownstreamSender.class);
+        when(sender.sendAndWaitForOutcome(any(Message.class), (SpanContext) any()))
+                .thenReturn(Future.failedFuture("some error"));
```
calohmn, line 1429: Previously, without this, there was an NPE thrown and caught via a Future.compose block here. Now, because of changed Future composition, this wouldn't get caught anymore. Therefore making it explicit that the test expects `sendAndWaitForOutcome` to fail here.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttConnectContext.java
```diff
@@ -0,0 +1,60 @@
...
+    private final MqttEndpoint deviceEndpoint;
+
+    private MqttConnectContext(final MqttEndpoint deviceEndpoint) {
+        this.deviceEndpoint = Objects.requireNonNull(deviceEndpoint);
+    }
+
+    @Override
+    public QoS getRequestedQos() {
+        return QoS.UNKNOWN;
+    }
```
sophokles73, line 37: maybe we can move this method's definition from *ExecutionContext* to *TelemetryExecutionContext*?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttRequestTenantAndAuthIdExtractor.java
```diff
@@ -0,0 +1,96 @@
...
+     * @return A future indicating the outcome of the operation.
+     * @throws NullPointerException if endpoint is {@code null}.
+     */
+    public Future<TenantObjectWithAuthId> getFromMqttEndpoint(final MqttEndpoint endpoint, final SpanContext spanContext) {
+
+        return getFromClientCertificate(endpoint.sslSession(), null, spanContext)
+                .recover(thr -> getFromAuthHeader(endpoint, spanContext));
+    }
+
+    private Future<TenantObjectWithAuthId> getFromAuthHeader(final MqttEndpoint endpoint, final SpanContext spanContext) {
```
sophokles73, line 61: getFromConnectPacket?

sophokles73 (commented):
Can't we get rid of the TenantObjectWithAuthId, MqttRequestTenantAndAuthIdExtractor and the like altogether?
FMPOV we could simply do all that in the factory methods of the protocol specific ExecutionContext implementations, e.g. *MqttConnectExecutionContext.fromConnectPacket()* could also accept a *TenantClientFactory* as a parameter and then store all relevant data in context fields like *authId*, *tenantObject* etc. WDYT?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttRequestTenantAndAuthIdExtractor.java
calohmn: I've used getFromMqttAuth now because the method may also get called in the context of a PUBLISH packet.

calohmn:
I've now changed it so that `tenantObject` and `authId` are fields of the `MqttContext` and are determined in the `MqttContext` factory method.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttContext.java
```diff
@@ -181,22 +229,24 @@ public ResourceIdentifier topic() {
...
-        } else if (topic != null) {
-            return topic.getTenantId();
-        } else {
-            return null;
-        }
+    /**
+     * Gets the authentication identifier used in the MQTT connection.
+     * <p>
+     * In case of an unauthenticated connection, the identifier of the device that published the message is returned
+     * here.
```
sophokles73, line 244: IMHO that is very misleading as we clearly distinguish between authentication identifier and device identifier ...
FMPOV this method should return `null` in that case.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttRequestTenantAndAuthIdExtractor.java
```diff
@@ -0,0 +1,105 @@
...
+     *         the connection attempt to determine the tenant.
+     * @throws NullPointerException if endpoint is {@code null}.
+     */
+    public Future<TenantObjectWithAuthId> getFromMqttEndpoint(final MqttEndpoint endpoint, final SpanContext spanContext) {
+
+        return getFromClientCertificate(endpoint.sslSession(), null, spanContext)
+                .recover(thr -> getFromMqttAuth(endpoint.auth(), spanContext));
+    }
+
+    private Future<TenantObjectWithAuthId> getFromMqttAuth(final MqttAuth auth, final SpanContext spanContext) {
```
sophokles73, line 67: didn't we want to get rid of the dreaded `TenantObjectWithAuthId` class?
How about passing in the context to set the resulting objects on instead?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttContext.java
```diff
@@ -34,92 +38,136 @@
...
      * Creates a new context for a published message.
      *
-     * @param publishedMessage The MQTT message to process.
-     * @param deviceEndpoint The endpoint representing the device
-     *                       that has published the message.
-     * @return The context.
-     * @throws NullPointerException if message or endpoint are {@code null}.
+     * @param publishedMessage The published MQTT message.
+     * @param deviceEndpoint The endpoint representing the device that has published the message.
+     * @param tenantAndAuthIdExtractor The extractor to determine tenant and authentication identifier from the CONNECT
```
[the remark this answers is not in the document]
calohmn: Since the authentication identifier isn't determined from the device, I'd rather use:
````
The extractor to determine tenant and authentication identifier from the
authentication data used during connection establishment or from the topic of the PUBLISH message in case
of an unauthenticated connection.
````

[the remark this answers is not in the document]
Comment (sophokles73) [reply adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttContext.java:75]: In case of an unauthenticated connection there is no *authentication identifier* but only a *device identifier* ...

File: adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java [hunk 8]
```diff
@@ -834,59 +808,55 @@ protected final void onUnsubscribe(
...
-                            span.log("closing connection to device");
-                            context.deviceEndpoint().close();
-                        }
+        MqttContext.fromPublishPacket(message, endpoint, tenantAndAuthIdExtractor, authenticatedDevice, span.context())
+                .compose(context -> {
+                    context.setTimer(getMetrics().startTimer());
+                    applyTenantTracingOptions(context.tenantObject(), context.authId(), span, context);
+
+                    return onPublishedMessage(context)
+                            .recover(thr -> {
```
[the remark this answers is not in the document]
calohmn: The future returned by the `recover` function shall be composed with the `recover` block below (the one with the statusCode handling). Therefore I rather use a proper Future composition instead of one Future with multiple handlers.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -834,59 +808,55 @@ protected final void onUnsubscribe(
...
+                                if (thr instanceof ClientErrorException) {
+                                    // nothing to do
+                                } else {
+                                    onMessageUndeliverable(context);
+                                }
+                                return Future.failedFuture(thr);
+                            }).map(ok -> {
```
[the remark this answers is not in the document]
calohmn: Same as above: the returned Future will be used further down, therefore I rather use a proper Future composition instead of one Future with multiple handlers.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -834,59 +808,55 @@ protected final void onUnsubscribe(
...
+                                Tags.HTTP_STATUS.set(span, HttpURLConnection.HTTP_ACCEPTED);
+                                onMessageSent(context);
+                                return null;
+                            });
+                }).recover(thr -> {
```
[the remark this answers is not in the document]
calohmn: Similar to the above: There's an `onComplete` coming afterwards which has to run after this `recover` case. Therefore I'd rather use composed Futures instead of one Future with multiple handlers.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttContext.java
calohmn: I don't really like the inclusion of the device id case here either.
The motivation here was to also enable using the "sampling-mode-per-auth-id" configuration (defined here:
https://github.com/eclipse/hono/blob/3ebcff93588f0294f1df62da1de1147369577cb9/site/documentation/content/api/management/device-registry-v1.yaml#L859-L868
) in the case of using unauthenticated devices, where then simply the device id would be used for that lookup.

I think we could either add a `MqttContext#deviceId` field then and pass along both ids, or, as this case of using the device id in the "sampling-mode-per-auth-id" lookup isn't documented anyway, we could simply remove this feature, also as we consider using unauthenticated requests as not to be the standard scenario.
I tend to prefer doing the latter, WDYT?

sophokles73: I would also prefer to simply omit the feature for unauthenticated devices (for the time being) ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttRequestTenantAndAuthIdExtractor.java
calohmn: > How about passing in the context to set the resulting objects on instead?

Then the `MqttContext#tenantObject` and `MqttContext#authId` fields couldn't be final any more. And in my view, having such a method in the extractor class that 'enhances' an `MqttContext` like that wouldn't be very straightforward.
Then I'd rather have the `TenantObjectWithAuthId` class and see it only be used in the *Extractor classes.

As an alternative, we could introduce a `Tuple` class in analogy to the Hono `TriTuple` class and use that instead of `TenantObjectWithAuthId`.

calohmn: I've now pushed a change using `Pair<TenantObject, String>` instead of `TenantObjectWithAuthId`. Once all the adapters have been changed like this, the `TenantObjectWithAuthId` class can be removed then.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttContext.java
calohmn: Ok, I've removed the feature in this PR for MQTT then.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttContext.java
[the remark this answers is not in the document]
calohmn: Yes, ok, adapted it accordingly.

service-base/src/main/java/org/eclipse/hono/service/BasicTenantAndAuthIdExtractor.java
```diff
@@ -0,0 +1,192 @@
...
+                tenantId = userComponents[1];
+                authId = userComponents[0];
+            } else {
+                tenantId = null;
+                authId = null;
+            }
+        }
+        if (tenantId == null) {
+            return fail(invalidUserNameExceptionSupplier, "unsupported user name format");
+        }
```
sophokles73, line 164: I have seen this code in at least three places now within Hono:
1. here
1. UsernamePasswordCredentials
1. PreSharedKeyDeviceIdentity (CoAP adapter)

I wonder if this is really necessary ...
How about extending *DeviceCredentials* with support for holding a TenantObject instead of just a tenant ID and returning a *DeviceCredentials* object here instead of a Pair/Tuple?

calohmn: I've resolved the code duplication with `UsernamePasswordCredentials`. The duplication in `PreSharedKeyDeviceIdentity` can be resolved against `UsernamePasswordCredentials` in a separate PR.
Regarding use of _DeviceCredentials_: yes, returning an object of a `TenantDetailsDeviceCredentials` class (or something like that) would be a way to make dealing with the result value more convenient (not having to use pair.one(), pair.two()) and would make it more easily extensible. Have to think about the case where there are no credentials, though (unauthenticated request). I'm preparing an update...

calohmn: @sophokles73 I've pushed an update, also changing the naming in some places (e.g. from `MqttRequestTenantAndAuthIdExtractor` to `MqttRequestTenantDetailsProvider`, its mostly about the tenant details after all).

service-base/src/main/java/org/eclipse/hono/service/auth/device/TenantDetailsDeviceCredentials.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.util.Objects;
+
+import org.eclipse.hono.util.TenantObject;
+import org.eclipse.hono.util.TenantObjectContainer;
+
+/**
+ * Credentials, provided by a device for authentication, along with detailed tenant information.
+ */
+public class TenantDetailsDeviceCredentials extends AbstractDeviceCredentials implements TenantObjectContainer {
```
sophokles73, line 24: why don't we simply add a TenantObject typed field and a corresponding getter to *AbstractDeviceCredentials*?

calohmn: That `tenantObject` field would be `null` in all the places that an `AbstractDeviceCredentials` object is currently used. For the one use case introduced in this PR where the `tenantObject` is not `null`, I'd rather use a separate class where `tenantObject` is always set.

sophokles73: FMPOV this will be the standard use case for all adapters, won't it? We create the (adapter specific) DeviceCredentials in order to authenticate the device and then put its values to the (adapter specific) context. I do not see why we wouldn't *always* include the TenantObject in the credentials ...

calohmn: That's not how I had envisioned it at this point. In the existing Hono code, the DeviceCredentials objects are created for example in `AuthProvider` classes (e.g. `CredentialsApiAuthProvider`) as part of the authentication handling. These `AuthProvider` classes could of course be changed to use the TenantClientFactory there to fetch and add the TenantObject to the DeviceCredentials there. But the DeviceCredentials object created there isn't returned by the `AuthProvider` methods, namely the `authenticate` methods. So these `authenticate` methods would have to be changed there (or the returned `DeviceUser` would have to get a TenantObject as well or something like that).

The whole goal of this PR is to fetch the TenantObject early on, i.e. _before_ the authentication check, so that e.g. a connection limit check can be done before the (expensive) authentication check. But when using an approach with the tenant-enhanced DeviceCredentials in the `AuthProvider` classes, the TenantObject would be fetched as part of the authentication check, not before.
As a way around this, I could imagine using some kind of "preAuthenticationCheckHandler" parameter, used inside the "AuthProvider.authenticate" methods _after_ having fetched the credentials with tenantObject and _before_ calling "getCredentialsForDevice(deviceCredentials)" and "validateCredentials(deviceCredentials". Inside this "preAuthenticationCheckHandler" (given to the AuthProvider from the protocol adapter class) e.g. the connection limit check could be done.

But this would require further changes to the `AuthProvider` classes.
The advantage would of course be to not have the tenantId/authId extraction logic duplicated as it is now, performing it once for getting the TenantObject and then for doing the authentication. I had thought about reusing the UsernamePasswordAuthProvider/X509AuthProvider classes for that, but found this not to be a good solution. Hence my approach to with a more lightweight, pluggable tenantId/authId extraction logic (without certificate check for example) to be used alongside/before the authentication logic.

But changing `AuthProvider` instead and adding such a "preAuthenticationCheckHandler" mechanism for the "authenticate" method, now also sounds like a good solution, probably even a better one. Before doing that I would just want to see agreement, that we want to change the authentication interfaces in such a way. @sophokles73 WDYT?


(As a side note: in the AMQP adapter there is already a "preAuthenticationHandler" mechanism, there it's in the AmqpAdapterSaslAuthenticatorFactory.)

calohmn: Having looked deeper into that, I think for the MQTT adapter we would need to use different, new AuthHandler/AuthProvider methods that don't rely on the credentials being passed around between "parseCredentials()" and "authenticate()" in a JsonObject as intermediate format, with no possibility to add further non-String parameters (like e.g. the ExecutionContext). Otherwise the OpenTracing root span of the current MQTT CONNECT handling can't be made available to the "preAuthenticationCheckHandler".

calohmn: I've created #2140 to prepare the Hono `AuthHandler` for this, using the `ExecutionContext` in the `AuthProvider` `authenticate` method, so that the `ExecutionContext` can be given to the "preAuthenticationCheckHandler".

I now think we can use this in the MQTT and probably also AMQP adapter to get rid of the duplicate "get tenant/authId from auth credentials" logic there. For the HTTP adapter that would not be so straightforward, as that would require overwriting and copying `io.vertx.ext.web.handler.impl.AuthHandlerImpl.handle` to add the "preAuthenticationHandler" hook.

That means I would now go ahead with the `AuthProvider` approach and close this PR (applying parts of it later).

calohmn:
Closing this PR (see https://github.com/eclipse/hono/pull/2119#discussion_r479935170 above).

PR #2122, opened 2020-08-19
[#2112] Add unit tests for search devices operation

kaniyan:


services/device-registry-base/src/test/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementSearchDevicesTest.java
```diff
@@ -0,0 +1,244 @@
...
+import io.opentracing.noop.NoopSpan;
+import io.vertx.core.Future;
+import io.vertx.junit5.VertxTestContext;
+
+/**
+ * As suite of tests for verifying implementations of the Device management's 
+ * search devices operation.
+ * <p>
+ * Concrete subclasses need to provide the service implementations under test
+ * by means of implementing the {@link #getDeviceManagementService()} method.
```
sophokles73, line 35: you should also mention that subclasses need to make sure to remove any fixture in the DB that has been created by individual test cases

services/device-registry-base/src/test/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementSearchDevicesTest.java
```diff
@@ -0,0 +1,244 @@
...
+     * Verifies that a request to search devices fails with a {@value HttpURLConnection#HTTP_NOT_FOUND}
+     * when no matching devices are found.
+     *
+     * @param ctx The vert.x test context.
+     */
+    @Test
+    public void testSearchDevicesWhenNoDevicesAreFound(final VertxTestContext ctx) {
+        final int pageSize = 10;
+        final int pageOffset = 0;
+        final Filter filter = new Filter("/enabled", false);
```
sophokles73, line 65: how about using a random value for an *ext* property in order to make sure that the result set will always be scoped to devices that have this property? This should also safeguard against any side effects that might occur when running test cases concurrently.

sophokles73 (commented):
looks good
how about using a additional random value for an *ext* property in all test cases in order to make sure that the result set will always be scoped to devices that have this property? This should also safeguard against any side effects that might occur when running test cases concurrently.

kaniyan:
> how about using a additional random value for an _ext_ property in all test cases in order to make sure that the result set will always be scoped to devices that have this property?

Good idea. Something like this needed for the concurrent execution of  test cases. Instead of adding a random value for an _ext_ property, how about using a random _tenant identifier_ for devices in each tests. It should also solve the problem. There by the tests need not have an extra filter based on the `ext` property that has random value. WDYT?

kaniyan:
I have pushed a commit containing changes based on above comments.

services/device-registry-base/src/test/java/org/eclipse/hono/service/management/device/AbstractDeviceManagementSearchDevicesTest.java
```diff
@@ -0,0 +1,245 @@
...
+    @Test
+    public void testSearchDevicesWithAFilterSucceeds(final VertxTestContext ctx) {
+        final String tenantId = DeviceRegistryUtils.getUniqueIdentifier();
+        final int pageSize = 10;
+        final int pageOffset = 0;
+        final Filter filter = new Filter("/enabled", true);
+
+        createDevices(tenantId, Map.of(
+                "testDevice1", new Device().setEnabled(true),
+                "testDevice2", new Device().setEnabled(false)))
```
sophokles73, line 87: IMHO we should also use random device IDs in order to minimize the potential for conflicts ...

Issue #2120, opened 2020-08-19
Resource limit checks use wrong resource usage period


sophokles73:
The `PrometheusBasedResourceLimitChecks` calculates the beginning of the most recent resource usage period based on whole days from the exact instant when the resource limits had become effective.
This is wrong in two regards:
1. Every resource usage period after the initial one begins at 12 AM UTC, regardless of the instant at which the limits became effective.
2. When calculating the resources that have already been used up during the current (most recent) resource usage period, a granularity of full days might lead to situations where resources used during the previous period would be accounted for the current period as well, thus preventing clients from using the full amount of resources they are entitled to.

PR #2123, opened 2020-08-20
[#2120] Fix calculation of duration of current accounting period

sophokles73:
The calculation had been using days granularity when determining how
much of the current accounting period's contingent has been used up
already.
If a lot of resources had been used towards the end of the previous
period, this could lead to situations where resources used during
tte previous period would be accounted for the current accounting period
as well.


service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -627,76 +647,146 @@ private Long extractLongValue(final JsonObject response, final Span span) {
...
-                    .between(effectiveSince, lastDayOfMonth) + 1;
-            return Double.valueOf(Math.ceil(daysBetween * configuredLimit / lastDayOfMonth.getDayOfMonth()))
-                    .longValue();
+
+        Objects.requireNonNull(effectiveSince, "effective since");
+        Objects.requireNonNull(targetDateTime, "target date-time");
+        Objects.requireNonNull(periodType, "period mode");
+
+        if (targetDateTime.isBefore(effectiveSince)) {
+            throw new IllegalArgumentException("target date-time must not be before effective since");
```
kaniyan, line 681: FMPOV we should also allow to set a future date as an _effectiveSince_ date. For example, I would like to start a plan from the next month. In this case, I should be able to set it now and not wait until 12 AM UTC on the first day of the next month. IMHO if the effective date is in the future, we can simply ignore and no need to calculate any limits. 

If we don't want to allow any future date as an _effectiveSince_, then I think a corresponding validation should be  there while _adding/updating_ tenant information. WDYT?

kaniyan (changes requested):

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -627,76 +647,146 @@ private Long extractLongValue(final JsonObject response, final Span span) {
...
+                        .withHour(0)
+                        .withMinute(0)
+                        .withSecond(0)
+                        .withNano(0);
+                final long minutesTillStartOfNextAccountingPeriod = Math.max(1, Duration
+                        .between(effectiveSinceZonedDateTime, startOfNextAccountingPeriod)
+                        .toMinutes());
+                final long lengthOfCurrentMonthInMinutes = 60 * 24 * effectiveSinceZonedDateTime
+                        .range(ChronoField.DAY_OF_MONTH).getMaximum();
+                return Double.valueOf(Math.ceil(minutesTillStartOfNextAccountingPeriod * configuredLimit / lengthOfCurrentMonthInMinutes))
```
calohmn, line 704: The result won't be the rounded up division result here because the division is done in an integer context and so the division result is truncated before it gets used in `Math.ceil()`.

core/src/main/java/org/eclipse/hono/util/ResourceLimitsPeriod.java
```diff
@@ -64,11 +64,11 @@ public final int getNoOfDays() {
...
      * @param noOfDays The number of days for which resource usage is calculated.
      * @return  a reference to this for fluent use.
-     * @throws IllegalArgumentException if the number of days is negative.
+     * @throws IllegalArgumentException if the number of days is &lt;= 0.
      */
     public final ResourceLimitsPeriod setNoOfDays(final int noOfDays) {
-        if (noOfDays < 0) {
-            throw new IllegalArgumentException("Number of days property must be  set to value >= 0");
+        if (noOfDays <= 0) {
+            throw new IllegalArgumentException("Number of days property must be  set to value > 0");
```
calohmn, line 71: double whitespace used here

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
sophokles73: very good point :+1:  I will change it accordingly

service-base/src/test/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecksTest.java
```diff
@@ -402,80 +427,110 @@ public void testMessageLimitNotExceededForMissingMetrics(final VertxTestContext
...
         final long maxBytes = 9300;
 
         // Monthly mode
-        // The case where the effectiveSince lies on the past months of the target date.
-        assertEquals(maxBytes,
+        // target date lies within the initial accounting period.
+        final long numberOfMinutesInSeptemer = 24 * 60 * 30;
+        // next accounting period starts Oct 1st 12 AM UTC
+        final long remainingMinutesTillStartOfNextAccountingPeriod = 24 * 60 * 28 + 9 * 60 + 30;
+        final long expectedEffectiveLimit = (long) Math.floor(remainingMinutesTillStartOfNextAccountingPeriod * maxBytes / numberOfMinutesInSeptemer );
```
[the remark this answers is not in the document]
sophokles73: floor() is the wrong one anyway ;-)

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
sophokles73: true, I have changed that

sophokles73:
@calohmn @kaniyan would you mind taking another look?

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
```diff
@@ -678,7 +678,7 @@ long calculateEffectiveLimit(
         Objects.requireNonNull(periodType, "period mode");
 
         if (targetDateTime.isBefore(effectiveSince)) {
-            throw new IllegalArgumentException("target date-time must not be before effective since");
+            return 0;
```
kaniyan, line 681: As per the master branch, the configured _maxBytes_ value is returned when the _effectiveSince_ date lies in the future. Now by returning zero, it means that no messages are accepted for this tenant (as the `allowedMaxBytes` is zero bytes) until the plan actually starts as per the _effectiveSince_ date. I think it makes more sense not to allow any messages until the plan actually starts.

kaniyan (changes requested):

service-base/src/main/java/org/eclipse/hono/service/resourcelimits/PrometheusBasedResourceLimitChecks.java
sophokles73: I agree. That's why I made that change ;-)

kaniyan:
> @calohmn @kaniyan would you mind taking another look?

Looks good to me :+1:

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -930,13 +931,19 @@ components:
          properties:
             "mode":
                type: string
-               description: The mode of the data usage caluclation. The supported modes by
-                            the default resource limit checks implementation are "days"
-                            and "monthly".
+               description: |
+                 The accounting period mode of recurrence. The following modes are defined:
+                 * `monthly`: Accounting periods start at 12 AM UTC on the 1st day of each month.
```
calohmn, line 936: Using `midnight (start of day) UTC` instead of `12 AM UTC` would be clearer and easier to understand here, IMHO. There might be some [confusion](https://en.wikipedia.org/wiki/12-hour_clock#Confusion_at_noon_and_midnight) otherwise.

sophokles73: Yes, I guess that will make it easier to understand for people from Europe ...

sophokles73:
@calohmn I have updated the PR ...

calohmn (approved):
LGTM

PR #2124, opened 2020-08-20
Add metrics sending messages downstream

ctron:
This change adds functionality to capture the duration of "send message" operations. Focusing on the downstream channel. It captures the times from sending the message, to receiving the message disposition.

It also records "timeout" and "queue full" conditions.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/SendMessageSampler.java
+++ b/client/src/main/java/org/eclipse/hono/client/SendMessageSampler.java
@@ -1,0 +1,153 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.client;
+
+import org.apache.qpid.proton.amqp.transport.DeliveryState;
+
+/**
+ * An interface for sampling <em>send message</em> operations.
+ */
+public interface SendMessageSampler {
+
+    /**
+     * A factory for creating samplers.
+     */
+    interface Factory {
+
+        /**
+         * Create a new sampler.
+         *
+         * @param messageType The message type to create a sampler for.
+         * @return A new sampler.
+         */
+        SendMessageSampler create(String messageType);
+
+        /**
+         * Get a default, no-op implementation.
+         *
+         * @return A no-op implementation, never returns {@code null}.
+         */
+        static SendMessageSampler.Factory noop() {
+            return Noop.FACTORY;
+        }
+
+    }
+
+    /**
+     * Get a default, no-op implementation.
+     *
+     * @return A no-op implementation, never returns {@code null}.
+     */
+    static SendMessageSampler noop() {
+        return Noop.SAMPLER;
+    }
+
+    /**
+     * A default no-op implementations.
+     */
+    class Noop {
+
+        private static final Sample SAMPLE = new Sample() {
+
+            @Override
+            public void completed(final String outcome) {
+            }
+
+            @Override
+            public void timeout() {
+            }
+
+        };
+
+        private static final SendMessageSampler SAMPLER = new SendMessageSampler() {
+
+            @Override
+            public Sample start(final String tenantId) {
+                return SAMPLE;
+            }
+
+            @Override
+            public void queueFull(final String tenantId) {
+            }
+
+        };
+
+        private static final Factory FACTORY = new Factory() {
+
+            @Override
+            public SendMessageSampler create(final String messageType) {
+                return SAMPLER;
+            }
+
+        };
+
+        private Noop() {
+        }
+
+    }
+
+    /**
+     * An active sample instance.
+     */
+    interface Sample {
+
+        String OUTCOME_ABORTED = "aborted";
+
+        /**
+         * Call when the message was processed by the remote peer.
+         *
+         * @param outcome The outcome of the message. This is expected to be one of AMQP dispositions.
+         */
+        void completed(String outcome);
+
+        /**
+         * Call when the message was processed by the remote peer.
+         * <p>
+         * This method will use the <em>simple class name</em> of the delivery state parameter. If the
+         * parameter is {@code null} the operating is considered <em>aborted</em> and the value {@link #OUTCOME_ABORTED}
+         * will be used instead.
+         *
+         * @param remoteState The remote state.
+         */
+        default void completed(final DeliveryState remoteState) {
+            if (remoteState == null) {
+                completed(OUTCOME_ABORTED);
+            } else {
+                completed(remoteState.getClass().getSimpleName().toLowerCase());
+            }
+        }
+
+        /**
+         * Call when the operation timed out.
+         */
+        void timeout();
+
+    }
+
+    /**
+     * Start operation.
+     *
+     * @param tenantId The tenant ID to sample for.
+     * @return A sample instance.
+     */
+    Sample start(String tenantId);
+
+    /**
+     * Record a case of "queue full".
+     *
+     * @param tenantId The tenant ID to sample for.
+     */
+    void queueFull(String tenantId);
+
+}
--- a/client/src/main/java/org/eclipse/hono/client/device/amqp/AmqpAdapterClientFactory.java
+++ b/client/src/main/java/org/eclipse/hono/client/device/amqp/AmqpAdapterClientFactory.java
@@ -19,6 +19,7 @@ import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.client.ConnectionLifecycle;
 import org.eclipse.hono.client.HonoConnection;
 import org.eclipse.hono.client.MessageConsumer;
+import org.eclipse.hono.client.SendMessageSampler;
 import org.eclipse.hono.client.impl.AmqpAdapterClientFactoryImpl;
 
 import io.vertx.core.Future;
@@ -33,11 +34,24 @@ public interface AmqpAdapterClientFactory extends ConnectionLifecycle<HonoConnec
      *
      * @param connection The connection to use.
      * @param tenantId The ID of the tenant for which the connection is authenticated.
+     * @param samplerFactory The sampler factory to use.
+     * @return The factory.
+     * @throws NullPointerException if any of the parameters is {@code null}
+     */
+    static AmqpAdapterClientFactory create(final HonoConnection connection, final String tenantId, final SendMessageSampler.Factory samplerFactory) {
+        return new AmqpAdapterClientFactoryImpl(connection, tenantId, samplerFactory);
+    }
+
+    /**
+     * Creates a new factory for an existing connection.
+     *
+     * @param connection The connection to use.
+     * @param tenantId The ID of the tenant for which the connection is authenticated.
      * @return The factory.
      * @throws NullPointerException if any of the parameters is {@code null}
      */
     static AmqpAdapterClientFactory create(final HonoConnection connection, final String tenantId) {
-        return new AmqpAdapterClientFactoryImpl(connection, tenantId);
+        return new AmqpAdapterClientFactoryImpl(connection, tenantId, SendMessageSampler.Factory.noop());
     }
 
     /**
--- a/client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
@@ -31,6 +31,7 @@ import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.client.ClientErrorException;
 import org.eclipse.hono.client.HonoConnection;
 import org.eclipse.hono.client.MessageSender;
+import org.eclipse.hono.client.SendMessageSampler;
 import org.eclipse.hono.client.ServerErrorException;
 import org.eclipse.hono.client.ServiceInvocationException;
 import org.eclipse.hono.config.ClientConfigProperties;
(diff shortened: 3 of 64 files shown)


client/src/main/java/org/eclipse/hono/client/SendMessageSampler.java
```diff
@@ -0,0 +1,141 @@
...
+         * Call when the message was accepted by the remote peer.
+         */
+        void accepted();
+
+        /**
+         * Call when the message was rejected by the remote peer.
+         *
+         * @param errorCode The rejection code. Expected to be a valid HTTP error code.
+         */
+        void rejected(int errorCode);
```
sophokles73, line 118: IMHO *relreased* would also be an interesting state to record ....

client/src/main/java/org/eclipse/hono/client/SendMessageSampler.java
```diff
@@ -0,0 +1,141 @@
...
+     * @param tenantId The tenant ID to sample for.
+     * @return A sample instance.
+     */
+    Sample start(String tenantId);
+
+    /**
+     * Record a case of "queue full".
+     * @param tenantId The tenant ID to sample for.
+     */
+    void queueFull(String tenantId);
```
sophokles73, line 139: does this represent the case when the client is out of credit?

client/src/main/java/org/eclipse/hono/client/impl/DownstreamSenderFactoryImpl.java
```diff
@@ -65,7 +69,7 @@ protected void onDisconnect() {
                 .compose(v -> connection.executeOnContext(result -> {
                     clientFactory.getOrCreateClient(
                             AddressHelper.getTargetAddress(TelemetryConstants.TELEMETRY_ENDPOINT, tenantId, null, connection.getConfig()),
-                            () -> TelemetrySenderImpl.create(connection, tenantId,
+                            () -> TelemetrySenderImpl.create(connection, tenantId, samplerFactory.create("telemetry"),
```
sophokles73, line 72: how about using *TelemetryConstants.TELEMETRY_ENDPOINT* instead of a literal string?

client/src/main/java/org/eclipse/hono/client/impl/DownstreamSenderFactoryImpl.java
```diff
@@ -84,7 +88,7 @@ protected void onDisconnect() {
                 .compose(v -> connection.executeOnContext(result -> {
                     clientFactory.getOrCreateClient(
                             AddressHelper.getTargetAddress(EventConstants.EVENT_ENDPOINT, tenantId, null, connection.getConfig()),
-                            () -> EventSenderImpl.create(connection, tenantId,
+                            () -> EventSenderImpl.create(connection, tenantId, samplerFactory.create("event"),
```
sophokles73, line 91: how about using *EventConstants.EVENT_ENDPOINT* instead of a literal string?

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -81,6 +82,18 @@
...
+     */
+    public static final String METER_DOWNSTREAM_FULL = "hono.downstream.full";
+    /**
+     * The name of the meter for sent messages.
+     */
+    public static final String METER_DOWNSTREAM_SENT = "hono.downstream.sent";
+    /**
+     * The name of the meter for timed out messages.
+     */
+    public static final String METER_DOWNSTREAM_TIMEOUT = "hono.downstream.timeout";
```
sophokles73, line 96: for the other metrics we followed an approach of using a tag to indicate the outcome of the operation. Is there a particular reason why we wouldn't want to follow that approach here as well?

client/src/main/java/org/eclipse/hono/client/SendMessageSampler.java
ctron: I think so. It is derived from the fact that the `sendQueueFull()` method returns `true`, thus the name.

client/src/main/java/org/eclipse/hono/client/SendMessageSampler.java
ctron: That is what the error code was intended to. `Released` would translate into `503`. The again `Modified` would translate into the same status code.

Maybe using the actual name (`Rejected`, …) is a better approach here?!

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
ctron: Yes.  I put a note into the `timeout()` method, maybe there is a better place:

~~~java
/*
 * We report timeouts with a different meter, since the message might still be
 * accepted by the remote peer, at a time after the timeout expired. And so we
 * can still track those times.
 */
~~~

I an nutshell, timeout and "completion" are actually two different events. For timeout we simply decide to, no longer listen to the outcome of the "send message" operation. Still we should get back a disposition, and thus would have the total run time of the operation, which we can log. This would give us an understanding of how long operations actually run.

For the "queue full" situation we don't even start sending. So it doesn't seem to make sense to capture any time.

ctron:
I updated the code, implementing the proposed changes.

client/src/main/java/org/eclipse/hono/client/SendMessageSampler.java
sophokles73: Maybe we can combine both *accepted()* and *rejected()* into one method, e.g. *completed(String outcome)* which takes the delivery state as an argument?

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -431,4 +448,75 @@ private void trackDeviceConnectionDuration(final String tenantId, final long dev
...
+
+                        complete(
+                                "type", messageType,
+                                "success", "true"
+                        );
+
+                    }
+
+                    @Override
+                    public void rejected(final int errorCode) {
```
sophokles73, line 478: if we use a single method *completed(String outcome)* instead of *accepted()* and *rejected(int)* then we can also omit the *success* tag ..

ctron:
Updated.

client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
```diff
@@ -280,6 +292,7 @@ protected final Span startSpan(final Message message) {
                         messageId, getMessageAddress(message));
             } else if (deliveryUpdated.remotelySettled()) {
                 logUpdatedDeliveryState(currentSpan, message, deliveryUpdated);
+                sample.completed(remoteState.getClass().getSimpleName());
```
sophokles73, line 295: :+1:

client/src/main/java/org/eclipse/hono/client/impl/ApplicationClientFactoryImpl.java
```diff
@@ -167,6 +171,7 @@ private void removeCommandClient(final String key) {
                     () -> AsyncCommandClientImpl.create(
                             connection,
                             tenantId,
+                            samplerFactory.create(CommandConstants.NORTHBOUND_COMMAND_REQUEST_ENDPOINT),
```
sophokles73, line 174: is there a particular reason why you only want to add the sampler to the async command client and not the one returned by *getOrCreateCommandClient()*? In general, I wonder if an northbound application client should be required to provide a sampler factory when creating the ApplicationClientFactory. IMHO we should at least also provide a constructor that doesn't require a sampler factory but instead simply uses the Noop one under the hood. WDYT?

client/src/main/java/org/eclipse/hono/client/SendMessageSampler.java
```diff
@@ -0,0 +1,149 @@
...
+         * Call when the message was processed by the remote peer.
+         *
+         * @param outcome The outcome of the message. This is expected to be one of AMQP dispositions.
+         */
+        void completed(String outcome);
+
+        /**
+         * Call when the message was processed by the remote peer.
+         * <p>
+         * This method will also handle the case when the delivery state is {@code null}.
```
sophokles73, line 115: IMHO it would be worthwhile to also document which outcome will be used in this case and we should probably use a constant instead of literal string for that value, shouldn't we?

ctron: Done

ctron:
And rebased.

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -436,4 +453,59 @@ private void trackDeviceConnectionDuration(final String tenantId, final long dev
...
+                         * We report timeouts with a different meter, since the message might still be
+                         * accepted by the remote peer, at a time after the timeout expired. And so we
+                         * can still track those times.
+                         */
+
+                        registry
+                                .counter(
+                                        METER_DOWNSTREAM_TIMEOUT,
+                                        "type", messageType,
+                                        "tenant", tenantId)
```
sophokles73, line 490: *MetricsTags.TAG_TYPE*, *MetricsTags.TAG_TENANT* ?

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -436,4 +453,59 @@ private void trackDeviceConnectionDuration(final String tenantId, final long dev
...
+            }
+
+            @Override
+            public void queueFull(final String tenantId) {
+
+                registry
+                        .counter(
+                                METER_DOWNSTREAM_FULL,
+                                "type", messageType,
+                                "tenant", tenantId)
```
sophokles73, line 505: *MetricsTags.TAG_TYPE*, *MetricsTags.TAG_TENANT* ?

service-base/src/main/java/org/eclipse/hono/service/metric/MicrometerBasedMetrics.java
```diff
@@ -436,4 +453,59 @@ private void trackDeviceConnectionDuration(final String tenantId, final long dev
...
+
+                final Timer.Sample sample = Timer.start(registry);
+
+                return new Sample() {
+
+                    @Override
+                    public void completed(final String outcome) {
+
+                        sample.stop(registry.timer(METER_DOWNSTREAM_SENT,
+                                "type", messageType,
```
sophokles73, line 471: *MetricsTags.TAG_TYPE* ?

client/src/main/java/org/eclipse/hono/client/device/amqp/AmqpAdapterClientFactory.java
```diff
@@ -33,11 +34,12 @@
      *
      * @param connection The connection to use.
      * @param tenantId The ID of the tenant for which the connection is authenticated.
+     * @param samplerFactory The sampler factory to use.
      * @return The factory.
      * @throws NullPointerException if any of the parameters is {@code null}
      */
-    static AmqpAdapterClientFactory create(final HonoConnection connection, final String tenantId) {
-        return new AmqpAdapterClientFactoryImpl(connection, tenantId);
+    static AmqpAdapterClientFactory create(final HonoConnection connection, final String tenantId, final SendMessageSampler.Factory samplerFactory) {
```
sophokles73, line 41: IMHO we should also provide a factory method that doesn't require a sampler factory because AMQP adapter clients will usually be external to a Hono instance and thus most likely not integrate with Hono's metrics backend. WDYT?

ctron:
Updated.

sophokles73 (commented):
looks good, I think the new metrics should also be added to the Metrics API spec. WDYT?

ctron:
Added some docs

site/documentation/content/api/Metrics.md
```diff
@@ -65,7 +65,8 @@ Additional tags for protocol adapters are:
...
+| *outcome* (for *hono.downstream.sent*) | *string* (AMQP 1.0 disposition) | Any of the AMQP 1.0 disposition states, as well as `Aborted`, in the case the connection/link was closed before the disposition could be read. |
```
sophokles73, line 69: I find it quite confusing to have two descriptions of the same tag name here. Maybe we should structure the metrics documentation differently so that we describe the tags in the context of the individual metrics instead of having a single place for all tags? We could also keep a list of shared tags which includes that standard tags which have the same semantics across all metrics. The descriptions of the metrics could then refer to them but could also define other tags in the context of the particular metric (e.g. the *outcome* tag). WDYT?

ctron: Like this?

site/documentation/content/api/Metrics.md
```diff
@@ -65,8 +65,19 @@ Additional tags for protocol adapters are:
...
 
+Additional tags for *hono.downstream.sent*:
+
+| Name        | Value                                              | Description |
+| ----------- | -------------------------------------------------- | ----------- |
+| *outcome*   | *string* (AMQP 1.0 disposition) | Any of the AMQP 1.0 disposition states, as well as `Aborted`, in the case the connection/link was closed before the disposition could be read. |
```
sophokles73, line 79: I think this is better. However, it is unclear, which values are actually allowed here.
The AMQP 1.0 spec defines the delivery states as `received`, `accepted`, `rejected`, `released` and `modified`.
The code currently uses the simple class names, i.e. starting with a capital letter, plus the non-AMQP state `Aborted`.
For the other tags we only use all lowercase values so far so I would prefer to stick to that pattern and explicitly list the possible values here. WDYT?

ctron: It would have been more efficient to do this right away, I will amend the PR … anything else?

ctron:
updated

site/documentation/content/api/Metrics.md
sophokles73: No, that should be it FMPOV. Sorry for the inconvenience ...

ctron:
Is this ready to merge now?

sophokles73 (approved):
LGTM

PR #2128, opened 2020-08-24
Use Vert.x WebClient API in integration tests

Alfusainey:
The `HttpClient#handler` API has been deprecated and the deprecation warning suppressed in revision ed756f8f. This patch replaces the usage of the `HttpClient` in the integration tests module with the new `WebClient` API 


kaniyan:
`HttpClient#handler` deprecation is indication of change in the method signature from Vert.x 4. But I don't think that's the only reason we should switch to the `Webclient`. I am in favour of using `Webclient` as it makes request/response handling much easier and also offers methods to manage query parameters. It would be quite useful for IT tests for _search devices operation_ where the request contains several query parameters. @sophokles73 WDYT?

sophokles73 (approved):
LGTM
Thanks for contributing, @Alfusainey :+1:

PR #2129, opened 2020-08-25
Minor fixes and improvements in device registry integration tests

kaniyan:


tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsManagementIT.java
```diff
@@ -81,10 +81,10 @@
         ORIG_BCRYPT_PWD = encoder.encode("thePassword");
     }
 
+    private final AtomicReference<String> resourceVersion = new AtomicReference<>();
```
sophokles73, line 84: this would probably better be initialized as part of the `@BeforeEach` method ...

kaniyan: Done.

sophokles73 (approved):
LGTM

PR #2132, opened 2020-08-26
Update spec. Add missing information.

ctron:
This corrects a note, making it more explicit that a tenant must exist
when creating a device. It also lists the error code to report in such
cases.

### DIFF
--- a/site/documentation/content/api/management/device-registry-v1.yaml
+++ b/site/documentation/content/api/management/device-registry-v1.yaml
@@ -323,6 +323,8 @@ paths:
                $ref: '#/components/responses/Unauthorized'
             403:
                $ref: '#/components/responses/NotAllowed'
+            404:
+               $ref: '#/components/responses/TenantNotFound'
 
       get:
          tags:
@@ -359,7 +361,8 @@ paths:
             - devices
          summary: Create a new device registration
          description: |
-            Clients use this operation to register a new device with a given identifier.
+            Clients use this operation to register a new device for an existing tenant with
+            a given identifier
          operationId: createDeviceRegistrationWithId
          requestBody:
             description: The configuration properties to register for the device.
@@ -384,6 +387,8 @@ paths:
                $ref: '#/components/responses/NotAllowed'
             409:
                $ref: '#/components/responses/AlreadyExists'
+            404:
+               $ref: '#/components/responses/TenantNotFound'
 
       get:
          tags:
@@ -550,7 +555,7 @@ paths:
             401:
                $ref: '#/components/responses/Unauthorized'
             404:
-               $ref: '#/components/responses/NotFound'
+               $ref: '#/components/responses/DeviceNotFound'
 
       put:
          tags:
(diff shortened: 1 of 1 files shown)


sophokles73:
From the very first version of the file based registry we didn't want to make this check mandatory. It still makes sense to add the corresponding response status codes to the spec. However, we should also discuss if we actually want the corresponding check(s) to be mandatory to implement. So far, they weren't.

ctron:
Well I think the file based registry is something for playing around. In the meantime we added the Tenant Information service, which actually provide the layer to validate this. Assuming the file base devices registry will be gone in the not so distant future, I don't think it makes sense to add all this new stuff to it.

And I really don't think there is a good story having the Hono system tests enforcing the behavior that you must be able to create devices without a tenant … providing a tenant ID in the process. I think that makes no sense at all.

dejanb (approved):
LGTM

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -270,8 +270,9 @@ paths:
          summary: Delete tenant
          description: |
             Clients use this operation to remove an existing tenant from the registry.
-            Registry implementations may optionally perform a *cascading* delete of all
-            the tenant's devices and credentials as well.
+
+            When the operation succeeds, the data owned by this tenant (including devices and their
+            credentials) must not longer be accessible. These entities should be reported as *not found*.
```
kaniyan, line 275: > When the operation succeeds, the data owned by this tenant (including devices and their
> credentials) must not longer be accessible. These entities should be reported as *not found*.

With the above comment, I understood that the tenant existence must be checked not only during _create device_ operation but for all device/credentials management (HTTP) operations and device/credentials services (AMQP). Shouldn't we specify the outcome with the corresponding response code (404) under all those operations in the spec?

If we are checking for tenant existence even during _delete device_ operation, then it is no more possible to delete the devices of a tenant that has been already deleted. Does it mean, in order to clean up the devices, the tenant need to be registered again? May be we should allow deleting devices even though the tenant doesn't exist. WDYT?

kaniyan (changes requested):

site/documentation/content/api/management/device-registry-v1.yaml
sophokles73: How about
> When the operation succeeds, clients must no longer be able to retrieve nor update the devices (including their credentials) that belong to the deleted tenant. Implementations of this API should respond with a *404 Not Found* to any such request.

This would make it more explicit that we do not want such data to be used any longer but still allow for implementations to support deleting such entities.

ctron: So aside from the spec: the tenant information should always be used, in order to look up the tenants, and use the tenant id that it provides in `TenantKey`, because that maybe been rewritten by the backing service to generate unique ID.

I don't see much reason here to make "deleting" a special operation in the public spec. Internally we can handle this differently. We can provide a privileged operation "clear tenant". For JDBC this can be handled: a) with a foreign key constraint b) with a separate process executing the "delete" operation.

So in the case of Kubernetes, the tenant will be deleted using the Kubernetes API, and that will trigger `b)`. Also would the internal tenant ID be the Kubernetes UID, to ensure uniqueness of the tenant.

I would propose that the MongoDB follows the same strategy. If the MongoDB registry instance is configured for both Tenant + Devices, then you can take an internal shortcut to delete tenant data when the tenant gets deleted. Otherwise there is specialized operation for cleaning tenant data.

sophokles73 (approved):
LGTM

PR #2136, opened 2020-08-27
Fix sorting of devices during search devices operation in MongoDB device registry

kaniyan:
The issue while converting the sorting field of type _JsonPointer_ to _MongoDB_ specific field in `MongoDbDocumentBuilder.withDeviceSortOptions(...)` is fixed with this PR. Also the corresponding test is updated.

### DIFF
--- a/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDocumentBuilder.java
+++ b/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDocumentBuilder.java
@@ -143,12 +143,7 @@ public final class MongoDbDocumentBuilder {
 
         filters.forEach(filter -> {
             // TODO: To implement when filter values contain patterns such as * or %
-            if (FIELD_ID.equals(filter.getField())) {
-                document.put(RegistryManagementConstants.FIELD_PAYLOAD_DEVICE_ID, filter.getValue());
-            } else {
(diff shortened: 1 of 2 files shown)


services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDocumentBuilder.java
```diff
@@ -143,12 +143,7 @@ public MongoDbDocumentBuilder withDeviceFilters(final List<Filter> filters) {
 
         filters.forEach(filter -> {
             // TODO: To implement when filter values contain patterns such as * or %
-            if (FIELD_ID.equals(filter.getField())) {
-                document.put(RegistryManagementConstants.FIELD_PAYLOAD_DEVICE_ID, filter.getValue());
-            } else {
-                document.put(MongoDbDeviceRegistryUtils.FIELD_DEVICE + filter.getField().toString().replace("/", "."),
-                        filter.getValue());
-            }
+            document.put(mapDeviceField(filter.getField()), filter.getValue());
```
calohmn, line 146: How about adding an "/id" filter to the `testSearchDevicesWithMultipleFiltersSucceeds` test, so that this code path is covered by tests as well.

kaniyan: I have added an`/id` filter to that test.

calohmn (approved):
LGTM

Issue #2135, opened 2020-08-27
Deprecate/Remove "single tenant" mode


calohmn:
The configuration options for a protocol adapter or for different services in a Hono component in general support a **single tenant** mode:
https://github.com/eclipse/hono/blob/a2fb22b4ffb8aee26b2e54ec8096d5ac2f07f5d4/core/src/main/java/org/eclipse/hono/config/ServiceConfigProperties.java#L96-L107

This setting is currently not documented in the [Protocol Adapter options](https://www.eclipse.org/hono/docs/admin-guide/common-config/#protocol-adapter-options) or anywhere else in the documentation.

The effect of this flag currently seems to be that the tenant can be omitted in the credentials (certificate or username/password) presented by a device, so that "DEFAULT_TENANT" will be taken instead. Besides that, Hono's AMQP services seem to add "DEFAULT_TENANT" to the address when having received requests to open sender/receiver links. (The corresponding logic of omitting the tenant on the part that opens the links seems to be missing though (?).)

Since it's questionable whether this feature works (it's also not being tested by integration tests) and since Hono can be used with just one tenant by configuring only one tenant, this configuration option should better be declared deprecated (Hono 1.4) and eventually be removed (1.5).

PR #2139, opened 2020-08-28
[#2135] Mark single-tenant mode methods as deprecated

calohmn:
This is the first step for #2135.


sophokles73 (commented):
How about adding a corresponding entry in the release notes?

calohmn:
@sophokles73 I've added an entry in the release notes.

sophokles73 (approved):
LGTM

PR #2140, opened 2020-08-31
Pass on execution context from Hono AuthHandler to AuthProvider

calohmn:
The Hono `AuthHandler` (in package `org.eclipse.hono.service.auth.device`) now uses an AuthProvider that also supports working with an `ExecutionContext`. This makes `TraceContext` serialization/deserialization unnecessary in scenarios where a
`HonoClientBasedAuthProvider` is used with such a Hono `AuthHandler`.

This is a preparation for #2115.

As a next step, `HonoClientBasedAuthProvider<T>` should better be renamed to the more generic `DeviceCredentialsAuthProvider<T>` (`ExecutionContextDeviceCredentialsAuthProvider<T>` would be more specific but a bit long).


sophokles73 (approved):
LGTM

Issue #2131, opened 2020-08-26
MongoDB implementation does not check for existing tenant


ctron:
The current mongodb implementation does not check for an existing tenant. It allows to create devices, without checking for the existence of a tenant first.

It should however use the `TenantInformationService` to ensure that a tenant exists.

Also, many of our tests currently do not set up a tenant, before registering a device.

kaniyan:
IMHO I don't see it as a bug. The MongoDB device registry is implemented as per the OpenAPI spec. We don't have this tenant check specified in our Device Management API spec. I guess for the same reason we don't have corresponding tests. IMHO if we like to bring this tenant check, then its better to start with the OpenAPI spec. So that we can also define the appropriate response codes for the operations. Currently if the tenant doesn't exist, the `AbstractDeviceManagementService` returns 404. But 404 is not defined as one of the response codes for many device operations. I think its better to create an issue for bringing in tenant check feature and start with the OpenAPI spec. And also the same is applicable for the credentials operations. @calohmn @ctron @dejanb @sophokles73 WDYT?

ctron:
As mentioned in the chat, I disagree:

The spec says:

> Clients use this operation to register a new device for an existing tenant with an identifier that is generated by the registry.

The second create call misses the explicit wording on "existing", which we will fix by updated the spec.

The missing items in the spec have been addresses with: https://github.com/eclipse/hono/pull/2132

kaniyan:
As I also wrote in the chat, I differ in opinion as the spec is inconsistent and also didn't mention what is expected if a tenant doesn't exist. With no existing tests so far and also from the discussion in #2132, I see that this check is not mandatory yet. Its good that you have raised that PR #2132 to fix the spec and include the expected response codes.

sophokles73:
@ctron can we close this based on #2141 ?

ctron:
> @ctron can we close this based on #2141 ?

To me it looks like this only fixes the registration, but not the credentials.

kaniyan:
> > @ctron can we close this based on #2141 ?
> 
> To me it looks like this only fixes the registration, but not the credentials.

The `MongoDbBasedCredentialsService` extends `AbstractCredentialsManagementService` where it uses the tenant informatiomn service to check if a tenant exists or not. Hence #2141 also fixes the credentials part too.

ctron:
> > > @ctron can we close this based on #2141 ?
> > 
> > 
> > To me it looks like this only fixes the registration, but not the credentials.
[...]

PR #2141, opened 2020-08-31
[#2131] Check for tenant existence in device registration operations in MongoDB device registry

kaniyan:
With this PR, while registering devices, the existence of the tenant is now being verified using `TenantInformationService`.

### DIFF
--- a/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
+++ b/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
@@ -20,6 +20,7 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 
 import org.eclipse.hono.client.ClientErrorException;
+import org.eclipse.hono.client.StatusCodeMapper;
 import org.eclipse.hono.deviceregistry.mongodb.config.MongoDbBasedRegistrationConfigProperties;
 import org.eclipse.hono.deviceregistry.mongodb.model.DeviceDto;
 import org.eclipse.hono.deviceregistry.mongodb.utils.MongoDbCallExecutor;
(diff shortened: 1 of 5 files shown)


services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -508,4 +521,12 @@ private static JsonArray getSearchDevicesAggregatePipelineQuery(final String ten
                     return Future.succeededFuture();
                 });
     }
+
+    private Future<Void> tenantExists(final String tenantId, final Span span) {
+
+        return tenantInformationService.tenantExists(tenantId, span)
+                .compose(result -> result.isError()
+                        ? Future.failedFuture(new ClientErrorException(result.getStatus()))
```
sophokles73, line 529: IMHO you should not assume the status code to be in the [400, 500) range here. Instead, use *StatusCodeMapper.from(int, String)* which takes care of this ...
For readability I would also explicitly check for OK instead, i.e.
```java
result.isOK() ? Future.succeededFuture() : StatusCodeMapper.from(result.getStatus(), null);
```

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -212,12 +221,15 @@ public MongoDbBasedRegistrationService(
         Objects.requireNonNull(deviceKey);
         Objects.requireNonNull(span);
 
-        return findDeviceDocument(deviceKey.getTenantId(), deviceKey.getDeviceId())
+        return tenantExists(deviceKey.getTenantId(), span)
```
sophokles73, line 224: I do not think that this is necessary because the assert Registration operation explicitly defines how to handle the case when there is no device matching the criteria ...

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
```diff
@@ -227,7 +239,8 @@ public MongoDbBasedRegistrationService(
         Objects.requireNonNull(viaGroups);
         Objects.requireNonNull(span);
 
-        return processResolveGroupMembers(tenantId, viaGroups, span);
+        return tenantExists(tenantId, span)
```
sophokles73, line 242: this is an internal operation that is invoked as part of one of the public API operations. The existence of the tenant should already have been checked there ...

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedRegistrationService.java
kaniyan: Thanks for pointing it out. I have changed it now and pushed a commit addressing all the comments.

sophokles73 (approved):
LGTM

Issue #2125, opened 2020-08-20
Allow to configure usage of piggybacked response for CoAP adapter on the tenant level


alinaserg:
The CoAP adapter can send response to the device in the piggybacked mode. And currently this parameter is configurable only on the adapter level.

We would like to have a possibility to configure this parameter additionally on the tenant level, being flexible in the usage of such functionality.

sophokles73:
I would propose to define a corresponding property in the *adapters* property of the *Tenant* object in the Registry Management API, e.g.
```json
{
  "adapters": [{
    "type": "coap",
    "enabled": true,
    "ext": {
      "timeoutToAck": 0
    }
  }]
}
```
The CoAP adapter would then use the config value specified at the adapter level by default with the value specified at the tenant level overriding that default value.

@alinaserg @calohmn @boaks WDYT?

alinaserg:
@sophokles73, thanks for the proposal. For me it sounds reasonable to proceed as you described.

PR #2145, opened 2020-09-01
[#2125] Support tenant specific configuration of timeoutToAck

sophokles73:
The CoAP adapter now supports the configiuration of the timeoutToAck
property at the tenant level. If not specified, the value set at the
adapter level is used.


adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapContext.java
```diff
@@ -158,24 +159,37 @@ public String getAuthId() {
...
     public void startAcceptTimer(
             final Vertx vertx,
+            final TenantObject tenant,
             final long timeoutMillis) {
 
+        Objects.requireNonNull(vertx, "vert.x");
+        Objects.requireNonNull("tenant");
+
+        final long ackTimeout = Optional.ofNullable(tenant.getAdapter(Constants.PROTOCOL_ADAPTER_TYPE_COAP))
+                .map(config -> (Long) config.getExtensions().get(CoapConstants.TIMEOUT_TO_ACK))
```
calohmn, line 181: Given a JSON string with the "timeoutToAck" value set to something <= `Integer.MAX_VALUE`, this will lead to a `ClassCastException` because the Map value will be an `Integer`, not a `Long`. Using the primitive `long` type for the cast would work.

Besides that, I think it would be better to deal with a possible configuration error here, the "timeoutToAck" value being defined as a `String` for example. For the standard TenantObject fields, we deal with that gracefully, catching ClassCastExceptions via `JsonHelper.getValue`. I think for extension values, we should do the same, so that there can't be any ClassCastException thrown here.

site/homepage/content/release-notes.md
```diff
@@ -15,7 +15,10 @@ title = "Release Notes"
 * The LoraWAN protocol adapter has been extended with support for the *Actility Enterprise* provider.
 * The LoraWAN protocol adapter has been extended with support for the *Orbiwise* provider.
 * Added metrics for tracking the RTT between sending an AMQP message to receiving the disposition.
-
+* The CoAP adapter now supports configuration of the *timeoutToAck* parameter at the tenant level
+  in addition to the adapter level. If set, the tenant specific value is used for all devices of the
+  particular tenant. if not set, the value defined at the adapter level is used.
```
calohmn, line 20: "If" with uppercase "i".

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CoapContext.java
```diff
@@ -158,24 +159,37 @@ public String getAuthId() {
...
      * <p>
+     * This method first tries to find a {@value CoapConstants#TIMEOUT_TO_ACK} property value
+     * in the tenant's CoAP adapter configuration.
+     * If no such property is found, the value given in the timeoutMillis parameter is used instead.
      *
-     * @param vertx Vertx to schedule the timer
+     * @param vertx The vert.x instance to set the timer on.
+     * @param tenant The tenant that the device belongs to which the request being processed originates from.
      * @param timeoutMillis The number of milliseconds to wait for a separate ACK. {@code -1}, never use separate
      *            response, {@code 0}, always use separate response.
```
calohmn, line 169: "[...] to wait for a separate ACK, if not configured for the tenant."

sophokles73:
@calohmn I have pushed some fixes ... can you take another look?

calohmn (approved):
LGTM

PR #2146, opened 2020-09-01
[#2144] Fix "authenticate device" span having no parent

### LINKED ISSUE #2144: HTTP adapter: "authenticate device" span has no parent (certificate authentication)
For an incoming HTTP request using certificate authentication, the HTTP adapter creates an "authenticate device" span that doesn't have the corresponding "POST" (or "PUT", etc.) span as its parent.
That means the "authenticate device" span is not part of the whole HTTP request trace.

calohmn:
This fixes #2144.


sophokles73 (approved):
LGTM

PR #2151, opened 2020-09-03
Set 'cache_hit' tag consistently

calohmn:
For the resource limit checks, the `cache_hit` tracing tag is currently only set to `false`, i.e. it is only set in case there was no cache hit.
This PR now sets the `cache_hit` tag consistently in all cases.


sophokles73 (approved):
LGTM

PR #2153, opened 2020-09-04
Improve Command & Control related tracing.

calohmn:
The HTTP adapter now tracks the reception of a command in a separate "wait for command" span.
This allows for easier filtering in Jaeger for HTTP requests that have opened a command consumer.
In this span the reception of the command message is now logged explicitly, with details of the command message.

For the HTTP adapter, there are generally 2 separate traces involved in processing a command:
- The trace for the HTTP request from the device
- The trace that gets started when the command message initially gets received in Hono. (This trace is always separate because at the point of starting it, there is no association to the HTTP request from the device yet.)

This PR adds logging events to these two traces that were previously only set in one of them.

(Corresponding changes to the CoAP adapter will be done in a separate PR.)

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -32,6 +32,7 @@ import org.eclipse.hono.client.CommandContext;
 import org.eclipse.hono.client.CommandResponse;
 import org.eclipse.hono.client.DownstreamSender;
 import org.eclipse.hono.client.ProtocolAdapterCommandConsumer;
+import org.eclipse.hono.client.impl.CommandConsumer;
 import org.eclipse.hono.service.AbstractProtocolAdapterBase;
 import org.eclipse.hono.service.http.ComponentMetaDataDecorator;
 import org.eclipse.hono.service.http.DefaultFailureHandler;
@@ -55,6 +56,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 
 import io.micrometer.core.instrument.Timer.Sample;
 import io.opentracing.Span;
+import io.opentracing.SpanContext;
 import io.opentracing.tag.Tags;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.CompositeFuture;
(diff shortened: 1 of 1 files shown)


calohmn:
Example trace:
HTTP_cmd_trace

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -975,24 +974,34 @@ protected void setNonEmptyResponsePayload(final HttpServerResponse response, fin
...
+                    CommandConsumer.logReceivedCommandToSpan(command, waitForCommandSpan);
+                    checkMessageLimit(tenantObject, command.getPayloadSize(), waitForCommandSpan.context())
                     .onComplete(result -> {
                         if (result.succeeded()) {
                             addMicrometerSample(commandContext, commandSample);
                             // put command context to routing context and notify
                             ctx.put(CommandContext.KEY_COMMAND_CONTEXT, commandContext);
                         } else {
                             commandContext.reject(getErrorCondition(result.cause()));
+                            waitForCommandSpan.log("rejected command for device: " + result.cause());
```
sophokles73, line 1004: FMPOV the cause (Throwable) should be logged here as intended by the OpenTracing spec, i.e. using *Fields.ERROR_KIND* and *Fields.ERROR_OBJECT* to be consistent with other cases where we use *TracingHelper.log(Throwable)*. We might want to consider adding *Fields.STACK* to that method as well ...

calohmn: I think the stack is not important here. The log message is supposed to contain `result.cause().toString()`.

Concerning the "ERROR_KIND" and "ERROR_OBJECT" fields: These would of course be set if  `Tracing.logError` was used here. I didn't want to use this because I didn't want the HTTP PUT span to be marked as an error span in this case (that wasn't done so before).

But thinking again, I think this case should indeed be marked as an error span, reflecting the "check message limit failed" error. I have adapted the commit accordingly, also setting an error for the "received invalid command message" case.

ctron (approved):
Looks reasonable to me

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -1020,6 +1029,9 @@ protected void setNonEmptyResponsePayload(final HttpServerResponse response, fin
                 }
 
             } else {
+                if (!requestProcessed.get()) {
+                    CommandConsumer.logReceivedCommandToSpan(command, waitForCommandSpan);
```
sophokles73, line 1033: The command is invalid in this case, isn't it? Shouldn't this result in an error being logged to the span?

calohmn: Yes, `logReceivedCommandToSpan` does  that.

sophokles73: I see. Then FMPOV we should pull up the CommandConsumer.logReceivedCommandToSpan(command, waitForCommandSpan) invocation before the `if (isCommandValid(command, waitForCommandSpan))` statement and explicitly log the fact that the request has already been processed and the command is being released. WDYT?

calohmn: I've amended the commit accordingly.

sophokles73 (approved):
LGTM

PR #2154, opened 2020-09-04
Let removeCommandHandlingAdapterInstance return 412 again

calohmn:
The recent change towards returning 404 for the "Remove command-handling adapter instance" operation (see #2066) didn't
match the most recent API definition as introduced in #1940. Therefore reverting that change here and again returning 412 instead of 404.

See the corresponding comment in #2066.


sophokles73 (approved):
LGTM

PR #2157, opened 2020-09-07
Refactor base classes to allow other implementations (JDBC)

ctron:


services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/CommonCredentials.java
```diff
@@ -0,0 +1,66 @@
...
+package org.eclipse.hono.service.management.credentials;
+
+import java.util.Collection;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * Helper methods for {@link CommonCredential}.
+ */
+public final class CommonCredentials {
```
sophokles73, line 23: how about simply adding these static methods to the existing *CommonCredential* class?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/AutoProvisioning.java
```diff
@@ -0,0 +1,107 @@
...
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     */
+    public static Future<OperationResult<String>> provisionDevice(
+            final DeviceManagementService deviceManagementService,
+            final CredentialsManagementService credentialsManagementService,
+            final String tenantId,
+            final X509Certificate clientCertificate,
+            final Span span) {
+
+        Objects.requireNonNull(tenantId);
```
sophokles73, line 65: maybe we should add non-null checks for the services as well?

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/device/AbstractRegistrationService.java
```diff
@@ -103,11 +104,39 @@ public final void setTenantInformationService(final TenantInformationService ten
...
      *            parent for any spans created in this method.
-     * @return A future indicating the outcome of the operation. A failed future with 
+     * @return A future indicating the outcome of the operation. A failed future with
      *           a {@link ServiceInvocationException} is returned, if there was an error resolving the groups.
      * @throws NullPointerException if any of the parameters is {@code null}.
      */
-    protected abstract Future<JsonArray> resolveGroupMembers(String tenantId, JsonArray viaGroups, Span span);
+    Future<JsonArray> resolveGroupMembers(final String tenantId, final JsonArray viaGroups, final Span span) {
+
+        final var viaGroupsAsString = viaGroups.stream()
```
sophokles73, line 113: how about adding null checks?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/CommonCredentials.java
ctron: I was following the pattern of splitting actual objects and helper classes apart, like `java.util.Collections`, `java.lang.Objects`, `com.google.collect.Streams`, …

ctron:
Added null checks.

kaniyan (changes requested):

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsManagementService.java
```diff
@@ -41,23 +43,34 @@
...
+    public AbstractCredentialsManagementService(
+            final Vertx vertx,
+            final HonoPasswordEncoder passwordEncoder,
+            final int maxBcryptIteration,
+            final Set<String> hashAlgorithmsWhitelist) {
+
         this.vertx = Objects.requireNonNull(vertx);
         this.passwordEncoder = Objects.requireNonNull(passwordEncoder);
+        this.maxBcryptIteration = maxBcryptIteration;
+        this.hashAlgorithmsWhitelist = hashAlgorithmsWhitelist;
```
kaniyan, line 72: Null check is missing for `hashAlgorithmsWhitelist`. `hashAlgorithmsWhitelist` is passed as an argumet to 
 `DeviceRegistryUtils.checkCredential(...)` and it does not allow _null_ value for this argument.

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsService.java
```diff
@@ -32,8 +44,13 @@
  */
 public abstract class AbstractCredentialsService implements CredentialsService {
 
+    private static final Logger log = LoggerFactory.getLogger(AbstractCredentialsService.class);
```
kaniyan, line 47: How about naming this static variable in upper case?

kaniyan (changes requested):

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsService.java
ctron: First of all, it is not a constant, in the sense of a scalar value constant. Second, we use lowercase `log` in all other areas of Hono, so I would leave it as is.

ctron:
Added javadoc about null and added "optional" handling for the allowed hash algorithm list

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsManagementService.java
```diff
@@ -43,21 +46,31 @@
...
-    public AbstractCredentialsManagementService(final Vertx vertx, final HonoPasswordEncoder passwordEncoder) {
+    public AbstractCredentialsManagementService(
+            final Vertx vertx,
+            final HonoPasswordEncoder passwordEncoder,
+            final int maxBcryptIteration,
+            final Set<String> hashAlgorithmsWhitelist) {
+
         this.vertx = Objects.requireNonNull(vertx);
         this.passwordEncoder = Objects.requireNonNull(passwordEncoder);
+        this.maxBcryptIteration = maxBcryptIteration;
```
sophokles73, line 71: FMPOV we should also make sure that the iterations are within the supported value range, in our case [4, 31].
Having read throught the bcrypt documentation again, it seems like the parameter should better be called *maxBcryptLogRounds* as the number of actual iterations is computed as 2^logRounds ...

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/CommonCredentials.java
sophokles73: I see

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsManagementService.java
ctron: I would say this is already covered by the configuration, which explicitly checks for that. Also would renaming mean, that we should do that consistently throughout the project. Not only here.

So I would say, we keep this PR focused on the current items. And create a new one later for renaming that parameter, which has an impact on documentation and configuration as well.

sophokles73: fine with me

sophokles73 (commented):
this looks good to me. However, I do not think that we will be able to properly review the JDBC implementation in time for 1.4.0 so FMPOV there is no need to rush this into 1.4.0 as well. WDYT? Can this wait till 1.5.0?

ctron:
> this looks good to me. However, I do not think that we will be able to properly review the JDBC implementation in time for 1.4.0 so FMPOV there is no need to rush this into 1.4.0 as well. WDYT? Can this wait till 1.5.0?

As long as we can agree on not merging anything which would break this. And that the release is still on for this week. Absolutely.

sophokles73 (approved):
LGTM

ready to be merged FMPOV

PR #2158, opened 2020-09-07
Refactor JDBC storage layer, add tenant store

ctron:


ctron:
This build of this will fail, because it already misses some features from #2157

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/device/TableAdapterStore.java
```diff
@@ -114,4 +195,52 @@ public TableAdapterStore(final SQLClient client, final Tracer tracer, final Stat
...
+     * @param spanContext The span to contribute to.
+     *
+     * @return A future tracking the outcome of the operation.
+     *
+     * @see org.eclipse.hono.deviceregistry.service.device.AbstractRegistrationService#resolveGroupMembers(String, JsonArray, Span)
+     */
+    public Future<Set<String>> resolveGroupMembers(final String tenantId, final Set<String> viaGroups, final SpanContext spanContext) {
+
+        final Span span = TracingHelper.buildChildSpan(this.tracer, spanContext, "resolve group members", getClass().getSimpleName())
+                .withTag("tenant_instance_id", tenantId)
```
calohmn, line 212: Could you have a look at the comment in the old PR (https://github.com/eclipse/hono/pull/2118#discussion_r482336686) and elaborate a bit on the internal tenant id you mentioned?
Do you mean the `TenantKey#name` value (javadoc: "tenant name in an external system")? 

To me, the `resolveGroupMembers` method above doesn't seem to be invoked in a way that such a name/identifier is used for the first parameter here, though.

ctron: Taking a closer look at the code, I think we have a problem here with the way this was migrated from EnMasse.

The "name" used to be the user facing name, where "id" is the internal ID, e.g. in the database. The tenant information service can (if implemented that way) translate from the external name to the internal name.

The JDBC device registry must only use the internal name.

So the `resolveGroupMembers` method, should either receive the internal ID (which is the current code), receive a `TenantKey` and use the internal ID, or do the lookup (via the tenant information service) itself.

I think it would be ok to ignore this for the moment, and simply use the "ID" as "tenant_id". And sort that out later.

ctron:
I rebased the PR and changed `tenant_instance_id` to `tenant_id`.

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/device/TableManagementStore.java
```diff
@@ -122,67 +183,414 @@ public TableManagementStore(final SQLClient client, final Tracer tracer, final S
...
+     * {@code data} and optionally current resource version from the column {@code version}.
+     *
+     * @param key The key of the device to read.
+     * @param spanContext The span to contribute to.
+     * @return A future, tracking the outcome of the operation.
+     */
+    public Future<Optional<DeviceReadResult>> readDevice(final DeviceKey key, final SpanContext spanContext) {
+
+        final Span span = TracingHelper.buildChildSpan(this.tracer, spanContext, "read device", getClass().getSimpleName())
+                .withTag("tenant_id", key.getTenantId())
```
sophokles73, line 432: please use Tags from TracingHelper

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/device/TableManagementStore.java
```diff
@@ -122,67 +183,414 @@ public TableManagementStore(final SQLClient client, final Tracer tracer, final S
...
+     *
+     * @param key The key of the device to delete.
+     * @param resourceVersion An optional resource version.
+     * @param spanContext The span to contribute to.
+     * @return A future, tracking the outcome of the operation.
+     */
+    public Future<UpdateResult> deleteDevice(final DeviceKey key, final Optional<String> resourceVersion, final SpanContext spanContext) {
+
+        final Span span = TracingHelper.buildChildSpan(this.tracer, spanContext, "delete device", getClass().getSimpleName())
+                .withTag("tenant_id", key.getTenantId())
```
sophokles73, line 472: please use Tags from TracingHelper

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/device/TableManagementStore.java
```diff
@@ -122,67 +183,414 @@ public TableManagementStore(final SQLClient client, final Tracer tracer, final S
...
+     * @param resourceVersion The optional resource version to update.
+     * @param spanContext The span to contribute to.
+     * @return A future, tracking the outcome of the operation.
+     */
+    public Future<Versioned<Boolean>> setCredentials(final DeviceKey key, final List<CommonCredential> credentials, final Optional<String> resourceVersion,
+                                                     final SpanContext spanContext) {
 
         final Span span = TracingHelper.buildChildSpan(this.tracer, spanContext, "set credentials", getClass().getSimpleName())
-                .withTag("tenant_instance_id", key.getTenantId())
+                .withTag("tenant_id", key.getTenantId())
```
sophokles73, line 549: please use Tags from TracingHelper

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/tenant/ManagementStore.java
```diff
@@ -0,0 +1,422 @@
...
+     * @param spanContext The span to contribute to.
+     * @return A future, tracking the outcome of the operation.
+     */
+    public Future<Versioned<Void>> create(final String tenantId, final Tenant tenant, final SpanContext spanContext) {
+
+        final var json = tenantToJson(tenant);
+        final var version = UUID.randomUUID().toString();
+
+        final Span span = TracingHelper.buildChildSpan(this.tracer, spanContext, "create tenant", getClass().getSimpleName())
+                .withTag("tenant_id", tenantId)
```
sophokles73, line 171: please use Tags from TracingHelper

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/tenant/ManagementStore.java
```diff
@@ -0,0 +1,422 @@
...
+     * @param resourceVersion An optional resource version.
+     * @param spanContext The span to contribute to.
+     * @return A future, tracking the outcome of the operation.
+     */
+    public Future<Versioned<Void>> update(final String tenantId, final Tenant tenant, final Optional<String> resourceVersion, final SpanContext spanContext) {
+
+        final var json = tenantToJson(tenant);
+
+        final Span span = TracingHelper.buildChildSpan(this.tracer, spanContext, "update tenant", getClass().getSimpleName())
+                .withTag("tenant_id", tenantId)
```
sophokles73, line 335: please use Tags from TracingHelper

sophokles73 (commented):
I did not review the code at a semantic level. Just from flying over the code it looks like this is a complete O/R mapping framework. If so, is that really necessary or can't we simply use one of the existing frameworks for that purpose?

ctron:
Should be fixed.

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/tenant/ManagementStore.java
```diff
@@ -0,0 +1,422 @@
...
+     * @param spanContext The span to contribute to.
+     * @return A future, tracking the outcome of the operation.
+     */
+    public Future<Versioned<Void>> update(final String tenantId, final Tenant tenant, final Optional<String> resourceVersion, final SpanContext spanContext) {
+
+        final var json = tenantToJson(tenant);
+
+        final Span span = TracingHelper.buildChildSpan(this.tracer, spanContext, "update tenant", getClass().getSimpleName())
+                .withTag(TracingHelper.TAG_TENANT_ID, tenantId)
+                .withTag("data", json)
```
sophokles73, line 336: Do you intend to provide full text searching on all CRUD method payloads via the tracing system? We do nothing like this for the other device registries so I wonder why we would want/need to do this in the JDBC registry.
FMPOV the Tags are intended as markers for the spans by which it is easy to group and filter them. Adding full payloads here doesn't seem right to me. If at all, the payload should be logged to the span but it should not go into a tag ....

ctron: Is there anything else on this PR to fix?

sophokles73: As I wrote above, FMPOV the data should not be written into Tags ...

ctron:
> I did not review the code at a semantic level. Just from flying over the code it looks like this is a complete O/R mapping framework. If so, is that really necessary or can't we simply use one of the existing frameworks for that purpose?

We don't really need a full O/R mapper. As we don't do O/R mapping. This is more a configurable, fixed set, of SQL statements.

It also allows to use database specific functionalities, like JSON types, which isn't supported by some O/R mappers.

Additionally, not using an O/R mapper also doesn't add any new dependencies.

services/base-jdbc/src/main/java/org/eclipse/hono/service/base/jdbc/store/tenant/ManagementStore.java
ctron: Yes, I understand that. But with every round of changes I add, new comments come up. Instead of endlessly updating the PR, I would like to understand "what else" I need to change on the PR, in order get it merged.

sophokles73: Sorry for the inconvenience. Nothing else from my side.

sophokles73 (approved):

PR #2159, opened 2020-09-07
Add JDBC device registry

ctron:
The CI is expected to fail on this one as well, as it misses changes from: https://github.com/eclipse/hono/pull/2157 and https://github.com/eclipse/hono/pull/2158

### DIFF
--- a/services/device-registry-jdbc/src/main/java/org/eclipse/hono/deviceregistry/jdbc/impl/CredentialsServiceImpl.java
+++ b/services/device-registry-jdbc/src/main/java/org/eclipse/hono/deviceregistry/jdbc/impl/CredentialsServiceImpl.java
@@ -1,0 +1,150 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+
+package org.eclipse.hono.deviceregistry.jdbc.impl;
+
+import java.net.HttpURLConnection;
+import java.time.Instant;
+import java.time.OffsetDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Objects;
+import java.util.function.BiFunction;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import org.eclipse.hono.deviceregistry.service.credentials.AbstractCredentialsService;
+import org.eclipse.hono.deviceregistry.service.credentials.CredentialKey;
+import org.eclipse.hono.deviceregistry.service.tenant.TenantKey;
+import org.eclipse.hono.service.base.jdbc.store.device.TableAdapterStore;
+import org.eclipse.hono.util.CacheDirective;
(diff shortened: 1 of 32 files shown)


ctron:
Rebased

tests/pom.xml
```diff
@@ -267,6 +286,74 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
...
+                          <includes>
+                            <include>**/*</include>
+                          </includes>
+                        </fileSet>
+                      </inline>
+                    </assembly>
+                    <runCmds>
+
+                      <run>microdnf -y install java-11-openjdk-headless</run>
+                      <run>curl -sL https://repo1.maven.org/maven2/com/h2database/h2/1.4.200/h2-1.4.200.jar -o h2.jar</run>
```
sophokles73, line 314: can we use the version defined in `bom/POM.xml` here instead?

tests/pom.xml
```diff
@@ -314,6 +401,7 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
                     </network>
                     <memorySwap>524288000</memorySwap>
                     <memory>524288000</memory>
+                    <memory>524288000</memory>
```
sophokles73, line 404: this seems to be a duplicate

sophokles73 (commented):
looks very good :-)

ctron:
Ok, I rebased, amended the feedback, and (most likely) fixes the test. Which was due to the fact that the "max bcrypt iterations" variable was renamed to "cost factor". But let's wait for travis to check.

sophokles73:
@ctron is there anything else I can help with? FMPOV the only thing that prevents this PR from merging is the failing unit test ...

fkaltner:
> @ctron is there anything else I can help with? FMPOV the only thing that prevents this PR from being merged is the failing unit test ...

@ctron I am also happy to help, if there's anything I could do FYPOV. @dejanb would like to wait with merging the pull request for the status property (https://github.com/eclipse/hono/pull/2189) until this PR is merged...

services/device-registry-jdbc/src/main/java/org/eclipse/hono/deviceregistry/jdbc/impl/CredentialsServiceImpl.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        if (!secret.getBoolean(CredentialsConstants.FIELD_ENABLED, true)) {
+            return false;
+        }
+
+        if (!validTime(secret, CredentialsConstants.FIELD_SECRETS_NOT_BEFORE, Instant::isAfter)) {
+            return false;
+        }
+        if (!validTime(secret, CredentialsConstants.FIELD_SECRETS_NOT_AFTER, Instant::isBefore)) {
```
kaniyan, line 106: @sophokles73 @ctron 

Sorry for raising this comment now. I noticed this just now with the merge of the PR #2241 for adjusting the tests.

The Credentials API docs says:

> Adapters MUST only consider secrets for authentication which
> 
>     have their not-before property set to either null or the current or a past point in time and
>     have their not-after property set to either null or the current or a future point in time.

AFAIK the credentials API includes all the secrets irrespective of the _not-before_ & _not-after_ property. Based on the above quotes, the adapters are responsible for filtering them out. Various _AuthProvider_ implementations already do this by using `CredentialsObject.getCandidateSecrets(...)` where the secrets are already being filtered out based on the validity period. IMHO we need not filter secrets based on the validity period in the credentials API implementation. I don't think we need to adapt the tests as in #2241

kaniyan (changes requested):

PR #2160, opened 2020-09-07
Fix command delivery not getting released, improve tracing

### LINKED ISSUE #2155: CoAP adapter: no disposition update sent for command
Consider the following scenario:
A telemetry message is sent to the CoAP adapter with type `CON` (_at least once_ delivery semantics) and a `ttd` parameter.
The downstream application does *not* acknowledge the message. There is however a command being sent to the device.

Because the telemetry message is not acknowledged, the CoAP adapter returns an error response back to the originating device after the configured `sendMessageTimeout` has elapsed.

But, nothing is done to the received command, no disposition update ("released") is sent back to the downstream application.
The same occurs if the telemetry or event message gets a disposition update other than "accepted".

calohmn:
This fixes #2155.

The first commit fixes the command delivery potentially not getting updated in the CoAP adapter.
The second commit improves the overall Command & Control related tracing in the CoAP adapter by adopting the corresponding changes done for the HTTP adapter (#2153).

### DIFF
--- a/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
+++ b/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
@@ -742,7 +742,8 @@ public abstract class AbstractVertxBasedCoapAdapter<T extends CoapAdapterPropert
                             responseReady,
                             currentSpan));
 
-            return CompositeFuture.join(senderTracker, commandConsumerTracker).compose(ok -> {
+            return CompositeFuture.join(senderTracker, commandConsumerTracker)
+            .compose(ok -> {
                 final DownstreamSender sender = senderTracker.result();
                 final Integer ttd = ttdTracker.result();
                 final Message downstreamMessage = newMessage(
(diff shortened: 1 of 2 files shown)


adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractVertxBasedCoapAdapter.java
```diff
@@ -745,15 +749,16 @@ protected final String getAuthId(final CoapExchange exchange) {
...
                 final Message downstreamMessage = newMessage(
                         context.getRequestedQos(),
-                        ResourceIdentifier.from(endpoint.getCanonicalName(), context.getOriginDevice().getTenantId(),
-                                context.getOriginDevice().getDeviceId()),
+                        ResourceIdentifier.from(endpoint.getCanonicalName(), tenantId, deviceId),
                         "/" + context.getExchange().getRequestOptions().getUriPathString(),
                         contentType,
                         payload,
-                        tenantValidationTracker.result(),
+                        tenantTracker.result(),
```
boaks, line 761: Is that change from `tenantValidationTracker` to `tenantTracker` intended?

calohmn: Yes. It makes the code more in line with the corresponding code from the HTTP adapter and makes it more obvious that it is about the returned `TenantObject` here. In the following lines the tenant object is accessed via `tenantTracker.result()` as well.

boaks: Is there a test, which demonstrates, that a failure in `tenantValidationTracker` is handled proper by that code?

calohmn: A failure in the `tenantValidationTracker` would cause the `ttdTracker` to fail and subsequently the `commandConsumerTracker`. A failed `commandConsumerTracker` will cause the above code block (sending the telemetry/event message) to not be entered anyway.

A unit test in which a failed `tenantValidationTracker` is tested is `AbstractVertxBasedCoapAdapterTest#testMessageLimitExceededForAnEventMessage`.

sophokles73 (approved):
LGTM

Issue #2134, opened 2020-08-27
MongoDB implementation does not support client context matching


sophokles73:
The Credentials API's *get* operation provides means to include a *client context* which the service implementation may use to determine matching credentials on record.
The file based registry does support this already. In particular, it will not return the following credentials on record
```json
{
  "auth-id": "4711",
  "type": "hashed-password",
  "secrets": [ ... ],
  "ext": {
    "client-id": "AA"
  }
}
```

for a request with the following search criteria:
```json
{
  "auth-id": "4711",
  "type": "hashed-password",
  "client-id": "BB"
}
```
However, the MongoDB currently does not consider the client context and thus (erroneously) returns the credentials mentioned above.

kaniyan:
This is not yet implemented and a _TODO_ is still open. If it can wait, I can start after finishing the IT tests for search devices operation and implementing tenant existence check while registering devices.
https://github.com/eclipse/hono/blob/e30ba4b192b5be4b34315c1325a2a113092b0322/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedCredentialsService.java#L415-L422

PR #2164, opened 2020-09-08
[#2134] Make use of the given client context to match credentials in MongoDB registry

kaniyan:


services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedCredentialsService.java
```diff
@@ -382,6 +383,8 @@ private CompositeFuture createIndices() {
                         .flatMap(ok -> Optional
                                 .ofNullable(result.getJsonArray(MongoDbDeviceRegistryUtils.FIELD_CREDENTIALS))
                                 .map(credential -> credential.getJsonObject(0))
+                                .filter(credential -> DeviceRegistryUtils.matchesWithClientContext(credential,
```
sophokles73, line 386: if the result is a JsonArray (as the previous line seems to indicate), then we should probably try to find the matching credentials among the elements of that array instead of just matching the first element, right?

kaniyan: AFAIK, at any case there won't be more than one element. The result is obtained by matching _tenantId_, _authId_ and _type_. No two records can have same value for those fields.

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
```diff
@@ -242,4 +242,39 @@ public static void checkCredential(
             }
         }
     }
+
+    /**
+     * Checks if the properties in the given client context matches with the extension properties of
```
sophokles73, line 247: It would be good to describe how the client context is matched against the extension properties as I do not think that this is straight forward.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedCredentialsService.java
sophokles73: Yes, that is true for our implementation at hand.

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
kaniyan: How about describing (how is being matched) in the paragraph next to the description as below?  WDYT? 

```
    /**
     * Checks if the properties in the given client context matches with the extension properties of the
     * given credential.
     * <p>
     * The matching properties in the client context and credential extensions are filtered out based 
     * on their property names. If their values are equal then {@code true} is returned. Otherwise {@code false}.
```

sophokles73: we'll also need to mention the special handling of properties that only exist in the context but not in the extensions ...

kaniyan: I have also added the above to the JavaDoc.

sophokles73 (approved):
LGTM

Issue #2156, opened 2020-09-07
Transport protocol agnostic interface for Hono-Client


b-abel:
To use [Kafka for messaging](#8 ), the Hono client must be refactored so that its interface is abstracted from AMQP 1.0 and allows the use of a different protocol.

sophokles73:
@b-abel  now that we have added the new interfaces, can we close this issue?

b-abel:
> @b-abel now that we have added the new interfaces, can we close this issue?

@sophokles73 Yes.

sophokles73:
closed via #2168

Issue #1276, opened 2019-06-06
Add explicit support for request/response communication pattern (device to application)


calohmn:
This issue is about a use case where a device sends a request message and expects a corresponding response from a northbound application.

Such a scenario can already be implemented with Hono. The device can send a telemetry/event message with a correlation id property/parameter and with a `ttd` parameter (signaling its readiness to receive a command message). The northbound application can send a one-way command message as response, including the given correlation id. Upon receiving the command message, the device can correlate the message to its initial telemetry/event message using the correlation id.

However, such a solution is not very intuitive and requires manual correlation of the messages.

Therefore it would be good to explore the feasibility of adding explicit support for such a communication pattern, including correlation support and matching endpoint names.

calohmn:
Initial idea:

**Device side: Sending the request message**
- to the MQTT adapter:
  - device chooses a correlation id
  - device subscribes to a topic `req_resp/${tenant-id}/${device-id}/resp/${correlation-id}/#`
  - device publishes request to `req_resp/${tenant-id}/${device-id}/req/${correlation-id}`
- to the HTTP adapter:
  - device chooses a correlation id (could also be described as request id)
  - device sends request to `/req_resp/${correlation-id}` and receives corresponding response

---
Handling request and response messages in terms of AMQP messages could either be done with `correlation-id` based links (ie. with the overhead of extra link creation):

**Shared Protocol Adapter logic**
- protocol adapter forwards a request message on a sender link with address `req_resp/${tenant-id}`, sets `req_resp/${tenant-id}/${correlation-id}` as `reply-to` address
- protocol adapter opens a receiver link for the response on `req_resp/${tenant-id}/${correlation-id}`

**Application side**
- application opens receiver link on `req_resp/${tenant-id}`
- application receives a request and sends the response message on a sender link with address `req_resp/${tenant-id}/${device-id}/${correlation-id}` (according to `reply-to` of request message)

---
*Or*, AMQP messages could be handled using a fixed number of reusable links (along with mapping incoming responses to the right protocol adapter instance):

**Shared Protocol Adapter logic**
- protocol adapter forwards a request message on a sender link with address `req_resp/${tenant-id}`, sets `req_resp/${tenant-id}/${correlation-id}` as `reply-to` address
- protocol adapter opens a receiver link on `req_resp/${tenant-id}/${protocolAdapterUUID}` (see #1272)
[...]

PR #2168, opened 2020-09-10
[#2156] Add Hono API client interfaces for protocol adapters

sophokles73:
This is just a first step towards a major refactoring of the client. In particular, it will be split up into separate modules targeting different client roles: devices/gateways, protocol adapters and downstream applications.

I have a created a new module *clients* which I would like to use as the parent for all the other client role specific modules, the first one being the *adapter* module containing the clients required for implementing protocol adapters.


clients/adapter/src/main/java/org/eclipse/hono/client/registry/DeviceRegistrationClient.java
```diff
@@ -0,0 +1,81 @@
...
+import io.vertx.core.Future;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * A client for accessing Hono's Device Registration API.
+ * <p>
+ * See Hono's <a href="https://www.eclipse.org/hono/docs/api/device-registration/">
+ * Device Registration API</a> for a description of the status codes returned.
+ */
+public interface DeviceRegistrationClient {
```
b-abel, line 27: I haven't spent much time on the interfaces for the registry clients. We haven't talked about whether we want registry clients with other transport protocols. But the way the interfaces look, there is nothing AMQP-specific in them anyway.

clients/adapter/src/main/java/org/eclipse/hono/client/upstream/CommandResponseSender.java
```diff
@@ -0,0 +1,41 @@
...
+package org.eclipse.hono.client.upstream;
+
+import io.opentracing.SpanContext;
+import io.vertx.core.Future;
+import io.vertx.proton.ProtonDelivery;
+
+/**
+ * A client for publishing a device's response to a command received from a downstream application.
+ */
+public interface CommandResponseSender {
```
b-abel, line 22: The interfaces in the `upstream` package still depend heavily on AMQP. Do we change them completely again if we want to support Kafka for commands? I think that if you use Kafka for downstream messaging, you might want to use Hono in the future entirely without AMQP.

clients/adapter/src/main/java/org/eclipse/hono/client/downstream/EventSender.java
```diff
@@ -0,0 +1,76 @@
...
+     * @return A future indicating the outcome of the operation.
+     *         <p>
+     *         The future will be succeeded if the event has been sent downstream.
+     *         <p>
+     *         The future will be failed with a {@link org.eclipse.hono.client.ServerErrorException} if the data
+     *         could not be sent. The error code contained in the exception indicates the
+     *         cause of the failure.
+     * @throws NullPointerException tenant ID, device ID or contentType are {@code null}.
+     */
+    Future<ProtonDelivery> sendEvent(
```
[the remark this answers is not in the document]
sophokles73: This is a typo. Both variants should return a Future<Void>.

clients/adapter/src/main/java/org/eclipse/hono/client/downstream/EventSender.java
```diff
@@ -0,0 +1,76 @@
...
+     * @return A future indicating the outcome of the operation.
+     *         <p>
+     *         The future will be succeeded if the event has been sent downstream .
+     *         <p>
+     *         The future will be failed with a {@link org.eclipse.hono.client.ServerErrorException} if the event
+     *         could not be sent. The error code contained in the exception indicates the
+     *         cause of the failure.
+     * @throws NullPointerException tenant ID, device ID or contentType are {@code null}.
+     */
+    Future<Void> sendEvent(
```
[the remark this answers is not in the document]
sophokles73: That is indeed arguable. So we either make the properties param in the second variant mandatory or we keep the first variant only. Which one would you prefer?

clients/adapter/src/main/java/org/eclipse/hono/client/registry/CredentialsClient.java
```diff
@@ -0,0 +1,84 @@
...
+     *         The future will succeed if a response with a status code in the [200, 300) range
+     *         has been received from the Credentials service. The JSON object will then contain values as
+     *         defined in <a href="https://www.eclipse.org/hono/docs/api/credentials/#get-credentials">
+     *         Get Credentials</a>.
+     *         <p>
+     *         Otherwise, the future will fail with a {@link org.eclipse.hono.client.ServiceInvocationException}
+     *         containing the (error) status code returned by the service.
+     * @throws NullPointerException if tenant ID, type or auth ID are {@code null}.
+     */
+    Future<CredentialsObject> get(
```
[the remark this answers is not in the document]
sophokles73: It is not strictly *necessary* but its presence allows invoking a method that is specific to the use case at hand without providing `null` params ...

clients/adapter/src/main/java/org/eclipse/hono/client/registry/DeviceRegistrationClient.java
sophokles73: Yes, that was the intention. So, in the future we might as well implement the interfaces based on HTTP ...

clients/adapter/src/main/java/org/eclipse/hono/client/upstream/CommandResponseSender.java
sophokles73: Yes, that is true. We might want to create `Command` and `CommandResponse` to be independent from the underlying transport.

clients/adapter/src/main/java/org/eclipse/hono/client/downstream/EventSender.java
```diff
@@ -0,0 +1,76 @@
...
+     * @param payload The data to send.
+     * @param contentType The content type of the data.
+     * @return A future indicating the outcome of the operation.
+     *         <p>
+     *         The future will be succeeded if the event has been sent downstream .
+     *         <p>
+     *         The future will be failed with a {@link org.eclipse.hono.client.ServerErrorException} if the event
+     *         could not be sent. The error code contained in the exception indicates the
+     *         cause of the failure.
+     * @throws NullPointerException tenant ID, device ID or contentType are {@code null}.
```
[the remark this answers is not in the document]
sophokles73: Currently, the Telemetry API defines the content-type to be mandatory. If a device wants to send an empty notification, then the content-type to use is [application/vnd.eclipse-hono-empty-notification](https://www.eclipse.org/hono/docs/1.4/api/event/#empty-notification). If a (device) transport does not support content type, then the protocol adapter should use `application/octet-stream` ...

clients/adapter/src/main/java/org/eclipse/hono/client/downstream/EventSender.java
```diff
@@ -0,0 +1,76 @@
...
+
+import io.opentracing.SpanContext;
+import io.vertx.core.Future;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.proton.ProtonDelivery;
+
+/**
+ * A client for publishing events originating from devices to downstream consumers.
+ */
+public interface EventSender {
```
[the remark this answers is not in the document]
sophokles73: Yes, that might be a good idea. I would prefer extending a separate interface for that purpose.

clients/adapter/src/main/java/org/eclipse/hono/client/downstream/TelemetrySender.java
```diff
@@ -0,0 +1,88 @@
...
+     * @return A future indicating the outcome of the operation.
+     *         <p>
+     *         The future will be succeeded if the data has been sent downstream according to
+     *         the delivery semantics indicated by the qos parameter.
+     *         <p>
+     *         The future will be failed with a {@link org.eclipse.hono.client.ServerErrorException} if the data
+     *         could not be sent. The error code contained in the exception indicates the
+     *         cause of the failure.
+     * @throws NullPointerException tenant ID, device ID or contentType are {@code null}.
+     * @throws IllegalArgumentException if qos is neither 0 nor 1.
```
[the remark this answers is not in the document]
sophokles73: That is true. We should probably define an enum ...

sophokles73:
@b-abel I have pushed some changes ...

clients/adapter/src/main/java/org/eclipse/hono/client/downstream/EventSender.java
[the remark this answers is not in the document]
b-abel: I expect metadata to be set wherever messages are sent in the protocol adapters. Then the variant without the properties parameter would be obsolete. But I am having a hard time figuring out all the metadata currently set in the messages, so I could be wrong about their usage frequency.
In general, I would provide as few overloaded methods as possible in advance and prefer to extend the interface later, if the need arises.

b-abel:
The downstream interfaces work fine for me now. Just one thing is not clear to me yet. Here, the tenant is passed ion when the message is sent, whereas up to now a separate sender was created for each tenant. 
Have you already thought about how you want to manage the sender instances? Do you want to continue using the `CachingClientFactory`?

sophokles73:
> The downstream interfaces work fine for me now.

That's good to know :-)

> Just one thing is not clear to me yet. Here, the tenant is passed ion when the message is sent, whereas up to now a separate sender was created for each tenant.

That is the whole purpose of the new interfaces. Do not require users to handle multiple client instances (one for each tenant) but instead have  a single client that can be used to access all tenants (if appropriate and authority permits).

> Have you already thought about how you want to manage the clients for the protocol adapters? Do you want to continue using the CachingClientFactory?

I have only thought about some additional principles that I deem worthwhile when implementing the client(s):
* Do not require users to actively manage the client's lifecycle. For example, a client implementation might open up one or more connections to the server after instantiation or it might open connection(s) on demand, i..e when the first method invocation happens.
* Links should be opened implicitly when needed. Users do not need to know whether a separate link for each tenant is used or not. This also allows a client implementation to make better use of cached values, e.g. a cached value could be returned while a link is not yet established.

clients/adapter/src/main/java/org/eclipse/hono/client/upstream/CommandResponse.java
```diff
@@ -0,0 +1,195 @@
...
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.client.upstream;
```
calohmn, line 14: Sticking to the downstream/upstream structure, I think this class as well as `CommandResponseSender` should be in the `downstream` package.

sophokles73: I see your point. However, my idea was not so much to organize by technical aspects but rather by means of communication patterns. Most devices will support the telemetry (downstream) direction, i.e. they will report sensor readings etc. Some will additionally support receiving commands (upstream) and (optionally) responding to commands.
Maybe it would be better to rename the packages to `org.eclipse.hono.client.telemetry` and `org.eclipse.hono.client.command` respectively? WDYT @calohmn

calohmn: I think I would still prefer the downstream/upstream packages, with "CommandResponseSender" in the downstream package.
To me, grouping classes by message flow direction offers more value here than using the communication pattern as the grouping theme. For the request/response pattern (#1276) we might also reuse the existing senders/consumers, so the Hono communication patterns wouldn't be directly reflected in the packages here anyway.
And having `org.eclipse.hono.client.telemetry` contain telemetry *and* event senders is also a bit odd. A different term encompassing both telemetry and events would be better, although I guess we don't have that.

sophokles73: > And having org.eclipse.hono.client.telemetry contain telemetry and event senders is also a bit odd. A different term encompassing both telemetry and events would be better, although I guess we don't have that.

We have been calling it the *telemetry* direction from the very beginning ...

> For the request/response pattern (#1276) we might also reuse the existing senders/consumers, so the Hono communication patterns wouldn't be directly reflected in the packages here anyway.

Well, FMPOV re-designing the client provides us with the opportunity to provide interfaces that are taylored to the the use case at hand. If we want to support a request/response pattern then FMPOV we should add a new package with new interfaces that contain method names and wording describing the actual problem being addressed. This should make it much easier to understand the exchange pattern compared to re-using the Telemetry and Command related interfaces for this purpose.

b-abel:
You may want to consider having a common interface (like an empty `DownstreamSender`) for `EventSender` and `TelemetrySender`, or else you may need to duplicate (or refactor otherwise) the method [AbstractVertxBasedHttpProtocolAdapter#doUploadMessage](https://github.com/eclipse/hono/blob/4a15feefb40696b83d7c2241e5d44b9c932ee7b1/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java#L595).

clients/adapter/src/main/java/org/eclipse/hono/client/downstream/QoS.java
```diff
@@ -0,0 +1,26 @@
...
+
+
+package org.eclipse.hono.client.downstream;
+
+
+/**
+ * A definition of the delivery semantics used for transferring messages.
+ *
+ */
+public enum QoS {
```
fkaltner, line 22: Just stumbled upon this via https://github.com/eclipse/hono/pull/2216 : there's already a QoS class in `org.eclipse.hono.util` (and another one in `org.eclipse.hono.service.metric`). I think we could use the one in `org.eclipse.hono.util` here. It looks similar to me.

sophokles73: nice catch :+1: I'll replace it with the `org.eclipse.hono.util.QoS` enum

b-abel: I thought you created a new one on purpose because they are actually quite different: `org.eclipse.hono.util.QoS` additionally contains `UNKNOWN`. I think you should set a valid QoS when invoking the client and I do not see value in opening the possibility of a new error case.

sophokles73: @b-abel That might be true. However, I think we can also easily get rid of the existing `QoS.UNKNOWN` value as it doesn't add much value anyway.

sophokles73: See #2233

clients/adapter/src/main/java/org/eclipse/hono/client/downstream/EventSender.java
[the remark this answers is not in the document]
sophokles73: Ok, then let's remove the variant without the properties for the time being and allow the properties map to be `null`.

clients/adapter/src/main/java/org/eclipse/hono/client/upstream/CommandResponse.java
sophokles73: @calohmn can we agree to group the interfaces into packages based on semantics instead of (technical) message flow direction? And then rename packages *downstream* and *upstream* to *telemetry* and *command* respectively for the time being and see if we find better names in the future?

clients/adapter/src/main/java/org/eclipse/hono/client/downstream/EventSender.java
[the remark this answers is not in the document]
b-abel: :+1:

sophokles73:
@calohmn @b-abel I would like to merge this PR ASAP so that we can use it as the initial version for implementing the new Kafka- and AMQP-based clients. We can make adaptions later on ...

b-abel:
> I would like to merge this PR ASAP so that we can use it as the initial version for implementing the new Kafka- and AMQP-based clients.

I'm all for it. What about my proposal to add a common interface `DownstreamSender` to `TelemetrySender` and `EventSender`? This would be no extra effort but significantly reduce the need for refactorings the abstract protocol adapter base classes.

clients/adapter/src/main/java/org/eclipse/hono/client/upstream/CommandResponse.java
calohmn: Yes, I'm ok with using *telemetry* and *command* for now.

calohmn (approved):

PR #2172, opened 2020-09-11
[#2115] Add ExecutionContext#getTracingSpan

calohmn:
This is for #2115:
Require the `MapBasedExecutionContext` classes to have a non-null `Span` object. This is a preparation for having the authentication handler classes apply a trace sampling priority on the span of the `ExecutionContext`.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/MapBasedExecutionContext.java
+++ b/core/src/main/java/org/eclipse/hono/util/MapBasedExecutionContext.java
@@ -14,8 +14,10 @@ package org.eclipse.hono.util;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 
+import io.opentracing.Span;
 import io.opentracing.SpanContext;
 
 /**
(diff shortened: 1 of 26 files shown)


core/src/main/java/org/eclipse/hono/util/MapBasedExecutionContext.java
```diff
@@ -35,25 +47,30 @@
...
+    /**
+     * Gets the <em>OpenTracing</em> root span that is used to
+     * track the processing of this context.
+     *
+     * @return The span.
+     */
     @Override
-    public SpanContext getTracingContext() {
-        return spanContext;
+    public Span getTracingSpan() {
```
sophokles73, line 67: final ?

core/src/main/java/org/eclipse/hono/util/MapBasedExecutionContext.java
```diff
@@ -35,25 +47,30 @@
...
+        return span;
     }
 
     @Override
-    public void setTracingContext(final SpanContext spanContext) {
-        this.spanContext = spanContext;
+    public SpanContext getTracingContext() {
```
sophokles73, line 72: final ?

sophokles73 (approved):
LGTM

Issue #2175, opened 2020-09-14
Device Registries do not accept credentials with auth-id containing underscore


sophokles73:
The Device Registry Management API defines the authentication identifier pattern as `[a-zA-Z0-9-_=.]+` which includes the underscore character. However, the registry implementations do not accept `4711_authid` as an authentication identifier.

So, either the pattern in the API needs to be changed to exclude the underscore or the implementations need to be changed to accept the underscore.

sophokles73:
@ctron @dejanb is there any reason not to accept underscores in the auth-id property value from your point of view? If not, then I would like to change the pattern in the implementations accordingly ...

calohmn:
I don't see scenarios where underscore would cause problems, therefore I would stick with the API definition here.

dejanb:
I don't see why that would be, I'd stick with the spec as well

@jbtrystram you spent the most time in this area. Do you see any reason for this?

jbtrystram:
I probably have removed the underscore in the implementation because I saw a couple of REST API designs guides that advised against using underscore in resources that where accessible through a URL. The argument being that it may cause confusion when linking such a resource, because the usual underline of URL resources may "hide" the underscore and someone could mistake it for a space. 

IIRC we discussed it afterwards and chose to ignore that, and i probably failed to update the implementation. I am fine sticking to the spec if you guys also think the issue mentioned above is not a big deal.

sophokles73:
closed via #2177

PR #2177, opened 2020-09-14
[#2175] Accept underscore in auth-id property

sophokles73:
The Device Registry Management API allows the underscore character to be
used in the auth-id property of credentials. The corresponding pattern
used for validating the authId property in the CommonCredential class
has been updated accordingly.


dejanb (approved):
LGTM

Issue #2166, opened 2020-09-08
HTTP adapter: Improve 503 error message in response in certain cases


calohmn:
When there is a HTTP 503 `Service Unavailable` error while processing a telemetry/event message in the HTTP adapter, the response payload only contains the error message `temporarily unavailable`.

I would be helpful to have a more specific error message especially in these cases:
- a telemetry message has been sent and there is no consumer (internally, there is a "no credit" error)
- a telemetry message with QoS 1 has been sent and there was a timeout waiting for the disposition update from the northbound application.

PR #2179, opened 2020-09-15
Add unit tests for HTTP 503 responses

calohmn:
Add unit tests for HTTP adapter responses with status code 503.
Extend `ApplicationClientFactory` with extra consumer methods to allow setting the delivery state explicitly.
Preparation for #2166.


client/src/main/java/org/eclipse/hono/client/ApplicationClientFactory.java
```diff
@@ -56,6 +56,29 @@ static ApplicationClientFactory create(final HonoConnection connection, final Se
      *
      * @param tenantId The tenant to consume data for.
      * @param telemetryConsumer The handler to invoke with every message received.
+     * @param autoAccept {@code true} if received deliveries should be automatically accepted (and settled)
```
sophokles73, line 59: under which circumstances would you want to set this to false?

client/src/main/java/org/eclipse/hono/client/ApplicationClientFactory.java
```diff
@@ -67,11 +90,32 @@ static ApplicationClientFactory create(final HonoConnection connection, final Se
             Consumer<Message> telemetryConsumer,
             Handler<Void> closeHandler);
 
+    /**
+     * Creates a client for consuming events from Hono's north bound <em>Event API</em>.
+     *
+     * @param tenantId The tenant to consume events for.
+     * @param eventConsumer The handler to invoke with every event received.
+     * @param closeHandler The handler invoked when the peer detaches the link.
+     * @param autoAccept {@code true} if received deliveries should be automatically accepted (and settled)
```
sophokles73, line 99: why would you want to set this to false?

client/src/main/java/org/eclipse/hono/client/ApplicationClientFactory.java
calohmn: In case accepting/rejecting the delivery should be done asynchronously.

In the case of the added test case here, a scenario should be simulated where the consumer takes a long time to accept the delivery - longer than the sendMessageTimeout. Without the `autoAccept=false` flag, this would mean having to block the event loop thread in the message handler, requiring the use of an extra HonoConnection/ApplicationClientFactory instance in the test.

client/src/main/java/org/eclipse/hono/client/ApplicationClientFactory.java
calohmn: See above.

tests/src/test/java/org/eclipse/hono/tests/http/TelemetryHttpIT.java
```diff
@@ -125,4 +127,102 @@ public void testUploadMessageFailsForLargePayload(final VertxTestContext ctx) {
...
+                            getEndpointUri(),
+                            Buffer.buffer("hello"),
+                            requestHeaders,
+                            ResponsePredicate.status(HttpURLConnection.HTTP_UNAVAILABLE));
+
+                })
+                // THEN the message gets rejected by the HTTP adapter with a 503
+                .onComplete(ctx.succeeding(response -> {
+                    ctx.verify(() -> {
+                        assertThat(response.bodyAsString()).isEqualTo("temporarily unavailable");
```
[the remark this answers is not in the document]
calohmn: I mainly added this here so that with the fix for #2166, the more specific error message can be checked upon here.

I have removed the check now and will add it again with the fix for #2166 (using a constant then).

client/src/main/java/org/eclipse/hono/client/ApplicationClientFactory.java
sophokles73: What you describe as the use is actually an anti pattern, isn't it? It means that the messages sent downstream will remain *in flight* for an indeterminate amount of time :-(
IMHO we should be very explicit in the JavaDoc that a consumer should not perform any time consuming work in the handler but should spawn a new thread or something like that in such a case, but otherwise quickly settle the incoming transfer.

calohmn: I've added a corresponding note in the javadocs.

PR #2185, opened 2020-09-17
[#2166] Improve 503 error messages in HTTP adapter

### LINKED ISSUE #2166: HTTP adapter: Improve 503 error message in response in certain cases

calohmn:
Also improve the corresponding documentation for the HTTP adapter and CoAP adapter.

This fixes #2166.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/NoConsumerException.java
+++ b/client/src/main/java/org/eclipse/hono/client/NoConsumerException.java
@@ -1,0 +1,41 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.client;
+
+import java.net.HttpURLConnection;
+
+/**
+ * A {@link ServerErrorException} indicating that there is no consumer for the request message.
+ */
+public class NoConsumerException extends ServerErrorException {
+
+    /**
+     * Resource key for the client facing error message.
+     */
+    public static final String CLIENT_FACING_MESSAGE_KEY = "SERVER_ERROR_NO_CONSUMER";
+
+    /**
+     * Creates a new NoConsumerException.
+     * <p>
+     * The exception will have a <em>503: Service unavailable</em> status code
+     * and its client facing error message will be set to the localized message
+     * with the key defined in {@link #CLIENT_FACING_MESSAGE_KEY}.
+     *
+     * @param msg The detail message.
+     */
+    public NoConsumerException(final String msg) {
+        super(HttpURLConnection.HTTP_UNAVAILABLE, msg);
+        setClientFacingMessageWithKey(CLIENT_FACING_MESSAGE_KEY);
+    }
+}
--- a/client/src/main/java/org/eclipse/hono/client/ServerErrorException.java
+++ b/client/src/main/java/org/eclipse/hono/client/ServerErrorException.java
@@ -21,6 +21,8 @@ public class ServerErrorException extends ServiceInvocationException {
 
     private static final long serialVersionUID = 1L;
 
+    private String clientFacingMessage;
+
     /**
      * Creates a new exception for a server error code.
      *
@@ -114,5 +116,23 @@ public class ServerErrorException extends ServiceInvocationException {
         if (errorCode < 500 || errorCode >= 600) {
             throw new IllegalArgumentException("client error code must be >= 500 and < 600");
         }
+    }
+
+    /**
+     * Gets the error message suitable to be propagated to an external client.
+     *
+     * @return The message or {@code null}.
+     */
+    public final String getClientFacingMessage() {
+        return clientFacingMessage;
+    }
+
+    /**
+     * Sets the error message suitable to be propagated to an external client.
+     *
+     * @param clientFacingMessageKey The key to get the localized message for.
+     */
+    public final void setClientFacingMessageWithKey(final String clientFacingMessageKey) {
+        this.clientFacingMessage = getLocalizedMessage(clientFacingMessageKey);
     }
 }
--- a/client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
@@ -30,8 +30,12 @@ import org.apache.qpid.proton.amqp.transport.DeliveryState;
 import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.client.ClientErrorException;
 import org.eclipse.hono.client.HonoConnection;
+import org.eclipse.hono.client.MessageNotProcessedException;
 import org.eclipse.hono.client.MessageSender;
+import org.eclipse.hono.client.MessageUndeliverableException;
+import org.eclipse.hono.client.NoConsumerException;
 import org.eclipse.hono.client.SendMessageSampler;
+import org.eclipse.hono.client.SendMessageTimeoutException;
 import org.eclipse.hono.client.ServerErrorException;
 import org.eclipse.hono.client.ServiceInvocationException;
 import org.eclipse.hono.config.ClientConfigProperties;
@@ -166,7 +170,7 @@ public abstract class AbstractSender extends AbstractHonoClient implements Messa
 
         return connection.executeOnContext(result -> {
             if (sender.sendQueueFull()) {
-                final ServiceInvocationException e = new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "no credit available");
+                final ServerErrorException e = new NoConsumerException("no credit available");
                 logMessageSendingError("error sending message [ID: {}, address: {}], no credit available",
                         rawMessage.getMessageId(), getMessageAddress(rawMessage));
                 logError(span, e);
@@ -271,9 +275,9 @@ public abstract class AbstractSender extends AbstractHonoClient implements Messa
         final Long timerId = connection.getConfig().getSendMessageTimeout() > 0
                 ? connection.getVertx().setTimer(connection.getConfig().getSendMessageTimeout(), id -> {
                     if (!result.future().isComplete()) {
-                        final ServerErrorException exception = new ServerErrorException(
-                                HttpURLConnection.HTTP_UNAVAILABLE,
-                                "waiting for delivery update timed out after " + connection.getConfig().getSendMessageTimeout() + "ms");
+                        final ServerErrorException exception = new SendMessageTimeoutException(
+                                "waiting for delivery update timed out after "
+                                        + connection.getConfig().getSendMessageTimeout() + "ms");
                         logMessageSendingError("waiting for delivery update timed out for message [ID: {}, address: {}] after {}ms",
                                 messageId, getMessageAddress(message), connection.getConfig().getSendMessageTimeout());
                         result.fail(exception);
@@ -303,11 +307,14 @@ public abstract class AbstractSender extends AbstractHonoClient implements Messa
                                 ? new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST)
                                 : new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST, rejected.getError().getDescription());
                     } else if (Released.class.isInstance(remoteState)) {
-                        e = new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE);
+                        e = new MessageNotProcessedException();
                     } else if (Modified.class.isInstance(remoteState)) {
                         final Modified modified = (Modified) deliveryUpdated.getRemoteState();
-                        e = modified.getUndeliverableHere() ? new ClientErrorException(HttpURLConnection.HTTP_NOT_FOUND)
-                                : new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE);
+                        if (modified.getUndeliverableHere()) {
+                            e = new MessageUndeliverableException();
+                        } else {
+                            e = new MessageNotProcessedException();
+                        }
                     }
                     result.fail(e);
                 }
--- a/site/documentation/content/user-guide/coap-adapter.md
+++ b/site/documentation/content/user-guide/coap-adapter.md
@@ -45,7 +45,7 @@ if the [message limit]({{< relref "/concepts/resource-limits.md" >}}) that has b
 
 ## CoAP Content Format Codes
 
-CoAP doesn't use a textual identifier for content types. Instead numbers are used, which are maintained by the [IANA](https://www.iana.org/).
+CoAP doesn't use a textual identifier for content types. Instead, numbers are used, which are maintained by the [IANA](https://www.iana.org/).
 The [IANA - CoAP Content Formats](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#content-formats) page lists all
 (currently) registered codes and the corresponding media types.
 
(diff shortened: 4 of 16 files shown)


client/src/main/java/org/eclipse/hono/client/ServerErrorException.java
```diff
@@ -21,6 +21,8 @@
 
     private static final long serialVersionUID = 1L;
 
+    private String clientFacingMessage;
```
sophokles73, line 24: this looks like it would be helpful in both ClientErrorException and ServerErrorException. Thus, IMHO we should add this to ServiceInvocationException.

client/src/main/java/org/eclipse/hono/client/ServerErrorException.java
```diff
@@ -115,4 +117,22 @@ public ServerErrorException(final String tenant, final int errorCode, final Stri
             throw new IllegalArgumentException("client error code must be >= 500 and < 600");
         }
     }
+
+    /**
+     * Gets the error message suitable to be propagated to an external client.
+     *
+     * @return The message or {@code null}.
+     */
+    public String getClientFacingMessage() {
```
sophokles73, line 126: final?

client/src/main/java/org/eclipse/hono/client/ServerErrorException.java
```diff
@@ -115,4 +117,22 @@ public ServerErrorException(final String tenant, final int errorCode, final Stri
...
+        return clientFacingMessage;
+    }
+
+    /**
+     * Sets the error message suitable to be propagated to an external client.
+     *
+     * @param clientFacingMessage The message.
+     */
+    public void setClientFacingMessage(final String clientFacingMessage) {
```
sophokles73, line 135: final ?

client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
```diff
@@ -56,6 +56,24 @@
  */
 public abstract class AbstractSender extends AbstractHonoClient implements MessageSender {
 
+    /**
+     * Client facing error message for the case of a message sender not receiving a delivery update in time.
+     * May be used as {@link ServerErrorException#clientFacingMessage}.
+     */
+    public static final String ERROR_SEND_MESSAGE_TIMEOUT = "consumer failed to indicate in time whether it has processed the message";
```
sophokles73, line 63: FMPOV you have now opened Pandora's Box of (localized) messages ;-)
I do not think that we should populate our constants classes with full fledged error messages ...
Instead, we should probably use resource bundles which have been designed for this purpose.

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -86,15 +86,18 @@ The device is authenticated using PSK.
     delivered to any potential consumer (yet). However, if the message type is `CON` (*at least once* semantics), then the adapter waits
     for the message to be delivered and accepted by a downstream consumer before responding with this status code.
   * 4.00 (Bad Request): The request cannot be processed. Possible reasons include:
-         * the request body is empty and the *URI-query* option doesn't contain the `empty` parameter
+         * the request body is empty, and the *URI-query* option doesn't contain the `empty` parameter.
```
sophokles73, line 89: no comma here

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -131,7 +134,7 @@ coap-client -u sensor1@DEFAULT_TENANT -k hono-secret -m POST coaps://hono.eclips
 ~~~
 
 {{% note %}}
-In the example above the response actually contains payload that should be used as input to a command to be executed by the device.
+In the example above the response actually contains a payload that should be used as input to a command to be executed by the device.
```
sophokles73, line 137: payload is unspecific here, analogously to how *people* is usually used...

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -171,16 +174,19 @@ send any commands to the device.
     delivered to any potential consumer (yet). However, if the message type is `CON` (*at least once* semantics), then the adapter waits
     for the message to be delivered and accepted by a downstream consumer before responding with this status code.
   * 4.00 (Bad Request): The request cannot be processed. Possible reasons include:
-         * the request body is empty and the *URI-query* option doesn't contain the `empty` parameter
+         * the request body is empty, and the *URI-query* option doesn't contain the `empty` parameter.
```
sophokles73, line 177: no comma

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -245,7 +251,7 @@ coap-client -m PUT coap://hono.eclipseprojects.io/telemetry/DEFAULT_TENANT/4711?
...
```
sophokles73, line 254: no comma

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -329,15 +338,14 @@ The device is authenticated using PSK.
...
```
sophokles73, line 341: no comma

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -329,15 +338,14 @@ The device is authenticated using PSK.
...
   * 4.03 (Forbidden): The request cannot be processed because the device's registration status cannot be asserted.
     Possible reasons for this include:
         * The given tenant is not allowed to use this protocol adapter.
   * 4.04 (Not Found): The request cannot be processed because the device is disabled or does not exist.
   * 4.13 (Request Entity Too Large): The request cannot be processed because the request body exceeds the maximum supported size.
   * 4.29 (Too Many Requests): The request cannot be processed because the tenant's message limit for the current period is exceeded.
-  * 5.03 (Service Unavailable): The request cannot be processed because there is no consumer of telemetry data for the given tenant
-    connected to Hono.
+  * 5.03 (Service Unavailable): The request cannot be processed because there is no consumer of events for the given tenant connected to Hono, or the consumer didn't process the event.
```
sophokles73, line 348: no comma

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -368,7 +376,7 @@ coap-client -u sensor1@DEFAULT_TENANT -k hono-secret -m POST coaps://hono.eclips
 ~~~
 
 {{% note %}}
-In the example above the response actually contains payload that should be used as input to a command to be executed by the device.
+In the example above the response actually contains a payload that should be used as input to a command to be executed by the device.
```
sophokles73, line 379: unspecific

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -406,16 +414,15 @@ send any commands to the device.
     delivered to any potential consumer (yet). However, if the message type is `CON` (*at least once* semantics), then the adapter waits
     for the message to be delivered and accepted by a downstream consumer before responding with this status code.
   * 4.00 (Bad Request): The request cannot be processed. Possible reasons include:
-         * the request body is empty and the *URI-query* option doesn't contain the `empty` parameter
+         * the request body is empty, and the *URI-query* option doesn't contain the `empty` parameter.
```
sophokles73, line 417: no comma

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -406,16 +414,15 @@ send any commands to the device.
...
   * 4.03 (Forbidden): The request cannot be processed because the device's registration status cannot be asserted.
     Possible reasons for this include:
         * The given tenant is not allowed to use this protocol adapter.
         * The given device does not belong to the given tenant.
   * 4.04 (Not Found): The request cannot be processed because the device is disabled or does not exist.
   * 4.13 (Request Entity Too Large): The request cannot be processed because the request body exceeds the maximum supported size.
   * 4.29 (Too Many Requests): The request cannot be processed because the tenant's message limit for the current period is exceeded.
-  * 5.03 (Service Unavailable): The request cannot be processed because there is no consumer of telemetry data for the given tenant
-    connected to Hono.
+  * 5.03 (Service Unavailable): The request cannot be processed because there is no consumer of events for the given tenant connected to Hono, or the consumer didn't process the event.
```
sophokles73, line 425: no comma

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -472,7 +479,7 @@ coap-client -m PUT coap://hono.eclipseprojects.io/event/DEFAULT_TENANT/4711?hono
     delivered to any potential consumer (yet). However, if the message type is `CON` (*at least once* semantics), then the adapter waits
     for the message to be delivered and accepted by a downstream consumer before responding with this status code.
   * 4.00 (Bad Request): The request cannot be processed. Possible reasons include:
-         * the request body is empty and the *URI-query* option doesn't contain the `empty` parameter
+         * the request body is empty, and the *URI-query* option doesn't contain the `empty` parameter.
```
sophokles73, line 482: no comma

site/documentation/content/user-guide/coap-adapter.md
```diff
@@ -482,8 +489,7 @@ coap-client -m PUT coap://hono.eclipseprojects.io/event/DEFAULT_TENANT/4711?hono
   * 4.04 (Not Found): The request cannot be processed because the device is disabled or does not exist.
   * 4.13 (Request Entity Too Large): The request cannot be processed because the request body exceeds the maximum supported size.
   * 4.29 (Too Many Requests): The request cannot be processed because the tenant's message limit for the current period is exceeded.
-  * 5.03 (Service Unavailable): The request cannot be processed because there is no consumer of telemetry data for the given tenant
-    connected to Hono.
+  * 5.03 (Service Unavailable): The request cannot be processed because there is no consumer of events for the given tenant connected to Hono, or the consumer didn't process the event.
```
sophokles73, line 492: no comma

site/documentation/content/user-guide/http-adapter.md
```diff
@@ -306,7 +318,7 @@ content-length: 23
   * 404 (Not Found): The request cannot be processed because the device is disabled or does not exist.
   * 413 (Request Entity Too Large): The request cannot be processed because the request body exceeds the maximum supported size.
   * 429 (Too Many Requests): The request cannot be processed because the tenant's message limit for the current period is exceeded.
-  * 503 (Service Unavailable): The request cannot be processed because there is no consumer of events for the given tenant connected to Hono.
+  * 503 (Service Unavailable): The request cannot be processed because there is no consumer of events for the given tenant connected to Hono, or the consumer didn't process the event.
```
sophokles73, line 321: no comma

site/documentation/content/user-guide/http-adapter.md
```diff
@@ -350,7 +362,7 @@ content-length: 0
...
```
sophokles73, line 365: no comma

site/documentation/content/user-guide/http-adapter.md
```diff
@@ -399,7 +411,7 @@ content-length: 0
...
```
sophokles73, line 414: no comma

site/documentation/content/user-guide/coap-adapter.md
calohmn: Comma used before `and` linking two independent clauses here.

client/src/main/java/org/eclipse/hono/client/ServerErrorException.java
calohmn: The motivation to put it only in the `ServerErrorException` class was that the standard error message of a `ClientErrorException` is already a client-facing one, i.e. it is used in the HTTP response for example:
https://github.com/eclipse/hono/blob/0e440efe3e6fa1d28f5bd4543137fbdda3b609ea/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java#L787-L789

I see a `ServerErrorException` as more of an internal exception, where it makes sense to add a separate error message that abstracts from the internals.

client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
calohmn: Using a resource bundle just for these error messages seems a bit overkill to me. We also don't use a resource bundle for the `ClientErrorException` messages already being included in HTTP responses.
I mainly introduced these constants to match the error message in the unit tests. As we rarely use such error message checks in unit tests, I could just remove them and then also remove the constants.

client/src/main/java/org/eclipse/hono/client/ServerErrorException.java
sophokles73: I see

client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
sophokles73: FMPOV it would actually be the right choice for the problem at hand. I also do not think that it is much effort to implement.

calohmn: So you mean introducing a resource bundle, used then also for the `ClientErrorException` messages?

sophokles73: You mean for the already existing messages? IMHO that should be the goal then, yes. We could maybe use a step-by-step approach here but in the end we should use the same mechanism. FMPOV this is even more desirable if we want to provide more elaborate messages than *bad JSON* ...

sophokles73:
@calohmn any news on this?

calohmn:
@sophokles73 Just pushed an update, using a resource bundle now.

sophokles73 (commented):
The more I look at it the more I think that we should actually simply use specific exceptions representing the three error codes that you have introduced. Each of the exception classes could then define its resource bundle key and provide a (localizable) message via its standard *Throwable.getMessage()* method. Client code could choose whether it wants the exception to provide a custom message or its standard client facing message by means of providing a message in the constructor or not.
IMHO this would make the code much easier to understand and use. WDYT?

client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
```diff
@@ -304,10 +307,15 @@ protected final Span startSpan(final Message message) {
                                 : new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST, rejected.getError().getDescription());
                     } else if (Released.class.isInstance(remoteState)) {
                         e = new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE);
+                        ((ServerErrorException) e).setClientFacingMessage(ServerErrorMessage.MESSAGE_NOT_PROCESSED);
                     } else if (Modified.class.isInstance(remoteState)) {
                         final Modified modified = (Modified) deliveryUpdated.getRemoteState();
-                        e = modified.getUndeliverableHere() ? new ClientErrorException(HttpURLConnection.HTTP_NOT_FOUND)
-                                : new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE);
+                        if (modified.getUndeliverableHere()) {
+                            e = new ClientErrorException(HttpURLConnection.HTTP_NOT_FOUND);
```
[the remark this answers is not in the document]
calohmn: yes, but I think the "not found/don't try again" semantics of the exception would then need to be put into the message as well. How about using
`SERVER_ERROR_MESSAGE_UNDELIVERABLE=consumer declared message as undeliverable, it should not be redelivered`
here?

calohmn:
>Each of the exception classes could then define its resource bundle key and provide a (localizable) message via its standard *Throwable.getMessage()* method. Client code could choose whether it wants the exception to provide a custom message or its standard client facing message by means of providing a message in the constructor or not.

As for the *ClientErrorExceptions*, I agree that the standard *Throwable.getMessage()* method can be used.

For the *ServerErrorExceptions*, I would rather keep a separate field for the *client facing message*. The *Throwable.getMessage()* method should return the *internal* error message then (for logging). With the extra field/method, it can also be easily decided whether the exception instance actually has a message that shall be given to an external client.

I have amended the commit so that there is a separate exception class for each of the new messages now.

tests/src/test/java/org/eclipse/hono/tests/http/TelemetryHttpIT.java
```diff
@@ -214,6 +221,10 @@ public void testUploadQos1MessageFailsIfDeliveryStateNotUpdated(final VertxTestC
 
         httpResponseFuture
                 .onComplete(ctx.succeeding(response -> {
+                    ctx.verify(() -> {
+                        assertThat(response.bodyAsString())
+                                .isEqualTo(new SendMessageTimeoutException("").getClientFacingMessage());
+                    });
```
sophokles73, line 227: wow, this looks terrible, doesn't it ;-)
Maybe we can add a *static String ServiceInvocationException.getLocalizedMessage(String)* method which accepts a key instead?

calohmn: Well, I agree that it doesn't look pretty and I was actually a bit hesitant, using that. However, it's in a test and as an advantage, the resource handling and keys can be kept internal to the exception classes that way.

But ok, I've changed it to use *ServiceInvocationException.getLocalizedMessage()* now.

client/src/main/java/org/eclipse/hono/client/NoConsumerException.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public class NoConsumerException extends ServerErrorException {
+
+    /**
+     * Resource key for the client facing error message.
+     */
+    public static final String CLIENT_FACING_MESSAGE_KEY = "SERVER_ERROR_NO_CONSUMER";
+
+    /**
+     * Creates a new NoConsumerException.
```
sophokles73, line 29: IMHO it would be helpful to explain that apart from the given message, the *clientFacingMessage* will be set to a certain key. This is also true for the other exception classes. WDYT?

calohmn: I've amended the commit accordingly.

sophokles73 (approved):
LGTM

PR #2186, opened 2020-09-17
Remove redundant stripping of private info while fetching credentials in MongoDB device registry

kaniyan:
Recently removing of private info from credentials using `CommonCredential.stripPrivateInfo()` has been incorporated in `AbstractCredentialsManagementService`. _MongoDB_ registry implementation already makes use of this abstract class. Hence invocation of `stripPrivateInfo()` has been removed from the _MongoDB_ registry implementation.


sophokles73 (approved):
LGTM

Issue #2184, opened 2020-09-16
File and Mongo DB based registries return pre-shared key via management API 


sophokles73:
The file based and Mongo DB based registries both include PSK credentials' shared *key* property in the response body of the *read credentials* operation. However, the key should not be included, analogously to the hashed password credentials

kaniyan:
@sophokles73 Could you please assign this to me, if none is already working on it.

sophokles73:
I have already fixed this, PR pending ...

kaniyan:
> I have already fixed this, PR pending ...

:+1:

PR #2188, opened 2020-09-17
[#2184] Strip shared key from PSK credentials

sophokles73:
The device registry implementations do no longer include the (secret)
shared key of PSK credentials in the result set of the Management API's
get credentials operation.

### DIFF
--- a/services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedCredentialsService.java
+++ b/services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedCredentialsService.java
@@ -551,7 +551,7 @@ public final class FileBasedCredentialsService implements CredentialsManagementS
                                     || secret.containsKey(CredentialsConstants.FIELD_SECRETS_PWD_HASH)
                                     || secret.containsKey(CredentialsConstants.FIELD_SECRETS_SALT)) {
 
-                                removePasswordDetailsFromSecret(newSecret);
+                                removeConfidentialDataFromSecret(newSecret);
                             }
                             // then copy the new details.
                             for (String field : secret.fieldNames()) {
@@ -652,8 +652,8 @@ public final class FileBasedCredentialsService implements CredentialsManagementS
         for (final Object credential : matchingCredentials) {
             final JsonObject credentialsObject = (JsonObject) credential;
             credentialsObject.remove(CredentialsConstants.FIELD_PAYLOAD_DEVICE_ID);
-            removePasswordDetailsFromCredential(credentialsObject);
             final CommonCredential cred = credentialsObject.mapTo(CommonCredential.class);
+            cred.stripPrivateInfo();
             credentials.add(cred);
         }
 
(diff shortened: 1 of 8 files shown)


services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedCredentialsService.java
```diff
@@ -652,7 +652,7 @@ private void removeAllForDevice(final String tenantId, final String deviceId, fi
         for (final Object credential : matchingCredentials) {
             final JsonObject credentialsObject = (JsonObject) credential;
             credentialsObject.remove(CredentialsConstants.FIELD_PAYLOAD_DEVICE_ID);
-            removePasswordDetailsFromCredential(credentialsObject);
+            removeConfidentialDataFromCredential(credentialsObject);
```
kaniyan, line 655: We could reuse `CommonCredential.stripPrivateInfo(...)` instead of `removeConfidentialDataFromCredential(...)` after mapping to `CommonCredential` in the following line.

kaniyan (changes requested):

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedCredentialsService.java
sophokles73: that is an excellent idea :+1:

dejanb (approved):
LGTM

akofoed:
Hi Bosch IoT team,
Removing the PSK (secrets array) from the Device Registry HTTP GET operation, makes it impossible for the Cloud Protocol Gateway to authenticate the sensor-device:
https://docs.bosch-iot-suite.com/asset-communication/Cloud-Protocol-Gateway---operated-by-a-customer.html

According to the documentation on the above link, the Cloud Protocol Gateway should be able to access the Device Registry in order to authenticate the device:
> In case the devices connected via the Cloud Protocol Gateway prefer to store their own credentials in the Hub service, please consider that the credential registration is only supported via HTTP API at the moment.

This pull removes that ability.

The question therefore becomes; How do the Cloud Protocol Gateway authenticate a sensor-device? Using auth-id (user-id) is not enough as any attacker can claim this user-id. The Protocol Gateway clearly need a way to retrieve the PSK or crypt password in order to securely authenticate the sensor-device.

sophokles73:
> Hi Bosch IoT team,

It is true that many people working on Hono are employed by Bosch. However, we still consider this to be the *Hono team* :-)
Any questions regarding Bosch's commercial offering(s) should be asked via those offerings' support channels.

That said, if you were to implement a protocol gateway as described in [Hono's documentation](https://www.eclipse.org/hono/docs/concepts/connecting-devices/) then the gateway should retrieve the device credentials using the [Credentials API](https://www.eclipse.org/hono/docs/api/credentials/) as the protocol adapters do. This might require configuration of network infrastructure and authorizations in order to allow the gateway to access the Credentials service implementation's AMQP endpoint.

akofoed:
Sorry for that.

For people like me, developing a Gateway solution with some sensor-devices attached to it (wireless or otherwise), it makes sense for the gateway/sensors to be cloud-agnostic. The user then have the choice to connect the gateway to his own Hono instance or Bosch IoT or RedHat IoT or some other Hono-hosted environment. This is how standards are created. Take the history of VHS vs Betamax or Ethernet vs Tokenring. The cheapest solution that solves the problem wins.

To the issue of authenticating a sensor-device behind a gateway connected to Hono; Right now my Gateway solution fetches the PSK from the Device Registry using HTTP GET. The neat thing about this is the portability between clouds; No Hono Adaptor is needed, it just works straight out of the box. No proprietary Hono Adaptor just for my particular Gateway solution is needed. Again - standards.

I do understand the reasoning behind this pull, and I do see that it makes sense not to be able to fetch the password. However, for gateway solutions, these **must** have a way to authenticate their connected sensor-devices. Since these sensor-device are created in Hono Device Registry anyway, the only thing that is needed is a way for the gateway to retrieve the PSK (or crypt/x509).

Say the Gateway provider were to provide an Adaptor (like LoRa or SigFox), what are the chances of asking the cloud-provider (Bosch/RedHat/other) to install this adaptor in their cloud? None is my guess. The only function of the Gateway Adaptor would then be to authenticate the devices, as, in my opinion, payload normalization should not be done in the adaptor.

My hope is that the Hono Team will help find a solution to this problem.

Kind regards
Anders

akofoed:
A solution could be to allow the Gateway to retrieve the device credentials through the Adaptors:
COAP: GET /credentials/tenant/device-id
HTTP: GET /credentials/tenant/device-id
MQTT: publish to credentials/tenant/ topic - subscribe to credentials/tenant/+
AMQP: publish to credentials/tenant/ topic - subscribe to credentials/tenant/+

The Hono Adaptor will then check that the device has the requesting Gateway in its 'via' or 'viaGroups' and if so, send the device secrets[] back to the Gateway.

sophokles73:
> To the issue of authenticating a sensor-device behind a gateway connected to Hono; Right now my Gateway solution fetches the PSK from the Device Registry using HTTP GET. The neat thing about this is the portability between clouds; No Hono Adaptor is needed, it just works straight out of the box. No proprietary Hono Adaptor just for my particular Gateway solution is needed. Again - standards.

I am not sure if I fully understand your use case and setup. Hono's *raison d'etre* are its protocol adapters. That is where devices and/or gateways connect to. Hono defines a set of APIs that the adapters use to authenticate devices and authorize messages. These are the Tenant, Device Registration, Credentials and (depending on the use case) the Device Connection API.

The Device Registry Management API is intended to be used for provisioning devices. However, Hono does not require any particular registry to actually implement this API. In fact, in many situations there will already be an existing system which contains the device identities and/or credentials from which Hono's adapters should retrieve the required information. In such a case, the Tenant, Device Registration and Credentials APIs would need to be implemented as adapters for the existing (legacy) system. Again, Hono's protocol adapters make no assumption whatsoever as to how the implementation of these APIs looks like.

My understanding of your setup is that you have implemented a component that relies on Hono's Device Registry Management API in order to retrieve credentials for authenticating devices/sensors. This, however, is not what that API is intended for. Instead, the Credentials API has been designed for that purpose. There should be nothing preventing your component from using the Credentials API directly, i.e. no need to connect to any protocol adapter in this case.

The Bosch IoT Hub's device registry currently doesn't implement Hono's Device Registry Management API. The documentation you are referring to seems to mention IoT Hub's registry API only in the context of *provisioning* devices. It also explicitly mentions that authentication of devices/sensors is the responsibility of the protocol gateway. I do not know if the IoT Hub provides access to its Credentials API for protocol gateways. You should probably ask the IoT Hub support team about that.

I agree that standardization of APIs is very important in this context and provides value and I think that is exactly what we are trying to achieve here.

PR #2192, opened 2020-09-17
[#2115] Add PreCredentialsValidationHandler (obsolete)

calohmn:
This is for #2115.

Adds an optional handler to the AuthHandler classes that is invoked before the actual (possibly expensive) credentials validation is performed. This handler can do checks using credentials and TenantObject obtained from the authentication data.

### DIFF
--- a/service-base/src/main/java/org/eclipse/hono/service/auth/device/ExecutionContextAuthHandler.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/auth/device/ExecutionContextAuthHandler.java
@@ -47,12 +47,21 @@ public abstract class ExecutionContextAuthHandler<T extends ExecutionContext> im
 
     private final DeviceCredentialsAuthProvider<?> authProvider;
 
+    private final PreCredentialsValidationHandler<T> preCredentialsValidationHandler;
+
     /**
      * Creates a new handler for authenticating MQTT clients.
      *
      * @param authProvider The auth provider to use for verifying a client's credentials.
+     * @param preCredentialsValidationHandler An optional handler to invoke after the credentials got determined and
+     *            before they get validated. Can be used to perform checks using the credentials and tenant information
+     *            before the potentially expensive credentials validation is done. A failed future returned by the
+     *            handler will fail the corresponding authentication attempt.
      */
-    protected ExecutionContextAuthHandler(final DeviceCredentialsAuthProvider<?> authProvider) {
+    protected ExecutionContextAuthHandler(
+            final DeviceCredentialsAuthProvider<?> authProvider,
+            final PreCredentialsValidationHandler<T> preCredentialsValidationHandler) {
+        this.preCredentialsValidationHandler = preCredentialsValidationHandler;
         this.authProvider = authProvider;
     }
 
@@ -68,21 +77,30 @@ public abstract class ExecutionContextAuthHandler<T extends ExecutionContext> im
 
         return parseCredentials(context)
                 .compose(authInfo -> {
+                    final DeviceCredentialsAuthProvider<?> authProvider = getAuthProvider(context);
+                    if (authProvider == null) {
+                        return Future.failedFuture(new IllegalStateException("no auth provider found"));
+                    }
                     final Promise<DeviceUser> authResult = Promise.promise();
-                    getAuthProvider(context).authenticate(authInfo, context, authResult);
+                    authProvider.authenticate(authInfo, context, authResult, preCredentialsValidationHandler);
                     return authResult.future();
                 });
     }
 
     private DeviceCredentialsAuthProvider<?> getAuthProvider(final T ctx) {
 
+        // check whether a compatible provider was put in the context;
+        // the ChainAuthHandler does this so that the AuthProvider of one of its contained AuthHandlers is used
         final Object obj = ctx.get(AUTH_PROVIDER_CONTEXT_KEY);
         if (obj instanceof DeviceCredentialsAuthProvider<?>) {
-            log.debug("using auth provider found in context [type: {}]", obj.getClass().getName());
+            log.debug("using auth provider found in context");
             // we're overruling the configured one for this request
             return (DeviceCredentialsAuthProvider<?>) obj;
         } else {
-            // bad type, ignore and return default
+            // no provider in context or bad type
+            if (obj != null) {
+                log.warn("unsupported auth provider found in context [type: {}]", obj.getClass().getName());
+            }
             return authProvider;
         }
     }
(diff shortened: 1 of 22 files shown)


calohmn:
(Amended to use a generic type in `PreCredentialsValidationHandler`.)

service-base/src/main/java/org/eclipse/hono/service/auth/device/PreCredentialsValidationHandler.java
```diff
@@ -0,0 +1,38 @@
...
+public interface PreCredentialsValidationHandler<T extends ExecutionContext> {
+
+    /**
+     * Invokes the handler.
+     *
+     * @param credentials The credentials
+     * @param executionContext The execution context. It has to be ensured that invoking
+     *              {@link ExecutionContext#getTenantObject()} on the given context doesn't return {@code null}.
+     * @return A future indicating the outcome of the operation.
+     * @throws NullPointerException If any of the parameters or the tenant of the executionContext is {@code null}.
```
sophokles73, line 35: shouldn't a context that has no tenant object better yield an `IllegalArgumentException`?

service-base/src/main/java/org/eclipse/hono/service/auth/device/ExecutionContextAuthHandler.java
```diff
@@ -47,12 +47,21 @@
...
      * Creates a new handler for authenticating MQTT clients.
      *
      * @param authProvider The auth provider to use for verifying a client's credentials.
+     * @param preCredentialsValidationHandler An optional handler to invoke after the credentials got determined and
+     *            before they get validated. Can be used to perform checks using the credentials and tenant information
+     *            before the potentially expensive credentials validation is done. A failed future returned by the
+     *            handler will fail the corresponding authentication attempt.
      */
-    protected ExecutionContextAuthHandler(final DeviceCredentialsAuthProvider<?> authProvider) {
+    protected ExecutionContextAuthHandler(
```
sophokles73, line 61: The *authProvider* is mandatory, isn't it?
Why do we inject the PreCredentialsValidationHandler into the AuthHandler where it is not used at all? Wouldn't it be more straightforward to add it to the AuthProviders where it is actually being used/invoked?

calohmn: > The authProvider is mandatory, isn't it?

No, it may actually be `null`. That is currently the case in the `ChainAuthHandler`.

> Why do we inject the PreCredentialsValidationHandler into the AuthHandler where it is not used at all? Wouldn't it be more straightforward to add it to the AuthProviders where it is actually being used/invoked?

Good point. Initially, I had the `PreCredentialsValidationHandler` as a field in the AuthProviders. But I found these downsides:
- as in Hono oftentimes the ChainAuthHandler and thereby multiple AuthProvider instances are used, it looks convenient to set the `PreCredentialsValidationHandler` only once for the ChainAuthHandler, making sure all its AuthProviders use the same PreCredentialsValidationHandler.
- the existing Hono AuthProvider implementations can easily be instantiated as beans (`@Autowired` constructors) and I wouldn't want to require a `PreCredentialsValidationHandler` bean for these constructors here. One could argue though that this isn't really relevant because the Hono AuthProvider instances don't seem to be defined as beans in Hono currently.

Actually, I'll be having a second look on this topic, on what it would mean also for the HTTP AuthProviders (concerning changes aimed for a subsequent PR)...

calohmn: I've had another look here: 
In this PR, I have parameterized the `PreCredentialsValidationHandler` class with the type of `ExecutionContext` used, when the Hono AuthHandler invokes the `authenticateDevice(executionContext)` method. This looks quite useful for the specific PreCredentialsValidationHandler implementations.

When putting the `PreCredentialsValidationHandler` instance in an `AuthProvider` *field* like suggested above, I would have to parameterize the `AuthProvider` class with the type of `ExecutionContext` (`E extends ExecutionContext`). But when the `authenticate(JsonObject, Handler<AsyncResult<User>>)` method of the vert.x `AuthProvider` base class is used, I can only create a `GenericExecutionContext` object, to pass along to the `CredentialsApiAuthProvider#authenticate(jsonObject, E executionContext, handler)` method where the `PreCredentialsValidationHandler` is used. 
So, this doesn't really work.

With the solution in this PR, the above problem doesn't exist because the ExecutionContext type is deduced here (in `DeviceCredentialsAuthProvider`) which is more flexible:
````
<S extends ExecutionContext> void authenticate(
            JsonObject authInfo,
            S executionContext,
            Handler<AsyncResult<DeviceUser>> resultHandler,
            PreCredentialsValidationHandler<S> preCredentialsValidationHandler);
````

So, I would stick to the approach in the PR.

Comment (sophokles73) [reply service-base/src/main/java/org/eclipse/hono/service/auth/device/ExecutionContextAuthHandler.java:61]: I see. Alternatively, we could already invoke the *PreCredentialsValidationHandler* in the AuthHandler's *authenticateDevice* method, couldn't we? That way we wouldn't need to pass it on to the AuthProvider ...

Comment (calohmn) [reply service-base/src/main/java/org/eclipse/hono/service/auth/device/ExecutionContextAuthHandler.java:61]: The `org.eclipse.hono.service.auth.device.AuthHandler#authenticateDevice(T)` method implementation only has the `authInfo` JsonObject available. The `Credentials` object used in the `PreCredentialsValidationHandler` is only obtained in the AuthProvider. Therefore the handler has to be invoked in the AuthProvider.

File: service-base/src/main/java/org/eclipse/hono/service/auth/device/ExecutionContextAuthHandler.java [hunk 3]
```diff
@@ -69,7 +78,7 @@ protected ExecutionContextAuthHandler(final DeviceCredentialsAuthProvider<?> aut
         return parseCredentials(context)
                 .compose(authInfo -> {
                     final Promise<DeviceUser> authResult = Promise.promise();
-                    getAuthProvider(context).authenticate(authInfo, context, authResult);
+                    getAuthProvider(context).authenticate(authInfo, context, authResult, preCredentialsValidationHandler);
```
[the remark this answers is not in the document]
calohmn: Yes, I've now improved error handling here.

service-base/src/main/java/org/eclipse/hono/service/auth/device/ExecutionContextAuthHandler.java
sophokles73: Well, then why don't we simply add the *getCredentials(JsonObject)* method to the *DeviceCredentialsAuthProvider* interface?

calohmn: After trying it out, I now see that this is indeed the better approach.
I have created  #2199 with it to replace this PR.

calohmn:
Closed in favour of  #2199.

PR #2196, opened 2020-09-21
Orig address and content type added to headers which are passed to the mapper

BobClaerhout:

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/impl/HttpBasedMessageMapping.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/impl/HttpBasedMessageMapping.java
@@ -35,6 +35,7 @@ import io.vertx.core.Handler;
 import io.vertx.core.MultiMap;
 import io.vertx.core.Promise;
 import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.HttpHeaders;
 import io.vertx.core.json.Json;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.client.HttpResponse;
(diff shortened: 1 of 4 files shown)


adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/impl/HttpBasedMessageMapping.java
```diff
@@ -121,6 +121,8 @@ private void mapMessageRequest(
                 headers.add(property.getKey(), Json.encode(value));
             }
         });
+        headers.add(MessageHelper.APP_PROPERTY_ORIG_ADDRESS, ctx.message().topicName());
+        headers.add(MessageHelper.SYS_PROPERTY_CONTENT_TYPE, ctx.contentType());
```
sophokles73, line 125: Why don't we use the standard HTTP *Content-Type* header?

BobClaerhout: you're right, the standard HTTP header is better, updated.

sophokles73:
@BobClaerhout not sure if you noticed but the unit tests are failing for your changes ...

BobClaerhout:
> @BobClaerhout not sure if you noticed but the unit tests are failing for your changes ...

Fixed it, test now fails due to something with mongo?

sophokles73:
@BobClaerhout there doesn't seem to be any assertions in the unit tests that verify the inclusion of the new properties.
I also think that it would be good to amend the documentation of the mapping feature in the MQTT adapter guide with the properties that are being forwarded to the mapper so that implementers know what to expect. WDYT?

BobClaerhout:
> @BobClaerhout there doesn't seem to be any assertions in the unit tests that verify the inclusion of the new properties.
> I also think that it would be good to amend the documentation of the mapping feature in the MQTT adapter guide with the properties that are being forwarded to the mapper so that implementers know what to expect. WDYT?

I updated and added some basic documentation as well. Don't know whether this is enough.

sophokles73 (approved):
LGTM

PR #2199, opened 2020-09-23
[#2115] Add PreCredentialsValidationHandler

calohmn:
This is for #2115 and replaces the previous PR #2192.
Adds an optional handler to the AuthHandler classes that is invoked before the actual (possibly expensive) credentials validation is performed. This handler can do checks using tenant information obtained from the authentication data.
Recent changes to the DeviceCredentialsAuthProvider interface have been reverted (removing the ExecutionContext dependency) and a getCredentials(JsonObject) method has been added.

In comparison to the previous PR #2192, I have removed the code that has obtained the TenantObject before calling the PreCredentialsValidationHandler (it was put in the ExecutionContext then). This removes the need to pass on the TenantClientFactory in all AuthHandler constructors.


service-base/src/main/java/org/eclipse/hono/service/auth/device/CredentialsApiAuthProvider.java
```diff
@@ -206,47 +203,19 @@ public final void authenticate(
...
             resultHandler.handle(Future.failedFuture(new ClientErrorException(HttpURLConnection.HTTP_UNAUTHORIZED, "malformed credentials")));
         } else {
-            authenticate(credentials, executionContext, resultHandler);
+            authenticate(credentials, TracingHelper.extractSpanContext(tracer, authInfo), s -> {
+                if (s.succeeded()) {
+                    resultHandler.handle(Future.succeededFuture(s.result()));
+                } else {
+                    resultHandler.handle(Future.failedFuture(s.cause()));
+                }
+            });
```
sophokles73, line 217: `authenticate(credentials, TracingHelper.extractSpanContext(tracer, authInfo), resultHandler)` ?

calohmn: This won't work because of the generic AsyncResult type:
`resultHandler` is of type `Handler<AsyncResult<User>>` (coming from the vert.x `AuthProvider` interface, whereas the `DeviceCredentialsAuthProvider.authenticate` method expects the more specific `Handler<AsyncResult<DeviceUser>>`.

sophokles73: I always get this wrong ;-)

sophokles73 (approved):
LGTM

PR #2200, opened 2020-09-23
Add missing dependency for printing test log output

calohmn:
Fixes no logs being printed out when running the unit tests in module hono-adapter-http-vertx-base and hono-adapter-mqtt-vertx-base.


sophokles73 (approved):
LGTM

PR #2203, opened 2020-09-23
[#2115] Use PreCredentialsValidationHandler in AMQP adapter

calohmn:
This is for #2115.
Setting the trace-sampling-priority according to the tenant configuration when processing a telemetry/event message in the AMQP adapter is now done via the new PreCredentialsValidationHandler. This removes duplicate code related to extracting the tenant information from the authentication data.


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -197,6 +196,37 @@ protected void doStart(final Promise<Void> startPromise) {
...
+        final String tenantId = credentials.getTenantId();
+        final Span span = executionContext.getTracingSpan();
+        final String authId = credentials.getAuthId();
+
+        return getTenantConfiguration(tenantId, span.context())
+                .compose(tenantObject -> {
+                    TracingHelper.setDeviceTags(span, tenantId, null, authId);
+                    final OptionalInt traceSamplingPriority = TenantTraceSamplingHelper.applyTraceSamplingPriority(
+                            tenantObject, authId, span);
+                    executionContext.getProtonConnection().attachments().set(
```
sophokles73, line 224: FMPOV we should not rely on the internal structure of the SaslResponseContext here but instead simply invoke something like *SaslResponseContext.setTraceSamplingPriority(TenantObject)* to improve encapsulation ...

calohmn: Yes, however it's not so straightforward at the moment because the places where you would want to call `getTraceSamplingPriority` don't have the `SaslResponseContext` object available.
I wanted to replace the various different objects being put into the connection attachments with one single "ConnectionContext" object anyway. I would rather do this in a follow-up PR then though.

sophokles73 (approved):
LGTM

PR #2204, opened 2020-09-23
Extend search devices operation to support wildcards in the MongoDB device registry

kaniyan:
With this PR, the MongoDB device registry supports wildcard matching in search devices operation. The supported wildcard characters are  `*`  and `?`.

### DIFF
--- a/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
+++ b/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
@@ -18,10 +18,13 @@ import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
 import java.time.Instant;
+import java.util.Collections;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.StringTokenizer;
 import java.util.UUID;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import javax.security.auth.x500.X500Principal;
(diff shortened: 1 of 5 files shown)


services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDocumentBuilder.java
```diff
@@ -142,10 +143,18 @@ public MongoDbDocumentBuilder withAuthId(final String authId) {
     public MongoDbDocumentBuilder withDeviceFilters(final List<Filter> filters) {
 
         filters.forEach(filter -> {
-            // TODO: To implement when filter values contain patterns such as * or %
-            document.put(mapDeviceField(filter.getField()), filter.getValue());
+            if (filter.getValue() instanceof String) {
+                final String value = (String) filter.getValue();
+                if (value.contains("*") || value.contains("?")) {
```
sophokles73, line 148: does it make a big difference to handle the two cases differently? I would assume that MongoDB internally treats both cases the same anyway, i.e. wrapping a literal string with `^` and `$`, thus turning it into a regex.
We could then simply let *DeviceRegistryUtils.getRegexExpressionForSearchOperation* handle all filters ...

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
```diff
@@ -284,4 +287,36 @@ public static boolean matchesWithClientContext(final JsonObject credential,
...
+     */
+    public static String getRegexExpressionForSearchOperation(final String filterValue) {
+        Objects.requireNonNull(filterValue);
+
+        return Collections.list(new StringTokenizer(filterValue, "*?", true))
+                .stream()
+                .map(token -> (String) token)
+                .map(token -> {
+                    if (token.equals("*")) {
+                        return "(.*)";
```
sophokles73, line 312: is there a particular reason why you want to use the brackets here?

kaniyan: The grouping has no special meaning here. I personally find that it improves the readability in log statements.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDocumentBuilder.java
kaniyan: I didn't find any official info on how it is being handled internally by MongoDB. Hence I used the above approach. If the value doesn't contain '*' or '?', then the query wouldn't involve _regex_ keyword. Based on the above approach it would be 
`{"device-id" : "4711"}}`. Else it would be`{"field",  {"device-id" : {"regex" : "^4711$"}}`. If you think it doesn't make any difference, I can treat both the cases same.

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/util/DeviceRegistryUtils.java
sophokles73: I would assume that it makes evaluating the regex more expensive because of the grouping, but that might be negligible

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDocumentBuilder.java
sophokles73: we can refactor it if we run into another place where we need to transform wildcards into a regex ...

sophokles73 (approved):
LGTM

PR #2209, opened 2020-09-24
Remove CoapContextTenantAndAuthIdProvider

calohmn:
The CoAP adapter shouldn't rely on the `BaseExecutionContextTenantAndAuthIdProvider` base class anymore because that will soon be removed (as part of #2115).


sophokles73 (approved):
LGTM

PR #2210, opened 2020-09-24
[#2115] Use PreCredentialsValidationHandler in MQTT adapter

calohmn:
This is for #2115.
Setting the trace-sampling-priority according to the tenant configuration when processing a telemetry/event message in
the MQTT adapter is now done via the new PreCredentialsValidationHandler. This removes duplicate code related to extracting the tenant information from the authentication data in the CONNECT packet.


sophokles73 (approved):
LGTM

PR #2213, opened 2020-09-25
Update monitoring-tracing-config.md

vasanth-bk:
Typo in line number 85. "nor" should be used there instead of "not"
The component/service will fail to start if neither the secure nor the insecure server have been configured properly.


sophokles73:
You will need to add the `Signed-off-by` footer to the commit message. See https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for

PR #2214, opened 2020-09-25
MQTT adapter: Fix NPE in case command payload is null

calohmn:
This fixes a potential NPE when sending a command with an empty (ie. unsupported) payload to an MQTT device.
(This NPE caused the command to be released via the exception handling.)

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
@@ -1460,7 +1460,8 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends MqttProtoc
         Tags.MESSAGE_BUS_DESTINATION.set(commandContext.getTracingSpan(), publishTopic);
         TracingHelper.TAG_QOS.set(commandContext.getTracingSpan(), subscription.getQos().name());
 
-        endpoint.publish(publishTopic, command.getPayload(), subscription.getQos(), false, false, sentHandler -> {
+        final Buffer payload = Optional.ofNullable(command.getPayload()).orElseGet(Buffer::buffer);
+        endpoint.publish(publishTopic, payload, subscription.getQos(), false, false, sentHandler -> {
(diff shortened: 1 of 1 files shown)


calohmn:
On a more general note: 
The return value of `MessageHelper.getPayload()`, which is used here, is null if the message doesn't contain a supported body type
or if the body is null (which I think would mean an invalid message according to the AMQP spec).
See https://github.com/eclipse/hono/blob/1.4.0/core/src/main/java/org/eclipse/hono/util/MessageHelper.java#L410-L416

Therefore I think we should in general consider such a command message invalid and reject it early on.
@sophokles73 WDYT?

sophokles73:
What about a very simple command like *turnOn*? This doesn't necessarily need input parameters and thus should not require a body in the AMQP message, should it?

sophokles73 (approved):
LGTM

calohmn:
> What about a very simple command like turnOn? This doesn't necessarily need input parameters and thus should not require a body in the AMQP message, should it?

I think the [AMQP spec](http://docs.oasis-open.org/amqp/core/v1.0/amqp-core-messaging-v1.0.html#section-message-format) is not very clear on whether the body section can be omitted. It doesn't explicitly say "Zero or one body sections" as with the other sections, it just says 
`The body consists of one of the following three choices: one or more data sections, one or more amqp-sequence sections, or a single amqp-value section. `.

But, yes, better to allow a `null` body here. 

However, in my above comment I was more thinking about the other cases where `MessageHelper.getPayload` returns `null`. Namely, when the `amqp-value` section contains some kind of numeric value, not a String or byte array, for example. I think we should consider this an invalid command, or alternatively use a String representation when passing the command to the device.

sophokles73:
Actually, the C&C API allows commands to contain a single *data* section only. The only case where we used to support an *amqp-value* was in request messages to the registry APIs in order to support JMS clients to use text messages without specifying a content-type.
Thus, FMPOV we can (and should) reject command messages containing a non-data section as they are not in line with the API anyway.

calohmn:
Removing the (undocumented) support for command messages with a (string/byte array) amqp-value section may break existing northbound applications (which are possibly also JMS based and choosing not to set the optional content-type). I don't really see the need to do that.
If we really want to do that, I would rather postpone that to Hono 1.6 and add a note in the C&C API about that in Hono 1.5.

PR #2216, opened 2020-09-25
[#8] Preview: Kafka based client.

b-abel:
I would like to discuss here the early stages of the development of the Kafka based client ([#8]). This is not intended to be merged, but it can be tested.

It uses the new client interfaces from #2168 which are not in master yet. The support of those interfaces in the protocol adapters is out of scope for this PR. Therefore I just "hacked" support for the new interfaces into the HTTP and MQTT adapters. This is the first of my commits here. I do not recommend to spend time reviewing this commit :smile:


b-abel:
# Howto use the Kafka clients in Minikube

This brief walk-through guide shows how to build Hono with the Kafka client and run it against a Kafka cluster in Minikube.

## Prerequisites

* Minikube installed
* Helm 3 installed

## Create Kubernetes Cluster

Start / create Minikube instance with

    minikube start --cpus 4 --memory 8192

Open a second shell to start a `tunnel`

    minikube tunnel


## Install Strimzi

The [Strimzi](https://strimzi.io/) project provides a way to run a Kafka cluster on Kubernetes. We deploy it in Minikube.

Add the Helm repo

    eval $(minikube docker-env)
    helm repo add strimzi https://strimzi.io/charts/
    helm repo update

Deploy the operator to default name space using release name `strimzi`:

    helm install strimzi strimzi/strimzi-kafka-operator

### Deploy kafka

Now we deploy a Kafka cluster. For details see [Strimzi's deployment guide](https://strimzi.io/docs/operators/latest/deploying.html#kafka-cluster-str).

* Download strimzi: https://github.com/strimzi/strimzi-kafka-operator/releases/download/0.19.0/strimzi-0.19.0.tar.gz
* extract the archive
* edit file `strimzi-0.19.0/examples/kafka/kafka-ephemeral-single.yaml`: set property `metadata.name` to `hono-kafka-cluster`
* deloy the Kafka cluster `kubectl apply -f examples/kafka/kafka-ephemeral-single.yaml`

To verify the Kafka deployment use the example clients.
Produce messages with:

    kubectl run kafka-producer -ti --image=strimzi/kafka:0.19.0-kafka-2.5.0 --rm=true --restart=Never -- bin/kafka-console-producer.sh --broker-list hono-kafka-cluster-kafka-bootstrap:9092 --topic my-topic

and consume them with:

    kubectl run kafka-consumer -ti --image=strimzi/kafka:0.19.0-kafka-2.5.0 --rm=true --restart=Never -- bin/kafka-console-consumer.sh --bootstrap-server hono-kafka-cluster-kafka-bootstrap:9092 --topic my-topic --from-beginning


## Deploy hono

Create name space `hono`:

    kubectl create namespace hono

Build Hono images locally. Change to Hono's base directory. Take care to in the correct branch (and if in a new shell, do not forget to invoke `eval $(minikube docker-env)`).

    # in hono's base dir
    mvn clean install -Pbuild-docker-image,metrics-prometheus


The Helm chart in the [Eclipse Packages repository](https://github.com/eclipse/packages/tree/master/charts/hono) does not support the configuration for Kafka yet, so it must be adapted.

Get Hono's Helm chart:

    git clone  https://github.com/eclipse/packages.git

Change in that directory:

  cd packages/charts

and edit the file `hono/templates/hono-adapter-http/hono-adapter-http-vertx-secret.yaml`: add the following at the end of the `application.yml` part:

```yaml
      kafka:
        producerConfig:
          bootstrap.servers: hono-kafka-cluster-kafka-bootstrap.default:9092
```

**NB:** make sure that `kafka` is indented one level below `hono`.

Deploy the HTTP adapter and only required components of Hono (remember to issue `eval $(minikube docker-env)` if in a new shell):

```
# in the "charts" directory of the packages repo
helm install --dependency-update -n hono --set honoImagesTag=1.5.0-SNAPSHOT --set adapters.amqp.enabled=false --set adapters.mqtt.enabled=false --set prometheus.createInstance=false --set grafana.enabled=false eclipse-hono ./hono
```

Verify the installation:

    kubectl get -n hono pods

and wait until all pods have status 'Running'.

Check the services with:

    kubectl get -n hono pods

Hint: If you see problems here, check that `minikube tunnel` is running.

## Send messages

Create tenant, device and credentials:

```bash
export REGISTRY_IP=$(kubectl get service eclipse-hono-service-device-registry-ext --output="jsonpath={.status.loadBalancer.ingress[0]['hostname','ip']}" -n hono)
export HTTP_ADAPTER_IP=$(kubectl get service eclipse-hono-adapter-http-vertx --output="jsonpath={.status.loadBalancer.ingress[0]['hostname','ip']}" -n hono)

export MY_DEVICE=TEST_DEVICE
export MY_TENANT=TEST_TENANT
export MY_PWD=secret
# add tenant
curl -i -X POST http://$REGISTRY_IP:28080/v1/tenants/$MY_TENANT
# add device
curl -i -X POST http://$REGISTRY_IP:28080/v1/devices/$MY_TENANT/$MY_DEVICE
# add credentials
curl -i -X PUT -H "content-type: application/json" --data-binary '[{
  "type": "hashed-password",
  "auth-id": "'$MY_DEVICE'",
  "secrets": [{
      "pwd-plain": "'$MY_PWD'"
  }]
}]' http://$REGISTRY_IP:28080/v1/credentials/$MY_TENANT/$MY_DEVICE
```

Send a telemetry message with QoS 0:

    curl -i -u $MY_DEVICE@$MY_TENANT:$MY_PWD -H 'Content-Type: application/json' --data-binary '{"temp": 6}' http://$HTTP_ADAPTER_IP:8080/telemetry?qos-level=0

Start Kafka consumer to consume the messages:

    kubectl run kafka-consumer -ti --image=strimzi/kafka:0.19.0-kafka-2.5.0 --rm=true --restart=Never -- bin/kafka-console-consumer.sh --bootstrap-server hono-kafka-cluster-kafka-bootstrap:9092 --property print.timestamp=true --property print.key=true --topic telemetry-$MY_TENANT --from-beginning

The output should look similar to this:

    If you don't see a command prompt, try pressing enter.
    CreateTime:1600975825255        TEST_DEVICE     {"temp": 5}

The first column shown the timestamp, the second the key and the third the value.
Unfortunately the `kafka-console-consumer.sh` does not support to print the headers which contain relevant metadata.

b-abel:
Rebased on master.

b-abel:
I added a Kafka consumer to the CLI. To consume messages enter:

```bash
# in Hono's base directory 
eval $(minikube docker-env)
export KAFKA_PORT=$(kubectl get service hono-kafka-external-bootstrap -o=jsonpath='{.spec.ports[0].nodePort}{"\n"}')
export KAFKA_IP=$(kubectl get nodes --output=jsonpath='{range .items[*]}{.status.addresses[?(@.type=="InternalIP")].address}{"\n"}{end}')

cd cli/target
java -jar hono-cli-*-exec.jar --spring.profiles.active=receiver-kafka --tenant.id=$MY_TENANT --hono.kafka.consumerConfig.bootstrap.servers=$KAFKA_IP:$KAFKA_PORT
```

The output should look contain messages printed like this:

```
16:25:07.538 [vert.x-eventloop-thread-0] INFO  o.e.hono.cli.app.KafkaBasedReceiver - topic: telemetry-TEST_TENANT, partition: 0, offset: 5, timestamp: 1601302574691 (2020-09-28T16:16:14.691)
        Headers:
                orig_adapter=hono-mqtt
                device_id=TEST_DEVICE
                qos=0
                x-opt-retain=false
                content-encoding=null
                creation-time=1601302574690
                message-id=null
                content-type=application/octet-stream
                orig_address=telemetry
                ttl=0
                reply-to=null
        Key:
                TEST_DEVICE
        Value:
                {"temp": 5}
```

sophokles73:
we need to be careful not to let this grow too big.
I would propose to merge #2168 ASAP so that it can be omitted from this PR. We can then adapt the client interfaces as we go and as we require.

b-abel:
> we need to be careful not to let this grow too big.

I fully agree. This PR is not intended to be merged at all. I only created it to enable collaboration and in case someone wants to try it out (I think that even some initial performance tests might make sense at this stage).

> I would propose to merge #2168 ASAP so that it can be omitted from this PR. 

That would be great. The really interesting part for me will be the actual use of these new clients in the protocol adapters (I don't think we have a ticket for that yet). Once this is done, my first commit here can be dropped as well. The rest is not really big.

cli/src/main/resources/application.yml
```diff
@@ -38,6 +38,22 @@ message:
...
 
+spring:
+  profiles: receiver-kafka
+
+hono:
+  kafka:
+    consumerConfig:
+      bootstrap.servers: localhost:9092
+      group.id: hono-cli
+      auto.offset.reset: earliest
```
fkaltner, line 49: We need to be careful about this. On the one hand I think the intention was not to miss any messages if the offset on a consumer somehow "got lost".  

On the other hand however this may lead to production problems, depending on how long we keep messages in topics: if we keep a million messages and the offset got lost when a consumer is restarted it will re-process everything from the very beginning. This can potentially take a very long time and seem to a client like a "downtime" as new messages are not yet processed.

fkaltner: Ah just noticed it is for the CLI. Still I think using latest might be better ;-)

fkaltner:
This also worked on my machine (tm) with 1 zookeeper replica saving precious CPU cycles ;-)

```
zookeeper:
    replicas: 1
```

fkaltner:
Also I reused the device with id '4711' (auth id 'sensor1') which is already in the default data of the packages project. This makes creating a tenant & devices obsolete.

cli/src/main/resources/application.yml
b-abel: > Ah just noticed it is for the CLI. Still I think using latest might be better ;-)

Yes, I do not see a productive use case for the CLI. I used `earliest` because it is easier for beginners to play around and immediately see messages produced before the consumer started. But you are right, it might be better to play it safe even in a beginner's tutorial.

hono-client-kafka/src/main/java/org/eclipse/hono/client/kafka/impl/KafkaBasedDownstreamSender.java
```diff
@@ -0,0 +1,187 @@
...
+     * @return A future indicating the outcome of the operation.
+     *         <p>
+     *         The future will be succeeded if the message has been sent downstream.
+     *         <p>
+     *         The future will be failed with a {@link org.eclipse.hono.client.ServerErrorException} if the data could
+     *         not be sent. The error code contained in the exception indicates the cause of the failure.
+     */
+    protected Future<Void> send(final KafkaProducer<String, Buffer> producer, final String topic,
+            final String deviceId, final QoS qos, final String contentType, final Buffer payload,
+            final Map<String, ?> properties, final SpanContext context) {
```
fkaltner, line 114: I noticed that the `SpanContext` is not used yet. Regarding tracing I think we should take a look at https://github.com/opentracing-contrib/java-kafka-client. By using e.g. the interceptor based approach it seems like a quick win to enable distributed tracing for Kafka.

b-abel:
Rebased to remove conflict and include the fix for missing log output in protocol adapters (https://github.com/eclipse/hono/commit/50ab23231f8213540d4b97a1b3c9d38888550acd).

b-abel:
Added drafts for the specifications of the Kafka-based Telemetry and Event APIs.

b-abel:
Rebased because #2168 has been merged to master.

b-abel:
I just realized that it might be better to separate the API specification from the implementation and move it to a new PR. There are no dependencies between the two, and it would be easier to follow the discussions if we kept them separate. WDYT?

b-abel:
Moved API specification to new PR #2254.

hono-client-kafka/src/main/java/org/eclipse/hono/client/kafka/impl/KafkaBasedDownstreamSender.java
b-abel: I added tracing. I did not use the library you mentioned, because does not add much to the tracing logic we already have in place and does not provide the flexibility to adapt to our needs

b-abel:
Rebased on master.

b-abel:
PR #2303 contains a new implementation of Kafka-based downstream senders and replaces this PR. It uses the new client interfaces so that the hacks in the protocol adapters are no longer needed.
I have also completely reworked the implementation:
1. the telemetry sender now uses only a single producer to ensure that telemetry data with all QoS levels is sent within the same transactions in the Kafka client
2. the downstream sender has been redesigned to support lifecycle and can be started again and again. The Kafka Producer is managed by the sender and if necessary (e.g. if it has to be closed after a terminal error) a new one is generated.

b-abel:
# Howto use the Kafka clients in Minikube

This brief walk-through guide shows how to build Hono with the Kafka client and run it against a Kafka cluster in Minikube.

## Prerequisites

* Minikube installed
* Helm 3 installed

## Create Kubernetes Cluster

Start/create Minikube instance with
```bash
minikube start --cpus 2 --memory 8192
```

## Install Strimzi

The [Strimzi](https://strimzi.io/) project provides a way to run a Kafka cluster on Kubernetes. We deploy it to Minikube.

Add the Helm repo

```bash
eval $(minikube docker-env)
helm repo add strimzi https://strimzi.io/charts/
helm repo update
```

Deploy the operator to default namespace using release name `strimzi`:

```bash
helm install strimzi strimzi/strimzi-kafka-operator --version 0.19.0
```

### Deploy Kafka

Now we deploy an ephemeral Kafka cluster with an external listener. For details see [Strimzi's deployment guide](https://strimzi.io/docs/operators/latest/deploying.html#kafka-cluster-str).

```bash
cat << EOF | kubectl create -f -
apiVersion: kafka.strimzi.io/v1beta1
kind: Kafka
metadata:
  name: hono
spec:
  kafka:
    version: 2.5.0
    replicas: 1
    listeners:
      plain: {}
      tls: {}
      external:
        type: nodeport
        tls: false
    config:
      offsets.topic.replication.factor: 1
      transaction.state.log.replication.factor: 1
      transaction.state.log.min.isr: 1
      log.message.format.version: "2.5"
    storage:
      type: ephemeral
  zookeeper:
    replicas: 1
    storage:
      type: ephemeral
  entityOperator:
    topicOperator: {}
    userOperator: {}
EOF
```

Wait while Kubernetes starts all pods:

```bash
kubectl wait kafka/hono --for=condition=Ready --timeout=300s
```

To verify the Kafka deployment use the example clients.
Produce messages with:

```bash
kubectl run kafka-producer -ti --image=strimzi/kafka:0.19.0-kafka-2.5.0 --rm=true --restart=Never -- bin/kafka-console-producer.sh --broker-list hono-kafka-bootstrap:9092 --topic my-topic
```

and consume them with:

```bash
kubectl run kafka-consumer -ti --image=strimzi/kafka:0.19.0-kafka-2.5.0 --rm=true --restart=Never -- bin/kafka-console-consumer.sh --bootstrap-server hono-kafka-bootstrap:9092 --topic my-topic --from-beginning
```

## Deploy Hono

Build Hono images from this branch locally:

```bash
# in hono's base dir
eval $(minikube docker-env)
mvn clean install -Pbuild-docker-image,metrics-prometheus,jaeger
```

While the images are being built, you can prepare the Helm chart.

The Helm chart in the [Eclipse Packages repository](https://github.com/eclipse/packages/tree/master/charts/hono) does not support the configuration for Kafka yet, so it must be adapted.

Get Hono's Helm chart:

```bash
git clone  https://github.com/eclipse/packages.git
```

Edit the file `packages/charts/hono/templates/hono-adapter-http/hono-adapter-http-vertx-secret.yaml`: add the following at the end of the `application.yml` part:

```yaml
      kafka:
        producerConfig:
          bootstrap.servers: hono-kafka-bootstrap.default:9092
```

**NB:** make sure that `kafka` is indented one level below `hono`.

Create namespace `hono`:

```bash
kubectl create namespace hono
```

Deploy the HTTP adapter and only required components of Hono:

```bash
# in the directory of the packages repo
cd charts/
eval $(minikube docker-env)
helm install --dependency-update -n hono --set honoImagesTag=1.5.0-SNAPSHOT --set adapters.amqp.enabled=false --set adapters.mqtt.enabled=false --set prometheus.createInstance=false --set grafana.enabled=false eclipse-hono ./hono
```

Wait while Kubernetes starts all pods:

```bash
kubectl wait pods --all --for=condition=Ready --timeout=300s -n hono
```

## Start the Downstream Consumer

First, start the example application that will print the messages.
Open a new terminal window and enter:

```bash
export KAFKA_IP=$(kubectl get kafka hono -o jsonpath='{.status.listeners[?(@.type=="external")].bootstrapServers}')
java -jar hono-cli-*-exec.jar --spring.profiles.active=receiver-kafka --tenant.id=DEFAULT_TENANT --hono.kafka.consumerConfig.bootstrap.servers=$KAFKA_IP  --print.verbose=true
```

The output should contain a line similar to this:

```
10:04:02.856 [vert.x-eventloop-thread-0] INFO  o.eclipse.hono.cli.app.KafkaReceiver - Receiver [tenant: DEFAULT_TENANT, mode: all] subscribed successfully, hit ctrl-c to exit
```


## Send a Message

Open a new terminal window to start a `tunnel`:

```bash
minikube tunnel
```

Send a telemetry message with QoS 0 from the predefined example device:

```bash
export HTTP_ADAPTER_IP=$(kubectl get service eclipse-hono-adapter-http-vertx --output="jsonpath={.status.loadBalancer.ingress[0]['hostname','ip']}" -n hono)
curl -i -u sensor1@DEFAULT_TENANT:hono-secret -H 'Content-Type: application/json' -H 'qos-level: 1' --data-binary '{"temp": 6}' http://$HTTP_ADAPTER_IP:8080/telemetry
```

The output should look something like this:
```
HTTP/1.1 202 Accepted
content-length: 0
```

Otherwise (e.g. if the `curl` command does not terminate), check that `minikube tunnel` is running.

In the terminal window with the example application you should see the messages similar to this:

```
10:21:39.747 [vert.x-eventloop-thread-0] INFO  o.eclipse.hono.cli.app.KafkaReceiver - topic: hono.telemetry.DEFAULT_TENANT, partition: 0, offset: 5, timestamp: 1605777699741 (2020-11-19T10:21:39.741)
  Headers:
    content-type=application/json
    orig_adapter=hono-http
    orig_address=/telemetry
    qos=1
    device_id=4711
    uber-trace-id=1a32de3a47df1341:b9cab790176837fd:8f3c92ed670927c4:0
  Key:
    4711
  Value:
    {"temp": 6}
```


## Clean up

To shut down the Kubernetes cluster enter:

    minikube stop

and when you are done you can delete the whole Kubernetes cluster with:

    minikube delete

b-abel:
I would like to leave this draft PR open for a while. Here I can add the CLI on top of the implementation of the client from #2303. So this draft offers the possibility to test the client manually.

sophokles73:
@b-abel is this still needed?

b-abel:
> @b-abel is this still needed?

No, I will open a new PR for the CLI.

PR #2218, opened 2020-09-28
Release and settle delivery when waiting for update timed out

calohmn:
This should free up some resources.

EDIT:
When a telemetry message with Qos1 gets sent and the northbound consumer doesn't acknowledge the message by sending a delivery update, a timeout will occur in the protocol adapter after some time. 

The problem here is, that the message is considered as "in flight" still, most notably from the point of view of the AMQP messaging network (e.g. the Qpid dispatch router). This message counts towards the link capacity configured in the router.

This all means that if one sends as many messages as configured as the router link capacity to the north bound application, and if the northbound application doesn't send a delivery update for any of these, then any further messages will result in a "no credit" error when the adapter tries to send them to the router.
This will be the case irrespective of the northbound application having sent new credits to the router.

I think we should consider releasing/settling the message from the *sender* side (i.e. the protocol adapter) in this case, in order to free up resources in the router.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
@@ -21,6 +21,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.qpid.proton.amqp.messaging.Accepted;
 import org.apache.qpid.proton.amqp.messaging.Modified;
@@ -53,6 +54,7 @@ import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.Promise;
 import io.vertx.proton.ProtonDelivery;
+import io.vertx.proton.ProtonHelper;
 import io.vertx.proton.ProtonSender;
 
 /**
@@ -267,6 +269,7 @@ public abstract class AbstractSender extends AbstractHonoClient implements Messa
         Objects.requireNonNull(message);
         Objects.requireNonNull(currentSpan);
 
+        final AtomicReference<ProtonDelivery> deliveryRef = new AtomicReference<>();
         final Promise<ProtonDelivery> result = Promise.promise();
         final String messageId = String.format("%s-%d", getClass().getSimpleName(), MESSAGE_COUNTER.getAndIncrement());
         message.setMessageId(messageId);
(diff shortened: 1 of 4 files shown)


sophokles73:
@calohmn did you check with vert.x proton whether this is actually beneficial/supported?

calohmn:
@sophokles73 I'm still doing some further tests on this.

calohmn:
Just settling the delivery locally doesn't bring much benefit, as far as I see it.

I've updated the commit and the PR title and description to suggest sending out a disposition update from the sender in case of this timeout, in order to *cancel* the disposition and tell the router that the message shouldn't be in flight anymore. 

It is still somewhat unclear whether doing this would be the recommended approach with regards to the AMQP spec and to what the router supports. Tests using an integration test (no committed here because it would be a quite long-running test) have shown the behaviour to be as expected. To be sure, I'll ask in the Qpid mailing list about this.

calohmn:
According to the [response](http://qpid.2158936.n2.nabble.com/How-to-quot-cancel-quot-a-delivery-from-the-sender-side-td7694547.html) on the Qpid mailinglist, the sender should indeed settle the delivery and optionally set an outcome.
As Proton-J doesn't seem to support sending a delivery update with settled=true and no outcome, I think we should stick to settling and releasing the delivery (as done in this PR).

client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
```diff
@@ -282,13 +285,15 @@ protected final Span startSpan(final Message message) {
                                         + connection.getConfig().getSendMessageTimeout() + "ms");
                         logMessageSendingError("waiting for delivery update timed out for message [ID: {}, address: {}] after {}ms",
                                 messageId, getMessageAddress(message), connection.getConfig().getSendMessageTimeout());
+                        Optional.ofNullable(deliveryRef.get())
```
sophokles73, line 288: how about adding a comment here explaining why we are doing this?

client/src/main/java/org/eclipse/hono/client/impl/DelegatedCommandSenderImpl.java
```diff
@@ -142,13 +146,15 @@ public String getEndpoint() {
                                         + connection.getConfig().getSendMessageTimeout() + "ms");
                         logMessageSendingError("waiting for delivery update timed out for message [ID: {}, address: {}] after {}ms",
                                 messageId, getMessageAddress(message), connection.getConfig().getSendMessageTimeout());
+                        Optional.ofNullable(deliveryRef.get())
```
sophokles73, line 149: comment?

client/src/main/java/org/eclipse/hono/client/impl/TelemetrySenderImpl.java
```diff
@@ -180,6 +184,8 @@ protected String getTo(final String deviceId) {
                         logMessageSendingError(
                                 "waiting for delivery update timed out for message [ID: {}, address: {}] after {}ms",
                                 messageId, getMessageAddress(message), connection.getConfig().getSendMessageTimeout());
+                        Optional.ofNullable(deliveryRef.get())
```
sophokles73, line 187: comment?

tests/src/test/java/org/eclipse/hono/tests/http/TelemetryHttpIT.java
```diff
@@ -194,8 +199,22 @@ public void testUploadQos1MessageFailsIfDeliveryStateNotUpdated(final VertxTestC
...
                         (delivery, msg) -> {
                             logger.debug("received {}", msg);
+                            ctx.verify(() -> {
+                                assertThat(delivery.remotelySettled()).isFalse();
+                                assertThat(delivery.getRemoteState()).isNull();
+                            });
                             messageReceived.flag();
                             // don't update the delivery state here
+                            // wait for the delivery to get remotely settled via the timeout handling in the adapter
+                            vertx.setTimer(ClientConfigProperties.DEFAULT_SEND_MESSAGE_TIMEOUT + 50, tid -> {
```
sophokles73, line 209: IMHO we should tie this check to *after* the HTTP client has gotten back its 501 response (and then wait some) in order to make it more robust against timing issues induced by scarce computing resources on the CI system ...

sophokles73 (approved):
LGTM

Issue #2215, opened 2020-09-25
Device Registry Management API accepts credentials without secrets


sophokles73:
The Management API currently defines the *secrets* property of the credentials types as optional. However, there is not much use for credentials that do not contain any secrets. FMPOV the property should therefore be mandatory and the secrets array should be required to contain at least one element. Formally, this would constitute a *breaking API change*. However, I wonder if there is a use case that would require adding credentials without secrets and which we want to (continue to) support.
 
The Credentials API already requires the response payload of the *get Credentials* operation to contain a non-empty secrets array, so it should make implementation much easier, if no credentials need to be *invented* by the registry ...

@ctron @dejanb @calohmn WDYT?

dejanb:
@sophokles73 this change sounds reasonable to me

sophokles73:
closed via #2219

PR #2219, opened 2020-09-28
[#2215] Make secrets array mandatory

sophokles73:
The supported credentials types now require the secrets array to be
non-empty.

### DIFF
--- a/services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordCredential.java
+++ b/services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordCredential.java
@@ -12,8 +12,10 @@
  *******************************************************************************/
 package org.eclipse.hono.service.management.credentials;
 
+import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Objects;
 
 import org.eclipse.hono.util.RegistryManagementConstants;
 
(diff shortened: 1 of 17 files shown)


services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/PasswordCredential.java
```diff
@@ -32,19 +34,23 @@
 
     static final String TYPE = RegistryManagementConstants.SECRETS_TYPE_HASHED_PASSWORD;
 
-    @JsonProperty
-    @JsonInclude(value = JsonInclude.Include.NON_EMPTY)
     private List<PasswordSecret> secrets = new LinkedList<>();
```
calohmn, line 37: final?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/PskCredential.java
```diff
@@ -32,19 +34,22 @@
 
     static final String TYPE = RegistryManagementConstants.SECRETS_TYPE_PRESHARED_KEY;
 
-    @JsonProperty(value = RegistryManagementConstants.FIELD_SECRETS)
-    @JsonInclude(value = JsonInclude.Include.NON_EMPTY)
     private List<PskSecret> secrets = new LinkedList<>();
```
calohmn, line 37: final?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/X509CertificateCredential.java
```diff
@@ -37,8 +39,6 @@
 
     static final String TYPE = RegistryManagementConstants.SECRETS_TYPE_X509_CERT;
 
-    @JsonProperty
-    @JsonInclude(value = JsonInclude.Include.NON_EMPTY)
     private List<X509CertificateSecret> secrets = new LinkedList<>();
```
calohmn, line 42: final?

calohmn (commented):
FMPOV the argument presented in #2215 is not specific to a credentials type. Therefore it does make sense to adapt the `GenericCredential` class and the `CommonCredentials` schema definition as well, doesn't it?

sophokles73:
> FMPOV the argument presented in #2215 is not specific to a credentials type. Therefore it does make sense to adapt the GenericCredential class and the CommonCredentials schema definition as well, doesn't it?

That is true. However, the `GenericCredential` class is never used (and cannot be used) based on the way the *TypedCredential* type is defined in the API. I therefore wonder if it wouldn't make more sense to remove the `GenericCredential` class altogether. Not sure what should be adapted in the `CommonCredentials` type, though ...

dejanb (approved):

PR #2222, opened 2020-09-29
[#2201] Add colon in device id pattern

### LINKED ISSUE #2201: Allow colon in default device id pattern
From https://github.com/eclipse/hono/issues/1956#issuecomment-696689998:
As it is a common use case to use Hono with Eclipse Ditto, a Hono device identifier should by default also be allowed to contain a colon, since Ditto requires this.

Also the device and tenant identifier pattern should be removed from the device management API definition, as these patterns may be adapted via the device registry config.

calohmn:
This fixes #2201.
Colon in device id should be supported for compatibility with Eclipse Ditto.
Also remove the tenant and device id patterns from the Device Registry Management API specification.
And add missing release notes entries regarding the pattern properties for Hono 1.3 and 1.4.

FMPOV, this fix should also be applied in the 1.3.x and 1.4.x branches, adding a release notes entry like this:
````
* The default pattern for valid device identifiers used for the file based and the MongoDB based registry
  implementations now also contains a colon for compatibility with Eclipse Ditto.
````

### CODE REVIEW

File: site/documentation/content/admin-guide/file-based-device-registry-config.md [hunk 1]
```diff
@@ -58,7 +58,7 @@ The following table provides an overview of the configuration variables and corr
...
 | `HONO_REGISTRY_REST_TENANT_ID_PATTERN`<br>`--hono.registry.rest.tenantIdPattern` | no | `^[a-zA-Z0-9-_\.]+$` | The regular expression to use to validate tenant ID. Please refer to the [java pattern documentation](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html).<br>**Deprecated** Use `HONO_REGISTRY_HTTP_TENANT_ID_REGEX` instead. |
-| `HONO_REGISTRY_REST_DEVICE_ID_PATTERN`<br>`--hono.registry.rest.deviceIdPattern` | no | `^[a-zA-Z0-9-_\.]+$` | The regular expression to use to validate device ID. Please refer to the [java pattern documentation](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html).<br>**Deprecated** Use `HONO_REGISTRY_HTTP_DEVICE_ID_REGEX` instead. |
+| `HONO_REGISTRY_REST_DEVICE_ID_PATTERN`<br>`--hono.registry.rest.deviceIdPattern` | no | `^[a-zA-Z0-9-_\.:]+$` | The regular expression to use to validate device ID. Please refer to the [java pattern documentation](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html).<br>**Deprecated** Use `HONO_REGISTRY_HTTP_DEVICE_ID_REGEX` instead. |
```
sophokles73, line 61: can you also fix the name of the property to use instead ... and the one referred to by the *HONO_REGISTRY_REST_TENANT_ID_PATTERN* as well?

site/homepage/content/release-notes.md
```diff
@@ -148,6 +150,14 @@ title = "Release Notes"
...
   what the method is used for. The same change has been applied to the other `customize[*]Config`
   methods in the `AbstractAdapterConfig` class.
+* The file based as well as the MongoDB based device registry implementations now disallow certain
+  special characters in device and tenant identifiers. The configuration properties
+  `HONO_REGISTRY_HTTP_TENANT_ID_PATTERN` and `HONO_REGISTRY_HTTP_DEVICE_ID_PATTERN` (and corresponding ones with
+  prefix `HONO_REGISTRY_REST` for the file based registry) can be used to override the patterns for matching
+  valid identifiers. Please refer to the 
+  [file based registry]({{% doclink "/admin-guide/file-based-device-registry-config/#service-configuration" %}}) or
+  [MongoDB based registry]({{% doclink "/admin-guide/mongodb-device-registry-config/#service-configuration" %}})
+  configuration guide for details.
```
sophokles73, line 160: is this supposed to be documenting the status quo of the 1.3.0 release only? My understanding is that you want to remove these restrictions again in 1.5.0, 1.4.1 and 1.3.1. I guess we then should also add a corresponding entry to the *API Changes* section for the upcoming 1.5.0, right?

calohmn: This shall document the fact that there are restrictions on the allowed characters now, not talking about the colon in particular. So this is still valid after the patch releases.

> I guess we then should also add a corresponding entry to the API Changes section for the upcoming 1.5.0, right?

I've added an entry there.

sophokles73 (approved):
LGTM

PR #2224, opened 2020-09-30
Add GitHub Actions workflow for CI build and running integration tests

sophokles73:
The job uses a build matrix for the different device registry implementations. The jobs are run in parallel on separate VMs.


calohmn (approved):
LGTM

PR #2225, opened 2020-09-30
Update to Artemis 2.13.0, Qpid router 1.14.0 image versions

calohmn:
Update to latest image versions for use in the integration tests.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -28,13 +28,13 @@
   <description>A &quot;bill-of-materials&quot; for Hono.</description>
 
   <properties>
-    <artemis.image.name>quay.io/enmasse/artemis-base:2.10.1</artemis.image.name>
+    <artemis.image.name>quay.io/enmasse/artemis-base:2.13.0</artemis.image.name>
     <assertj-core.version>3.17.2</assertj-core.version>
     <c3p0.version>0.9.5.4</c3p0.version>
     <caffeine.version>2.8.4</caffeine.version>
     <californium.version>2.3.1</californium.version>
     <commons-cli.version>1.2</commons-cli.version>
-    <dispatch-router.image.name>quay.io/interconnectedcloud/qdrouterd:1.12.0</dispatch-router.image.name>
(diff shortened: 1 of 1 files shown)


sophokles73:
we will also need to assure license compatibility for the new Dispatch Router and Artemis versions ...

sophokles73:
@calohmn I have seen that you have created CQs for Artemis and Dispatch Router. However, I couldn't find CQs for Artemis' transitive dependencies. We will also need those ....

calohmn:
@sophokles73 I checked the transitive dependencies of Artemis 2.13.0 and found no relevant changes compared to the Artemis version used in Hono up to now.

calohmn:
CQs have been resolved now.

sophokles73:
Did you run the dash-license tool successfully?

calohmn:
Running the dash-license tool on the Hono dependencies extracted via `mvn dependency:list` obviously doesn't show any difference since there are no maven dependencies for Artemis and the Qpid router in Hono.
Running the dash-license tool on the artemis-distribution module maven dependencies also shows no difference. It seems the Artemis 2.13.0 maven coordinates don't get matched with the Artemis CQ created as part of this PR (most of them get matched with ClearlyDefined though, e.g. artemis-server).
So, I guess the dash-license tool isn't of much help here.

calohmn:
@sophokles73 
Checked out tag 2.13.0 of the artemis repo.
$ cd activemq-artemis/artemis-distribution
$ mvn dependency:list | grep -Poh "\S+:(runtime|compile)" | sort | uniq > _maven_rt_compile_2.13.0.deps
$ java -jar <path>/org.eclipse.dash.licenses-0.0.1-SNAPSHOT.jar _maven_rt_compile_2.13.0.deps 

Then the corresponding steps for tag 2.10.1.

The output for 2.13.0 has added/changed dependencies only for "org.apache.activemq/*artemis*" entries (except "org.apache.activemq/activemq-branding").

sophokles73:
I did the same thing and it looks like we have CQs for all of the non-artemis components on record which are part of the artemis Docker image.

sophokles73 (approved):
LGTM

Issue #2183, opened 2020-09-16
File based registry fails to replace credentials


sophokles73:
The registry should support replacing a device's credentials with a new set of credentials by means of omitting the existing credentials from the body of a PUT request to `/v1/credentials/tenant/device`.
However, the file based registry does not remove the omitted credentials but merely adds the (new) credentials from the request body to the set of existing credentials.

Issue #2174, opened 2020-09-14
File based registry does not accept X.509 credentials


sophokles73:
The file based device registry fails to update a device with X.509 type credentials that do not contain a *secrets* array.

PR #2227, opened 2020-10-01
[#2174][#2183] Accept X.509 credentials

sophokles73:
The registry implementations now accept X.509 credentials in update
Credentials requests. The registry now also removes existing credentials
on record that do not match any of the credentials in an update
Credentials request's body.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
+++ b/core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
@@ -12,10 +12,6 @@
  *******************************************************************************/
 package org.eclipse.hono.util;
 
-import java.util.Objects;
-
-import io.vertx.core.json.JsonObject;
-
 /**
  * Constants &amp; utility methods used throughout the Device Management API.
  */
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -350,4 +354,16 @@
...
     }
+
+    /**
+     * Checks if a given secret is of type {@value #SECRETS_TYPE_HASHED_PASSWORD}.
+     *
+     * @param secret The secret to check.
+     * @return {@code true} if the secret has a string property with name {@value #FIELD_TYPE}
+     *         and value {@value #SECRETS_TYPE_HASHED_PASSWORD}.
+     */
+    public static boolean isHashedPasswordSecret(final JsonObject secret) {
```
kaniyan, line 365: Missing info regarding `NullPointerException` in _JavaDoc_ if the secret is _null_.

kaniyan (changes requested):

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
sophokles73: This method is not used anyway, I'll remove it ...

dejanb (approved):
LGTM

PR #2229, opened 2020-10-02
Extend search devices operation to support wildcards in the file-based device registry

kaniyan:
With this PR, the file-based device registry supports wildcard matching in search devices operation. The supported wildcard characters are  `*`  and `?`.


sophokles73 (approved):
LGTM

PR #2233, opened 2020-10-07
Remove UNKNOWN QoS level

sophokles73:
The UNKNOWN QoS level has no particular meaning but prevents the QoS
class from being used in the new client implementation (#2168).


calohmn (approved):
LGTM

PR #2235, opened 2020-10-08
[#2234] Reject command message with unsupported body section

### LINKED ISSUE #2234: Command with unsupported AMQP message body isn't rejected
Brought forward in the discussion in #2214:
A command message with an unsupported AMQP message body is forwarded to the device with an empty payload.

Instead of payload data getting lost, such a command message should get rejected as invalid.

calohmn:
This fixes #2234:
Commands now get rejected if they have a non-null body section that is not supported.
This is a conservative approach in that it still allows the body section to be an `amqp-value` (not documented in the API), not wanting to break existing consumer applications.

Also introduce a getPayloadSize() method, preventing copy of body section bytes.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/MessageHelper.java
+++ b/core/src/main/java/org/eclipse/hono/util/MessageHelper.java
@@ -391,7 +391,7 @@ public final class MessageHelper {
      *
      * @param msg The AMQP 1.0 message to parse the body of.
      * @return The message body parsed into a JSON object or {@code null} if the message does not have a <em>Data</em>
-     *         nor an <em>AmqpValue</em> section.
+     *         nor an <em>AmqpValue</em> section or if the body section is empty.
      * @throws NullPointerException if the message is {@code null}.
      * @throws DecodeException if the payload cannot be parsed into a JSON object.
      */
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/eclipse/hono/util/MessageHelper.java
```diff
@@ -459,39 +440,88 @@ public static Buffer getPayload(final Message msg) {
...
+    /**
+     * Checks whether the payload data contained in a message's body is returned as {@code null} by
+     * {@link #getPayload(Message)} or {@link #getPayloadAsString(Message)} and, if that is the case,
+     * returns the reason for that.
+     *
+     * @param msg The AMQP 1.0 message to parse.
+     * @return A reason string or {@code null} if the payload is not {@code null}.
+     * @throws NullPointerException if the message is {@code null}.
+     */
+    public static String getReasonIfPayloadNull(final Message msg) {
```
sophokles73, line 482: This looks a little odd to me. How about renaming this to something like *assertSupportedPayloadType* and throw an exception holding a description in case of an invalid payload?

sophokles73:
@calohmn any news on this one?

core/src/main/java/org/eclipse/hono/util/MessageHelper.java
calohmn: Throwing an exception in that message doesn't fit the use case of the `Command` class and just adds clutter there.
I've renamed the method to `getUnsupportedPayloadReason` now (returning an Optional).
When the need for a `assertSupportedPayloadTyp` method arises, we can add that, simply invoking `getUnsupportedPayloadReason` internally.

sophokles73: Is this method being used anywhere else? If this is a very Command specific requirement, then we could also simply embed this method into the Command class. WDYT?

calohmn: Ok, I've moved the method to the `Command` class.
I've also updated the new `org.eclipse.hono.adapter.client.command.Command` class.

sophokles73 (approved):
LGTM
how about adding an entry to the release notes?

calohmn:
Release notes entry added in https://github.com/eclipse/hono/commit/bbe53bd27470e71f118412ebc48060a15366fe1b.

Issue #2230, opened 2020-10-02
Define limits for adapters receiving commands


sophokles73:
The protocol adapters currently receive commands over a link that does not limit the max-message-size using the connection's default session which does not limit the session window. These limits should be set based on configuration properties with reasonable defaults in order to prevent the adapters from running into OOM exceptions when command bursts occur.

PR #2236, opened 2020-10-08
[#2230] Support limiting resource usage in hono-client

sophokles73:
The client now supports setting
- the max frame size that should be used for outgoing connections and
- the max-message-size to be used for (receiver) links created by the
client and
- the incoming window size of sessions created by the client.


core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -575,4 +603,140 @@ public final Pattern getAddressRewritePattern() {
...
+     * The default value of this property is {@value #MIN_MESSAGE_SIZE_NONE}.
+     * <p>
+     * The value of this property will be used by the client during sender link establishment.
+     * Link establishment will fail, if the peer indicates a max-message-size in its <em>attach</em>
+     * frame that is smaller than the configured minimum message size.
+     *
+     * @return The message size in bytes or {@value #MIN_MESSAGE_SIZE_NONE}, indicating no minimum size at all.
+     */
+    public final long getMinMessageSize() {
+        return minMessageSize;
```
calohmn, line 619: I think this property is quite confusing. Just guessing from the name, I would think that with such a setting, messages with a smaller size than configured here would be dropped/rejected. However, its actual semantics would be described by a name like `getMinimumRequiredPeerMaxMessageSize()` (which is maybe a bit long), right?
Also the fact that `getMaxMessageSize` only applies to a receiver whereas `getMinMessageSize` is only for a sender makes things somewhat hard to grasp at first sight.

Can there be a better name? Or, do we really need this property?

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -769,9 +832,20 @@ public void closeAndFree(
...
+                        final long remoteMaxMessageSize = Optional.ofNullable(sender.getRemoteMaxMessageSize())
+                                .map(UnsignedLong::longValue)
+                                .orElse(0L);
+                        if (remoteMaxMessageSize < clientConfigProperties.getMinMessageSize()) {
+                            // peer won't accept our (biggest) messages
+                            sender.close();
+                            log.debug("peer does not support minimum message size [required: {}, supported: {}",
+                                    clientConfigProperties.getMinMessageSize(), remoteMaxMessageSize);
+                            senderPromise.tryFail(new ClientErrorException(HttpURLConnection.HTTP_PRECON_FAILED,
+                                    "peer does not meet sender's minimum max-message-size requirement"));
```
calohmn, line 846: I think throwing a `ClientErrorException` here means that we have to adapt the code for sending telemetry messages for example by catching this `ClientErrorException` and rethrowing it back as a ServerErrorException, so that the device gets a 5xx error here.

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
```diff
@@ -575,4 +603,140 @@ public final Pattern getAddressRewritePattern() {
...
+     * Sets the maximum size of an AMQP 1.0 message that this client should accept from a peer.
+     * <p>
+     * The default value of this property is {@value #MAX_MESSAGE_SIZE_UNLIMITED}.
+     *
+     * @param size The message size in bytes or {@value #MAX_MESSAGE_SIZE_UNLIMITED}, indicating messages of any size.
+     * @throws IllegalArgumentException if size is &lt; 0 and not {@value #MAX_MESSAGE_SIZE_UNLIMITED}.
+     */
+    public final void setMaxMessageSize(final long size) {
+        if (size != MAX_MESSAGE_SIZE_UNLIMITED && size < 0) {
+            throw new IllegalArgumentException("max-message-size must be >= 0");
```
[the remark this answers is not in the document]
sophokles73: That is true. However, I wanted to use the same semantics as with MAX_FRAME_SIZE, i.e. use `-1` in both cases to indicate *no limit*. In fact, the `ProtonClientOptions` require a negative value to indicate *no limit* ...

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
sophokles73: > Or, do we really need this property?

What is the alternative? Having the sender link being closed on each message being sent that is bigger than the peer's max-message-size?

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
sophokles73: Well, both the `DownstreamSender.send()` and `HonoConnection.createSender()` JavaDoc allow the returned Future to be failed with a ClientErrorException. Or am I missing something?

core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java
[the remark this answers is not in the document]
calohmn: Ok. I that case I would suggest adding a check when applying the MaxMessageSize value, either not using the value at all:
````
                if (clientConfigProperties.getMaxMessageSize() != ClientConfigProperties.MAX_MESSAGE_SIZE_UNLIMITED) {
                    receiver.setMaxMessageSize(new UnsignedLong(clientConfigProperties.getMaxMessageSize()));
                }
````
or applying `0` in the "unlimited size" case, in order to not transport `18446744073709551615` on the wire. 
And I would also change the exception text to `max-message-size must be >= -1`.

File: core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java (hunk 1 above)
Comment (calohmn) [reply core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java:619]: > What is the alternative? Having the sender link being closed on each message being sent that is bigger than the peer's max-message-size?

Right, having that check on link establishment already is better. For the use case of the telemetry/event sender clients in the protocol adapters, I think this config value should internally be modified to be at least the configured `ServiceConfigProperties#getMaxPayloadSize` value.

I still think the name should be changed, though. How about `getMinMaxMessageSize` or `getMinMessageSizeLimit`?
(MinMaxMessageSize would be similar in naming to the MIN-MAX-FRAME-SIZE constant of the AMQP spec).

File: client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java (hunk 2 above)
Comment (calohmn) [reply client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java:846]: Yes, I see there is a case already where a ClientErrorException is thrown in `HonoConnection.createSender()`.
So that is an issue independent of this PR really, preventing a ClientErrorException on `createSender()` for a telemetry/event sender to be propagated to the device. I'll create an issue for that.

File: core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java (hunk 1 above)
Comment (sophokles73) [reply core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java:619]: let's use *minMaxMessageSize* then ...

File: core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java (hunk 3 above)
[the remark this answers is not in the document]
Comment (sophokles73) [reply core/src/main/java/org/eclipse/hono/config/ClientConfigProperties.java:663]: yes, that makes sense ...

**Comment by sophokles73** [comment]
@calohmn can you take another look?

File: client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java [hunk 4]
```diff
@@ -436,8 +441,14 @@ private void connect(
...
                                             newConnection.getRemoteContainer());
-                                    setConnection(newConnection);
-                                    wrappedConnectionHandler.handle(Future.succeededFuture(this));
+                                    createDefaultSession(newConnection)
+                                        .onSuccess(newSession -> {
+                                            setConnection(newConnection, newSession);
+                                            wrappedConnectionHandler.handle(Future.succeededFuture(this));
+                                        })
+                                        .onFailure(t -> {
+                                            wrappedConnectionHandler.handle(Future.failedFuture(t));
```
calohmn, line 450: Instead of failing all reconnect attempts like this, close the connection and call `reconnect` in the `onFailure` case here instead?

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -769,9 +832,20 @@ public void closeAndFree(
...
-                        // wait on credits a little time, if not already given
-                        if (sender.getCredit() <= 0) {
+                        log.debug("sender open [target: {}, sendQueueFull: {}, remote max-message-size: {}]",
+                                targetAddress, sender.sendQueueFull(), sender.getRemoteMaxMessageSize());
+                        final long remoteMaxMessageSize = Optional.ofNullable(sender.getRemoteMaxMessageSize())
+                                .map(UnsignedLong::longValue)
+                                .orElse(0L);
+                        if (remoteMaxMessageSize < clientConfigProperties.getMinMaxMessageSize()) {
+                            // peer won't accept our (biggest) messages
+                            sender.close();
```
calohmn, line 842: I think in the condition here there is an added `remoteMaxMessageSize > 0` missing.

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -769,9 +832,20 @@ public void closeAndFree(
...
+                            log.debug("peer does not support minimum message size [required: {}, supported: {}",
+                                    clientConfigProperties.getMinMaxMessageSize(), remoteMaxMessageSize);
+                            senderPromise.tryFail(new ClientErrorException(HttpURLConnection.HTTP_PRECON_FAILED,
+                                    "peer does not meet sender's minimum max-message-size requirement"));
```
calohmn, line 846: Maybe include the `remoteMaxMessageSize` in the exception text so that it appears in the surrounding tracing span (if one exists) via the logged error message? (So that the application doesn't needed to be restarted if not run with DEBUG log level in order to find out about the actual advertised value.)

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -707,6 +718,58 @@ public void closeAndFree(
...
+                            connectionFactory.getPort(),
+                            connectionFactory.getServerRole());
+                } else {
+                    log.debug("server [{}:{}, role: {}] closed session with error [condition: {}, description: {}]",
+                            connectionFactory.getHost(),
+                            connectionFactory.getPort(),
+                            connectionFactory.getServerRole(),
+                            error.getCondition(),
+                            error.getDescription());
+                }
```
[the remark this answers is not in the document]
sophokles73: yes, closing the connection seems to be the right thing to do. I will set a corresponding handler ...

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
sophokles73: If the peer does not accept the session I do not really see why it should accept it in subsequent attempts. FMPOV this is comparable to wrong credentials, the peer either is willing to open a session or not, or am I mistaken?
I agree that we should close the newly opened connection but then I guess we will need to fail the connection attempt. WDYT?

calohmn: For a session to not get established by the remote peer, here Qpid in particular, I could think of 2 reasons:
- maxSessions having been reached already (shouldn't happen here since we only use one session per connection)
- or Qpid not having enough memory to create a new session.
For the 2nd case, I think it would be good to retry the connection attempt, potentially ending up at another Qpid instance or Qpid having been restarted then.
Failing the connection attempt instead in this case would be fatal and cause the whole adapter to need a restart in case of the telemetry/event sender connection for example, if I'm not mistaken.

sophokles73: I see your point and it makes sense to me. The default session in vertx-proton only has a close handler which is used to trigger the underlying connection's close handler. This mechanism is already being used by Hono implicitly, so I guess we could use the same approach here as well, i.e. do not register an explicit open handler for the session at all but simply rely on the (standard) behavior of the connection's close handler to trigger reconnect. WDYT?

calohmn: Yes, I agree.

sophokles73:
@calohmn I have changed HonoConnectionImpl to only register a session close handler now ... would you mind taking another look?

calohmn (approved):
LGTM

PR #2240, opened 2020-10-09
Use the existing static method to get status codes for exceptions

kaniyan:
Make use of `ServiceInvocationException.extractStatusCode()` method instead of duplicating the same code.

### DIFF
--- a/examples/hono-client-examples/src/main/java/org/eclipse/hono/devices/AmqpExampleDevice.java
+++ b/examples/hono-client-examples/src/main/java/org/eclipse/hono/devices/AmqpExampleDevice.java
@@ -19,7 +19,7 @@ import java.util.Map;
 import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.client.HonoConnection;
 import org.eclipse.hono.client.SendMessageSampler;
-import org.eclipse.hono.client.ServerErrorException;
+import org.eclipse.hono.client.ServiceInvocationException;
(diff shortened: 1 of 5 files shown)


examples/hono-client-examples/src/main/java/org/eclipse/hono/devices/AmqpExampleDevice.java
```diff
@@ -135,14 +135,12 @@ private void sendTelemetryMessageWithQos2() {
                     if (ar.succeeded()) {
                         System.out.println("Telemetry message with QoS 'AT_LEAST_ONCE' sent: " + payload);
                     } else {
-                        final Throwable cause = ar.cause();
                         String hint = "";
-                        if (cause instanceof ServerErrorException
-                                && ((ServerErrorException) cause).getErrorCode() == 503) {
+                        if (ServiceInvocationException.extractStatusCode(ar.cause()) == 503) {
```
sophokles73, line 139: can you also fix the naming of the methods?
*sendTelemetryMessageWithQos1* needs to be *sendTelemetryMessageWithQos0* and
*sendTelemetryMessageWithQos2* needs to be *sendTelemetryMessageWithQos1* ...

kaniyan: Nice catch. I have renamed them now.

sophokles73 (approved):
LGTM

PR #2242, opened 2020-10-09
Fix the static method to extract status code from exceptions

kaniyan:
* Fix by replacing `Optional.of(...)` with `Optional.ofNullable(...)`
* Add tests to verify status code extraction

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/ServiceInvocationException.java
+++ b/client/src/main/java/org/eclipse/hono/client/ServiceInvocationException.java
@@ -165,10 +165,11 @@ public class ServiceInvocationException extends RuntimeException {
      * Extract the HTTP status code from an exception.
      *
      * @param t The exception to extract the code from.
-     * @return The HTTP status code, or 500 if the exception is not of type {@link ServiceInvocationException}.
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/eclipse/hono/client/ServiceInvocationException.java
```diff
@@ -165,10 +165,11 @@ private static String providedOrDefaultMessage(final int errorCode, final String
      * Extract the HTTP status code from an exception.
      *
      * @param t The exception to extract the code from.
-     * @return The HTTP status code, or 500 if the exception is not of type {@link ServiceInvocationException}.
+     * @return The HTTP status code. Otherwise {@link HttpURLConnection#HTTP_INTERNAL_ERROR} 
+     *         if the exception is {@code null} or not of type {@link ServiceInvocationException}.
      */
     public static int extractStatusCode(final Throwable t) {
-        return Optional.of(t).map(cause -> {
+        return Optional.ofNullable(t).map(cause -> {
```
sophokles73, line 172: nice catch

sophokles73 (approved):
LGTM

PR #2248, opened 2020-10-12
Fix retrieval of valid secrets by filtering out disabled secrets

kaniyan:
`CredentialsObject.getCandidateSecrets(...)` is used to retrieve valid secrets.  Currently the secrets are filtered based on their validity period regardless of whether the status is enabled or not. This has been fixed now, so that the disabled secrets are filtered out.


sophokles73:
@kaniyan can you also add en entry to the release notes?

kaniyan:
> @kaniyan can you also add en entry to the release notes?

I have now added an entry to the release notes.

sophokles73 (approved):
LGTM

Issue #2244, opened 2020-10-09
[MQTT/AMQP] Issues with devices not properly disconnecting and then reconnecting


calohmn:
When an MQTT device connection gets broken without the MQTT adapter noticing, and when the device then reconnects (before the MQTT keepAlive period has elapsed) and re-subscribes for commands, one of these issues may arise:
- **A:** If reconnect happened on different adapter/verticle instance:
The late disconnect (after the keepAlive period) will trigger a "disconnectedTtdEvent" message, superceding the "connectedTtdEvent" message from the reconnect, meaning the latest such event message doesn't reflect the "connected" state of the device.
- **B:** If reconnect happened on *same* adapter/verticle instance:
The late disconnect will also cause the command consumer from the reconnect attempt to be removed. This means the reconnected device won't receive any commands.

Scenario A described in more detail:
- An MQTT device has subscribed for receiving command messages. 
- The MQTT connection then gets broken somehow, without a proper disconnect getting sent, so that the MQTT adapter first doesn't notice that the device is disconnected.
- Then the MQTT device reconnects with another MQTT adapter instance (or verticle instance) and subscribes again. This means an event message (with `ttd=-1`) is sent downstream, notifying the northbound application about the connected device.
- Then the MQTT adapter, that the device was initially connected with, notices the broken connection via the MQTT keep alive mechanism. This means again an event message is sent downstream, this time meaning the device isn't connected anymore (`ttd=0`). This means the latest event message that the northbound application receives (ie. the "disconnected" event), doesn't reflect the actual connection status of the device. This possibly means that the application doesn't send commands anymore although the device is connected and has subscribed for commands.

When testing with `mosquitto_sub`: The case of the broken (half-open) MQTT connection can be simulated by suspending the `mosquitto_sub` process (CTRL+Z).

Both issues probably also apply to a corresponding AMQP device scenario.

calohmn:
Ideas for fixing this:

#### A: If reconnect happened on different adapter/verticle instance
Upon disconnect, when de-registering the command consumer, the command handling adapter instance mapping gets removed.
As there is already a new mapping (from the reconnect/re-subscribe) with a different instance id, this operation will return a 404. In that case the "disconnectedTtdEvent" shall not get sent.

#### B: If reconnect happened on same adapter/verticle instance
[...]

PR #2251, opened 2020-10-14
[#2244] Skip sendDisconnectedTtdEvent if consumer not mapped

calohmn:
No 'disconnectedTtdEvent' is now sent anymore when a command consumer is obviously not registered anymore
because its associated command handling adapter instance entry has already been removed or overwritten.

This resolves both cases described in #2244.

### DIFF
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
@@ -384,7 +384,7 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
     }
 
     private void onConnectionLoss(final ProtonConnection con) {
-        final Span span = newSpan("handle closing of command connection", getAuthenticatedDevice(con),
+        final Span span = newSpan("handle closing of connection", getAuthenticatedDevice(con),
                 getTraceSamplingPriority(con));
         @SuppressWarnings("rawtypes")
         final List<Future> handlerResults = getConnectionLossHandlers(con).stream().map(handler -> handler.apply(span))
@@ -577,7 +577,7 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
                 Device.class);
         final OptionalInt traceSamplingPriority = getTraceSamplingPriority(conn);
 
-        final Span span = newSpan("attach receiver", authenticatedDevice, traceSamplingPriority);
+        final Span span = newSpan("attach device sender link", authenticatedDevice, traceSamplingPriority);
         span.log(Map.of("snd-settle-mode", receiver.getRemoteQoS()));
 
         final String remoteTargetAddress = Optional.ofNullable(receiver.getRemoteTarget())
@@ -721,7 +721,7 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
                 Device.class);
         final OptionalInt traceSamplingPriority = getTraceSamplingPriority(connection);
 
-        final Span span = newSpan("attach Command receiver", authenticatedDevice, traceSamplingPriority);
+        final Span span = newSpan("attach device command receiver link", authenticatedDevice, traceSamplingPriority);
 
         getResourceIdentifier(sender.getRemoteSource())
         .compose(address -> validateAddress(address, authenticatedDevice))
@@ -731,14 +731,11 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
                 return openCommandSenderLink(connection, sender, validAddress, authenticatedDevice, span, traceSamplingPriority)
                         .map(consumer -> {
                             setConnectionLossHandler(connection, validAddress.toString(), connectionLossSpan -> {
-                                // do not use the above created span for onCommandConnectionClose()
+                                // do not use the above created span for closing the command consumer
                                 // because that span will (usually) be finished long before the
                                 // connection is closed/lost
-                                final Future<ProtonDelivery> sendEventFuture = sendDisconnectedTtdEvent(validAddress.getTenantId(), 
-                                        validAddress.getResourceId(), authenticatedDevice, connectionLossSpan.context());
-                                final Future<Void> closeConsumerFuture = consumer.close(connectionLossSpan.context());
-
-                                return CompositeFuture.join(sendEventFuture, closeConsumerFuture).mapEmpty();
+                                return closeCommandConsumer(consumer, validAddress.getTenantId(),
+                                        validAddress.getResourceId(), authenticatedDevice, connectionLossSpan);
                             });
                             return consumer;
                         });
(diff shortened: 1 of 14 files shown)


calohmn:
Note that with this change, currently there will also be no "disconnectedTtdEvent" getting sent when there was an exception removing the command handling adapter instance.
On second thought, I think it might be better sending an event in that case. I'll look into changing the commit accordingly.

client/src/main/java/org/eclipse/hono/client/impl/ProtocolAdapterCommandConsumerFactoryImpl.java
```diff
@@ -196,19 +198,29 @@ protected void onDisconnect() {
...
-            final SpanContext onCloseSpanContext) {
+            final SpanContext onCloseSpanContext, final boolean isUnlimitedLifespanConsumer) {
         log.trace("remove command consumer [tenant-id: {}, device-id: {}]", tenantId, deviceId);
         adapterInstanceCommandHandler.removeDeviceSpecificCommandHandler(tenantId, deviceId);
 
         return deviceConnectionClientFactory.getOrCreateDeviceConnectionClient(tenantId)
                 .compose(client -> client.removeCommandHandlingAdapterInstance(deviceId, adapterInstanceId,
                         onCloseSpanContext))
+                .compose(removed -> {
+                    if (!removed && isUnlimitedLifespanConsumer) {
```
sophokles73, line 209: I find it extremely confusing that *DeviceConnectionClientImpl.removeCommandHandlingAdapterInstance* method does not comply with the Device Connection API in that the returned future actually succeeds for status codes 404 and 412 but fails for e.g. a status code 404 being returned from the service.
I am even more confused seeing that a few lines below we actually map the successful result to a 404?

sophokles73 (changes requested):

calohmn:
(rebased and added tests)

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -823,6 +819,26 @@ private Span newSpan(final String operationName, final Device authenticatedDevic
...
+            final ProtocolAdapterCommandConsumer consumer,
+            final String tenantId,
+            final String deviceId,
+            final Device authenticatedDevice,
+            final Span span) {
+
+        return consumer.close(span.context())
+                .recover(thr -> {
+                    // ignore all but not-found errors
+                    if (ServiceInvocationException.extractStatusCode(thr) == HttpURLConnection.HTTP_NOT_FOUND) {
```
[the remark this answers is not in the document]
calohmn: I've changed it now that the relevant code to match here is indeed 412. (See also the `ProtocolAdapterCommandConsumer#close` javadoc.)

client/src/main/java/org/eclipse/hono/client/impl/ProtocolAdapterCommandConsumerFactoryImpl.java
calohmn: The change to let *DeviceConnectionClientImpl.removeCommandHandlingAdapterInstance* return a `Boolean` future value was done when introducing the lifespan for the corresponding cache entries, with which 404 exceptions when trying to remove a just expired entry can easily occur (when the ttd has elapsed). PR for that was #1926. With that change, a failure to remove a (potentially expired) entry will also not be traced as an error in the client anymore.
I think the corresponding server method wasn't changed back then for compatibility reasons and following up on this later on was abandoned because of the new command router concept.
I see that it may be confusing to use that client method with having the API definition in mind. I've changed the javadoc now to add a note about that difference.
We could also revert #1926 and just adapt the client tracing to skip logging a 412 error there. If doing so, that should be in a follow-up PR, but frankly, I don't see that much benefit in that, also since the plan is to abandon the device connection API for the command routing component concept.

> I am even more confused seeing that a few lines below we actually map the successful result to a 404?

I've changed that now to 412.

calohmn:
Amended the commit to fix also case B described in #2244. I'll still be adding integration tests here.

client/src/main/java/org/eclipse/hono/client/impl/ProtocolAdapterCommandConsumerFactoryImpl.java
```diff
@@ -195,11 +202,19 @@ protected void onDisconnect() {
...
-        log.trace("remove command consumer [tenant-id: {}, device-id: {}]", tenantId, deviceId);
-        adapterInstanceCommandHandler.removeDeviceSpecificCommandHandler(tenantId, deviceId);
+    private Future<Void> removeCommandConsumer(final CommandHandlerWrapper commandHandlerWrapper, final Duration lifespan,
+            final Instant lifespanStart, final SpanContext onCloseSpanContext) {
+
+        final String tenantId = commandHandlerWrapper.getTenantId();
+        final String deviceId = commandHandlerWrapper.getDeviceId();
 
+        log.trace("remove command consumer [tenant-id: {}, device-id: {}]", tenantId, deviceId);
+        if (!adapterInstanceCommandHandler.removeDeviceSpecificCommandHandler(commandHandlerWrapper)) {
```
sophokles73, line 212: this is the check for scenario B, right (device has re-connected to same verticle instance)? IMHO it would be very helpful to describe this scenario in a comment here

client/src/main/java/org/eclipse/hono/client/impl/ProtocolAdapterCommandConsumerFactoryImpl.java
```diff
@@ -208,7 +223,18 @@ protected void onDisconnect() {
                             deviceId, thr);
                     return Future.failedFuture(thr);
                 })
-                .mapEmpty();
+                .compose(removed -> {
+                    final boolean entryNotExpired = lifespan.isNegative() || Instant.now().isBefore(lifespanStart.plus(lifespan));
+                    if (!removed && entryNotExpired) {
+                        // entry wasn't actually removed and entry hasn't expired (yet)
```
sophokles73, line 229: this covers adapters which use stateful protocols like AMQP and MQTT where the TTD is -1, right?
again, IMHO it would be helpful to describe the circumstances under which this happens here in the comment in some more detail

sophokles73 (commented):
just some small hints

sophokles73 (approved):
LGTM

client/src/main/java/org/eclipse/hono/client/impl/ProtocolAdapterCommandConsumerFactoryImpl.java
calohmn: > this covers adapters which use stateful protocols like AMQP and MQTT where the TTD is -1, right?

Most importantly, but not exclusively. There could also be 2 overlapping HTTP adapter TTD requests, the above "if" block getting reached when calling `removeCommandConsumer` on the 1st request before the 2nd ttd period has elapsed. It's not that important throwing an exception in that case though, since the HTTP adapter won't send a "disconnectedTtd" event. But could be informative in the tracing span anyway.

PR #2252, opened 2020-10-14
Fix Hono CI build by encoding unsafe characters while creating URI

kaniyan:
Some tests as below fail during the Hono CI build. The reasons is that the URL namely `jdbc:h2:/jobs/genie.hono/Hono CI Pipeline/workspace/services/device-registry-jdbc/target/data/d222ff86-346d-413b-b40e-17fb48004579` contains space characters. This causes issue while creating an URI object to get the JDBC database type. This has been fixed by encoding unsafe characters in the URL.

```
[ERROR] Tests run: 16, Failures: 0, Errors: 16, Skipped: 0, Time elapsed: 3.971 s <<< FAILURE! - in org.eclipse.hono.deviceregistry.jdbc.impl.JdbcBasedCredentialsServiceTest
[ERROR] testUpdateCredentialsSupportsRemovingExistingCredentialsAndSecrets{VertxTestContext}  Time elapsed: 0.009 s  <<< ERROR!
java.lang.IllegalArgumentException: Illegal character in opaque part at index 29: jdbc:h2:/jobs/genie.hono/Hono CI Pipeline/workspace/services/device-registry-jdbc/target/data/d222ff86-346d-413b-b40e-17fb48004579
```

PR #2253, opened 2020-10-15
[#2252] Fix error while creating URI by encoding unsafe characters

kaniyan:
This is related to the PR #2252. There is one more place where the URI creation should be fixed by encoding unsafe characters such as space in the URL string. This PR fixes it.


sophokles73 (approved):
LGTM

PR #2259, opened 2020-10-22
Make tenant ID property mandatory

sophokles73:
The TenantObject now requires the mandatory properties tenant-id and
enabled to be specified in the constructor.


kaniyan (approved):
LGTM

Issue #2260, opened 2020-10-23
Do "adapter enabled for tenant" check before credentials validation


calohmn:
Followup of #2115:
Currently, a connection attempt from a device means that the target protocol adapter does some tenant related checks only *after* the credentials given by the device have been validated.

As the credentials validation may be quite expensive, it would be better to do the tenant related checks *before* that already, as soon as tenant information has been obtained from the credentials.

The tenant related checks are:
- check whether  the adapter is enabled for the tenant,
- check whether the connection limit for the tenant is not exceeded,
- check whether the message limit for the tenant is not exceeded by the request payload (HTTP/CoAP adapter).

EDIT: only the check whether the adapter is enabled for the tenant is to be done here - see below.

calohmn:
PRs for the HTTP and MQTT adapter have been created. 
The changes there only cause a minor change of behaviour towards an HTTP device: A connection attempt with wrong credentials and with one of the above checks failing, will now result in an error outcome pointing to the failed check (e.g. "403 (Forbidden)" for the HTTP adapter not being enabled for the tenant), not pointing to the wrong credentials ("401 (Unauthorized)"). This doesn't break the behaviour outlined in the HTTP adapter documentation FMPOV, as there is no order of precedence defined for the different possible error scenarios.

For the AMQP adapter, however, letting the checks be done as part of the credentials validation would result in a more noticeable change of behaviour towards the AMQP device:
Exceeding the tenant connection limit for example would then be met with a SASL handshake error (probably SASL_PERM). The current behaviour on the other hand means that the SASL handshake succeeds (provided credentials are correct) and then the connection is getting closed with an "UNAUTHORIZED_ACCESS" error condition (and possibly an error string, but that isn't used as of now).
I still think it would be good to make that change also for the AMQP adapter, adding an explicit note in the AMQP adapter guide about the possible SASL handshake error reasons.

sophokles73:
> I still think it would be good to make that change also for the AMQP adapter, adding an explicit note in the AMQP adapter guide about the possible SASL handshake error reasons.

That is the behavior that we had in previous version of Hono already. We deliberately chose to change that and do the capacity check etc *after* the SASL handshake because AMQP does not provide means to convey an error message to the client as part of a (failed) SASL handshake. So FMPOV we should keep the AMQP adapter as-is ...

calohmn:
Ok, I see. I'll skip changes to the AMQP adapter here then.

kaniyan:
[...]

PR #2261, opened 2020-10-26
[#2260] HTTP adapter: Check tenant message limit earlier

calohmn:
This is for #2260:
The check in the HTTP adapter whether a request payload exceeds the tenant message limit is now done before the (potentially expensive) credentials validation. The check whether the adapter is enabled for the tenant is done at that point already as well.


calohmn:
Closing this - see https://github.com/eclipse/hono/issues/2260#issuecomment-717274557.

Issue #2133, opened 2020-08-26
Message mapping API does not return a failed future


Alfusainey:
Maybe I do not understand the mapper endpoint feature added recently, so please bear with me :-) I will close the issue once my doubt is clarified ;-)

According to HttpBasedMessageMapping#mapMessage API docs, the method is supposed to returned
1. a `mapped message` if the device payload can be successfully mapped to a `new payload type` OR
2. a `failed future`, with a ServiceInvocationException, if the message cannot be mapped. This happens when the device registration info contains a mapper in its config but the adapter is not configured with an Endpoint corresponding to the configured mapper for the device/gateway

However, the current default implementation present with Hono does not do that. In the default impl., if a mapper is not configured for a device/gateway, or it is configured but endpoint is either missing or not configured for the adapter, then the original message is mapped `as-is`. No failure is returned. If this is the intended behavior, should the API docs be adapted accordingly?

If you look at where the API is used (i.e the `onPublishedMessage` handler of the MQTT adapter), I can see why the original message is mapped `as-is` instead of returning a failure (makes the caller code easier I think). IMHO, the mapper should not map all incoming MQTT messages: either it successfully maps an incoming message (if it can) or fails. In the case of a failure, the MQTT adapter can proceed to forward the message like before, ignoring the failed attempt to mapped the message.

sophokles73:
@BobClaerhout Hi Bob, would you mind taking a look at this and get in touch with @Alfusainey ?

BobClaerhout:
Thanks for the notification @sophokles73.
Hi @Alfusainey,

The primary goal of adding the mapper is/was making sure that the other processing continued without any problems at all costs. This means that it should not fail in any way when nothing is configured, wrongly configured, ...

If I'm correct, the mapper will return a failure when the actual mapping doesn't work (in the pure sense of the word, this is a failure to me). I think what you are talking about here is more of a semantic discussion on the word "failure". Did the mapping really fail if e.g. the mapping was not configured at all or wrongly configured? If so, we can return a failure as well. If we believe this is not really a failure but wrong configuration and for that reason the mapper is doing just fine, we can just return the original payload as-is.

Personally, I don't really have a preference, I can live with both implementations, it's just a different take on this but the end result is the same. I have one remark though: If the mapper is capable of doing error handling (like catching wrong configuration) and returning the original payload, we do not need to repeat ourselves in all the adapter doing the same error handling. But as I said, either way is good for me.

[...]

PR #2262, opened 2020-10-26
adapt MessageMapping#mapMessage API documentation

Alfusainey:
According to discussions in #2133, it appears that the `MessageMapping#mapMessage` API should always return a successful future, which is not currently reflected in the API documentation.


sophokles73:
We should probably wait for the result of the discussion in #2133 before starting the review ...

Alfusainey:
closing this PR. a new PR will be needed to address/fix #2133

PR #2264, opened 2020-10-26
[#2260] MQTT adapter: Check tenant connection limit earlier

calohmn:
This is for #2260:
The check whether a connection attempt exceeds the tenant connection limit is now done before the (potentially expensive) credentials validation. The check whether the adapter is enabled for the tenant is now done at that point as well.


calohmn:
Closing this - see https://github.com/eclipse/hono/issues/2260#issuecomment-717274557.

PR #2265, opened 2020-10-26
Workaround for quarkus system logging issue

dejanb:
The issue with the Quarkus system logging is rooted in the jboss-logmanager-embedded library. The PR to fix it is already in (https://github.com/dmlloyd/jboss-logmanager-embedded/pull/2).

This workaround helps the test pass, but I'm not sure it'll be fully configurable in Quarkus per (https://groups.google.com/g/quarkus-dev/c/g0pSv77NNno/m/K9ZijGDmBwAJ). But it should get us going until jbos-logmanager-embedded:1.0.6 is released.

### DIFF
--- a/service-base-quarkus/pom.xml
+++ b/service-base-quarkus/pom.xml
@@ -35,6 +35,12 @@
       <artifactId>quarkus-arc</artifactId>
       <version>${quarkus.platform.version}</version>
       <scope>compile</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.jboss.logmanager</groupId>
+          <artifactId>jboss-logmanager-embedded</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>io.jaegertracing</groupId>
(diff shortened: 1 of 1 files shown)


service-base-quarkus/pom.xml
```diff
@@ -149,6 +155,14 @@
   <profiles>
     <profile>
       <id>jboss-logmanager</id>
+      <!-- alternative until the problem with jboss-logmanager-embedded is fixed -->
+      <dependencies>
+        <dependency>
+          <groupId>org.jboss.logmanager</groupId>
+          <artifactId>jboss-logmanager</artifactId>
```
sophokles73, line 162: I had also tried to replace the embedded logmanager with the real thing. However, the *real thing* doesn't contain the `org.jboss.logmanager.formatters.ColorPatternFormatter` which is used by Quarkus during startup :-(

sophokles73:
I guess we will need to wait for jboss-logmanager-1.0.6 ...

service-base-quarkus/pom.xml
dejanb: Ah, I remember now hitting that one at some point as well.

dejanb:
OK, I'll close this one then and I'll keep track on 1.0.6 release

PR #2269, opened 2020-10-27
Remove obsolete code

kaniyan:


core/src/main/java/org/eclipse/hono/util/Constants.java
```diff
@@ -309,17 +295,6 @@ public static void copyProperties(final ProtonConnection source, final ProtonLin
...
 
-    /**
-     * Gets the (surrogate) identifier of the AMQP connection that a link is part of.
-     *
-     * @param link The link to determine the connection id for.
-     * @return The identifier retrieved from the link's <em>attachment</em> using key {@link #KEY_CONNECTION_ID}
-     *         or {@code null} if the attachments do not contain a value for that a key.
-     */
-    public static String getConnectionId(final ProtonLink<?> link) {
-        return link.attachments().get(KEY_CONNECTION_ID, String.class);
```
calohmn, line 320: It seems to me the `KEY_CONNECTION_ID` constant itself and the other remaining places where it is used can be removed as well. As far as I see it, the id was used in connection with the long removed Hono messaging component (?).

kaniyan: As I understood, `KEY_CONNECTION_ID` is being used by `copyProperties(...)` and `getConnectionId(final ProtonConnection connection)`. And `copyProperties(...)` is in use in `AmqpServiceBase` and `SimpleAuthenticationServer`.

calohmn: Yes, but the ID is only ever *set* in the attachments. Except for the `copyProperties` method, the ID is not read (see usage of `getConnectionId`). See also the long removed [HonoMessaging](https://github.com/eclipse/hono/blob/4575c30f1cee5b3e537c9a11b904e1f29d01c716/services/messaging/src/main/java/org/eclipse/hono/messaging/HonoMessaging.java#L90) class.

kaniyan: Good point. I have now removed the obsolete code regarding `connection-id` which is being set in AMQP connection attachments.

calohmn (approved):
LGTM

PR #2270, opened 2020-10-27
[#2260] Check if adapter enabled for tenant earlier on

### LINKED ISSUE #2260: Do "adapter enabled for tenant" check before credentials validation

calohmn:
This fixes #2260.

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -61,6 +61,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import io.micrometer.core.instrument.Timer.Sample;
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
+import io.opentracing.noop.NoopSpan;
 import io.opentracing.tag.Tags;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.CompositeFuture;
@@ -331,6 +332,8 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
      * information before the potentially expensive credentials validation is done
      * <p>
      * The default implementation updates the trace sampling priority in the execution context tracing span.
+     * It also verifies that the tenant provided via the credentials is enabled and that the adapter is enabled for
+     * that tenant, failing the returned future if either is not the case.
      * <p>
      * Subclasses should override this method in order to perform additional operations after calling this super method.
      *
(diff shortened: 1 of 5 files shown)


service-base/src/main/java/org/eclipse/hono/service/http/AuthHandlerTools.java
```diff
@@ -40,13 +40,19 @@ private AuthHandlerTools() {
...
+     * In all other cases, the routing context is failed with the given exception.
+     * <p>
+     * Note that the routing context is failed with just the exception, no status
+     * code. Setting the status code on the response corresponding to the exception
+     * is to be done in the failure handler, e.g. as implemented in the {@link DefaultFailureHandler}.
      *
      * @param ctx The routing context.
      * @param exception The cause of failure to process the request.
      * @param authenticateHeader The value to return in the HTTP Authenticate header.
+     * @see io.vertx.ext.web.handler.impl.AuthHandlerImpl#processException(RoutingContext, Throwable)
```
kaniyan, line 55: `AuthHandlerImpl#processException` is inaccessible from this class as that method is _protected_.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
```diff
@@ -331,6 +332,7 @@ protected Router createRouter() {
      * information before the potentially expensive credentials validation is done
      * <p>
      * The default implementation updates the trace sampling priority in the execution context tracing span.
+     * It also verifies that the adapter is enabled for the tenant, failing the returned future if that is not the case.
```
kaniyan, line 335: In addition to the adapter, `isAdapterEnabled(...)` also verifies if the tenant is _enabled_ or not. How about including this info about tenant also in the JavaDoc ?  It is also applicable for the MQTT adpter too. WDYT?

kaniyan (changes requested):

service-base/src/main/java/org/eclipse/hono/service/http/AuthHandlerTools.java
calohmn: There was a discussion on such a case earlier already (regarding `TableAdapterStore` class) and the outcome was, that the javadoc tool won't reject this with an error anyway, so it was deemed ok.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
calohmn: I've adapted the javadocs.

kaniyan (approved):
LGTM

sophokles73 (approved):
LGTM

PR #2271, opened 2020-10-27
Fix javadoc in TelmetrySender and EventSender

kaniyan:
* With PR #2259, the tenant identifier is made mandatory for the `TenantObject`. The `EventSender.sendEvent(...)` and `TelmetrySender.sendTelemetry(...)` methods throwing `IllegalArgumentException` in case of no tenant identifier, is obsolete. This has been fixed in the JavaDoc.
 * Minor improvement by replacing `stream().forEach()` with `forEach()`


sophokles73 (approved):
LGTM

PR #2275, opened 2020-10-30
Create downstream message from properties gathered from context

sophokles73:
The protocol adapters now create downstream messages using the new
MessageHelper.newMessage() method which accepts a Map of properties to
include in the message.

This is in preparation for the adapters being refactored to use the new
TelemetrySender and EventSender instead of the legacy DownstreamSender.

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttContext.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttContext.java
@@ -13,11 +13,14 @@
 
 package org.eclipse.hono.adapter.mqtt;
 
+import java.time.Duration;
 import java.util.Objects;
 import java.util.Optional;
 
 import org.eclipse.hono.auth.Device;
 import org.eclipse.hono.service.metric.MetricsTags;
+import org.eclipse.hono.service.metric.MetricsTags.EndpointType;
+import org.eclipse.hono.util.Constants;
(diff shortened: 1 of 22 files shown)


adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttContext.java
```diff
@@ -152,11 +156,22 @@ public Device authenticatedDevice() {
...
+     * @return The type of the message payload.
      */
     public String contentType() {
-        return contentType;
+        if (contentType != null) {
+            return contentType;
+        }
+        return Optional.ofNullable(propertyBag)
+                .flatMap(bag -> Optional.ofNullable((String) bag.getProperty(HttpHeaders.CONTENT_TYPE.toString())))
+                .orElse(MessageHelper.CONTENT_TYPE_OCTET_STREAM);
```
calohmn, line 174: (Redundant String cast.)
The extra propertyBag check looks superfluous here since all `MqttContext` fields except `timer` are effectively final. (And there is a different content-type constant being used compared to the constructor code.)

sophokles73: Yes, I stumbled across that as well during the refactoring. I will remove the *setContentType* method and make the contentType field final, initializing it from the property bag during creation ...

sophokles73:
@calohmn I have pushed some changes. Would you mind taking another look?

calohmn (approved):
LGTM

Issue #2029, opened 2020-06-11
Introduce CommandRouting service/component (replacing DeviceConnectionService)


calohmn:
The current Command & Control implementation uses tenant scoped `command/[tenantId]` links to initially receive commands from a downstream application. After having received a command message on that link, the protocol adapter instance that the command target device is connected to is being identified and the command is sent/routed to that adapter instance.

Currently, each adapter supports these tenant-scoped links and the corresponding routing logic.

This means added complexity for protocol adapter implementations and makes command processing also somewhat unintuitive (a command may be received first by an AMQP adapter and then routed to an MQTT adapter for example).
It also means that potential issues in a custom protocol adapter could have consequences for commands directed at one of the standard Hono protocol adapters (if these command messages are received by the custom protocol adapter first).

This all leads to the idea of introducing a separate component that will first receive **all** command messages and then route them to the appropriate protocol adapter instance.
Such a `CommandRouting` component could then include also the implementation of the current `DeviceConnectionService`. That means that the `CommandRouting` implementation has the service to identify the target protocol adapter instance of a command message right inside its own component. Thinking further along this line, we can just as well replace the `DeviceConnectionService` with the `CommandRouting` service.

Operations of the new `CommandRouting` service (to be used by the protocol adapters):
- `setLastKnownGatewayForDevice(tenant, device, gateway)`
- `registerCommandConsumer(tenant, device, consumerId, lifespan)`
- `unregisterCommandConsumer(tenant, device, consumerId)`

(~~`consumerId` here means the same as what was called `adapterInstanceId` in the *DeviceConnectionAPI*; `adapterInstanceId` was a bit misleading, as there are usually multiple such consumers per adapter instance, corresponding to the number of vert.x verticle instances~~ EDIT: using `adapterInstanceId` is probably more appropriate since we want to keep the number of related resources (e.g. Kafka topics) low. That requires mapping incoming commands to the vert.x verticles (via vert.x event bus), though.)

For #1276, this method will probably be needed as well:
- `registerCommandConsumer(tenant, device, correlationId, consumerId, lifespan)`

The `AdapterInstancesLivenessService` (#2028) will then also be incorporated into the `CommandRouting` service. For that, an additional, optional operation could be added:
- `ping(consumerId)`

This would then be used by custom protocol adapters that are not in the same kubernetes cluster as the other adapters and the `CommandRouting` component.

calohmn:
The architecture and command message flow would look like this then:
[...]

PR #2280, opened 2020-11-04
[#2029] Add Command Router API specification

calohmn:
This is for #2029. 

Adds the specification for a new `Command Router API`, providing methods to configure the Command Router component so that it can route command & control messages to their target.


site/documentation/content/api/command-router/index.md
```diff
@@ -0,0 +1,174 @@
...
+linkTitle: "Command Router API"
+weight: 425
+resources:
+  - src: preconditions.svg
+  - src: register_cmd_consumer.svg
+  - src: unregister_cmd_consumer.svg
+  - src: set_last_known_gateway_success.svg
+---
+
+The *Command Router API* is used by *Protocol Adapters* to configure the Command Router service component, enabling it to route command & control messages to the protocol adapters that the target devices are connected to.
```
sophokles73, line 12: I do not think that the adapters *configure* the component itself, that would be done by environment variables during startup, wouldn't it?
How about
> Protocol Adapters use the *Command Router API* to supply information during run time that a Command Router service component uses to route command & control messages ...

site/documentation/content/api/command-router/index.md
```diff
@@ -0,0 +1,174 @@
...
+<!--more-->
+
+The Command Router API is defined by means of AMQP 1.0 message exchanges, i.e. a client needs to connect to Hono using an AMQP 1.0 client in order to invoke operations of the API as described in the following sections.
```
sophokles73, line 15: needs to connect to the service component?

site/documentation/content/api/command-router/index.md
```diff
@@ -0,0 +1,174 @@
...
+
+1. Client has established an AMQP connection with the Command Router service.
+1. Client has established an AMQP link in role *sender* on the connection using target address `cmd_router/${tenant_id}`. This link is used by the client to send request messages to the Command Router service.
+1. Client has established an AMQP link in role *receiver* on the connection using source address `cmd_router/${tenant_id}/${reply-to}` where *reply-to* may be any arbitrary string chosen by the client. This link is used by the client to receive responses to the requests it has sent to the Command Router service. This link's source address is also referred to as the *reply-to* address for the request messages.
+
+{{< figure src="preconditions.svg" alt="A client establishes an AMQP connection and the links required to invoke operations of the Command Router service" title="Client connecting to Command Router service" >}}
+
+## Register command consumer for device
+
+Clients use this command to *register* a consumer for command & control messages for a device, providing the protocol adapter instance that the given device or gateway is currently connected to.
```
sophokles73, line 28: This sounds like the consumer and the protocol adapter instance are different things here. How about
> Clients use this command to *register* a protocol adapter instance as the consumer of command & control messages for a device or gateway that is currently connected to that particular adapter instance.

site/documentation/content/api/command-router/index.md
```diff
@@ -0,0 +1,174 @@
...
+
+A response to a *unregister command consumer for device* request contains the [Standard Response Properties]({{< relref "#standard-response-properties" >}}).
+
+The response message's *status* property may contain the following codes:
+
+| Code  | Description |
+| :---- | :---------- |
+| *204* | OK, the command consumer registration entry for the device has been removed. |
+| *400* | Bad Request, the request message does not contain all required properties. |
+| *412* | Precondition failed, the registration entry for the device has not been removed because there is no command consumer registration entry matching the adapter instance assigned to the device. |
```
sophokles73, line 98: may we can add an example here for a sequence of actions that might lead to 412 being returned, e.g. a device losing its connection and re-connecting to a different adapter instance?

site/documentation/content/api/command-router/index.md
```diff
@@ -0,0 +1,174 @@
...
+{{< figure src="set_last_known_gateway_success.svg" title="Client sets the last known gateway for a device" alt="A client sends a request message for setting the last known gateway and receives a response containing a confirmation" >}}
+
+**Request Message Format**
+
+The following table provides an overview of the properties a client needs to set on a message to set the last known gateway for a device in addition to the [Standard Request Properties]({{< relref "#standard-request-properties" >}}).
+
+| Name         | Mandatory | Location                 | AMQP Type | Description |
+| :----------- | :-------: | :----------------------- | :-------- | :---------- |
+| *subject*    | yes       | *properties*             | *string*  | MUST be set to `set-last-gw`. |
+| *gateway_id* | yes       | *application-properties* | *string*  | The identifier of the gateway that last acted on behalf of the device identified by the *device_id* property. If a device connects directly instead of through a gateway, the device's identifier MUST be specified here. |
```
sophokles73, line 122: I think we can make the last sentence a little more explicit. How about
> For a device that connects to the adapter directly instead of through a gateway, the value of this property MUST be the same as the value of the *device_id* application property.

site/documentation/content/api/command-router/index.md
calohmn: Yes. That's something to be changed in some of the other API docs as well (credentials, device-registration, etc).

site/documentation/content/api/command-router/index.md
calohmn: Yes, I've added:
> This may happen if the mapping entry has already been removed or its lifespan has elapsed. The entry could also have been updated with a different adapter instance identifier before, which could for example mean the device lost its connection to the protocol adapter and has reconnected to a different adapter instance. The original adapter instance trying to remove the mapping entry at some point after that will result in a *412* error response because the instance identifier doesn't match.

site/documentation/content/api/command-router/index.md
calohmn: Yes, maybe *configure* is a bit confusing here.
I've adapted the proposal slightly to:
> *Protocol Adapters* use the *Command Router API* to supply information with which a Command Router service component can route command & control messages to the protocol adapters that the target devices are connected to.

sophokles73 (approved):
LGTM

PR #2281, opened 2020-11-04
Use Maven profile for running Quarkus based adapters in ITs

sophokles73:
The ITs can now be run by activating the "protocol-adapters-quarkus"
Maven profile.

### DIFF
--- a/tests/pom.xml
+++ b/tests/pom.xml
@@ -33,23 +33,69 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch 
     <maven.source.skip>true</maven.source.skip>
     <maven.install.skip>true</maven.install.skip>
     <gpg.skip>true</gpg.skip>
+
+    <!--
+      Hono service names
+    -->
+    <hono.amqp-network.host>hono-dispatch-router.hono</hono.amqp-network.host>
+    <hono.device-connection.host>hono-service-device-connection.hono</hono.device-connection.host>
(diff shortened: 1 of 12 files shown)


tests/pom.xml
```diff
@@ -243,15 +277,31 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
...
         <hono.mongodb.port>27017</hono.mongodb.port>
         <hono.mongodb.username>device-registry@HONO</hono.mongodb.username>
         <hono.mongodb.password>device-registry-secret</hono.mongodb.password>
         <hono.mongodb.database.name>hono-it-tests</hono.mongodb.database.name>
-        <hono.deviceregistry.resources.folder>deviceregistry-mongodb</hono.deviceregistry.resources.folder>
-        <deviceregistry.supportsSearchDevices>true</deviceregistry.supportsSearchDevices>
+      </properties>
+    </profile>
+    <profile>
+      <id>protocol-adapters-quarkus</id>
```
dejanb, line 291: Maybe keep profile name a bit more generic as I can see we use Quarkus for other services not just protocol adapters? Or you want to configure that with separate profiles?

dejanb (approved):
Looks good

tests/pom.xml
sophokles73: No, I do not think that we need to use separate profiles for it. How about changing it to a more generic name once we start migrating the first non-adapter component?

Issue #2267, opened 2020-10-27
Add AMQP 1.0 based implementation of Hono client for protocol adapters


sophokles73:
Needs to implement the client interfaces defined as part of #2156

sophokles73:
FMPOV this can be closed. WDYT? @calohmn @kaniyan @b-abel

PR #2285, opened 2020-11-09
[#2267] Decouple command consumer from device connection client factory

sophokles73:
The ProtocolAdapterCommandConsumerFactory's dependency on the legacy
DeviceConnectionClientFactory has been replaced by an interface that
provides access to the required command handling adapter instance
information. This is a preliminary step for introducing a new
DeviceConnectionClient to the adapter-client module.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/ProtocolAdapterCommandConsumerFactoryImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/ProtocolAdapterCommandConsumerFactoryImpl.java
@@ -23,7 +23,6 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
-import org.eclipse.hono.client.BasicDeviceConnectionClientFactory;
 import org.eclipse.hono.client.ClientErrorException;
 import org.eclipse.hono.client.CommandContext;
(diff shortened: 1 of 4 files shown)


client/src/main/java/org/eclipse/hono/client/impl/ProtocolAdapterCommandConsumerFactoryImpl.java
```diff
@@ -81,7 +80,7 @@
     private final AtomicBoolean recreatingConsumers = new AtomicBoolean(false);
     private final AtomicBoolean tryAgainRecreatingConsumers = new AtomicBoolean(false);
 
-    private BasicDeviceConnectionClientFactory deviceConnectionClientFactory;
+    private CommandHandlingAdapterInfoAccess commandHandlingAdapterInfoAccessor;
```
calohmn, line 83: In the comment a few lines above:
````
     * Will be used to represent the protocol adapter instance that this factory instance is used in, when registering
     * command handlers with the Device Connection service.
     */
    private final String adapterInstanceId;
````
"Device Connection service" could be replaced with "CommandHandlingAdapterInfoAccess service".

sophokles73: I have pushed an update ...

calohmn (approved):
LGTM

Issue #2053, opened 2020-06-24
Implicit registration of devices connected via gateways


mhemmeter:
As of today all devices that connect to Hono via a gateway are supposed to be explicitly registered in the device registry with the via flag pointing to the gateway device. In this setup only gateway devices are authenticated and not the devices themselves. In order to simplify the management of devices in such a setup an _implicit_ registration of devices behind a gateway would be beneficial form my perspective. 

At first glance from security perspective an _implicit_ registration should be possible as after successful authentication of the gateway device a trust relation has been established.

For example such a feature would help in case devices connect via Bluetooth to a gateway that then communicates with Hono. Due to the features in Bluetooth devices may show up in a deployment dynamically and it may not be possible to "pre-register" the devices. It would be easier in a such a dynamic scenario if Hono registers these devices _implicitly_ upon connection instead requiring the gateway to invoke the Device Registry Management API for explicit registration.

mhemmeter:
Remark: An interesting question is if and how an implicit registration can work in a setup where Eclipse Hono is used in combination with Eclipse Ditto (see [Eclipse IoT Packages](https://www.eclipse.org/packages/). In such a setup devices must not only be registered in the Eclipse Hono device registry but also as a Thing in Ditto.

geglock:
Regarding the combination with Eclipse Ditto (via Eclipse IoT Packages or somehow else), a proposal could be to emit some kind of "welcome" message to the north-bound application for each implicit registration. 
If such a message could be defined freely per tenant including placeholder replacement, then it could be a "Thing Create" message - as defined here: https://www.eclipse.org/ditto/protocol-specification-things-create-or-modify.html
This way each implicitly registered device on Hono level would also trigger the creation of a respective Thing on Ditto level.

sophokles73:
I understand the use case and I also see the benefit of being able to register such devices on the fly. Devices being registered in this way will never be able to connect directly to a protocol adapter, though, because we would not register any credentials for them. This is consistent with the fact that the gateway device authenticates to the adapter and then acts on behalf of the edge device(s).
The biggest challenge in this context will be the management of the (edge) device identifiers.

I can imagine several approaches:
1. The gateway *generates* an identifier for the edge device to be registered and uses it when uploading data on behalf of the edge device. The protocol adapter and registry implementation determine that the device doesn't exist (yet) and then provision the device on the fly.
[...]

PR #2287, opened 2020-11-09
[#2053] Adding registry-internal auto-provisioning properties to devi…

fkaltner:
…ce status.

These properties are implemented as a preparation for the implementation of the auto-provisionoing of edge devices connected to gateways which relies on those properties in order to increase the robustness of the implementation.


services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceDto.java
```diff
@@ -32,6 +32,12 @@
     @JsonProperty(value = RegistryManagementConstants.FIELD_PAYLOAD_DEVICE_ID)
     private String deviceId;
 
+    @JsonProperty(RegistryManagementConstants.FIELD_AUTO_PROVISIONED)
+    private Boolean autoProvisioned;
+
+    @JsonProperty(RegistryManagementConstants.FIELD_AUTO_PROVISIONING_NOTIFICATION_SENT)
+    private Boolean autoProvisioningNotificationSent;
```
sophokles73, line 39: is it important that these are not of the primitive type? IMHO the code will be easier to understand if we do not need to account for `null` values all over the place ...

sophokles73 (commented):
I had expected to also find the *DeviceStatus* class. Am I missing something?

fkaltner:
> I had expected to also find the _DeviceStatus_ class. Am I missing something?

No, you are absolutely right about that. Forgot to add that file to the commit.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceDto.java
fkaltner: I think we could use the primitive type here. 

The only consequence as far as I can see would be: the autoProvisioningNotificationSent property would always be returned, also in cases of not-auto-provisioned devices it would also be returned as "false" (instead of being left out).

But I guess that's o.k. from a usability point of view?

sophokles73: I see. Then how about letting the *isAutoProvisioned* and *isAutoProvisioningNotificationSent* methods return a primitive boolean instead?

fkaltner:
Shall I squash all commits and convert to "real" (meaning non-draft) pr?

sophokles73:
Sure, if you think that it is *ready for prime time* now.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceDto.java
```diff
@@ -53,12 +61,13 @@ public DeviceDto() {
      *
      * @return A DTO instance for creating a new entry.
      */
-    public static <P extends Device, T extends DeviceDto> T forCreation(final Supplier<T> supplier, final String tenantId, final String deviceId, final Device device, final String version) {
+    public static <P extends Device, T extends DeviceDto> T forCreation(final Supplier<T> supplier, final String tenantId, final String deviceId, final Boolean autoProvisioned, final Device device, final String version) {
```
sophokles73, line 64: there doesn't seem to be any *P* involved in this method, or am I mistaken?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceDto.java
sophokles73: wouldn't it be easier to keep a *DeviceStatus* typed field instead of duplicating the *DeviceStatus* properties here?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceStatus.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Marks this device as auto-provisioned/manually provisioned.
+     *
+     * @param autoProvisioned {@code true}, if auto-provisioned. May be {@code null} which then defaults to false.
+     *
+     * @return a reference to this for fluent use.
+     */
+    public DeviceStatus setIsAutoProvisioned(final Boolean autoProvisioned) {
```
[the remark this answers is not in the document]
fkaltner: Okay, will rename.

The reason for accepting null are device registrations which were made before these two properties were introduced. 
For MongoDb this means that there are documents which do not contain that property and for the JDBC registry it depends on how the migration is done (either adding two columns and setting all present values to false or null).

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceDto.java
fkaltner: No, you're right. Thanks for spotting 👍

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceStatus.java
```diff
@@ -0,0 +1,81 @@
...
+import org.eclipse.hono.util.RegistryManagementConstants;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * A status used in Device value object.
+ */
+@JsonInclude(value = JsonInclude.Include.NON_NULL)
+public class DeviceStatus extends Status<DeviceStatus> {
```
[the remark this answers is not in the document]
fkaltner: No, added final modifier.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceDto.java
```diff
@@ -91,8 +89,10 @@ public static DeviceDto forRead(final String tenantId, final String deviceId, fi
         final DeviceDto deviceDto = BaseDto.forRead(DeviceDto::new, device, created, updated, version);
         deviceDto.setTenantId(tenantId);
         deviceDto.setDeviceId(deviceId);
-        deviceDto.setIsAutoProvisioned(autoProvisioned);
-        deviceDto.setIsAutoProvisioningNotificationSent(autoProvisioningNotificationSent);
+        deviceDto.setDeviceStatus(new DeviceStatus()
```
sophokles73, line 92: is there a particular reason why we do not simply pass a *DeviceStatus* instance as a parameter to this method? Or do you want to hide the existence of *DeviceStatus* to client code?

fkaltner: No, I hadn't information hiding in mind when I implemented this. Thinking a second time about this, your proposal makes sense as it shortens the parameter list.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/DeviceDto.java
```diff
@@ -76,23 +76,19 @@ public DeviceDto() {
...
+     * @param deviceStatus The registry internal status of the device.
      * @param created The instant when the object was created.
      * @param updated The instant of the most recent update.
      * @param version The version of the DTO
      *
      * @return A DTO instance for reading an entry.
      */
     public static DeviceDto forRead(final String tenantId, final String deviceId, final Device device,
-                                    final Boolean autoProvisioned, final Boolean autoProvisioningNotificationSent, final Instant created, final Instant updated, final String version) {
+                                    final DeviceStatus deviceStatus, final Instant created, final Instant updated, final String version) {
```
sophokles73, line 87: what about *forCreation* and *forUpdate*? It seems odd that one method accepts the DeviceStatus as arg but the others require DeviceStatus' properties explicitly, doesn't it?

fkaltner: _forCreation_ only should set the _autoProvisioned_ flag, _forUpdate_ only the _autoProvisioningNotificationSent_ flag.

If we made the method accept a DeviceStatus instance we would invoke the impression to the caller that one can set both flags in those calls which isn't true.

sophokles73: I see

sophokles73 (approved):
LGTM

sophokles73:
@fkaltner IMHO it would be good to have some assertions added to our device registry integration tests regarding the presence/absence of the status properties in responses returned by the registry. WDYT?

fkaltner:
> @fkaltner IMHO it would be good to have some assertions added to our device registry integration tests regarding the presence/absence of the status properties in responses returned by the registry. WDYT?

If I understood correctly the subclasses of _RegistrationServiceTests_ (to which I added assertions verifying the status properties) already spin up a local MongoDb instance / a local H2 instance. So they are pretty close to an integration test in my eyes.

In general I am open to add an additional test. But I am struggling to see which aspect you see being additionally taken into account when implementing a "real" integration test on top? 🤔

sophokles73:
I am talking about our *existing* integration tests in the *tests* module. In particular, the tests in the `org.eclipse.hono.tests.registry` package which verify the external behavior of all of the device registry implementations that we have. I think that we could easily add some assertions regarding the presence of the status properties that we expect to be included in the response body there. Thus, making sure that an external client actually gets what it expects, based on the Device Registry Management API.

fkaltner:
Understood. Added a new issue for the integration tests.

PR #2289, opened 2020-11-10
Reduce code duplication in Quarkus based adapters

sophokles73:
Introduced common base classes for protocol adapter applications.
Rearranged hono-service-base-quarkus classes to match package structure
of hono-service-base.


dejanb (approved):
Looks good

PR #2290, opened 2020-11-11
Add MockSupport classes in core-test-utils

calohmn:
Move `VertxMockSupport` from `hono-client` to `core-test-utils`; introduce `TracingMockSupport` class.


kaniyan (approved):
LGTM

PR #2293, opened 2020-11-12
[#2029] Add Command Router client

calohmn:
This is for #2029.

### DIFF
--- a/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedTenantCommandRouterClient.java
+++ b/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedTenantCommandRouterClient.java
@@ -1,0 +1,328 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.adapter.client.command.amqp;
+
+import java.net.HttpURLConnection;
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.UUID;
+
+import org.apache.qpid.proton.amqp.messaging.ApplicationProperties;
+import org.apache.qpid.proton.message.Message;
+import org.eclipse.hono.client.HonoConnection;
+import org.eclipse.hono.client.SendMessageSampler;
+import org.eclipse.hono.client.ServiceInvocationException;
+import org.eclipse.hono.client.StatusCodeMapper;
+import org.eclipse.hono.client.impl.AbstractRequestResponseClient;
+import org.eclipse.hono.tracing.TracingHelper;
+import org.eclipse.hono.util.CacheDirective;
+import org.eclipse.hono.util.CommandRouterConstants;
+import org.eclipse.hono.util.MessageHelper;
+import org.eclipse.hono.util.RequestResponseResult;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.opentracing.Span;
+import io.opentracing.SpanContext;
+import io.opentracing.tag.Tags;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.Promise;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.json.DecodeException;
+import io.vertx.core.json.JsonObject;
+import io.vertx.proton.ProtonDelivery;
+import io.vertx.proton.ProtonReceiver;
+import io.vertx.proton.ProtonSender;
+
+/**
+ * A Vertx-Proton based client for Hono's Command Router API, scoped to a specific tenant.
+ *
+ */
+class ProtonBasedTenantCommandRouterClient extends AbstractRequestResponseClient<RequestResponseResult<JsonObject>> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ProtonBasedTenantCommandRouterClient.class);
+
+    /**
+     * Creates a new client for accessing the Command Router service.
+     * <p>
+     * The client will be ready to use after invoking {@link #createLinks()} or
+     * {@link #createLinks(Handler, Handler)} only.
+     *
+     * @param connection The connection to the Command Router service.
+     * @param tenantId The identifier of the tenant for which the client should be created.
+     * @param sampler The sampler to use.
+     */
+    protected ProtonBasedTenantCommandRouterClient(final HonoConnection connection, final String tenantId, final SendMessageSampler sampler) {
+        super(connection, tenantId, sampler);
+    }
+
+    /**
+     * Creates a new client for accessing the Command Router service.
+     *
+     * @param connection The connection to the Command Router service.
+     * @param tenantId The identifier of the tenant for which the client should be created.
+     * @param sender The AMQP link to use for sending requests to the service.
+     * @param receiver The AMQP link to use for receiving responses from the service.
+     * @param sampler The sampler to use.
+     */
+    protected ProtonBasedTenantCommandRouterClient(
+            final HonoConnection connection,
+            final String tenantId,
+            final ProtonSender sender,
+            final ProtonReceiver receiver,
+            final SendMessageSampler sampler) {
+
+        super(connection, tenantId, sender, receiver, sampler);
+    }
+
+    /**
+     * Gets the AMQP <em>target</em> address to use for sending requests to Hono's Command Router API endpoint.
+     *
+     * @param tenantId The tenant to upload data for.
+     * @return The target address.
+     * @throws NullPointerException if tenant is {@code null}.
+     */
+    public static final String getTargetAddress(final String tenantId) {
+        return String.format("%s/%s", CommandRouterConstants.COMMAND_ROUTER_ENDPOINT, Objects.requireNonNull(tenantId));
+    }
+
+    @Override
+    protected final String getName() {
+        return CommandRouterConstants.COMMAND_ROUTER_ENDPOINT;
+    }
+
+    @Override
+    protected final String createMessageId() {
+        return String.format("cmd-router-client-%s", UUID.randomUUID());
+    }
+
+    @Override
+    protected final RequestResponseResult<JsonObject> getResult(
+            final int status,
+            final String contentType,
+            final Buffer payload,
+            final CacheDirective cacheDirective,
+            final ApplicationProperties applicationProperties) {
+
+        if (payload == null) {
+            return new RequestResponseResult<>(status, null, null, applicationProperties);
+        } else {
+            try {
+                // ignoring given cacheDirective param here - command router results shall not be cached
+                return new RequestResponseResult<>(status, new JsonObject(payload),
+                        CacheDirective.noCacheDirective(), applicationProperties);
+            } catch (final DecodeException e) {
+                LOG.warn("received malformed payload from Command Router service", e);
+                return new RequestResponseResult<>(HttpURLConnection.HTTP_INTERNAL_ERROR, null, null,
+                        applicationProperties);
+            }
+        }
+    }
+
+    /**
+     * Creates a new command router client for a tenant.
+     *
+     * @param con The connection to the server.
+     * @param tenantId The tenant to consumer events for.
+     * @param sampler The sampler to use.
+     * @param senderCloseHook A handler to invoke if the peer closes the sender link unexpectedly.
+     * @param receiverCloseHook A handler to invoke if the peer closes the receiver link unexpectedly.
+     * @return A future indicating the outcome of the creation attempt.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     */
+    public static final Future<ProtonBasedTenantCommandRouterClient> create(
+            final HonoConnection con,
+            final String tenantId,
+            final SendMessageSampler sampler,
+            final Handler<String> senderCloseHook,
+            final Handler<String> receiverCloseHook) {
+
+        LOG.debug("creating new command router client for [{}]", tenantId);
+        final ProtonBasedTenantCommandRouterClient client = new ProtonBasedTenantCommandRouterClient(con, tenantId, sampler);
+        // no response cache being set on client here - command router results shall not be cached
+        return client.createLinks(senderCloseHook, receiverCloseHook)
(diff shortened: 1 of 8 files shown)


clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterClient.java
```diff
@@ -0,0 +1,87 @@
...
+ * <p>
+ * See Hono's <a href="https://www.eclipse.org/hono/docs/api/command-router">
+ * Command Router API</a> for a description of the status codes returned.
+ */
+public interface CommandRouterClient extends Lifecycle {
+
+    /**
+     * Sets the given gateway as the last gateway that acted on behalf of the given device.
+     * <p>
+     * If a device connects directly instead of through a gateway, the device identifier itself is to be used as value
```
sophokles73, line 34: just out of curiosity: when would an adapter need to invoke this method with gatewayId = deviceId?

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterClient.java
```diff
@@ -0,0 +1,87 @@
...
+     * @param deviceId The device id.
+     * @param adapterInstanceId The protocol adapter instance id.
+     * @param lifespan The lifespan of the registration entry. Using a negative duration or {@code null} here is
+     *                 interpreted as an unlimited lifespan. Only the number of seconds in the given duration
+     *                 will be taken into account.
+     * @param context The currently active OpenTracing span context or {@code null} if no span is currently active.
+     *            An implementation should use this as the parent for any span it creates for tracing
+     *            the execution of this operation.
+     * @return A future indicating whether the operation succeeded or not.
+     * @throws NullPointerException if device id or adapter instance id is {@code null}.
```
sophokles73, line 62: or tenant id

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterClient.java
```diff
@@ -0,0 +1,87 @@
...
+     */
+    Future<Void> registerCommandConsumer(String tenantId, String deviceId, String adapterInstanceId, Duration lifespan,
+            SpanContext context);
+
+    /**
+     * Unregisters a command consumer for a device.
+     * <p>
+     * The registration entry is only deleted if its value contains the given protocol adapter instance id.
```
sophokles73, line 70: how about 
> ... if the device is currently mapped to the given adapter instance.

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterClient.java
```diff
@@ -0,0 +1,87 @@
...
+     * @param adapterInstanceId The protocol adapter instance id that the entry to be removed has to contain.
+     * @param context The currently active OpenTracing span context or {@code null} if no span is currently active.
+     *            An implementation should use this as the parent for any span it creates for tracing
+     *            the execution of this operation.
+     * @return A future indicating the outcome of the operation, with its value indicating whether the protocol
+     *         adapter instance value was removed or not.
+     *         <p>
+     *         The future will be failed with a {@link org.eclipse.hono.client.ServiceInvocationException} if there
+     *         was an error removing the value.
+     * @throws NullPointerException if device id or adapter instance id is {@code null}.
```
sophokles73, line 83: or tenant id

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterClient.java
```diff
@@ -0,0 +1,87 @@
...
+     * @param tenantId The tenant id.
+     * @param deviceId The device id.
+     * @param adapterInstanceId The protocol adapter instance id that the entry to be removed has to contain.
+     * @param context The currently active OpenTracing span context or {@code null} if no span is currently active.
+     *            An implementation should use this as the parent for any span it creates for tracing
+     *            the execution of this operation.
+     * @return A future indicating the outcome of the operation, with its value indicating whether the protocol
+     *         adapter instance value was removed or not.
+     *         <p>
+     *         The future will be failed with a {@link org.eclipse.hono.client.ServiceInvocationException} if there
```
sophokles73: The future will be failed with a {@code org.eclipse.hono.client.ServiceInvocationException} if there

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterClient.java
```diff
@@ -0,0 +1,87 @@
...
+     * If a device connects directly instead of through a gateway, the device identifier itself is to be used as value
+     * for the <em>gatewayId</em> parameter.
+     *
+     * @param tenantId The tenant id.
+     * @param deviceId The device id.
+     * @param gatewayId The gateway id (or the device id if the last message came from the device directly).
+     * @param context The currently active OpenTracing span context or {@code null} if no span is currently active.
+     *            An implementation should use this as the parent for any span it creates for tracing
+     *            the execution of this operation.
+     * @return A future indicating whether the operation succeeded or not.
```
sophokles73, line 43: IMHO we could/should be a little more specific about the type of exception that the future might be failed with

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterClient.java
```diff
@@ -0,0 +1,87 @@
...
+     * @param tenantId The tenant id.
+     * @param deviceId The device id.
+     * @param adapterInstanceId The protocol adapter instance id.
+     * @param lifespan The lifespan of the registration entry. Using a negative duration or {@code null} here is
+     *                 interpreted as an unlimited lifespan. Only the number of seconds in the given duration
+     *                 will be taken into account.
+     * @param context The currently active OpenTracing span context or {@code null} if no span is currently active.
+     *            An implementation should use this as the parent for any span it creates for tracing
+     *            the execution of this operation.
+     * @return A future indicating whether the operation succeeded or not.
```
sophokles73, line 61: IMHO we could/should be a little more specific about the type of exception that the future might be failed with

core/src/main/java/org/eclipse/hono/util/CommandRouterConstants.java
```diff
@@ -0,0 +1,94 @@
...
+         */
+        REGISTER_COMMAND_CONSUMER("register-command-consumer"),
+        /**
+         * The <em>unregister command consumer</em> operation.
+         */
+        UNREGISTER_COMMAND_CONSUMER("unregister-command-consumer"),
+        /**
+         * The <em>unknown</em> operation.
+         */
+        UNKNOWN("unknown");
```
sophokles73, line 46: do we need this?

core/src/main/java/org/eclipse/hono/util/CommandRouterConstants.java
```diff
@@ -0,0 +1,94 @@
...
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.util;
```
sophokles73, line 14: is this class required to be in the core module? IMHO it will only ever be needed in the context of the client itself, or am I mistaken?

core/src/main/java/org/eclipse/hono/util/CommandRouterResult.java
```diff
@@ -0,0 +1,102 @@
...
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.util;
```
sophokles73, line 14: is this class required to be in the core module? IMHO it will only ever be needed in the context of the client itself, or am I mistaken?

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedTenantCommandRouterClient.java
```diff
@@ -0,0 +1,326 @@
...
+    /**
+     * Creates a new command router client for a tenant.
+     *
+     * @param con The connection to the server.
+     * @param tenantId The tenant to consumer events for.
+     * @param sampler The sampler to use.
+     * @param senderCloseHook A handler to invoke if the peer closes the sender link unexpectedly.
+     * @param receiverCloseHook A handler to invoke if the peer closes the receiver link unexpectedly.
+     * @return A future indicating the outcome of the creation attempt.
+     * @throws NullPointerException if any of the parameters other than cache provider is {@code null}.
```
sophokles73, line 143: there is no cache provider param ...

clients/adapter-amqp/src/test/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedTenantCommandRouterClientTest.java
```diff
@@ -0,0 +1,510 @@
...
+                        // and the span is finished
+                        verify(span).finish();
+                    });
+                    ctx.completeNow();
+                }));
+    }
+
+    /**
+     * Verifies that a client invocation of the <em>unregister-command-consumer</em> operation
+     * returns a <em>Boolean.FALSE</em> value if a <em>PRECON_FAILED</em> response was returned.
```
sophokles73, line 241: we do not return a boolean value here, do we?

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterClient.java
```diff
@@ -0,0 +1,87 @@
...
+     * for the <em>gatewayId</em> parameter.
+     *
+     * @param tenantId The tenant id.
+     * @param deviceId The device id.
+     * @param gatewayId The gateway id (or the device id if the last message came from the device directly).
+     * @param context The currently active OpenTracing span context or {@code null} if no span is currently active.
+     *            An implementation should use this as the parent for any span it creates for tracing
+     *            the execution of this operation.
+     * @return A future indicating whether the operation succeeded or not.
+     * @throws NullPointerException if device id or gateway id is {@code null}.
```
sophokles73, line 44: or tenant id

core/src/main/java/org/eclipse/hono/util/CommandRouterConstants.java
calohmn: I've followed the same approach here as with the corresponding classes from the DeviceConnection, Tenant and Credentials APIs. It's used so that `CommandRouterAction.from` doesn't return `null`, meaning no extra null check is needed to use the return value in a "switch .. case".

core/src/main/java/org/eclipse/hono/util/CommandRouterResult.java
calohmn: If sticking to the way the other API services are implemented, the Result class is also used in the service interface in the form of `Future<CommandRouterResult>` method return values. That means it's also used in the command router server implementation.

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterClient.java
calohmn: Whether there really is a real *need* is a good question.

When a device has connected directly to a protocol adapter and has subscribed for commands there, incoming commands will always be routed to that device directly, not taking any gateways into account and not using the `lastKnownGateway` information.
The `lastKnownGateway` information is currently only used, if there was no explicit subscription for the device id of the command message, so that the `via` gateways of the device are checked, using the `lastKnownGateway` information to choose from potentially multiple subscribed gateways.

In that sense there is no need to invoke `lastKnownGateway` with gatewayId = deviceId with respect to the command routing algorithm.

Independent of that, I think the original idea of the `lastKnownGateway` information was just that it should reflect the most up-to-date information about the `device connection`, whether it be a direct connection or via a gateway, in so far as can be deduced from telemetry/event messages received at the adapter.

If the `lastKnownGateway` information continues to not be used for purposes other than command routing, we might as well skip setting it for cases with gatewayId = deviceId. But I think this would probably not have that many performance benefits over all, since a device having non-empty `via` entries and connecting directly to an adapter is probably not the usual case.

core/src/main/java/org/eclipse/hono/util/CommandRouterResult.java
calohmn: Actually, in the client the `CommandRouterResult` isn't really needed. It's only used internally in the `ProtonBasedTenantCommandRouterClient` class and can be replaced with `RequestResponseResult<JsonObject>` there.
So, I've done the replacement and removed the class for now.

core/src/main/java/org/eclipse/hono/util/CommandRouterConstants.java
calohmn: There are 2 use cases of the class:
The client uses the endpoint name and the action names ("set-last-known-gw", etc.) when opening the link/sending messages. (And the endpoint name is used as a bean qualifier.)
The server uses the endpoint name and actions in `DelegatingCommandRouterAmqpEndpoint` to receive/map incoming messages. That class would go into the `service-base` module if sticking to current conventions.

So, if not wanting to duplicate endpoint name and actions, I don't see a different, common place where these things can be put in. The `service-base` module isn't an option because it isn't referenced by `hono-client-adapter-amqp`.

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterClient.java
calohmn: Yes, I've added a note about a ServiceInvocationException getting used then.

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterClient.java
calohmn: Yes, I've added a note about a ServiceInvocationException getting used then.

core/src/main/java/org/eclipse/hono/util/CommandRouterConstants.java
sophokles73: I see. You're right ...

core/src/main/java/org/eclipse/hono/util/CommandRouterResult.java
sophokles73: I did not want to suggest to remove the class altogether but to simply move it to the adapter-amqp module or even make it a private class of ProtonBasedTenantCommandRouterClient. Your choice ...

sophokles73 (approved):
LGTM

PR #2294, opened 2020-11-12
[#2292] Adopt authInfo properties in credentials clientContext

### LINKED ISSUE #2292: Putting MQTT client identifier in credentials API 'get' operation
In the [MQTT adapter user guide "Authentication" chapter](https://www.eclipse.org/hono/docs/user-guide/mqtt-adapter/#authentication) it is stated that the MQTT client identifier is put as client-id in the Credentials API *get* request payload.

Looking at the implementation, this is only true for the client certificate authentication, not for username/password authentication.

Should this feature be added to the username/password authentication? This would mean less credentials request caching if the MQTT client uses a different client id for each connection..

Or should it be removed from the certificate authentication implementation and the MQTT adapter documentation in general? Such a feature only exists for the MQTT adapter and a special device registry implementation would be needed to make use of this.

calohmn:
This fixes #2292.


sophokles73 (approved):
LGTM

sophokles73:
@calohmn can you add this being fixed to the release notes?

calohmn:
@sophokles73 yes, done in c6a88fd3f7667dd02778642aecd517ce37a056d4.

PR #2297, opened 2020-11-13
[#2267] Do not use Spring autowiring for adapters anymore

sophokles73:
The collaborators required by protocol adapters are not injected using
autowiring anymore. This allows for shifting responsibility of object
instantiation/initialization from AbstractProtocolAdapterBase to the
Spring Boot/Quarkus configuration classes which helps with hiding
implementation details from the adapter implementation classes.

### DIFF
--- a/adapters/mqtt-vertx-quarkus/src/main/java/org/eclipse/hono/adapter/mqtt/quarkus/Application.java
+++ b/adapters/mqtt-vertx-quarkus/src/main/java/org/eclipse/hono/adapter/mqtt/quarkus/Application.java
@@ -12,39 +12,26 @@
  */
 package org.eclipse.hono.adapter.mqtt.quarkus;
 
-import java.time.Duration;
-import java.util.List;
-import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.event.Observes;
 import javax.inject.Inject;
 
-import org.eclipse.hono.adapter.client.command.DeviceConnectionClient;
-import org.eclipse.hono.adapter.client.registry.DeviceRegistrationClient;
 import org.eclipse.hono.adapter.mqtt.MessageMapping;
 import org.eclipse.hono.adapter.mqtt.MqttAdapterMetrics;
 import org.eclipse.hono.adapter.mqtt.MqttContext;
 import org.eclipse.hono.adapter.mqtt.MqttProtocolAdapterProperties;
 import org.eclipse.hono.adapter.mqtt.impl.HttpBasedMessageMapping;
 import org.eclipse.hono.adapter.mqtt.impl.VertxBasedMqttProtocolAdapter;
-import org.eclipse.hono.client.CommandTargetMapper;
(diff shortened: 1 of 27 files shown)


adapters/mqtt-vertx-quarkus/src/main/java/org/eclipse/hono/adapter/mqtt/quarkus/Application.java
```diff
@@ -83,16 +96,73 @@ void onStop(final @Observes ShutdownEvent ev) {
...
     private VertxBasedMqttProtocolAdapter adapter() {
 
+        final DeviceRegistrationClient registrationClient = registrationClient();
+        final DeviceConnectionClient deviceConnectionClient = deviceConnectionClient();
+        final CommandTargetMapper commandTargetMapper = commandTargetMapper();
+
+        commandTargetMapper.initialize(new CommandTargetMapperContext() {
+
+            @Override
+            public Future<List<String>> getViaGateways(
```
calohmn, line 106: It seems to me, the content of this method can be reduced to just:
````
        final VertxBasedMqttProtocolAdapter adapter = new VertxBasedMqttProtocolAdapter();
        setCollaborators(adapter);
        adapter.setConfig(adapterProperties);
        adapter.setMetrics(metrics);
        adapter.setMessageMapping(messageMapping());
        return adapter;
````
using the parent `setCollaborators(AbstractProtocolAdapterBase)` method.

Same goes for the corresponding class/method in the `hono-adapter-http-vertx-quarkus` module.

Comment (sophokles73) [reply adapters/mqtt-vertx-quarkus/src/main/java/org/eclipse/hono/adapter/mqtt/quarkus/Application.java:106]: That's true and that is what I wanted to do. Don't know why it ended up this way ...

File: service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java [hunk 2]
```diff
@@ -55,22 +51,15 @@
...
 import org.eclipse.hono.util.QoS;
 import org.eclipse.hono.util.RegistrationAssertion;
-import org.eclipse.hono.util.RegistrationConstants;
 import org.eclipse.hono.util.ResourceIdentifier;
 import org.eclipse.hono.util.Strings;
-import org.eclipse.hono.util.TelemetryConstants;
 import org.eclipse.hono.util.TelemetryExecutionContext;
-import org.eclipse.hono.util.TenantConstants;
 import org.eclipse.hono.util.TenantObject;
-import org.springframework.beans.factory.annotation.Autowired;
```
calohmn, line 72: There are still subclasses of `AbstractProtocolAdapterBase` that use the `@Autowired` annotation (e.g. `VertxBasedMqttProtocolAdapter`), I guess it should be removed there as well.

sophokles73: Good point. I will remove those as well ...

sophokles73: I have pushed changes ...

calohmn (approved):
LGTM

PR #2298, opened 2020-11-13
[#2029] Add Command Router component

calohmn:
This is for #2029:
Adds a new component that implements the new Command Router API and serves as the first-level command message consumer and forwarder.

The Command Router implementation delegates to the corresponding Device Connection API methods, implemented using an Infinispan cache, like it is done in the "hono-service-device-connection" module. 

The command consumer part is adopted from the corresponding implementation in the protocol adapters.
Incoming command messages are analyzed as to which device or gateway and which adapter instance they are targeted at. They are then forwarded on a protocol adapter specific address from where protocol adapters can receive them. The new corresponding consumer logic in the protocol adapters is still to be added, along with usage of the Command Router API there.


service-base/src/main/java/org/eclipse/hono/service/commandrouter/DelegatingCommandRouterAmqpEndpoint.java
```diff
@@ -0,0 +1,239 @@
...
+     * @throws NullPointerException if any of the parameters is {@code null};
+     */
+    public DelegatingCommandRouterAmqpEndpoint(final Vertx vertx, final S service) {
+        super(vertx, service);
+    }
+
+    @Override
+    protected Future<Message> handleRequestMessage(final Message requestMessage, final ResourceIdentifier targetAddress,
+            final SpanContext spanContext) {
+        Objects.requireNonNull(requestMessage);
```
sophokles73, line 62: based on *processSetLastKnownGatewayRequest* we should also check *targetAddress* for `null` or at least handle a `null` address accordingly in *processSetLastKnownGatewayRequest* ...

services/command-router/pom.xml
```diff
@@ -0,0 +1,146 @@
...
+                          </goals>
+                      </execution>
+                  </executions>
+              </plugin>
+          </plugins>
+      </build>
+    </profile>
+  </profiles>
+
+</project>
```
sophokles73, line 146: missing EOL

services/pom.xml
```diff
@@ -29,6 +29,7 @@
     <module>auth</module>
     <module>base-jdbc</module>
     <module>device-connection</module>
+    <module>command-router</module>
```
sophokles73, line 32: so far we have sorted the modules lexicographically ...

services/command-router/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
```diff
@@ -0,0 +1,394 @@
...
+    @Override
+    public void start(final Promise<Void> startPromise) throws Exception {
+
+        if (registrationClient == null) {
+            startPromise.fail(new IllegalStateException("Device Registration client must be set"));
+        } else if (deviceConnectionInfo == null) {
+            startPromise.fail(new IllegalStateException("Device Connection info client must be set"));
+        } else {
+            startServiceClient(registrationClient, "Device Registration service");
+            if (deviceConnectionInfo instanceof ConnectionLifecycle) {
```
sophokles73, line 149: *CacheBasedDeviceConnectionInfo* is no longer an instance of *ConnectionLifecycle* but is an instance of *org.eclipse.hono.util.Lifecylce* instead ...

sophokles73:
My understanding is that once we have the *Command Router* component, the *adaper-amqp* client module's *CommandConsumerFactory* implementation can/will be replaced with a much simpler version which simply connects to the AMQP 1.0 messaging network's endpoint for receiving adapter instance specific command messages, right?

calohmn:
> 
> 
> My understanding is that once we have the _Command Router_ component, the _adaper-amqp_ client module's _CommandConsumerFactory_ implementation can/will be replaced with a much simpler version which simply connects to the AMQP 1.0 messaging network's endpoint for receiving adapter instance specific command messages, right?

Yes. The existing _CommandConsumerFactory_ implementation should still be kept though for compatibility reasons, to be used in case no _command router_ configuration options are configured for an adapter.

sophokles73 (approved):
LGTM

Issue #2288, opened 2020-11-10
HTTP Adapter forwards wrong QoS default level as application property when sending events


fkaltner:

fkaltner:
@sophokles73 May I work on this?

sophokles73:
sure, IMHO the first step would be to add corresponding assertions to the existing integration tests in order to demonstrate the problem. Which version(s) are affected?

fkaltner:
I added a test based on the 1.4.X branch in order to demonstrate the problem

As far as I can see, the bug was introduced with a43f572904420a004f4b21d5c247b66c24a5f023 on 2020-06-29.
So I think 1.2.x, 1.3.x and 1.4.x are affected.

To my understanding 1.5.x should not be affected as there was a code change which fixed this bug.

sophokles73:
Based on our *support policy* we should fix this on branches 1.4.x and 1.3.x. So I guess we should have
* a PR for adding the corresponding integration test (extensions) to master in order to assert this behavior for 1.5.0 and future versions
* a PR for adding the corresponding integration test (extensions) to 1.4.x along with the fix for 1.4.x
* a PR for adding the corresponding integration test (extensions) to 1.3.x along with the fix for 1.3.x

I agree that it is desirable to fix it on 1.3.x and 1.4.x with as little code changes as possible.

sophokles73:
@fkaltner can this be closed?

fkaltner:
Yes, this should be fixed as discussed.

PR #2299, opened 2020-11-16
[#2288] Fixing device level QoS forwarding of events for HTTP adapter.

fkaltner:

### DIFF
--- a/tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpUploadTestBase.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpUploadTestBase.java
@@ -81,8 +81,10 @@ public abstract class AmqpUploadTestBase extends AmqpAdapterTestBase {
         assertAdditionalMessageProperties(ctx, msg);
     }
 
-    private void assertQosLevel(final VertxTestContext ctx, final Message msg, final ProtonQoS expectedQos) {
-        ctx.verify(() -> assertThat(MessageHelper.getQoS(msg)).isEqualTo(expectedQos.ordinal()));
+    private void assertQosLevel(final VertxTestContext ctx, final Message msg) {
+        // AMQP adapter only supports opening links with sender-settle-mode unsettled.
+        // Thus all messages will be sent as AT_LEAST_ONCE.
+        ctx.verify(() -> assertThat(MessageHelper.getQoS(msg)).isEqualTo(ProtonQoS.AT_LEAST_ONCE.ordinal()));
     }
 
     /**
@@ -345,7 +347,7 @@ public abstract class AmqpUploadTestBase extends AmqpAdapterTestBase {
                             msg.getContentType(), MessageHelper.getPayloadAsString(msg));
                 }
                 assertMessageProperties(messageSending, msg);
-                assertQosLevel(messageSending, msg, senderQoS);
+                assertQosLevel(messageSending, msg);
                 callback.handle(null);
             }).map(c -> {
                 consumer = c;
@@ -360,7 +362,6 @@ public abstract class AmqpUploadTestBase extends AmqpAdapterTestBase {
             msg.setAddress(getEndpointName());
             final Promise<?> sendingComplete = Promise.promise();
             final Handler<ProtonSender> sendMsgHandler = replenishedSender -> {
-                replenishedSender.setQoS(senderQoS);
                 replenishedSender.sendQueueDrainHandler(null);
                 switch (senderQoS) {
                 case AT_LEAST_ONCE:
--- a/tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpUploadTestBase.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpUploadTestBase.java
@@ -81,8 +81,8 @@ public abstract class AmqpUploadTestBase extends AmqpAdapterTestBase {
         assertAdditionalMessageProperties(ctx, msg);
     }
 
-    private void assertQosLevel(final VertxTestContext ctx, final Message msg, final ProtonQoS qos) {
-        ctx.verify(() -> assertThat(MessageHelper.getQoS(msg)).isEqualTo(qos.ordinal()));
+    private void assertQosLevel(final VertxTestContext ctx, final Message msg, final ProtonQoS expectedQos) {
+        ctx.verify(() -> assertThat(MessageHelper.getQoS(msg)).isEqualTo(expectedQos.ordinal()));
     }
 
     /**
--- a/tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpUploadTestBase.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpUploadTestBase.java
@@ -81,6 +81,10 @@ public abstract class AmqpUploadTestBase extends AmqpAdapterTestBase {
         assertAdditionalMessageProperties(ctx, msg);
     }
 
+    private void assertQosLevel(final VertxTestContext ctx, final Message msg, final ProtonQoS qos) {
+        ctx.verify(() -> assertThat(MessageHelper.getQoS(msg)).isEqualTo(qos.ordinal()));
+    }
+
     /**
      * Perform additional checks on a received message.
      * <p>
@@ -341,6 +345,7 @@ public abstract class AmqpUploadTestBase extends AmqpAdapterTestBase {
                             msg.getContentType(), MessageHelper.getPayloadAsString(msg));
                 }
                 assertMessageProperties(messageSending, msg);
+                assertQosLevel(messageSending, msg, senderQoS);
                 callback.handle(null);
             }).map(c -> {
                 consumer = c;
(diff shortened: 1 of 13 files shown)


tests/src/test/java/org/eclipse/hono/tests/http/EventHttpIT.java
```diff
@@ -52,4 +63,46 @@ protected void assertAdditionalMessageProperties(final Message msg) {
...
+
+    /**
+     * Checks that device QoS level is ignored for events.
+     *
+     * @param ctx The test context.
+     *
+     * @throws InterruptedException if the test fails.
+     */
+    @Test
+    public void testDeviceQosLevelIsIgnored(final VertxTestContext ctx) throws InterruptedException {
```
sophokles73, line 75: I had imagined to add some assertions to the existing integration tests that we have for uploading telemetry/event messages to adapters. On the consumer side we would simply need to assert that the consumed messages contain the correct value for the *qos* application property, right?

fkaltner: I took a look on how this could possibly be implemented.

In Hono's integration tests base classes there is a method called _assertAdditionalMessageProperties_. As far as I can see this could be the right place for verifying the qos application property. 
For this an additional abstract method, say _getExpectedQoS_, could be defined in the base classes and overridden in the test classes.

For this approach to work each test subclass needs to focus on testing with a defined qos (as the abstract method can only be implemented once per subclass), like done in the MQTT Adapter tests: there is a subclass _TelemetryMqttQoS1IT_ and _TelemetryMqttQoS0IT_.

Consequently I would need to refactor test classes of all adapters so that they follow the pattern in the MQTT to put this to work.

WDYT?

sophokles73: I can also imagine adding an *assertQosLevel(Message, int)* method to the base test classes which is then invoked in addition to the *assertAdditionalMessageProperties* method. We would then pass the expected QoS level into the generic *testUploadMessages* method(s) where the downstream consumer is created. We would then not need to refactor all test classes ...

fkaltner: Turned out I couldn't implemented strictly following either my or your proposal, since the test cases the different adapters are implement following different approaches.

I managed to find something a regard as a minimal as possible change by fitting the approach to the test case at hand.

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpUploadTestBase.java
```diff
@@ -81,6 +81,10 @@ private void assertMessageProperties(final VertxTestContext ctx, final Message m
         assertAdditionalMessageProperties(ctx, msg);
     }
 
+    private void assertQosLevel(final VertxTestContext ctx, final Message msg, final ProtonQoS qos) {
```
sophokles73, line 84: can we rename *qos* to *expectedQos*?

tests/src/test/java/org/eclipse/hono/tests/mqtt/MqttPublishTestBase.java
```diff
@@ -77,6 +85,7 @@
      *         message. The future will succeed if the message has been
      *         published successfully.
      */
+    // here
```
sophokles73, line 88: ???

tests/src/test/java/org/eclipse/hono/tests/coap/CoapTestBase.java
```diff
@@ -922,6 +926,25 @@ private void assertMessageProperties(final VertxTestContext ctx, final Message m
...
+            case CON:
+                return QoS.AT_LEAST_ONCE;
+            case NON:
+                return QoS.AT_MOST_ONCE;
+            default:
+                throw new IllegalArgumentException("Either QoS must be non-null or message type must be CON or NON!");
+        }
+    }
+
+    private void  assertQosLevel(final VertxTestContext ctx, final Message msg, final QoS qos) {
```
sophokles73, line 944: qos -> expectedQos

tests/src/test/java/org/eclipse/hono/tests/coap/TelemetryCoapIT.java
```diff
@@ -80,7 +81,7 @@ protected Type getMessageType() {
      * @throws InterruptedException if the test fails.
      */
     @Test
-    public void testUploadUsingQoS1(final VertxTestContext ctx) throws InterruptedException {
+        public void testUploadUsingQoS1(final VertxTestContext ctx) throws InterruptedException {
```
sophokles73, line 84: indentation?

tests/src/test/java/org/eclipse/hono/tests/http/HttpTestBase.java
```diff
@@ -1241,6 +1265,10 @@ private void assertMessageProperties(final VertxTestContext ctx, final Message m
         });
     }
 
+    private void  assertQosLevel(final VertxTestContext ctx, final Message msg, final QoS qos) {
```
sophokles73, line 1268: qos -> expectedQos

tests/src/test/java/org/eclipse/hono/tests/http/EventHttpIT.java
```diff
@@ -52,4 +63,47 @@ protected void assertAdditionalMessageProperties(final Message msg) {
...
+        assertThat(setup.awaitCompletion(5, TimeUnit.SECONDS)).isTrue();
+        if (setup.failed()) {
+            ctx.failNow(setup.causeOfFailure());
+            return;
+        }
+
+        testUploadMessages(ctx, tenantId,
+                msg -> {
+                    ctx.verify(() -> {
+                        assertThat(msg.getApplicationProperties().getValue().get(MessageHelper.APP_PROPERTY_QOS))
```
sophokles73, line 95: this is not needed anymore, right?

tests/src/test/java/org/eclipse/hono/tests/mqtt/MqttPublishTestBase.java
```diff
@@ -66,6 +67,13 @@
     // <MQTT message ID, PUBACK handler>
     private final Map<Integer, Handler<Integer>> pendingMessages = new HashMap<>();
 
+    /**
+     * Gets the QoS level with which the MQTT message shall be sent.
+     *
+     * @return The QoS level.
+     */
+    protected abstract MqttQoS getExpectedQos();
```
sophokles73, line 75: then it is probably not the *expected* QoS but simply the QoS, right?

tests/src/test/java/org/eclipse/hono/tests/mqtt/MqttPublishTestBase.java
fkaltner: Thanks for spotting that was a "note" I overlooked before committing...

tests/src/test/java/org/eclipse/hono/tests/mqtt/MqttPublishTestBase.java
fkaltner: Actually it is both as the QoS to be expected should be the one with the message was sent. I am fine with renaming though.

tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpUploadTestBase.java
```diff
@@ -355,6 +360,7 @@ private void testUploadMessages(
             msg.setAddress(getEndpointName());
             final Promise<?> sendingComplete = Promise.promise();
             final Handler<ProtonSender> sendMsgHandler = replenishedSender -> {
+                replenishedSender.setQoS(senderQoS);
```
sophokles73, line 363: the QoS of the sender is being set when opening the sender link. Setting this property after the link has been established has no impact.

fkaltner: While it might be true what you are saying, setting this property does have an impact:

If you comment this line the test will actually fail, because the forwarded application QoS property does not match the expected value.

In order to double check I inserted two debug log statements in _VertxBasedAmqpProtocolAdapter_ on line 1112 and 1116 printing out whether the message is sent as AT_LEAST_ONCE or as AT_MOST_ONCE.

When also removing the AT_LEAST_ONCE QoS in _TelemetryAmqpIT_'s _senderQoSTypes()_ method I observe that all messages are sent with AT_LEAST_ONCE QoS in the AMQP protocol adapter's log and also the test fails.

I ran the test with: _mvn verify -Prun-tests,jaeger,useRunningContainers -Ddocker.keepRunning -Dit.test=TelemetryAmqpIT#testUploadMessagesUsingSaslPlain_

Maybe this is a side-effect? Or is it a problem with my machine?

sophokles73: Oh, you are working on the 1.3.x branch. In 1.3.x the AMQP adapter only supports opening link with *sender-settle-mode* `unsettled` (see https://github.com/eclipse/hono/blob/c63b042d2a79b7e8480e1995ce2db73175200159/tests/src/test/java/org/eclipse/hono/tests/amqp/AmqpAdapterTestBase.java#L123). We are therefore not able to send *settled* messages from the test case. I guess it only makes sense to check for the correct QoS level in downstream messages for message being sent with AT_LEAST_ONCE semantics ...
For 1.5.0, we have changed the AMQP adapter to now also accept links with *snd-settle-mode* `settled`, your test should therefore work on master. However, I am afraid that it will also not work on 1.4.x

sophokles73 (commented):
looks good, can you also please add a record to the release notes regarding this issue having been fixed?

sophokles73 (approved):
LGTM

PR #2300, opened 2020-11-17
[#2267] Add AMQP 1.0 based CommandConsumerFactory implementation

sophokles73:
Added transport protocol agnostic CommandConsumerFactory interface to
the adapter client module.
Added AMQP 1.0 based implementation which simply wraps the existing
vertx-proton based "legacy" AdapterProtocolCommandConsumerFactoryImpl.

### DIFF
--- a/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/CacheBasedDeviceConnectionClient.java
+++ b/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/CacheBasedDeviceConnectionClient.java
@@ -39,6 +39,7 @@ import io.vertx.ext.healthchecks.HealthCheckHandler;
  */
 public final class CacheBasedDeviceConnectionClient implements DeviceConnectionClient, ServiceClient {
 
+    private static final String SPAN_NAME_GET_LAST_GATEWAY = "get last known gateway";
     private static final String SPAN_NAME_SET_LAST_GATEWAY = "set last known gateway";
     private static final String SPAN_NAME_GET_CMD_HANDLING_ADAPTER_INSTANCES = "get command handling adapter instances";
     private static final String SPAN_NAME_SET_CMD_HANDLING_ADAPTER_INSTANCE = "set command handling adapter instance";
@@ -80,25 +81,24 @@ public final class CacheBasedDeviceConnectionClient implements DeviceConnectionC
 
     /**
      * {@inheritDoc}
-     * <p>
-     * Simply delegates to {@link #registerCommandConsumer(String, String, String, Duration, SpanContext)}.
      */
     @Override
-    public Future<Void> setCommandHandlingAdapterInstance(
+    public Future<JsonObject> getLastKnownGatewayForDevice(
             final String tenantId,
             final String deviceId,
-            final String adapterInstanceId,
-            final Duration lifespan,
             final SpanContext context) {
 
-        return registerCommandConsumer(tenantId, deviceId, adapterInstanceId, lifespan, context);
+        final Span span = newSpan(context, SPAN_NAME_GET_LAST_GATEWAY);
+        TracingHelper.setDeviceTags(span, tenantId, deviceId);
+
+        return finishSpan(connectionInfoCache.getLastKnownGatewayForDevice(tenantId, deviceId, span), span);
     }
 
     /**
      * {@inheritDoc}
      */
     @Override
-    public Future<Void> registerCommandConsumer(
+    public Future<Void> setCommandHandlingAdapterInstance(
             final String tenantId,
             final String deviceId,
             final String adapterInstanceId,
@@ -116,24 +116,9 @@ public final class CacheBasedDeviceConnectionClient implements DeviceConnectionC
 
     /**
      * {@inheritDoc}
-     * <p>
-     * Simply delegates to {@link #unregisterCommandConsumer(String, String, String, SpanContext)}.
      */
     @Override
     public Future<Void> removeCommandHandlingAdapterInstance(
-            final String tenantId,
-            final String deviceId,
-            final String adapterInstanceId,
-            final SpanContext context) {
-
-        return unregisterCommandConsumer(tenantId, deviceId, adapterInstanceId, context);
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public Future<Void> unregisterCommandConsumer(
             final String tenantId,
             final String deviceId,
             final String adapterInstanceId,
--- a/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/CacheBasedDeviceConnectionClient.java
+++ b/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/CacheBasedDeviceConnectionClient.java
@@ -39,7 +39,6 @@ import io.vertx.ext.healthchecks.HealthCheckHandler;
  */
 public final class CacheBasedDeviceConnectionClient implements DeviceConnectionClient, ServiceClient {
 
-    private static final String SPAN_NAME_GET_LAST_GATEWAY = "get last known gateway";
     private static final String SPAN_NAME_SET_LAST_GATEWAY = "set last known gateway";
     private static final String SPAN_NAME_GET_CMD_HANDLING_ADAPTER_INSTANCES = "get command handling adapter instances";
     private static final String SPAN_NAME_SET_CMD_HANDLING_ADAPTER_INSTANCE = "set command handling adapter instance";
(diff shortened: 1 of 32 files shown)


clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedCommandConsumerFactory.java
```diff
@@ -0,0 +1,247 @@
...
+        super(connection, samplerFactory, adapterConfig);
+
+        Objects.requireNonNull(commandTargetMapper);
+        Objects.requireNonNull(commandRoutingInfoAccess);
+
+        factory = ProtocolAdapterCommandConsumerFactory.create(connection, samplerFactory);
+        factory.initialize(commandTargetMapper, commandRoutingInfoAccess);
+    }
+
+    private static class CommandContextAdapter implements CommandContext {
```
calohmn, line 81: IMHO it would help readability to move this inner class to the bottom of the class.

sophokles73:
@calohmn I have pushed changes ...

calohmn (approved):
Looks ok to me. I haven't fully grasped the full extent of the changes yet, I think it will become clearer when adding the implementation to use the Command Router component. FMPOV this can be merged now, and potential further changes can be discussed when adding the new consumer factory implementation.

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
```diff
@@ -619,71 +624,97 @@ public HonoConnection commandConsumerConnection() {
...
+            return new ProtonBasedCommandConsumerFactory(
+                    commandConsumerConnection(),
+                    samplerFactory,
+                    adapterProperties,
+                    commandTargetMapper,
+                    new CommandHandlingAdapterInfoAccess() {
+
+                        @Override
+                        public Future<Void> setCommandHandlingAdapterInstance(
+                                final String tenant,
```
calohmn, line 678: The creation of the `CommandHandlingAdapterInfoAccess` object here and in the Quarkus `AbstractProtocolAdapterApplication` can be moved to the `ProtonBasedCommandConsumerFactory` constructor, passing along the  `DeviceConnectionClient` object instead of `CommandHandlingAdapterInfoAccess`.

service-base-quarkus/src/main/java/org/eclipse/hono/service/quarkus/AbstractProtocolAdapterApplication.java
```diff
@@ -283,10 +287,22 @@ protected HonoConnection commandConsumerConnection() {
...
      * @return The factory.
      */
-    protected ProtocolAdapterCommandConsumerFactory commandConsumerFactory() {
-        return ProtocolAdapterCommandConsumerFactory.create(commandConsumerConnection());
+    protected CommandConsumerFactory commandConsumerFactory(
+            final CommandTargetMapper commandTargetMapper,
+            final CommandHandlingAdapterInfoAccess commandRoutingInfoAccess) {
+
+        return new ProtonBasedCommandConsumerFactory(
+                HonoConnection.newConnection(vertx, config.command),
```
calohmn, line 301: `commandConsumerConnection()` can be used here instead, it is unused otherwise.

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedDeviceConnectionClient.java
```diff
@@ -122,41 +122,43 @@ protected void onDisconnect() {
...
             final SpanContext context) {
 
-        Objects.requireNonNull(tenant);
+        Objects.requireNonNull(tenantId);
         Objects.requireNonNull(deviceId);
         Objects.requireNonNull(adapterInstanceId);
 
-        return getOrCreateDeviceConnectionClient(tenant)
+        return getOrCreateDeviceConnectionClient(tenantId)
                 .compose(client -> client.setCommandHandlingAdapterInstance(deviceId, adapterInstanceId, lifespan, context));
```
calohmn, line 155: FMPOV, an `UnsupportedOperationException` should be thrown here. The `registerCommandConsumer` method should not be invoked on the DeviceConnectionClient.

Same goes for the `unregisterCommandConsumer` method.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/CacheBasedDeviceConnectionClient.java
```diff
@@ -81,21 +80,25 @@ public CacheBasedDeviceConnectionClient(final DeviceConnectionInfo cache, final
...
-        return finishSpan(connectionInfoCache.getLastKnownGatewayForDevice(tenantId, deviceId, span), span);
+        return registerCommandConsumer(tenantId, deviceId, adapterInstanceId, lifespan, context);
     }
 
+    /**
+     * {@inheritDoc}
+     */
     @Override
-    public Future<Void> setCommandHandlingAdapterInstance(
+    public Future<Void> registerCommandConsumer(
```
calohmn, line 101: FMPOV, an `UnsupportedOperationException` should be thrown here. The `registerCommandConsumer` method should not be invoked on the DeviceConnectionClient.

Same goes for the `unregisterCommandConsumer` method.

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/DeviceConnectionClient.java
```diff
@@ -28,47 +26,7 @@
...
-     *         The future will succeed if a response with status 200 has been received from the device connection service.
-     *         In that case the value of the future will contain a <em>gateway-id</em> property with the
-     *         gateway id.
-     *         <p>
-     *         In case a status other then 200 is received, the future will fail with a
-     *         {@code org.eclipse.hono.client.ServiceInvocationException} containing the (error) status code returned by the service.
-     * @throws NullPointerException if tenant or device id are {@code null}.
-     */
-    Future<JsonObject> getLastKnownGatewayForDevice(String tenant, String deviceId, SpanContext context);
+public interface DeviceConnectionClient extends CommandRouterClient {
```
calohmn, line 29: I'm still not convinced that having the possibility to use just *one* bean (CommandRouterClient) for both the old commandConsumerFactory implementation and the new, CommandRouter-based one is worth the added confusion about the DeviceConnectionClient and CommandRouterClient being mixed up like this.
FMPOV, having a CommandRouterClient bean and a separate DeviceConnectionClient bean and allowing one or the other to be `null` would probably be easier to understand.

At least, when mixing the interfaces, I think it would be good to document the motivation in the DeviceConnectionClient javadoc.

sophokles73: I'll give it another thought ...

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
calohmn: > But it makes the factory (again) dependent on a particular other client which we already saw to be problematic.

As far as I remember, such a case was problematic when creating the factory in a bean method directly, where e.g. the used registration client bean was created via a `@Scope("prototype")` bean definition and there were therefore issues when requiring the very same registration client bean instance in the adapter instance that used the above factory bean.

But, with the `setCollaborators` method, from which the above `commandConsumerFactory()` method is called, we have already decoupled that. I don't see that another layer of decoupling is needed here, seeing that usage of the `DeviceRegistrationClient` is the primary way to create the factory. Therefore I think creating an additional `ProtonBasedCommandConsumerFactory` constructor with a `DeviceRegistrationClient` parameter would be a good way to remove some clutter and duplicate code from the `AbstractAdapterConfig` and `AbstractProtocolAdapterApplication` classes.

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/DeviceConnectionClient.java
sophokles73: In any case we would have two different *CommandConsumerFactory* implementations, right?
1. The *old* one which simply wraps `org.eclipse.hono.client.ProtocolAdapterCommandConsumerFactory` and requires a *DeviceConnectionClient* and *DeviceRegistrationClient* implementation, and 
2- the new one which only requires the *CommandRouterClient*.

Based on whether the DeviceConnectionClient bean or the CommandRouterClient bean has been configured, we would instantiate the former or the latter, right?

calohmn: Yes.

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
sophokles73: Ok, I'll do that.

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/DeviceConnectionClient.java
sophokles73: Ok, I have pushed corresponding changes ...

calohmn (approved):
LGTM

PR #2302, opened 2020-11-17
[#2029] Refactor CommandConsumerFactory in the Command Router

calohmn:
For #2029:
Using `org.eclipse.hono.util.Lifecycle` instead of `ConnectionLifecycle` in `CommandConsumerFactory`.

### DIFF
--- a/services/command-router/src/main/java/org/eclipse/hono/commandrouter/CommandConsumerFactory.java
+++ b/services/command-router/src/main/java/org/eclipse/hono/commandrouter/CommandConsumerFactory.java
@@ -13,49 +13,20 @@
 
 package org.eclipse.hono.commandrouter;
 
-import java.time.Duration;
-
 import org.eclipse.hono.client.CommandTargetMapper;
-import org.eclipse.hono.client.ConnectionLifecycle;
-import org.eclipse.hono.client.HonoConnection;
-import org.eclipse.hono.client.SendMessageSampler;
 import org.eclipse.hono.client.ServiceInvocationException;
-import org.eclipse.hono.commandrouter.impl.amqp.ProtonBasedCommandConsumerFactoryImpl;
+import org.eclipse.hono.util.Lifecycle;
 
 import io.opentracing.SpanContext;
 import io.vertx.core.Future;
 
 /**
- * A factory for creating clients for the <em>AMQP 1.0 Messaging Network</em> to
- * receive commands and send responses.
+ * A factory for creating consumers of command &amp; control messages.
  */
-public interface CommandConsumerFactory extends ConnectionLifecycle<HonoConnection> {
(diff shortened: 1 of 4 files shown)


services/command-router/src/main/java/org/eclipse/hono/commandrouter/CommandConsumerFactory.java
```diff
@@ -13,49 +13,20 @@
...
-     * Creates a new factory for an existing connection.
-     *
-     * @param connection The connection to the AMQP network.
-     * @return The factory.
-     * @throws NullPointerException if connection or gatewayMapper is {@code null}.
-     */
-    static CommandConsumerFactory create(final HonoConnection connection) {
-        return create(connection, SendMessageSampler.Factory.noop());
-    }
+public interface CommandConsumerFactory extends Lifecycle {
```
calohmn, line 26: Regarding the interface name and the comment that it might get confused with `org.eclipse.hono.client.adapter.command.CommandConsumerFactory`:
I thought about naming it `CommandRouterCommandConsumerFactory` instead, but then the name for the implementation class would get even longer still. 
As the Command Router module isn't added as dependency in any other module, the confusion could only arise in the Command Router module itself. But I don't think it's much of an issue there.

sophokles73: ok

services/command-router/src/main/java/org/eclipse/hono/commandrouter/CommandConsumerFactory.java
```diff
@@ -13,49 +13,20 @@
...
-     * @return The factory.
-     * @throws NullPointerException if connection or gatewayMapper is {@code null}.
-     */
-    static CommandConsumerFactory create(final HonoConnection connection, final SendMessageSampler.Factory samplerFactory) {
-        return new ProtonBasedCommandConsumerFactoryImpl(connection, samplerFactory);
-    }
-
-    /**
-     * Initializes the ProtocolAdapterCommandConsumerFactory with the given components.
+     * Initializes the CommandConsumerFactory with the given commandTargetMapper.
```
sophokles73, line 29: I do not really see the value of the initialize method. It only seems to make implementation harder. Why don't we simply pass in the mapper in the implementation class's constructor? The factory interface itself doesn't require the mapper anyway ...

calohmn: That will be a bigger change. I'd rather do this in a separate commit/PR.

sophokles73: ok

sophokles73 (approved):
LGTM

Issue #2273, opened 2020-10-28
Support using Kafka for Command & Control messages


calohmn:
For a Hono instance that uses Kafka as messaging infrastructure for sending telemetry/event messages (see the [corresponding proposal](https://github.com/eclipse/hono/issues/8#issuecomment-683628747)), it makes sense to provide Kafka support also for Command & Control messages. 

This would mean northbound applications don't have to use both a Kafka API (for telemetry/events) and an AMQP-based API (for commands) to interact with Hono. It also means removing the necessity to use an AMQP messaging network _in addition_ to a Kafka instance, if Kafka is to be used for telemetry/event messages.

calohmn:
General notes and possible approaches:

### Topics to be used by northbound applications

The general idea is for northbound applications to publish command messages on a Kafka topic and to receive command response messages by subscribing to a Kafka topic. In order for applications to only be able to publish/receive command (response) messages for their specific tenant, the topic names should contain the tenant identifier, so that corresponding ACLs can be put in place in Kafka. This could mean using **`command-$tenantId`** as topic for sending commands and **`command_response-$tenantId`** for receiving command responses. For these topics, the device identifier would be used as partition key to ensure correct ordering of messages related to one specific device.

### Command acknowledgements

For the AMQP-based Command & Control API, the outcome of sending a command message was indicated back to the sending application by means of AMQP delivery states. As when using Kafka, there is no such protocol-provided acknowledgement mechanism, the question is whether or how to support acknowledgements then.

Either:

- Don't support (technical-level) command acknowledgements: applications will no longer know whether commands were actually forwarded to or even acknowledged by the target device (depending on device protocol). Applications will then have to use application-level mechanisms to find out, whether a command has reached the device. Usually that would mean requiring a command response message from the device. As for the alternatives for implementing the command message routing listed below, this option would allow command messages to be directed to multiple (potentially matching) protocol adapter instances, letting each instance decide whether to handle the command message.

- Or mimic/implement technical-level command acknowledgements by letting the protocol adapters publish corresponding messages on a topic like `command_ack-$tenantId`, which the northbound application will have to subscribe to. This option will reduce the alternatives for implementing the command message routing listed below, as it has to be made sure that the command message is only directed to _one_ specific protocol adapter instance that the device/gateway is connected to.

## Alternatives for routing command messages

### 1. Receiving commands on all protocol adapters (that have subscriptions for command tenant)

[...]

PR #2303, opened 2020-11-17
Add Kafka based downstream client

b-abel:
Replaces #2216. 

* I have not tested it yet with the Quarkus based adapters.
* Only tested manually, integration tests are still missing.
* I have added new Spring `Condition`s to enable a clean control of the selection between AMQP and Kafka. @sophokles73, now I have seen that you have taken steps in `AbstractAdapterConfig` to move away from Spring Boot. What is your strategy there? Do you want to remove the bean creation by Spring Boot entirely?

### DIFF
--- a/adapters/pom.xml
+++ b/adapters/pom.xml
@@ -74,6 +74,10 @@
       <groupId>io.opentracing</groupId>
       <artifactId>opentracing-util</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.wildfly.security</groupId>
+      <artifactId>wildfly-elytron-password-impl</artifactId>
+    </dependency>
 
     <!-- Testing -->
     <dependency>
--- a/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/telemetry/kafka/AbstractKafkaBasedDownstreamSender.java
+++ b/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/telemetry/kafka/AbstractKafkaBasedDownstreamSender.java
@@ -1,0 +1,329 @@
+/*
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.hono.adapter.client.telemetry.kafka;
+
+import java.net.HttpURLConnection;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+import org.apache.kafka.clients.producer.ProducerConfig;
+import org.eclipse.hono.client.ServerErrorException;
+import org.eclipse.hono.config.ProtocolAdapterProperties;
+import org.eclipse.hono.kafka.client.HonoTopic;
+import org.eclipse.hono.kafka.client.KafkaProducerFactory;
+import org.eclipse.hono.kafka.client.tracing.KafkaTracingHelper;
+import org.eclipse.hono.tracing.TracingHelper;
+import org.eclipse.hono.util.Lifecycle;
+import org.eclipse.hono.util.MessageHelper;
+import org.eclipse.hono.util.QoS;
+import org.eclipse.hono.util.RegistrationAssertion;
+import org.eclipse.hono.util.TenantObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.opentracing.References;
+import io.opentracing.Span;
+import io.opentracing.SpanContext;
+import io.opentracing.Tracer;
+import io.opentracing.tag.Tags;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.json.EncodeException;
+import io.vertx.core.json.Json;
+import io.vertx.kafka.client.producer.KafkaHeader;
+import io.vertx.kafka.client.producer.KafkaProducer;
+import io.vertx.kafka.client.producer.KafkaProducerRecord;
+import io.vertx.kafka.client.producer.RecordMetadata;
+import io.vertx.kafka.client.producer.impl.KafkaHeaderImpl;
+
+/**
+ * A client for publishing messages to a Kafka cluster.
+ */
+public abstract class AbstractKafkaBasedDownstreamSender implements Lifecycle {
+
+    /**
+     * A logger to be shared with subclasses.
+     */
+    protected final Logger log = LoggerFactory.getLogger(getClass());
+
+    private final KafkaProducerFactory<String, Buffer> producerFactory;
+    private final String producerName;
+    private final Map<String, String> config;
+    private final ProtocolAdapterProperties adapterConfig;
+    private final Tracer tracer;
+
+    /**
+     * Creates a new Kafka-based telemetry sender.
+     *
+     * @param producerFactory The factory to use for creating Kafka producers.
+     * @param producerName The producer name to use.
+     * @param config The Kafka producer configuration properties to use.
+     * @param adapterConfig The protocol adapter's configuration properties.
+     * @param tracer The OpenTracing tracer.
+     * @throws NullPointerException if any of the parameters are {@code null}.
+     */
+
+    public AbstractKafkaBasedDownstreamSender(final KafkaProducerFactory<String, Buffer> producerFactory,
+            final String producerName, final Map<String, String> config, final ProtocolAdapterProperties adapterConfig,
+            final Tracer tracer) {
+
+        Objects.requireNonNull(producerFactory);
+        Objects.requireNonNull(producerName);
+        Objects.requireNonNull(config);
+        Objects.requireNonNull(adapterConfig);
+        Objects.requireNonNull(tracer);
+
+        this.producerFactory = producerFactory;
+        this.producerName = producerName;
+        this.config = config;
+        this.adapterConfig = adapterConfig;
+        this.tracer = tracer;
+    }
+
+    /**
+     * Sends a message downstream.
+     * <p>
+     * Default properties defined either at the device or tenant level are added to the message headers.
+     *
+     * @param topic The topic to send the message to.
+     * @param tenant The tenant that the device belongs to.
+     * @param device The registration assertion for the device that the data originates from.
+     * @param qos The delivery semantics to use for sending the data.
+     * @param contentType The content type of the data. If {@code null}, the content type be taken from the following
+     *            sources (in that order, the first one that is present is used):
+     *            <ol>
+     *            <li>the <em>contentType</em> parameter</li>
+     *            <li>the property with key {@link org.eclipse.hono.util.MessageHelper#SYS_PROPERTY_CONTENT_TYPE} in the
+     *            <em>properties</em> parameter</li>
+     *            <li>the device default</li>
+     *            <li>the tenant default</li>
+     *            <li>the {@linkplain org.eclipse.hono.util.MessageHelper#CONTENT_TYPE_OCTET_STREAM default content
+     *            type}</li>
+     *            </ol>
+     * @param payload The data to send.
+     * @param properties Additional meta data that should be included in the downstream message.
+     * @param context The currently active OpenTracing span (may be {@code null}). An implementation should use this as
+     *            the parent for any span it creates for tracing the execution of this operation.
+     * @return A future indicating the outcome of the operation.
+     *         <p>
+     *         The future will be succeeded if the message has been sent downstream.
+     *         <p>
+     *         The future will be failed with a {@link org.eclipse.hono.client.ServerErrorException} if the data could
+     *         not be sent. The error code contained in the exception indicates the cause of the failure.
+     * @throws NullPointerException if topic, tenant, device, or qos are {@code null}.
+     */
+    protected Future<Void> send(final HonoTopic topic, final TenantObject tenant, final RegistrationAssertion device,
+            final QoS qos, final String contentType, final Buffer payload, final Map<String, Object> properties,
+            final SpanContext context) {
+
+        Objects.requireNonNull(topic);
+        Objects.requireNonNull(tenant);
+        Objects.requireNonNull(device);
+        Objects.requireNonNull(qos);
+
+        final String tenantId = tenant.getTenantId();
+        final String deviceId = device.getDeviceId();
+
+        log.trace("sending to Kafka [topic: {}, tenantId: {}, deviceId: {}, qos: {}, contentType: {}, properties: {}]",
+                topic, tenantId, deviceId, qos, contentType, properties);
+        final Span span = startSpan(topic, tenantId, deviceId, qos, contentType, context);
+
+        final KafkaProducerRecord<String, Buffer> record = KafkaProducerRecord.create(topic.toString(), deviceId,
+                payload);
+
+        final Map<String, Object> propsWithDefaults = addDefaults(tenant, device, qos, contentType, properties);
+        record.addHeaders(encodePropertiesAsKafkaHeaders(propsWithDefaults, span));
+
+        KafkaTracingHelper.injectSpanContext(tracer, record, span.context());
+        logProducerRecord(span, record);
+
+        final Promise<RecordMetadata> promise = Promise.promise();
+        getOrCreateProducer().send(record, promise);
+
+        final Future<Void> producerFuture = promise.future()
+                .recover(t -> {
+                    logError(span, topic, tenantId, deviceId, qos, t);
+                    span.finish();
+                    return Future.failedFuture(new ServerErrorException(getErrorCode(t), t));
+                })
+                .map(recordMetadata -> {
+                    logRecordMetadata(span, deviceId, recordMetadata);
+                    span.finish();
+                    return null;
+                });
+
+        return qos.equals(QoS.AT_MOST_ONCE) ? Future.succeededFuture() : producerFuture;
+    }
+
+    /**
+     * {@inheritDoc}
+     * <p>
+     * Starts the producer.
+     */
+    @Override
+    public Future<Void> start() {
+        configureUniqueClientId();
+        getOrCreateProducer();
+        return Future.succeededFuture();
+    }
+
+    /**
+     * {@inheritDoc}
+     * <p>
+     * Closes the producer.
+     */
+    @Override
+    public Future<Void> stop() {
+        return producerFactory.closeProducer(producerName);
+    }
+
+    private KafkaProducer<String, Buffer> getOrCreateProducer() {
+        return producerFactory.getOrCreateProducer(producerName, config);
+    }
+
+    private Map<String, Object> addDefaults(final TenantObject tenant, final RegistrationAssertion device,
+            final QoS qos, final String contentType, final Map<String, Object> properties) {
+
+        final Map<String, Object> headerProperties = new HashMap<>();
+        if (adapterConfig.isDefaultsEnabled()) {
+            headerProperties.putAll(tenant.getDefaults().copy().getMap()); // (1) add tenant defaults
+            headerProperties.putAll(device.getDefaults()); // (2) overwrite with device defaults
+        }
+
+        // (3) overwrite with properties provided by protocol adapter
+        Optional.ofNullable(properties).ifPresent(headerProperties::putAll);
+
+        // (4) overwrite by values of separate parameters
+        headerProperties.put(MessageHelper.APP_PROPERTY_DEVICE_ID, device.getDeviceId());
+        headerProperties.put(MessageHelper.APP_PROPERTY_QOS, qos.ordinal());
+        if (contentType != null) {
+            headerProperties.put(MessageHelper.SYS_PROPERTY_CONTENT_TYPE, contentType);
+        }
+
+        // (5) if still no content type present, set the default content type
+        headerProperties.putIfAbsent(MessageHelper.SYS_PROPERTY_CONTENT_TYPE, MessageHelper.CONTENT_TYPE_OCTET_STREAM);
+
+        if (headerProperties.containsKey(MessageHelper.APP_PROPERTY_DEVICE_TTD)
+                && !headerProperties.containsKey(MessageHelper.SYS_PROPERTY_CREATION_TIME)) {
+            // TODO set this as creation time in the KafkaRecord?
+
+            // must match http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-types-v1.0-os.html#type-timestamp
+            // as defined in https://www.eclipse.org/hono/docs/api/telemetry/#forward-telemetry-data
+            final long timestamp = Instant.now().toEpochMilli();
+            headerProperties.put(MessageHelper.SYS_PROPERTY_CREATION_TIME, timestamp);
+        }
+
+        return headerProperties;
+    }
+
+    private List<KafkaHeader> encodePropertiesAsKafkaHeaders(final Map<String, Object> properties, final Span span) {
+        final List<KafkaHeader> headers = new ArrayList<>();
+        properties.forEach((k, v) -> {
+            try {
+                final Buffer headerValue = (v instanceof String)
+                        ? Buffer.buffer((String) v)
+                        : Buffer.buffer(Json.encode(v));
+
+                headers.add(new KafkaHeaderImpl(k, headerValue));
+            } catch (final EncodeException e) {
+                log.info("failed to serialize property with key [{}] to Kafka header", k);
+                span.log("failed to create Kafka header from property: " + k);
+            }
+        });
+
+        return headers;
+    }
+
+    private Span startSpan(final HonoTopic topic, final String tenantId, final String deviceId, final QoS qos,
+            final String contentType, final SpanContext context) {
+
+        final String referenceType = QoS.AT_MOST_ONCE.equals(qos) ? References.FOLLOWS_FROM : References.CHILD_OF;
+        return KafkaTracingHelper.newProducerSpan(tracer, topic, referenceType, context)
+                .setTag(TracingHelper.TAG_TENANT_ID.getKey(), tenantId)
+                .setTag(TracingHelper.TAG_DEVICE_ID.getKey(), deviceId)
+                .setTag(TracingHelper.TAG_QOS.getKey(), qos.name())
+                .setTag(MessageHelper.SYS_PROPERTY_CONTENT_TYPE, contentType);
+    }
+
+    private void logProducerRecord(final Span span, final KafkaProducerRecord<String, Buffer> record) {
+        final String headersAsString = record.headers()
+                .stream()
+                .map(header -> header.key() + "=" + header.value())
+                .collect(Collectors.joining(",", "{", "}"));
+
+        log.trace("producing message [topic: {}, key: {}, partition: {}, timestamp: {}, headers: {}]",
+                record.topic(), record.key(), record.partition(), record.timestamp(), headersAsString);
+
+        span.log("producing message with headers: " + headersAsString);
+    }
+
+    private void logRecordMetadata(final Span span, final String recordKey, final RecordMetadata metadata) {
+
+        log.trace("message produced to Kafka [topic: {}, key: {}, partition: {}, offset: {}, timestamp: {}]",
+                metadata.getTopic(), recordKey, metadata.getPartition(), metadata.getOffset(), metadata.getTimestamp());
+
+        span.log("message produced to Kafka");
+        KafkaTracingHelper.setRecordMetadataTags(span, metadata);
+        Tags.HTTP_STATUS.set(span, HttpURLConnection.HTTP_ACCEPTED);
+
+    }
+
+    private void logError(final Span span, final HonoTopic topic, final String tenantId, final String deviceId,
+            final QoS qos, final Throwable cause) {
+        log.debug("sending message failed [topic: {}, key: {}, qos: {}, tenantId: {}, deviceId: {}]",
+                topic, deviceId, qos, tenantId, deviceId, cause);
+
+        Tags.HTTP_STATUS.set(span, getErrorCode(cause));
+        TracingHelper.logError(span, cause);
+    }
+
+    private int getErrorCode(final Throwable t) {
+        /*
+         * TODO set error code depending on exception?
+         *
+         * Possible thrown exceptions include:
+         *
+         * Non-Retriable exceptions (fatal, the message will never be sent):
+         *
+         * InvalidTopicException OffsetMetadataTooLargeException RecordBatchTooLargeException RecordTooLargeException
+         * UnknownServerException UnknownProducerIdException
+         *
+         * Retriable exceptions (transient, may be covered by increasing #.retries):
+         *
+         * CorruptRecordException InvalidMetadataException NotEnoughReplicasAfterAppendException
+         * NotEnoughReplicasException OffsetOutOfRangeException TimeoutException UnknownTopicOrPartitionException
+         */
+
+        return HttpURLConnection.HTTP_UNAVAILABLE;
+    }
+
+    private void configureUniqueClientId() {
+        final UUID uuid = UUID.randomUUID();
+
+        final String clientId = config.get(ProducerConfig.CLIENT_ID_CONFIG);
+        if (clientId == null) {
+            config.put(ProducerConfig.CLIENT_ID_CONFIG, String.format("%s-%s", producerName, uuid));
+        } else {
+            config.put(ProducerConfig.CLIENT_ID_CONFIG, String.format("%s-%s-%s", clientId, producerName, uuid));
+        }
+    }
+}
--- a/clients/adapter-kafka/src/test/java/org/eclipse/hono/adapter/client/telemetry/kafka/TestHelper.java
+++ b/clients/adapter-kafka/src/test/java/org/eclipse/hono/adapter/client/telemetry/kafka/TestHelper.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.hono.adapter.client.telemetry.kafka;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import org.apache.kafka.clients.producer.MockProducer;
+import org.apache.kafka.clients.producer.Producer;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.clients.producer.RecordMetadata;
+import org.apache.kafka.common.header.internals.RecordHeader;
+import org.apache.kafka.common.serialization.StringSerializer;
+import org.eclipse.hono.kafka.client.CachingKafkaProducerFactory;
+import org.eclipse.hono.test.VertxMockSupport;
+import org.eclipse.hono.util.QoS;
+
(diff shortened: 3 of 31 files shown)


clients/kafka-common/pom.xml
```diff
@@ -0,0 +1,143 @@
...
+    </parent>
+    <artifactId>hono-client-kafka-common</artifactId>
+
+    <name>Hono Client Kafka Common</name>
+    <description>Classes required for implementing Kafka based Hono clients</description>
+
+    <properties>
+        <maven.compiler.source>1.8</maven.compiler.source>
+        <maven.compiler.target>1.8</maven.compiler.target>
+    </properties>
```
sophokles73, line 30: why do we need to keep this at Java 1.8 level?

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/HonoTopic.java
```diff
@@ -0,0 +1,135 @@
...
+        return tenantId;
+    }
+
+    /**
+     * Returns the string representation of the topic as used by the Kafka client.
+     *
+     * @return The topic as a string.
+     */
+    @Override
+    public String toString() {
```
sophokles73, line 91: this seems to be the most often used method, isn't it? Given that this class is immutable, we should probably create the string representation once during construction instead of concatenating the strings over and over again ...

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/test/FakeProducer.java
```diff
@@ -0,0 +1,306 @@
...
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.hono.kafka.client.test;
```
sophokles73, line 14: do we really need this in the main source tree or can this go into the test folder?

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/test/TestHelper.java
```diff
@@ -0,0 +1,69 @@
...
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.hono.kafka.client.test;
```
sophokles73, line 14: test folder?

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/tracing/KafkaTracingHelper.java
```diff
@@ -0,0 +1,134 @@
...
+     * <li>{@link #TAG_PARTITION}</li>
+     * <li>{@link #TAG_TIMESTAMP}</li>
+     * </ul>
+     * <p>
+     * <em>It does not set the topic, as this is expected to be already already set.</em>
+     *
+     * @param span The span to set the tags on.
+     * @param recordMetadata The record metadata.
+     */
+    public static void setRecordMetadataTags(final Span span, final RecordMetadata recordMetadata) {
```
sophokles73, line 106: IMHO these values should be logged to a span but should not be set as tags ... I doubt that you will filter/search spans based on these values, will you?

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/tracing/LongTag.java
```diff
@@ -0,0 +1,37 @@
...
+
+package org.eclipse.hono.kafka.client.tracing;
+
+import io.opentracing.Span;
+import io.opentracing.tag.AbstractTag;
+
+/**
+ * An OpenTracing tag type for {@code long} values.
+ */
+public class LongTag extends AbstractTag<Long> {
```
sophokles73, line 22: this doesn't seem to be kafka specific, is it?

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
```diff
@@ -248,11 +281,24 @@ protected ClientConfigProperties getDownstreamSenderFactoryConfigDefaults() {
...
+     *
+     * @return The factory.
+     */
+    @Bean
+    @Scope("prototype")
+    @ConditionalOnKafkaMessaging
+    public CachingKafkaProducerFactory<String, Buffer> kafkaProducerFactory() {
+        return CachingKafkaProducerFactory.sharedProducerFactory(vertx());
+    }
+
```
sophokles73, line 301: can we keep the AMQP related beans together?

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
```diff
@@ -266,13 +312,30 @@ public HonoConnection downstreamConnection() {
      */
     @Qualifier(TelemetryConstants.TELEMETRY_ENDPOINT)
     @Bean
+    @ConditionalOnAmqpMessaging
```
sophokles73, line 315: IMHO we do not need the conditional annotations because we only instantiate the beans directly by invoking the methods explicitly ...

clients/kafka-common/pom.xml
b-abel: FMPOV we should add a client for northbound applications as well. The current one is at Java 1.8, so I just kept it that way.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/HonoTopic.java
b-abel: Done.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/test/FakeProducer.java
b-abel: FMPOV this adds a missing _feature_ in the Vert.x Kafka client. The (Apache) Kafka client ships with a great class `MockProducer`, which cannot be used in Vert.x's client.
I see it as a _fake_ in the sense of https://martinfowler.com/bliki/TestDouble.html. Although it is used in test scenarios, it is a working implementation, which itself needs to be tested. Also, in practice, if I move it to the test folder, I cannot use it in other modules.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/test/TestHelper.java
b-abel: Done.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/tracing/KafkaTracingHelper.java
b-abel: I'm not sure. I aimed for compatibility with https://github.com/opentracing-contrib/java-kafka-client which does set _partition_ and _offset_ as tags. 
I just learned that Vert.x 4 will soon be released with support for OpenTracing. We may want Hono to take advantage of those APIs. So maybe this implementation should be aligned with the tracing that Vert.x's Kafka client will produce, which seems to be this https://github.com/vert-x3/vertx-kafka-client/blob/master/src/main/java/io/vertx/kafka/client/common/tracing/TraceTags.java
WDYT?

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
b-abel: Do you want to remove Spring Boot? Otherwise, we could refactor `ProtonBasedDownstreamSender` into two classes, one implementing `TelemetrySender` and one implementing `EventSender` and then look up the beans from the ApplicationContext, like you do it with the DeviceConnectionClient. This would remove the ugly if-else block from _setCollaborators()_ and instead take advantage of conditional bean instantiation based on config properties (which are used by the conditional annotations).

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
b-abel: Done.

b-abel:
To try out this implementation please follow the instructions in https://github.com/eclipse/hono/pull/2216#issuecomment-730255279.

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
b-abel: Another option for the conditional bean creation would be to use _profiles_. This is probably the more common way to handle it with Spring Boot. Quarkus has a concept of profiles as well: https://quarkus.io/guides/config#configuration-profiles

sophokles73:
@b-abel can you please rebase this PR?
regarding Spring Boot: we stopped using autowiring for setting the adapters' collaborators and instead now manually create and configure protocol adapter instances. As long as it does not interfere with that approach, we can still take advantage of conditional creation of beans ...

b-abel:
> @b-abel can you please rebase this PR?

Done.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/test/FakeProducer.java
b-abel: Maybe this could go into the "test-utils" module. WDYT?

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/CachingKafkaProducerFactory.java
```diff
@@ -0,0 +1,171 @@
...
+    /**
+     * Creates an instance of the factory which produces {@link FakeProducer}s.
+     * <p>
+     * This is intended for tests only.
+     *
+     * @param <K> The type for the record key serialization.
+     * @param <V> The type for the record value serialization.
+     * @return An instance of the factory.
+     */
+    public static <K, V> CachingKafkaProducerFactory<K, V> testProducerFactory() {
```
sophokles73, line 85: If we want to be able to create fake producers then I guess we should have a *KafkaProducerFactory* interface which can be mocked for test cases ...

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/KafkaConsumerConfigProperties.java
```diff
@@ -0,0 +1,113 @@
...
+     * <ul>
+     * <li>{@code key.deserializer=org.apache.kafka.common.serialization.StringDeserializer}: defines how message keys
+     * are deserialized</li>
+     * <li>{@code value.deserializer=io.vertx.kafka.client.serialization.BufferDeserializer}: defines how message values
+     * are deserialized</li>
+     * <li>{@code client.id} if the property is not already present in the configuration and a value has been set with
+     * {@link #setClientId(String)}, this value will be taken</li>
+     * </ul>
+     *
+     * @return a copy of the consumer configuration with the applied properties or {@code null} if no consumer
```
sophokles73, line 77: FMPOV this method should never return `null` ...

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/KafkaProducerConfigProperties.java
```diff
@@ -0,0 +1,119 @@
...
+     * <li>{@code key.serializer=org.apache.kafka.common.serialization.StringSerializer}: defines how message keys are
+     * serialized</li>
+     * <li>{@code value.serializer=io.vertx.kafka.client.serialization.BufferSerializer}: defines how message values are
+     * serialized</li>
+     *
+     * <li>{@code client.id} if the property is not already present in the configuration and a value has been set with
+     * {@link #setClientId(String)}, this value will be taken</li>
+     * </ul>
+     *
+     * @return a copy of the producer configuration with the applied properties or {@code null} if no producer
```
sophokles73, line 79: same here

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/tracing/KafkaHeaderInjectAdapter.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public Iterator<Entry<String, String>> iterator() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void put(final String key, final String value) {
+        headers.add(new KafkaHeaderImpl(key, value));
```
sophokles73, line 48: IMHO this should use `KafkaHeader.header(key, value)` instead ...

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/tracing/KafkaHeaderInjectAdapter.java
```diff
@@ -0,0 +1,50 @@
...
+
+import io.opentracing.propagation.TextMap;
+import io.vertx.kafka.client.producer.KafkaHeader;
+import io.vertx.kafka.client.producer.impl.KafkaHeaderImpl;
+
+/**
+ * An adapter for injecting properties as a new {@link KafkaHeader} to a list of Vert.x Kafka producer headers.
+ *
+ */
+public class KafkaHeaderInjectAdapter implements TextMap {
```
sophokles73, line 28: final ?

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/test/FakeProducer.java
sophokles73: Is this class required by other modules than the adapter-kafka module?

site/documentation/content/admin-guide/hono-kafka-client-configuration.md
```diff
@@ -0,0 +1,43 @@
...
+The configuration needs to be provided in the form `hono.kafka.producerConfig.${property}`, where `${property}` is any of the 
+Kafka client's [producer properties](https://kafka.apache.org/documentation/#producerconfigs). 
+The provided configuration is passed directly to the Kafka producer without Hono parsing or validating it.
+The following table shows which properties _are_ changed by Hono.
+
+| Kafka Producer Config Property | Mandatory Value | Description |
+| :----------------------------- | :-------------- | :-----------|
+| `--hono.kafka.producerConfig.key.serializer` | `org.apache.kafka.common.serialization.StringSerializer` | The record keys in Hono are always strings. Any other specified value is ignored. |
+| `--hono.kafka.producerConfig.value.serializer` | `io.vertx.kafka.client.serialization.BufferSerializer` | The record values in Hono are always byte arrays.  Any other specified value is ignored. |
+| `--hono.kafka.producerConfig.enable.idempotence` | `true` | The Hono Kafka client uses only idempotent producers.  Any other specified value is ignored. |
```
sophokles73, line 24: can you also add the environment variable names?

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
```diff
@@ -134,14 +141,24 @@ protected void setCollaborators(
             adapter.setCommandConsumerFactory(commandConsumerFactory(adapterProperties, samplerFactory, commandRouterClient));
         }
 
+        final KafkaProducerConfigProperties kafkaProducerConfig = kafkaProducerConfig();
```
sophokles73, line 144: does this work (i.e. doesn't throw a runtime exception) if no `hono.kafka` properties are set at all?

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/CachingKafkaProducerFactory.java
```diff
@@ -0,0 +1,171 @@
...
+ * <p>
+ * Created producers are being cached.
+ * <p>
+ * Producers are closed and removed from the cache if they throw a {@link #isFatalError(Throwable) fatal exception}. A
+ * following invocation of {@link #getOrCreateProducer(String, Map)} will then return a new instance.
+ *
+ * @param <K> The type for the record key serialization.
+ * @param <V> The type for the record value serialization.
+ */
+public class CachingKafkaProducerFactory<K, V> {
```
kaniyan, line 46: AFAIK `CachingClientFactory` is quite generic and I think it can also accommodate Kafka clients and in this case `KafkaProducer`? Then we could reuse the code and IMHO having a uniform style throughout, makes the code easier to understand. I am just curious to know if there are any reasons to have a new  `CachingKafkaProducerFactory`.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/test/FakeProducer.java
```diff
@@ -0,0 +1,306 @@
...
+                        exceptionHandler.handle(err);
+                    }
+                    if (handler != null) {
+                        handler.handle(Future.failedFuture(err));
+                    }
+                } else if (handler != null) {
+                    handler.handle(Future.succeededFuture(Helper.from(metadata)));
+                }
+            });
+        } catch (Exception e) {
```
kaniyan, line 216: How about `final Exception e`?

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
```diff
@@ -134,14 +141,24 @@ protected void setCollaborators(
...
+            // look up via bean factory is not possible because EventSender and TelemetrySender are implemented by
+            // ProtonBasedDownstreamSender
+            adapter.setEventSender(downstreamEventSender(samplerFactory, adapterProperties));
+            adapter.setTelemetrySender(downstreamTelemetrySender(samplerFactory, adapterProperties));
+        } else {
+            final CachingKafkaProducerFactory<String, Buffer> kafkaProducerFactory = kafkaProducerFactory();
+            adapter.setEventSender(downstreamEventKafkaSender(kafkaProducerFactory, kafkaProducerConfig));
+            adapter.setTelemetrySender(downstreamTelemetryKafkaSender(kafkaProducerFactory, kafkaProducerConfig));
+        }
+
```
b-abel, line 155: I added #2326. This simplifies this part to the following, by leveraging conditional bean creation:  
```java
// look up client via bean factory in order to take advantage of conditional bean instantiation based
// on config properties
adapter.setEventSender(context.getBean(EventSender.class));
adapter.setTelemetrySender(context.getBean(TelemetrySender.class));
```

Additionally, I can then change the class `KafkaProducerConfig` to no longer return `null`.

@sophokles73 WDYT?

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
b-abel: Yes. That is the case when `KafkaProducerConfigProperties.getProducerConfig()` returns `null` (see discussion above).

But I would prefer to use the change in #2326 which would make this construct obsolete.

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
b-abel: > Otherwise, we could refactor `ProtonBasedDownstreamSender` into two classes, one implementing `TelemetrySender` and one implementing `EventSender` and then look up the beans from the ApplicationContext, like you do it with the DeviceConnectionClient.

This is what #2326 provides.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/KafkaConsumerConfigProperties.java
b-abel: What are you aiming at? Would you prefer an empty map or do you think an exception should be thrown?

sophokles73: Such methods usually return an empty map, I'd say. Would that work in this context?

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/test/FakeProducer.java
b-abel: I think this is useful for all unit tests involving a Kafka client. I expect to have at least one module for a client for backend applications (something like a module _application-kafka_ inside the module _clients_). And then there will probably be a Kafka client in the _command routing component_ discussed in #2273.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/tracing/LongTag.java
b-abel: No, it's not. Let's decide on [how to proceed](https://github.com/eclipse/hono/pull/2303#discussion_r526028936) with tracing, then I will move it to a more appropriate location.

sophokles73: I do not see any reason why we shouldn't move it to the place where our other OpenTracing related (helper) classes are currently located, or is there? FMPOV this should go into `org.eclipse.hono.tracing` in the hono-core module ...

clients/kafka-common/pom.xml
sophokles73: The current one will be deprecated ASAP :-) FMPOV there is no need to need to keep compatibility with Java 1.8.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/test/FakeProducer.java
sophokles73: Then you should indeed move it to the test-utils module

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
sophokles73: I do not mind the if-else block in *setCollaborators* for the time being. We will most probably need to do something like that for the Quarkus adapters anyway, as it doesn't support conditional instantiation based on runtime configuration ...

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
```diff
@@ -297,13 +332,58 @@ public TelemetrySender downstreamTelemetrySender(
...
+     * @param kafkaProducerFactory The producer factory to use.
+     * @param kafkaProducerConfig The producer configuration to use.
+     */
+    @Bean
+    @ConditionalOnKafkaMessaging
+    @Scope("prototype")
+    public EventSender downstreamEventKafkaSender(
+            final CachingKafkaProducerFactory<String, Buffer> kafkaProducerFactory,
+            final KafkaProducerConfigProperties kafkaProducerConfig) {
+        return new KafkaBasedEventSender(kafkaProducerFactory, kafkaProducerConfig, getTracer());
```
sophokles73, line 384: both the event sender  and the telemetry sender seem to use the same configuration, or am I mistaken? If that is the case, what advantage does it have to create two separate senders? is there a reason why they shouldn't use the same KafkaProducer instance (and the same connection to the Kafka cluster)?

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/tracing/LongTag.java
b-abel: I agree. The only reason I did not do it yet is that we discussed if we shouldn't drop the tags for the `RecordMetadata` and if we do so, I would remove the `LongTag` as it is not used anywhere else.

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
b-abel: These are very valid questions. I just had no idea where I could explain the reasoning behind it.

In general, a Kafka client does not open a single connection but rather connects to the brokers to whose partitions it sends messages, i.e. between 1 and the cluster size. Since the topics for events and telemetry do not have to be on the same brokers, on average there are not many more connections through the separate sender instances.

I see it as an advantage that the "rare but important" events do not have to wait in the same queue as the much more frequent telemetry messages. Since the messages are collected in batches before sending and are sent in sequence with retries and rather generous timeouts (the default timeout for sending is [2 minutes](https://kafka.apache.org/documentation/#delivery.timeout.ms)), I expect a lower latency for events under load. 
I would also like us to have the possibility to decide later that both types of messages should be configured differently.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/CachingKafkaProducerFactory.java
b-abel: CachingClientFactory was not public when I wrote the code.
FMPOV the `CachingKafkaProducerFactory` is not comparable to `CachingClientFactory` but more to the sender factories like `org.eclipse.hono.client.impl.DownstreamSenderFactoryImpl`. It might be that the naming causes confusion.
I could have used `CachingClientFactory` inside of `CachingKafkaProducerFactory` to hold the producer instances. But it does not fit well because it is designed around managing a connection and synchronizing expensive client creation. Using a shared KafkaProducer all of this is already provided by `io.vertx.kafka.client.producer.impl.KafkaProducerImpl`, so I think a simple map is sufficient to hold the producer.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/test/FakeProducer.java
b-abel: @sophokles73 I now found a way to use the MockProducer of the Kafka client in the client of Vert.x. Therefore I removed the FakeProducer after all.

clients/kafka-common/src/test/java/org/eclipse/hono/kafka/client/CachingKafkaProducerFactoryTest.java
```diff
@@ -49,7 +59,25 @@
...
+            final Handler<Future<RecordMetadata>> blockingCode = invocation.getArgument(0);
+            blockingCode.handle(result.future());
+            return null;
+        }).when(context).executeBlocking(VertxMockSupport.anyHandler(), any());
+
+        final BiFunction<String, Map<String, String>, KafkaProducer<String, Buffer>> instanceSupplier = (n, c) -> {
+            final MockProducer<String, Buffer> mockProducer = new MockProducer<>(true, new StringSerializer(),
+                    new BufferSerializer());
+            return KafkaProducer.create(vertxMock, mockProducer);
+        };
```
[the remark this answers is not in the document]
b-abel: This is only the _standard producer supplier_ for unit tests. For productive usage, you would use KafkaProducerFactory.sharedProducerFactory()`. Does this make sense from your POV?

clients/adapter-kafka/src/test/java/org/eclipse/hono/adapter/client/telemetry/kafka/TestHelper.java
```diff
@@ -30,41 +48,65 @@ private TestHelper() {
...
-     *
-     * @throws NoSuchElementException if the given factory does not contain the expected producer (e.g. when the
-     *             producer got closed after a fatal error).
-     * @throws ClassCastException if the provided producer implementation is not an instance of {@link FakeProducer}.
+     * @param producer The (mock) producer to be wrapped.
+     * @return The producer factory.
      */
-    public static <K, V> MockProducer<K, V> getUnderlyingMockProducer(
-            final CachingKafkaProducerFactory<K, V> producerFactory, final String producerName) {
+    public static CachingKafkaProducerFactory<String, Buffer> newProducerFactory(
```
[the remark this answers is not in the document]
b-abel: This method intentionally returns the implementation. The idea is that the factory manages the lifecycle of the producers. In unit tests, I can easily inspect the internal state by invoking `CachingKafkaProducerFactory.getProducer()`, a method that is not part of the interface.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/KafkaProducerFactory.java
```diff
@@ -0,0 +1,67 @@
...
+    KafkaProducer<K, V> getOrCreateProducer(String producerName, Map<String, String> config);
+
+    /**
+     * Closes the producer with the given producer name if it exists.
+     *
+     * @param producerName The name of the producer to remove.
+     * @return A future that is completed when the close operation completed or a succeeded future if no producer
+     *         existed with the given name.
+     */
+    Future<Void> closeProducer(String producerName);
```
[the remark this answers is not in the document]
b-abel: I added explanations in my last commit. Are they helpful?

b-abel:
I added the application of tenant level and device level defaults to downstream messages. There are currently no further changes FMPOV that need to go into this PR.
Please let me know if you want me to do the rebase.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/tracing/KafkaHeaderInjectAdapter.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ */
+public final class KafkaHeaderInjectAdapter implements TextMap {
+
+    private final List<KafkaHeader> headers;
+
+    /**
+     * Creates an adapter for a list of {@link KafkaHeader} objects.
+     *
+     * @param headers The list of {@link KafkaHeader} objects.
```
sophokles73, line 34: I guess the headers param is mandatory and should therefore be checked for `null`?

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
```diff
@@ -134,14 +139,26 @@ protected void setCollaborators(
             adapter.setCommandConsumerFactory(commandConsumerFactory(adapterProperties, samplerFactory, commandRouterClient));
         }
 
+        final KafkaProducerConfigProperties kafkaProducerConfig = kafkaProducerConfig();
+        if (kafkaProducerConfig.getProducerConfig().isEmpty()) {
```
sophokles73, line 143: IMHO we should better add a `KafkaProducerConfigProperties.isConfigured()` method in order to decouple the test from the inner structure of the properties object. WDYT?

site/documentation/content/admin-guide/hono-kafka-client-configuration.md
```diff
@@ -0,0 +1,49 @@
...
+
+The configuration needs to be provided in the form `HONO_KAFKA_PRODUCERCONFIG_${PROPERTY}` as an environment variable or
+as a command line option in the form `hono.kafka.producerConfig.${property}`, where `${PROPERTY}` respectively 
+`${property}` is any of the Kafka client's [producer properties](https://kafka.apache.org/documentation/#producerconfigs). 
+The provided configuration is passed directly to the Kafka producer without Hono parsing or validating it.
+The following table shows which properties _are_ changed by Hono.
+
+| Environment Variable<br>Command Line Option | Mandatory Value | Description |
+| :------------------------------------------ | :-------------- | :---------- |
+| `HONO_KAFKA_PRODUCERCONFIG_KEY_SERIALIZER`<br>`--hono.kafka.producerConfig.key.serializer` | `org.apache.kafka.common.serialization.StringSerializer` | The record keys in Hono are always strings. Any other specified value is ignored. |
```
sophokles73, line 28: why do we make this configurable if we only accept `org.apache.kafka.common.serialization.StringSerializer` anyway?
Apart from that I am not sure if the name of the command line option will result in the desired outcome of putting key `key.serializer` to the Map. Have you tested if Spring Boot maps this as you expected?

site/documentation/content/admin-guide/hono-kafka-client-configuration.md
```diff
@@ -0,0 +1,49 @@
...
+| `HONO_KAFKA_PRODUCERCONFIG_VALUE_SERIALIZER`<br>`--hono.kafka.producerConfig.value.serializer` | `io.vertx.kafka.client.serialization.BufferSerializer` | The record values in Hono are always byte arrays.  Any other specified value is ignored. |
```
sophokles73, line 29: same here

site/documentation/content/admin-guide/hono-kafka-client-configuration.md
```diff
@@ -0,0 +1,49 @@
...
+| `HONO_KAFKA_PRODUCERCONFIG_ENABLE_IDEMPOTENCE`<br>`--hono.kafka.producerConfig.enable.idempotence` | `true` | The Hono Kafka client uses only idempotent producers.  Any other specified value is ignored. |
```
sophokles73, line 30: and here

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/tracing/KafkaHeaderInjectAdapter.java
b-abel: Right. Good catch, thank you.

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
b-abel: This is a good idea. I will do it.

site/documentation/content/admin-guide/hono-kafka-client-configuration.md
b-abel: > why do we make this configurable if we only accept `org.apache.kafka.common.serialization.StringSerializer` anyway?

FMPOV this is not configurable. If you have an idea of how I can make it more clear in the description here, I'll be happy to change it. 
The logic is the following: every property that starts with the prefix `hono.kafka.producerConfig` is put into a map (without the prefix of course). The map is finally provided to the "native" Kafka client. You can add arbitrary properties there. The Kafka client applies the known properties and ignores those that it does not know. The table lists the exceptions: properties that are effectively not configurable because Hono overwrites them. You can set the command line option `--hono.kafka.producerConfig.key.serializer=bumlux`. Hono (i.e. the class `KafkaProducerConfigProperties`) will change the value of the property to `org.apache.kafka.common.serialization.StringSerializer` (a message about this is printed on debug level). 
I don't see a way to exclude properties from that mechanism. Alternatively, Hono would have to support each Kafka client config property explicitly (which are many dozens) and would then be tied to specific versions of the client (because properties change between versions).
Unless someone comes up with a better idea, I think simply overwriting the properties where we expect a certain value is an acceptable solution.

> Apart from that I am not sure if the name of the command line option will result in the desired outcome of putting key `key.serializer` to the Map. Have you tested if Spring Boot maps this as you expected?

I verified that it works as expected. Btw.: I was more surprised that Spring Boot applies its transformation magic (changing to lower case and replacing underscores by dots) not only on the level of the variable names (the prefix) but additionally with the part that it uses as keys in the map (therefore I thought initially we could not provide generic environment variables here). But it works fine.

sophokles73: Ok, now I understand.
How about replacing
> The following table shows which properties _are_ changed by Hono.

and the table below with
> The following properties can *not* be set using this mechanism because the protocol adapters use fixed values instead in order to implement the message delivery semantics defined by Hono's Telemetry and Event APIs.

Then add a table with the kafka property names and the fixed values used by the adapters.

b-abel:
I pushed the requested changes.

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/telemetry/kafka/AbstractKafkaBasedDownstreamSender.java
```diff
@@ -72,22 +74,25 @@
...
         Objects.requireNonNull(producerFactory);
         Objects.requireNonNull(producerName);
         Objects.requireNonNull(config);
         Objects.requireNonNull(tracer);
 
-        this.config = config;
+        this.producerFactory = producerFactory;
         this.producerName = producerName;
+        this.config = config;
+        this.adapterConfig = adapterConfig;
```
kaniyan, line 97: Missing _null_ check for the `adapterConfig`.

adapters/pom.xml
```diff
@@ -82,6 +82,10 @@
       <groupId>io.opentracing</groupId>
       <artifactId>opentracing-noop</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.wildfly.security</groupId>
+      <artifactId>wildfly-elytron-password-impl</artifactId>
```
sophokles73, line 87: out of curiosity: which component uses/requires this dependency?

b-abel: I have not had a deep look into this. But it looked like it is the "native" Kafka client. When I configured the Kafka client to use the SASL mechanism _SCRAM-SHA-256_, the authentication failed with an exception related to Wildfly and password checking. Adding this dependency did the trick...

b-abel: @sophokles73 Is this sufficient from your POV? Or should we investigate further?

sophokles73: No, all good

sophokles73 (approved):
LGTM
nice work @b-abel :+1:

Issue #2301, opened 2020-11-17
Add integration tests for auto-provisioning status properties


fkaltner:
As discussed in https://github.com/eclipse/hono/pull/2287#issuecomment-728796657

fkaltner:
@sophokles73 Please assign this to me.

sophokles73:
@fkaltner can this be closed?

PR #2304, opened 2020-11-17
#2301 auto provisioning status properties integration tests

fkaltner:


fkaltner:
In the first commit I also replaced the "dummy" implementation in FileBasedRegistrationService by a more complete one as running the integration tests against it feels a bit odd otherwise...

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceManagementIT.java
```diff
@@ -734,11 +735,19 @@ private static String assertLocationHeader(final MultiMap responseHeaders, final
...
     }
 
-    private static void assertRegistrationInformation(final Device response, final Device expectedData) {
+    private static void assertRegistrationInformation(final HttpResponse<Buffer> response, final Device expectedData) {
 
+        final JsonObject actualDeviceJson = response.bodyAsJsonObject();
+
+        // internal status is not decoded from JSON as users should not be allowed to change internal status
+        final JsonObject actualDeviceStatus = actualDeviceJson.getJsonObject(RegistryManagementConstants.FIELD_STATUS);
+        assertThat(actualDeviceStatus.getBoolean(RegistryManagementConstants.FIELD_AUTO_PROVISIONED)).isFalse();
```
sophokles73, line 744: currently, the Device Registry Management API does not require the *status* property to be present in the response body. It also doesn't mention the *autoprovisioned* and * autoprovisioningnotificationsent* property at all.
So, IMHO we need to add these properties to the API spec, if they indeed can be returned by a registry implementation.

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
```diff
@@ -707,47 +705,57 @@ public String toString() {
...
         return tempDeviceId;
     }
 
-    private static final class FileBasedDeviceDto extends BaseDto<Versioned<Device>> {
+    private static final class FileBasedDeviceDto extends DeviceDto {
 
         FileBasedDeviceDto() {
         }
 
+        public static DeviceDto forRead(final String tenantId, final String deviceId, final JsonObject entry) {
```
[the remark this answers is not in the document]
fkaltner: Yes, I was reluctant to change the signature of DeviceDto's _forUpdate_ method taking a _Supplier_ as this leads to changes in JDBC and MongoDb registry and I didn't want to do this because of a the file based registry which is just for "playing around" 😐.

But as you said below, it seems like this is too confusing. I adapted the implementation to use _FileBasedDeviceDto_ everywhere.

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationService.java
```diff
@@ -75,7 +73,7 @@
     private static final Logger LOG = LoggerFactory.getLogger(FileBasedRegistrationService.class);
 
     // <tenantId, <deviceId, registrationData>>
-    private final ConcurrentMap<String, ConcurrentMap<String, FileBasedDeviceDto>> identities = new ConcurrentHashMap<>();
+    private final ConcurrentMap<String, ConcurrentMap<String, DeviceDto>> identities = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
fkaltner: see above.

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceManagementIT.java
fkaltner: Sure, makes sense.

I just became confused again: should the properties be camelCased or kebab-cased? I am asking since I noticed that _last-user_ is kebab-cased the rest is camelCased.

By majority decision counting the properties by case I think it is camelCased? 😄

sophokles73: camel case, yes. `last-user` should be changed to `lastUser` FMPOV. But we will handle that in another issue: #2309

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedDeviceDto.java
```diff
@@ -44,7 +44,7 @@ public MongoDbBasedDeviceDto() {
      */
     public static DeviceDto forRead(final String tenantId, final String deviceId, final JsonObject recordJson) {
 
-        return DeviceDto.forRead(tenantId,
+        return DeviceDto.forRead(DeviceDto::new, tenantId,
```
sophokles73, line 47: shouldn't this method return a `MongoDbBasedDeviceDto`?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -1030,6 +1032,13 @@ components:
             "last-user":
                type: string
                description: The user who made the last edit on this device.
+            "autoProvisioned":
```
sophokles73, line 1035: can you also please bump the spec's version to 1.5.0

sophokles73 (approved):
LGTM

PR #2307, opened 2020-11-19
Fix search devices operation and integration tests

kaniyan:
* The integration tests for the search devices operation have not been running as it expects
  the system property `hono.deviceregistry.supportsSearchDevices` set to true. This property
  has been renamed to  `deviceregistry.supportsSearchDevices` in the `pom.xml` of _tests_ 
  module but not in the corresponding test class. This has been fixed now to use the renamed
  property in the test class.

* Integration tests corresponding to the search devices operation have been disabled for the 
 JDBC based device registry as it doesn't implement this operation.

* When the data type of a filter value is string and the value doesn't contain any search wild 
  card characters then that filter has been ignored by the MongoDB device registry. This has 
  been fixed.

* When there are no filters specified in the request, then at least the results should be 
  filtered based on the tenant identifier in the request URL. This has been fixed in the 
  MongoDB device registry.

* Fix integration tests corresponding to the search devices operation.

### DIFF
--- a/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDocumentBuilder.java
+++ b/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDocumentBuilder.java
@@ -207,12 +207,9 @@ public final class MongoDbDocumentBuilder {
         filters.forEach(filter -> {
             if (filter.getValue() instanceof String) {
                 final String value = (String) filter.getValue();
-                if (value.contains("*") || value.contains("?")) {
-                    // if the value contains * and ? then use regex matching
-                    document.put(mapDeviceField(filter.getField()),
-                            new JsonObject().put("$regex",
-                                    DeviceRegistryUtils.getRegexExpressionForSearchOperation(value)));
-                }
+                document.put(mapDeviceField(filter.getField()),
+                        new JsonObject().put("$regex",
+                                DeviceRegistryUtils.getRegexExpressionForSearchOperation(value)));
             } else {
                 document.put(mapDeviceField(filter.getField()), filter.getValue());
             }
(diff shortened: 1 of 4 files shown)


services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDocumentBuilder.java
```diff
@@ -207,12 +207,9 @@ public MongoDbDocumentBuilder withDeviceFilters(final List<Filter> filters) {
         filters.forEach(filter -> {
             if (filter.getValue() instanceof String) {
                 final String value = (String) filter.getValue();
-                if (value.contains("*") || value.contains("?")) {
-                    // if the value contains * and ? then use regex matching
-                    document.put(mapDeviceField(filter.getField()),
-                            new JsonObject().put("$regex",
-                                    DeviceRegistryUtils.getRegexExpressionForSearchOperation(value)));
-                }
+                document.put(mapDeviceField(filter.getField()),
```
sophokles73, line 210: would you mind creating an issue for this that we can refer to in this PR as this seems to be clearly a bug, right?

tests/src/test/java/org/eclipse/hono/tests/registry/DeviceManagementIT.java
```diff
@@ -403,7 +401,7 @@ public void testDeregisterDeviceFailsForNonExistingDevice(final VertxTestContext
      *      Device Registry Management API - Search Devices</a>
      */
     @Nested
-    @EnabledIfSystemProperty(named = "hono.deviceregistry.supportsSearchDevices", matches = "true")
+    @EnabledIfSystemProperty(named = "deviceregistry.supportsSearchDevices", matches = "true")
```
sophokles73, line 404: nice catch :+1:

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDocumentBuilder.java
kaniyan: I have created an issue  #2308 for the above.

sophokles73:
@kaniyan Did you see the failing integration tests for the Mongo DB registry?

kaniyan:
> @kaniyan Did you see the failing integration tests for the Mongo DB registry?

yes, I am working on it.

sophokles73 (approved):
LGTM

PR #2312, opened 2020-11-20
[#2029] Use Command Router in adapters and integration tests

calohmn:
This is for #2029.

This adds a `org.eclipse.hono.adapter.client.command.CommandConsumerFactory` implementation that uses the new Command Router component.
For the integration tests, a new maven profile `command-router` is added to let the tests run using the Command Router component.
The GitHub action workflow has been adapted to use that profile in the test-run that uses the jdbc device registry (so that the other test-runs still use the old command routing mechanism).

### DIFF
--- a/tests/pom.xml
+++ b/tests/pom.xml
@@ -39,6 +39,7 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch 
     -->
     <hono.amqp-network.host>hono-dispatch-router.hono</hono.amqp-network.host>
     <hono.device-connection.host>hono-service-device-connection.hono</hono.device-connection.host>
+    <hono.commandrouter.host>hono-service-command-router.hono</hono.commandrouter.host>
     <hono.registration.host>hono-service-device-registry.hono</hono.registration.host>
     <hono.auth.host>hono-service-auth.hono</hono.auth.host>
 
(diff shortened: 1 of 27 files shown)


services/command-router/src/main/resources/logback-spring.xml
```diff
@@ -27,29 +27,35 @@
...
 
-    <logger name="io.vertx.proton.impl" level="INFO"/>
-    <logger name="io.vertx.core.net.impl" level="INFO"/>
-  </springProfile>
+<!--    <logger name="io.netty.handler.logging.LoggingHandler" level="INFO"/>-->
+<!--    <logger name="io.netty.resolver.dns" level="INFO"/>-->
+
+<!--    <logger name="io.vertx.proton.impl" level="INFO"/>-->
+<!--    <logger name="io.vertx.core.net.impl" level="INFO"/>-->
+<!--  </springProfile>-->
```
sophokles73: this doesn't look right ...

tests/pom.xml
```diff
@@ -875,6 +902,79 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
                     </wait>
                   </run>
                 </image>
+                <!-- ##### Command Router service (may also act as Device Connection service via 'enable-device-connection-endpoint') ##### -->
+                <image>
+                  <name>${docker.repository}/hono-service-command-router-test</name>
+                  <alias>command-router</alias>
+                  <build>
+                    <imagePullPolicy>IfNotPresent</imagePullPolicy>
```
sophokles73, line 910: we should also skip building the image if the command router is disabled ...

tests/src/test/resources/commandrouter/logback-spring.xml
```diff
@@ -0,0 +1,68 @@
...
+<!--  </springProfile>-->
+
+<!--  <springProfile name="dev">-->
+<!--    <logger name="org.eclipse.hono" level="DEBUG"/>-->
+<!--  </springProfile>-->
+
+<!--  <springProfile name="prod">-->
+<!--    <logger name="org.eclipse.hono" level="INFO"/>-->
+<!--  </springProfile>-->
+
```
sophokles73, line 62: this doesn't look right

calohmn: This is only for debugging the integration test errors (I didn't get them locally). It's in the 2nd commit which won't be merged.

calohmn: I now also get the test errors locally (sometimes occurring in CoAP tests). I think it's an issue with the CommandRouter component. I'm looking into it and will see to fix it outside of this PR.

calohmn: I've removed the above log changes. PR with the fix for the intermittent test failures is in #2323.

calohmn: I've rebased now to include the fix here.

tests/pom.xml
calohmn: Done

sophokles73 (approved):
LGTM

PR #2313, opened 2020-11-23
Fix tracing for Quarkus based adapters

calohmn:
HonoConnectionImpl#tracer needs to be set explicitly, otherwise the NoopTracer is used.


sophokles73 (approved):
LGTM

PR #2318, opened 2020-11-23
Support using Device Connection cache client with Quarkus adapters

sophokles73:
The Quarkus based adapters now also support connecting to a remote cache
containing Device Connection data directly using the Infinispan Hotrod
client.


service-base-quarkus/src/main/java/org/eclipse/hono/service/quarkus/AbstractProtocolAdapterApplication.java
```diff
@@ -216,10 +224,30 @@ protected CommandRouterClient commandRouterClient() {
      * @return The client.
      */
     protected DeviceConnectionClient deviceConnectionClient() {
-        return new ProtonBasedDeviceConnectionClient(
-                HonoConnection.newConnection(vertx, config.deviceConnection, tracer),
-                messageSamplerFactory,
-                protocolAdapterProperties);
+
+        LOG.debug(deviceConnectionCacheConfig.toString());
+        LOG.debug(config.deviceConnection.toString());
```
calohmn, line 229: Was that intended to be kept here? If so, maybe put some more context info in the log output?

dejanb (approved):
LGTM

service-base-quarkus/src/main/java/org/eclipse/hono/service/quarkus/AbstractProtocolAdapterApplication.java
sophokles73: I have removed the log statements

PR #2319, opened 2020-11-23
[#2288] Fixing device level QoS forwarding of events for HTTP adapter

fkaltner:

### DIFF
--- a/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
+++ b/adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/AbstractVertxBasedHttpProtocolAdapter.java
@@ -675,7 +675,7 @@ public abstract class AbstractVertxBasedHttpProtocolAdapter<T extends HttpProtoc
                     .map(c -> ttdTracker.result())
                     .orElse(null);
             final Message downstreamMessage = newMessage(
-                    ctx.getRequestedQos(),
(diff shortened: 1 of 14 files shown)


fkaltner:
Can you please re-trigger the build? I think the failing test case (CredentialsAmqpIT#testGetCredentialsWithAutoProvisioning) is not 100% stable as I could not reproduce the error locally.

sophokles73:
The request timeout for the AMQP clients is set to 200ms in the IntegrationTestSupport class. Maybe you can increase that value to 1000ms in order to compensate for the fewer CPU resources available on the CI system ...

fkaltner:
> The request timeout for the AMQP clients is set to 200ms in the IntegrationTestSupport class. Maybe you can increase that value to 1000ms in order to compensate for the fewer CPU resources available on the CI system ...

Not sure, if I got everything right, but on 1.4.x I couldn't find that timeout setting in IntegrationTestSupport. I guess you were referring to master (where it is set to 400ms as far as I can see).

I guess it makes sense to set it to 1000ms on master and leave it like it is on 1.4.x (as it involves merging the changes to set this in the client configuration properties in IntegrationTestSupport to 1.4.x)?

sophokles73:
This is the place: https://github.com/eclipse/hono/blob/5a5deb5654716197c205634c36a16bd8d27d851e/tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java#L456

sophokles73 (approved):
LGTM

PR #2320, opened 2020-11-23
[#2288] Add integration tests verifying forwarding of QoS level in downstream messages

fkaltner:

### DIFF
--- a/tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
@@ -73,7 +73,7 @@ public final class IntegrationTestSupport {
     /**
      * The default number of milliseconds to wait for a response to an AMQP 1.0 performative.
      */
-    public static final int DEFAULT_AMQP_TIMEOUT = 1000;
+    public static final int DEFAULT_AMQP_TIMEOUT = 400;
     /**
      * The default port exposed by the AMQP adapter.
      */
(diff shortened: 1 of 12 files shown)


tests/src/test/java/org/eclipse/hono/tests/http/EventHttpIT.java
```diff
@@ -52,4 +63,37 @@ protected void assertAdditionalMessageProperties(final Message msg) {
...
+        if (setup.failed()) {
+            ctx.failNow(setup.causeOfFailure());
+            return;
+        }
+
+        httpClient.create(
+                getEndpointUri(),
+                Buffer.buffer("hello"),
+                requestHeaders,
+                ResponsePredicate.status(HttpURLConnection.HTTP_BAD_REQUEST))
```
fkaltner, line 96: Note that there was a code change which leads to a different behaviour in Hono's AMQP adapter in 1.5.:
Previously the message was accepted (and the QoS was ignored), now it is rejected.

It's not specified as far as I can see how the adapter will react on this, but this may impact client if they had accidentally set a _qos_level_ HTTP header for events to AT_MOST_ONCE.

sophokles73: you mean HTTP adapter, right?

tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
```diff
@@ -73,7 +73,7 @@
     /**
      * The default number of milliseconds to wait for a response to an AMQP 1.0 performative.
      */
-    public static final int DEFAULT_AMQP_TIMEOUT = 400;
+    public static final int DEFAULT_AMQP_TIMEOUT = 1000;
```
sophokles73, line 76: please revert this change, we are already applying a factor > 1 to this value when running on test environment

tests/src/test/java/org/eclipse/hono/tests/http/EventHttpIT.java
fkaltner: Yes, of course 😅 Sorry for the confusion.

PR #2323, opened 2020-11-23
[#2029] Fix Command Router AMQP server vert.x context issues

calohmn:
When processing Command Router API requests, the original vert.x context of the CommandRouterAmqpServer needs to be restored at the end, because the commandConsumerFactory has switched to the context in which the downstream AMQP connection is used.

This fixes issues with the Command Router component sending invalid AMQP frames or throwing exceptions when handling the requests.

I'm also working on a more general fix concerning verticle handling/communication in the Command Router component, which shall use multiple `ProtonBasedCommandConsumerFactoryImpl` instances and thereby use multiple downstream AMQP connections.

### DIFF
--- a/services/command-router/src/main/java/org/eclipse/hono/commandrouter/ApplicationConfig.java
+++ b/services/command-router/src/main/java/org/eclipse/hono/commandrouter/ApplicationConfig.java
@@ -33,6 +33,7 @@ import org.eclipse.hono.config.VertxProperties;
 import org.eclipse.hono.deviceconnection.infinispan.client.BasicCache;
 import org.eclipse.hono.deviceconnection.infinispan.client.CacheBasedDeviceConnectionInfo;
 import org.eclipse.hono.deviceconnection.infinispan.client.CommonCacheConfig;
+import org.eclipse.hono.service.HealthCheckProvider;
 import org.eclipse.hono.service.HealthCheckServer;
 import org.eclipse.hono.service.VertxBasedHealthCheckServer;
 import org.eclipse.hono.service.amqp.AmqpEndpoint;
@@ -63,6 +64,7 @@ import io.opentracing.contrib.tracerresolver.TracerResolver;
 import io.opentracing.noop.NoopTracerFactory;
 import io.vertx.core.Vertx;
 import io.vertx.core.VertxOptions;
+import io.vertx.ext.healthchecks.HealthCheckHandler;
 
 /**
  * Spring Boot configuration for the Command Router service.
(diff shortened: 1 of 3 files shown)


sophokles73:
IMHO the problem here is that the service implementation is being deployed as a Verticle and thus initialized with a different Context than the AmqpServer exposing the endpoint. There is no need to do so as slong as the service implementation implements `Lifecycle` it will be started automatically during deployment of the AmqpServer which starts up all its endpoints implicitly. This way we can be sure that each incoming request message will be processed on a single event loop thread. 
Based on that, I believe that the service implementation should not need to implement Verticle at all.

calohmn:
The service implementation being a Verticle is not the issue here, IMHO. 
We have here the scenario of the CommandRouter API AMQPServer deployed as ${maxInstances} Verticle instances.
Then we have the (for now singleton) CommandConsumerFactory using a downstream HonoConnection tied to one particular Vert.x context (and switching to it for each of its operations).
Therefore, the processing of a CommandRouter API request message will probably result in 2 vert.x contexts being involved. And that's why I've added the switch back to the AMQP server context here.

The service implementation being a separate Verticle means that the CommandConsumerFactory and its HonoConnection will be initialized in that Verticle - separate from the AMQPServer verticles. That makes sense since processing of incoming command messages via the CommandConsumerFactory should be separate from handling of CommandRouter API requests via the AMQPServer.

Admittedly, the implementation with this PR, containing the multiple "runOnContext" calls, is not very clean. It's still close to the implementation in the protocol adapters. That's why I want to refactor it further, creating multiple CommandConsumerFactory verticles (allowing multiple downstream AMQP connections) and decoupling them from the AMQP server (via basically just a ConcurrentHashSet that keeps the tenants for which there need to be links, and info about the current 'connected' state of downstream connections).

sophokles73:
Why don't you simply create one CommandConsumerFactory (along with its HonoConnection) per AmqpServer (and corresponding service) instance? Then everything would run on the same event loop thread, wouldn't it?

calohmn:
> Why don't you simply create one CommandConsumerFactory (along with its HonoConnection) per AmqpServer (and corresponding service) instance? Then everything would run on the same event loop thread, wouldn't it?

I've now changed it accordingly, seeing that it's better to have multiple CommandConsumerFactory instances now already.
Decoupling the CommandConsumerFactory instances and allowing for more optimized `command/$tenant` link handling can be done later then.

services/command-router/src/main/java/org/eclipse/hono/commandrouter/ApplicationConfig.java
```diff
@@ -172,13 +174,28 @@ public ObjectFactoryCreatingFactoryBean amqpServerFactory() {
...
     @Bean
     @Scope("prototype")
-    public AmqpEndpoint commandRouterAmqpEndpoint(final CommandRouterService service) {
-        return new DelegatingCommandRouterAmqpEndpoint<>(vertx(), service);
+    public AmqpEndpoint commandRouterAmqpEndpoint() {
+        final CommandRouterService commandRouterService = commandRouterService();
+        return new DelegatingCommandRouterAmqpEndpoint<>(vertx(), commandRouterService) {
+
+            @Override
+            public void registerLivenessChecks(final HealthCheckHandler handler) {
```
sophokles73, line 186: how about adding this as the default implementation to `org.eclipse.hono.service.amqp.AbstractDelegatingRequestResponseEndpoint`?
It seems to make a lot of sense to register these checks automatically, doesn't it?

calohmn: I also thought about this. But that would have implications on the `hono-service-device-connection` module and its usage of the `DelegatingDeviceConnectionAmqpEndpoint` class, where the given service implementation is already registered with the health check by other means (done in the `Application` class there).

sophokles73: I see, then let's keep it like this for the time being ...

sophokles73 (approved):
LGTM

PR #2326, opened 2020-11-24
Refactor ProtonBasedDownstreamSender into separate subclasses.

b-abel:
Having separate types for the downstream senders makes the qualifiers obsolete and allows leveraging conditional bean creation for multiple implementations by retrieving the beans from Spring's application context.


calohmn:
That of course means the number of AMQP connections for telemetry/event senders is doubled. See also https://github.com/eclipse/hono/pull/2276#discussion_r515776778.

b-abel:
> The way it is implemented here means the number of AMQP connections for telemetry/event senders is doubled. See also [#2276 (comment)](https://github.com/eclipse/hono/pull/2276#discussion_r515776778).

My understanding of the current code is that there are two instances of `ProtonBasedDownstreamSender` created, each with a separate instance of `HonoConnection`. What I am missing?

sophokles73:
It is at least possible to use the same connection for events and telemetry ... we could easily instantiate only one instance of the downstream sender and use it for both.

b-abel:
We agreed on another solution (https://github.com/eclipse/hono/pull/2303#discussion_r529501341), making this PR obsolete.

PR #2327, opened 2020-11-24
Fix credentials caching and assert cache directives in AbstractCredentialsServiceTest

lhotari:
- Credential caching was not used when using JDBC device registry

- the pre-existing `org.eclipse.hono.service.credentials.AbstractCredentialsServiceTest#getExpectedCacheDirective`
  was unused. Modify tests to assert the cache directives of read operations
   - for `CredentialsManagementService.readCredentials` calls
   - for `CredentialsService.get` calls

- also fix gaps in `FileBasedCredentialsService` & `MonoDbBasedCredentialsService`
  - cache `org.eclipse.hono.service.management.credentials.CredentialsManagementService#readCredentials`
    method, this resolves a TODO in `FileBasedCredentialsService`
    - use the maximum cache directive (or no cache) for a result with multiple credentials
    - cache an empty result by resolving the cache directive of `CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD`

### DIFF
--- a/services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedCredentialsService.java
+++ b/services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedCredentialsService.java
@@ -22,6 +22,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -608,6 +609,17 @@ public final class FileBasedCredentialsService extends AbstractCredentialsManage
                 .map(CommonCredential::stripPrivateInfo)
                 .collect(Collectors.toList());
 
+        // resolve cache directive for results.
+        // Limit caching to the case when cache directives are the same for all results
+        final Set<CacheDirective> cacheDirectives = result.stream().map(CommonCredential::getType)
+                .distinct()
+                .map(this::getCacheDirective)
+                .collect(Collectors.collectingAndThen(Collectors.toUnmodifiableSet(),
+                        // use cache options for hashed password when results is empty so that empty result gets cached
+                        set -> set.isEmpty() ? Set.of(getCacheDirective(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD)) : set));
+        final Optional<CacheDirective> cacheDirective = cacheDirectives.size() == 1 ?
+                cacheDirectives.stream().findFirst() : Optional.empty();
+
         // note that the result set might be empty
         // however, in this case an empty set of credentials has been
         // registered for the device explicitly and we return
@@ -615,8 +627,7 @@ public final class FileBasedCredentialsService extends AbstractCredentialsManage
         return Future.succeededFuture(
                 OperationResult.ok(HttpURLConnection.HTTP_OK,
                         result,
-                        // TODO check cache directive
-                        Optional.empty(),
+                        cacheDirective,
                         Optional.of(currentVersion)));
     }
 
--- a/services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedCredentialsService.java
+++ b/services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedCredentialsService.java
@@ -22,7 +22,6 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -609,16 +608,7 @@ public final class FileBasedCredentialsService extends AbstractCredentialsManage
                 .map(CommonCredential::stripPrivateInfo)
                 .collect(Collectors.toList());
 
-        // resolve cache directive for results.
-        // Limit caching to the case when cache directives are the same for all results
-        final Set<CacheDirective> cacheDirectives = result.stream().map(CommonCredential::getType)
-                .distinct()
-                .map(this::getCacheDirective)
-                .collect(Collectors.collectingAndThen(Collectors.toUnmodifiableSet(),
-                        // use cache options for hashed password when results is empty so that empty result gets cached
-                        set -> set.isEmpty() ? Set.of(getCacheDirective(CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD)) : set));
-        final Optional<CacheDirective> cacheDirective = cacheDirectives.size() == 1 ?
-                cacheDirectives.stream().findFirst() : Optional.empty();
+        final CacheDirective cacheDirective = resolveCacheDirective(result);
 
         // note that the result set might be empty
         // however, in this case an empty set of credentials has been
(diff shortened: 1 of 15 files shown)


services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedCredentialsService.java
```diff
@@ -775,6 +786,7 @@ private CacheDirective getCacheDirective(final String type) {
             switch (type) {
             case CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD:
             case CredentialsConstants.SECRETS_TYPE_X509_CERT:
+            case CredentialsConstants.SECRETS_TYPE_PRESHARED_KEY:
```
sophokles73, line 789: we deliberately chose to not allow caching of PSK credentials as they are plain text ...

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedCredentialsService.java
```diff
@@ -608,15 +609,25 @@ private boolean hasUniqueSecretIds(final JsonObject credentialObject) {
                 .map(CommonCredential::stripPrivateInfo)
                 .collect(Collectors.toList());
 
+        // resolve cache directive for results.
+        // Limit caching to the case when cache directives are the same for all results
+        final Set<CacheDirective> cacheDirectives = result.stream().map(CommonCredential::getType)
```
sophokles73, line 614: FMPOV we do not need to support caching of this response as it is part of the Registry Management API which we expect to be invoked rarely only. The CredentialsService, on the other hand, is being invoked permanently by the protocol adapters during runtime. Thus it is more important to support caching in the adapters.

services/device-registry-jdbc/src/main/java/org/eclipse/hono/deviceregistry/jdbc/impl/CredentialsServiceImpl.java
```diff
@@ -86,7 +90,7 @@ public CredentialsServiceImpl(final TableAdapterStore store) {
                             .put(CredentialsConstants.FIELD_AUTH_ID, key.getAuthId())
                             .put(CredentialsConstants.FIELD_SECRETS, new JsonArray(secrets));
 
-                    return CredentialsResult.from(HttpURLConnection.HTTP_OK, payload, CacheDirective.noCacheDirective());
+                    return CredentialsResult.from(HttpURLConnection.HTTP_OK, payload, ttl);
```
sophokles73, line 93: The cache directive included in the response should be determined dynamically (based on the credentials type) as in the other registry implementations

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedCredentialsService.java
```diff
@@ -377,6 +394,7 @@ private CacheDirective getCacheDirective(final String type) {
             switch (type) {
             case CredentialsConstants.SECRETS_TYPE_HASHED_PASSWORD:
             case CredentialsConstants.SECRETS_TYPE_X509_CERT:
+            case CredentialsConstants.SECRETS_TYPE_PRESHARED_KEY:
```
sophokles73, line 397: see above

sophokles73 (changes requested):

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedCredentialsService.java
lhotari: fixed

services/device-registry-jdbc/src/main/java/org/eclipse/hono/deviceregistry/jdbc/impl/CredentialsServiceImpl.java
lhotari: fixed

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedCredentialsService.java
lhotari: fixed

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedCredentialsService.java
lhotari: In the tests I was doing, this was causing a bottleneck. I have improved the change in the most recent commits. Please re-review.

sophokles73: Would you mind elaborating a little on what you had been doing in your tests and why you believe that the fact that the registry doesn't include a cache directive in responses to the management API operations caused a bottleneck?

lhotari: @sophokles73 The reason why I started making changes to cache `CredentialsManagementService.readCredentials` calls was because of the TODO that was present in FileBasedCredentialService.processReadCredentials method: https://github.com/eclipse/hono/blob/251b1472f0275b2174ea07d9ec00bd66d55fd4d0/services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedCredentialsService.java#L611-L620 .

Adding caching headers to CredentialsManagementService.readCredentials calls most likely didn't fix the performance bottleneck I was experiencing. IIRC, it was the fix (also part of this PR) to add caching headers to `TenantManagementService.readTenant`. I assume that there's a bug somewhere that the TenantManagementService.readTenant gets called on the "hot path", when new telemetry is being received. I haven't had the chance to verify my assumptions.

sophokles73: All implementations of the Tenant service include a proper max-age cache directive in its responses. So that should not be your problem. However, the JDBC based Credentials service implementation always includes a no-cache directive which prevents the protocol adapters from caching credentials. Depending on the kind of (device) protocol you have been using in your tests, this could have a major impact. For instance, when using the HTTP protocol adapter, each request from a device needs to verify the device's credentials again. So if your device used Basic Auth for authentication, the credentials are being retrieved from the Credentials service for each request.
Based on that I would very much like to accept your changes adding the cache directive to the JDBC based Credentials service implementation. But I would still rather not add cache directives to the responses returned by the Registry Management API. I can understand if you do not consider making this change worth your time anymore. In that case I will be happy to add the missing code to the JDBC registry myself. I just wanted to give you the chance to get your code merged ... 
Let me know how you want to proceed. Thanks in any case for your interest in Hono and your contributions so far :+1:

sophokles73: @lhotari any opinion? How should we proceed?

lhotari: I have started a new job and I don't have time to work on this PR. I'll close this PR.

sophokles73: I see. Thanks for your contributions and good luck and all the best for your new job :+1:

lhotari: Thanks @sophokles73

PR #2328, opened 2020-11-26
[#2068] Use cached connected check result in HotrodCache

### LINKED ISSUE #2068: Make HotrodCache connection check less expensive

calohmn:
This fixes #2068:
The `HotrodCache#checkForCacheAvailability()` method will now determine its outcome based on whether the last cache operation was successful or not. It will only "put" the connectionCheckKey/Value if the last cache operation took place more than 30s ago.
The `EmbeddedCache#checkForCacheAvailability()` method now only checks whether the cache instance was successfully obtained, not doing the "put" operation with connectionCheckKey/Value anymore.

### DIFF
--- a/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/EmbeddedCache.java
+++ b/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/EmbeddedCache.java
@@ -24,6 +24,7 @@ import org.slf4j.LoggerFactory;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
+import io.vertx.core.json.JsonObject;
 
 /**
  * An embedded cache.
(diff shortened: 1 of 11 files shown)


client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/EmbeddedCache.java
```diff
@@ -104,10 +99,22 @@ protected boolean isStarted() {
...
     }
 
+    @Override
+    public Future<JsonObject> checkForCacheAvailability() {
+
+        if (isStarted()) {
+            return Future.succeededFuture(new JsonObject());
+        } else {
+            // try to (re-)establish connection
+            connectToGrid();
```
sophokles73, line 115: maybe this method should also be renamed to *connectToCache*?

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/HotrodCache.java
```diff
@@ -127,4 +145,80 @@ protected boolean isStarted() {
         });
     }
 
+    @Override
+    protected <T> Future<T> withCache(
+            final Function<org.infinispan.commons.api.BasicCache<K, V>, CompletionStage<T>> futureSupplier) {
+        return super.withCache(futureSupplier)
```
sophokles73, line 151: IMHO invoking the super class implementation is an anti pattern as you have no control over what is happening there.
FMPOV *withCache* should be declared final and should be changed to invoke e.g. a *postCacheAccess(Handler<AsyncResult<?>>)* method with the outcome of the cache access. Subclasses can then override this method or not. But its purpose is well defined in any case ...

calohmn: Yes, that's better. I've changed it accordingly.

sophokles73 (approved):
LGTM

PR #2330, opened 2020-11-30
Include native transport libs by default

sophokles73:
All of Hono's service components now include Netty's native transport
libraries by default. The libraries are license compatible with EPL so
there is no reason to not include them by default and then allow users
to simply enable native transport using the existing configuration
property.


dejanb (approved):
Looks good

site/documentation/content/admin-guide/common-config.md
```diff
@@ -26,23 +26,18 @@ The vert.x framework instance used to run Hono's components on can be configured
...
 | `HONO_VERTX_MAX_EVENT_LOOP_EXECUTE_TIME_MILLIS`<br>`--hono.vertx.maxEventLoopExecuteTimeMillis` | no | `2000` | The maximum number of milliseconds that a task on the event loop may run without being considered to block the event loop. |
-| `HONO_VERTX_PREFER_NATIVE`<br>`--hono.vertx.preferNative`| no | `false` | Tries to enable *epoll()* support on Linux (if available). See the [notes below](./#epoll) for an explanation of the benefits of enabling *epoll*. |
+| `HONO_VERTX_PREFER_NATIVE`<br>`--hono.vertx.preferNative`| no | `false` | Enables/disables *epoll()* support on Linux/MacOS. See the [notes below](./#epoll) for an explanation of the benefits of enabling *epoll*. It is generally safe to set this property to `true` because Netty will disable native transport if the platform doesn't support it. |
```
calohmn, line 29: Is there a way to let this also be applied to the quarkus adapters?

Maybe this could also be a reason to rename `preferNative` to `preferNativeTransport` and deprecate the `preferNative` option (seeing this option in the config files without any context, I was first wondering whether this has something to do with quarkus native images; `preferNativeTransport` has at least the analogy with the vert.x option and the "netty-transport-native" module).

PR #2331, opened 2020-11-30
Use run-java.sh for starting Hono components

sophokles73:
Hono components no longer use the Spring Boot Maven plugin to create an
executable uber-jar. Instead, all runtime dependencies are now copied to
the container image and the run-java.sh script is used to start the
Spring Boot application.

The run-java.sh script provides more flexibility with configuring the
classpath and sets some reasonable default JVM parameters if
appropriate. This is also a first step towards harmonizing the container
image creation process between the Spring Boot and Quarkus based
components.


dejanb (approved):
Looks good

PR #2334, opened 2020-11-30
[#2333] Use ANYCAST routing type in broker config

### LINKED ISSUE #2333: Sandbox: Event messages aren't queued when there is no consumer
Sending `event` messages to a Sandbox Hono protocol adapter should result in the messages getting queued in the broker when there is no northbound consumer for the messages currently connected.

But that doesn't seem to be the case. `event` messages only get delivered to a northbound consumer application when the application is connected while the message is being sent.

See also corresponding issue in IoT Packages: https://github.com/eclipse/packages/issues/168.

calohmn:
This fixes #2333:
This fixes the issue of broker queues not getting created when event messages are sent and of event messages getting forwarded to multiple consumers.

Integration tests verifying the behaviour will be added in a separate PR (prepared by @kaniyan).

### DIFF
--- a/deploy/src/main/sandbox/artemis/broker.xml
+++ b/deploy/src/main/sandbox/artemis/broker.xml
@@ -1,7 +1,7 @@
 <?xml version='1.0'?>
 
 <!--
-    Copyright (c) 2016, 2018 Contributors to the Eclipse Foundation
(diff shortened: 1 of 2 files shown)


sophokles73 (approved):
LGTM

PR #2335, opened 2020-11-30
[#2333] Add integration tests

kaniyan:
When no north bound application is connected, the Event messages, which have been sent, are supposed to be queued by the Artemis broker. Once the north bound application connects those queued messages should be delivered. This was not working as
expected due to a bug described in the issue #2333. Integration tests are added to test the expected behavior.

These newly added tests will fail until the PR #2334 has been merged.

### DIFF
--- a/tests/src/test/java/org/eclipse/hono/tests/amqp/EventAmqpIT.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/amqp/EventAmqpIT.java
@@ -14,21 +14,21 @@ package org.eclipse.hono.tests.amqp;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import java.util.UUID;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 import java.util.stream.Stream;
 
 import org.apache.qpid.proton.amqp.messaging.Accepted;
 import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.client.MessageConsumer;
+import org.eclipse.hono.tests.IntegrationTestSupport;
 import org.eclipse.hono.util.AmqpErrorException;
 import org.eclipse.hono.util.MessageHelper;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 import io.vertx.core.Future;
-import io.vertx.core.Promise;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.junit5.VertxExtension;
 import io.vertx.junit5.VertxTestContext;
@@ -68,7 +68,7 @@ public class EventAmqpIT extends AmqpUploadTestBase {
 
     /**
      * Verifies that an event message from a device has been successfully sent and a north bound application, 
-     * which connects after the event has been sent, can successfully receive those event message.
+     * which connects after the event has been sent, can successfully receive this event message.
      *
      * @param ctx The vert.x test context.
      * @throws InterruptedException if test execution gets interrupted.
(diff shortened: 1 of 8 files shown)


tests/src/test/java/org/eclipse/hono/tests/amqp/EventAmqpIT.java
```diff
@@ -56,4 +65,65 @@ protected void assertAdditionalMessageProperties(final VertxTestContext ctx, fin
             assertThat(msg.isDurable()).isTrue();
         });
     }
+
+    /**
+     * Verifies that an event message from a device has been successfully sent and a north bound application, 
+     * which connects after the event has been sent, can successfully receive those event message.
```
calohmn, line 71: "... receive this event message"

tests/src/test/java/org/eclipse/hono/tests/amqp/EventAmqpIT.java
```diff
@@ -56,4 +65,65 @@ protected void assertAdditionalMessageProperties(final VertxTestContext ctx, fin
...
+                sendingComplete.fail(AmqpErrorException.from(delivery.getRemoteState()));
+            }
+        });
+
+        sendingComplete
+                .future()
+                .compose(eventSent -> {
+                    //THEN create a consumer once the event message has been successfully sent
+                    final Promise<MessageConsumer> consumerCreated = Promise.promise();
+                    vertx.setTimer(4000, tid -> {
```
calohmn, line 112: Is it necessary to wait 4s here? I would have guessed at most 200ms should be enough here, if waiting at all.

tests/src/test/java/org/eclipse/hono/tests/amqp/EventAmqpIT.java
```diff
@@ -56,4 +65,65 @@ protected void assertAdditionalMessageProperties(final VertxTestContext ctx, fin
...
+                        log.info("opening event consumer for tenant [{}]", tenantId);
+                        createConsumer(tenantId, msg -> receivedMessageCount.incrementAndGet())
```
sophokles73, line 114: how about using a checkpoint here instead and simply relying on the test timeout instead of setting another vert.x timer to wait for the message to arrive?

tests/src/test/java/org/eclipse/hono/tests/amqp/EventAmqpIT.java
sophokles73, line 112: why do you need to set a timer here?

tests/src/test/java/org/eclipse/hono/tests/amqp/EventAmqpIT.java
```diff
@@ -56,4 +65,65 @@ protected void assertAdditionalMessageProperties(final VertxTestContext ctx, fin
...
+                                .onComplete(consumerCreated);
+                    });
+                    return consumerCreated.future();
+                })
+                .compose(consumer -> {
+                    final Promise<Void> done = Promise.promise();
+                    vertx.setTimer(1000, tid -> {
+                        //THEN verify if the message is received by the consumer
+                        assertThat(receivedMessageCount.get()).isEqualTo(1);
```
sophokles73, line 123: instead of verifying the number of messages here I'd rather assert a particular property of the message that has been sent in the consumer itself, before flagging the checkpoint ...

sophokles73 (commented):
I have to admit that I do not really like the timer based approach as it makes certain assumptions about the underlying infrastructure which are not needed FMPOV

tests/src/test/java/org/eclipse/hono/tests/amqp/EventAmqpIT.java
```diff
@@ -56,4 +65,65 @@ protected void assertAdditionalMessageProperties(final VertxTestContext ctx, fin
...
+                    vertx.setTimer(4000, tid -> {
+                        log.info("opening event consumer for tenant [{}]", tenantId);
+                        createConsumer(tenantId, msg -> receivedMessageCount.incrementAndGet())
+                                .onComplete(consumerCreated);
+                    });
+                    return consumerCreated.future();
+                })
+                .compose(consumer -> {
+                    final Promise<Void> done = Promise.promise();
+                    vertx.setTimer(1000, tid -> {
```
calohmn, line 121: Instead of waiting a second, how about finishing the test in the consumer message handler, and making sure there is a timeout for the whole test method execution?

tests/src/test/java/org/eclipse/hono/tests/amqp/EventAmqpIT.java
kaniyan: It has been corrected.

tests/src/test/java/org/eclipse/hono/tests/amqp/EventAmqpIT.java
kaniyan: I have removed the timer approach and rewritten the tests.

tests/src/test/java/org/eclipse/hono/tests/amqp/EventAmqpIT.java
kaniyan: I have made use of a random UUID as a payload message. The event consumer verifies the payload content with that of the random UUID value.

kaniyan:
@calohmn @sophokles73 I have removed the timer based approach and rewritten those tests. Would you mind taking a look again?

sophokles73 (approved):
LGTM

calohmn (approved):

PR #2336, opened 2020-12-01
Support building of native images using Docker Maven plugin

sophokles73:
The build process of the Quarkus based adapters has been changed to make
use of the fabric8 Docker Maven plugin as we do with the other
components as well.

Also updated to Caffeine 2.8.6 and Quarkus 1.9.2 to get rid of manual registration of Caffeine cache classes for reflection when building native images.


dejanb (approved):
LGTM

PR #2337, opened 2020-12-01
[#2029] Add documentation for the Command Router component

calohmn:
This is for #2029.


site/documentation/content/concepts/command-and-control/index.md
```diff
@@ -121,3 +129,20 @@ Handling of the command response is done in the same way as shown in the chapter
...
 The particular device subscription has precedence then in choosing over which subscription protocol/channel to send the command to the gateway.
+
+<a name="command-router-service"></a>
+## New approach using the Command Router service
+
+The following diagrams show the message flow if protocol adapters are configured to use the [Command Router service]({{< relref "/admin-guide/command-router-config" >}}) instead of the Device Connection service.
+
+{{< figure src="command_control_concept_cmdrouter_1.svg" title="Command subscription" >}}
+
+In the scenario the device subscribes for commands (1) and the protocol adapter [registers the command consumer]({{< relref "/api/command-router#register-command-consumer-for-device" >}}) with the Command Router service, associating the device with its protocol adapter instance identifier (2). The Command Router service creates the receiver link (3) on the tenant-scoped link if it doesn't already exist. Following that, the notification about the device subscription is sent to the application via the AMQP messaging network (4).
```
sophokles73, line 140: how about 
> ... The Command Router service creates a receiver link scoped to the device's tenant (3) if it doesn't exist yet. ...

site/documentation/content/concepts/command-and-control/index.md
```diff
@@ -121,3 +129,20 @@ Handling of the command response is done in the same way as shown in the chapter
...
+
+{{< figure src="command_control_concept_cmdrouter_2.svg" title="Command handling" >}}
+
+Upon receiving the notification, the application prepares sender and command response receiver links (1,2) and sends the command message to the AMQP messaging network. The message is received by the Command Router service component (3), which will determine the protocol adapter instance #1 that is able to handle the command message. The command then gets forwarded to the AMQP messaging network on the address for adapter instance #1 (4). The protocol adapter instance #1 will receive the message (5) and forward it to the device (6). As the last step, an "accepted" disposition will be sent back to the application (7).
```
sophokles73, line 144: > .... an `accepted` disposition will be sent ...

site/documentation/content/admin-guide/command-router-config.md
```diff
@@ -0,0 +1,154 @@
...
+## Data Grid Connection Configuration
+
+The Command Router component requires either an embedded cache or a remote
+data grid, using the Infinispan Hotrod protocol to store device information.
+
+The following table provides an overview of the configuration variables and corresponding command line options for configuring the common aspects of the service:
+
+| Environment Variable<br>Command Line Option | Mandatory | Default | Description                                                             |
+| :------------------------------------------ | :-------: | :------ | :-----------------------------------------------------------------------|
+| `HONO_DEVICECONNECTION_COMMON_CACHENAME`<br>`--hono.deviceConnection.common.cacheName` | no | `device-connection` | The name of the cache |
```
sophokles73, line 105: `_DEVICECONNECTION_` ?

calohmn: I have used that deliberately, combining configuration data used by the `(CacheBased)DeviceConnectionInfo` class (which will probably still be there even when the DeviceConnection API has been removed).

Another option would be using "commandRouter.cache" instead (as in `hono.commandRouter.cache.common.cacheName` for example). I'd be fine with that as well. WDYT?

calohmn:
Changed the 2 parts mentioned above and added the missing "Command & Control Connection Configuration" in the Command Router config page.

site/documentation/content/admin-guide/command-router-config.md
sophokles73: FMPOV it would be rather confusing to have configuration properties with two different prefixes for configuring the command router service, don't you think?

calohmn: Well, configuring access to the (possibly external) cache *could* be seen as analogous to the configuration for the other external services (registration, credentials, etc), having their own prefix.
But as it's no defined external Hono *service*, I see that the "hono.commandRouter.cache" prefix may be better.

I've added commits now changing the prefix in the Command Router component and in the above documentation and also changing the default cache name.

site/documentation/content/admin-guide/command-router-config.md
```diff
@@ -0,0 +1,160 @@
...
+## Data Grid Connection Configuration
+
+The Command Router component requires either an embedded cache or a remote
+data grid, using the Infinispan Hotrod protocol to store device information.
+
+The following table provides an overview of the configuration variables and corresponding command line options for configuring the common aspects of the service:
+
+| Environment Variable<br>Command Line Option | Mandatory | Default | Description                                                             |
+| :------------------------------------------ | :-------: | :------ | :-----------------------------------------------------------------------|
+| `HONO_COMMANDROUTER_CACHE_COMMON_CACHENAME`<br>`--hono.commandRouter.cache.common.cacheName` | no | `command-router` | The name of the cache |
```
sophokles73, line 113: this change of the prefix will also need to be reflected in the configuration classes, won't it?

sophokles73: ah, just saw that it is ...

sophokles73 (approved):
LGTM

PR #2339, opened 2020-12-01
[#2338] Fix NPE in AMQP adapter when content-type is null.

### LINKED ISSUE #2338: NPE in AMQP adpter if no content-type set
The AMQP adapter throws a nullpointer exception when a telemetry message without a content-type is received.
This results in a log like the following: 

    18:16:53.031 [vert.x-eventloop-thread-3] WARN  o.eclipse.hono.tracing.TracingHelper - An unexpected error occurred!
    at java.base/java.util.Objects.requireNonNull(Unknown Source)
    at org.eclipse.hono.adapter.client.telemetry.amqp.ProtonBasedDownstreamSender.sendTelemetry(ProtonBasedDownstreamSender.java:121)
    at org.eclipse.hono.adapter.amqp.impl.VertxBasedAmqpProtocolAdapter.lambda$doUploadMessage$64(VertxBasedAmqpProtocolAdapter.java:1122)
    at io.vertx.core.Future.lambda$compose$3(Future.java:368)

b-abel:
This patch fixes #2338 by setting "application/octet-stream" as the content-type if nothing is set (analogous to the MQTT adapter).

### DIFF
--- a/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/AmqpContext.java
+++ b/adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/AmqpContext.java
@@ -101,11 +101,14 @@ public class AmqpContext extends MapBasedTelemetryExecutionContext {
 
     /**
      * Gets the content type of the AMQP 1.0 message.
+     * <p>
+     * If the message does not contain a content type, the {@linkplain MessageHelper#CONTENT_TYPE_OCTET_STREAM default
+     * content type} is returned.
      *
      * @return The content type of the AMQP 1.0 message.
      */
     final String getMessageContentType() {
-        return message.getContentType();
+        return Optional.ofNullable(message.getContentType()).orElse(MessageHelper.CONTENT_TYPE_OCTET_STREAM);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/impl/AmqpContext.java
```diff
@@ -105,7 +105,7 @@ final int getPayloadSize() {
      * @return The content type of the AMQP 1.0 message.
      */
     final String getMessageContentType() {
-        return message.getContentType();
+        return Optional.ofNullable(message.getContentType()).orElse(MessageHelper.CONTENT_TYPE_OCTET_STREAM);
```
calohmn, line 108: A hint in the javadoc about the default value would be good here (like in `MqttContext#contentType`).

b-abel: Sure. Will add.

b-abel: Done.

calohmn (approved):
LGTM

PR #2340, opened 2020-12-03
Fix getContentType() returning empty string

calohmn:
A not-set 'Content-Type' header caused an empty String to be returned by getContentType() although 'null' was
documented in the javadoc.
Since the AMQP message anyway gets the "application/octet-stream" content type in that case, the getContentType() method now returns "application/octet-stream" as the default (returning 'null' would have produced an NPE down the line).

Related to #2338.

Another option would have been returning `null` here and throwing a "BAD_REQUEST" error in the calling method when sending telemetry/events. That would be in accordance with the documentation as the content-type header is required, but I'd rather not introduce such a behaviour change that may break clients, instead keeping the behaviour analogous with MQTT and AMQP adapters (the CoAP adapter returns "BAD_REQUEST" though).

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttContext.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttContext.java
@@ -171,8 +171,8 @@ public final class MqttContext extends MapBasedTelemetryExecutionContext {
     /**
      * Gets the content type of the message payload.
      * <p>
-     * The type determined from the message topic's property
-     * bag, if if contains a content type.
+     * The type is determined from the message topic's property
+     * bag, if it contains a content type.
      * Otherwise, the {@linkplain MessageHelper#CONTENT_TYPE_OCTET_STREAM default
      * content type} is used.
      *
--- a/service-base/src/main/java/org/eclipse/hono/service/http/HttpContext.java
+++ b/service-base/src/main/java/org/eclipse/hono/service/http/HttpContext.java
@@ -22,6 +22,7 @@ import org.eclipse.hono.service.auth.DeviceUser;
 import org.eclipse.hono.service.metric.MetricsTags;
 import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.EventConstants;
+import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.QoS;
 import org.eclipse.hono.util.Strings;
 import org.eclipse.hono.util.TelemetryExecutionContext;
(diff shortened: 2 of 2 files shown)


sophokles73 (approved):
LGTM

sophokles73:
@calohmn should this go into 1.4.x or 1.3.x as well?

sophokles73:
In any case, can you add an entry to the release notes?

calohmn:
First looking into it, I thought there was no behaviour change here, as for telemetry/events the "application/octet-stream" content type is set here:
https://github.com/eclipse/hono/blob/b5ecf44168def95e52083d6aec509cc92b70048c/core/src/main/java/org/eclipse/hono/util/MessageHelper.java#L1323

But, for command response messages, that code is not invoked. So, I've also applied the fix in 1.4.x (https://github.com/eclipse/hono/commit/76c349df271cdd4c1d4c78f548624f76d16333b3) and 1.3.x and added a release notes entry.

sophokles73:
For command response messages we should continue to explicitly support *void* (i.e. empty) results, shouldn't we? In that case the message has no body and (consequently) no content-type. So I guess always returning the default content-type is not really correct in this case, is it?

calohmn:
Good point.
For an empty payload we have otherwise used the "application/vnd.eclipse-hono-empty-notification" content type:
https://github.com/eclipse/hono/blob/4533442e42935bf966a375e8ddcc72ca56b7dc4d/service-base/src/main/java/org/eclipse/hono/service/AbstractProtocolAdapterBase.java#L1228-L1234
But, on the other hand, usage of that type is only [documented](https://www.eclipse.org/hono/docs/api/event/#empty-notification) for *event* messages, so it may come as a bit of a surprise for command response messages.

How about:
- letting `HttpContext.getContext()` return `null` if no header is set
- allowing a `null` contentType in `org.eclipse.hono.adapter.client.telemetry.EventServer#sendEvent` and `org.eclipse.hono.adapter.client.telemetry.TelemetrySender#sendTelemetry`. Implementations of these methods should use "application/octet-stream" if contentType is `null` and given payload is not empty. Question is what to do if payload is empty - we could omit the contentType property in that case (which would be a change in behaviour).
- for command response messages
  - don't set the content-type property if given contentType is `null` and payload is empty
  - use "application/octet-stream" if contentType is `null` and given payload is not empty.

sophokles73:
> letting HttpContext.getContext() return null if no header is set

agreed

> allowing a null contentType in org.eclipse.hono.adapter.client.telemetry.EventServer#sendEvent and org.eclipse.hono.adapter.client.telemetry.TelemetrySender#sendTelemetry. Implementations of these methods should use "application/octet-stream" if contentType is null and given payload is not empty.

agreed

> Question is what to do if payload is empty - we could omit the contentType property in that case (which would be a change in behaviour)

I'd rather set the given content-type (if any). Analogous to our usage of `application/vnd.eclipse-hono-empty-notification` an application might understand other application specific content types that are used with an empty payload. I see no reason not to allow for that.

> for command response messages
> * don't set the content-type property if given contentType is null and payload is empty
> * use "application/octet-stream" if contentType is null and given payload is not empty.

agreed

calohmn:
> > Question is what to do if payload is empty - we could omit the contentType property in that case (which would be a change in behaviour)

> I'd rather set the given content-type (if any). Analogous to our usage of `application/vnd.eclipse-hono-empty-notification` an application might understand other application specific content types that are used with an empty payload. I see no reason not to allow for that.

Yes, I'd set the given non-null content-type also for an empty payload. I meant the above in the context of the given contentType being `null` (so, contentType `null` and payload empty). While currently this would set the `application/vnd.eclipse-hono-empty-notification` content type, I would change that so that the content type property is omitted then.

PR #2343, opened 2020-12-04
1.4.x: Let getContentType() return null if no header is set

calohmn:
This reverts the behaviour introduced in the last commit of using CONTENT_TYPE_OCTET_STREAM as default.
Instead, null is returned to also enable usage of the tenant default content type if set.

Applies the getContentType() change suggested in #2340 (but omits the other changes to keep this only a small change in 1.4.x).


sophokles73 (approved):
LGTM

PR #2344, opened 2020-12-04
Allow null contentType in Event/Telemetry Sender

calohmn:
This makes it possible that a content-type defined in the defaults of a tenant can be used as fallback.
This provides an alternative fix for #2338.

See also discussion in #2340 (changes regarding usage of "application/octet-stream" as default will go in separate PR).

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/impl/HttpBasedMessageMapping.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/impl/HttpBasedMessageMapping.java
@@ -120,7 +120,9 @@ public final class HttpBasedMessageMapping implements MessageMapping<MqttContext
             }
         });
         headers.add(MessageHelper.APP_PROPERTY_ORIG_ADDRESS, ctx.message().topicName());
-        headers.add(HttpHeaders.CONTENT_TYPE.toString(), ctx.contentType());
+        if (ctx.contentType() != null) {
+            headers.add(HttpHeaders.CONTENT_TYPE.toString(), ctx.contentType());
+        }
 
         final Promise<MappedMessage> result = Promise.promise();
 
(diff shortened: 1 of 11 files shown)


clients/adapter/src/main/java/org/eclipse/hono/adapter/client/telemetry/EventSender.java
```diff
@@ -47,7 +47,7 @@
      *         The future will be failed with a {@code org.eclipse.hono.client.ServerErrorException} if the data
      *         could not be sent. The error code contained in the exception indicates the
      *         cause of the failure.
-     * @throws NullPointerException if tenant ID, device ID or contentType are {@code null}.
+     * @throws NullPointerException if tenant ID or device ID are {@code null}.
```
sophokles73, line 50: how about adding a note to the *contentType* param that it will be set to the fallback content-type if a payload is given but no content-type is specified?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/impl/HttpBasedMessageMapping.java
```diff
@@ -120,7 +120,8 @@ private void mapMessageRequest(
             }
         });
         headers.add(MessageHelper.APP_PROPERTY_ORIG_ADDRESS, ctx.message().topicName());
-        headers.add(HttpHeaders.CONTENT_TYPE.toString(), ctx.contentType());
+        headers.add(HttpHeaders.CONTENT_TYPE.toString(),
```
sophokles73, line 123: FMPOV we should not set the header at all if the device did not provide a content-type. Maybe the mapper wants to handle this case specifically?

sophokles73 (commented):
didn't we want to set the content-type to the fall-back type in case of payload != null and content-type == null?
I had expected to see this handling being added to the MessageHelper class. Or am I mistaken?

calohmn:
> didn't we want to set the content-type to the fall-back type in case of payload != null and content-type == null?
I had expected to see this handling being added to the MessageHelper class. Or am I mistaken?

Yes, but that change isn't in here (see above: "changes regarding usage of "application/octet-stream" as default will go in separate PR"). I found the changes for that to be a bit tricky, that's why I want to use a separate commit only with that change in behaviour. So this PR here does only what the title says.

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/telemetry/EventSender.java
calohmn: Yes. I'll do that in the next commit (see below), then aligned with the updated javadoc for `MessageHelper#newMessage`.

sophokles73 (approved):
LGTM

PR #2345, opened 2020-12-07
[#2133] Improve message mapping API

Alfusainey:

@BobClaerhout @sophokles73 sorry, this took quiet sometime :-) This patch addresses the cases discussed in #2133

> No mapping service is configured for the gateway.

The default mapper now returns a `null` result in this case. The MQTT adapter checks for the presence of a null value and forwards the original message without any mapping

> Mapping service is configured for the gateway but invocation of the service results in an error status code.

We distinguish between success codes (2XX) and error codes (4XX, 5XX). The 200 status code results in a mapped message with the mapped payload and target address while status codes > 200 results in a mapped message with the original payload  and target address

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MessageMapping.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MessageMapping.java
@@ -31,15 +31,23 @@ public interface MessageMapping<T extends ExecutionContext> {
 
     /**
      * Maps a message uploaded by a device.
+     * <p>
+     * If a mapping service is not configured for a gateway or a protocol adapter, this method returns
+     * a successful future containing the original device payload and target address. If a mapping service is configured for a gateway
+     * or a protocol adapter and the mapping service returns a 200 OK HTTP status code, then this method returns a successful
+     * future containing the mapped payload.
+     * <p>
+     * For all other 2XX HTTP status codes, this method returns a mapped message containing the original device payload and 
+     * target address. In all other cases, this method returns a failed future with a {@link org.eclipse.hono.client.ServiceInvocationException}.
      *
-     * @param ctx The context in which the message has been uploaded.
-     * @param targetAddress The downstream address that the message will be forwarded to.
+     * @param ctx              The context in which the message has been uploaded.
+     * @param targetAddress    The downstream address that the message will be forwarded to.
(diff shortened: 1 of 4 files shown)


adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MessageMapping.java
```diff
@@ -31,15 +31,23 @@
 
     /**
      * Maps a message uploaded by a device.
+     * <p>
+     * If a mapping service is not configured for a gateway or a protocol adapter, this method returns
+     * a successful future containing a {@code null} value. If a mapping service is configured for a gateway
```
sophokles73, line 36: why do we need to introduce this additional case that the client needs to check for? Can't we simply do the same as for a 2xx != 200 status code, i.e. completing the future with a mapped message containing the original device payload and target address? That way the client wouldn't need to handle a `null` result specifically, would it?

sophokles73:
@Alfusainey sorry for the delay, we are in the middle of preparing the 1.5.0 release and are quite busy ;-)
Can you also check regarding the failing integration test(s)?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MessageMapping.java
Alfusainey: make sense. i was interpreting this requirement you stated in the issue

> No mapping service is configured for the gateway -> Nothing to be done at all

I interpreted doing nothing to mean returning null as a way to inform the client that nothing has been done. but will fix

Alfusainey:
> @Alfusainey sorry for the delay, we are in the middle of preparing the 1.5.0 release and are quite busy ;-)
> Can you also check regarding the failing integration test(s)?

@sophokles73 sure! am on holiday now but will resume next week and will take a look

sophokles73:
@Alfusainey Hi Alfu, we are planning to release 1.5.0 on Dec 17th, 2020. It would be great if we could include this PR ...

Alfusainey:
> @Alfusainey Hi Alfu, we are planning to release 1.5.0 on Dec 17th, 2020. It would be great if we could include this PR ...

@sophokles73 sure, i updated the PR

sophokles73 (approved):
LGTM

Issue #2346, opened 2020-12-07
Error sending "Connection Event" event messages


calohmn:
When configuring Connection Events with producer "events" (instead of "logging"), so that event messages are sent, there is an exception sending the event if the tenant isn't configured in a specific way.

Sending the events only works if the tenant has a `maxTtl` configured (with value other than -1).
Otherwise there is either a NPE (if no resourselimits are configured) or a NumberFormatException (if `maxTtl` isn't set to another value than -1).
````
MQTT17:38:45.343 [vert.x-eventloop-thread-3] WARN  o.e.h.a.m.i.VertxBasedMqttProtocolAdapter - failed to send connection event for client [clientId: d93938c0-ac22-48f4-967c-4670ee68470e]
MQTTjava.lang.NullPointerException: null
MQTT	at org.eclipse.hono.service.monitoring.AbstractMessageSenderConnectionEventProducer.lambda$sendNotificationEvent$0(AbstractMessageSenderConnectionEventProducer.java:87)
MQTT	at io.vertx.core.Future.lambda$compose$3(Future.java:368)
MQTT	at io.vertx.core.impl.FutureImpl.dispatch(FutureImpl.java:105)
MQTT	at io.vertx.core.impl.FutureImpl.onComplete(FutureImpl.java:83)
MQTT	at io.vertx.core.Future.compose(Future.java:364)
MQTT	at io.vertx.core.Future.compose(Future.java:336)
MQTT	at org.eclipse.hono.service.monitoring.AbstractMessageSenderConnectionEventProducer.sendNotificationEvent(AbstractMessageSenderConnectionEventProducer.java:76)
MQTT	at org.eclipse.hono.service.monitoring.AbstractMessageSenderConnectionEventProducer.connected(AbstractMessageSenderConnectionEventProducer.java:45)
MQTT	at org.eclipse.hono.service.AbstractProtocolAdapterBase.sendConnectedEvent(AbstractProtocolAdapterBase.java:1083)
````

(relates to #1714).

calohmn:
Fixed via d7e03b13f743c51bd68cf4b163a5ac5113269e13 (and commits in 1.3.x, 1.4.x).

PR #2347, opened 2020-12-07
1.3.x: [#2346] Fix ConnectionEvent messages not getting sent

calohmn:
Fix for #2346 (1.3.x):
Using the tenant maxTtl as the ttl parameter value in the `MessageHelper.newMessage()` invocation has been removed since
the tenant maxTtl will be taken into account in `MessageHelper.newMessage()` anyway.


sophokles73 (approved):
LGTM

PR #2349, opened 2020-12-07
Apply default content type in setPayload()

calohmn:
Based on the discussion in #2340:
The "application/octet-stream" content type now gets applied as default in the setPayload() methods, provided the payload isn't null. The "payload not null" condition is now also used in the other places where the content type gets set.

This effectively means:
- non telemetry/event messages (e.g. command response messages), that aren't empty, get the "application/octet-stream" content type as default as well
- a telemetry/event AMQP message without payload and content type, sent to the AMQP adapter, will now be forwarded without a content type.
- a default content type defined in the tenant configuration will now only get applied if the payload isn't null (this probably only applies to telemetry/event AMQP messages without payload sent to the AMQP adapter).

Note the extra handling of "application/vnd.eclipse-hono-empty-notification" in setPayload, getting set also for a `null` payload: This allows for the places where such empty events get sent to keep using a `null` payload.

### DIFF
--- a/core/src/test/java/org/eclipse/hono/util/MessageHelperTest.java
+++ b/core/src/test/java/org/eclipse/hono/util/MessageHelperTest.java
@@ -153,6 +153,7 @@ public class MessageHelperTest {
     public void testAddDefaultsAddsDefaultContentType() {
 
         final Message message = ProtonHelper.message();
+        message.setBody(new Data(new Binary("test".getBytes(StandardCharsets.UTF_8))));
         final ResourceIdentifier target = ResourceIdentifier.from(TelemetryConstants.TELEMETRY_ENDPOINT, Constants.DEFAULT_TENANT, "4711");
         final Map<String, Object> defaults = new JsonObject()
                 .put(MessageHelper.SYS_PROPERTY_CONTENT_TYPE, "application/hono")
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/eclipse/hono/util/MessageHelperTest.java
```diff
@@ -325,6 +341,64 @@ public void testNewMessageUsesGivenTtlValue() {
         assertThat(message.getTtl()).isEqualTo(timeToLive.toMillis());
     }
 
+    /**
+     * Verifies that the content-type of a newly created event message
+     * is set to the "application/vnd.eclipse-hono-empty-notification" content-type
+     * even if the payload is {@code null}.
```
sophokles73, line 347: this sounds like the content-type is set *despite* the fact that the body is null. However, my understanding is that this should happen *only* if the body is null, right? So, FMPOV this should read
> ... is set to the "application/vnd.eclipse-hono-empty-notification" content-type if the payload is null

calohmn: > this should happen only if the body is null, right?

From an overall perspective yes, but this condition is currently not enforced in the MessageHelper methods (there is also no automatic setting of "application/vnd.eclipse-hono-empty-notification" if the payload is null). 
(Throwing an IllegalArgumentException for "eclipse-hono-empty-notification" with payload != null (or not empty) would be the method doing too much, I think.)

So it's more like this content-type gets applied *even though* the payload is null, because normally the content-type parameter isn't applied if the payload is null. There's only the special case of the "application/vnd.eclipse-hono-empty-notification" content type for which this is done.

Anyway, I've changed the text now to:
> Verifies that the "application/vnd.eclipse-hono-empty-notification" content-type
can be set on a newly created event message with a `null` payload.

sophokles73 (approved):
LGTM

PR #2350, opened 2020-12-07
Add 'to' address in AMQP device command message

calohmn:
Preserving the behaviour of Hono 1.4, a command message sent to an AMQP device should have the 'to' address set as it was set on the AMQP message sent by the northbound application.


sophokles73 (approved):
LGTM

Issue #2250, opened 2020-10-13
Include Jaeger client in published Hono images


calohmn:
The Hono docker images published on Dockerhub should include the Jaeger OpenTracing client from Hono 1.5 onwards.
This makes it possible to easily setup a Hono deployment that includes the Jaeger components via the IoT Packages Hono helm chart.

Using a different OpenTracing client will still be possible by extending the image with the corresponding jars and making sure the client is found/loaded by means of the `TracerResolver`.

calohmn:
As building and pushing the Hono images to Dockerhub is done manually on every release, the corresponding build will just have to be done with the added `jaeger` maven profile.

calohmn:
To make it more explicit that we now see 'jaeger' as a standard build option (just like 'metrics-prometheus'), the documentation could be adapted, replacing the parts where `mvn clean install -Pbuild-docker-image,metrics-prometheus` is mentioned with `mvn clean install -Pbuild-docker-image,metrics-prometheus,jaeger`.

PR #2351, opened 2020-12-07
[#2250] Use Jaeger profile in standard build instructions

calohmn:
This is for #2250.

Also adds release notes entry about inclusion of Jaeger in images on Docker Hub.


sophokles73 (approved):
LGTM

PR #2352, opened 2020-12-08
[#2247] Map client to server error on sender creation

### LINKED ISSUE #2247: Don't propagate ClientErrorExceptions on telemetry sender creation to device
When creating a telemetry/event sender in a protocol adapter via `HonoConnection.createSender()` fails with a ClientErrorException, that exception should be propagated as a ServerErrorException to the device.

calohmn:
This fixes #2247.


sophokles73 (approved):
LGTM

PR #2353, opened 2020-12-08
Use logging.profile for integration test log levels

calohmn:
When running the integration tests with the `logging.profile` parameter (e.g. `-Dlogging.profile=trace` ), the logging profile is currently only applied to the components started in the docker containers, not the integration tests themselves.
This PR fixes that.

Uses `<include resource="logback-test-include-${logging.profile}.xml"/>` to select/apply the loggers for the selected profile.

(Using `<if condition="some conditional expression">` instead would have allowed putting it all into logback-test.xml, but that would require the Janino library as an extra dependency.)


calohmn:
@sophokles73 Could you have a look?

sophokles73 (approved):
LGTM

PR #2094, opened 2020-07-31
[#2053] Implementing auto provisioning of devices for gateways

fkaltner:
I am opening this PR (issue is still WIP) to get a first feedback on my approach.

Especially:

- I had to introduce a dependency to `DeviceBackend` in `AbstractRegistrationService`. Consequently there's now a circular dependency between `FileBasedRegistrationService` and `FileBasedDeviceBackend`. Unfortunately I do not see an obvious solution for it. Any hints?
- I tried to minimize the impact (in terms of amount of change and pretty similar, but not duplicated code) in protocol adapters. Is `AbstractProtocolAdapterBase` the right place for this protocol adapter independent code?
- Also I had to put some "flags" in the execution context in order to set the same flags again when the downstream message is constructed. I also noticed some other methods like `customizeDownstreamMessage`. Not sure if I should have put it there? Or a complete different approach I missed?
- I decided to use an Empty Event (with a corresponding application property being set) in case a telemetry message lead to an auto provisioning. Does that make sense from your POV?

PR #2358, opened 2020-12-09
#2053 implicit device registration file registry

fkaltner:
This pull request depends on https://github.com/eclipse/hono/pull/2094 being merged first.

Note however that this PR partially shares changes with #2359 and #2360 in order to not depend these being merged before.

### DIFF
--- a/services/device-registry-base/src/test/java/org/eclipse/hono/service/management/device/DelegatingDeviceManagementHttpEndpointTest.java
+++ b/services/device-registry-base/src/test/java/org/eclipse/hono/service/management/device/DelegatingDeviceManagementHttpEndpointTest.java
@@ -14,6 +14,7 @@
 
 package org.eclipse.hono.service.management.device;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.anyString;
@@ -37,6 +38,7 @@ import org.eclipse.hono.service.management.device.Sort.Direction;
 import org.eclipse.hono.util.RegistryManagementConstants;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
 
 import io.opentracing.Span;
 import io.vertx.core.Future;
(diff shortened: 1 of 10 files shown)


services/device-registry-base/src/test/java/org/eclipse/hono/service/management/device/DelegatingDeviceManagementHttpEndpointTest.java
```diff
@@ -196,6 +198,88 @@ public void testCreateDeviceRejectsInvalidPayload() {
...
+        final ArgumentCaptor<Device> deviceArgumentCaptor = ArgumentCaptor.forClass(Device.class);
+        verify(service).createDevice(anyString(), any(Optional.class), deviceArgumentCaptor.capture(), any(Span.class));
+
+        assertThat(deviceArgumentCaptor.getValue().getStatus()).isNull();
+    }
+
+    /**
+     * Makes sure that status properties are ignored when updating a device management via the REST API.
+     */
+    @SuppressWarnings("unchecked")
```
calohmn, line 245: not needed here

services/device-registry-file/src/test/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationServiceTest.java
```diff
@@ -87,6 +90,21 @@ public void setUp() {
...
+        final AutoProvisioner autoProvisioner = new AutoProvisioner(registrationService, registrationService);
+        autoProvisioner.setConfig(new AutoProvisionerConfigProperties());
+
+        final DownstreamSenderFactory downstreamSenderFactoryMock = mock(DownstreamSenderFactory.class);
+        when(downstreamSenderFactoryMock.connect()).thenReturn(Future.succeededFuture());
+        doAnswer(invocation -> {
+            final Handler<AsyncResult<Void>> handler = invocation.getArgument(0);
+            handler.handle(Future.succeededFuture());
+            return null;
+        }).when(downstreamSenderFactoryMock).disconnect(any(Handler.class));
```
calohmn, line 103: `VertxMockSupport.anyHandler()` can be used instead of `any(Handler.class)` to prevent the "unchecked assignment" warning.

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedServiceConfig.java
```diff
@@ -139,6 +199,13 @@ public FileBasedDeviceBackend deviceBackend(final Vertx vertx) {
         final FileBasedRegistrationService registrationService = new FileBasedRegistrationService(vertx);
         registrationService.setConfig(registrationProperties());
 
+        final AutoProvisioner autoProvisioner = new AutoProvisioner(registrationService, registrationService);
+        autoProvisioner.setVertx(vertx);
+        autoProvisioner.setDownstreamSenderFactory(downstreamSenderFactory(vertx));
+        autoProvisioner.setConfig(autoProvisionerConfigProperties());
```
calohmn, line 205: It looks to me as if invoking `autoProvisioner.setTracer()` and `autoProvisioner.setTenantInformationService()` is missing here.

calohmn: Have you tried defining `AutoProvisioner` as a Bean here instead, to make use of the Autowired annotations in the `AutoProvisioner` class?

services/device-registry-file/src/test/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationServiceTest.java
fkaltner: Thanks for the hint 👍

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedServiceConfig.java
fkaltner: Yes, to make it consistent I set them explicitly here as well.

fkaltner: It turned out that the _Tracer_ instance is defined in the parent application context. The _TenantInformationService_ was missing though.

fkaltner: Good catch, thanks! 😄 

It turned out that the _Tracer_ bean has to be taken from the root application context of the file based device registry and the TenantInformationService wasn't defined in the context so far leading to the NOOP service being used.

I could define AutoProvisioner as a bean instead, but then it must be refactored not to take any constructors arguments and use setters instead. 

But as it seems @sophokles73 changed the declaration explicitly to use setters in 17179adb6933c0eb3d1eeeccc26518e27297d977 I am wondering if we should do this in the scope of this PR or re-discuss this in a separate issue and change the "bean-declaration-style" not just for the beans affected in this PR but for all beans (and for all other registries as well)...

sophokles73:
@calohmn anything else to be done here or can this be merged?

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedServiceConfig.java
calohmn: Ok, yes, we can discuss a possible refactoring here once all registry changes have been merged.

calohmn (approved):
Can be merged FMPOV.

PR #2359, opened 2020-12-09
#2053 implicit device registration jdbc registry

fkaltner:
This pull request depends on #2094 being merged first.

Note however that this PR partially shares changes with #2358 and #2360  in order to not depend these being merged before.

### DIFF
--- a/services/base-jdbc/src/main/resources/org/eclipse/hono/service/base/jdbc/store/device/base.sql.yaml
+++ b/services/base-jdbc/src/main/resources/org/eclipse/hono/service/base/jdbc/store/device/base.sql.yaml
@@ -43,7 +43,7 @@ updateRegistrationVersioned: |
       data=:data,
       version=:next_version,
       updated_on=:updated_on,
-      auto_provisioning_notification_sent=IsNull(:auto_provisioning_notification_sent,auto_provisioning_notification_sent)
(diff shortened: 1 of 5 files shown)


services/base-jdbc/src/main/resources/org/eclipse/hono/service/base/jdbc/store/device/base.sql.yaml
```diff
@@ -43,7 +43,7 @@ updateRegistrationVersioned: |
       data=:data,
       version=:next_version,
       updated_on=:updated_on,
-      auto_provisioning_notification_sent=:auto_provisioning_notification_sent
+      auto_provisioning_notification_sent=IsNull(:auto_provisioning_notification_sent,auto_provisioning_notification_sent)
```
sophokles73, line 46: is this H2 specific syntax or is this supposed to be standard SQL?

fkaltner: Good point! I did some further research and it turned out that it is actual T-SQL specific (https://docs.microsoft.com/de-de/sql/t-sql/functions/isnull-transact-sql?view=sql-server-ver15). 
I wasn't aware that H2 supports anything beyond ANSI SQL at all.

COALESCE is ANSI SQL hence it makes sense to use COALESCE here as well in my opinion.

sophokles73:
@ctron would you mind taking a look as well?

PR #2360, opened 2020-12-09
#2053 implicit device registration mongodb registry

fkaltner:
This pull request depends on #2094 being merged first.

Note however that this PR partially shares changes with #2358 and #2359 in order to not depend on these being merged before.

### DIFF
--- a/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/ApplicationConfig.java
+++ b/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/ApplicationConfig.java
@@ -17,7 +17,10 @@ import java.util.Optional;
 
 import org.eclipse.hono.auth.HonoPasswordEncoder;
 import org.eclipse.hono.auth.SpringBasedHonoPasswordEncoder;
+import org.eclipse.hono.client.DownstreamSenderFactory;
+import org.eclipse.hono.client.HonoConnection;
 import org.eclipse.hono.config.ApplicationConfigProperties;
+import org.eclipse.hono.config.ClientConfigProperties;
 import org.eclipse.hono.config.ServerConfig;
 import org.eclipse.hono.config.ServiceConfigProperties;
 import org.eclipse.hono.config.VertxProperties;
@@ -31,6 +34,8 @@ import org.eclipse.hono.deviceregistry.mongodb.service.MongoDbBasedRegistrationS
 import org.eclipse.hono.deviceregistry.mongodb.service.MongoDbBasedTenantService;
 import org.eclipse.hono.deviceregistry.server.DeviceRegistryAmqpServer;
 import org.eclipse.hono.deviceregistry.server.DeviceRegistryHttpServer;
+import org.eclipse.hono.deviceregistry.service.device.AutoProvisioner;
+import org.eclipse.hono.deviceregistry.service.device.AutoProvisionerConfigProperties;
 import org.eclipse.hono.deviceregistry.service.tenant.AutowiredTenantInformationService;
 import org.eclipse.hono.service.HealthCheckServer;
 import org.eclipse.hono.service.VertxBasedHealthCheckServer;
@@ -287,18 +292,86 @@ public class ApplicationConfig {
     }
 
     /**
+     * Exposes a factory for creating clients for the <em>AMQP Messaging Network</em> as a Spring bean.
+     * <p>
+     * The factory is initialized with the connection provided by {@link #downstreamConnection(Vertx)}.
+     *
+     * @param vertx The vert.x instance to run on.
+     *
+     * @return The factory.
+     */
+    @Bean
+    @Scope("prototype")
+    public DownstreamSenderFactory downstreamSenderFactory(final Vertx vertx) {
+        return DownstreamSenderFactory.create(downstreamConnection(vertx));
+    }
+
+    /**
+     * Exposes the connection to the <em>AMQP Messaging Network</em> as a Spring bean.
+     * <p>
+     * The connection is configured with the properties provided by {@link #downstreamSenderFactoryConfig()}.
+     *
+     * @param vertx The vert.x instance to run on.
+     *
+     * @return The connection.
+     */
+    @Bean
+    @Scope("prototype")
+    public HonoConnection downstreamConnection(final Vertx vertx) {
+        return HonoConnection.newConnection(vertx, downstreamSenderFactoryConfig());
+    }
+
+    /**
+     * Exposes configuration properties for accessing the AMQP Messaging Network as a Spring bean.
+     *
+     * @return The properties.
+     */
+    @ConfigurationProperties(prefix = "hono.messaging")
+    @Bean
+    public ClientConfigProperties downstreamSenderFactoryConfig() {
+        return new ClientConfigProperties();
+    }
+
+    /**
+     * Gets properties for configuring gateway based auto-provisioning.
+     *
+     * @return The properties.
+     */
+    @Bean
+    @ConfigurationProperties(prefix = "hono.autoprovisioning")
+    public AutoProvisionerConfigProperties autoProvisionerConfigProperties() {
+        return new AutoProvisionerConfigProperties();
+    }
+
+    /**
      * Exposes the MongoDB registration service as a Spring bean.
+     *
+     * @param vertx The vert.x instance to run on.
+     * @param tracer The tracer to be used.
      *
      * @return The MongoDB registration service.
      */
     @Bean
     @Scope("prototype")
-    public MongoDbBasedRegistrationService registrationService() {
-        return new MongoDbBasedRegistrationService(
-                vertx(),
+    public MongoDbBasedRegistrationService registrationService(final Vertx vertx, final Tracer tracer) {
+        final AutowiredTenantInformationService tenantInformationService = tenantInformationService();
+        final MongoDbBasedRegistrationService mongoDbBasedRegistrationService = new MongoDbBasedRegistrationService(
+                vertx,
                 mongoClient(),
                 registrationServiceProperties(),
-                tenantInformationService());
+                tenantInformationService);
+
+        final AutoProvisioner autoProvisioner = new AutoProvisioner();
+        autoProvisioner.setVertx(vertx);
+        autoProvisioner.setTracer(tracer);
+        autoProvisioner.setDeviceManagementService(mongoDbBasedRegistrationService);
+        autoProvisioner.setTenantInformationService(tenantInformationService);
+        autoProvisioner.setDownstreamSenderFactory(downstreamSenderFactory(vertx));
+        autoProvisioner.setConfig(autoProvisionerConfigProperties());
+
+        mongoDbBasedRegistrationService.setAutoProvisioner(autoProvisioner);
+
+        return mongoDbBasedRegistrationService;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/ApplicationConfig.java
```diff
@@ -286,19 +291,87 @@ public ObjectFactoryCreatingFactoryBean amqpServerFactory() {
...
+     * <p>
+     * The factory is initialized with the connection provided by {@link #downstreamConnection(Vertx)}.
+     *
+     * @param vertx The vert.x instance to run on.
+     *
+     * @return The factory.
+     */
+    @Bean
+    @Scope("prototype")
+    public DownstreamSenderFactory downstreamSenderFactory(final Vertx vertx) {
```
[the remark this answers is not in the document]
fkaltner: Do you mind if I create an issue and do this after this PR, the JDBC PR and especially the integration tests are merged, so that I can test the migration more easily?

As this involves changing the AutoProvisioner this will not only affect the MongoDb registry but also the JDBC and File registry and would involve semi-manual re-testing on my machine for all of them.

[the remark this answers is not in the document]
sophokles73: sure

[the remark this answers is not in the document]
fkaltner: Created #2393

services/device-registry-mongodb/src/test/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedDeviceDtoTest.java
```diff
@@ -0,0 +1,68 @@
...
+                .setAutoProvisioned(true)
+                .setAutoProvisioningNotificationSent(true)
+                // make sure that these values do not interfere with the corresponding properties of DeviceDto's parent classes
+                .setCreationTime(Instant.now().plusSeconds(3600))
+                .setLastUpdate(Instant.now().plusSeconds(3600));
+        final Instant created = Instant.now().minusSeconds(1).truncatedTo(ChronoUnit.SECONDS);
+        final Instant updated = Instant.now().truncatedTo(ChronoUnit.SECONDS);;
+        final String version = "spam";
+
+        final var deviceDto = MongoDbBasedDeviceDto.forRead(MongoDbBasedDeviceDto::new, tenantId, deviceId, device, deviceStatus,
```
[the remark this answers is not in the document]
fkaltner: The reason I didn't use one of those is that I wanted to test if every property of the DTO is serialised as expected. If I had used forCreate() or forUpdate() not all fields would be populated and the test code would be partially duplicated.

Maybe it would help if I put what I stated above into a comment?

services/device-registry-mongodb/src/test/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedCredentialServiceTest.java
```diff
@@ -69,6 +74,9 @@ public void startService(final VertxTestContext testContext) {
 
         vertx = Vertx.vertx();
         mongoClient = MongoDbTestUtils.getMongoClient(vertx, "hono-credentials-test");
+        final DownstreamSenderFactory downstreamSenderFactoryMock = mock(DownstreamSenderFactory.class);
```
sophokles73, line 77: is this used anywhere in this test?

services/device-registry-mongodb/src/test/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDBBasedDeviceManagementSearchDevicesTest.java
```diff
@@ -66,6 +71,8 @@ public void setup(final VertxTestContext testContext) {
                 mongoClient,
                 config,
                 new NoopTenantInformationService());
+        final DownstreamSenderFactory downstreamSenderFactoryMock = mock(DownstreamSenderFactory.class);
```
sophokles73, line 74: is this used anywhere?

services/device-registry-mongodb/src/test/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedDeviceDtoTest.java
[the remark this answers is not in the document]
sophokles73: FMPOV it would be more helpful to have separate test methods which test the DTO in the way(s) that it is supposed to be used by client code ...

services/device-registry-mongodb/src/test/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedCredentialServiceTest.java
fkaltner: Yes in the line below:

`when(downstreamSenderFactoryMock.connect()).thenReturn(Future.succeededFuture());`

services/device-registry-mongodb/src/test/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDBBasedDeviceManagementSearchDevicesTest.java
fkaltner: See https://github.com/eclipse/hono/pull/2360#discussion_r554910673

services/device-registry-mongodb/src/test/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedCredentialServiceTest.java
sophokles73: Sure, but why should `downstreamSenderFactoryMock.connect()` ever be invoked? I do not see any place where the created downstreamSenderFactoryMock is being used by or injected into any of the objects under test. Or am I mistaken?

services/device-registry-mongodb/src/test/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedDeviceDtoTest.java
```diff
@@ -0,0 +1,97 @@
...
+        final var deviceDto = MongoDbBasedDeviceDto.forCreation(MongoDbBasedDeviceDto::new, tenantId, deviceId, true, device, version);
+        final var json = JsonObject.mapFrom(deviceDto);
+
+        assertThat(json).isNotNull();
+        // size should match number of assertions below, to make sure no unexpected properties are serialized
+        assertThat(json.getMap()).hasSize(7);
+
+        assertThat(json.getString(RegistryManagementConstants.FIELD_PAYLOAD_TENANT_ID)).isEqualTo(tenantId);
+        assertThat(json.getString(RegistryManagementConstants.FIELD_PAYLOAD_DEVICE_ID)).isEqualTo(deviceId);
+        assertThat(json.getInstant(RegistryManagementConstants.FIELD_STATUS_CREATION_DATE)).isNotNull();
```
sophokles73, line 57: should this be related to the instant set in the `DeviceStatus`?

sophokles73: what about the *updated on* field? Shouldn't that contain the same instant as the *created on* field in this case?

services/device-registry-mongodb/src/test/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedCredentialServiceTest.java
fkaltner: Sorry - didn't see the wood before the trees. You're right that is obsolete.

services/device-registry-mongodb/src/test/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedDeviceDtoTest.java
fkaltner: > should this be related to the instant set in the DeviceStatus?

No, actually the device status should be ignored.  

I added a DeviceStatus object to the test since the properties of DeviceStatus partially have the same name as the properties in the DTO subclasses. Before adding the _@JsonIgnore_ annotation to the _deviceStatus_ property in _DeviceDto_ they were accidentally serialised but the error was hard to spot since everything was still working but there was an additional, unwanted property being serialized. 

To make sure this doesn't accidentally happen in future (e.g. when someone changes the Jackson annotation again), I added that to the test case as well.

> what about the updated on field? Shouldn't that contain the same instant as the created on field in this case?

If the DTO was created by using forCreation() it can't have been updated yet which is why there is no update timestamp set. We could also set the same timestamp for created and updated but it is more cumbersome for someone viewing the entry to spot whether it has been updated at all yet (one needs to compare the created and updated timestamp for equality then).

services/device-registry-mongodb/src/test/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedDeviceDtoTest.java
```diff
@@ -0,0 +1,97 @@
...
+        assertThat(json.getString(BaseDto.FIELD_VERSION)).isEqualTo(version);
+        assertThat(json.getBoolean(RegistryManagementConstants.FIELD_AUTO_PROVISIONED)).isTrue();
```
sophokles73, line 59: but this is supposed to be the same value as in the DeviceStatus, right? If so, we could also make it more explicit by comparing to `DeviceStatus.isAutoProvisioned()`. WDYT?

services/device-registry-mongodb/src/test/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedDeviceDtoTest.java
sophokles73: > No, actually the device status should be ignored.
>
> I added a DeviceStatus object to the test since the properties of DeviceStatus partially have the same name as the properties in the DTO subclasses. Before adding the @JsonIgnore annotation to the deviceStatus property in DeviceDto they were accidentally serialised but the error was hard to spot since everything was still working but there was an additional, unwanted property being serialized.

That's a good idea :+1:  IMHO we should make it explicit then and add corresponding assertions as well.

fkaltner: It's already there (if I got you right):

`// size should match number of assertions below, to make sure no unexpected properties are serialized
  assertThat(json.getMap()).hasSize(7);`

By that it is made sure there is no other property serialized besides the one for which the test case contains assertions.

sophokles73: I was thinking about
```
// make sure that the creation date set on the new device is not the one contained in the DeviceStatus
assertThat(json.getInstant(RegistryManagementConstants.FIELD_STATUS_CREATION_DATE)).isBefore(deviceStatus.getCreationTime());
```

sophokles73 (approved):
LGTM

PR #2362, opened 2020-12-10
Remove non-working HonoConnectionImpl test

calohmn:
The connection attempt doesn't actually fail here (`isTrue()` missing after `assertThat(result.failed())`).
The connection getting closed after closing the session is already getting tested by `testRemoteSessionCloseTriggersReconnection()`.

In order to establish the behaviour that this test wants to verify, the `connect()` invocation would have to wait for the session `openHandler` to be called and only then finish the `connect` Future.
But that is also not done in the `ProtonConnectionImpl` default session handling.

### DIFF
--- a/client/src/test/java/org/eclipse/hono/client/impl/HonoConnectionImplTest.java
+++ b/client/src/test/java/org/eclipse/hono/client/impl/HonoConnectionImplTest.java
@@ -134,34 +134,6 @@ public class HonoConnectionImplTest {
     }
 
     /**
-     * Verifies that the client fails a connection attempt if no AMQP session can be established
-     * with the peer.
-     *
-     * @param ctx The vert.x test client.
-     */
-    @SuppressWarnings("unchecked")
-    @Test
-    public void testConnectFailsIfSessionCannotBeEstablished(final VertxTestContext ctx) {
-
-        // GIVEN a client attempting to connect to a peer that does not allow opening
-        // a session
-        final Future<HonoConnection> result = honoConnection.connect();
-        ctx.verify(() -> {
-            verify(session).open();
-            // WHEN the peer closes the session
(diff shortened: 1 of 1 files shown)


sophokles73:
> In order to establish the behaviour that this test wants to verify, the connect() invocation would have to wait for the session openHandler to be called and only then finish the connect Future.
But that is also not done in the ProtonConnectionImpl default session handling.

I agree with the analysis. But shouldn't the fix then be to change HonoConnectionImpl accordingly, instead of removing the test case that verifies the intended behavior?

calohmn:
> I agree with the analysis. But shouldn't the fix then be to change HonoConnectionImpl accordingly, instead of removing the test case that verifies the intended behavior?

Well, I wouldn't describe the current behaviour as necessarily wrong or problematic.

Current behaviour if peer doesn't open session:
- connect() returns succeeded Future; the connection then immediately gets closed again (via session closeHandler), triggering a reconnect

Potential new behaviour if peer doesn't open session:
- connect() fails internally (similar to the case of a connection error) and tries to reconnect. If then the session still isn't opened and at some point the reconnect attempts are exhausted, the connect() Future fails.

Sure, the described new behaviour concerning the connect() result Future looks better.
However, the overall outcome of reconnects getting triggered then is the same.

In the initial version of the changes in #2236, there was already the session openHandler getting set (see [here](https://github.com/eclipse/hono/pull/2236/commits/b2c6bb313886735a40b043449c94d3fa35fcc403#diff-b00dc82976b903763fc7bd9995fcadadd2fdf2d1e52bdc6c470418b1e3c73c28L738)). But that solution didn't trigger the reconnects (see [comment here](https://github.com/eclipse/hono/pull/2236#discussion_r503865667)). So, this would need to get adapted.
I can create an issue for this kind of improvement of HonoConnectionImpl.

sophokles73:
Ah, yes, I remember ... then let's keep it the way it is and indeed remove the test case ...

sophokles73 (approved):
LGTM

PR #2364, opened 2020-12-10
[#2357] Skip closing tenant link on tenant timeout if still needed

### LINKED ISSUE #2357: Tenant timeout mechanism mustn't close links if there are still subscriptions
The tenant-timeout mechanism (#1188/#1475) is currently used to close tenant-scoped command consumer links in the protocol adapters. It is triggered when there were no messages sent or received for a tenant for some time (1h per default).

But closing the tenant-scoped command consumer link when there are still command subscriptions from (AMQP, MQTT) devices, means that these devices won't get commands any more and are also not informed about this.

calohmn:
This fixes #2357.


sophokles73 (approved):
LGTM

Issue #2341, opened 2020-12-03
AMQP adapter: Message properties aren't adopted in downstream message anymore


calohmn:
The AMQP adapter documentation states that "Arbitrary properties (content-type, correlation-id, …)" can be set on Telemetry/Event messages, implying that these will be adopted in the downstream message.

With the 1.5 changes, such properties (apart from the content type) of the device AMQP message are not taken into account anymore.

That also breaks the contract of the AMQP adapter client interfaces:
https://github.com/eclipse/hono/blob/e07eb4752c9e2d8d45fe380bc999f7ce3e734060/client/src/main/java/org/eclipse/hono/client/device/amqp/TelemetrySender.java#L39

I think this should be fixed by adopting the AMQP message *application-properties*.
About the `correlation-id` AMQP message *property* I'm not so sure. On the one hand it's currently documented to be adopted, on the other hand it's rather an internal, protocol-specific property to maybe better not be taken into account.

(relates to #2338)

sophokles73:
FMPOV the AMQP adapter user guide currently does not explicitly state that all properties and application-properties of a message uploaded by a client are copied to the message being forwarded downstream. However, the guide currently is quite unspecific about which properties are actually required or optional.

Issue #1571, opened 2019-10-15
Add introductory Concept Page to Documentation


b-abel:
The documentation could benefit from an introductory concept page that gives a brief overview of Hono's idea, core features, and key terms. Ideally, it would specify the order in which the other concept pages should be read.

b-abel:
As far as I remember, @dejanb volunteered to write this.

b-abel:
Includes #1559.

b-abel:
Should IMHO explain the terms "downstream/upstream" and "northbound/southbound".

b-abel:
@dejanb, are you working on this? Otherwise, I volunteer to provide a PR for it.

sophokles73:
closed via #2580

PR #2367, opened 2020-12-11
[#2341] Improve AMQP adapter user guide

sophokles73:
The user guide for the AMQP adapter was unclear about which properties
are required/optional in messages uploaded by clients.

### DIFF
--- a/site/documentation/content/user-guide/amqp-adapter.md
+++ b/site/documentation/content/user-guide/amqp-adapter.md
@@ -3,7 +3,8 @@ title = "AMQP Adapter"
 weight = 220
 +++
 
-The AMQP protocol adapter allows clients (devices or gateway components) supporting the AMQP 1.0 protocol to publish messages to Eclipse Hono&trade;'s Telemetry, Event and Command & Control endpoints.
+The AMQP protocol adapter allows clients (devices or gateway components) supporting the AMQP 1.0 protocol to publish
+messages to Eclipse Hono&trade;'s Telemetry, Event and Command & Control endpoints.
 <!--more-->
 
 ## Device Authentication
@@ -21,9 +22,7 @@ This guide provides examples for publishing telemetry and events for *authentica
 ### SASL PLAIN Authentication
 
 The AMQP adapter supports authenticating clients using a *username* and *password*. This means that clients need to provide a *username* and a *password*
-when connecting to the AMQP adapter. If the adapter is configured for multi-tenancy (i.e `HONO_AMQP_SINGLE_TENANT` is set to `false`),
-then the *username* must match the pattern [*auth-id@tenant*], e.g. `sensor1@DEFAULT_TENANT`. Otherwise, the tenant-id can be omitted from the username
-and the client is assumed to belong to the `DEFAULT_TENANT`.
+when connecting to the AMQP adapter. The *username* must match the pattern [*auth-id@tenant*], e.g. `sensor1@DEFAULT_TENANT`.
 
 The adapter verifies the credentials provided by the client against the credentials that the
 [configured Credentials service]({{< relref "/admin-guide/common-config#credentials-service-connection-configuration" >}}) has on record for the client.
@@ -104,7 +103,7 @@ The client indicates its preferred message delivery mode by means of the *snd-se
 
 ## Error Handling
 
-The AMQP adapter distinguishes between two types of errors when a message is published using *at least once*:
+The AMQP adapter distinguishes between two types of errors when a message is published using *AT LEAST ONCE* delivery semantics:
 
 * An error caused by the client side, e.g invalid message address, content-type, adapter disabled for tenant etc.
 * An error caused by the server side, e.g no downstream consumers registered, downstream connection loss etc.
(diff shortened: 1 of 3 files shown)


site/documentation/content/user-guide/amqp-adapter.md
```diff
@@ -141,39 +140,45 @@ There are two JAR files in the hono/cli/target directory. The JAR to use for the
...
-        * (`amqp:unauthorized-access`): Request rejected because the adapter is disabled for tenant.
-        * (`amqp:precondition-failed`): Request does not fulfill certain requirements e.g adapter cannot assert device registration etc.
-        * (`amqp:resource-limit-exceeded`): Request rejected because the message limit for the given tenant is exceeded.
-
-When a device publishes data to the `telemetry` address, the AMQP adapter automatically determines the device's identity and tenant during the authentication process.
+  * (optional) Arbitrary payload contained in either a single AMQP *Data* or *AmqpValue* section.
+* Outcomes:
+  * `accepted`: The message has been successfully forwarded downstream.
+  * `released`: The message could not be processed by the adapter due to a (temporary) problem at the infrastructure level.
+    A common reason for this is that there is no active downstream consumer for messages from the device.
```
calohmn, line 168: I think "by the adapter" and "at the infrastructure level" are potentially misleading for the telemetry case here. The adapter could have forwarded it just fine, but the northbound application for example may not have sent a disposition update in time, leading to the (initially unsettled) message to get released.

site/documentation/content/user-guide/amqp-adapter.md
```diff
@@ -237,27 +247,36 @@ In this example, we are using message address `t/DEFAULT_TENANT/4711` which cont
...
-        * (`amqp:unauthorized-access`): The adapter is disabled for tenant.
-        * (`amqp:precondition-failed`): Request not fulfilling certain requirements.
-        * (`amqp:resource-limit-exceeded`): Request rejected because the message limit for the given tenant is exceeded.
-
-This is the preferred way for devices to publish events. It is available only if the protocol adapter has been configured to require devices to authenticate (which is the default).
+  * (optional) Arbitrary payload contained in either a single AMQP *Data* or *AmqpValue* section.
+* Outcomes:
+  * `accepted`: The message has been successfully forwarded downstream.
+  * `released`: The message could not be processed by the adapter due to a (temporary) problem at the infrastructure level.
+    A common reason for this is that there is no active downstream consumer for messages from the device.
```
calohmn, line 268: The "no active downstream consumer" scenario doesn't usually apply for event messages here.

site/documentation/content/user-guide/amqp-adapter.md
```diff
@@ -141,39 +140,45 @@ There are two JAR files in the hono/cli/target directory. The JAR to use for the
...
-  * Accepted: Message successfully processed by the adapter.
-  * Released: Message cannot be processed and should be redelivered.
-  * Rejected: Adapter rejects the message due to one of the following:
-        * (`hono:bad-request`): Request rejected due to a bad client request.
-        * (`amqp:unauthorized-access`): Request rejected because the adapter is disabled for tenant.
-        * (`amqp:precondition-failed`): Request does not fulfill certain requirements e.g adapter cannot assert device registration etc.
-        * (`amqp:resource-limit-exceeded`): Request rejected because the message limit for the given tenant is exceeded.
-
-When a device publishes data to the `telemetry` address, the AMQP adapter automatically determines the device's identity and tenant during the authentication process.
+  * (optional) Arbitrary payload contained in either a single AMQP *Data* or *AmqpValue* section.
```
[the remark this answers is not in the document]
sophokles73: *arbitrary* here refers to the fact that the semantics of the data contained is not of interest. But I will add some text further restricting the supported object type in the AmqpValue.

site/documentation/content/user-guide/amqp-adapter.md
```diff
@@ -141,39 +140,45 @@ There are two JAR files in the hono/cli/target directory. The JAR to use for the
...
-* Message Address: `telemetry` or `t`
-  * This refers to the `to` property of the message.
-* Settlement Mode: `presettled` (*AT MOST ONCE*) or `unsettled` (*AT LEAST ONCE*)
 * Authentication: SASL PLAIN or SASL EXTERNAL
+* Message *properties*:
+  * (required) `to`: either `telemetry` or `t`
+  * (optional) `content-type`: The type of payload contained in the message body. The given content type will be used
+    in the AMQP message being forwarded downstream if not empty. Otherwise, the content type of the downstream
+    message will be set to `application/octet-stream` if the payload is not empty and no default content type has been
+    defined for the origin device or its tenant (see [Downstream Meta Data]({{< relref "#downstream-meta-data" >}}).
```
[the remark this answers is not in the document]
sophokles73: yes, that is what I would prefer to do if that is ok for you as well

site/documentation/content/user-guide/amqp-adapter.md
sophokles73: From the client's view, it is the adapter which processes its uploaded message. The client does not know (nor need to know) if and how the adapter is forwarding the message. However, the message will also be released if one of the registry services cannot be reached by the adapter and therefore I think that it is appropriate to refer to *infrastructure* level in general here.

site/documentation/content/user-guide/amqp-adapter.md
sophokles73: The adapter has no influence on how the AMQP 1.0 Messaging Network is configured/implemented, has it? The statement seems to be true in any case, even if it never occurs ;-)

site/documentation/content/user-guide/amqp-adapter.md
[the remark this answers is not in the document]
calohmn: yes, that's ok with me.

site/documentation/content/user-guide/amqp-adapter.md
calohmn: Then I would at least say "potentially at the infrastructure level".
A developer responsible both for the device and the northbound application side could otherwise mistakenly read this error as "not my fault" while perhaps having forgotten to do a delivery update in the northbound application.

site/documentation/content/user-guide/amqp-adapter.md
calohmn: Well, ok.
On that note: I think, we haven't clearly stated the differences between event and telemetry messages or the motivation for having both concepts anywhere in the documentation, have we? I think it's at least missing from the Telemetry/Event API pages. I can create a PR...

sophokles73: I guess this could be added as part of #1571

sophokles73:
@calohmn I have pushed changes

calohmn (approved):
LGTM

PR #2372, opened 2020-12-17
[#2273] Prototype: Kafka based command router

kaniyan:
This draft PR provides a prototype of the command router component that uses Kafka as the messaging infrastructure. The aim of this draft PR is to provide a preview and a prototype to test various command & control scenarios using Kafka. This is an very early stage and hence it is not ready for review or merge. This draft uses Kafka clients from PR #2303 

- North bound applications should use the Kafka topic `hono.command.<tenant-Id>` with `deviceId` as the key to send commands.
- This Kafka based command router subscribes to receive those commands using a topic pattern `hono.command.*`. 
- The Kafka consumer in these command router instances have the same group id to ensure that a command is received by only one of the command router instances.
- On receiving a command, this router finds out to which adapter the device/gateway is connected to. Then forwards the command to the topic `hono.command_internal.<adapterInstanceId>`.

In order to deploy and tryout please follow the below steps:
1. You can also deploy a kafka cluster in _minikube_ as mentioned in the [PR #2216](https://github.com/eclipse/hono/pull/2216#issuecomment-730255279).

2. Use the helm charts workaround from the [IoT packages branch](https://github.com/bosch-io/packages/tree/%232273_Workaround_Kafka_based_command_router) to deploy Hono with this command router component.
      `helm install --dependency-update -n hono --set honoImagesTag=1.6.0-SNAPSHOT  eclipse-hono ./hono`

3. Ensure that a device (e.g. _4711_ belonging to the tenant _DEFAULT_TENANT_) is subscribed to receive commands.

4. Send commands to the device _4711_ using the below kafka command. Ensure that the command messages are prefixed with the _deviceId_ as key. Example: **4711:turnOnLight**

```
kubectl run kafka-producer -ti --image=strimzi/kafka:0.19.0-kafka-2.5.0 --rm=true --restart=Never -- bin/kafka-console-producer.sh --broker-list hono-kafka-bootstrap:9092  --property "parse.key=true" --property "key.separator=:" --topic hono.command.DEFAULT_TENANT
```
5. Use the below command to subscribe to the Kafka topic `hono.command_internal.<replace with adapter instance id>` to receive those commands. Make sure to use the correct adapter instance id in the above mentioned topic. Here is an example topic `hono.command_internal.Hono_adapter-http-vertx-Command_Control-1294e794-6dd0-4fbf-b908-d8b0dc0dd939` to demonstrate how it looks like.
```
kubectl run kafka-consumer -ti --image=strimzi/kafka:0.19.0-kafka-2.5.0 --rm=true --restart=Never -- bin/kafka-console-consumer.sh --bootstrap-server hono-kafka-bootstrap:9092  --topic hono.command_internal.<replace with adapter instance id>
```

### DIFF
--- a/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/telemetry/kafka/AbstractKafkaBasedMessageSender.java
+++ b/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/telemetry/kafka/AbstractKafkaBasedMessageSender.java
@@ -52,7 +52,7 @@ import io.vertx.kafka.client.producer.impl.KafkaHeaderImpl;
 /**
  * A client for publishing messages to a Kafka cluster.
  */
-public abstract class AbstractKafkaBasedDownstreamSender implements Lifecycle {
+public abstract class AbstractKafkaBasedMessageSender implements Lifecycle {
 
     /**
      * A logger to be shared with subclasses.
@@ -74,7 +74,7 @@ public abstract class AbstractKafkaBasedDownstreamSender implements Lifecycle {
      * @throws NullPointerException if any of the parameters are {@code null}.
      */
 
-    public AbstractKafkaBasedDownstreamSender(final KafkaProducerFactory<String, Buffer> producerFactory,
+    public AbstractKafkaBasedMessageSender(final KafkaProducerFactory<String, Buffer> producerFactory,
             final String producerName, final Map<String, String> config, final Tracer tracer) {
 
         Objects.requireNonNull(producerFactory);
--- a/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/telemetry/kafka/AbstractKafkaBasedMessageSender.java
+++ b/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/telemetry/kafka/AbstractKafkaBasedMessageSender.java
@@ -58,11 +58,11 @@ public abstract class AbstractKafkaBasedMessageSender implements Lifecycle {
      * A logger to be shared with subclasses.
      */
     protected final Logger log = LoggerFactory.getLogger(getClass());
+    protected final Tracer tracer;
 
     private final KafkaProducerFactory<String, Buffer> producerFactory;
     private final String producerName;
     private final Map<String, String> config;
-    private final Tracer tracer;
 
     /**
      * Creates a new Kafka-based telemetry sender.
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/CachingKafkaConsumerFactory.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/CachingKafkaConsumerFactory.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.hono.kafka.client;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.BiFunction;
+import java.util.stream.Collectors;
+
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+import io.vertx.core.Vertx;
+import io.vertx.kafka.client.consumer.KafkaConsumer;
+
+/**
+ * TODO.
+ *
+ * @param <K> The type for the record key serialization.
+ * @param <V> The type for the record value serialization.
+ */
+public class CachingKafkaConsumerFactory<K, V> implements KafkaConsumerFactory<K, V> {
+    private final Map<String, KafkaConsumer<K, V>> activeConsumers = new HashMap<>();
+    private final BiFunction<String, Map<String, String>, KafkaConsumer<K, V>> consumerInstanceSupplier;
+
+    /**
+     * Creates a new consumer factory.
+     * <p>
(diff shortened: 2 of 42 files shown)


clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/telemetry/kafka/AbstractKafkaBasedMessageSender.java
```diff
@@ -0,0 +1,288 @@
...
+import io.vertx.kafka.client.producer.KafkaHeader;
+import io.vertx.kafka.client.producer.KafkaProducer;
+import io.vertx.kafka.client.producer.KafkaProducerRecord;
+import io.vertx.kafka.client.producer.RecordMetadata;
+import io.vertx.kafka.client.producer.impl.KafkaHeaderImpl;
+
+/**
+ * A client for publishing messages to a Kafka cluster.
+ */
+public abstract class AbstractKafkaBasedMessageSender implements Lifecycle {
```
b-abel, line 55: The class `AbstractKafkaBasedDownstreamSender` has changed quite significantly since you created this PR. The class cannot simply be used to forward commands (e.g. it sets the headers for downstream messages). I would like to see it refactored so that it can be used more generically, but I'm not sure that there is much to reuse, as the class `CachingKafkaProducerFactory` is the one that does the heavy lifting like reopening failed producer instances.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/CachingKafkaConsumerFactory.java
```diff
@@ -0,0 +1,92 @@
...
+import io.vertx.core.Vertx;
+import io.vertx.kafka.client.consumer.KafkaConsumer;
+
+/**
+ * TODO.
+ *
+ * @param <K> The type for the record key serialization.
+ * @param <V> The type for the record value serialization.
+ */
+public class CachingKafkaConsumerFactory<K, V> implements KafkaConsumerFactory<K, V> {
```
b-abel, line 32: I added a quite generic Kafka consumer for AT LEAST ONCE consumption: `AbstractAtLeastOneKafkaConsumer` from the PR #2420 (it does not require the factory). Would it FYPOV make sense to use it here?

services/command-router/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/KafkaBasedCommandConsumerFactoryImpl.java
```diff
@@ -0,0 +1,240 @@
...
+                .getTargetGatewayAndAdapterInstance(command.getTenantId(), command.getDeviceId(), currentSpan.context())
+                .onSuccess(result -> {
+                    final String targetAdapterInstanceId = result
+                            .getString(DeviceConnectionConstants.FIELD_ADAPTER_INSTANCE_ID);
+                    final String targetDeviceId = result.getString(DeviceConnectionConstants.FIELD_PAYLOAD_DEVICE_ID);
+                    final String gatewayId = targetDeviceId.equals(command.getDeviceId()) ? null : targetDeviceId;
+                    forwardCommand(command, gatewayId, targetAdapterInstanceId, incomingCommandRecord.value(),
+                            currentSpan);
+                    // TODO to explore options for committing manually based on the QoS once the command is
+                    // successfully forwarded
```
b-abel, line 161: You cannot mix AT MOST ONCE and AT LEAST ONCE semantics in one consumer. The consumer either commits the offset *before* (AT MOST ONCE) or *after* (AT LEAST ONCE) processing the message. To be able to decide this based on a property of the message, we would have to fetch each message individually from the broker (opposed to fetching batches) which is not desirable. In other words, you have to decide if messages should be consumed twice or if you accept message loss if the consumer is interrupted or an error occurs. We decided to live with duplicates because the application layer must be able to handle them anyway. Long story short, I would always commit the offsets after processing the message. This is what `AbstractAtLeastOneKafkaConsumer` already provides (see #2420).

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/HonoTopic.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * The type of a Hono specific Kafka topic.
+     */
+    public enum Type {
+
+        TELEMETRY(NAMESPACE + TelemetryConstants.TELEMETRY_ENDPOINT + SEPARATOR),
+        EVENT(NAMESPACE + EventConstants.EVENT_ENDPOINT + SEPARATOR),
+        COMMAND(NAMESPACE + CommandConstants.COMMAND_ENDPOINT + SEPARATOR),
+        COMMAND_INTERNAL(NAMESPACE + CommandConstants.INTERNAL_COMMAND_ENDPOINT + SEPARATOR),
```
b-abel, line 122: I'm not sure Hono's internal topics should be handled in the class `HonoTopic`. Do you see other options?

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/HonoTopic.java
```diff
@@ -0,0 +1,138 @@
...
+     */
+    public static HonoTopic fromString(final String topicString) {
+        if (topicString.startsWith(Type.TELEMETRY.prefix)) {
+            return new HonoTopic(Type.TELEMETRY, topicString.substring(Type.TELEMETRY.prefix.length()));
+        } else if (topicString.startsWith(Type.EVENT.prefix)) {
+            return new HonoTopic(Type.EVENT, topicString.substring(Type.EVENT.prefix.length()));
+        } else if (topicString.startsWith(Type.COMMAND.prefix)) {
+            return new HonoTopic(Type.COMMAND, topicString.substring(Type.COMMAND.prefix.length()));
+        } else if (topicString.startsWith(Type.COMMAND_INTERNAL.prefix)) {
+            return new HonoTopic(Type.COMMAND_INTERNAL, topicString.substring(Type.COMMAND_INTERNAL.prefix.length()));
```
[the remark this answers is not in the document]
kaniyan: This is a workaround that I used for the prototype and I will modify it in the productive version.

kaniyan:
This has been closed in favor of PR #2454.

Issue #2363, opened 2020-12-10
The received command message's topic missing tenant ID


gongzhao2:
Hi, I try to send command to my device, but the message topic I received missing tenant ID, it's just a blank. like below picture.
![image](https://user-images.githubusercontent.com/66044894/101737541-fe276c00-3aff-11eb-8df0-25b865925f34.png)

Why the message.topic I received missing tenant ID, how can I add it to the topic?

Thanks in advance.

sophokles73:
which version of Hono are you using?

gongzhao2:
Hi,
Hono version: 1.2.3
mosquitto version: 1.6.8

sophokles73:
You might want to refer to the [MQTT adapter user guide](https://www.eclipse.org/hono/docs/1.2/user-guide/mqtt-adapter/#command-control), which describes the adapter's behavior for different Command & Control client types. First you will need to figure out, which type matches your use case.

gongzhao2:
I want to **receive commands with authenticated gateway**, that's why the message I received don't have tenant id.
![image](https://user-images.githubusercontent.com/66044894/101791442-7f9def00-3b3e-11eb-8506-790a655b04ef.png)
I want to know why the topic of the command received in this way does not contain tenant-id, because the authentication method of the device already contains the tenant-id?
Can I modify some configurations to make Hono return a complete topic in this way?

sophokles73:
> Can I modify some configurations to make Hono return a complete topic in this way?

No, you can not.

calohmn:
Since you are using an authenticated gateway, you have established the MQTT connection in the context of the tenant given in the username (DEFAULT_TENANT in your example) anyway. So, all the commands that you receive via this connection will be for devices of that particular tenant. That's why there is no need for the tenant to be in the topic to which commands are published for an authenticated gateway.

gongzhao2:
Thanks sophokles73, calohmn.

gongzhao2:
Hi there,
We are using lora network server of Chirpstack as our MQTT client to receive command(authenticated gateway), but the lora network server report an error "topic regex match error" because the topic hono returned is missing "tenant-ID".
![chirpstack_lns_error](https://user-images.githubusercontent.com/66044894/102176117-bdeb3380-3edb-11eb-9e8f-46d5df97ebaf.jpg)
Has the topic format returned by Hono considered the compatibility of other products?
Actually, I tested the mosquitto broker and it will return the full address of the topic.
[...]

PR #2374, opened 2020-12-18
Make inclusion of tenant ID configurable

sophokles73:
This addresses #2363 

The MQTT adapter now supports a tenant specific boolean configuration
property that allows to always include the tenant ID in the topic name
used for publishing command messages.

This can be used to allow protocol gateways to determine a device's
tenant from the topic name.

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandSubscription.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandSubscription.java
@@ -15,6 +15,7 @@ package org.eclipse.hono.adapter.mqtt;
 
 import java.util.Objects;
 
+import org.eclipse.hono.adapter.client.command.Command;
 import org.eclipse.hono.auth.Device;
 import org.eclipse.hono.util.CommandConstants;
 import org.eclipse.hono.util.ResourceIdentifier;
@@ -81,8 +82,16 @@ public final class CommandSubscription {
         this.req = resource.elementAt(3);
 
         this.isAuthenticated = authenticatedDevice != null;
-        this.authenticatedDeviceId = authenticatedDevice != null ? authenticatedDevice.getDeviceId() : null;
-        if (authenticatedDevice == null) {
+
+        if (isAuthenticated) {
+            if (resourceTenant != null && !authenticatedDevice.getTenantId().equals(resourceTenant)) {
+                throw new IllegalArgumentException(
+                        "tenant in topic filter does not match authenticated device");
+            }
+            this.tenant = authenticatedDevice.getTenantId();
+            this.authenticatedDeviceId = authenticatedDevice.getDeviceId();
+            this.deviceId = Strings.isNullOrEmpty(resourceDeviceId) ? authenticatedDevice.getDeviceId() : resourceDeviceId;
+        } else {
             if (Strings.isNullOrEmpty(resourceTenant)) {
                 throw new IllegalArgumentException(
                         "for unauthenticated devices the tenant needs to be given in the subscription");
@@ -92,14 +101,8 @@ public final class CommandSubscription {
                         "for unauthenticated devices the device-id needs to be given in the subscription");
             }
             this.tenant = resourceTenant;
+            this.authenticatedDeviceId = null;
             this.deviceId = resourceDeviceId;
-        } else {
-            if (resourceTenant != null && !authenticatedDevice.getTenantId().equals(resourceTenant)) {
-                throw new IllegalArgumentException(
-                        "tenant in topic filter does not match authenticated device");
-            }
-            this.tenant = authenticatedDevice.getTenantId();
-            this.deviceId = Strings.isNullOrEmpty(resourceDeviceId) ? authenticatedDevice.getDeviceId() : resourceDeviceId;
         }
     }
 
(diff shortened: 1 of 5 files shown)


kaniyan:
This PR for including tenant ID on the topic name while publishing commands looks good to me. For me what is not clear is the topic structure that devices/gateways use to subscribe for receiving commands in combination with this feature. I have summarised the various scenarios in the below table. 
image

Lets us assume that the adapter level configuration `always-include-tenant-id` is set to _true_. Now an authenticated device (Eg: 4711 belongs to DEFAULT_TENANT) has subscribed to a topic `command///req/#`. In this case, the command is sent to the topic `command/DEFAULT_TENANT//q/req-id/commandName`. The device(4711) subscribed to the topic `command///req/#` won't receive that command unless it subscribes to the topic `command/+//req/#` or `command/DEFAULT_TENANT//req/#`.

As per the [doc](https://www.eclipse.org/hono/docs/user-guide/mqtt-adapter/#receiving-commands-authenticated-device), in future, the command topic should start with prefix `command//` and not `command/+/`. Since we allow the tenant ID be included in the topic name, I think the command subscription topic should start with prefix `command/+/`. Otherwise we should explicitly mention in the document, that the devices should subscribe with tenant ID in the topic name if the `always-include-tenant-id` is set to _true_. I think the former one is straightforward. 

@sophokles73 @calohmn WDYT?

sophokles73:
> The device(4711) subscribed to the topic command///req/# won't receive that command unless it subscribes to the topic command/+//req/# or command/DEFAULT_TENANT//req/#

Very good catch :+1: 

I think we can make this even more straightforward by letting the client decide, whether it wants to have the tenant ID included or not by simply including the tenant ID in the topic name when subscribing. i.e. if an authenticated client subscribes using `command///req/#` then the adapter doesn't include the tenant ID in the topic it publishes commands to. But if it does include the tenant ID (`command/DEFAULT_TENANT//req/#`), then the adapter does include the tenant ID in the publish topic. We then do not need the configuration property at all, which makes things even easier ...
WDYT?

calohmn:
I agree that making the decision whether to include the tenant can then better be made using the subscription topic.

We might as well apply this principle for the device id then as well, so that an authenticated device subscribing to `command/DEFAULT_TENANT/4711/req/#` will get its commands published to `command/DEFAULT_TENANT/4711/req/${req-id}/${command}`. The scenario of #2363 with the MQTT client expecting non-empty topic levels might also occur with the client acting as an authenticated device, not as an authenticated gateway.

kaniyan:
> I think we can make this even more straightforward by letting the client decide, whether it wants to have the tenant ID included or not by simply including the tenant ID in the topic name when subscribing. i.e. if an authenticated client subscribes using `command///req/#` then the adapter doesn't include the tenant ID in the topic it publishes commands to. But if it does include the tenant ID (`command/DEFAULT_TENANT//req/#`), then the adapter does include the tenant ID in the publish topic. We then do not need the configuration property at all, which makes things even easier ...

:+1: This makes more sense and also no need to configure any extra property.

kaniyan:
> We might as well apply this principle for the device id then as well, so that an authenticated device subscribing to `command/DEFAULT_TENANT/4711/req/#` will get its commands published to `command/DEFAULT_TENANT/4711/req/${req-id}/${command}`.

I think so, too. It is good to have this principle extended to the device identifier.

sophokles73:
@calohmn @kaniyan I have pushed corresponding changes ...

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandSubscription.java
```diff
@@ -32,13 +33,19 @@
...
+ * <li>{@code c///q/#} - authenticated device using short names</li>
  * </ol>
  */
 public final class CommandSubscription {
 
+    /**
+     * The name of the MQTT adapter configuration property that defines if the tenant ID is
+     * always included in the topic name that is used for publishing a command to a client.
+     */
+    public static final String CONFIG_PROPERTY_ALWAYS_INCLUDE_TENANT_ID = "always-include-tenant-id";
```
calohmn, line 47: Not needed anymore.

sophokles73: true

site/documentation/content/user-guide/mqtt-adapter.md
```diff
@@ -274,25 +274,38 @@ The following variables are used:
 
 ### Receiving Commands (authenticated Device)
 
-An authenticated device MUST use the topic filter `command///req/#` to subscribe to commands.
+An authenticated device MUST use one of the following topic filters for subscribing to commands:
+
+* `command///req/#`
+* `command/${tenant-id}//req/#`
+* `command/${tenant-id}/${device-id}/req/#`
```
calohmn, line 281: Strictly speaking, usage of `command//${device-id}/req/#` would work as well. So, maybe better to just use
`command///req/#`
`command/${tenant-id}/${device-id}/req/#`
noting that in the latter case  either tenant or device id may either be set to the value of the authenticated device or may be kept empty.

site/documentation/content/user-guide/mqtt-adapter.md
```diff
@@ -301,14 +314,45 @@ command///q/1010f8ab0b53-bd96-4d99-9d9c-56b868474a6a/setBrightness
...
 
+The following command can be used to subscribe to commands resulting in command messages being published to a
+topic that includes the tenant ID:
+
+```sh
+mosquitto_sub -v -u 'sensor1@DEFAULT_TENANT' -P hono-secret -t c/DEFAULT_TENANT//q/#
+```
+
+Note the usage of the abbreviated names (`c` and `q` instead of `command` and `req`) the the inclusion of the tenant ID
+in the topic filter.
```
calohmn, line 333: "the the" -> "and the"

kaniyan (approved):
LGTM

calohmn (approved):

PR #2377, opened 2020-12-22
Invoke assertRegistration before creating command consumer

calohmn:


tests/src/test/java/org/eclipse/hono/tests/amqp/CommandAndControlAmqpIT.java
```diff
@@ -849,6 +849,45 @@ public void testSendCommandFailsForCommandNotAcknowledgedByDevice(
...
+                .compose(ok -> connectToAdapter(IntegrationTestSupport.getUsername(deviceId, tenantId), password))
+                .compose(conAck -> {
+                    final Promise<Void> result = Promise.promise();
+                    context.runOnContext(go -> {
+                        final ProtonReceiver recv = connection
+                                .createReceiver(endpointConfig.getSubscriptionAddress(tenantId, otherDeviceId));
+                        recv.setQoS(ProtonQoS.AT_LEAST_ONCE);
+                        recv.openHandler(openResult -> {
+                            ctx.verify(() -> {
+                                assertThat(openResult.failed()).isTrue();
```
sophokles73, line 879: I am not sure if the open handler will be invoked with a failed AsyncResult in any case. However, the closeHandler and/or the detachHandler should contain an error that we can assert.

calohmn: Yes, I think in theory the open handler could succeed with the receiver link remote source being null. I'm not sure whether that can actually happen during a somehow delayed test run.
In any case, I've adapted the tests here accordingly.

sophokles73 (approved):
LGTM

Issue #2376, opened 2020-12-21
Error creating Tenant with Tenant CA information


vasanth-bk:
Hono version: 1.5.0
Environment: Plain Kubernetes cluster

When I try to create a tenant with name "TEST_TENANT" with the following Tenant CA information
```
curl -i -X POST -H "content-type: application/json" --data-binary '[{
   "tenant-id": "TEST_TENANT",
   "enabled": true,
   "trusted-ca": [{
     "subject-dn": "subject=CN=TEST_TENANT,OU=Digital,O=L&T-Nxt Pvt Ltd,ST=TN,C=IN",
     "public-key": "MIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEAt3UMpvxQkr3WdBBh6gvBzngrOzPMqJWPBSUWN3EUMBNciS2T34nXlLfl/ti5ZbejVdUcaOIQZDjHG61I5A5w20k6lohXYonYpMLiY1bLHnojvTc8HWel4FifgFR10jZAjwZ4FmS9kUpYsbGcRRxE0ZwsOgrfMngexBBaheHtN6jnjCKS8iGnHbTjEFw1PBI3Nelc7MXQi4srdyc9O3tAHPRzDEWiQ9XJZwDNhWt3MPGdWj5NdUcn2JT23ORcY1no6BbgbFq+0M8/MTKn3VWuLeCBro8ROiN4hiuzm1Iprq129daU/bUfilvWWeGP3jSV2oU4pBOnVOhcP1euIHcanFYE9/LPgq4kOjwrVrF/2UxR5I3kiYdDc/HM5SVWWa2r2aIHQKRRjIKdq8OkKAY8q9nAVGOv7rmnlKI208NU4sbIyfyS25XHBDOV0oUpEd/2Ef/evpNFJXGw6eAzaM1O+QnNF0xXnjmAp6imMM5xp/YOPfSRltFh7EpKNMp2HV4bMLK1zx8GnhZF97BjKMB7KTYf1TiVgTgvhqLHkaAmSfo/2zPmLVafRFJHZdgn52yN5vWtU9/+xwGdOAyCTxCWrJ+d4H/bMefPn1DVaR+Jhe1L2np5Hqo6AP0IiuL9gS79tP4gFBTy3WDXyzit9KSeXrCJ7hCHYUZouktSRhGUuZkCAwEAAQ=="
   }]
[...]

PR #2378, opened 2020-12-22
Add information on accepted date/time format in the device registry management API

kaniyan:


sophokles73 (commented):
actually it needs to be a string that complies with the format defined in [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6) as indicated by the `date-time` format used for the properties (see http://spec.openapis.org/oas/v3.0.3#dataTypeFormat) ...

kaniyan:
> actually it needs to be a string that complies with the format defined in [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6) as indicated by the `date-time` format used for the properties (see http://spec.openapis.org/oas/v3.0.3#dataTypeFormat) ...

I am not aware that this has been already indicated by the `date-time` format in the OpenAPI spec. Since it is already defined in OpenAPI spec, there is no need to define again and this PR can be closed. WDYT?

sophokles73:
> Since it is already defined in OpenAPI spec, there is no need to define again and this PR can be closed. WDYT?

In general, I agree and that is what I had hoped readers would determine as well. However, it might not be a bad idea to make this more explicit as we have seen in #2376 ...

kaniyan:
@sophokles73 I have updated accordingly and pushed those changes. Would you mind taking an another look?

sophokles73 (approved):
LGTM

PR #2379, opened 2020-12-22
Invoke assertRegistration before sending command to gateway

calohmn:
Only done if gateway has subscribed specifically for the device the command is targeted at.

Not needed if command is forwarded to a gateway that has subscribed to commands for all its devices.
In that case the command is explicitly mapped to the gateway via the DeviceConnection API `getCommandHandlingAdapterInstances(tenant, deviceId, viaGateways, spanContext)` method, using the up-to-date via-gateway information.
(A gateway having subscribed specifically for the device is handled in the DeviceConnection service, as if the device had subscribed directly. The assignment to the gateway is only done in the target adapter instance logic in that case.)

I wanted to add integration tests here, clearing the via-gateways in the device registration data after a command subscription was done. But I don't see a way to circumvent usage of a cached device registration entry in that case (still containing the via-gateway), therefore the test would need to wait for 5min in between for the entry to expire.


sophokles73 (approved):
LGTM

Issue #2381, opened 2020-12-23
Support response caching at service client level


sophokles73:
The new AMQP 1.0 adapter client for accessing Hono's service APIs currently simply wraps the legacy client implementation under the hood. This also means that responses from the services are being cached at the tenant scoped link level. As a consequence, the cache is being removed/cleared if the underlying connection to the service is lost. This is in stark contrast to the overall intention of caching the responses in the first place, which is to be able to return values from the cache until the connection has been re-established.

The new client interfaces all accept the tenant ID as a parameter and therefore allow to cache values for all tenants at the service client level. The (new) client implementations need to be updated accordingly.

PR #2382, opened 2020-12-23
Introduce response caching at Tenant service client level

sophokles73:
This is the first step of #2381 

Responses from the Tenant service are now being stored in a cache that
is independent from the life cycle of the underlying connection to the
service.

### DIFF
--- a/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/amqp/AbstractRequestResponseServiceClient.java
+++ b/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/amqp/AbstractRequestResponseServiceClient.java
@@ -1,0 +1,369 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.hono.adapter.client.amqp;
+
+import java.net.HttpURLConnection;
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Function;
+
+import org.apache.qpid.proton.amqp.messaging.ApplicationProperties;
+import org.apache.qpid.proton.message.Message;
+import org.eclipse.hono.cache.CacheProvider;
+import org.eclipse.hono.cache.ExpiringValueCache;
+import org.eclipse.hono.client.HonoConnection;
+import org.eclipse.hono.client.RequestResponseClientConfigProperties;
+import org.eclipse.hono.client.SendMessageSampler;
+import org.eclipse.hono.client.ServiceInvocationException;
+import org.eclipse.hono.client.StatusCodeMapper;
+import org.eclipse.hono.client.impl.CachingClientFactory;
+import org.eclipse.hono.config.ProtocolAdapterProperties;
+import org.eclipse.hono.tracing.TracingHelper;
+import org.eclipse.hono.util.CacheDirective;
+import org.eclipse.hono.util.MessageHelper;
+import org.eclipse.hono.util.RequestResponseResult;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.opentracing.Span;
+import io.opentracing.tag.Tags;
+import io.vertx.core.Future;
+import io.vertx.core.buffer.Buffer;
+
+/**
+ * A vertx-proton based parent class for the implementation of API clients that follow the request response pattern.
+ * <p>
+ * Provides support for caching response messages from a service in an {@link ExpiringValueCache}.
+ *
+ * @param <R> The type of response this client expects the peer to return.
+ * @param <T> The type of object contained in the peer's response.
+ *
+ */
+public abstract class AbstractRequestResponseServiceClient<T, R extends RequestResponseResult<T>> extends AbstractServiceClient {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractRequestResponseServiceClient.class);
+    private static final int[] CACHEABLE_STATUS_CODES = new int[] {
+                            HttpURLConnection.HTTP_OK,
+                            HttpURLConnection.HTTP_NOT_AUTHORITATIVE,
+                            HttpURLConnection.HTTP_PARTIAL,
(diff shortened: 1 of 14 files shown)


clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/amqp/AbstractRequestResponseServiceClient.java
```diff
@@ -0,0 +1,369 @@
...
+    };
+
+    /**
+     * Contains the AMQP link sender/receiver link pairs for invoking the service.
+     */
+    protected final CachingClientFactory<RequestResponseClient<R>> clientFactory;
+    /**
+     * A cache to use for responses received from the service.
+     */
+    private final ExpiringValueCache<Object, R> responseCache;
```
calohmn, line 75: Not directly related to the changes here, but:
Have you considered letting the same cache instance be used for all client instances, i.e. across verticle instances?
Then we would not only get cached tenant values when a connection is down, but also when the tenant object has already been retrieved via another connection/verticle.
I guess from a thread-safety point of view, this would be ok using CaffeineCache.

sophokles73: In general, that should already be possible because the verticle instances get access to the `ExpiringvalueCache` via a `CacheProvider`. If they are all configured to use the same `CacheProvider` instance, then they should all use the same cache instance as well.

sophokles73:

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/amqp/AbstractRequestResponseServiceClient.java
```diff
@@ -0,0 +1,369 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2020 Contributors to the Eclipse Foundation
```
calohmn, line 2: 2020

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/amqp/AbstractRequestResponseServiceClient.java
```diff
@@ -0,0 +1,369 @@
...
+
+import io.opentracing.Span;
+import io.opentracing.tag.Tags;
+import io.vertx.core.Future;
+import io.vertx.core.buffer.Buffer;
+
+/**
+ * A vertx-proton based parent class for the implementation of API clients that follow the request response pattern.
+ * <p>
+ * Provides access to a {@link CacheProvider} which can be used to create caches for service response messages.
```
calohmn, line 50: This sentence originating from `AbstractRequestResponseClient` doesn't really fit here.

clients/adapter-amqp/src/test/java/org/eclipse/hono/adapter/client/amqp/RequestResponseClientTest.java
```diff
@@ -0,0 +1,379 @@
+/*******************************************************************************
+ * Copyright (c) 2016, 2020 Contributors to the Eclipse Foundation
```
calohmn, line 2: 2020

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/amqp/RequestResponseClient.java
```diff
@@ -0,0 +1,647 @@
...
+        return String.format("%s-client-%s", endpointName, UUID.randomUUID());
+    }
+
+    /**
+     * Creates the sender and receiver links to the peer for sending requests
+     * and receiving responses.
+     *
+     * @return A future indicating the outcome. The future will succeed if the links
+     *         have been created.
+     * @throws NullPointerException if con is {@code null}.
```
calohmn, line 255: There is no con parameter.

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/amqp/RequestResponseClient.java
```diff
@@ -0,0 +1,647 @@
...
+     * @param properties The headers to include in the request message as AMQP application properties.
+     * @param payload The payload to include in the request message as an AMQP Value section.
+     * @param contentType The content type of the payload.
+     * @param responseMapper A function mapping a raw AMQP message to the response type.
+     * @param currentSpan The <em>Opentracing</em> span used to trace the request execution.
+     * @return A future indicating the outcome of the operation.
+     *         The future will be failed with a {@link ServerErrorException} if the request cannot be sent to
+     *         the remote service, e.g. because there is no connection to the service or there are no credits
+     *         available for sending the request or the request timed out.
+     * @throws NullPointerException if any of action, result handler or currentSpan is {@code null}.
```
calohmn, line 424: No `result handler` param here. It looks like the `responseMapper` param mustn't be null instead here.

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/amqp/RequestResponseClient.java
```diff
@@ -0,0 +1,647 @@
...
```
calohmn, line 467: No `result handler` param here. 
It looks like the `responseMapper` param mustn't be null instead here. The corresponding null check is missing below.

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/amqp/RequestResponseClient.java
```diff
@@ -0,0 +1,647 @@
...
+    public final boolean isOpen() {
+        return sender != null && sender.isOpen() && receiver != null && receiver.isOpen();
+    }
+
+    /**
+     * Closes the links.
+     *
+     * @param closeHandler The handler to invoke with the result.
+     */
+    public final void close(final Handler<AsyncResult<Void>> closeHandler) {
```
calohmn, line 637: Could be refactored to return `Future<Void>` instead of taking the handler parameter (which would be in line with `createLinks` for example).

calohmn:
I've added comments about some minor points. Otherwise OK from my side.

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/amqp/AbstractRequestResponseServiceClient.java
sophokles73: I had deliberately kept the origin year because most of the code has been copied over from the legacy client.

clients/adapter-amqp/src/test/java/org/eclipse/hono/adapter/client/amqp/RequestResponseClientTest.java
sophokles73: I had deliberately kept the origin year because most of the code has been copied over from the legacy client.

calohmn (approved):
LGTM

PR #2394, opened 2021-01-06
Use Testcontainers for running MongoDB registry unit tests

sophokles73:
The flapdoodle library used for starting an embedded Mongo DB before
running the unit tests of the Mongo DB based registry doesn't support
Mongo DB versions > 4.0.
The testcontainers framework provides more flexibility in that regard.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-# Copyright (c) 2020 Contributors to the Eclipse Foundation
+# Copyright (c) 2020, 2021 Contributors to the Eclipse Foundation
 #
 # See the NOTICE file(s) distributed with this work for additional
 # information regarding copyright ownership.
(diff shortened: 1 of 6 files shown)


calohmn:
Also allows running the MongoDB tests from the IDE now, not having to be run via maven.

calohmn:
This part:
````
    - name: Cache embedded MongoDB for unit tests
      uses: actions/cache@v2
      with:
        path: ~/.embedmongo/**/*.tgz
        key: ${{ runner.os }}-mongodb-${{ hashFiles('**/pom.xml') }}
        restore-keys: |
          ${{ runner.os }}-mongodb-
````
in `.github/workflows/ci.yml` is obsolete then.

I think it would also be good to explore, whether a docker image (layer) cache be be used instead (e.g. [Docker Layer Caching](https://github.com/marketplace/actions/docker-layer-caching)) because otherwise each complete test run will download 3 * 157MB.

sophokles73:
@calohmn I have removed the cache from the CI build.

> I think it would also be good to explore, whether a docker image (layer) cache be be used instead (e.g. Docker Layer Caching) because otherwise each complete test run will download 3 * 157MB.

I'd rather do this in a separate PR as it also affects other parts of the build.

calohmn (approved):
LGTM

PR #2396, opened 2021-01-06
[#2395] Upgrade to vert.x 3.9.5

sophokles73:
Sadly, this requires addition of RX Java libraries to the test scope as
described here:
https://github.com/vert-x3/wiki/wiki/3.9.5-Deprecations-and-breaking-changes#vertx-junit5-extensions

We could wait for 3.9.6 which supposedly will fix the issue above. However, I would rather take advantage of
the other fixes that come with 3.9.5 in the meantime and remove the additional dependencies again, once
3.9.6 is here.


calohmn:
I think we should also consider updating to Quarkus 1.10.5.Final, which also updates the vert.x dependency to 3.9.5.

sophokles73:
@calohmn we can update Quarkus in another PR, can't we?

calohmn:
sure

calohmn (approved):
LGTM

PR #2397, opened 2021-01-06
Fix usage of embedded cache with default config

calohmn:
Enable using an embedded cache for the device connection / command router service with the default options, without having to use a cache configuration file.


sophokles73 (approved):
LGTM

calohmn:
Also applied to the 1.5.x branch.

PR #2399, opened 2021-01-07
Add own CommandTargetMapper in Command Router

calohmn:
Thereby removing usage of the deprecated `org.eclipse.hono.client.CommandTargetMapper` class.

### DIFF
--- a/services/command-router/src/main/java/org/eclipse/hono/commandrouter/CommandTargetMapper.java
+++ b/services/command-router/src/main/java/org/eclipse/hono/commandrouter/CommandTargetMapper.java
@@ -1,0 +1,109 @@
+/*******************************************************************************
+ * Copyright (c) 2019, 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.commandrouter;
+
+import org.eclipse.hono.adapter.client.registry.DeviceRegistrationClient;
+import org.eclipse.hono.client.ClientErrorException;
+import org.eclipse.hono.client.ServiceInvocationException;
+import org.eclipse.hono.commandrouter.impl.CommandTargetMapperImpl;
+import org.eclipse.hono.deviceconnection.infinispan.client.DeviceConnectionInfo;
+import org.eclipse.hono.util.DeviceConnectionConstants;
+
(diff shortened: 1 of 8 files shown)


services/command-router/src/main/java/org/eclipse/hono/commandrouter/impl/CommandTargetMapperImpl.java
```diff
@@ -0,0 +1,169 @@
...
+     * @param tracer The tracer instance.
+     * @throws NullPointerException if tracer is {@code null}.
+     */
+    public CommandTargetMapperImpl(final Tracer tracer) {
+        this.tracer = Objects.requireNonNull(tracer);
+    }
+
+    @Override
+    public void initialize(final DeviceRegistrationClient registrationClient, final DeviceConnectionInfo deviceConnectionInfo) {
+        this.registrationClient = registrationClient;
```
sophokles73, line 67: these should be checked for `null`, I guess

services/command-router/src/main/java/org/eclipse/hono/commandrouter/impl/CommandTargetMapperImpl.java
```diff
@@ -0,0 +1,169 @@
...
+
+    @Override
+    public final Future<JsonObject> getTargetGatewayAndAdapterInstance(final String tenantId, final String deviceId,
+            final SpanContext context) {
+        Objects.requireNonNull(tenantId);
+        Objects.requireNonNull(deviceId);
+
+        if (!initialized.get()) {
+            LOG.error("not initialized");
+            return Future.failedFuture(new ServerErrorException(HttpURLConnection.HTTP_INTERNAL_ERROR));
```
sophokles73, line 80: FMPOV this calls for an `IllegalStateException` (which will then be mapped to an internal error, I believe)

services/command-router/src/main/java/org/eclipse/hono/commandrouter/CommandTargetMapper.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * A component for determining where a incoming command message should be targeted at when processing the message in a
+ * protocol adapter.
+ * <p>
+ * This refers to finding the <em>protocol adapter instance</em> that a device or gateway has connected to in order to
+ * receive a command message. Furthermore, it is determined to which <em>gateway</em> a command is to be mapped to or
+ * whether it will be sent to the target device directly.
+ * <p>
+ * For obtaining both information, the <em>getCommandHandlingAdapterInstances</em> operation of the Device Connection
```
sophokles73, line 36: isn't that an implementation detail from now on? I was under the impression that we are replacing the Device Connection API with the Command Router API ...

calohmn: Yes, good point. I've changed the javadoc now.

sophokles73 (approved):
LGTM

PR #2401, opened 2021-01-08
[#8] Fix display of Kafka-based API specifications.

b-abel:
This fixes a  404 in https://www.eclipse.org/hono/docs/api/kafka/ (resp. https://www.eclipse.org/hono/docs/dev/api/kafka/).

The content folder "kafka" needed an `_index.md` (chapter) file. I found no way to hide the chapter page from the menu while showing the child pages. This leads to a new level of pages in the menu which looks OK FMPOV.


site/documentation/content/api/kafka/_index.md
```diff
@@ -0,0 +1,9 @@
++++
+title="Kafka-based APIs"
+weight = 416
+chapter = true
++++
+
+# Kafka-based APIs
```
sophokles73, line 7: why don't we simply rename the `api/kafka/telemetry` folder to `api/telemetry-kafka` and the `api/kafka/event` folder to `api/event-kafka` and update the links from the Event API to the Telemetry API accordingly? Then all documents are at the same level ...

b-abel: FMPOV, it's a good thing to have a central URL for the Kafka-based APIs that we can reference. WDYT?

sophokles73: well, we don't have one for the AMQP based ones either, do we?

b-abel: > well, we don't have one for the AMQP based ones either, do we?

It's the base URL because everything is tangled with AMQP ;)

But having all APIs on one level is also OK for me. If you prefer, I'll update the PR.

sophokles73: Until we all talk about and agree on another structure, I do.
Maybe you want to create an issue which we can use to discuss an alternative structure?

b-abel: > Maybe you want to create an issue which we can use to discuss an alternative structure?

I would definitively not touch the existing (AMQP-based) API specifications, because that would require to update a lot of internal links and add redirects for each page (so that external links do not break). 
Whether we have the 3 affected APIs (event, telemetry, and C&C) on one level next to each other or arrange them hierarchically is not a big difference in my eyes. If you want I'll create a ticket, but I don't think there are many arguments pro and con. I would rather adapt the PR here according to your suggestion.

sophokles73: ok, then go ahead ...

b-abel:
Now I moved the Kafka-based APIs to the same level as the others.

sophokles73 (approved):
LGTM

PR #2404, opened 2021-01-11
[#2273] Add Command & Control API for Kafka spec

calohmn:
This is for #2273:
Adds the specification for the Command & Control API for Kafka.

One of the main differences to the AMQP spec is that command response messages aren't targeted at a resource containing a "replyId" (like `command_response/${tenant_id}/${reply_id}`) but instead are written into the more general `hono.command_response.${tenant_id}` Kafka topic.

This means that no `reply-to` property/header is needed for the command message. Instead, to indicate that a command is supposed to trigger a response, ~~a `expect-response` header with a boolean value is used~~ the existence of the `correlation-id` header is used.

### DIFF
--- a/site/documentation/content/api/command-and-control-kafka/index.md
+++ b/site/documentation/content/api/command-and-control-kafka/index.md
@@ -1,0 +1,120 @@
+---
+title: "Command & Control API for Kafka Specification"
+linkTitle: "Command & Control API for Kafka"
+weight: 419
+---
+
+The *Command & Control* API of Eclipse Hono&trade; is used by *Business Applications* to send commands to connected devices.
+
+Commands can be used to trigger actions on devices. Examples include updating a configuration property, installing a software component or switching the state of an actuator.
+<!--more-->
+
+Hono distinguishes two types of commands. The first type is *one-way* only. In this case the sender of the command does not expect the device to send back a message in response to the command. This type of command is referred to as a *one-way command* in the remainder of this page. One-way commands may be used to e.g. *notify* a device about a change of state.
+
+The second type of commands expects a *response* to be sent back from the device as a result of processing the command. In this case the response contains a *status* code which indicates whether the command could be processed successfully. If so, the response may also include data representing the result of processing the command. This type of command is plainly referred to as a *command* because it represents the default case.
+
+The Command & Control API for Kafka is an alternative to the [Command & Control API for AMQP]({{< relref "/api/command-and-control" >}}).
+With this API, clients send command messages to an Apache Kafka&reg; cluster instead of an AMQP Messaging Network.
+
+See [Kafka-based APIs]({{< relref "/api/kafka-api" >}}) for fundamental information about Hono's Kafka-based APIs.
+The statements there apply to this specification.
+
+{{% note title="Tech preview" %}}
+The support of Kafka as a messaging system is currently a preview and not yet ready for production. The APIs are subject to change without prior notice.
+{{% /note %}}
+
+## Send a One-Way Command
+
+Business Applications use this operation to send a command to a device for which they do not expect to receive a response from the device.
+For that, the Business Application connects to the *Kafka Cluster* and writes a message to the tenant-specific topic `hono.command.${tenant_id}` where `${tenant_id}` is the ID of the tenant that the client wants to send the command for.
+
+**Preconditions**
+
+1. Either the topic `hono.command.${tenant_id}` exists, or the broker is configured to automatically create topics on demand.
+1. The *Business Application* is authorized to write to the topic.
+
+**Message Flow**
+
+1. The *Business Application* writes a command message to the topic `hono.command.${tenant_id}` on the *Kafka Cluster*.
+1. Hono consumes the message from the *Kafka Cluster* and forwards it to the device, provided that the target device is connected and is accepting commands.
+
+**Message Format**
+
+The key of the message MUST be the ID of the device that the command is targeted at.
+
+Metadata MUST be set as Kafka headers on a message.
+The following table provides an overview of the headers the *Business Application* needs to set on a one-way command message.
+
+| Name               | Mandatory | Type      | Description |
+| :----------------- | :-------: | :-------- | :---------- |
+| *device_id*        | yes       | *string*  | The identifier of the device that the command is targeted at. |
+| *subject*          | yes       | *string*  | The name of the command to be executed by the device. |
+| *content-type*     | no        | *string*  | If present, MUST contain a *Media Type* as defined by [RFC 2046](https://tools.ietf.org/html/rfc2046) which describes the semantics and format of the command's input data contained in the message payload. However, not all protocol adapters will support this property as not all transport protocols provide means to convey this information, e.g. MQTT 3.1.1 has no notion of message headers. |
+
+The command message MAY contain arbitrary payload, set as message value, to be sent to the device. The value of the message's *subject* header may provide a hint to the device regarding the format, encoding and semantics of the payload data.
+
+## Send a (Request/Response) Command
+
+*Business Applications* use this operation to send a command to a device for which they expect the device to send back a response.
+For that, the Business Application connects to the *Kafka Cluster* and writes a message to the tenant-specific topic `hono.command.${tenant_id}` where `${tenant_id}` is the ID of the tenant that the client wants to send the command for.
+The Business Application can consume the corresponding command response from the `hono.command_response.${tenant_id}` topic.
+
+In contrast to a one-way command, a request/response command contains a *response-expected* header with value `true` and a *correlation-id* header, providing the identifier that is used to correlate a response message to the original request.
+
+**Preconditions**
+
+1. Either the topics `hono.command.${tenant_id}` and `hono.command_response.${tenant_id}` exist, or the broker is configured to automatically create topics on demand.
+1. The *Business Application* is authorized to write to the `hono.command.${tenant_id}` topic and read from the `hono.command_response.${tenant_id}` topic.
+1. The *Business Application* is subscribed to the `hono.command_response.${tenant_id}` topic with a Kafka consumer.
+
+**Message Flow**
+
+1. The *Business Application* writes a command message to the topic `hono.command.${tenant_id}` on the *Kafka Cluster*.
+1. Hono consumes the message from the *Kafka Cluster* and forwards it to the device, provided that the target device is connected and is accepting commands.
+1. The device sends a command response message. Hono writes that message to the `hono.command_response.${tenant_id}` topic on the *Kafka Cluster*.
+1. The *Business Application* consumes the command response message from the `hono.command_response.${tenant_id}` topic.
+
+**Command Message Format**
+
+The key of the message MUST be the ID of the device that the command is targeted at.
+
+Metadata MUST be set as Kafka headers on a message.
+The following table provides an overview of the headers the *Business Application* needs to set on a request/response command message.
+
+| Name                | Mandatory | Type      | Description |
+| :------------------ | :-------: | :-------- | :---------- |
+| *correlation-id*    | yes       | *string*  | The identifier used to correlate a response message to the original request. It is used as the *correlation-id* header in the response. |
+| *device_id*         | yes       | *string*  | The identifier of the device that the command is targeted at. |
+| *response-expected* | yes       | *boolean* | MUST be set with a value of `true`, meaning that a response from the device is expected for the command. |
+| *subject*           | yes       | *string*  | The name of the command to be executed by the device. |
+| *content-type*      | no        | *string*  | If present, MUST contain a *Media Type* as defined by [RFC 2046](https://tools.ietf.org/html/rfc2046) which describes the semantics and format of the command's input data contained in the message payload. However, not all protocol adapters will support this property as not all transport protocols provide means to convey this information, e.g. MQTT 3.1.1 has no notion of message headers. |
+
+The command message MAY contain arbitrary payload, set as message value, to be sent to the device. The value of the message's *subject* header may provide a hint to the device regarding the format, encoding and semantics of the payload data.
+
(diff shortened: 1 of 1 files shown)


site/documentation/content/api/command-and-control-kafka/index.md
```diff
@@ -0,0 +1,117 @@
...
+With this API, clients publish command messages to an Apache Kafka&reg; cluster instead of an AMQP Messaging Network.
+
+{{% note title="Tech preview" %}}
+The support of Kafka as a messaging system is currently a preview and not yet ready for production. The APIs may change with the next version.
+{{% /note %}}
+
+## Send a One-Way Command
+
+Business Applications use this operation to send a command to a device for which they do not expect to receive a response from the device.
+For that, the Business Application writes a message to the tenant-specific topic `hono.command.${tenant_id}` where `${tenant_id}` is the ID of the tenant that the client wants to send the command for.
```
sophokles73, line 26: is this really called *writing to a topic* in the Kafka universe? I could also imagine
> ... publishes a message to the tenant-specific topic ...

site/documentation/content/api/command-and-control-kafka/index.md
```diff
@@ -0,0 +1,117 @@
...
+
+**Preconditions**
+
+1. Either the topic `hono.command.${tenant_id}` exists, or the broker is configured to automatically create topics on demand.
+1. The *Business Application* is authorized to write to the topic.
+
+**Message Flow**
+
+1. The *Business Application* produces a command message to the *Kafka Cluster* on the topic `hono.command.${tenant_id}`.
```
sophokles73, line 35: publishes? IMHO we should use consistent terminology throughout the document ...

site/documentation/content/api/command-and-control-kafka/index.md
```diff
@@ -0,0 +1,117 @@
...
+1. Hono consumes the message from the *Kafka Cluster* and forwards it to the device, provided that device is connected and accepting commands.
```
sophokles73, line 36: > provided that the target device is connected and accepts commands,

site/documentation/content/api/command-and-control-kafka/index.md
```diff
@@ -0,0 +1,117 @@
...
+
+Metadata MUST be set as Kafka headers on a message.
+The following table provides an overview of the headers the *Business Application* needs to set on a one-way command message.
+
+| Name               | Mandatory | Type      | Description |
+| :----------------- | :-------: | :-------- | :---------- |
+| *device_id*        | yes       | *string*  | The identifier of the device that the command is targeted at. |
+| *subject*          | yes       | *string*  | The name of the command to be executed by the device. |
+| *content-type*     | no        | *string*  | If present, MUST contain a *Media Type* as defined by [RFC 2046](https://tools.ietf.org/html/rfc2046) which describes the semantics and format of the command's input data contained in the message payload. However, not all protocol adapters will support this property as not all transport protocols provide means to convey this information, e.g. MQTT 3.1.1 has no notion of message headers. |
+| *message-id*       | no        | *string*  | An identifier that uniquely identifies the message at the sender side. |
```
sophokles73, line 50: what is this good for?

site/documentation/content/api/command-and-control-kafka/index.md
```diff
@@ -0,0 +1,117 @@
...
+
+*Business Applications* use this operation to send a command to a device for which they expect the device to send back a response.
+For that, the Business Application writes a message to the tenant-specific topic `hono.command.${tenant_id}` where `${tenant_id}` is the ID of the tenant that the client wants to send the command for.
+The Business Application can consume the corresponding command response from the `hono.command_response.${tenant_id}` topic.
+
+**Preconditions**
+
+1. Either the topics `hono.command.${tenant_id}` and `hono.command_response.${tenant_id}` exist, or the broker is configured to automatically create topics on demand.
+1. The *Business Application* is authorized to write to the `hono.command.${tenant_id}` topic and read from the `hono.command_response.${tenant_id}` topic.
+1. The *Business Application* subscribes to the `hono.command_response.${tenant_id}` topic with a Kafka consumer.
```
sophokles73, line 64: ... is subscribed to the ...

site/documentation/content/api/command-and-control-kafka/index.md
```diff
@@ -0,0 +1,117 @@
...
+
+**Message Flow**
+
+1. The *Business Application* produces a command message to the *Kafka Cluster* on the topic `hono.command.${tenant_id}`.
```
sophokles73, line 68: terminology

site/documentation/content/api/command-and-control-kafka/index.md
```diff
@@ -0,0 +1,117 @@
...
+1. Hono consumes the message from the *Kafka Cluster* and forwards it to the device, provided that device is connected and accepting commands.
```
sophokles73, line 69: see above

site/documentation/content/api/command-and-control-kafka/index.md
```diff
@@ -0,0 +1,117 @@
...
+1. The device sends a command response message. Hono writes that message into the `hono.command_response.${tenant_id}` topic on the *Kafka Cluster*.
```
sophokles73, line 70: terminology

site/documentation/content/api/command-and-control-kafka/index.md
```diff
@@ -0,0 +1,117 @@
...
+The key of the message MUST be the ID of the device that the command is targeted at.
+
+Metadata MUST be set as Kafka headers on a message.
+The following table provides an overview of the headers the *Business Application* needs to set on a request/response command message.
+
+| Name               | Mandatory | Type      | Description |
+| :----------------- | :-------: | :-------- | :---------- |
+| *device_id*        | yes       | *string*  | The identifier of the device that the command is targeted at. |
+| *subject*          | yes       | *string*  | The name of the command to be executed by the device. |
+| *expect-response*  | yes       | *boolean* | MUST be set with a value of `true` for a command for which a response from the device is expected. |
```
sophokles73, line 84: why don't we simply use the presence of the *correlation-id* for this purpose?

site/documentation/content/api/command-and-control-kafka/index.md
```diff
@@ -0,0 +1,117 @@
...
+| *content-type*     | no        | *string*  | If present, MUST contain a *Media Type* as defined by [RFC 2046](https://tools.ietf.org/html/rfc2046) which describes the semantics and format of the command's input data contained in the message payload. However, not all protocol adapters will support this property as not all transport protocols provide means to convey this information, e.g. MQTT 3.1.1 has no notion of message headers. |
+| *correlation-id*   | no        | *string*  | MAY contain an ID used to correlate a response message to the original request. If set, it is used as the *correlation-id* header in the response, otherwise the value of the *message-id* header is used. Either this or the *message-id* header MUST be set. |
+| *message-id*       | no        | *string*  | MAY contain an identifier that uniquely identifies the message at the sender side. Either this or the *correlation-id* header MUST be set. |
```
sophokles73, line 87: do we need this? why don't we simply require the use of *correlation-id*?

site/documentation/content/api/command-and-control-kafka/index.md
calohmn: I first I thought it would be better to not just use the `correlation-id` (which could be routinely set for all messages by the northbound application) but instead make it more explicit if actually a response is expected.

But looking more closely here, marking the command as a request/response command doesn't have much of an impact anyway. It's mostly used for the command metrics.
So, yes, I'll remove the `expect-response` header.

site/documentation/content/api/command-and-control-kafka/index.md
calohmn: Right, for AMQP it might make sense to have the `message-id` fallback since that is a system-property there, but not so for Kafka, since it's another custom header there. I'll remove it.

site/documentation/content/api/command-and-control-kafka/index.md
calohmn: The Kafka quickstart uses "write some events into the topic", and also "events being written to the topic". We are also using the phrase in the Kafka telemetry/event API specs.

So, I'd suggest using "writing to a topic" and removing the "publish/produce a message to a topic" variants.

site/documentation/content/api/command-and-control-kafka/index.md
calohmn: I've adapted the sentence.

calohmn:
@sophokles73 I've committed changes addressing the comments above.

site/documentation/content/api/command-and-control-kafka/index.md
```diff
@@ -0,0 +1,116 @@
...
+The key of the message MUST be the ID of the device that the command response is from.
+
+Metadata MUST be set as Kafka headers on a message.
+The following table provides an overview of the headers set on a message sent in response to a command.
+
+| Name               | Mandatory | Type      | Description |
+| :----------------- | :-------: | :-------- | :---------- |
+| *device_id*        | yes       | *string*  | The identifier of the device that sent the response. |
+| *status*           | yes       | *integer* | MUST indicate the status of the execution. See table below for possible values. |
+| *correlation-id*   | yes       | *string*  | MUST contain the correlation ID used to match the command message with the response message containing the result of execution on the device. |
```
sophokles73, line 103: how about
> MUST contain the value of the *correlation-id* header of the request message that this is the response for.

calohmn: Changed accordingly.

sophokles73 (approved):
LGTM

site/documentation/content/api/command-and-control-kafka/index.md
```diff
@@ -0,0 +1,116 @@
...
+
+The second type of commands expects a *response* to be sent back from the device as a result of processing the command. In this case the response contains a *status* code which indicates whether the command could be processed successfully. If so, the response may also include data representing the result of processing the command. This type of command is plainly referred to as a *command* because it represents the default case.
+
+The Command & Control API for Kafka is an alternative to the [Command & Control API for AMQP]({{< relref "/api/command-and-control" >}}).
+With this API, clients send command messages to an Apache Kafka&reg; cluster instead of an AMQP Messaging Network.
+
+{{% note title="Tech preview" %}}
+The support of Kafka as a messaging system is currently a preview and not yet ready for production. The APIs may change with the next version.
```
b-abel, line 20: In the specification of the Kafka-based downstream APIs the phrasing of the last sentence is now "The APIs are subject to change without prior notice". You may want to use it here as well.

site/documentation/content/api/command-and-control-kafka/index.md
```diff
@@ -0,0 +1,116 @@
...
+<!--more-->
+
+Hono distinguishes two types of commands. The first type is *one-way* only. In this case the sender of the command does not expect the device to send back a message in response to the command. This type of command is referred to as a *one-way command* in the remainder of this page. One-way commands may be used to e.g. *notify* a device about a change of state.
+
+The second type of commands expects a *response* to be sent back from the device as a result of processing the command. In this case the response contains a *status* code which indicates whether the command could be processed successfully. If so, the response may also include data representing the result of processing the command. This type of command is plainly referred to as a *command* because it represents the default case.
+
+The Command & Control API for Kafka is an alternative to the [Command & Control API for AMQP]({{< relref "/api/command-and-control" >}}).
+With this API, clients send command messages to an Apache Kafka&reg; cluster instead of an AMQP Messaging Network.
```
b-abel, line 17: General explanations about Hono's Kafka-based APIs are now on a separate page. FMPOV it would make sense to add a reference to it like the one in the downstream APIs:

```
See [Kafka-based APIs]({{< relref "/api/kafka-api" >}}) for fundamental information about Hono's Kafka-based APIs.
The statements there apply to this specification.
```

site/documentation/content/api/command-and-control-kafka/index.md
calohmn: @sophokles73 I've revisited this topic here and now think, that we should indeed introduce some distinguishing property other than the `correlation-id`.

The `correlation-id` property is something that should possibly correlate a one-way command message with the previous telemetry/event `ttd` message that triggered the command. That would be the straightforward way to implement a device request/response communication pattern (device to application/cloud - see #1276).

Therefore I would want to have a separate boolean property, making it explicit that a command-response is expected.

I would adapt the name slightly and choose `response-expected`.

Other options:
`response-required` - doesn't fit that well because it would imply that a timeout/error is thrown if no response is received (which isn't the case)
`one-way` - omitting the property would mean the request/response scheme as default, that would be different from AMQP, where the default without added property is one-way

site/documentation/content/api/command-and-control-kafka/index.md
calohmn: I've changed it accordingly.

site/documentation/content/api/command-and-control-kafka/index.md
calohmn: I've added that.

site/documentation/content/api/command-and-control-kafka/index.md
sophokles73: using `response-expected` sounds good to me

calohmn: Ok, I've added that.
I've also added a commit to hide the menu entry for now. When it is clear that the spec implementation will be in the next Hono release, that commit can be reverted.

sophokles73: Do you really want to make it mandatory? We could also make it optional and let it default to `false` (saving a few bytes in the message) ...

calohmn: It's marked as mandatory in the table above because that is in the "Send a (Request/Response) Command" chapter.
In the headers table for the "Send a One-Way Command" chapter, the header isn't listed.

So, one-way commands do not have to have a response-expected=false header.

A combined headers table for both kinds of commands would make this more straightforward, I guess. But with the structure of having separate one-way and request-response chapters (which makes sense to me, and is also used in the AMQP C&C spec), this would look like duplicate information.

PR #2405, opened 2021-01-11
[#2170] Support upload of client certificate for X.509 credentials

### LINKED ISSUE #2170: Registry should support upload of client certificate for X.509 credentials type
The Device Registry Management API currently requires clients to register the subject DN of a device's client certificate as the authentication identifier of X.509 credentials. When doing so, the registry requires the client to use the string format defined in RFC 2253. However, it is not really straight forward to produce from a given X.509 certificate. In fact, it requires using the openssl command line tool with multiple switches etc. Things get worse when a non-standard attribute type like *emailAddress* is part of the subject DN. This often leads to erroneous configuration of the device's credentials. 

In order to make registration of X.509 credentials easier for clients, the management API should accept a (base64 encoded) client certificate instead of the subject DN and then extract the public key's subject DN from the certificate directly. Thus, preventing the client from doing the RFC 2253 compliant formatting.

sophokles73:
The Device Registry Management API's "update credentials" operation has
been extended to allow specifying the auth-id and validity period
implicitly by means of including a (Base64 encoded) client certificate
in the new "cert" property. This can be used instead of specifying the
client certificate's subject DN and public key's validity period
explicitly in the "auth-id" and "secrets" properties. This should make
setting the correct "auth-id" value much less error prone.

Fixes #2170

### DIFF
--- a/services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/X509CertificateCredential.java
+++ b/services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/X509CertificateCredential.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019, 2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2019, 2021 Contributors to the Eclipse Foundation
  *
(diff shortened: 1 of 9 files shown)


dejanb (approved):
Looks good

PR #2406, opened 2021-01-11
[#2053] Adding documentation for gateway-based auto provisioning.

fkaltner:


site/documentation/content/api/event/index.md
```diff
@@ -88,10 +88,14 @@ The AMQP 1.0 properties an event sender needs to set for an *empty notification*
...
+| :------------------------- | :--------------: | :----------------------- | :-------- | :---------- |
+| *content-type*             | yes              | *properties*             | *symbol*  | MUST be set to *application/vnd.eclipse-hono-empty-notification* |
+| *ttd*                      | no               | *application-properties* | *int*     | The *time 'til disconnect* as described in the [*Telemetry API*]({{< relref "/api/telemetry" >}}). |
+| *hono_registration_status* | no               | *application-properties* | *string*  | If set to `NEW` this indicates that a device was auto-provisioned in the course of sending telemetry data via gateway. |
+| *tenant_id*                | no               | *application-properties* | *string*  | The tenant id denoting the tenant for which the device was auto-provisioned. |
+| *device_id*                | no               | *application-properties* | *string*  | The device id denoting of the auto-provisioned device. |
+| *gateway_id*               | no               | *application-properties* | *string*  | The gateway id denoting the gateway for which the device was auto-provisioned. |
```
sophokles73, line 98: Is there a particular reason why we are (ab)using the empty notification for reporting the provisioning of a device? Why don't we use a specific event with its own content-type for this purpose. Wouldn't that make processing of the event easier?

fkaltner: Well, I guess at the time we discussed we thought that it is an acceptable approach (see: https://github.com/eclipse/hono/pull/2094#issuecomment-668473083).

I didn't hear any concerns from the Ditto team (which is the only client of this event sub type to my knowledge) regarding the complexity of the processing of the event. Nevertheless it is right that clients would save one conditional if we had used a dedicated event.

Nevertheless I don't think that it is significantly more complex with the empty event...

sophokles73: > Well, I guess at the time we discussed we thought that it is an acceptable approach (see: #2094 (comment)).

Now I see the problem. There seems to have been a misunderstanding on my side regarding what you were planning to do. My understanding was that we will add a *new type of* event which will have no payload, hence an *empty event* but with its own *content-type*, hence my answer
> However, we should add this particular event to the Well known events ...

FMPOV the chosen approach is inconsistent with the way we have designed the *well known events*, which is to indicate the purpose of the event by means of the content-type and then define the properties according to the purpose.
Now we re doing it the other way around: by inspecting the presence/absence of particular properties, we need to deduce the purpose of the event :-(. I would like this to be changed so that we have a dedicated event (type) for indicating that a device has been provisioned. But we should do this in another PR ... I have created #2408 for tracking this.

fkaltner: > Now I see the problem. There seems to have been a misunderstanding on my side regarding what you were planning to do. My understanding was that we will add a new type of event which will have no payload, hence an empty event but with its own content-type, hence my answer

Sorry for that 😞, it would have been better if I had written "Empty Notification" instead but I mixed that one up with "Empty Event"...

sophokles73: No worries, we'll get it fixed ...

site/documentation/content/concepts/device-provisioning/index.md
```diff
@@ -83,3 +81,33 @@ The newly created credentials are returned to the protocol adapter in the respon
...
+Refer to the [Management API]({{< ref "/api/management" >}}) for creating or updating a 
+corresponding device.
+
+##### Step 2: Connect an unregistered Device via Gateway to Hono
+
+{{< figure src="auto-provisioning-gateway.svg" alt="A unregistered device connects to a gateway which uses the 
+Device Registration API to provision it" title="Automatic Provisioning of a Device via gateway" >}}
+
+The yet unregistered edge device sends telemetry data via a gateway (1) to a protocol adapter which then checks if the 
+edge device is already registered by calling the "assert" operation of Device Registration API (2). If the device
```
sophokles73, line 101: we do not use quotes for referring to names but use *italics* instead, i.e.
> ... by calling the *assert* operation of the Device Registration API (2).

site/documentation/content/concepts/device-provisioning/index.md
```diff
@@ -83,3 +81,33 @@ The newly created credentials are returned to the protocol adapter in the respon
...
+
+{{< figure src="auto-provisioning-gateway.svg" alt="A unregistered device connects to a gateway which uses the 
+Device Registration API to provision it" title="Automatic Provisioning of a Device via gateway" >}}
+
+The yet unregistered edge device sends telemetry data via a gateway (1) to a protocol adapter which then checks if the 
+edge device is already registered by calling the "assert" operation of Device Registration API (2). If the device 
+registration service finds that the device isn't registered yet and the gateway has the `auto-provisioning-enabled` 
+authority set, it creates the edge device (3). 
+
+Subsequently, after it made sure that it hasn't already done so, it sends an empty event
```
sophokles73, line 105: > ..., it sends an [*Empty Notification*]({{< ref "/api/event##empty-notification" >}}) with the ...

site/documentation/content/concepts/device-provisioning/index.md
```diff
@@ -83,3 +81,33 @@ The newly created credentials are returned to the protocol adapter in the respon
...
+Device Registration API to provision it" title="Automatic Provisioning of a Device via gateway" >}}
+
+The yet unregistered edge device sends telemetry data via a gateway (1) to a protocol adapter which then checks if the 
+edge device is already registered by calling the "assert" operation of Device Registration API (2). If the device 
+registration service finds that the device isn't registered yet and the gateway has the `auto-provisioning-enabled` 
+authority set, it creates the edge device (3). 
+
+Subsequently, after it made sure that it hasn't already done so, it sends an empty event 
+(see [Event API]({{< ref "/api/event" >}})) with the `hono_registration_status` application property being set to `NEW` 
+to the AMQP network (4). Upon receiving the disposition for the sent event (5), it marks the event as having been
```
sophokles73, line 107: The disposition is the generic mechanism for conveying the outcome of the transfer which might also be `rejected` or `released`. Thus, IMHO we should change this to
> Once the event has been accepted by the peer (5), the registration service marks ...

site/documentation/content/concepts/device-provisioning/index.md
```diff
@@ -83,3 +81,33 @@ The newly created credentials are returned to the protocol adapter in the respon
...
+
+The yet unregistered edge device sends telemetry data via a gateway (1) to a protocol adapter which then checks if the 
+edge device is already registered by calling the "assert" operation of Device Registration API (2). If the device 
+registration service finds that the device isn't registered yet and the gateway has the `auto-provisioning-enabled` 
+authority set, it creates the edge device (3). 
+
+Subsequently, after it made sure that it hasn't already done so, it sends an empty event 
+(see [Event API]({{< ref "/api/event" >}})) with the `hono_registration_status` application property being set to `NEW` 
+to the AMQP network (4). Upon receiving the disposition for the sent event (5), it marks the event as having been 
+delivered (6) in order to prevent sending duplicate events
```
sophokles73, line 108: this is not so much about avoiding duplicates but about making sure that we send the event at least once, e.g. when the registration service has crashed before it could mark the event as delivered, right?

fkaltner: Yes, I rephrased this as it is more about making sure the Empty Notification is sent at least once. A crash of the registration service in the very moment it starts to send this event would be one possible case being mitigated by the persistent flag as it would re-send the notification when the next telemetry is received.

sophokles73 (approved):
LGTM

PR #2410, opened 2021-01-13
[#2273] Add KafkaBasedCommand class

calohmn:
This is for #2273:
Also improve Javadocs of Command related classes and consolidate replyToId handling in Commands class.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/Command.java
+++ b/client/src/main/java/org/eclipse/hono/client/Command.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2021 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -72,7 +72,7 @@ public final class Command {
     }
 
     /**
-     * Creates a command for an AMQP 1.0 message that should be sent to a device.
+     * Creates a command from an AMQP 1.0 message.
      * <p>
      * The message is expected to contain
      * <ul>
@@ -305,9 +305,12 @@ public final class Command {
 
     /**
      * Gets the request identifier of this command.
+     * <p>
+     * May be {@code null} for a one-way command.
      *
      * @return The identifier or {@code null} if not set.
      * @throws IllegalStateException if this command is invalid.
+     * @see #getRequestId(String, String, String)
      */
     public String getRequestId() {
         if (isValid()) {
@@ -355,12 +358,13 @@ public final class Command {
     }
 
     /**
-     * Gets this command's reply-to-id as given in the incoming command message.
+     * Gets this command's reply-to-id. It is the last part of the command message's <em>reply-to</em> property
+     * value {@code command_response/${tenant_id}/${reply_id}}.
      * <p>
      * Note that an outgoing command message targeted at the device will contain an
      * adapted reply-to address containing the device id.
      *
-     * @return The identifier.
+     * @return The identifier or {@code null} if not set (meaning the command is a one-way command).
      * @throws IllegalStateException if this command is invalid.
      */
     public String getReplyToId() {
@@ -418,13 +422,17 @@ public final class Command {
             throw new IllegalStateException("command is invalid");
         }
     }
+
     /**
      * Creates a request ID for a command.
+     * <p>
+     * Incorporates the given correlationId and replyToId (minus deviceId if contained in the replyToId).
      *
      * @param correlationId The identifier to use for correlating the response with the request.
      * @param replyToId An arbitrary identifier to encode into the request ID.
      * @param deviceId The target of the command.
      * @return The request identifier or {@code null} if correlationId or deviceId is {@code null}.
+     * @deprecated Use {@code org.eclipse.hono.adapter.client.command.Commands#getRequestId(String, String, String)} instead.
      */
     public static String getRequestId(final String correlationId, final String replyToId, final String deviceId) {
 
@@ -491,10 +499,9 @@ public final class Command {
     /**
      * Gets the reply-to-id that will be set when forwarding the command to the device.
      * <p>
-     * It is ensured that this id starts with the device id.
+     * It is ensured that this id starts with {@code ${deviceId}/}.
      *
-     * @param replyToId The reply-to-id as extracted from the 'reply-to' of the command AMQP message. Potentially not
-     *            containing the device id.
+     * @param replyToId The reply-to-id as extracted from the 'reply-to' of the command AMQP message.
      * @param deviceId The device id.
      * @return The reply-to-id, starting with the device id.
      */
--- a/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommand.java
+++ b/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommand.java
@@ -1,0 +1,261 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.adapter.client.command.kafka;
+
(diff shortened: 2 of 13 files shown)


client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -355,12 +358,13 @@ public String getContentType() {
     }
 
     /**
-     * Gets this command's reply-to-id as given in the incoming command message.
+     * Gets this command's reply-to-id. It is the last part of the command message's <em>reply-to</em> property
+     * value "command_response/${tenant_id}/${reply_id}".
```
sophokles73, line 362: how about using `{@code ... }` instead of quotes

client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -491,10 +499,9 @@ public String toString() {
     /**
      * Gets the reply-to-id that will be set when forwarding the command to the device.
      * <p>
-     * It is ensured that this id starts with the device id.
+     * It is ensured that this id starts with "${deviceId}/".
```
sophokles73, line 502: How about
> The returned identifier always start with {@code ${deviceId}/}.

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommand.java
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    /**
+     * Creates a command from a Kafka consumer record.
+     * <p>
+     * The record is expected to contain
+     * <ul>
+     * <li>a <em>topic</em> of the format <em>hono.command.${tenant_id}</em></li>
+     * <li>a <em>key</em> with the command target device id, matching the value of the <em>device_id</em> header</li>
+     * <li>a non-empty <em>correlation-id</em> and/or <em>message-id</em> header if an <em>expect-response</em> header
```
sophokles73, line 82: we decided to only use the *correlation-id* header for this purpose, didn't we?

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommand.java
```diff
@@ -0,0 +1,285 @@
...
+     * is set with value {@code true}.</li>
+     * </ul>
+     * <p>
+     * If any of the requirements above are not met, then the returned command's {@link #isValid()}
+     * method will return {@code false}.
+     *
+     * @param record The record containing the command.
+     * @param deviceId The identifier of the device that the command will be sent to. If the command has been mapped
+     *                 to a gateway, this id is the gateway id and the original command target device is given in
+     *                 the message address.
```
sophokles73, line 92: ... is given by the record's <em>key</em> ?

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommand.java
```diff
@@ -0,0 +1,285 @@
...
+    /**
+     * Creates a command from a Kafka consumer record.
+     * <p>
+     * The record is expected to contain
+     * <ul>
+     * <li>a <em>topic</em> of the format <em>hono.command.${tenant_id}</em></li>
+     * <li>a <em>key</em> with the command target device id, matching the value of the <em>device_id</em> header</li>
+     * <li>a non-empty <em>correlation-id</em> and/or <em>message-id</em> header if an <em>expect-response</em> header
+     * is set with value {@code true}.</li>
+     * </ul>
```
sophokles73, line 84: the *subject* header also seems to be required, isn't it?

clients/adapter-kafka/src/test/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommandTest.java
```diff
@@ -0,0 +1,190 @@
...
+ * Verifies behavior of {@link KafkaBasedCommand}.
+ *
+ */
+public class KafkaBasedCommandTest {
+
+    /**
+     * Verifies that a command can be created from a valid message.
+     */
+    @Test
+    public void testFromMessageSucceeds() {
```
sophokles73, line 45: testFromRecord?

clients/adapter-kafka/src/test/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommandTest.java
```diff
@@ -0,0 +1,190 @@
...
+        assertTrue(cmd.isValid());
+        assertThat(cmd.getName()).isEqualTo(subject);
+        assertThat(cmd.getDeviceId()).isEqualTo(deviceId);
+        assertThat(cmd.getOriginalDeviceId()).isEqualTo(deviceId);
+        assertThat(cmd.getCorrelationId()).isEqualTo(correlationId);
+        assertFalse(cmd.isOneWay());
+    }
+
+    /**
+     * Verifies that a command can be created from a valid message, containing a message address with device id
```
sophokles73, line 63: needs to be adapted to Kafka record terminology

clients/adapter/pom.xml
```diff
@@ -126,5 +126,20 @@
...
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+    </dependency>
   </dependencies>
 </project>
```
sophokles73, line 145: missing EOL

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/Command.java
```diff
@@ -97,7 +97,10 @@
     String getOriginalDeviceId();
 
     /**
-     * Gets the request identifier of this command.
+     * Gets the request identifier of this command. Can be used to correlate this command
+     * with a corresponding command response.
+     * <p>
+     * May be {@code null} for a one-way command.
```
sophokles73, line 103: It *will* be `null` for a one-way command, won't it?

calohmn: For the `KafkaBasedCommand`, yes.
But not in the case of a `ProtonBasedCommand` with a `correlation-id` property but no `reply-to` property (i.e. a one-way command with a given correlation-id).

calohmn:
@sophokles73 I've pushed changes addressing the comments above.

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommand.java
```diff
@@ -0,0 +1,285 @@
...
+     * <li>a non-empty <em>correlation-id</em> header if a response is expected for the command.</li>
+     * </ul>
+     * <p>
+     * If any of the requirements above are not met, then the returned command's {@link #isValid()}
+     * method will return {@code false}.
+     *
+     * @param record The record containing the command.
+     * @param deviceId The identifier of the device that the command will be sent to. If the command has been mapped
+     *                 to a gateway, this id is the gateway id and the original command target device is given in
+     *                 the <em>device_id</em> header and the key of the record.
```
sophokles73, line 92: FMPOV this is very confusing. Why isn't all the necessary information contained in the headers? Why do we need to pass in the device ID as a separate parameter? Can't we simply read the target device ID from the *device_id* header and, if the device is connected via a gateway, read the gateway ID from e.g. the *gateway_id* header that has been added by the Command Router?

This is also very confusing in the `Command` interface as well. FMPOV the *getDeviceId()* method should *always* return the identifier of the device that is the target of the command. We should then replace *getOriginalDeviceId* with *getGatewayId* which returns the gateway ID if the command is routed via a gateway. WDYT?

calohmn: Yes, such a change is something I already had on my TODO list. As there are quite many files involved in that change, I would rather do this in a separate PR, created as a followup of this one.

>Why do we need to pass in the device ID as a separate parameter?

Because there is one scenario, where the gateway id cannot be deduced from the message/record (i.e. the `via` property/header set by the command router). That's when a gateway subscribes for commands of a specific device. The gateway information is only available at the target protocol adapter in this case.
But I want to remove the device id parameter from the factory method and either set the `via` property/header beforehand or introduce a `setGatewayId()` setter for that.

calohmn:
Pushed minor changes in `KafkaBasedCommand`.

sophokles73 (approved):
LGTM

PR #2411, opened 2021-01-13
Fix build of documentation.

b-abel:
Hugo fails with the following error:

```
`"hono/site/documentation/content/concepts/device-provisioning/index.md:97:53": unterminated quoted string in shortcode parameter-argument: 'A unregistered device connects to a gateway which uses the 
'
```
This causes the Hono build to fail as well.

This fix removes a line break inside the "alt" string of a figure declaration.


calohmn (approved):
LGTM

PR #2414, opened 2021-01-13
[#2413] Auto-provisioning must fail when done concurrently by two gat…

### LINKED ISSUE #2413: Auto-provisioning must not succeed if the same device is concurrently sending via two different gateways
See: https://github.com/eclipse/hono/commit/fe27f781158b4161c0aa03db796643ef667fa0a4#r45925380

fkaltner:
…eways for the same device.

This fixes #2413

### DIFF
--- a/services/device-registry-base/src/test/java/org/eclipse/hono/deviceregistry/service/device/AutoProvisionerTest.java
+++ b/services/device-registry-base/src/test/java/org/eclipse/hono/deviceregistry/service/device/AutoProvisionerTest.java
@@ -33,6 +33,7 @@ import java.util.Optional;
 import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.client.DownstreamSender;
 import org.eclipse.hono.client.DownstreamSenderFactory;
+import org.eclipse.hono.client.ServiceInvocationException;
 import org.eclipse.hono.deviceregistry.service.tenant.TenantInformationService;
 import org.eclipse.hono.service.management.Id;
 import org.eclipse.hono.service.management.OperationResult;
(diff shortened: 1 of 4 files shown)


services/device-registry-base/src/test/java/org/eclipse/hono/deviceregistry/service/device/AutoProvisionerTest.java
```diff
@@ -153,6 +153,29 @@ public void testAutoProvisioningIsSuccessfulForAlreadyPresentEdgeDevice(final Ve
...
+    @Test
+    public void testAutoProvisioningFailsWhenTriggeredConcurrentlyByDifferentGatewaysForTheSameDevice(final VertxTestContext ctx) {
+        mockAssertRegistration(GATEWAY_ID, Collections.singletonList(GATEWAY_GROUP_ID), Collections.singletonList(RegistryManagementConstants.AUTHORITY_AUTO_PROVISIONING_ENABLED));
+        mockAssertRegistration(DEVICE_ID, true);
+        mockAddEdgeDevice(HttpURLConnection.HTTP_CONFLICT);
+
+        autoProvisioner.performAutoProvisioning(Constants.DEFAULT_TENANT, DEVICE_ID, "another-" + GATEWAY_ID, NEW_EDGE_DEVICE, span.context())
+                .onComplete(ctx.failing(throwable -> {
+                    ctx.verify(() -> {
+                        verify(deviceManagementService).createDevice(any(), any(), any(), any());
```
sophokles73, line 170: it helps understanding the test better, if you are as specific as possible in the verification. In this case, I guess this would be
```
verify(deviceManagementService).createDevice(eq(Constants.DEFAULT_TENANT), eq(Optional.of(DEVICE_ID)), any(Device.class), any(Span.class));
```
or am I mistaken?

fkaltner: No, this is correct. 

One could be more specific by adding an argument captor checking if the given device hast GATEWAY_ID in its "via" property. Not sure where to draw the line though regarding "_as specific as possible_".
Personally I it feels a bit odd to verify what I just wrote in the same test, but I am open to add that as well if it helps.

I also added the assertions regarding the status code.

sophokles73 (approved):
LGTM

PR #2416, opened 2021-01-15
Refactor Command classes

calohmn:
Refactor device/gateway related Command methods:
`getDeviceId()` has been renamed to `getGatewayOrDeviceId()` and `getOriginalDeviceId()` has been renamed to `getDevice()`. Also a `getGatewayId()` method has been added.

Refactor Command classes:
The new `Command` class now requires tenant/device information to be set and returns this kind of information even if the
command is otherwise invalid. This makes it easier to choose the command handler for a potentially invalid command.
Usage of the legacy command class inside the` ProtonBasedCommand` has been removed.

For the review it may be easier to look at the 2 commits separately.

### DIFF
--- a/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedAdapterInstanceCommandHandler.java
+++ b/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedAdapterInstanceCommandHandler.java
@@ -24,8 +24,6 @@ import org.eclipse.hono.adapter.client.command.CommandHandlers;
 import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.Constants;
 import org.eclipse.hono.util.MessageHelper;
-import org.eclipse.hono.util.ResourceIdentifier;
-import org.eclipse.hono.util.Strings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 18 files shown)


clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/Command.java
```diff
@@ -87,14 +89,20 @@
     /**
      * Gets the device identifier used in the original command. It is extracted from the
      * <em>to</em> property of the command AMQP message.
```
sophokles73, line 82: Since this is a generic class, IMHO we should remove the reference to the AMQP specific *to* property ...
I would also suggest to make this more explicit:
> Gets the identifier of the device that is supposed to execute the command.

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/Command.java
```diff
@@ -61,16 +61,18 @@
     String getTenant();
 
     /**
-     * Gets the device's identifier.
+     * Gets the identifier of the gateway or edge device that this
+     * command is to be sent to.
```
sophokles73, line 58: how about
> Gets the identifier of the gateway or edge device that this command needs to be forwarded to for delivery.

calohmn:
@sophokles73 I've made the suggested changes.

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedCommand.java
```diff
@@ -31,190 +40,308 @@
...
     @Override
-    public String getInvalidCommandReason() {
-        return command.getInvalidCommandReason();
+    public boolean isValid() {
+        return !validationError.isPresent();
     }
 
     /**
-     * {@inheritDoc}
+     * {@inheritDoc}}
```
sophokles73, line 197: typo

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommand.java
```diff
@@ -45,99 +44,120 @@
...
     private static String getHeader(final KafkaConsumerRecord<String, Buffer> record, final String header) {
         return RecordUtil.getStringHeaderValue(record.headers(), header);
     }
 
+    /**
+     * Sets the identifier of the gateway this command is to be sent to.
+     *
+     * @param gatewayId The gateway identifier.
+     */
+    public void setGatewayId(final String gatewayId) {
```
sophokles73, line 157: why do you need to have a setter for the gateway ID? Isn't it determined from the record?

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedCommand.java
```diff
@@ -31,190 +40,308 @@
...
+        return command;
     }
 
     /**
-     * {@inheritDoc}
+     * Sets the identifier of the gateway this command is to be sent to.
+     *
+     * @param gatewayId The gateway identifier.
      */
+    public void setGatewayId(final String gatewayId) {
```
sophokles73, line 182: why do we need a setter for the gateway ID? Isn't it determined from the message?

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommand.java
calohmn: Not exclusively - see the comment in `ProtonBasedAdapterInstanceCommandHandler#handleCommandMessage`:
````
// Gateway information set in command handler means a gateway has subscribed for commands for a specific device.
// This information isn't getting set in the message (by the Command Router) and therefore has to be adopted manually here.
````

File: clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedCommand.java (hunk 5 above)
Comment (calohmn) [reply clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedCommand.java:182]: See above - same applies to the Kafka record.

File: clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommand.java (hunk 4 above)
Comment (sophokles73) [reply clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommand.java:157]: I see. I guess it would make sense to add this information to the JavaDoc here in case other people (or me again in 3 months) are wondering as well :-)

Comment (calohmn) [reply clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommand.java:164]: Yes, I've added the description of the scenario to the JavaDoc.

File: clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedAdapterInstanceCommandHandler.java [hunk 6]
```diff
@@ -70,38 +68,28 @@ public void handleCommandMessage(final Message msg, final ProtonDelivery deliver
...
+        if (commandHandler != null && commandHandler.getGatewayId() != null) {
+            // Gateway information set in command handler means a gateway has subscribed for commands for a specific device.
+            // This information isn't getting set in the message (by the Command Router) and therefore has to be adopted manually here.
+            command.setGatewayId(commandHandler.getGatewayId());
+        }
 
         final SpanContext spanContext = TracingHelper.extractSpanContext(tracer, msg);
-        final Span currentSpan = org.eclipse.hono.client.impl.CommandConsumer.createSpan("handle command", tenantId,
-                originalDeviceId, gatewayId, tracer, spanContext);
+        final Span currentSpan = org.eclipse.hono.client.impl.CommandConsumer.createSpan("handle command",
```
sophokles73, line 91: I assume this dependency will be removed in an additional PR?

sophokles73 (approved):
LGTM

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedAdapterInstanceCommandHandler.java
calohmn: Yes.

PR #2419, opened 2021-01-18
Use the new API for sending to internal command endpoint

calohmn:
The Command Router now uses the new API for sending command messages to the target protocol adapter instance.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/Command.java
+++ b/client/src/main/java/org/eclipse/hono/client/Command.java
@@ -48,10 +48,10 @@ public final class Command {
     private final Optional<String> validationError;
     private final Message message;
     private final String tenantId;
-    private final String deviceId;
     private final String correlationId;
     private final String replyToId;
     private final String requestId;
+    private String deviceId;
 
     private Command(
             final Optional<String> validationError,
@@ -230,7 +230,8 @@ public final class Command {
     }
 
     /**
-     * Gets the device's identifier.
+     * Gets the identifier of the gateway or edge device that this command
+     * needs to be forwarded to for delivery.
      * <p>
      * In the case that the command got redirected to a gateway,
      * the id returned here is a gateway id. See {@link #getOriginalDeviceId()}
@@ -240,6 +241,18 @@ public final class Command {
      */
     public String getDeviceId() {
         return deviceId;
+    }
+
+    /**
+     * Sets the identifier of the gateway this command is to be sent to.
+     * <p>
+     * Using {@code null} as parameter means that the command is to be forwarded directly
+     * to the device given in the original command message, without using a gateway.
+     *
+     * @param gatewayId The gateway identifier.
+     */
+    public void setGatewayId(final String gatewayId) {
+        this.deviceId = Optional.ofNullable(gatewayId).orElseGet(this::getOriginalDeviceId);
     }
 
     /**
@@ -468,10 +481,10 @@ public final class Command {
         if (isValid()) {
             final String originalDeviceId = getOriginalDeviceId();
             if (!getDeviceId().equals(originalDeviceId)) {
-                return String.format("Command [name: %s, tenant-id: %s, device-id %s, original device-id %s, request-id: %s]",
+                return String.format("Command [name: %s, tenant-id: %s, gateway-id: %s, device-id: %s, request-id: %s]",
                         getName(), getTenant(), getDeviceId(), originalDeviceId, getRequestId());
             } else {
-                return String.format("Command [name: %s, tenant-id: %s, device-id %s, request-id: %s]",
+                return String.format("Command [name: %s, tenant-id: %s, device-id: %s, request-id: %s]",
                         getName(), getTenant(), getDeviceId(), getRequestId());
             }
         } else {
--- a/clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/Command.java
+++ b/clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/Command.java
@@ -93,6 +93,16 @@ public interface Command {
     String getGatewayId();
 
     /**
+     * Sets the identifier of the gateway this command is to be sent to.
+     * <p>
+     * Using {@code null} as parameter means that the command is to be forwarded directly
+     * to the device given in the original command message, without using a gateway.
+     *
+     * @param gatewayId The gateway identifier.
+     */
+    void setGatewayId(String gatewayId);
+
+    /**
      * Gets the name of this command.
      *
      * @return The name.
--- a/clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/InternalCommandSender.java
+++ b/clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/InternalCommandSender.java
@@ -1,0 +1,42 @@
+/**
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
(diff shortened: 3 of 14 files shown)


clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/InternalCommandSender.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * A client to send commands to the internal Command and Control API endpoint provided by protocol adapters.
+ */
+public interface InternalCommandSender extends Lifecycle {
+
+    /**
+     * Sends a command to the downstream peer, targeting the protocol adapter Command and Control API endpoint
+     * identified by the given protocol adapter instance id.
+     * <p>
+     * The result of the send operation will lead to one of the {@link CommandContext#accept()},
```
sophokles73, line 29: FMPOV this needs to be re-phrased to make this a requirement towards implementers, e.g. something like 
> Implementations are required to invoke one of the CommandContext's accept, release, modify or reject methods based on the outcome of sending the command to the protocol adapter instance.

services/command-router/src/main/java/org/eclipse/hono/commandrouter/impl/amqp/MappingAndDelegatingCommandHandler.java
```diff
@@ -112,124 +103,77 @@ public void mapAndDelegateIncomingCommandMessage(
...
+            commandContext.reject("malformed command message");
         }
     }
 
-    private void mapAndDelegateIncomingCommand(final String tenantId, final String originalDeviceId,
-            final CommandContext originalCommandContext) {
-        final Command originalCommand = originalCommandContext.getCommand();
+    private void mapAndDelegateIncomingCommand(final String tenantId, final String deviceId,
+            final ProtonBasedCommandContext commandContext) {
+        final ProtonBasedCommand command = commandContext.getCommand();
```
sophokles73, line 137: does this really need to be a `ProtonBasedCommand` instead of just a `Command`?

services/command-router/src/main/java/org/eclipse/hono/commandrouter/impl/amqp/ProtonBasedCommandConsumerFactoryImpl.java
```diff
@@ -157,13 +156,17 @@ protected void onDisconnect() {
...
                     log.debug("successfully created MappingAndDelegatingCommandConsumer [{}]", address);
                     consumerLinkTenants.add(tenantId);
-                    final CommandConsumer consumer = new CommandConsumer(connection, receiver);
-                    consumer.setLocalCloseHandler(sourceAddress -> {
-                        log.debug("MappingAndDelegatingCommandConsumer receiver link [tenant-id: {}] closed locally", tenantId);
-                        mappingAndDelegatingCommandConsumerFactory.removeClient(tenantId);
-                        consumerLinkTenants.remove(tenantId);
-                    });
-                    return (MessageConsumer) consumer;
+                    return (CommandConsumer) new CommandConsumer() {
```
sophokles73, line 159: do we really need the cast here?

calohmn: It seems to be needed, yes. Without it there is this compile error:
````
incompatible types: 
io.vertx.core.Future<<anonymous org.eclipse.hono.adapter.client.command.CommandConsumer>> 
cannot be converted to 
io.vertx.core.Future<org.eclipse.hono.adapter.client.command.CommandConsumer>
````

File: clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/InternalCommandSender.java (hunk 1 above)
Comment (calohmn) [reply clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/InternalCommandSender.java:29]: As with every interface method contract, it can be seen from both sides. 
I've added the implementers point of view then as well, in the 2nd sentence:

> Note that the result of the send operation is getting applied to the command context here. That means implementations are required to invoke one of the given command context's CommandContext.accept(), CommandContext.release(), CommandContext.modify(boolean, boolean) or CommandContext.reject(String) methods based on the outcome of sending the command to the protocol adapter instance.

File: services/command-router/src/main/java/org/eclipse/hono/commandrouter/impl/amqp/MappingAndDelegatingCommandHandler.java (hunk 2 above)
Comment (calohmn) [reply services/command-router/src/main/java/org/eclipse/hono/commandrouter/impl/amqp/MappingAndDelegatingCommandHandler.java]: `setGatewayId()` not being in the `Command` interface stood in the way.
But I've pulled that method into the interface now and added a legacy implementation.

File: services/command-router/src/main/java/org/eclipse/hono/commandrouter/impl/amqp/ProtonBasedCommandConsumerFactoryImpl.java (hunk 3 above)
Comment (sophokles73) [reply services/command-router/src/main/java/org/eclipse/hono/commandrouter/impl/amqp/ProtonBasedCommandConsumerFactoryImpl.java:159]: interesting

File: clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/InternalCommandSender.java [hunk 4]
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * A client to send commands to the internal Command and Control API endpoint provided by protocol adapters.
+ */
+public interface InternalCommandSender extends Lifecycle {
+
+    /**
+     * Sends a command to the downstream peer, targeting the protocol adapter Command and Control API endpoint
+     * identified by the given protocol adapter instance id.
+     * <p>
+     * Note that the result of the send operation is getting applied to the command context here.
```
sophokles73, line 29: It is not, unless the implementation does so. That is why I urged to make this more explicit. FMPOV the first sentence should be removed as it suggests that something is happening *automagically* here ...

calohmn: Well, "Sends a command to the downstream peer" is also not happening unless the implementation does so.

From the viewpoint of someone *invoking* this interface method, I'm interested in what the method does - send a command *and* apply the result to the command context. That is an important piece of information for *invoking* the method, to emphasize that the command context doesn't need to get "accepted" by the invoking method afterwards.

Therefore I think this kind of target audience of the javadoc should be addressed here, not only the other viewpoint of the interface implementer.

(I initially wanted to not have the "apply the result to the command context" logic in that method, but the alternative would be passing the resulting proton delivery in some result value object, making things more complicated for the proton-independent interface.)

sophokles73: I see your point.

client/src/main/java/org/eclipse/hono/client/Command.java
```diff
@@ -242,6 +243,15 @@ public String getDeviceId() {
         return deviceId;
     }
 
+    /**
+     * Sets the identifier of the gateway this command is to be sent to.
+     *
+     * @param gatewayId The gateway identifier.
+     */
+    public void setGatewayId(final String gatewayId) {
+        this.deviceId = Optional.ofNullable(gatewayId).orElseGet(this::getOriginalDeviceId);
```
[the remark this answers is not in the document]
calohmn: I've adapted the javadoc here.

> BTW is it really expected/intended that this method is being invoked with `null` as the argument?

I've mainly added that to comply with the corresponding new `Command` interface method. Since `getGatewayId` may return `null` there, I think it's ok to allow resetting a command to such a state.

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/Command.java
```diff
@@ -92,6 +92,13 @@
      */
     String getGatewayId();
 
+    /**
+     * Sets the identifier of the gateway this command is to be sent to.
+     *
+     * @param gatewayId The gateway identifier.
```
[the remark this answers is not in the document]
calohmn: Yes, it's supported. As stated in the other comment above, since `getGatewayId` may return `null`, I think it's ok to allow resetting a command to such a state.
I've extended the javadoc here for that.

sophokles73 (approved):
LGTM

PR #2420, opened 2021-01-18
Add an abstract Kafka Consumer

b-abel:
This PR provides a pretty generic Kafka consumer with AT LEAST ONCE semantics.

### DIFF
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/consumer/AbstractAtLeastOnceKafkaConsumer.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/consumer/AbstractAtLeastOnceKafkaConsumer.java
@@ -1,0 +1,289 @@
+/*
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.hono.kafka.client.consumer;
+
+import java.time.Duration;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import org.eclipse.hono.util.Lifecycle;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.Promise;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.kafka.client.consumer.KafkaConsumer;
+import io.vertx.kafka.client.consumer.KafkaConsumerRecord;
+import io.vertx.kafka.client.consumer.KafkaConsumerRecords;
+
+/**
+ * A client to consume data from a Kafka cluster.
+ * <p>
+ * This consumer continuously polls for batches of messages from Kafka. Each message is passed to a message handler for
+ * processing. When all messages of a batch are processed, this consumer commits the offset and polls for a new batch.
+ * <p>
+ * This consumer provides AT LEAST ONCE message processing. When the message handler completes without throwing an
+ * exception, the message is considered to be processed and will be "acknowledged" with the following commit. If an
+ * error occurs, the underlying Kafka consumer will be closed. In this case, other consumer instances in the same
+ * consumer group will consume the messages from the last committed offset. Already processed messages will then be
+ * again in the next batch and be passed to the message handler. If de-duplication of messages is required, it must be
+ * handled by the message handler.
+ * <p>
+ * The consumer starts consuming when {@link #start()} is invoked. It needs to be closed by invoking {@link #stop()} to
+ * release the resources.
+ * <p>
+ * </p>
+ * ERROR CASES:
+ * <p>
+ * Errors can happen when polling, in message processing, and when committing the offset to Kafka.
+ *
+ * If a fatal error occurs, the underlying Kafka consumer will be closed and the close-handler invoked with an exception
+ * indicating the cause. Therefore, the provided Kafka consumer must not be used anywhere else.
+ * <p>
+ * If {@link KafkaConsumer#poll(Duration, Handler)} fails during the start, {@link #start()} will return a failed
+ * future. For subsequent {@code poll} operations, the Kafka consumer will be closed and the close handler will be
+ * passed a {@link KafkaConsumerPollException}.
+ *
+ * <p>
+ * If the message processing fails because either {@link #createMessage(KafkaConsumerRecord)} or the message handler
+ * throws an exception, the Kafka consumer will be closed and the exception will be passed to the close handler.
+ * <p>
+ * If {@link KafkaConsumer#commit(Handler)} times out, the commit will be retried once. If the retry fails or the commit
+ * fails with another exception, the Kafka consumer will be closed and the close handler will be passed a
+ * {@link KafkaConsumerCommitException}. For example, commits could regularly fail during consumer rebalance.
+ *
+ * @param <T> The type of message to be created from the Kafka records.
+ */
+public abstract class AbstractAtLeastOnceKafkaConsumer<T> implements Lifecycle {
+
+    // TODO add unit test
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractAtLeastOnceKafkaConsumer.class);
+
+    private final KafkaConsumer<String, Buffer> kafkaConsumer;
+    private final Set<String> topics;
+    private final Pattern topicPattern;
+    private final Consumer<T> messageHandler;
+    private final Handler<Throwable> closeHandler;
+    private final Duration pollTimeout;
+
+    /**
+     * Creates a new consumer.
+     *
+     * @param kafkaConsumer The Kafka consumer to be exclusively used by this instance to consume records.
+     * @param topic The Kafka topic to consume records from.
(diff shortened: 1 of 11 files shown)


b-abel:
Consumers will be needed at least in the following places:

- business application client
- integration tests
- command router
- protocol adapters for commands

I would like to use this as a common implementation for all those places.

calohmn:
_Commit handling / when is a message considered as "safe":_
The `AbstractAtLeastOnceKafkaConsumer` implementation here simply invokes a messageHandler for each received message (`messageHandler.accept(message)`), implying that message handling is synchronous here and that the message is "safe" to be committed after the "accept" invocation has finished. But I think we need to think in asynchronous terms here and only consider the message "safe" once the message handler has explicitly said so by completing a passed in `Future`.
This of course makes things more complicated in order to fulfill the at-least-once constraint. It also emphasizes the need to think about how to handle a failure in message handling, which could also be a timeout in not completing the result Future in time.
I think the article [here](https://quarkus.io/blog/kafka-commit-strategies/) gives a good overview of possible approaches, specifically when handling messages asynchronously. I think we should choose one of the strategies presented there (possibly "throttled" or "latest"), or even implement different strategies and make the one to use configurable.

_Polling logic:_
As I see it, the `AbstractAtLeastOnceKafkaConsumer` class now implements its own automatic polling mechanism, resembling the default polling mechanism in the vert.x KafkaConsumer but adding integrated commit handling.
As I think the commit handling needs to be adapted anyway (see above), possibly ending up using the `kafkaConsumer#commit(offsets, handler)` method, the self-implemented polling mechanism may not be needed anymore.
This would enable us to use the default polling mechanism, which to me seems preferable, as it would allow more features of the KafkaConsumer to be used.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/consumer/AbstractAtLeastOnceKafkaConsumer.java
```diff
@@ -0,0 +1,231 @@
...
+     * @param topic The Kafka topic to consume records from.
+     * @param messageHandler The function to be invoked for each message created from a record.
+     * @param closeHandler The handler to be invoked when the Kafka consumer has been closed due to an error.
+     * @param pollTimeout The maximal number of milliseconds to wait for messages during a poll operation.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     * @see #start()
+     * @see #stop()
+     */
+    public AbstractAtLeastOnceKafkaConsumer(final KafkaConsumer<String, Buffer> kafkaConsumer, final String topic,
+            final Consumer<T> messageHandler, final Handler<Throwable> closeHandler, final long pollTimeout) {
```
calohmn, line 98: The way the `AbstractAtLeastOnceKafkaConsumer` currently is implemented means the given `kafkaConsumer` can only be used exclusively with this `AbstractAtLeastOnceKafkaConsumer` instance, and the one given topic, right?

I think that should be pointed out here. And for scenarios where multiple topics are needed, corresponding constructors with `Set<String> topics` and `Pattern topicPattern` parameters would be good.

b-abel:
> _Commit handling / when is a message considered as "safe":_
> The `AbstractAtLeastOnceKafkaConsumer` implementation here simply invokes a messageHandler for each received message (`messageHandler.accept(message)`), implying that message handling is synchronous here and that the message is "safe" to be committed after the "accept" invocation has finished. 

Exactly. To me, this is the most straight forward way to consume messages from Kafka. It provides a clean interface that is easy to reason about. This is probably also why the Vert.x KafkaConsumer promotes it. The `AbstractAtLeastOnceKafkaConsumer` only adds AT LEAST ONCE semantics to it. Asynchronous message consumption with Kafka is really hard, even the Vert.x KafkaConsumer expects the messages to be consumed synchronously. The common strategies I saw are to have either a quick consumption or to persist records again (in Kafka or even relational databases) for further processing later on. 

> But I think we need to think in asynchronous terms here and only consider the message "safe" once the message handler has explicitly said so by completing a passed in `Future`.

Are you sure that this is a must? While you could consume messages in parallel, you can only commit the batch once the consumption of _every_ message in the batch has finished. What do you want to do if the consumption of one message takes very long or fails with an error? The whole consumption from the partition would be halted. So, I don't see an easy alternative to expecting the message consumption to complete reasonably fast. We should indeed have a further discussion about failures in the message consumption.

> I think we should choose one of the strategies presented there (possibly "throttled" or "latest"), or even implement different strategies and make the one to use configurable.

It could be interesting to see how Quarkus implements the _throttled_ strategy and what are the implications of message _nack_ and _poison pills_ there. I hope that we can avoid the _latest_ strategy because committing for every message pretty much destroys much of the performance that Kafka provides (and you are pretty much back to synchronous message handling).

I would prefer if we start with the synchronous message processing at least for northbound consumers of downstream messages. We can add more complex and asynchronous consumer implementations any time later.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/consumer/AbstractAtLeastOnceKafkaConsumer.java
b-abel: > The way the `AbstractAtLeastOnceKafkaConsumer` currently is implemented means the given `kafkaConsumer` can only be used exclusively with this `AbstractAtLeastOnceKafkaConsumer` instance, and the one given topic, right?

Yes. Providing the native `KafkaConsumer` as a parameter is mainly for testability (in unit tests I would use `org.apache.kafka.clients.consumer.MockConsumer` here). Maybe it would be an improvement to add a second constructor without the parameter. Then the class would manage the `KafkaConsumer` while I can still provide the MockConsumer in unit tests. WDYT?

calohmn:
> Exactly. To me, this is the most straight forward way to consume messages from Kafka. It provides a clean interface that is easy to reason about. This is probably also why the Vert.x KafkaConsumer promotes it. The AbstractAtLeastOnceKafkaConsumer only adds AT LEAST ONCE semantics to it. 

But without support for taking the asynchronous message handling into account when committing the offset, I don't think we should call it "at least once" consumer.
Simple example:
- Command & Control message X gets received
- messageHandler.accept(messageX) gets invoked
  - handler logic calls `assertRegistration`, event loop thread will do other things while waiting for the result
- offset gets committed
- protocol adapter crashes, restarts
- consumer will not read message X again, so it is lost, which doesn't fit the "at least once" guarantee.

I'm not saying that the message X above absolutely has to be read again here (for our initial Kafka implementation anyway). In the AMQP scenario it would be lost as well.
But emphasizing an "at-least-once" guarantee while the typical usage scenario of the consumer doesn't support that, is misleading to me.

> > But I think we need to think in asynchronous terms here and only consider the message "safe" once the message handler has explicitly said so by completing a passed in `Future`.

> Are you sure that this is a must? While you could consume messages in parallel, you can only commit the batch once the consumption of _every_ message in the batch has finished. 

For the asynchronous approach, you wouldn't necessarily think in terms of batches when committing. You could do a periodical commit (similar to the auto-commit) but only take the already handled/acknowledged messages into account (like implemented [here](https://github.com/smallrye/smallrye-reactive-messaging/blob/e16c139a515bc3b01683112dcf076f182c308fac/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/commit/KafkaThrottledLatestProcessedCommit.java)).

As to whether I see it as a must: I think we can start out with a synchronous solution. Then we can evaluate, whether/when to move towards a solution with asynchronous support.

> What do you want to do if the consumption of one message takes very long or fails with an error? The whole consumption from the partition would be halted.

Consumption taking too long should lead to a timeout exception. Failure handling in general would lead to the same question as already exists with the solution in this PR: What to do when `closeWithError` gets called?

b-abel:
> But without support for taking the asynchronous message handling into account when committing the offset, I don't think we should call it "at least once" consumer.
> Simple example:

Thank you for the example, now I get the use case you have in mind. For such long-running message processing the class is indeed not well suited. I think we need another consumer implementation for that.

> I'm not saying that the message X above absolutely has to be read again here (for our initial Kafka implementation anyway). In the AMQP scenario it would be lost as well.

Maybe an AT MOST ONCE behavior is better suited for the consumption of command messages.

> But emphasizing an "at-least-once" guarantee while the typical usage scenario of the consumer doesn't support that, is misleading to me.

I will clarify the scope of the AT MOST ONCE guarantee in the JavaDoc.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/consumer/AbstractAtLeastOnceKafkaConsumer.java
```diff
@@ -0,0 +1,291 @@
...
+/**
+ * A client to consume data from a Kafka cluster.
+ * <p>
+ * This consumer continuously polls for batches of messages from Kafka. Each message is passed to a message handler for
+ * processing. When all messages of a batch are processed, this consumer commits the offset and polls for a new batch.
+ * <p>
+ * This consumer provides AT LEAST ONCE message processing. When the message handler completes without throwing an
+ * exception, the message is considered to be processed and will be "acknowledged" (with the following commit). If the
+ * processing of messages or the commit of the offsets fails, other consumer instances in the same consumer group will
+ * consume the messages from the last committed offset. Already processed messages will then be again in the next batch
```
calohmn, line 48: > other consumer instances in the same consumer group will consume the messages from the last committed offset

This only happens because in case of the commit error, the consumer is stopped here, right? Without that information, the sentence is a bit confusing here.

calohmn:
> For such long-running message processing the class is indeed not well suited. I think we need another consumer implementation for that.

Yes, ok.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/consumer/AbstractAtLeastOnceKafkaConsumer.java
b-abel: Please have a look if it is now more understandable.

calohmn (approved):
LGTM

calohmn:
@b-abel Can you squash this to the commits you want to see merged?

b-abel:
> @b-abel Can you squash this to the commits you want to see merged?

Sure. Thanks.

PR #2422, opened 2021-01-19
Fix -DskipTests not working for docker related tests

calohmn:
The "no.docker" property has been replaced by "skipDockerTests", initialized to the "skipTests" value by default.
This re-enables the possibility to skip all tests via the -DskipTests maven argument.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -188,7 +188,7 @@
     <!--
       Needs to be set to true on platforms where no Docker daemon is available.
     -->
-    <no.docker>false</no.docker>
+    <skipDockerTests>${skipTests}</skipDockerTests>
 
     <!--
       Build images defined in modules with packaging type "pom".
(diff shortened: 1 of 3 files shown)


sophokles73 (changes requested):
If you want to change the name of the maven property then you will also need to adapt the Jenkins build jobs accordingly, where the property is used to disable the Docker related tests when running on Eclipse Foundation Jenkins infrastructure ...

calohmn:
Do you mean the "-DnoDocker" parameter here?

https://github.com/eclipse/hono/blob/944697cdb4c434284ea2d08214831211630cb3d1/jenkins/Hono-Nightly-Pipeline-Declarative.groovy#L113

That would be unaffected by the change in this PR. Only the `no.docker` property (with a dot) has been renamed.

sophokles73 (approved):
LGTM

PR #2423, opened 2021-01-19
[#2408] Sending an event with a dedicated content type for notifying …

### LINKED ISSUE #2408: Device registry should use dedicated event type for indicating provisioning of a device
The device registry currently uses the well-known *empty notification* in order to indicate the (implicit) provisioning of a device. However, the purpose of the *empty notification* has been defined differently and consumers of the event now need to determine the real purpose of the event by interpreting the presence/absence of particular properties instead of simply inspecting the event's content-type as originally intended.

fkaltner:
…applications about auto-provisioned devices.

This fixes #2408


site/documentation/content/api/event/index.md
```diff
@@ -134,3 +130,21 @@ The example below might be used by the MQTT adapter to indicate that a connectio
   }
 }
 ~~~
+
+### Device Provisioning Notification
+
+Device registries may send this event to indicate that a new device has been provisioned. This may be used as a hook for
```
sophokles73, line 136: FMPOV we should not tie this event to the gateway based auto-provisioning process here. The event can be used to convey the provisioning status of a device in a very generic sense. As such, it could also be used to indicate that a device has been de-provisioned (deleted) and it should also be possible to be used regardless of the trigger that caused the device's status to be changed, i.e. we could also send this event if the device had been created using the Device Registry Management API, right?

site/documentation/content/api/event/index.md
```diff
@@ -134,3 +130,21 @@ The example below might be used by the MQTT adapter to indicate that a connectio
...
+implemented upon a newly provisioned device.
+
+An AMQP message containing this type of event does not have any payload, so the body of the message MUST be empty.
+
+The relevant properties are listed again in the following table:
+
+| Name                       | Mandatory        | Location                 | Type      | Description |
+| :------------------------- | :--------------: | :----------------------- | :-------- | :---------- |
+| *content-type*             | yes              | *properties*             | *symbol*  | MUST be set to *application/vnd.eclipse-device-provisioning-notification* |
+| *hono_registration_status* | no               | *application-properties* | *string*  | If set to `NEW` this indicates that a device was auto-provisioned in the course of sending telemetry data via gateway. |
```
sophokles73, line 147: FMPOV this is mandatory as it indicates whether the device is newly added, removed, updated etc.
We should not mention *auto-provisioned* here as the event is also applicable for manual provisioning cases. How about
> Contains `NEW` if the device has been newly provisioned.

site/documentation/content/api/event/index.md
```diff
@@ -134,3 +130,21 @@ The example below might be used by the MQTT adapter to indicate that a connectio
...
+| *tenant_id*                | no               | *application-properties* | *string*  | The tenant id denoting the tenant for which the device was auto-provisioned. |
```
sophokles73, line 148: FMPOV this is mandatory as it qualifies the device identifier
We should not mention *auto-provisioned* here as the event is also applicable for manual provisioning cases ...

site/documentation/content/api/event/index.md
```diff
@@ -134,3 +130,21 @@ The example below might be used by the MQTT adapter to indicate that a connectio
...
+| *device_id*                | no               | *application-properties* | *string*  | The device id denoting of the auto-provisioned device. |
```
sophokles73, line 149: FMPOV this is mandatory because without it a consumer does not know which device we are talking about, does it?
We should not mention *auto-provisioned* here as the event is also applicable for manual provisioning cases ...

site/documentation/content/api/event/index.md
```diff
@@ -134,3 +130,21 @@ The example below might be used by the MQTT adapter to indicate that a connectio
...
+| *gateway_id*               | no               | *application-properties* | *string*  | The gateway id denoting the gateway for which the device was auto-provisioned. |
```
sophokles73, line 150: How about
> This property contains a value only if the device's registration status has been changed by a gateway acting on behalf of the device. In such a case, the property contains the identifier of the gateway, otherwise the property will not be included.

site/documentation/content/api/event/index.md
fkaltner: Yes, if we intend to use the event in a more generic fashion that is correct.

sophokles73 (approved):
LGTM

PR #2424, opened 2021-01-19
Fix queue full handling

sophokles73:
The AMQP based adapter client has been changed to throw a
ServerErrorException with a status code of 503 in case a downstream AMQP
container rejects a message with error condition
amqp:resource-limit-exceeded.

This allows protocol adapters to notify devices accordingly.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2021 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -39,6 +39,7 @@ import org.eclipse.hono.client.SendMessageSampler;
 import org.eclipse.hono.client.SendMessageTimeoutException;
 import org.eclipse.hono.client.ServerErrorException;
 import org.eclipse.hono.client.ServiceInvocationException;
+import org.eclipse.hono.client.StatusCodeMapper;
 import org.eclipse.hono.config.ClientConfigProperties;
 import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.MessageHelper;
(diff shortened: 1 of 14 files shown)


client/src/main/java/org/eclipse/hono/client/impl/AbstractSender.java
```diff
@@ -314,9 +315,15 @@ protected final Span startSpan(final Message message) {
...
                         final Rejected rejected = (Rejected) remoteState;
-                        e = rejected.getError() == null
-                                ? new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST)
-                                : new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST, rejected.getError().getDescription());
+                        e = Optional.ofNullable(rejected.getError())
+                                .map(error -> {
+                                    if (AmqpError.RESOURCE_LIMIT_EXCEEDED.equals(error.getCondition())) {
+                                        return new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, error.getDescription());
+                                    } else {
+                                        return new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST, error.getDescription());
```
calohmn, line 323: There actually already is a method mapping an errorCondition to a `ServiceInvocationException`:

https://github.com/eclipse/hono/blob/e0ca00cc8b2925c79e91f35cd07a3ae217beeb39/client/src/main/java/org/eclipse/hono/client/StatusCodeMapper.java#L122-L127

It is only used in integration tests and its original intention seems to have been, to map errors created when opening an AMQP link (i.e. the failed result of a link openHandler).
Using the FORBIDDEN code seems appropriate in such a scenario, in contrast to the scenario addressed in this PR.

Therefore I don't want to suggest using that method here. Instead, to avoid confusion, I think the `StatusCodeMapper` method could be renamed to something like `fromConnectError`.

sophokles73: Nice catch :+1: 
I will change the method name according to your suggestion ...

sophokles73: @calohmn I have pushed a corresponding change ...

calohmn: Shouldn't this then also use `StatusCodeMapper#fromTransferError`?

sophokles73: yes, it should

sophokles73:
@calohmn I have pushed corresponding changes. Hope I got it all covered now ;-)

calohmn (approved):
LGTM

PR #2426, opened 2021-01-20
Create AMQP adapter base module as a first step towards introducing Q…

dejanb:
…uarkus alternative


adapters/amqp-vertx-base/pom.xml
```diff
@@ -0,0 +1,35 @@
+<!--
+    Copyright (c) 2016, 2020 Contributors to the Eclipse Foundation
```
sophokles73, line 2: 2021?

sophokles73 (commented):
IMHO we should also move the *AmqpContext* class to the *org.eclipse.hono.adapter.amqp* package as part of this because it is not implementation specific ...

dejanb:
@sophokles73 fixed

sophokles73 (commented):
IMHO we cannot keep the *AmqpContext* class in the *org.eclipse.hono.adapter.amqp.impl* package because we would then end up with a *split package*, i.e. classes from the same package which need to be loaded from different artifacts (jars). This results in an error when loading classes form signed jars (as we  do) ...
That is why I proposed to move the class to *org.eclipse.hono.adapter.amqp* which will then be only contained in the amqp adapter base artifact, right?

dejanb:
@sophokles73 yeah, it makes sense. Fixed

sophokles73 (approved):
LGTM

PR #2427, opened 2021-01-20
Drop dependency on ProtocolAdpaterProperties

sophokles73:
Mayn classes of the adapter-amqp module required a
ProtocolAdapterProperties instance in their constructors. However, some
classes will also be used outside the context of a protocol adapter. In
such cases, it seems odd to provide a properties instance which does not
actually represent the configuration of a protocol adapter.

The base classes for implementing AMQP based clients have been changed
to require simple booleans instead of the ProtocolAdapterProperties in
their constructors. The concrete client implementations have been
updated accordingly.

### DIFF
--- a/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedCommandRouterClient.java
+++ b/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedCommandRouterClient.java
@@ -67,8 +67,6 @@ public class ProtonBasedCommandRouterClient extends AbstractRequestResponseServi
             final ProtocolAdapterProperties adapterConfig) {
         super(connection,
                 samplerFactory,
-                adapterConfig.isDefaultsEnabled(),
-                adapterConfig.isJmsVendorPropsEnabled(),
                 new CachingClientFactory<>(connection.getVertx(), RequestResponseClient::isOpen),
                 null);
         connection.getVertx().eventBus().consumer(
--- a/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedCommandRouterClient.java
+++ b/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedCommandRouterClient.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2020, 2021 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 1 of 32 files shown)


clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/registry/amqp/ProtonBasedDeviceRegistrationClient.java
```diff
@@ -76,8 +76,12 @@ public ProtonBasedDeviceRegistrationClient(
             final ProtocolAdapterProperties adapterConfig,
             final Cache<Object, RegistrationResult> responseCache) {
 
-        super(connection, samplerFactory, adapterConfig, new CachingClientFactory<>(
-                connection.getVertx(), RequestResponseClient::isOpen), responseCache);
+        super(connection,
+                samplerFactory,
+                adapterConfig.isDefaultsEnabled(),
+                adapterConfig.isJmsVendorPropsEnabled(),
```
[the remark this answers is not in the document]
sophokles73: > In this case here, this would mean whether the device registry implementation is JMS based and would actually need to read such properties. FMPOV, the latter isn't the case with our device registry APIs (the contentType being always JSON or empty anyway, depending on the operation).

I guess you are right here. We took great care to allow registry services being implemented using JMS without requiring the vendor properties being used.

> With the other comment above in mind, I think it would even be better to remove both new constructor params from AbstractServiceClient altogether and only use them directly in ProtonBasedDownstreamSender and ProtonBasedCommandResponseSender (usage is missing there still).

I guess you are right here as well. Great feedback :+1:

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedCommandRouterClient.java
```diff
@@ -65,9 +65,12 @@ public ProtonBasedCommandRouterClient(
             final HonoConnection connection,
             final SendMessageSampler.Factory samplerFactory,
             final ProtocolAdapterProperties adapterConfig) {
```
calohmn, line 67: The `adapterConfig` parameter is now not used anymore.

This also applies to:
ProtonBasedDeviceConnectionClient, ProtonBasedCredentialsClient, ProtonBasedDeviceRegistrationClient, ProtonBasedTenantClient
(so overall 5 classes)

sophokles73:

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedCommandRouterClient.java
sophokles73: damn, missed that ...

sophokles73:
@calohmn hope I got it all covered now ...

calohmn (approved):
LGTM

PR #2428, opened 2021-01-20
[#2273] Add usage of "response-expected" Kafka header

calohmn:
For #2273.


sophokles73 (approved):
LGTM

dguggemos:
Hi, Ditto already defines a header `response-required` with the same purpose. Using the same name for the same thing would surely be less confusing for users of Ditto and Hono.

calohmn:
I've chosen "response-expected" instead of "response-required" since Hono doesn't really require that there is a command response. I.e. there is no waiting for the response and hence no error if there is no response within a certain timeframe.

The "response-expected" only has an effect on the metrics (if the command is counted as request/response command) and on some different header sent to the device for certain protocols.

calohmn:
Since we have defined that
> An application should consider execution of a command to have failed, if it does not receive a response within a reasonable amount of time.

it actually fits better to name the header "response-required" from an application point of view.

The header has now been renamed accordingly - see #2452.

PR #2430, opened 2021-01-22
[#2273] Prepare base classes for implementing Kafka based command router

kaniyan:
`AbstractKafkaBasedDownstreamSender` is refactored and some methods are pulled up to the new base class `AbstractKafkaBasedMessageSender`. This is to prepare for the Kafka based command router component.

### DIFF
--- a/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/telemetry/kafka/AbstractKafkaBasedDownstreamSender.java
+++ b/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/telemetry/kafka/AbstractKafkaBasedDownstreamSender.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2020, 2021 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -13,66 +13,35 @@
 
 package org.eclipse.hono.adapter.client.telemetry.kafka;
 
-import java.net.HttpURLConnection;
 import java.time.Instant;
-import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.UUID;
-import java.util.stream.Collectors;
 
-import org.apache.kafka.clients.producer.ProducerConfig;
-import org.eclipse.hono.client.ServerErrorException;
 import org.eclipse.hono.client.kafka.HonoTopic;
 import org.eclipse.hono.client.kafka.KafkaProducerFactory;
-import org.eclipse.hono.client.kafka.tracing.KafkaTracingHelper;
+import org.eclipse.hono.client.kafka.producer.AbstractKafkaBasedMessageSender;
 import org.eclipse.hono.config.ProtocolAdapterProperties;
-import org.eclipse.hono.tracing.TracingHelper;
-import org.eclipse.hono.util.Lifecycle;
 import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.QoS;
 import org.eclipse.hono.util.RegistrationAssertion;
 import org.eclipse.hono.util.TenantObject;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
-import io.opentracing.References;
-import io.opentracing.Span;
 import io.opentracing.SpanContext;
 import io.opentracing.Tracer;
-import io.opentracing.tag.Tags;
 import io.vertx.core.Future;
-import io.vertx.core.Promise;
 import io.vertx.core.buffer.Buffer;
-import io.vertx.core.json.EncodeException;
-import io.vertx.core.json.Json;
-import io.vertx.kafka.client.producer.KafkaHeader;
-import io.vertx.kafka.client.producer.KafkaProducer;
-import io.vertx.kafka.client.producer.KafkaProducerRecord;
-import io.vertx.kafka.client.producer.RecordMetadata;
-import io.vertx.kafka.client.producer.impl.KafkaHeaderImpl;
 
 /**
- * A client for publishing messages to a Kafka cluster.
+ * A client for publishing downstream messages to a Kafka cluster.
  */
-public abstract class AbstractKafkaBasedDownstreamSender implements Lifecycle {
+public abstract class AbstractKafkaBasedDownstreamSender extends AbstractKafkaBasedMessageSender {
 
-    /**
-     * A logger to be shared with subclasses.
-     */
-    protected final Logger log = LoggerFactory.getLogger(getClass());
-
-    private final KafkaProducerFactory<String, Buffer> producerFactory;
-    private final String producerName;
-    private final Map<String, String> config;
     private final ProtocolAdapterProperties adapterConfig;
-    private final Tracer tracer;
 
     /**
-     * Creates a new Kafka-based telemetry sender.
+     * Creates a new Kafka-based downstream sender.
      *
      * @param producerFactory The factory to use for creating Kafka producers.
      * @param producerName The producer name to use.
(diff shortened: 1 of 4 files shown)


clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/telemetry/kafka/AbstractKafkaBasedMessageSender.java
```diff
@@ -0,0 +1,304 @@
...
+                }
+            });
+        }
+
+        return headers;
+    }
+
+    private Span startSpan(final String topic, final String tenantId, final String deviceId, final QoS qos,
+            final SpanContext context) {
+        final String referenceType = QoS.AT_MOST_ONCE.equals(qos) ? References.FOLLOWS_FROM : References.CHILD_OF;
```
b-abel, line 238: With my proposal above (providing separate `send` methods for both QoS levels), the QoS could be completely removed from the class and the reference type here determined depending on which `send` method is invoked.

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/telemetry/kafka/AbstractKafkaBasedMessageSender.java
```diff
@@ -0,0 +1,304 @@
...
+     *            the parent for any span it creates for tracing the execution of this operation.
+     * @return A future indicating the outcome of the operation.
+     *         <p>
+     *         The future will be succeeded if the message has been sent downstream.
+     *         <p>
+     *         The future will be failed with a {@link org.eclipse.hono.client.ServerErrorException} if the data could
+     *         not be sent. The error code contained in the exception indicates the cause of the failure.
+     * @throws NullPointerException if topic, tenantId, deviceId or qos are {@code null}.
+     */
+    public Future<Void> send(final String topic, final String tenantId, final String deviceId,
```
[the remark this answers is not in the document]
kaniyan: 1. I mainly focused on changes required from the perspective of the command router. I like this idea of having two methods namely `send()` and `sendAndWaitForOutcome()` rather than passing the QoS and handling internally inside a single send method. With these new methods, the intentions are made clear.

2. In case of command router, the headers in the `KafkaConsumerRecord` can be reused while forwarding messages to the internal command topic. By having methods that accept headers, it avoids unnecessary conversion to properties from headers and then convert again to headers. FMPOV its good to have the above new methods overloaded also to accept headers.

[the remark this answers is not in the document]
kaniyan: With the latest commit, I have introduced two types of methods namely `send()` and `sendAndWaitForOutcome()`.

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/tracing/KafkaTracingHelper.java
```diff
@@ -68,23 +67,23 @@ private KafkaTracingHelper() {
...
      */
-    public static Span newProducerSpan(final Tracer tracer, final HonoTopic topic, final String referenceType,
+    public static Span newProducerSpan(final Tracer tracer, final String topic, final String referenceType,
             final SpanContext parent) {
         Objects.requireNonNull(tracer);
         Objects.requireNonNull(topic);
         Objects.requireNonNull(referenceType);
 
-        return TracingHelper.buildSpan(tracer, parent, "To_" + topic.getType().endpoint, referenceType)
+        return TracingHelper.buildSpan(tracer, parent, "send message", referenceType)
```
b-abel, line 82: Could you use "kafka_send" (instead of "send message") as the operation name? This seems to be what the built-in tracing support of the Vert.x 4 Kafka client uses.

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/telemetry/kafka/AbstractKafkaBasedDownstreamSender.java
```diff
@@ -13,66 +13,42 @@
...
- * A client for publishing messages to a Kafka cluster.
+ * A client for publishing downstream messages to a Kafka cluster.
  */
-public abstract class AbstractKafkaBasedDownstreamSender implements Lifecycle {
+public abstract class AbstractKafkaBasedDownstreamSender extends AbstractKafkaBasedMessageSender {
 
     /**
      * A logger to be shared with subclasses.
      */
     protected final Logger log = LoggerFactory.getLogger(getClass());
```
b-abel, line 46: Why not use the logger provided by the parent class?

clients/kafka-common/src/main/java/org/eclipse/hono/kafka/client/tracing/KafkaTracingHelper.java
calohmn: @b-abel Do you see a practical advantage in using that pattern, being unlike the one used elsewhere in Hono?
FMPOV, having a common naming scheme across all spans created within Hono looks preferable.
In that sense, "send message" fits better IMHO.

b-abel: Vert.x 4.0 comes with [build-in](https://vertx.io/docs/vertx-kafka-client/java/#_automatic_trace_propagation) tracing support. I hope that we can use this and remove our custom tracing at least for the Kafka clients. Following the patterns used there could bring the advantage that the stored spans look the same instead of a sudden change.

calohmn: I think that remains to be seen, whether we should really be using the generic built-in tracing support with Vert.x 4 and abandon our Hono-specific tracing there (which still gives us more flexibility, I guess).
And if and when we do it, I think changing this one span name won't be a big deal.
Until then, I'd rather keep the span names consistent.

kaniyan:
The commits have been squashed and rebased due to conflicts with the merge of other PRs.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/producer/AbstractKafkaBasedMessageSender.java
```diff
@@ -0,0 +1,331 @@
...
+     * @param topic The topic to send the message to.
+     * @param tenantId The tenant that the device belongs to.
+     * @param deviceId The device identifier.
+     * @param payload The data to send.
+     * @param headers Additional meta data that should be included in the message.
+     * @param context The currently active OpenTracing span (may be {@code null}). An implementation should use this as
+     *            the parent for any span it creates for tracing the execution of this operation.
+     * @throws NullPointerException if topic, tenantId, deviceId or headers are {@code null}.
+     */
+    protected void send(final String topic, final String tenantId, final String deviceId, final Buffer payload,
```
calohmn, line 145: I was first wondering whether we should better use a `Future<Void>` return value here - similar to corresponding AMQP `AbstractSender` method. That would allow us to return a failed Future if the process of sending the message itself has failed (e.g. because the producer is already closed or there was a serialization error).

Unfortunately, the Vert.x Kafka client doesn't provide this feature, to distinguish between such a producer side error and an error related to the asynchronous acknowledgement.
So, for the time being, we have to accept not being able to get these early exceptions here and can only define a `void` send method here.

clients/kafka-common/pom.xml
```diff
@@ -64,7 +64,11 @@
       </exclusions>
     </dependency>
     <dependency>
-      <groupId>io.vertx</groupId>
+      <groupId>org.eclipse.hono</groupId>
+            <artifactId>hono-client</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.vertx</groupId>
```
calohmn, line 71: indentation

clients/kafka-common/pom.xml
```diff
@@ -64,7 +64,11 @@
       </exclusions>
     </dependency>
     <dependency>
-      <groupId>io.vertx</groupId>
+      <groupId>org.eclipse.hono</groupId>
+            <artifactId>hono-client</artifactId>
```
calohmn, line 68: Exclusions should be applied here like the ones for `hono-core` above (e.g. excluding `vertx-proton`).

kaniyan: I have included the exclusions now.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/producer/AbstractKafkaBasedMessageSender.java
kaniyan: Initially I had the same thought of returning `Future<Void>` and I didn't see any advantage in case of Vertx Kafka clients and hence changed to `void`.

calohmn (approved):
LGTM

PR #2431, opened 2021-01-22
Refactor Command Router based command consumer

calohmn:
The functionality of the `ProtonBasedCommandRouterCommandConsumerFactoryImpl` class (used in the protocol adapters to receive commands forwarded by the Command Router) has been moved into separate classes:
One generic `CommandRouterCommandConsumerFactory` and a proton based consumer class, providing the receiver on the internal command endpoint targeted by the Command Router. This is a preparation for adding a corresponding Kafka based consumer.

I've put the new Proton/Kafka-independent `CommandRouterCommandConsumerFactory` class into the `hono-client-adapter` module. Unfortunately, this required adding the hono-client module dependency there (needed for the `ServiceInvocationException` classes).
The alternative would be putting the class into the service-base module. But putting it in a `org.eclipse.hono.service.command` package there (corresponding to the location of the other classes in this module), as a class implementing the  `org.eclipse.hono.adapter.client.command.CommandConsumerFactory` interface, wouldn't fit that well.


clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/InternalCommandConsumer.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+package org.eclipse.hono.adapter.client.command;
+
+import org.eclipse.hono.util.Lifecycle;
+
+/**
+ * A client to receive commands to the internal Command and Control API endpoint provided by protocol adapters.
+ */
+public interface InternalCommandConsumer extends Lifecycle {
```
sophokles73, line 21: why do we need this in addition to the already existing *org.eclipse.hono.adapter.client.command.CommandConsumer*?

calohmn: The `InternalCommandConsumer` needs to extend `Lifecycle` and it doesn't need the `close(spanContext)` method of the `CommandConsumer` interface.

sophokles73: Then why don't we simply use Lifecycle directly? *InternalCommandConsumer* doesn't seem to define any other behavior anyway, does it?

calohmn: It was meant more as a marker interface. But ok, it doesn't provide much benefit here. I've removed it.

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterCommandConsumerFactory.java
```diff
@@ -11,100 +11,100 @@
...
-                    recreateConsumer();
-                });
+        @SuppressWarnings("rawtypes")
+        final List<Future> futures = internalCommandConsumers.stream()
+                .map(Lifecycle::start)
+                .collect(Collectors.toList());
+        if (futures.isEmpty()) {
+            return Future.failedFuture("no command consumer registered");
+        }
+        return CompositeFuture.all(futures).mapEmpty();
```
sophokles73, line 98: I do not understand this start-up behavior. When is the star method being invoked and by whom? Why should there be any *internalCommandConsumers* being (already) registered at this point? Why should start-up fail if there aren't any? IMHO all of this deserves proper documentation in the method's JavaDoc as it is far from obvious what this method is supposed to do ...

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterCommandConsumerFactory.java
```diff
@@ -11,100 +11,111 @@
...
+     * That is the endpoint that the Command Router forwards received commands to.
+     * <p>
+     * Note that this method needs to be called before invoking {@link #start()}. The {@link #start()} and
+     * {@link #stop()} methods of this factory will invoke the corresponding methods on the internal command consumer.
+     *
+     * @param internalCommandConsumerSupplier Function that returns the consumer. Parameters are the adapter instance
+     *            id that identifies the internal command endpoint and the command handlers to choose from when handling
+     *            a received command.
+     */
+    public void registerInternalCommandConsumer(
```
sophokles73, line 87: is this method intended to be invoked once only or several times?

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterCommandConsumerFactory.java
calohmn: The `CommandRouterCommandConsumerFactory` instance (or a `ProtonBasedDelegatingCommandConsumerFactory` instance) is kept in the `AbstractProtocolAdapterBase#commandConsumerFactory` field. 
In `AbstractProtocolAdapterBase`, the `start` method is invoked on that `commandConsumerFactory`.

The `CommandRouterCommandConsumerFactory` needs the (started) _InternalCommandConsumer(s)_ (proton and/or kafka based) to work. Therefore I've associated the startup of the consumer factory with the startup of the _InternalCommandConsumer(s)_. Otherwise, the `AbstractProtocolAdapterBase` class would have to deal with the startup of the _InternalCommandConsumer(s)_ explicitly (and only do so if a `CommandRouterCommandConsumerFactory`, not a `ProtonBasedDelegatingCommandConsumerFactory` is set).

I've adapted the javadocs of the `registerInternalCommandConsumer` method and the above `start` method to describe more clearly what's done and required.

clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterCommandConsumerFactory.java
calohmn: It's intended to be invoked once only if the Hono deployment uses an AMQP messaging solution *or* a Kafka cluster.
If both kinds of messaging solutions are configured for the adapter, it's intended to be invoked twice, once for each Kafka or Proton specific internal consumer implementation.

sophokles73: got it, thanks for explaining

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedInternalCommandConsumer.java
```diff
@@ -0,0 +1,191 @@
...
+        super(connection, SendMessageSampler.Factory.noop());
+
+        this.commandHandlers = Objects.requireNonNull(commandHandlers);
+        this.adapterInstanceId = Objects.requireNonNull(adapterInstanceId);
+        tracer = connection.getTracer();
+    }
+
+    @Override
+    public Future<Void> start() {
+        return super.start()
```
sophokles73, line 82: FMPOV we should not rely on any particular behavior of the super class' method's default implementation here. The *start* method is explicitly designed to be overridden and its default implementation could therefore be changed at any time as well. IMHO we should therefore make sure to start the connection here ourselves ...

calohmn: I think in the particular case of the `AbstractServiceClient` class, all its sub classes rely on the fact that invocation of `connection.connect()` and `connection.disconnect()` on startup and shutdown is done in `AbstractServiceClient`, so that there is no need to invoke these methods in start()/stop() in the sub classes.
In that sense, I would rather make this behaviour more explicit in the javadoc of `AbstractServiceClient` and then keep the invocation of `super.start()` in `ProtonBasedInternalCommandConsumer`.

Otherwise, if I explicitly call `connection.connect()` in `ProtonBasedInternalCommandConsumer#start`, it would look odd not to override the `stop()` method as well, calling `connection.disconnect()` there. But this method would then be the same as in `AbstractServiceClient`.

calohmn: I've adapted the javadoc of the start/stop methods of `AbstractServiceClient` to make the connect/disconnect invocation there more explicit.

sophokles73: I still do not agree with invoking super.start() here. Overriding AbstractServiceClient.start() without overriding AbstractServiceClient.stop() is perfectly alright IMHO as it simply means that the implementation of AbtstractServiceClient.stop() is still in effect and behaves as specified (in AbstractServiceClient).

But I will not be religious about it ...

sophokles73 (approved):

Issue #2392, opened 2021-01-05
Transport protocol agnostic interface for northbound Hono-Client


b-abel:
To use [Kafka for messaging](https://github.com/eclipse/hono/issues/8), we need a northbound client interface that does not depend on AMQP specific types, analogous to the changes in  #2156.

b-abel:
I would like to use the consumer from #2420 for the implementation of the Kafka-based northbound client. This is very little code then for telemetry and events.

b-abel:
PR #2432 offers a Kafka-based implementation of the protocol-agnostic northbound interfaces.

b-abel:
@sophokles73 What about the AMQP-based implementation? Strangling the _hono-cllient_ module is FMPOV best left with you. 
FMPOV, resolving the old "client" module best left in your hands. However, I could provide (as an intermediate step) an implementation that "wraps" the `org.eclipse.hono.client.application.ApplicationClientFactory` and provides it with the new interface `org.eclipse.hono.client.application.ApplicationClientFactory`. Does that make sense?

sophokles73:
@b-abel yes, you can leave the Proton based implementation to me.

sophokles73:
@b-abel @kaniyan Anything left to do here or can this be closed (and be added to 1.7.0)?

kaniyan:
FMPOV this can be closed. We have a separate issue #2478 for C&C.

PR #2432, opened 2021-01-25
[#2392] Implementation of the Kafka-based northbound Client

b-abel:
This PR belongs to #2392. It adds the implementation of the Kafka-Based northbound application client.


b-abel:
I changed the package names in the _hono-client-application*_ sub-modules to follow the pattern of the _clients_ module, according to [this comment](https://github.com/eclipse/hono/commit/4f32409ef080c5d5b0306a9e773136382b86a701#r46336494).

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/KafkaMessageProperties.java
```diff
@@ -36,16 +36,18 @@
...
         record.headers().forEach(header -> properties.put(header.key(), header.value()));
     }
 
     /**
      * {@inheritDoc}
+     * <p>
+     * The values in the map are of type {@link Buffer}.
      *
-     * @return The headers of the {@link ConsumerRecord}.
+     * @return The headers of the {@link KafkaConsumerRecord}.
```
sophokles73, line 50: is the map returned supposed to be writable as well?

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/DownstreamMessageConsumer.java
```diff
@@ -0,0 +1,119 @@
...
+
+        final String tenantId = getTenantId(record);
+        final String deviceId = record.key();
+        final MessageProperties properties = new KafkaMessageProperties(record);
+        final String contentType = getContentType(record.headers());
+        final MessageContext messageContext = new KafkaMessageContext(record);
+        final QoS qos = getQoS(record.headers());
+        final Buffer payload = record.value();
+
+        return new DownstreamMessageImpl(tenantId, deviceId, properties, contentType, messageContext, qos, payload);
```
sophokles73, line 87: IMHO it would improve encapsulation if you created a kafka specific DownstreamMessage which simply wrapped a KafkaConsumerRecord, thus hiding all the code for retrieving the property values within it ...
The DownstreamMessageImpl class can then be removed FMPOV.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaApplicationClientFactoryImpl.java
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * {@inheritDoc}
+     *
+     * @return a future indicating the outcome. When {@link TelemetryConsumer#start()} completes successfully, the
+     *         future will be completed with the consumer. Otherwise the future will fail with the cause.
+     */
+    @Override
+    public Future<MessageConsumer<DownstreamMessage>> createTelemetryConsumer(final String tenantId,
+            final Consumer<DownstreamMessage> telemetryConsumer, final Handler<Throwable> closeHandler) {
```
sophokles73, line 66: it looks strange to me that we use both `java.util.function.Consumer` and `io.vertx.core.Handler` in a single method. They serve the same purpose so IMHO we should either use Consumer or Handler but not both in parallel.
I also do not think that the closeHandler accepting a Throwabe provides much benefit because (a) the client has no clue what exceptions to expect and should be able to handle and (b) the whole point of introducing the abstract consumers is to decouple the client from the messaging infrastructure implementation in the first place, isn't it?
FMPOV we should replace it with a `Consumer<Void` or `Handler<Void>` and thus allow the client to react to the channel having been closed in a generic way.

clients/application-kafka/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright (c) 2020 Contributors to the Eclipse Foundation
```
sophokles73, line 3: 2021

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/DownstreamMessageConsumer.java
b-abel: Yes, makes sense.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaApplicationClientFactoryImpl.java
b-abel: > it looks strange to me that we use both `java.util.function.Consumer` and `io.vertx.core.Handler` in a single method. They serve the same purpose so IMHO we should either use Consumer or Handler but not both in parallel.

You are right.

> I also do not think that the closeHandler accepting a Throwabe provides much benefit because (a) the client has no clue what exceptions to expect and should be able to handle 

I'm not sure that this will work well. In the Kafka consumers, there are a couple of different error cases to consider which cannot necessarily be handled in the same way. I tried to group them here: https://github.com/eclipse/hono/blob/7a9e374af0a2f1acd81aa433c951efab0d9fc657/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java#L55
Starting a new consumer might help in some cases (e.g. if a consumer group rebalance takes very long, the offset commit may fail but succeed later on) but not in others (e.g. if the message handler throws an exception, it will probably be thrown on each consumer restart). Therefore I don't know what a completely generic error handling would look like. 
Maybe we should focus on informing the client on how to handle the errors. 

> and (b) the whole point of introducing the abstract consumers is to decouple the client from the messaging infrastructure implementation in the first place, isn't it?

That's what I aim for, but I think there are cases where it might be useful to have more control than the lowest common denominator offers.

sophokles73: But is there anything that could be done by client code to remedy the situation, given that it has very limited access to the underlying kafka-client anyway? If I stick to the API defined so far and register a close handler that can accept a Throwable, what can I actually do about it?

b-abel: I don't have the silver bullet here but I think that because the cause of the error is passed to the client application, it can do as much or as little as if it were directly controlling the underlying Kafka consumer. Thinkable options are retrying (by opening a new consumer), stopping the application, triggering an alarm, letting the Kubernetes health check fail, putting the message to a dead-letter-queue. Besides that, it could decide to write the error to logs, traces, or metrics.
FMPOV the critical point is to decide whether retrying or skipping a message makes sense which depends on the type of error. Retries do not come for free. If I have a consumer that polls 100 messages from a topic and fails to process the last one or, similarly, to commit the offset afterwards, then all the messages from that batch will be processed again with each retry. So, with simply restarting, the client application would constantly get the same 100 messages in a loop. Alternatively, if the using application does not know the cause, it could only completely stop processing messages, while there are a lot of potentially temporary root causes.

Some common error handling strategies for the handling of messages consumed from Kafka are provided by [Kafka Connect](https://www.confluent.io/blog/kafka-connect-deep-dive-error-handling-dead-letter-queues/) and, similarly, [Quarkus](https://quarkus.io/blog/kafka-failure-strategy/). They are:

1. Fail fast
2. Ignore message
3. Dead letter queue

I could imagine that at a later stage, when we have more experience with Kafka-based messaging in Hono and when the need becomes apparent, we could offer such strategies directly in the client. Until then, I would take hints in the documentation and leave the implementation to the using application.

b-abel: @sophokles73 I added documentation to the interface `KafkaApplicationClientFactory` about the error cases to expect in the close handler. Does this improve the situation from your POV?

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/KafkaApplicationClientFactory.java
```diff
@@ -14,10 +14,82 @@
...
+     * with an exception indicating the cause. There are error cases that might disappear later on and where it makes
+     * sense to create a new consumer and other cases that need to be resolved externally.
+     * <p>
+     * </p>
+     * ERROR CASES:
+     * <p>
+     * Errors can happen when polling, in message processing, and when committing the offset to Kafka. If a {@code poll}
+     * operation fails, the consumer will be closed and the close handler will be passed a
+     * {@link KafkaConsumerPollException} indicating the cause. If the message processing fails, the consumer will be
+     * closed and the exception will be passed to the close handler. If the offset commit fails, the consumer will be
```
sophokles73, line 47: it is unclear to me here, what you are referring to with *message processing fails*. Is this about the provided handler throwing an exception or is this referring to processing within the kafka-client? In any case, is there a specific type of exception being passed in to the close handler?

b-abel: > Is this about the provided handler throwing an exception or is this referring to processing within the kafka-client?

The former one. It is the exception thrown by the message handler. Do you see value in wrapping this exception?

sophokles73: No, and I also do not understand why the close handler will be invoked in this case in the first place. But if this is really necessary, then the wording should be changed to better describe which exception it is that is passed in. At least I did not understand without your explaining ...

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/EventConsumer.java
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Creates and starts a message consumer for events.
+     *
+     * @param kafkaConsumer The Kafka consumer to be exclusively used by this instance to consume records.
+     * @param config The Kafka consumer configuration properties to use.
+     * @param tenantId The tenant to consume events for.
+     * @param messageHandler The handler to be invoked for each message created from a record.
+     * @param closeHandler The handler to be invoked when the Kafka consumer has been closed due to an error.
+     * @return a future indicating the outcome. When {@link #start()} completes successfully, the future will be
+     *         completed with the consumer. Otherwise the future will fail with the cause.
```
sophokles73, line 52: we should add a `@throws NullPointerException` tag for the required parameters ...

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/EventConsumer.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public static Future<MessageConsumer> create(final KafkaConsumer<String, Buffer> kafkaConsumer,
+            final KafkaConsumerConfigProperties config, final String tenantId,
+            final Handler<DownstreamMessage<KafkaMessageContext>> messageHandler,
+            final Handler<Throwable> closeHandler) {
+
+        Objects.requireNonNull(kafkaConsumer);
```
sophokles73, line 59: IMHO the messageHandler is also required, isn't it?

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/TelemetryConsumer.java
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Creates and starts a message consumer for telemetry data.
+     *
+     * @param kafkaConsumer The Kafka consumer to be exclusively used by this instance to consume records.
+     * @param config The Kafka consumer configuration properties to use.
+     * @param tenantId The tenant to consume telemetry data for.
+     * @param messageHandler The handler to be invoked for each message created from a record.
+     * @param closeHandler The handler to be invoked when the Kafka consumer has been closed due to an error.
+     * @return a future indicating the outcome. When {@link #start()} completes successfully, the future will be
+     *         completed with the consumer. Otherwise the future will fail with the cause.
```
sophokles73, line 52: `@throws NullPointerException`?

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/TelemetryConsumer.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    public static Future<MessageConsumer> create(final KafkaConsumer<String, Buffer> kafkaConsumer,
+            final KafkaConsumerConfigProperties config, final String tenantId,
+            final Handler<DownstreamMessage<KafkaMessageContext>> messageHandler,
+            final Handler<Throwable> closeHandler) {
+
+        Objects.requireNonNull(kafkaConsumer);
```
sophokles73, line 59: messageHandler is also required

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/DownstreamMessageConsumer.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Creates a new downstream consumer.
+     *
+     * @param kafkaConsumer The Kafka consumer to be exclusively used by this instance to consume records.
+     * @param topic The Kafka topic to consume records from.
+     * @param messageHandler The handler to be invoked for each message created from a record.
+     * @param closeHandler The handler to be invoked when the Kafka consumer has been closed due to an error.
+     * @param pollTimeout The maximal number of milliseconds to wait for messages during a poll operation.
+     * @throws NullPointerException if any of the parameters is {@code null}.
```
sophokles73, line 45: IMHO closeHandler can be  `null`

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
```diff
@@ -91,15 +90,15 @@
      *
      * @param kafkaConsumer The Kafka consumer to be exclusively used by this instance to consume records.
      * @param topic The Kafka topic to consume records from.
-     * @param messageHandler The function to be invoked for each message created from a record.
+     * @param messageHandler The handler to be invoked for each message created from a record.
      * @param closeHandler The handler to be invoked when the Kafka consumer has been closed due to an error.
      * @param pollTimeout The maximal number of milliseconds to wait for messages during a poll operation.
      * @throws NullPointerException if any of the parameters is {@code null}.
```
sophokles73, line 96: closeHandler can be `null`

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
```diff
@@ -112,15 +111,15 @@ public AbstractAtLeastOnceKafkaConsumer(final KafkaConsumer<String, Buffer> kafk
      *
      * @param kafkaConsumer The Kafka consumer to be exclusively used by this instance to consume records.
      * @param topics The Kafka topics to consume records from.
-     * @param messageHandler The function to be invoked for each message created from a record.
+     * @param messageHandler The handler to be invoked for each message created from a record.
      * @param closeHandler The handler to be invoked when the Kafka consumer has been closed due to an error.
      * @param pollTimeout The maximal number of milliseconds to wait for messages during a poll operation.
      * @throws NullPointerException if any of the parameters is {@code null}.
```
sophokles73, line 117: closeHandler can be `null`

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
```diff
@@ -133,22 +132,22 @@ public AbstractAtLeastOnceKafkaConsumer(final KafkaConsumer<String, Buffer> kafk
      *
      * @param kafkaConsumer The Kafka consumer to be exclusively used by this instance to consume records.
      * @param topicPattern The pattern of Kafka topic names to consume records from.
-     * @param messageHandler The function to be invoked for each message created from a record.
+     * @param messageHandler The handler to be invoked for each message created from a record.
      * @param closeHandler The handler to be invoked when the Kafka consumer has been closed due to an error.
      * @param pollTimeout The maximal number of milliseconds to wait for messages during a poll operation.
      * @throws NullPointerException if any of the parameters is {@code null}.
```
sophokles73, line 138: closeHandler can be `null`

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/KafkaApplicationClientFactory.java
b-abel: I changed it so that the close handler is no longer invoked when the message handler throws an exception.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/DownstreamMessageConsumer.java
b-abel: I handled the case of a `closeHandler` that is `null` (as allowed by `ApplicationClientFactory`) on the highest possible level (`KafkaApplicationClientFactoryImpl`). FMPOV it does not make much sense to pass `null` for the `closeHandler`, because you would not be informed if the consumer stops working.

sophokles73 (approved):
LGTM

Can you squash to a reasonable amount of commits?

PR #2440, opened 2021-01-27
Add KafkaAdminClientConfigProperties class

calohmn:
Also provide a way to define common config properties used in all the producer, consumer and admin client types.

### DIFF
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.hono.client.kafka;
+
+import java.util.Map;
+import java.util.Objects;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Common configuration properties for Kafka clients.
+ */
+public abstract class AbstractKafkaConfigProperties {
+
+    protected final Logger log = LoggerFactory.getLogger(getClass());
+
+    /**
+     * Common client config properties.
+     */
+    protected Map<String, String> commonClientConfig;
+    /**
+     * Client id prefix to be applied if no {@code client.id} property is set in the common
+     * or specific client config properties.
+     */
+    protected String defaultClientIdPrefix;
+
+    /**
+     * Sets the common Kafka client config properties to be used.
+     *
+     * @param commonClientConfig The config properties.
+     * @throws NullPointerException if the config is {@code null}.
+     */
+    public final void setCommonClientConfig(final Map<String, String> commonClientConfig) {
+        this.commonClientConfig = Objects.requireNonNull(commonClientConfig);
+    }
+
+    /**
+     * Sets the prefix for the client ID that is passed to the Kafka server to allow application specific server-side
+     * request logging.
+     * <p>
+     * If the common or specific client config already contains a value for key {@code client.id}, that one
+     * will be used and the parameter here will be ignored.
+     *
+     * @param clientId The client ID prefix to set.
+     * @throws NullPointerException if the client ID is {@code null}.
+     */
+    public final void setDefaultClientIdPrefix(final String clientId) {
+        this.defaultClientIdPrefix = Objects.requireNonNull(clientId);
+    }
+
+    /**
+     * Overrides a property in the given map.
+     *
+     * @param config The map to set the property in.
+     * @param key The key of the property.
+     * @param value The property value.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     */
+    protected final void overrideConfigProperty(final Map<String, String> config, final String key, final String value) {
+        Objects.requireNonNull(config);
+        Objects.requireNonNull(key);
+        Objects.requireNonNull(value);
+
+        log.trace("setting Kafka config property [{}={}]", key, value);
+        final Object oldValue = config.put(key, value);
+        if (oldValue != null) {
+            log.debug("provided Kafka configuration contains property [{}={}], changing it to [{}]", key,
+                    oldValue, value);
+        }
+    }
+
+}
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaAdminClientConfigProperties.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaAdminClientConfigProperties.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.hono.client.kafka;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+import org.apache.kafka.clients.admin.AdminClientConfig;
+
+/**
+ * Configuration properties for Kafka admin clients.
+ * <p>
+ * This class is intended to be as agnostic to the provided properties as possible in order to be forward-compatible
+ * with changes in new versions of the Kafka client.
+ *
+ * @see <a href="https://kafka.apache.org/documentation/#adminclientconfigs">Kafka AdminClient Configs</a>
+ * @see <a href="https://www.eclipse.org/hono/docs/api/telemetry-kafka">Telemetry API for Kafka Specification</a>
+ * @see <a href="https://www.eclipse.org/hono/docs/api/event-kafka">Event API for Kafka Specification</a>
+ * @see <a href="https://www.eclipse.org/hono/docs/api/command-and-control-kafka/">Command &amp; Control API for Kafka Specification</a>
+ */
+public class KafkaAdminClientConfigProperties extends AbstractKafkaConfigProperties {
+
+    private Map<String, String> adminClientConfig;
+
+    /**
+     * Sets the Kafka admin client config properties to be used.
+     *
+     * @param adminConfig The config properties.
+     * @throws NullPointerException if the config is {@code null}.
+     */
+    public final void setAdminClientConfig(final Map<String, String> adminConfig) {
+        this.adminClientConfig = Objects.requireNonNull(adminConfig);
+    }
+
+    /**
+     * Checks if a configuration has been set.
+     *
+     * @return true if configuration is present.
+     */
+    public final boolean isConfigured() {
+        return commonClientConfig != null || adminClientConfig != null;
+    }
+
+    /**
+     * Gets the Kafka admin client configuration to which additional properties were applied. The following properties are
+     * set here to the given configuration:
+     * <ul>
+     * <li>{@code client.id} if the property is not already present in the configuration and a value has been set with
+     * {@link #setDefaultClientIdPrefix(String)}, this value will be taken</li>
+     * </ul>
+     *
+     * @return a copy of the admin client configuration with the applied properties or an empty map if neither an admin
+     *         client configuration was set with {@link #setAdminClientConfig(Map)} nor common configuration properties were
+     *         set with {@link #setCommonClientConfig(Map)}.
+     */
+    public final Map<String, String> getAdminClientConfig() {
+
+        if (commonClientConfig == null && adminClientConfig == null) {
+            return Collections.emptyMap();
+        }
+
+        final Map<String, String> newConfig = new HashMap<>();
+        if (commonClientConfig != null) {
+            newConfig.putAll(commonClientConfig);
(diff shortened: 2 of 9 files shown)


clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaProducerConfigProperties.java
```diff
@@ -83,47 +66,42 @@ public boolean isConfigured() {
...
      */
     public Map<String, String> getProducerConfig() {
 
-        if (producerConfig == null) {
+        if (commonClientConfig == null && producerConfig == null) {
             return Collections.emptyMap();
         }
 
-        final HashMap<String, String> newConfig = new HashMap<>(producerConfig);
+        final HashMap<String, String> newConfig = new HashMap<>();
```
sophokles73, line 84: Map ?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaAdminClientConfigProperties.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private Map<String, String> adminClientConfig;
+
+    /**
+     * Sets the Kafka admin client config properties to be used.
+     *
+     * @param adminConfig The config properties.
+     * @throws NullPointerException if the config is {@code null}.
+     */
+    public void setAdminClientConfig(final Map<String, String> adminConfig) {
```
sophokles73, line 44: final ?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaAdminClientConfigProperties.java
```diff
@@ -0,0 +1,89 @@
...
+        this.adminClientConfig = Objects.requireNonNull(adminConfig);
+    }
+
+    /**
+     * Checks if a configuration has been set.
+     *
+     * @return true if configuration is present.
+     */
+    public boolean isConfigured() {
```
sophokles73, line 53: final ?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaAdminClientConfigProperties.java
```diff
@@ -0,0 +1,89 @@
...
+     * <ul>
+     * <li>{@code client.id} if the property is not already present in the configuration and a value has been set with
+     * {@link #setClientIdFallback(String)}, this value will be taken</li>
+     * </ul>
+     *
+     * @return a copy of the admin client configuration with the applied properties or an empty map if neither an admin
+     *         client configuration was set with {@link #setAdminClientConfig(Map)} nor common configuration properties were
+     *         set with {@link #setCommonClientConfig(Map)}.
+     */
+    public Map<String, String> getAdminClientConfig() {
```
sophokles73, line 69: final ?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaAdminClientConfigProperties.java
```diff
@@ -0,0 +1,89 @@
...
+
+        if (commonClientConfig == null && adminClientConfig == null) {
+            return Collections.emptyMap();
+        }
+
+        final HashMap<String, String> newConfig = new HashMap<>();
```
sophokles73, line 75: Map ?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaProducerConfigProperties.java
```diff
@@ -50,26 +46,13 @@ public void setProducerConfig(final Map<String, String> producerConfig) {
...
-    }
-
     /**
      * Checks if a configuration has been set.
      *
      * @return true if configuration is present.
      */
     public boolean isConfigured() {
-        return producerConfig != null;
+        return commonClientConfig != null || producerConfig != null;
```
sophokles73, line 55: final method?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaProducerConfigProperties.java
```diff
@@ -83,47 +66,42 @@ public boolean isConfigured() {
...
+     * @return a copy of the producer configuration with the applied properties or an empty map if neither a producer
+     *         client configuration was set with {@link #setProducerConfig(Map)} nor common configuration properties were
+     *         set with {@link #setCommonClientConfig(Map)}.
      * @see <a href="https://kafka.apache.org/documentation/#enable.idempotence">The Kafka documentation -
      *      "Producer Configs" - enable.idempotence</a>
      */
     public Map<String, String> getProducerConfig() {
 
-        if (producerConfig == null) {
+        if (commonClientConfig == null && producerConfig == null) {
```
sophokles73, line 80: final method?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/KafkaConsumerConfigProperties.java
```diff
@@ -56,26 +52,13 @@ public void setConsumerConfig(final Map<String, String> consumerConfig) {
...
-    public final void setClientId(final String clientId) {
-        this.clientId = Objects.requireNonNull(clientId);
-    }
-
     /**
      * Checks if a configuration has been set.
      *
      * @return true if configuration is present.
      */
     public boolean isConfigured() {
```
sophokles73, line 60: final ?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/KafkaConsumerConfigProperties.java
```diff
@@ -87,28 +70,35 @@ public boolean isConfigured() {
...
+     * {@link #setClientIdFallback(String)}, this value will be taken</li>
      * </ul>
      *
-     * @return a copy of the consumer configuration with the applied properties an empty map if no consumer
-     *         configuration was set with {@link #setConsumerConfig(Map)}.
+     * @return a copy of the consumer configuration with the applied properties or an empty map if neither a consumer
+     *         client configuration was set with {@link #setConsumerConfig(Map)} nor common configuration properties were
+     *         set with {@link #setCommonClientConfig(Map)}.
      */
     public Map<String, String> getConsumerConfig() {
```
sophokles73, line 80: final ?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/KafkaConsumerConfigProperties.java
```diff
@@ -87,28 +70,35 @@ public boolean isConfigured() {
...
 
-        if (consumerConfig == null) {
+        if (commonClientConfig == null && consumerConfig == null) {
             return Collections.emptyMap();
         }
 
-        final HashMap<String, String> newConfig = new HashMap<>(consumerConfig);
+        final HashMap<String, String> newConfig = new HashMap<>();
```
sophokles73, line 86: Map ?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Common client config properties.
+     */
+    protected Map<String, String> commonClientConfig;
+    /**
+     * Client id to be applied if no {@code client.id} property is set in the common
+     * or specific client config properties.
+     */
+    protected String clientIdFallback;
```
b-abel, line 37: Instead of "fallback", I would rather call it a "default". Usually, this should not be set in config but in code. 
And: since each client needs an individual one (meaning each verticle) they are suffixed by a UUID (at least for producers, I'll add this for the consumer). So, maybe `clientIdDefaultPrefix` might be a better name. WDYT?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Sets the common Kafka client config properties to be used.
+     *
+     * @param commonClientConfig The config properties.
+     * @throws NullPointerException if the config is {@code null}.
+     */
+    public final void setCommonClientConfig(final Map<String, String> commonClientConfig) {
```
b-abel, line 45: Having this abstract base class for all types of Kafka client configuration makes a lot of sense to me.

I'm not convinced, however, that the introduction of a common section where you configure properties for _all_ producers and consumers in an application provides many benefits, but it definitely makes code and documentation more complex and you need to have more knowledge for the configuration. Then there are only a few properties that can be set on both and having them separately, even if they could be equal, does not hurt.

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
```diff
@@ -261,7 +263,37 @@ public ClientConfigProperties downstreamSenderConfig() {
...
+    }
+
+    /**
+     * Exposes configuration properties for an admin client accessing the Kafka cluster as a Spring bean.
+     *
+     * @return The properties.
+     */
+    @ConfigurationProperties(prefix = "hono.kafka")
+    @Bean
+    public KafkaAdminClientConfigProperties kafkaAdminClientConfig() {
```
b-abel, line 293: Just out of curiosity, what is the Admin client used for by the protocol adapter?

calohmn: It is to be used for creating the `hono.command_internal.${adapterInstanceId}` topic (with one partition) on startup, and to delete this topic on shutdown. 
That would at least be the preferred setup. The topic could also be auto-created but then there would be the need for cleanup when the adapter instance isn't there anymore.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
calohmn: My intention was mainly to prevent having to have three identical sets of `bootstrap.servers` and `ssl.*` property sections in each protocol adapter properties. It's about one and the same Kafka cluster here anyhow (and authentication would also be the same, at least for producer and consumer, I guess).
Usage of the common properties is also optional, so, for a configuration with only one of the client types, it doesn't have to be used.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
calohmn: Right, "default" is better. I would use `defaultClientIdPrefix` then.

calohmn: Using `prefix` here might be a bit confusing though, because it isn't used as a prefix here. The UUID suffix is applied at a different place.
How about changing this the getProducerConfig/getConsumerConfig/getAdminClientConfig methods to already include the UUID and an additional "name" parameter?

b-abel: I see the point with the term "prefix". You mean "name" as a third part of the effective client ID, like this is done in the following snipped? 
https://github.com/eclipse/hono/blob/18ba130bb51610c3f546a9d21621ef4a0e7aa1ad/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/producer/AbstractKafkaBasedMessageSender.java#L246

And: when to generate the UUID? It must always be ensured that two clients cannot have the same client ID.

service-base/src/main/java/org/eclipse/hono/service/AbstractAdapterConfig.java
b-abel: > It is to be used for creating the `hono.command_internal.${adapterInstanceId}` topic (with one partition) on startup, and to delete this topic on shutdown.

Ah, thanks. That makes sense.

> The topic could also be auto-created but then there would be the need for cleanup when the adapter instance isn't there anymore.

Yes, this would be suboptimal. Kafka admins might want to disable the broker config `auto.create.topics.enable` (and for managed Kafka instances there might be no way to enable this).

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
calohmn: Yes, that's the snippet I mean.

> And: when to generate the UUID?

Generate it when it is needed in `KafkaConsumerConfigProperties#getConsumerConfig()`.
Anyway, I think that change can be done in a separate PR.
As a preparation, I've already renamed the field/setter to `defaultClientIdPrefix` here.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * Overrides a property in the given map.
+     *
+     * @param config The map to set the property in.
+     * @param key The key of the property.
+     * @param value The property value.
+     */
+    protected final void overrideConfigProperty(final Map<String, String> config, final String key, final String value) {
```
kaniyan, line 70: How about _null_ check for the parameters?

calohmn:
I've made the suggested changes.

kaniyan (approved):
LGTM

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
calohmn: @b-abel I've made the aforementioned changes in PR https://github.com/eclipse/hono/pull/2441.

PR #2441, opened 2021-01-27
Set a unique client.id property in AbstractKafkaConfigProperties

calohmn:
Corresponding discussion:
https://github.com/eclipse/hono/pull/2440#discussion_r565363789

### DIFF
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/CachingKafkaProducerFactory.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/CachingKafkaProducerFactory.java
@@ -40,7 +40,8 @@ import io.vertx.kafka.client.producer.KafkaProducer;
  * Producers are closed and removed from the cache if they throw a {@link #isFatalError(Throwable) fatal exception}.
  * This is triggered by {@link KafkaProducer#exceptionHandler(Handler)} and run asynchronously after the
  * {@link io.vertx.kafka.client.producer.impl.KafkaWriteStreamImpl#send(ProducerRecord, Handler) send operation} has
- * finished. A following invocation of {@link #getOrCreateProducer(String, Map)} will then return a new instance.
+ * finished. A following invocation of {@link #getOrCreateProducer(String, KafkaProducerConfigProperties)} will then
+ * return a new instance.
  *
  * @param <K> The type for the record key serialization.
  * @param <V> The type for the record value serialization.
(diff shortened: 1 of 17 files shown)


b-abel:
I'm a little worried about name conflicts with the client ID. If a producer connects to a Kafka cluster and presents a client ID that is already known in the cluster, the producer will be rejected. For consumers, the situation is likely to be analogous. The mechanism in the PR uses the UUID to ensure that there are no conflicts between the IDs of different verticles (and application instances). But the question is how can we ensure that every time a client is terminated and reopened, a new client ID is generated. In `AbstractKafkaBasedMessageSender` this is done every time `start()` is invoked, with this PR it only happens at instantiation. There you can solve the situation by passing in the config object and invoking `getProducerConfig()` in `start()` each time. It would be nicer to have an API that enforces that, but at the moment I have no idea how this could be acchieved.

calohmn:
How about using `KafkaProducerConfigProperties` instead of `Map<String, String>` in the `AbstractKafkaBasedDownstreamSender` and `AbstractKafkaBasedMessageSender` constructors then?

Then there should also be a `KafkaProducerFactory#getOrCreateProducer` method with a `KafkaProducerConfigProperties` instead of Map parameter.

b-abel:
> How about using `KafkaProducerConfigProperties` instead of `Map<String, String>` in the `AbstractKafkaBasedDownstreamSender` and `AbstractKafkaBasedMessageSender` constructors then?
> 
> Then there should also be a `KafkaProducerFactory#getOrCreateProducer` method with a `KafkaProducerConfigProperties` instead of Map parameter.

Yes. That is what I meant. I would _replace_ the map parameter in all of the classes you mentioned. 

This does not solve the general problem that it is easy to use the interface of the config properties classes wrong, as you cannot _enforce_ that for each new Kafka client instance new properties are fetched. So, this should be at least documented very clearly at the getters (`getProducerConfig()` etc.).

calohmn:
@b-abel I've amended the commit with the added `getProducerConfig` javadoc and added a 2nd commit containing the changes towards using `KafkaProducerConfigProperties` instead of a Map.

calohmn:
(rebased)

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/CachingKafkaProducerFactory.java
```diff
@@ -86,6 +86,29 @@ public CachingKafkaProducerFactory(
...
+     * a new instance is created using the given factory and put to the cache.
+     * <p>
+     * The given config is ignored when an existing producer is returned.
+     *
+     * @param producerName The name to identify the producer.
+     * @param config The Kafka configuration with which the producer is to be created.
+     * @return an existing or new producer.
+     */
+    @Override
+    public KafkaProducer<K, V> getOrCreateProducer(final String producerName, final KafkaProducerConfigProperties config) {
```
b-abel, line 102: FMPOV the other variant with the map parameter should be entirely be removed.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaProducerFactory.java
```diff
@@ -67,6 +67,25 @@
...
+     * <p>
+     * When the producer is no longer needed, it should be closed to release the resources (like connections). NB: Do
+     * not close the returned producer directly. Instead, call {@link #closeProducer(String)} to let the factory close
+     * the producer and update its internal state.
+     *
+     * @param producerName The name to identify the producer.
+     * @param config The Kafka configuration with which the producer is to be created.
+     * @return an existing or new producer.
+     */
+    KafkaProducer<K, V> getOrCreateProducer(String producerName, KafkaProducerConfigProperties config);
```
b-abel, line 87: FMPOV the other variant with the map parameter should be entirely be removed.

calohmn: done

calohmn:
@sophokles73 @b-abel Further comments here?

sophokles73:
not from my side

b-abel:
> @sophokles73 @b-abel Further comments here?

no

PR #2443, opened 2021-01-28
Implement Quarkus based AMQP protocol adapter

dejanb:


sophokles73:
@dejanb can you rebase? I think/hope that the CI build will then succeed again ...

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/Application.java
```diff
@@ -10,7 +10,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  *******************************************************************************/
-package org.eclipse.hono.adapter.amqp.impl;
+package org.eclipse.hono.adapter.amqp;
```
sophokles73, line 13: IMHO this will create the *split package* problem when loading classes of the same package from two different signed jar files ...
For the other adapters that you have migrated already, we kept the spring related classes in the `impl` package, so I suggest that we do that here as well.

adapters/amqp-vertx/src/main/java/org/eclipse/hono/adapter/amqp/Config.java
```diff
@@ -10,13 +10,14 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  *******************************************************************************/
-package org.eclipse.hono.adapter.amqp.impl;
+package org.eclipse.hono.adapter.amqp;
```
sophokles73, line 13: see above

sophokles73 (commented):
IMHO we should also add logging configuration to the `tests/src/test/resources/amqp/application.yaml` file as we did for the HTTP and MQTT adapters ...

adapters/amqp-vertx-quarkus/src/main/java/org/eclipse/hono/adapter/amqp/quarkus/QuarkusAmqpProtocolAdapterProperties.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package org.eclipse.hono.adapter.amqp.quarkus;
+
+import org.eclipse.hono.adapter.amqp.AmqpAdapterProperties;
+
+import io.quarkus.arc.config.ConfigProperties;
+import io.quarkus.arc.config.ConfigProperties.NamingStrategy;
+
+/**
+ * Configuration properties for the HTTP adapter.
```
calohmn, line 21: AMQP Adapter

sophokles73 (approved):
LGTM

calohmn (approved):

PR #2444, opened 2021-01-28
[#8] Commit offsets when the Kafka consumer is interrupted.

b-abel:
This change aims to reduce the cases where the same Kafka record is consumed a second time.

If an error occurs while processing a message or when the consumer is stopped, the offsets for the already processed messages from the current batch are now committed. This is to prevent all messages from being processed again. Especially in case of errors during message processing, the same messages could otherwise be consumed again and again.

### DIFF
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
@@ -319,8 +319,7 @@ public abstract class AbstractAtLeastOnceKafkaConsumer<T> implements Lifecycle {
 
     private void closeAndCallHandler(final Throwable exception) {
         LOG.error("Closing consumer with cause", exception);
-        closeHandler.handle(exception);
-        closeConsumer();
+        closeConsumer().onComplete(v -> closeHandler.handle(exception));
     }
 
     private Future<Void> tryCommitAndClose() {
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
@@ -247,7 +247,7 @@ public abstract class AbstractAtLeastOnceKafkaConsumer<T> implements Lifecycle {
                 messageHandler.handle(message);
                 addToCurrentOffsets(record);
             }
-            commit(true).compose(ok -> poll()).onSuccess(this::handleBatch);
+            commit(true).compose(ok -> poll()).onSuccess(this::handleBatch); // this is "the loop"
         } catch (final Exception ex) {
             LOG.error("Error handling record, closing the consumer: ", ex);
             tryCommitAndClose().onComplete(v -> closeHandler.handle(ex));
@@ -284,13 +284,7 @@ public abstract class AbstractAtLeastOnceKafkaConsumer<T> implements Lifecycle {
             return Future.failedFuture("consumer already stopped"); // stop consuming
         }
 
-        final Promise<Void> commitPromise = Promise.promise();
-        kafkaConsumer.commit(commitPromise);
-        return commitPromise.future()
-                .onSuccess(ok -> {
-                    LOG.debug("Committed offsets");
-                    offsetsToBeCommitted.clear();
-                })
+        return commitCurrentOffsets()
                 .recover(cause -> {
                     LOG.error("Error committing offsets: " + cause);
                     if ((cause instanceof TimeoutException) && retry) {
@@ -304,7 +298,7 @@ public abstract class AbstractAtLeastOnceKafkaConsumer<T> implements Lifecycle {
                 });
     }
 
-    private Future<Map<TopicPartition, OffsetAndMetadata>> commitCurrentOffsets() {
+    private Future<Void> commitCurrentOffsets() {
         if (offsetsToBeCommitted.isEmpty()) {
             LOG.debug("no offsets to commit");
             return Future.succeededFuture();
@@ -315,7 +309,12 @@ public abstract class AbstractAtLeastOnceKafkaConsumer<T> implements Lifecycle {
 
         final Promise<Map<TopicPartition, OffsetAndMetadata>> completionHandler = Promise.promise();
         kafkaConsumer.commit(offsetsToBeCommitted, completionHandler);
-        return completionHandler.future();
+        return completionHandler.future()
+                .map(committedOffsets -> {
+                    LOG.debug("successfully committed offsets");
+                    offsetsToBeCommitted.clear();
+                    return null;
+                });
     }
 
     private void closeAndCallHandler(final Throwable exception) {
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
@@ -220,21 +220,7 @@ public abstract class AbstractAtLeastOnceKafkaConsumer<T> implements Lifecycle {
      */
     @Override
     public Future<Void> stop() {
-        final Promise<Void> returnFuture = Promise.promise();
-
-        commitCurrentOffsets()
-                .onComplete(commitResult -> {
-                    closeConsumer()
-                            .onComplete(v -> {
-                                if (commitResult.succeeded()) {
-                                    returnFuture.complete();
-                                } else {
-                                    returnFuture.fail(commitResult.cause());
-                                }
-                            });
-                });
-
-        return returnFuture.future();
+        return tryCommitAndClose();
     }
 
     private void handleBatch(final KafkaConsumerRecords<String, Buffer> records) {
@@ -254,7 +240,7 @@ public abstract class AbstractAtLeastOnceKafkaConsumer<T> implements Lifecycle {
                 .recover(cause -> {
                     LOG.error("Error polling messages: " + cause);
                     final KafkaConsumerPollException exception = new KafkaConsumerPollException(cause);
-                    closeWithError(exception);
+                    closeAndCallHandler(exception);
                     return Future.failedFuture(exception);
                 });
     }
@@ -275,7 +261,7 @@ public abstract class AbstractAtLeastOnceKafkaConsumer<T> implements Lifecycle {
             return Future.succeededFuture();
         } catch (final Exception ex) {
             LOG.error("Error handling record, closing the consumer: ", ex);
-            stop(); // commit currentOffsets and close consumer
+            tryCommitAndClose();
             return Future.failedFuture(ex);
         }
     }
@@ -302,7 +288,7 @@ public abstract class AbstractAtLeastOnceKafkaConsumer<T> implements Lifecycle {
                         return commit(false); // retry once
                     } else {
                         final KafkaConsumerCommitException exception = new KafkaConsumerCommitException(cause);
-                        closeWithError(exception);
+                        closeAndCallHandler(exception);
                         return Future.failedFuture(exception);
                     }
                 });
@@ -317,10 +303,30 @@ public abstract class AbstractAtLeastOnceKafkaConsumer<T> implements Lifecycle {
         return completionHandler.future();
     }
 
-    private void closeWithError(final Throwable exception) {
+    private void closeAndCallHandler(final Throwable exception) {
         LOG.error("Closing consumer with cause", exception);
         closeHandler.handle(exception);
         closeConsumer();
+    }
+
+    private Future<Void> tryCommitAndClose() {
+        final Promise<Void> returnFuture = Promise.promise();
+
+        commitCurrentOffsets()
+                .onComplete(commitResult -> {
+                    // always close consumer
+                    closeConsumer()
+                            .onComplete(v -> {
+                                // return outcome of commit
+                                if (commitResult.succeeded()) {
+                                    returnFuture.complete();
+                                } else {
+                                    returnFuture.fail(commitResult.cause());
+                                }
+                            });
+                });
+
+        return returnFuture.future();
     }
 
     private Future<Void> closeConsumer() {
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
@@ -246,7 +246,7 @@ public abstract class AbstractAtLeastOnceKafkaConsumer<T> implements Lifecycle {
                     addToCurrentOffsets(record);
                 } catch (final RuntimeException messageHandlingError) {
                     LOG.debug("Message handler failed", messageHandlingError);
-                    // will commit the offset of the failed record and then resume polling (will include failed record)
+                    // resume with committing the current offsets and then polling (failed record will be polled again)
                     break;
                 }
             }
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
@@ -21,6 +21,7 @@ import java.util.Set;
 import java.util.regex.Pattern;
 
 import org.apache.kafka.common.errors.TimeoutException;
+import org.eclipse.hono.client.ServerErrorException;
 import org.eclipse.hono.util.Lifecycle;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -41,11 +42,13 @@ import io.vertx.kafka.client.consumer.OffsetAndMetadata;
  * This consumer continuously polls for batches of messages from Kafka. Each message is passed to a message handler for
  * processing. When all messages of a batch are processed, this consumer commits the offset and polls for a new batch.
  * <p>
- * This consumer provides AT LEAST ONCE message processing. When the message handler completes, the message is
- * considered to be processed and will be "acknowledged" with the following commit. If an error occurs, the underlying
- * Kafka consumer will be closed. In this case, other consumer instances in the same consumer group will consume the
- * messages from the last committed offset. Already processed messages will then be again in the next batch and be
- * passed to the message handler. If de-duplication of messages is required, it must be handled by the message handler.
+ * This consumer provides AT LEAST ONCE message processing. When the message handler completes without throwing an
+ * exception, the message is considered to be processed and will be "acknowledged" with the following commit.
+ * <p>
+ * If a fatal error occurs, the underlying Kafka consumer will be closed. In this case, other consumer instances in the
+ * same consumer group will consume the messages from the last committed offset. Already processed messages might be
+ * polled again with the next batch and then passed to the message handler. If de-duplication of messages is required,
+ * it must be handled by the message handler.
  * <p>
  * The consumer starts consuming when {@link #start()} is invoked. It needs to be closed by invoking {@link #stop()} to
  * release the resources. A stopped instance cannot be started again.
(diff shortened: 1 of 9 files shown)


sophokles73 (approved):
LGTM

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
```diff
@@ -256,43 +257,83 @@ private void handleBatch(final KafkaConsumerRecords<String, Buffer> records) {
...
             final T message = createMessage(record);
             messageHandler.handle(message);
+            addToCurrentOffsets(record);
             return Future.succeededFuture();
         } catch (final Exception ex) {
             LOG.error("Error handling record, closing the consumer: ", ex);
-            commitCurrentOffset(record); // prevents that already processed records from the batch are consumed again
-            stop();
+            tryCommitAndClose();
             return Future.failedFuture(ex);
```
calohmn, line 265: In case this "catch" block getting run, will the failed Future here cause the block with "LOG.error("Error polling messages: " + cause);" inside `poll`  above to be run?
If so, `close` would get called twice.
If no, the `closeHandler` wouldn't get called here.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
```diff
@@ -209,17 +211,16 @@ private AbstractAtLeastOnceKafkaConsumer(final KafkaConsumer<String, Buffer> kaf
...
-     * @return A future that will complete when the Kafka consumer is closed.
+     * @return A future that will complete when the Kafka consumer is closed. If the offsets have been committed
+     *         successfully, the future will succeed, otherwise, it will fail.
      */
     @Override
     public Future<Void> stop() {
-        final Promise<Void> promise = Promise.promise();
-        kafkaConsumer.close(promise);
-        return promise.future();
+        return tryCommitAndClose();
```
calohmn, line 223: When testing this, I always seem to be getting an exception when stopping the consumer:
````
# consumer.stop() called here
12:53:45.061 [vert.x-eventloop-thread-0] DEBUG o.e.h.c.k.c.AbstractAtLeastOnceKafkaConsumer - committing the current offsets
12:53:45.425 [vert.x-eventloop-thread-0] DEBUG o.e.h.c.k.c.AbstractAtLeastOnceKafkaConsumer - Polled 0 records
12:53:45.445 [vert.x-eventloop-thread-0] ERROR o.e.h.c.k.c.AbstractAtLeastOnceKafkaConsumer - Error committing offsets: org.apache.kafka.common.errors.WakeupException
12:53:45.458 [vert.x-kafka-consumer-thread-0] INFO  o.a.k.c.c.i.ConsumerCoordinator - [Consumer clientId=test-a5bee52f-ee67-4381-b682-7c62b95bad76, groupId=myGroup] Revoke previously assigned partitions hono.command_internal.myAdapterInstanceId-0
12:53:45.459 [vert.x-kafka-consumer-thread-0] INFO  o.a.k.c.c.i.AbstractCoordinator - [Consumer clientId=test-a5bee52f-ee67-4381-b682-7c62b95bad76, groupId=myGroup] Member test-a5bee52f-ee67-4381-b682-7c62b95bad76-d9a878d8-899f-4819-8ca5-fe6b13e018f3 sending LeaveGroup request to coordinator localhost:49186 (id: 2147483646 rack: null) due to the consumer is being closed
12:53:45.462 [vert.x-eventloop-thread-0] ERROR o.e.h.c.k.c.AbstractAtLeastOnceKafkaConsumer - Closing consumer with cause
org.eclipse.hono.client.kafka.consumer.KafkaConsumerCommitException: org.apache.kafka.common.errors.WakeupException
	at org.eclipse.hono.client.kafka.consumer.AbstractAtLeastOnceKafkaConsumer.lambda$commit$6(AbstractAtLeastOnceKafkaConsumer.java:290)
	at io.vertx.core.Future.lambda$recover$6(Future.java:499)
	at io.vertx.core.impl.FutureImpl$Handlers.handle(FutureImpl.java:228)
	at io.vertx.core.impl.FutureImpl$Handlers.handle(FutureImpl.java:224)
	at io.vertx.core.impl.FutureImpl.tryFail(FutureImpl.java:195)
	at io.vertx.core.impl.FutureImpl.fail(FutureImpl.java:125)
	at io.vertx.core.impl.FutureImpl.handle(FutureImpl.java:178)
	at io.vertx.core.impl.FutureImpl.handle(FutureImpl.java:21)
	at io.vertx.kafka.client.consumer.impl.KafkaConsumerImpl.lambda$commit$5(KafkaConsumerImpl.java:366)
	at io.vertx.kafka.client.consumer.impl.KafkaReadStreamImpl.lambda$null$1(KafkaReadStreamImpl.java:123)
````

File: clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java [hunk 3]
```diff
@@ -256,43 +257,83 @@ private void handleBatch(final KafkaConsumerRecords<String, Buffer> records) {
         try {
             final T message = createMessage(record);
             messageHandler.handle(message);
+            addToCurrentOffsets(record);
             return Future.succeededFuture();
```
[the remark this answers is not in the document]
b-abel: Done.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
b-abel: I don't see a problem here. Error handling of the `poll` operation should not be reached. And it is intended that that close handler is not invoked if the message handler failed (the code that uses this class sets both handlers, so it "knows" that it threw an exception).

b-abel:
I took the chance to add some small changes that improve the code FMPOV even if this extends the initial scope of this PR. @sophokles73 I'm sorry for any inconvenience caused by this.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
calohmn: > so it "knows" that it threw an exception

If it's an unexpected exception and there's no try/catch inside the message handler, the message handler might not know. I think it would be more consistent, to also call the close handler then. And the "onClose" logic would only need to be in one place.

In any case, the class javadoc currently says, that the close handler is called:
````
 * If the message processing fails because either {@link #createMessage(KafkaConsumerRecord)} or the message handler
 * throws an exception, the Kafka consumer will be closed and the exception will be passed to the close handler.
````

Comment (b-abel) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: > If it's an unexpected exception and there's no try/catch inside the message handler, the message handler might not know. I think it would be more consistent, to also call the close handler then. And the "onClose" logic would only need to be in one place.

These are valid points. On the other hand: how would the caller handle the unexpected exception? The generic failure handling could be to start a new consumer. That would lead to the same exception, causing a failure loop...

Initially, it was like that. @sophokles73 seem to have expected the opposite: https://github.com/eclipse/hono/pull/2432#discussion_r565301540
Therefore I changed it (and obviously missed to adapt this point in that place of the JavaDoc). 
FMPOV both variants are valid. Of course, the documentation must match the behavior.

Comment (calohmn) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: > On the other hand: how would the caller handle the unexpected exception? The generic failure handling could be to start a new consumer. 

You mean, if there's a try/catch inside the message handler, what to do in the "catch" block? I don't see a reason for starting a new consumer there. When an exception happens there, the issue is with the record or some other consumer independent state, not the consumer itself, isn't it? And besides, actually starting a new consumer there while the current consumer is still alive would be error-prone.
So, this means that I think it's not very practical to close the consumer when there is an exception in `createMessage` or `messageHandler.handle(message);`. I would rather invoke a `onMessageHandlingError(record)` method which by default just logs the error. That method can then be overridden by actual Consumer implementations if needed.

Comment (calohmn) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: Since using the `AbstractAtLeastOnceKafkaConsumer` means having to implement the closeHandler anyway (to handle commit errors, etc.), I would also be ok with closing the consumer on messageHandler errors (which should be seen as programming errors). But then I think indeed the closeHandler should be called there.
And I think it would be good to document that the messageHandler shouldn't explicitly throw an exception (because otherwise the consumer gets closed).

Comment (b-abel) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: @calohmn I agree, it would be better to invoke the close handler on every kind of error. Otherwise, at least for programming errors, the consumer would be closed silently. And it would improve the testability of the class if I could see from the outside when the consumer closes.

Comment (b-abel) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: @sophokles73 Invoking the close handler on errors during message consumption would also affect the behavior of the Kafka-based downstream consumers returned by the application client factory. I would change the documentation in https://github.com/eclipse/hono/blob/82a600c770b68e719c429ec1c4cc8f8f49a4a09e/clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/KafkaApplicationClientFactory.java#L46-L47 accordingly. Are you OK with that?

Comment (sophokles73) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: >  Invoking the close handler on errors during message consumption would also affect the behavior of the Kafka-based downstream consumers returned by the application client factory.

Are we talking about exceptions being thrown accidentally or exceptions thrown deliberately by the consumer to indicate that it cannot/will not process the message? I guess the latter behavior should be ruled out explicitly (i.e. be documented properly).

Comment (b-abel) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: Yes, I agree.

Comment (sophokles73) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: So then this is indeed something that we want to prevent the message handler from doing. That is in contrast to the AMQP based consumer where we explicitly allow the user code to throw an (unchecked) exception to indicate that the message should be settled with the rejected/released outcome. Allowing it there but preventing it here will probably make it much harder to reach the goal of implementing a consumer once and use it for both AMQP and Kafka ...

Comment (b-abel) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: Yes, I don't really see a way around that. AMQP allows rejecting a single message. Kafka's streaming semantics do not provide this. You can ignore a message (which could break the AT LEAST ONCE guarantee) or you can stop the stream and wait for help :-) So, if one wants to use Kafka-based messaging one should accept all messages and do not rely on a "backchannel" to the sender. 

I would say that the message handler *should* not throw an exception. Because this would stop the consumption. If users of the consumer decide to throw an exception in the message handler nevertheless, they get their consumer closed and need to handle them in the close handler. Do you have a better idea?

Comment (sophokles73) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: Maybe we can use the MessageContext to provide an abstraction for the consumer in this case as well. I can imagine to add *MessageContext.accept()*, *reject()* and *release()* methods which for AMQP have the expected consequences regarding the disposition update and for Kafka may or may not have an impact on which index is being committed or the Kafka consumer being closed.

Comment (b-abel) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: The question remains what the Kafka consumer should do to reject or release a message. 

FMPOV the goal is not that consumers for AMQP and Kafka behave exactly the same but to have an API that makes it easy to switch from AMQP to Kafka. It should also allow it to implement a solution that supports both variants (which might involve type checking and casting when things should be handled differently).
The alternative is to aim for the lowest common denominator. In this case, this would mean to refrain from transmitting the status to the sender.

**Comment by b-abel** [comment]
@ctron @dejanb I accidentally pushed a branch where the rebase went wrong. This caused a lot of files to be marked as changed and automatically requested reviews from you. I fixed it. I am sorry for the inconvenience.

File: clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java (hunk 1 above)
Comment (b-abel) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: I pushed the change that invokes the close handler if the message hanlder throws an exception. And I changed the JavaDoc in the northbound application client accordingly. @sophokles73 Does this make sense from your POV?

Comment (sophokles73) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: I am still missing support for a very simple scenario that I believe to be quite common:
When the user provided consumer receives a message for processing it might want to retrieve some data from a database or a remote service first. Let's say that step fails with a runtime exception. My understanding is that the consumer needs to implement retry behavior itself in this case, right? In the AMQP case the runtime exception would trigger re-delivery of the (event) message and thus (implicitly) re-run the consumer with the message (AT LEAST ONCE). In the Kafka case, however, the client will close the Kafka Consumer and notify the close handler with the runtime exception. Alternatively, the consumer can catch the RE itself and re-run its handling logic. When the client crashes before it has committed an already processed message, however, it will get the message again for processing after a restart. Is this how it is supposed to work or am I missing something?

Comment (b-abel) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: > Alternatively, the consumer can catch the RE itself and re-run its handling logic. 

That would probably be the preferred way. But if this takes very long, it could also close the consumer (either by invoking the `stop()` method or by throwing an exception) and have the message re-delivered when it starts a new consumer.

> Is this how it is supposed to work 

Yes. Is there anything that should be added?

Comment (b-abel) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: > I am still missing support for a very simple scenario that I believe to be quite common:

I'm afraid I don't understand what you're missing. Do you think we should implement an automatic restart in the consumer?

Comment (sophokles73) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: I was referring to the automatic replay of messages as in the AMQP case. It still strikes me as odd that the programming model for implementing the consumer is so different between AMQP and Kafka. We could let the Kafka based DownstreamMessageConsumer re-deliver the message automatically to the client provided consumer, if it throws a runtime exception. Then it would be the same behavior as in the AMQP case, wouldn't it?

Comment (b-abel) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: I'll think about it. I guess it wouldn't be necessary to poll the record again from Kafka but instead invoke the message handler with the already created `DownstreamMessage` object. The result will be that if the error is not transient it will infinitely call the message handler and create and throw an exception (and no other event respectively telemetry message will be processed for this tenant). This will happen for NPEs in the message handler, other programming errors, or malformed messages. 
Wouldn't it be better, if we want to go this route, that based on the exception thrown, we can distinguish between expected, potentially transient, errors and unexpected permanent ones?

Comment (sophokles73) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: Sounds like a good idea. In any case, I would like to align the behavior of the Kafka and AMQP based clients wrt this as much as (reasonably) possible in order to minimize differences in the programming model.

Comment (b-abel) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: We could then define an exception type that the handler is expected to throw and which indicates that it wants a "re-try". All other exceptions are then unexpected and lead to closing the consumer. WDYT?

Comment (sophokles73) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: In the ServiceErrorException hierarchy we already distinguish between client induced and server side induced errors. In the AMQP based client a `ClientErrorException` leads to a `rejected` outcome whereas a `ServerErrorException` leads to a `released` outcome. IMHO we could therefore do a re-tried delivery for `ServerErrorException`s thrown by the consumer as they usually indicate a transient problem (at least in Hono). All other exceptions could then cause the consumer to be closed and the close handler being invoked. Would that work?

Comment (b-abel) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: @sophokles73 Please have a look at my last commit. I added this there.

Comment (b-abel) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: @sophokles73 There is still a difference between `ProtonBasedApplicationClientFactory` and `KafkaApplicationClientFactory` regarding error handling. Do you want to adjust the behavior of the AMQP based client factory as well?

Comment (sophokles73) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: IMHO we should never let the client close the Kafka consumer in case of runtime exceptions thrown when processing messages. Can you please change the code and update the JavaDocs accordingly @b-abel ?

Comment (b-abel) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: Sure. We are talking about the message _processing_ step (i.e. the message handler invocation), right? And not about errors in the communication with the Kafka cluster (i.e. _poll_ and _commit_)?

Comment (sophokles73) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: yes

Comment (b-abel) [reply clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java]: Done. Please see my last commit.

File: clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java [hunk 4]
```diff
@@ -209,90 +218,140 @@ private AbstractAtLeastOnceKafkaConsumer(final KafkaConsumer<String, Buffer> kaf
...
+
+                final KafkaConsumerRecord<String, Buffer> record = records.recordAt(i);
+                final T message = createMessage(record);
+                try {
+                    messageHandler.handle(message);
+                    addToCurrentOffsets(record);
+                } catch (final RuntimeException messageHandlingError) {
+                    LOG.debug("Message handler failed", messageHandlingError);
+                    // will commit the offset of the failed record and then resume polling (will include failed record)
+                }
```
sophokles73, line 250: why are we processing the remaining records here? we want to commit the offsets up to the first record that has caused the exception, don't we?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
```diff
@@ -170,6 +171,10 @@ private AbstractAtLeastOnceKafkaConsumer(final KafkaConsumer<String, Buffer> kaf
     /**
      * Creates a message from the given Kafka consumer record. The message will be passed to the message handler
      * afterward.
+     * <p>
+     * If possible, implementations should not throw exceptions. Any exception thrown here, will stop message
+     * consumption permanently, because the consumer will not skip the message and new consumer instances will start
+     * consuming from the failed message again.
```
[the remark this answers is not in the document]
b-abel: With the current implementation, the consumer is stopped if the `createMessage()` method throws an exception (in contrast to exceptions in the message handling where the consumer stays open). This enables rolling updates in cases where a new consumer instance will be able to parse newer versions of the record.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractAtLeastOnceKafkaConsumer.java
b-abel: Uh, that's bad. You are totally right. I missed adding a `break` statement :worried:

PR #2445, opened 2021-01-29
Fix firefly lora provider encoding

BobClaerhout:

### DIFF
--- a/adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/providers/FireflyProvider.java
+++ b/adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/providers/FireflyProvider.java
@@ -13,6 +13,7 @@
 
 package org.eclipse.hono.adapter.lora.providers;
 
+import java.nio.charset.StandardCharsets;
 import java.util.Objects;
 import java.util.Set;
 
@@ -20,8 +21,6 @@ import org.eclipse.hono.adapter.lora.GatewayInfo;
 import org.eclipse.hono.adapter.lora.LoraMessageType;
 import org.eclipse.hono.adapter.lora.LoraMetaData;
 import org.springframework.stereotype.Component;
-
-import com.google.common.io.BaseEncoding;
 
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.json.JsonArray;
@@ -83,7 +82,7 @@ public class FireflyProvider extends JsonBasedLoraProvider {
 
         Objects.requireNonNull(loraMessage);
         return LoraUtils.getChildObject(loraMessage, FIELD_FIREFLY_PAYLOAD, String.class)
-                .map(s -> Buffer.buffer(BaseEncoding.base16().decode(s.toUpperCase())))
+                .map(s -> Buffer.buffer(s.getBytes(StandardCharsets.UTF_8)))
                 .orElseThrow(() -> new LoraProviderMalformedPayloadException("message does not contain String valued payload property"));
     }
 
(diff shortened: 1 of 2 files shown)


adapters/lora-vertx/src/main/java/org/eclipse/hono/adapter/lora/providers/FireflyProvider.java
```diff
@@ -83,7 +82,7 @@ protected Buffer getPayload(final JsonObject loraMessage) {
 
         Objects.requireNonNull(loraMessage);
         return LoraUtils.getChildObject(loraMessage, FIELD_FIREFLY_PAYLOAD, String.class)
-                .map(s -> Buffer.buffer(BaseEncoding.base16().decode(s.toUpperCase())))
+                .map(s -> Buffer.buffer(s.getBytes(StandardCharsets.UTF_8)))
```
sophokles73, line 85: I find it hard to believe that the payload will always be UTF8 encoded text. Maybe the (Base16 encoded?) payload should be extracted from field *raw_payload* instead?

BobClaerhout: I captured some real data from our firefly installation and following contents represent the same message:
"raw_payload": "809C905F29809B02027B8F490D84532E6AC78FA58AC6771BC89BA1D03A",
"payload": "2043000CED010100310000000000FFFF",

I couldn't find the link between those two so I went with the UTF8 encoding.

sophokles73: The *payload* property's value clearly represents Base16 encoded bytes to me, don't you think?

BobClaerhout: I agree, the reason why we have a problem with this provider in particular is that payloads from other providers are parsed properly (in our business app), while this isn't. So there should be a difference

sophokles73: I am not sure if I can follow. The Lora adapter always puts opaque binary data in the payload of the message being sent downstream. To me this looks like what we are doing here already (Base16 decoding the property value). IMHO we need a way to verify that the content of the *payload* property really represents a Base16 encoded byte vector. Maybe you can ask on a Firefly support forum if it is not properly documented?

BobClaerhout: i'm currently in the process of writing some more tests for our business application to verify the statement above with 2 different providers.

BobClaerhout: Closing the PR. Looks like the error is situated somewhere else. Sorry for jumping into conclusions to fast :) Thanks for your support!

sophokles73: I remember faintly that we had a discussion about Firefly before in which we realized that the concrete format used for the uplink messages sent by Firefly depends on the *template* being configured for the *queue* over which the data is being sent to the protocol adapter. Sadly, though, the documentation does not provide any hint regarding a default format that is being used if no specific *template* is being used, as far as I can see :-(

BobClaerhout: yes indeed. I remember this as well. documentation is not that good. That's why I am implementing automated system tests to verify every time we do an update. Otherwise we're in doubt each time

Issue #2028, opened 2020-06-11
Introduce AdapterInstancesLivenessService, use for getCmdHandlingAdapterInstances


calohmn:
This came out of the discussions for #1858:

In order to detect and remove stale `commandHandlingAdapterInstance` entries (used in the Device Connection service), a service shall get introduced that keeps track of the currently alive protocol adapter instances. In the implementation of the Device Connection API `getCommandHandlingAdapterInstances` method, such a `AdapterInstancesLivenessService` is to be used in order to skip dead protocol adapter instance entries in the result and to also delete such stale entries.

The default implementation of the `AdapterInstancesLivenessService` shall support protocol adapters deployed in the same Kubernetes cluster and shall use the Kubernetes API to query the corresponding Pods/Containers.

calohmn:
Supporting not only protocol adapters in the same Kubernetes cluster but also (custom) protocol adapters outside of it, requires an extra, new API method for these adapters to announce that they are still alive. Instead of integrating this in the Device Connection API, I would rather do this in the new `CommandRouting` API (see #2029). 
Integrating the new `AdapterInstancesLivenessService` into the `CommandRouting` component also seems more straightforward compared to having to integrate the new service also in the protocol adapters (if the `HotrodBasedDeviceConnectionClient` and no separate Device Connection API component is used).
Therefore I would go for Hono 1.4 here.

calohmn:
FMPOV, there are 2 approaches here, the first requiring the adapters to be part of the same K8s cluster, the second not having that requrirement:

## Option 1: Liveness Check Using a Kubernetes API client in the Command Router

The Command Router places a watch on the list of Kubernetes pods so that it keeps a list of currently alive protocol adapter pods.
A given `commandHandlingAdapterInstance` entry is checked against that list to determine whether the instance is alive.
NOTE: The `commandHandlingAdapterInstance` ID is referring to one verticle instance inside the container inside the protocol adapter pod. That means:
- there are usually multiple `commandHandlingAdapterInstance` IDs per pod (corresponding to the number of verticles);
- using a `commandHandlingAdapterInstance` ID naming pattern like `[podName]_[verticleNo]` (e.g. `hono-adapter-amqp-vertx-cbcf8f967-9wfrw_1`) wouldn't unique identify the verticle instance over time, since the container can be restarted inside a given pod. 
In that case stale device-to-commandHandlingAdapterInstance mappings after a container was restarted couldn't be identified as such.

[...]

PR #2447, opened 2021-01-29
[#2273] Add KafkaBasedInternalCommandConsumer

calohmn:
This is for #2273.
Adds a Kafka consumer used in protocol adapters for receiving commands forwarded by the Command Router component.

### DIFF
--- a/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommand.java
+++ b/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommand.java
@@ -92,7 +92,6 @@ public final class KafkaBasedCommand implements Command {
      * <li>a non-empty <em>correlation-id</em> header if the <em>response-required</em> header is set to {@code true}.</li>
      * </ul>
      * or otherwise the returned command's {@link #isValid()} method will return {@code false}.
-     * <p>
      *
      * @param record The record containing the command.
      * @return The command.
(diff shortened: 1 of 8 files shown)


clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedInternalCommandConsumer.java
```diff
@@ -0,0 +1,237 @@
...
+    private String getTopicName() {
+        return new HonoTopic(HonoTopic.Type.COMMAND_INTERNAL, adapterInstanceId).toString();
+    }
+
+    @Override
+    public Future<Void> stop() {
+        final String topicName = getTopicName();
+        final Promise<Void> adminClientClosePromise = Promise.promise();
+        LOG.debug("stop: delete topic [{}]", topicName);
+        adminClient.deleteTopics(List.of(topicName), ar -> {
```
sophokles73, line 186: what if the process simply  crashes and thus the topics are not deleted during shut down of the component? Is there any other safeguard against *topic inflation* on the Kafka cluster?

calohmn: Cleaning up obsolete topics will be the task of the Command Router component. Should be prepared/implemented as part of https://github.com/eclipse/hono/issues/2028.

sophokles73: I see

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedInternalCommandConsumer.java
```diff
@@ -0,0 +1,237 @@
...
+        this.commandHandlers = Objects.requireNonNull(commandHandlers);
+        this.tracer = Objects.requireNonNull(tracer);
+
+        final Map<String, String> adminClientConfig = adminClientConfigProperties.getAdminClientConfig("internal-cmd");
+        adminClient = KafkaAdminClient.create(vertx, adminClientConfig);
+
+        final Map<String, String> consumerConfig = consumerConfigProperties.getConsumerConfig("internal-cmd");
+        // disable auto commit; offsets aren't committed here because topic and consumer are tied to
+        // the particular protocol adapter instance, with topic and consumer becoming obsolete once the
+        // adapter instance is stopped
```
b-abel, line 97: This comment does not cover that consumption could be stopped on the server-side. What do you expect to happen on rebalancing? Rebalancing happens even if there is only one consumer in the group when the consumer fails to send a heartbeat (https://kafka.apache.org/documentation/#consumerconfigs_heartbeat.interval.ms).

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedInternalCommandConsumer.java
```diff
@@ -0,0 +1,237 @@
...
+        consumerConfig.put("enable.auto.commit", "false");
```
kaniyan, line 98: How about `ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG` instead of `"enable.auto.commit"`?

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedInternalCommandConsumer.java
```diff
@@ -0,0 +1,237 @@
...
+        consumerConfig.put("group.id", adapterInstanceId);
```
kaniyan, line 99: How about `ConsumerConfig.GROUP_ID_CONFIG` instead of `"group.id"`?

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedInternalCommandConsumer.java
calohmn: Good point!
With the current implementation, I would actually rely on the "auto.offset.reset=latest" behaviour, getting the latest offset after a rebalance has happened. This could potentially cause some records to be skipped.
A way to fix this would be to manually keep track of the offset and use it explicitly after partitionsRevoked/partitionsAssigned events have happened. 

Or of course the standard commit handling could be used. For now I've changed the code to do just that and added a TODO to revisit this later.

b-abel: Yes, the options I see are also that you can either keep track internally of the offsets or you could simply use the standard mechanism with committing the offsets. Depending on the delivery semantics that you aim for you could use the `AbstractAtLeastOnceKafkaConsumer` or you could easily create an _at most once_ variant based on that. It might even be an option for the medium term to check if it makes sense to unify both delivery semantics in one consumer implementation.

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommand.java
```diff
@@ -110,6 +109,45 @@ public static KafkaBasedCommand from(final KafkaConsumerRecord<String, Buffer> r
...
+     * <p>
+     * If the <em>via</em> header is set, its value will be set as target gateway of the created command.
+     *
+     * @param record The record containing the command.
+     * @return The command.
+     * @throws NullPointerException if record is {@code null}.
+     * @throws IllegalArgumentException if the record doesn't correctly reference a target device.
+     */
+    public static KafkaBasedCommand fromRoutedCommandRecord(final KafkaConsumerRecord<String, Buffer> record) {
+        final String tenantId = getHeader(record, MessageHelper.APP_PROPERTY_TENANT_ID);
```
kaniyan, line 137: Missing _null_ check for the `record` as described in the JavaDoc. If the `record` is null, then the `getHeader(...)` will throw a _null_ pointer exception.

calohmn: I've added the explicit null check.

kaniyan (approved):
LGTM

PR #2450, opened 2021-02-01
Update user guide regarding unsupported operations by the file and JDBC device registries

kaniyan:
Add a warning to the JDBC and file based registry user guides, indicating which operations defined by the API are not supported by these registries.


sophokles73 (approved):
LGTM

PR #2451, opened 2021-02-01
[#2392] Use protocol-agnostic client factory in CLI

b-abel:
This belongs to #2392.
The "Receiver" in the CLI now uses the new application client for consuming downstream telemetry and event messages.


cli/src/main/java/org/eclipse/hono/cli/app/Receiver.java
```diff
@@ -52,20 +54,30 @@
...
-    private BiConsumer<String, Message> messageHandler = (endpoint, msg) -> handleMessage(endpoint, msg);
+    private BiConsumer<String, DownstreamMessage<? extends MessageContext>> messageHandler = this::handleMessage;
+
+    // TODO remove the client factory from here and use the one from parent once
+    //  org.eclipse.hono.application.client.ApplicationClientFactory supports C&C
+    private ApplicationClientFactory<? extends MessageContext> clientFactory;
+
+    @Autowired
+    public final void setProtocolAgnosticApplicationClientFactory(
+            final ApplicationClientFactory<? extends MessageContext> factory) {
```
calohmn, line 70: How about naming it just `setApplicationClientFactory` now already? The actual given factory is not protocol-agnostic (it's either an `AmqpApplicationClientFactory` or a `KafkaApplicationClientFactory`).
And with `setApplicationClientFactory` no further renaming is needed when moving towards one clientFactory here.  (And for now, just typing `setApplicationClientFactory` fits both factories, no need to remember the name for the "new" factory setter method.)

b-abel:

calohmn (approved):
LGTM

PR #2452, opened 2021-02-02
Use response-required instead of response-expected command header

calohmn:
Rename the Kafka record header used for request/response commands. 

The new name better reflects the need for a response to be sent by the device because otherwise the sending application has to consider the execution of the command to have failed. The new name is also in line with the corresponding name in Eclipse Ditto.


sophokles73 (approved):
LGTM

PR #2455, opened 2021-02-03
Create Authentication service base module

sophokles73:
Added auth-base module as a first step towards introducing a Quarkus
based implementation of the Authentication service.
The module contains common functionality which is independent from the
framework being used.


services/auth-base/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+          <artifactId>guava</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.springframework.security</groupId>
+          <artifactId>spring-security-crypto</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+  </dependencies>
+</project>
```
dejanb, line 78: No newline

dejanb (approved):

PR #2458, opened 2021-02-08
Move HTTP adapter base classes to HTTP adapter module

sophokles73:
There are several classes in the service-base module which are only
relevant to HTTP based protocol adapters. These classes have therefore
been moved to the HTTP adapter base module.

This is also the first step towards splitting up the service-base module into a service-base and an adapter-base module. The latter is supposed to contain classes only which are useful for implementing protocol adapters while the former will be useful for implementing both (plain) services and protocol adapters. This way we will be able to more precisely define the service component's (transitive) dependencies, e.g. protocol adapters will have a dependency on the Infinispan based client while the Authentication server won't. In a further step, the service-base and adapter-base modules will then be refined to service-base, service-base-spring and service-base-quarkus on the one hand and adapter-base, adapter-base-spring and adapter-base-quarkus on the other hand. This will allow to keep Spring Boot dependencies separate from Quarkus dependencies when adding Quarkus based service implementations.

### DIFF
--- a/adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/impl/VertxBasedHttpProtocolAdapterTest.java
+++ b/adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/impl/VertxBasedHttpProtocolAdapterTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2021 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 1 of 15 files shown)


adapters/http-vertx-base/src/test/java/org/eclipse/hono/adapter/http/impl/VertxBasedHttpProtocolAdapterTest.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2021 Contributors to the Eclipse Foundation
```
kaniyan, line 2: I don't see any changes in this class. FMPOV the copyright year need not be updated to _2021_.

kaniyan (approved):
LGTM

PR #2459, opened 2021-02-08
[#8] Use KafkaMessageHelper in AbstractKafkaBasedMessageSender.

b-abel:
Use KafkaMessageHelper.createKafkaHeader() for the encoding of Kafka headers in AbstractKafkaBasedMessageSender. Encoding and decoding are error-prone operations that must produce stable results. Therefore I want to use a single class where both operations are implemented.


kaniyan (approved):
LGTM

PR #2461, opened 2021-02-08
Upgrade Hugo theme for documentation

sophokles73:
The Learn theme has been upgraded to latest minor version 2.5.0.
Also replaced deprecated Hugo function/property references.

### DIFF
--- a/site/documentation/layouts/partials/header.html
+++ b/site/documentation/layouts/partials/header.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    {{ .Hugo.Generator }}
+    {{ hugo.Generator }}
     {{ partial "meta.html" . }}
     {{ partial "favicon.html" . }}
     <title>{{ .Title | safeHTML }} :: {{ .Site.Title | safeHTML }}</title>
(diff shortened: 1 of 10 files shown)


sophokles73:
@b-abel would you mind taking a look at this?

site/documentation/layouts/partials/header.html
```diff
@@ -15,15 +15,19 @@
...
+    <link href="{{"css/atom-one-dark-reasonable.css" | relURL}}{{ if $assetBusting }}?{{ now.Unix }}{{ end }}" rel="stylesheet">
     <link href="{{"css/theme.css" | relURL}}{{ if $assetBusting }}?{{ now.Unix }}{{ end }}" rel="stylesheet">
     <link href="{{"css/hugo-theme.css" | relURL}}{{ if $assetBusting }}?{{ now.Unix }}{{ end }}" rel="stylesheet">
     {{with .Site.Params.themeVariant}}
-      <link href="{{(printf "css/theme-%s.css" .) | relURL}}{{ if $assetBusting }}?{{ now.Unix }}{{ end }}" rel="stylesheet">
+    <link href="{{(printf "css/theme-%s.css" .) | relURL}}{{ if $assetBusting }}?{{ now.Unix }}{{ end }}" rel="stylesheet">
     {{end}}
+    {{ range .Site.Params.custom_css -}}
+    <link href="{{(printf "%s" .) | relURL}}{{ if $assetBusting }}?{{ now.Unix }}{{ end }}" rel="stylesheet">
+    {{- end }}
```
b-abel, line 26: This looks like a useful new feature. We could leverage it by adding the following to the `[params]` section in `hono/site/documentation/config.toml` under :

```toml
  # Provide a list of custom css files to load relative from the `static/` folder in the site root.
  custom_css = ["css/hono.css", "css/theme-hono.css", "css/swagger-ui-3.23.11.css"]
```

b-abel (approved):

site/documentation/layouts/partials/header.html
sophokles73: To me it looks like `theme-hono.css` is already automatically loaded by the `header.html` partial and we only want to load the swagger stuff when we are on a page that actually contains a swagger UI. It is therefore being loaded by our `custom-header.html` in that case only. We could remove the loading of `hono.css` from `custom-header.html`, though ...

sophokles73: @b-abel I have added the custom_css Site param ...

b-abel: > To me it looks like `theme-hono.css` is already automatically loaded by the `header.html` partial and we only want to load the swagger stuff when we are on a page that actually contains a swagger UI. It is therefore being loaded by our `custom-header.html` in that case only. We could remove the loading of `hono.css` from `custom-header.html`, though ...

You are right. I had mainly the cache busting feature in mind and didn't look closely enough at the stylesheet loading aspect.

b-abel:
LGTM

PR #2463, opened 2021-02-09
[#2273] Improve createCommandConsumer() in Command Router

calohmn:
Ensure that once createCommandConsumer() in the Kafka-based CommandConsumerFactory succeeds, the relevant Kafka topic exists and is in the list of subscribed topics of the consumer.


sophokles73 (approved):
I have no clue about whether the Kafka topic/rebalancing etc code is correct. But the rest looks good to me :-)

PR #2467, opened 2021-02-10
Add adapter-base modules

sophokles73:
Added the adapter-base, adapter-base-spring and adapter-base-quarkus
modules which are supposed to contain classes used for building protocol
adapters based on Spring and/or Quarkus.

Initially, the base classes for implementing adapters and the
Spring/Quarkus specific configuration classes have been moved to the new
modules.


adapters/base-quarkus/pom.xml
```diff
@@ -48,6 +48,10 @@
   </dependencyManagement>
 
   <dependencies>
+    <dependency>
+      <groupId>org.eclipse.hono</groupId>
+      <artifactId>hono-adapter-base-quarkus</artifactId>
+    </dependency>
     <dependency>
       <groupId>org.eclipse.hono</groupId>
       <artifactId>hono-service-base-quarkus</artifactId>
```
calohmn, line 57: Wouldn't it be enough to have just `hono-adapter-base-quarkus` here?

Will the `hono-adapters-quarkus` module defined here ("Defines common dependencies for Quarkus-based protocol adapter implementations") then be become obsolete? It's a bit confusing to have both that module and the new `hono-adapter-base-quarkus` ("Base classes for implementing Hono protocol adapters using Quarkus.").
Same goes for the corresponding spring based modules.

adapters/base-spring/pom.xml
```diff
@@ -26,6 +26,10 @@
   <description>Defines common dependencies for Spring-based protocol adapter implementations</description>
 
   <dependencies>
+    <dependency>
+      <groupId>org.eclipse.hono</groupId>
+      <artifactId>hono-adapter-base-spring</artifactId>
+    </dependency>
     <dependency>
       <groupId>org.eclipse.hono</groupId>
       <artifactId>hono-service-base-spring</artifactId>
```
calohmn, line 35: Wouldn't it be enough to have just `hono-adapter-base-spring` here? (see also other comment)

sophokles73: In general, yes. However, the Spring based adapter modules have explicit dependencies on classes in hono-service-base-spring, which is what I want to make explicit here.

adapters/base-quarkus/pom.xml
sophokles73: > Wouldn't it be enough to have just hono-adapter-base-quarkus here?

see other comment

> Will the hono-adapters-quarkus module defined here ("Defines common dependencies for Quarkus-based protocol adapter implementations") then be become obsolete? It's a bit confusing to have both that module and the new hono-adapter-base-quarkus ("Base classes for implementing Hono protocol adapters using Quarkus.").
Same goes for the corresponding spring based modules.

I don't think I get your point here.

sophokles73:
@calohmn any additional comments?

adapters/base-quarkus/pom.xml
calohmn: I was thinking of a module hierarchy with e.g. "service-base"->"adapter-base"->"adapter-spring"->"adapter-mqtt-spring" (with parent relationships). But I guess this would need more changes and would probably be less flexible. Maybe could be something to consider when we don't have to support spring anymore and only use quarkus.

calohmn (approved):
LGTM

PR #2468, opened 2021-02-10
Upgrade Hugo theme used for homepage to latest 1.1.1 version

sophokles73:
I have tried to minimize the customizations we do in the partials in order to allow for updating to newer versions of the theme more easily.


sophokles73:
@b-abel would you mind taking a look?

site/homepage/layouts/partials/head.html
```diff
@@ -3,40 +3,37 @@
   <meta name="robots" content="all,follow">
   <meta name="googlebot" content="index,follow,snippet,archive">
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <title>{{ .Title | safeHTML }} :: {{ .Site.Title | safeHTML }}</title>
+  {{ $title_plain := .Title | markdownify | plainify }}
+  <title>{{ $title_plain }} :: {{ .Site.Title | safeHTML }}</title>
   <meta name="author" content="{{ .Site.Author.name }}" />
```
b-abel, line 8: This should be:
`<meta name="author" content="{{ .Param "author" }}" />`

site/homepage/layouts/partials/head.html
```diff
@@ -60,25 +57,56 @@
...
 
-  <!-- owl carousel css -->
-
+  <!-- owl carousel CSS -->
   <link href="{{ "css/owl.carousel.css" | relURL}}" rel="stylesheet">
   <link href="{{ "css/owl.theme.css" | relURL}}" rel="stylesheet">
 
-  <link rel="alternate" href="{{ .Site.BaseURL }}/index.xml" type="application/rss+xml" title="{{ .Site.Title }}">
+  <!-- RSS feed -->
+  <link rel="alternate" href="{{ "/index.xml" | absURL }}" type="application/rss+xml" title="{{ .Site.Title }}">
```
b-abel, line 65: This resolves to `https://www.eclipse.org/index.xml` but should be `https://www.eclipse.org/hono/index.xml`. I think that this should do the trick:
`<link rel="alternate" href="{{ .Site.BaseURL }}index.xml" type="application/rss+xml" title="{{ .Site.Title }}">`

site/homepage/layouts/partials/head.html
sophokles73: Yes, I didn't change it because it had been like that before. I will change that in the next PR that I have prepared for implementing the adivce for SEO that we got some time back from marketing ...

sophokles73:
@b-abel thanks for the feedback :+1: I have pushed corresponding changes

b-abel (approved):
LGTM

Issue #2469, opened 2021-02-10
Prepare Kafka integration tests


fkaltner:
In parallel to the ongoing Kafka migration tasks the Hono integration tests need to spin up a Kafka (and Zookeeper) container which is properly configured for Docker-internal access (by Protocol Adapters and Device Registries) and Docker-external access (by the integration tests themselves)

Furthermore the integration test helpers need to be adapted to use the currently developed _org.eclipse.hono.application.client.ApplicationClientFactory_ interface.

In addition to that we need to make up our minds about the test concept itself considering the fact that it is likely that Hono will support AMQP and Kafka in parallel based on a configuration property of a tenant. Strictly considered this means that we need to run every test once with AMQP and once with Kafka. This will obviously at least double the runtime of our tests.

fkaltner:
@sophokles73 may I work on this?

fkaltner:
@b-abel Here's a draft PR which can be used to play around https://github.com/eclipse/hono/pull/2473

Note that the integration tests will still fail as northbound consumers still connect to AMQP. I'll continue to work on that and update the PR as soon as I am done.

Regarding the device registries: will you update the configuration of those similar to _AbstractAdapterConfig_, so that Kafka can be enabled by configuration properties?

b-abel:
> @b-abel Here's a draft PR which can be used to play around #2473

Cool. I'll have a look.

> Regarding the device registries: will you update the configuration of those similar to _AbstractAdapterConfig_, so that Kafka can be enabled by configuration properties?

This is currently not on my priority list :smile:  Could you please create an issue for it?

fkaltner:
> This is currently not on my priority list 😄 Could you please create an issue for it?

Sure. Created #2475.

sophokles73:
@fkaltner can this be closed?

fkaltner:
Integration tests for request/response commands are still not there as the implementation is not complete yet. But we can also create a new issue for that as this issue has the 1.7.0. milestone.

sophokles73:
Yes, that is what I would propose as well. Let's close this issue as it has indeed laid out the groundwork necessary for implementing the integration tests. can you create a new issue for the request/response command tests and then close this issue?

PR #2473, opened 2021-02-12
#2469 Adding Kafka to integration test infrastructure 

fkaltner:
Kafka for ITs may be enabled by running `mvn` with:  `-Dhono.messaging.type=kafka`.

For example: `mvn verify -Prun-tests,jaeger -Ddocker.keepRunning -Dit.test=EventHttpIT -Dhono.messaging.type=kafka`

In order to access the Kafka cluster from your development machine get the exposed Kafka port by 
`docker ps -a | grep -a cp-` 

and connect to Kafka using that port by using `<docker-host-ip>:<docker-kafka-port>` as `bootstrap.servers`

For example:
`kafkacat -b <docker-host-ip>:<docker-kafka-port> -L` 

or 

`kafka-console-consumer --bootstrap-server <docker-host-ip>:<docker-kafka-port> --property print.timestamp=true --property print.key=true --topic hono.event.<tenant-id> --from-beginning`


b-abel:
If you want to consume the events without the need have Kafka scripts installed, or want to see the headers, you can do this alternatively with the Hono CLI like that:

```sh
java -jar hono-cli-*-exec.jar --spring.profiles.active=receiver,kafka --tenant.id=<tenant-id> --hono.kafka.consumerConfig.bootstrap.servers=<docker-host-ip>:<docker-kafka-port>  --hono.kafka.consumerConfig.auto.offset.reset=earliest
```

tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
```diff
@@ -481,6 +490,8 @@
      */
     public AmqpApplicationClientFactory amqpApplicationClient;
 
+    public KafkaApplicationClientFactory kafkaApplicationClientFactory;
```
sophokles73, line 493: is there also a way to retrieve a factory?

tests/src/test/resources/amqp/application.yml
```diff
@@ -102,3 +102,13 @@ management:
...
+
+---
+
+spring:
+  profiles: kafka
+hono:
+  kafka:
+    producerConfig:
+      bootstrap:
+        servers: ${hono.kafka.bootstrap.servers}
```
sophokles73, line 114: missing EOL

tests/src/test/resources/coap/application.yml
```diff
@@ -85,3 +85,13 @@ management:
...
+
+---
+
+spring:
+  profiles: kafka
+hono:
+  kafka:
+    producerConfig:
+      bootstrap:
+        servers: ${hono.kafka.bootstrap.servers}
```
sophokles73, line 97: missing EOL

tests/src/test/resources/mqtt/application.yml
```diff
@@ -102,3 +102,13 @@ management:
...
+
+---
+
+spring:
+  profiles: kafka
+hono:
+  kafka:
+    producerConfig:
+      bootstrap:
+        servers: ${hono.kafka.bootstrap.servers}
```
sophokles73, line 114: missing EOL

sophokles73 (commented):
My understanding is that only a subset of the integration tests can be run using the kafka profile for the time being, right?

tests/pom.xml
```diff
@@ -422,6 +453,63 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
...
+                    <portName>kafka.port</portName>
+                  </portNames>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+          <plugin>
+            <!--
+              The standard Maven plugin for writing properties cannot be used as it is only capable of writing
+              all project properties to file but not specific ones only.
```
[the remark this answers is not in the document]
fkaltner: The other components use the docker-maven-plugin for writing dynamically assigned ports to property files. But for Kafka we cannot rely on the docker-maven-plugin to pick a port because of how Kafka's advertised listener works (see comment on the build-helper-maven-plugin and its reserve-network-port goal)

tests/pom.xml
```diff
@@ -782,6 +870,68 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
                     </env>
                   </run>
                 </image>
+                <!-- ##### Zookeeper ##### -->
+                <image>
+                  <name>${zookeeper.image.name}</name>
+                  <alias>hono-zookeeper-test</alias>
+                  <run>
+                    <skip>${hono.zookeeper.disabled}</skip>
```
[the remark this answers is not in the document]
fkaltner: Yes, it is. We can use the same property for skip.

tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
fkaltner: I'll remove this. This can be added later on when test code is adapted to Kafka.

tests/pom.xml
```diff
@@ -42,16 +42,21 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
...
     <hono.auth.host>hono-service-auth.hono</hono.auth.host>
+    <hono.kafka.bootstrap.servers>kafka.hono:9092</hono.kafka.bootstrap.servers>
 
     <!-- duration before vert.x complains about blocking the event loop -->
     <max.event-loop.execute-time>PT5S</max.event-loop.execute-time>
 
     <!-- logging profile used by all Hono components: prod, dev or trace -->
     <logging.profile>prod</logging.profile>
+    <!-- messaging profile configurable for integration tests: kafka or empty (defaults to AMQP) -->
+    <messaging.profile></messaging.profile>
```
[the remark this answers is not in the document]
fkaltner: This translates to Spring profile names. Currently there is no profile for AMQP as it is the default. If we set it to amqp then we will set a Spring profile name which does not exist. This will still work, but is confusing in my opinion. So I would rather leave it as empty.

tests/pom.xml
```diff
@@ -42,16 +42,21 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
...
     <!-- maximum time (milliseconds) to wait for successful startup of a single service component -->
     <service.startup.timeout>120000</service.startup.timeout>
     <!-- color used by components of the AMQP Messaging Network for logging -->
     <log.color.amqp-network>magenta</log.color.amqp-network>
+    <!-- color used by components of the AMQP Messaging Network for logging -->
```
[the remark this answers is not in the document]
fkaltner: Done.

tests/src/test/resources/amqp/application.yml
fkaltner: Added EOLs.

fkaltner:
> My understanding is that only a subset of the integration tests can be run using the kafka profile for the time being, right?

The adaption of the tests is not done in this PR in order to have smaller PRs. At the moment the tests would fail if they are ran with the Kafka profile. But one can connect to the local Kafka and observe telemetry messages being written to the corresponding topics. This PR just focuses on the "infrastructure" (but was validated with the branch mentioned below).

The adaption of telemetry & events has been started (see: https://github.com/bosch-io/hono/tree/%232469_kafka_integration_tests_wip). 
As a second step we could merge this (after polishing 😉 ), then some of the tests will succeed.

As soon as command and control is included in _ApplicationClientFactory_ (to my knowledge @kaniyan is currently working on this) we could add the necessary adaptions for making all tests successful when using Kafka as messaging network.

tests/pom.xml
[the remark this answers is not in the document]
sophokles73: I see

tests/readme.md
```diff
@@ -142,3 +142,15 @@ Router service component instead, the `command-router` maven profile can be set:
...
+
+By default, the integration tests are run using an AMQP 1.0 based messaging network. In order to use Kafka instead, 
+the `hono.messaging.type` maven property can be set to `kafka`:
+
+```sh
+mvn verify -Dhono.messaging.type=kafka
+```
+
+If the docker containers are kept running, a connection to Kafka for debugging purposes can be made by using the 
+machine's `DOCKER_HOST` as IP address and the dynamically exposed port of the Kafka container.
```
sophokles73, line 156: IMHO we should remove this for now and add it once we can indeed run the tests with Kafka. WDYT?

fkaltner: Fine by me.

sophokles73 (approved):
LGTM

Issue #2472, opened 2021-02-12
[1.5] Protocol adapters reuse obsolete downstream senders after disconnect


calohmn:
After a disconnect with the AMQP messaging network has occurred, protocol adapters don't clear the cache containing the downstream sender objects.
This leads to telemetry/event messages not getting sent.

Bug only exists in 1.5, not in 1.6 and master.

fkaltner:
Regarding the labels you added: aren't the device registries affected as well since they also may send events when they use gateway-based auto-provisioning?

calohmn:
Fixed via 8126063e795d06c0de839c09d344653db274cb22.

PR #2474, opened 2021-02-12
Also check connection when checking for open link

calohmn:
This serves as an extra safeguard against using a link whose underlying connection was already disconnected.
Relates to #2472.


sophokles73 (approved):
LGTM

Issue #2064, opened 2020-07-03
MQTT Adapter "Close behavior" in case of hono internal errors


Christian-Schmid:
The MQTT adapter closes the MQTT connection to the device in case of any errors during publishing telemetry / events / answering commands. This includes also more internal errors, e.g. timeouts on getting the assertion. 
 
As the initial connection establishment with a client is somehow resources intense (SSL Handshake, resource allocation, initial get, assert of device and tenant). Does it really make sense to close the mqtt connection in such error cases? I saw the discussion in [1262](https://github.com/eclipse/hono/issues/1262) and the following [PR](https://github.com/eclipse/hono/pull/1263) to close the connection. 

I'm aware that the specifications says with "[MQTT-4.8.0-2]", that the server should close the connection in case of transient errors, but in scenarios where you have a very large number of devices connected, a short issue e.g. in the registration service could have some cascade effect resulting in the disconnect of a large number of devices. 

In the issue [1262](https://github.com/eclipse/hono/issues/1262) there was a discussion ongoing regarding if this could be configurable per tenant. I think such option would be very nice. 

If the clients of the tenant are implemented in a way where they rely on the PubAck and not only on the client disconnect we still would fulfill the QoS guarantees towards the client:

- **Telemetry with QoS 0**: It could leave the connection open and just drop the telemetry message, which would be in line with the guarantees towards the client.

- **Telemetry with QoS 1**: It could leave the connection open and don't pub ack the message. Client would / could try to resend the message.

- **Event**: Same as Telemetry QoS 1 

For cases in which the cause of the failure is a well-known final cause and the cause is successfully received by the adapter (e.g. device is disabled in registry), the adapter should still close the connection. But for other cases such as timeouts, status codes 500) I would like to have the option to not close the connection.

Thanks
Chris

kaniyan:
@sophokles73 Could you please assign this to me, if no one is already working on it.

sophokles73:
I agree with the handling of unprocessable Telemetry QoS 0 messages being in line with the spec and the guarantees made by Hono.

> Telemetry with QoS 1: It could leave the connection open and don't pub ack the message. Client would / could try to resend the message.

Most people I have talked to believe that this behavior is actually *not* compliant with the spec, i.e. the server only has the options to send a PUBACK or close the connection. We have decided for closing the connection as most MQTT client implementations are not prepared to *not* get a PUBACK in response to a PUBLISH. However, they can always handle the server close the connection.

[...]

PR #2479, opened 2021-02-15
Improve MQTT subscription handling

calohmn:
As a preparation for #2064, the command subscription handling in the MQTT adapter has been refactored. 

Invalid subscriptions will now log an error entry in the tracing span, tracing when unsubscribing has been improved and proper handling of duplicate subscriptions (with same device id but different QoS or endpoint id) has been implemented. 
Also, an MQTT connection getting closed before the SUBACK can be sent will now prevent an event notification to be sent downstream.


sophokles73 (approved):
LGTM

calohmn:
(changed commit message slightly)

Issue #2402, opened 2021-01-08
Make messaging network in use configurable on tenant level


mhemmeter:
Eclipse Hono 1.6.0 introduces support for Apache Kafka as the messaging network for downstream communication (device-to-cloud communication). In a managed service offering customers should be able to decide which messaging network they want to use (Apache Kafka or AMQP 1.0). Therefore it would be great to make the messaging network configurable on a tenant level. 

I don't see the need to distinguish between Telemetry, Events and C&C. An either or configuration (Apache Kafka or AMQQP 1.0) for the tenant is sufficient.

sophokles73:
@b-abel is this something for you?

b-abel:
> @b-abel is this something for you?

Yes. Please assign to me.

b-abel:
The first idea would be:
1. adding a flag to the tenant configuration that indicates if Kafka should be used for that tenant
2. the protocol adapters start the support for AMQP- _and_ Kafka-based messaging on startup if configurations for both networks are present
3. when a protocol adapter receives a downstream message, it decides  which messaging to use based on the flag (1.)
4. once we have support for C&C with Kafka, it should be not too hard to find a similar solution there

mhemmeter:
Sounds very good to me!

sophokles73:
@b-abel @fkaltner @calohmn Anything else that needs to be done here?

b-abel:
FMPOV this can be closed.

calohmn:
I also think this is done.

fkaltner:
Looks good to me!

PR #2481, opened 2021-02-15
[#2402] Add interfaces to downstream senders.

b-abel:
This is a preparation step for #2402.
To allow the selection of the messaging network type on the tenant level, the Kafka-based senders, as well as the AMQP-based ones, need to be set on the protocol adapters. This PR adds interfaces that allow on the one hand to distinguish between AMQP- and Kafka-based senders and prevent on the other hand that the protocol adapters depend on the implementation classes.

### DIFF
--- a/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/telemetry/amqp/AmqpEventSender.java
+++ b/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/telemetry/amqp/AmqpEventSender.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.hono.adapter.client.telemetry.amqp;
+
+import org.eclipse.hono.adapter.client.telemetry.EventSender;
+
+/**
+ * A client for publishing events originating from devices to downstream AMQP consumers.
+ */
+public interface AmqpEventSender extends EventSender {
+}
--- a/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/telemetry/amqp/AmqpTelemetrySender.java
+++ b/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/telemetry/amqp/AmqpTelemetrySender.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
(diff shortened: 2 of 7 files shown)


b-abel:
Uh, the error does not seem related to my change. It builds on my machine ;-)

sophokles73:
I do not understand (yet) why we need these additional interfaces. Can you elaborate? A rebase should fix the CI build ...

b-abel:
> I do not understand (yet) why we need these additional interfaces. Can you elaborate? 

Currently, the only one `EventSender` is set on a protocol adapter. With #2402 we need to set one for AMQP _and_ one for Kafka (same goes for telemetry, but let me stick to event for the example). We need two different setters. Currently, we could either set an `EventSender` on both setters, then both setters differ only in the names, but you could easily set an AMQP-based sender where one for Kafka is expected. Or alternatively, the setters could expect the implementation types but I prefer if the protocol adapters do not depend on the concrete implementation. 

The new interfaces allow to replace `AbstractProtocolAdapterBase#setEventSender(EventSender)` by `AbstractProtocolAdapterBase#setEventSender(AmqpEventSender)` and `AbstractProtocolAdapterBase#setEventSender(KafkaEventSender)`. Then we have both senders available in the adapter. 

The next step would be FMPOV to change the signature of `AbstractProtocolAdapterBase#getEventSender()` to `AbstractProtocolAdapterBase#getEventSender(TenantObject)` and resolve the event sender to use according to the tenant configuration. This method will return a `EventSender` so that the subclasses of `AbstractProtocolAdapterBase` would only use the event `EventSender` interaface and know nothing about different senders for AMQP and Kafka.
I this comprehensible?

sophokles73:
> The new interfaces allow to replace AbstractProtocolAdapterBase#setEventSender(EventSender) by AbstractProtocolAdapterBase#setEventSender(AmqpEventSender) and AbstractProtocolAdapterBase#setEventSender(KafkaEventSender). Then we have both senders available in the adapter.

Is that really worth the effort of introducing four new interfaces? I do not really see a problem with simply replacing *setEventSender()* with *setAmqpEventSender()* and *setKafkaEventSender()*.

@calohmn @kaniyan WDYT?

b-abel:
> I do not really see a problem with simply replacing _setEventSender()_ with _setAmqpEventSender()_ and _setKafkaEventSender()_.

FMPOV this is not about the names of the setters. The names you suggest are fine. For the logic in the protocol adapters it is essential to distinguish between the AMQP-based and the Kafka-based adapters. This could be done vie the variable names or, with added type safety, with the interfaces.

sophokles73:
> This could be done vie the variable names or, with added type safety, with the interfaces.

I know, still the same question remains.

calohmn:
> We need two different setters. Currently, we could either set an `EventSender` on both setters, then both setters differ only in the names, but you could easily set an AMQP-based sender where one for Kafka is expected. Or alternatively, the setters could expect the implementation types but I prefer if the protocol adapters do not depend on the concrete implementation. 

I also think that adding the 4 extra interfaces just for the sake of these 2 setter methods looks not worth the effort, especially since they don't have methods defined in them. If we want an extra safety measure, we could also add `KafkaBased/ProtonBasedClient` marker interfaces instead and add `instanceof` checks for these inside the setter methods (throwing IllegalArgumentExceptions).

> The next step would be FMPOV to change the signature of `AbstractProtocolAdapterBase#getEventSender()` to `AbstractProtocolAdapterBase#getEventSender(TenantObject)` and resolve the event sender to use according to the tenant configuration. This method will return a `EventSender` so that the subclasses of `AbstractProtocolAdapterBase` would only use the event `EventSender` interaface and know nothing about different senders for AMQP and Kafka.

I guess this step wouldn't require the extra interfaces, would it?

b-abel:
> I guess this step wouldn't require the extra interfaces, would it?

Right.

b-abel:
I add #2489 that does not use the interfaces. So this can closed, right @sophokles73 @calohmn?

sophokles73:
@b-abel are you going to close this PR?

PR #2483, opened 2021-02-15
Refactor MqttEndpoint related methods

calohmn:
The abstract MQTT adapter base class has been refactored, moving all `MqttEndpoint` related methods into a new `MqttDeviceEndpoint` inner class. This means the parameters related to the MQTT device connection don't have to be passed around in these methods anymore. This also makes it easier to add further MQTT device connection related fields (e.g. for #2064).

As a next step, the `MqttDeviceEndpoint` class could be moved out of the abstract MQTT adapter base class. 
This would then however require some visibility changes in `AbstractProtocolAdapterBase`, making methods public instead of protected.


calohmn:
@kaniyan I've committed corresponding changes.

kaniyan (approved):
LGTM

PR #2488, opened 2021-02-19
[2273] Support sending command response via Kafka

kaniyan:
This PR contains a Kafka based command response sender which enables the protocol adapters to forward command responses via the configured Kafka cluster. Also protocol adapters can be configured now to use this Kafka based command router.


sophokles73 (approved):
LGTM

PR #2493, opened 2021-02-23
[#2064] Extract AbstractSubscription out of CommandSubscription

calohmn:
For #2064:
This is in preparation of adding support for different kinds of MQTT subscriptions, notably a subscription on an error topic.

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandSubscriptionsManager.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandSubscriptionsManager.java
@@ -70,18 +70,25 @@ public final class CommandSubscriptionsManager<T extends MqttProtocolAdapterProp
      * Invoked when a device sends an MQTT <em>PUBACK</em> packet.
      *
      * @param msgId The message/packet id of the command published with QoS 1.
+     * @return {@code true} if a pending command request for the PUBACK was found.
      * @throws NullPointerException if msgId is {@code null}.
      */
-    public void handlePubAck(final Integer msgId) {
+    public boolean handlePubAck(final Integer msgId) {
         Objects.requireNonNull(msgId);
-        LOG.trace("acknowledgement received for command sent to device [packet-id: {}]", msgId);
-        Optional.ofNullable(removeFromWaitingForAcknowledgement(msgId))
-                .ifPresentOrElse(pendingCommandRequest -> {
+        return Optional.ofNullable(removeFromWaitingForAcknowledgement(msgId))
+                .map(pendingCommandRequest -> {
+                    LOG.trace("acknowledgement received for command sent to device [packet-id: {}]", msgId);
                     if (pendingCommandRequest.timerId != null) {
                         cancelTimer(pendingCommandRequest.timerId);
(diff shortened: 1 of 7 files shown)


adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandSubscriptionsManager.java
```diff
@@ -70,18 +70,25 @@ public CommandSubscriptionsManager(final Vertx vertx, final T config) {
      * Invoked when a device sends an MQTT <em>PUBACK</em> packet.
      *
      * @param msgId The message/packet id of the command published with QoS 1.
+     * @return {@code true} if a pending command request for the PUBACK was found.
      * @throws NullPointerException if msgId is {@code null}.
      */
-    public void handlePubAck(final Integer msgId) {
+    public boolean handlePubAck(final Integer msgId) {
```
kaniyan, line 76: I see that the return value is not being used anywhere. Is it here in consideration with the upcoming PRs?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandSubscriptionsManager.java
```diff
@@ -70,18 +70,25 @@ public CommandSubscriptionsManager(final Vertx vertx, final T config) {
...
+                .map(pendingCommandRequest -> {
+                    LOG.trace("acknowledgement received for command sent to device [packet-id: {}]", msgId);
                     if (pendingCommandRequest.timerId != null) {
                         cancelTimer(pendingCommandRequest.timerId);
                     }
                     pendingCommandRequest.onAckHandler.handle(msgId);
-                }, () -> LOG.debug("no active command request found for received acknowledgement [packet-id: {}]", msgId));
+                    return true;
+                }).orElseGet(() -> {
+                    if (!subscriptions.isEmpty()) {
```
kaniyan, line 87: Is this check needed? It seems to me that this log message makes sense even when there are no subscriptions (i.e the _map_ is empty) .

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandSubscriptionsManager.java
calohmn: Yes, right. The PUBACK could have been sent for a published command *or* a published message on the new error topic for #2064. So, the PUBACK isn't necessarily only handled in the CommandSubscriptionsManager any more.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/CommandSubscriptionsManager.java
calohmn: This is an extra check in preparation for the upcoming error topic handling (see also other comment).
A PUBACK received while `subscriptions` is empty is most certainly not a PUBACK for a command message (but instead for a published error message for example).

kaniyan (approved):
LGTM

Issue #2478, opened 2021-02-12
Add Command&Control support to new northbound Client Interface


b-abel:
The new northbound client interface `org.eclipse.hono.application.client.ApplicationClientFactory` currently only supports downstream messaging and misses support for Command&Control.

sophokles73:
@b-abel will this go into 1.7.0?

b-abel:
@calohmn Is this something for you? 
The next issue I plan to work on is #2402. I'm not sure that I can finish both this month.

kaniyan:
@b-abel @calohmn I am working on this.

sophokles73:
@b-abel @kaniyan Can this be closed?

kaniyan:
> @b-abel @kaniyan Can this be closed?

@sophokles73  This can be closed now.

kaniyan:
This has been closed via PRs #2498, #2529, #2533 and #2583.

PR #2498, opened 2021-02-24
[#2478] Extend the northbound application client to support sending commands asynchronously

kaniyan:
This PR is for #2478. It extends the northbound application client to support sending commands asynchronously.  Support for sending one-way/synchronous commands and consuming command responses will be addressed in the next PRs. 

This PR has been implemented on top of code from PR #2497 and that should be merged first.

### DIFF
--- a/cli/src/main/java/org/eclipse/hono/cli/AppConfiguration.java
+++ b/cli/src/main/java/org/eclipse/hono/cli/AppConfiguration.java
@@ -151,7 +151,7 @@ public class AppConfiguration {
     public ApplicationClient<KafkaMessageContext> kafkaApplicationClient(
             final Vertx vertx,
             final KafkaConsumerConfigProperties kafkaConsumerConfigProperties,
-            final KafkaProducerFactory producerFactory,
+            final KafkaProducerFactory<String, Buffer> producerFactory,
             final KafkaProducerConfigProperties kafkaProducerConfigProperties) {
         return new KafkaApplicationClientImpl(vertx, kafkaConsumerConfigProperties, producerFactory,
                 kafkaProducerConfigProperties);
--- a/cli/src/main/java/org/eclipse/hono/cli/AppConfiguration.java
+++ b/cli/src/main/java/org/eclipse/hono/cli/AppConfiguration.java
@@ -144,7 +144,7 @@ public class AppConfiguration {
      * @param kafkaConsumerConfigProperties The consumer configuration properties.
      * @param producerFactory The factory to use for creating Kafka producers.
      * @param kafkaProducerConfigProperties The producer configuration properties.
-     * @return The factory.
+     * @return The client.
      */
     @Profile("kafka")
     @Bean
--- a/cli/src/main/java/org/eclipse/hono/cli/AppConfiguration.java
+++ b/cli/src/main/java/org/eclipse/hono/cli/AppConfiguration.java
@@ -20,6 +20,8 @@ import org.eclipse.hono.application.client.kafka.KafkaMessageContext;
 import org.eclipse.hono.application.client.kafka.impl.KafkaApplicationClientImpl;
 import org.eclipse.hono.client.ApplicationClientFactory;
 import org.eclipse.hono.client.HonoConnection;
+import org.eclipse.hono.client.kafka.KafkaProducerConfigProperties;
+import org.eclipse.hono.client.kafka.KafkaProducerFactory;
 import org.eclipse.hono.client.kafka.consumer.KafkaConsumerConfigProperties;
 import org.eclipse.hono.config.ClientConfigProperties;
 import org.springframework.boot.context.properties.ConfigurationProperties;
@@ -29,6 +31,7 @@ import org.springframework.context.annotation.Profile;
 
 import io.vertx.core.Vertx;
 import io.vertx.core.VertxOptions;
+import io.vertx.core.buffer.Buffer;
 import io.vertx.core.dns.AddressResolverOptions;
 
 /**
@@ -87,6 +90,29 @@ public class AppConfiguration {
     }
 
     /**
+     * Exposes Kafka producer configuration properties as a Spring bean.
+     *
+     * @return The properties.
+     */
+    @ConfigurationProperties(prefix = "hono.kafka")
+    @Profile("kafka")
+    @Bean
+    public KafkaProducerConfigProperties kafkaProducerConfig() {
+        return new KafkaProducerConfigProperties();
+    }
+
+    /**
+     * Exposes a factory for creating producers for sending messages via the Kafka cluster.
+     *
+     * @return The factory.
+     */
+    @Profile("kafka")
+    @Bean
+    public KafkaProducerFactory<String, Buffer> kafkaProducerFactory() {
+        return KafkaProducerFactory.sharedProducerFactory(vertx());
+    }
+
+    /**
      * Exposes a factory for creating clients for Hono's northbound APIs as a Spring bean.
      *
      * @return The factory.
@@ -116,13 +142,18 @@ public class AppConfiguration {
      *
      * @param vertx The vertx instance to be used.
      * @param kafkaConsumerConfigProperties The consumer configuration properties.
+     * @param producerFactory The factory to use for creating Kafka producers.
+     * @param kafkaProducerConfigProperties The producer configuration properties.
      * @return The factory.
      */
     @Profile("kafka")
     @Bean
     public ApplicationClient<KafkaMessageContext> kafkaApplicationClient(
             final Vertx vertx,
-            final KafkaConsumerConfigProperties kafkaConsumerConfigProperties) {
-        return new KafkaApplicationClientImpl(vertx, kafkaConsumerConfigProperties);
+            final KafkaConsumerConfigProperties kafkaConsumerConfigProperties,
+            final KafkaProducerFactory producerFactory,
+            final KafkaProducerConfigProperties kafkaProducerConfigProperties) {
+        return new KafkaApplicationClientImpl(vertx, kafkaConsumerConfigProperties, producerFactory,
+                kafkaProducerConfigProperties);
     }
 }
--- a/clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedCommandSender.java
+++ b/clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedCommandSender.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.hono.application.client.amqp;
+
+import java.util.Map;
+import java.util.Objects;
+
+import org.apache.qpid.proton.message.Message;
+import org.eclipse.hono.application.client.CommandSender;
+import org.eclipse.hono.client.HonoConnection;
+import org.eclipse.hono.client.SendMessageSampler;
+import org.eclipse.hono.client.StatusCodeMapper;
+import org.eclipse.hono.client.amqp.SenderCachingServiceClient;
+import org.eclipse.hono.util.AddressHelper;
+import org.eclipse.hono.util.CommandConstants;
+import org.eclipse.hono.util.MessageHelper;
+
+import io.opentracing.SpanContext;
+import io.vertx.core.Future;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.proton.ProtonHelper;
+
+/**
+ * A vertx-proton based client for sending commands.
+ *
+ * @see <a href="https://www.eclipse.org/hono/docs/api/command-and-control/">
+ *      Command &amp; Control API for AMQP 1.0 Specification</a>
+ */
+public class ProtonBasedCommandSender extends SenderCachingServiceClient implements CommandSender {
+
+    /**
+     * Creates a new vertx-proton based command sender.
+     *
+     * @param connection The connection to the Hono server.
+     * @param samplerFactory The factory for creating samplers for tracing AMQP messages being sent.
+     * @throws NullPointerException if any of the parameters except the closeHandler is {@code null}.
+     */
+    public ProtonBasedCommandSender(
+            final HonoConnection connection,
+            final SendMessageSampler.Factory samplerFactory) {
+        super(connection, samplerFactory);
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * @throws NullPointerException if tenantId, deviceId, command, correlationId or replyId is {@code null}.
+     */
+    @Override
+    public Future<Void> sendAsyncCommand(final String tenantId, final String deviceId, final String command,
+            final String contentType, final Buffer data, final String correlationId, final String replyId,
+            final Map<String, Object> properties, final SpanContext context) {
+        Objects.requireNonNull(tenantId);
+        Objects.requireNonNull(deviceId);
+        Objects.requireNonNull(command);
+        Objects.requireNonNull(correlationId);
+        Objects.requireNonNull(replyId);
+
+        return getOrCreateSenderLink(CommandConstants.NORTHBOUND_COMMAND_REQUEST_ENDPOINT, tenantId)
+                .recover(thr -> Future.failedFuture(StatusCodeMapper.toServerError(thr)))
+                .compose(sender -> {
+                    final String targetAddress = AddressHelper
+                            .getTargetAddress(CommandConstants.NORTHBOUND_COMMAND_REQUEST_ENDPOINT, tenantId, deviceId,
+                                    connection.getConfig());
+                    final Message msg = createMessage(tenantId, deviceId, command, contentType, data, correlationId,
+                            replyId, targetAddress, properties);
+                    return sender.sendAndWaitForOutcome(msg, newChildSpan(context, "send Command " + command));
+                })
+                .mapEmpty();
+    }
+
+    private static Message createMessage(final String tenantId, final String deviceId, final String command,
+            final String contentType, final Buffer data, final String correlationId, final String replyId,
+            final String targetAddress, final Map<String, Object> properties) {
+        final Message msg = ProtonHelper.message();
+
+        MessageHelper.setCreationTime(msg);
+        msg.setAddress(targetAddress);
+        msg.setReplyTo(
+                String.format("%s/%s/%s", CommandConstants.NORTHBOUND_COMMAND_RESPONSE_ENDPOINT, tenantId, replyId));
+        MessageHelper.setApplicationProperties(msg, properties);
+        msg.setCorrelationId(correlationId);
+        msg.setSubject(command);
+        MessageHelper.setPayload(msg, contentType, data);
+        MessageHelper.addTenantId(msg, tenantId);
+        MessageHelper.addDeviceId(msg, deviceId);
+
+        return msg;
+    }
+}
--- a/clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedCommandSender.java
+++ b/clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedCommandSender.java
@@ -74,7 +74,7 @@ public class ProtonBasedCommandSender extends SenderCachingServiceClient impleme
                                     connection.getConfig());
                     final Message msg = createMessage(tenantId, deviceId, command, contentType, data, correlationId,
                             replyId, targetAddress, properties);
-                    return sender.sendAndWaitForOutcome(msg, newChildSpan(context, "send command " + command));
+                    return sender.sendAndWaitForOutcome(msg, newChildSpan(context, "send command"));
                 })
                 .mapEmpty();
     }
--- a/clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedCommandSender.java
+++ b/clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedCommandSender.java
@@ -74,7 +74,7 @@ public class ProtonBasedCommandSender extends SenderCachingServiceClient impleme
                                     connection.getConfig());
                     final Message msg = createMessage(tenantId, deviceId, command, contentType, data, correlationId,
                             replyId, targetAddress, properties);
-                    return sender.sendAndWaitForOutcome(msg, newChildSpan(context, "send Command " + command));
+                    return sender.sendAndWaitForOutcome(msg, newChildSpan(context, "send command " + command));
                 })
                 .mapEmpty();
     }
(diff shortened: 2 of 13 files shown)


clients/application/src/main/java/org/eclipse/hono/application/client/AsyncCommandClient.java
```diff
@@ -0,0 +1,133 @@
...
+     * @param command The command name.
+     * @param data The command data to send to the device or {@code null} if the command has no input data.
+     * @param correlationId The identifier to use for correlating the response with the request. Note: This parameter is
+     *            security sensitive. To ensure secure request response mapping choose correlationId carefully, e.g.
+     *            {@link java.util.UUID#randomUUID()}.
+     * @param replyId This identifier is to be set depending upon the underlying implementation (i.e AMQP or Kafka based).
+     *                <ul>
+     *                <li>In Kafka based implementation, this is not used and hence it can be set as {@code null}.</li>
+     *                <li>In AMQP based implementation, this is an arbitrary string which gets used for the response 
+     *                link address in the form of <em>command_response/${tenantId}/${replyId}</em>. Must match the
```
sophokles73, line 49: Is there any support method for creating a consumer for that address anywhere?

clients/application/src/main/java/org/eclipse/hono/application/client/AsyncCommandClient.java
```diff
@@ -0,0 +1,133 @@
...
+     * @return A future indicating the result of the operation.
+     *         <p>
+     *         If the command was accepted, the future will succeed.
+     *         <p>
+     *         The future will fail with a {@link ServiceInvocationException} if the command could not
+     *         be forwarded upstream.
+     * @throws NullPointerException if deviceId, command or correlationId is {@code null}. 
+     *                              Also if the replyId is {@code null} provided the implementation is AMQP based.
+     */
+    Future<Void> sendAsyncCommand(String deviceId, String command, Buffer data, String correlationId, String replyId);
```
b-abel, line 61: FMPOV only the last method is necessary. I don't see enough benefit in sparing the `properties` and `context` parameters to justify the increased complexity that comes with overloading.

clients/application/src/main/java/org/eclipse/hono/application/client/ApplicationClientFactory.java
```diff
@@ -79,4 +79,16 @@
...
+     * get received by a separate receiver.
+     *
+     * @param tenantId The tenant of the devices to which commands shall be sent.
+     * @return A future that will complete with the command client (if successful) or
+     *         fail if the client cannot be created, e.g. because the underlying connection
+     *         is not established or if a concurrent request to create a client for the same
+     *         tenant is already being executed.
+     * @throws NullPointerException if tenantId is {@code null}.
+     */
+    Future<AsyncCommandClient> createAsyncCommandClient(String tenantId);
```
sophokles73, line 93: I do not quite understand why we need to create another object here. Can't we simply let the *ApplicationClientFactory* implement *AsyncCommandClient*? One of the goals of the new client was to no longer create tenant specific instances of clients but instead allow users to simply pass in a tenant ID to the commands directly instead. Or is there a particular reason that would require use to use the old approach here?

clients/application/src/main/java/org/eclipse/hono/application/client/AsyncCommandClient.java
kaniyan: My intention is to keep this PR small so that it will be easier to review. Hence I added only the `AsyncCommandClient` part in this PR. The consumer part is planned for the next PR.

clients/application/src/main/java/org/eclipse/hono/application/client/AsyncCommandClient.java
kaniyan: I tried to keep the methods as close to the old `AsyncCommandClient` considering the users who might be already using those methods.

b-abel: A good point. On the other hand, this is our chance to improve the user-facing interface as users need to adapt client code anyways.

clients/application/src/main/java/org/eclipse/hono/application/client/ApplicationClientFactory.java
kaniyan: > Can't we simply let the _ApplicationClientFactory_ implement _AsyncCommandClient_?

With implement do you mean  _ApplicationClientFactory_ to have directly  `sendAsyncCommand(...)`. FMPOV then this approach doesn't approach the name _ApplicationClientFactory_ as it is a factory for creating clients.


> One of the goals of the new client was to no longer create tenant specific instances of clients but instead allow users to simply pass in a tenant ID to the commands directly instead.

I understand the intention now. In this case, instead of `Future<AsyncCommandClient> createAsyncCommandClient(String tenantId)` I would change it to `Future<AsyncCommandClient> createAsyncCommandClient()`.

clients/application/src/main/java/org/eclipse/hono/application/client/AsyncCommandClient.java
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * A client for sending asynchronous commands.
+ *
+ * @see <a href="https://www.eclipse.org/hono/docs/api/command-and-control/">
+ *      Command &amp; Control API for AMQP 1.0 Specification</a>
+ * @see <a href="https://www.eclipse.org/hono/docs/api/command-and-control-kafka/">
+ *      Command &amp; Control API for Kafka Specification</a>
+ */
+public interface AsyncCommandClient {
```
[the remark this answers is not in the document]
kaniyan: I tried to keep the same name. I think so too that the `AsyncCommandSender` sounds better and also in sync with the other senders.

clients/application/src/main/java/org/eclipse/hono/application/client/ApplicationClientFactory.java
sophokles73: > I understand the intention now. In this case, instead of Future<AsyncCommandClient> createAsyncCommandClient(String tenantId) I would change it to Future<AsyncCommandClient> getOrCreateAsyncCommandClient().

FMPOV there is no need for a factory method here.

> FMPOV then this approach doesn't approach the name ApplicationClientFactory as it is a factory for creating clients.

I agree, we should rename it to just *ApplicationClient* :-)

kaniyan: > I agree, we should rename it to just _ApplicationClient_ :-)

:+1:

`ApplicationClient` can extend only one class. In that case, I would group `CommandClient` and `AsyncCommandClient` methods under `CommandSender`. Then the `ApplicationClient` can extend `CommandSender`

kaniyan: @sophokles73 I have renamed now to `ApplicationClient`. Also the `ApplicationClient` now implements the `CommandSender`. Would you mind taking a look?

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedAsyncCommandClient.java
```diff
@@ -0,0 +1,160 @@
...
+            final String correlationId,
+            final String replyId,
+            final Map<String, Object> properties,
+            final SpanContext context) {
+        Objects.requireNonNull(deviceId);
+        Objects.requireNonNull(command);
+        Objects.requireNonNull(correlationId);
+
+        final List<KafkaHeader> headers = getHeaders(deviceId, command, contentType, correlationId);
+        return sendAndWaitForOutcome(topic.toString(), tenantId, deviceId, data, headers, context);
```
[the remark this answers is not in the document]
kaniyan: In the above place since we already have a map of properties, it makes sense to add the additional properties to the Map pass the `Map` as an argument.

> TBH I still think the existence of this overloaded method is not justified.

In command router, we forward commands to the internal Kafka topic and there it is helpful to pass a list of headers from the `KafkaConsumerRecord`. Otherwise the headers need to be converted to a map of properties, which will finally get converted to a list of headers.

sophokles73:
@kaniyan can you rebase in order to reduce the number of changed files?

kaniyan:
> @kaniyan can you rebase in order to reduce the number of changed files?

sure. I am working on it.

kaniyan:
> > @kaniyan can you rebase in order to reduce the number of changed files?
> 
> sure. I am working on it.

@sophokles73 I have rebased it.

sophokles73:
@kaniyan can we doe the renaming of *ApplicationClientFactory* in a separate PR first?

kaniyan:
> @kaniyan can we doe the renaming of _ApplicationClientFactory_ in a separate PR first?

I have created a PR #2509 for that.

b-abel:
@kaniyan @sophokles73 If you rename `ApplicationClientFactory` to `ApplicationClient`, do you want to rename the methods like `createTelemetryConsumer()` too? And change all the JavaDoc that says things like "Creates a client for consuming data from Hono's north bound Event API" or the JavaDoc of the interface `MessageConsumer` that currently is "A client that consumes messages from Hono's north bound APIs."? Because having a client which methods return clients sounds a bit confusing to me.

clients/application/src/main/java/org/eclipse/hono/application/client/CommandSender.java
```diff
@@ -0,0 +1,163 @@
...
+     * @param correlationId The identifier to use for correlating the response with the request. Note: This parameter is
+     *            security sensitive. To ensure secure request response mapping choose correlationId carefully, e.g.
+     *            {@link java.util.UUID#randomUUID()}.
+     * @param replyId This identifier is to be set depending upon the underlying implementation (i.e AMQP or Kafka based).
+     *                <ul>
+     *                <li>In Kafka based implementation, this is not used and hence it can be set as {@code null}.</li>
+     *                <li>In AMQP based implementation, this is an arbitrary string which gets used for the response 
+     *                link address in the form of <em>command_response/${tenantId}/${replyId}</em>. Must match the
+     *                {@code replyId} passed to the command response receiver. The replyId must not be {@code null}</li>
+     *                </ul>
```
sophokles73, line 52: It feels very strange to refer to a particular (AMQP based) implementation here when the whole point of the interface is to abstract away the underlying transport protocol. If we refer to a corresponding *command response receiver* then it also feels strange that there is no method/way of getting one. Based on the fact that the replyId seems ot be relevant for the AMQP based implementation, I wonder why that is. If in fact we do not need the replyId at all, then why not simply remove it from here completely?

kaniyan:
> @kaniyan @sophokles73 If you rename `ApplicationClientFactory` to `ApplicationClient`, do you want to rename the methods like `createTelemetryConsumer()` too? And change all the JavaDoc that says things like "Creates a client for consuming data from Hono's north bound Event API" or the JavaDoc of the interface `MessageConsumer` that currently is "A client that consumes messages from Hono's north bound APIs."? Because having a client which methods return clients sounds a bit confusing to me.

@b-abel Good point.  FMPOV we could use the below methods instead of`createTelemetryConsumer()` and `createEventConsumer()`. IMHO the below methods provide more abstraction as the north bound users need not deal with the _MessageConsumer_ class. A small overhead is, that in addition to the senders (are already being cached), we should also store the consumers references in a collection in order to facilitate `unsubscribe` methods.

```
    Future<Void> subscribeTelemetry(
            String tenantId,
            Handler<DownstreamMessage<T>> messageHandler,
            Handler<Throwable> closeHandler);

    Future<Void> subscribeEvent(
            String tenantId,
            Handler<DownstreamMessage<T>> messageHandler,
            Handler<Throwable> closeHandler);

   // closes the consumer for the given tenant
    Future<Void> unSubscribeTelemetry(String tenantId);

    Future<Void> unSubscribeEvent(String tenantId);

    //closes all the consumers, senders and the connection
    Future<Void> close();
  
```
@sophokles73 @calohmn WDYT?

sophokles73:
I do not see a problem with simply keeping the *createTelemetryConsumer* and *createEventConsumer* methods the way they are. Just because the class is called *ApplicationClient* doesn't mean that it may not return Objects as the result of any of its methods IMHO.

clients/application/src/main/java/org/eclipse/hono/application/client/CommandSender.java
calohmn: The `replyId` is needed for AMQP because we support command response addresses with the custom `replyId` suffix there (`command_response/${tenantId}/${replyId}`). 
For the Kafka implementation, there is only the fixed `hono.command_response.${tenantId}` topic on which command responses will be sent and received (thereby not increasing the overall number of topics unnecessarily). 

Of course we _could_ get rid of the `replyId` in the AMQP implementation but that would be a bigger change (would also be an API break) and would be something to discuss separately.

For now, I think the javadoc could also be rephrased to keep it more generic: E.g. something like "The `replyId` is used as part of the resource to which a corresponding command response should be sent, if the messaging network specific implementation defines the use of such a variable part of the endpoint identifier. If not, `null` is to be used here."

sophokles73: IMHO we should actually describe it the other way around, i.e.
> If the underlying messaging infrastructure does not require a replyId, the specified value will be ignored.

WDYT?

calohmn: Since it's not a requirement of the messaging infrastructure but rather of the Command & Control implementation on top of it, I would prefer:

> If the messaging network specific Command & Control implementation does not require a replyId, the specified value will be ignored.

kaniyan: It sounds good. Just to be sure, I would add to the JavaDoc as below.

```
     * @param replyId An arbitrary string which will be used to create the reply-to address to be included in
     *                commands sent to devices of the tenant. If the messaging network specific Command & Control
     *                implementation does not require a replyId, the specified value will be ignored.
```

kaniyan:
> I do not see a problem with simply keeping the _createTelemetryConsumer_ and _createEventConsumer_ methods the way they are. Just because the class is called _ApplicationClient_ doesn't mean that it may not return Objects as the result of any of its methods IMHO.

Currently the existing methods return `MessageConsumer` and there it has only `close()` operation. IMHO refactoring the methods as in the above comment provides more abstraction and it seems to me more intuitive than what we have now.  If everyone like to keep the methods as it is, I am ok with it.  @calohmn What is your view on this?

clients/application/src/main/java/org/eclipse/hono/application/client/CommandSender.java
kaniyan: @sophokles73 I have updated it and pushed the changes. Would you mind taking a look?

clients/application-amqp/src/test/java/org/eclipse/hono/application/client/amqp/ProtonBasedApplicationClientTest.java
```diff
@@ -75,7 +77,9 @@
     @BeforeEach
     void setUp() {
         final var vertx = mock(Vertx.class);
+        when(vertx.eventBus()).thenReturn(mock(EventBus.class));
```
sophokles73, line 80: what do we need the event bus for?

clients/application-amqp/src/test/java/org/eclipse/hono/application/client/amqp/ProtonBasedApplicationClientTest.java
```diff
@@ -75,7 +77,9 @@
...
         connection = AmqpClientUnitTestHelper.mockHonoConnection(vertx);
+        when(connection.getVertx()).thenReturn(vertx);
```
sophokles73, line 82: this is already set up by *AmqpClientUnitTestHelper.mockHonoConnection(vertx)*

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -0,0 +1,101 @@
...
+            final String contentType, final String correlationId, final Map<String, Object> properties) {
+        final Map<String, Object> props = Optional.ofNullable(properties)
+                .orElse(new HashMap<>());
+
+        props.put(MessageHelper.APP_PROPERTY_DEVICE_ID, deviceId);
+        props.put(MessageHelper.SYS_PROPERTY_SUBJECT, subject);
+        props.put(MessageHelper.SYS_PROPERTY_CONTENT_TYPE,
+                Objects.nonNull(contentType) ? contentType : MessageHelper.CONTENT_TYPE_OCTET_STREAM);
+        props.put(MessageHelper.SYS_PROPERTY_CORRELATION_ID, correlationId);
+        props.put(KafkaMessageHelper.HEADER_RESPONSE_REQUIRED, true);
```
sophokles73, line 97: shouldn't this be set only if this is a request-response command?

clients/application-amqp/src/test/java/org/eclipse/hono/application/client/amqp/ProtonBasedApplicationClientTest.java
kaniyan: `SenderCachingServiceClient` invokes the below piece of code if `isTenantSpecificLink` is true. Hence the event bus is needed in the test.
```
        if (isTenantSpecificLink) {
            connection.getVertx().eventBus().consumer(Constants.EVENT_BUS_ADDRESS_TENANT_TIMED_OUT,
                    this::handleTenantTimeout);
        }
```

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
kaniyan: _HEADER_RESPONSE_REQUIRED_ will be set to _false_ when we send one way commands. In the consequent PRs, I will add methods to send one-way commands. In the already existing API (`AsyncCommandClient`) , the method `sendAsyncCommand()` is used to send only request-response commands in asynchronous manner.

calohmn:
> > I do not see a problem with simply keeping the _createTelemetryConsumer_ and _createEventConsumer_ methods the way they are. Just because the class is called _ApplicationClient_ doesn't mean that it may not return Objects as the result of any of its methods IMHO.

> Currently the existing methods return `MessageConsumer` and there it has only `close()` operation. IMHO refactoring the methods as in the above comment provides more abstraction and it seems to me more intuitive than what we have now.  If everyone like to keep the methods as it is, I am ok with it.  @calohmn What is your view on this?

I'm also ok with keeping the methods as is.
Using separate `(un)subscribeTelemetry` methods instead doesn't really seem that much more intuitive to me - it is just different. I admit that getting a `MessageConsumer` with only a `close()` method does perhaps seem a bit odd, but possibly offers more flexibility in that the user can cast to a KafkaMessageConsumer that possibly offers more methods (could be useful in integration tests).

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
kaniyan: I have set it to _true_ as there are `sendAsyncCommand(..)` methods in this PR that support only _request-response_ commands. When adding _one-way_ commands in the next PRs, I wanted to get the value for _HEADER_RESPONSE_REQUIRED_ using a parameter of `getHeaderProperties()`. I have now made these changes and pushed a commit.

clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedCommandSender.java
```diff
@@ -0,0 +1,100 @@
...
+
+        return getOrCreateSenderLink(CommandConstants.NORTHBOUND_COMMAND_REQUEST_ENDPOINT, tenantId)
+                .recover(thr -> Future.failedFuture(StatusCodeMapper.toServerError(thr)))
+                .compose(sender -> {
+                    final String targetAddress = AddressHelper
+                            .getTargetAddress(CommandConstants.NORTHBOUND_COMMAND_REQUEST_ENDPOINT, tenantId, deviceId,
+                                    connection.getConfig());
+                    final Message msg = createMessage(tenantId, deviceId, command, contentType, data, correlationId,
+                            replyId, targetAddress, properties);
+                    return sender.sendAndWaitForOutcome(msg, newChildSpan(context, "send Command " + command));
```
calohmn, line 77: Just use "send command" as span name here - the span name should be fixed (will occur in the "operation" dropdown in the Jaeger UI).

clients/application/src/main/java/org/eclipse/hono/application/client/CommandSender.java
```diff
@@ -0,0 +1,156 @@
...
+     *            security sensitive. To ensure secure request response mapping choose correlationId carefully, e.g.
+     *            {@link java.util.UUID#randomUUID()}.
+     * @param replyId An arbitrary string which will be used to create the reply-to address to be included in commands
+     *            sent to devices of the tenant. If the messaging network specific Command &amp; Control implementation does
+     *            not require a replyId, the specified value will be ignored.
+     * @return A future indicating the result of the operation.
+     *         <p>
+     *         If the command was accepted, the future will succeed.
+     *         <p>
+     *         The future will fail with a {@link ServiceInvocationException} if the command could not be forwarded
```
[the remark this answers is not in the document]
kaniyan: A corresponding _import_ statement  for `ServiceInvocationException` is already there in  `CommandSender`. I think it should be already resolved.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -0,0 +1,102 @@
...
+        return sendAndWaitForOutcome(topic.toString(), tenantId, deviceId, data, headerProperties, context);
+    }
+
+    private Map<String, Object> getHeaderProperties(final String deviceId, final String subject,
+            final String contentType, final String correlationId, final boolean responseRequired,
+            final Map<String, Object> properties) {
+        final Map<String, Object> props = Optional.ofNullable(properties)
+                .orElse(new HashMap<>());
+
+        props.put(MessageHelper.APP_PROPERTY_DEVICE_ID, deviceId);
```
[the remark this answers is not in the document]
kaniyan: @calohmn I have pushed a new commit with the changes. Would you mind taking a look?

cli/src/main/java/org/eclipse/hono/cli/AppConfiguration.java
```diff
@@ -116,13 +142,18 @@ public ApplicationClientFactory clientFactory() {
      *
      * @param vertx The vertx instance to be used.
      * @param kafkaConsumerConfigProperties The consumer configuration properties.
+     * @param producerFactory The factory to use for creating Kafka producers.
+     * @param kafkaProducerConfigProperties The producer configuration properties.
      * @return The factory.
```
calohmn, line 147: "The client."

clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedCommandSender.java
calohmn: The " + command" part seems to be still in there.

kaniyan: Ah ok, now I see what you meant. Done

calohmn (approved):
LGTM

PR #2499, opened 2021-02-25
Use "client facing error message" in AMQP adapter

calohmn:
Also move "getErrorCondition(Throwable)" out of adapter base class.
That method is only needed for the AMQP adapter.


sophokles73 (approved):
LGTM

PR #2500, opened 2021-02-25
Do not delete images from Hugo theme when building docs locally

b-abel:
A recent change in the build scripts for the documentation causes that on each build images of the Hugo theme are deleted. While this step makes sense in the Jenkins pipeline, it is not necessary for local builds. Because the images are simply deleted from a directory under version control, this is always displayed as changes in my IDE, which is unnecessarily annoying. So just stopping to delete the files in the local builds is the simplest solution.


sophokles73 (approved):
LGTM

Issue #2501, opened 2021-02-25
Support TLS 1.3 in Hono components by default


sophokles73:
The Hono components all support setting the supported TLS protocol versions by means of a configuration property.
By default, i.e. if not set explicitly, only TLS 1.2 is supported. With TLS 1.3 having been around for quite some time now, it seems reasonable to add it to the protocols supported by default.

sophokles73:
Fixed via #2502

PR #2502, opened 2021-02-25
Enable TLS 1.3 by default in all components

sophokles73:
This addresses #2501

Changed default configuration properties to prefer TLS 1.3 and 1.2 in
descending order.
Also added integration tests for verifying that devices and downstream
applications can use TLS 1.3 to connect to adapters and AMQP Messaging
Network.


dejanb (approved):
Looks good

Issue #2495, opened 2021-02-23
NPE in MQTT adapter when trying to parse property bag


calohmn:
The following exception was observed in MQTT adapter logs:
````
java.lang.NullPointerException: null
	at java.base/java.lang.StringBuilder.<init>(Unknown Source)
	at org.eclipse.hono.util.ResourceIdentifier.createStringRepresentation(ResourceIdentifier.java:150)
	at org.eclipse.hono.util.ResourceIdentifier.setResourcePath(ResourceIdentifier.java:101)
	at org.eclipse.hono.util.ResourceIdentifier.<init>(ResourceIdentifier.java:60)
	at org.eclipse.hono.util.ResourceIdentifier.fromString(ResourceIdentifier.java:175)
	at org.eclipse.hono.adapter.mqtt.PropertyBag.fromTopic(PropertyBag.java:71)
	at org.eclipse.hono.adapter.mqtt.AbstractVertxBasedMqttProtocolAdapter.lambda$handlePublishedMessage$46(AbstractVertxBasedMqttProtocolAdapter.java:872)
	at java.base/java.util.Optional.flatMap(Unknown Source)
	at org.eclipse.hono.adapter.mqtt.AbstractVertxBasedMqttProtocolAdapter.handlePublishedMessage(AbstractVertxBasedMqttProtocolAdapter.java:872)
````

The issue is in the way the `PropertyBag` class treats the MQTT topic parameter.

To reproduce:
`PropertyBag.fromTopic("/?")` 
or `PropertyBag.fromTopic("/?test=1")` 
or `PropertyBag.fromTopic("/test")`.

`PropertyBag.fromTopic("/")` produces an `ArrayIndexOutOfBoundsException`.

The `IllegalArgumentException` produced by `PropertyBag.fromTopic("test/?%%")` should also be taken care of.

PR #2503, opened 2021-02-25
 [#2495] Fix ResourceIdentifier NPE in case of empty endpoint

calohmn:
This is for #2495:
First commit:
Trying to create a `ResourceIdentifier` with an empty endpoint segment caused an NPE.
Since I don't see a use case for supporting an empty endpoint segment in `ResourceIdentifier` objects, I've now made it explicit that the endpoint segment must not be empty.
Therefore, an `IllegalArgumentException` is now thrown instead, documented in the javadocs of the factory methods.

Places where a `ResourceIdentifier` gets created from a String should be adapted now, replacing the parameter String not-null-or-empty check with `ResourceIdentifier.isValid(resource)`. I'll see to that after this PR gets merged.

2nd commit:
Refactor ResourceIdentifier constructors/factory methods: Using only one constructor now, making fields final. Also improving javadocs.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/ResourceIdentifier.java
+++ b/core/src/main/java/org/eclipse/hono/util/ResourceIdentifier.java
@@ -12,10 +12,9 @@
  *******************************************************************************/
 package org.eclipse.hono.util;
 
-import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 
 /**
  * A unique identifier for a resource within Hono.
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/eclipse/hono/util/ResourceIdentifier.java
```diff
@@ -144,49 +82,62 @@ private String createStringRepresentation(final int startIdx) {
...
      */
     public static ResourceIdentifier fromString(final String resource) {
         Objects.requireNonNull(resource);
-        if (resource.isEmpty()) {
-            throw new IllegalArgumentException("resource must not be empty");
-        }
-        return new ResourceIdentifier(resource);
+        final String[] path = resource.split("/");
+        final List<String> pathSegments = new ArrayList<>(Arrays.asList(path));
+        return new ResourceIdentifier(pathSegments.toArray(new String[pathSegments.size()]));
```
sophokles73, line 112: why don't we simply pass in the *path* variable?

calohmn: Right, I've changed it now.

sophokles73: There seem to be some unused imports remaining ..

sophokles73 (approved):
LGTM

PR #2508, opened 2021-02-26
[#2064] Add support for MQTT error topic subscription

calohmn:
For #2064:
This PR adds support for MQTT devices subscribing on a new `error` topic, on which the MQTT adapter will publish error messages in case processing of a telemetry, event or command response message by the device fails.

This also introduces new ways in which errors thrown while processing such device messages are handled, meaning the MQTT device connection isn't necessarily closed then anymore.

A remaining TODO for #2064 is the detection of a kind of terminal error, for which the MQTT connection should always be closed (e.g. the device/tenant not existing anymore or being disabled).


adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1274,6 +1321,119 @@ protected final void onSubscribe(final MqttSubscribeMessage subscribeMsg) {
...
+                                : "published error message");
+                    }).onFailure(thr -> {
+                        log.debug("error publishing error message to {} [tenant-id: {}, MQTT client-id: {}, QoS: {}, topic: {}]",
+                                targetInfo, subscription.getTenant(), endpoint.clientIdentifier(),
+                                subscription.getQos(), publishTopic, thr);
+                        TracingHelper.logError(span, "failed to publish error message", thr);
+                    }).compose(msgId -> {
+                        if (MqttQoS.AT_LEAST_ONCE.equals(subscription.getQos())) {
+                            final Promise<Void> ackOrTimeoutPromise = Promise.promise();
+                            pendingAcks.add(
```
kaniyan, line 1419: We are adding to `pendingAcks` once the message has been successfully sent. I think there is still chance that a `PUBACK` has already arrived even before adding to `pendingAcks`. In this case, we will wait for `PUBACK` and then time out. Even though we have a timeout mechanism, but this is misleading as it seems no `PUBACK` has arrived. The same behavior is also there while we publish commands.

I think it would be appropriate to add to the `pendingAcks` first and then publish the message using the below method. In this case, we have to generate the message identifier, add it to `pendingAcks` followed by publishing the message. We can also handle this in a separate PR.

```
MqttEndpoint publish(String topic, Buffer payload, MqttQoS qosLevel, boolean isDup, boolean isRetain, int messageId, Handler<AsyncResult<Integer>> publishSentHandler);
```

kaniyan:
In this PR, there are IT tests to test this feature for telemetry and event messages. FMPOV it would be good also to have some tests for command response.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/ErrorSubscription.java
```diff
@@ -0,0 +1,263 @@
...
+/**
+ * A device's MQTT subscription for error messages.
+ * <p>
+ * Supported topic filters: {@code error|e/[TENANT]/[DEVICE_ID]/#}
+ * <br>(empty tenant/device segments can also be omitted)
+ * <p>
+ * Examples:
+ * <ol>
+ * <li>{@code error/DEFAULT_TENANT/4711/#} unauthenticated device</li>
+ * <li>{@code error/#} - authenticated device</li>
```
sophokles73, line 40: based on the pattern above, wouldn't this rather be `error///#`? This would also better fit the pattern we use for subscribing to commands, wouldn't it?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
kaniyan: I have tested the above scenario with introducing some extra statements so that after successfully publishing an error message it takes some time till `pendingAcks.add(` is called. I have noticed that the `PUBACKs` have been always processed by the same event-loop thread. So we wouldn't end up with the above situation of missing `PUBACKs`

calohmn: The above "publish()" invocation will complete the result promise directly (without the event loop thread calling other handlers in between). Therefore it is safe to invoke "pendingAcks.add" here since there can't be an incoming PUBACK getting handled between "publish" and "pendingAcks.add". This is also because only one event loop thread is involved in the handling of requests (like SUBSCRIBE/PUBACK) for the MQTT connection here (`AbstractVertxBasedMqttProtocolAdapter` is a verticle).

This is of course somewhat hidden in the `MqttEndpointImpl.publish` implementation. Therefore things could be made more obvious by doing the `pendingAcks.add` invocation *before* the *publish* invocation. I would leave that as something that can be done in a subsequent PR.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/ErrorSubscription.java
calohmn: The minimal topic filter for command subscriptions is `command///req/#`, so there's still a non-empty segment after the empty tenant and device id parts.
I just think `error/#` is more intuitive and feels more natural than `error///#`. For an authenticated device it's just as short as using e.g. "event" for the other direction.
Of course `error/#` must not be interpreted as "get all error messages for *all* tenants/devices". But I don't think that this would be the intuitive expectation anyway since for authenticated (non-gateway) devices everything is done in the context of the device derived from the authentication data.
And users still can use `error///#` if they prefer.

calohmn:
> In this PR, there are IT tests to test this feature for telemetry and event messages. FMPOV it would be good also to have some tests for command response.

@kaniyan Good point. I thought about deriving a new `CommandResponseMqttIT` from `MqttPublishTestBase` for that. But that requires some `helper.amqpApplicationClient.createCommandResponseConsumer()` method which isn't there yet. So I would leave that for later.

Besides, I've made changes corresponding to your comments above.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/ErrorSubscription.java
sophokles73: Is it a supported use case that a gateway device subscribes to errors for all devices that it may act on behalf of? If so, what is the topic that it would need to subscribe to? `error//+/#`?

calohmn: Yes. The possible topic filters for such a case are `error/#`, `error//+/#` and `error/${tenant-id}/+/#` (see also the "Receiving Error Messages (authenticated Gateway)" chapter in `mqtt-adapter.md`).
The gateway would otherwise use `error//${device-id}/#` or `error/${tenant-id}/${device-id}#` to get error messages for a specific device it acts on behalf of.

sophokles73: So in these cases we follow MQTT wildcard semantics. However, `error/#` does not, right? FMPOV that is not very intuitive but confusing ...

calohmn: I don't see what you mean. `error/#` with the multi-level # wildcard would match all error topic strings that the device may receive messages for, meaning the # matches the tenant, device and following topic levels.

sophokles73: Well, with the command subscription topics we deliberately made the change from `command/+/+/req/#` to `command///req/#` because the `+` wildcard character suggests that *any* tenant and device ID would match. FMPOV the same argument applies here as well, doesn't it?

calohmn: Well, the topic with empty tenant/device id segments are special anyway. in the sense that they make it clear that there is no single big, tenant/device-independent topic tree to tap into with the subscription topic filter. The `command///req/[*]/[*]` publish topic is used *across* devices.

For example, if a person manages 2 different (non-gateway) devices "deviceA" and "deviceB" and uses `command///req/#` for both their subscriptions, the observation to make is that both devices get their commands with topics like `command///req/111/setBrightness`. 
But, that person will notice that "deviceA" will only get the  `command///req/[*]/[*]` messages for "deviceA", although command messages for "deviceB" also fit the `command///req/#` pattern.

So a user has to understand that the messages received via a subscription are always just the ones for the authenticated device (or the gateway-managed devices).

In that sense, the understanding that `error/#` will also just match the authenticated device or gateway-managed devices isn't that unintuitive.

I see that mandating `error///#` will make it just a bit more obvious, what kind of topics a subscription of authenticated "deviceA" of "tenantX" will get (in the sense that it's necessarily `error///something` instead of potentially `error/tenantX/deviceA/something`). But I still think it still looks a bit odd to use `error///#` instead of `error/#` (with the realization in mind that the received messages are always device-specific anyway).

site/documentation/content/user-guide/mqtt-adapter.md
```diff
@@ -499,6 +507,139 @@ After a command has arrived as in the above example, the response is sent using
...
+
+* `${endpoint-type}`: The endpoint type of the device message that caused the error. Its value is either `telemetry`, `event` or the respective shorthand version. In case of a command response device message `command-response` or `c-s` is used.
+* `${correlation-id}`: The identifier that may be used to correlate the error message with the device message that caused the error. The identifier is either the value of a *correlation-id* property bag value contained in the device message topic, or the identifier is the *packet-id* of the device message if it was sent with QoS 1. Otherwise, a value of `-1` is used.
+* `${error-status}`: The HTTP status code of the error that was caused by the device message.
+
+{{% note title="Examples" %}}
+Since the subscription on the error topic needs to be done on the same MQTT connection that is also used for publishing the telemetry, event or command response messages, the Mosquitto MQTT Command Line Client cannot be used. The [MQTT CLI](https://hivemq.github.io/mqtt-cli/) tool with its [shell mode](https://hivemq.github.io/mqtt-cli/docs/shell.html) is an alternative that supports using one MQTT connection for both subscribing and publishing.
```
calohmn, line 525: BTW, I've tried using the `mosquitto_rr` client (mosquitto >= v1.6), but that doesn't work since mosquitto_rr already follows the MQTT 5 spec for Request/Response here (even for MQTT 3.11 servers) in that it doesn't allow wildcards in the response topic. Considering the error topic as a response topic wouldn't fit that well here anyway.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/ErrorSubscription.java
sophokles73: Based on your arguments we could also simply let devices subscribe to `command/#` in order to receive commands, couldn't we? The problem with this approach is FMPOV that we start introducing *special cases* that we need to handle differently than usual. We deliberately chose to use the common addressing scheme *endpoint/tenant/deviceId/...* throughout our APIs. I do not see any convincing reason to divert from it now, in particular since nobody seems to have a problem with subscribing to `command///req/#` ...

calohmn: Ok, to keep it in accordance to the command pattern, I've made the corresponding changes so that `error///#` needs to be used now and `error/#` isn't supported.

site/documentation/content/user-guide/mqtt-adapter.md
```diff
@@ -499,6 +507,137 @@ After a command has arrived as in the above example, the response is sent using
...
+
+An alternative way of dealing with errors involves keeping the connection intact and letting the MQTT adapter publish a corresponding error message on a specific error topic to the device. To enable that behaviour, the device sends an MQTT *SUBSCRIBE* packet with a topic filter as described below on *the same MQTT connection* that is also used for publishing the telemetry, event or command response messages. Devices can subscribe with QoS 1 or QoS 0. The adapter indicates the outcome of the subscription request by sending back a corresponding *SUBACK* packet. The SUBACK packet will contain *Success - QoS 0* (`0x00`) or *Success - QoS 1* (`0x01`) for a valid error topic filter indicating QoS 0 or 1 and will contain the *Failure* (`0x80`) value for an invalid or unsupported filter. In order to again activate the default error handling behaviour, the device can send an MQTT *UNSUBSCRIBE* packet to the adapter, including the same topic filter that has been used to subscribe.
```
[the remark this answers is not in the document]
calohmn: I explicitly want to support using QoS 1 here:
Consider a scenario where a device, whose device registration entry got deleted after the device opened the MQTT connection, sends a telemetry message and has a QoS 1 error subscription.
With the still-to-be-added changes of doing a disconnect in that case (deleted tenant/device is to be considered a terminal error), the behaviour will then be:
First the device gets the error message ("device unknown"), then the adapter waits for the PUBACK of that error message from the device (with timeout), and then the adapter does the disconnect. That way it is assured that the device actually properly got the error message.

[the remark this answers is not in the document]
sophokles73: Why shouldn't that also be the case with QoS 0? The only important thing here is that the PUBLISH packet is sent *before* the DISCONNECT packet, isn't it? But FMPOV this is independent from the QoS in use but depends only on the order of the messages that the MQTT adapter receives the command and the disconnect event in, or am I mistaken?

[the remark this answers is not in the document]
calohmn: The DISCONNECT packet is only meant to be sent from the client to the server. So what is called here is `MqttEndpoint#close` which simply closes the socket. Looking at the implementation, it seems indeed that `flush` gets called beforehand, so all previous data (i.e. the error PUBLISH message) *should* probably have been sent when the socket gets closed.

So, yes, I guess it should be enough to leave it to the netty layer to ensure that the error message gets sent. That means we shouldn't really need QoS 1. I'll update the commit.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1154,6 +1182,21 @@ protected final void handlePublishedMessage(final MqttPublishMessage message) {
...
+                // first check for a subscription with the device id taken from the topic or the authenticated device
+                // (handles case where a gateway is publishing for specific edge devices and has individual error subscriptions for them)
+                result = errorSubscriptions.get(ErrorSubscription.getKey(context.tenant(), context.deviceId()));
+            }
+            if (result == null && context.authenticatedDevice() != null) {
+                // otherwise check using the authenticated device (ie. potentially gateway) id
+                result = errorSubscriptions
+                        .get(ErrorSubscription.getKey(context.tenant(), context.authenticatedDevice().getDeviceId()));
+            }
+            return result;
```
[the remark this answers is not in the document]
calohmn: This method potentially uses two different keys derived from the MqttContext, so `ErrorSubscription.getKey(MqttContext)` without an extra parameter wouldn't work. And adding that parameter (or a 2nd getKey method) wouldn't make the code much easier to understand compared to the above code.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/MqttProtocolAdapterProperties.java
```diff
@@ -77,6 +77,9 @@ public final void setCommandAckTimeout(final int timeout) {
      * This timeout is used by the MQTT adapter for commands published with QoS 1. If there is no acknowledgement
      * within this time limit, then the command is settled with the <em>released</em> outcome.
      * <p>
+     * This timeout is also used for error messages published with QoS 1, to wait before further processing of
```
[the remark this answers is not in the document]
calohmn: Based on the other comment, I've removed the sentence.

tests/src/test/java/org/eclipse/hono/tests/mqtt/MqttPublishTestBase.java
```diff
@@ -353,9 +650,32 @@ private void doTestUploadMessages(
         assertMessageReceivedRatio(messagesReceived, messageCount.get(), ctx);
     }
 
+    private Future<Void> subscribeToErrorTopic(final String errorTopic) {
+        final Promise<Void> result = Promise.promise();
+        final AtomicReference<Integer> subMessageIdRef = new AtomicReference<>();
+        mqttClient.subscribeCompletionHandler(subAckMsg -> {
+            if (subAckMsg.grantedQoSLevels().contains(0) && subAckMsg.messageId() == subMessageIdRef.get()) {
+                result.tryComplete();
```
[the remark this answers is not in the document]
calohmn: Only if there is another SUBACK for another SUBSCRIBE coming later without the `subscribeCompletionHandler` having been set again. I've adapted the code and removed the "try" parts.

sophokles73 (approved):
LGTM

Issue #2460, opened 2021-02-08
[#8] Duplicate logic for parsing of Kafka headers


b-abel:
I found that the classes `RecordUtil` and `KafkaMessageHelper` implement the same use case: getting and decoding Kafka headers. It would be good to have this implemented only once.

b-abel:
The main difference I found is, that `RecordUtil` does not return a value if it found another header with the same key while `KafkaMessageHelper` always returns the first header with a matching key. (On the producer-side we should anyways always ensure that we cannot have duplicate keys).

kaniyan:
It is better to avoid duplication and reuse the code. IMHO `KafkaMessageHelper` is already being used for the `telemetry/event` side and it could also be extended to process `commands` too.

calohmn:
How about renaming `KafkaMessageHelper` to `RecordHelper` then (`Message` is more of an AMQP term) and adding the duplicate header check?

b-abel:
> How about renaming `KafkaMessageHelper` to `RecordHelper` then (`Message` is more of an AMQP term) 

I'm fine with that.

> and adding the duplicate header check?

My implementation does not enforce this on purpose. I would rather accept a message with a duplicate key in the header than dropping it as there is no way to inform the initial sender about it. We should better spend our energy on the producer-side and implement producer in a way that prevents duplicate header keys and clearly state that requirement in the command API.

calohmn:
> I would rather accept a message with a duplicate key in the header 

What if the producer is the northbound application (sending commands and not using our Kafka client classes)? I would rather be more strict here than work with the potentially wrong header value. And about informing the initial sender: this is a problem we face also concerning other ways in which the record might be invalid (missing headers or invalid values). So, I would rather not make a special case for duplicate headers here. 
In order to inform the sender, an idea would be to introduce an error topic to which the northbound application can subscribe and get informed about such errors.

kaniyan:
@b-abel  @calohmn We already deal with handling of multiple headers having the same key in Hono. In case of HTTP headers, we take the first header value. We could apply the same here in case of kafka headers too and keep it uniform. IMHO what is important, is to clearly document the behavior in the specifications.

calohmn:
Well, taking the first header and not causing an error is still not my preference, but at least there's some consistency to it. So, ok.
Interestingly, the Kafka `Headers` interface only has a `lastHeader` method for getting a single header.

PR #2510, opened 2021-02-26
[#2460] Reuse methods for processing Kafka headers

kaniyan:
This is for issue #2460


kaniyan:
@b-abel As we discussed, I have raised this PR for issue #2460. Would you mind also taking a look?

calohmn:
You could replace `new KafkaHeaderImpl` by `KafkaHeader.header` in `KafkaMessageHelper`.
Do you want to do the suggested renaming to `RecordHelper` also here as a 2nd commit?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaMessageHelper.java
```diff
@@ -199,4 +202,18 @@ public static boolean isTtlElapsed(final List<KafkaHeader> headers) {
         }
     }
 
+    /**
+     * Gets the String value of a specific <em>header</em>.
+     * <p>
+     * If the headers contains multiple occurrences of the same key, the value of its first
```
calohmn, line 208: "contain" (also in `getHeaderValue`)

kaniyan:
> You could replace `new KafkaHeaderImpl` by `KafkaHeader.header` in `KafkaMessageHelper`.

It was already there in the existing code and it is good to replace it with `KafkaHeader.header`

kaniyan:
> Do you want to do the suggested renaming to `RecordHelper` also here as a 2nd commit?

FMPOV `RecordHelper` seems to be quite generic. How about `KafkaRecordHelper`?

calohmn:
> How about KafkaRecordHelper?

I'm fine with that.

kaniyan:
> Do you want to do the suggested renaming to `RecordHelper` also here as a 2nd commit?

@calohmn I have added a 2nd commit regarding renaming of that file and also updated in accordance with the above comments.

calohmn (approved):
LGTM

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaRecordHelper.java
```diff
@@ -199,4 +201,18 @@ public static boolean isTtlElapsed(final List<KafkaHeader> headers) {
...
+     * <p>
+     * If the headers contain multiple occurrences of the same key, the value of its first
+     * occurrence is returned.
+     *
+     * @param headers The headers to retrieve the value from.
+     * @param key The header key.
+     * @return The value or {@code null} if the headers do not contain a value of the expected type for the given key.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     */
+    public static String getStringHeaderValue(final List<KafkaHeader> headers, final String key) {
```
b-abel, line 215: Why does this not return an optional like the other methods here?

kaniyan: This method has been moved from the `RecordUtil` class (as that class has been deleted). The classes which use this utility method expect that this method returns a `String` value if present else _null_.

b-abel: I know that. But should not `KafkaRecordHelper` define its interface and code that uses it be adapted instead?

kaniyan: FMPOV it is more of a taste if to return a _String_ or an _Optional_ for a utility method. Considering that we already return _Optional_ for an already existing method in that class, I find it reasonable to have the same style. If `getStringHeaderValue(...)` is to return an _Optional_, I find it there is not much value in having that method and rather better to use the `getHeaderValue(headers, key, String.class)`.  I have made changes to use the  `getHeaderValue(headers, key, String.class)`.

b-abel:
LGTM

kaniyan:
The conflicts due to merge of PR #2498 has been resolved.

PR #2513, opened 2021-03-01
#2469 Adding getTimeUntilDisconnectNotification() to DownstreamMessage

fkaltner:
This is done as preparation for adapting ITs to Kafka. When ITs are based upon DownstreamMessage, `DownstreamMessage#getTimeUntilDisconnectNotification ` can be used to obtain the time until disconnect notification instead of `TimeUntilDisconnectNotification.fromMessage(msg)`.

### DIFF
--- a/clients/application/src/main/java/org/eclipse/hono/application/client/DownstreamMessage.java
+++ b/clients/application/src/main/java/org/eclipse/hono/application/client/DownstreamMessage.java
@@ -144,8 +144,7 @@ public interface DownstreamMessage<T extends MessageContext> extends Message<T> 
 
             if (tenantId != null && deviceId != null) {
                 final TimeUntilDisconnectNotification notification =
-                        new TimeUntilDisconnectNotification(tenantId, deviceId, ttd,
-                                TimeUntilDisconnectNotification.getReadyUntilInstantFromTtd(ttd, creationTime), creationTime);
+                        new TimeUntilDisconnectNotification(tenantId, deviceId, ttd, creationTime);
                 return Optional.of(notification);
             }
         }
--- a/clients/application/src/main/java/org/eclipse/hono/application/client/DownstreamMessage.java
+++ b/clients/application/src/main/java/org/eclipse/hono/application/client/DownstreamMessage.java
@@ -16,7 +16,6 @@ package org.eclipse.hono.application.client;
 import java.time.Instant;
 import java.util.Optional;
 
-import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.QoS;
 import org.eclipse.hono.util.TimeUntilDisconnectNotification;
 
@@ -137,7 +136,7 @@ public interface DownstreamMessage<T extends MessageContext> extends Message<T> 
 
         if (ttd == null) {
             return Optional.empty();
-        } else if (ttd == 0 || MessageHelper.isDeviceCurrentlyConnected(ttd,
+        } else if (ttd == 0 || TimeUntilDisconnectNotification.isDeviceCurrentlyConnected(ttd,
                 creationTime != null ? creationTime.toEpochMilli() : null)) {
             final String tenantId = getTenantId();
             final String deviceId = getDeviceId();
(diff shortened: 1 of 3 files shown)


clients/application/src/main/java/org/eclipse/hono/application/client/DownstreamMessage.java
```diff
@@ -123,4 +125,31 @@ default boolean isSenderConnected(final Instant now) {
...
+            return Optional.empty();
+        } else if (ttd == 0 || MessageHelper.isDeviceCurrentlyConnected(ttd,
+                creationTime != null ? creationTime.toEpochMilli() : null)) {
+            final String tenantId = getTenantId();
+            final String deviceId = getDeviceId();
+
+            if (tenantId != null && deviceId != null) {
+                final TimeUntilDisconnectNotification notification =
+                        new TimeUntilDisconnectNotification(tenantId, deviceId, ttd,
+                                TimeUntilDisconnectNotification.getReadyUntilInstantFromTtd(ttd, creationTime), creationTime);
```
sophokles73, line 148: it looks very starnge to pass in ttd and creation time as parameters and then also being required to use the same parameters to compute another one. Shouldn't the constructor take care of that?

fkaltner: Absolutely! It's implemented in `TimeUntilDisconnectNotification#fromMessage` which I used as a template, I'll fix it there as well.

clients/application/src/main/java/org/eclipse/hono/application/client/DownstreamMessage.java
```diff
@@ -123,4 +125,30 @@ default boolean isSenderConnected(final Instant now) {
...
+     * @return A notification if the message contains a TTD value {@link Optional#empty()} otherwise.
+     */
+    default Optional<TimeUntilDisconnectNotification> getTimeUntilDisconnectNotification() {
+
+        final Integer ttd = getTimeTillDisconnect();
+        final Instant creationTime = getCreationTime();
+
+        if (ttd == null) {
+            return Optional.empty();
+        } else if (ttd == 0 || MessageHelper.isDeviceCurrentlyConnected(ttd,
```
sophokles73, line 140: I'd rather not introduce this dependency on *MessgeHelper* here as it is basically a helper class for dealing with AMQP messages and pulls in a lot of transitive dependencies. Maybe we can encapsulate the required logic in *TimeUntilDisconnectNotification* as well?

fkaltner: Done.

core/src/main/java/org/eclipse/hono/util/TimeUntilDisconnectNotification.java
```diff
@@ -187,4 +185,54 @@ public long getMillisecondsUntilExpiry() {
...
+     *
+     * @param ttd The TTD value.
+     * @param creationTime The creation time of the message. If {@code null} the device is considered as disconnected.
+     * @return {@code true} if the TTD value contained in the message indicates that the device will
+     *         stay connected for some additional time.
+     */
+    public static boolean isDeviceCurrentlyConnected(final Integer ttd, final Long creationTime) {
+
+        return Optional.ofNullable(ttd).map(ttdValue -> {
+            if (ttdValue == MessageHelper.TTD_VALUE_UNLIMITED) {
```
sophokles73, line 205: Hmmm, same problem, I guess. If we really want to get rid of the AMQP dependencies we will probably need to move the field constants to some more technology-agnostic class, e.g. *Constants*. But I guess we should do that in another PR ...

sophokles73 (approved):

core/src/main/java/org/eclipse/hono/util/TimeUntilDisconnectNotification.java
fkaltner: I think, if we really want to get rid of unnecessary dependencies of the client module we need to cut the Maven dependency to hono-core which is still AMQP-specific.

sophokles73: If we only use classes from hono-core which do not depend on e.g. vertx-proton and/or proton-j then we can at least exclude those artifacts from the declared dependency to hono-core. But that is not possible if we need e.g. MessageHelper and MessageHelper uses proton-j ...

PR #2514, opened 2021-03-01
#2469 Only setting bootstrap.servers property when Kafka profile is e…

fkaltner:
This is required so that IntegrationTestSupport can establish a connection by evaluating the `DOWNSTREAM_BOOTSTRAP_SERVERS` property to the AMQP network or Kafka depending on the active Maven profile.

This also removes unnecessary plugin executions when the Kafka profile is not set.


sophokles73 (approved):
LGTM

PR #2516, opened 2021-03-02
Support gathering of code coverage information

sophokles73:
The build has been extended to support running the JaCoCo plugin for
gathering code coverage information when running unit tests.
By default, the plugin will be skipped. It can be activated by means of
setting the "jacoco.skip" Maven property to value "false".

This might be helpful for some downstream consumers of our code who need
some "proof-of-quality". We can also use this data to (sporadically)
asses whether we have some "blind spots" in our code that should be
tested more thoroughly.


kaniyan (approved):
LGTM

PR #2517, opened 2021-03-02
#2469 Adapt integration tests for telemetry and events to Kafka

fkaltner:
This PR adapts the telemetry and event based integration tests to the new `ApplicationClient` interface so that they can be executed with an AMQP 1.0 network _or_ Kafka.

All Command and Control specific tests are skipped for the Kafka profile. The goal is to have the Kafka ITs in a state where they can be executed successfully. 

Skipped Tests:
- CommandAndControlAmqpIT:*
- CommandAndControlMqttIT:*
- CoapTestBase:testUploadMessagesWithTtdThatReplyWithCommand
- CoapTestBase:testUploadMessagesWithTtdThatReplyWithOneWayCommand
- HttpTestBase:testHandleConcurrentUploadWithTtd
- HttpTestBase:testUploadMessagesWithTtdThatReplyWithCommand
- HttpTestBase:testUploadMessagesWithTtdThatReplyWithOneWayCommand

As soon as the Command and Control implementation is complete those tests can be included/migrated to the Kafka profile as well.

### DIFF
--- a/tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
@@ -909,18 +909,13 @@ public final class IntegrationTestSupport {
             legacyClientResult.complete();
         }
 
-        final Promise<Void> newClientResult = Promise.promise();
-        if (applicationClient instanceof AmqpApplicationClient) {
-            ((AmqpApplicationClient) applicationClient).disconnect(newClientResult);
-        } else {
-            newClientResult.complete();
-        }
+        final Future<Void> stopResult = applicationClient.stop();
 
         return CompositeFuture.all(
                 legacyClientResult.future()
                     .onSuccess(ok -> LOGGER.info("legacy client's connection to AMQP Messaging Network closed")),
-                newClientResult.future()
-                    .onSuccess(ok -> LOGGER.info("new client's connection to AMQP Messaging Network closed")))
+                stopResult
+                    .onSuccess(ok -> LOGGER.info("new client's connection to messaging network closed")))
                 .mapEmpty();
     }
 
--- a/tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
@@ -42,13 +42,21 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
+import org.apache.kafka.clients.consumer.ConsumerConfig;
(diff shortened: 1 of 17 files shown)


tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
```diff
@@ -841,9 +903,19 @@ public void deleteObjects(final VertxTestContext ctx) {
...
+            legacyClientResult.complete();
+        }
+
         final Promise<Void> newClientResult = Promise.promise();
-        amqpApplicationClient.disconnect(newClientResult);
+        if (applicationClient instanceof AmqpApplicationClient) {
+            ((AmqpApplicationClient) applicationClient).disconnect(newClientResult);
+        } else {
+            newClientResult.complete();
+        }
```
sophokles73, line 917: looks like `ApplicationClient` is lacking a generic *close()* method to release underlying resources, doesn't it?

tests/src/test/java/org/eclipse/hono/tests/http/TelemetryHttpIT.java
```diff
@@ -199,6 +201,8 @@ public void testUploadQos1MessageFailsIfDeliveryStateNotUpdated(
             final VertxTestContext ctx)
             throws InterruptedException {
 
+        final AmqpApplicationClient amqpApplicationClient = (AmqpApplicationClient) helper.applicationClient;
```
sophokles73, line 204: why is this cast necessary?

fkaltner: Because the test makes verifications on AMQP message specific conditions (by checking the remotely settled flag and the message delivery). Hence it needs to create an AMQP specific telemetry consumer.

sophokles73: Ah, I see. It is further down below in the code that has not been changed. That's why I didn't see it at first in the GitHub diff ...

tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
fkaltner: Hmm. I was about to start implementing this method, but then I noticed that `ApplicationClient` actually implements `Lifecycle#stop`.

The `ProtonBasedApplicationClient` calls `HonoConnection#shutdown` in its `Lifecycle#stop` method. Both, the `shutdown` and the `disconnect` method end up calling `HonoConnectionImpl#closeConnection`.

So I think calling `Lifecycle#stop` actually is already the generic _close()_ method?

sophokles73 (approved):
LGTM

PR #2518, opened 2021-03-02
Fix internal method name in MultiMessagingSenders

b-abel:
Fixes a wrong name of a private method inside the class `MultiMessagingSenders` (it is not only for event senders but also for other senders).


calohmn (approved):
LGTM

PR #2520, opened 2021-03-03
[#2478] Pull up Kafka test helper class for reuse

kaniyan:
Pull up Kafka test helper class so that it can be used for the Kafka based north bound application tests


sophokles73 (approved):
LGTM

PR #2523, opened 2021-03-04
Handle received message delivery getting updated from remote

calohmn:
Use a handler that will be invoked when a AMQP message delivery of a received message gets updated from the remote peer before the local receiver updates the delivery.

A corresponding scenario would be that the remote sender of the message only waits a limited time for the disposition update and afterwards aborts (i.e. releases) the delivery from the sender side. With the handler here, the receiver can get notified of such a delivery update.

### DIFF
--- a/adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
@@ -35,6 +35,7 @@ import org.apache.qpid.proton.amqp.transport.AmqpError;
 import org.apache.qpid.proton.amqp.transport.DeliveryState;
 import org.apache.qpid.proton.amqp.transport.ErrorCondition;
 import org.apache.qpid.proton.amqp.transport.Source;
+import org.apache.qpid.proton.amqp.transport.Target;
 import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.adapter.AbstractProtocolAdapterBase;
 import org.eclipse.hono.adapter.AdapterConnectionsExceededException;
@@ -577,11 +578,11 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
         final Span span = newSpan("attach device sender link", authenticatedDevice, traceSamplingPriority);
         span.log(Map.of("snd-settle-mode", receiver.getRemoteQoS()));
 
-        final String remoteTargetAddress = Optional.ofNullable(receiver.getRemoteTarget())
-                .map(target -> target.getAddress())
+        final String remoteTargetAddress = Optional.ofNullable(receiver.getRemoteTarget()).map(Target::getAddress)
                 .orElse(null);
         if (!Strings.isNullOrEmpty(remoteTargetAddress)) {
-            log.debug("client provided target address [{}] in open frame, closing link", remoteTargetAddress);
+            log.debug("client provided target address [{}] in open frame, closing link [container: {}, {}]",
+                    remoteTargetAddress, conn.getRemoteContainer(), authenticatedDevice);
             span.log(Map.of("target address", remoteTargetAddress));
             final Exception ex = new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST,
                     "container supports anonymous terminus only");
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
```diff
@@ -308,4 +310,21 @@ public static void closeAndFree(
             link.free();
         });
     }
+
+    /**
+     * Sets a handler that will be invoked when the given message delivery of a received message gets updated
+     * from the remote peer before the local receiver updates the delivery.
+     * <p>
+     * A corresponding scenario would be that the remote sender of the message only waits a limited time for
```
sophokles73, line 318: > A scenario where such a handler could be useful would be that ...

core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
```diff
@@ -308,4 +310,21 @@ public static void closeAndFree(
...
+     * the disposition update and afterwards aborts (i.e. releases) the delivery from the sender side. With
+     * the handler here, the receiver can get notified of such a delivery update.
+     *
+     * @param delivery The delivery to set the handler on. Must be a delivery provided by a <em>ProtonReceiver</em> handler.
```
sophokles73, line 322: IMHO it is worth mentioning that the delivery is actually expected to be of type `ProtonDeliveryImpl` ...

adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -600,6 +601,11 @@ protected void handleRemoteReceiverOpen(final ProtonConnection conn, final Proto
                 try {
                     final SpanContext spanContext = TracingHelper.extractSpanContext(tracer, message);
                     final Span msgSpan = newSpan("upload message", authenticatedDevice, traceSamplingPriority, spanContext);
+                    HonoProtonHelper.onReceivedMessageDeliveryUpdatedFromRemote(delivery, d -> {
+                        log.debug("got unexpected disposition update for message received from device [remote state: {}, container: {}, {}]",
+                                delivery.getRemoteState(), conn.getRemoteContainer(), authenticatedDevice);
+                        TracingHelper.logError(msgSpan, "got unexpected disposition from device [remote state: " + delivery.getRemoteState() + "]");
```
sophokles73, line 607: if we consider this to be an error, shouldn't we then also prevent the message from being processed? Otherwise I would opt for not logging this as an error but simply as an information.

client/src/main/java/org/eclipse/hono/client/CommandContext.java
```diff
@@ -175,31 +185,48 @@ public void disposition(final DeliveryState deliveryState) {
...
+                LOG.warn("unexpected delivery state [{}] when settling command message [{}]", deliveryState, getCommand());
+                TracingHelper.logError(span, "unexpected delivery state: " + deliveryState);
+            }
         }
         span.finish();
     }
+
+    private boolean checkDeliveryNotAlreadySettled(final DeliveryState.DeliveryStateType toBeAppliedOutcome) {
+        if (delivery.isSettled() || delivery.remotelySettled()) {
+            final String msg = String.format("cannot apply '%s' outcome to command delivery - delivery already settled %s",
```
sophokles73, line 222: how about
> cannot complete transfer of command from upstream using outcome '%s' - delivery already settled

client/src/main/java/org/eclipse/hono/client/CommandContext.java
```diff
@@ -106,8 +106,10 @@ public ProtonDelivery getDelivery() {
     public void accept() {
         final Span span = getTracingSpan();
         LOG.trace("accepting command message [{}]", getCommand());
-        ProtonHelper.accepted(delivery, true);
-        span.log("accepted command for device");
+        if (checkDeliveryNotAlreadySettled(DeliveryState.DeliveryStateType.Accepted)) {
```
sophokles73, line 109: why shouldn't we settle the message locally if the peer has sent an (unexpected) disposition frame? IMHO we should always settle in order to let the AMQP protocol stack to do its housekeeping, e.g. removing the transfer from the *unsettled* list.

adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
calohmn: Message processing is most likely already in progress at the point the handler is called. 
A way to let that delivery update from the remote peer influence the message processing would be checking the `remotelySettled` flag of the delivery as part of message processing - as done in `ProtonBasedCommandContext`. But that needs to be implemented separately. Therefore having an error logged in the span in any case is something I would prefer here.

sophokles73: But there is no error, is there?

client/src/main/java/org/eclipse/hono/client/CommandContext.java
calohmn: I would suggest
> cannot complete incoming delivery of command message with outcome '%s' - delivery already settled %s

adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
calohmn: Well, ok, not necessarily. I'll just add a span log entry.

client/src/main/java/org/eclipse/hono/client/CommandContext.java
calohmn: I've refactored the CommandContext classes now, making sure the delivery gets settled locally.

sophokles73 (approved):
LGTM

PR #2524, opened 2021-03-04
Ensure no vertx context switch done on resource limit check

calohmn:
Ensure that concurrent resource limit checks are finished on their original vert.x context.


sophokles73:
Out of curiosity: Did you observe any erroneous behavior that this is the cause for?

calohmn:
No, nothing came to my attention there. But I did notice that there were actually cases where a different vert.x context was getting used after the query.

PR #2525, opened 2021-03-04
[#2402] Encapsulate the messaging clients in protocol adapters.

b-abel:
This refactors the handling of messaging clients in the protocol adapters in order to be able to easily support different messing systems while allowing to configure on a tenant which messaging to use (#2402). To support a new messaging system, the protocol adapter code no longer needs to be changed. This should make it easier to add support for RabbitMQ based messaging.

I would be happy if anyone comes up with better names for the classes `MessagingClient` and `MessagingClients`.


b-abel:
@calohmn Could you take a look at this and see if there's any problem from your perspective with adding Command Consumer there as well?

adapter-base/src/main/java/org/eclipse/hono/adapter/MessagingClients.java
```diff
@@ -0,0 +1,156 @@
...
+     *
+     * @param tenant The tenant to select the messaging cleint for.
+     * @return The client that is configured at the tenant or, if this is missing and only one client is set, that one,
+     *         otherwise, the AMQP-based client.
+     * @throws IllegalStateException if no client is set.
+     */
+    public MessagingClient getClientForTenant(final TenantObject tenant) {
+
+        if (clients.isEmpty()) {
+            throw new IllegalStateException("No messaging client present");
```
sophokles73, line 74: If we consider the fact that no clients have been set a misconfiguration, then we should (also) throw an IllegalStateException in the *start* method.

adapter-base/src/main/java/org/eclipse/hono/adapter/MessagingClient.java
```diff
@@ -0,0 +1,160 @@
...
+import org.slf4j.LoggerFactory;
+
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Future;
+import io.vertx.ext.healthchecks.HealthCheckHandler;
+
+/**
+ * A client for messaging used by a protocol adapter.
+ */
+public class MessagingClient implements Lifecycle {
```
sophokles73, line 34: how about naming it `MessagingClientSet`?
and how about making it final?

adapter-base/src/main/java/org/eclipse/hono/adapter/MessagingClients.java
```diff
@@ -0,0 +1,156 @@
...
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.healthchecks.HealthCheckHandler;
+
+/**
+ * The set of messaging clients present in a protocol adapter.
+ * <p>
+ * It contains one client for each connected messaging system and allows to get the messaging to be used for a tenant
+ * based on the tenants configuration.
+ */
+public class MessagingClients implements Lifecycle {
```
sophokles73, line 40: final?

adapter-base/src/main/java/org/eclipse/hono/adapter/MessagingClientSet.java
```diff
@@ -34,12 +34,12 @@
 /**
  * The set of messaging clients present in a protocol adapter.
  * <p>
- * It contains one client for each connected messaging system and allows to get the messaging to be used for a tenant
- * based on the tenants configuration.
+ * It contains one client for each connected messaging system type and allows to get the messaging to be used for a
+ * tenant based on the tenants configuration.
  */
-public class MessagingClients implements Lifecycle {
+public final class MessagingClientSet implements Lifecycle {
```
sophokles73, line 40: FMPOV you renamed the wrong class. IIRC I had proposed to rename `MessageClient` to `MessageClientSet` because it holds a set of clients. `MessageClients` can remain as-is FMPOV ...

b-abel: Oh yes, I see now, your proposal was to rename the other class. It made perfect sense to me :laughing: 
I'll change it.

sophokles73 (approved):
LGTM

PR #2529, opened 2021-03-05
[#2478] Extend north bound client to receive command responses

kaniyan:
This PR is for #2478. It extends the northbound application client to support receiving command responses.


clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaDownstreamMessage.java
```diff
@@ -69,6 +71,8 @@ public KafkaDownstreamMessage(final KafkaConsumerRecord<String, Buffer> record)
         payload = record.value();
         creationTime = getCreationTime(record.headers());
         timeTillDisconnect = getTimeTillDisconnect(record.headers());
+        correlationId = getCorrelationId(record.headers());
+        status = getStatus(record.headers());
```
sophokles73, line 75: FMPOV we should not try to retrieve these values as part of the constructor because by far the most messages will not be command responses and we will simply be waisting CPU and generate useless log entries ...
Instead, we should retrieve the values when the getters are invoked or at least lazily initialize the fields.

clients/application-kafka/src/test/java/org/eclipse/hono/application/client/kafka/impl/KafkaApplicationClientImplTest.java
```diff
@@ -0,0 +1,233 @@
...
+        correlationId = UUID.randomUUID().toString();
+
+        mockConsumer = new MockConsumer<>(OffsetResetStrategy.EARLIEST);
+        kafkaConsumer = KafkaConsumer.create(vertx, mockConsumer);
+
+        consumerConfig = new KafkaConsumerConfigProperties();
+        consumerConfig.setConsumerConfig(Map.of("client.id", "application-test-consumer"));
+        producerConfig.setCommonClientConfig(Map.of("client.id", "application-test-sender"));
+
+        client = spy(new KafkaApplicationClientImpl(vertx, consumerConfig, producerFactory, producerConfig));
```
sophokles73, line 111: why do you need to create a spy here?

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaApplicationClientImpl.java
```diff
@@ -125,7 +125,8 @@ public KafkaApplicationClientImpl(
                 closeHandler);
     }
 
-    private Future<MessageConsumer> createKafkaBasedDownstreamMessageConsumer(
+    // visible for testing
+    Future<MessageConsumer> createKafkaBasedDownstreamMessageConsumer(
```
sophokles73, line 129: why can't you use one of the public factory methods?

kaniyan: I need to use a `MockConsumer` instead of a `KafkaConsumer` for the tests. Whenever the above method is being called, I answer that call using a mocked method which uses a `MockConsumer` to create a `MessageConsumer` as in the below code. For that purpose I also have to create a spy of `KafkaApplicationClientImpl`. 

```
doAnswer(x -> KafkaBasedDownstreamMessageConsumer
        .create(tenantId, type, kafkaConsumer, consumerConfig, msgHandler, closeHandler))
                .when(client)
                .createKafkaBasedDownstreamMessageConsumer(tenantId, type, msgHandler, closeHandler);
```

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaDownstreamMessage.java
kaniyan: I have updated now, so that those values are being retrieved only when the _getters_ are invoked.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaApplicationClientImpl.java
kaniyan: I have changed the tests so that `spy(...)` is not used anymore. Would you mind taking a look again?

clients/application-kafka/src/test/java/org/eclipse/hono/application/client/kafka/impl/KafkaApplicationClientImplTest.java
```diff
@@ -0,0 +1,232 @@
...
+     * Verifies that the message consumer created by the application client receives and processes the
+     * downstream message using the given message handler.
+     *
+     * @param msgType The message type (telemetry, event or command_response)
+     * @param ctx The vert.x test context.
+     */
+    @SuppressWarnings("unchecked")
+    @ParameterizedTest(name = PARAMETERIZED_TEST_NAME_PATTERN)
+    @MethodSource("messageTypes")
+    public void testCreateConsumerAcceptsMessage(final Type msgType, final VertxTestContext ctx) {
```
sophokles73, line 142: I wonder if the test cases in this class are in the right place. It looks like we are actually testing the correct behavior of KafkaBasedDownstreamMessageConsumer or am I mistaken?

In this class, I would expect to find tests for creating a consumer and sending commands etc.

kaniyan: In this PR for command response consumer, I have focused only on the consumers. In the above test, create consumer is also taken care of(eg. error in creating consumer leads to test failure). I think I went ahead and also tested if the messages are received by those consumers. If you think that this part is not needed here, I can remove it.

The next PR is for sending one way commands and then for synchronous command sending. In the consecutive PRs, I will add tests for sending commands.

sophokles73: > If you think that this part is not needed here, I can remove it.

I do. As I said, I would expect this to be tested in *KafkaBasedDownstreamMessageConsumerTest* ...

clients/application-kafka/src/test/java/org/eclipse/hono/application/client/kafka/impl/KafkaApplicationClientImplTest.java
```diff
@@ -0,0 +1,232 @@
...
+
+        consumerConfig = new KafkaConsumerConfigProperties();
+        consumerConfig.setConsumerConfig(Map.of("client.id", "application-test-consumer"));
+        producerConfig.setCommonClientConfig(Map.of("client.id", "application-test-sender"));
+
+        client = new KafkaApplicationClientImpl(vertx, consumerConfig, producerFactory, producerConfig) {
+
+            //Override to use the mocked Kafka consumer
+            @Override
+            Future<MessageConsumer> createKafkaBasedDownstreamMessageConsumer(final String tenantId, final Type type,
```
sophokles73, line 113: If I understand the code correctly, the underlying problem is the hard wired dependency on *KafkaConsumer.create(Vertx, Consumer)* which makes it very hard to test. How about adding something like
*KafkaApplicationClientImpl.setKafkaConsumerFactory(Function<Consumer, KafkaConsumer>)* which can be used by the test to set a function that produces a mocked KafkaConsumer?

kaniyan: Having some factory method is a good idea. I think instead of passing a `Consumer` we can directly supply a `KafkaConsumer` as below.

```   
 KafkaApplicationClientImpl.setKafkaConsumerFactory(final Supplier<KafkaConsumer<String, Buffer>> kafkaConsumerSupplier)
```

kaniyan: I have updated and pushed the changes, please take a look.

clients/application/src/main/java/org/eclipse/hono/application/client/MessageProperties.java
```diff
@@ -29,4 +29,14 @@
...
+    /**
+     * Gets a property value.
+     *
+     * @param name The property name.
+     * @param type The expected value type.
+     * @param <T> The expected type of the property value.
+     * @return The property value or {@code null} if the property is not set or is of an unexpected type.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     */
+    <T> T getProperty(String name, Class<T> type);
```
sophokles73, line 41: IMHO this is a valuable addition as it better encapsulates the internal implementation :+1:

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaApplicationClientImpl.java
```diff
@@ -41,6 +42,7 @@
 
     private final Vertx vertx;
     private final KafkaConsumerConfigProperties consumerConfig;
+    private Supplier<KafkaConsumer<String, Buffer>> kafkaConsumerSupplier;
```
sophokles73, line 45: IMHO this should be a function accepting a String (type) so that you can use a default implementation here that invokes KafkaConsumer.create(vertx, String) ...

kaniyan: For the test we need the method with parameters as below to create a Kafka consumer
`KafkaConsumer.create(vertx, consumer)`

And in the client implementation the below one. 
`KafkaConsumer.create(vertx,  configMap)`

Both are methods with different signature. Sorry I didn't get it what you mean by to use a default implementation that invokes `KafkaConsumer.create(vertx, String)`

kaniyan:
@sophokles73 FMPOV all the comments have been addressed. Is there anything still open?

sophokles73 (approved):
LGTM

PR #2530, opened 2021-03-08
#2469 Making test execution messaging system specific by annotation

fkaltner:
This is a further preparation for migrating the ITs to Kafka.

Some tests do not make sense to be executed when ITs are run with Kafka as the Kafka "protocol" does not offer the same functionality as AMQP 1.0. 

For instance, it is not possible to determine if a consumer has actually processed the message (testUploadMessageFailsForNoConsumer) or there is no delivery state to be updated (testUploadQos1MessageFailsIfDeliveryStateNotUpdated).

The tests for the JMS client of Apache Qpid do also not make sense in a Kafka context as far as I can judge.

The implemented annotation can be used to skip those tests automatically when the Kafka profile is set.


sophokles73 (approved):
LGTM

PR #2533, opened 2021-03-10
[#2478] Extend northbound application client to send one way commands

kaniyan:
This PR extends northbound application client to send one way commands.

### DIFF
--- a/clients/application-amqp/src/test/java/org/eclipse/hono/application/client/amqp/ProtonBasedCommandSenderTest.java
+++ b/clients/application-amqp/src/test/java/org/eclipse/hono/application/client/amqp/ProtonBasedCommandSenderTest.java
@@ -1,0 +1,175 @@
+/**
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.hono.application.client.amqp;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyBoolean;
+import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.qpid.proton.amqp.messaging.Accepted;
+import org.apache.qpid.proton.message.Message;
+import org.eclipse.hono.client.HonoConnection;
+import org.eclipse.hono.client.SendMessageSampler;
+import org.eclipse.hono.client.amqp.test.AmqpClientUnitTestHelper;
+import org.eclipse.hono.test.VertxMockSupport;
+import org.eclipse.hono.util.MessageHelper;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentCaptor;
+
+import io.opentracing.noop.NoopSpan;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.eventbus.EventBus;
+import io.vertx.junit5.Timeout;
+import io.vertx.junit5.VertxExtension;
+import io.vertx.junit5.VertxTestContext;
+import io.vertx.proton.ProtonDelivery;
+import io.vertx.proton.ProtonMessageHandler;
+import io.vertx.proton.ProtonQoS;
+import io.vertx.proton.ProtonReceiver;
+import io.vertx.proton.ProtonSender;
+
+/**
+ * Tests verifying behavior of {@link ProtonBasedCommandSender}.
+ *
+ */
+@ExtendWith(VertxExtension.class)
+@Timeout(value = 5, timeUnit = TimeUnit.SECONDS)
+public class ProtonBasedCommandSenderTest {
+
+    private ProtonBasedCommandSender commandSender;
+    private ProtonSender sender;
+    private ProtonDelivery protonDelivery;
+
+    /**
+     * Sets up the fixture.
+     */
+    @BeforeEach
+    void setUp() {
+        final var vertx = mock(Vertx.class);
+        when(vertx.eventBus()).thenReturn(mock(EventBus.class));
+        final HonoConnection connection = AmqpClientUnitTestHelper.mockHonoConnection(vertx);
+
+        final ProtonReceiver receiver = AmqpClientUnitTestHelper.mockProtonReceiver();
+        when(connection.createReceiver(
+                anyString(),
+                any(ProtonQoS.class),
+                any(ProtonMessageHandler.class),
+                anyInt(),
+                anyBoolean(),
+                VertxMockSupport.anyHandler())).thenReturn(Future.succeededFuture(receiver));
+
+        protonDelivery = mock(ProtonDelivery.class);
+        when(protonDelivery.remotelySettled()).thenReturn(true);
+        when(protonDelivery.getRemoteState()).thenReturn(new Accepted());
+
+        sender = AmqpClientUnitTestHelper.mockProtonSender();
+        when(sender.send(any(Message.class), VertxMockSupport.anyHandler())).thenReturn(protonDelivery);
+        when(connection.createSender(anyString(), any(), any())).thenReturn(Future.succeededFuture(sender));
+
+        commandSender = new ProtonBasedCommandSender(connection, SendMessageSampler.Factory.noop());
+    }
+
+    /**
+     * Verifies that a one-way command sent to a device succeeds.
+     *
+     * @param ctx The vert.x test context.
+     */
+    @Test
+    public void testSendOneWayCommandSucceeds(final VertxTestContext ctx) {
+        final String tenantId = UUID.randomUUID().toString();
+        final String deviceId = UUID.randomUUID().toString();
+        final String subject = "setVolume";
+        final Map<String, Object> applicationProperties = Map.of("appKey", "appValue");
+        final ArgumentCaptor<Handler<ProtonDelivery>> dispositionHandlerCaptor = VertxMockSupport.argumentCaptorHandler();
+        final ArgumentCaptor<Message> messageCaptor = ArgumentCaptor.forClass(Message.class);
+
+        // WHEN sending a one-way command with some application properties and payload
+        final Future<Void> sendCommandFuture = commandSender
+                .sendOneWayCommand(tenantId, deviceId, subject, null, Buffer.buffer("{\"value\": 20}"),
+                        applicationProperties, NoopSpan.INSTANCE.context())
+                .onComplete(ctx.completing());
+
+        // VERIFY that the command is being sent
+        verify(sender).send(messageCaptor.capture(), dispositionHandlerCaptor.capture());
+
+        // VERIFY that the future waits for the disposition to be updated by the peer
+        assertThat(sendCommandFuture.isComplete()).isFalse();
+
+        // THEN the disposition is updated and the peer accepts the message
+        dispositionHandlerCaptor.getValue().handle(protonDelivery);
+
+        //VERIFY if the message properties are properly set
+        final Message message = messageCaptor.getValue();
+        assertThat(MessageHelper.getDeviceId(message)).isEqualTo(deviceId);
+        assertThat(message.getSubject()).isEqualTo(subject);
+        assertThat(MessageHelper.getApplicationProperty(message.getApplicationProperties(), "appKey", String.class))
+                .isEqualTo("appValue");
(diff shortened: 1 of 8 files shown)


clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/AmqpBasedCommandSender.java
```diff
@@ -0,0 +1,66 @@
...
+     * @return A future indicating the result of the operation:
+     *         <p>
+     *         If the one-way command was accepted, the future will succeed.
+     *         <p>
+     *         The future will fail with a {@link ServiceInvocationException} if the one-way command could 
+     *         not be forwarded to the device.
+     * @throws NullPointerException if any of tenantId, deviceId or command are {@code null}.
+     * @throws IllegalArgumentException if the sendMessageTimeout value is &lt; 0
+     */
+    Future<Void> sendOneWayCommand(
```
sophokles73, line 57: is there anything AMQP specific about this method? If not, what do we need this interface for?

kaniyan: This method has an extra parameter `sendMessageTimeout` and currently this method is used only in the IT test in the context of AMQP (explicitly `AmqpApplicationClient` is being used in the `IntegrationTestSupport` as below). Hence I added it to the AMQP specific interface to support that test.

```
amqpApplicationClient.sendOneWayCommand(tenantId, deviceId, command, contentType, payload, properties,
                requestTimeout, NoopSpan.INSTANCE.context())
```

sophokles73: That I understood/saw. The question is: why does this need to be in an AMQP specific interface when there is nothing AMQP specific about it. Why can't we simply add it to `CommandSender`?

kaniyan: Even though the `sendMessageTimeout` is not AMQP specific and since the above method is only being used in tests (involving AMQP), I didn't pull up to the `CommandSender`.

We can simply add it to the `CommandSender`. But I doubt that all the underlying implementations can support it. For example, in case of Kafka we won't be able to use the `sendMessageTimeOut` at each transaction level. Kafka offers timeout properties which can be set at the producer configuration. Kafka normally works by collecting messages and sending them in batches and there is no provision in vertx kafka client API to achieve this timeout for each message transaction. In that case, we can add a note to the _JavaDoc_ saying that this timeout may be used if the underlying implementation supports it. FMPOV adding such disclaimers reduces clarity in understanding how a API would behave with various implementations.

sophokles73: > But I doubt that all the underlying implementations can support it.

FMPOV this is not about taking advantage of the underlying transport's capabilities but simply about providing the client the opportunity to limit the time to wait for the completion of the returned Future. Otherwise, the client would need to implement the timeout logic itself. It doesn't matter whether AMQP sends each message individually or Kafka sends it in batches. If the client is not willing to wait more than 1 second for the Future to complete, then it will probably resend the message once the second has passed. If that means that a client is never able to (successfully) send a message via Kafka, then either the Kafka producer config needs to be adapted or the client needs to increase the timeout. The same is true for AMQP, i.e. if the underlying messaging network takes more than a second to forward a message from the client to the device, then the client will run into the timeout.
Again, FMPOV there is nothing transport specific about this method. However, the Kafka transport can not determine if the command could be forwarded to the device because it writes the command to the message log instead of forwarding it to the protocol adapter.

kaniyan: > FMPOV this is not about taking advantage of the underlying transport's capabilities but simply about providing the client the opportunity to limit the time to wait for the completion of the returned Future. Otherwise, the client would need to implement the timeout logic itself.

The client need not implement the timeout logic itself. The client already has this flexibility to limit the time to wait for completion but at the `ApplicationClient` level. In case of AMQP it is the `sendMessageTimeout` at the connection config. In case of Kafka it can be set using the producer config property `delivery.timeout.ms`. FMPOV if a user wants to change the existing value, then he/she can create a new client instance with the new timeout value.

> FMPOV there is nothing transport specific about this method

I agree that there is nothing transport specific. I can move it to the `CommandSender`. 

But in case of Kafka, once `KafkaProducer.send(...)` is invoked, currently vertx Kafka client provides no way to abort that transaction. We can finish the promise, but it doesn't mean that Kafka producer won't forward that message so that the user can resend it again. The `delivery.timeout.ms` can be set only when we create the _producer_.  In our implementation we reuse the cached producers. Lets say for this method implementation that we create a new Kafka Producer with the given timeout value. But still it is not as simple as changing the value of `delivery.timeout.ms`. Because this value has some prerequisite conditions with other timeout properties namely `request.timeout.ms` and `linger.ms`. I think the user should deal with the Kafka properties directly and we providing an abstraction doesn't make sense to me due to dependeny of one property with that of the others. Since there is already a provision to set the timeout value at the `ApplicationClient` level, IMHO I am not fully convinced that we should also provide on a more granular level in methods.

kaniyan: Previously my focus was mainly to ensure that the command is not being sent to the device in case timeout happens. I see  that the interest behind is to protect the client and the client need not wait more than the specified timeout value. It means that there is no guarantee that the command will not be sent to the device, when a request is timed out . I will pull up the method to the `CommandSender` and make use of a timer to implement it.

sophokles73: > Since there is already a provision to set the timeout value at the ApplicationClient level, IMHO I am not fully convinced that we should also provide on a more granular level in methods.

Sounds reasonable to me. So FMPOV we should probably simply omit this variant of the method altogether. WDYT?

kaniyan: Sounds good to me. I will omit this variant. And also I will update the IT test support to with a _vertx_ timer solution so that this method won't be used there anymore.

kaniyan: I have updated and pushed the changes. Would you mind taking a look again?

clients/application-amqp/src/test/java/org/eclipse/hono/application/client/amqp/ProtonBasedApplicationClientTest.java
```diff
@@ -68,8 +74,14 @@
 @Timeout(value = 5, timeUnit = TimeUnit.SECONDS)
 class ProtonBasedApplicationClientTest {
 
+    @SuppressWarnings("unchecked")
+    private final ArgumentCaptor<Handler<ProtonDelivery>> dispositionHandlerCaptor = ArgumentCaptor
+            .forClass(Handler.class);
```
sophokles73, line 79: org.eclipse.hono.test.VertxMockSupport.argumentCaptorHandler() ?

clients/application-amqp/src/test/java/org/eclipse/hono/application/client/amqp/ProtonBasedApplicationClientTest.java
```diff
@@ -262,4 +282,77 @@ void testCreateEventConsumerAcceptsMessage(final VertxTestContext ctx) {
...
+
+        // THEN the disposition is updated and the peer accepts the message
+        dispositionHandlerCaptor.getValue().handle(protonDelivery);
+
+        //VERIFY if the message properties are properly set
+        final Message message = messageCaptor.getValue();
+        assertThat(MessageHelper.getDeviceId(message)).isEqualTo(deviceId);
+        assertThat(message.getSubject()).isEqualTo(subject);
+        assertThat(MessageHelper.getApplicationProperty(message.getApplicationProperties(), "appKey", String.class))
+                .isEqualTo("appValue");
```
sophokles73, line 317: add
```
assertThat(sendCommandFuture.isComplete()).isTrue();
```
?

clients/application-amqp/src/test/java/org/eclipse/hono/application/client/amqp/ProtonBasedApplicationClientTest.java
```diff
@@ -262,4 +282,77 @@ void testCreateEventConsumerAcceptsMessage(final VertxTestContext ctx) {
...
+        // THEN the disposition is updated and the peer accepts the message
+        dispositionHandlerCaptor.getValue().handle(protonDelivery);
+
+        //VERIFY if the message properties are properly set
+        final Message message = messageCaptor.getValue();
+        assertThat(MessageHelper.getDeviceId(message)).isEqualTo(deviceId);
+        assertThat(message.getSubject()).isEqualTo(subject);
+        assertThat(MessageHelper.getCorrelationId(message)).isEqualTo(correlationId);
+        assertThat(MessageHelper.getApplicationProperty(message.getApplicationProperties(), "appKey", String.class))
+                .isEqualTo("appValue");
```
sophokles73, line 355: add
```
assertThat(sendCommandFuture.isComplete()).isTrue();
```
?

clients/application-kafka/src/test/java/org/eclipse/hono/application/client/kafka/impl/KafkaApplicationClientImplTest.java
```diff
@@ -136,6 +147,63 @@ public void testCloseConsumer(final Type msgType, final VertxTestContext ctx) {
                 })));
     }
 
+    /**
+     * Verifies that a command sent asynchronously to a device succeeds.
+     *
+     * @param ctx The vert.x test context.
+     */
+    @Test
+    public void testSendAsyncCommandSucceeds(final VertxTestContext ctx) {
```
sophokles73, line 156: shoudn't these tests better be in *KafkaBasedCommandSenderTest*?

kaniyan: I have moved these tests to the `KafkaBasedCommandSenderTest` and also the tests for the proton based implementation to the `ProtonBasedCommandSenderTest`.

clients/application-amqp/src/test/java/org/eclipse/hono/application/client/amqp/ProtonBasedCommandSenderTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+        protonDelivery = mock(ProtonDelivery.class);
+        when(protonDelivery.remotelySettled()).thenReturn(true);
+        when(protonDelivery.getRemoteState()).thenReturn(new Accepted());
+
+        sender = AmqpClientUnitTestHelper.mockProtonSender();
+        when(sender.send(any(Message.class), VertxMockSupport.anyHandler())).thenReturn(protonDelivery);
+        when(connection.createSender(anyString(), any(), any())).thenReturn(Future.succeededFuture(sender));
+
+        client = new ProtonBasedApplicationClient(connection);
```
sophokles73, line 94: this should be a new *ProtonBasedCommandSender*, shouldn't it?

clients/application-amqp/src/test/java/org/eclipse/hono/application/client/amqp/ProtonBasedCommandSenderTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Tests verifying behavior of {@link ProtonBasedCommandSender}.
+ *
+ */
+@ExtendWith(VertxExtension.class)
+@Timeout(value = 5, timeUnit = TimeUnit.SECONDS)
+public class ProtonBasedCommandSenderTest {
+
+    private final ArgumentCaptor<Handler<ProtonDelivery>> dispositionHandlerCaptor = VertxMockSupport.argumentCaptorHandler();
```
sophokles73, line 62: you might indeed be able to reuse these across tests, but since it is not used in the set up method, these captors should be created in the particular test cases FMPOV ...

clients/application-kafka/src/test/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSenderTest.java
```diff
@@ -0,0 +1,139 @@
...
+        consumerConfig = new KafkaConsumerConfigProperties();
+        consumerConfig.setConsumerConfig(Map.of("client.id", "application-test-consumer"));
+        mockConsumer = new MockConsumer<>(OffsetResetStrategy.EARLIEST);
+
+        producerConfig = new KafkaProducerConfigProperties();
+        producerConfig.setProducerConfig(Map.of("client.id", "application-test-sender"));
+        mockProducer = KafkaClientUnitTestHelper.newMockProducer(true);
+        producerFactory = KafkaClientUnitTestHelper.newProducerFactory(mockProducer);
+
+        client = new KafkaApplicationClientImpl(vertx, consumerConfig, producerFactory, producerConfig);
```
sophokles73, line 78: KafkaBasedCommandSender?

clients/application-amqp/src/test/java/org/eclipse/hono/application/client/amqp/ProtonBasedCommandSenderTest.java
kaniyan: My bad, I just moved those tests to the new class and missed to replace it with `ProtonBasedCommandSender`. Thanks for pointing out and I fixed it now.

sophokles73: No worries, that's what the review is there for ;-)

sophokles73 (approved):
LGTM

Issue #2475, opened 2021-02-12
Make Kafka connection of device registries configurable


fkaltner:
Like already done for AbstractAdapterConfig: use Kafka if config properties are set, use AMQP otherwise.

sophokles73:
@fkaltner can this be closed?

PR #2534, opened 2021-03-11
#2475 Migrating device registries to new messaging client

fkaltner:
This is for #2475

### DIFF
--- a/adapter-base-spring/src/main/java/org/eclipse/hono/adapter/spring/AbstractMessagingClientConfig.java
+++ b/adapter-base-spring/src/main/java/org/eclipse/hono/adapter/spring/AbstractMessagingClientConfig.java
@@ -1,0 +1,261 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.hono.adapter.spring;
+
+import java.util.Optional;
+
+import org.eclipse.hono.adapter.AdapterConfigurationSupport;
+import org.eclipse.hono.adapter.MessagingClientSet;
+import org.eclipse.hono.adapter.MessagingClients;
+import org.eclipse.hono.adapter.client.command.CommandResponseSender;
+import org.eclipse.hono.adapter.client.command.amqp.ProtonBasedCommandResponseSender;
+import org.eclipse.hono.adapter.client.command.kafka.KafkaBasedCommandResponseSender;
+import org.eclipse.hono.adapter.client.telemetry.EventSender;
+import org.eclipse.hono.adapter.client.telemetry.TelemetrySender;
+import org.eclipse.hono.adapter.client.telemetry.amqp.ProtonBasedDownstreamSender;
+import org.eclipse.hono.adapter.client.telemetry.kafka.KafkaBasedEventSender;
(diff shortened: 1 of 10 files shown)


adapter-base-spring/src/main/java/org/eclipse/hono/adapter/spring/AbstractMessagingClientConfig.java
```diff
@@ -0,0 +1,255 @@
...
+import io.opentracing.Tracer;
+import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
+
+
+/**
+ * A base class that provides helper methods for configuring messaging clients.
+ */
+@Configuration
+public abstract class AbstractMessagingClientConfig extends AdapterConfigurationSupport {
```
b-abel, line 53: The naming sounds a bit strange to me. My understanding is that `AbstractMessagingClientConfig` should not only be used in adapters but also in other services and therefore should not extend a class with "Adapter" in the name. Accordingly, the name of the` AdapterConfigurationSupport#getAdapterName()` method does not fit anymore.

fkaltner: Yes, I also thought about changing this but was reluctant as this will add more changes to this PR. Hence I wanted to wait upon other opinions on that and change it in a separate PR.

It is also strange to me that it is an abstract class an not an interface...

adapter-base-spring/src/main/java/org/eclipse/hono/adapter/spring/AbstractMessagingClientConfig.java
```diff
@@ -0,0 +1,255 @@
...
+     * Creates new messaging clients according to the configuration in use.
+     *
+     * @param samplerFactory The sampler factory to use.
+     * @param tracer The tracer instance.
+     * @param vertx The Vert.x instance to use.
+     * @param adapterProperties The adapter's configuration properties.
+     *
+     * @return The created messaging clients.
+     */
+    protected MessagingClients messagingClients(final SendMessageSampler.Factory samplerFactory,
```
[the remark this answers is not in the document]
fkaltner: Makes sense.

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedServiceConfig.java
```diff
@@ -226,7 +179,7 @@ public FileBasedDeviceBackend deviceBackend() {
         autoProvisioner.setVertx(vertx);
         autoProvisioner.setTracer(tracer);
         autoProvisioner.setTenantInformationService(tenantInformationService);
-        autoProvisioner.setEventSender(eventSender());
+        autoProvisioner.setMessagingClients(messagingClients(SendMessageSampler.Factory.noop(), tracer, vertx, new ProtocolAdapterProperties()));
```
sophokles73, line 182: passing in a *ProtocolAdapterProperties* instance from within a non-adapter component looks very, very strange

sophokles73 (approved):
Based on the current implementation and package structuring of the *clients* modules, this looks good.
IMHO it was a mistake to group/package the registry service clients according to the client side role that would be using them, in this case *protocol adapters*. It turns out that clients for accessing registry services are used by other components as well. IMHO we should therefore refactor the adapter *clients* modules to no longer make any assumptions regarding what type of component would actually use them. But that is, of course, outside the scope of this PR.

fkaltner:
> Based on the current implementation and package structuring of the _clients_ modules, this looks good.
> IMHO it was a mistake to group/package the registry service clients according to the client side role that would be using them, in this case _protocol adapters_. It turns out that clients for accessing registry services are used by other components as well. IMHO we should therefore refactor the adapter _clients_ modules to no longer make any assumptions regarding what type of component would actually use them. But that is, of course, outside the scope of this PR.

Yes, that sounds good. I created https://github.com/eclipse/hono/issues/2540, so that we don't forget about this.

PR #2539, opened 2021-03-12
[#2064] Close MQTT connection on terminal errors

kaniyan:
This PR is to close the MQTT connection when one of the following terminal errors occur.

* The adapter is disabled for the tenant that the client belongs to.
* The device is disabled or not registered (provided that the device is not connected via a gateway).
* The gateway is disabled or not registered.
* The tenant is disabled or does not exist.

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
@@ -37,6 +37,9 @@ import org.eclipse.hono.adapter.AbstractProtocolAdapterBase;
 import org.eclipse.hono.adapter.AdapterConnectionsExceededException;
 import org.eclipse.hono.adapter.AdapterDisabledException;
 import org.eclipse.hono.adapter.AuthorizationException;
+import org.eclipse.hono.adapter.DeviceDisabledOrNotRegisteredException;
+import org.eclipse.hono.adapter.GatewayDisabledOrNotRegisteredException;
+import org.eclipse.hono.adapter.TenantDisabledOrNotRegisteredException;
 import org.eclipse.hono.adapter.auth.device.AuthHandler;
 import org.eclipse.hono.adapter.auth.device.ChainAuthHandler;
 import org.eclipse.hono.adapter.auth.device.CredentialsApiAuthProvider;
@@ -536,13 +539,9 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends MqttProtoc
         return authAttempt
                 .compose(authenticatedDevice -> CompositeFuture.all(
                         getTenantConfiguration(authenticatedDevice.getTenantId(), currentSpan.context())
-                            .compose(tenantObj -> CompositeFuture.all(
-                                    isAdapterEnabled(tenantObj).recover(t -> Future.failedFuture(
-                                            new AdapterDisabledException(
-                                                    authenticatedDevice.getTenantId(),
-                                                    "adapter is disabled for tenant",
-                                                    t))),
-                                    checkConnectionLimit(tenantObj, currentSpan.context()))),
+                                .compose(tenantObj -> CompositeFuture.all(
+                                        isAdapterEnabled(tenantObj),
+                                        checkConnectionLimit(tenantObj, currentSpan.context()))),
                         checkDeviceRegistration(authenticatedDevice, currentSpan.context()))
                         .map(authenticatedDevice))
                 .map(authenticatedDevice -> {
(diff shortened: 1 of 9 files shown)


site/documentation/content/user-guide/mqtt-adapter.md
```diff
@@ -621,7 +621,17 @@ Example payload:
...
 
 If no error subscription is in place and no *on-error* parameter was set, the default error handling behaviour is to close the MQTT connection to the device.
-If the device has a subscription on the error topic (on the same MQTT connection the device uses for sending messages), the default behaviour is to keep the MQTT connection open. 
+If the device has a subscription on the error topic (on the same MQTT connection the device uses for sending messages), the default behaviour is to keep the 
+MQTT connection open unless a terminal error happens. The errors that are classified as terminal are listed below.
+
+* The adapter is disabled for the tenant that the client belongs to.
+* The device is disabled or not registered (provided that the device is not connected via a gateway).
+* The gateway is disabled or not registered.
```
calohmn, line 629: I think replacing the above 2 entries with
`The authenticated device or gateway is disabled or not registered.`
would be easier to understand.
(I think in case an *unauthenticated* device has sent a message, the connection shouldn't be disconnected anyway - see my other comment.)

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
```diff
@@ -1106,48 +1105,74 @@ protected final void handlePublishedMessage(final MqttPublishMessage message) {
...
+                if (Objects.nonNull(authenticatedDevice) && !authenticatedDevice.getDeviceId().equals(deviceId)) {
+                    return getRegistrationAssertion(tenantId, authenticatedDevice.getDeviceId(), null, spanContext)
+                            .map(ok -> false)
+                            .recover(e -> {
+                                // if the gateway is not registered then it is a terminal error
+                                return Future.succeededFuture(e instanceof DeviceRegistrationException);
+                            });
+                }
+                // if the device is not connected via a gateway
+                return Future.succeededFuture(true);
```
calohmn, line 1169: (The `if` in the comment is a bit confusing.)
I think the return statement should be changed to
````
return Future.succeededFuture(authenticatedDevice != null);
````
An *unauthenticated* client (which could potentially use a different deviceId for every telemetry/event message it publishes) should not be disconnected, FMPOV.

Comment (kaniyan) [reply adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java:1169]: IIRC anonymous devices are not supported. An unauthenticated device still should be registered, isn't it?  I see from the code that we don't allow unauthenticated devices that are not registered to send messages.

File: site/documentation/content/user-guide/mqtt-adapter.md (hunk 1 above)
Comment (kaniyan) [reply site/documentation/content/user-guide/mqtt-adapter.md:629]: The unauthenticated devices are not disconnected unless they are disabled or not registered, see the below comment.

File: adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java (hunk 2 above)
Comment (calohmn) [reply adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java:1169]: FMPOV, the client that sent the MQTT CONNECT packet *without* authentication data doesn't necessarily represent just one device. The connected client will provide the tenant/device id in the topic path for each message that it publishes.
Therefore the client could be used for different devices, similar to a gateway publishing messages on behalf of different devices. That's why I wouldn't close the MQTT connection in that case.

Comment (kaniyan) [reply adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java:1169]: > FMPOV, the client that sent the MQTT CONNECT packet _without_ authentication data doesn't necessarily represent just one device.

Do you mean that it's an gateway? If so, then it should be registered, isn't it?


>  The connected client will provide the tenant/device id in the topic path for each message that it publishes.

In that case, the connected client (i.e gateway) should be registered, isn't it?

Comment (kaniyan) [reply adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java]: I have updated and pushed the commit. Would you mind taking a look again?

File: adapter-base/src/main/java/org/eclipse/hono/adapter/exception/DeviceDisabledOrNotRegisteredException.java [hunk 3]
```diff
@@ -0,0 +1,37 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.eclipse.hono.adapter.exception;
```
calohmn, line 13: Regarding the extra `exception` package:
In the parent `org.eclipse.hono.adapter` package, there are already some exception classes (`AdapterConnectionsExceededException`, `ConnectionDurationExceededException`, etc.).
I think it would be better to move the new exception classes to the `org.eclipse.hono.adapter` package.

adapter-base/src/main/java/org/eclipse/hono/adapter/exception/DeviceDisabledOrNotRegisteredException.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.eclipse.hono.client.ClientErrorException;
+
+/**
+ * An exception indicating that the device is either not registered or disabled.
+ */
+public class DeviceDisabledOrNotRegisteredException extends ClientErrorException {
```
calohmn, line 20: There is already a `RegistrationAssertionException` class in the parent package with a similar purpose.

kaniyan: `RegistrationAssertionException` extends `AuthorizationException` and it is used in the AMQP and MQTT adapters while a device makes a connection request (`AbstractProtocolAdapterBase.checkDeviceRegistration`). As the parent class indicates, the error code is always `HTTP_UNAUTHORIZED`. 

In our case before processing messages, `AbstractProtocolAdapterBase.getRegistrationAssertion` is being invoked and the expected error code are `HTTP_NOT_FOUND` or `HTTP_FORBIDDEN`. Also I wanted to differentiate the device from the gateway. Hence I have created classes `DeviceDisabledOrNotRegisteredException` and `GatewayDisabledOrNotRegisteredException`

As I responded in the other comment, I would like to move all the exception classes under adapter module to the _exception_ package and if it is feasible and makes sense, I would replace the `RegistrationAssertionException` with the new ones in similar to the `AdapterDisabledException`.

adapter-base/src/main/java/org/eclipse/hono/adapter/exception/DeviceDisabledOrNotRegisteredException.java
kaniyan: My idea is to move the other exceptions in a separate PR to the _exception_ package and I didn't do that in this PR as I tried to keep this PR short. IMHO it is better to group all of them under `exception` package.

calohmn: Grouping exceptions in an `exception` package seems rather unusual. I think they should better be put in the package that corresponds to the context in which they are usually used in or here the general `org.eclipse.hono.adapter` package.

kaniyan: I have moved them to the `org.eclipse.hono.adapter` package.

adapter-base/src/main/java/org/eclipse/hono/adapter/exception/DeviceDisabledOrNotRegisteredException.java
calohmn: Ok, so `RegistrationAssertionException` should be replaced/adapted afterwards then.

PR #2541, opened 2021-03-15
#2469 Adapting ITs to tenant-based messaging configuration and includ…

fkaltner:
This PR adapts the ITs to the recent changes:

* The ITs now create tenants with a proper "ext" field enabling Kafka, if the Kafka Maven profile is set
* The application contexts of the ITs registries were extended by a Kafka profile, similar to the protocol adapters

By this the ITs can be executed using Kafka but will still fail. However one can observe the messages flowing through Kafka again.


tests/src/test/resources/deviceregistry-mongodb/application.yml
```diff
@@ -56,3 +56,13 @@ management:
...
+
+---
+
+spring:
+  profiles: kafka
+hono:
+  kafka:
+    producerConfig:
+      bootstrap:
+        servers: ${hono.kafka.bootstrap.servers}
```
sophokles73, line 68: missing EOL

tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
```diff
@@ -722,11 +725,20 @@ public static long getSendCommandTimeout() {
...
+                    IntegrationTestSupport.HONO_DEVICEREGISTRY_HTTP_PORT,
+                    Maps.of(TenantConstants.FIELD_EXT_MESSAGING_TYPE, (Object) MessagingType.kafka.name())
+                        .build()
+            );
+        } else {
+            registry = new DeviceRegistryHttpClient(
+                    vertx,
+                    IntegrationTestSupport.HONO_DEVICEREGISTRY_HOST,
+                    IntegrationTestSupport.HONO_DEVICEREGISTRY_HTTP_PORT);
+        }
```
sophokles73, line 741: how about simply always adding the messaging type?

tests/src/test/java/org/eclipse/hono/tests/registry/TenantManagementIT.java
```diff
@@ -455,6 +455,8 @@ public void testGetTenantSucceeds(final VertxTestContext context)  {
             .compose(ar -> getHelper().registry.getTenant(tenantId))
             .onComplete(context.succeeding(httpResponse -> {
                 final JsonObject json = httpResponse.bodyAsJsonObject();
+                // ignore messaging type extension as it is set dynamically by generic test logic
+                json.getJsonObject(TenantConstants.FIELD_EXT).remove(TenantConstants.FIELD_EXT_MESSAGING_TYPE);
```
sophokles73, line 459: why do we need to remove the messaging type? In fact, we should actually assert that it has a particular value, shouldn't we?

fkaltner: This test focuses on tenant specific operations and verifies that the tenant is equal to the JSON which is used as a fixture in the test cases.

The fixture tenant JSON is enriched "under the hood" by _DeviceRegistryHttpClient_. Hence the returned JSON can never be equal the fixture JSON. 

The only other option I see would be to add the the corresponding _messaging_type_ field to _ext_ in _buildTenantPayload_  (which is then overwritten with the same value by _DeviceRegistryHttpClient_). As this involves duplicating code I decided to remove it from the returned JSON.

(The verification of the _messaging_type_ field is done implicitly as all Kafka tests would fail if it wasn't set properly)

tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
```diff
@@ -728,7 +743,9 @@ public void initRegistryClient() {
         registry = new DeviceRegistryHttpClient(
                 vertx,
                 IntegrationTestSupport.HONO_DEVICEREGISTRY_HOST,
-                IntegrationTestSupport.HONO_DEVICEREGISTRY_HTTP_PORT);
+                IntegrationTestSupport.HONO_DEVICEREGISTRY_HTTP_PORT,
+                Collections.singletonMap(TenantConstants.FIELD_EXT_MESSAGING_TYPE, getConfiguredMessagingType().name())
```
sophokles73, line 747: Map.of() ?

tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
```diff
@@ -542,6 +545,18 @@ private static ClientConfigProperties getClientConfigProperties(
         }
     }
 
+    private MessagingType getConfiguredMessagingType() {
+        if (getConfiguredApplicationClientType().equals(KafkaApplicationClient.class)) {
```
sophokles73, line 549: IMHO we should completely replace the *getConfiguredApplicationClientType* method with this method ...

tests/src/test/java/org/eclipse/hono/tests/registry/TenantManagementIT.java
sophokles73: The Tenant service implementation is always invoked via AMQP and it is expected to return the configured *messaging_type* extension field, isn't it? So FMPOV we should make sure (i.e. assert) that it does.

sophokles73: > The fixture tenant JSON is enriched "under the hood" by DeviceRegistryHttpClient. Hence the returned JSON can never be equal the fixture JSON.

IMHO this actually makes it hard(er) to understand the integration tests ...

fkaltner: > The Tenant service implementation is always invoked via AMQP and it is expected to return the configured messaging_type extension field, isn't it? So FMPOV we should make sure (i.e. assert) that it does.

I am not sure if I got you right. `TenantManagementIT` calls `IntegrationTestSupport.registry` (via `getHelper().registry`) which is of type `DeviceRegistryHttpClient`. Also the JavaDoc of states it is making calls to the Tenant management's HTTP endpoint.

Do I miss something obvious? Or are you talking about the protocol adapters? If so shouldn't we add the tests to `TenantAmqpIT` then?

fkaltner: > IMHO this actually makes it hard(er) to understand the integration tests ...

Yes. It was a trade-off in my opinion. I looked into making this more explicit, but then many test cases which create a new tenant need to be adapted to add the corresponding field to the tenant's `ext` field which would add a lot of boilerplate.

The idea was that `IntegrationTestSupport` takes care of setting up the messaging system specific stuff, so that test cases can be focused on the concrete scenario and code does not need to be duplicated.

tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
fkaltner: Yes, that was already on my list as well. I refactored this in a separate PR (https://github.com/eclipse/hono/pull/2544) on which I'll rebase this one as soon as it is merged.

tests/src/test/java/org/eclipse/hono/tests/registry/TenantManagementIT.java
sophokles73: > Do I miss something obvious? Or are you talking about the protocol adapters? If so shouldn't we add the tests to TenantAmqpIT then?

I'd say we should verify the returned result for both the Registry Management API's Tenant endpoint and the Tenant API.

tests/src/test/resources/deviceregistry-jdbc/application.yml
```diff
@@ -77,3 +77,13 @@ management:
...
 server:
   port: -1
+
+---
+
+spring:
+  profiles: kafka
+hono:
+  kafka:
+    producerConfig:
```
sophokles73, line 87: shouldn't this be part of *commonClientConfig*?

tests/src/test/java/org/eclipse/hono/tests/registry/TenantManagementIT.java
fkaltner: Done!

tests/src/test/resources/deviceregistry-jdbc/application.yml
fkaltner: It could be. But this makes only sense to me if this component had a producer and a consumer, so that one doesn't need to configure `bootstrap.servers` twice. This way it's clear that the component is a producer only.

sophokles73: I see

sophokles73 (approved):
LGTM

Issue #2521, opened 2021-03-04
Disconnecting an AMQP connection doesn't abort ongoing connection attempt


calohmn:
Calling `honoConnection.disconnect` on a HonoConnection, where a connection attempt is currently in progress, doesn't lead to the connection attempt to be aborted.
This means the HonoConnection will possibly be connected after that although the invoking code doesn't expect it to be and although the HonoConnection object possibly isn't being worked with anymore.

For the Hono Integration tests this could mean that unused connections get created that aren't getting disconnected and thereby fill up the Qdrouter `maxConnectionsPerHost` limit defined in the `qdrouterd-with-broker.json`.

calohmn:
This exception may trigger #2543.

PR #2546, opened 2021-03-16
HonoConnectionImpl: Remove unused code, prevent duplicate disconnect handler invocation

calohmn:
This removes the unused 'disconnectHandler' param of the private connect() method.
Also makes sure that a disconnect doesn't cause the notifyDisconnectHandlers() method to be invoked twice if no reconnect attempt shall be done.

Preparation for  #2521.


sophokles73 (approved):
LGTM

PR #2548, opened 2021-03-17
[#2521] Let disconnect() abort an ongoing connection attempt

### LINKED ISSUE #2521: Disconnecting an AMQP connection doesn't abort ongoing connection attempt

calohmn:
This fixes #2521:
The connect() method has been refactored so that a connection attempt along with potential retries is handled in a separate inner class named "ConnectionAttempt". The old implementation allowed for separate connect() invocations to potentially lead to separate, concurrent retry loops (if invoked while the other attempt was in the reconnectDelay timer). This is not possible any more. With the encapsulation in the ConnectionAttempt class, an attempt can now be cancelled, aborting a potentially ongoing reconnectDelay timer. This is used in the disconnect() and shutdown() methods.

### DIFF
--- a/client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
+++ b/client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
@@ -25,6 +25,7 @@ import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
 import javax.net.ssl.SSLException;
 import javax.security.sasl.AuthenticationException;
@@ -108,17 +109,16 @@ public class HonoConnectionImpl implements HonoConnection {
     private final List<DisconnectListener<HonoConnection>> disconnectListeners = new ArrayList<>();
     private final List<DisconnectListener<HonoConnection>> oneTimeDisconnectListeners = Collections.synchronizedList(new ArrayList<>());
     private final List<ReconnectListener<HonoConnection>> reconnectListeners = new ArrayList<>();
-    private final AtomicBoolean connecting = new AtomicBoolean(false);
     private final AtomicBoolean shuttingDown = new AtomicBoolean(false);
     private final AtomicBoolean disconnecting = new AtomicBoolean(false);
     private final ConnectionFactory connectionFactory;
     private final Object connectionLock = new Object();
-    private final AtomicInteger connectAttempts = new AtomicInteger(0);
+    private final AtomicReference<ConnectionAttempt> currentConnectionAttempt = new AtomicReference<>();
 
     private final String containerId;
     private final DeferredConnectionCheckHandler deferredConnectionCheckHandler;
 
-    private ProtonClientOptions clientOptions;
+    private ProtonClientOptions lastUsedClientOptions;
     private List<Symbol> offeredCapabilities = Collections.emptyList();
     private Tracer tracer = NoopTracerFactory.create();
     private ProtonSession session;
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -1155,23 +1013,199 @@ private int getCloseConnectionTimeout() {
...
+         *
+         * @param isReconnect if {@code true}, the connection attempt is started after the configured minimum
+         *            reconnect delay.
+         * @return {@code true} if the attempt was started, {@code false} if there already is an ongoing attempt.
+         */
+        public boolean start(final boolean isReconnect) {
+            if (!currentConnectionAttempt.compareAndSet(null, this)) {
+                // there already is another ongoing attempt
+                return false;
+            }
```
sophokles73, line 1052: shouldn't this better be done in HonoConnectionImpl.connect ? There seems to be no need to first instantiate a ConnectionAttempt and then figuring out that it should be discarded, is there?

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
```diff
@@ -1155,23 +1013,199 @@ private int getCloseConnectionTimeout() {
...
+        ConnectionAttempt(final ProtonClientOptions clientOptions, final Handler<AsyncResult<HonoConnection>> connectionHandler) {
+            this.clientOptions = clientOptions;
+
+            this.connectionHandler = ar -> {
+                if (ar.failed()) {
+                    clearState();
+                }
+                currentConnectionAttempt.compareAndSet(this, null);
+                connectionHandler.handle(ar);
+                deferredConnectionCheckHandler.setConnectionAttemptFinished(ar);
```
sophokles73, line 1036: IMHO this could/should be done in HonoConnectionImpl.connect, couldn't it?

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
calohmn: > shouldn't this better be done in HonoConnectionImpl.connect ?

I prefer to make both kind of updates to the `currentConnectionAttempt` reference close to one another in the `ConnectionAttempt` inner class.

> There seems to be no need to first instantiate a ConnectionAttempt

I want to use `currentConnectionAttempt.compareAndSet` here, so an instance of `ConnectionAttempt` is indeed needed.
The (usually exceptional) case that it is created unnecessarily should be no big deal here.

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
calohmn: Same argument as above: I prefer to make all updates to both `currentConnectionAttempt` and `deferredConnectionCheckHandler` close to one another for better readability (the other place is `ConnectionAttempt.start()` then).

sophokles73: I see

client/src/main/java/org/eclipse/hono/client/impl/HonoConnectionImpl.java
sophokles73: I see

sophokles73 (approved):
LGTM

Issue #2549, opened 2021-03-17
Kafka configuration cannot be set using Quarkus based images


sophokles73:
The classes used for configuring Kafka producers and consumers take advantage of a feature of Spring Boot which allows to set Map valued properties via environment/system variables. However, there is no such mechanism in place in Quarkus which effectively prevents Kafka connections being configured at all when using Quarkus based images.

PR #2550, opened 2021-03-17
[#2549] Load Kafka client configuration from property file(s)

sophokles73:
This addresses #2549 

The classes used for configuring Kafka producers and consumers have been
refactored to read (Kafka) configuration properties from file(s) that
have been configured via a setter accepting a list of file names instead
of a Map of configuration properties.

This way, the configuration can also be set when using Quarkus based
images. The existing setters accepting Map<String, String> have been
removed in order to reduce mimgration efforts when migrating to Quarkus
completely.

Don't worry, most files changed are property files that are needed for running the integration tests.
Code changes affect only about ten files.

### DIFF
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaConfigProperties.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaConfigProperties.java
@@ -13,39 +13,113 @@
 
 package org.eclipse.hono.client.kafka;
 
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Properties;
 import java.util.UUID;
 
+import org.eclipse.hono.util.Strings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * Common configuration properties for Kafka clients.
  */
-public abstract class AbstractKafkaConfigProperties {
+public class KafkaConfigProperties {
+
+    /**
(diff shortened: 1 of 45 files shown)


sophokles73:
@b-abel would you mind to also take a look?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaConfigProperties.java
```diff
@@ -13,39 +13,113 @@
...
+    private void loadPropertiesFromFile(final String path, final Properties props) {
+        try (InputStream in = new FileInputStream(path)) {
+            props.load(in);
+            log.info("successfully loaded Kafka client configuration from file [{}]", path);
+        } catch (final FileNotFoundException e) {
+            log.info("could not find property file [{}], ignoring ...", path);
+            throw new IllegalArgumentException("could not find property file", e);
+        } catch (IOException e) {
+            log.info("failed to load Kafka client configuration from file [{}], ignoring ...", path, e);
+            throw new IllegalArgumentException("failed to load properties from file", e);
```
calohmn, line 91: `Ignoring ...` (2 times) doesn't fit here because of the thrown exception.

b-abel:
@sophokles73 I already implemented a solution for that but didn't provide a PR yet because there are other problems with the Kafka client in the native image that I wanted to fix before adding Kafka support to the Quarkus adapters. My solution was to add the Quarkus Kafka extension and overwrite the configuration parsing. The Quarkus extension contains a couple of fixes that are necessary to use the Kafka client with GraalVM.
I'll try to find time today to look at this PR and to create a PR with my solution. Then we can see which way we prefer. WDYT?

sophokles73:
sounds good

b-abel:
@sophokles73  I added #2553.

sophokles73:
closed in favor of https://github.com/eclipse/hono/pull/2553

PR #2552, opened 2021-03-17
add command mapper to mqtt adapter

BobClaerhout:

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/RegistrationConstants.java
+++ b/core/src/main/java/org/eclipse/hono/util/RegistrationConstants.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016, 2021 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -37,10 +37,16 @@ public final class RegistrationConstants extends RequestResponseApiConstants {
     public static final String FIELD_VIA = "via";
 
     /**
-     * The name of the mapper used. This mapper should be configured for the adapter and can be referenced using
+     * The name of the downstream mapper used. This mapper should be configured for the adapter and can be referenced using
      * this field.
      */
-    public static final String FIELD_MAPPER = "mapper";
+    public static final String FIELD_DOWNSTREAM_MESSAGE_MAPPER = "downstream-message-mapper";
+
+    /**
+     * The name of the upstream mapper used. This mapper should be configured for the adapter and can be referenced using
+     * this field.
+     */
+    public static final String FIELD_UPSTREAM_MESSAGE_MAPPER = "upstream-message-mapper";
 
     /**
      * The name of the Device Registration API endpoint.
--- a/core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
+++ b/core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019, 2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2019, 2021 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -73,10 +73,16 @@ public final class RegistryManagementConstants extends RequestResponseApiConstan
     public static final String FIELD_STATUS_LAST_USER = "last-user";
 
     /**
-     * The name of the field that contains the name of a service that can be used to transform messages
+     * The name of the field that contains the name of a service that can be used to transform downstream messages
      * uploaded by the device before they are forwarded to downstream consumers.
      */
-    public static final String FIELD_MAPPER = "mapper";
+    public static final String FIELD_DOWNSTREAM_MESSAGE_MAPPER = "downstream-message-mapper";
+
+    /**
+     * The name of the field that contains the name of a service that can be used to transform upstream commands
+     * to be sent to the device.
+     */
+    public static final String FIELD_UPSTREAM_MESSAGE_MAPPER = "upstream-message-mapper";
 
     /**
      * The name of the field that contains the names of the gateway groups that the (gateway) device is a member of.
--- a/site/documentation/content/api/management/device-registry-v1.yaml
+++ b/site/documentation/content/api/management/device-registry-v1.yaml
@@ -1057,11 +1057,18 @@ components:
                   The authorities granted to a device. Applies to gateways only currently: the authority
                   `auto-provisioning-enabled` entitles a gateway device to perform auto-provisioning for its
                   edge devices.
-            "mapper":
+            "downstream-message-mapper":
                type: string
                description: |
-                  The name of a service that can be used by protocol adapters to transform messages
+                  The name of a service that can be used by protocol adapters to transform downstream messages
                   uploaded by devices before sending them to downstream consumers.
+                  The protocol adapter needs to map this name to the particular service to invoke.
+                  NB: This is an experimental feature and thus this property is subject to change without
+                  notice.
+            "upstream-message-mapper":
+               type: string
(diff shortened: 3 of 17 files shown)


BobClaerhout:
It's working as intended right now.

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -1065,6 +1065,13 @@ components:
                   The protocol adapter needs to map this name to the particular service to invoke.
                   NB: This is an experimental feature and thus this property is subject to change without
                   notice.
+            "command-mapper":
```
calohmn, line 1068: Having now a "mapper" and a "command-mapper" field doesn't look particularly well designed here.
We *could* make a clearer distinction here and rename the "mapper" field to something like "downstream-message-mapper" (or "telemetry-mapper", meaning both telemetry and event messages as used in some other places in Hono, although I think that would be rather confusing).
On the other hand one could think of the telemetry/event direction as the more important, default direction, so using just "mapper" would match that. So, I'm not sure here.
@sophokles73 Any opinion on this?

core/src/main/java/org/eclipse/hono/util/RegistrationConstants.java
```diff
@@ -42,6 +42,12 @@
      */
     public static final String FIELD_MAPPER = "mapper";
 
+    /**
+     * The name of the mapper used. This mapper should be configured for the adapter and can be referenced using
+     * this field.
```
calohmn, line 47: "command mapper" instead of "mapper"? To have that javadoc be different from the one for `FIELD_MAPPER`.

site/documentation/content/api/management/device-registry-v1.yaml
sophokles73: Do we actually need to define two mappers? Will it be the usual case to use two different services for mapping the payload?

BobClaerhout: @sophokles73, yes we require 2 different mappers. Mapping the downstream payload is sort of decoding the payload while mapping the upstream payload encodes the payload.
In our usecase some additional information is added to the payload which is not required when sending to an LNS (because that specific json payload is added by the lora providers)
Considering the name: I thought it would be clear enough to have a mapper for downstream and specifying it for upstream although maybe `downstream-mapper` and `upstream-mapper` might suffice as well?

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/Mapping.java
```diff
@@ -13,19 +13,21 @@
...
+import io.vertx.core.buffer.Buffer;
 
 /**
  * A service for processing messages uploaded by devices before they are being
  * forwarded downstream.
  *
  * @param <T> The type of execution context supported by this mapping service.
  */
-public interface MessageMapping<T extends ExecutionContext> {
+public interface Mapping<T extends ExecutionContext> {
```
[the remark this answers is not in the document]
BobClaerhout: I was struggling with this as well. `MessageMapping` as it was before didn't cover anymore what is was doing so that's why I renamed it. But I like your approach to rename the method and clarifying it that way.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/impl/HttpBasedMapping.java
```diff
@@ -123,6 +124,86 @@ public HttpBasedMessageMapping(
...
+
+        webClient.post(mapperEndpoint.getPort(), mapperEndpoint.getHost(), mapperEndpoint.getUri())
+            .putHeaders(headers)
+            .ssl(mapperEndpoint.isTlsEnabled())
+            .sendBuffer(command.getPayload(), httpResponseAsyncResult -> {
+                if (httpResponseAsyncResult.failed()) {
+                    LOG.debug("failed to map message [origin: {}] using mapping service [host: {}, port: {}, URI: {}]",
+                        command.getDeviceId(),
+                        mapperEndpoint.getPort(), mapperEndpoint.getHost(), mapperEndpoint.getUri(),
+                        httpResponseAsyncResult.cause());
```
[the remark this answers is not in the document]
BobClaerhout: next level reviewing! nice catch. Was a copy paste from another typo :)

site/documentation/content/api/management/device-registry-v1.yaml
sophokles73: I am not sure if we are on the same page here. i am not arguing for using the same logic to map downstream and upstream payload. I am just saying that we could (should?) use the same service component for both directions which could be selected by means of using different URIs, e.g. using `http://my-mapper/map/downstream` vs. `http://my-mapper/map/upstream`. The `downstream` and `upstream` could even be appended to the generic endpoint URI (`http://my-mapper/map`) by the protocol adapter by convention.

BobClaerhout: @sophokles73, am I correct that you are talking about the actual implementation while @calohmn was talking about the configuration part?
The implementation is a bit different because it has another type of context and the result is also slightly handled differently. As for the configuration: I would rather have 2 different configuration settings instead of using convention. Generating the commands might be a computational intensive operation...

sophokles73: I see.
Based on the assumption that the whole *mapping functionality* is still to be considered experimental, I think that we should indeed rename the *mapper* property to *telemetry-mapper* then.

adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/impl/HttpBasedMapping.java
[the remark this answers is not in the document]
calohmn: It looks like there are still two log lines to be fixed here (L197, 260).

site/documentation/content/api/management/device-registry-v1.yaml
calohmn: Looking at the current implementation in the MQTT adapter, the mapper also gets applied to *command response* messages. I think that also makes sense. With respect to that, `downstream-message-mapper` would fit better.

sophokles73: If we no longer use the message semantics for the names, then the *command-mapper* should probably be named *upstream-message-mapper* accordingly, right?
I am OK with both approaches, I just do not want to mix up both concepts, i.e. we either use a mapper *per direction of message flow* or we use mappers *per message type/semantics*.
Does that make sense @calohmn @BobClaerhout ?

BobClaerhout: Ok, I'm ok with `downstream-message-mapper` and `upstream-message-mapper`. I think this is descriptive.

calohmn: I'm ok with that as well.

BobClaerhout: I changed it and adopted the documentation accordingly

core/src/main/java/org/eclipse/hono/util/RegistryManagementConstants.java
```diff
@@ -73,10 +73,16 @@
     public static final String FIELD_STATUS_LAST_USER = "last-user";
 
     /**
-     * The name of the field that contains the name of a service that can be used to transform messages
+     * The name of the field that contains the name of a service that can be used to transform downstream messages
      * uploaded by the device before they are forwarded to downstream consumers.
      */
-    public static final String FIELD_MAPPER = "mapper";
+    public static final String FIELD_DOWNSTREAM_MAPPER = "downstream-message-mapper";
```
calohmn, line 79: `FIELD_DOWNSTREAM_MESSAGE_MAPPER`?

calohmn (commented):
Looks good FMPOV. Can you update the copyright header year of changed files where needed and add a release notes entry (also noting the renamed field) ?

calohmn (approved):
LGTM

PR #2553, opened 2021-03-17
[#2490] Support Kafka messaging in Quarkus based Adapters.

### LINKED ISSUE #2490: Support messaging with Kafka in Quarkus-based Protocol Adapters
While the protocol adapters that use Spring Boot already support downstream messaging with Kafka, the Quarkus-based protocol adapters miss this support.

b-abel:
This fixes #2490. It adds support for messaging with Kafka to the Quarkus-based protocol adapters. It uses the Quarkus Kafka Extension which provides two features:
a) it parses the Kafka configuration and b) it provides fixes to the original Kafka client that are necessary to run it in GraalVM. The class `KafkaRuntimeConfigProducer` overwrites the config parsing so that it supports the Hono-specific prefixes for the Kafka configuration.

Initially, I implemented a custom Quarkus extension that parses Hono's Kafka configuration properties and provides them as beans. When testing it in native images, I found that the Kafka client needs a couple of fixes to run in GraalVM. Instead of implementing them on our own, we can simply include `quarkus-kafka-client` which already contains the fixes. So, I simply overwrite the parsing log for the Kafka config. The price is that produces and consumer properties are added to the same map so that one cannot use different values for the same common properties. But FMPOV this should be no real issue in practice. 

An alternative approach is provided by #2550.

Open: 
The Vert.x `BufferSerializer` needs to be added to the reflections file  (and maybe other classes as well) for the native image to work.

### DIFF
--- a/adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
+++ b/adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
@@ -33,6 +33,7 @@ import org.eclipse.hono.adapter.client.command.amqp.ProtonBasedCommandRouterClie
 import org.eclipse.hono.adapter.client.command.amqp.ProtonBasedDelegatingCommandConsumerFactory;
 import org.eclipse.hono.adapter.client.command.amqp.ProtonBasedDeviceConnectionClient;
 import org.eclipse.hono.adapter.client.command.amqp.ProtonBasedInternalCommandConsumer;
+import org.eclipse.hono.adapter.client.command.kafka.KafkaBasedCommandResponseSender;
 import org.eclipse.hono.adapter.client.registry.CredentialsClient;
 import org.eclipse.hono.adapter.client.registry.DeviceRegistrationClient;
 import org.eclipse.hono.adapter.client.registry.TenantClient;
@@ -42,6 +43,8 @@ import org.eclipse.hono.adapter.client.registry.amqp.ProtonBasedTenantClient;
 import org.eclipse.hono.adapter.client.telemetry.EventSender;
 import org.eclipse.hono.adapter.client.telemetry.TelemetrySender;
 import org.eclipse.hono.adapter.client.telemetry.amqp.ProtonBasedDownstreamSender;
+import org.eclipse.hono.adapter.client.telemetry.kafka.KafkaBasedEventSender;
+import org.eclipse.hono.adapter.client.telemetry.kafka.KafkaBasedTelemetrySender;
 import org.eclipse.hono.adapter.monitoring.ConnectionEventProducer;
 import org.eclipse.hono.adapter.monitoring.HonoEventConnectionEventProducer;
 import org.eclipse.hono.adapter.monitoring.LoggingConnectionEventProducer;
@@ -49,6 +52,8 @@ import org.eclipse.hono.adapter.monitoring.quarkus.ConnectionEventProducerConfig
 import org.eclipse.hono.adapter.resourcelimits.ResourceLimitChecks;
 import org.eclipse.hono.client.HonoConnection;
 import org.eclipse.hono.client.SendMessageSampler;
+import org.eclipse.hono.client.kafka.KafkaProducerConfigProperties;
+import org.eclipse.hono.client.kafka.KafkaProducerFactory;
 import org.eclipse.hono.client.quarkus.RequestResponseClientConfigProperties;
 import org.eclipse.hono.client.util.MessagingClient;
 import org.eclipse.hono.config.ClientConfigProperties;
@@ -80,6 +85,7 @@ import io.quarkus.runtime.StartupEvent;
 import io.vertx.core.DeploymentOptions;
 import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
 import io.vertx.core.impl.cpu.CpuCoreSensor;
 
 /**
@@ -117,6 +123,9 @@ public abstract class AbstractProtocolAdapterApplication<C extends ProtocolAdapt
 
     @Inject
     protected ApplicationConfigProperties appConfig;
+
+    @Inject
+    protected KafkaProducerConfigProperties kafkaProducerConfig;
 
     @ConfigPrefix("hono.messaging")
     protected RequestResponseClientConfigProperties downstreamSenderConfig;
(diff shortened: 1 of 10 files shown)


adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/KafkaRuntimeConfigProducer.java
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Gets the properties with the given prefix from the provided config.
+     *
+     * @param config The provided configuration.
+     * @param configPrefix The prefix of the properties to be returned.
+     * @return The properties with the prefix removed.
+     */
+    protected final Map<String, Object> createPropertiesFromConfig(final Config config, final String configPrefix) {
+        final Map<String, Object> properties = new HashMap<>();
```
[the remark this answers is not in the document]
b-abel: Yes, that should be changed. Although I don't think we can change the return type of the method that exposes the result as a bean: `Map<String, Object> createKafkaRuntimeConfig()`.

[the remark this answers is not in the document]
b-abel: Because we are "overwriting" a bean that is created by the Quarkus Kafka extension. While the injection uses a name, I assume that we cannot change the type, but I have not verified this yet. If your suggestion works, I am more than happy to change it.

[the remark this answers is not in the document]
sophokles73: My understanding is that we do not even need the name anymore once we change the type ...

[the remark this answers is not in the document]
b-abel: You are right. So, I went a step further and create beans with our configuration classes instead of a map.

sophokles73:
@b-abel I like this approach a lot :+1: Can you rebase?
We then need to address the problem of the kafka config properties being picked up by Quarkus even if we want to run the AMQP based integration tests.

adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
```diff
@@ -127,8 +125,10 @@
     protected ApplicationConfigProperties appConfig;
 
     @Inject
-    @Named("default-kafka-broker")
-    protected Instance<Map<String, Object>> kafkaConfig;
+    protected Instance<KafkaProducerConfigProperties> kafkaProducerConfig;
```
sophokles73, line 128: why do we need to use *Instance* here instead of a field of type *KafkaProducerConfigProperties*?

adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/KafkaRuntimeConfigProducer.java
```diff
@@ -0,0 +1,143 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Parses Hono's Kafka configuration properties and provides them as beans.
+ * <p>
+ * This replaces the behavior of {@link io.quarkus.kafka.client.runtime.KafkaRuntimeConfigProducer} from Quarkus' Kafka
+ * extension. The dependency also provides some patches that are required to use the Kafka clients in native images with
+ * GraalVM.
+ */
+@Dependent
```
sophokles73, line 39: any particular reason why this should have scope *Dependent* instead of *ApplicationScoped*?

adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
b-abel: We don't. I changed it.

adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/KafkaRuntimeConfigProducer.java
b-abel: Not a good one :smile:

b-abel:
> We then need to address the problem of the kafka config properties being picked up by Quarkus even if we want to run the AMQP based integration tests.

I don't know exactly the current state of integration testing regarding various messaging systems. But we want to select the messaging per tenant. So shouldn't we use a tenant in each IT that is configured for the messaging it should be tested with?

adapter-base-quarkus/src/test/java/org/eclipse/hono/adapter/quarkus/KafkaRuntimeConfigProducerTest.java
```diff
@@ -0,0 +1,115 @@
...
+ */
+
+package org.eclipse.hono.adapter.quarkus;
+
+import javax.inject.Inject;
+
+import org.eclipse.hono.client.kafka.KafkaAdminClientConfigProperties;
+import org.eclipse.hono.client.kafka.KafkaProducerConfigProperties;
+import org.eclipse.hono.client.kafka.consumer.KafkaConsumerConfigProperties;
+import org.junit.jupiter.api.Assertions;
```
sophokles73, line 21: can we please use `import static org.assertj.core.api.Assertions.assertThat` instead?

adapter-base-quarkus/src/test/java/org/eclipse/hono/adapter/quarkus/KafkaRuntimeConfigProducerTest.java
```diff
@@ -0,0 +1,115 @@
...
+    @Inject
+    KafkaAdminClientConfigProperties kafkaAdminClientConfig;
+
+    /**
+     * Asserts that the Kafka configuration is exposed as a bean.
+     */
+    @Test
+    public void testThatBeanIsProvided() {
+        Assertions.assertNotNull(kafkaProducerConfig);
+        Assertions.assertNotNull(kafkaProducerConfig);
```
sophokles73, line 48: any particular reason why we need to assert this twice?

sophokles73:
> But we want to select the messaging per tenant. So shouldn't we use a tenant in each IT that is configured for the messaging it should be tested with?

That will be the final step. For now, we want to be able to run the tests with either AMQP or Kafka. But never mind, this issue has already been solved.

sophokles73 (approved):
LGTM

adapter-base-quarkus/src/test/java/org/eclipse/hono/adapter/quarkus/KafkaRuntimeConfigProducerTest.java
b-abel: I've been 10  minutes too late. Please see #2560.

adapter-base-quarkus/src/test/java/org/eclipse/hono/adapter/quarkus/KafkaRuntimeConfigProducerTest.java
b-abel: Please see #2560.

PR #2555, opened 2021-03-17
[#2478] Part 1: Extend ApplicationClient to send commands and wait for command responses

kaniyan:
This PR contains new methods added to the `ApplicationClient` which enables the client to send a command, wait and receive a response for that command. In order to keep this PR short and easier for review, it contains only AMQP 1.0 based implementation of these API methods. Kafka based implementation will be provided in the consecutive PR.

### DIFF
--- a/clients/amqp-common/pom.xml
+++ b/clients/amqp-common/pom.xml
@@ -34,6 +34,10 @@
       <groupId>org.eclipse.hono</groupId>
       <artifactId>hono-client-common</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.github.ben-manes.caffeine</groupId>
+      <artifactId>caffeine</artifactId>
+    </dependency>
     <!-- Testing -->
     <dependency>
       <groupId>org.eclipse.hono</groupId>
--- a/tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
+++ b/tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
@@ -78,7 +78,6 @@ import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
 import io.opentracing.noop.NoopSpan;
 import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
-import io.vertx.core.Handler;
 import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
 import io.vertx.core.buffer.Buffer;
@@ -1045,7 +1044,7 @@ public final class IntegrationTestSupport {
     }
 
     /**
-     * Sends a command to a device.
+     * Sends a command to a device and receives a response.
      *
      * @param tenantId The tenant that the device belongs to.
      * @param deviceId The identifier of the device.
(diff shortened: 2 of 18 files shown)


sophokles73:
Is all this worth the effort? What is wrong with sending commands asynchronously, which we already support?

kaniyan:
> Is all this worth the effort? What is wrong with sending commands asynchronously, which we already support?

When sending commands asynchronously, the users have to correlate the command that has been sent with its response as the future returned doesn't contain a command response. The command responses are received separately by a command response consumer. They may need to implement some logic to keep track of correlation ids to correlate.

Using API methods in this PR, there is no need for users to correlate as the returned future result contains the command response. Also this feature already exists in the existing north-bound APIs.

kaniyan:
> The question is: does this (waiting for the response) represent the pre-dominant use case? 

Since we don't have any usage statistics on that part, it is quite hard for me to say that if it is pre-dominant or not.  But we showcase this feature in our `hono-examples` and `hono-cli`. 

> It looks like we need to make quite some changes for this. So I wonder if it is really worth the effort or if we shouldn't simply wait with this until somebody explicitly asks about it. I would not be surprised if nobody would ever do :-)

In one hand, it also makes sense to wait until someone requests this feature. In that case, we have to replace usage of these methods in examples, cli and integration tests. It also needs some effort. Also Hono users like IoT Hub  who have already created system tests and health checks  based on such API methods also need to adapt them.

I have already analyzed how it can be implemented in case of Kafka. FMPOV it would be more or less the same effort as for changing integration tests, cli and examples. 

The idea is to hold the correlation ids of commands in a collection and receive responses using (a) Kafka consumer(s) and then correlate the responses using the stored correlation ids. Users can send commands to devices across **various** tenants. This can be handled by creating either a single consumer subscribed to multiple topics (tenants). In this case, every time the user sends a command to a device belonging to a new tenant, then we have to do resubscribe with the updated topics list (tenants). Alternatively each consumer can be created for each tenant so that there won't be any lag in receiving responses. But this approach creates as many connections as the number of tenants.  The downside that I see in this Kafka based implementation is that it consumes more memory than that of the AMQP version as it stores the correlation ids e.g lot of commands being sent but  responses are not sent or delayed, provided the timeout is also set to a higher value.

> The other question is if we need this desperately for implementing our integration tests ...

We don't need this desperately for our integration tests. Rather we need to modify those tests to test asynchronous command sending and receiving responses and no need to test the correlation then.

tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
```diff
@@ -999,29 +998,28 @@ public String setupGatewayDeviceBlocking(
...
+
+        LOGGER.trace("sending command [name: {}, contentType: {}, payload: {}]", command, contentType, payload);
+        // send the command upstream to the device and receive the command response
+        final Future<BufferResult> sendCommandTracker = amqpApplicationClient
+                .sendCommand(tenantId, deviceId, command, contentType, payload, properties)
+                .onComplete(ar -> {
+                    vertx.cancelTimer(timerId);
+                    timeOutTracker.tryComplete();
+                    if (ar.succeeded()) {
+                        LOGGER.debug("successfully sent command [name: {}, payload: {}] and received response [payload: {}]",
```
sophokles73, line 1013: this is kind of misleading as it is also logged if the time out has already occurred, i.e. if the response arrives after the timeout, or am I mistaken?

sophokles73:
@kaniyan Then I guess, we'll go on with this PR, right?
@calohmn Any thoughts on this?

kaniyan:
> @kaniyan Then I guess, we'll go on with this PR, right?

I don't see any major technical hindrances in implementing this and we can go with this PR.

clients/amqp-common/pom.xml
```diff
@@ -34,6 +34,10 @@
       <groupId>org.eclipse.hono</groupId>
       <artifactId>hono-client-common</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.github.ben-manes.caffeine</groupId>
+      <artifactId>caffeine</artifactId>
```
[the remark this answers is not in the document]
kaniyan: `com.github.benmanes.caffeine.cache.Cache` is being  used in `AbstractRequestResponseServiceClient`. Setting to optional leads to compilation failure as `ProtonBasedRequestResponseCommandClient` extends `AbstractRequestResponseServiceClient`. 

```
 (default-compile) on project hono-client-application-amqp: Compilation failure
[ERROR] /home/hono/tryout/tmp/hono/clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedRequestResponseCommandClient.java:[46,7] cannot access com.github.benmanes.caffeine.cache.Cache
[ERROR]   class file for com.github.benmanes.caffeine.cache.Cache not found

```

tests/src/test/java/org/eclipse/hono/tests/IntegrationTestSupport.java
kaniyan: You are right. I have updated it now and also for other comments. Would you mind taking a look again?

sophokles73 (approved):
LGTM

PR #2562, opened 2021-03-22
Run command handler in correct vert.x context

calohmn:
Command processing in the adapters needs to be done in the correct vert.x context. Otherwise there will for example be concurrency issues in the AMQP adapter concerning the AMQP connection to the device used for sending the command. 

Since there is only *one* KafkaBasedInternalCommandConsumer instance where commands are received, command handlers explicitly need to be run on the correct context. (That isn't a problem with the  ProtonBasedInternalCommandConsumer since its (multiple) instances use the same context as the corresponding adapter verticles.)

### DIFF
--- a/clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterCommandConsumerFactory.java
+++ b/clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterCommandConsumerFactory.java
@@ -34,6 +34,7 @@ import io.opentracing.SpanContext;
 import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
+import io.vertx.core.Vertx;
 
 /**
  * A factory for creating consumers of command messages.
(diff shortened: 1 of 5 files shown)


clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandRouterCommandConsumerFactory.java
```diff
@@ -172,7 +167,7 @@ public void registerInternalCommandConsumer(
 
         // register the command handler
         final CommandHandlerWrapper commandHandlerWrapper = new CommandHandlerWrapper(tenantId, deviceId, gatewayId,
-                commandHandler);
+                commandHandler, Vertx.currentContext());
```
sophokles73, line 170: this makes it hard to test, doesn't it? Wouldn't it be better to pass in a vert.x Context instead?

calohmn: That would mean a change in the `CommandConsumerFactory` methods and would involve quite many places.
And testing with using `Vertx.currentContext()` would still be possible using the `VertxExtension` for the test.

sophokles73: I see.

sophokles73 (approved):
LGTM

calohmn:
The changes in this PR serve as an extra safeguard. The original reason for the different context is fixed in #2570.

PR #2563, opened 2021-03-22
[#2542] Use tenant trace sampling mode in Command Router

### LINKED ISSUE #2542: Enable forced sampling of traces for a specific tenant (Command Router)
The new config options introduced in #896 to enable forced sampling of traces (i.e. recording of traces in the tracing backend) for a specific tenant should also be taken into account in the Command Router.

calohmn:
This fixes #2542.

### DIFF
--- a/services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
+++ b/services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
@@ -20,6 +20,7 @@ import java.util.List;
 import java.util.Objects;
 
 import org.eclipse.hono.adapter.client.registry.DeviceRegistrationClient;
+import org.eclipse.hono.adapter.client.registry.TenantClient;
 import org.eclipse.hono.client.ServiceInvocationException;
 import org.eclipse.hono.client.util.ServiceClient;
 import org.eclipse.hono.commandrouter.CommandConsumerFactory;
@@ -31,6 +32,7 @@ import org.eclipse.hono.service.commandrouter.CommandRouterResult;
 import org.eclipse.hono.service.commandrouter.CommandRouterService;
 import org.eclipse.hono.util.Lifecycle;
 import org.eclipse.hono.util.RegistrationConstants;
+import org.eclipse.hono.util.TenantConstants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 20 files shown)


services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/amqp/ProtonBasedCommandConsumerFactoryImpl.java
```diff
@@ -82,9 +83,10 @@ public ProtonBasedCommandConsumerFactoryImpl(
     }
 
     @Override
-    public void initialize(final CommandTargetMapper commandTargetMapper) {
+    public void initialize(final TenantClient tenantClient, final CommandTargetMapper commandTargetMapper) {
```
sophokles73, line 86: can't we simply add *tenantClient* and *commandTargetMapper* to the constructor and get rid of this method? We already have life cycle methods and having an additional *initialize* before we can actually start the component seems really awkward to me ...

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/KafkaBasedCommandConsumerFactoryImpl.java
```diff
@@ -112,10 +113,11 @@ public KafkaBasedCommandConsumerFactoryImpl(
     }
 
     @Override
-    public void initialize(final CommandTargetMapper commandTargetMapper) {
+    public void initialize(final TenantClient tenantClient, final CommandTargetMapper commandTargetMapper) {
```
sophokles73, line 116: same here, IMHO we can get rid of this method, can't we?

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/amqp/ProtonBasedCommandConsumerFactoryImpl.java
calohmn: I've refactored the corresponding classes now to remove the `initialize` methods, also removing `@Autowired` annotations in `CommandRouterServiceImpl` (in the command router base module).

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
```diff
@@ -53,82 +48,38 @@
...
-     * If not set explicitly, the {@code NoopTracer} from OpenTracing will
-     * be used.
+     * Creates a new CommandRouterServiceImpl.
      *
-     * @param opentracingTracer The tracer.
+     * @param config The command router service configuration.
+     * @param registrationClient The device registration client.
+     * @param tenantClient The tenant client.
+     * @param deviceConnectionInfo The client for accessing device connection data.
+     * @param commandConsumerFactory The factory to use for creating clients to receive commands.
```
sophokles73, line 69: IMHO we should document the NullPointerException if params are `null`

sophokles73 (approved):
LGTM

Issue #2545, opened 2021-03-16
Close AMQP adapter device connection on terminal errors


calohmn:
When any of the following errors occurs while an AMQP device sends a telemetry/event/command response message, the connection to the device should be closed:

* The adapter is disabled for the tenant that the client belongs to,
* The authenticated device or gateway is disabled or not registered,
* The tenant is disabled or does not exist.

Corresponds to the changes made for the MQTT adapter in #2539.

kaniyan:
This has been closed via PR #2564.

PR #2564, opened 2021-03-22
 [#2545] Close AMQP device connection on terminal errors

kaniyan:
This PR closes the issue #2545.

Now the _AMQP_ adapter closes the connection to the device if one of the following terminal errors occur.

* The adapter is disabled for the tenant that the client belongs to.
* The authenticated device or gateway is disabled or not registered.
* The tenant is disabled or does not exist.

### DIFF
--- a/adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
@@ -615,11 +615,14 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
 
                     spanPreparationFuture
                             .compose(ar -> onMessageReceived(ctx)
-                            .onComplete(r -> msgSpan.finish()));
-
+                                    .onSuccess(ok -> msgSpan.finish())
+                                    .onFailure(error -> closeConnectionOnTerminalError(error, conn, ctx, msgSpan)));
                 } catch (final Exception ex) {
-                    log.warn("error handling message [container: {}, {}]", conn.getRemoteContainer(), authenticatedDevice, ex);
-                    ProtonHelper.released(delivery, true);
+                    log.warn("error handling message [container: {}, {}]", conn.getRemoteContainer(),
+                            authenticatedDevice, ex);
+                    if (!conn.isDisconnected()) {
+                        ProtonHelper.released(delivery, true);
+                    }
                 }
             });
             receiver.open();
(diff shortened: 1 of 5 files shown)


adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -615,7 +615,23 @@ protected void handleRemoteReceiverOpen(final ProtonConnection conn, final Proto
...
+                                                    address.getResourceId(),
+                                                    ctx.getAuthenticatedDevice(),
+                                                    msgSpan.context())
+                                                            .map(isTerminalError -> {
+                                                                span.log("closing connection to device ["
+                                                                        + ctx.getAddress().getResourceId() + "]");
+                                                                conn.close();
+                                                                conn.disconnect();
+                                                                return null;
+                                                            })
```
calohmn, line 632: The `if(isTerminalError)` check is missing here.

I think the whole block, starting with `Optional.ofNullable(ctx.getAddress()).ifPresentOrElse` can be made much more readable by introducing an extra 
````
private Future<Boolean> isTerminalError(final Throwable error, final AmqpContext context, final SpanContext spanContext)
````
method, doing the `ctx.getAddress()` null check in there (either returning `Future.failedFuture(..)` or just `Future.succeededFuture(false)` then).

And the above `.map(isTerminalError -> {` can be replaced with `.onSuccess(isTerminalError -> {`, skipping the `return null;` at the end.

Comment (kaniyan) [reply adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java]: I have extracted into a new method and incorporated other changes too. Would you mind taking a look again?

File: adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java [hunk 2]
```diff
@@ -754,6 +765,17 @@ protected void handleRemoteSenderOpenForCommands(
         });
     }
 
+    private Future<Boolean> isTerminalError(final Throwable error, final AmqpContext ctx,
+            final SpanContext spanContext) {
+        final ResourceIdentifier address = ctx.getAddress();
+        if (address == null) {
+            return Future.succeededFuture(false);
+        } else {
+            return isTerminalError(error, address.getTenantId(), address.getResourceId(), ctx.getAuthenticatedDevice(),
```
calohmn, line 774: I think you can't use `ctx.getAddress().getTenantId()` here.
The address is the original address from the request, so it could just be `event`.

When the event message request gets processed, a separate ResourceIdentifier with included tenant and deviceId gets created in `AbstractProtocolAdapterBase.validateAddress()` and that one is used then, see
https://github.com/eclipse/hono/blob/f227638f19988d034153812d9a462a50e10c152e/adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java#L1119-L1130

adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -754,6 +757,22 @@ protected void handleRemoteSenderOpenForCommands(
...
+            isTerminalError(error, address.getResourceId(), ctx.getAuthenticatedDevice(), span.context())
+                    .onSuccess(isTerminalError -> {
+                        if (isTerminalError) {
+                            span.log("closing connection to device");
+                            conn.close();
+                            conn.disconnect();
+                        }
+                    })
+                    .onComplete(o -> span.finish());
+        }
```
calohmn, line 773: The `span` is not getting closed here in the (theoretical) case of the address being null. So, it would look better to add a corresponding "else" block here or return a Future here and do the closing where the method gets called.

calohmn (approved):
LGTM

PR #2568, opened 2021-03-24
Use "hono.kafka" config prefix in Command Router

calohmn:
Using "hono.kafka" instead of "hono.command.kafka" follows the convention used in the protocol adapters.


sophokles73 (approved):
LGTM

PR #2573, opened 2021-03-25
Rename postRegisterServiceVerticles() to postDeployVerticles()

calohmn:
The old name of that method in AbstractBaseApplication is a relict from when the method was called after deployServiceVerticles() in the run() method. But deployServiceVerticles() is now only contained in the AbstractApplication subclass.


sophokles73 (approved):
LGTM

Issue #2576, opened 2021-03-26
Command Router receives no commands after restart


calohmn:
The Command Router component uses tenant-specific receiver links to receive command messages.

When a Command Router crashes and restarts, these links don't get recreated automatically, so that commands for still-subscribed devices are not forwarded to the adapters.

This is a problem when

- all command routers are restarted at the same time
- or one command router gets restarted, and it previously had a tenant-specific receiver link not existing in one of the other command routers.

If a tenant has many devices and they only subscribe for commands with a short ttd (e.g. HTTP/CoAP devices), then this probably won't cause much trouble because the other command routers probably will have been involved with this tenant as well and the restarted command router will soon get a new command-subscription so that the tenant-specific receiver link is recreated.

But for a tenant with a single MQTT/AMQP device, where a command subscription was registered at the now restarted Command Router, commands won't reach the device and only a restart of the MQTT/AMQP adapter or a new subscription by the device will fix this.

calohmn:
A fix for this should address these cases:

**Command Router with embedded cache**
The device connection/command subscription data will need to be restored from the adapters.
The commandRouter config in the adapters will need to have a "restoreOnReconnect: true" setting in this case.
With this flag set, the adapter will handle a reconnect of the AMQP connection to the Command Router by going through its current command-subscriptions and invoking `commandRouterClient.registerCommandSubscription` for each of them.

Since a command-subscription in a adapter could have been made obsolete by a subsequent subscription for the same device from another adapter, the `registerCommandSubscription` operation needs an additional "timestamp" parameter (in the sense of "timestamp of the original request"), so that the Command Router can ignore requests with a timestamp older than one it already got. With this timestamp, the remaining lifespan can then also be determined.

**Command Router with data grid**
Upon startup, the Command Router queries the Device Connection cache entries for the contained tenant IDs and recreates the corresponding AMQP receiver links.
The Hono Cache interface needs to be extended accordingly, allowing retrieval of distinct key prefixes. To be implemented for Infinispan using a keySet() [stream](https://infinispan.org/docs/stable/titles/developing/developing.html#streams).

sophokles73:
For the second case with the Data Grid:
I am not sure if it is a feasible approach to iterate over all keys in the cache to determine the distinct tenant IDs ...
[...]

PR #2578, opened 2021-03-30
[#2576] Support re-enabling of command routing after restart

sophokles73:
This addresses #2576 

The Command Router API has been extended with an operation to be used by
protocol adadpters for submitting tenant IDs for which the Command
Router should (re-)enable routing of commands after an unexpected
re-start. The command router will then try to (re-)create a command
consumer for the submitted tenant IDs.


services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
```diff
@@ -176,6 +200,46 @@ private Duration getSanitizedLifespan(final Duration lifespan) {
...
     }
 
+    @Override
+    public Future<CommandRouterResult> enableCommandRouting(final List<String> tenantIds, final Span span) {
+        Objects.requireNonNull(tenantIds);
+        final boolean isProcessingRequired = tenantsToEnable.isEmpty();
+        tenantsToEnable.addAll(tenantIds);
+        if (isProcessingRequired) {
+            processTenantQueue();
+        }
```
calohmn, line 210: As the case of an `enableCommandRouting` invocation while `processTenantQueue()` is still in progress is expected here, I think it would be good to skip already handled tenant entries in `processTenantQueue` (e.g. by means of an `alreadyProcessed` Set as `processTenantQueue()` parameter).

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
```diff
@@ -176,6 +200,46 @@ private Duration getSanitizedLifespan(final Duration lifespan) {
                 .otherwise(t -> CommandRouterResult.from(ServiceInvocationException.extractStatusCode(t)));
     }
 
+    @Override
+    public Future<CommandRouterResult> enableCommandRouting(final List<String> tenantIds, final Span span) {
+        Objects.requireNonNull(tenantIds);
+        final boolean isProcessingRequired = tenantsToEnable.isEmpty();
+        tenantsToEnable.addAll(tenantIds);
+        if (isProcessingRequired) {
+            processTenantQueue();
```
calohmn, line 209: I think the tracing output could be structured a bit better by creating a parent span for the "re-enable command routing for tenant" spans created in `processTenantQueue()`.
The span could be created here as a FOLLOWS_FROM span of the `enableCommandRouting` `span` parameter.
(In the `isProcessingRequired` `else` case, the fact that the tenant queue is already being processed could be logged to the `span` (of the `enableCommandRouting` parameter).)

sophokles73: Why don't we simply create all of the *re-enable command routing for tenant* spans as *FOLLOWS_FROM* spans of the *enableCommandRouting* span?

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
sophokles73: I do not really see the value of this information. If tenant IDs are being processed, there will be corresponding FOLLOWS_FROM spans. If not, there won't be any.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
calohmn: Would be ok as well.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
calohmn: I don't quite follow. To rephrase my comment: When there are for example 10 adapters making `enableCommandRouting` requests almost in parallel with the same 10 tenants, it would be good for the `processTenantQueue()` loop to skip duplicate tenant entries, e.g. by keeping track of already handled tenant ids via a `alreadyProcessed` `Set` passed around as `processTenantQueue()` parameter.

sophokles73:
@calohmn I have pushed changes ..

calohmn (approved):
LGTM

PR #2580, opened 2021-03-30
[#1571] Add a introduction page to the documentation.

b-abel:
This addresses  #1571. 
It adds a short introduction to Hono to the landing page of the documentation which explains briefly what Hono is about and how to start learning its concepts.
I feel that having something visual for such an introduction. Therefore I added a graphic. It is quite similar to the one on the homepage but emphasizes different aspects.


sophokles73 (commented):
To be honest, I would be happier with re-using the SVG from the landing page. Both for consistency and for aesthetic reasons ;-)

b-abel:
I replaced the diagram.

sophokles73 (commented):
Thanks for changing the diagram.
How about also linking the mentioned API names to the corresponding specifications?

b-abel:
> How about also linking the mentioned API names to the corresponding specifications?

I did not link the APIs on purpose because, at that point of the reading journey, you should know that those APIs exist, but I don't think you should read them yet. I tried with this introduction page to link pages in the order it might make sense to read them. The specifications are very detailed by nature and full of details relevant for implementing a new protocol adapter which is not the first thing you do when you learn about Hono. WDYT?

sophokles73 (approved):
I see

PR #2582, opened 2021-03-31
[#2194] Clarify documentation regarding password hashes.

### LINKED ISSUE #2194: PasswordSecret's pwd-hash type is not used as declared
The API declares the *pwd-hash* as type *String* of format *byte* which means that the property value is supposed to be a Base64 encoded byte array. However, if the hash-function is `bcrypt` then the hash is a plain String which itself contains the Base64 encoding of the pwd-hash and the salt used to create the hash.

FMPOV we have the follwing options:
1. Keep the property declared as String with format byte. We then need to adapt the registry implementation to always Base64 encode/decode its value.
2. Declare the property as a plain String and amend its documentation that the value is expected to be a plain String if hash-function is bcrypt and a Base64 encoded byte array otherwise.
3. Declare a separate plain String property *bcrypt-hash* which is only used if hash-function is bcrypt.

IMHO option one is the most consistent one.

b-abel:
This fixes #2194.

### DIFF
--- a/site/documentation/content/api/management/device-registry-v1.yaml
+++ b/site/documentation/content/api/management/device-registry-v1.yaml
@@ -1351,8 +1351,12 @@ components:
                                  In this case the id field must be set instead.
                  "pwd-hash":
                     type: string
-                    format: byte
-                    description: The Base64 encoding of the password hash created using the `hash-function` and (optional) `salt`.
+                    description: The password hash as a string, either the output of applying the Bcrypt hash function
+                                 to the clear text password if the `hash-function` is `bcrypt`, or otherwise the Base64 encoding
+                                 of the bytes resulting from applying the hash function to the byte array consisting of the salt bytes
+                                 (if a salt is used) and the UTF-8 encoding of the clear text password.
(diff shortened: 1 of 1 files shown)


site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -1351,8 +1351,11 @@ components:
                                  In this case the id field must be set instead.
                  "pwd-hash":
                     type: string
-                    format: byte
-                    description: The Base64 encoding of the password hash created using the `hash-function` and (optional) `salt`.
+                    description: The password hash as a string, either the output of applying the Bcrypt hash function
+                                 to the clear text password and (optional) `salt` if the `hash-function` is `bcrypt`, or otherwise
```
sophokles73, line 1355: the value in the *salt* field is actually only used if the hash function is *not* `bcrypt`. Bcrypt generates a random salt value as part of the hash algorithm and includes it in the produced hash.

b-abel: Thank you. I changed it.

sophokles73 (approved):
LGTM

PR #2583, opened 2021-04-01
[#2478] Add kafka based implementation to send commands and receive responses by the north bound side

kaniyan:
This PR adds a Kafka based implementation to send commands and receive responses by the north bound side.

In order keep the PR small, the other PR #2571 needs to be merged first.

### DIFF
--- a/clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
+++ b/clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
@@ -12,48 +12,110 @@
  */
 package org.eclipse.hono.application.client.kafka.impl;
 
+import java.time.Duration;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
 
+import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.eclipse.hono.application.client.CommandSender;
 import org.eclipse.hono.application.client.DownstreamMessage;
+import org.eclipse.hono.application.client.MessageConsumer;
 import org.eclipse.hono.application.client.kafka.KafkaMessageContext;
+import org.eclipse.hono.client.SendMessageTimeoutException;
 import org.eclipse.hono.client.kafka.HonoTopic;
 import org.eclipse.hono.client.kafka.KafkaProducerConfigProperties;
 import org.eclipse.hono.client.kafka.KafkaProducerFactory;
 import org.eclipse.hono.client.kafka.KafkaRecordHelper;
+import org.eclipse.hono.client.kafka.consumer.KafkaConsumerConfigProperties;
 import org.eclipse.hono.client.kafka.producer.AbstractKafkaBasedMessageSender;
+import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.MessageHelper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import io.opentracing.Span;
 import io.opentracing.SpanContext;
 import io.opentracing.Tracer;
+import io.opentracing.tag.Tags;
+import io.vertx.core.AsyncResult;
+import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.Promise;
+import io.vertx.core.Vertx;
 import io.vertx.core.buffer.Buffer;
+import io.vertx.kafka.client.consumer.KafkaConsumer;
 
 /**
- * A Kafka based client for sending commands.
+ * A Kafka based client for sending commands and receiving command responses.
  *
  * @see <a href="https://www.eclipse.org/hono/docs/api/command-and-control-kafka/">
  *      Command &amp; Control API for Kafka Specification</a>
  */
 public class KafkaBasedCommandSender extends AbstractKafkaBasedMessageSender
         implements CommandSender<KafkaMessageContext> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(KafkaBasedCommandSender.class);
+    //TODO to make this timeout configurable
+    private static final Duration COMMAND_TIMEOUT = Duration.ofSeconds(30);
+
+    private final KafkaConsumerConfigProperties consumerConfig;
+    private Function<Handler<DownstreamMessage<KafkaMessageContext>>, Future<MessageConsumer>> commandResponseConsumerFactory;
+    // The tenant identifiers are stored as keys in the map. The message consumers which are used to receive command
+    // responses for the tenant mentioned in the key are stored as values.
+    private final ConcurrentHashMap<String, MessageConsumer> commandResponseSubscriptions = new ConcurrentHashMap<>();
+    // The tenant identifiers are stored as keys in the map. The value is an another map with correlation ids as keys
+    // and expiring command promises as values. This correlation ids are used to correlate the response messages with 
+    // that of the command.
+    private final ConcurrentHashMap<String, ConcurrentHashMap<String, ExpiringCommandPromise>> pendingCommandResponses = new ConcurrentHashMap<>();
+    private Supplier<String> correlationIdSupplier = () -> UUID.randomUUID().toString();
+    private final Vertx vertx;
 
     /**
      * Creates a new Kafka-based command sender.
      *
+     * @param vertx The vert.x instance to use.
+     * @param consumerConfig The Kafka consumer configuration properties to use.
+     *                       The value of the config parameter 'auto.offset.reset' should be 'latest'. 
+     *                       If any other value is set then it is overridden with 'latest'.
      * @param producerFactory The factory to use for creating Kafka producers.
-     * @param config          The Kafka producer configuration properties to use.
-     * @param tracer          The OpenTracing tracer.
+     * @param producerConfig The Kafka producer configuration properties to use.
+     * @param tracer The OpenTracing tracer.
      * @throws NullPointerException if any of the parameters are {@code null}.
      */
     public KafkaBasedCommandSender(
+            final Vertx vertx,
+            final KafkaConsumerConfigProperties consumerConfig,
             final KafkaProducerFactory<String, Buffer> producerFactory,
-            final KafkaProducerConfigProperties config,
+            final KafkaProducerConfigProperties producerConfig,
             final Tracer tracer) {
-        super(producerFactory, "command-sender", config, tracer);
+        super(producerFactory, "command-sender", producerConfig, tracer);
+        this.vertx = Objects.requireNonNull(vertx);
+        this.consumerConfig = Objects.requireNonNull(consumerConfig);
+    }
+
+    @SuppressWarnings("rawtypes")
+    @Override
+    public Future<Void> stop() {
+        final List<Future> closeKafkaClientsTracker = commandResponseSubscriptions
+                .keySet()
+                .stream()
+                .map(commandResponseSubscriptions::remove)
+                .filter(Objects::nonNull)
+                .map(MessageConsumer::close)
+                .collect(Collectors.toList());
+
+        closeKafkaClientsTracker.add(super.stop());
+
+        return CompositeFuture.all(closeKafkaClientsTracker)
+                .mapEmpty();
     }
 
     /**
@@ -126,8 +188,52 @@ public class KafkaBasedCommandSender extends AbstractKafkaBasedMessageSender
         Objects.requireNonNull(deviceId);
         Objects.requireNonNull(command);
 
-        // TODO to implement.
-        return null;
+        final String correlationId = correlationIdSupplier.get();
+        final Span span = TracingHelper
+                .buildChildSpan(tracer, context, "send command and wait for response", getClass().getSimpleName())
+                .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_CLIENT)
+                .withTag(TracingHelper.TAG_TENANT_ID, tenantId)
+                .withTag(TracingHelper.TAG_DEVICE_ID, deviceId)
+                .withTag(TracingHelper.TAG_CORRELATION_ID, correlationId)
+                .start();
+        final ExpiringCommandPromise expiringCommandPromise = new ExpiringCommandPromise(
+                correlationId,
+                COMMAND_TIMEOUT,
+                // Remove the correlation id if times out
+                x -> Optional.ofNullable(pendingCommandResponses.get(tenantId))
+                        .ifPresent(ids -> ids.remove(correlationId)),
+                span);
+
+        subscribeForCommandResponse(tenantId, span)
+                .compose(ok -> sendCommand(tenantId, deviceId, command, contentType, data, correlationId, properties,
+                        true, span.context())
+                                .onSuccess(sent -> {
+                                    //Store the correlation id and the expiring command promise
+                                    pendingCommandResponses.computeIfAbsent(tenantId, k -> new ConcurrentHashMap<>())
+                                            .put(correlationId, expiringCommandPromise);
+                                    LOGGER.debug("sent command [correlation-id: {}] and waiting for response", correlationId);
+                                    span.log("sent command and waiting for response");
+                                })
+                                .onFailure(error -> {
+                                    LOGGER.debug("error sending command", error);
+                                    TracingHelper.logError(span, "error sending command", error);
+                                    expiringCommandPromise.tryCompleteAndCancelTimer(Future.failedFuture(error));
+                                }));
+
+        return expiringCommandPromise.future()
+                .onComplete(o -> span.finish());
+    }
+
+    // visible for testing
+    void setCommandResponseConsumerFactory(
+            final Function<Handler<DownstreamMessage<KafkaMessageContext>>, Future<MessageConsumer>> commandResponseConsumerFactory) {
+        this.commandResponseConsumerFactory = commandResponseConsumerFactory;
+    }
+
+    // visible for testing
+    void setCorrelationIdSupplier(final Supplier<String> correlationIdSupplier) {
+        Objects.requireNonNull(correlationIdSupplier);
+        this.correlationIdSupplier = correlationIdSupplier;
     }
 
     private Future<Void> sendCommand(final String tenantId, final String deviceId, final String command,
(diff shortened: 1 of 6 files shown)


clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -12,46 +12,104 @@
...
-public class KafkaBasedCommandSender extends AbstractKafkaBasedMessageSender implements CommandSender {
+public class KafkaBasedCommandSender extends AbstractKafkaBasedMessageSender
+        implements CommandSender<KafkaMessageContext> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(KafkaBasedCommandSender.class);
+    //TODO to make this timeout configurable
+    private static final long COMMAND_TIMEOUT = TimeUnit.SECONDS.toMillis(30);
+
+    private final KafkaConsumerConfigProperties consumerConfig;
+    private Function<Handler<DownstreamMessage<KafkaMessageContext>>, Future<MessageConsumer>> commandResponseConsumerFactory;
+    private final ConcurrentHashMap<String, Pair<MessageConsumer, List<String>>> commandResponseSubscriptions = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
kaniyan: I have added a comment explaining what is stored in the `commandResponseSubscriptions` and pushed the changes.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -157,4 +284,43 @@ public KafkaBasedCommandSender(
 
         return props;
     }
+
+    private Future<Void> subscribeForCommandResponse(final String tenantId,
+            final Handler<DownstreamMessage<KafkaMessageContext>> messageHandler, final Span span) {
+        if (commandResponseSubscriptions.get(tenantId) != null
+                && commandResponseSubscriptions.get(tenantId).one() != null) {
```
calohmn, line 291: The `commandResponseSubscriptions.get(tenantId).one() != null` check looks superfluous and rather confusing here.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -124,8 +187,72 @@ public KafkaBasedCommandSender(
...
+                .withTag(TracingHelper.TAG_TENANT_ID, tenantId)
+                .withTag(TracingHelper.TAG_DEVICE_ID, deviceId)
+                .start();
+        final String correlationId = correlationIdSupplier.get();
+        final Promise<Void> requestTracker = Promise.promise();
+        final Promise<DownstreamMessage<KafkaMessageContext>> responseTracker = Promise.promise();
+        final Promise<Void> timeOutTracker = Promise.promise();
+        final long timerId = vertx.setTimer(COMMAND_TIMEOUT, id -> {
+            timeOutTracker.fail(new SendMessageTimeoutException("sending command/waiting for response timed out after "
+                    + COMMAND_TIMEOUT + "ms"));
```
calohmn, line 202: I think the correlation id should be removed here from the `commandResponseSubscriptions.get(tenantId)` value list.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -124,8 +182,72 @@ public KafkaBasedCommandSender(
...
+                    span.log("sent command and waiting for response");
+                    LOGGER.debug("sent command [correlation-id: {}] and waiting for response", correlationId);
+                    requestTracker.complete();
+                })
+                .onFailure(error -> {
+                    TracingHelper.logError(span, "error sending command", error);
+                    LOGGER.debug("error sending command", error);
+                    requestTracker.fail(error);
+                    responseTracker.tryComplete();
+                    timeOutTracker.tryComplete();
```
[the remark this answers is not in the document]
kaniyan: `CompositeFuture.all()` will fail and when at-least one future fails. But it doesn't mean the other 2 futures are also get completed. The other futures still await for completion and I thought of proactively completing them.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -124,8 +182,72 @@ public KafkaBasedCommandSender(
...
+                .start();
+        final String correlationId = correlationIdSupplier.get();
+        final Promise<Void> requestTracker = Promise.promise();
+        final Promise<DownstreamMessage<KafkaMessageContext>> responseTracker = Promise.promise();
+        final Promise<Void> timeOutTracker = Promise.promise();
+        final long timerId = vertx.setTimer(COMMAND_TIMEOUT, id -> {
+            timeOutTracker.fail(new SendMessageTimeoutException("sending command/waiting for response timed out after "
+                    + COMMAND_TIMEOUT + "ms"));
+            requestTracker.tryComplete();
+            responseTracker.tryComplete();
```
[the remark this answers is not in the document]
kaniyan: Please refer the answer for the previous comment.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
kaniyan: I have incorporated this and the other comments and pushed the changes.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -157,4 +290,42 @@ public KafkaBasedCommandSender(
...
+                    .getConsumerConfig(HonoTopic.Type.COMMAND_RESPONSE.toString());
+
+            // Use a unique group-id so that all command responses for this tenant are received by this consumer.
+            // Thereby the responses can be correlated with the command that has been sent.
+            consumerConfig.put(ConsumerConfig.GROUP_ID_CONFIG, tenantId + "-" + UUID.randomUUID().toString());
+
+            return Optional.ofNullable(commandResponseConsumerFactory)
+                    .map(t -> t.apply(messageHandler))
+                    .orElseGet(
+                            () -> KafkaBasedDownstreamMessageConsumer.create(
```
fkaltner, line 311: If I understood correctly this call also starts the consumer. Maybe I missed it, but I couldn't find a call to `KafkaBasedDownstreamMessageConsumer`'s `close()` method.

Is that still missing or does the consumer stop in some other way?

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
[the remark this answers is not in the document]
calohmn: > But it doesn't mean the other 2 futures are also get completed. The other futures still await for completion and I thought of proactively completing them.

But there's nothing waiting on these futures after `CompositeFuture.all()` has failed, so whether these futures get completed or not doesn't matter.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -157,4 +290,42 @@ public KafkaBasedCommandSender(
...
     }
+
+    private Future<Void> subscribeForCommandResponse(final String tenantId,
+            final Handler<DownstreamMessage<KafkaMessageContext>> messageHandler, final Span span) {
+        if (commandResponseSubscriptions.get(tenantId) != null) {
+            span.log("command response consumer already exists");
+            LOGGER.debug("command response consumer already exists for tenant[{}]", tenantId);
+            return Future.succeededFuture();
+        } else {
+            final Map<String, String> consumerConfig = this.consumerConfig
```
fkaltner, line 301: Does `KafkaBasedCommandSender` make any assumptions about the value of [https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html#consumerconfigs_auto.offset.reset](auto.offset.reset)?

In the unit test I saw this is being set to `earliest`. The default value however is `latest`. Independent from that it can be set by configuration of the Kafka consumer, right?

Hence, if this class makes an assumption on the value of `auto.offset.reset` I guess it makes sense to make sure it can at least not be misconfigured.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -157,4 +290,42 @@ public KafkaBasedCommandSender(
...
+                    .getConsumerConfig(HonoTopic.Type.COMMAND_RESPONSE.toString());
+
+            // Use a unique group-id so that all command responses for this tenant are received by this consumer.
+            // Thereby the responses can be correlated with the command that has been sent.
+            consumerConfig.put(ConsumerConfig.GROUP_ID_CONFIG, tenantId + "-" + UUID.randomUUID().toString());
```
fkaltner, line 306: Does this mean that there will be one consumer per in-flight command?

While I don't think this would be a serious issue for the Kafka broker itself, this would increase the network traffic quite a lot. With every in-flight command there would be a new consumer which will poll all currently present command responses.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
kaniyan: The consumers are closed inside the `stop()` method.

fkaltner: But that `stop()` method isn't called when the command response was processed successfully or failed? At least I could not find a reference in `KafkaBasedCommandSender` to it.

kaniyan: The consumers are created for every _new tenant_ and I think its a disadvantage to close and create a new consumer for every command being sent. So they are not closed after every command.

fkaltner: Ah! I see, that's done in line 296. Thanks! 😄

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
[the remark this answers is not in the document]
kaniyan: > so whether these futures get completed or not doesn't matter.

Since we already knew that these futures need not wait any further and FMPOV it's good to proactively complete them. I don't see any disadvantages in completing them.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -124,8 +187,78 @@ public KafkaBasedCommandSender(
...
+            //remove the correlation id if timed out
+            Optional.ofNullable(commandResponseSubscriptions.get(tenantId))
+                    .map(Pair::two)
+                    .map(ids -> ids.remove(correlationId));
+            TracingHelper.logError(span, timeoutException);
+            timeOutTracker.fail(timeoutException);
+            requestTracker.tryComplete();
+            responseTracker.tryComplete();
+        });
+        final Handler<DownstreamMessage<KafkaMessageContext>> messageHandler = m -> {
```
calohmn, line 212: The `messageHandler` here uses variables specific to this `sendCommand` invocation (`responseTracker`, `timerId`, etc.) but the `messageHandler` is added to the `KafkaBasedDownstreamMessageConsumer` which is only created *once* per tenant.
So it looks to me this doesn't work for any `sendCommand` invocations where `commandResponseSubscriptions.get(tenantId) != null`.

I think `timerId`, `timeOutTracker` and `responseTracker` need to be stored alongside the `correlationId`.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
kaniyan: The command responses are correlated using the correlation id. Once a response has been correlated then corresponding entry has been removed from the list. If the same response is received again then it is simply ignored. Setting `auto.offset.reset` to `earliest` or `latest` doesn't change the behavior of the implementation. 

@fkaltner As we discussed, setting to `earliest` could possibly lead to some overhead when the north bound application is restarted. Then the consumer receives responses that have been correlated already, tries to correlate them again and at the end ignore them. In order to avoid this, we can override in the code that the value of the `auto.offset.reset`is always `latest`.@calohmn WDYT?

calohmn: I agree, better to explicitly set `auto.offset.reset` to `latest` here.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
kaniyan: @calohmn I have replaced it with a wrapper for a promise with expiring mechanism and store them in a map. Would you mind taking a look?

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -157,4 +264,123 @@ public KafkaBasedCommandSender(
...
+        };
+    }
+
+    private Future<Void> subscribeForCommandResponse(final String tenantId, final Span span) {
+        final Handler<DownstreamMessage<KafkaMessageContext>> messageHandler = getCommandResponseHandler(tenantId, span);
+
+        if (commandResponseSubscriptions.get(tenantId) != null) {
+            span.log("command response consumer already exists");
+            LOGGER.debug("command response consumer already exists for tenant[{}]", tenantId);
+            return Future.succeededFuture();
```
calohmn, line 290: In this case here the created `messageHandler` with the span parameter doesn't get used.
That means that in this case a previously created `messageHandler` with a different span is used. That span is then the wrong one here.

So it looks to me the span parameter should be removed from `getCommandResponseHandler()`.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -157,4 +264,123 @@ public KafkaBasedCommandSender(
...
     }
+
+    private Handler<DownstreamMessage<KafkaMessageContext>> getCommandResponseHandler(final String tenantId,
+            final Span span) {
+        return message -> {
+            if (message.getCorrelationId() != null) {
+                // if the correlation id of the received response matches with that of the command
+                Optional.ofNullable(pendingCommandResponses.get(tenantId))
+                        .map(ids -> ids.remove(message.getCorrelationId()))
+                        .ifPresent(expiringPromise -> {
```
calohmn, line 275: How about using `ifPresentOrElse` here instead and adding a `LOGGER.debug(` line about a command response message having been received that couldn't be associated with a sent command (potentially because waiting for the response has timed out).

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -157,4 +264,123 @@ public KafkaBasedCommandSender(
...
+         * @throws NullPointerException if the timeout is {@code null}
+         */
+        ExpiringPromise(final Duration timeout, final Handler<Void> timeOutHandler) {
+            Objects.requireNonNull(timeout);
+
+            timerId = vertx.setTimer(timeout.toMillis(), id -> {
+                LOGGER.debug("cancelling sending command and waiting for response after {} ms", timeout);
+                timerId = null;
+                tryCompleteAndCancelTimer(Future.failedFuture(new SendMessageTimeoutException(
+                        "send command/wait for response timed out after " + timeout + "ms")));
```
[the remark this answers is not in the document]
kaniyan: Originally I had it named as `ExpiringCommandPromise` and then renamed to `ExpiringPromise`. I will rename back to that.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -157,4 +264,123 @@ public KafkaBasedCommandSender(
...
+         * @param timeOutHandler The operation to run after this promise got failed as part of a timeout.
+         * @throws NullPointerException if the timeout is {@code null}
+         */
+        ExpiringPromise(final Duration timeout, final Handler<Void> timeOutHandler) {
+            Objects.requireNonNull(timeout);
+
+            timerId = vertx.setTimer(timeout.toMillis(), id -> {
+                LOGGER.debug("cancelling sending command and waiting for response after {} ms", timeout);
+                timerId = null;
+                tryCompleteAndCancelTimer(Future.failedFuture(new SendMessageTimeoutException(
```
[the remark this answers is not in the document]
kaniyan: We already have such way of handling in our existing code base. Any way I will update that in this PR and  also the below place in a separate commit.

https://github.com/eclipse/hono/blob/b0cf8245d8a9a4af44decd0045eeb5a40317c092/client/src/main/java/org/eclipse/hono/client/impl/DeferredConnectionCheckHandler.java#L180-L181

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
kaniyan: I updated and pushed the changes related to the comments.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -157,4 +265,132 @@ public KafkaBasedCommandSender(
...
+        ExpiringCommandPromise(final Duration timeout, final Handler<Void> timeOutHandler, final Span span) {
+            Objects.requireNonNull(timeout);
+            Objects.requireNonNull(span);
+
+            this.span = span;
+            timerId = vertx.setTimer(timeout.toMillis(), id -> {
+                final SendMessageTimeoutException error = new SendMessageTimeoutException(
+                        "send command/wait for response timed out after " + timeout + "ms");
+                timerId = null;
+                LOGGER.debug("cancelling sending command and waiting for response after {} ms", timeout);
```
calohmn, line 358: It would be good to log the correlation id here.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -157,4 +265,132 @@ public KafkaBasedCommandSender(
...
+         * @throws NullPointerException if the result is {@code null}
+         */
+        final void tryCompleteAndCancelTimer(final AsyncResult<T> result) {
+            Objects.requireNonNull(result);
+
+            Optional.ofNullable(timerId)
+                    .ifPresent(vertx::cancelTimer);
+
+            if (result.succeeded()) {
+                LOGGER.trace("received command response");
```
calohmn, line 379: It would be good to log the correlation id here.

kaniyan: I have pushed the changes related to the above comments.

kaniyan:
The conflicts due to the merge of PR #2571 has been resolved with the latest commit.

calohmn (approved):
LGTM

clients/application-kafka/src/test/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSenderTest.java
```diff
@@ -126,4 +148,65 @@ public void testSendOneWayCommandSucceeds(final VertxTestContext ctx) {
...
                 }));
     }
+
+    /**
+     * Verifies that a command sent to a device succeeds and also a response is received from the device.
+     *
+     * @param ctx The vert.x test context.
+     */
+    @Test
+    public void testSendCommandAndReceiveResponse(final VertxTestContext ctx) {
```
calohmn, line 158: @kaniyan This test seems to always fail when run as part of the nightly jenkins job (see e.g. [here](https://ci.eclipse.org/hono/job/Hono-Nightly-Declarative-Pipeline/966/console)).
````
[ERROR] Tests run: 3, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 5.227 s <<< FAILURE! - in org.eclipse.hono.application.client.kafka.impl.KafkaBasedCommandSenderTest
[ERROR] testSendCommandAndReceiveResponse{VertxTestContext}  Time elapsed: 5.209 s  <<< ERROR!
java.util.concurrent.TimeoutException: The test execution timed out. Make sure your asynchronous code includes calls to either VertxTestContext#completeNow(), VertxTestContext#failNow() or Checkpoint#flag()
````
Can you have a look?

PR #2585, opened 2021-04-07
Improve docs for deploying in minikube

calohmn:
Note that this requires https://github.com/eclipse/packages/pull/215, otherwise there will be image-pull errors if using SNAPSHOT images.


sophokles73 (approved):
LGTM
we should also apply this to the 1.7.x branch ...

calohmn:
Also committed in 1.7.x.

PR #2586, opened 2021-04-07
Fix automatic numbering of steps in SNI documentation.

b-abel:
The admin guide documents steps required to enable SNI: https://www.eclipse.org/hono/docs/admin-guide/secure_communication/#server-name-indication-sni 
In the version currently online, the automatic numbering of the steps is broken and every step is displayed with the number 1. 
My assumption is that this is caused by an indentation of the paragraphs with 3 spaces, so this PR changes it to 4. Although I cannot verify if this will help because I cannot reproduce the error on my machine (my Hugo version renders the numbering OK), this PR will do no harm.

### DIFF
--- a/site/documentation/content/admin-guide/secure_communication.md
+++ b/site/documentation/content/admin-guide/secure_communication.md
@@ -173,33 +173,33 @@ The following steps a re necessary in order to configure the protocol adapters w
 
 1. Create Server Certificate(s)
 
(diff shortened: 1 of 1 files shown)


sophokles73:
I have updated the hugo binary used for building the website to 0.81.0. Let's see if this does the trick already ...

sophokles73:
@b-abel looks like this has fixed the issue ...

b-abel:
> @b-abel looks like this has fixed the issue ...

Great news! So, this is obsolete.

PR #2589, opened 2021-04-08
Fail sending messages if sender already stopped

calohmn:


clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/producer/AbstractKafkaBasedMessageSender.java
```diff
@@ -100,6 +102,7 @@ public AbstractKafkaBasedMessageSender(
      */
     @Override
     public Future<Void> stop() {
+        stopped = true;
```
kaniyan, line 105: This should be set to `stopped = false` in the `start()` method. Otherwise, no messages can be sent after stopping and starting the sender.

kaniyan (approved):
LGTM

PR #2591, opened 2021-04-09
[#2590] Use broker default replication factor for internal topic

### LINKED ISSUE #2590: Support non-static replication factor for internal C&C Kafka topic
Currently the replication factor used when creating the internal C&C Kafka topic is always set to 1.
This may cause errors depending on the Kafka cluster setup.
Instead, the broker "default.replication.factor" setting should be used.

calohmn:
This fixes #2590.

### DIFF
--- a/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedInternalCommandConsumer.java
+++ b/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedInternalCommandConsumer.java
@@ -13,12 +13,16 @@
 
 package org.eclipse.hono.adapter.client.command.kafka;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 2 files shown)


clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedInternalCommandConsumer.java
```diff
@@ -129,8 +134,25 @@ public KafkaBasedInternalCommandConsumer(
...
     }
 
+    /**
+     * Sets the vert.x context to run on.
+     * <p>
+     * If not set explicitly, the context is determined during startup.
+     *
+     * @param context The context.
+     */
+    void setContext(final Context context) {
```
kaniyan, line 144: I see that this method is used only for unit tests. Can't we just add the `context` as an parameter to the constructor with package level visibility, which is only being used by the tests. So that the tests need not invoke this method additionally.

calohmn: Good point, I've changed the commit accordingly.

kaniyan (approved):
LGTM

PR #2593, opened 2021-04-09
Add support for an In-Memory database to the JDBC registry

### LINKED ISSUE #2587: Support bundled Database in JDBC Device Registry
Using the JDBC device registry with a bundled database would provide a simple example registry and allow to replace the file-based registry.

b-abel:
This resolves #2587.
Since an in-memory database does not store data beyond the runtime of the application, a DB with the expected schema must be created each time the Registry Application is started. For the example registry, the example data needs to be created afterward as well. Both tasks can be elegantly solved with a DB migration tool like Flyway.
As far as I can see, there is no logic for the former in Hono yet. For creating the example data, the IoT Packages project contains a job that creates the data via the HTTP API. However, this cannot be used with the in-memory DB, because the data must be created after each start of the registry and not only during the initial deployment. Flyway supports this purpose as well.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -35,7 +35,7 @@
     <californium.version>2.6.2</californium.version>
     <commons-cli.version>1.4</commons-cli.version>
     <dispatch-router.image.name>quay.io/interconnectedcloud/qdrouterd:1.14.0</dispatch-router.image.name>
-    <flyway-core.version>7.7.2</flyway-core.version>
+    <flyway-core.version>7.3.2</flyway-core.version>
     <guava.version>29.0-jre</guava.version>
     <gson.version>2.8.5</gson.version>
     <h2.version>1.4.200</h2.version>
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -35,6 +35,7 @@
     <californium.version>2.6.2</californium.version>
     <commons-cli.version>1.4</commons-cli.version>
     <dispatch-router.image.name>quay.io/interconnectedcloud/qdrouterd:1.14.0</dispatch-router.image.name>
+    <flyway-core.version>7.7.2</flyway-core.version>
     <guava.version>29.0-jre</guava.version>
     <gson.version>2.8.5</gson.version>
     <h2.version>1.4.200</h2.version>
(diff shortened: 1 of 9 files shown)


b-abel:
I suggest that we do the following in the _IoT Packages_ project:

* set the JDBC implementation with in-memory DB as the default example device registry
* remove the file-based registry
* start the job that creates the example data only for the MengoDB based implementation

bom/pom.xml
```diff
@@ -35,6 +35,7 @@
     <californium.version>2.6.2</californium.version>
     <commons-cli.version>1.4</commons-cli.version>
     <dispatch-router.image.name>quay.io/interconnectedcloud/qdrouterd:1.14.0</dispatch-router.image.name>
+    <flyway-core.version>7.7.2</flyway-core.version>
```
calohmn, line 38: It looks like using this specific version would require an Eclipse Contribution Questionnaire (CQ) entry.
There is already an approved entry for version 7.3, so, if the newest version isn't explicitly needed, we could use `7.3.2` here for now (version is from December 2020).

A dependency license check can be done like this:
- run `create_dependencies_list.sh`; this will update `legal/src/main/resources/legal/hono-maven.deps`
- run the [Dash License Tool](https://github.com/eclipse/dash-licenses/): `java -jar <path>/org.eclipse.dash.licenses-0.0.1-SNAPSHOT.jar -summary legal/src/main/resources/legal/DEPENDENCIES legal/src/main/resources/legal/hono-maven.deps`

b-abel: Thank you @calohmn! I will check this.

calohmn: You can do a rebase here and run `create_dependencies_list.sh` with the Dash License Tool jar path as parameter (see #2595).

b-abel: I changed the Flyway version to `7.3.2`.

sophokles73:
> Since an in-memory database does not store data beyond the runtime of the application, a DB with the expected schema must be created each time the Registry Application is started. 

I am not sure about this. As with the file based registry, it would be nice to be able to persist the data in the H2 DB to the file system as well. I also believe that H2 supports this out-of-the-box, doesn't it?

b-abel:
> I also believe that H2 supports this out-of-the-box, doesn't it?

So, your suggestion is to use an [embedded](http://www.h2database.com/html/features.html#embedded_databases) H2 database instead of the [in-memory](http://www.h2database.com/html/features.html#in_memory_databases) mode?

sophokles73:
If that is the H2 specific terminology, then yes.

b-abel:
Close in favor of #2631.

PR #2594, opened 2021-04-09
[#2478] Use ApplicationClient to send commands in integration tests

kaniyan:
The integration tests for AMQP adapter still use methods from `ApplicationClientFactory`
to send commands. This has been replaced now with methods from `ApplicationClient`.


calohmn (approved):
LGTM

PR #2595, opened 2021-04-09
Create DEPENDENCIES file via script

calohmn:
Also sort DEPENDENCIES file content making it better comparable.

Dash License Tool jar is to be given as parameter to the script.

This should serve as an intermediate solution before integration of the Dash License Tool Maven plugin.


sophokles73 (approved):
LGTM

Issue #2480, opened 2021-02-15
CommandRouter/Kafka: prevent lost/duplicate message deliveries


calohmn:
The Kafka command message consumer in the CommandRouter component is currently configured to auto-commit partition offsets.

This means there is a high likelihood of duplicate command messages when a CommandRouter instance is restarted.

On the other hand, it also means that a command message could get lost, if the commit takes place *before* a command message is forwarded on the internal command topic. To prevent this, the offset commit should take place *after* the ack of the message forwarding operation has been received.

sophokles73:
@calohmn do we need this in 1.7.0 or can/should we postpone this issue?

calohmn:
Yes, we can postpone it.

sophokles73:
@calohmn can this be closed?

calohmn:
@sophokles73 Not yet, I'll create one more PR (manual offset commits in command router).

calohmn:
Implemented via 8569ed9d976bae43d1816db33d077e6861f4ca20, adf43a253be0753e1460093176c96634ceba43f1 and e867fdc84aea3f0c884868c97e5d8cab39c063f6.

PR #2596, opened 2021-04-12
[#2480] Skip duplicate command records

calohmn:
This is for #2480:
After a command router has crashed (via a hard kill, not properly closing the Kafka consumer, so that last partition offsets aren't committed), some command records may get sent again on the internal topic.
This change here skips these records in the consumer that reads from this topic by keeping track of the already handled offsets concerning the original tenant-based command topic.

### DIFF
--- a/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedInternalCommandConsumer.java
+++ b/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedInternalCommandConsumer.java
@@ -67,6 +67,10 @@ public class KafkaBasedInternalCommandConsumer implements Lifecycle {
     private final CommandHandlers commandHandlers;
     private final Tracer tracer;
     private final Admin adminClient;
+    /**
+     * Key is the tenant id, value is a Map with partition index as key and offset as value.
+     */
+    private final Map<String, Map<Integer, Long>> lastHandledPartitionOffsetsPerTenant = new HashMap<>();
 
     private KafkaConsumer<String, Buffer> consumer;
     private Context context;
@@ -232,17 +236,30 @@ public class KafkaBasedInternalCommandConsumer implements Lifecycle {
 
     void handleCommandMessage(final KafkaConsumerRecord<String, Buffer> record) {
 
+        // get partition/offset of the command record - related to the tenant-based topic the command was originally received in
+        final Integer commandPartition = KafkaRecordHelper
+                .getHeaderValue(record.headers(), KafkaRecordHelper.HEADER_ORIGINAL_PARTITION, Integer.class)
+                .orElse(null);
+        final Long commandOffset = KafkaRecordHelper
+                .getHeaderValue(record.headers(), KafkaRecordHelper.HEADER_ORIGINAL_OFFSET, Long.class)
+                .orElse(null);
+        if (commandPartition == null || commandOffset == null) {
+            LOG.warn("command record is invalid - missing required original partition/offset headers");
+            return;
+        }
+
         final KafkaBasedCommand command;
         try {
             command = KafkaBasedCommand.fromRoutedCommandRecord(record);
         } catch (final IllegalArgumentException e) {
-            LOG.debug("command record is invalid [tenant-id: {}, device-id: {}]",
+            LOG.warn("command record is invalid [tenant-id: {}, device-id: {}]",
                     KafkaRecordHelper
                             .getHeaderValue(record.headers(), MessageHelper.APP_PROPERTY_TENANT_ID, String.class)
                             .orElse(""),
                     KafkaRecordHelper
                             .getHeaderValue(record.headers(), MessageHelper.APP_PROPERTY_DEVICE_ID, String.class)
-                            .orElse(""), e);
+                            .orElse(""),
+                    e);
             return;
         }
         final CommandHandlerWrapper commandHandler = commandHandlers.getCommandHandler(command.getTenant(),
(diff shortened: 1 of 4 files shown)


calohmn:
@b-abel Can you have a look?

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedInternalCommandConsumer.java
```diff
@@ -260,9 +274,22 @@ void handleCommandMessage(final KafkaConsumerRecord<String, Buffer> record) {
...
         if (commandHandler != null) {
-            LOG.trace("using [{}] for received command [{}]", commandHandler, command);
-            // command.isValid() check not done here - it is to be done in the command handler
-            commandHandler.handleCommand(commandContext);
+            // partition index and offset here are related to the *tenant-based* topic the command was originally received in
+            // therefore they are stored with the tenant as key
+            final Map<Integer, Long> lastHandledPartitionOffsets = lastHandledPartitionOffsetsPerTenant
+                    .computeIfAbsent(command.getTenant(), k -> new HashMap<>());
+            final Long lastHandledOffset = lastHandledPartitionOffsets.get(commandPartition);
+            if (lastHandledOffset != null && commandOffset <= lastHandledOffset) {
```
kaniyan, line 282: The commands are forwarded by the command routers to this topic. The command records are handled by a void method `mapAndDelegateIncomingCommandMessage`. Inside this method we are making several future calls (get tenant, get adapter instance and then send command). It seems to me that we cannot ensure that the commands are forwarded in the same order we have received. In such cases, there is a chance of losing commands when we compare the offsets using `commandOffset <= lastHandledOffset`.

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedInternalCommandConsumer.java
```diff
@@ -67,6 +67,7 @@
     private final CommandHandlers commandHandlers;
     private final Tracer tracer;
     private final Admin adminClient;
+    private final Map<String, Map<Integer, Long>> lastHandledPartitionOffsetsPerTenant = new HashMap<>();
```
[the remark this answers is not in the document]
calohmn: I've added a description.

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedInternalCommandConsumer.java
b-abel: > It seems to me that we cannot ensure that the commands are forwarded in the same order we have received. 

If I understand correctly, this can even mess up the order of commands to a particular device. That would not be acceptable. So, maybe the code in the command router needs to be changed. Then the logic here would work. The alternative, keeping _all_ handled offsets instead of only the last one,  seems not very practicable.

calohmn: > If I understand correctly, this can even mess up the order of commands to a particular device. That would not be acceptable. 

Well, strictly speaking, I think we don't guarantee that the input/output order is kept the same - also not in the case of AMQP commands and also not for telemetry/event messages as they are processed in the protocol adapters. The chances of this happening are however rather slim, I think (I find it difficult imagining a theoretical scenario where this would happen necessarily). But I agree that with the change in this PR, a message would be dropped in that case, and this should be prevented. I'll have a look at guaranteeing the order in the CommandRouter.

b-abel: > also not in the case of AMQP commands and also not for telemetry/event messages as they are processed in the protocol adapters

https://www.eclipse.org/hono/docs/api/telemetry-kafka/#consume-telemetry-data says: "Hono delivers messages containing telemetry data reported by a particular device in the same order that they have been received in". But maybe in the case of a device sending multiple messages at the same time, the execution in the adapter could be out of order.

sophokles73: > But maybe in the case of a device sending multiple messages at the same time, the execution in the adapter could be out of order.

FMPOV this could only happen if a device has *multiple* connections to one or more protocol adapters at the same time. Within each Verticle instance, the messages would still be processed in the order that they have been received. This is due to the fact that each Verticle runs on a single Context.

calohmn: Yes, but what about external service invocations done during the processing of e.g. a telemetry message? I'm thinking about device registry and prometheus requests for example. Theoretically, one such request could take longer than the other. Although with caching in place and with the way such requests are handled in the corresponding components, thinking about a case where this would necessarily appear doesn't seem easy.

calohmn: Concerning possible scenarios relevant to this PR, there might be several commands from different devices, being received by the command router in quick succession and getting mapped to the same gateway and target protocol adapter instance. The number of data grid requests needed to do the mapping could be different for these devices, depending on the number of `via` entries of the device and whether the "last known gateway" is set (see the `CacheBasedDeviceConnectionInfo` class). Hence a later command could overtake an earlier command during processing in the command router. This would most probably involve commands for different devices, but if they are mapped to the same gateway and target adapter instance (and if they originate from the same topic partition), this becomes relevant with regard to the changes in this PR here.
To resolve this, I've now created #2617, ensuring that the command router sends the commands originating from the same topic partition in the order they were received.

calohmn:
Rebased so that GitHub Actions are also run using Kafka.

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedInternalCommandConsumer.java
```diff
@@ -232,17 +236,30 @@ private String getTopicName() {
...
 
+        // get partition/offset of the command record - related to the tenant-based topic the command was originally received in
+        final Integer commandPartition = KafkaRecordHelper
+                .getHeaderValue(record.headers(), KafkaRecordHelper.HEADER_ORIGINAL_PARTITION, Integer.class)
+                .orElse(null);
+        final Long commandOffset = KafkaRecordHelper
+                .getHeaderValue(record.headers(), KafkaRecordHelper.HEADER_ORIGINAL_OFFSET, Long.class)
+                .orElse(null);
+        if (commandPartition == null || commandOffset == null) {
+            LOG.info("command record is invalid - missing required original partition/offset headers");
```
kaniyan, line 247: FMPOV this should be `LOG.warn()` as it indicates something went wrong with the command router.

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedInternalCommandConsumer.java
```diff
@@ -232,17 +236,30 @@ private String getTopicName() {
...
+            return;
+        }
+
         final KafkaBasedCommand command;
         try {
             command = KafkaBasedCommand.fromRoutedCommandRecord(record);
         } catch (final IllegalArgumentException e) {
-            LOG.debug("command record is invalid [tenant-id: {}, device-id: {}]",
+            LOG.info("command record is invalid [tenant-id: {}, device-id: {}]",
```
kaniyan, line 255: FMPOV this should be `LOG.warn()` as it indicates something went wrong with the command router.

kaniyan (approved):
LGTM

PR #2603, opened 2021-04-13
Apply minor improvements to KafkaBasedCommandSender

calohmn:
Minor refactoring and logging improvements.

### DIFF
--- a/clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
+++ b/clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
@@ -69,15 +69,17 @@ public class KafkaBasedCommandSender extends AbstractKafkaBasedMessageSender
     private static final Duration COMMAND_TIMEOUT = Duration.ofSeconds(30);
 
     private final KafkaConsumerConfigProperties consumerConfig;
-    private Function<Handler<DownstreamMessage<KafkaMessageContext>>, Future<MessageConsumer>> commandResponseConsumerFactory;
-    // The tenant identifiers are stored as keys in the map. The message consumers which are used to receive command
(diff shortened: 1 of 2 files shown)


clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -268,62 +275,62 @@ void setCorrelationIdSupplier(final Supplier<String> correlationIdSupplier) {
...
                         .ifPresentOrElse(
                                 expiringCommandPromise -> expiringCommandPromise
                                         .tryCompleteAndCancelTimer(Future.succeededFuture(message)),
-                                () -> LOGGER.trace(
-                                        "ignoring received command response [correlation-id: {}] as there is no matching correlation id found",
-                                        message.getCorrelationId()));
+                                () -> LOGGER.trace("ignoring received command response - no response pending for correlation id [{}], tenant [{}]",
+                                        message.getCorrelationId(), tenantId));
+            } else {
+                LOGGER.trace("ignoring received command response - no responses pending for tenant [correlation-id: {}, tenant: {}]",
```
kaniyan, line 292: I think we could leave the code as before(from line 273)  and we could avoid adding the above log statement and we could append the tenant id to the existing log statement and could achieve the same result.

calohmn: My intention here was to explicitly distinguish between "no entry found for correlation id" and "pendingCommandResponses.get(tenantId) is null".
A "pendingCommandResponses.get(tenantId) is null" case log entry would have probably helped in identifying the issue with the recent `KafkaBasedCommandSenderTest` failures.

kaniyan (approved):
LGTM

PR #2604, opened 2021-04-13
Respect status when sending synchronous command via Kafka

b-abel:
The JavaDoc of the `sendCommand()` methods in the class `CommandSender` defines that the returned future fails if the response does not contain a status code with 2xx. This commit adds that the status code is taken into account in `KafkaBasedCommandSender`.

### DIFF
--- a/clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
+++ b/clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
@@ -12,6 +12,7 @@
  */
 package org.eclipse.hono.application.client.kafka.impl;
 
+import java.net.HttpURLConnection;
 import java.time.Duration;
 import java.util.HashMap;
 import java.util.List;
@@ -30,6 +31,7 @@ import org.eclipse.hono.application.client.DownstreamMessage;
 import org.eclipse.hono.application.client.MessageConsumer;
 import org.eclipse.hono.application.client.kafka.KafkaMessageContext;
 import org.eclipse.hono.client.SendMessageTimeoutException;
+import org.eclipse.hono.client.StatusCodeMapper;
 import org.eclipse.hono.client.kafka.HonoTopic;
 import org.eclipse.hono.client.kafka.KafkaProducerConfigProperties;
 import org.eclipse.hono.client.kafka.KafkaProducerFactory;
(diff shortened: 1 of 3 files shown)


clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -274,14 +276,38 @@ void setCorrelationIdSupplier(final Supplier<String> correlationIdSupplier) {
...
+     * {@link CommandSender#sendCommand(String, String, String, String, Buffer, String, Map, SpanContext)}.
+     *
+     * @param message The received command response.
+     * @return The outcome: a succeeded future if the message contains a 2xx status and a failed future otherwise.
+     */
+    private Future<DownstreamMessage<KafkaMessageContext>> mapResponseResult(
+            final DownstreamMessage<KafkaMessageContext> message) {
+
+        final int status = Optional.ofNullable(message.getStatus()).orElseGet(() -> {
+            LOGGER.trace(
```
calohmn, line 298: I think, this can be logged on DEBUG level (at least) because such a case would most likely be a bug in the protocol adapter that forwarded the command response.

b-abel: I wanted it to match the AMQP pendant. But I agree, this violates the C&C API. If it indicates a bug in the adapter, we should consider an even higher level. What about WARN? And then change it in `ProtonBasedRequestResponseCommandClient` as well.

calohmn: Yes, good.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -274,14 +276,38 @@ void setCorrelationIdSupplier(final Supplier<String> correlationIdSupplier) {
...
+                    "response message has no status code header [tenant ID: {}, device ID: {}, correlation ID: {}]",
+                    message.getTenantId(), message.getDeviceId(), message.getCorrelationId());
+            return HttpURLConnection.HTTP_INTERNAL_ERROR;
+        });
+
+        if (status >= 200 && status < 300) {
```
kaniyan, line 304: It would be good to reuse `StatusCodeMapper.isSuccessful(statusCode)`.

b-abel:

kaniyan (approved):
LGTM

PR #2606, opened 2021-04-14
Add basic LwM2M registration support to CoAP adapter

sophokles73:
The CoAP adapter has been extended with a "Resource Directory" resource
as used by LwM2M 1.0.2. It currently supports registration, updating of
a registration and de-registration of devices using binding modes U and
UQ only.

Upon registration the adapter creates a command consumer for the device
and sends an empty notification downstream which contains a TTD
reflecting the binding mode, i.e. the TTD will be the registered
lifetime for binding mode U and a fixed 20 seconds for mode UQ.

However, the adapter does not support forwarding commands to devices
yet.

The adapter also creates observations of LwM2M objects Device and
Firmware upon registration. In a future version the particular resources
to observe should be configurable at the device level.

Notifications received for the observations are being forwarded
downstream as telemetry or event messages, depending on the type of
object being observed.

The CoAP adapter's configuration properties have been extended with an
undocumented feature flag ("lwm2mEnabled") for enabling support for
LwM2M. The feature is disabled by default.

### DIFF
--- a/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/lwm2m/LwM2MResourceDirectory.java
+++ b/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/lwm2m/LwM2MResourceDirectory.java
@@ -1,0 +1,336 @@
+/**
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+
+package org.eclipse.hono.adapter.coap.lwm2m;
+
+import java.net.HttpURLConnection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+
+import org.eclipse.californium.core.coap.CoAP.ResponseCode;
+import org.eclipse.californium.core.coap.Request;
+import org.eclipse.californium.core.server.resources.CoapExchange;
+import org.eclipse.hono.adapter.coap.CoapContext;
+import org.eclipse.hono.adapter.coap.CoapProtocolAdapter;
+import org.eclipse.hono.adapter.coap.TracingSupportingHonoResource;
+import org.eclipse.hono.auth.Device;
+import org.eclipse.hono.client.ClientErrorException;
+import org.eclipse.hono.tracing.TracingHelper;
+import org.eclipse.hono.util.Lifecycle;
+import org.eclipse.hono.util.Strings;
+import org.eclipse.leshan.core.Link;
+import org.eclipse.leshan.core.californium.EndpointContextUtil;
+import org.eclipse.leshan.core.request.BindingMode;
+import org.eclipse.leshan.server.registration.Registration;
+import org.eclipse.leshan.server.registration.RegistrationUpdate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.opentracing.Span;
+import io.opentracing.Tracer;
+import io.opentracing.tag.Tags;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+
+/**
+ * A CoAP resource implementing basic parts of the
+ * <a href="https://tools.ietf.org/html/draft-ietf-core-resource-directory-23">
+ * CoRE Resource Directory</a>.
+ * <p>
+ * The registration, update and removal requests of devices are mapped to <em>empty
+ * notification</em> events containing the <em>lifetime</em> provided by the device
+ * as TTD value.
+ *
+ */
+public final class LwM2MResourceDirectory extends TracingSupportingHonoResource implements Lifecycle {
+
+    private static final String QUERY_PARAM_BINDING_MODE = "b";
+    private static final String QUERY_PARAM_ENDPOINT = "ep";
+    private static final String QUERY_PARAM_LIFETIME = "lt";
+    private static final String QUERY_PARAM_LWM2M_VERSION = "lwm2m";
+    private static final Logger LOG = LoggerFactory.getLogger(LwM2MResourceDirectory.class);
+
+    private final Set<BindingMode> supportedBindingModes = Set.of(BindingMode.U, BindingMode.UQ);
+    private final LwM2MRegistrationStore registrationStore;
+
+    /**
+     * Creates a new resource directory resource.
+     *
+     * @param adapter The protocol adapter that this resource is part of.
+     * @param registrationStore The component to use for managing registration information.
+     * @param tracer Open Tracing tracer to use for tracking the processing of requests.
+     */
+    public LwM2MResourceDirectory(
+            final CoapProtocolAdapter adapter,
+            final LwM2MRegistrationStore registrationStore,
+            final Tracer tracer) {
+        super(adapter, tracer, "rd");
+        getAttributes().addResourceType("core.rd");
+        this.registrationStore = Objects.requireNonNull(registrationStore);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Future<Void> start() {
+        if (registrationStore instanceof Lifecycle) {
+            ((Lifecycle) registrationStore).start();
+        }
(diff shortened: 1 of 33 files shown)


adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/lwm2m/LwM2MResourceDirectory.java
```diff
@@ -0,0 +1,694 @@
...
+     * Handles a request to create or update a device's registration.
+     */
+    @Override
+    protected Future<?> handlePost(final CoapContext ctx) {
+
+        final Request request = ctx.getExchange().advanced().getRequest();
+        LOG.debug("processing POST request: {}", request);
+
+        // we only support confirmable messages
+        if (!ctx.isConfirmable()) {
```
boaks, line 236: Is it possible to add some diagnose text to the response?
e.g. "lwm2m doesn't support NON requests!".

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/lwm2m/LwM2MResourceDirectory.java
```diff
@@ -0,0 +1,694 @@
...
+                .map(BindingMode::valueOf)
+                .orElse(BindingMode.U);
+        final Link[] objectLinks = Link.parse(request.getPayload());
+
+        final Promise<ResponseCode> result = Promise.promise();
+        if (uri.size() == 1) {
+            handleRegister(ctx, lwm2mVersion, endpoint, bindingMode, lifetime, objectLinks).onComplete(result);
+        } else if (uri.size() == 2) {
+            handleUpdate(ctx, bindingMode, lifetime, objectLinks, uri.get(1)).onComplete(result);
+        } else {
```
boaks, line 271: Maybe this case could also be moved above to

```java
if (uri.size() > 2) {
```

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/lwm2m/LwM2MResourceDirectory.java
```diff
@@ -0,0 +1,694 @@
...
+                    ctx.getOriginDevice().getDeviceId(),
+                    commandContext -> {
+                        Tags.COMPONENT.set(commandContext.getTracingSpan(), getAdapter().getTypeName());
+                     // TODO add logic for sending commands to device
+                    },
+                    null,
+                    currentSpan.context())
+                .compose(commandConsumer -> {
+                    commandConsumers.put(registration.getId(), commandConsumer);
+                    return sendLifetimeAsTtdEvent(ctx, registration, currentSpan);
```
boaks, line 343: The "lwm2m lifetime" is in my opinion no "hono-ttd". that time defines, how frequently update-registrations are expected, not how long the device stays receivable

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/lwm2m/LwM2MResourceDirectory.java
```diff
@@ -0,0 +1,694 @@
...
+
+    private Future<?> sendLifetimeAsTtdEvent(
+            final CoapContext ctx,
+            final Registration registration,
+            final Span currentSpan) {
+
+        final int lifetime = registration.getBindingMode() == BindingMode.U
+                ? registration.getLifeTimeInSec().intValue()
+                : 20;
+        return getAdapter().sendTtdEvent(
```
boaks, line 511: OK, now I get the "lifetime" stuff. I'm not sure, if all devices will do that proper. In fact, sometimes the device itself stays "reachable", but k8s or other network-components closes the ip-route. We will see ...

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/lwm2m/LwM2MResourceDirectory.java
```diff
@@ -0,0 +1,694 @@
...
+     * {@inheritDoc}
+     *
+     * Handles a request to delete a device's registration.
+     */
+    @Override
+    protected Future<?> handleDelete(final CoapContext ctx) {
+
+        final List<String> uri = ctx.getExchange().getRequestOptions().getUriPath();
+
+        if (uri.size() >= 2) {
```
boaks, line 529: `handlePost` is more strict about the `uri.size()`. Is that intended?

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/lwm2m/LwM2MResourceDirectory.java
```diff
@@ -0,0 +1,694 @@
...
+        final Span currentSpan = TracingHelper
+                .buildChildSpan(getTracer(), ctx.getTracingContext(), "update LwM2M device registration", getAdapter().getTypeName())
+                .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_SERVER)
+                .withTag(TracingHelper.TAG_TENANT_ID, ctx.getTenantId())
+                .withTag(TracingHelper.TAG_DEVICE_ID, ctx.getOriginDevice().getDeviceId())
+                .withTag(TAG_LWM2M_REGISTRATION_ID, registrationId)
+                .start();
+
+        LOG.debug("processing LwM2M registration update request [tenant-id: {}, device-id: {}, registration ID]",
+                ctx.getTenantId(), ctx.getOriginDevice().getDeviceId(), registrationId);
```
calohmn, line 471: Missing placeholder after "registration ID".

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/lwm2m/LwM2MResourceDirectory.java
sophokles73: I will change it to only accept URIs with two path segments ....

sophokles73:
I realized that I missed out on some basic things regarding setup of the leshan stack. I wil also split the LwM2M resource directory into a plain CoAP request handling part and a request processing part in order to make it easier to (unit) test these aspects separately from each other. Thus I turned this into a draft for the time being ...

sophokles73:
@boaks @calohmn I have made several changes and also incorporated your feedback. Would you mind taking another look?

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/lwm2m/LwM2MResourceDirectory.java
```diff
@@ -0,0 +1,336 @@
...
+                .map(BindingMode::valueOf)
+                .orElse(BindingMode.U);
+        final Link[] objectLinks = Link.parse(request.getPayload());
+
+        final Promise<ResponseCode> result = Promise.promise();
+        if (uri.size() == 1) {
+            handleRegister(ctx, lwm2mVersion, endpoint, bindingMode, lifetime, objectLinks).onComplete(result);
+        } else {
+            // URI path can have at most 2 segments as already verified above
+            handleUpdate(ctx, bindingMode, lifetime, objectLinks, uri.get(1)).onComplete(result);
```
boaks, line 181: Hard to see, at least 

```java
if (uri.size() > 2)
```

doesn't cover 

```java
if (uri.size() == 0)
```

FMPOV, that should be

```java
if (uri.size() > 2 || uri.size() < 1)
```

sophokles73: uri.size also cannot be 0, it contains at least `rd`, otherwise we wouldn't be handling the request in this Resource, would we?

boaks: Yes, you're right.

sophokles73:
this turned out to not be such a great idea in the end ...

endianjesper:
> this turned out to not be such a great idea in the end ...

Out of curiosity, why? And is lwm2m support still on the table for hono? :)

sophokles73:
It would have required us to jump through quite some hoops in order to add proper lwm2m support. In particular, managing the state involved with maintaining observations of device resources as defined by LwM2M 1.0.2 turns out to not be trivial in a supposedly stateless architecture.

> And is lwm2m support still on the table for hono?

Never say never, but at least I will not spend any more time on it in the foreseeable future. I guess things could be different with LwM2M 1.1 but then we would first need to add support for that to leshan ...
Do you have use cases that would benefit from LwM2M support?

PR #2611, opened 2021-04-16
isValidUri should not be serialized in device registry

BobClaerhout:

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/CommandEndpoint.java
+++ b/core/src/main/java/org/eclipse/hono/util/CommandEndpoint.java
@@ -21,6 +21,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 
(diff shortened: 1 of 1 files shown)


BobClaerhout:
@sophokles73 this bug is present in 1.7.0 release. Any chance this might be added to a 1.7.1 release somehow? If so, do you have any idea when this 1.7.1 release might happen?

sophokles73:
@BobClaerhout We can surely add this to 1.7.1. We usually do a service release if there is  a severe bug to be fixed and/or if a reasonable amount of issues have been fixed. In general, we can do a service release whenever we want. That said, we already have four issues in https://github.com/eclipse/hono/projects/53 so we could do the release as soon as these are fixed ...

sophokles73 (approved):
LGTM

BobClaerhout:
@sophokles73, ok. thanks for explanation. Created the PR: https://github.com/eclipse/hono/pull/2613

PR #2612, opened 2021-04-16
Fix sending commands and receiving responses over Kafka

kaniyan:
Currently the correlation id is stored in a collection after successfully sending a
command. There is a chance that the command response is received by the Kafka
consumer before the correlation id is even stored. This has been fixed now by storing
the correlation id before calling the send command method. If sending a command
fails, then the corresponding correlation id is removed from the collection.

### DIFF
--- a/clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
+++ b/clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
@@ -202,26 +202,31 @@ public class KafkaBasedCommandSender extends AbstractKafkaBasedMessageSender
         final ExpiringCommandPromise expiringCommandPromise = new ExpiringCommandPromise(
                 correlationId,
                 COMMAND_TIMEOUT,
-                // Remove the correlation id if times out
-                x -> Optional.ofNullable(pendingCommandResponses.get(tenantId))
-                        .ifPresent(ids -> ids.remove(correlationId)),
+                // Remove the corresponding pending response entry if times out
+                x -> removePendingCommandResponse(tenantId, correlationId),
                 span);
 
         subscribeForCommandResponse(tenantId, span)
-                .compose(ok -> sendCommand(tenantId, deviceId, command, contentType, data, correlationId, properties,
-                        true, span.context())
-                                .onSuccess(sent -> {
-                                    //Store the correlation id and the expiring command promise
-                                    pendingCommandResponses.computeIfAbsent(tenantId, k -> new ConcurrentHashMap<>())
-                                            .put(correlationId, expiringCommandPromise);
-                                    LOGGER.debug("sent command [correlation-id: {}] and waiting for response", correlationId);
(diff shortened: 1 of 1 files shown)


clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -321,6 +324,11 @@ void setCorrelationIdSupplier(final Supplier<String> correlationIdSupplier) {
         }
     }
 
+    private Optional<ExpiringCommandPromise> removeCorrelationId(final String tenantId, final String correlationId) {
```
calohmn, line 327: Maybe `removePendingCommandResponse` would fit better here.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -203,25 +203,28 @@ public KafkaBasedCommandSender(
...
-                                })
-                                .onFailure(error -> {
-                                    LOGGER.debug("error sending command", error);
-                                    TracingHelper.logError(span, "error sending command", error);
-                                    expiringCommandPromise.tryCompleteAndCancelTimer(Future.failedFuture(error));
-                                }));
+                .compose(ok -> {
+                    //Store the correlation id and the expiring command promise
+                    pendingCommandResponses.computeIfAbsent(tenantId, k -> new ConcurrentHashMap<>())
+                            .put(correlationId, expiringCommandPromise);
```
calohmn, line 213: With that, I think the extra `pendingResponsesForTenant` null/empty check in `getCommandResponseHandler` can be replaced with the `Optional.ofNullable` code like before. The original motivation of logging this case explicitly because of integration test errors is probably obsolete now.

kaniyan: I was not aware that you have introduced these log statements for only the integration tests. It's good to remove it now.

clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -203,25 +203,28 @@ public KafkaBasedCommandSender(
...
+                    return sendCommand(tenantId, deviceId, command, contentType, data, correlationId, properties,
+                            true, span.context())
+                            .onSuccess(sent -> {
+                                LOGGER.debug("sent command [correlation-id: {}] and waiting for response", correlationId);
+                                span.log("sent command and waiting for response");
+                            })
+                            .onFailure(error -> {
+                                LOGGER.debug("error sending command", error);
+                                TracingHelper.logError(span, "error sending command", error);
```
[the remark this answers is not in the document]
kaniyan: I have pushed the changes. Would you mind taking a look again?

calohmn (approved):
LGTM

PR #2617, opened 2021-04-20
[#2480] Ensure command records get routed in the original order

calohmn:
This is for #2480:
This is needed because records with an offset smaller than one already received get skipped in the target protocol adapter (see #2596).

Records handled in the Command Router `mapAndDelegateIncomingCommandMessage` method may overtake each other because determining the target protocol adapter instance may involve a different number of data grid requests and therefore may vary in duration.

### DIFF
--- a/services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/AbstractMappingAndDelegatingCommandHandler.java
+++ b/services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/AbstractMappingAndDelegatingCommandHandler.java
@@ -82,9 +82,10 @@ public abstract class AbstractMappingAndDelegatingCommandHandler implements Life
      * and delegates the command to the resulting protocol adapter instance.
      *
      * @param commandContext The context of the command to send.
+     * @return A future indicating the output of the operation.
      * @throws NullPointerException if the commandContext is {@code null}.
      */
-    protected final void mapAndDelegateIncomingCommand(final CommandContext commandContext) {
+    protected final Future<Void> mapAndDelegateIncomingCommand(final CommandContext commandContext) {
         final Command command = commandContext.getCommand();
 
         // determine last used gateway device id
@@ -92,7 +93,7 @@ public abstract class AbstractMappingAndDelegatingCommandHandler implements Life
 
         final Future<TenantObject> tenantObjectFuture = tenantClient
                 .get(command.getTenant(), commandContext.getTracingContext());
-        tenantObjectFuture
+        return tenantObjectFuture
                 .map(tenantObject -> {
                     TenantTraceSamplingHelper.applyTraceSamplingPriority(tenantObject, null,
                             commandContext.getTracingSpan());
@@ -100,26 +101,7 @@ public abstract class AbstractMappingAndDelegatingCommandHandler implements Life
                 })
                 .compose(tenantObject -> commandTargetMapper.getTargetGatewayAndAdapterInstance(command.getTenant(),
                         command.getDeviceId(), commandContext.getTracingContext()))
-                .onSuccess(result -> {
-                    final String targetAdapterInstanceId = result
-                            .getString(DeviceConnectionConstants.FIELD_ADAPTER_INSTANCE_ID);
-                    final String targetDeviceId = result.getString(DeviceConnectionConstants.FIELD_PAYLOAD_DEVICE_ID);
-                    final String targetGatewayId = targetDeviceId.equals(command.getDeviceId()) ? null : targetDeviceId;
-
-                    if (Objects.isNull(targetGatewayId)) {
-                        log.trace("command not mapped to gateway, use original device id [{}]", command.getDeviceId());
-                    } else {
-                        command.setGatewayId(targetGatewayId);
-                        log.trace("determined target gateway [{}] for device [{}]", targetGatewayId,
-                                command.getDeviceId());
-                        commandContext.getTracingSpan().log("determined target gateway [" + targetGatewayId + "]");
-                    }
-
-                    log.trace("delegate command to target adapter instance '{}' [command: {}]", targetAdapterInstanceId,
-                            commandContext.getCommand());
-                    internalCommandSender.sendCommand(commandContext, targetAdapterInstanceId);
-                })
-                .onFailure(cause -> {
+                .recover(cause -> {
                     final String errorMsg;
 
                     if (tenantObjectFuture.failed()) {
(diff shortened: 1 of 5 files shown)


calohmn:
Rebased so that GitHub Actions are also run using Kafka.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/KafkaBasedMappingAndDelegatingCommandHandler.java
```diff
@@ -83,12 +123,222 @@ public void mapAndDelegateIncomingCommandMessage(final KafkaConsumerRecord<Strin
...
+                } else {
+                    sendNextCommandInQueueIfPossible();
+                }
+            } else if (timerId != null && queue.size() < 2) {
+                vertx.cancelTimer(timerId);
+                timerId = null;
+            }
+        }
+
+        private void onFirstEntryReadyToSend(final Supplier<Future<Void>> commandSendActionSupplier,
```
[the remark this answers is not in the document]
calohmn: I had this before (`sendGivenCommandAndNextInQueue`), but that didn't accurately describe what the method is doing.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/KafkaBasedMappingAndDelegatingCommandHandler.java
```diff
@@ -83,12 +123,222 @@ public void mapAndDelegateIncomingCommandMessage(final KafkaConsumerRecord<Strin
...
+
+            if (commandContext.equals(queue.peek())) {
+                onFirstEntryReadyToSend(sendActionSupplier, sendActionCompletedPromise);
+            } else {
+                // given command is not next-in-line;
+                // that means determining its target adapter instance has finished sooner (maybe because of fewer data-grid requests)
+                // compared to a command that was received earlier
+                removeLeadingEntriesWithBlockingTimeoutReached();
+                if (!queue.contains(commandContext)) {
+                    log.trace("command won't get sent - already cancelled [{}]", commandContext.getCommand());
```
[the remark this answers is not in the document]
calohmn: The span is already finished at that point.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/KafkaBasedMappingAndDelegatingCommandHandler.java
```diff
@@ -83,12 +123,222 @@ public void mapAndDelegateIncomingCommandMessage(final KafkaConsumerRecord<Strin
...
+            } else if (timerId != null && queue.size() < 2) {
+                vertx.cancelTimer(timerId);
+                timerId = null;
+            }
+        }
+
+        private void onFirstEntryReadyToSend(final Supplier<Future<Void>> commandSendActionSupplier,
+                final Promise<Void> sendActionCompletedPromise) {
+            final KafkaBasedCommandContext commandContext = queue.remove();
+            log.trace("sending [{}]", commandContext.getCommand());
```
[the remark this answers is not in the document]
calohmn: There is already a whole new span getting created as part of the subsequent "send" operation ("delegate Command request").

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/KafkaBasedMappingAndDelegatingCommandHandler.java
```diff
@@ -12,47 +12,86 @@
...
 public class KafkaBasedMappingAndDelegatingCommandHandler extends AbstractMappingAndDelegatingCommandHandler {
+
+    /**
+     * Timeout applied to determining target gateway and adapter instance for a command, <i>if</i> that
+     * command is blocking the delegation of a newer command for which these values have already
+     * been determined. An elapsed timeout in that scenario means that the processing of the
+     * corresponding command is cancelled, so that the next-in-line command can be delegated to its
+     * target adapter.
+     */
+    static final Duration BLOCKING_COMMAND_QUEUE_ITEM_TIMEOUT = Duration.ofSeconds(30);
```
calohmn, line 61: The javadoc above describes a safeguard feature I've added here.
I wanted to handle a case where the `getTargetGatewayAndAdapterInstance` operation (which includes device registry and data grid requests) is stuck and blocking the processing of later commands for which `getTargetGatewayAndAdapterInstance` has already returned a value.
I've therefore implemented a timeout mechanism to remove/cancel such command records so that the later commands can get sent.

However, on second thought, I think this is probably not really needed here. The data grid request duration could be limited via HotRod-client timeout configuration values and a scenario where one request is stuck for that long and later ones have already succeeded is probably not very realistic.

I'll therefore remove this feature here, making the code a lot simpler.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/AbstractMappingAndDelegatingCommandHandler.java
```diff
@@ -128,12 +110,46 @@ protected final void mapAndDelegateIncomingCommand(final CommandContext commandC
...
+                })
+                .compose(result -> {
+                    final String targetAdapterInstanceId = result
+                            .getString(DeviceConnectionConstants.FIELD_ADAPTER_INSTANCE_ID);
+                    final String targetDeviceId = result.getString(DeviceConnectionConstants.FIELD_PAYLOAD_DEVICE_ID);
+                    final String targetGatewayId = targetDeviceId.equals(command.getDeviceId()) ? null : targetDeviceId;
+
+                    if (Objects.isNull(targetGatewayId)) {
+                        log.trace("determined target adapter instance [{}] for [{}] (command not mapped to gateway)",
+                                targetAdapterInstanceId, command);
```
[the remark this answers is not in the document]
calohmn: The determined adapter instance id is already set as a tag on the span (by the commandTargetMapper).

calohmn:
@kaniyan Can you have another look? Code is a lot simpler now with the timeout handling removed.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/KafkaBasedMappingAndDelegatingCommandHandler.java
```diff
@@ -83,12 +100,143 @@ public void mapAndDelegateIncomingCommandMessage(final KafkaConsumerRecord<Strin
...
+        }
+
+        private void sendGivenCommandAndNextInQueueIfPossible(
+                final Pair<Supplier<Future<Void>>, Promise<Void>> sendActionSupplierAndResultPromise) {
+            final KafkaBasedCommandContext commandContext = queue.remove();
+            log.trace("sending [{}]", commandContext.getCommand());
+            // TODO use the supplier result future to keep track of the latest fully processed command record with regard to its partition offset
+            sendActionSupplierAndResultPromise.one().get()
+                    .onComplete(sendActionSupplierAndResultPromise.two());
+            sendNextCommandInQueueIfPossible();
```
kaniyan, line 220: Wouldn't it be more appropriate to call  `sendNextCommandInQueueIfPossible()` after `sendActionSupplierAndResultPromise()` completes?

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/KafkaBasedMappingAndDelegatingCommandHandler.java
```diff
@@ -83,12 +100,143 @@ public void mapAndDelegateIncomingCommandMessage(final KafkaConsumerRecord<Strin
...
+            Objects.requireNonNull(commandContext);
+            Objects.requireNonNull(sendActionSupplier);
+
+            final Promise<Void> resultPromise = Promise.promise();
+            final Pair<Supplier<Future<Void>>, Promise<Void>> sendActionSupplierAndResultPromise = Pair
+                    .of(sendActionSupplier, resultPromise);
+            if (commandContext.equals(queue.peek())) {
+                sendGivenCommandAndNextInQueueIfPossible(sendActionSupplierAndResultPromise);
+            } else if (!queue.contains(commandContext)) {
+                log.warn("command won't get sent - already removed from queue [{}]", commandContext.getCommand());
```
kaniyan, line 199: How about also logging this warning in a span too?

calohmn: I currently don't see a scenario where this case can actually happen, so this is just meant as a safeguard here. I think logging is enough for that case. The command context span would also probably be already finished here.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/KafkaBasedMappingAndDelegatingCommandHandler.java
calohmn: That would make it extra sure that the command records are written to Kafka in the right order.
But I don't think it is really needed to wait for the ack from the Kafka cluster here to send the next command. I think it's enough that the `KafkaProducer.send` invocations are in the right order here, which is the case.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/KafkaBasedMappingAndDelegatingCommandHandler.java
kaniyan: It's a very rare scenario. FMPOV such scenarios are very difficult to debug and hence a trace log would be very useful there too.  We can also ensure that the span is not yet closed by checking if the promise corresponding to the KEY_COMMAND_SEND_ACTION_SUPPLIER_AND_RESULT_PROMISE is not finished.

calohmn: I've made the method now more consistent overall by failing the result promise in this case, releasing the command context and also adding a trace log.

kaniyan (approved):
LGTM

PR #2625, opened 2021-04-23
Extend northbound interface to make command time out configurable

kaniyan:


clients/application-kafka/src/main/java/org/eclipse/hono/application/client/kafka/impl/KafkaBasedCommandSender.java
```diff
@@ -375,26 +378,24 @@ void setCorrelationIdSupplier(final Supplier<String> correlationIdSupplier) {
...
-        ExpiringCommandPromise(final String correlationId, final Duration timeout, final Handler<Void> timeOutHandler,
+        ExpiringCommandPromise(final String correlationId, final long timeoutInMs, final Handler<Void> timeOutHandler,
                 final Span span) {
-            Objects.requireNonNull(timeout);
             Objects.requireNonNull(span);
 
             this.span = span;
-            timerId = vertx.setTimer(timeout.toMillis(), id -> {
-                final long timeOutInMs = timeout.toMillis();
+            timerId = vertx.setTimer(timeoutInMs, id -> {
```
calohmn, line 393: This will cause an `IllegalArgumentException` if the timeout is configured as `0`. 
For the `RequestResponseClient#setRequestTimeout` we also allow `0`, meaning no timeout, so I think there should be no exception here in that case.

kaniyan: @calohmn Also `RequestResponseClient#setRequestTimeout` doesn't allow negative values. The `Duration` class accepts negative values too. I have also introduced validation to check if the timeout is not negative in par with the `RequestResponseClient#setRequestTimeout`. I have pushed a new commit for it. Would you mind taking a look?

calohmn (approved):
LGTM

Issue #2099, opened 2020-08-04
Support authentication of devices using X.509 certificates in the CoAP adapter


mostafa-eltaher:
The CoAP adapter should support devices using an X.509 client certificate for authenticating.

sophokles73:
I would suggest to also include support for raw public keys in this issue.

boaks:
x509 - PR #816

"Long ago", I tried to implement it. In the meantime, the hono code base changed a lot, so I'm not sure, if that contains still useful ideas. 

RPK - Issue #733

Contains some ideas about RPK.

For both, x509 and RPK, the validation in Californium is currently still "blocking/sync", so I'm not sure, if the validation could be postponed into the "naturally" none-blocking forwarding of the message.

sophokles73:
@boaks 

> x509 - PR #816
> 
> "Long ago", I tried to implement it. In the meantime, the hono code base changed a lot, so I'm not sure, if that contains still useful ideas.

I took a look at your PR and I believe that it basically does the right thing(s). It would need to be adapted to the current state of the CoAP adapter but in general it looks quite good. Would you like to give it another shot?

boaks:
I guess, that I will not have time for it until the end of this year ;-(.

sophokles73:
@boaks Then how about I give it a shot? I can rebase and adapt to current CoAP adapter. WDYT?

boaks:
Sounds good.

boaks:
@sophokles73 
@mostafa-eltaher 

I started to implement a "none blocking" `CertificateVerifier` and `RPK lookup`. That's only the first step to get an initial support. The steps for handling multiple server certificates based on DNS or signature algorithms are not planned from my side for now. Also steps to make the "session cache" really cluster ready, are not planned (see [Evolving the SessionCache](https://github.com/eclipse/californium/issues/1345) ).

FMPOV, using expensive crypto-functions as x509 should therefore also use CID. Hopefully with the [cid-load-balancing](https://github.com/bosch-io/californium/tree/wip_dtls_cid_cluster) the cluster topic will be sorted out.

I hope, that approach works for you.

boaks:
@sophokles73 

I created a first PR for x.509 in californium which also supports a custom argument may be used for the device id.

About RPK support:
[...]

PR #2629, opened 2021-04-27
Add dtls-x509 support for coap-adapter.

boaks:
Adapt callback custom argument from String to AdditionalInfo to support
both PSK and x509.
Introduce DeviceInfoSupplier as common callback for PSK and x509.

### DIFF
--- a/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CertificateDeviceResolver.java
+++ b/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CertificateDeviceResolver.java
@@ -1,0 +1,264 @@
+/**
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.hono.adapter.coap;
+
+import java.security.cert.CertPath;
+import java.security.cert.Certificate;
+import java.security.cert.X509Certificate;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+import javax.security.auth.x500.X500Principal;
+
+import org.eclipse.californium.elements.auth.AdditionalInfo;
+import org.eclipse.californium.elements.util.CertPathUtil;
+import org.eclipse.californium.scandium.dtls.AlertMessage;
+import org.eclipse.californium.scandium.dtls.AlertMessage.AlertDescription;
+import org.eclipse.californium.scandium.dtls.AlertMessage.AlertLevel;
+import org.eclipse.californium.scandium.dtls.CertificateMessage;
+import org.eclipse.californium.scandium.dtls.CertificateType;
+import org.eclipse.californium.scandium.dtls.CertificateVerificationResult;
+import org.eclipse.californium.scandium.dtls.ConnectionId;
+import org.eclipse.californium.scandium.dtls.DTLSSession;
+import org.eclipse.californium.scandium.dtls.HandshakeException;
+import org.eclipse.californium.scandium.dtls.HandshakeResultHandler;
+import org.eclipse.californium.scandium.dtls.x509.NewAdvancedCertificateVerifier;
+import org.eclipse.californium.scandium.util.ServerNames;
+import org.eclipse.hono.adapter.auth.device.DeviceCredentials;
+import org.eclipse.hono.adapter.auth.device.DeviceCredentialsAuthProvider;
+import org.eclipse.hono.adapter.auth.device.SubjectDnCredentials;
+import org.eclipse.hono.adapter.auth.device.X509Authentication;
+import org.eclipse.hono.adapter.client.registry.TenantClient;
+import org.eclipse.hono.auth.Device;
+import org.eclipse.hono.tracing.TenantTraceSamplingHelper;
+import org.eclipse.hono.tracing.TracingHelper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.opentracing.Span;
+import io.opentracing.Tracer;
+import io.opentracing.tag.Tags;
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+
+/**
+ * A Hono Credentials service based implementation of Scandium's x509 authentication related interfaces.
+ *
+ */
+public class CertificateDeviceResolver
+        implements NewAdvancedCertificateVerifier {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CertificateDeviceResolver.class);
+    /**
+     * Use empty issuer list. Device is required to send the proper certificate without this hint. Otherwise, if all
+     * trusted issuers for the tenants are included, the client certificate request will get easily too large.
+     */
+    private static final List<X500Principal> ISSUER = Collections.emptyList();
+    /**
+     * Support only x509 (for now).
+     */
+    private static final List<CertificateType> SUPPORTED_TYPES = Arrays.asList(CertificateType.X_509);
+
+    /**
+     * The vert.x context to run interactions with Hono services on.
+     */
+    private final Context context;
+    private final Tracer tracer;
+    private final String adapterName;
+    private final TenantClient tenantClient;
+    private final X509Authentication auth;
+    private final DeviceCredentialsAuthProvider<SubjectDnCredentials> authProvider;
+
+    private volatile HandshakeResultHandler californiumResultHandler;
+
+    /**
+     * Creates a new resolver.
+     *
+     * @param vertxContext The vert.x context to run on.
+     * @param tracer The OpenTracing tracer.
+     * @param adapterName The name of the protocol adapter.
+     * @param tenantClient The service to use for trace sampling priority.
+     * @param clientAuth The service to use for validating the client's certificate path.
+     * @param authProvider The authentication provider to use for verifying the device identity.
+     * @throws NullPointerException if any of the parameters are {@code null}.
+     */
+    public CertificateDeviceResolver(
+            final Context vertxContext,
+            final Tracer tracer,
+            final String adapterName,
+            final TenantClient tenantClient,
+            final X509Authentication clientAuth,
+            final DeviceCredentialsAuthProvider<SubjectDnCredentials> authProvider) {
+
+        this.context = Objects.requireNonNull(vertxContext);
+        this.tracer = Objects.requireNonNull(tracer);
+        this.adapterName = Objects.requireNonNull(adapterName);
+        this.tenantClient = Objects.requireNonNull(tenantClient);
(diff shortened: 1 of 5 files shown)


boaks:
Fixes issue #2099.

boaks:
One pending question is, if in the mid-future a dtls session resumption should be supported or not.

If it should be supported, there may be two levels:

1. only per coap-adapter instance
2. clustered for all coap-adapter instances

the 2. will obviously be more complex and would require much more time to be implemented.

Regardless of the support-level, the second question supporting it, is,if the "credentials" should be validated again, e.g. to check, if they are expired. If such a check is wanted, a additional API is required in Californium in order to do that none-blocking. The current API `ApplicationLevelInfoSupplier` is blocking and therefore not hat useful.

If resumption must not be considered, the `DeviceInfoSupplier` can be simplified.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CertificateDeviceResolver.java
```diff
@@ -0,0 +1,302 @@
...
+
+        auth.validateClientCertificate(certChain, span.context())
+                .onSuccess(ar -> {
+                    final SubjectDnCredentials credentials = authProvider.getCredentials(ar);
+                    if (credentials == null) {
+                        authResult.fail("x509: no valid subject-dn!");
+                    } else {
+                        LOG.debug("x509: subject-dn: {}", credentials.getAuthId());
+                        applyTraceSamplingPriority(credentials, span)
+                                .onSuccess(v -> {
```
sophokles73, line 179: What if *applyTraceSamplingPriority* fails? It looks like *authResult* will not be completed in that case, or am I mistaken?

boaks: AFAIK, it maps all failures to succeeded.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DeviceInfoSupplier.java
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Key in the extended info of the peer identity containing the authenticated Device.
+     */
+    public static final String EXT_INFO_KEY_HONO_DEVICE = "hono-device";
+
+    private static final Logger LOG = LoggerFactory.getLogger(DeviceInfoSupplier.class);
+
+    /**
+     * The vert.x context to run interactions with Hono services on.
+     */
```
sophokles73, line 53: this comment doesn't seem to match here

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DeviceInfoSupplier.java
```diff
@@ -0,0 +1,135 @@
...
+    private final Tracer tracer;
+    private final String adapterName;
+    private final List<AdvancedApplicationLevelInfoSupplier> suppliers = new CopyOnWriteArrayList<>();
+
+    /**
+     * Creates a new resolver.
```
sophokles73, line 59: a new supplier?

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DeviceInfoSupplier.java
```diff
@@ -0,0 +1,135 @@
...
+        return getInfo(clientIdentity, null);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public AdditionalInfo getInfo(final Principal clientIdentity, final Object customArgument) {
+        if (customArgument instanceof AdditionalInfo) {
+            final Span newSpan = newSpan("get device info");
```
sophokles73, line 105: this span is not getting finished, is it? It doesn't really seem helpful to create a Span here in the first place as there is no real processing to be traced, is there? It should be sufficient to simply let a client log the outcome of this method in a span, if necessary ....

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DefaultDeviceResolver.java
```diff
@@ -55,19 +55,10 @@
...
-
-    /**
-     * Key in the extended info of the peer identity containing the auth-id used for authentication.
-     */
-    public static final String EXT_INFO_KEY_HONO_AUTH_ID = "hono-auth-id";
-    /**
-     * Key in the extended info of the peer identity containing the authenticated Device.
-     */
-    public static final String EXT_INFO_KEY_HONO_DEVICE = "hono-device";
+public class DefaultDeviceResolver implements AdvancedPskStore, AdvancedApplicationLevelInfoSupplier {
```
sophokles73, line 61: shouldn't we rename this to e.g. `PskDeviceResolver` now that we also have a `CertificateDeviceResolver`?

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DeviceInfoSupplier.java
boaks: That class may change much more, when we know, if resumption must be supported and if a non-blocking API will be used. 

My favorite is then just

```
    public AdditionalInfo getInfo(final Principal clientIdentity, final Object customArgument) {
        if (customArgument instanceof AdditionalInfo) {
            return (AdditionalInfo) customArgument;
        }
        LOG.warn("missing custom argument of type AdditionalInfo");
        return AdditionalInfo.empty();
    }
```

Which should obsolete the Tracer and adapter name.

boaks: Comment removed.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DeviceInfoSupplier.java
boaks: changed to supplier.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DeviceInfoSupplier.java
boaks: See my comment above ... if there is a decision about the resumption and behavior, that may make it easier.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DefaultDeviceResolver.java
boaks: Done.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/DeviceInfoSupplier.java
boaks: [This one](https://github.com/eclipse/hono/pull/2629#discussion_r621280534).

You may also consider to read [questions](https://github.com/eclipse/hono/pull/2629#issuecomment-827611731).

> What purpose does the span have?

That depends on the intended functionality based on the (short/mid term) requirements. Therefore the questions.
  
> Why do you want to log an error if the passed in object is an AdditionalInfo instance?

The cleanup (instead of delete) depends on the answers.

boaks:
FMPOV:

- resumption with Californium 2.6.x seems to be no option. Too much work on an old Californium ...
- for the upcoming Californium 3.0 I would go for 
--   changing the definition of the return value of `ApplicationLevelInfoSupplier.getInfo()` to `null` := don't change the additional info. That would help to resume, if the principal has already that info. 
--   Depending on the plans/requirements (resumption on cluster, not only protocol-adapter level), add a new non-blocking API to Californium to check, if the resumption is OK or should turn into a full handshake.

sophokles73:
> One pending question is, if in the mid-future a dtls session resumption should be supported or not.

Are we talking about the standard DTLS *abbreviated handshake* or using the *Connection ID*?

boaks:
About the  standard (D)TLS abbreviated handshake according [TLS / RFC5246](https://tools.ietf.org/html/rfc5246#section-7.3) (frequently also referred as "session resumption" in that RFC5246).

sophokles73:
I was under the impression that the abbreviated handshake had already (always?) been supported by Californium.

boaks:
Sure ... but not in the cluster.
And maybe with trouble using the AdditionalInfo feature (depending on serialize session tickets or direct use of the connection). And for all that odds, I think disable it for hono using Californium 2.6.x will be the right approach.

And, if hono is intended to support an "abbreviated handshake", then it maybe useful to know, on which level (node or cluster) in order to adapt the APi of Californium 3.0.

boaks:
I thought, that I already mentioned that [limitations](https://github.com/eclipse/hono/issues/2099#issuecomment-685727041).

boaks:
Disabled  abbreviated handshake and removed code for providing additional info during resumption.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CertificateDeviceResolver.java
```diff
@@ -0,0 +1,264 @@
...
+        return tracer.buildSpan(operation)
+                .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_SERVER)
+                .withTag(Tags.COMPONENT.getKey(), adapterName)
+                .start();
+    }
+
+    /**
+     * Validate certificate used by a device in a x509 based DTLS handshake and load device.
+     *
+     * @param certChain certificate chain.
```
calohmn, line 125: `certChain` doesn't match parameter name (`certPath`).

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CertificateDeviceResolver.java
```diff
@@ -0,0 +1,264 @@
...
+                .onFailure(t -> authResult.fail(t));
+        return authResult.future();
+    }
+
+    /**
+     * Validate certificate used by a device in a x509 based DTLS handshake, load device, and pass the device back as
+     * result's custom argument.
+     *
+     * @param cid the connection id to report the result.
+     * @param certChain certificate chain.
```
calohmn, line 170: `certChain` doesn't match parameter name (`certPath`).

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CertificateDeviceResolver.java
boaks: Done.

adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/CertificateDeviceResolver.java
boaks: Done.

sophokles73 (approved):
LGTM

boaks:
In order to use it, the packages must adapt `hono-adapter-coap-vertx-secret.yaml` by adding

```sh
application.yml: |
    hono:
      coap:
        keyPath: /etc/hono/key.pem
        certPath: /etc/hono/cert.pem
```

PR #2631, opened 2021-04-27
[#2587] Support automatic schema creation in the JDBC registry.

### LINKED ISSUE #2587: Support bundled Database in JDBC Device Registry

b-abel:
This resolves #2587. It adds the application profile "create-schema". If this profile is active, the tables and indices for the device registry are created. If the profile "tenant-service" is additionally active, the tables and indices for the tenant registry are added as well. The SQL scripts only create the tables and indices if they do not already exist.


b-abel:
I suggest replacing the file-based registry in the _packages_ project with an embedded H2 database on a persistent volume.

b-abel:
Rebased with a fix of the module `tests`.

services/device-registry-jdbc/src/main/java/org/eclipse/hono/deviceregistry/jdbc/impl/ClasspathSchemaCreator.java
```diff
@@ -0,0 +1,106 @@
...
+            return readScript(jdbcClient, "'classpath:/sql/postgresql/02-create.devices.sql'");
+        case SQL.DIALECT_H2:
+            return readScript(jdbcClient, "'classpath:/sql/h2/02-create.devices.sql'");
+        default:
+            return Future.failedFuture(new IllegalArgumentException(
+                    "Unknown database dialect '" + SQL.getDatabaseDialect(deviceProperties.getUrl()) + "'"));
+        }
+    }
+
+    private Future<Void> readScript(final JDBCClient client, final String scriptPath) {
```
sophokles73, line 94: this should probably be called *executeScript* or *runScript* instead, shouldn't it?

services/device-registry-jdbc/src/test/java/org/eclipse/hono/deviceregistry/jdbc/impl/AbstractJdbcRegistryTest.java
```diff
@@ -103,7 +103,8 @@ void startDevices(final Vertx vertx) throws IOException, SQLException {
                 properties
         );
         this.registrationAdapter = new RegistrationServiceImpl(
-                DeviceStores.adapterStoreFactory().createTable(vertx, TRACER, jdbc, Optional.empty(), Optional.empty(), Optional.empty())
+                DeviceStores.adapterStoreFactory().createTable(vertx, TRACER, jdbc, Optional.empty(), Optional.empty(), Optional.empty()),
+                new NoOpSchemaCreator()
```
sophokles73, line 107: why don't we use the SchemaCreator here as well?

b-abel: > why don't we use the SchemaCreator here as well?

I tried this. But that resulted in sporadic, hard-to-analyze test failures, and I'd rather not invest any more time in it at this point.

sophokles73: So now we end up with two different approaches being used to create the tables in the DB? I do not think I understand why you introduced `SchemaCreator` in the first place, if we want (need) to also keep the existing mechanism for creating the schemas ...

sophokles73: No, you're right. We shouldn't spend any more time on it for now.
@ctron any further comments?

sophokles73 (approved):
LGTM

sophokles73:
@b-abel At the moment we start a PostgreSQL DB in the integration tests and let the DB run the schema creation scripts during startup (https://github.com/eclipse/hono/blob/fbe015b3f0d535773f04a59be3572d274446a05d/tests/pom.xml#L580).
FMPOV we should modify that to take advantage of the changes you have introduced and we no longer need to copy the sql scripts to the container image.

b-abel:
>  At the moment we start a PostgreSQL DB in the integration tests and let the DB run the schema creation scripts during startup

I removed that part from the integration tests.

Issue #2623, opened 2021-04-22
Integration tests: Kafka consumers should be closed, Kafka topics deleted after each test


calohmn:
Currently, when running the integration tests, the Kafka topics related to each temporary tenant are not deleted.
This results in an ever growing list of topics during one test run, as shown by this log message:
````
2021-04-21T08:12:58.5519977Z CMDROUTER08:12:58.494 [vert.x-kafka-consumer-thread-0] INFO  o.a.k.c.c.i.ConsumerCoordinator - 
[Consumer clientId=cmd-router-cmd-router-8f678277-31cb-4160-a366-12ddae328487, groupId=cmd-router-group] Notifying assignor about the new Assignment(partitions=
[hono.command.c632c807-157b-4017-8077-c9178f5072c5-0, hono.command.f3c7d3b0-ce57-46e3-8261-5eadceb73357-0, hono.command.1cb17432-726b-4027-b24e-90e86a76987a-0, 
hono.command.f05aa099-f16d-449f-899a-5f9e7e23052e-0, hono.command.99491c6e-7c25-4d6c-91bd-d73d7ef1ad2a-0, hono.command.8ff53b56-04b9-42f6-84a2-8006105612eb-0, 
hono.command.de5375d4-2b74-4102-b108-8a696c5a652c-0, hono.command.d380d419-6f92-46ef-91b4-85986dcdbbdc-0, hono.command.dc1330b7-0061-4a53-92cf-2678ac96b43b-0, 
hono.command.4c97d4f9-b5f9-4cc0-9ae5-eed44d51d3e4-0, hono.command.7484c78e-42fa-4d4c-abc9-f6870ed37805-0, hono.command.e49cd3cd-44ab-4009-8e97-14a107302526-0, [...]
`````
Also the Kafka consumers created for one particular test method should be closed at the end of the method.

calohmn:
When deleting the topics for the temporary tenants, the Command Router "command.*" topic subscription doesn't seem to get updated. This results in countless offset commit errors to be logged:
````
CMDROUTER14:01:10.013 [vert.x-kafka-consumer-thread-0] WARN  o.a.k.c.c.i.ConsumerCoordinator - 
[Consumer clientId=cmd-router-cmd-router-28227f71-4143-46d5-895a-e49bff69bac4, groupId=cmd-router-group] 
[...]

PR #2632, opened 2021-04-28
[#2623] Close created Kafka consumers when client is stopped

calohmn:
This is for #2623.

Makes sure the created Kafka consumers get stopped when the `KafkaApplicationClient` is stopped.

This behaviour is analogeous to closing a `ProtonBasedApplicationClient`, where by closing the underlying AMQP connection all receivers are also stopped.


kaniyan (approved):
LGTM

PR #2636, opened 2021-04-29
[#2620] Add cipher suite tag to hono.connections.attempts metric

### LINKED ISSUE #2620: Include cipher suite name when reporting connection attempts
The protocol adapters currently report connection attempts with an outcome (successful/unsuccessful) tag. It would be nice to also have a tag `cipher` which contains the cipher suite used when a connection has been established on top of TLS.

sophokles73:
The metric used for reporting devices' connection attempts has been
extended with tag "cipher-suite" which takes the name of the cipher
suite used by a device's connection to a protocol adapter.

The AMQP and MQTT adapters have been updated to report the cipher suite
as part of connection attempt processing.

Fixes #2620


service-base/src/main/java/org/eclipse/hono/service/metric/MetricsTags.java
```diff
@@ -418,4 +422,18 @@ public static Tags forService(final String name) {
     public static Tag getTenantTag(final String tenant) {
         return Tag.of(MetricsTags.TAG_TENANT, Strings.isNullOrEmpty(tenant) ? "UNKNOWN" : tenant);
     }
+
+    /**
+     * Creates a tag for a tenant identifier.
```
kaniyan, line 427: The _JavaDoc_  mistakenly describes about the _tenant identifier_  instead of the _cipher suite_.

sophokles73: Thanks for spotting :-)

kaniyan (approved):
LGTM

Issue #2272, opened 2020-10-27
Auto-Provisioning of Gateways


alinaserg:
There're two patterns in Hono how devices can be registered on-the-fly.

1) For devices directly connected to Hono adapters auto-provisoning is done based on X.509 certificates.
2) For devices connected via gateways to Hono adapters, so called implicit device registration (see #2053 ) will be provided.

I want to ask you, if the combination of these two patterns is possible for complete on-the-fly registration of the gateways and devices communicating via these gateways. I.e. there's an edge gateway configured with X.509 certificate for authentication and it's necessary to provision this gateway as well as the edge devices communicating via this gateway on-the-fly.
Meaning the gateway sends on behalf of the edge device one telemetry or event message, firstly the gateway and its credentials will be registered in Device Registry (x.509 cert and device-id provided by the gateway), after that the request will be sent to Ditto to create digital twin for this gateway. Secondly, the edge device for which the gateway trying to send a message is implicitly registered as it is implemented in  #2053. 

@b-abel @fkaltner @sophokles73 could you please provide your opinion, thoughts to this topic?

fkaltner:
I already had a similar idea a couple of months ago, see: https://github.com/eclipse/hono/issues/2053#issuecomment-668609655

After a short chat with @b-abel we came to the following suggestion:
In general it should be possible to combine both approaches, the main challenge ist the authorisation:

1. Auto-provisioning of gateways via client certificates is only allowed if the tenant has the corresponding flag set.
2. Currently auto-provisioning of edge devices will only be allowed if the gateway has the "auto-provisioning-enabled" authority set in its device registration.

We could solve this by taking the tenant flag also into consideration when auto-provisioning edge devices (2), meaning it will "override" the gateway's authority / be a replacement for it.

@alinaserg & @sophokles73 WDYT?

alinaserg:
Thanks @fkaltner for the proposal, additional question regarding replacement of gateways' authority. See below 3 use cases how auto-provisioning can be activated:

**Activating auto-provisioning:**

1. Devices connected directly to Hono protocol adapters -> enabling auto-provisioning on tenant level for CA cert.
2. Devices connected via gateways to Hono protocol adapters, gateways **don't use** client certificates -> enabling auto-provisioning on gateway level by setting  "auto-provisioning-enabled" authority during device registration.
[...]

PR #2637, opened 2021-04-29
[#2272] Prepare for auto-provisioning of gateways

kaniyan:


sophokles73:
This addresses #2272

adapter-base/src/main/java/org/eclipse/hono/adapter/auth/device/TenantServiceBasedX509Authentication.java
```diff
@@ -256,6 +256,8 @@ public TenantServiceBasedX509Authentication(
         if (tenant.isAutoProvisioningEnabled(issuerDn)) {
             try {
                 authInfo.put(CredentialsConstants.FIELD_CLIENT_CERT, deviceCert.getEncoded());
+                authInfo.put(CredentialsConstants.FIELD_AUTO_PROVISION_AS_GATEWAY,
+                        tenant.isAutoProvisioningAsGatewayEnabled(issuerDn));
```
sophokles73, line 260: why does a protocol adapter need to know whether the device will be auto-provisioned as a gateway or as a normal device? Isn't that the business of the registry implementation to determine? Having the adapter include this flag also allows the adapter to simply put in any value it wants to, right? FMPOV the registry must read the value of this flag from the tenant configuration on record and then act accordingly.

kaniyan: Currently the protocol adapters decide if a device need to be auto-provisoned or not as below . So why not to decide if that should be a device or a gateway.


https://github.com/eclipse/hono/blob/86d0f934e1a1a409383ed48d8d30504362622326/adapter-base/src/main/java/org/eclipse/hono/adapter/auth/device/TenantServiceBasedX509Authentication.java#L256-L259


Also in the device registry implementations we check if the certificate is available in the _clientContext_ or not in order to determine if auto-provisioning is enabled or not. We don't retrieve the corresponding tenant's CA from the database and validate it. In the above approach in this PR, the device registry need not search again the underlying database for the corresponding tenant's CA to determine if a device or a gateway is being auto-provisioned.

sophokles73: Obviously. But that is a problem FMPOV as it lets the protocol adapter decide if a device may be auto provisioned or not. The reason why the protocol adapter needs to know if auto-provisioning is enabled at all for the CA is because it needs to determine if it needs to include the client cert in the get Credentials request. However, the decision whether a device can actually be auto provisioned (and as what, a *device* or *gateway*) must be made by the registry, not the adapter.

kaniyan: Do you suggest that the device registry should make use of the tenant's CA entry instead of verifying if a certificate is available in the _clientContext_ to determine if auto-provisioning is enabled or not? And to verify if a device or a gateway is to be auto-provisioned in the same fashion.

sophokles73: exactly

kaniyan:
This has been closed in favor of PR #2647

PR #2640, opened 2021-05-04
Expose tenant information service needed for tenant existence check in device registry

kaniyan:
Any device registration operation first checks if the given tenant exists or not
using tenant information service. Currently the device registry implementations
use a `NoopTenantInformationService` as there is no tenant information bean is
exposed in their `ApplicationConfig`. This has been fixed now.


sophokles73 (approved):
LGTM

PR #2641, opened 2021-05-04
[#2480] Implement and use Kafka consumer with own auto-commit

calohmn:
This is for #2480.
This adds an extended HonoKafkaConsumer that automatically commits partition offsets corresponding to the latest record per partition whose (asynchronous) processing has been marked as completed (provided all its preceding records are also completed).

This consumer is used for receiving commands in the Command Router.

In contrast to the default `enable.auto.commit` behaviour, identical offsets are skipped during successive commits. (They are recommitted after a certain time to prevent them from reaching their retention time.) This helps in case a corresponding partition just got deleted. Not trying to commit the offset then means the consumer won't block/retry for a long time in this case - see https://github.com/eclipse/hono/issues/2623#issuecomment-824906383.

### DIFF
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AsyncHandlingAutoCommitKafkaConsumer.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AsyncHandlingAutoCommitKafkaConsumer.java
@@ -1,0 +1,454 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.client.kafka.consumer;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Collection;
+import java.util.Deque;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.kafka.common.TopicPartition;
+import org.eclipse.hono.util.Pair;
+import org.eclipse.hono.util.Strings;
+
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.kafka.client.consumer.KafkaConsumerRecord;
+
+/**
+ * A Kafka consumer that automatically commits partition offsets corresponding to the latest record per
+ * partition whose (asynchronous) handling has been marked as completed.
+ * <p>
+ * A scenario that this consumer addresses is one of a received record taking quite long to be handled asynchronously
+ * while the consumer is closed and the application exits. With the standard <em>enable.auto.commit</em> behaviour,
+ * an offset commit for that record will already be done on closing the consumer so that another consumer of the
+ * consumer group won't read that record again, meaning the record potentially only gets partially handled.
+ * Using the auto-commit implementation in this class, only the offset of a record whose handling was marked as
+ * completed gets committed. This prevents incompletely handled records and thereby enables at-least-once semantics.
+ * <p>
+ * In terms of when offsets are committed, the behaviour is similar to the one used for a consumer with
+ * <em>enable.auto.commit</em>. Commits are done periodically (using <em>commitAsync</em>) and when a rebalance
+ * happens or the consumer is stopped (using <em>commitSync</em>). The periodic commit interval is defined via
+ * the standard <em>auto.commit.interval.ms</em> configuration property.
+ * <p>
+ * In order to not fall behind with the position of the committed offset vs. the last received offset, users of this
+ * class have to make sure that the record handling function, which provides the completion Future, is completed in time.
+ * <p>
+ * In contrast to the <em>enable.auto.commit</em> behaviour, identical offsets are skipped during successive commits.
+ * It is only after a period defined by <em>hono.offsets.skip.recommit.period.seconds</em> has elapsed, that such offsets
+ * are committed again. This is to make sure that such offsets don't reach their retention time, provided the recommit
+ * period is lower than the <em>offsets.retention.minutes</em> broker config value.
+ */
+public class AsyncHandlingAutoCommitKafkaConsumer extends HonoKafkaConsumer {
+
+    /**
+     * The name of the configuration property to define the period for which committing an already committed offset is
+     * skipped, until it may be recommitted if it is still the latest offset.
+     */
+    public static final String CONFIG_HONO_OFFSETS_SKIP_RECOMMIT_PERIOD_SECONDS = "hono.offsets.skip.recommit.period.seconds";
+    /**
+     * The default periodic commit interval.
+     */
+    public static final Duration DEFAULT_COMMIT_INTERVAL = Duration.ofSeconds(5);
+    /**
+     * The default period for which committing an already committed offset is skipped,
+     * until it may be recommitted if it is still the latest offset.
+     */
+    public static final Duration DEFAULT_OFFSETS_SKIP_RECOMMIT_PERIOD = Duration.ofMinutes(30);
+
+    private final long commitIntervalMillis;
+    private final long skipOffsetRecommitPeriodSeconds;
+    private final Map<TopicPartition, TopicPartitionOffsets> offsetsMap = new HashMap<>();
+    private final AtomicBoolean periodicCommitInvocationInProgress = new AtomicBoolean();
+
+    private Long periodicCommitTimerId;
+
+    /**
+     * Creates a consumer to receive records on the given topics.
+     * <p>
+     * Partition offsets returned by invoking the given supplier are committed periodically and
+     * when a rebalance happens. The periodic commit interval is taken from the
+     * <em>auto.commit.interval.ms</em> config value.
+     *
+     * @param vertx The Vert.x instance to use.
+     * @param topics The Kafka topic to consume records from.
+     * @param recordHandler The function to be invoked for each received record. The completion of its Future return
+     *                      value marks the record handling as finished so that the corresponding offset may be committed.
+     * @param consumerConfig The Kafka consumer configuration.
+     * @throws NullPointerException if any of the parameters is {@code null}.
+     * @throws IllegalArgumentException if the consumerConfig is missing a "group.id" entry.
+     */
+    public AsyncHandlingAutoCommitKafkaConsumer(
+            final Vertx vertx,
+            final Set<String> topics,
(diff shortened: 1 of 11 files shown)


calohmn:
@b-abel Can you also have a look?

b-abel:
> @b-abel Can you also have a look?

Sure. Will do.

calohmn:
I've pushed some changes:
I've made it more explicit that offsets are also committed when the consumer is stopped (done via the `onPartitionsRevoked` invocation then). And I've replaced usage of the term "processing" with "handling" in the code..

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AsyncHandlingAutoCommitKafkaConsumer.java
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    private static long getCommitInterval(final Map<String, String> consumerConfig) {
+        return Optional.ofNullable(consumerConfig.get(ConsumerConfig.AUTO_COMMIT_INTERVAL_MS_CONFIG))
+                .map(Long::parseLong)
+                .orElse(DEFAULT_COMMIT_INTERVAL.toMillis());
+    }
+
+    private static long getOffsetsRecommitSeconds(final Map<String, String> consumerConfig) {
+        return Optional.ofNullable(consumerConfig.get(CONFIG_HONO_OFFSETS_RECOMMIT_SECONDS))
```
b-abel, line 177: I would not mix this property into the map with the standard Kafka properties. They are usually filled by Spring or Quarkus based on a prefix and you would probably have to use non-obvious tricks to add custom properties into it.

calohmn: I meant this property to be set like the other Kafka properties:
````
hono:
  kafka:
    consumerConfig:
      bootstrap.servers: ${hono.kafka.bootstrap.servers}
      hono.offsets.skip.recommit.period.seconds: 600
````
It's a non-intrusive way to add such a property and it is something to be set for all consumers (using the above class) since it depends on the server `offsets.retention.minutes` config setting.
On the other hand I agree that it's not optimal to mix official and hono-defined properties on that level. I tried to at least make the distinction clear with the `hono.` prefix.
However, when thinking about an alternative place where to put such a property, there currently seems to be no particularly good place either. The `hono.command` section (see [here](https://github.com/eclipse/hono/blob/f69b2cb0d56b18a9288bb5b8e7cb5e7e95eb4c23/tests/src/test/resources/commandrouter/application.yml#L40)) contains only AMQP related properties.

I see there's a similar case with the `AbstractAtLeastOnceKafkaConsumer#pollTimeout` property, being defined in a `KafkaConsumerConfigProperties` field. But from what I see, there's currently no way to set this property in the `application.yml`.

File: clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/HonoKafkaConsumer.java [hunk 2]
```diff
@@ -69,19 +72,23 @@
 
     protected final Logger log = LoggerFactory.getLogger(getClass());
     protected final Vertx vertx;
-    protected final AtomicBoolean stopped = new AtomicBoolean();
+    protected final AtomicBoolean stopCalled = new AtomicBoolean();
```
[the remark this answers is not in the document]
calohmn: At the point in time when `stopCalled` is set to `true` (on the vert.x event loop thread), the consumer isn't stopped yet, just the `stop()` method has been called.
This distinction is relevant now because the variable is also checked from the Kafka polling thread, meaning it may be read between `stop()` getting invoked and the inner invocation of `consumer.close()` on the Kafka polling thread.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/KafkaBasedCommandConsumerFactoryImpl.java
```diff
@@ -92,12 +94,14 @@ public KafkaBasedCommandConsumerFactoryImpl(
...
-        commandHandler = new KafkaBasedMappingAndDelegatingCommandHandler(tenantClient, commandTargetMapper,
-                internalCommandSender, tracer);
+        commandHandler = new KafkaBasedMappingAndDelegatingCommandHandler(tenantClient, commandQueue,
+                commandTargetMapper, internalCommandSender, tracer);
         final Map<String, String> consumerConfig = kafkaConsumerConfig.getConsumerConfig("consumer");
         consumerConfig.put(ConsumerConfig.GROUP_ID_CONFIG, "cmd-router-group");
-        kafkaConsumer = new HonoKafkaConsumer(vertx, COMMANDS_TOPIC_PATTERN,
+        kafkaConsumer = new AsyncHandlingAutoCommitKafkaConsumer(vertx, COMMANDS_TOPIC_PATTERN,
                 commandHandler::mapAndDelegateIncomingCommandMessage, consumerConfig);
+        kafkaConsumer.setOnPartitionsAssignedHandler(
```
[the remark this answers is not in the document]
calohmn: I don't think intervening with the command processing via extra synchronization ("don't block the event loop") or waiting for some Futures added in between would be a good idea. And when "kafkaProducer.send()" has been invoked to forward the command on the internal command topic, there is no stopping that anyway.

> During re-balancing, the commands are still processed and consumers try to commit them too. Won't those commit fails during re-balancing?

During the rebalance, happening on the Kafka polling thread, the commands are indeed still processed on the event loop thread, but they are not committed there. The commands are just marked as "completed". The next commit happening on the Kafka polling thread will then take these completed commands into account when determining the offsets to be committed.

The scenario of a partition getting revoked (and afterwards not getting re-assigned) while a command is in the process of getting published on the internal command topic, might happen. But I think it's a narrow timeframe in which that can happen. And in case a command gets sent twice in such a case (the 2nd time by the consumer that the partition assignment has switched to), the duplicate command will indeed be ignored by the target protocol adapter (via #2596). So I don't think we should add further synchronization/locks in the command router for that.

calohmn:
@kaniyan I've pushed changes. I've also added a fix for the `KafkaCommandProcessingQueue.setCurrentlyHandledPartitions` not being involved with all current partitions. There is a new `onRebalanceDoneHandler` for that now, providing all current partitions. (That is difficult to test with a unit test because of the limited `MockConsumer` features.)

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AsyncHandlingAutoCommitKafkaConsumer.java
```diff
@@ -0,0 +1,449 @@
...
+                        lastSequentiallyCompletedOffset, lastCommitTime, topicPartition);
+                return Optional.of(lastSequentiallyCompletedOffset);
+            } else {
+                return Optional.empty();
+            }
+        }
+
+        private void cleanupAndUpdateLastCompletedOffset() {
+            while (Optional.ofNullable(queue.peek()).map(OffsetsQueueEntry::isHandlingComplete).orElse(false)) {
+                lastSequentiallyCompletedOffset = queue.remove().getOffset();
```
kaniyan, line 373: `cleanupAndUpdateLastCompletedOffset()` is invoked during several independent events (eg. commit at regular intervals, arrival of a new record and during re-balancing).
1.  In order to ensure that the correct item is being removed I think it's better to use `queue.remove(offsetsQueueEntry)` instead of invoking `queue.remove()`.
2. The operations `peek()` and `remove()` are not done atomically. Since there are several events that trigger this, wouldn't that affect the sequence of the offsets being committed. For example `cleanupAndUpdateLastCompletedOffset()` is also being invoked by a non _synchronized_ method namely `setRecordReceived(...)`

calohmn: The `cleanupAndUpdateLastCompletedOffset()` method is invoked from these methods (in the outer class):
- `setRecordReceived`
- `getOffsetsToCommit`
- `clearObsoleteTopicPartitionOffsets`
All of these are `synchronized`. And the used `OffsetsQueueEntry#handlingComplete` field is an `AtomicBoolean`. So I don't see any potential concurrency issues here.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AsyncHandlingAutoCommitKafkaConsumer.java
```diff
@@ -0,0 +1,449 @@
...
+
+    // synchronized because offsetsMap is accessed from vert.x event loop and kafka polling thread
+    private synchronized Map<TopicPartition, OffsetAndMetadata> getOffsetsToCommit() {
+        final Map<TopicPartition, OffsetAndMetadata> offsets = offsetsMap.entrySet().stream()
+                // map each to key/value Pair only if offsets needs to be committed
+                .flatMap(entry -> entry.getValue().getLastSequentiallyCompletedOffsetForCommit().stream()
+                        .map(uncommittedOffset -> Pair.of(entry.getKey(),
+                                new OffsetAndMetadata(uncommittedOffset + 1, ""))))
+                .collect(Collectors.toMap(Pair::one, Pair::two));
+        log.trace("getOffsetsToCommit: {}", offsets);
```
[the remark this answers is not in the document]
calohmn: I've removed that log line and used `getOffsetsDebugString` to limit the number of logged offsets in the other places where the offsets get logged.

clients/kafka-common/src/test/java/org/eclipse/hono/client/kafka/consumer/AsyncHandlingAutoCommitKafkaConsumerTest.java
```diff
@@ -0,0 +1,494 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(AsyncHandlingAutoCommitKafkaConsumerTest.class);
+
+    private static final String TOPIC = "test.topic";
+    private static final String TOPIC2 = "test.topic2";
+    private static final String TOPIC3 = "test.topic3";
+    private static final Pattern TOPIC_PATTERN = Pattern.compile(Pattern.quote("test.") + ".*");
+    private static final int PARTITION = 0;
+    private static final TopicPartition topicPartition = new TopicPartition(TOPIC, PARTITION);
+    private static final TopicPartition topic2Partition = new TopicPartition(TOPIC2, PARTITION);
+    private static final TopicPartition topic3Partition = new TopicPartition(TOPIC3, PARTITION);
```
kaniyan, line 72: Variables `topic3Partition` and `TOPIC3` are unused.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AsyncHandlingAutoCommitKafkaConsumer.java
b-abel: Yes, `AbstractAtLeastOnceKafkaConsumer#pollTimeout` is a similar case. The property is picked up when you use an `application.yml` containing this:
```
hono:
  kafka:
    pollTimeout: 200
    consumerConfig:
      bootstrap.servers: ${hono.kafka.bootstrap.servers}
```

So, the difference is that it is a separate member of the `KafkaConsumerConfigProperties` and not part of the configuration of the (native) Kafka consumer. Even though the Kafka client ignores properties that it does not know and only logs a warning,  it seems cleaner to me to keep Hono's properties and Kafka's properties separate.

calohmn: Ah ok, I only had a look at the quarkus implementation here for which this doesn't seem to be implemented yet.

In any case, I agree that we should separate the Hono and standard Kafka config properties. I think to achieve this properly, we should reconsider the property structure in general, e.g. by using a 
````
consumer:
  pollTimeout: 200
  skipOffsetRecommitPeriodSeconds: 600
  clientConfig: 
    bootstrap.servers: ${hono.kafka.bootstrap.servers}
    [...]
````
structure, possibly put e.g. under `hono.command.kafka`, to enable defining different properties for different consumer use cases.
I think this should be done in a follow-up PR then. Until then, I would prefer to keep the code here as is.

kaniyan (approved):
LGTM

Issue #2084, opened 2020-07-15
Prometheus message resource limit check performance issues


bordeuax:
**Production Problem**
Daily in production i am facing with regular CPU/memory spikes for Prometheus, as a results of such behavior we have messaging/connection timeouts and huge performance degradation of message resource limit check. 
**Particular use case**
A tenant pushing events/telemetry/command each 15 minutes via HTTP (i think this should be available for other adapters) number of devices  that doing connections are around 30-40. Prometheus cache timeout is 600s, 3 instances of HTTP adapter and 4 CPU core per pod (4 verticles). Prometheus has been configured to log all the queries into a file , for more info see [query log](https://prometheus.io/docs/guides/query-log/). 
**Prometheus query log**
```
{"httpRequest":{"clientIP":"100.108.0.17","method":"GET","path":"/api/v1/query"},"params":{"end":"2020-07-15T09:30:00.239Z","query":"floor(sum(increase(hono_messages_payload_bytes_sum{status=~\"forwarded|unprocessable\", tenant=\"tenant_id\"} [15d]) or hono_commands_payload_bytes_sum*0) + sum(increase(hono_commands_payload_bytes_sum{status=~\"forwarded|unprocessable\", tenant=\"tenant_id\"} [15d]) or hono_messages_payload_bytes_sum*0))","start":"2020-07-15T09:30:00.239Z","step":0},"stats":{"timings":{"evalTotalTime":3.700122281,"resultSortTime":0,"queryPreparationTime":0.002982689,"innerEvalTime":3.697125523,"execQueueTime":0.000013985,"execTotalTime":3.700148945}},"ts":"2020-07-15T09:30:03.939Z"}
[...]

PR #2643, opened 2021-05-05
Reduce load on Prometheus server when checking limits

sophokles73:
This addresses #2084

The Prometheus based resource limit checks were prone to generating a
lot of load on a Prometheus server under certain conditions.
In particular, the load would spike, if a protocol adapter to which many
devices of many different tenants were connected would crash (or be shut
down) and the devices would immediately re-connect to another instance
of the protocol adapter. In this case, for the first device per tenant
up to three queries would be issued to Prometheus' query API, depending
on whether the tenant's current limited resource values were already
available in the cache or not. For multiple devices from different
tenants connecting in parallel, this could lead to a high load of
queries on the Prometheus server.

The limit checks have been changed to reduce (if not prevent) these
spikes:
1. The Java Executor used for running the queries has been changed from
the default (multi-threaded) implementation to a single-threaded
implementation. This results in one query being issued to Prometheus at
a time (per adapter instance).
2. When invoking any of the methods to check if a limit is reached, the
methods now immediately return FALSE (passing the check) if the required
limited resource value for the tenant is not yet in the cache.
Regardless of this behavior, a task to load the value into the cache is
scheduled using the Executor in the background.

In addition to these changes, the query to determine the used data
volume for the current accounting period has been slightly optimized to
reduce the amount of data that needs to be processed on the Prometheus
server.

### DIFF
--- a/adapter-base/src/test/java/org/eclipse/hono/adapter/resourcelimits/PrometheusBasedResourceLimitChecksTest.java
+++ b/adapter-base/src/test/java/org/eclipse/hono/adapter/resourcelimits/PrometheusBasedResourceLimitChecksTest.java
@@ -12,13 +12,14 @@
  */
 package org.eclipse.hono.adapter.resourcelimits;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.ArgumentMatchers.anyString;
-import static org.mockito.ArgumentMatchers.argThat;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
@@ -30,11 +31,12 @@ import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.ZoneOffset;
-import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiFunction;
 
 import org.eclipse.hono.test.TracingMockSupport;
@@ -52,11 +54,11 @@ import org.junit.jupiter.api.Timeout;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 import com.github.benmanes.caffeine.cache.AsyncCache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
 import io.opentracing.Tracer;
-import io.opentracing.log.Fields;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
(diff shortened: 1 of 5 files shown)


adapter-base/src/test/java/org/eclipse/hono/adapter/resourcelimits/PrometheusBasedResourceLimitChecksTest.java
```diff
@@ -557,6 +557,39 @@ public void testMessageLimitUsesValueFromCache(final VertxTestContext ctx) {
...
+     */
+    @Test
+    @SuppressWarnings("unchecked")
+    public void testMessageLimitFallsBackToDefaultValueIfQueryStillRunning(final VertxTestContext ctx) {
+
+        when(dataVolumeCache.get(anyString(), any(BiFunction.class)))
+            .then(invocation -> {
+                final CompletableFuture<LimitedResource<Long>> result = new CompletableFuture<>();
+                return result;
+            });
```
kaniyan, line 574: It could be simplified as `.then(invocation -> new CompletableFuture<LimitedResource<Long>>());`

sophokles73: Anything else?

kaniyan:
>     2. When invoking any of the methods to check if a limit is reached, the
>         methods now immediately return FALSE (passing the check) if the required
>         limited resource value for the tenant is not yet in the cache.
>         Regardless of this behavior, a task to load the value into the cache is
>         scheduled using the Executor in the background.


In case of shutdown adapter scenario, it's good to have such solution to avoid overloading of the Prometheus server. In normal scenario, let us assume that devices under a tenant are denied to connect as the limit is exceeded. Those devices try to reconnect at regular short intervals. Lets us assume when they try to reconnect, the cache entry is already expired and the query is being executed. If I understand correctly, with this change, we don't wait until query execution and we allow those devices to connect (depending upon how long it takes for the query to execute). This case is also applicable for devices regularly trying to resend messages that have been failed due to exceeding message limit. From the customer POV, it seems to be confusing as sometimes the devices are able to connect and send messages even though resource limits are exceeded.

sophokles73:
> This case is also applicable for devices regularly trying to resend messages that have been failed due to exceeding message limit. From the customer POV, it seems to be confusing as sometimes the devices are able to connect and send messages even though resource limits are exceeded.

We already have that problem when devices try to send via adapter instance A (which denies the request due to a very up-to-date cache value) and then retries via adapter instance B (which allows the request due to an older cache value) ...

kaniyan:
> > This case is also applicable for devices regularly trying to resend messages that have been failed due to exceeding message limit. From the customer POV, it seems to be confusing as sometimes the devices are able to connect and send messages even though resource limits are exceeded.
> 
> We already have that problem when devices try to send via adapter instance A (which denies the request due to a very up-to-date cache value) and then retries via adapter instance B (which allows the request due to an older cache value) ...

But that is for a brief moment of time (around config time out interval) after that the adapters synchronize their cache data. Lets us assume that the connection duration and message limits are exceeded for a tenant. Once all the adapters have the actual data, no devices belonging to the tenant are allowed to connect or send messages. But in the approach in this PR, it's not the case, whenever cache times out and query is under execution the devices always are able to connect and send messages. 

I think whenever a query is under execution, if we could use the old cache data to validate instead of simply allowing the requests would address this situation.

sophokles73:
> But in the approach in this PR, it's not the case, whenever cache times out and query is under execution the devices always are able to connect and send messages.

They will be able to connect and send messages only until the query result is available and is put into the cache. We do the check if the message limit is exceeded for each message, so there should only be a short period of time where this is the case.

> I think whenever a query is under execution, if we could use the old cache data to validate instead of simply allowing the requests would address this situation.

That might be an option. I will need to take a look if the Caffeine cache supports that ...

sophokles73:
I will post another PR using a Caffeine `AsyncCacheLoader` for refreshing stale entries in the background.

PR #2647, opened 2021-05-07
[#2272] Add support to auto-provision gateways

kaniyan:


core/src/main/java/org/eclipse/hono/util/TenantObject.java
```diff
@@ -173,6 +181,7 @@ public TenantObject addTrustAnchor(final byte[] publicKey,
         trustedCa.put(TenantConstants.FIELD_PAYLOAD_PUBLIC_KEY, publicKey);
         trustedCa.put(TenantConstants.FIELD_PAYLOAD_KEY_ALGORITHM, publicKeyAlgorithm);
         trustedCa.put(TenantConstants.FIELD_AUTO_PROVISIONING_ENABLED, autoProvisioningEnabled);
+        trustedCa.put(TenantConstants.FIELD_AUTO_PROVISION_AS_GATEWAY, autoProvisioningAsGatewayEnabled);
```
sophokles73, line 184: this property is not relevant for the protocol adapters, is it?
I understand that we use the TenantObject as the vehicle to convey the tenant's configuration within the device registry's services, in particular it is used as the return value of *TenantInformationService.getTenant()*. However, *TenantInformationService* is a strictly registry internal service so to me it feels very strange that we use TenantObject here instead of the *Tenant* object that is used by the TenantManagementService. FMPOV we should keep the information that we expose to the adapters to the required minimum. IMHO we should therefore change *TenantInformationService.getTenant()* to return a *Tenant* instance instead of a *TenantObject* (which is intended to be consumed by adapters). WDYT?

kaniyan: At the very first sight I also thought the same seeing that the tenant information returns `TenantObject`. I also think that it should return `Tenant` instance. Currently tenant information service makes use of `TenantService` internally and in my opinion it should be `TenantManagementService`.   I already wanted to to address this in a separate PR as it involves changes in several other places too after finishing #2272.

kaniyan: I think it makes more sense to change the `TenantInformationService.getTenant()` to return `Tenant` instance and also to change the `AutowiredTenantInformationService` to make use of the `TenantManagmentService` instead of `TenantService` before further proceeding with this PR.  I will create a PR for that.

kaniyan: > @sophokles73 Currently I am working on the refactoring as we discussed above. What I noticed is that the edge way device provisioning uses `EventSender.sendEvent(TenantObject tenant ...)` to send auto-provisioned events. In order to get rid of the `TenantObject` there we need an alternative option for sending events without involving `TenantObject`. For me it seems that starting with change in `EventSender` would slow down the progress of #2272. What I can think of now, is to leave the usage of the `TenantObject` as it is only for the above case.
> 
> That means we need a workaround by converting `Tenant` to `TenantObject` as in the below code only for the sake of `EventSender.sendEvent(TenantObject tenant ...)`. And once the #2272 is over, we could come with a solution to over come that. Or do you have other suggestions?
> 
> ```
>  DeviceRegistryUtils.convertTenant(tenantId, tenant)
>                     .mapTo(TenantObject.class);
> ```

sophokles73: Wow, looks like we have opened a can of worms ;-)
I had already been thinking about refactoring the telemetry/event sender from *client-adapter* into a separate module which is not intended to be used by adapters only. As part of that we could introduce a kind of *internal* sender which does not require the information from the `TenantObject`. But I agree that it will be better to postpone that to a subsequent issue/PR.

kaniyan: @sophokles73 I have created a PR #2650 for changing the tenant information service to make use of `Tenant` and `TenantManagementService` as we discussed above. Would you mind taking a look?

kaniyan:
I have also updated this PR in accordance with the other #2650 and this PR also includes a commit from #2650.

sophokles73:
@kaniyan can you rebase?

kaniyan:
> @kaniyan can you rebase?

@sophokles73 I have rebased.

core/src/main/java/org/eclipse/hono/util/TenantConstants.java
```diff
@@ -66,6 +66,12 @@
      */
     public static final String FIELD_AUTO_PROVISIONING_ENABLED = "auto-provisioning-enabled";
 
+    /**
+     * The name of the property that indicates whether a unregistered device that authenticates with
+     * a client certificate should be auto-provisioned as a gateway. 
+     */
+    public static final String FIELD_AUTO_PROVISION_AS_GATEWAY = "auto-provision-as-gateway";
+
```
sophokles73, line 74: this is not needed anymore, or is it?

services/device-registry-base/src/test/java/org/eclipse/hono/service/management/device/AutoProvisioningTest.java
```diff
@@ -219,4 +197,57 @@ public void testDeviceRegistrationIsRemovedWhenAutoProvisionFails(final VertxTes
...
+        if (isGateway) {
+            // The property auto-provision-as-gateway is set to true
+            trustedCA.setAutoProvisioningAsGatewayEnabled(true);
+        }
+
+        when(deviceManagementService.createDevice(eq(tenantId), any(), any(), any()))
+                .thenReturn(Future.succeededFuture(OperationResult.ok(HttpURLConnection.HTTP_OK, Id.of(deviceId),
+                        Optional.empty(), Optional.empty())));
+        when(credentialsManagementService.updateCredentials(eq(tenantId), eq(deviceId), any(), any(),
+                any())).thenReturn(Future.succeededFuture(OperationResult.empty(HttpURLConnection.HTTP_CREATED)));
```
sophokles73, line 218: this should return HTTP_NO_CONTENT (204) if I am not mistaken ...

services/device-registry-base/src/test/java/org/eclipse/hono/service/management/device/AutoProvisioningTest.java
```diff
@@ -219,4 +197,57 @@ public void testDeviceRegistrationIsRemovedWhenAutoProvisionFails(final VertxTes
...
+        // GIVEN a tenant CA with auto-provisioning enabled
+        trustedCA.setAutoProvisioningEnabled(true);
+
+        if (isGateway) {
+            // The property auto-provision-as-gateway is set to true
+            trustedCA.setAutoProvisioningAsGatewayEnabled(true);
+        }
+
+        when(deviceManagementService.createDevice(eq(tenantId), any(), any(), any()))
+                .thenReturn(Future.succeededFuture(OperationResult.ok(HttpURLConnection.HTTP_OK, Id.of(deviceId),
```
sophokles73, line 215: this should return HTTP_CREATED (201) if I am not mistaken ...

services/device-registry-base/src/test/java/org/eclipse/hono/service/management/device/AutoProvisioningTest.java
```diff
@@ -219,4 +197,57 @@ public void testDeviceRegistrationIsRemovedWhenAutoProvisionFails(final VertxTes
...
+                .onComplete(ctx.succeeding(result -> {
+                    ctx.verify(() -> {
+                        // VERIFY that that the device/gateway has been registered.
+                        final ArgumentCaptor<Device> deviceCaptor = ArgumentCaptor.forClass(Device.class);
+                        verify(deviceManagementService).createDevice(eq(tenantId), any(),
+                                deviceCaptor.capture(), any());
+
+                        if (isGateway) {
+                            // VERIFY that a gateway has been provisioned by checking the relevant property
+                            assertThat(deviceCaptor.getValue().getAuthorities()
```
sophokles73, line 233:
```java
assertThat(deviceCaptor.getValue().getAuthorities()).contains(RegistryManagementConstants.AUTHORITY_AUTO_PROVISIONING_ENABLED);
```

services/device-registry-base/src/test/java/org/eclipse/hono/service/management/device/AutoProvisioningTest.java
```diff
@@ -219,4 +197,57 @@ public void testDeviceRegistrationIsRemovedWhenAutoProvisionFails(final VertxTes
...
+                                    .contains(RegistryManagementConstants.AUTHORITY_AUTO_PROVISIONING_ENABLED))
+                                    .isTrue();
+                        }
+
+                        //VERIFY that the correct credentials are stored
+                        verify(credentialsManagementService).updateCredentials(eq(tenantId), eq(deviceId), any(), any(),
```
sophokles73, line 239: since this an external service being invoked, we should make sure that the credentials passed into this method are of the expected type and have the expected properties.

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsApiTests.java
```diff
@@ -183,6 +184,65 @@ public void testGetCredentialsWithAutoProvisioning(final VertxTestContext ctx)
...
+        final JsonObject clientCtx = new JsonObject().put(CredentialsConstants.FIELD_CLIENT_CERT, cert.getEncoded());
+        final String authId = cert.getSubjectX500Principal().getName(X500Principal.RFC2253);
+
+        tenantId = getHelper().getRandomTenantId();
+        getHelper().registry
+                .addTenant(tenantId, tenant)
+                .compose(ok -> getClient()
+                        // WHEN getting credentials
+                        .get(tenantId, CredentialsConstants.SECRETS_TYPE_X509_CERT, authId, clientCtx, spanContext))
+                .onComplete(ctx.succeeding(result -> {
```
sophokles73, line 219: how about changing this into a `compose` invocation doing all the assertions and returning the output of
```
getHelper().registry.getRegistrationInfo(tenantId, result.getDeviceId());
```
and then do the final assertions in an *onComplete* block?
FMPOV that would improve readability because it prevents the nesting of the async processing ...

core/src/main/java/org/eclipse/hono/util/TenantConstants.java
kaniyan: This is currently used in the test classes `AbstractTenantServiceTest` and `FileBasedTenantServiceTest`.

If we don't want to this property to showup in `TenantObject` then we could remove that while mapping to `TenantObject` using `DeviceRegistryUtils.convertTenant()`. But I don't see any drawbacks why it should be removed.

tests/src/test/java/org/eclipse/hono/tests/registry/CredentialsApiTests.java
kaniyan: I have pushed the changes in accordance with the comments. Would you mind taking a look?

core/src/main/java/org/eclipse/hono/util/TenantConstants.java
sophokles73: IMHO we should *not* include the property in the TenantObject returned to protocol adapters because this is information that is *internal* to the registry and is of no concern to the adapters. This way we also prevent adapters to *suddenly* depend on this information ...

kaniyan: Now this property is not included in the `TenantObject` anymore. I have pushed the changes for it.

sophokles73 (approved):
LGTM

PR #2650, opened 2021-05-10
[#2272] Refactor tenant information service

kaniyan:
For more details refer to the [discussion thread](https://github.com/eclipse/hono/pull/2647#discussion_r628183627).

* `TenantInformation.getTenant(...)` returns a result that contains `Tenant` instead of `TenantObject`.
* `AutowiredTenantInformationService` internally makes use of `TenantManagementService` instead of `TenantService`.

### DIFF
--- a/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/tenant/TenantInformationService.java
+++ b/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/tenant/TenantInformationService.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019, 2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2019, 2021 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
@@ -13,8 +13,7 @@
 package org.eclipse.hono.deviceregistry.service.tenant;
 
 import org.eclipse.hono.service.management.Result;
-import org.eclipse.hono.util.TenantObject;
-import org.eclipse.hono.util.TenantResult;
+import org.eclipse.hono.service.management.tenant.Tenant;
 
 import io.opentracing.Span;
 import io.vertx.core.Future;
@@ -53,15 +52,10 @@ public interface TenantInformationService {
      *            it may set tags and use this span as the parent for any spans created in this method.
      *
      * @return A future indicating the outcome of the operation.
-     *             The <em>status</em> will be
-     *             <ul>
-     *             <li><em>200 OK</em> if a tenant with the given ID is registered.
-     *             The <em>payload</em> will contain the tenant's configuration information.</li>
-     *             <li><em>404 Not Found</em> if no tenant with the given identifier exists.</li>
-     *             </ul>
-     *
+     *         If succeeds the future contains the tenant information. Otherwise, the future will fail with
+     *         a {@link org.eclipse.hono.client.ServiceInvocationException} containing a corresponding status code.
      * @throws NullPointerException if any argument is {@code null}.
      */
-     Future<TenantResult<TenantObject>> getTenant(String tenantId, Span span);
+     Future<Tenant> getTenant(String tenantId, Span span);
 
 }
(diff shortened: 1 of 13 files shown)


services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/tenant/AutowiredTenantInformationService.java
```diff
@@ -29,17 +28,18 @@
 import io.vertx.core.Future;
 
 /**
- * A default implementation of {@link TenantInformationService} that uses embedded {@link TenantService} to verify if tenant exists.
+ * A default implementation of {@link TenantInformationService} that uses embedded {@link TenantManagementService} to
```
sophokles73, line 31: How about leaving this class as it is but instead create a new `TenantServiceImplementation` that uses `TenantManagementService` and does not rely on Spring auto-wiring anymore?

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/tenant/TenantInformationService.java
```diff
@@ -56,12 +56,12 @@
...
      *             <li><em>200 OK</em> if a tenant with the given ID is registered.
-     *             The <em>payload</em> will contain the tenant's configuration information.</li>
+     *             The <em>payload</em> will contain the tenant information.</li>
      *             <li><em>404 Not Found</em> if no tenant with the given identifier exists.</li>
      *             </ul>
      *
      * @throws NullPointerException if any argument is {@code null}.
      */
-     Future<TenantResult<TenantObject>> getTenant(String tenantId, Span span);
+     Future<OperationResult<Tenant>> getTenant(String tenantId, Span span);
```
sophokles73, line 65: Is it really useful to return an `OperationResult` here? In the past it has merely caused confusion about how to interpret it correctly. FMPOV it is much more straightforward to either succeed the future with a *Tenant* or fail it with a *ServiceInvocationException* ...

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/tenant/AutowiredTenantInformationService.java
kaniyan: Ok, but still anyway the `AutoWiredTenantInformationService` need to updated with `TenantManagementService` as we use now `Tenant` instead of `TenantObject`.

sophokles73: obviously

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/tenant/TenantInformationService.java
kaniyan: I have pushed changes for the above comment .  Also created a `DefaultTenantInformationService` without using `@Autowired` and the device registries are configured use this new class.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/AutoProvisioning.java
```diff
@@ -208,17 +204,17 @@ private AutoProvisioning() {
...
      *
-     * @param tenantConfig The tenant configuration to check if auto-provisioning is enabled or not.
+     * @param tenant The tenant information to check if auto-provisioning is enabled or not.
      * @param certificate The client certificate that devices used for authentication.
      *                    If the certificate is {@code null} then {@code false} is returned.
      * @param span The active OpenTracing span for this operation.
      * @return {@code true} if auto-provisioning is enabled.
      */
-    private static boolean isEnabled(final TenantObject tenantConfig, final X509Certificate certificate,
+    private static boolean isEnabled(final Tenant tenant, final X509Certificate certificate,
```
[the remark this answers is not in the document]
sophokles73: That seems to be a much better method name :+1:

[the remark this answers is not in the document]
kaniyan: I have updated with the above method name and pushed those changes.

sophokles73 (approved):
LGTM

PR #2652, opened 2021-05-11
Use Caffeine AsyncCacheLoader for populating cache in the background

sophokles73:
This addresses #2084

The Prometheus based resource limit checks were prone to generating a
lot of load on a Prometheus server under certain conditions.
In particular, the load would spike, if a protocol adapter to which many
devices of many different tenants were connected would crash (or be shut
down) and the devices would immediately re-connect to another instance
of the protocol adapter. In this case, for the first device per tenant
up to three queries would be issued to Prometheus' query API, depending
on whether the tenant's current limited resource values were already
available in the cache or not. For multiple devices from different
tenants connecting in parallel, this could lead to a high load of
queries on the Prometheus server.

The limit checks have been changed to reduce (if not prevent) these
spikes:
1. The Java Executor used for running the queries has been changed from
the default (multi-threaded) implementation to a single-threaded
implementation. This results in one query being issued to Prometheus at
a time (per adapter instance).
2. When invoking any of the methods to check if a limit is reached, the
methods now immediately return FALSE (passing the check) if the required
limited resource value for the tenant is not yet in the cache.
Regardless of this behavior, a task to load the value into the cache is
scheduled using the Executor in the background.
3. The cache implementation now refreshes values in the background that
have reached their expiration time but are still of interest. The cache
returns the existing (possibly stale) value while the refresh task is
running, thus making sure that checks can always be performed based on
data from the cache.

In addition to these changes, the query to determine the used data
volume for the current accounting period has been slightly optimized to
reduce the amount of data that needs to be processed on the Prometheus
server.


adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
```diff
@@ -472,4 +484,49 @@ private ClientConfigProperties commandResponseSenderConfig() {
...
+
+            final var webClient = WebClient.create(vertx, webClientOptions);
+            final var cacheTimeout = Duration.ofSeconds(config.getCacheTimeout());
+            final Caffeine<Object, Object> builder = Caffeine.newBuilder()
+                    // make sure we run one Prometheus query at a time
+                    .executor(Executors.newSingleThreadExecutor())
+                    .initialCapacity(config.getCacheMinSize())
+                    .maximumSize(config.getCacheMaxSize())
+                    .expireAfterWrite(cacheTimeout)
+                    .refreshAfterWrite(cacheTimeout);
```
ben-manes, line 520: This will result in expiration still taking precedence, because it indicates that the entry is invalid to return. On the other hand, a refresh is when the entry is stale but still acceptable to use. The two are meant to be used together so that active entries are reloaded and hide the latency, while inactive ones are allowed to fade away.

This data freshness is the same as a loaf of bread used for lunch:
1. just backed: enjoy!
2. stale: eat & order for delivery
3. moldy: throw away
  a) having a sandwich? go buy break and be hungry
  b) switching to a salad? no reason to keep or get more

You might therefore use `cacheTimeout / 2` or similar as a threshold. If a read request occurs within the refresh interval then it will be reloaded asynchronously. But if the entry has passed the expiration threshold, the cache won't serve that entry and penalize the caller as if it was absent.

You can use `Caffeine.ticker(t)` to control logical time, and use Guava's `FakeTicker` in your tests. That would allow you to write a test case demonstrating those different scenarios and observe Caffeine's behavior.

adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
```diff
@@ -472,4 +484,49 @@ private ClientConfigProperties commandResponseSenderConfig() {
...
+            webClientOptions.setDefaultPort(config.getPort());
+            webClientOptions.setTrustOptions(config.getTrustOptions());
+            webClientOptions.setKeyCertOptions(config.getKeyCertOptions());
+            webClientOptions.setSsl(config.isTlsEnabled());
+
+            final var webClient = WebClient.create(vertx, webClientOptions);
+            final var cacheTimeout = Duration.ofSeconds(config.getCacheTimeout());
+            final Caffeine<Object, Object> builder = Caffeine.newBuilder()
+                    // make sure we run one Prometheus query at a time
+                    .executor(Executors.newSingleThreadExecutor())
```
ben-manes, line 516: You might prefer a daemon thread pool. If you have an orderly server shutdown, you might also register it explicitly.

adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
sophokles73: Got it. Thanks for the hint :-) May I ask how you became aware of our using of Caffeine in Hono?

adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
sophokles73: Not sure what you mean by
> you might also register it explicitly

What would need to be done for that?

adapter-base/src/main/java/org/eclipse/hono/adapter/resourcelimits/ConnectedDevicesAsyncCacheLoader.java
```diff
@@ -0,0 +1,101 @@
...
+                } else {
+                    final long maxConnections = tenant.getResourceLimits().getMaxConnections();
+                    LOG.trace("connection limit for tenant [{}] is [{}]", tenant.getTenantId(), maxConnections);
+
+                    if (maxConnections == TenantConstants.UNLIMITED_CONNECTIONS) {
+                        span.log(Map.of(Fields.MESSAGE, "no connection limit configured"));
+                        result.complete(new LimitedResource<>(null, 0L));
+                    } else {
+                        final String queryParams = String.format("sum(%s{tenant=\"%s\"})", METRIC_NAME_CONNECTIONS, key.getTenantId());
+                        executeQuery(queryParams, null)
```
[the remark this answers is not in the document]
sophokles73: thanks for spotting

adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
ben-manes: I use GitHub’s global search to monitor recent activity. That way I can try to catch misunderstandings early before users get frustrated and also get feedback on where improvements are needed. I saw this PR and skimming it caught this setting.

sophokles73:
@kaniyan would you mind taking another look?

adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
ben-manes: In my work code base, we use Guava’s ServiceManager to start/stop the application. An orderly server shutdown could call the executor’s shutdown to stop processing new requests. This isn’t necessary, just an FYI.

The non daemon thread could cause an orderly shutdown to hang as it won’t die until killed. So you either change the type or shutdown cleanly.

kaniyan (approved):
LGTM

adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
sophokles73: Ok got it. I had already changed the *ThreadFactory* to return daemon threads so this should work now.

PR #2653, opened 2021-05-12
AMQP adapter: Fix command not getting released in case device already disconnected

calohmn:
This fixes a potential NPE in `HonoProtonHelper.isLinkOpenAndConnected`.

It occurs for example when a command message got sent to an AMQP device and there is a timeout waiting for the disposition update, while the transport of the underlying AMQP connection already got closed (with the AMQP link still locally marked as open however). The NPE in that case leads to the command message not getting released.

````
  at org.eclipse.hono.util.HonoProtonHelper.isLinkOpenAndConnected(HonoProtonHelper.java:183)
  at org.eclipse.hono.adapter.amqp.VertxBasedAmqpProtocolAdapter.lambda$onCommandReceived$65(VertxBasedAmqpProtocolAdapter.java:998)
  at io.vertx.core.impl.VertxImpl$InternalTimerHandler.handle(VertxImpl.java:923)
````

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
+++ b/core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
@@ -154,6 +154,8 @@ public final class HonoProtonHelper {
 
     /**
      * Checks if a link is established.
+     * <p>
+     * Note that this only applies to a link which has originally been created by the local peer.
      *
      * @param link The link to check.
      * @return {@code true} if the link has been established.
@@ -178,14 +180,26 @@ public final class HonoProtonHelper {
         if (link != null && link.isOpen()) {
             if (link.getSession() != null && link.getSession().getConnection() != null
                     && !link.getSession().getConnection().isDisconnected()) {
-                return true;                
+                return true;
             }
-            final String localLinkAddress = link instanceof ProtonSender ? link.getTarget().getAddress()
-                    : link.getSource().getAddress();
-            LOG.debug("{} link [address: {}] is locally open but underlying transport is disconnected",
-                    link instanceof ProtonSender ? "sender" : "receiver", localLinkAddress);
+            LOG.debug("{} link [source: {}, target: {}] is locally open but underlying transport is disconnected",
+                    link instanceof ProtonSender ? "sender" : "receiver", getRemoteOrLocalSourceAddress(link), getRemoteOrLocalTargetAddress(link));
         }
         return false;
+    }
+
+    private static String getRemoteOrLocalSourceAddress(final ProtonLink<?> link) {
+        if (link != null && link.getRemoteSource() != null) {
+            return link.getRemoteSource().getAddress();
+        }
+        return link != null && link.getSource() != null ? link.getSource().getAddress() : null;
+    }
+
+    private static String getRemoteOrLocalTargetAddress(final ProtonLink<?> link) {
+        if (link != null && link.getRemoteTarget() != null) {
+            return link.getRemoteTarget().getAddress();
+        }
+        return link != null && link.getTarget() != null ? link.getTarget().getAddress() : null;
     }
 
     /**

### CODE REVIEW

File: core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java [hunk 1]
```diff
@@ -178,16 +180,22 @@ public static boolean isLinkOpenAndConnected(final ProtonLink<?> link) {
...
             if (link.getSession() != null && link.getSession().getConnection() != null
                     && !link.getSession().getConnection().isDisconnected()) {
-                return true;                
+                return true;
             }
-            final String localLinkAddress = link instanceof ProtonSender ? link.getTarget().getAddress()
-                    : link.getSource().getAddress();
-            LOG.debug("{} link [address: {}] is locally open but underlying transport is disconnected",
-                    link instanceof ProtonSender ? "sender" : "receiver", localLinkAddress);
+            LOG.debug("{} link [source: {}, target: {}] is locally open but underlying transport is disconnected",
```
sophokles73, line 185: the labels in the log statement should probably be *remote source* and *remote target* in order to better reflect the origin of the information.

calohmn: I mainly want to output *any* kind of relevant source or target *address* here to distinguish the link (whether source or target is relevant depends on whether the link was initiated by the local or remote peer - the old code assumed it was the local peer).
In case the link had been established before, the interesting part is the remote source/target. 
What is not taken into account above: In case a locally initiated link had not been established, it would be the local source/target.
I've adapted the commit now to use `getRemoteOrLocalSourceAddress` and `getRemoteOrLocalTargetAddress` methods.

sophokles73: FMPOV this doesn't make much sense. IMHO we should distinguish between a sender and a receiver link. For the former, we should log the remote and local source address, for the latter we should log the remote and local target address.
If you do not want to distinguish between the types of links then we should include all four addresses in the log statement.

calohmn: > IMHO we should distinguish between a sender and a receiver link. For the former, we should log the remote and local source address, for the latter we should log the remote and local target address.

Apart from including the remote info, the previous code does this kind of distinguishing - with source and target switched (see list below).
https://github.com/eclipse/hono/blob/fbe015b3f0d535773f04a59be3572d274446a05d/core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java#L183-L184

But: We also need to distinguish between a link initiated by the local peer and one initiated by the remote peer. (The case of a remote-triggered link is the scenario in which the NPE occurs here.)
That results in 4 cases. Since that would make the code here quite complicated, I chose to just log the source and target addresses. 

The 4 cases along with the relevant data would be
- sender
  - initiated by local peer: remote target (if null, i.e. link not established yet, local target)
  - initiated by remote peer: remote source
- receiver
  - initiated by local peer: remote source (if null, i.e. link not established yet, local source)
  - initiated by remote peer: remote target

sophokles73: And you think anybody reading a log statement like that will then know which of the four cases it is? Whatever ...

sophokles73 (commented):
if you want to merge it like this, be my guest ...

core/src/main/java/org/eclipse/hono/util/HonoProtonHelper.java
calohmn: I think so, yes. Together with the information about the component the log appears in, it should be clear enough which kind of link it is (just sender/receiver + address should usually be enough). It's just a debug log message anyway.

Examples:
````
AMQP adapter: sender link [source: null, target: registration/myTenant]
AMQP adapter: sender link [source: command, target: null]
AMQP adapter: receiver link [source: registration/myTenant/myDevice, target: null]
AMQP adapter: receiver link [source: null, target: null]        (-> link on anonymous-relay address)
````

sophokles73:
@calohmn this seems to be relevant for 1.6.x as well, isn't it? If so, can you also please commit the fix to the 1.6.x branch (which is still *in service*)?

calohmn:
@sophokles73 The `isLinkOpenAndConnected` method was only added as recently as February (see 1da34ea7e1adc3e9a1c92b83a71371a438c6171a) and doesn't exist in the 1.6.x branch. So this fix isn't needed in 1.6.x.

PR #2655, opened 2021-05-14
Enable reporting of jaeger client metrics

calohmn:
This is done via the added `jaeger-micrometer` dependency.

This includes for example `jaeger_tracer_reporter_spans_total` metrics. Analyzing such metrics with value `dropped` can help in determining whether the reporter max queue size (`JAEGER_REPORTER_MAX_QUEUE_SIZE` env variable, default 100) should potentially be increased.

### DIFF
--- a/services/command-router/pom.xml
+++ b/services/command-router/pom.xml
@@ -73,6 +73,10 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
           <groupId>io.jaegertracing</groupId>
           <artifactId>jaeger-client</artifactId>
         </dependency>
+        <dependency>
+          <groupId>io.jaegertracing</groupId>
+          <artifactId>jaeger-micrometer</artifactId>
+        </dependency>
       </dependencies>
     </profile>
   </profiles>
--- a/adapters/base-spring/pom.xml
+++ b/adapters/base-spring/pom.xml
@@ -151,6 +151,10 @@
           <groupId>io.jaegertracing</groupId>
           <artifactId>jaeger-client</artifactId>
         </dependency>
+        <dependency>
+          <groupId>io.jaegertracing</groupId>
+          <artifactId>jaeger-micrometer</artifactId>
+        </dependency>
       </dependencies>
     </profile>
 
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -692,6 +692,12 @@
         <scope>runtime</scope>
       </dependency>
       <dependency>
+        <groupId>io.jaegertracing</groupId>
+        <artifactId>jaeger-micrometer</artifactId>
+        <version>${jaeger.version}</version>
+        <scope>runtime</scope>
+      </dependency>
+      <dependency>
         <groupId>com.squareup.okhttp3</groupId>
         <artifactId>okhttp</artifactId>
         <version>4.9.0</version>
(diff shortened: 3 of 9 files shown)


services/command-router/pom.xml
```diff
@@ -73,6 +73,10 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
           <groupId>io.jaegertracing</groupId>
           <artifactId>jaeger-client</artifactId>
         </dependency>
+        <dependency>
+          <groupId>io.jaegertracing</groupId>
+          <artifactId>jaeger-micrometer</artifactId>
+        </dependency>
```
sophokles73, line 79: IMHO we also need to add the dependency to the Quarkus based version ...

calohmn: It looks to me as if jaeger client metrics are already supported via the `quarkus-jaeger` dependency and can be enabled by setting `quarkus.jaeger.metrics.enabled` to `true`.

I think this property should be added in the default quarkus config in the IoT Packages chart. I'll create a separate PR there then.

sophokles73: My understanding is that the Quarkus Jaeger extension does not per se pull in the dependency but can make use of it, if the Quarkus Metrics extension is being used. Since we do not use that extension, I do not think that the jar is being pulled in as a quick run of the Maven Dependency plugin in for `command-router-quarkus` seems to confirm:
```bash
mvn dependency:list -Pbuild-docker-image,metrics-prometheus,jaeger | grep jaeger
[INFO]    io.quarkus:quarkus-jaeger:jar:1.11.3.Final:compile -- module quarkus.jaeger (auto)
[INFO]    io.jaegertracing:jaeger-core:jar:1.5.0:runtime -- module jaeger.core (auto)
[INFO]    io.jaegertracing:jaeger-thrift:jar:1.5.0:runtime -- module jaeger.thrift (auto)
```
and
```bash
mvn dependency:list -Pbuild-native-image,metrics-prometheus,jaeger | grep jaeger
[INFO]    io.quarkus:quarkus-jaeger:jar:1.11.3.Final:compile -- module quarkus.jaeger (auto)
[INFO]    io.jaegertracing:jaeger-core:jar:1.5.0:runtime -- module jaeger.core (auto)
[INFO]    io.jaegertracing:jaeger-thrift:jar:1.5.0:runtime -- module jaeger.thrift (auto)
```

calohmn: The `jaeger-micrometer` dependency only contains one class - `MicrometerMetricsFactory`.
In the quarkus jaeger code, I see a corresponding `QuarkusJaegerMicrometerFactory` class in the quarkus-jaeger module itself getting used instead (see [here](https://github.com/quarkusio/quarkus/blob/59b7449ffb90304ef7c17cf9175ea2d10b74b80a/extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/QuarkusJaegerMicrometerFactory.java#L16), instantiated [here](https://github.com/quarkusio/quarkus/blob/b96973dcbcbff76ec6842448ba802e1269b5ca53/extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/JaegerDeploymentRecorder.java#L39)). So I don't think the `jaeger-micrometer` dependency is needed for quarkus.

sophokles73 (approved):
LGTM

Issue #2556, opened 2021-03-18
Disable server-side support of specific TLS/SSL cipher suites by configuration


jnbsi:
As SSL cipher suites' level of security decreases over time and SSL engines might not get adapted promptly it can become necessary to be able to deactivate server-side support for certain SSL cipher suites explicitly.
I.e. there could be an additional configuration property on each service/adapter that takes a list of cipher suite names which will be disabled on the corresponding SSL endpoint.

sophokles73:
I would be more in favor of specifying a white list, i.e. a list of supported cipher suites, instead. Would that work for you?

jnbsi:
Yes, that would work as well!

PR #2658, opened 2021-05-17
Add support for white listing cipher suites

sophokles73:
This addresses #2556 

The Hono components now support configuration properties for setting
supported TLS cipher suites. The cipher suites can be configured
separately for both the endpoints exposed by the components as well as
the clients used for accessing service endpoints exposed by other
components.


core/src/main/java/org/eclipse/hono/connection/impl/ConnectionFactoryImpl.java
```diff
@@ -404,6 +404,12 @@ private void addTlsTrustOptions(final ProtonClientOptions clientOptions) {
                 protocols.add(p);
             });
             clientOptions.setEnabledSecureTransportProtocols(protocols);
+
+            config.getSupportedCipherSuites().stream()
+                .forEach(suiteName -> {
```
kaniyan, line 409: Instead of `stream().forEach()` simply `forEach()` can be used.

service-base/src/main/java/org/eclipse/hono/service/AbstractServiceBase.java
```diff
@@ -437,6 +437,12 @@ protected final void addTlsKeyCertOptions(final NetServerOptions serverOptions)
             });
             serverOptions.setEnabledSecureTransportProtocols(protocols);
 
+            getConfig().getSupportedCipherSuites().stream()
+                .forEach(suiteName -> {
```
kaniyan, line 441: Instead of `stream().forEach()` simply `forEach()` can be used.

sophokles73:
I have pushed changes ...

kaniyan (approved):
LGTM

PR #2662, opened 2021-05-17
Use HonoKafkaConsumer for KafkaApplicationClient

calohmn:
Currently, the `KafkaApplicationClient` implementation uses the `AbstractAtLeastOnceKafkaConsumer` class.
There are several downsides to this:
The `AbstractAtLeastOnceKafkaConsumer` performs synchronous partition offset commits after each `poll` invocation. For once, offset commits could be not needed at all if the `KafkaApplicationClient` is used with no `group.id` or some random group id only used for that consumer instance. On the other hand the frequent commits seem unncessary when compared to the standard behaviour of the vert.x KafkaConsumer with auto-commit enabled, where offsets are committed periodically (every 5s by default) and when a rebalance occurs.
With the standard vert.x KafkaConsumer, the automatic poll loop involves
- putting a task on the vert.x event loop thread to trigger the `consumer.poll()` invocation (which itself is done on the separate polling thread)
- letting the records get handled on the vert.x event loop thread (invoking the handler set on the vert.x KafkaConsumer for each of them)

Since auto-commits are done in the beginning of the `consumer.poll()` invocation, and since `poll()` and handler invocations are all triggered one after the other on the vert.x event loop thread, only offsets for records will be committed, whose handler has finished already (synchronously). That means there is already an at-least-once guarantee.
For the underlying KafkaConsumer, this is also described in the [javadoc](https://kafka.apache.org/27/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html).
<details>
<summary>(details)</summary>

> Note: Using automatic offset commits can also give you "at-least-once" delivery, but the requirement is that you must consume all data returned from each call to poll(Duration) before any subsequent calls, or before closing the consumer. If you fail to do either of these, it is possible for the committed offset to get ahead of the consumed position, which results in missing records. The advantage of using manual offset control is that you have direct control over when a record is considered "consumed."

</details>

Another downside is that when the auto.offset.reset config is set to LATEST and no offsets were committed yet it may be that after the `start()` future has succeeded any records published *before* the consumer has gotten its subsequent rebalance may not be received by the consumer. This can be a problem if the `start()` future in the application triggers some records to get published which should get received by the consumer.

---

This PR replaces usage of the `AbstractAtLeastOnceKafkaConsumer` with the `HonoKafkaConsumer` class, which is only a thin wrapper of the vert.x Kafka consumer, using no explicit offset commit handling and letting `start()` only get finished after the first rebalance.

I've simplified the `KafkaApplicationClient` interface, removing the error scenario descriptions concerning the explicit commit errors. I currently see no scenario where the consumer will stop functioning or where the remote peer causes the consumer to stop. In case of network errors, the consumer will anyway try to reconnect under the hood, so that an "unconnected" state isn't even easily detectable from our code.

I've also removed the handler retry in case of a handler RuntimeException. I think this could produce unwanted sideeffects and is also something which we don't do with the AMQP based implementation.


clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/HonoKafkaConsumer.java
```diff
@@ -238,7 +262,18 @@ void setKafkaConsumerSupplier(final Supplier<Consumer<String, Buffer>> supplier)
...
                     .orElseGet(() -> KafkaConsumer.create(vertx, consumerConfig, String.class, Buffer.class));
-            kafkaConsumer.handler(recordHandler);
+            kafkaConsumer.handler(record -> {
+                if (respectTtl && KafkaRecordHelper.isTtlElapsed(record.headers())) {
+                    onRecordHandlerSkippedForExpiredRecord(record);
+                } else {
+                    try {
+                        recordHandler.handle(record);
+                    } catch (final Exception e) {
+                        log.warn("error handling record [topic: {}, partition: {}, offset: {}, headers: {}]",
```
b-abel, line 272: FMPOV skipping a record in an _at least once_ scenario is an essential point. It would be good to provide more options to react on that, e.g. put it into the traces.

b-abel (commented):
`AbstractAtLeastOnceKafkaConsumer` (and some corresponding tests) is no longer used and should be removed.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/HonoKafkaConsumer.java
calohmn: This try-catch block is really meant as a last-resort safety net so that an unhandled exception here doesn't prevent the next polling operation. I think it should be the responsibility of the code that uses the HonoKafkaConsumer to make sure this doesn't happen, or to handle it properly, potentially also creating trace output.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/HonoKafkaConsumer.java
```diff
@@ -111,6 +113,8 @@
      * @param recordHandler The handler to be invoked for each received record.
      * @param consumerConfig The Kafka consumer configuration.
      * @throws NullPointerException if any of the parameters is {@code null}.
+     * @throws IllegalArgumentException if the consumerConfig contains an "enable.auto.commit" property with value
+     *             "true" but no "group.id" property.
```
kaniyan, line 117: How about using `{@value ConsumerConfig#GROUP_ID_CONFIG}` and `{@value ConsumerConfig#ENABLE_AUTO_COMMIT_CONFIG}` in the _JavaDoc_?

calohmn: @kaniyan I've adapted the Javadocs accordingly.

kaniyan (approved):
LGTM

PR #2666, opened 2021-05-18
Make the AMQP network config non-mandatory in the adapters

calohmn:
The protocol adapter AMQP consumers on the internal command address will now only be used if the AMQP consumer is actually configured.
Needed for a setup that only uses a Kafka messaging network.


sophokles73 (approved):
LGTM

PR #2673, opened 2021-05-20
Add Kafka support in CLI and example application to release notes.

b-abel:
This adds to the release notes that the CLI and the example application now support Kafka.


sophokles73 (approved):
LGTM

Thanks for updating ...

PR #2674, opened 2021-05-21
Fix jaeger RemoteControlledSampler not working in quarkus native images

calohmn:
Jaeger tracing currently doesn't work when using quarkus-native images and a jaeger sampling configuration that uses the sampling strategies defined in the Jaeger Collector component (i.e. using the RemoteControlledSampler).

This seems caused by the reflection used in conjunction with the `io.jaegertracing.internal.samplers.HttpSamplingManager` class.
There, a `SamplingStrategyResponse` object is deserialized from the JSON response using gson, which makes use of reflection to do so. That means that, at least with the current `quarkus-jaeger` code, a reflection config has to be defined in places where the JaegerTracer is used (see also https://github.com/quarkusio/quarkus/issues/10402#issuecomment-845717069).

This PR adds that. 

Hope would be that https://github.com/quarkusio/quarkus/issues/10402 gets solved in a way that doesn't make such a configuration necessary, so that this reflection config can be removed again with a quarkus update.

Relates to:
https://github.com/eclipse/packages/pull/245

---
I've triggered a test build here also without the `jaeger` maven profile. That causes some warnings (e.g. `WARNING: Could not resolve io.jaegertracing.internal.samplers.http.OperationSamplingParameters for reflection configuration.`), but the build succeeds.

### DIFF
--- a/service-base-quarkus/pom.xml
+++ b/service-base-quarkus/pom.xml
@@ -25,6 +25,10 @@
   <description>Base classes for implementing Hono services using Quarkus.</description>
   <url>https://www.eclipse.org/hono</url>
 
+  <properties>
+    <jaeger-reflection-config/>
+  </properties>
+
   <dependencyManagement>
     <dependencies>
       <dependency>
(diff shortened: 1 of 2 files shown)


adapter-base-quarkus/src/main/resources/META-INF/native-image/org.eclipse.hono/hono-adapter-base-quarkus/native-image.properties
```diff
@@ -8,4 +8,4 @@
 # http://www.eclipse.org/legal/epl-2.0
 #
 # SPDX-License-Identifier: EPL-2.0
-Args = -H:ReflectionConfigurationResources=${.}/reflection-config.json, --initialize-at-run-time=io.netty.internal.tcnative.SSL
+Args = -H:ReflectionConfigurationResources=${.}/reflection-config.json, -H:+AllowIncompleteClasspath --initialize-at-run-time=io.netty.internal.tcnative.SSL
```
sophokles73, line 11: why do we need to allow an incomplete classpath here?

calohmn: Otherwise there build be such an error when building the native image without using the `jaeger` maven profile:
````
Error: Error parsing reflection configuration in jar:file:/project/lib/org.eclipse.hono.hono-adapter-base-quarkus-1.9.0-SNAPSHOT.jar!/META-INF/native-image/org.eclipse.hono/hono-adapter-base-quarkus/reflection-config.json:
Could not resolve io.jaegertracing.internal.samplers.http.OperationSamplingParameters for reflection configuration. To allow unresolvable reflection configuration, use option -H:+AllowIncompleteClasspath
````

Comment (sophokles73) [reply adapter-base-quarkus/src/main/resources/META-INF/native-image/org.eclipse.hono/hono-adapter-base-quarkus/native-image.properties:11]: I see. On the other hand, it will prevent the build from detecting missing classes that are being referenced during build time, i.e. such errors will only become apparent during runtime. But that is probably not such a big issue as most problems that we are facing are related to missing classes that are referred to via reflection only anyway ....

File: adapter-base-quarkus/src/main/resources/META-INF/native-image/org.eclipse.hono/hono-adapter-base-quarkus/reflection-config.json [hunk 2]
```diff
@@ -13,5 +13,70 @@
     "name" : "org.eclipse.hono.adapter.resourcelimits.DataVolumeAsyncCacheLoader",
     "allDeclaredMethods" : true,
     "allPublicMethods" : true
+  },
```
sophokles73, line 16: it should be sufficient to only add the reflection config to the *service-base-quarkus* module as *adapter-base-quarkus* depends on it and thus pulls it in anyway

adapter-base-quarkus/src/main/resources/META-INF/native-image/org.eclipse.hono/hono-adapter-base-quarkus/native-image.properties
calohmn: I agree that it's better to not have the AllowIncompleteClasspath option in there.
I've adapted the commit now so that it isn't needed anymore.

adapter-base-quarkus/src/main/resources/META-INF/native-image/org.eclipse.hono/hono-adapter-base-quarkus/reflection-config.json
calohmn: Good point, I've adapted the commit accordingly.

sophokles73 (approved):
LGTM

sophokles73:
@calohmn how about putting this into a 1.8.1 as well?

calohmn:
@sophokles73 Sure, will do.

PR #2677, opened 2021-05-26
Adapt connection limit strategy to VM being used

### LINKED ISSUE #2675: Quarkus native image based adapters compute wrong max number of connections
The `MemoryBasedConnectionLimitStrategy` that is used by the protocol adapters determines the recommended maximum number of connections per adapter instance based on the amount of memory that the adapter needs to run and the memory required for each connection.
The memory to run the adapter depends on whether the adapter runs on a standard JVM or on the Substrate VM. The native images run on the Substrate VM and require a lot less memory to run than the JVM variant.
When configuring a native image based adapter container with less memory than the minimum amount of memory required to run the adapter (which is hard coded into the adapter), then startup of the container fails because the maximum number of connections is determined to be 0.

sophokles73:
The protocol adapters use the MemoryBasedConnectionLimitStrategy to
determine the number of connections that can be established with devices
based on the memory available to the adapter process.

The adapters currently assume to be running on a standard Java VM and
therefore assess the minimum amount of memory required for running to be
100MB (excluding memory allocated for device connections).

When using the Quarkus based native images, the memory footprint of the
adapters is significantly lower. This is now reflected in the MQTT and
AMQP adapters which first check the type of VM they are running on and
then adapt the minimum required memory accordingly.

Fixes #2675

### DIFF
--- a/adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
@@ -105,14 +105,19 @@ public final class VertxBasedAmqpProtocolAdapter extends AbstractProtocolAdapter
 
     // These values should be made configurable.
     /**
-     * The minimum amount of memory that the adapter requires to run.
+     * The minimum amount of memory (bytes) that the adapter requires to run on a standard Java VM.
      */
-    private static final int MINIMAL_MEMORY = 100_000_000; // 100MB: minimal memory necessary for startup
+    private static final int MINIMAL_MEMORY_JVM = 100_000_000;
+    /**
+     * The minimum amount of memory (bytes) that the adapter requires to run on a Substrate VM (i.e. when running
+     * as a GraalVM native image).
+     */
+    private static final int MINIMAL_MEMORY_SUBSTRATE = 35_000_000;
 
     /**
-     * The amount of memory required for each connection.
+     * The amount of memory (bytes) required for each connection.
      */
-    private static final int MEMORY_PER_CONNECTION = 20_000; // 20KB: expected avg. memory consumption per connection
+    private static final int MEMORY_PER_CONNECTION = 20_000;
 
     /**
      * The AMQP server instance that maps to a secure port.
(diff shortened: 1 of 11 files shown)


kaniyan:
I see that the connection limit strategy is also being used to set the maximum DTLS connections accepted by COAP adapter. Anyway the stale connections are evicted first if max limit is reached. COAP adapter test config still contain `maxConnections` configured and may be that can be removed too to make use of the memory strategy.


Quarkus native images are also used in Sandbox deployment and may be `maxConnections` to be removed from the AMQP, MQTT and COAP adpaters config there as well so that the memory strategy get employed.

sophokles73:
> Quarkus native images are also used in Sandbox deployment and may be maxConnections to be removed from the AMQP, MQTT and COAP adpaters config there as well so that the memory strategy get employed.

We can do that once we have release 1.9.0 and these changes are in effect.

adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -105,14 +105,19 @@
 
     // These values should be made configurable.
     /**
-     * The minimum amount of memory that the adapter requires to run.
+     * The minimum amount of memory (bytes) that the adapter requires to run on a standard Java VM.
      */
-    private static final int MINIMAL_MEMORY = 100_000_000; // 100MB: minimal memory necessary for startup
+    private static final int MINIMAL_MEMORY_JVM = 100_000_000;
```
dejanb, line 110: Should we move these to the abstract adapter or even limit strategy class, so we don't have to repeat it for every adapter?

sophokles73: That would, of course, make sense if the values would be the same for all adapters by definition. However, in general, I guess these are adapter (type) specific, aren't they?

dejanb: I supposed these were reasonable defaults for most of the adapters, that could be overridden if needed. Anyhow, the PR looks good to me.

kaniyan (approved):
LGTM

PR #2680, opened 2021-05-27
Catch exception publishing command to MQTT device

calohmn:
Most notably catch an IllegalStateException thrown by MqttEndpointImpl.publish() if the connection is
already disconnected.

Up to now this was misleadingly written as an "Connection not accepted yet" error in the trace, with such an exception being logged:
````
WARN org.eclipse.hono.tracing.TracingHelper An unexpected error occurred!
java.lang.IllegalStateException: Connection not accepted yet
  at io.vertx.mqtt.impl.MqttEndpointImpl.checkConnected(MqttEndpointImpl.java:734)
  at io.vertx.mqtt.impl.MqttEndpointImpl.write(MqttEndpointImpl.java:712)
  at io.vertx.mqtt.impl.MqttEndpointImpl.publish(MqttEndpointImpl.java:458)
  at io.vertx.mqtt.impl.MqttEndpointImpl.publish(MqttEndpointImpl.java:440)
  at io.vertx.mqtt.impl.MqttEndpointImpl.publish(MqttEndpointImpl.java:50)
  at org.eclipse.hono.adapter.mqtt.AbstractVertxBasedMqttProtocolAdapter$MqttDeviceEndpoint.lambda$onCommandReceived$37(AbstractVertxBasedMqttProtocolAdapter.java:1532)
  at io.vertx.core.Future.lambda$onSuccess$0(Future.java:150)
  at io.vertx.core.impl.FutureImpl.dispatch(FutureImpl.java:105)
  at io.vertx.core.impl.FutureImpl.onComplete(FutureImpl.java:83)
  at io.vertx.core.Future.onSuccess(Future.java:148)
  at org.eclipse.hono.adapter.mqtt.AbstractVertxBasedMqttProtocolAdapter$MqttDeviceEndpoint.onCommandReceived(AbstractVertxBasedMqttProtocolAdapter.java:1531)
  at org.eclipse.hono.adapter.mqtt.AbstractVertxBasedMqttProtocolAdapter$MqttDeviceEndpoint.lambda$createCommandConsumer$30(AbstractVertxBasedMqttProtocolAdapter.java:1458)
````

### DIFF
--- a/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
+++ b/adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java
@@ -1414,9 +1414,7 @@ public abstract class AbstractVertxBasedMqttProtocolAdapter<T extends MqttProtoc
                     ? String.format("gateway [%s], device [%s]", authenticatedDevice.getDeviceId(), context.deviceId())
                     : String.format("device [%s]", context.deviceId());
 
-            final Promise<Integer> resultPromise = Promise.promise();
-            endpoint.publish(publishTopic, errorJson.toBuffer(), subscription.getQos(), false, false, resultPromise);
-            return resultPromise.future()
+            return publish(publishTopic, errorJson.toBuffer(), subscription.getQos())
                     .onSuccess(msgId -> {
                         log.debug("published error message [packet-id: {}] to {} [tenant-id: {}, MQTT client-id: {}, QoS: {}, topic: {}]",
                                 msgId, targetInfo, subscription.getTenant(), endpoint.clientIdentifier(),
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: adapters/mqtt-vertx-base/src/main/java/org/eclipse/hono/adapter/mqtt/AbstractVertxBasedMqttProtocolAdapter.java [hunk 1]
```diff
@@ -1534,43 +1534,53 @@ protected final void onCommandReceived(
...
+        }
+
+        private Future<Integer> publish(final String topic, final Buffer payload, final MqttQoS qosLevel) {
+            final Promise<Integer> publishSentPromise = Promise.promise();
+            try {
+                endpoint.publish(topic, payload, qosLevel, false, false, publishSentPromise);
+            } catch (final Exception e) {
+                publishSentPromise.fail(!endpoint.isConnected()
+                        ? new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE, "connection to device already closed")
+                        : new ServerErrorException(HttpURLConnection.HTTP_INTERNAL_ERROR, e));
```
kaniyan, line 1581: Isn't it better to check if device is connected or not even before publishing instead of catching the exception and then categorizing it?

calohmn: The case that the endpoint may not be connected here is already rather exceptional (means the disconnect happened right *after* the command was received by the adapter but *before* the extra checks before forwarding the command to the device were completed). Therefore I prefer the solution above to adding an extra `isConnected` check (which is synchronized in `MqttEndpointImpl`) for each `publish` invocation here.

calohmn:
The added `publish` method is now also used when publishing error messages to the device.

kaniyan (approved):
LGTM

Issue #2601, opened 2021-04-12
Update Hono Documentation to include Kafka


fkaltner:
With the upcoming Kafka support the Hono documentation should be adapted, e.g. the CLI documentation, Getting Started and all other places where the messaging network is used.

b-abel:
Regarding the _Getting Started_ guide: the simplest solution would probably be to make a complete copy of the guide with adaptions to Kafka. The two documents can always be integrated later if it becomes apparent that both variants of messaging should be promoted permanently. WDYT?

fkaltner:
>  The two documents can always be integrated later if it becomes apparent that both variants of messaging should be promoted permanently.

@sophokles73 Maybe I missed something, butI thought both variants actually were supposed to coexist permanently?

sophokles73:
FMPOV there is no reason to discontinue the AMQP support. However, we should still have a *preferred* setup that we use by default and which we should then also use in the Getting Started guide, right?

b-abel:
I am not sure now: can we all agree on the proposal to create two separate documents?

I think there needs to be some sort of end-to-end documentation for each type of messaging system that shows how to use the overall system with clients. The existing Getting Started Guide FMPOV does a great job at this. The differences between AMQP-based and Kafka-based messaging are significant enough that I think it makes sense to have two separate documents for them. 
How to name them and where and how prominently they are displayed can still be changed at any time.

fkaltner:
If the documents do differ significantly there's no reason no to have them 😃 

Maybe I misunderstood what you wrote. I got the impression the argument was to have a (mostly) duplicated copy for now and postpone the effort for integrating both documents later, since the assumption was that one copy may become obsolete...

sophokles73:
@b-abel @fkaltner can this be closed?

b-abel:
FMPOV, yes.

PR #2682, opened 2021-05-28
[#2601] Add Getting Started guide with Kafka based messaging.

b-abel:
This adds an adapted Getting Started guide for Kafka-based messaging. The diagram showing the containers involved in an example deployment is also adapted.
~It references the documentation of deploying Hono with Kafka in the _IoT Packages_ project, which is included in a [PR](eclipse/packages#237) in the repository there. This PR should be merged after eclipse/packages#237.~


b-abel:
This PR is currently blocked by #2683 and eclipse/packages#237.

b-abel:
@sophokles73 I'm interested in your opinion on the question raised in https://github.com/eclipse/packages/pull/237#issuecomment-849560483. Which of the following options do you prefer? 

1. Investing time to configure Kafka so that it allows northbound clients to connect with TLS or plaintext depending on the port (as it is currently the case with the qdrouter) or 
2. support only TLS connections for Kafka and document (in this guide here) how to configure northbound clients to connect with TLS. It should be quite easy to provide a second `values.yaml` file with the Helm chart that can be used as a "profile" to deploy a plaintext setup instead.

I would prefer option 2. I don't see very much value in enabling TLS and plaintext at the same time. The Kafka deployment on the Sandbox on the other hand could be plaintext only because FMPOV it is mainly used for the Getting Started guide, which currently only uses plaintext connections.

sophokles73:
@b-abel I agree, we should start with option 2.

b-abel:
This is now ready to be reviewed.

b-abel:
Note that merging this PR will instantly (i.e. with the next run of the homepage deployment) make the content available on Hono's landing page.

b-abel:
I assume that Kafka-based messaging will be "fully supported" from version 1.10.0 of Hono. Therefore I suggest merging this PR with the release of that version.

site/homepage/content/getting-started-kafka/index.md
```diff
@@ -0,0 +1,440 @@
...
+kubectl get secrets eclipse-hono-kafka-jks --template="{{index .data \"kafka.truststore.jks\" | base64decode}}" -n hono > $KAFKA_TRUSTSTORE_PATH
+~~~
+
+and then start the client (make sure to replace `my-tenant`):
+
+~~~sh
+# in directory where the hono-cli-*-exec.jar file has been downloaded to
+export MY_TENANT=my-tenant
+export KAFKA_IP=$(kubectl get service eclipse-hono-kafka-0-external --output="jsonpath={.status.loadBalancer.ingress[0]['hostname','ip']}" -n hono)
+java -jar hono-cli-*-exec.jar --hono.kafka.commonClientConfig.bootstrap.servers=$KAFKA_IP:9092 --hono.kafka.commonClientConfig.security.protocol=SASL_SSL --hono.kafka.commonClientConfig.sasl.jaas.config="org.apache.kafka.common.security.scram.ScramLoginModule required username=\"hono\" password=\"hono-secret\";" --hono.kafka.commonClientConfig.sasl.mechanism=SCRAM-SHA-512 --spring.profiles.active=receiver,kafka --tenant.id=$MY_TENANT --hono.kafka.commonClientConfig.ssl.truststore.location=$KAFKA_TRUSTSTORE_PATH --hono.kafka.commonClientConfig.ssl.truststore.password=honotrust --hono.kafka.commonClientConfig.ssl.endpoint.identification.algorithm=""
```
sophokles73, line 254: this looks very complicated :-( Are you sure that it still reflects the command client client's current state/requirements?

b-abel: Alternatively, we could skip all the parameters for the authentication and use the sandbox profile from the CLI that already contains them as defaults. Would you prefer that?

sophokles73: Wasn't that the overall idea when introducing the profile(s)?

b-abel: Please have a look at my latest commit.

sophokles73 (approved):
LGTM

PR #2684, opened 2021-05-28
Update Spring Boot config in CLI.

b-abel:
In current Spring Boot versions the property "spring.profiles" is deprecated and should be replaced by "spring.config.activate.on-profile". 
This PR does that change in the application.yml of the CLI.


sophokles73 (approved):
LGTM

PR #2687, opened 2021-05-31
Deprecate file based device registry implementation

### LINKED ISSUE #2657: Deprecate file-based Device Registry
With #2587 the device registry can be used without providing an external database. This allows to use it instead of the file-based registry. FMPOV the file-based can be deprecated now. I would remove it in a future version to reduce the amount of code that needs to be maintained. This should be done once the JDBC-based registry has been used in the Helm chart as the default for a while.

sophokles73:
The file based registry had been devised as a simple, single-instance
only implementation which persists data to the local file system. The
same can now be achieved using the JDBC based registry configured with
an embedded H2 database. Maintaining an additional code base for the
same purpose therefore does not seem reasonable anymore.

Fixes #2657


sophokles73:
@b-abel would you also mind taking a look?

tests/pom.xml
```diff
@@ -415,9 +415,14 @@ Test cases are run against Docker images of Hono server + (Apache Qpid Dispatch
       </activation>
       <properties>
         <hono.mongodb.disabled>true</hono.mongodb.disabled>
-        <hono.deviceregistry.image>hono-service-device-registry-file</hono.deviceregistry.image>
-        <hono.deviceregistry.resources.folder>deviceregistry</hono.deviceregistry.resources.folder>
+        <hono.postgres.disabled>true</hono.postgres.disabled>
+        <hono.deviceregistry.image>hono-service-device-registry-jdbc</hono.deviceregistry.image>
```
b-abel, line 419: What about renaming the profile "device-registry-file" to something like "device-registry-embedded" or "device-registry-embedded-db"? The old name still suggests the file-based implementation is used. As far as I can tell, the profile is not referenced anywhere in the project, so the impact should be minimal anyway.

b-abel: The profile name should correspond to the property that activates it. Can we change the value of "hono.deviceregistry.type" to "embedded"?

sophokles73: I deliberately kept the original value in order to not require any configuration changes in other places like the CI build definition. The data is still being persisted to the the local file system so I thought it to (still) be an appropriate name. The name of the profile is also only relevant within the Hono build process so I do not think that it has much of an overall impact anyway.

kaniyan (approved):
LGTM

Issue #2540, opened 2021-03-15
Refactor client module so that it does not assume a certain component as client


fkaltner:
See https://github.com/eclipse/hono/pull/2534#pullrequestreview-611886500

I think it makes also sense to convert `AdapterConfigurationSupport` to an interface, as well as `ProtocolAdapterProperties`(only the part which is actually required). This would also enable us to give them proper names illustrating what is actually required for configuration for the client.

NB: This issue should not be started before the Kafka migration itself is completed (at the time of this writing this means not before 1.7.0 is released) as we might run into merge hell otherwise...

sophokles73:
An obvious candidate seems to be to factor out the telemetry client into a separate module because the device registry for example uses requires it to publish events. The Command Router, on the other hand, also needs to consume commands and needs to query the registry services but is (obviously) not a protocol adapter. We could therefore simply use a less opinionated name for the *adapter* client module or we could factor out the command and registry clients into dedicated modules as well.

WDYT? @calohmn @fkaltner @kaniyan

fkaltner:
I think the main problem was already addressed by bccad71b6737e62d6c6865136b0a413a0abdc4b9 making the reference to ProtocolAdapterProperties obsolete. But the separate modules do also make sense 👍 as the naming of the imports still was a bit odd...

sophokles73:
@calohmn @kaniyan I do have two more PRs on hold for factoring out the C&C and the Device Registry client(s) into separate modules. FMPOV we should go all the way now and resolve the (still) existing *adapter* client module into dedicated *registry* and *command* client modules as well. WDYT?

calohmn:
@sophokles73 Ok, good. There will probably be a conflict with #2711 - I can adapt that PR after your PR got merged then.

kaniyan:
> resolve the (still) existing _adapter_ client module into dedicated _registry_ and _command_ client modules as well.

+1

sophokles73:
closed via #2721

PR #2693, opened 2021-06-01
Move telemetry client classes into separate module

sophokles73:
This addresses #2540

In order to better reflect such cases, the classes for sending Telemetry
and Event messages have been moved into dedicated telemetry,
telemetry-amqp and telemetry-kafka modules.

Subsequent PRs can be used to factor out the registry and C&C clients into
dedicated modules if desirable.


clients/telemetry/src/main/java/org/eclipse/hono/client/telemetry/EventSender.java
```diff
@@ -12,7 +12,7 @@
  */
 
 
-package org.eclipse.hono.adapter.client.telemetry;
+package org.eclipse.hono.client.telemetry;
```
kaniyan, line 15: How about updating copyright notice with current year?

clients/telemetry/src/main/java/org/eclipse/hono/client/telemetry/TelemetrySender.java
```diff
@@ -12,7 +12,7 @@
  */
 
 
-package org.eclipse.hono.adapter.client.telemetry;
+package org.eclipse.hono.client.telemetry;
```
kaniyan, line 15: How about updating copyright notice with current year?

clients/telemetry/src/main/java/org/eclipse/hono/client/telemetry/EventSender.java
sophokles73: well, I haven't changed the classes but merely renamed the package. But I am glad to adapt the year as well ...

clients/telemetry/src/main/java/org/eclipse/hono/client/telemetry/TelemetrySender.java
kaniyan: The same is also applicable to `ConnectionEventProducer` and `HonoEventConnectionEventProducerTest`.

sophokles73: I have adapted the headers accordingly ...

kaniyan (approved):
LGTM

PR #2697, opened 2021-06-01
Remove support for disabling modification of data

### LINKED ISSUE #2696: Remove modificationEnabled config property from MongoDB based registry
The property had originally been introduced with the file based registry in order to prevent users of the Sandbox installation to modify other people's data. In the MongoDB based registry this feature is not really required FMPOV. Apart from that, it is not implemented 100% correctly anyway, so I think we should simply remove it altogether.

sophokles73:
The MongoDB based registry has been changed to no longer support the
modificationEnabled configuration variables which could be used to
prevent existing data from being changed.

The feature did not make much sense in production environments where
access to data should be authorized at the tenant level instead.

Fixes #2696


kaniyan (approved):
LGTM

Issue #2665, opened 2021-05-18
Send event notification after successful device/gateway auto-provisioning


kaniyan:
This is to extend support to send an empty notification event with _content-type_ as `application/vnd.eclipse-hono-device-provisioning-notification` once a device or gateway is successfully auto-provisioned. For more information refer to the [discussion thread](https://github.com/eclipse/hono/issues/2272#issuecomment-723094832) and [documentation](https://www.eclipse.org/hono/docs/api/event/#device-provisioning-notification) on device provisioning notification.

kaniyan:
What should be done if sending an auto-provisioning event notification fails even after specified number of retries?
1. As I understood, in many cases auto-provisioning may not be useful until the north-bound side (e.g ditto) is notified with an provisioning event. FMPOV sending an auto-provisioning event is also a part of the auto-provisioning process. If sending an event fails, then auto-provisioning should fail too. In that case, we should revert all the changes that we made during the auto-provisioning process (i.e remove device registration/credentials).

2. In case of auto-provisioning implementation for edge devices, a property `autoProvisioningNotificationSent`, which is part of the status section in device registration information, is being employed to ensure that a provisioning event is sent at least once. When sending a provisioning event fails, _the auto-provisioning method still returns a failed future_ but auto-provisioning related changes are _not_ being reverted (i.e remove device registration/credentials) and the `autoProvisioningNotificationSent` remains _false_. Every time when devices connect and protocol adapters assert device registration, the device registry checks for the value of the property `autoProvisioningNotificationSent`. If it is _false_, it sends a provisioning event notification and updates this property to _true_ and persists the device registration information. I am not aware of the background why the changes are not being reverted (i.e remove device registration/credentials) and rather making use of a property to resend the event. As far as I understood, if we revert those changes when sending an event fails, then we don't need this property to ensure sending of events and also no need to check this property every time a device registration is being asserted by protocol adapters.  If auto-provisioning of devices and gateways is to follow the same pattern, then credentials record should also contain a status section involving property `autoProvisioningNotificationSent` similar to device registration. When every time credentials are retrieved by protocol adapters, this property should be checked to resend provisioning event. After re-sending a event, credentials should be updated to reflect the actual status.

[...]

PR #2699, opened 2021-06-01
[#2665] Send auto-provisioning event when a device/gateway is auto-provisioned

kaniyan:


kaniyan:
The build fails due to the bug #2700.

kaniyan:
> The build fails due to the bug #2700.

I have pushed a commit to fix this bug. @sophokles73 Would you mind taking a look?

sophokles73 (approved):
LGTM

Issue #2705, opened 2021-06-07
Command Router: endless "Enable Command Routing" loop


calohmn:
We've encountered a scenario, where an "Enable Command Routing" invocation in the Command Router (see #2576) resulted in an endless loop, trying to invoke the command consumer registration for a large list of tenants again and again.

The scenario included a few subsequent "Enable Command Routing" requests with a lot of tenants (several hundred entries all in all, also caused by #2704 in connection with already deleted test tenants).
This resulted in many "get Tenant" invocations, which eventually led to a `org.eclipse.hono.client.ServerErrorException: no credit available for sending request` exception.
The problem is that this caused the corresponding tenant to be again put into the queue of tenants for which "get Tenant" and "register command consumer" was to be invoked (as part of a retry mechanism).
As there is no delay in processing implemented here, the situation could never recover, leading to constant high CPU usage and errors in processing the single incoming "register command consumer" Command Router API requests for yet uncached tenants.
Also the Jaeger Agent was put under high load, leading to most of the tracing spans to get dropped.
(This made it difficult to analyze the original exception, as there is no log output here).

An indication of the problem in the Jaeger UI was that there was no "re-enable command routing for tenants" entry listed in the operations combo box, as this span never got finished.

sophokles73:
@calohmn If I understand correctly, then the root issue is that there are (many) test tenants that do not exist anymore and which are therefore endlessly re-scheduled in the loop, right?

calohmn:
@sophokles73 No, the non-existing tenants result in `ClientErrorExceptions`, meaning there is no retry for that tenant.

There are 2 problems FMPOV: 
- There is no limit to the retry mechanism, so that any re-occuring 500 or 503 error for example will result in endless retries.
- There is no delay in the tenants queue processing, meaning that there may be a "no credit available" error if there are lots of (yet uncached) tenants in the queue. And there is no delay in between retries. This would be relevant in case the device registry AMQP connection is temporarily not there for example.

sophokles73:
@calohmn are you working on this already or should I take a shot?

calohmn:
@sophokles73 I started with some ideas about adding a retry delay.

<details>
<summary>details</summary>

~~~
    @Override
    public Future<CommandRouterResult> enableCommandRouting(final List<String> tenantIds, final Span span) {
        Objects.requireNonNull(tenantIds);
        final List<String> tenantIdsToUse = new LinkedList<>(tenantIds);
[...]

PR #2709, opened 2021-06-07
Prevent endless loop when re-activating command routing

sophokles73:
This addresses #2705

Actually, there has not been an endless loop as far as I can tell but the list of tenant IDs to re-activate routing for had grown larger and larger over time ...

### DIFF
--- a/services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
+++ b/services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
@@ -42,6 +42,7 @@ import org.slf4j.LoggerFactory;
 
 import io.opentracing.References;
 import io.opentracing.Span;
+import io.opentracing.SpanContext;
 import io.opentracing.Tracer;
 import io.opentracing.log.Fields;
 import io.vertx.core.CompositeFuture;
@@ -212,27 +213,29 @@ public class CommandRouterServiceImpl implements CommandRouterService, HealthChe
         tenantsToEnable.addAll(tenantIds);
         if (isProcessingRequired) {
             final Span processingSpan = tracer.buildSpan("re-enable command routing for tenants")
-                    .addReference(References.FOLLOWS_FROM, span.context())
+                    .addReference(References.CHILD_OF, span.context())
                     .start();
-            processTenantQueue(new ConcurrentHashSet<>(), processingSpan);
+            processingSpan.log(Map.of("number of tenant IDs", tenantIds.size()));
+            // make sure to report the span in a timely fashion
+            // so that it can be seen in the tracing back end
+            processingSpan.finish();
+            processTenantQueue(new ConcurrentHashSet<>(), processingSpan.context());
         }
         return Future.succeededFuture(CommandRouterResult.from(HttpURLConnection.HTTP_NO_CONTENT));
     }
 
-    private void processTenantQueue(final Set<String> processedTenants, final Span parentSpan) {
-        final String tenantId = tenantsToEnable.poll();
+    private void processTenantQueue(final Set<String> processedTenants, final SpanContext tracingContext) {
+        final String tenantId = tenantsToEnable.pollFirst();
         if (tenantId == null) {
-            parentSpan.finish();
             return;
         }
         context.runOnContext(go -> {
             if (processedTenants.contains(tenantId)) {
-                parentSpan.log(Map.of(
-                        Fields.MESSAGE, "skipping tenant, already processed ...",
-                        TracingHelper.TAG_TENANT_ID.getKey(), tenantId));
+                LOG.debug("skipping tenant [tenant-id: {}], already processed ...", tenantId);
+                processTenantQueue(processedTenants, tracingContext);
             } else {
                 final Span span = tracer.buildSpan("re-enable command routing for tenant")
-                        .addReference(References.CHILD_OF, parentSpan.context())
+                        .addReference(References.FOLLOWS_FROM, tracingContext)
                         .withTag(TracingHelper.TAG_TENANT_ID, tenantId)
                         .start();
                 tenantClient.get(tenantId, span.context())
@@ -247,14 +250,14 @@ public class CommandRouterServiceImpl implements CommandRouterService, HealthChe
                         if (t instanceof ServerErrorException) {
                             // add to end of queue in order to retry at a later time
                             span.log("marking tenant for later re-try to create command consumer");
-                            tenantsToEnable.add(tenantId);
+                            tenantsToEnable.addLast(tenantId);
                         }
                     })
                     .onComplete(r -> {
                         span.finish();
+                        processTenantQueue(processedTenants, tracingContext);
                     });
             }
-            processTenantQueue(processedTenants, parentSpan);
         });
     }
 
--- a/services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
+++ b/services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
@@ -248,6 +248,7 @@ public class CommandRouterServiceImpl implements CommandRouterService, HealthChe
                         TracingHelper.logError(span, "failed to create command consumer", t);
                         if (t instanceof ServerErrorException) {
                             // add to end of queue in order to retry at a later time
+                            LOG.info("failed to create command consumer", t);
                             span.log("marking tenant for later re-try to create command consumer");
                             tenantsToEnable.addLast(tenantId);
                         }
--- a/services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
+++ b/services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
@@ -243,8 +243,12 @@ public class CommandRouterServiceImpl implements CommandRouterService, HealthChe
 
         final var attempt = tenantsToEnable.pollFirst();
         if (attempt == null) {
-            reenabledTenants.clear();
-            LOG.debug("finished re-enabling of command routing");
+            // at this point there might still be pending (re-try) tasks on the event loop,
+            // thus we need to wait for those to have finished before declaring victory
+            if (tenantsInProcess.isEmpty()) {
+                reenabledTenants.clear();
+                LOG.debug("finished re-enabling of command routing");
+            }
         } else {
             tenantsInProcess.add(attempt.one());
             final long delay = calculateDelayMillis(attempt.two());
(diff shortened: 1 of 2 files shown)


services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
```diff
@@ -247,14 +250,14 @@ private void processTenantQueue(final Set<String> processedTenants, final Span p
                         if (t instanceof ServerErrorException) {
```
calohmn, line 250: I'm still not sure whether there aren't kinds of ServerErrorExceptions here where we _don't_ want to retry adding the command consumer indefinitely.

I guess most scenarios would be fine. E.g. also for a "amqp:unauthorized-access - link disallowed by local policy" or "amqp:not-found - Node not found" failure when trying to open the command consumer receiver link, ClientServiceExceptions are thrown.

But I think it is a bit risky to make the assumption here that any kind of ServerErrorExceptions (e.g. status 500) is resolvable and stops reoccurring here at some point.
When using Kafka for example, there are no replies to the consumer creation request like "amqp:unauthorized-access" which we would map to a ClientServiceException.
If a tenant-specific Kafka topic has been deleted at some point while the tenant still exists, creation of the Kafka-based command consumer will fail with an UNAVAILABLE ServerErrorException. To keep retrying here, checking for the existence of the topic would cost unnecessary Kafka server requests.

sophokles73: Do you have a suggestion for a better approach?

calohmn: I would limit the number of retries per specific tenant overall.
And I would keep a list of tenants for which processing failed and trigger the retry-operation concerning these tenants after a certain delay.
See https://github.com/eclipse/hono/issues/2705#issuecomment-855740463.

calohmn: Ok, good.

sophokles73: I have pushed corresponding changes ...

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
```diff
@@ -65,7 +69,7 @@
     private final DeviceConnectionInfo deviceConnectionInfo;
     private final MessagingClient<CommandConsumerFactory> commandConsumerFactories;
     private final Tracer tracer;
-    private final Deque<String> tenantsToEnable = new LinkedList<>();
+    private final Deque<Pair<String, Integer>> tenantsToEnable = new LinkedList<>();
```
calohmn, line 72: How about clearing `tenantsToEnable` and cancelling a potentially existing timer in the `stop()` method?

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
```diff
@@ -209,53 +220,94 @@ private Duration getSanitizedLifespan(final Duration lifespan) {
     public Future<CommandRouterResult> enableCommandRouting(final List<String> tenantIds, final Span span) {
         Objects.requireNonNull(tenantIds);
         final boolean isProcessingRequired = tenantsToEnable.isEmpty();
-        tenantsToEnable.addAll(tenantIds);
+        tenantIds.forEach(id -> {
+            tenantsToEnable.addLast(Pair.of(id, 1));
```
calohmn, line 224: How about checking for each to-be-added tenantId first, if it's already in `tenantsToEnable`?
Otherwise the `attemptNo` counter wouldn't be accurate.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
```diff
@@ -209,53 +220,94 @@ private Duration getSanitizedLifespan(final Duration lifespan) {
...
+                logEntries.put(Fields.MESSAGE, "failed to create command consumer");
+                logEntries.put(Fields.ERROR_OBJECT, t);
+                TracingHelper.logError(span, logEntries);
+                if (t instanceof ServerErrorException) {
+                    // add to end of queue in order to retry at a later time
+                    LOG.info("failed to create command consumer [attempt#: {}]", attempt.two(), t);
+                    span.log("marking tenant for later re-try to create command consumer");
+                    // with the maximum delay of 10s before re-trying,
+                    // the attempt number will increase to 
+                    tenantsToEnable.addLast(Pair.of(attempt.one(), attempt.two() + 1));
```
calohmn, line 304: I would suggest checking first whether `tenantsToEnable` already contains an entry for that tenant (the entry could have been added while processing above took place or during the timer period) and removing that entry first.

sophokles73: I have pushed some changes

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
```diff
@@ -207,55 +219,103 @@ private Duration getSanitizedLifespan(final Duration lifespan) {
...
         if (isProcessingRequired) {
             final Span processingSpan = tracer.buildSpan("re-enable command routing for tenants")
-                    .addReference(References.FOLLOWS_FROM, span.context())
+                    .addReference(References.CHILD_OF, span.context())
                     .start();
-            processTenantQueue(new ConcurrentHashSet<>(), processingSpan);
+            processingSpan.log(Map.of("number of tenant IDs", tenantIds.size()));
+            // make sure to report the span in a timely fashion
+            // so that it can be seen in the tracing back end
+            processingSpan.finish();
```
calohmn, line 243: I think this span isn't really needed - there is already the "enable command routing" parent span, with a `no_of_tenants` log entry.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
```diff
@@ -207,55 +219,103 @@ private Duration getSanitizedLifespan(final Duration lifespan) {
...
+            .map(factory -> factory.createCommandConsumer(attempt.one(), span.context()))
+            .onSuccess(ok -> {
+                logEntries.put(Fields.MESSAGE, "successfully created command consumer");
+                span.log(logEntries);
+                reenabledTenants.add(attempt.one());
+            })
+            .onFailure(t -> {
+                logEntries.put(Fields.MESSAGE, "failed to create command consumer");
+                logEntries.put(Fields.ERROR_OBJECT, t);
+                TracingHelper.logError(span, logEntries);
```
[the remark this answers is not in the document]
sophokles73: `TracingHelper.logError(Span, Map)` sets the *error* tag to `true`.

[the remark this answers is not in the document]
calohmn: Yes, I think that should be done here. The `re-enable command routing for tenant` action failed so the corresponding span gets the error tag.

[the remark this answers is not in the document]
sophokles73: my point is that there is nothing that needs to be changed because the span is already marked as erroneous with the current invocation of `TracingHelper.logError(Span, Map)`

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
```diff
@@ -207,55 +219,103 @@ private Duration getSanitizedLifespan(final Duration lifespan) {
...
-            if (processedTenants.contains(tenantId)) {
-                parentSpan.log(Map.of(
-                        Fields.MESSAGE, "skipping tenant, already processed ...",
-                        TracingHelper.TAG_TENANT_ID.getKey(), tenantId));
+    private void processTenantQueue(final SpanContext tracingContext) {
+
+        final var attempt = tenantsToEnable.pollFirst();
+        if (attempt == null) {
+            reenabledTenants.clear();
+            tenantsInProcess.clear();
```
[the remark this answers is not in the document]
sophokles73: right

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
```diff
@@ -207,55 +219,103 @@ private Duration getSanitizedLifespan(final Duration lifespan) {
...
+        }
+    }
+
+    private void activateCommandRouting(
+            final Pair<String, Integer> attempt,
+            final SpanContext tracingContext) {
+
+        if (!running.get()) {
+            // component has been stopped, no need to create command consumer in this case
+            tenantsInProcess.remove(attempt.one());
```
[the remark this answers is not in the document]
sophokles73: probably not, but it also doesn't cause any harm, IMHO

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
sophokles73: I have pushed corresponding changes ...

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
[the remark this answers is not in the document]
calohmn: Yes, I got that wrong about the error tag.
(My point was anyway more about the usage of the `logEntries` map, with the above mentioned alternatives being more straightforward and shorter. But, that may also be a matter of taste.)

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java
```diff
@@ -207,55 +218,97 @@ private Duration getSanitizedLifespan(final Duration lifespan) {
...
-            if (processedTenants.contains(tenantId)) {
-                parentSpan.log(Map.of(
-                        Fields.MESSAGE, "skipping tenant, already processed ...",
-                        TracingHelper.TAG_TENANT_ID.getKey(), tenantId));
+    private void processTenantQueue(final SpanContext tracingContext) {
+
+        final var attempt = tenantsToEnable.pollFirst();
+        if (attempt == null) {
+            reenabledTenants.clear();
+            LOG.debug("finished re-enabling of command routing");
```
calohmn, line 247: nit pick: It's only finished if `tenantsInProcess` is empty.

sophokles73: true

sophokles73: I have pushed a fix

calohmn (approved):
LGTM

PR #2710, opened 2021-06-08
[#2705] Fix pendingRequests handling in tenant client

### LINKED ISSUE #2705: Command Router: endless "Enable Command Routing" loop

calohmn:
A "get tenant" service invocation that failed immediately (with no asynchronous decoupling in between) caused that failure to be put in the `ProtonBasedTenantClient#pendingRequests` map and be used for later "get tenant" invocations.

This is the main fix for #2705.

The added test that reproduces the bug with the old `ProtonBasedTenantClient` code is `testGetTenantSucceedsWhenCreditRestored`.

### DIFF
--- a/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/registry/amqp/ProtonBasedTenantClient.java
+++ b/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/registry/amqp/ProtonBasedTenantClient.java
@@ -177,7 +177,7 @@ public final class ProtonBasedTenantClient extends AbstractRequestResponseServic
             final Span currentSpan) {
 
         final Future<TenantResult<TenantObject>> resultTracker = getResponseFromCache(responseCacheKey, currentSpan)
-                .recover(cacheMiss -> addResultHandler(
+                .recover(cacheMiss -> executeOrUsePendingRequestResult(
                         responseCacheKey,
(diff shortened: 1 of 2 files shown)


sophokles73 (approved):
LGTM

nice catch :+1:

Issue #2661, opened 2021-05-17
Kafka: Provide feedback on failed Command messages


calohmn:
One main difference between the Kafka and AMQP Command & Control implementations is that with Kafka, there is no protocol-provided way to inform the northbound application about a failed delivery, like there is with AMQP using the delivery disposition update mechanism (using a `rejected`, `modified` or `released` outcome).

A failure to forward a command message to the device could for example be caused by
- the command message being invalid
- the command target device or its tenant being disabled/deleted
- there being no currently subscribed command consumer for the target device
- some network error when forwarding the message
- etc.

## Request/response command messages

When there is an error forwarding a request/response command message, a command response message with the corresponding error could be published on the command-response topic.
That means either the Command Router component or the Protocol Adapter will publish such a message.
The message should have some special header so that it can be distinguished from a command response coming from the device.

## One-way command messages

For one-way command messages, using the command-response topic for reporting errors doesn't seem appropriate, "one way" and "response" being mutually exclusive.
However, having a backchannel for an error notification in a scenario where only one-way commands can be used because the device can't send responses still looks like a useful feature. 

One-way and request-response commands are currently distinguished via the `response-required` header.
For a scenario of sending commands to a device that doesn't support sending response messages, we could use an additional header - `use-delivery-status-response` (or `only-delivery-status-response`). 
With `response-required` and `use-delivery-status-response` set to `true`, the command will be forwarded like a one-way command to the device. But any error during command processing in the command router or the adapter will be reported on the command-response topic and a successful delivery will result in a "status 202" command-response message. Such a response would be equivalent to an `accepted` delivery outcome returned in the AMQP one-way command case - with the same caveat, that a northbound application can't be certain that the device has actually received the command message in that case. Nonetheless, being able to quickly get an error response makes the overall feature worthwhile FMPOV.

calohmn:
The idea is now to (first) implement sending error command responses for _request/response_ command messages.

I would make these messages distinguishable from command responses sent from the device itself by using a special content type, namely `application/vnd.eclipse-hono-delivery-failure-notification+json`.

calohmn:
Implemented for request/response command messages - see #2728, #2712.

sophokles73:
@calohmn can you also add an entry to the release notes?

calohmn:
@sophokles73 done (fa52f41161241b11868e33ecf3ee1c937f6c99bb).

PR #2711, opened 2021-06-08
[#2661] Use Throwable param in CommandContext methods

calohmn:
This is for #2661.

This is a preparation for letting the KafkaBasedCommandContext send Command Response messages in the release/modify/reject methods. The new Throwable param allows an error message and an error status code to be given.

In the first commit:
Move/rename the `AbstractProtocolAdapterBase#getClientFacingErrorMessage()` method to the `ServiceInvocationException` class.
Also fix setting of an error message in CoapErrorResponse.

### DIFF
--- a/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedLegacyCommandContextWrapper.java
+++ b/clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedLegacyCommandContextWrapper.java
@@ -15,9 +15,9 @@ package org.eclipse.hono.adapter.client.command.amqp;
 
 import java.util.Objects;
 
-import org.apache.qpid.proton.amqp.transport.ErrorCondition;
 import org.eclipse.hono.adapter.client.command.Command;
 import org.eclipse.hono.adapter.client.command.CommandContext;
+import org.eclipse.hono.client.ServiceInvocationException;
 import org.eclipse.hono.util.Constants;
 
 import io.opentracing.Span;
@@ -64,14 +64,24 @@ public class ProtonBasedLegacyCommandContextWrapper implements CommandContext {
     }
 
     @Override
+    public void release(final Throwable error) {
+        Objects.requireNonNull(error);
+        ctx.release();
+    }
+
+    @Override
     public void modify(final boolean deliveryFailed, final boolean undeliverableHere) {
         ctx.modify(deliveryFailed, undeliverableHere);
     }
 
     @Override
     public void reject(final String cause) {
-        final ErrorCondition error = ProtonHelper.condition(Constants.AMQP_BAD_REQUEST, cause);
-        ctx.reject(error);
+        ctx.reject(ProtonHelper.condition(Constants.AMQP_BAD_REQUEST, cause));
+    }
+
+    @Override
+    public void reject(final Throwable error) {
+        reject(ServiceInvocationException.getErrorMessageForExternalClient(error));
     }
 
     @Override
--- a/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommandContext.java
+++ b/clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommandContext.java
@@ -13,9 +13,11 @@
 
 package org.eclipse.hono.adapter.client.command.kafka;
 
+import java.net.HttpURLConnection;
 import java.util.Objects;
 
 import org.eclipse.hono.adapter.client.command.CommandContext;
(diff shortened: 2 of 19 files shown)


clients/adapter/src/main/java/org/eclipse/hono/adapter/client/command/CommandContext.java
```diff
@@ -59,17 +61,18 @@
     /**
      * Indicates to the sender that the command message could not be delivered to its target due to
      * reasons that are not the responsibility of the sender of the command.
+     *
+     * @param error The delivery error. May be {@code null}.
      */
-    void release();
+    void release(Throwable error);
```
sophokles73, line 67: I'd rather keep the existing variant without a param and instead overload the method. Explicitly passing in a `null` argument into methods doesn't look too good FMPOV, in particular when the method has a single parameter only ...

calohmn: I've added the extra `release()` method.

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedLegacyCommandContextWrapper.java
```diff
@@ -63,15 +63,25 @@ public void release() {
         ctx.release();
     }
 
+    @Override
+    public void release(final Throwable error) {
+        Objects.requireNonNull(error);
+        ctx.release();
```
sophokles73, line 69: shouldn't this invoke `ctx.release(Throwable)`?

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommandContext.java
```diff
@@ -57,21 +61,34 @@ public KafkaBasedCommand getCommand() {
 
     @Override
     public void accept() {
+        if (!completed.compareAndSet(false, true)) {
```
sophokles73, line 64: out of curiosity: why are you guarding this with a boolean?

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedLegacyCommandContextWrapper.java
calohmn: The deprecated `org.eclipse.hono.client.CommandContext` class doesn't have such a method.

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommandContext.java
calohmn: The guard is mainly for the other methods (reject/modify/release). It's a preparation for when in the KafkaBasedCommandContext class a command response is to be sent, making sure it doesn't get sent twice.
(In the AMQP context, invoking these methods twice has no effect, so there might be a case where this may happen.)

clients/adapter-amqp/src/main/java/org/eclipse/hono/adapter/client/command/amqp/ProtonBasedLegacyCommandContextWrapper.java
sophokles73: Ah, I see. It's the deprecated one ...

clients/adapter-kafka/src/main/java/org/eclipse/hono/adapter/client/command/kafka/KafkaBasedCommandContext.java
sophokles73: I see. Aren't you a little worried that two *different* methods might be invoked instead of the same method being invoked twice? This would raise the question of which invocation actually reflects the proper outcome, right?

calohmn: I'm thinking more about cases where "released" might be invoked twice. I don't know about actual cases where this may happen, but I'm not 100% certain it can't happen (cases where a timeout for getting device registry info has kicked in but command processing was already cancelled because the corresponding command partition got unassigned, something like that).
So this is just a safety measure. I've also just changed the code here a bit, logging both outcomes.

sophokles73 (approved):
LGTM

calohmn:
@sophokles73 Should I merge this now or wait for the refactoring regarding moving the `org.eclipse.hono.adapter.client.command.amqp` classes?

sophokles73:
@calohmn please go ahead, I can rebase if necessary ...

PR #2712, opened 2021-06-09
[#2661] Document Hono sending Kafka error command responses

calohmn:
This is for #2661:

Adds documentation for the new feature that the Command Router or a protocol adapter sends an error command response message if a Kafka request/response command message couldn't be delivered to the device (or can't be processed by it).

Also improves the documentation of the AMQP command & control spec regarding the `released` and `rejected` delivery outcomes.


sophokles73 (approved):
LGTM

PR #2713, opened 2021-06-09
Move registry client classes into separate module

sophokles73:
This is step two of #2540

In order to better reflect such cases, the classes for querying the
registry services have been moved into dedicated registry and
registry-amqp modules.


calohmn:
Javadoc references beginning with `{@code org.eclipse.hono.adapter.client.registry` should be adapted as well.

Logback files containing "org.eclipse.hono.adapter.client" loggers but no "org.eclipse.hono.client" loggers should also be adapted FMPOV. Could also be done in the followup PR when getting rid of the "org.eclipse.hono.adapter.client" package, though.

sophokles73:

calohmn (approved):
LGTM

PR #2714, opened 2021-06-09
Rework cli profiles

b-abel:
I noticed a couple of properties that are set in the configuration file of the CLI component which set values that do not really match a recommended Hono setup. So, this PR does not resolve the issue #1486 and does not change the logic of the CLI. It is an intermediary step that updates some defaults.

The goal is to improve the comprehensibility of the file `application.yml` and thereby also the usability of the tool.

The main points are:

1. all usage instructions documented within the Hono project should behave as before
1. this PR does only change defaults for the profiles for the usage as a northbound client and not for the profiles `amqp-send` and `amqp-command`
1. one exception to the point above is the hostname. Which changes from localhost to the sandbox address in the southbound profile
1. there are 2 setups for which we should provide defaults: a) the sandbox and b) a deployment with the Helm chart
1. the default setup should be the sandbox (instead of localhost which does not match any recommended setup)
1. default IDs for tenant and device make sense for custom setups but should not be used on the Sandbox
1. the profile `ssl` does not work for any of the setups and should be removed
1. enable TLS for the sandbox, disable for local setups (using TLS locally would require getting the truststore from the deployment in the right version)
1. I could not resist fixing some issues in the documentation that I found while testing the examples

The solution proposed here is having 3 types of profiles. One can use one of each type:

1. select the "mode": `receiver`, `command`
2. select the "target" `sandbox`, `local` (aims for deployment in Minikube but works for every deployment of the Helm chart)
3. select the "messaging-type": `kafka` (if not set, it defaults to AMQP-based messaging)

(A new CLI tool would better use separate parameters for each of those instead of leaking Spring profiles)

### DIFF
--- a/cli/src/main/resources/application.yml
+++ b/cli/src/main/resources/application.yml
@@ -1,32 +1,38 @@
+# The file contains some default configurations for Hono's CLI tool.
+#
+# The Hono CLI serves two different use cases:
+# For the first one, it can be used as a southbound client to Hono's AMQP protocol adapter
+# and simulates an AMQP enabled device. This southbound client is activated by setting
+# one of the profiles "amqp-send" or "amqp-command".
+#
+# For the second use case, it serves as a northbound client that simulates a business application
+# that receives telemetry data and events or sends commands upstream to devices.
+# The northbound client supports a couple of profiles that provide some defaults:
+# 1. select the mode: "receiver" or "command"
+# 2. select the target: "sandbox" or "local"
+# 3. optionally add "kafka" to enable Kafka based messaging (otherwise AMQP is used for messaging).
+
 spring:
   jmx:
     enabled: false
   profiles:
-    active: receiver,ssl
+    active: receiver,sandbox
 
 hono:
   client:
-    host: localhost
     reconnectAttempts: 5
 
 connection:
   retryInterval: 1000
 
-tenant:
-  id: DEFAULT_TENANT
-device:
-  id: 4711
-
 ---
+# Enables the "receiver" mode of the northbound client used to receive telemetry data and/or event messages.
+# Use this profile together with either "sandbox" or "local".
 
 spring:
   config:
     activate:
       on-profile: receiver
-
-hono:
-  client:
-    port: 15671
 
 address:
   resolver:
@@ -39,6 +45,79 @@ message:
   type: all
 
 ---
+# Enables the "command" mode of the northbound client used to send commands to devices.
+# Use this profile together with either "sandbox" or "local".
+
+spring:
+  config:
+    activate:
+      on-profile: command
+
+command:
+  timeoutInSeconds: 60
+
+---
+# Configuration for the Hono sandbox - see:
+# https://www.eclipse.org/hono/getting-started/
+# Use this profile together with either "receiver" or "command".
+
+spring:
+  config:
+    activate:
+      on-profile: sandbox
+
+hono:
+  client:
+    host: hono.eclipseprojects.io
+    port: 15672
+    tlsEnabled: false
+    username: consumer@HONO
+    password: verysecret
+  kafka:
+    commonClientConfig:
+      bootstrap.servers: hono.eclipseprojects.io:9092
+      security.protocol: SASL_PLAINTEXT
+      sasl.mechanism: SCRAM-SHA-256
+      sasl.jaas.config: "org.apache.kafka.common.security.scram.ScramLoginModule required username=\"hono\" password=\"hono-secret\";"
+
+---
+# Configuration for example deployment in local Kubernetes cluster - see:
+# https://www.eclipse.org/hono/getting-started/
+# Use this profile together with either "receiver" or "command".
+# If used with Kafka based messaging, bootstrap servers and the truststore path need to be provided additionally.
+
+spring:
+  config:
+    activate:
+      on-profile: local
+
+hono:
+  client:
+    port: 15672
+    tlsEnabled: false
+    username: consumer@HONO
+    password: verysecret
+  kafka:
+    commonClientConfig:
+      bootstrap.servers: UNCONFIGURED
+      security.protocol: SASL_SSL
+      sasl.jaas.config: "org.apache.kafka.common.security.scram.ScramLoginModule required username=\"hono\" password=\"hono-secret\";"
+      sasl.mechanism: SCRAM-SHA-512
+      ssl.truststore.location: UNCONFIGURED
+      ssl.truststore.password: honotrust
+      ssl.endpoint.identification.algorithm: ""
+
+tenant:
+  id: DEFAULT_TENANT
+device:
+  id: 4711
+
+---
+# This profile enables Kafka based messaging - see:
+# https://www.eclipse.org/hono/getting-started-kafka/
+# Use this profile together with either "receiver" or "command".
+# Additionally activate either "sandbox" or "local".
+# If not using the sandbox, bootstrap servers and the truststore path need to be provided additionally.
 
 spring:
   config:
@@ -48,11 +127,14 @@ spring:
 hono:
   kafka:
     commonClientConfig:
-      bootstrap.servers: localhost:9092
+      client.id: hono-cli
     consumerConfig:
-      group.id: hono-cli
+      group.id: cli
 
 ---
+# This profile enables the collection of statistics of received messages from
+# Hono's northbound Telemetry and Event APIs.
+# Use this profile together with "receiver".
 
 spring:
   config:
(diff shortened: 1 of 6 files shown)


cli/src/main/resources/application.yml
```diff
@@ -43,73 +33,106 @@ message:
...
+    port: 15671
+    tlsEnabled: true
+    username: consumer@HONO
+    password: verysecret
+  kafka:
+    commonClientConfig:
+      bootstrap.servers: hono.eclipseprojects.io:9092
+      security.protocol: SASL_PLAINTEXT
+      sasl.mechanism: SCRAM-SHA-256
+      sasl.jaas.config: "org.apache.kafka.common.security.scram.ScramLoginModule required username=\"hono\" password=\"hono-secret\";"
```
sophokles73, line 63: isn't it possible to use the same user/pwd for kafka as we do for AMQP?

cli/src/main/resources/application.yml
```diff
@@ -43,73 +33,106 @@ message:
...
+  config:
+    activate:
+      on-profile: amqp-send,amqp-command
 
-device:
-  id: 4711
+hono:
+  client:
+    host: hono.eclipseprojects.io
+    port: 5672
```
sophokles73, line 134: why shouldn't we use the TLS port here? My understanding was that we use secured ports with the sandbox and unsecured ones with the local deployment, or am I mistaken?

site/documentation/content/user-guide/amqp-adapter.md
```diff
@@ -197,7 +197,7 @@ Publish some JSON data for device `4711` using a client certificate for authenti
 
 ~~~sh
 # in directory: hono/cli/target/
-java -jar hono-cli-*-exec.jar --spring.profiles.active=amqp-send --hono.client.port=5671 --hono.client.certPath=config/hono-demo-certs-jar/device-4711-cert.pem --hono.client.keyPath=config/hono-demo-certs-jar/device-4711-key.pem --hono.client.trustStorePath=config/hono-demo-certs-jar/trusted-certs.pem --hono.client.hostnameVerificationRequired=false
+java -jar hono-cli-*-exec.jar --spring.profiles.active=amqp-send --hono.client.port=5671 --hono.client.certPath=config/hono-demo-certs-jar/device-4711-cert.pem --hono.client.keyPath=config/hono-demo-certs-jar/device-4711-key.pem --hono.client.tlsEnabled=true --hono.client.hostnameVerificationRequired=false
```
sophokles73, line 200: in this example we indeed use the secured port (5671)

cli/src/main/resources/application.yml
b-abel: The character '@' does not work in the username. I do not know the exact cause. I am open to other ideas for usernames. 
Apart from that, this needs to be changed in _IoT Packages_ first, I would like to decouple that from this PR.

cli/src/main/resources/application.yml
b-abel: This Spring profile is for the usage as a southbound client for the AMQP adapter (which should be in a separate app FMPOV). This PR targets the northbound client, so I have not thought much about the southbound client. 
The problem is here again the same as with the northbound client: there are a couple of properties (`host`, `port`, `tlsEnabled`, `hostnameVerificationRequired`) that must then be different between the sandbox and the local deployment and then the documentation needs to address them separately. But you're right and I'll change it and extend the documentation (I don't want to add separate profiles for the environments for the southbound client too).

sophokles73:
@b-abel any news on this one?

cli/src/main/resources/application.yml
b-abel: I have found that the situation is more complex:

In the Getting Started Guide, only insecure ports are used. This guide is IMHO particularly important because error sources here easily discourage beginners. Besides, it is the only place in the documentation where the use of the (northbound) CLI is mentioned to some extent.

In the user guide of the AMQP adapter, there is one example with TLS, the one you mentioned below. But in this example, TLS is mandatory, because it should demonstrate the authentication with client certificates. This makes it independent from the other examples in that guide (8x insecure port, 1x secure port). 

Then there is the readme of the CLI module, there also only the insecure port is used.

On the sandbox page is the recommendation to use only the secure ports.

The defaults of the CLI were inconsistent: if you didn't specify a profile at all, the TLS port was selected (but nothing worked, because localhost was set and an invalid truststore was set). In southbound client mode, however, the insecure port has always been the default.

In addition, we had discussed elsewhere to offer only the insecure port on Kafka on the sandbox. 

I wonder if it's not best to use the insecure ports as defaults everywhere. It's not pretty, but it's consistent and changes the previous behavior the least. WDYT?

sophokles73: > In the Getting Started Guide, only insecure ports are used. This guide is IMHO particularly important because error sources here easily discourage beginners. Besides, it is the only place in the documentation where the use of the (northbound) CLI is mentioned to some extent.

Then I guess, we should keep the Getting Started guide using the insecure ports.

> I wonder if it's not best to use the insecure ports as defaults everywhere. It's not pretty, but it's consistent and changes the previous behavior the least. WDYT?

Not sure what you are referring to here. Is this about using the insecure port(s) in the CLI's Spring profiles by default?

b-abel: > Not sure what you are referring to here. Is this about using the insecure port(s) in the CLI's Spring profiles by default?

Yes, I mean only the defaults in the CLI's Spring profiles.

sophokles73: Then let's do that.

sophokles73:
It looks like the client now always will try to connect to a Kafka cluster, regardless of whether the *kafka* profile has been activated or not. Or am I mistaken?

b-abel:
> It looks like the client now always will try to connect to a Kafka cluster, regardless of whether the _kafka_ profile has been activated or not. Or am I mistaken?

No, it won't. The configuration for Kafka is only considered (by Spring's configuration mechanisms) if the profile "kafka" is activated.

b-abel:
I added some comments to the `application.yml` to make it easier to understand how to combine the profiles.

sophokles73:
> No, it won't. The configuration for Kafka is only considered (by Spring's configuration mechanisms) if the profile "kafka" is activated.

Well, my understanding is that the *sandbox* profile is active by default and also defines kafka bootstrap servers, doesn't it?

b-abel:
> > No, it won't. The configuration for Kafka is only considered (by Spring's configuration mechanisms) if the profile "kafka" is activated.
> 
> Well, my understanding is that the _sandbox_ profile is active by default and also defines kafka bootstrap servers, doesn't it?

That's right. However, Spring will not try to apply these configuration values to ConfigurationProperties because in Spring configuration checks for the profile "kafka". By this I mean the following excerpts on the `AppConfiguration.java` file:  

```java
@ConfigurationProperties(prefix = "hono.kafka")
@Profile("kafka")
@Bean
public KafkaProducerConfigProperties kafkaProducerConfig() {
    return new KafkaProducerConfigProperties();
}

@Profile("kafka")
@Bean
public ApplicationClient<KafkaMessageContext> kafkaApplicationClient(
        final Vertx vertx,
        final KafkaConsumerConfigProperties kafkaConsumerConfigProperties,
        final KafkaProducerFactory<String, Buffer> producerFactory,
        final KafkaProducerConfigProperties kafkaProducerConfigProperties) {
    return new KafkaApplicationClientImpl(vertx, kafkaConsumerConfigProperties, producerFactory,
            kafkaProducerConfigProperties);
}
```

This means that the beans required for Kafka-based messaging are not created, including the configuration properties. Therefore the client won't try to connect to a Kafka cluster without having the profile "kafka" activated regardless of the properties provided. 
The same is true for AMQP: if the profile "kafka" is activated, the configuration for AMQP is ignored and the AMQP-based client is not created. This is not new with this PR.
In other words: Spring ignores the properties for the unused messaging and does not pass it to the application code. 

The idea of this PR is to have independent "dimensions" for the mode of the (northbound) client (receiver vs. command), for the target (sandbox vs. other), and for the messaging system (AMQP vs. Kafka). I tried to explain this concept at the beginning of the file `application.yml`. The alternative would be having dedicated profiles for each scenario like "receiver-sandbox-amqp" or "command-local-kafka" which looks quite cumbersome to me.

Do you think that this too confusing? Or have a better idea of how to structure the defaults for the different "dimensions".

sophokles73 (approved):
LGTM

sophokles73:
@b-abel can you squash this to a reasonable number of commits?

b-abel:
> @b-abel can you squash this to a reasonable number of commits?

Done.

Issue #2704, opened 2021-06-07
LoRa Adapter: command subscriptions never get removed


calohmn:
With the command & control implementation in #2482, the LoRa adapter adds a command subscription for the LoraWAN network server (LNS) when the adapter has received a telemetry message from the LNS.

But, this command subscription never gets removed, also not when the tenant of the LNS or the gateway device entry is deleted.

This means that when  the LoRa adapter loses the connection to the Command Router and then reconnects, the subsequent "Enable Command Routing" (see #2576) potentially contains obsolete tenants.

In a system where many test tenants get created and deleted over time, this could lead to unnecessarily large "Enable Command Routing" requests.

calohmn:
With the upcoming new notification API, informing protocol adapters of tenant/device entry change (see #293), the command subscriptions should be removed when gateway device or corresponding tenant get deleted.

Up til then, we could consider adding a periodic tenant/device check to clean up command subscriptions.

calohmn:
With the changes from #2720, the LoRa adapter now checks whenever a tenant has been idle for a given time period (with no telemetry/event/command messages received during that time) whether that tenant actually still exists and is enabled. If either is not the case, the command consumers for that tenant will be closed.

Note that for that mechanism to actually be used, the corresponding config property `HONO_LORA_TENANT_IDLE_TIMEOUT` / `--hono.lora.tenantIdleTimeout` has to be set explicitly (the default property value is zero, meaning the above check and consumer removal isn't done).

The 2nd part of the fix here would be applying the upcoming device/tenant changes notification API as described above. That should be handled via a new issue then.

calohmn:
> The 2nd part of the fix here would be applying the upcoming device/tenant changes notification API as described above. That should be handled via a new issue then.

Actually, the tenant-timeout based approach can be replaced with one that uses the `DeviceChangeNotification` mechanism now.

PR #2720, opened 2021-06-10
[#2704] Check LoRa adapter command consumers on tenant timeout

calohmn:
Fix for #2704:
If tenant is already removed or disabled, corresponding command consumers will be closed.


sophokles73 (approved):
LGTM

PR #2721, opened 2021-06-10
Move C&C client classes into separate module

sophokles73:
This is the 3rd (and last) step for #2540

In order to better reflect such cases, the classes for consuming
commands and publishing command responses have been moved into dedicated
command, command-amqp and command-kafka modules.


calohmn:
There are still `{@code org.eclipse.hono.adapter.client` javadoc entries and `<logger name="org.eclipse.hono.adapter.client"` logger definitions.

sophokles73:

calohmn (approved):
LGTM

PR #2722, opened 2021-06-10
LoRa adapter: Check for command endpoint before creating consumer

calohmn:
The registration of a command consumer is skipped now during handling of a message from a LoRa network gateway if there is no command endpoint defined in the gateway device registration data. 
Handling of command messages for such a gateway would fail anyway because of the missing command endpoint.

It is also made sure that the LoRa adapter "process message" span gets finished before the root span.

This relates to #2704, not creating a consumer in the first place if the gateway device isn't configured accordingly.

### DIFF
--- a/adapters/lora-vertx-base/src/main/java/org/eclipse/hono/adapter/lora/LoraProtocolAdapter.java
+++ b/adapters/lora-vertx-base/src/main/java/org/eclipse/hono/adapter/lora/LoraProtocolAdapter.java
@@ -58,6 +58,7 @@ import org.slf4j.LoggerFactory;
 
 import io.micrometer.core.instrument.Timer.Sample;
 import io.opentracing.Span;
+import io.opentracing.SpanContext;
 import io.opentracing.log.Fields;
 import io.opentracing.tag.StringTag;
 import io.opentracing.tag.Tag;
@@ -81,6 +82,8 @@ import io.vertx.ext.web.handler.ChainAuthHandler;
  * A Vert.x based protocol adapter for receiving HTTP push messages from a LoRa provider's network server.
  */
 public final class LoraProtocolAdapter extends AbstractVertxBasedHttpProtocolAdapter<LoraProtocolAdapterProperties> {
+
+    static final String SPAN_NAME_PROCESS_MESSAGE = "process message";
 
     private static final Logger LOG = LoggerFactory.getLogger(LoraProtocolAdapter.class);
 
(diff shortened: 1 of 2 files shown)


calohmn:
@BobClaerhout Could you have a look here?

A scenario that wouldn't anymore with this change would involve a gateway device being first provisioned *without* a command endpoint and the gateway forwarding telemetry messages already.
When at a later point the "command endpoint" information gets set for the gateway device, and no telemetry messages get sent anymore, incoming command messages are not processed.
FMPOV, such a scenario isn't really realistic. WDYT?

BobClaerhout:
Does this mean that every device to which you want to send a command to, needs to send a telemetry/event message before being able to receive a command? Or is 1 message from 1 device via that gateway/LNS enough?

calohmn:
> Or is 1 message from 1 device via that gateway/LNS enough?

Yes, nothing has changed in that regard.
What is added here is the condition that when the telemetry/event message sent via the gateway/LNS is handled, the gateway/LNS needs to have a non-empty "command endpoint" set in its registration data, otherwise the ability to receive commands won't be activated.

BobClaerhout:
Ok, in that case I do not have any objections. LGTM

adapters/lora-vertx-base/src/main/java/org/eclipse/hono/adapter/lora/LoraProtocolAdapter.java
```diff
@@ -235,78 +238,106 @@ protected void customizeDownstreamMessageProperties(final Map<String, Object> pr
...
+        final String tenantId = gatewayDevice.getTenantId();
+        final String gatewayId = gatewayDevice.getDeviceId();
+        final SubscriptionKey key = new SubscriptionKey(tenantId, gatewayId);
+        if (commandSubscriptions.containsKey(key)) {
+            return;
         }
-        currentSpan.finish();
+        // use FOLLOWS_FROM span since this operation is decoupled from the rest of the request handling
+        final Span currentSpan = TracingHelper.buildFollowsFromSpan(tracer, context, "create command consumer")
+                .withTag(Tags.SPAN_KIND.getKey(), Tags.SPAN_KIND_SERVER)
```
sophokles73, line 314: IMHO this span's kind should rather be *client* as we are now creating the consumer and invoke the Command Router service ...

sophokles73 (approved):
LGTM

PR #2723, opened 2021-06-10
[#2706] Document required versions of Kafka

### LINKED ISSUE #2706: Document compatibility of Kafka versions
While Hono's APIs for Kafka only require at least Kafka version `1.0` (brokers and clients), Hono's clients require a much newer version. 

I found the following to be a concise summary of the rules that determine the compatibility of Kafka clients and brokers ([Source](https://docs.cloudera.com/cdp-private-cloud-base/7.1.6/kafka-managing/topics/kafka-manage-client-broker-comp.html)):

> * Newer Kafka brokers can talk to older Kafka clients. The reverse is not true: older Kafka brokers cannot talk to newer Kafka clients.
> * Changes in either the major part or the minor part of the upstream version major.minor determines whether the client and broker are compatible. Differences among the maintenance versions are not considered when determining compatibility.

Hono's clients (in Protocol Adapters, Command Router, Device Registry, ...) use the Kafka clients with a version determined by the Vert.x Kafka client. Currently, Kafka clients in version 2.6 are used. This effectively requires that the brokers have at least version 2.6 too.

This information needs to be added to Hono's documentation. 
A good place for this would probably be a separate "Kafka" page in the admin guide.

b-abel:
This resolves #2706.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -51,6 +51,8 @@
     <jgroups.version>4.2.9.Final</jgroups.version>
     <jjwt.version>0.11.2</jjwt.version>
     <junit.jupiter.version>5.7.1</junit.jupiter.version>
+    <!-- when updating the Kafka clients, update the documentation in
+    site/documentation/content/admin-guide/hono-kafka-client-configuration.md as well -->
     <kafka-clients.version>2.6.0</kafka-clients.version>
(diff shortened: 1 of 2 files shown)


sophokles73 (approved):
LGTM

PR #2730, opened 2021-06-14
Fix command tenant/device 404 handling, adapt command tracing

calohmn:
Relates to #2661:
The Command Router handles commands, for which the corresponding tenant or device is disabled or doesn't exist, with a 'rejected' disposition now (instead of 'released').
Tracing of command handling errors has been moved into the CommandContext classes.


calohmn:
Commit amended to add setting a `status` span tag.

clients/command-amqp/src/main/java/org/eclipse/hono/client/command/amqp/ProtonBasedCommandContext.java
```diff
@@ -70,38 +74,56 @@ public ProtonBasedCommand getCommand() {
...
+        TracingHelper.logError(getTracingSpan(), "command could not be delivered or processed");
+        Tags.HTTP_STATUS.set(getTracingSpan(), HttpURLConnection.HTTP_UNAVAILABLE);
         updateDelivery(Released.getInstance());
     }
 
     @Override
     public void release(final Throwable error) {
         Objects.requireNonNull(error);
+        TracingHelper.logError(getTracingSpan(), "command could not be delivered or processed", error);
+        final int status = error instanceof ServerErrorException
```
sophokles73, line 92: ServiceInvocationException.extractStatusCode(Throwable) ?

clients/command-amqp/src/main/java/org/eclipse/hono/client/command/amqp/ProtonBasedCommandContext.java
```diff
@@ -70,38 +74,56 @@ public ProtonBasedCommand getCommand() {
...
+                : HttpURLConnection.HTTP_INTERNAL_ERROR;
+        Tags.HTTP_STATUS.set(getTracingSpan(), status);
         updateDelivery(Released.getInstance());
     }
 
     @Override
     public void modify(final boolean deliveryFailed, final boolean undeliverableHere) {
+        final int status = undeliverableHere ? HttpURLConnection.HTTP_NOT_FOUND
+                : HttpURLConnection.HTTP_UNAVAILABLE;
+        Tags.HTTP_STATUS.set(getTracingSpan(), status);
```
sophokles73, line 103: is there a particular reason why we do not log an error to the Span here?

clients/command-amqp/src/main/java/org/eclipse/hono/client/command/amqp/ProtonBasedInternalCommandSender.java
```diff
@@ -93,12 +95,10 @@ public ProtonBasedInternalCommandSender(final HonoConnection connection) {
...
                 })
-                .recover(thr -> {
+                .onFailure(thr -> {
                     LOG.debug("failed to send command [{}] to downstream peer", commandContext.getCommand(), thr);
-                    TracingHelper.logError(commandContext.getTracingSpan(),
-                            "failed to send command message to downstream peer: " + thr);
-                    commandContext.release(thr);
-                    return Future.succeededFuture();
+                    commandContext.release(new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE,
+                            "failed to send command message to downstream peer: " + thr));
```
sophokles73, line 101: FMPOV the *thr* should be included as the root cause here instead of its String serialization...

clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedInternalCommandSender.java
```diff
@@ -72,7 +74,8 @@ public KafkaBasedInternalCommandSender(final KafkaProducerFactory<String, Buffer
                 getHeaders((KafkaBasedCommand) command),
                 commandContext.getTracingContext())
                         .onSuccess(v -> commandContext.accept())
-                        .onFailure(thr -> commandContext.release(thr));
+                        .onFailure(thr -> commandContext.release(new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE,
+                                "failed to publish command message on internal command topic: " + thr)));
```
sophokles73, line 78: see above

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/AbstractMappingAndDelegatingCommandHandler.java
```diff
@@ -102,20 +106,28 @@ public AbstractMappingAndDelegatingCommandHandler(
...
                     } else if (ServiceInvocationException.extractStatusCode(cause) == HttpURLConnection.HTTP_NOT_FOUND) {
-                        errorMsg = "no target adapter instance found for command with device id "
-                                + command.getDeviceId();
+                        log.debug("no target adapter instance found for command with device id "
+                                + command.getDeviceId(), cause);
+                        error = new NoConsumerException("no target adapter instance found");
+                    } else {
+                        log.debug("error getting target gateway and adapter instance for command with device id "
+                                + command.getDeviceId(), cause);
+                        error = new ServerErrorException(HttpURLConnection.HTTP_UNAVAILABLE,
```
sophokles73, line 123: see above

clients/command-amqp/src/main/java/org/eclipse/hono/client/command/amqp/ProtonBasedInternalCommandSender.java
calohmn: That would prevent the `thr` root cause to be logged because the `error.object` span log field only contains the `exception.toString` representation.

sophokles73: Then we should probably change that in *TracingHelper.logError()*, right?

clients/command-amqp/src/main/java/org/eclipse/hono/client/command/amqp/ProtonBasedCommandContext.java
calohmn: I've added log output.

clients/command-amqp/src/main/java/org/eclipse/hono/client/command/amqp/ProtonBasedInternalCommandSender.java
calohmn: That would mean adding an extra (non-standard) `error.cause.object` span log item, I guess.

I don't think that we need this in most cases where a Throwable with a non-null `cause` is given to `TracingHelper.logError()`.
Oftentimes the nested exception will have been logged in a child span already anyway.
Or it could be considered too low-level to include in tracing output (we are also not including stacktraces there).

So, I would rather say that if for a particular span the nested exception message should be logged, it should be done by adding an explicit `TracingHelper.logError` invocation.

sophokles73: > That would mean adding an extra (non-standard) error.cause.object span log item, I guess.

Sounds good :+1:

sophokles73: > I don't think that we need this in most cases where a Throwable with a non-null cause is given to TracingHelper.logError().

In most cases we do not specify a root cause anyway. So there shouldn't be a problem logging it using the *error.cause.object* field in the rare cases where we do set it, right?

calohmn: I've had a look here and yes, in most cases there's no root cause given.
And there are quite some cases, where logging the nested exception would indeed be beneficial - e.g. for an exception like `new ClientErrorException(HttpURLConnection.HTTP_BAD_REQUEST, e)`.

So, I've made the corresponding changes now, adding a `error.cause.object` field.

sophokles73 (approved):
LGTM

PR #2733, opened 2021-06-15
Make ServiceInvocationException abstract

calohmn:
We have many `exception instanceof ServerErrorException` and `exception instanceof ClientErrorException` conditions in the code.
These wouldn't match for plain `ServiceInvocationException` instances. Therefore, this PR makes `ServiceInvocationException` abstract.

I've also removed 3 constructors with tenant param in order to reduce the overall number of constructors.


sophokles73 (approved):
LGTM

PR #2734, opened 2021-06-15
Add hono user to docker images

michaelkrebs94:
This adds a user `hono` to the docker images of services and adapters, and uses it to run the application.
This provides additional security, because currently all applications run as root.

### DIFF
--- a/adapters/base-quarkus/pom.xml
+++ b/adapters/base-quarkus/pom.xml
@@ -152,6 +152,10 @@
                   <image>
                     <name>${docker.repository}/%a:%v</name>
                     <build>
+                      <runCmds>
+                        <runCmd>adduser --system --disabled-password -u 1000 hono</runCmd>
+                      </runCmds>
+                      <user>hono</user>
                       <env>
                         <JAVA_MAJOR_VERSION>${maven.compiler.release}</JAVA_MAJOR_VERSION>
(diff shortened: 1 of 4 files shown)


sophokles73:
@michaelkrebs94 The integration tests do not seem to work with your changes. Would you mind taking a look?
Is there a particular reason why you want to specify the UID when creating the hono user? I also wonder if we could/should add the `--no-create-home` switch as we do not need a home directory for the user ...

michaelkrebs94:
@sophokles73 I am looking into the integration tests.
The UID is specified in order to use it in a [Kubernetes SecurityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/), which requires specifying the UID instead of the username.

sophokles73:
My understanding is that if you configure the kubernetes security context, then kubernetes makes sure to run the container using the configured UID, GID (instead of the root user). In this case, you do not need to specify a particular UID, GID at all in the Dockerfile when building the image, or am I mistaken?

michaelkrebs94:
@sophokles73 you are right, the kubernetes securityContext does not need the UID to exist in the container.
Would you recommend creating the user anyway, or can we close this PR?

sophokles73:
FMPOV we can close it ...

Issue #2736, opened 2021-06-15
Support receiving AMQP commands after tenant switched to Kafka


calohmn:
The following scenario is given:
Hono is runnning with both AMQP and Kafka messaging systems and a tenant configuration has just been updated to use Kafka instead of the AMQP messaging system.
This means that from now on all messages sent by a device of that tenant are published to Kafka only.
Concerning the other direction, commands will at first still be received also via AMQP - at least as long as there are Command Router instances with already established command AMQP receiver links for that tenant.

This behaviour will allow for more time for the northbound application to switch from sending commands via AMQP to using Kafka.

It would be good if that support for still also receiving commands via AMQP would be made explicit by letting the Command Router instance also create a _new_ tenant AMQP receiver link if needed (e.g. if the instance was just started).
The definition of that added support could be either
- implicit, i.e. always creating AMQP command receiver links,
- configured in the Command Router config
- or configured for the tenant.

calohmn:
Btw: for the other way round, having a tenant configured for AMQP, we currently implicitly also support receiving commands via Kafka because of the wildcard topic Kafka consumer.

PR #2739, opened 2021-06-17
Use new MessagingClient interface for messaging clients

calohmn:
The clients managed by the ~~`org.eclipse.hono.client.util.MessagingClients`~~ `org.eclipse.hono.client.util.MessagingClientProvider` class (renamed from `MessagingClient`) now support invoking "getMessagingType()" on them, making it explicit whether e.g. a returned telemetry sender is Kafka or AMQP based.
The old `org.eclipse.hono.adapter.MessagingClients` class has been renamed to ~~`AdapterMessagingClients`~~ `MessagingClientProviders`.

This is a preparation for the fix for #2736.

(EDIT: used new class names based on the discussion below.)


clients/client-common/src/main/java/org/eclipse/hono/client/util/MessagingClients.java
```diff
@@ -36,7 +37,7 @@
  *
  * @param <T> The type of client to wrap.
  */
-public final class MessagingClient<T extends Lifecycle> implements Lifecycle, ServiceClient {
+public final class MessagingClients<T extends MessagingClient & Lifecycle> implements Lifecycle, ServiceClient {
```
sophokles73, line 40: I'd rather keep the name of this class as `MessagingClient` and instead rename the new `MessagingClient` interface to `TypedClient`. This would
1. prevent the massive renaming effort and
2. be abetter fit anyway because the new interface does not define any messaging client specific behavior anyway.

WDYT?

calohmn: Concerning the above class:
I think `MessagingClients` is a better fit there, because it actually contains multiple clients. This can also be seen when looking at the names chosen for instances, e.g. in `AbstractProtocolAdapterApplication`:
`final MessagingClient<TelemetrySender> telemetrySenders = new MessagingClient<>();`.

Concerning the name for the new Interface with the `getMessagingType` method:
`TypedClient` sounds very generic and is a name where I wouldn't associate `Type` with `MessagingType` - the "messaging" part is missing in there. `TypedMessagingClient`, `MessagingTypeAwareClient` or `MessagingTypeSpecificClient` as alternatives are all not very convincing to me either. 
I would still see `MessagingClient` as the most intuitive choice, also in combination with `MessagingClients` above.

sophokles73: > I think MessagingClients is a better fit there, because it actually contains multiple clients. 

It actually is a client for a particular API which *may* contain multiple implementations using different transports.

> This can also be seen when looking at the names chosen for instances, e.g. in AbstractProtocolAdapterApplication:
final MessagingClient<TelemetrySender> telemetrySenders = new MessagingClient<>();.

FMPOV those variables are named wrongly as well ...

sophokles73: > I would still see MessagingClient as the most intuitive choice, also in combination with MessagingClients above.

I am actually not sure if I understood, why we need an additional `MessagingClient` interface with a *getMessagingType* method in the first place.

calohmn: > It actually is a client for a particular API which *may* contain multiple implementations using different transports.

I think when it comes to the `client` notion here, using it for the contained implementations (e.g. telemetry senders) comes more naturally than using it for the `MessagingClient` container class itself (as it's named in master currently).
`MessagingClient` implements `ServiceClient` but that is quite an abstract kind of client with only the `registerReadinessChecks`, `registerLivenessChecks` methods. You really want to deal with the telemetry sender kind of clients when using this class.

So, I find these two kinds of `client` notions here rather confusing - also in connection with the `getClient` method names (e.g. when calling `messagingClient.getClient()`).

Maybe we can rename the `MessagingClient` class in master to something without `client` in the name? I don't have an actual suggestion here though.

Or maybe we at least name the class `MessagingClientWrapper`. `messagingClientWrapper.getClient()` sounds better than `messagingClient.getClient()`.
Or `MessagingClientContainer`.


> I am actually not sure if I understood, why we need an additional `MessagingClient` interface with a *getMessagingType* method in the first place.

This is useful for the fix for #2736, to directly query the messaging type of the client returned by the `getClient(tenant)` method. Otherwise an extra `getClientMessagingType(tenant)` would be needed, which of course wouldn't be a problem but I think the solution in this PR is more natural, also not having to pass around the object on which one would have to invoke `getClientMessagingType(tenant)`.

sophokles73: > Maybe we can rename the MessagingClient class in master to something without client in the name? I don't have an actual suggestion here though.

*MessagingClientProvider*?

> This is useful for the fix for #2736, to directly query the messaging type of the client returned by the getClient(tenant) method. 

In what context is this needed/done? I can also imagine creating a *MessageSender* interface in the *telemetry* module which defines the *getMessagingType* method and let *EventSender* and *TelemetrySender* extend it ...

calohmn: > *MessagingClientProvider*?

Yes, that's good. I've renamed the class and corresponding fields/methods accordingly.

> In what context is this needed/done? I can also imagine creating a *MessageSender* interface in the *telemetry* module which defines the *getMessagingType* method and let *EventSender* and *TelemetrySender* extend it ...

For #2736, it will be used in the context of the `CommandConsumerFactory` clients, potentially using both AMQP and Kafka factories.

https://github.com/eclipse/hono/blob/0dcf40933a6e42c6ec406982f8e012f281d9f5ed/services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/CommandRouterServiceImpl.java#L293-L294

adapter-base/src/main/java/org/eclipse/hono/adapter/AdapterMessagingClients.java
```diff
@@ -31,42 +31,42 @@
  * <p>
  * It contains clients for each of Hono's south bound API endpoints.
  */
-public final class MessagingClients implements Lifecycle {
+public final class AdapterMessagingClients implements Lifecycle {
```
sophokles73, line 34: MessagingClientProviders?

clients/client-common/src/main/java/org/eclipse/hono/client/util/MessagingClients.java
sophokles73: Then why don't you simply add the *getMessagingType* method to *CommandConsumerFactory*?

calohmn: That would also be a possible solution.

However, I found it then quite obvious to have the other kinds of messaging clients also implement such a `getMessagingType` method and then use the corresponding interface in what is here the `MessagingClientProvider` class.

clients/client-common/src/main/java/org/eclipse/hono/client/util/MessagingClientProvider.java
```diff
@@ -36,7 +37,7 @@
  *
  * @param <T> The type of client to wrap.
  */
-public final class MessagingClient<T extends Lifecycle> implements Lifecycle, ServiceClient {
+public final class MessagingClientProvider<T extends MessagingClient & Lifecycle> implements Lifecycle, ServiceClient {
```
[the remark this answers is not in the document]
calohmn: The `T extends MessagingClient` is needed for the `setClient(T client)` method (not having the `final MessagingType type` parameter anymore), so that `client.getMessagingType()` can be called there.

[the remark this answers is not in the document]
sophokles73: Oh yes, sure

sophokles73 (approved):
LGTM

PR #2742, opened 2021-06-21
[#2732] Set last known gateway as part of batch request

### LINKED ISSUE #2732: Limit rate at which setLastKnownGatewayForDevice requests are sent
Currently, for every telemetry/event message that a device with gateway support has sent, a `setLastKnownGatewayForDevice` request is made to the Command Router component.

If telemetry/event messages are sent at a high rate, this may cause a lot of traffic to the Command Router component.

Sending `setLastKnownGatewayForDevice` requests with the same gateway at a high rate (e.g. higher than 1 request per 50ms) seems unnecessary. A device switching between gateways every few milliseconds would be something we wouldn't be able to support with the `lastKnownGatewayForDevice` mechanism anyway, because of the latency associated with command forwarding.

Therefore, a protocol adapter should skip `setLastKnownGatewayForDevice` requests if the last request had the same gateway and isn't older than a configurable time period (e.g. 50ms or higher).

calohmn:
This fixes #2732.

The last known gateway of a device isn't set right away anymore. Instead, all such device-to-gateway mappings happening in a time frame of 400ms are sent to the Command Router as part of a batch request.
It is ensured that these requests don't exceed a certain size and not too many such requests are made at once.

### DIFF
--- a/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/CacheBasedDeviceConnectionInfo.java
+++ b/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/CacheBasedDeviceConnectionInfo.java
@@ -116,6 +116,33 @@ public final class CacheBasedDeviceConnectionInfo implements DeviceConnectionInf
             });
     }
 
+    @Override
+    public Future<Void> setLastKnownGatewayForDevice(final String tenantId,
+            final Map<String, String> deviceIdToGatewayIdMap, final Span span) {
+        Objects.requireNonNull(tenantId);
+        Objects.requireNonNull(deviceIdToGatewayIdMap);
+        Objects.requireNonNull(span);
+
+        if (deviceIdToGatewayIdMap.isEmpty()) {
+            return Future.succeededFuture();
+        }
+
+        final long lifespanMillis = LAST_KNOWN_GATEWAY_CACHE_ENTRY_LIFESPAN.toMillis();
+        final Map<String, String> mapToBePut = deviceIdToGatewayIdMap.entrySet().stream()
+                .collect(Collectors.toMap(entry -> getGatewayEntryKey(tenantId, entry.getKey()), Map.Entry::getValue));
+        return cache.putAll(mapToBePut, lifespanMillis, TimeUnit.MILLISECONDS)
+                .map(v -> {
+                    LOG.debug("set {} last known gateway entries [tenant: {}]", deviceIdToGatewayIdMap.size(), tenantId);
+                    return (Void) null;
+                })
+                .recover(t -> {
+                    LOG.debug("failed to set {} last known gateway entries [tenant: {}]", deviceIdToGatewayIdMap.size(),
+                            tenantId, t);
+                    TracingHelper.logError(span, "failed to set last known gateway entries", t);
+                    return Future.failedFuture(new ServerErrorException(HttpURLConnection.HTTP_INTERNAL_ERROR, t));
+                });
+    }
+
     /**
      * {@inheritDoc}
      */
--- a/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/DeviceConnectionInfo.java
+++ b/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/DeviceConnectionInfo.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ * Copyright (c) 2020, 2021 Contributors to the Eclipse Foundation
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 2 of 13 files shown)


client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/CacheBasedDeviceConnectionInfo.java
```diff
@@ -116,6 +116,33 @@ public CacheBasedDeviceConnectionInfo(final Cache<String, String> cache, final T
...
+
+        if (deviceIdToGatewayIdMap.isEmpty()) {
+            return Future.succeededFuture();
+        }
+
+        final long lifespanMillis = LAST_KNOWN_GATEWAY_CACHE_ENTRY_LIFESPAN.toMillis();
+        final Map<String, String> mapToBePut = deviceIdToGatewayIdMap.entrySet().stream()
+                .collect(Collectors.toMap(entry -> getGatewayEntryKey(tenantId, entry.getKey()), Map.Entry::getValue));
+        return cache.putAll(mapToBePut, lifespanMillis, TimeUnit.MILLISECONDS)
+                .map(replacedValue -> {
```
sophokles73, line 134: it doesn't look there is a *replacedValue* being returned by *cache.putAll* ...

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/DeviceConnectionInfo.java
```diff
@@ -47,6 +48,28 @@
...
+     * @param deviceIdToGatewayIdMap The map containing device identifiers and associated gateway identifiers. The
+     *                               gateway identifier may be the same as the device identifier if the device is
+     *                               (currently) not connected via a gateway but directly to a protocol adapter.
+     * @param span The active OpenTracing span for this operation. It is not to be closed in this method!
+     *            An implementation should log (error) events on this span and it may set tags and use this span as the
+     *            parent for any spans created in this method.
+     * @return A future indicating the outcome of the operation.
+     *         <p>
+     *         The future will be succeeded if the device connection information has been updated.
+     *         Otherwise the future will be failed with a {@link org.eclipse.hono.client.ServiceInvocationException}.
```
sophokles73, line 68: IMHO we should also document what happens if only a certain subset of the map's entries could be put to the cache.

clients/command-amqp/src/main/java/org/eclipse/hono/client/command/amqp/ProtonBasedCommandRouterClient.java
```diff
@@ -52,8 +62,34 @@
...
+     * sent after these requests are finished.
+     */
+    static final int SET_LAST_KNOWN_GATEWAY_UPDATE_MAX_PARALLEL_REQ = 50;
+
     private static final Logger LOG = LoggerFactory.getLogger(ProtonBasedCommandRouterClient.class);
 
+    /**
+     * Key is a Pair with tenant and device identifier, value is the gateway identifier.
+     */
+    private final LinkedHashMap<Pair<String, String>, String> lastKnownGatewaysWorkQueue = new LinkedHashMap<>();
```
sophokles73, line 84: Is there a particular reason why you are using the concrete class as the field's type here?

sophokles73 (commented):
Does this still work with a gateway publishing telemetry for one of its devices to the HTTP adapter including a TTD header? In particular, if the information about the mapping will be recorded before the downstream application gets the notification that it can send a command to the device ...

calohmn:
> Does this still work with a gateway publishing telemetry for one of its devices to the HTTP adapter including a TTD header? In particular, if the information about the mapping will be recorded before the downstream application gets the notification that it can send a command to the device ...

A gateway publishing telemetry for one of its devices along with a TTD header means a command subscription for _that particular (edge) device_ will be created, i.e. a mapping of that edge device to the adapter instance the gateway is connected to. 

Such a direct mapping of the command target device always has precedence when choosing the command target adapter instance. So in such a scenario the last-known-gateway information wouldn't be used for routing incoming commands.

The last-known-gateway information is only needed when multiple gateways have subscribed to receive commands for _any_ device that has the gateway in its 'via' property value.

clients/command-amqp/src/main/java/org/eclipse/hono/client/command/amqp/ProtonBasedCommandRouterClient.java
calohmn: To emphasize that it's a map where the iteration order is defined by the insertion order, hence also the `queue` part of the variable name.

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/DeviceConnectionInfo.java
calohmn: I don't see details in the Infinispan HotRod client docs on how `remoteCache#putAllAsync` behaves if any of the internal put operations failed.

sophokles73: hmm, then maybe we can document that, i.e.
> The outcome is indeterminate if any of the entries cannot be processed by an implementation. In such a case, client code should assume that none of the entries have been updated.

sophokles73 (approved):
LGTM

Issue #2735, opened 2021-06-15
Secret in credential can be added without password


mbaeuerle:
Observed behavior:
It is possible to create a credential with secrets which do not contain a password.

Expected behavior:
Creating a credential without any passwords should result in an error.

Steps to reproduce:

1. Make sure the credential is empty. If the credential is not empty to my knowledge everything works as expected
  ![grafik](https://user-images.githubusercontent.com/1345394/122050032-5d35a880-cde3-11eb-8ba5-78cf2eba381e.png)
1. Create a credential with one or more secrets just containing an arbitrary `id` and some other field, e.g. `not-after`
      ```json
      [
          {
              "type": "hashed-password",
              "auth-id": "auth-my-test-device",
              "secrets": [
                  {
                      "id": "1065959d-8541-4c71-b475-58f248464bfd",
                      "not-after": "2020-12-31T23:00:00Z"
                  },
                  {
                      "id": "26ffefe0-c0ec-4d42-af2a-70c4e0925a61",
                      "not-after": "2020-12-31T23:00:00Z"
                  }
              ],
              "enabled": true
          }
      ]
      ```
    ![grafik](https://user-images.githubusercontent.com/1345394/122050072-69216a80-cde3-11eb-83b6-7c68a9add701.png)
1. The request succeeds and the credential is stored in the database without passwords
  ![grafik](https://user-images.githubusercontent.com/1345394/122050101-70e10f00-cde3-11eb-9511-c81429bfdd1d.png)

sophokles73:
This problem also exists in the JDBC based and the file based registry.

PR #2744, opened 2021-06-24
Fix updating empty set of credentials

sophokles73:
This addresses #2735 

The registry implementations failed to reject a request to update a
device's empty set of credentials, e.g. right after the device has been
created, if the request contained a secret having an ID. In fact, the
registry implementations would accept such a request even if the secrets
did not contain a password/key etc at all but only the identifier.

The registries have been changed to now explicitly check if a device's
credentials on record are empty before merging in updated credentials.


services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/CredentialsDto.java
```diff
@@ -201,25 +242,30 @@ public void createMissingSecretIds() {
...
         Objects.requireNonNull(otherCredentialsDto);
 
-        Optional.ofNullable(otherCredentialsDto.getCredentials())
-                .ifPresent(credentialsToMerge -> this.getData()
-                        .forEach(credential -> findCredentialByIdAndType(credential.getAuthId(), credential.getType(),
-                                credentialsToMerge)
-                                        .ifPresent(credential::merge)));
+        final var credentialsToMerge = otherCredentialsDto.getCredentials();
+        if (credentialsToMerge == null || credentialsToMerge.isEmpty()) {
+            throw new IllegalArgumentException("DTO does not contain any credentials");
```
kaniyan, line 258: With respect to issue #2735, FMPOV the issue is that we are able to update with secret-ids that do not exist already provided the existing credentials list is empty. If the existing credentials list is not empty, then currently we send the error message `other credential has no secret with id: xyz`. But with this fix, in the case of empty list of existing credentials the error message is `DTO does not contain any credentials` and IMHO it is bit misleading. Shouldn't we keep it consistent with the other one?

kaniyan (changes requested):

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/CredentialsDto.java
sophokles73: What is misleading about it? We could still change it to e.g.
> no credentials on record for device

Or did you have anything specific in mind?

kaniyan: FMPOV the issue is that a user is trying to update with secret-ids that do not exist. In this scenario, IMHO an error message such as  `no secrets are available for the given secret ids` or `other credential has no secret with id: xyz` (in consistent with the other scenario). Also `no credentials on record for device` suits too.

sophokles73: I have rebased and updated the error message ...

kaniyan (approved):
LGTM

PR #2747, opened 2021-06-24
Support reading key store password from file

### LINKED ISSUE #2731: Read truststore password from file
Currently, a password can be provided for a JKS truststore file by setting the property `trustStorePassword`. The client configuration should additionally allow providing a path to a file that contains the password. This would allow reading it from a separate secret instead of putting it plain into a configuration file.

sophokles73:
The AbstractConfig class' trustStorePassword and keyStorePassword
properties now support reading the password from a file by means of
setting the property value to "file:/path/to/password"

The HTTP adapter used in the ITs has been configured to make use of this
feature.

Fixes #2731


sophokles73:
@b-abel would mind to take a look as well?

core/src/main/java/org/eclipse/hono/config/AbstractConfig.java
```diff
@@ -91,6 +103,45 @@ public AbstractConfig(final AbstractConfig other) {
...
+     * This method determines the password as follows:
+     * <ol>
+     * <li>If the given value does not start with {@value #PREFIX_FILE}, then the password is the given value.</li>
+     * <li>Otherwise the password is the UTF-8 encoded string represented by the first line read from the
+     * file indicated by the property's value after the {@value #PREFIX_FILE} prefix.</li>
+     * </ol>
+     *
+     * @param purpose A (very) short description of the context in which the password is being used.
+     * @param value The property value to determine the password from.
+     * @return The password.
```
kaniyan, line 118: Missing information about `NullPointerException` in the _JavaDoc_.

kaniyan (changes requested):

kaniyan (approved):
LGTM

core/src/main/java/org/eclipse/hono/config/AbstractConfig.java
```diff
@@ -91,6 +103,45 @@ public AbstractConfig(final AbstractConfig other) {
...
+        Objects.requireNonNull(value);
+
+        if (!value.startsWith(PREFIX_FILE)) {
+            return value;
+        }
+
+        final String fsPath = value.substring(PREFIX_FILE.length());
+        final File file = new File(fsPath);
+        if (!file.exists()) {
+            LOG.warn("cannot read {} password, file [{}] does not exist", purpose, fsPath);
```
b-abel, line 133: What if an (existing) password starts by accident with "file:"? (I admit that this is not very likely). One could return the value as a literal password if the file is not found. The downside would be of course that you don't have such a clear result in error cases.

sophokles73: Then it doesn't work :-) I am willing to take that risk. BTW the behavior is documented so users should be able to prevent this from happening by simply choosing a more appropriate password for their key store ...

b-abel:
LGTM

Issue #2592, opened 2021-04-09
Command responses need to be forwarded using correct messaging infrastructure


sophokles73:
The messaging infrastructure to be used by devices of a tenant can potentially be changed during runtime by updating the tenant configuration. The following might happen:
1. a protocol adapter receives a command request via AMQP
1. the adapter sends the command request to the device
1. the preference is changed from AMQP to Kafka
1. the device sends its command response to the adapter
1. the adapter determines the messaging infrastructure to be used for devices of the tenant as Kafka and forwards the command response to the Kafka broker
1. the downstream application which had sent the request via AMQP never gets the response

sophokles73:
@calohmn @b-abel @kaniyan Can you confirm that this might be an issue?

calohmn:
I think whether we should see this as an issue depends on the kinds of requirements we associate with changing the messaging network preference from AMQP to Kafka.

I think one could argue that as soon as this switch is done, the northbound application has to have the corresponding Kafka consumers in place and the Kafka topics for the tenant must exist. If that is the case, the northbound application can receive the command response message. Of course the application has to be able to correlate the response message to the original command message (given the correlation-id header of the command response record), which might be a challenge depending on the implementation.

In order to send the command response via AMQP in the above scenario, we would first have to encode the information about which messaging system to be used in the request-id that is part of the command sent to the device. When getting that request-id as part of the command-response in the adapter, the corresponding messaging system would be used.
But what if the command-response comes a long time of the command was sent - when the northbound application has already closed the AMQP command response receiver?

I would prefer, if we could require the northbound application to be able to already receive the Kafka-based command response - or the northbound application has to send the command again, that time via Kafka.

b-abel:
> I would prefer, if we could require the northbound application to be able to already receive the Kafka-based command response - or the northbound application has to send the command again, that time via Kafka.

I agree.
Does Hono have a hard guarantee that every command response will be delivered? FMPOV, the messaging infrastructure to be used should not be changed regularly, but at most once per tenant. Therefore, I would like to keep the logic that supports such a one-time event to an absolute minimum.

sophokles73:
This is all fine as long as we make sure that an application (and Ditto in particular) has a chance to detect the problem and act accordingly.

calohmn:
As discussed in the Hono community call:
The preferred solution here is to send the command response using the same messaging network that the command was received from. The used messaging system therefore has to be encoded in the request id contained in the command forwarded to the device.

[...]

PR #2753, opened 2021-06-29
[#2736][#2592] Support receiving commands via both AMQP and Kafka

### LINKED ISSUE #2736: Support receiving AMQP commands after tenant switched to Kafka

calohmn:
This fixes #2736 and #2592:
If both AMQP and Kafka messaging systems are configured, then the command router will receive command messages via both
systems.
The used messaging system is encoded in the command message forwarded to the device (as part of the request id or the command address in case of an AMQP device).
A command response message will then be forwarded using the messaging system that the command was received in. If that system isn't configured any more, the (tenant/global) default messaging system will be used.

### DIFF
--- a/adapter-base/src/main/java/org/eclipse/hono/adapter/ProtocolAdapter.java
+++ b/adapter-base/src/main/java/org/eclipse/hono/adapter/ProtocolAdapter.java
@@ -23,6 +23,7 @@ import org.eclipse.hono.client.registry.CredentialsClient;
 import org.eclipse.hono.client.registry.TenantClient;
 import org.eclipse.hono.client.telemetry.EventSender;
 import org.eclipse.hono.client.telemetry.TelemetrySender;
+import org.eclipse.hono.util.MessagingType;
 import org.eclipse.hono.util.RegistrationAssertion;
 import org.eclipse.hono.util.TelemetryExecutionContext;
 import org.eclipse.hono.util.TenantObject;
@@ -98,11 +99,16 @@ public interface ProtocolAdapter {
 
     /**
      * Gets the client being used for sending command response messages downstream.
+     * <p>
+     * Per default, the client using the given messaging type will be chosen. Only if
+     * no such client is available, the client will be determined using the messaging
+     * type configured for the given tenant (if set) or using the overall default type.
      *
+     * @param messagingType The type of messaging system to use.
      * @param tenant The tenant for which to send messages.
      * @return The sender.
      */
-    CommandResponseSender getCommandResponseSender(TenantObject tenant);
+    CommandResponseSender getCommandResponseSender(MessagingType messagingType, TenantObject tenant);
 
         /**
      * Gets an assertion of a device's registration status.
--- a/clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedCommandResponseSender.java
+++ b/clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedCommandResponseSender.java
@@ -60,6 +60,9 @@ public class KafkaBasedCommandResponseSender extends AbstractKafkaBasedMessageSe
 
         final HonoTopic topic = new HonoTopic(HonoTopic.Type.COMMAND_RESPONSE, response.getTenantId());
         log.trace("publish command response [{}]", response);
+        if (response.getMessagingType() != getMessagingType()) {
+            // TODO log to span
+        }
 
         return sendAndWaitForOutcome(topic.toString(), response.getTenantId(), response.getDeviceId(),
                 response.getPayload(), getHeaders(response), context);
--- a/clients/command/src/main/java/org/eclipse/hono/client/command/CommandResponse.java
+++ b/clients/command/src/main/java/org/eclipse/hono/client/command/CommandResponse.java
@@ -16,6 +16,7 @@ package org.eclipse.hono.client.command;
 import java.util.Objects;
 import java.util.function.Predicate;
 
+import org.eclipse.hono.util.MessagingType;
 import org.eclipse.hono.util.Pair;
 import org.eclipse.hono.util.ResourceIdentifier;
 import org.slf4j.Logger;
@@ -42,6 +43,7 @@ public final class CommandResponse {
     private final int status;
     private final String correlationId;
     private final String replyToId;
+    private final MessagingType messagingType;
 
     /**
      * Creates a command response.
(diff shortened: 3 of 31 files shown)


clients/command/src/main/java/org/eclipse/hono/client/command/Commands.java
```diff
@@ -107,76 +120,112 @@ public static String getRequestId(final String correlationId) {
...
+
+    private static int getMessagingTypeBitmask() {
+        // currently only encoding 2 messaging types here (BITFLAG_WIDTH_MESSAGING_TYPE is 1)
+//        return IntStream.range(BITFLAG_INDEX_MESSAGING_TYPE, BITFLAG_INDEX_MESSAGING_TYPE + BITFLAG_WIDTH_MESSAGING_TYPE)
+//                .reduce(0, (mask, index) -> mask | (1 << index));
+        return 1 << BITFLAG_INDEX_MESSAGING_TYPE;
+    }
+
+    private static int getMessagingTypeIndex(final MessagingType messagingType) {
+        return messagingType.equals(MessagingType.kafka) ? 1 : 0;
```
sophokles73, line 225: how about using MessagingType.ordinal() instead?

clients/command/src/main/java/org/eclipse/hono/client/command/CommandResponse.java
```diff
@@ -231,6 +241,15 @@ public int getStatus() {
         return status;
     }
 
+    /**
+     * Gets the type of the messaging system on which the command was received.
+     *
+     * @return The messaging type.
+     */
+    public MessagingType getMessagingType() {
```
sophokles73, line 249: how about adding the messaging type to the output of *toString()* as well?

clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedCommandResponseSender.java
```diff
@@ -60,6 +60,9 @@ public KafkaBasedCommandResponseSender(
 
         final HonoTopic topic = new HonoTopic(HonoTopic.Type.COMMAND_RESPONSE, response.getTenantId());
         log.trace("publish command response [{}]", response);
+        if (response.getMessagingType() != getMessagingType()) {
+            // TODO log to span
```
sophokles73, line 64: did you leave this out intentionally?

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/AbstractMappingAndDelegatingCommandHandler.java
```diff
@@ -103,8 +114,21 @@ public AbstractMappingAndDelegatingCommandHandler(
...
+                .compose(tenantObject -> {
+                    // check whether the handler messaging type is equal to the messaging type of the tenant (if set)
+                    final MessagingType tenantMessagingType = Optional
+                            .ofNullable(tenantObject.getProperty(TenantConstants.FIELD_EXT, JsonObject.class))
+                            .map(ext -> ext.getString(TenantConstants.FIELD_EXT_MESSAGING_TYPE))
+                            .map(MessagingType::valueOf).orElse(null);
+                    if (tenantMessagingType != null && getMessagingType() != tenantMessagingType) {
+                        log.info("command received via {} but tenant is configured to use {} [{}]", getMessagingType(),
+                                tenantMessagingType, commandContext.getCommand());
+                        commandContext.getTracingSpan().log(String.format("command received via %s but tenant is configured to use %s",
```
sophokles73, line 126: FMPOV we should log these two values to the span in any case at the beginning of the method. During migration this will probably become handy to track down routing problems ...

clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedCommandResponseSender.java
calohmn: Yes. The span is currently created in the parent class as part of the following `sendAndWaitForOutcome` invocation. I want to change that in a followup commit, creating the span here already, so that also a specific span name can be used.

adapter-base/src/main/java/org/eclipse/hono/adapter/ProtocolAdapter.java
```diff
@@ -99,10 +100,11 @@
     /**
      * Gets the client being used for sending command response messages downstream.
      *
+     * @param messagingType The type of messaging system to use.
```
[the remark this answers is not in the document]
calohmn: Yes, I'll add:
````
     * Per default, the client using the given messaging type will be chosen. Only if
     * no such client is available, the client will be determined using the messaging
     * type configured for the given tenant (if set) or using the overall default type.
````

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/AbstractMappingAndDelegatingCommandHandler.java
calohmn: I've extended the span log message created in `Command#logtoSpan` ("received command message via AMQP/Kafka").
The tenant default messaging type is only available after having fetched the tenant object, so I would keep it here. Whether this gets logged a few milliseconds earlier or later doesn't matter much anyway, I think.

clients/command/src/main/java/org/eclipse/hono/client/command/Commands.java
sophokles73: I see

clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedCommandResponseSender.java
sophokles73: ok

sophokles73 (approved):
LGTM

PR #2755, opened 2021-06-30
Improve HTTP/CoAP C&C tracing

calohmn:
This adds a `FOLLOWS_FROM` span reference from the HTTP/CoAP ttd request trace to the command-handling trace started in the Command Router.
Also decouples the closing of the command consumer from the request handling trace spans by means of an extra `FOLLOWS_FROM` span, the motivation being that the request handling trace span ("POST"/"PUT") gets finished before the span for closing the command consumer anyway.

Before:
HttpSpan_Before

With this PR:
HttpSpan_link_to_CmdRouter

### DIFF
--- a/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractHonoResource.java
+++ b/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractHonoResource.java
@@ -50,6 +50,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.micrometer.core.instrument.Timer.Sample;
+import io.opentracing.References;
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
 import io.opentracing.Tracer;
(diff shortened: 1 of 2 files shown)


adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/AbstractHonoResource.java
```diff
@@ -562,6 +571,7 @@ protected void addCommandToResponse(
                         commandSample);
                 LOG.debug("command message is invalid: {}", command);
                 commandContext.reject("malformed command message");
+                processCommandSpan.finish();
```
sophokles73, line 574: can't we simply pull this out of the top level *if* statement?

calohmn: Not really - the first of the 3 `processCommandSpan.finish()` invocations (in line 548) is inside an `onComplete` block, so it's invoked some time later.

sophokles73: oh, right

sophokles73 (approved):
LGTM

PR #2756, opened 2021-06-30
Use specific tracing span names when publishing Kafka records

calohmn:
Span names for sending messages downstream using Kafka are now the same as when using AMQP (not always "send message" anymore).


sophokles73 (commented):
Is this relevant to Kafka only, or are you planning to create a corresponding PR for AMQP as well?

calohmn:
> Is this relevant to Kafka only, or are you planning to create a corresponding PR for AMQP as well?

This is only relevant for Kafka. The span names used in this PR are the ones already used for the corresponding AMQP tracing spans.

sophokles73 (approved):
LGTM

PR #2758, opened 2021-07-01
Skip error command response if partition revoked

calohmn:
Related to  #2751:
Skips sending an error command response message in case command handling got cancelled because the topic partition assignment got revoked. In this case the command will be handled again by the other consumer that the topic partition got newly assigned to.


calohmn:
@kaniyan I've rebased and removed the changes in the `KafkaBasedCommandConsumerFactoryImplIT` class. They made the test unstable (because of possible interference with the consumer from the integration tests command router component).
I'll see whether to integrate the added test logic for the scenario here in that class at a later point.
For now, there is the `KafkaCommandProcessingQueueTest` unit test to test the changes.

kaniyan (approved):
LGTM

PR #2759, opened 2021-07-01
Improve Command Router C&C tracing

calohmn:
This adds a `FOLLOWS_FROM` span reference from the "handle command" span in the command handling trace started in the Command Router to the span in the preceding trace where the command consumer was created in the protocol adapter.
This is only enabled for short-lived command consumers, i.e. those that get created in the HTTP/CoAP adapters for ttd requests.

CommandRouter_http_coap_spanRef

Compared with the span reference added in  #2755, this is the link in the other direction.

### DIFF
--- a/clients/command/src/main/java/org/eclipse/hono/client/command/CommandContext.java
+++ b/clients/command/src/main/java/org/eclipse/hono/client/command/CommandContext.java
@@ -23,6 +23,7 @@ import org.eclipse.hono.tracing.TracingHelper;
 import org.eclipse.hono.util.ExecutionContext;
 import org.eclipse.hono.util.MessageHelper;
 
+import io.opentracing.References;
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
 import io.opentracing.Tracer;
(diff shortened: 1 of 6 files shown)


clients/command/src/main/java/org/eclipse/hono/client/command/CommandContext.java
```diff
@@ -117,11 +118,14 @@ default void reject(final String error) {
      *
      * @param tracer The tracer to use.
      * @param command The command for which the span should be started.
-     * @param spanContext Existing span context.
+     * @param spanContext The span context to set as parent span.
```
sophokles73, line 121: how about renaming this to *parentSpan* then ...

clients/command/src/main/java/org/eclipse/hono/client/command/CommandContext.java
```diff
@@ -117,11 +118,14 @@ default void reject(final String error) {
...
+     * @param followsFromSpanContext A span context for which to add a <em>follows-from</em> reference,
```
sophokles73, line 122: ... and this could simply be *followsFromSpan*

clients/command/src/main/java/org/eclipse/hono/client/command/CommandHandlerWrapper.java
```diff
@@ -42,15 +44,21 @@
      *                  parameter.)
      * @param commandHandler The command handler.
      * @param context The vert.x context to run the handler on or {@code null} to invoke the handler directly.
+     * @param consumerCreationSpanContext The span context of the command consumer creation operation to reference as
+     *                                    <em>follow-from</em> span in the command handling spans or {@code null} if
+     *                                    that context is not available or shouldn't be used. Note that providing the
+     *                                    context here only makes sense for short-lived command consumers.
```
sophokles73, line 50: how about
> The span for tracking the creation of the command consumer that this wrapper belongs to.
If not {@code null}, then the spans created by this wrapper for handling command request messages will have a <em>follows-from</em> reference to the given span. Note that providing a span context here only makes sense for short-lived command consumers as created by the HTTP and CoAP adapters for example.

clients/command/src/main/java/org/eclipse/hono/client/command/CommandContext.java
calohmn: Renamed to `parentSpanContext` - I'd rather make it clear that it is a span _context_, not a span here (as used in some other places as method parameter).

clients/command/src/main/java/org/eclipse/hono/client/command/CommandContext.java
calohmn: see above

clients/command/src/main/java/org/eclipse/hono/client/command/CommandHandlerWrapper.java
calohmn: I've adopted parts and rephrased others.

sophokles73 (approved):
LGTM

PR #2762, opened 2021-07-01
Add Credentials DAO for accessing credentials collection

### LINKED ISSUE #2110: Clean up class structure of Device Registry
Below is the class structure for the device management part of the file based device registry which relies heavily on inheritance instead of aggregation.

image

I believe that we have reached a point where we need to first unclutter the base classes before adding more functionality. In particular, I believe that the *FileBasedRegistrationService* implementing both *DeviceRegistrationService* and *DeviceManagementService* at the same time is problematic. The same is true for the *FileBasedCredentialsService* which implements *CredentialsService* and at the same time *CredentialsManagementService*. Analogously this applies to the *FileBasedTenantService*.

I think that we would be better off with a dedicated DAO layer for accessing the data store and then have independent implementation classes for *DeviceRegistrationService* and *DeviceManagementService*. IMHO it would then be possible to always use dedicated instances for the aggregating AMQP and HTTP endpoints, i.e. they would never need to share the same instance of a collaborator. The only synchronization necessary would then be at the DAO level as can be seen in the diagram below which shows how the classes could be re-structured.

image

Note that we would also no longer need the *DeviceBackend* interface and the corresponding implementation.

sophokles73:
Also moved credentials management related functionality from
MongoDbBasedCredentialsService to separate
MongoDbBasedCredentialsManagementSerice.

Also removed MongoDbBasedDeviceBackend and MongoDbCallExecutor as they
are no longer needed.

Fixes #2110


sophokles73:
@kaniyan would you mind taking a look?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedCredentialsDao.java
```diff
@@ -0,0 +1,460 @@
...
+                .addReference(References.CHILD_OF, tracingContext)
+                .withTag(TracingHelper.TAG_TENANT_ID, tenantId)
+                .withTag(TracingHelper.TAG_DEVICE_ID, deviceId)
+                .start();
+        resourceVersion.ifPresent(v -> span.setTag("resource_version", v));
+
+        final JsonObject removeCredentialsQuery = MongoDbDocumentBuilder.builder()
+                .withTenantId(tenantId)
+                .withDeviceId(deviceId)
+                .document();
```
kaniyan, line 434: `resourceVersion` is unused while deleting the entry from the database.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedDao.java
```diff
@@ -101,6 +105,25 @@ public final void close() {
...
     }
 
+    /**
+     * Checks if the given error is caused due to duplicate keys.
+     *
+     * @param error The error to check.
+     * @return {@code true} if the given error is caused by duplicate keys.
+     * @throws NullPointerException if the error is {@code null}.
+     */
+    protected static boolean isDuplicateKeyError(final Throwable error) {
```
kaniyan, line 115: By moving these methods to this class, `MongoDbDeviceRegistryUtils` is no more used and we can remove it.

services/device-registry-mongodb/src/test/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedCredentialServiceTest.java
```diff
@@ -170,44 +169,12 @@ public CredentialsService getCredentialsService() {
...
-    }
-
-    /**
-     * Verifies that a request to add credentials for a device fails with a 403 status code
-     * if the number of credentials exceeds the tenant's configured limit.
-     *
-     * @param ctx The vert.x test context.
-     */
-    @Test
-    public void testAddCredentialsFailsForExceededCredentialsPerDeviceLimit(final VertxTestContext ctx) {
```
kaniyan, line 188: Don't we need this test anymore?

kaniyan (approved):

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedCredentialsDao.java
```diff
@@ -0,0 +1,460 @@
...
+                                .partialFilterExpression(new JsonObject()
+                                        .put(KEY_AUTH_ID, new JsonObject().put("$exists", true))
+                                        .put(KEY_CREDENTIALS_TYPE, new JsonObject().put("$exists", true)))))
+                .onSuccess(ok -> indicesCreated.set(true))
+                .onComplete(r -> {
+                    creatingIndices.set(false);
+                    result.handle(r);
+                });
+        } else {
+            LOG.debug("already trying to create indices");
```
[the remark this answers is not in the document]
sophokles73: actually I think that we should adapt the other DAOs to use the approach used here. If the indices are already being created then we simply let the client wait for its completion based on the existing future instead of telling it that the indices cannot be created. Doesn't that make more sense?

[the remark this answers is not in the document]
kaniyan: It makes sense to wait instead of throwing `ConcurrentModificationException`.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedCredentialsDao.java
sophokles73: Hmm, we do remove credentials as part of deleting a device only, so in that case the credentials are being removed *unconditionally*. However, the DAO exposes generic CRUD operations and we should probably support optimistic locking in the same way here as we do elsewhere so I would tend to modifying the code to use the resource version instead of ignoring it. WDYT?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedDao.java
sophokles73: Absolutely. I thought I already did ...

services/device-registry-mongodb/src/test/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedCredentialServiceTest.java
sophokles73: I don't think so. The MongoDbBasedCredentialsManagementService does not have an *addCredentials* method anymore and when we create a device, an empty set of credentials is being created ...

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedCredentialsDao.java
kaniyan: Since DAO exposes generic CRUD operations, I think so too that we should make use of the resource version in the delete query to support optimistic locking.

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedCredentialsDao.java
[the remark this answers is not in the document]
sophokles73: Ok, then I will adapt the other DAOs accordingly ...

sophokles73:
@kaniyan I have pushed corresponding changes ...

kaniyan (approved):
LGTM

PR #2771, opened 2021-07-05
[#2767] Add rate-limit to AsyncHandlingAutoCommitKafkaConsumer

### LINKED ISSUE #2767: Limit rate at which Command Router handles Kafka command records
Currently there is no limit to the rate with which the Command Router receives and handles incoming Kafka command records (as provided via the `poll` operations of the `KafkaConsumer`).
This may lead to errors and eventually an OOM error if the records can't be processed at the same rate they are read (e.g. due to the device registry link hitting a `no credits` error).

Therefore the polling operations should be done in such a way, that not too many records are processed at the same time.

calohmn:
This fixes #2767.

### DIFF
--- a/clients/kafka-common/src/test/java/org/eclipse/hono/client/kafka/consumer/AsyncHandlingAutoCommitKafkaConsumerTest.java
+++ b/clients/kafka-common/src/test/java/org/eclipse/hono/client/kafka/consumer/AsyncHandlingAutoCommitKafkaConsumerTest.java
@@ -16,6 +16,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


clients/kafka-common/src/test/java/org/eclipse/hono/client/kafka/consumer/AsyncHandlingAutoCommitKafkaConsumerTest.java
```diff
@@ -159,6 +160,76 @@ public void testConsumerCreationWithTopicPatternSucceeds(final VertxTestContext
...
+            // schedule the poll tasks
+            for (int i = 0; i < numTestBatches; i++) {
+                mockConsumer.schedulePollTask(() -> {
+                    for (int j = 0; j < maxPollRecords; j++) {
+                        final int offset = offsetCounter.getAndIncrement();
+                        mockConsumer.addRecord(new ConsumerRecord<>(TOPIC, PARTITION, offset, "key_" + offset, Buffer.buffer()));
+                    }
+                });
+            }
+            final long timerId = vertx.setTimer(8000, tid -> {
```
sophokles73, line 216: is it really necessary to wait for this task for 8 seconds? Can't we use a mocked vert.x instance instead?

calohmn: The 8s here are just the timeout value (lower than the overall test timeout). Using this extra timer makes it possible to print some extra output in case this timeout occurs here.
And better to do this on the vert.x event loop thread. The alternative of using a `CountdownLatch` on the `main` thread seemed problematic - doing some things on the event loop thread and others on the `main` thread.

And concerning the vert.x instance, I think it wouldn't be easy to use a mocked instance with the `KafkaConsumer`. And it would introduce a change in behaviour in the KafkaConsumer compared with using a proper vert.x instance. So, I wouldn't want to use a mocked instance here.

sophokles73: I see

sophokles73 (approved):
LGTM

PR #2778, opened 2021-07-08
Reduce complexity of DTO usage

sophokles73:
This is a follow-up of #2776

The *deviceStatus* property of the DeviceDto has been replaced by explicit *autoProvisioned* and * autoProvisioningNotificationSent* properies. This also makes the *MongoDbBasedDeviceDto* obsolete.


calohmn (approved):
LGTM

Issue #2777, opened 2021-07-08
Integration tests module compilation takes very long


calohmn:
Something that has baffled me for quite a while is how long it takes to compile the "tests" Hono module with the integration tests.
On my machine that takes about 60 seconds (for just 63 classes)!
This means there is quite some patience needed when developing integration tests.

I've tried to narrow it down a bit, to see which classes take up the most time.
Removing everything except the `org.eclipse.hono.tests.http.HttpTestBase` class and classes it depends on leads to a compilation time of still about 40 seconds.

Testing some more, the compilation time is increased by several seconds when duplicating the last `assertThat` line multiple times here:
https://github.com/eclipse/hono/blob/969751067daf83a1e4f6dfde5b0b0c7f7cbe5c13/tests/src/test/java/org/eclipse/hono/tests/http/HttpTestBase.java#L1300-L1310

Narrowing it down to just this piece of code
```
    public void test(final java.util.Optional<java.util.List<String>> listOptional) {
        listOptional
                .map(list -> {
                    verify(() -> {
                        org.assertj.core.api.Assertions.assertThat(list.get(1)).isNull();
                        org.assertj.core.api.Assertions.assertThat(list.get(1)).isNull();
                        org.assertj.core.api.Assertions.assertThat(list.get(1)).isNull();
                        org.assertj.core.api.Assertions.assertThat(list.get(1)).isNull();
                        org.assertj.core.api.Assertions.assertThat(list.get(1)).isNull();
                        org.assertj.core.api.Assertions.assertThat(list.get(1)).isNull();
                        org.assertj.core.api.Assertions.assertThat(list.get(1)).isNull();
                        org.assertj.core.api.Assertions.assertThat(list.get(1)).isNull();
                        org.assertj.core.api.Assertions.assertThat(list.get(1)).isNull();
                        org.assertj.core.api.Assertions.assertThat(list.get(1)).isNull();
                    });
                    return list;
                });
    }
    private void verify(final Runnable block) {
        //
    }
```
I arrive at 6 seconds:
```
javac -verbose -version -cp [path]/assertj-core-3.20.2.jar:. Test.java
javac 14.0.2
...
[total 6073ms]
```
Trying out other JDK (versions), there is no significant difference.

This all leads me to believe that the combination of lambdas + generics + the heavily overloaded AssertJ `Assertions.assertThat` methods lead to these long compilation times. Using the latest AssertJ version (3.20.2) has brought no difference.

[...]

PR #2779, opened 2021-07-09
[#2777] Use Truth assertThat() instead of AssertJ in 'tests' module

calohmn:
For #2777:
This makes compilation of the 'tests' module up to 5x faster.

I've kept a few AssertJ `assertThat()` invocations where objects get compared recursively.

The adaptations made here are either related to
- the differences in the assertion methods (e.g. `containsAllEntriesOf` vs. `containsExactlyEntriesIn`) 
- or the different way of handling an `assertThat` description message - `assertThat(...).as("message")` vs. `assertWithMessage("message").that(...)`

Concerning these `assertThat` messages, I've rephrased them in most cases, so that they fit the format of the AssertionError output, which is e.g.:
```
myAssertThatMessage
expected to be true
---
myAssertThatMessage
expected to be greater than: 7
but was                    : 5
---
myAssertThatMessage
expected not to be: null
---
myAssertThatMessage
expected to contain: sdss
but was            : test
```

Complete example:
AssertJ:
```
final UnsignedLong maxMessageSize = s.getRemoteMaxMessageSize();
assertThat(maxMessageSize).as("check adapter's attach frame includes max-message-size").isNotNull();
```
produces:
```
java.lang.AssertionError: 
java.lang.AssertionError: [check adapter's attach frame includes max-message-size] 
Expecting actual not to be null
Caused by: java.lang.AssertionError: 
[check adapter's attach frame includes max-message-size] 
Expecting actual not to be null
```

Truth:
```
final UnsignedLong maxMessageSize = s.getRemoteMaxMessageSize();
assertWithMessage("max-message-size included in adapter's attach frame")
        .that(maxMessageSize).isNotNull();
```
produces:
```
java.lang.AssertionError: 
max-message-size included in adapter's attach frame
value of          : getRemoteMaxMessageSize()
expected not to be: null
Caused by: com.google.common.truth.AssertionErrorWithFacts: 
max-message-size included in adapter's attach frame
value of          : getRemoteMaxMessageSize()
expected not to be: null
```

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -92,6 +92,7 @@
     <spring-boot.version>2.4.8</spring-boot.version>
     <spring-security-crypto.version>5.4.7</spring-security-crypto.version>
     <testcontainers.version>1.15.2</testcontainers.version>
+    <truth.version>1.1.3</truth.version>
     <!-- when updating Vert.x, check if the Kafka clients need to be updated as well -->
     <vertx.version>3.9.7</vertx.version>
     <wildfly-common.version>1.5.4.Final-format-001</wildfly-common.version>
@@ -1021,6 +1022,19 @@
         <artifactId>assertj-core</artifactId>
         <version>${assertj-core.version}</version>
         <scope>test</scope>
+      </dependency>
+      <dependency>
+        <groupId>com.google.truth</groupId>
+        <artifactId>truth</artifactId>
+        <version>${truth.version}</version>
+        <scope>test</scope>
+        <!-- exclude JUnit 4, see https://github.com/google/truth/issues/333 -->
+        <exclusions>
+          <exclusion>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+          </exclusion>
+        </exclusions>
       </dependency>
       <dependency>
         <groupId>org.hamcrest</groupId>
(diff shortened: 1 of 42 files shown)


sophokles73 (approved):
LGTM

Can you also please create a corresponding Build-/Test-Only CQ

sophokles73:
@calohmn we should probably unschedule this from 1.9.0 until the CQ has been approved ...

calohmn:
@sophokles73 I don't think we need a CQ here. Using the `create_dependencies_list.sh` script, all added entries seem to be covered either by a ClearlyDefined or CQ entry already:
````
maven/mavencentral/com.google.auto.value/auto-value-annotations/1.8.1, Apache-2.0, approved, clearlydefined
maven/mavencentral/com.google.errorprone/error_prone_annotations/2.2.0, Apache-2.0, approved, clearlydefined
maven/mavencentral/com.google.truth/truth/1.1.3, Apache-2.0, approved, clearlydefined
maven/mavencentral/org.ow2.asm/asm/9.1, BSD-3-Clause, approved, CQ23029
````
To make that explicit, I could already add a `Truth` "compile" dependency in one of the test-utils modules, and add the regenerated DPENDENCIES file here. WDYT?

PR #2781, opened 2021-07-14
Fix wrong wildfly elytron version used in NOTICE file

calohmn:
The actual wildfly elytron version used is `1.15.1.Final`.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -96,7 +96,6 @@
     <!-- when updating Vert.x, check if the Kafka clients need to be updated as well -->
     <vertx.version>3.9.7</vertx.version>
     <wildfly-common.version>1.5.4.Final-format-001</wildfly-common.version>
-    <wildfly-elytron.version>1.14.1.Final</wildfly-elytron.version>
     <zookeeper.image.name>confluentinc/cp-zookeeper:6.1.1</zookeeper.image.name>
 
     <!-- The port at which the Prometheus scraping endpoint is exposed -->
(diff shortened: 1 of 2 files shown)


calohmn:
@sophokles73 Or should we remove the NOTICE file entry since we don't explicitly declare the "wildfly-elytron" dependency in the POM anymore (`wildfly-elytron.version` is currently only used in the NOTICE file)?

sophokles73:
I would suggest to remove the version definition from the `bom/pom.xml` file and remove the version reference from the `legal/src/main/resources/legal/NOTICE.md` file and instead simply refer to the wildfly project on GitHub (without any particular version).

sophokles73 (approved):
LGTM

calohmn:
> I would suggest to remove the version definition from the `bom/pom.xml` file and remove the version reference from the `legal/src/main/resources/legal/NOTICE.md` file and instead simply refer to the wildfly project on GitHub (without any particular version).

Ok, yes. I've updated the PR accordingly.

---
In general, I've had a closer look at the `wildfly-elytron` dependency here.

In Hono, the used Infinispan version is `11.0.9.Final` which uses `wildfly-elytron:1.12.1.Final`.

But the dependency tree of the Hono `client-device-connection-infinispan` module shows that actually `wildfly-elytron:1.15.1` is currently being used.
That is because the `quarkus-bom:1.13.7.Final` dependency in Hono uses Infinispan `12.0.1.Final`, which itself uses `wildfly-elytron:1.15.1`. That means with the update from quarkus 1.11.7 to 1.13.7 in Hono we got the newer wildfly-elytron version while keeping the 11.0.9 Infinispan version because we override only the Infinispan version explicitly.

It seems that the combination of Infinispan 11.0,.9 with the newer wildfly-elytron version works in at least the standard Hono deployment configuration. I'm not sure though whether that is the case for all possible custom Hotrod client configurations.

In general, I think we should upgrade to Infinispan 12.0.1 in Hono to prevent potential issues here.

sophokles73:
> In general, I think we should upgrade to Infinispan 12.0.1 in Hono to prevent potential issues here.

Sounds good. That would be an additional PR which is independent from this one, right?

calohmn:
yes

Issue #2727, opened 2021-06-11
Migrate Sandbox deployment to Helm chart


sophokles73:
The Sandbox currently uses a shell script and the Docker command line client to deploy Hono to Docker Swarm running on the Sandbox machine. It would be much more convenient and more consistent with our general approach for deploying Hono if we could run Kubernetes on the Sandbox (e.g. [k3s](https://k3s.io/)) and then simply use the Hono Helm chart to deploy.

sophokles73:
@b-abel I think we should get this done before adding kafka to the Sandbox (#2683).

b-abel:
I would like to work on this issue.

PR #2782, opened 2021-07-15
[#2727] Improve Sandbox deployment scripts.

b-abel:
This moves the commands for creating a certificate into another script than the one for installing cert-manager. While the cert-manager installation should be performed only once, requesting certificates may happen more often. The commit contains other small improvements to the sandbox deployment.


b-abel:
@sophokles73 Would you mind taking a look?

deploy/src/main/sandbox-k8s/10-deploy-cert-manager.sh
```diff
@@ -12,43 +12,20 @@
...
+helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --set installCRDs=true --wait --timeout 5m0s
 
-# checking when cert-manager is ready is tricky: https://cert-manager.io/docs/installation/kubernetes/#verifying-the-installation
-# we just retry the following command
 echo ""
-echo "Adding issuer Let's Encrypt to cert-manager with API: $LE_API"
-LE_CONFIG=$(helm template ./letsencrypt-chart -f ./letsencrypt-chart/values-$LE_API.yaml --set spec.acme.email=$EMAIL)
-retry echo "$LE_CONFIG" | kubectl apply -f - $KUBECONFIG || true # do not abort although '-e' is set
+echo "Creating namespace 'hono'..."
+kubectl create namespace hono
```
sophokles73, line 31: why do we create the *hono* namespace here? This doesn't seem to be related to the cert-manager, is it?

b-abel: Right, it seems a little misplaced. Therefore it is explicitly mentioned in the comments of the script and in the readme. The point is that the following script needs the namespace to create the secret for the certificate in it. Having an additional script only for creating the namespace seemed a little bit like overkill to me. But maybe that would be cleaner...

sophokles73: why don't we simply mention it in the readme that you need to create the namespace using `kubectl create namespace hono`?

b-abel: Done.

deploy/src/main/sandbox-k8s/30-deploy-hono.sh
```diff
@@ -12,17 +12,18 @@
...
+echo ""
+echo "Adding the Helm repository for Hono..."
+helm repo add eclipse-iot https://eclipse.org/packages/charts
+helm repo update
 
 echo ""
-echo "Deploying hono..."
-helm install eclipse-hono --dependency-update -f hono-values.yml -n hono eclipse-iot/hono $KUBECONFIG $HELM_WAIT
+echo "Deploying Hono..."
+helm install eclipse-hono --dependency-update -f $SCRIPTPATH/hono-values.yml -n hono eclipse-iot/hono --wait --timeout 5m0s
```
sophokles73, line 29: why don't we use the `--create-namespace` and `--namespace` options here as well (as we do when installing cert-manager)? We do not seem to be deploying the certificates to namespace *hono*, though, are we?

b-abel: The certificates are indeed put into a secret in the namespace _hono_. You can find this configuration in the file `letsencrypt-certificate.yml`.

deploy/src/main/sandbox-k8s/readme.md
```diff
@@ -9,17 +9,19 @@ This folder contains scripts and configuration files for setting up a Hono *sand
...
-execute the script again with `production` as the second argument to request a valid certificate from the productive API.
-
-The scripts need to be executed with `sudo` because the Kube config is only readable by root.
+The scripts need to be executed in the order of the numbers in the file names.
+
+* Install Kubernetes with: `./00-install-k3s.sh`. Export the Kubernetes config as an environment variable, 
+   e.g., like: `export KUBECONFIG=/etc/rancher/k3s/k3s.yaml`. 
+* Deploy the [cert-manager](https://cert-manager.io/): `./10-deploy-cert-manager.sh`.
+* Create the namespace "hono": `kubectl create namespace hono`.
+* Request a public certificate from Let's Encrypt: `./20-create-certificate.sh $LE_EMAIL`. Verify that the certificate
```
sophokles73, line 18: `$LE_EMAIL` ?

deploy/src/main/sandbox-k8s/30-deploy-hono.sh
sophokles73: I see

deploy/src/main/sandbox-k8s/readme.md
b-abel: Better now?

deploy/src/main/sandbox-k8s/readme.md
```diff
@@ -9,17 +9,21 @@ This folder contains scripts and configuration files for setting up a Hono *sand
...
+The scripts need to be executed in the order of the numbers in the file names.
+
+* Install Kubernetes with: `./00-install-k3s.sh`. Export the Kubernetes config as an environment variable, 
+   e.g., like: `export KUBECONFIG=/etc/rancher/k3s/k3s.yaml`. 
+* Deploy the [cert-manager](https://cert-manager.io/): `./10-deploy-cert-manager.sh`.
+* Create the namespace "hono": `kubectl create namespace hono`.
+* Requesting a public certificate from Let's Encrypt requires an email address that Let's Encrypt uses to contact you
+  Please make sure to replace "<your-email-address>" in the following commands by an email address that is actually used.
+  Request the certificate with: `./20-create-certificate.sh <your-email-address>`. Verify that the certificate
+  has successfully been issued with `kubectl describe certificates hono-eclipseprojects-io`. When this looks good,
```
sophokles73, line 21: this will probably not work if the certificates have been deployed to namespace *hono*, will it?

deploy/src/main/sandbox-k8s/readme.md
sophokles73: yes

deploy/src/main/sandbox-k8s/readme.md
b-abel: Right. My bad. I added the namespace.

sophokles73 (approved):
LGTM

Issue #2785, opened 2021-07-19
Run ITs using a dedicated Infinispan server


sophokles73:
The ITs are currently run using an embedded Infinispan cache in the Command Router and Device Connection components. However, in production a dedicated Infinispan server/cluster will usually be employed. It therefore seems reasonable to run ITs using a dedicated Infinispan server as well, in order to identify issues early on in the development process.

PR #2786, opened 2021-07-19
Run ITs with dedicated Infinispan server

sophokles73:
This addresses #2785

The integration tests now use a dedicated Infinispan server for storing
command routing information by default.

An additional profile for using an embedded cache instead has been
added.

One of the CI build configurations has been updated to make use of the
Infinispan server.


sophokles73:

PR #2788, opened 2021-07-20
Upgrade to Infinispan Hot Rod client 12.0.2

### LINKED ISSUE #2787: Upgrade to Infinispan 12.0

sophokles73:
Quarkus 1.13.7 uses Infinispan 12.0.1 so using 12.0.2 will reduce the
(potential) conflicts between dependencies defined by Quarkus and Hono.

I have also tested native image build and ran ITs with the native images.

Fixes #2787


legal/src/main/resources/legal/NOTICE.md
```diff
@@ -676,7 +676,7 @@ is also available at http://www.apache.org/licenses/LICENSE-2.0.html.
 
 The source code is available from [Maven Central](http://search.maven.org/remotecontent?filepath=io/vertx/vertx-web-client/${vertx.version}/vertx-web-client-${vertx.version}-sources.jar).
 
-### WildFly Common ${wildfly-common.version}
+### WildFly Common
```
kaniyan, line 679: In this PR, _wildfly-common_  dependency has been removed from the _bom_. Is there any reason why we should still specify it here?

kaniyan (changes requested):

legal/src/main/resources/legal/NOTICE.md
sophokles73: Removing it from the BOM doesn't mean that it is no longer used. It is still pulled in by the Hot Rod client but we no longer explicitly manage its version. So FMPOV it is still necessary for us to specify it in the NOTICE file.

kaniyan (approved):
LGTM

PR #2791, opened 2021-07-23
Return proper error detail in tenants resource responses

### LINKED ISSUE #2471: Device Registry: Error responses do not comply with the specification

```
      Error:
         type: object
         additionalProperties: true
         required:
            - error
         properties:
            "error":
               type: string
```

sophokles73:
This is the first PR in a series to fix #2471

"tenants" resources.

### DIFF
--- a/services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedTenantService.java
+++ b/services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedTenantService.java
@@ -18,7 +18,6 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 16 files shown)


services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedTenantService.java
```diff
@@ -360,64 +357,56 @@ private void addTenant(final JsonObject tenantToAdd) {
...
+                            tenantId,
+                            HttpURLConnection.HTTP_PRECON_FAILED,
+                            "resource Version mismatch"));
                 }
             } else {
                 TracingHelper.logError(span, "Tenant not found.");
-                return Result.from(HttpURLConnection.HTTP_NOT_FOUND);
+                return Future.failedFuture(new ClientErrorException(
+                        tenantId,
+                        HttpURLConnection.HTTP_NOT_FOUND));
```
kaniyan, line 378: How about also adding a corresponding error message in case of _HTTP_NOT_FOUND_ ? So that the error json has some meaningful message instead of `N/A`.

kaniyan (changes requested):

services/device-registry-file/src/main/java/org/eclipse/hono/deviceregistry/file/FileBasedTenantService.java
sophokles73: I have pushed corresponding changes

kaniyan (approved):
LGTM

PR #2794, opened 2021-07-27
Improve error tracing in MongoDB classes

calohmn:
A MongoDB error will now be traced before mapping it to a ServerErrorException, so that the 'error.object' in the trace contains more info than just "org.eclipse.hono.client.ServerErrorException: Error Code: 500".


sophokles73 (approved):
LGTM

PR #2796, opened 2021-07-27
Use index when finding credentials by type and auth-id

### LINKED ISSUE #2792: Finding credentials by tenant-id, auth-id and type does not use MongoDb index
The [MongoDbDocumentBuilder](https://github.com/eclipse/hono/blob/0f0e4cae45601a25f37109de42db6e319ee7743e/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDocumentBuilder.java#L140) class uses MongoDb's $elemMatch operator for filtering credentials by tenant-id, auth-id and type. For this purpose an index on those fields was created. When using MongoDb's explain it turned out that this index is in fact not used:

```
db.getCollection('device-credentials').find({
    "tenant-id": "foobar",
    "credentials": {
      "$elemMatch": {
        "auth-id": "spamegg",
        "type": "hashed-password"
      }
    }
  }).explain()
  
# RETURNS
{
    "queryPlanner" : {
        ...
        },
        "winningPlan" : {
            ...
            },
            "inputStage" : {
                "stage" : "IXSCAN",
                "keyPattern" : {
                    "tenant-id" : 1,
                    "device-id" : 1
                },
                "indexName" : "tenant-id_1_device-id_1",
                "isMultiKey" : false,
                "multiKeyPaths" : {
                    "tenant-id" : [],
                    "device-id" : []
                },
                "isUnique" : true,
                "isSparse" : false,
                "isPartial" : false,
                "indexVersion" : 2,
                "direction" : "forward",
                "indexBounds" : {
                    "tenant-id" : [ 
                        "[\"foobar\", \"foobar\"]"
                    ],
                    "device-id" : [ 
                        "[MinKey, MaxKey]"
                    ]
                }
            }
        },
        "rejectedPlans" : []
    },
    ...
}
```

Actually an index on tenant-id and device-id is used and a scan is done for all device ids (see the "[MinKey, MaxKey]" value for "device-id".
 
As far as I can see there is no need to the $elemMatch operator for this query as it is only a simple match of two values and not more complicated match resorting to other MongoDb funtions ($gt, $and, $or,...).
 
If the query is done without the $elemMatch operator the index is used as expected:

```
db.getCollection('device-credentials').find({
    "tenant-id": "foobar",
    "credentials.auth-id": "spamegg",
    "credentials.type": "hashed-password"
  }).explain()
  
# RETURNS
{
    "queryPlanner" : {
        ...
        },
        "winningPlan" : {
            "stage" : "FETCH",
            "filter" : {
                "credentials.type" : {
                    "$eq" : "hashed-password"
                }
            },
            "inputStage" : {
                "stage" : "IXSCAN",
                "keyPattern" : {
                    "tenant-id" : 1,
                    "credentials.auth-id" : 1,
                    "credentials.type" : 1
                },
                "indexName" : "tenant-id_1_credentials.auth-id_1_credentials.type_1",
                "isMultiKey" : true,
                "multiKeyPaths" : {
                    "tenant-id" : [],
                    "credentials.auth-id" : [ 
                        "credentials"
                    ],
                    "credentials.type" : [ 
                        "credentials"
                    ]
                },
                "isUnique" : true,
                "isSparse" : false,
                "isPartial" : true,
                "indexVersion" : 2,
                "direction" : "forward",
                "indexBounds" : {
                    "tenant-id" : [ 
                        "[\"foobar\", \"foobar\"]"
                    ],
                    "credentials.auth-id" : [ 
                        "[\"spamegg\", \"spamegg\"]"
                    ],
                    "credentials.type" : [ 
                        "[MinKey, MaxKey]"
                    ]
                }
            }
        },
        "rejectedPlans" : [ 
            ...
        ]
    },
    ...
}
```

From https://github.com/eclipse/hono/commit/d75a28e7276a055fd5dc72264f8016355cb80223 I understand that the $elemMatch operator was used since the credentials to be returned should not contain sensitive information. Hence MongoDb's projection operator was used. This did not work with a "plain" filter which is why the $elemMatch operator was introduced.

Since the result set returned by the query should usually be only one credential (consequently there's no big advantage using the projection so that traffic is reduced), I think that commit could be reverted and the sensitive fields could be removed "in code". 

WDYT?

sophokles73:
The MongoDB based Credentials service implementation does not use the
existing index for finding credentials by tenant, auth-id and type.
Instead, all documents of the tenant are scanned one by one for matching
auth-id and type.

The credentials DAO has been changed to create another compund index on
the credentials array documents' auth-id and type fields which is then
used by Mongo DB to determine the (small) set of documents matching the
auth-id and type before comparing with the required tenant ID.

Fixes #2792

### DIFF
--- a/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedCredentialsDao.java
+++ b/services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedCredentialsDao.java
@@ -50,15 +50,19 @@ import io.vertx.ext.mongo.UpdateOptions;
  */
 public final class MongoDbBasedCredentialsDao extends MongoDbBasedDao implements CredentialsDao, HealthCheckProvider {
 
-    private static final Logger LOG = LoggerFactory.getLogger(MongoDbBasedCredentialsDao.class);
-
-    private static final String KEY_AUTH_ID = String.format("%s.%s", CredentialsDto.FIELD_CREDENTIALS,
+    public static final String KEY_AUTH_ID = String.format("%s.%s", CredentialsDto.FIELD_CREDENTIALS,
             RegistryManagementConstants.FIELD_AUTH_ID);
-    private static final String KEY_CREDENTIALS_TYPE = String.format("%s.%s", CredentialsDto.FIELD_CREDENTIALS,
+    public static final String KEY_CREDENTIALS_TYPE = String.format("%s.%s", CredentialsDto.FIELD_CREDENTIALS,
             RegistryManagementConstants.FIELD_TYPE);
-    private static final String CREDENTIALS_FILTERED_POSITIONAL_OPERATOR = String.format("%s.$",
+    public static final String CREDENTIALS_FILTERED_POSITIONAL_OPERATOR = String.format("%s.$",
             CredentialsDto.FIELD_CREDENTIALS);
+    public static final JsonObject PROJECTION_CREDS_BY_TYPE_AND_AUTH_ID = new JsonObject()
+            .put(RegistryManagementConstants.FIELD_PAYLOAD_DEVICE_ID, 1)
+            .put(CREDENTIALS_FILTERED_POSITIONAL_OPERATOR, 1)
+            .put("_id", 0);
+    public static final String IDX_CREDENTIALS_TYPE_AND_AUTH_ID = "credentials_type_and_auth_id";
 
+    private static final Logger LOG = LoggerFactory.getLogger(MongoDbBasedCredentialsDao.class);
 
     private final AtomicBoolean creatingIndices = new AtomicBoolean(false);
     private final AtomicBoolean indicesCreated = new AtomicBoolean(false);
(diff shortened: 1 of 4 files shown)


sophokles73:
@fkaltner Would you mind taking a look as well?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/model/MongoDbBasedCredentialsDao.java
```diff
@@ -50,15 +50,19 @@
  */
 public final class MongoDbBasedCredentialsDao extends MongoDbBasedDao implements CredentialsDao, HealthCheckProvider {
 
-    private static final Logger LOG = LoggerFactory.getLogger(MongoDbBasedCredentialsDao.class);
-
-    private static final String KEY_AUTH_ID = String.format("%s.%s", CredentialsDto.FIELD_CREDENTIALS,
+    public static final String KEY_AUTH_ID = String.format("%s.%s", CredentialsDto.FIELD_CREDENTIALS,
```
fkaltner, line 53: Is there a reason why you made all these field public? Usage in my IDE shows only this class...

sophokles73: Some of them are used by the PROJECTION constant which is declared public. So the used constants need to be declared before that one. And since the PROJECTION is public and we want to stick with the declaration order (public, protected, private), they also need to be public ...

services/device-registry-mongodb/src/test/java/org/eclipse/hono/deviceregistry/mongodb/service/MongoDbBasedCredentialServiceTest.java
```diff
@@ -208,4 +216,85 @@ public void testUpdateCredentialsFailsForExceededCredentialsPerDeviceLimit(final
...
+                if (LOG.isTraceEnabled()) {
+                    LOG.trace("result:{}{}", System.lineSeparator(), result.encodePrettily());
+                }
+                ctx.verify(() -> {
+                    final var obj = JsonPointer.from("/queryPlanner/winningPlan/inputStage/inputStage")
+                            .queryJson(result);
+                    assertThat(obj).isInstanceOf(JsonObject.class);
+                    final var indexScan = (JsonObject) obj;
+                    assertThat(indexScan.getString("indexName"))
+                        .isEqualTo(MongoDbBasedCredentialsDao.IDX_CREDENTIALS_TYPE_AND_AUTH_ID);
```
fkaltner, line 294: Good idea to check that the index is actually used! 😄 

Nitpick:

It may also make sense to check the execution stats for "totalKeysExamined" / "totalDocsExamined" being 1. 

At least when using the hint (see https://github.com/eclipse/hono/issues/2792#issuecomment-887266164) one could also observe that using the index isn't sufficient; it is the number of scanned document that counts in the end.

sophokles73: I have pushed corresponding changes

PR #2798, opened 2021-07-27
Keep original registration when rejecting duplicate

### LINKED ISSUE #2795: MongoDB Device Registry: Creating a device with the same ID will delete the device
## Description
When a device with a specific device-id and type already exists in the database then on a subsequent create request this device will be deleted
 
## Actual behavior
A subsequent request to create the same device results in the deletion of the device

## Expected behavor
A subsequent request to create the same device should not alter or delete an already existing device

## Steps to reproduce
1. Create a device with device-id `abc` and type `hashed-password`.
   `POST v1/devices/tenant-1/abc` -> 200 OK
3. Make a call to GET the device, it will show that the device was created
   `GET v1/devices/tenant-1/abc` -> 200 OK
2. When an identical request through the device management API is made with the same device-id and type it will return a status `409 conflict`.
   `POST v1/devices/tenant-1/abc` -> 409 Conflict
3. Make a call to GET the device, it will show that the device is deleted
   `GET v1/devices/tenant-1/abc` -> 404

Screenshot 2021-07-26 at 16 21 08

sophokles73:
The MongoDb based DeviceManagementService erroneously removed the
original device registration when trying to register a new device using
the existing device's ID.

Fixes #2795

### DIFF
--- a/services/device-registry-base/src/test/java/org/eclipse/hono/service/registration/AbstractRegistrationServiceTest.java
+++ b/services/device-registry-base/src/test/java/org/eclipse/hono/service/registration/AbstractRegistrationServiceTest.java
@@ -340,11 +340,23 @@ public interface AbstractRegistrationServiceTest {
         final String deviceId = randomDeviceId();
 
         createDevice(deviceId, new Device())
-            .compose(ok -> getDeviceManagementService()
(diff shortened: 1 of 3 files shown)


calohmn (approved):
LGTM

PR #2799, opened 2021-07-27
Return proper error detail in devices resource responses

### LINKED ISSUE #2471: Device Registry: Error responses do not comply with the specification

sophokles73:
This is step 2 in a series of PR to fix #2471 

"devices" resources.

### DIFF
--- a/services/device-registry-base/src/test/java/org/eclipse/hono/service/registration/AbstractRegistrationServiceTest.java
+++ b/services/device-registry-base/src/test/java/org/eclipse/hono/service/registration/AbstractRegistrationServiceTest.java
@@ -343,6 +343,7 @@ public interface AbstractRegistrationServiceTest {
         final String deviceId = randomDeviceId();
 
         createDevice(deviceId, new Device())
+            .onFailure(ctx::failNow)
             .compose(ok -> getDeviceManagementService().createDevice(
                     TENANT,
                     Optional.of(deviceId),
--- a/services/device-registry-base/src/test/java/org/eclipse/hono/service/registration/AbstractRegistrationServiceTest.java
+++ b/services/device-registry-base/src/test/java/org/eclipse/hono/service/registration/AbstractRegistrationServiceTest.java
@@ -19,10 +19,13 @@ import java.time.Instant;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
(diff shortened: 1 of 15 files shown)


sophokles73:
@calohmn would you mind taking a look? @kaniyan will probably not find the time ...

services/device-registry-base/src/test/java/org/eclipse/hono/service/registration/AbstractRegistrationServiceTest.java
```diff
@@ -340,26 +343,23 @@ default void testCreateDeviceFailsForExistingDeviceId(final VertxTestContext ctx
         final String deviceId = randomDeviceId();
 
         createDevice(deviceId, new Device())
-            .onComplete(ctx.succeeding(ok -> {}))
```
calohmn, line 343: It would be in line with the other tests to keep this here, making sure the first device creation succeeded.

services/device-registry-file/src/test/java/org/eclipse/hono/deviceregistry/file/FileBasedRegistrationServiceTest.java
```diff
@@ -424,69 +421,98 @@ public void testDoStartIgnoreIdentitiesIfStartEmptyIsSet(final VertxTestContext
...
+                    Optional.of("newDevice"),
+                    new Device(),
+                    NoopSpan.INSTANCE))
+            .onComplete(ctx.failing(t -> {
+                // THEN the result contains a FORBIDDEN status code
+                ctx.verify(() -> Assertions.assertServiceInvocationException(t, HttpURLConnection.HTTP_FORBIDDEN));
+                registrationService.readDevice(TENANT, "newDevice", NoopSpan.INSTANCE)
+                    .onComplete(ctx.failing(error -> {
+                        // and the device has not been added to the registry
+                        ctx.verify(() -> org.eclipse.hono.deviceregistry.util.Assertions.assertServiceInvocationException(
```
calohmn, line 446: `org.eclipse.hono.deviceregistry.util.` superfluous here.

sophokles73:

calohmn (approved):
LGTM

PR #2802, opened 2021-07-29
Return proper error detail in credentials resource responses

### LINKED ISSUE #2471: Device Registry: Error responses do not comply with the specification

sophokles73:
"credentials" resources.

Fixes #2471

### DIFF
--- a/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsManagementService.java
+++ b/services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsManagementService.java
@@ -108,7 +108,9 @@ public abstract class AbstractCredentialsManagementService implements Credential
      *             as the parent for additional spans created as part of this method's execution.
      * @return A future indicating the outcome of the operation.
      *         <p>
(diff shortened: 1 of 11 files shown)


services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsManagementService.java
```diff
@@ -167,7 +170,7 @@ public void setTenantInformationService(final TenantInformationService tenantInf
                                     resourceVersion,
                                     span));
                 })
-                .otherwise(t -> DeviceRegistryUtils.mapErrorToResult(t, span));
```
calohmn, line 170: `DeviceRegistryUtils.mapErrorToResult` can now be removed.

sophokles73: I have pushed corresponding changes

calohmn (approved):
LGTM

Issue #2805, opened 2021-08-03
Upgrade to vert.x 4.1.x


sophokles73:
vert.x 4.x has been around for some time now and I think we should start migration.
This will probably entail upgrading to Quarkus 2 as well.

PR #2806, opened 2021-08-03
Upgrade to Vert.x 4.1.2

sophokles73:
This addresses #2805 

Also upgraded to Quarkus 2.0.3 and Infinispan 12.1.4
This PR is only about updating the frameworks and making the changes that are required to make our code run.
I am not 100% if we still achieve our goals that we had in mind with our own AuthHandler classes. Maybe you can have a particular look on that area @calohmn ?

I do have another PR for replacing the deprecated `ConfigProperties` class usage.


adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/HonoHttpAuthenticationHandler.java
```diff
@@ -27,21 +27,21 @@
...
+import io.vertx.ext.auth.authentication.AuthenticationProvider;
 import io.vertx.ext.web.RoutingContext;
 
 /**
- * Interface for Hono {@link io.vertx.ext.web.handler.AuthHandler} implementations.
+ * Interface for Hono {@link io.vertx.ext.web.handler.AuthenticationHandler} implementations.
  *
  */
-public interface HonoHttpAuthHandler {
+public interface HonoHttpAuthenticationHandler {
```
calohmn, line 37: This interface seems obsolete now - it isn't used and therefore could be removed.

tests/src/test/java/org/eclipse/hono/tests/http/EventHttpIT.java
```diff
@@ -98,7 +98,7 @@ public void testEventIsRejectedForUnsupportedQosLevel(final VertxTestContext ctx
                 Buffer.buffer("hello"),
                 requestHeaders,
                 ResponsePredicate.status(HttpURLConnection.HTTP_BAD_REQUEST))
-                .onComplete(ctx.completing());
+                .onComplete(ctx.succeedingThenComplete());
     }
```
calohmn, line 103: Regarding `EventHttpIT` in general:
I've run the integration tests like this: `mvn verify -Prun-tests,jaeger -Ddocker.keepRunning -Dit.test=EventHttpIT -Dmsg.count=1`.

When looking at the HTTP adapter traces with tags `cache_hit=true credentials_type=x509-cert` (corresponding Jaeger UI URL: http://localhost:18080/search?limit=200&lookback=1h&service=hono-adapter-http&tags=%7B%22cache_hit%22%3A%22true%22%2C%22credentials_type%22%3A%22x509-cert%22%7D) there were 117 traces using current master but only 59 traces when using the code from this PR. The overall number of traces with `credentials_type=x509-cert` was the same (121).

This makes me wonder whether changes in this PR somehow lead to less cache hits.

calohmn:
Minor:
Looking at the log output of the integration tests, there is quite a lot of noise on INFO level, like:
```
HTTP16:46:31.008 [vert.x-eventloop-thread-1] INFO  io.vertx.ext.web.RoutingContext - RoutingContext failure (500)
HTTPorg.eclipse.hono.client.ClientErrorException: validation of device certificate failed
HTTP	at org.eclipse.hono.adapter.auth.device.TenantServiceBasedX509Authentication.lambda$validateClientCertificate$1(TenantServiceBasedX509Authentication.java:185)
HTTP	at io.vertx.core.impl.future.Composition.onFailure(Composition.java:50)
HTTP	at io.vertx.core.impl.future.FutureBase.emitFailure(FutureBase.java:79)
[...]
```
This could be prevented by setting the `io.vertx.ext.web.RoutingContext` log level to WARN for the HTTP adapter and Device Registry. The next vert.x web 4.1.x version will have this set to DEBUG (see [here](https://github.com/vert-x3/vertx-web/commit/4cc2a9ff2e7aba07d5cd25afeece9e7e9870d95e)).

tests/src/test/java/org/eclipse/hono/tests/http/EventHttpIT.java
sophokles73: Hm, it is probably not a coincidence that this is close to half of what we had before, right?

calohmn: Not sure here yet. The issue can be further tracked down by just concentrating on the `EventHttpIT#testConnectSucceedsWithAutoProvisioning` test.

sophokles73: Not sure if I understand what you mean ...

calohmn: It seems that when trying to get the same credentials mulitple times, in `CredentialsApiAuthProvider#getCredentialsForDevice` the given `SubjectDnCredentials` have the same `clientContext` JsonObject content (e.g. `{"client-certificate":"MIIC4TCC..."}`), but the HashCode of these JsonObjects seems to be different now with every `getCredentialsForDevice` invocation.
With the older vert.x version it was the same.

calohmn: It looks like this is caused by changes to the `JsonObject.copy()` behaviour.

For a tenant with auto-provisioning enabled, `TenantServiceBasedX509Authentication#getCredentials` will return a JsonObject with an `client-certificate` entry containing a byte array.
This JsonObject is then copied in `X509AuthProvider#getCredentials`, and the copy is put into the `SubjectDnCredentials` object. The hashcode of that copy is then relevant for cache hit or miss.

Consider this example:
```
            final JsonObject authInfo = new JsonObject().put("client-certificate", new byte[] { 0x00, 0x01 })
                    .copy();
            System.out.println("hashcode: " + authInfo.hashCode() + "; map: " + authInfo.getMap());

            final JsonObject authInfo2 = new JsonObject().put("client-certificate", new byte[] { 0x00, 0x01 })
                    .copy();
            System.out.println("hashcode: " + authInfo2.hashCode() + "; map: " + authInfo2.getMap());
```

Output with vert.x 3.9.7:
```
hashcode: -207395283; map: {client-certificate=AAE=}
hashcode: -207395283; map: {client-certificate=AAE=}
```
With vert.x 4.1.2:
```
hashcode: -1623841546; map: {client-certificate=[B@6c8bca63}
hashcode: -864800432; map: {client-certificate=[B@3fc9dfc5}
```

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/HonoHttpAuthenticationHandler.java
calohmn: And then `org.eclipse.hono.adapter.http.HttpAuthProviderAdapter` can be removed as well.

adapters/http-vertx-base/src/main/java/org/eclipse/hono/adapter/http/X509AuthHandler.java
```diff
@@ -117,13 +117,9 @@ public void authenticate(final RoutingContext context, final Handler<AsyncResult
...
-                        authHandler.authenticateDevice(HttpContext.from(context)).onComplete(ar -> {
-                            if (ar.succeeded()) {
-                                handler.handle(Future.succeededFuture(ar.result()));
-                            } else {
-                                handler.handle(Future.failedFuture(ar.cause()));
-                            }
-                        });
+                        authHandler.authenticateDevice(HttpContext.from(context))
+                            .map(deviceUser -> (User) deviceUser)
+                            .onComplete(handler);
```
calohmn, line 122: This kind of change can also be done in `HonoBasicAuthHandler#authenticate`.

tests/src/test/java/org/eclipse/hono/tests/http/EventHttpIT.java
sophokles73: Yes, it seems that when copying a binary value, 3.9.x performed a Base64 encoding of the byte array whereas 4.1.x simply copies the byte array itself. I have added a *nomalization* step to the Credentials client where the caching key is created,

calohmn (approved):
LGTM

sophokles73:
IMHO we can/should apply https://github.com/eclipse/hono/pull/2806/commits/4a5ee8fee7307cfd9c4fab2612399f681c0372b1 to master already as the current code on master already depends on the (undocumented) specifics of the *JsonObject.copy()* implementation. WDYT @calohmn ?

calohmn:
> IMHO we can/should apply 4a5ee8f to master already as the current code on master already depends on the (undocumented) specifics of the JsonObject.copy() implementation. WDYT @calohmn ?

@sophokles73 I agree.

sophokles73:
I have scheduled this for 1.11.0 ...

sophokles73:
replaced by #2847

PR #2808, opened 2021-08-05
Remove Quarkus specific Infinispan client project

sophokles73:
The standard Infinispan client project has been extended with SmallRye
@ConfigMapping annotated interfaces to which Quarkus will bind
configuration properties.

The Quarkus specific project is now obsolete.


sophokles73:

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/CommonCacheOptions.java
```diff
@@ -0,0 +1,52 @@
...
+
+import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.ConfigMapping.NamingStrategy;
+import io.smallrye.config.WithDefault;
+
+/**
+ * Common options for configuring a cache.
+ *
+ */
+@ConfigMapping(prefix = "hono.commandRouter.cache.common", namingStrategy = NamingStrategy.VERBATIM)
```
calohmn, line 26: Having the `commandRouter` reference here doesn't fit the nature of the module this class is in very well. (Although I'm aware it works because the only place where the module is used in a quarkus component is the command-router.)

But, would it also work if this annotation gets moved to a sub-interface, defined in the command router module?

sophokles73: I guess so, let me check it out ...

sophokles73: But we can also simply use a more generic (and unused) prefix here like `hono.infinispan` and then only override the prefix in the command router module. We then do not need to define an empty interface just for hosting the annotation ...

calohmn: Sounds good.

sophokles73: I have pushed corresponding changes

sophokles73: Does this work for you?

calohmn: Yes.

calohmn (approved):
LGTM

PR #2810, opened 2021-08-05
Replace Quarkus specific property binding mechanism

sophokles73:
Changed ConnectionEventProducer configuration to use SmallRye
@ConfigMapping instead of Quarkus specific @ConfigProperties.

This will allow for easier migration to Quarkus 2 where
@ConfigProperties has been deprecated.


calohmn (approved):
LGTM

PR #2813, opened 2021-08-06
Use SpanContext param in ConnectionEventProducer methods

calohmn:
This adds a `SpanContext` parameter to the methods of the `ConnectionEventProducer` interface.

This prevents `get Tenant by ID` spans in the AMQP/MQTT adapter with no parent span, where it is unclear just by looking at the tracing span to which overall operation/request this belongs to.


sophokles73 (approved):
LGTM

PR #2816, opened 2021-08-06
Remove Docker Swarm based Sandbox Deployment

### LINKED ISSUE #2814: Remove old Sandbox deployment based on Docker Swarm
As a follow-up of https://github.com/eclipse/hono/issues/2727, the scripts and configuration files for the old Docker Swarm deployment can be removed.

b-abel:
This fixes #2814. The Helm-based deployment no longer requires resource filtering, therefore the `pom.xml` has been removed.


deploy/src/main/deploy/example-credentials.json
```diff
@@ -49,7 +49,7 @@
       {
         "device-id": "4711",
         "type": "x509-cert",
-        "auth-id": "${cert.device-4711.subject}",
+        "auth-id": "CN=Device 4711,OU=Hono,O=Eclipse IoT,L=Ottawa,C=CA",
```
sophokles73, line 52: IMHO we can the remove the example-credentials, -tenants and -device-identities altogether. They are not being deployed by the helm chart and we do not rely on nor use them in the getting started guide, do we?

sophokles73 (commented):
any reason you want to change the file permissions for the `services.sh` script and make it no longer executable?

deploy/src/main/deploy/example-credentials.json
b-abel: The example files are only referenced in the _admin_ guide of the file-based device registry.

sophokles73: Meaning?

b-abel: I would wait with removing the example configurations until the file-based registry is (finally!) removed. Otherwise, we need to fix the links in the admin guide.

sophokles73: Ok, so we'll keep the example files for reference for the time being ...

sophokles73 (approved):
LGTM

PR #2822, opened 2021-08-12
[#2028] Add AdapterInstanceStatusService in Command Router

### LINKED ISSUE #2028: Introduce AdapterInstancesLivenessService, use for getCmdHandlingAdapterInstances

calohmn:
This fixes #2028.


legal/src/main/resources/legal/DEPENDENCIES
```diff
@@ -144,10 +172,13 @@ maven/mavencentral/org.jboss.logmanager/jboss-logmanager-embedded/1.0.9, Apache-
 maven/mavencentral/org.jboss.logmanager/log4j2-jboss-logmanager/1.0.0.Beta1, Apache-2.0, approved, clearlydefined
 maven/mavencentral/org.jboss.slf4j/slf4j-jboss-logmanager/1.1.0.Final, Apache-2.0, approved, clearlydefined
 maven/mavencentral/org.jboss.spec.javax.transaction/jboss-transaction-api_1.2_spec/1.1.1.Final, GPL-2.0-only WITH Classpath-exception-2.0 OR CDDL-1.0, approved, clearlydefined
+maven/mavencentral/org.jboss.spec.javax.xml.bind/jboss-jaxb-api_2.3_spec/2.0.0.Final, NOASSERTION, restricted, clearlydefined
```
calohmn, line 175: A restricted dependency entry got added here. I'm looking into this.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/KubernetesBasedAdapterInstanceStatusService.java
```diff
@@ -0,0 +1,344 @@
...
+    private void onWatcherClosed(final WatcherException cause) {
+        if (active.compareAndExchange(1, 0) != 1) {
+            return; // watcher already closed or status service already stopped
+        }
+        containerIdToPodNameMap.clear();
+        podNameToContainerIdMap.clear();
+        LOG.error("Watcher closed with error", cause);
+        while (active.get() == 0) {
+            try {
+                Thread.sleep(WATCH_RECREATION_DELAY_MILLIS);
```
sophokles73, line 236: this looks very, very strange ...

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/AdapterInstanceStatusProvider.java
```diff
@@ -0,0 +1,44 @@
...
+    AdapterInstanceStatusProvider UNKNOWN_STATUS_PROVIDER = new UnknownStatusProvider();
+
+    /**
+     * Gets the status of the adapter identified by the given identifier.
+     *
+     * @param adapterInstanceId The identifier of the adapter.
+     * @return The status of the adapter instance.
+     * @throws NullPointerException if adapterInstanceId is {@code null}.
+     */
+    AdapterInstanceStatus getStatus(String adapterInstanceId);
```
sophokles73, line 32: How about returning a `Future<AdapterInstanceStatus>` here instead in order to also support asynchronous implementations?

core/src/main/java/org/eclipse/hono/util/AdapterInstanceStatus.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public enum AdapterInstanceStatus {
+    /**
+     * Adapter instance is alive.
+     */
+    ALIVE,
+    /**
+     * Adapter instance has probably been terminated.
+     */
+    SUSPECTED,
```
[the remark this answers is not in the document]
sophokles73: The *KubernetesBasedAdapterInstanceStatusService#getStatus()* method contains the following comment for the case when it returns SUSPECTED:
> (if the) container is new and the watcher hasn't informed us about it yet (watcher could be currently reconnecting internally)

Base don that, something like PENDING would better fit the purpose here ...

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/KubernetesBasedAdapterInstanceStatusService.java
sophokles73: I see. Does the kubernetes client say anything about the impact of blocking the thread that this method is invoked on? You might end up blocking the thread for quite some time ...

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/KubernetesBasedAdapterInstanceStatusService.java
```diff
@@ -0,0 +1,344 @@
...
+    private static final String ADAPTER_NAME_MATCH = "adapter";
+    private static final long WATCH_RECREATION_DELAY_MILLIS = 100;
+    private static final int MAX_LRU_MAP_ENTRIES = 200;
+
+    private final KubernetesClient client;
+    private final String namespace;
+    /**
+     * Values: 0: inactive, 1: active, -1: status service stopped.
+     */
+    private final AtomicInteger active = new AtomicInteger();
```
sophokles73, line 62: why do we need do differentiate three states? If the service needs to be started before it can be used, then FMPOV we would simply throw an IllegalStateException if somebody tries to invoke its methods before it has been started.

legal/src/main/resources/legal/DEPENDENCIES
```diff
@@ -144,10 +172,13 @@ maven/mavencentral/org.jboss.logmanager/jboss-logmanager-embedded/1.0.9, Apache-
...
 maven/mavencentral/org.jboss.threads/jboss-threads/3.2.0.Final, Apache-2.0, approved, clearlydefined
 maven/mavencentral/org.jetbrains/annotations/13.0, Apache-2.0, approved, clearlydefined
 maven/mavencentral/org.jetbrains.kotlin/kotlin-stdlib/1.4.32, Apache-2.0, approved, clearlydefined
 maven/mavencentral/org.jetbrains.kotlin/kotlin-stdlib-common/1.4.32, Apache-2.0, approved, clearlydefined
+maven/mavencentral/org.jetbrains.kotlin/kotlin-stdlib-jdk7/1.4.32, Apache-2.0, approved, clearlydefined
+maven/mavencentral/org.jetbrains.kotlin/kotlin-stdlib-jdk8/1.4.32, Apache-2.0, approved, clearlydefined
```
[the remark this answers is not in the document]
calohmn: I've excluded them.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/CommandRouterServiceConfigProperties.java
```diff
@@ -18,4 +18,27 @@
...
 public class CommandRouterServiceConfigProperties extends ServiceConfigProperties {
+
+    private boolean adapterInstanceStatusServiceEnabled = true;
+
+    /**
+     * Checks whether the service to get the status of an adapter instance is enabled.
+     *
+     * @return {@code true} if the status service is enabled.
+     */
+    public final boolean isAdapterInstanceStatusServiceEnabled() {
```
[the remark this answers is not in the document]
calohmn: Right, I've renamed this method accordingly.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/KubernetesBasedAdapterInstanceStatusService.java
calohmn: A switch between inactive and active is also done when the Watcher has been closed and is then re-initiated.
Apart from that there is the state of the whole service having been closed.
So, the `getStatus` method has to check both whether the service is currently inactive (Watcher being re-initialized) or stopped. In order not to check two Booleans here, I've chosen the AtomicInteger, to be updated atomically.
See also the other places where the `active` flag is used (e.g. Watcher re-init should be aborted if service already closed).

core/src/main/java/org/eclipse/hono/util/AdapterInstanceStatus.java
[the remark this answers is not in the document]
calohmn: PENDING would have this connotation of "container just being started", but that is just the unlikely scenario.
I've removed the above comment and put a new one in the SUSPECTED block:
```
                // container hasn't explicitly been marked as terminated
                // so it could either have been terminated before this service got started
                // or, in an unlikely case, could be newly started and alive, and the watcher hasn't informed us about it yet
                // (watcher could be currently reconnecting internally)
```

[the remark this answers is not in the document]
sophokles73: Ok, can we then use something like INDETERMINATE? SUSPECTED sounds like the pod had committed a crime ... ;-)

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/AdapterInstanceStatusProvider.java
calohmn: I initially had used a `Future` return value here, but then didn't use it because the implementation in this PR doesn't need it and the code is simpler without it. An implementation that isn't doing time-consuming things here that require asynchronous decoupling would be preferred anyway, since this method is invoked for every command message.
And if we really need a Future at some point here, I think we can still change it then.

core/src/main/java/org/eclipse/hono/util/AdapterInstanceStatus.java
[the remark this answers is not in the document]
calohmn: I've used SUSPECTED thinking about the same term being used for JGroups/Infinispan cluster nodes that are believed to be dead. INDETERMINATE would be quite similar to UNKNOWN.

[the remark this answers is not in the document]
sophokles73: Then maybe SUSPECTED_DEAD?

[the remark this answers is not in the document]
calohmn: I'm fine with that. I've renamed it accordlingly, also renaming TERMINATED to DEAD as a consequence.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/KubernetesBasedAdapterInstanceStatusService.java
calohmn: Looking at KubernetesClient / OkHttp DEBUG log output when running the `testServiceReactivatesAfterWatchClosed` test method, I don't see an issue here.
Testing with a sleep period of 60s, I still see maintenance work on the connection getting done in this period via other threads:
```
09:19:36.444 [OkHttp TaskRunner] DEBUG o.internal.concurrent.TaskRunner - Q10003 starting              : OkHttp WebSocket http://localhost:42369/... ping
09:19:36.446 [OkHttp WebSocket http://localhost:42369/... ping] DEBUG o.internal.concurrent.TaskRunner - Q10003 run again after  30 s : OkHttp WebSocket http://localhost:42369/... ping
09:19:36.447 [OkHttp TaskRunner] DEBUG o.internal.concurrent.TaskRunner - Q10003 finished run in   2 ms: OkHttp WebSocket http://localhost:42369/... ping
```
And in terms of the KubernetesClient, the watcher is closed at that point, so nothing to be done there.

bom/pom.xml
```diff
@@ -740,9 +742,21 @@
       <dependency>
         <groupId>com.squareup.okhttp3</groupId>
         <artifactId>okhttp</artifactId>
-        <version>4.9.0</version>
+        <version>${okhttp.version}</version>
         <scope>runtime</scope>
       </dependency>
+      <dependency>
+        <groupId>com.squareup.okhttp3</groupId>
+        <artifactId>logging-interceptor</artifactId>
```
sophokles73, line 750: is this really a *required* dependency? Or can we also omit this?

legal/src/main/resources/legal/DEPENDENCIES
```diff
@@ -66,6 +91,8 @@ maven/mavencentral/io.quarkus/quarkus-development-mode-spi/1.13.7.Final, Apache-
 maven/mavencentral/io.quarkus/quarkus-jackson/1.13.7.Final, Apache-2.0, approved, clearlydefined
 maven/mavencentral/io.quarkus/quarkus-jaeger/1.13.7.Final, Apache-2.0, approved, clearlydefined
 maven/mavencentral/io.quarkus/quarkus-kafka-client/1.13.7.Final, Apache-2.0, approved, clearlydefined
+maven/mavencentral/io.quarkus/quarkus-kubernetes-client/1.13.7.Final, Apache-2.0, approved, clearlydefined
+maven/mavencentral/io.quarkus/quarkus-kubernetes-client-internal/1.13.7.Final, Apache-2.0, approved, clearlydefined
```
sophokles73, line 95: this looks odd, doesn't it?

calohmn: It looks odd yes. But excluding the dependency will lead to a build error:
`Failed to load steps from class io.quarkus.kubernetes.client.deployment.KubernetesClientBuildStep`

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/KubernetesBasedAdapterInstanceStatusService.java
sophokles73: ok

legal/src/main/resources/legal/DEPENDENCIES
calohmn: I'll create a ClearlyDefined PR for this and update the DEPENDENCIES file later.

bom/pom.xml
calohmn: The OkHttp `logging-interceptor` artifact in version 3.14.9 is used by the `kubernetes-client` dependency.
Excluding the entry here would mean that we use `logging-interceptor:3.14.9` along with `okhttp:4.9.0`.
I would rather use one version for the OkHttp components.

I've therefore amended the commit to make this more explicit, excluding the OkHttp dependencies from `kubernetes-client` and `kubernetes-server-mock` dependency declarations and explicitly adding the OkHttp dependencies in the POM where the Kubernetes client/server-mock dependencies get used.

sophokles73: > Excluding the entry here would mean that we use logging-interceptor:3.14.9 along with okhttp:4.9.0.
I would rather use one version for the OkHttp components.

???

My question was if the kubernetes-client would also work if this artifact were not on the class path.

calohmn: No, it wouldn't work.
Exception in the Command Router component:
```
java.lang.ClassNotFoundException: okhttp3.logging.HttpLoggingInterceptor
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(Unknown Source)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(Unknown Source)
	at java.base/java.lang.ClassLoader.loadClass(Unknown Source)
	at io.quarkus.bootstrap.runner.RunnerClassLoader.loadClass(RunnerClassLoader.java:104)
	at io.quarkus.bootstrap.runner.RunnerClassLoader.loadClass(RunnerClassLoader.java:60)
	at io.fabric8.kubernetes.client.utils.HttpClientUtils.createHttpClient(HttpClientUtils.java:145)
	at io.fabric8.kubernetes.client.utils.HttpClientUtils.createHttpClient(HttpClientUtils.java:66)
	at io.fabric8.kubernetes.client.BaseClient.<init>(BaseClient.java:51)
	at io.fabric8.kubernetes.client.BaseClient.<init>(BaseClient.java:43)
	at io.fabric8.kubernetes.client.DefaultKubernetesClient.<init>(DefaultKubernetesClient.java:138)
	at org.eclipse.hono.commandrouter.impl.KubernetesBasedAdapterInstanceStatusService.<init>(KubernetesBasedAdapterInstanceStatusService.java:72)
```
`kubernetes-server-mock` used in the unit test also needs it:
```
java.lang.NoClassDefFoundError: okhttp3/logging/HttpLoggingInterceptor
	at io.fabric8.kubernetes.client.utils.HttpClientUtils.createHttpClient(HttpClientUtils.java:145)
	at io.fabric8.kubernetes.client.utils.HttpClientUtils.createHttpClientForMockServer(HttpClientUtils.java:70)
	at io.fabric8.kubernetes.client.server.mock.KubernetesMockServer.createClient(KubernetesMockServer.java:91)
	at io.fabric8.kubernetes.client.server.mock.KubernetesMockServerExtension.initializeKubernetesClientAndMockServer(KubernetesMockServerExtension.java:92)
	at io.fabric8.kubernetes.client.server.mock.KubernetesMockServerExtension.setKubernetesClientAndMockServerFields(KubernetesMockServerExtension.java:123)
	at io.fabric8.kubernetes.client.server.mock.KubernetesMockServerExtension.beforeAll(KubernetesMockServerExtension.java:72)
...
```

sophokles73: that's a pity

sophokles73 (approved):
LGTM

legal/src/main/resources/legal/DEPENDENCIES
calohmn: With the dependency changes in #2847, this entry isn't there anymore, so it looks like there is no further action needed here with regards to the upcoming Hono 1.10.0.

PR #2824, opened 2021-08-12
Fix Kafka configuration on Sandbox

b-abel:
The Kafka server had returned a wrong IP address of the broker to clients. To fix this, the domain name of the sandbox is now explicitly set as the "loadBalancerIP" at the external service for Kafka. The "auto-discovery" of the load balancer IP at the Bitnami Helm chart is now disabled.


sophokles73 (approved):
LGTM

Issue #2834, opened 2021-08-26
ApplicationClient consumer close handler not invoked on disconnect


calohmn:
The
`org.eclipse.hono.application.client.ApplicationClient#createTelemetryConsumer()` and
`org.eclipse.hono.application.client.ApplicationClient#createEventConsumer()`
methods have a `closeHandler` parameter that is invoked when the consumer gets closed by the remote peer.
A common usage pattern is to try to re-create the consumer in the `closeHandler`.

In the case of `AmqpApplicationClient` implementations, the consumer may also get implicitly closed when the underlying AMQP connection gets closed by the remote peer (without having closed the receiver link) or when the socket gets closed.
In that case the `closeHandler` currently doesn't get invoked.

An application using an `AmqpApplicationClient` may already use the `addDisconnectListener`/`addReconnectListener` methods (provided via the `ConnectionLifecycle` interface) to notice a disconnect and re-create the consumers. But IMHO it would be more consistent for the consumer `closeHandler` to be also invoked on disconnects.
The application may also use the `AmqpApplicationClient` implementation as a generic `ApplicationClient` object. Then it wouldn't have direct access to the `addDisconnectListener`/`addReconnectListener` methods.

PR #2835, opened 2021-08-26
[#2834] Invoke consumer close handler on remote disconnect

calohmn:
This is for #2834.

### DIFF
--- a/clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedApplicationClient.java
+++ b/clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedApplicationClient.java
@@ -14,13 +14,17 @@
 
 package org.eclipse.hono.application.client.amqp;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.function.BiConsumer;
 import java.util.function.Function;
 
 import org.apache.qpid.proton.amqp.messaging.Rejected;
 import org.apache.qpid.proton.amqp.messaging.Released;
 import org.apache.qpid.proton.amqp.transport.DeliveryState;
+import org.apache.qpid.proton.message.Message;
 import org.eclipse.hono.application.client.DownstreamMessage;
 import org.eclipse.hono.application.client.MessageConsumer;
 import org.eclipse.hono.client.ClientErrorException;
@@ -38,6 +42,7 @@ import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.Promise;
+import io.vertx.core.impl.NoStackTraceThrowable;
 import io.vertx.proton.ProtonDelivery;
 import io.vertx.proton.ProtonHelper;
 
@@ -51,6 +56,8 @@ public class ProtonBasedApplicationClient extends ProtonBasedCommandSender imple
 
     private static final Logger LOG = LoggerFactory.getLogger(ProtonBasedApplicationClient.class);
 
+    private final List<Handler<Throwable>> consumerCloseHandlers = new ArrayList<>();
+
     /**
      * Creates a new vertx-proton based based application client.
      *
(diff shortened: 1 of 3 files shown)


clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedApplicationClient.java
```diff
@@ -62,6 +68,18 @@ public ProtonBasedApplicationClient(final HonoConnection connection) {
         super(connection, SendMessageSampler.Factory.noop());
     }
 
+    @Override
+    protected void onDisconnect() {
+        // consumer close handlers shall be called upon disconnect from remote peer,
+        // so skip invocation here if disconnect is done during shutdown
+        if (!connection.isShutdown()) {
+            final Throwable error = Future.failedFuture("disconnected").cause();
```
sophokles73, line 76: what is the purpose of creating a Future here?

clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedApplicationClient.java
```diff
@@ -247,55 +265,69 @@ public final void disconnect(final Handler<AsyncResult<Void>> completionHandler)
...
-            .map(recv -> new MessageConsumer() {
-                @Override
-                public Future<Void> close() {
-                    return recv.close();
-                }
-            });
+                closeHandler);
+    }
+
+    private Future<MessageConsumer> createConsumer(
```
sophokles73, line 299: can we move this to right after the other private *createConsumer* method?

clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedApplicationClient.java
```diff
@@ -247,55 +265,69 @@ public final void disconnect(final Handler<AsyncResult<Void>> completionHandler)
...
+            final String sourceAddress,
+            final BiConsumer<ProtonDelivery, Message> messageConsumer,
+            final Handler<Throwable> closeHandler) {
+
+        // wrap the handler to ensure distinct objects are added to the consumerCloseHandlers list
+        // preventing issues if the same closeHandler instance is used for multiple createConsumer() invocations
+        final Handler<Throwable> wrappedCloseHandler = closeHandler != null ? (thr -> closeHandler.handle(thr)) : null;
```
sophokles73, line 306: what difference does this make?

calohmn: > preventing issues if the same closeHandler instance is used for multiple createConsumer() invocations

The scenario in which there is an issue is a bit theoretical:
- The same `closeHandler` instance is used with 2 `createConsumer` invocations.
- One consumer instance is closed explicitly via the `MessageConsumer#close` method, and for some reason the `close` method is invoked twice for that consumer.
- Then there is no closeHandler anymore in the `consumerCloseHandlers` list to be invoked for the _other_ consumer instance.

In more general terms, the aim here is to have distinct `consumerCloseHandlers` list entries for each `createConsumer` invocation, so that `consumerCloseHandlers.remove(handler)` doesn't remove the wrong handler.

clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedApplicationClient.java
calohmn: The aim is the create a `NoStackTraceThrowable` here. I've made that now more obvious.

clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedApplicationClient.java
calohmn: done

clients/application-amqp/src/main/java/org/eclipse/hono/application/client/amqp/ProtonBasedApplicationClient.java
```diff
@@ -62,6 +69,18 @@ public ProtonBasedApplicationClient(final HonoConnection connection) {
         super(connection, SendMessageSampler.Factory.noop());
     }
 
+    @Override
+    protected void onDisconnect() {
+        // consumer close handlers shall be called upon disconnect from remote peer,
+        // so skip invocation here if disconnect is done during shutdown
+        if (!connection.isShutdown()) {
+            final Throwable error = new NoStackTraceThrowable("disconnected");
```
sophokles73, line 77: :+1:

sophokles73 (approved):
LGTM

PR #2845, opened 2021-09-07
[#2836] Refactor KafkaProducerFactory

### LINKED ISSUE #2836: KafkaProducerFactory#sharedProducerFactory potentially misleading in how producers are shared
Consider a scenario where you want to use different `KafkaApplicationClient` instances, each working with Kafka configuration properties containing different login credentials, giving access to a specific set of topics.

For a these clients you need a `KafkaProducerFactory` instance. The intuitive way to create one is to use `KafkaProducerFactory#sharedProducerFactory`.

Even when using different `KafkaProducerFactory` instances created with that method for these client, the created Kafka producers with the same name still are getting shared between all client instances. This results in authentication/authorisation errors when working with the producers.

This is especially a problem concerning the command producer created in the `KafkaApplicationClient` because it gets created with a fixed name ("command-sender").

Caching of Kafka producer instances for such a scenario should only be done per `KafkaProducerFactory` instance.

EDIT: made it clear that this is about the `KafkaProducerFactory#sharedProducerFactory` factory method.

b-abel:
This fixes #2836.

### DIFF
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/CachingKafkaProducerFactory.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/CachingKafkaProducerFactory.java
@@ -54,13 +54,12 @@ public class CachingKafkaProducerFactory<K, V> implements KafkaProducerFactory<K
     /**
      * Creates a new producer factory.
      * <p>
-     * Use {@link #sharedFactory(Vertx)} to create producers that can safely be shared
-     * between verticle instances.
+     * Use {@link #sharedFactory(Vertx)} to create producers that can safely be shared between verticle instances.
      *
      * @param producerInstanceSupplier The function that provides new producer instances. Parameters are the producer
-     *                                 name and the producer configuration.
+     *            name and the producer configuration.
      */
-    public CachingKafkaProducerFactory(
+    private CachingKafkaProducerFactory(
             final BiFunction<String, Map<String, String>, KafkaProducer<K, V>> producerInstanceSupplier) {
         this.producerInstanceSupplier = producerInstanceSupplier;
     }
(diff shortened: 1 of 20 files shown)


b-abel:
@calohmn Could you have a look. This is the change requested by you. And it requested a lot of reviewers :laughing:

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/CachingKafkaProducerFactory.java
```diff
@@ -54,17 +54,76 @@
...
+     *            name and the producer configuration.
      */
-    public CachingKafkaProducerFactory(
+    private CachingKafkaProducerFactory(
             final BiFunction<String, Map<String, String>, KafkaProducer<K, V>> producerInstanceSupplier) {
         this.producerInstanceSupplier = producerInstanceSupplier;
     }
 
+    /**
+     * Creates a new factory that produces {@link KafkaProducer#createShared(Vertx, String, Map) sharedFactory
```
calohmn, line 68: "that produces *shared* producers"?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/CachingKafkaProducerFactory.java
```diff
@@ -54,17 +54,76 @@
...
+     * producers}. Shared producers can safely be sharedFactory between verticle instances and improve efficiency by
+     * leveraging the batching capabilities of the Kafka client.
+     *
+     * Producers with the same name will be shared (per Vert.x instance), meaning they are shared between instances of
+     * the factory.
+     * <p>
+     * Config must always be the same for the same key in
+     * {@link #getOrCreateProducer(String, KafkaProducerConfigProperties)}.
+     * <p>
```
calohmn, line 77: How about:
```
Instances returned here must use the same config in 
{@link #getOrCreateProducer(String, KafkaProducerConfigProperties)}
invocations that use the same key.
```
?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/CachingKafkaProducerFactory.java
```diff
@@ -54,17 +54,76 @@
...
      */
-    public CachingKafkaProducerFactory(
+    private CachingKafkaProducerFactory(
             final BiFunction<String, Map<String, String>, KafkaProducer<K, V>> producerInstanceSupplier) {
         this.producerInstanceSupplier = producerInstanceSupplier;
     }
 
+    /**
+     * Creates a new factory that produces {@link KafkaProducer#createShared(Vertx, String, Map) sharedFactory
+     * producers}. Shared producers can safely be sharedFactory between verticle instances and improve efficiency by
```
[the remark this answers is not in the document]
b-abel: Uh, the renaming support of IntelliJ can be a bit aggressive sometimes :smile:

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/CachingKafkaProducerFactory.java
b-abel: I think the whole sentence should be removed. This is about the `getOrCreateProducer` method where the JavaDoc already states "The given config is ignored when an existing producer is returned". It always applies to this method regardless of the producer instance supplier.

calohmn (approved):
LGTM

PR #2850, opened 2021-09-09
Documentation: Add the JDBC device registry to page "Device Registries".

b-abel:
The page "Device Registries" maintains the URL path that the file based registry used to have when it was the only supported implementation. It links to the current implementations but the JDBC based one was missing. This adds it.


sophokles73:
I do not seem to be able to find that page in the documentation menu. Can you help me out?

b-abel:
The page is hidden in the documentation. You only find it by opening the URL https://www.eclipse.org/hono/docs/user-guide/device-registry/
The reason that this page exists is that it uses the path that the file based registry had when it was the only implementation. 
Now the name does not fit anymore. But removing pages is a not good idea because it breaks (a) external links to it and (b) the version selection box (if you switch to version 1.2 you still see the old page).

sophokles73 (approved):
LGTM

PR #2853, opened 2021-09-16
Support deletion of all devices of a tenant

### LINKED ISSUE #2793: MongoDB device registry does not delete devices & credentials on tenant delete
The MongoDB device registry currently does not delete devices and credentials belonging to a tenant that gets deleted.
These devices and credentials are then orphaned in the database and it's not possible to delete them afterwards.

That's because the `delete` operation on devices of the device management API checks for the existance of the tenant.

It's  not possible to create a device without creating the tenant first therefore it seems like a device is bound to the tenant lifecycle just as a device-credential is bound to the device lifecycle.

The proposal is therefore to cascade the deletion of the tenant so devices (and credentials) belonging to it are deleted as well.

If the loose coupling of tenants and devices in Hono is wanted, a workaround could be the introduction of a new API request:
"delete all devices" which deletes all the devices and credentials of a tenant (preferrably this operation should also work when the tenant is already gone).

sophokles73:
The Registry Management API has been extended with an operation to
delete all devices (and their credentials) of a tenant.

Both the Mongo DB and the JDBC based registry implementations support
this operation.

Fixes #2793


sophokles73:
@kaniyan would you mind taking a look?

kaniyan (approved):
LGTM

PR #2856, opened 2021-09-20
Use ConfigMapping instead of deprecated ConfigProperties

### LINKED ISSUE #2831: MQTT Adapter/Quarkus: "mapperEndpoints" configuration not used
When using the Quarkus MQTT adapter, the 
```
hono.mqtt.mapperEndpoints.<mapperName>.host
hono.mqtt.mapperEndpoints.<mapperName>.port
hono.mqtt.mapperEndpoints.<mapperName>.uri
```
configuration options are not taken into account.

sophokles73:
The ConfigMapping class of SmallRye supports mapping of arbitrary
configuration properties to a Map.

I had to rename the Java system property used to select the messaging infrastructure for the integration tests because it conflicted with the property bound to Hono client configuration properties used in protocol adapters.

Fixes #2831


sophokles73:

core/pom.xml
```diff
@@ -88,6 +88,11 @@
       <artifactId>jaxb-api</artifactId>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>io.smallrye.config</groupId>
+      <artifactId>smallrye-config</artifactId>
+      <optional>true</optional>
```
calohmn, line 94: What is the motivation of marking this dependency as optional here? The `AbstractConfig` class, requiring this dependency, looks like an important part of the hono-core module.

services/auth-base/src/test/java/org/eclipse/hono/authentication/file/QuarkusPropertyBindingTest.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+@QuarkusTest
+public class QuarkusPropertyBindingTest {
+
+    @Inject
+    FileBasedAuthenticationServiceOptions serviceOptions;
+
+    /**
+     * Verifies that Quarkus correctly binds properties from a yaml file to a
+     * {@link ServiceOptions} instance.
```
calohmn, line 38: `ServiceOptions` cannot be resolved here.

service-base-quarkus/src/main/java/org/eclipse/hono/service/quarkus/metric/HealthCheckServerProducer.java
```diff
@@ -69,10 +70,11 @@
     @Produces
     HealthCheckServer healthCheckServer(
             final Vertx vertx,
-            @ConfigPrefix("hono.healthCheck")
-            final ServerConfig healthCheckServerConfig,
+            @ConfigMapping(prefix = "hono.healthCheck")
+            final ServerOptions healthCheckServerOptions,
```
calohmn, line 74: Parameter name in javadoc doesn't match.

adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
```diff
@@ -157,6 +145,80 @@
...
+            @ConfigMapping(prefix = "hono.tenant")
+            final RequestResponseClientOptions options) {
+        final var props = new RequestResponseClientConfigProperties(options);
+        props.setServerRoleIfUnknown("Tenant");
+        props.setNameIfNotSet(getComponentName());
+        this.tenantClientConfig = props;
+    }
+
+    @Inject
+    void setdeviceRegistrationClientConfig(
```
calohmn, line 186: Capital D (`setDevice[...]`)?

services/command-router-quarkus/src/main/java/org/eclipse/hono/commandrouter/quarkus/Application.java
```diff
@@ -127,10 +116,58 @@ public void onClose(final WatcherException cause) {
...
+            @ConfigMapping(prefix = "hono.tenant")
+            final RequestResponseClientOptions options) {
+        final var props = new RequestResponseClientConfigProperties(options);
+        props.setServerRoleIfUnknown("Tenant");
+        props.setNameIfNotSet(getComponentName());
+        this.tenantClientConfig = props;
+    }
+
+    @Inject
+    void setdeviceRegistrationClientConfig(
```
calohmn, line 162: Capital D (`setDevice[...]`)?

adapter-base-quarkus/src/test/java/org/eclipse/hono/adapter/quarkus/QuarkusPropertyBindingTest.java
```diff
@@ -107,4 +112,16 @@ public void testThatNumberArePresentAsStrings() {
         assertThat(kafkaProducerConfig.getProducerConfig("test").get("number")).isEqualTo("123");
     }
 
+    /**
+     * Verifies that Quarkus correctly binds properties from a yaml file to a
+     * {@link ProtocolAdapterOptions} instance.
+     */
+    public void testProtocolAdapterOptionsBinding() {
```
calohmn, line 119: Missing `@Test` annotation.

service-base-quarkus/src/test/resources/application.yaml
```diff
@@ -0,0 +1,64 @@
...
+    initialCredits: 55
+    keyPath: "/etc/key.pem"
+    linkEstablishmentTimeout: 1111
+    maxFrameSize: 32000
+    maxMessageSize: 64000
+    maxSessionFrames: 30
+    minMaxMessageSize: 65000
+    name: "client"
+    password: "secret"
+    pathSeparator: "-"
```
calohmn, line 40: `GenericOptions#pathSeparator` is documented as `character separating the segments of target addresses.`.
Using `-` here seems strange to me - I don't see how this would work.
In general, it seems to me this pathSeparator isn't actually used in non-test code. So how about removing it altogether?

core/pom.xml
sophokles73: Not necessarily. If you use Spring Boot based configuration, then you so not need it. Once we have removed all Spring Boot based code/components, we can/should make it a required dependency ...

service-base-quarkus/src/test/resources/application.yaml
sophokles73: In this test case we only verify the correct binding of properties. The particular value of the property is chosen arbitrarily.
Removing the pathSeparator altogether seems reasonable to me but should be done in a separate issue FMPOV.

core/pom.xml
calohmn: But as I understand it, the `hono-service-base-spring` module would still need this dependency because the `AbstractConfig` class is used there (e.g. using the `ServerConfig` sub class of `AbstractConfig` being used in the `hono-service-base-spring` `AbstractAdapterConfig` class).

sophokles73: As far as I can tell, there is no *AbstractAdapterConfig* class in the *hono-service-base-spring* module.

sophokles73:
@calohmn I have pushed fixes ...

core/pom.xml
calohmn: Sorry, I meant `AbstractAdapterConfig` in `hono-adapter-base-spring`. 
The point is that I think in practise, the spring modules still need the `smallrye-config` dependency via usage of the `AbstractConfig` class.

sophokles73: Why do you think so? FMPOV no classes from smallrye-config will be needed/used during runtime in a Spring Boot based component, or am I mistaken?

calohmn: As I see it, during runtime the `AbstractConfig` class needs to be loaded by the Spring Boot based components. And this means the `smallrye-config` classes will be loaded as well (via the `GenericOptions` class used in the corresponding `AbstractConfig` constructor).
You could for example delete the `AbstractConfig` class and see that there is a compile error in `hono-service-base` `AbstractServiceBase` and in `hono-adapter-base-spring` `AbstractAdapterConfig`, for example.

sophokles73: It is true that the AbstractConfig class is being loaded but that doesn't mean that all classes that have been required to compile it are being loaded as well. If the constructor accepting the GenericOptions argument is never invoked, then the GenericOptions class will not be loaded. As the Spring Boot based components to not bind any properties to the *Options classes, these constructors are never invoked. This is also consistent with the fact that the integration tests using the Spring Boot based components succeed but none of these components' Docker containers contain the *smallrye-config* artifact ...

calohmn (approved):
LGTM

core/src/main/java/org/eclipse/hono/config/quarkus/AuthenticatingClientOptions.java
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Gets the name or literal IP address of the host that the client is configured to connect to.
+     * <p>
+     * The default value of this property is <em>localhost</em>.
+     *
+     * @return The host name.
+     */
+    @WithDefault("localhost")
+    String host();
```
calohmn, line 45: @sophokles73 I found an issue here:
Setting the default to "localhost" will let `AuthenticatingClientConfigProperties#isHostConfigured()` always be `true`.

This means that if only Kafka is configured as messaging system, there will still be AMQP connection attempts.
I got corresponding connection attempt errors when deploying the Hono 1.10 SNAPSHOT components using the iot-packages hono helm chart with the 
`--set messagingNetworkTypes[0]=kafka --set kafkaMessagingClusterExample.enabled=true --set amqpMessagingNetworkExample.enabled=false`
params.

sophokles73: I see. Do you want to fix it yourself or do you want me to do it? I am fine with both options ...

calohmn: Could you do it? I'm still working on the added metrics.

sophokles73: Sure, will do ...

sophokles73: I have pushed a8320ac1c59171984f1d612ff16eb319178ae24d which (hopefully) fixes the issue

PR #2858, opened 2021-09-20
[#2857] Limit message TTL to tenant's max-ttl with Kafka.

### LINKED ISSUE #2857: TTL in downtream messages with Kafka is not capped by max-TTL
The tenant API provides the possibility to define a _max-ttl_ for a tenant. The Kafka-based downstream sender does not limit the TTL by this value.

b-abel:
Fixes #2857.

### DIFF
--- a/clients/telemetry-kafka/src/main/java/org/eclipse/hono/client/telemetry/kafka/AbstractKafkaBasedDownstreamSender.java
+++ b/clients/telemetry-kafka/src/main/java/org/eclipse/hono/client/telemetry/kafka/AbstractKafkaBasedDownstreamSender.java
@@ -26,6 +26,8 @@ import org.eclipse.hono.client.kafka.producer.AbstractKafkaBasedMessageSender;
 import org.eclipse.hono.util.MessageHelper;
 import org.eclipse.hono.util.QoS;
 import org.eclipse.hono.util.RegistrationAssertion;
+import org.eclipse.hono.util.ResourceLimits;
+import org.eclipse.hono.util.TenantConstants;
 import org.eclipse.hono.util.TenantObject;
 
 import io.opentracing.SpanContext;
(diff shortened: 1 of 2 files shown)


clients/telemetry-kafka/src/main/java/org/eclipse/hono/client/telemetry/kafka/AbstractKafkaBasedDownstreamSender.java
```diff
@@ -154,14 +156,34 @@ public AbstractKafkaBasedDownstreamSender(
...
+                    ttlSeconds = ttl;
+                }
             } else {
-                log.error("TTL property is not a number [{}]", v);
-                return v;
+                if (maxTtl != TenantConstants.UNLIMITED_TTL) {
+                    log.debug("setting TTL to tenant's max TTL [{}s]", maxTtl);
+                    ttlSeconds = maxTtl;
+                } else {
+                    log.error("TTL property is not a number [{}]", v);
```
sophokles73, line 182: this is also true in the case above, isn't it?

clients/telemetry-kafka/src/test/java/org/eclipse/hono/client/telemetry/kafka/AbstractKafkaBasedDownstreamSenderTest.java
```diff
@@ -294,44 +294,153 @@ private void assertContentType(final String expectedContentType, final String co
...
+     * configured, the latter one is used.
+     *
+     * @param ctx The vert.x test context.
+     */
+    @Test
+    public void testThatTtlFromDefaultsIsUsed(final VertxTestContext ctx) {
+        final MockProducer<String, Buffer> mockProducer = KafkaClientUnitTestHelper.newMockProducer(true);
+        final AbstractKafkaBasedDownstreamSender sender = newSender(mockProducer);
+
+        // GIVEN a tenant with maximum TTL configured and a lower one in the properties
```
sophokles73, line 344: .. a lower one configured in the defaults

clients/telemetry-kafka/src/main/java/org/eclipse/hono/client/telemetry/kafka/AbstractKafkaBasedDownstreamSender.java
b-abel: Yes. It was a mistake to add this log statement. The value `v` is especially not a number if it is `null`.

sophokles73 (approved):
LGTM

PR #2859, opened 2021-09-21
[#2846] Add support for defining failure-notification headers

### LINKED ISSUE #2846: Enhance failure notifications with header from original command
We want to use the recently added [failure notifications](https://www.eclipse.org/hono/docs/api/command-and-control-kafka/#send-a-requestresponse-command) in Eclipse Ditto to provide meaningful error responses to live messages (commands in Hono speak) sent to a device via Hono. However, when receiving such failure notifications, we do not have enough context to build a response that the original sender understands properly. One solution could be to add the required information to the command (e.g. a defined header) and Hono adds the header again to the failure notification when sending it back. What do you think?

calohmn:
This fixes #2846:
Kafka command message headers with the `delivery-failure-notification-metadata` prefix are now adopted for the corresponding command response message sent in case of a command delivery failure.

### DIFF
--- a/clients/command-amqp/src/main/java/org/eclipse/hono/client/command/amqp/ProtonBasedCommandResponseSender.java
+++ b/clients/command-amqp/src/main/java/org/eclipse/hono/client/command/amqp/ProtonBasedCommandResponseSender.java
@@ -14,6 +14,7 @@
 
 package org.eclipse.hono.client.command.amqp;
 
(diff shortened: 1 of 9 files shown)


sophokles73 (approved):
LGTM

calohmn:
(Applied fix in `ProtonBasedCommandResponseSender` following integration test failures.)

PR #2862, opened 2021-09-22
Merge device registry user guides

sophokles73:
The three user guides for the different registry implementations have
been merged into a single user guide in order to reduce confusion among
readers.

@b-abel would you mind to also take a look?


site/documentation/content/user-guide/file-based-device-registry.md
```diff
@@ -1,41 +1,51 @@
 +++
 title = "File Based Device Registry"
+hidden = true
```
b-abel, line 3: Why hiding the documentation? FMPOV stating that it is deprecated would be enough. I would hide it once it has been removed.

sophokles73: My thinking was that now that we have a single Device Registry user guide page and we do not want any new users to use the file based registry, we do not need to mention it in the 1.10.0 documentation anymore. It will still be available in the < 1.10.0 docs, won't it?

b-abel: > It will still be available in the < 1.10.0 docs, won't it?

Yes. 

And it will still be available in version 1.10.0. You can still visit the hidden page if you open it from a link or open it in an older version and then select the current one. It only won't be displayed in the menu.

sophokles73: Sounds reasonable to me.

b-abel:
LGTM

calohmn (approved):
LGTM

PR #2865, opened 2021-09-24
Improve tracing/logging on AMQP/MQTT server shutdown

calohmn:
Giving tracing spans created for closing client connections on server shutdown a dedicated operation name.
That makes them stand out from spans created when clients themselves close the connection.

### DIFF
--- a/adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
+++ b/adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
@@ -22,6 +22,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.OptionalInt;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


calohmn:
We *could* also create a span for the server shudown procedure and create the "connection-closing" spans as child spans. But this could create lots and lots of child spans if many clients are connected. So I think the change in this PR is enough for the time being.

adapters/amqp-vertx-base/src/main/java/org/eclipse/hono/adapter/amqp/VertxBasedAmqpProtocolAdapter.java
```diff
@@ -241,33 +243,37 @@ private ConnectionLimitManager createConnectionLimitManager() {
 
     @Override
     protected void doStop(final Promise<Void> stopPromise) {
+        if (!stopCalled.compareAndSet(false, true)) {
+            log.trace("stop already called");
+            stopPromise.complete();
```
sophokles73, line 248: FMPOV this should instead be tied to the outcome of the promise that had initially been used to shut down the adapter, e.g. by using an AtomicReference<Promise<Void>> instead ...

calohmn: I've amended the commit accordingly.

sophokles73 (approved):
LGTM

PR #2869, opened 2021-09-28
Use Kafka consumer 'max.poll.interval.ms' default

calohmn:
Setting a much lower 'max.poll.interval.ms' value may lessen the impact of UNKNOWN_TOPIC_OR_PARTITION errors while trying to commit offsets but may also cause "leave group" requests and rebalances if poll invocations get delayed for some reason. Therefore better to stick to the default 'max.poll.interval.ms' if no such value is configured explicitly.


calohmn:
The integration test
`KafkaBasedCommandConsumerFactoryImplIT.testCommandsGetForwardedIfOneConsumerInstanceGetsClosed`
seems to repeatably fail with this.
I'm having a look...

PR #2873, opened 2021-09-29
Adapt environment variable names to Quarkus requirements

sophokles73:
Quarkus uses SmallRye under the hood for binding environment variables
and/or Java system properties to configuration objects as described in
https://quarkus.io/guides/config-mappings. The VERBATIM naming strategy
that is used in Hono does not allow environment variables to contain an
underscore character for a change in case of the property name. For
example the property "useDefaultMetrics" of a configuration object
requires an environment variable "USEDEFAULTMETRICS" instead of
"USE_DEFAULT_METRICS" using this naming strategy.

The admin guides have been updated accordingly.


site/documentation/content/admin-guide/file-based-device-registry-config.md
```diff
@@ -177,11 +191,11 @@ can be ignored.
...
+| :---------------------------------------------- | :-------: | :------ | :------------------------------------------ |
+| `HONO_AUTH_VALIDATION_CERTPATH`<br>`hono.auth.validation.certPath` | no  | - | The absolute path to the PEM file containing the an X.509 certificate that the service should use for validating tokens issued by the Authentication service. Alternatively, a symmetric key can be used for validating tokens by setting the `HONO_AUTH_VALIDATION_SHAREDSECRET` variable. If none of these variables is set, startup of the service fails. |
+| `HONO_AUTH_VALIDATION_SHAREDSECRET`<br>`hono.auth.validation.sharedSecret` | no  | - | A string to derive a symmetric key from which is used for validating tokens issued by the Authentication service. The key is derived from the string by using the bytes of the String's UTF8 encoding. When setting the validation key using this variable, the Authentication service **must** be configured with the same key. Alternatively, an X.509 certificate can be used for validating tokens by setting the `HONO_AUTH_VALIDATION_CERTPATH` variable. If none of these variables _SUPPORTEDSASLMECHANISMSvice fails. |
```
calohmn, line 197: Error at the end (`If none of these variables _SUPPORTEDSASLMECHANISMSvice fails.`).

calohmn (approved):
LGTM

PR #2875, opened 2021-09-30
Publish hono-legal jar again

calohmn:
So that it can be used in the hono-extras project.

See https://github.com/eclipse/hono-extras/issues/12.


sophokles73 (approved):
LGTM

PR #2879, opened 2021-10-06
Count connection attempts to auth server

sophokles73:
The Authentication Server has been extended to report the number of
attempts made by clients to establish an AMQP 1.0 connection to the
server.


sophokles73:

PR #2902, opened 2021-10-15
Include reasonable detail message in exceptions

### LINKED ISSUE #2900: Use error text for 429 (too many requests) errors
When the resource limit concerning the maximum number of messages is reached, further messages will be responded to with a 429 Client error exception, but with no corresponding error message.

In the case of Kafka command messages for example, a delivery failure error response message with the `error` field value "Error Code: 429" will be generated.

Error reporting should be improved here by using an explicit error message like "message limit exceeded".

sophokles73:
The exceptions thrown in the context of resource limit violations have
been changed to include a reasonable, localizable detail message.

Fixes #2900


sophokles73:

Issue #2903, opened 2021-10-20
Kafka command message potentially getting skipped


calohmn:
**Scenario / Problem:**
A downstream application sends some Kafka command & control messages, then doesn't send any messages for a relatively long time (longer than the retention time of records in the Kafka command topic) and then again sends a command message.

If this last command message gets sent while a rebalance is happening concerning the Hono Command Router Kafka consumers, the command message might not get processed by the Hono Command Router.

**Explanation:**
If the time the downstream application doesn't send any command messages in the above scenario is longer than the retention time of records in the Kafka command topic ([1]), the committed offset for the last handled command record at some point will refer to a record that has already been deleted in Kafka.

When the corresponding topic partition gets reassigned during an upcoming rebalance, there will be this log message ("position is out of range"):
```
INFO  o.a.k.c.consumer.internals.Fetcher - [Consumer clientId=consumer-x, groupId=cmd-router-group] Fetch position FetchPosition{offset=30, [...]} is out of range for partition hono.command.myTenantId0-1, resetting offset
```
Since the Hono Kafka Consumer for command messages uses the "latest" offset reset policy, the consumer offset position will be reset to the latest offset.  (Unfortunately, that offset position won't be used for the manual offset commits happening from then on.)

If the downstream application sends another command in the (rather short) period after the rebalance has started and this offset reset has happened, this command won't get received by the consumer.

**Solution:**
The possibility of committed offsets referring to already deleted records should be reduced by (re-)committing offsets more often. This means reducing the `AsyncHandlingAutoCommitKafkaConsumer#DEFAULT_OFFSETS_SKIP_RECOMMIT_PERIOD` default and documenting that the "hono.offsets.skip.recommit.period.seconds" config property should be set if a short retention time is used.

Rebalance handling in the Hono Kafka consumer should detect the above "position is out of range" case and make sure that the manual offset commits done from then on don't use the obsolete offset.

---
[1]
Also if the time is shorter than the interval in which the Hono Kafka Consumer re-commits an already committed partition offset, see `AsyncHandlingAutoCommitKafkaConsumer#DEFAULT_OFFSETS_SKIP_RECOMMIT_PERIOD`.

PR #2907, opened 2021-10-21
[#2903] Ensure out-of-range position gets reset

calohmn:
This is for #2903:
For Kafka consumers with auto offset reset config "latest", the consumer position for a newly assigned partition will be reset to the latest offset if the committed offset refers to an already deleted record. This makes position handling after start/rebalance more consistent, ensuring that records sent just after consumer.start() completed or during a rebalance are actually getting received.


sophokles73:
@calohmn I do not think that I am qualified to review this PR. Maybe @b-abel can jump in?

b-abel:
> @calohmn I do not think that I am qualified to review this PR. Maybe @b-abel can jump in?

Sure, I'll have a look.

Issue #2837, opened 2021-08-27
Define Internal Notification API for communication among Hono components


calohmn:
Coming out of the discussion in #293, being a precondition for solving that issue:

Hono components should be able to send internal event messages among themselves.
The first use case would be the device registry sending event/notification messages when tenant/device registration data were updated/deleted. These messages can then be received by the adapters and command router components to perform various tasks like:
- clear tenant/device related cache data (#2984)
- disconnect MQTT/AMQP devices in case of deleted device/tenant (#293)
- prevent tenant Kafka topic related operations in the Command Router, like committing offsets (#2838).

The idea is to use the messaging network (AMQP / Kafka) for transporting these messages.
In the AMQP messaging network case, the messages would be sent on a fixed _multicast_ address (e.g. "internal_event"). If Kafka is used, a corresponding Kafka topic would be used.

Reading on this address/topic is only meant to be done from the Hono components, not northbound applications (hence no need to put a tenant id in the address/topic name).

b-abel:
I don't have another use case ATM. From my point of view, it is advantageous to have one unified mechanism and not several individual ones. My understanding of the discussion so far is that it was agreed that an event-driven mechanism makes sense for these use cases.  What has not been made explicit so far is that the cost is that the recipients have to manage their state in some way. E.g., for Use Case 2, the adapter must maintain a list of all connected devices in order to disconnect them when the notification arrives. But this is probably not a problem.
  
My proposal in #2837 builds on the concept of [Event Notifications](https://martinfowler.com/articles/201701-event-driven.html#EventNotification). the main feature of this is that the sending component makes no assumptions about whether and how possible recipients will use the information. This fits well with @calohmn suggestion to use Kafka or AMQP broadcasts. 

That said, @sophokles73  I'd love to hear your idea on how this could be simplified.

b-abel:
I think it should be closed with #2994.

calohmn:
#2994 is merged now.

PR #2917, opened 2021-10-27
[#2837] Add Generic Notification API and implementation for the Device Registry

b-abel:
This comes from the discussion in #2905. It contains: 
- a generic Notification API
- a concrete Notification API for the Device Registry
- abstract client implementations of the Notification API for Kafka 
- a concrete client implementation of the Device Registry Notification API for Kafka

The would put the AMQP based notification receiver for Device Registry notifications in the existing module _hono-client-registry-amqp_.

This is an alternative to #2918.


sophokles73:
closed in favor of #2918

PR #2918, opened 2021-10-27
[#2837] Add Device Registry specific Notification API

b-abel:
This comes from the discussion in #2905. It contains:

- a Notification API for the Device Registry
- client implementations of the Device Registry Notification API for Kafka

The would put the AMQP based notification receiver in the existing module _hono-client-registry-amqp_.

This is an alternative to #2917.

### DIFF
--- a/clients/notification-kafka/src/main/java/org/eclipse/hono/client/notification/KafkaBasedNotificationReceiver.java
+++ b/clients/notification-kafka/src/main/java/org/eclipse/hono/client/notification/KafkaBasedNotificationReceiver.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.hono.client.notification;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Supplier;
+
+import org.apache.kafka.clients.consumer.Consumer;
+import org.eclipse.hono.client.kafka.AbstractKafkaConfigProperties;
+import org.eclipse.hono.client.kafka.consumer.HonoKafkaConsumer;
+import org.eclipse.hono.notification.AbstractNotification;
+import org.eclipse.hono.notification.NotificationReceiver;
+import org.eclipse.hono.util.Strings;
+
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.json.Json;
+import io.vertx.kafka.client.consumer.KafkaConsumerRecord;
+
+/**
+ * A client for receiving notifications with Kafka.
+ */
+public class KafkaBasedNotificationReceiver implements NotificationReceiver {
+
+    private final Vertx vertx;
+    private final Map<String, String> consumerConfig = new HashMap<>();
+    private final Map<Class<? extends AbstractNotification>, Handler<? extends AbstractNotification>> handlerPerType = new HashMap<>();
+    private final Set<String> topics = new HashSet<>();
+
+    private Supplier<Consumer<String, Buffer>> kafkaConsumerSupplier;
+    private HonoKafkaConsumer honoKafkaConsumer;
+    private boolean started = false;
+
+    /**
+     * Creates a client for consuming notifications.
+     *
+     * @param vertx The vert.x instance to be used.
+     * @param consumerConfig The configuration for the Kafka consumer.
+     *
+     * @throws NullPointerException if any parameter is {@code null}.
+     * @throws IllegalArgumentException if consumerConfig does not contain a valid configuration (at least a bootstrap
+     *             server).
+     */
+    public KafkaBasedNotificationReceiver(final Vertx vertx, final Map<String, String> consumerConfig) {
+
+        Objects.requireNonNull(vertx);
+        Objects.requireNonNull(consumerConfig);
+
+        if (Strings.isNullOrEmpty(consumerConfig.get(AbstractKafkaConfigProperties.PROPERTY_BOOTSTRAP_SERVERS))) {
+            throw new IllegalArgumentException("No Kafka configuration found!");
+        }
+
+        this.vertx = vertx;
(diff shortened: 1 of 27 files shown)


b-abel:
I just changed the code to support generic types of notifications.

clients/notification-kafka/src/main/java/org/eclipse/hono/client/notification/NotificationTopicHelper.java
```diff
@@ -0,0 +1,68 @@
...
+        } else if (DeviceChangeNotification.class.equals(notificationType)) {
+            address = DeviceChangeNotification.ADDRESS;
+        } else if (CredentialsChangeNotification.class.equals(notificationType)) {
+            address = CredentialsChangeNotification.ADDRESS;
+        } else {
+            throw new IllegalArgumentException("Unknown notification type " + notificationType.getName());
+        }
+
+        return getTopicNameForAddress(address);
+    }
```
sophokles73, line 63: Is there a reason why we need both these methods? Can't we always derive the address from the `ADDRESS` field?

b-abel: > Can't we always derive the address from the `ADDRESS` field?

That is exactly what the methods do. The address is always simply a constant defined by the ADDRESS field. The topic name is the address prefixed by `hono.notification.`. On the receiver side, we need to determine which address field to use. The point here is that we need a static value because the client subscribes to it before having an instance of the notification at hand. 
We could move the logic into the Kafka sender respectively receiver implementations but I don't know any benefits of this. FMPOV it is clearer to have one place where the logic resides that maps between notification types and their topics.

b-abel: Ah... Now I get it. I'll do this.

clients/notification-kafka/src/main/java/org/eclipse/hono/client/notification/KafkaBasedNotificationReceiver.java
```diff
@@ -0,0 +1,136 @@
...
+    @SuppressWarnings("unchecked")
+    void handleNotification(final AbstractNotification abstractNotification) {
+        for (final var entry : handlerPerType.entrySet()) {
+            final Class<? extends AbstractNotification> type = entry.getKey();
+
+            if (type.isInstance(abstractNotification)) {
+                entry.getValue().handle(abstractNotification);
+                break;
+            }
+        }
```
sophokles73, line 116: can't we simply look up the handler by the notification's type directly, i.e.
```java
final var handler = handlerPerType.get(abstractNotification.getClass());
```

b-abel: Right, thank you. I included this,

clients/notification-kafka/src/main/java/org/eclipse/hono/client/notification/KafkaBasedNotificationReceiver.java
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * A client for receiving notifications with Kafka.
+ */
+public class KafkaBasedNotificationReceiver implements NotificationReceiver {
+
+    private final Vertx vertx;
+    private final Map<String, String> consumerConfig = new HashMap<>();
+    @SuppressWarnings("rawtypes")
+    private final Map<Class<? extends AbstractNotification>, Handler> handlerPerType = new HashMap<>();
```
sophokles73, line 46: how about
```java
private final Map<Class<? extends AbstractNotification>, Handler<? extends AbstractNotification>> handlerPerType = new HashMap<>();
```

clients/notification-kafka/src/main/java/org/eclipse/hono/client/notification/KafkaBasedNotificationSender.java
```diff
@@ -0,0 +1,123 @@
...
+
+        try {
+
+            final String topic = NotificationTopicHelper.getTopicName(notification.getClass());
+            final String key = getKey(notification);
+            final JsonObject value = JsonObject.mapFrom(notification);
+            final KafkaProducerRecord<String, JsonObject> record = KafkaProducerRecord.create(topic, key, value);
+
+            final Promise<RecordMetadata> sendPromise = Promise.promise();
+            producerFactory.getOrCreateProducer(PRODUCER_NAME, config).send(record, sendPromise);
```
sophokles73, line 78: will this create the topic implicitly or does the topic need to be created upfront?

clients/notification-kafka/src/main/java/org/eclipse/hono/client/notification/KafkaBasedNotificationReceiver.java
```diff
@@ -0,0 +1,129 @@
...
+        handlerPerType.put(notificationType, consumer);
+    }
+
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    private Handler<KafkaConsumerRecord<String, Buffer>> getRecordHandler() {
+        return record -> {
+            final AbstractNotification abstractNotification = Json.decodeValue(record.value(),
+                    AbstractNotification.class);
+            final Handler handler = handlerPerType.get(abstractNotification.getClass());
+            handler.handle(abstractNotification);
```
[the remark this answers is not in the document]
b-abel: FMPOV a notification should always have a value. So, throwing an exception is appropriate. The `HonoKafkaConsumer` catches all exceptions thrown by the record handler and prints them to the logs. Are you OK with this behavior?

clients/notification-kafka/src/main/java/org/eclipse/hono/client/notification/KafkaBasedNotificationSender.java
b-abel: This depends on the broker configuration. I would add that to the documentation.

clients/notification-kafka/src/main/java/org/eclipse/hono/client/notification/KafkaBasedNotificationReceiver.java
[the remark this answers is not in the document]
sophokles73: I was referring to the `handler` which might be `null` here ...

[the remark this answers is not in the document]
b-abel: Added that.

sophokles73 (approved):
LGTM

clients/notification-kafka/src/main/java/org/eclipse/hono/client/notification/KafkaBasedNotificationSender.java
sophokles73: The documentation will be in another PR, I guess?

b-abel: Yes. I think this is a concern of the component that publishes a notification.

Issue #2904, opened 2021-10-20
Use Kafka client specific consumer/producer config properties


calohmn:
Currently, the configuration of the Kafka consumer for receiving command messages in the Command Router component is done via the generic Kafka configuration properties:
```
hono:
  kafka:
    consumerConfig:
      bootstrap.servers: [...]
      [...]
```

Especially with the introduction of another kind of Kafka consumer in the Command Router (see #2837), there should be a way to configure the Kafka *command* consumer explicitly.

The Command Router Kafka command consumer config needs:
- standard Kafka config properties passed on to the Kafka consumer
- specific properties introduced by the Hono consumer classes, like `HonoKafkaConsumer#POLL_TIMEOUT` and the `AsyncHandlingAutoCommitKafkaConsumer#CONFIG[*]` properties
(see also the Hono `KafkaConsumerConfigProperties` class, currently not used for these classes)

The definition of the new config structure should be done taking into account how the AMQP messaging network command consumer is configured. This is currently done like this:
```
hono:
  command:
    name: 'Hono Command Router'
    host: someHost
    tlsEnabled: true
    trustStorePath: /etc/hono/trusted-certs.pem
    initialCredits: 2000
    sendMessageTimeout: 5000
    linkEstablishmentTimeout: 1000
```

b-abel:
What about using the same keywords as with AMQP with the difference that they are not prefixed by `hono.` but instead by `hono.kafka.`? For example: `hono.kafka.messaging` and  `hono.kafka.command`. 

~~~yaml
hono:
  kafka:
    command: 
      pollTimeout: 50L
      commonClientConfig:
        bootstrap.servers: ${hono.kafka.bootstrap.servers}
~~~

b-abel:
This could be considered breaking change because the components will no longer work with an existing configuration which also affects northbound clients like the CLI. On the other hand, I think there is no need to wait for Hono 2.0 because it only affects the Kafka support which is still under development. For me, the only consequence of this is that we should not rush this and avoid that we have to make changes in the structure of the configuration again in one of the next versions. 
WDYT @sophokles73, @calohmn?

b-abel:
My proposal above could for example look like the following (MQTT adapter, configured for AMQP _and_ Kafka):

~~~yaml
hono:
  messaging:
    name: Hono adapter-mqtt-vertx
    amqpHostname: hono-internal
    host: eclipse-hono-dispatch-router
    port: 5673
[...]

PR #2924, opened 2021-11-03
[#2837] Publish notifications from the Device Registry with Kafka.

b-abel:
This is for #2837. It adds notification support to the device registry using the Kafka client.


b-abel:
The Kafka producers and consumers for the notifcations use `JSON` as the type of the record values. The class `KafkaProducerConfigProperties` on the other hand expects the value to be of type `Buffer`. The difference between both is not so much technical, rather the APIs are different. We could change the type for notifications to `Buffer`, but this becomes obsolete with #2904 when we use different configuration properties. I tend to defer this PR until #2904 is resolved. @calohmn @kaniyan Are you planning to work on #2904?

services/device-registry-base/src/test/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsManagementServiceTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * Verifies that
+     * {@link AbstractCredentialsManagementService#updateCredentials(String, String, List, Optional, Span)} publishes
+     * the expected notification.
+     *
+     * @param context The vert.x test context.
+     */
+    @Test
+    public void testNotificationOnUpdateDevice(final VertxTestContext context) {
```
sophokles73, line 120: onUpdateCredentials?

sophokles73:
@b-abel can you rebase? I have already incorporated the fix for calculating the command sending timeout in the ITs into master ...

b-abel:
> @b-abel can you rebase? I have already incorporated the fix for calculating the command sending timeout in the ITs into master ...

Done.

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/device/AbstractDeviceManagementService.java
```diff
@@ -372,6 +406,7 @@ protected String generateDeviceId(final String tenantId) {
                     }
                     return processDeleteDevicesOfTenant(tenantId, span);
                 })
+                // TODO send notification
```
sophokles73, line 409: are you going to address this as part of this PR or will you create another PR for it?

b-abel: TBH I haven't thought much about this yet and have no idea how to address this. So, yes this should be part of another PR.

sophokles73: Ok, then please don't forget to do so ;-) I can imagine using a dedicated notification type for that purpose.

sophokles73 (approved):
LGTM

PR #2926, opened 2021-11-04
[#2921] Improve command processing rate-limiting

### LINKED ISSUE #2921: Fix handling of long CommandRouter Kafka consumer pause (due to rate-limiting)
A Command Router Kafka consumer gets paused if the number of concurrently handled, unfinished commands exceeds a given limit (see #2767).
A pause longer than 5min (`max.poll.interval.ms`) will let the consumer leave the active consumer group. This causes commit errors:
```
00:02:05.288 paused consumer record polling; max no. of records in processing exceeded (current: 501, max: 501) [client-id: hono-command-router-consumer-xxx]

00:07:05.294 [Consumer clientId=hono-command-router-consumer-xxx, groupId=cmd-router-group] Member hono-command-router-consumer-xxx sending LeaveGroup request to coordinator xxx:9092 (id: xxx rack: null) due to consumer poll timeout has expired. This means the time between subsequent calls to poll() was longer than the configured max.poll.interval.ms, which typically implies that the poll loop is spending too much time processing messages. You can address this either by increasing max.poll.interval.ms or by reducing the maximum size of batches returned in poll() with max.poll.records.

00:40:25.387 [Consumer clientId=hono-command-router-consumer-xxx, groupId=cmd-router-group] Failing OffsetCommit request since the consumer is not part of an active group
```
The rate-limiting mechanism needs to be adapted so that this doesn't happen.

calohmn:
This fixes #2921:
With the new rate-limiting approach, the poll invocations won't get paused for too long, meaning the consumer won't leave the active consumer group in case of prolonged high load.

### DIFF
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/HonoKafkaConsumer.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/HonoKafkaConsumer.java
@@ -55,6 +55,7 @@ import io.vertx.kafka.client.common.TopicPartition;
 import io.vertx.kafka.client.common.impl.Helper;
 import io.vertx.kafka.client.consumer.KafkaConsumer;
 import io.vertx.kafka.client.consumer.KafkaConsumerRecord;
+import io.vertx.kafka.client.consumer.KafkaConsumerRecords;
 import io.vertx.kafka.client.consumer.impl.KafkaReadStreamImpl;
 
 /**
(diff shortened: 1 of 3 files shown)


clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/HonoKafkaConsumer.java
```diff
@@ -264,15 +267,70 @@ public void setKafkaConsumerSupplier(final Supplier<Consumer<String, Buffer>> su
...
     }
 
+    /**
+     * Suspends fetching records from all partitions assigned to this consumer (if not already suspended).
+     * That means the next poll invocations won't return any records.
+     * For already fetched records, the record handler will still be invoked.
+     *
+     * @return {@code true} if the record fetching was active before and is now getting paused.
+     */
+    public final boolean pauseRecordFetching() {
```
calohmn, line 277: As an explanation here:
This method is the way to pause the consumer without pausing the poll loop, so that rebalances can still happen and the consumer doesn't leave the active consumer group.
This approach isn't exclusively used here since it isn't very fine grained. After calling `resumeRecordFetching()` a whole batch of records (per default up to 500 records) will be received with the next poll.
In order to flatten the line of records being currently in processing, a combination with calling the `pauseRecordHandlingAndPolling(timeout)` method is used. The latter method can pause on a per-record basis (implemented in the vert.x KafkaConsumer).

calohmn:
Made a few minor improvements, fixing an edge case where the `recordsInProcessing` exceeds the overall limit when repeatedly only 1 message per poll is getting received all while no records are getting completed.

kaniyan (approved):
LGTM

PR #2929, opened 2021-11-05
Refactor Kafka configuration properties.

b-abel:
This is needed for [ #2924].
Extracted new superclasses for the Kafka producer and consumer configuration properties that do not have the serialization types hardcoded or assume any particular configuration values.

The first commit contains the actual change. The second one only renames the Kafka client config properties implementation classes that are specific for messaging.

### DIFF
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/MessagingKafkaConsumerConfigProperties.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/MessagingKafkaConsumerConfigProperties.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2020, 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.hono.client.kafka.consumer;
+
+import org.apache.kafka.common.serialization.StringDeserializer;
+
+import io.vertx.kafka.client.serialization.BufferDeserializer;
+
+/**
+ * Configuration properties for Kafka consumers used for Hono's messaging.
+ *
+ * Record keys will be deserialized with {@link StringDeserializer}, the values with {@link BufferDeserializer}.
+ *
+ * @see <a href="https://kafka.apache.org/documentation/#consumerconfigs">Kafka Consumer Configs</a>
+ * @see <a href="https://www.eclipse.org/hono/docs/api/telemetry-kafka">Telemetry API for Kafka Specification</a>
+ * @see <a href="https://www.eclipse.org/hono/docs/api/event-kafka">Event API for Kafka Specification</a>
+ * @see <a href="https://www.eclipse.org/hono/docs/api/command-and-control-kafka/">Command &amp; Control API for Kafka Specification</a>
+ */
+public class MessagingKafkaConsumerConfigProperties extends KafkaConsumerConfigProperties {
+
+    /**
+     * Creates an instance.
+     */
+    public MessagingKafkaConsumerConfigProperties() {
+        super(StringDeserializer.class, BufferDeserializer.class);
+    }
+
+}
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/producer/MessagingKafkaProducerConfigProperties.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/producer/MessagingKafkaProducerConfigProperties.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) 2020, 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+
+package org.eclipse.hono.client.kafka.producer;
+
+import java.util.Map;
+
+import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.common.serialization.StringSerializer;
+
+import io.vertx.kafka.client.serialization.BufferSerializer;
+
+/**
+ * Configuration properties for Kafka producers used for Hono's messaging.
+ *
+ * Record keys will be serialized with {@link StringSerializer}, the values with {@link BufferSerializer}.
+ *
+ * The properties that are required by Hono's messaging APIs are will be set in
+ * {@link MessagingKafkaProducerConfigProperties#adaptConfiguration(Map)}.
+ *
+ * @see <a href="https://kafka.apache.org/documentation/#producerconfigs">Kafka Producer Configs</a>
+ * @see <a href="https://www.eclipse.org/hono/docs/api/telemetry-kafka">Telemetry API for Kafka Specification</a>
+ * @see <a href="https://www.eclipse.org/hono/docs/api/event-kafka">Event API for Kafka Specification</a>
+ * @see <a href="https://www.eclipse.org/hono/docs/api/command-and-control-kafka/">Command &amp; Control API for Kafka Specification</a>
+ */
+public class MessagingKafkaProducerConfigProperties extends KafkaProducerConfigProperties {
+
+    /**
+     * Creates an instance.
+     */
+    public MessagingKafkaProducerConfigProperties() {
+        super(StringSerializer.class, BufferSerializer.class);
+    }
+
+    /**
+     * Sets the required properties.
+     *
+     * The following properties are set here to the given configuration:
+     * <ul>
+     * <li>{@code enable.idempotence=true}: enables idempotent producer behavior</li>
+     * </ul>
+     *
+     * @see <a href="https://kafka.apache.org/documentation/#enable.idempotence">The Kafka documentation -
+     *      "Producer Configs" - enable.idempotence</a>
+     */
+    @Override
+    protected final void adaptConfiguration(final Map<String, String> config) {
+        overrideConfigProperty(config, ProducerConfig.ENABLE_IDEMPOTENCE_CONFIG, "true");
+    }
+
+}
--- a/adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
+++ b/adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
@@ -52,14 +52,14 @@ import org.eclipse.hono.client.command.kafka.KafkaBasedCommandResponseSender;
 import org.eclipse.hono.client.command.kafka.KafkaBasedInternalCommandConsumer;
 import org.eclipse.hono.client.kafka.KafkaAdminClientConfigProperties;
 import org.eclipse.hono.client.kafka.KafkaClientOptions;
-import org.eclipse.hono.client.kafka.consumer.KafkaConsumerConfigProperties;
+import org.eclipse.hono.client.kafka.consumer.MessagingKafkaConsumerConfigProperties;
 import org.eclipse.hono.client.kafka.metrics.KafkaClientMetricsSupport;
 import org.eclipse.hono.client.kafka.metrics.KafkaMetricsOptions;
 import org.eclipse.hono.client.kafka.metrics.MicrometerKafkaClientMetricsSupport;
 import org.eclipse.hono.client.kafka.metrics.NoopKafkaClientMetricsSupport;
 import org.eclipse.hono.client.kafka.producer.CachingKafkaProducerFactory;
-import org.eclipse.hono.client.kafka.producer.KafkaProducerConfigProperties;
 import org.eclipse.hono.client.kafka.producer.KafkaProducerFactory;
+import org.eclipse.hono.client.kafka.producer.MessagingKafkaProducerConfigProperties;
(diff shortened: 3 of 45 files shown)


clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/MessagingKafkaConsumerConfigProperties.java
```diff
@@ -27,13 +27,12 @@
...
  */
-// TODO rename to something that expresses that this is about the consumers used for *messaging*
-public class KafkaConsumerConfigProperties extends AbstractKafkaConsumerConfigProperties {
+public class MessagingKafkaConsumerConfigProperties extends AbstractKafkaConsumerConfigProperties {
 
     /**
      * Creates an instance.
      */
-    public KafkaConsumerConfigProperties() {
+    public MessagingKafkaConsumerConfigProperties() {
```
sophokles73, line 35: I do not quite see how this class is specific to Hono's Messaging infrastructure or am I missing something?

b-abel: It enforces the configuration that is required by Hono's messaging APIs as defined in https://www.eclipse.org/hono/docs/api/kafka-api/. While it might be possible to use this configuration in other contexts, this is not always the case. For notifications, we could use JSON as the value type, while the messaging APIs require it to be a plain byte array.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractKafkaConsumerConfigProperties.java
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * Creates an instance.
+     *
+     * @param keyDeserializerClass The class to be used for serializing the record keys, if {@code null} the
+     *            deserializer needs to be provided in the configuration at runtime.
+     * @param valueDeserializerClass The class to be used for serializing the record values, if {@code null} the
+     *            deserializer needs to be provided in the configuration at runtime.
+     */
+    @SuppressWarnings("rawtypes")
+    protected AbstractKafkaConsumerConfigProperties(final Class<? extends Deserializer> keyDeserializerClass,
```
sophokles73, line 56: I guess you want to pass in the type itself instead of just the class name to make sure that it actually is a Deserializer, right?
If so, I believe we can use `final Class<? extends Deserializer<?>> keyDeserializerClass)` and remove the `@SuppressWarning` annotations ...

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractKafkaConsumerConfigProperties.java
```diff
@@ -0,0 +1,162 @@
...
+     * @throws NullPointerException if consumerName is {@code null}.
+     */
+    public final Map<String, String> getConsumerConfig(final String consumerName) {
+        Objects.requireNonNull(consumerName);
+
+        if (commonClientConfig == null && consumerConfig == null) {
+            return new HashMap<>();
+        }
+
+        final Map<String, String> newConfig = new HashMap<>();
```
sophokles73: move this before the previous block?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractKafkaConsumerConfigProperties.java
```diff
@@ -0,0 +1,162 @@
...
+        Optional.ofNullable(valueDeserializerClass).ifPresent(serializerClass -> overrideConfigProperty(newConfig,
+                ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, serializerClass.getName()));
+
+        setUniqueClientId(newConfig, consumerName, ConsumerConfig.CLIENT_ID_CONFIG);
+        return newConfig;
+    }
+
+    /**
+     * Adapt the properties.
+     * <p>
```
sophokles73, line 126: would be good to know, when this method is being invoked so that subclasses can *do the right thing* :-)

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractKafkaConsumerConfigProperties.java
```diff
@@ -0,0 +1,162 @@
...
+            throw new IllegalArgumentException("poll timeout must not be negative");
+        } else {
+            this.pollTimeout = pollTimeoutMillis;
+        }
+    }
+
+    /**
+     * Gets the timeout for polling records.
+     * <p>
+     * The default value of this property is {@link #DEFAULT_POLL_TIMEOUT}.
```
sophokles73: you can/should use `{@value #DEFAULT_POLL_TIMEOUT}` here in order to get the constant's value instead of a link to it.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/producer/AbstractKafkaProducerConfigProperties.java
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Creates an instance.
+     *
+     * @param keySerializerClass The class to be used for serializing the record keys, if {@code null} the serializer
+     *            needs to be provided in the configuration at runtime.
+     * @param valueSerializerClass The class to be used for serializing the record values, if {@code null} the
+     *            serializer needs to be provided in the configuration at runtime.
+     */
+    @SuppressWarnings("rawtypes")
+    protected AbstractKafkaProducerConfigProperties(final Class<? extends Serializer> keySerializerClass,
```
sophokles73, line 51: `Class<? extends Serializer<?>>` ?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/producer/AbstractKafkaProducerConfigProperties.java
```diff
@@ -0,0 +1,131 @@
...
+        Optional.ofNullable(valueSerializerClass).ifPresent(serializerClass -> overrideConfigProperty(newConfig,
+                ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, serializerClass.getName()));
+
+        setUniqueClientId(newConfig, producerName, ProducerConfig.CLIENT_ID_CONFIG);
+        return newConfig;
+    }
+
+    /**
+     * Adapt the properties.
+     * <p>
```
sophokles73, line 121: see my comment for the consumer

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractKafkaConsumerConfigProperties.java
```diff
@@ -0,0 +1,162 @@
...
+
+    /**
+     * Sets the timeout for polling records.
+     * <p>
+     * The default value of this property is {@link #DEFAULT_POLL_TIMEOUT}.
+     *
+     * @param pollTimeoutMillis The maximum number of milliseconds to wait.
+     * @throws IllegalArgumentException if poll timeout is negative.
+     */
+    public final void setPollTimeout(final long pollTimeoutMillis) {
```
sophokles73: why is this modeled as a dedicated property? isn't this included in the map that contains the consumer configuration properties?

Where is this property being used?

sophokles73 (commented):
I seem to fail to understand why this refactoring is necessary. In particular, you claim that this is a pre-requisite for #2837 but I do not quite see why that is. What am I missing?

b-abel:
> I seem to fail to understand why this refactoring is necessary. In particular, you claim that this is a pre-requisite for #2837 but I do not quite see why that is. What am I missing?

Without this change, we cannot produce records with value type `JSON` in the class `KafkaBasedNotificationSender` because in the current `KafkaConsumerConfigProperties` the value serializer is hardcoded to `BufferSerializer` and cannot be changed to `JsonObjectSerializer`.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractKafkaConsumerConfigProperties.java
b-abel: > why is this modeled as a dedicated property? isn't this included in the map that contains the consumer configuration properties?

This is not a Kafka client config but instead an Hono specific property that was once intended to be used by `HonoKafkaConsumer`. This PR does not touch it but only pulls it u into the new base class.

> Where is this property being used?

@calohmn Do you know why it is no longer in use?

calohmn: This was used in the `AbstractAtLeastOnceKafkaConsumer` class.
At one point, that class got replaced by the `HonoKafkaConsumer` class, which doesn't use the `AbstractKafkaConsumerConfigProperties` though. So this is still a TODO, using this property in the `HonoKafkaConsumer` class.

See also the text in #2904:
> The Command Router Kafka command consumer config needs:
> [...]
> specific properties introduced by the Hono consumer classes, like `HonoKafkaConsumer#POLL_TIMEOUT` and the `AsyncHandlingAutoCommitKafkaConsumer#CONFIG[*]` properties
(see also the Hono `KafkaConsumerConfigProperties` class, currently not used for these classes)

b-abel:
> I seem to fail to understand why this refactoring is necessary. In particular, you claim that this is a pre-requisite for #2837 but I do not quite see why that is. What am I missing?

Maybe it is confusing that here the `KafkaBasedNotificationSender` (and `KafkaBasedNotificationReceiver`) are configured with `MessagingKafkaProducerConfigProperties`. My idea was to create in #2837 new configuration properties classes for notifications and use them in sender and receiver after this PR has been merged. If it is easier to follow, I can also add the commit here.

calohmn:
> Without this change, we cannot produce records with value type `JSON` in the class `KafkaBasedNotificationSender` because in the current ~`KafkaConsumerConfigProperties`~ `KafkaProducerConfigProperties` the value serializer is hardcoded to `BufferSerializer` and cannot be changed to `JsonObjectSerializer`.

So, is it, that the main reason for having a `MessagingKafkaConsumerConfigProperties` class and I guess a `NotificationKafkaConsumerConfigProperties` soon is that we need different Deserializers on these?
I think having one `KafkaConsumerConfigProperties` class instead would be more straightforward.
How about letting the deserializer fields be non-final (with StringDeserializer/BufferDeserializer as default) and letting them be set to the JSON deserializer if needed?

b-abel:
There are many ways to get the desired properties. The most important point in my eyes is:
The configuration of the clients used for messaging is crucial for Hono's core functionality, the messaging APIs. It depends on the configuration of the Kafka clients whether Hono complies with the behavior specified in the APIs or not. I think it is justifiable to give Hono messaging its own small configuration class, where the mandatory configuration is comprehensibly defined and which will not potentially be changed every time someone wants to use a Kafka client somewhere for other purposes.

calohmn:
I could have also imagined a static `MESSAGING_CONFIG_BASE` map with the messaging-related Kafka consumer default properties to be applied on a `KafkaConsumerConfigProperties` instance.
But, you're right, a specific properties class for that makes it more explicit and it is more flexible concerning applying the defaults. So, that's fine with me.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/AbstractKafkaConsumerConfigProperties.java
```diff
@@ -16,31 +16,47 @@
...
  * This class is intended to be as agnostic to the provided properties as possible in order to be forward-compatible
  * with changes in new versions of the Kafka client.
- *
- * @see <a href="https://kafka.apache.org/documentation/#consumerconfigs">Kafka Consumer Configs</a>
- * @see <a href="https://www.eclipse.org/hono/docs/api/telemetry-kafka">Telemetry API for Kafka Specification</a>
- * @see <a href="https://www.eclipse.org/hono/docs/api/event-kafka">Event API for Kafka Specification</a>
- * @see <a href="https://www.eclipse.org/hono/docs/api/command-and-control-kafka/">Command &amp; Control API for Kafka Specification</a>
  */
-public class KafkaConsumerConfigProperties extends AbstractKafkaConfigProperties {
+public abstract class AbstractKafkaConsumerConfigProperties extends AbstractKafkaConfigProperties {
```
calohmn, line 31: I don't quite see why this class needs to be abstract. Couldn't we just keep the existing signature/name?
I think this class should eventually also be used as parameter in the `HonoKafkaConsumer` constructors. And using a non-abstract `KafkaConsumerConfigProperties` class there would be better, IMHO.

b-abel:
Dear reviewers, what is the status here?

calohmn (approved):
LGTM. @sophokles73 Anything to be discussed still?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/producer/MessagingKafkaProducerConfigProperties.java
```diff
@@ -0,0 +1,61 @@
...
+     * <ul>
+     * <li>{@code enable.idempotence=true}: enables idempotent producer behavior</li>
+     * </ul>
+     *
+     * @see <a href="https://kafka.apache.org/documentation/#enable.idempotence">The Kafka documentation -
+     *      "Producer Configs" - enable.idempotence</a>
+     */
+    @Override
+    protected void adaptConfiguration(final Map<String, String> config) {
+        overrideConfigProperty(config, ProducerConfig.ENABLE_IDEMPOTENCE_CONFIG, "true");
```
sophokles73, line 58: My understanding of this class is that it should enforce the serializers and config properties required for successful usage with Hono's Kafka based messaging. based on that I guess that we should make sure that this configuration is always being applied, right? So I guess this method should either be final or we should support passing in some mandatory configuration properties in the superclass' constructor.

b-abel: I made the method final.

sophokles73 (approved):
LGTM

Issue #2910, opened 2021-10-22
Handle failure to create Kafka "command_internal" topic


calohmn:
A failure to create the Kafka `hono.command_internal.[id]` topic on protocol adapter startup is currently not getting handled.
This causes errors in the Command Router when trying to publish on this topic.

There should be retries on such failures and in general, the adapter shouldn't invoke the `Register command consumer for device` operation of the Command Router API if the topic doesn't exist yet.

kaniyan:
This has been closed via  #2930.

PR #2930, opened 2021-11-05
[#2910] Handle internal command topic creation failure

kaniyan:
* The creation of the internal command topic in Kafka is now coupled to the adapter's readiness check. If the creation
  of that topic fails then the adapter is not set to ready.
* If the topic creation fails, then the internal command consumer retries to create that topic.

### DIFF
--- a/clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedInternalCommandConsumer.java
+++ b/clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedInternalCommandConsumer.java
@@ -89,6 +89,7 @@ public class KafkaBasedInternalCommandConsumer implements InternalCommandConsume
     private Context context;
     private KafkaClientMetricsSupport metricsSupport;
     private final Vertx vertx;
+    private long retryCreateTopicTimerId;
 
     /**
      * Creates a consumer.
@@ -242,7 +243,7 @@ public class KafkaBasedInternalCommandConsumer implements InternalCommandConsume
     private Future<Void> retryCreateTopic() {
         final Promise<Void> createTopicRetryPromise = Promise.promise();
         // Retry at specified interval until the internal command topic is successfully created
-        vertx.setPeriodic(CREATE_TOPIC_RETRY_INTERVAL, id -> {
+        retryCreateTopicTimerId = vertx.setPeriodic(CREATE_TOPIC_RETRY_INTERVAL, id -> {
             if (retryCreateTopic.compareAndSet(true, false)) {
                 createTopic()
                         .onSuccess(ok -> {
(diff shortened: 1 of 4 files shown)


clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedInternalCommandConsumer.java
```diff
@@ -196,13 +229,31 @@ public final KafkaBasedInternalCommandConsumer setMetricsSupport(final KafkaClie
...
                 .onSuccess(v -> LOG.debug("created topic [{}]", topicName))
                 .onFailure(thr -> LOG.error("error creating topic [{}]", topicName, thr));
     }
 
+    private Future<Void> createTopicWithRetries() {
+        final Promise<Void> createTopicRetryPromise = Promise.promise();
+        // Retry at specified interval until the internal command topic is successfully created
+        vertx.setPeriodic(CREATE_TOPIC_RETRY_INTERVAL, id -> {
+            if (retryCreateTopic.compareAndSet(true, false)) {
+                createTopic()
```
calohmn, line 246: This means, there is always a one second delay before the first attempt to create the topic is made.
It would be better if the first attempt is made directly.

clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedInternalCommandConsumer.java
```diff
@@ -196,13 +229,31 @@ public final KafkaBasedInternalCommandConsumer setMetricsSupport(final KafkaClie
...
+                    context.runOnContext(v1 -> Optional.ofNullable(ex)
+                            .filter(e -> !(e instanceof TopicExistsException))
+                            .ifPresentOrElse(promise::fail, promise::complete));
                 });
         return promise.future()
                 .onSuccess(v -> LOG.debug("created topic [{}]", topicName))
                 .onFailure(thr -> LOG.error("error creating topic [{}]", topicName, thr));
     }
 
+    private Future<Void> createTopicWithRetries() {
```
calohmn, line 241: Retries should be stopped when `stop()` has been called.
Also there's no need to try to delete the topic in `stop()` if `isTopicCreated` is `false`.

clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedInternalCommandConsumer.java
```diff
@@ -265,10 +302,33 @@ private String getTopicName() {
...
+                            LOG.warn("error deleting topic [{}]", topicName, ex);
+                        }
+                        context.executeBlocking(future -> {
+                            adminClient.close();
+                            future.complete();
+                        }, adminClientClosePromise);
+                    });
+            return adminClientClosePromise.future().onComplete(ar -> LOG.debug("admin client closed"));
+        } else {
+            return Future.succeededFuture();
```
calohmn, line 328: In this case the admin client should also be closed.
Closing the admin client could also be moved out of this method entirely, then the `deleteTopicIfExists` method name would also fit better.

clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedInternalCommandConsumer.java
calohmn: > Retries should be stopped when `stop()` has been called.

How about setting `retryCreateTopic.set(false)` in `stop()` to ensure this?

clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedInternalCommandConsumer.java
```diff
@@ -239,34 +292,57 @@ private String getTopicName() {
...
-                        future.complete();
-                    }, adminClientClosePromise);
-                });
-        adminClientClosePromise.future().onComplete(ar -> LOG.debug("admin client closed"));
+        LOG.debug("stop: close admin client");
+        context.runOnContext(x -> {
+            adminClient.close();
+            LOG.debug("admin client closed");
+            adminClientClosePromise.complete();
+        });
```
calohmn, line 333: I think this should rather be invoked on a worker thread (like before):
```
                    context.executeBlocking(future -> {
                        adminClient.close();
                        future.complete();
                    }, adminClientClosePromise);
```

calohmn (approved):
LGTM

PR #2931, opened 2021-11-08
[#2925] Use "earliest" offset reset strategy for Kafka command consumer

### LINKED ISSUE #2925: Use "earliest" offset reset strategy for Kafka command consumer
The Kafka consumer for receiving command messages in the Command Router currently uses the "latest" [offset reset strategy](https://kafka.apache.org/documentation/#consumerconfigs_auto.offset.reset).

This means that when the Command Router starts, it will read from the latest offsets of the command topics if there are no committed offsets. 
The case of there being no committed offsets should usually only occur for newly created topics, where "latest" and "earliest" makes no difference. The default for the offsets retention time (`offsets.retention.minutes`, concerning deletion of committed offsets) is a week, which should be much more than the retention time one would configure for the command topic records (`log.retention.hours` default is also a week).

Therefore it should be no problem to switch to using "earliest" as default.

This would have the benefit that our start/rebalance handling in the consumer gets much easier.
All the logic to ensure that "latest" doesn't get applied when we don't want it to isn't needed then anymore.

calohmn:
This resolves #2925.

Along with this change, the `HonoKafkaConsumer#ensureTopicIsAmongSubscribedTopicPatternTopics` method could be changed so that the result Future is completed earlier (without waiting for the consumer rebalance). But this changes the timing in the integration tests (first command subscription finishes sooner, but first command takes longer to be received), so that adaptations are needed there. This will go in a separate PR then.

### DIFF
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/HonoKafkaConsumer.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/consumer/HonoKafkaConsumer.java
@@ -842,17 +842,18 @@ public class HonoKafkaConsumer implements Lifecycle {
      * Tries to ensure that the given topic is part of the list of topics this consumer is subscribed to.
      * <p>
      * The successful completion of the returned Future means that the topic exists and is among the subscribed topics.
-     * It also means that if partitions of this topic have been assigned to this consumer, the positions for these
-     * partitions have already been fetched and the consumer will receive records published thereafter.
+     * It also means that if partitions of this topic have been assigned to this consumer and if the offset reset config
+     * is set to "latest", the positions for these partitions have already been fetched and the consumer will receive
+     * records published thereafter.
      * If all topic partition are assigned to other consumers, a best-effort approach is taken to let the result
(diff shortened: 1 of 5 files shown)


calohmn:
Increased integration test timeouts because command consumer registration may now succeed before partition offsets have been applied. This means the test setup with the consumer registration is finished sooner, but receiving the command takes longer.

calohmn:
@b-abel Any comments on switching to "earliest" here?

b-abel:
> @b-abel Any comments on switching to "earliest" here?

This makes sense FMPOV if you apply a time-based retention policy to make sure that only commands are consumed that are still "valid" (to be defined from a business perspective). The default of 7 days is probably much too long. Maybe you can document this reasoning somewhere? A starting point could be the admin guide of the command router. But maybe we should add a documentation page dedicated to Kafka configurations where we collect all recommended configurations for the brokers as well as the clients.

calohmn:
> But maybe we should add a documentation page dedicated to Kafka configurations where we collect all recommended configurations for the brokers as well as the clients.

Yes, I think that would be good. I've added #2936 for that.

Issue #2825, opened 2021-08-12
Possibility to provide auth-ID pattern as tenant configuration


alinaserg:
With introduction of auto-registration feature there's possibility to configure device-id pattern, which is used during device registration process, see: [#2664](https://github.com/eclipse/hono/issues/2664).

I think it would be also helpful to provide similar configuration for auth-id which is created during device registration process and used for device authentication. Currently auth-id for certificate based device authentication completely corresponds to subject-DN of device certificate.

I'll try to explain it with the help of one example:

The client certificates of devices in the field can be renewed and updated. The renewed certificates can contain some changes in subject-DN, e.g. the department name is changed. In such case subject-DN of the client certificate will not be similar to the subject-DN from the old certificate. And when such device will try to connect with a new certificate the authentication will fail as auth-id (subject-DN) registered in Device Registry will not match with auth-id (subject-DN) from the new client certificate.

In such case it would be helpful to provide possibility to configure which parts from Subject-DN should be used for auth-id, which are not affected and changed by certificate renewal and stay stable and unique during the whole lifecycle of the device.

See below examples of possible auth-id patterns:

`auth-id-pattern: "{{subject-CN}}"` -> only use the subject CN from the device certificate

`auth-id-pattern: "{{subject-DN}}" `-> use the subject DN from the device certificate

WDYT?

b-abel:
At least in the example, it's more about selecting certain attributes from the subject to be used in the authentication. I wonder if it would be better to specify a list of attributes rather than a pattern. That would be easier to use because not all possible attributes and their order would have to be considered. @alinaserg @sophokles73 WDYT?

alinaserg:
@b-abel, for me your proposal sounds reasonable. I don't see anything that speaks against of this approach.

b-abel:
So, I summarize what we want here:
1. Define a list of placeholders and document them in the Management API. We start with the following:
  - {{subject-CN}}
  - {{subject-DN}}
  - {{subject-OU}}
2. The auth-id pattern is resolved with the values from the device cert during the (auto-)provisioning process.
3. During the authentication of a device, the pattern (the Management API calls it *template*, I think we should keep that) is taken from the tenant stored in the database, the placeholders are replaced with the values from the certificate of the device that tries to authenticate and then checked if the result matches the auth-id from the stored credentials.

Constraints that need to be documented:
- when a new certificate is issued for a device, the values for the used placeholders must not be the same that are used during the provisioning.
[...]

PR #2938, opened 2021-11-10
[#2825] Extend device id template to support more attributes

kaniyan:
The auto-provisioning _device-id_ template now supports one more subjectDN's attribute namely _Organizational Unit Name (OU)_. This PR also includes preparation for _auth-id_ template implementation #2825.

EDIT:
It also now supports attribute _Organization Name (O)_.


services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/TemplateHelper.java
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * A utility class for handling template used for generating device identifier during auto-provisioning.
+ */
+public final class TemplateHelper {
+
+    private static final String QUOTED_PLACEHOLDER_SUBJECT_DN = Pattern
+            .quote(RegistryManagementConstants.PLACEHOLDER_SUBJECT_DN);
+
+    private TemplateHelper() {
```
sophokles73, line 34: wouldn't it make sense to turn this into an `IdentityTemplate` class and create instances of it using the template string in the constructor? The *fill* method would then only accept a DN as the argument, right? It might be worth considering to rename the *fill* method to something like *apply* as we apply the DN to the template, producing the identity.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/TemplateHelper.java
```diff
@@ -0,0 +1,145 @@
...
+            result = fillAttribute(Attribute.OU, result, rdns);
+            return fillAttribute(Attribute.O, result, rdns);
+        } catch (InvalidNameException e) {
+            throw new IllegalArgumentException(String.format("subject DN [%s] is not valid", subjectDN));
+        }
+    }
+
+    private static String fillAttribute(final Attribute attribute, final String template,
+            final List<Rdn> rdns) {
+        if (template.contains(attribute.getPlaceHolder())) {
```
sophokles73, line 133: FMPOV it doesn't make much sense to first scan the template string for occurrence of the placeholder as that is what the String.replaceAll method already does as well, doesn't it?

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -885,11 +885,12 @@ components:
...
                   Specifies the template to use by the device registry to generate the `device-id` during 
-                  auto-provisioning of devices and gateways. The placeholders `{{subject-dn}}` and `{{subject-cn}}` 
-                  are currently supported. The template MUST contain at least one of these placeholders. The `device-id`
-                  is generated by the device registry by replacing any placeholders in the template string with their
-                  corresponding *Subject DN* or *Common Name* from the client certificate used for authentication by 
-                  unregistered devices and gateways. The *Subject DN* is formatted according to the rules defined in 
+                  auto-provisioning of devices and gateways. The placeholders `{{subject-cn}}`, `{{subject-dn}}`,
+                  `{{subject-ou}}` and `{{subject-o}}` are currently supported. The template MUST contain at least
+                  one of these placeholders. The `device-id` is generated by replacing any placeholders in the template
+                  string with their corresponding *Common Name (CN)*, *Subject DN (DN)*, *Organizational Unit Name (OU)*
```
sophokles73, line 891: IMHO we should make it more explicit, that the CN, OU and O are actually taken from the certificate's subject DN (instead of the issuer DN)

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/TemplateHelper.java
kaniyan: This check is there to handle the scenario when placeholder for an attribute is configured in the template but there is no corresponding attribute/value provided in the client certificate. In this case an `IllegalArgumentException` is thrown with an appropriate error reason.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/TemplateHelper.java
kaniyan: It makes sense. Since we name our classes normally using a noun, how about `TemplateIdentifer`?

kaniyan: I have pushed a commit with respective changes. Would you mind taking a look again?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/IdentityTemplate.java
```diff
@@ -0,0 +1,152 @@
...
+            }
+            return result;
+        } catch (final InvalidNameException e) {
+            throw new IllegalArgumentException(String.format("subject DN [%s] is not valid", subjectDN));
+        }
+    }
+
+    private static String applyAttribute(final Attribute attribute, final String template,
+            final List<Rdn> rdns) {
+        if (template.contains(attribute.getPlaceHolder())) {
```
sophokles73, line 140: the content of this conditional block is fine. My point is that we can simply run it unconditionally because the `template.replaceAll()` method will perform the same check, i.e. it will only replace occurrences of the attribute's corresponding placeholder if the placeholder is actually contained in the template, won't it?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/IdentityTemplate.java
```diff
@@ -0,0 +1,152 @@
...
+     * @throws NullPointerException if any of the parameters are {@code null}.
+     */
+    public String apply(final String subjectDN) {
+        Objects.requireNonNull(subjectDN, "subjectDN must not be null");
+
+        try {
+            final List<Rdn> rdns = new LdapName(subjectDN).getRdns();
+            String result = template.replaceAll(QUOTED_PLACEHOLDER_SUBJECT_DN, subjectDN);
+            for (final Attribute attribute : Attribute.values()) {
+                result = applyAttribute(attribute, result, rdns);
```
sophokles73, line 130: :+1:

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/IdentityTemplate.java
kaniyan: The `IllegalArgumentException` is to be thrown only when a placeholder for an attribute is configured in the template and also that attribute is not found in the subejct DN. If we remove that conditional statement, then an `IllegalArgumentException` will be thrown whenever an attribute is not found in _SubjectDN_ irrespective of  the corresponding placeholder is configured in the template or not. Hence we need that check. Sorry, I didn't get it how it can be achieved only with `template.replaceAll()`.

sophokles73: right, my bad

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -889,9 +889,8 @@ components:
                   `{{subject-ou}}` and `{{subject-o}}` are currently supported. The template MUST contain at least
                   one of these placeholders. The `device-id` is generated by replacing any placeholders in the template
                   string with their corresponding *Common Name (CN)*, *Subject DN (DN)*, *Organizational Unit Name (OU)*
-                  and *Organization Name (O)* from the client certificate used for authentication by unregistered
-                  devices and gateways. The *Subject DN* is formatted according to the rules defined in
-                  [RFC 2253](https://tools.ietf.org/html/rfc2253#section-2).
+                  and *Organization Name (O)* from the client certificate's subject DN. The *Subject DN* is formatted
```
sophokles73, line 892: the subject DN from the certificate's subject DN?

kaniyan: I have corrected it now and pushed a commit. Would you mind taking another look?

sophokles73 (approved):
LGTM

PR #2940, opened 2021-11-10
Let ensureTopicIsAmongSubscribedTopics() finish sooner

calohmn:
Relates to #2925:
If the Kafka consumer partition offset reset config is set to "earliest", the method to ensure that the consumer with a wildcard topic subscription knows about a given topic now may return earlier, not waiting for a rebalance.

This has the effect, that the first "createCommandConsumer" invocation for a tenant is done sooner, preventing possible timeouts there. On the other hand, if a command gets sent immediately after the "createCommandConsumer"/command subscription result, the time it takes for that command to be received is longer.


kaniyan (approved):
LGTM

Issue #2942, opened 2021-11-11
ConfigBasedCoapEndpointFactory.loadNetworkConfig: returned Future succeeded too early


boaks:
[ConfigBasedCoapEndpointFactory.loadNetworkConfig](https://github.com/eclipse/hono/blob/master/adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/impl/ConfigBasedCoapEndpointFactory.java#L164-L187) returns a `Future.succeededFuture(networkConfig);`, but the properties file will be loaded later.

PR #2943, opened 2021-11-11
Fix issue #2942.

boaks:
Postpone completion until file is read.


adapters/coap-vertx-base/src/main/java/org/eclipse/hono/adapter/coap/impl/ConfigBasedCoapEndpointFactory.java
```diff
@@ -181,9 +181,12 @@ private NetworkConfig newDefaultNetworkConfig() {
                 } else {
                     LOG.warn("error reading NetworkConfig file [{}]", fileName, readAttempt.cause());
                 }
+                result.complete(networkConfig);
```
sophokles73, line 184: FMPOV we should either complete or fail the result depending on the outcome of trying to read the file.

boaks: I considered to implement the behavior close to the behavior before this fix.
But failing will also be a possibility, if the resulting behavior change is acceptable.

sophokles73: Well, if I would explicitly configure a config file to be read and used then I would expect the adapter to fail to start, if the file could not be read, because otherwise I would get the impression that it is now using the configuration that I put into the file when it actually doesn't. So FMPOV we should fail the start up if the file cannot be read. Does that make sense to you as well?

boaks: Sure, that makes sense. My concern was more, if such a behavior change would be acceptable in one commit. It seems so. So I will adapt the PR.

boaks: done :-).

sophokles73: can you also adapt the CoAP adapter admin guide to mention that start up will fail if the configured config file cannot be read/found?

boaks: Yes. Working on migration to Californium 3.0, I also found, that the warning "skipping malformed NetworkConfig properties" is now also misleading.

boaks: Done.

sophokles73 (approved):
LGTM

Issue #2888, opened 2021-10-12
Improve timeout handling of cases where messages can't get sent via KafkaProducer


calohmn:
When using an AMQP messaging network, sending an event message will lead to a `SendMessageTimeoutException` after the configured sendMessageTimeout (default 1s) has elapsed without a delivery update having been received. 

When using Kafka, the behaviour is currently as follows:
For QoS 1 telemetry and event messages, the send operation in the protocol adapter will fail after 2min (i.e. `delivery.timeout.ms` producer config value) with:

> org.apache.kafka.common.errors.TimeoutException: Expiring 3 record(s) for hono.telemetry.my-tenant-0:120001 ms has passed since batch creation

The default timeout handling when using Kafka should be adapted to use either the same timeout as with AMQP or at least a much lower default value.

PR #2948, opened 2021-11-16
[#2919] Set overall timeout for Command Router command processing

### LINKED ISSUE #2919: Use timeout for command processing in Command Router
When the Command Router receives and processes a command message, there are Infinispan requests being done (concerning the `device-connection` cache). Currently there is no timeout getting applied on this.
That means when Infinispan requests take a long time, the whole command processing is getting impacted, in the end triggering the mechanism that limits the number of command records processed in parallel and pausing the Kafka consumer then (if Kafka messaging is used).

There should be a configurable timeout that limits the command processing time.

calohmn:
This fixes #2919.

### DIFF
--- a/core/src/main/java/org/eclipse/hono/util/Futures.java
+++ b/core/src/main/java/org/eclipse/hono/util/Futures.java
@@ -13,6 +13,7 @@
 
 package org.eclipse.hono.util;
 
+import java.util.Objects;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Supplier;
 
@@ -40,8 +41,10 @@ public final class Futures {
      * @param <T> The result data type.
      * @param futureSupplier The supplier of the future.
      * @return A vert.x future, never returns {@code null}.
+     * @throws NullPointerException if futureSupplier is {@code null}.
      */
     public static <T> Future<T> create(final Supplier<CompletionStage<T>> futureSupplier) {
+        Objects.requireNonNull(futureSupplier);
 
         final Context originalContext = Vertx.currentContext();
         final CompletionStage<T> future;
@@ -89,8 +92,12 @@ public final class Futures {
      * @param vertx The vertx context.
      * @param blocking The blocking code.
      * @return The future, reporting the result.
+     * @throws NullPointerException if any of the parameters is {@code null}.
      */
     public static <T> Future<T> executeBlocking(final Vertx vertx, final BlockingCode<T> blocking) {
+        Objects.requireNonNull(vertx);
+        Objects.requireNonNull(blocking);
+
         final Promise<T> result = Promise.promise();
 
         vertx.executeBlocking(promise -> {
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/eclipse/hono/util/Futures.java
```diff
@@ -104,4 +104,22 @@ private Futures() {
...
+     * Applies the given result on the given promise using {@link Promise#tryComplete(Object)} or
+     * {@link Promise#tryFail(Throwable)}.
+     * <p>
+     * Similar to {@link Promise#handle(AsyncResult)} but does nothing if the promise is already complete.
+     *
+     * @param promise The promise to complete or fail.
+     * @param asyncResult The result to apply.
+     * @param <T> The promise and result type.
+     */
+    public static <T> void tryHandleResult(final Promise<T> promise, final AsyncResult<T> asyncResult) {
```
sophokles73, line 117: `null` handling?

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/KafkaBasedMappingAndDelegatingCommandHandler.java
```diff
@@ -42,13 +49,25 @@
...
 
+    /**
+     * Duration after which record processing is cancelled.
+     * <p>
+     * Usually, the timeout values configured for the service clients involved in command processing should define the
+     * overall time limit. The timeout mechanism associated with this value here serves as a safeguard against
+     * exceptional cases where the other timeouts didn't get applied. It thereby prevents command processing from
+     * getting stuck, so that the command consumer doesn't get blocked.
+     */
+    private static final Duration PROCESSING_TIMEOUT = Duration.ofMinutes(1);
```
sophokles73, line 60: one minute? what's the reasoning behind such a long timeout? I would have expected something in the single digit second range ...

calohmn: The value here shall serve as an upper boundary of configured timeout values related to command processing. Therefore I've chosen a high value. (And in the Kafka world, timeouts in general seem to be quite high by default, e.g. producer.send() with 2min timeout. So, that was an influence here.)

But right, this may be unnecessarily high (also in connection with #2888, reducing producer default timeouts).
How about 8s?

sophokles73: Sounds a lot better to me :-) However, the issue indicates that this value should be configurable or am I mistaken?

calohmn: It would be best to have the value configurable, yes. However, I would wait for #2904 here.
In any case, configuration of the other timeouts (device registry, infinispan clients) is more important. The timeout above should only trigger in exceptional cases (like the Hot Rot client NPE mentioned in #2919).

sophokles73: ok

sophokles73 (approved):
LGTM

PR #2951, opened 2021-11-17
Check validity of identity templates

kaniyan:


services/device-registry-base/src/main/java/org/eclipse/hono/service/management/device/IdentityTemplate.java
```diff
@@ -135,6 +138,28 @@ public String apply(final String subjectDN) {
...
+     * @throws NullPointerException if template is {@code null}.
+     * @throws IllegalArgumentException if the template does not contain any placeholders.
+     */
+    public static void checkValidity(final String template) {
+        if (!template.contains(RegistryManagementConstants.PLACEHOLDER_SUBJECT_DN)) {
+            Arrays.stream(Attribute.values())
+                    .map(Attribute::getPlaceHolder)
+                    .filter(template::contains)
+                    .findFirst()
+                    .orElseThrow(() -> new IllegalArgumentException("template must contain at least one placeholder"));
```
sophokles73, line 154: why does it need to contain at least one placeholder? However, we should probably be able to detect a case where an unsupported placeholder is being used, right?

kaniyan: As per the spec,
> The template MUST contain at least one of these placeholders.

We already do this check inside `TrustedCertificateAuthority`. I have refactored it now to check for attributes dynamically.

sophokles73: What about checking for unsupported placeholders?

kaniyan: I have added a check for unsupported placeholders. Would you mind taking another look?

sophokles73 (approved):
LGTM

PR #2954, opened 2021-11-18
[#2901] Let command_internal topics be deleted in the Command Router

### LINKED ISSUE #2901: Command Router: Errors/Delays when internal command topic got deleted (2)
This is a follow-up of #2773:
After a protocol adapter pod has shut down and deleted its internal command topics (`hono.command_internal.[adapterInstanceId]`), the Command Router might still try to forward command messages on these topics, resulting in potentially long delays:
```
13:30:03.532 [vert.x-eventloop-thread-2] DEBUG o.e.h.c.c.k.KafkaBasedInternalCommandSender - sending message failed [topic: hono.command_internal.hono-adapter-mqtt-vertx-69f65b6b4b-ch7jh_595e80c45077_0, key: myDevice, tenantId: myTenant, deviceId: myDevice]
org.apache.kafka.common.errors.TimeoutException: Topic hono.command_internal.hono-adapter-mqtt-vertx-69f65b6b4b-ch7jh_595e80c45077_0 not present in metadata after 60000 ms.
```

With the introduction of the AdapterInstancesLivenessService (#2028), the command router already skips trying to forward commands to "dead" adapters, but it has revealed itself, that the above error might still occur from time to time (e.g. if the KafkaBasedInternalCommandSender takes longer or retries sending the messages).

Therefore I think we cannot safely let the protocol adapter delete its internal command topics on shutdown.
Deletion must happen with a delay of some minutes.

An idea here is to let the Command Router do the deletion, using the AdapterInstancesLivenessService to find the topics to be deleted.

calohmn:
This fixes #2901:
The `hono.command_internal.[adapterInstanceId]` Kafka topics associated with dead adapter instances are now deleted with some delay in the Command Router component.
This relies on the `KubernetesBasedAdapterInstanceStatusService` (as the only `AdapterInstanceStatusService` implementation right now) to recognize dead adapter instances.

### DIFF
--- a/services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/InternalKafkaTopicCleanupService.java
+++ b/services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/InternalKafkaTopicCleanupService.java
@@ -1,0 +1,183 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.hono.commandrouter.impl.kafka;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Supplier;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+import org.eclipse.hono.client.kafka.HonoTopic;
+import org.eclipse.hono.client.kafka.KafkaAdminClientConfigProperties;
+import org.eclipse.hono.commandrouter.AdapterInstanceStatusService;
+import org.eclipse.hono.util.Lifecycle;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+import io.vertx.core.Vertx;
+import io.vertx.kafka.admin.KafkaAdminClient;
+
+/**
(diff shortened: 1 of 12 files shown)


kaniyan:
@calohmn In this PR, internal command topics are being deleted within the scope of Kafka based command router. But those topics are created by adapters. Wouldn't it be more appropriate that adapters take care of this clean up service as they create those topics? Also adapters already have access using admin client. Is there any reason that this clean up is being done as part of the command router?

calohmn:
> Wouldn't it be more appropriate that adapters take care of this clean up service as they create those topics?

There are usually more adapter pods than Command Router pods. Since we currently have no mechanism to choose **_one_** pod to run the `InternalKafkaTopicCleanupService`, it must be run by all pods of a specific kind. Therefore having the `InternalKafkaTopicCleanupService` be run by the Command Router pods (from which there may even be just one pod) involves less overhead than having each adapter run the `InternalKafkaTopicCleanupService`.

And, the adapters are the ones to create the internal topic, yes. But the command routers are the ones to publish to that topic. So they might as well delete it once there is nothing to publish to that topic any more.

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/InternalKafkaTopicCleanupService.java
```diff
@@ -0,0 +1,183 @@
...
+import io.vertx.kafka.admin.KafkaAdminClient;
+
+/**
+ * A service to delete obsolete {@link HonoTopic.Type#COMMAND_INTERNAL} topics.
+ */
+public class InternalKafkaTopicCleanupService implements Lifecycle {
+
+    private static final Logger LOG = LoggerFactory.getLogger(InternalKafkaTopicCleanupService.class);
+
+    private static final long CHECK_INTERVAL_MILLIS = TimeUnit.MINUTES.toMillis(5);
```
kaniyan, line 48: FMPOV we can set this to even higher value. In real time, I think not so often those internal topics become obsolete and should to be deleted too. We can avoid making the Kubernetes call and also Kafka call to list topics every 5 minutes by all the command router instances. IMHO later we can consider making this value configurable, so that it can help optimizing test/performance-test setups.

kaniyan (changes requested):

services/command-router-base/src/main/java/org/eclipse/hono/commandrouter/impl/kafka/InternalKafkaTopicCleanupService.java
calohmn: I've now set the interval to 10min, so that obsolete topics will be deleted between 10 and 20 min after the corresponding adapter pods got shutdown. I think the K8s API requests won't matter much with these kinds of intervals.

kaniyan (approved):
LGTM

PR #2956, opened 2021-11-18
Run ITs with preferred profiles

sophokles73:
The integration test profiles to be run on GitHub Actions have been
modified to better reflect the preferred setup. The deprecated Device
Connection service has been removed from the three profiles altogether
in order to reuse the slot for an additional setup with the Command
Router.


calohmn (approved):
LGTM

PR #2958, opened 2021-11-18
[#2825] Introduce auth-id template for authentication and auto-provisioning

kaniyan:
This PR introduces a new tenant's trusted CA property namely `auth-id-template`.  When devices authenticate using client certificates, the adapters now generate `auth-id` based on this property and use the generated `auth-id` for authentication and also for auto-provisioning.

I have created this PR even though the below items are in progress, so that the concept can be reviewed earlier.

-  Currently `X509CertificateCredential` assumes that  _auth-id_ is always _SubjectDN_ and to be adapted.
- Integration tests
- Documentation


sophokles73:
@kaniyan can you rebase?

kaniyan:
> @kaniyan can you rebase?

Done.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/tenant/TrustedCertificateAuthority.java
```diff
@@ -339,4 +344,30 @@ public final TrustedCertificateAuthority setAutoProvisioningDeviceIdTemplate(fin
         this.autoProvisioningDeviceIdTemplate = template;
         return this;
     }
+
+    /**
+     * Gets the template for generating the authentication identifier.
+     *
+     * @return the template for generating the authentication identifier.
+     */
+    public String getAuthIdTemplate() {
```
sophokles73, line 353: final ?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/tenant/TrustedCertificateAuthority.java
```diff
@@ -339,4 +344,30 @@ public final TrustedCertificateAuthority setAutoProvisioningDeviceIdTemplate(fin
...
+     * Sets the template for generating the authentication identifier.
+     * <p>
+     * The validity of the template is verified using {@link IdentityTemplate#checkValidity(String)}.
+     *
+     * @param template the template for generating the authentication identifier.
+     * @return A reference to this for fluent use.
+     * @throws NullPointerException if the template is {@code null}.
+     * @throws IllegalArgumentException if the template is not valid.
+     */
+    public TrustedCertificateAuthority setAuthIdTemplate(final String template) {
```
sophokles73, line 367: I assume you expect this method to be invoked by Jackson when unmarshalling a JSON request body, right?
I am not sure if this is actually the case because you have annotated the authIdTemplate field with `@JonsProperty` instead of this setter. You should validate by means of adding a corresponding test case to `TenantTest`

kaniyan: As per my understanding, even though a field is annotated, the getter and setter method  (following the correct naming convention) are used for marshalling and unmarshalling of JSON by Jackson. I have also added a corresponding test in `TenantTest`.

adapter-base/src/main/java/org/eclipse/hono/adapter/auth/device/X509AuthProvider.java
```diff
@@ -52,6 +52,9 @@ public X509AuthProvider(final CredentialsClient credentialsClient, final Tracer
      * The JSON object passed in is required to contain a <em>subject-dn</em> and a
      * <em>tenant-id</em> property.
      * <p>
+     * The JSON property <em>auth-id-template</em> is available only if an auth-id template
+     * is configured in the tenant's trust anchor for the given subject DN.
+     * <p>
```
sophokles73, line 57: IMHO this should be modified to be more instructive to the reader, e.g.
> If the JSON object does not contain an <em>auth-id-template</em> property, the value of the <em>subject-dn</em> property is used as the authentication identifier of the returned credentials object. Otherwise, ...

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -1379,9 +1381,12 @@ components:
...
                        meta data in the *auth-id* and *secrets* explicitly.
                        When provided in a request body, implementors of this API MUST extract the subject DN
-                       from the certificate's public key and store its value properly formatted to the *auth-id*
-                       property. The certificate's validity period MUST be stored in a single secret.
-                       The certificate itself must then be discarded.
+                       from the certificate's public key. The *subject DN* MUST be formatted according to the
+                       rules defined in [RFC 2253](https://tools.ietf.org/html/rfc2253#section-2).
+                       The extracted *subject DN* is used as the *auth-id* if there is no *auth-id-template*
+                       configured in the tenant's trusted anchor. Otherwise, the *auth-id* MUST be generated
+                       using the *auth-id-template* and the extracted *subject DN*. The certificate's validity
```
sophokles73, line 1388: by applying the template to the extracted subject ?

kaniyan:
In the last commit, I have updated the _Credentials Management API_ spec in order to support the `auth-id` template concept. @sophokles73 Would you mind taking a look? I would like to finalize the spec before starting to implement.

site/documentation/content/api/management/device-registry-v1.yaml
```diff
@@ -1359,9 +1359,11 @@ components:
                  "auth-id":
                     type: string
                     description: |
-                       The *subject DN* of the public key contained in the device's X.509 certificate.
-                       The value MUST be formatted according to the rules defined in
-                       [RFC 2253](https://tools.ietf.org/html/rfc2253#section-2).
+                       The *subject DN* of the public key contained in the device's X.509 certificate
+                       if there is no *auth-id-template* configured in the tenant's trusted anchor.
+                       Else privide an authentifiation identifier that MUST be in compliance with the
```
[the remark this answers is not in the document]
kaniyan: I already thought of an alternative which is to leave the *subject DN* as it is. And add a note explaining that the `auth-id` will be generated  by applying the template to the provided *subject DN* if a corresponding template is configured. With this option there is no extra effort from user side to compute the _auth-id_. But I don't like this option as it is not intuitive.

Another option would be adding a new property _subject-dn_ and deprecating the existing property _auth-id_. The *auth-id* will be always computed by the implementer (either generated or the *subject DN*).  Until we remove *auth-id* property, we can add a note encouraging to use the property *subject DN* hereafter especially when a template is configured.  In the current spec, we provide a `subject-DN` to an `auth-id` property which is not very intuitive. This approach solves that and I am in favor of this option than the first. 

@sophokles73 WDYT? Any other ideas?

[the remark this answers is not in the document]
sophokles73: FMPOV the first option is totally fine. In order to not confuse people, we might want to add a comment to the secret when the registry applies the template to the auth-id (subject DN) and overwrites it.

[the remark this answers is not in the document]
kaniyan: I updated the spec with the first option.

sophokles73 (approved):
LGTM

Don't forget to squash before merging!

PR #2965, opened 2021-11-22
[#2945] Add tenant_id header for Kafka-based command response

### LINKED ISSUE #2945: Kafka based Command response does not contain tenant ID
The Command & Control API contains a definition of the response message format which includes an application property called `tenant_id` which contains the identifier of the tenant that the device belongs to.
The response message format defined by the Kafka based variant of the C&C API does not include a corresponding header which may lead to incompatibilities in downstream applications which expect to find it.

calohmn:
This fixes #2945.


sophokles73 (approved):
LGTM

@calohmn can you also please make sure to apply this to the 1.10.x branch?

calohmn:
yes, it's done.

PR #2966, opened 2021-11-23
Don't let stop() fail if cache already stopped

calohmn:
Correction regarding #2947:
If `stop()` was already called (e.g. from another verticle), the result Future shouldn't be failed.
Otherwise the whole verticle undeployment would be logged as failed:
```
ERROR  io.vertx.core.impl.DeploymentManager  Undeploy failed
io.vertx.core.impl.NoStackTraceThrowable: stop already called
```


sophokles73 (approved):
LGTM

PR #2962, opened 2021-11-19
Pick up Kafka client ID prefix configuration property

sophokles73:
The Quarkus based components did not support the "defaultClientIdPrefix"
configuration property of the Kafka client.

Fixes #2960

Issue #2906, opened 2021-10-20
Use Kafka client identifiers including pod/host name


calohmn:
Currently, the client identifiers used for the Kafka clients in the Hono components look like this:
`hono-command-router-consumer-consumer-185d8412-bd66-4bd8-ba89-4f978dbb6f50`,
`hono-amqp-adapter-consumer-internal-cmd-012f5312-1cae-40c0-b3ff-257ec7d22957`,
`hono-amqp-adapter-producer-event-0c503717-cdde-4120-b9bf-a55fcec5d9fa`.

When showing Kafka client related metrics, it would be helpful to make it clear in which component instance the client is running.
This, however, produces a very long id:
`hono-amqp-adapter-producer-event-0c503717-cdde-4120-b9bf-a55fcec5d9fa - hono-adapter-amqp-vertx-78b999b4b5-n8lvf` 
("client - host").

When Hono components are running in Kubernetes, the client identifier format could be adapted to include the pod name (and the (docker) container identifier and a counter to make the id unique).
This could for example be:
`[client-type]_[pod name]_[container id]_[counter]`, e.g.
`hono-event-producer_hono-adapter-amqp-vertx-78b999b4b5-n8lvf_196e4cf69446_5`.

PR #2969, opened 2021-11-24
[#2967] Remove support for Kafka client ID prefix configuration property

### LINKED ISSUE #2967: Remove support for configuring `defaultClientIdPrefix` on Kafka clients
The discussion in #2960 revealed that we do not want to provide a configuration property defaultClientIdPrefix for Kafka clients. The changes in the PR #2962 that added support for this property should therefore be reverted (other changes in the PR should not).

b-abel:
This fixes #2967. 
It reverts some changes of PR #2962. Property defaultClientIdPrefix is not intended to be changed by configuration.

### DIFF
--- a/adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
+++ b/adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
@@ -222,22 +222,18 @@ public abstract class AbstractProtocolAdapterApplication<C extends ProtocolAdapt
 
     @Inject
     void setKafkaClientOptions(final KafkaClientOptions options) {
-
-        final Optional<String> clientIdPrefix = Optional.ofNullable(
-                options.defaultClientIdPrefix().orElse(getComponentName()));
-
         this.kafkaProducerConfig = new MessagingKafkaProducerConfigProperties();
-        clientIdPrefix.ifPresent(this.kafkaProducerConfig::setDefaultClientIdPrefix);
+        Optional.ofNullable(getComponentName()).ifPresent(this.kafkaProducerConfig::setDefaultClientIdPrefix);
         this.kafkaProducerConfig.setCommonClientConfig(options.commonClientConfig());
         this.kafkaProducerConfig.setProducerConfig(options.producerConfig());
 
         this.kafkaConsumerConfig = new MessagingKafkaConsumerConfigProperties();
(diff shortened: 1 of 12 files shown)


sophokles73 (commented):
I would have expected to see
`AbstractKafkaConfigProperties.setDefaultClientIdPrefix(String)` being removed as well. Is there a particular reason why it isn't?

b-abel:
My understanding is that we still want to set the component name as a prefix _programmatically_ with this method, as we have always done.  But maybe we should change the name of the method to something like `AbstractKafkaConfigProperties.useClientIdPrefixIfNotConfigured(String)`. Then Spring would not use it as a setter for the property and you would see that it is not meant to be changed in the configuration. WDYT?
So, this PR only restores the behavior we had before and I understood that we will decide later about how to proceed with the prefix. #2906 will probably affect this too.

sophokles73:
FMPOV we do not need to restore a state that we do not need/want.

It would be great if you, @calohmn and @b-abel, could figure out if and how you want to set the Kafka client ID and then create a corresponding PR for implementing that desired state.

b-abel:
> FMPOV we do not need to restore a state that we do not need/want.

This sounds reasonable. However, I have a practical problem. I'm working on a PR for #2904, which unfortunately I can't really get cut small. In the last weeks, a lot has been done to the Kafka configuration, so I have had a bunch of merge conflicts. Currently, the PR for #2904 is blocked by this one and my interest is possible to get soon to a state on which I can provide it. 
@calohmn Any ideas for a quick solution here?

calohmn:
I would suggest merging this PR here as is, and considering further changes (also regarding the suggested method renaming) in the context of #2906. I wanted to have a go at that, but I think it can wait until #2904, to not cause further conflicts there.

adapter-base-quarkus/src/test/java/org/eclipse/hono/adapter/quarkus/QuarkusPropertyBindingTest.java
```diff
@@ -1,49 +0,0 @@
...
-import org.eclipse.hono.config.ProtocolAdapterProperties;
-import org.junit.jupiter.api.Test;
-
-import io.quarkus.test.junit.QuarkusTest;
-
-/**
- * Verifies that Quarkus binds properties from yaml files to configuration objects.
- */
-@QuarkusTest
-public class QuarkusPropertyBindingTest {
```
calohmn, line 31: Why should this test be deleted?

b-abel: Uh, that was a mistake.

calohmn (approved):
LGTM

Issue #2702, opened 2021-06-02
Support TTL for Telemetry


b-abel:
The [Telemetry API for Kafka](https://www.eclipse.org/hono/docs/api/telemetry-kafka/#produce-telemetry-data) defines a TTL record header to limit the _time-to-live_ of a message. The protocol adapters (at least HTTP and MQTT) already support that devices set a TTL for event messages. For Kafka-based messaging this should be supported for Telemetry messages as well because they are persisted.

b-abel:
FMPOV it would be a good first step to set the TTL header for each telemetry message at the protocol adapters. This allows Hono admins to configure the time period after which telemetry messages should be ignored. 
The alternative would be to completely rely on _retention time_, but that is much more heavyweight. Kafka does not delete individual records, but always a whole _log segment_. A newer record in a _log segment_ can therefore prevent the deletion of much older ones. The [interval of the Log Cleaner](https://kafka.apache.org/documentation/#brokerconfigs_log.retention.check.interval.ms) comes on top.
You might also want to configure a larger _retention time_, for example, to analyze the messages at times with lower throughput.

Setting a TTL by default on the protocol adapter would allow to let telemetry messages with QoS 0 expire earlier than those with QoS 1.
In a second step, we can allow devices to set the desired TTL per message, just like for _events_.

sophokles73:
Just to get it right: the TTL header information is intended to be read and acted upon by the consumer of the message, i.e. a downstream application, right? There are no technical means (apart from the kafka broker's retention policy) that prevent applications from processing messages having an expired TTL, or am I mistaken?

b-abel:
> Just to get it right: the TTL header information is intended to be read and acted upon by the consumer of the message, i.e. a downstream application, right? 

Right. The TTL header is defined by Hono's APIs and is not a mechanism by Kafka. 

> There are no technical means (apart from the kafka broker's retention policy) that prevent applications from processing messages having an expired TTL, or am I mistaken?

I am not aware of any other means. From Kafka's POV the consumer decides what records to read and how to handle (or ignore) them.

b-abel:
My proposal is:

1. it should be possible to define a _maximum_ TTL for telemetry messages at the tenant, which will always be set if there is not a lower individual value specified at the message. This is analogous to events where a `max-ttl` can be defined in the limits of the tenant.
[...]

PR #2971, opened 2021-11-24
Prep work for supporting ttl on all message types

sophokles73:
This is preparatory work for adding support for setting a TTL on downstream telemetry and command response messages (#2702).


kaniyan (approved):
LGTM

PR #2972, opened 2021-11-25
[#2904] Allow to configure Kafka clients individually

b-abel:
This addresses #2904.

The common properties block is unchanged. This is useful for configuring the bootstrap server, authentication and TLS. Instead of having specific blocks per client type (i.e. producer, consumer or admin client), there is now the possibility to create a separate configuration block for each individual Kafka client. It needs a name and is mapped to a client type.

The documentation of the properties supported by each Hono component is still missing here. It should probably best be put in the respective admin guides.

Example configuration for a protocol adapter:

~~~yaml
hono:
  kafka:
    commonClientConfig:
      bootstrap.servers: eclipse-hono-kafka-0.eclipse-hono-kafka-headless.hono:9094
      security.protocol: SASL_PLAIN
    event: # client name
      producerConfig:
        property: value # example Kafka producer config property
    telemetry:
      producerConfig:
        property: value
    command:
      pollTimeout: 200 # example for a Hono specific property
      consumerConfig:
        property: value
    command-response:
      producerConfig:
        property: value
    command-internal:
      adminClientConfig:
        property: value
    registry-notification:
      pollTimeout: 200
      consumerConfig:
        property: value

~~~


b-abel:
Discussion point:
So far the properties in configuration files were in camelCase. Both Spring and Quarkus advise using kebab-case instead. Changing existing configuration keys usually breaks compatibility. Since we are doing this here for the Kafka client configuration anyway, I wonder if we shouldn't take advantage of that by using kebab case for the new or changed properties instead of introducing new keys in camel case only for consistency. This affects: `commonClientConfig`, `producerConfig`, `consumerConfig`, `adminClientConfig`, `pollTimeout`.

calohmn:
Independent of the suggested discussion here, I think in this part
```
    command-internal:
      adminClientConfig:
```
the case should be consistent, i.e. using `commandInternal` if we stick to camel-case.

calohmn:
@b-abel Are there cases where using camel-case causes issues?

If there is no need for it, I wouldn't introduce new properties in kebab-case here, but rather keep it consistent. 
_If_ we want to switch to kebab-case, we could do this for Hono 2.0 and then do it across all properties. But again, I don't see much of a benefit in that, outweighing the effort needed for such a change.
@sophokles73 WDYT?

calohmn:
> That is probably also the reason why kebab case is used in Hono in many places for configuration prefixes.
In other words, this PR only contains configuration properties in camel case. Prefixes are in kebab case. This is perfectly consistent with what I find all over the Hono code base.

Yes, so this is referring to the notation used in the Spring `@ConfigurationProperties` annotation (e.g. `@ConfigurationProperties(prefix = "hono.command-router.amqp")`).
In my comment above, I was referring to the notation used in the YAML file. With Spring, this may also be in camel-case then - see also the example here:
https://github.com/eclipse/hono/blob/47b56d62c1cfa13fc45ddb0d0707c6d17f6679da/tests/src/test/resources/commandrouter/embedded-cache/application.yml#L15
So, the Spring-related code in this PR is already also supporting camel-case notation in the YAML file (because of the Spring relaxed binding).

For Quarkus, the corresponding annotations in the existing Hono code are used with camel-case prefixes, e.g. `@ConfigMapping(prefix = "hono.commandRouter.amqp")` or `@ConfigMapping(prefix = "hono.resourceLimits.prometheusBased", [...]`.
Therefore, I would suggest using the same notation in the `@ConfigMapping(prefix = "hono.kafka.[*]")` annotations used in this PR.

b-abel:
> Therefore, I would suggest using the same notation in the `@ConfigMapping(prefix = "hono.kafka.[*]")` annotations used in this PR.

Done.

services/command-router-quarkus/src/main/java/org/eclipse/hono/commandrouter/quarkus/Application.java
```diff
@@ -180,20 +183,19 @@ void setKafkaClientMetricsSupport(final KafkaMetricsOptions options) {
...
 
     @Inject
-    void setKafkaClientOptions(final KafkaClientOptions options) {
-        this.kafkaProducerConfig = new MessagingKafkaProducerConfigProperties();
-        this.kafkaProducerConfig.setCommonClientConfig(options.commonClientConfig());
-        this.kafkaProducerConfig.setProducerConfig(options.producerConfig());
+    void setKafkaClientOptions(
+            @ConfigMapping(prefix = "hono.kafka") final CommonKafkaClientOptions commonOptions,
+            @ConfigMapping(prefix = "hono.kafka.commandOut") final KafkaProducerOptions producerOptions,
+            @ConfigMapping(prefix = "hono.kafka.commandIn") final KafkaConsumerOptions consumerOptions,
```
calohmn, line 189: In the Command Router, a Kafka producer is used for "command_internal" and "command_response" messages.
The Kafka consumer is used for "command" messages.
Therefore, I think it would be better, to use the corresponding names in the prefixes (like also in the adapters), i.e.:
```
@ConfigMapping(prefix = "hono.kafka.commandInternal") final KafkaProducerOptions commandInternalProducerOptions,
@ConfigMapping(prefix = "hono.kafka.commandResponse") final KafkaProducerOptions commandResponseProducerOptions,
@ConfigMapping(prefix = "hono.kafka.command") final KafkaConsumerOptions consumerOptions,
```

b-abel: Are you suggesting that I refactor the class `KafkaBasedCommandConsumerFactoryImpl` to take two separate producers in this PR?

calohmn: Ok, this PR is big enough already. This should be done in a subsequent PR then.

calohmn (approved):
LGTM

PR #2973, opened 2021-11-25
[#2837] Publish notification if all devices of a tenant are deleted.

b-abel:
When deleting all devices of a tenant at once, the device registry implementations do not report which devices have been deleted. It is therefore impossible to send out individual change notifications for each device. 
This PR adds a new notification type `DevicesOfTenantDeletedNotification` to inform about this event as suggested by https://github.com/eclipse/hono/pull/2924#discussion_r754078670.

### DIFF
--- a/clients/notification-kafka/src/main/java/org/eclipse/hono/client/notification/NotificationTopicHelper.java
+++ b/clients/notification-kafka/src/main/java/org/eclipse/hono/client/notification/NotificationTopicHelper.java
@@ -15,6 +15,7 @@ package org.eclipse.hono.client.notification;
 
 import org.eclipse.hono.client.kafka.HonoTopic;
 import org.eclipse.hono.notification.AbstractNotification;
+import org.eclipse.hono.notification.deviceregistry.AllDevicesOfTenantDeletedNotification;
 import org.eclipse.hono.notification.deviceregistry.CredentialsChangeNotification;
 import org.eclipse.hono.notification.deviceregistry.DeviceChangeNotification;
 import org.eclipse.hono.notification.deviceregistry.TenantChangeNotification;
@@ -44,6 +45,8 @@ public final class NotificationTopicHelper {
             address = DeviceChangeNotification.ADDRESS;
         } else if (CredentialsChangeNotification.class.equals(notificationType)) {
             address = CredentialsChangeNotification.ADDRESS;
+        } else if (AllDevicesOfTenantDeletedNotification.class.equals(notificationType)) {
+            address = AllDevicesOfTenantDeletedNotification.ADDRESS;
         } else {
             throw new IllegalArgumentException("Unknown notification type " + notificationType.getName());
         }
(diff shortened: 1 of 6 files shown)


clients/notification/src/main/java/org/eclipse/hono/notification/deviceregistry/DevicesOfTenantDeletedNotification.java
```diff
@@ -0,0 +1,75 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Notification that informs that all devices of a tenant have been deleted.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class DevicesOfTenantDeletedNotification extends AbstractNotification {
```
kaniyan, line 31: IMHO naming to `AllDevicesOfTenantDeletedNotification` would be more suited as all the devices belonging to a tenant  are being deleted.

clients/notification/src/test/java/org/eclipse/hono/notification/deviceregistry/DevicesOfTenantDeletedNotificationTest.java
```diff
@@ -0,0 +1,106 @@
...
+        // When adding new properties to the data object, make sure not to break the existing API because messages might
+        // be persisted. For breaking changes add a new object mapper class instead.
+        final int expectedPropertiesCount = 4;
+        assertWithMessage("JSON contains unknown fields").that(json.size()).isEqualTo(expectedPropertiesCount);
+
+        assertThat(json.getString("type")).isEqualTo("devices-of-tenant-deleted-v1");
+        assertThat(json.getString("source")).isEqualTo("device-registry");
+        assertThat(json.getString("creation-time")).isEqualTo(CREATION_TIME);
+
+        assertThat(json.getString("tenant-id")).isNotNull();
```
kaniyan, line 83: How about using static constants instead of string literals as in the other tests? Example: NotificationConstants.JSON_FIELD_TENANT_ID instead of _tenant-id_?

kaniyan (changes requested):

kaniyan:
How about also extending the integration tests to check for this notification?

clients/notification/src/test/java/org/eclipse/hono/notification/deviceregistry/DevicesOfTenantDeletedNotificationTest.java
b-abel: That would contradict the intention of this test. There is already another test that checks that the JSON contains the values of the constants. This test is intended to detect when the constant is (accidentally) changed. This could lead to the fact that serialized (i.e. in Kafka stored) notifications cannot be de-serialized again.

b-abel:
> How about also extending the integration tests to check for this notification?

Integration tests for all notifications are missing. I created #2976 for that.

kaniyan (approved):
LGTM

PR #2974, opened 2021-11-25
Enable setting SSL ciphers in HotRod config

calohmn:
The [Infinispan HotRod client config](https://docs.jboss.org/infinispan/12.1/apidocs/org/infinispan/client/hotrod/configuration/package-summary.html) allows setting the ciphers to be used in the SSL handshake via the property `infinispan.client.hotrod.ssl_ciphers`.
Unfortunately, there are no corresponding getter/setter methods on the `ConfigurationProperties` object, so that the property isn't adopted in the Spring-based Command Router component.
For the Quarkus-based component, the property can also not be set.

As for the motivation for setting this property:
When using the Boring SSL library, on each connection attempt the following warning is logged:
```
BoringSSL doesn't allow to enable or disable TLSv1.3 ciphers explicitly. 
```
This can be quite misleading. Explicitly setting the ciphers prevents this.

### DIFF
--- a/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/InfinispanRemoteConfigurationOptions.java
+++ b/client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/InfinispanRemoteConfigurationOptions.java
@@ -29,152 +29,167 @@ import io.smallrye.config.WithDefault;
 public interface InfinispanRemoteConfigurationOptions {
 
     /**
-     * Gets the options.
+     * Gets the connection pool options.
      *
      * @return The options.
      */
     Map<String, String> connectionPool();
 
     /**
-     * Gets the options.
+     * Gets the SASL properties.
      *
-     * @return The options.
+     * @return The properties.
      */
     Map<String, String> saslProperties();
 
     /**
-     * Gets the options.
+     * Gets the cluster options.
      *
      * @return The options.
      */
     Map<String, String> cluster();
 
     /**
-     * Gets the value.
+     * Gets the list of remote servers as a string of the form <em>host1[:port][;host2[:port]]</em>.
      *
-     * @return The value.
+     * @return The servers.
      */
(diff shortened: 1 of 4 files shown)


client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/InfinispanRemoteConfigurationOptions.java
```diff
@@ -177,4 +177,11 @@
      */
     @WithDefault("false")
     boolean useSsl();
+
+    /**
+     * Gets the value.
```
sophokles73, line 182: IMHO this should be more specific to the purpose of the property

client-device-connection-infinispan/src/test/resources/application.yaml
```diff
@@ -23,3 +23,4 @@ hono:
       trustStoreType: "PKCS12"
       trustStorePassword: "trust-store-secret"
       useSsl: true
+      sslCiphers: "TLSv1.3 TLSv1.2"
```
sophokles73, line 26: This example value is at least misleading as its value contains TLS protocol versions but not TLS cipher suite names.

sophokles73 (changes requested):
can you please also add the getters/setters for configuring the TLS protocol versoin(s) to support?

client-device-connection-infinispan/src/main/java/org/eclipse/hono/deviceconnection/infinispan/client/InfinispanRemoteConfigurationOptions.java
calohmn: I've adapted the javadoc - also of the other method containing the same sentence.

sophokles73 (approved):
LGTM

sophokles73:
@calohmn how about adding a release notes entry and also applying this to the 1.10.x branch?

calohmn:
@sophokles73 Done.

PR #2977, opened 2021-11-26
Fix JavaDoc in NotificationTopicHelper

b-abel:
The method `NotificationTopicHelper.getTopicName()` is used on the sender side as well.


calohmn (approved):
LGTM

PR #2978, opened 2021-11-26
Fix handling of periodic commit failure while rebalance in progress

calohmn:
Retry of periodic commit should be done right after rebalance in that case.

Error handling was wrong here - the `RebalanceInProgressException` is actually returned in the `OffsetCommitCallback` parameter of `commitAsync`, not thrown directly in `commitAsync`.
This has become clear via logs like:
```
periodic commit failed: org.apache.kafka.common.errors.RebalanceInProgressException: Offset commit cannot be completed since the consumer is undergoing a rebalance for auto partition assignment. You can try completing the rebalance by calling poll() and then retry the operation. [client-id: cmd-router-consumer-4f4fae11-98ad-44e3-9fde-d9ef95130187]
```

This might actually resolve some integration test errors happening because a periodic offset commit was done too late - when the corresponding had already been deleted, this all leading to the consumer being blocked with internal offset commit retries.


kaniyan (approved):
LGTM

Issue #2953, opened 2021-11-17
Handle exception when creating Kafka clients


calohmn:
When deploying the Hono Helm chart, with the deployment of Kafka included, the protocol adapters may find the Kafka bootstrap server URLs to be non-resolvable initially, because it takes a while for Kafka to become ready.

Currently, the protocol adapters will shut down directly if encountering such a `ConfigException: No resolvable bootstrap urls given in bootstrap.servers` exception when creating the Kafka clients.
It would be better to retry the Kafka client creation, similar to the retries concerning the AMQP messaging network connection.

Log excerpt:
```
14:16:36.869 [main] INFO  o.a.k.c.admin.AdminClientConfig - AdminClientConfig values: 
	bootstrap.servers = [eclipse-hono-kafka-0.eclipse-hono-kafka-headless.hono:9094]
[...]

14:16:37.173 [main] WARN  org.apache.kafka.clients.ClientUtils - Couldn't resolve server eclipse-hono-kafka-0.eclipse-hono-kafka-headless.hono:9094 from bootstrap.servers as DNS resolution failed for eclipse-hono-kafka-0.eclipse-hono-kafka-headless.hono
14:16:37.198 [main] ERROR o.e.h.a.m.s.Application$$EnhancerBySpringCGLIB$$bb91eb70 - exception occurred during startup, shutting down ...
java.util.concurrent.ExecutionException: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'vertxBasedMqttProtocolAdapter' defined in org.eclipse.hono.adapter.mqtt.spring.Config: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.eclipse.hono.adapter.mqtt.impl.VertxBasedMqttProtocolAdapter]: Factory method 'vertxBasedMqttProtocolAdapter' threw exception; nested exception is org.apache.kafka.common.KafkaException: Failed to create new KafkaAdminClient
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2096)
	at org.eclipse.hono.service.spring.AbstractBaseApplication.run(AbstractBaseApplication.java:169)
	at org.springframework.boot.SpringApplication.callRunner(SpringApplication.java:782)
	at org.springframework.boot.SpringApplication.callRunners(SpringApplication.java:772)
[...]

PR #2980, opened 2021-11-29
[#2953] Use retries when trying to create Kafka consumer

calohmn:
The second part of the fix for #2953.


calohmn:
Also includes a fix for systems, where DNS resolution of host names like "some.hostname.that.is.invalid" (which is [supposed to be unresolvable](https://stackoverflow.com/questions/39282774/what-is-a-guaranteed-unresolvable-but-valid-url)) takes a long time.

kaniyan (approved):
LGTM

PR #2985, opened 2021-11-30
[#2825] Extend credentials management service to support auth-id template

kaniyan:
* Add _issuer-DN_ to credentials management specification.

*  Extend credentials management service implementations to generate
  the _auth-id_ by applying the _auth-id-template_ to the given _subject DN_.


services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/X509CertificateCredential.java
```diff
@@ -46,74 +48,121 @@
...
+     *                                  the given subject and issuer DNs are not a valid X.500 distinguished name or if
      *                                  secrets is empty.
      * @return The credentials.
      */
     @JsonCreator(mode = Mode.PROPERTIES)
     public static X509CertificateCredential fromProperties(
             @JsonProperty(value = RegistryManagementConstants.FIELD_PAYLOAD_CERT) final byte[] derEncodedX509Certificate,
-            @JsonProperty(value = RegistryManagementConstants.FIELD_AUTH_ID) final String distinguishedName,
+            @JsonProperty(value = RegistryManagementConstants.FIELD_AUTH_ID) final String authId,
+            @JsonProperty(value = RegistryManagementConstants.FIELD_IS_FOR_UPDATE) final boolean isForUpdate,
```
sophokles73, line 98: IMHO we do not need this parameter. Instead, we should be able to use the presence of the issuer DN as the indicator whether we need to derive the auth-id from the subject DN or not. If an issuer DN is given (or if a certificate is being uploaded), we apply the auth-id template to the given subject DN and store it in the auth-id field.
We would then also not need to keep the issuer DN in a persistent field, would we?

kaniyan: Initially I implemented as you suggested. But whenever an user updates credentials, the _auth-id_ should be always _subject-dn_ irrespective of _issuer-dn_ is provided or not. In that case, we should validate that the _auth-id_ is a valid  _subject-dn_ and format the _subject-dn_ as per RFC2253. The formatted _subject-dn_ is stored as the _auth-id_. Whenever we decode the credentials already stored in the database, we should not perform this validation as the _auth-id_ need not be always the _subject-dn_. In order to perform this, I needed that flag.

sophokles73: I have strong doubts that anybody of us will be able to understand any of this in two months from now. We (I, that is) already have a hard time understanding it now ;-)
Can we maybe keep the existing properties as they are and store the generated auth-id in a separate field instead, e.g. *generated-auth-id*, which we do not expose externally but which we use to find the matching credentials when an adapter invokes *get credentials*? A Management API client would, of course, still need to supply the issuer DN if he/she wants to have a *generated-auth-id* be created as well ...

kaniyan: > Can we maybe keep the existing properties as they are and store the generated auth-id in a separate field instead.


1. We can add a new database field to store the generated _auth-id_. During _get credentials_ operation we can first look into the generated auth-id field. If no value is there, we can look into the _auth-id_ field which always contain _subject-dn_
2. The _auth-id_ in the credentials object retrieved from _get credentials_ operation is compared with that of the _auth-id_ from the request during device authentication by adapters. In this case, we have to set the _generated-auth-id_ as the _auth-id_ in the credentials object.

sophokles73: Step one will be done by the registry implementation, I guess.
Step two is done on the adapter, right?
> In this case, we have to set the generated-auth-id as the auth-id in the credentials object.

You mean, the registry implementation needs to return the value of the *generated-auth-id* in the returned credentials object's *auth-id* property, right? If so, then we seem to agree.

FMPOV this would make things much easier to understand because on the adapter side, nothing changes at all and we do not need to introduce additional flags and/or fields in the Registry Management API (apart from the *issuer-dn* which we would need anyway). WDYT?

kaniyan: Initially I was hesitant to make any DDL changes as users need to add a new field with this Hono version. Now it seems to be a better choice as it improves the readability and helps to keep the concern separate to some extent.

sophokles73: Ok, does that mean that you want to proceed into that direction?

kaniyan: yes, I will implement this approach.

kaniyan:
@sophokles73 As we discussed, I added a new field `generated-auth-id` now to support the auth id template feature in JDBC and MongoDB registries. Also the _get credentials operation_ has been extended to support this feature. I have pushed a consolidated commit for it.

services/device-registry-base/src/main/java/org/eclipse/hono/deviceregistry/service/credentials/AbstractCredentialsService.java
```diff
@@ -190,8 +192,20 @@ protected final CacheDirective getCacheDirective(final String type, final long m
...
 
+    private static CredentialsResult<JsonObject> overrideAuthIdWithGeneratedAuthIdIfExists(
+            final CredentialsResult<JsonObject> credentialsResult) {
+        if (!credentialsResult.isError()) {
+            final JsonObject credential = credentialsResult.getPayload();
+            Optional.ofNullable(credential)
+                    .map(c -> c.getString(RegistryManagementConstants.FIELD_TYPE))
+                    .filter(RegistryManagementConstants.SECRETS_TYPE_X509_CERT::equals)
+                    .map(ok -> credential.getString(RegistryManagementConstants.FIELD_GENERATED_AUTH_ID))
+                    .ifPresent(id -> credential.put(RegistryManagementConstants.FIELD_AUTH_ID, id));
```
sophokles73, line 203: Since this is the Credentials service, I would assume that all these constants should be defined in CredentialsConstants, right?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/X509CertificateCredential.java
```diff
@@ -198,4 +258,45 @@ public final X509CertificateCredential setSecrets(final List<X509CertificateSecr
...
+
+    /**
+     * Gets the authentication identifier generated by applying the <em>auth-id-template</em> from the tenant's trust
+     * anchor to the client certificate's subject DN.
+     *
+     * @return The generated authentication identifier.
+     */
+    @JsonProperty(value = RegistryManagementConstants.FIELD_GENERATED_AUTH_ID)
+    public final String getGeneratedAuthId() {
+        return generatedAuthId;
```
sophokles73, line 290: this means that we are exposing this property to external clients, doesn't it?

kaniyan: My intention was to include the _generatedAuthId_ for serialization to persist in the database. By setting the `@JsonProperty`, the property is exposed to the external clients and I overlooked it.  I use a different approach now and pushed a commit for it. Would you mind taking a look again.

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/X509CertificateCredential.java
```diff
@@ -198,4 +261,68 @@ public final X509CertificateCredential setSecrets(final List<X509CertificateSecr
...
+            @Override
+            @JsonGetter(value = RegistryManagementConstants.FIELD_GENERATED_AUTH_ID)
+            protected String getGeneratedAuthId() {
+                return super.getGeneratedAuthId();
+            }
+        };
+
+        copyCommonAttributesTo(credential);
+        return credential;
+    }
```
sophokles73, line 295: I guess that this works but, pardon my French, it looks like a hack. Wouldn't it be easier to understand and also much cleaner from a design perspective to introduce two subclasses instead, one for handling credentials with a generated ID and the other for a non-generated ID?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/X509CertificateCredential.java
```diff
@@ -46,91 +49,165 @@
...
-     * @param distinguishedName The DN to use as the authentication identifier.
+     * @param subjectDN The subject DN of the client certificate to be used as the authentication identifier.
+     * @param issuerDN The issuer DN of the client certificate. 
+     *                 <p> It is used to retrieve the <em>auth-id-template</em> from the tenant's
+     *                 trust anchor to generate the authentication identifier.
+     * @param generatedAuthId the authentication identifier generated by applying the <em>auth-id-template</em>
+     *                        from the tenant's trust anchor to the client certificate's subject DN.
      * @param secrets The credential's secret(s).
-     * @throws NullPointerException if certificate bytes and any of distinguished name and secrets are {@code null}.
+     * @throws NullPointerException if certificate bytes, subject DN and secrets are {@code null}.
```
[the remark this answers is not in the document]
kaniyan: The distinguished name in the old wording is ambiguous as both _issuer-dn_ and _subject-dn_ are distinguished names.
 How about _if certificate bytes and any of  subject DN and secrets are {@code null}_?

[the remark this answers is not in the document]
sophokles73: it is not about distinguished name vs subject DN. it is about the expression built from the three parameters.
> if certificate bytes and any of distinguished name and secrets are {@code null}

is `true` if either certificate bytes AND distinguished name are `null`, OR if certificate bytes AND secrets are `null`

but

> if certificate bytes, subject DN and secrets are {@code null}.

is only `true` if all three parameters are `null`, or am I mistaken?

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/X509CertificateCredential.java
kaniyan: I noticed that anonymous class approach is employed in several places in Hono. I chose that approach than my initial one. Good that I still have a backup of `X509CertificateCredentialWithGeneratedAuthId` ;-)

services/device-registry-base/src/main/java/org/eclipse/hono/service/management/credentials/X509CertificateCredential.java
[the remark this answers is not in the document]
kaniyan: > if certificate bytes, subject DN and secrets are {@code null}.

We agree on the same. The above one is not correct.

In my previous comment, I proposed the below one.  I just replaced  `distinguished name` with `subject DN` in the original description as we now have two distinguished names in the list of paramteres(`final String subjectDN,  final String issuerDN`).

> if certificate bytes and any of subject DN and secrets are {@code null}

Another suggestions is as below:
> if certificate bytes and either subject DN or secrets are {@code null}

Or am I missing something?

services/device-registry-mongodb/src/main/java/org/eclipse/hono/deviceregistry/mongodb/utils/MongoDbDocumentBuilder.java
```diff
@@ -129,10 +148,57 @@ public MongoDbDocumentBuilder withType(final String type) {
...
+     * </pre>
+     *
+     * @param authId The auth id.
+     * @return a reference to this for fluent use.
+     */
+    private MongoDbDocumentBuilder withGeneratedAuthId(final String authId) {
+        final var credentialsArraySpec = document.getJsonObject(CredentialsDto.FIELD_CREDENTIALS, new JsonObject());
+        final var elementMatchSpec = credentialsArraySpec.getJsonObject(MONGODB_OPERATOR_ELEM_MATCH, new JsonObject());
+
+        elementMatchSpec.put(MONGODB_OPERATOR_OR, new JsonArray(List.of(
```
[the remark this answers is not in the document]
kaniyan: I think a List is needed there as there are two JsonObjects to be added to the JsonArray.  If we like to avoid using List, then it can be rewritten as below. One way of populating JsonArray is to pass a list to the constructor and another is to invoke add methods.

```
elementMatchSpec.put(MONGODB_OPERATOR_OR, new JsonArray().add(new JsonObject()).add(new JsonObject())
```

core/src/main/java/org/eclipse/hono/util/CredentialsConstants.java
```diff
@@ -39,6 +39,11 @@
      * The name of the field that contains the authentication identifier.
      */
     public static final String FIELD_AUTH_ID                     = "auth-id";
+    /**
+     * The name of the field that contains the generated authentication identifier
+     * by applying the template to the subject DN.
+     */
+    public static final String FIELD_GENERATED_AUTH_ID           = "generated-auth-id";
```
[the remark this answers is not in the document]
kaniyan: Yes it is still being used in `CredentialsServiceImpl`.

PR #2986, opened 2021-12-02
[#2702] Improve Kafka client class hierarchy

sophokles73:
This is preparatory work for #2702

The AbstractKafkaBasedMessageSender's send* methods have been changed to
require an OpenTracing Span to be passed in which will not be finished
by the send method anymore. The subclasses and test cases have been
adapted accordingly. With this change the classes are better in line
with the approach used in Hono in gerneral, i.e. the component creating
a Span should also be responsible for finishing it.

Some (now) obsolete send* methods have been removed as well.

### DIFF
--- a/clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedInternalCommandSender.java
+++ b/clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedInternalCommandSender.java
@@ -29,6 +29,7 @@ import org.eclipse.hono.client.kafka.producer.KafkaProducerFactory;
 import org.eclipse.hono.client.kafka.producer.MessagingKafkaProducerConfigProperties;
 import org.eclipse.hono.util.MessageHelper;
 
+import io.opentracing.Span;
 import io.opentracing.Tracer;
 import io.vertx.core.Future;
 import io.vertx.core.buffer.Buffer;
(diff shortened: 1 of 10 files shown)


clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedInternalCommandSender.java
```diff
@@ -69,17 +70,28 @@ public KafkaBasedInternalCommandSender(final KafkaProducerFactory<String, Buffer
...
+                command.getDeviceId(),
+                commandContext.getTracingContext());
+
         return sendAndWaitForOutcome(
-                getInternalCommandTopic(adapterInstanceId),
+                topicName,
                 command.getTenant(),
                 command.getDeviceId(),
                 command.getPayload(),
                 getHeaders((KafkaBasedCommand) command),
```
kaniyan, line 86: In this PR, the `KafkaBasedCommandResponseSender.getHeaders(..)` now return `Map<String, Object>` instead of `List<KafkaHeader>`. The same we can also apply in this class too to. This is just a minor side note.

kaniyan (changes requested):

clients/command-kafka/src/main/java/org/eclipse/hono/client/command/kafka/KafkaBasedInternalCommandSender.java
sophokles73: We could, but in this case we are only forwarding an existing command message to the internal topic and we can therefore take advantage of simply copying the existing headers from the incoming message to the outgoing message, i.e. no need to create a Map and re-encode the outgoing Kafka headers again.

sophokles73:
@kaniyan I have pushed corresponding changes

kaniyan (approved):
LGTM

PR #2987, opened 2021-12-02
[#2904] Use separate Kafka producer configs in Command Router

calohmn:
For #2904, addendum to #2972.

### DIFF
--- a/services/command-router-quarkus/src/main/java/org/eclipse/hono/commandrouter/quarkus/Application.java
+++ b/services/command-router-quarkus/src/main/java/org/eclipse/hono/commandrouter/quarkus/Application.java
@@ -128,7 +128,8 @@ public class Application extends AbstractServiceApplication {
     private RequestResponseClientConfigProperties deviceRegistrationClientConfig;
     private RequestResponseClientConfigProperties tenantClientConfig;
     private KafkaClientMetricsSupport kafkaClientMetricsSupport;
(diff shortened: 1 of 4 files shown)


calohmn:
@b-abel Could you have a look?

services/command-router-quarkus/src/main/java/org/eclipse/hono/commandrouter/quarkus/Application.java
```diff
@@ -288,10 +293,10 @@ private CommandRouterService commandRouterService() {
             final CommandTargetMapper commandTargetMapper) {
 
         final MessagingClientProvider<CommandConsumerFactory> commandConsumerFactoryProvider = new MessagingClientProvider<>();
-        if (kafkaProducerConfig.isConfigured() && kafkaConsumerConfig.isConfigured()) {
+        if (commandInternalKafkaProducerConfig.isConfigured() && kafkaConsumerConfig.isConfigured()) {
```
b-abel, line 296: Shouldn't this be `commandResponseKafkaProducerConfig` instead of `commandInternalKafkaProducerConfig`? The latter one is checked again a couple of lines below...
In general, I think the checks for Kafka configuration could be improved but that is independent of this PR.

calohmn: I've changed the condition now to also include `commandResponseKafkaProducerConfig`.

PR #2988, opened 2021-12-03
[#2906] Improve Kafka client ID creation

### LINKED ISSUE #2906: Use Kafka client identifiers including pod/host name

calohmn:
This fixed #2906.
Default client ID contains the Kubernetes Pod name and container ID now.

### DIFF
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
@@ -18,8 +18,10 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.kafka.clients.CommonClientConfigs;
+import org.eclipse.hono.util.KubernetesContainerUtil;
 import org.eclipse.hono.util.Strings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -34,16 +36,34 @@ public abstract class AbstractKafkaConfigProperties {
      */
     public static final String PROPERTY_BOOTSTRAP_SERVERS = "bootstrap.servers";
 
+    private static final AtomicInteger ID_COUNTER = new AtomicInteger();
+
     protected final Logger log = LoggerFactory.getLogger(getClass());
 
     /**
-     * Client id prefix to be applied if no {@code client.id} property is set in the common
-     * or specific client config properties.
+     * If set, this ID along with the incremented idCounter value will be used as client ID suffix.
      */
-    protected String defaultClientIdPrefix;
+    private String componentUId;
 
     private Map<String, String> commonClientConfig;
     private Map<String, String> specificClientConfig;
+
+    /**
+     * Creates a new instance.
+     */
+    protected AbstractKafkaConfigProperties() {
+        this.componentUId = getComponentUIdFromEnv();
+    }
+
+    private String getComponentUIdFromEnv() {
+        final String k8sContainerId = KubernetesContainerUtil.getContainerId();
+        if (k8sContainerId != null && k8sContainerId.length() >= 12) {
+            // running in Kubernetes: use HOSTNAME env var containing the pod name
+            final String podName = System.getenv("HOSTNAME");
+            return String.format("%s_%s", podName, k8sContainerId.substring(0, 12));
+        }
+        return null;
+    }
 
     /**
      * Sets the common Kafka client config to be used.
@@ -67,17 +87,20 @@ public abstract class AbstractKafkaConfigProperties {
     }
 
     /**
-     * Sets the prefix for the client ID that is passed to the Kafka server to allow application specific server-side
-     * request logging.
+     * Sets the component unique ID to be included in the client ID along with a counter value.
      * <p>
-     * If the common or specific client config already contains a value for key {@code client.id}, that one
-     * will be used and the parameter here will be ignored.
+     * This overrides the unique ID determined automatically in case this application is running in Kubernetes
+     * (consisting of Pod/Container ID then).
+     * <p>
+     * Setting {@code null} here means that a UUID will be used in the client ID instead of component unique ID and
+     * counter value.
+     * <p>
+     * This method is mainly intended for usage in non-Kubernetes environments and unit tests.
      *
-     * @param clientId The client ID prefix to set.
-     * @throws NullPointerException if the client ID is {@code null}.
+     * @param componentUId The component ID to use.
      */
-    public final void setDefaultClientIdPrefix(final String clientId) {
-        this.defaultClientIdPrefix = Objects.requireNonNull(clientId);
+    public final void overrideComponentUidUsedForClientId(final String componentUId) {
+        this.componentUId = componentUId;
     }
 
     /**
@@ -93,9 +116,10 @@ public abstract class AbstractKafkaConfigProperties {
     /**
      * Gets the Kafka client configuration. This is the common configuration on which the client specific configuration
      * has been applied.
-     *
-     * An already set client id property or alternatively the value set via {@link #setDefaultClientIdPrefix(String)}
-     * will be used as a prefix, to which the given clientName and a UUID will be added.
+     * <p>
+     * It is ensured that the returned map contains a unique {@code client.id}. The client ID will be created from the
+     * given client name, followed by a unique ID (containing component identifiers if running in Kubernetes).
+     * An already set {@code client.id} property value will be used as prefix for the client ID.
      *
      * @param clientName A name for the client to include in the added {@code client.id} property.
      * @return a copy of the client configuration with the applied properties.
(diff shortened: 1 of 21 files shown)


calohmn:
@b-abel Could you have a look?

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
```diff
@@ -67,17 +91,30 @@ protected final void setSpecificClientConfig(final Map<String, String> specificC
...
-     * Sets the prefix for the client ID that is passed to the Kafka server to allow application specific server-side
-     * request logging.
+     * Sets the component name to be included in the client ID along with a UUID in case no component unique ID
+     * has been set via {@link #setComponentUid(String)} and also no unique ID could be determined automatically
+     * because this application is not running in Kubernetes.
+     *
+     * @param componentName The component name to use.
+     * @throws NullPointerException if the component name is {@code null}.
+     */
+    public final void setComponentName(final String componentName) {
```
b-abel, line 101: My understanding is that this method should be called in code. It is not meant to map configuration properties (e.g. from Spring Boot). Maybe renaming it to a name that does not start with "set" would help here.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
```diff
@@ -67,17 +91,30 @@ protected final void setSpecificClientConfig(final Map<String, String> specificC
...
+     * Setting {@code null} here means that the name set via {@link #setComponentName(String)} along with a UUID
+     * will be used in the client ID.
      *
-     * @param clientId The client ID prefix to set.
-     * @throws NullPointerException if the client ID is {@code null}.
+     * @param componentUId The component ID to use.
      */
-    public final void setDefaultClientIdPrefix(final String clientId) {
-        this.defaultClientIdPrefix = Objects.requireNonNull(clientId);
+    public final void setComponentUid(final String componentUId) {
```
b-abel, line 116: Same as with `setComponentName()`.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/KafkaAdminClientConfigProperties.java
```diff
@@ -60,9 +60,13 @@ public final void setAdminClientConfig(final Map<String, String> adminConfig) {
...
      * Gets the Kafka admin client configuration. This is the result of applying the admin client configuration on the
-     * common configuration. The returned map will contain a property {@code client.id} which will be set to a unique
-     * value containing the already set client id or alternatively the value set via
-     * {@link #setDefaultClientIdPrefix(String)}, the given adminClientName and a newly created UUID.
+     * common configuration.
+     * <p>
+     * It is ensured that the returned map contains a unique {@code client.id}. The client ID will be created from the
+     * given client name, followed by a component identifier with included or added unique ID (either containing pod name
+     * and container ID if running in Kubernetes, or including component unique ID or name set via
+     * {@link #setComponentUid(String)}) or {@link #setComponentName(String)})).
```
b-abel, line 68: That's hard to understand in prose. Perhaps examples would help, or a schematic diagram listing the individual elements and their manifestations. 
Alternatively, one could perhaps refrain from explaining the whole algorithm in detail in the JavaDoc. If it is complex and you can never be sure if the documentation is up to date and covers all edge cases anyway, I might rather look directly into the implementation than work through the text first, then look at the implementation and then think about possible differences...

calohmn: I would opt for the alternative then: I think there's no need to explain the algorithm in detail here - it should be enough to document the fact that a unique id is used as a suffix.
I've therefore removed the part in parentheses above.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
calohmn: It wouldn't make much sense for a user to set this property via a configuration file, yes. 
But on the other hand, it wouldn't necessarily be a problem and we wouldn't put this property in the list of supported configuration properties in the documentation anyway. (And it can only be set using Spring which be want to move away from.)
So, I would rather keep this standard naming style.

b-abel: You are technically right. On the other hand, it is not only about Spring. The naming suggests that the class has a property `componentName` that can be changed in the configuration. FMPOV this naming has caused a lot of confusion in our discussions in the last couple of weeks. If I recall correctly, @sophokles73 has explicitly asked for more clarity.

calohmn: > The naming suggests that the class has a property `componentName` that can be changed in the configuration.

I see no problem per se in the `AbstractKafkaConfigProperties` having different kinds of properties.
There are the properties in the map given to the Kafka client constructor (no individual property setters there).
There may also be additional properties for configuring the Hono Kafka client classes (like `pollTimeout`).
And now we configure properties used for the client ID creation.

We could make the meaning of the new methods more explicit by renaming the methods to `setComponentNameForClientId` / `setComponentUIdForClientId`. WDYT?

b-abel: > And now we configure properties used for the client ID creation.

FMPOV the confusion arises from the fact that `componentName` is not a _configuration property_. It is not intended to be configured externally by providing a property in a configuration file, or an environment variable, or a Java system property. It is a method that you use to provide a value to this class. Usually, this is prefixed by "set". But this is not a random Java class but a specific one, namely one that implements configuration properties. And a method starting by "set" on this kind of class suggests that there is a property with that name that is intended to be configured with one of the mechanisms mentioned above. 

The fact that we are having this discussion is proof that it is not just another innocent setter. It was exactly this naming issue that first led to us accidentally setting the client ID by configuration in the IoT Packages Project, then even led to this bug being considered a feature. Then even effort was invested to implement this anti-feature for Quarkus as well. @sophokles73 WDYT?

sophokles73: I tend to agree with @b-abel here. 
Why do we need to set the component name on the properties object in the first place? Where is the component name being used/required?

b-abel:
The release notes entry should mention that the property `hono.kafka.defaultClientIdPrefix` needs to be removed from existing configurations.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
calohmn: > Why do we need to set the component name on the properties object in the first place? Where is the component name being used/required?

In the code before this PR, the client ID created in `AbstractKafkaConfigProperties` contained a "(default)ClientIdPrefix". As value for that prefix, the component name (e.g. "Hono AMQP Adapter") was set.
This means, when looking at client IDs in the Kafka broker logs (or e.g. Confluent UI), it was being made clear in which Hono component the client was operating.

With #2906, the intention was to improve on this by using the Kubernetes Pod/container IDs instead.

Now, with this PR, I've kept the "(default)ClientIdPrefix" but renamed it to "componentName(ForClientId)".
As stated in the "setComponentNameForClientId" javadoc, this name is used for the created client ID if the component isn't running in Kubernetes, i.e. if the Kubernetes Pod/container IDs can't be used in the client ID.

So, in practice, "setComponentNameForClientId" could also be removed if we want to ignore Docker (ie. integration test) or non-K8s scenarios. That would also be OK with me.

> And a method starting by "set" on this kind of class suggests that there is a property with that name that is intended to be configured with one of the mechanisms mentioned above.

For the Quarkus components this is not the case since we have separate `[*]Options` classes there (e.g. `KafkaConsumerOptions`).

calohmn: > So, in practice, "setComponentNameForClientId" could also be removed if we want to ignore Docker (ie. integration test) or non-K8s scenarios. That would also be OK with me.

I've removed it now.
`setComponentUidForClientId` has been renamed to `overrideComponentUidUsedForClientId` - it is only used in unit tests here.

calohmn:
@b-abel I've added a corresponding release notes entry.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
b-abel: > setComponentUidForClientId has been renamed to overrideComponentUidUsedForClientId - it is only used in unit tests here.

I like the name. But I'm not sure what the purpose of this method is now. Is it intended to be invoked by productive code in the future? Or only for unit tests?

calohmn: I've extended the javadoc now:
`This method is mainly intended for usage in non-Kubernetes environments and unit tests.`

PR #2991, opened 2021-12-06
Some notification API fixes/improvements

calohmn:
Move Kafka notification classes into new kafka package.
Add toString() methods for notification classes.
Fix missing AllDevicesOfTenantDeletedNotification key.


clients/notification/src/main/java/org/eclipse/hono/notification/deviceregistry/CredentialsChangeNotification.java
```diff
@@ -92,4 +92,11 @@ public final String getType() {
         return TYPE;
     }
 
+    @Override
+    public String toString() {
```
b-abel, line 96: What is the use case? Logging? Would it then make sense to add the creation time in the toString method as well? At least on the receiver side, this information isn't obvious.

calohmn: Yes, for logging, mainly on the sender side.
I've now added the creation time and also added error logging in `KafkaBasedNotificationSender`.

b-abel:
LGTM

PR #2993, opened 2021-12-08
[#2953] Use retries when trying to create Kafka producer

calohmn:
This is the last PR for #2953.

### DIFF
--- a/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
+++ b/clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
@@ -31,11 +31,6 @@ import org.slf4j.LoggerFactory;
  */
 public abstract class AbstractKafkaConfigProperties {
 
-    /**
-     * The name of the Kafka configuration property containing the servers to bootstrap from.
-     */
-    public static final String PROPERTY_BOOTSTRAP_SERVERS = "bootstrap.servers";
-
     private static final AtomicInteger ID_COUNTER = new AtomicInteger();
 
     protected final Logger log = LoggerFactory.getLogger(getClass());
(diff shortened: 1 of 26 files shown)


clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/AbstractKafkaConfigProperties.java
```diff
@@ -106,13 +101,30 @@ public final void overrideComponentUidUsedForClientId(final String componentUId)
     /**
      * Checks if a configuration has been set.
      *
-     * @return {@code true} if the {@value #PROPERTY_BOOTSTRAP_SERVERS} property has been configured with a non-null
-     *         value.
+     * @return {@code true} if the {@value CommonClientConfigs#BOOTSTRAP_SERVERS_CONFIG} property has been configured
```
sophokles73, line 104: When I try to create JavaDoc on my machine as part of the Maven build, this breaks the build because the JavaDoc generator cannot resolve the reference. Does this work on your machine?

sophokles73: My bad, I had only added the `CommonClientConfigs` but did not adapt the constant name ... seems to work.

clients/kafka-common/src/main/java/org/eclipse/hono/client/kafka/producer/KafkaProducerFactory.java
```diff
@@ -46,9 +49,37 @@
      * @param producerName The name to identify the producer.
      * @param config The Kafka configuration with which the producer is to be created.
      * @return an existing or new producer.
+     * @throws org.apache.kafka.common.KafkaException If creating the producer failed because of an invalid config or
+     *             because bootstrap servers could not be resolved.
```
sophokles73, line 53: do we still need this method?

calohmn: The `getOrCreateProducer` method is still used.
I haven't replaced all occurrences of this method with the new `getOrCreateProducerWithRetries` method.
The `[..]WithRetries` method is now used in places where it is called in a `start()` method. This is the main idea here, that when the component is just starting up, the other cluster components might also not be ready yet so that bootstrap URLs aren't resolvable yet.

For scenarios where the URLs stop being resolvable while the Hono component is already running, the corresponding method invocations will need to use a specific, configurable `retriesTimeout` value (instead of using `UNLIMITED_RETRIES_DURATION`), for example in the `KafkaBasedEventSender` class where sending an event should not be retried indefinitely.
Therefore, more changes would be needed for that. I would rather not do this as part of this PR.

sophokles73 (approved):
LGTM

PR #2996, opened 2021-12-08
[#2984] Clean Caches on Device Registry Notifications

### LINKED ISSUE #2984: Clear tenant/device related cache data after tenant/device got deleted/disabled
Relies on #2837:
The new notification API should be used to invalidate tenant/device-specific cache data used in the `DeviceRegistrationClient`/`TenantClient` when the corresponding tenant/device got deleted or disabled.

b-abel:
This fixes #2984.
The clients for the device registry APIs may register receivers for notifications of the device registry that are used to remove
objects from the response cache if the corresponding data has changed.

I have created new classes for the cache keys. The reason is that before, some elements of the keys were concatenated (like "${tenantId}-${deviceId}"). This makes it difficult to find e.g. all keys for a tenant.
Previously, the name of the API operation in question was part of the keys. I left it out because none of the clients provide multiple operations. If that ever changes, it's easy to add the operation to the key. @sophokles73 Is this ok for you? Or am I missing something?

Sometimes data that is not part of the key is needed to find affected cache entries. For example, the device ID was not stored with the credentials. 
In order to find the cache entries by means of the Device ID, I could have created an additional map in which I keep track of which credentials belong to which Device ID. The problem is that entries are evicted from the cache and then would have remained in the map. Instead, I created the class AnnotatedCacheKey, which allows storing additional attributes to a key without them having to become part of the key.

### DIFF
--- a/adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
+++ b/adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
@@ -63,6 +63,8 @@ import org.eclipse.hono.client.kafka.producer.CachingKafkaProducerFactory;
 import org.eclipse.hono.client.kafka.producer.KafkaProducerFactory;
 import org.eclipse.hono.client.kafka.producer.KafkaProducerOptions;
 import org.eclipse.hono.client.kafka.producer.MessagingKafkaProducerConfigProperties;
+import org.eclipse.hono.client.notification.kafka.KafkaBasedNotificationReceiver;
+import org.eclipse.hono.client.notification.kafka.NotificationKafkaConsumerConfigProperties;
 import org.eclipse.hono.client.registry.CredentialsClient;
 import org.eclipse.hono.client.registry.DeviceRegistrationClient;
 import org.eclipse.hono.client.registry.TenantClient;
@@ -79,6 +81,8 @@ import org.eclipse.hono.config.ClientConfigProperties;
 import org.eclipse.hono.config.ProtocolAdapterProperties;
 import org.eclipse.hono.config.quarkus.ClientOptions;
 import org.eclipse.hono.config.quarkus.RequestResponseClientOptions;
+import org.eclipse.hono.notification.NoOpNotificationReceiver;
+import org.eclipse.hono.notification.NotificationReceiver;
 import org.eclipse.hono.service.cache.Caches;
 import org.eclipse.hono.service.quarkus.AbstractServiceApplication;
 import org.eclipse.hono.util.CredentialsObject;
@@ -139,6 +143,7 @@ public abstract class AbstractProtocolAdapterApplication<C extends ProtocolAdapt
     private MessagingKafkaProducerConfigProperties kafkaCommandResponseConfig;
     private MessagingKafkaConsumerConfigProperties kafkaCommandConfig;
     private KafkaAdminClientConfigProperties kafkaCommandInternalConfig;
+    private NotificationKafkaConsumerConfigProperties kafkaNotificationConfig;
 
     private Cache<Object, TenantResult<TenantObject>> tenantResponseCache;
     private Cache<Object, RegistrationResult> registrationResponseCache;
@@ -232,13 +237,15 @@ public abstract class AbstractProtocolAdapterApplication<C extends ProtocolAdapt
             @ConfigMapping(prefix = "hono.kafka.event") final KafkaProducerOptions eventOptions,
             @ConfigMapping(prefix = "hono.kafka.commandResponse") final KafkaProducerOptions commandResponseOptions,
             @ConfigMapping(prefix = "hono.kafka.command") final KafkaConsumerOptions commandOptions,
-            @ConfigMapping(prefix = "hono.kafka.commandInternal") final KafkaAdminClientOptions commandInternalOptions) {
+            @ConfigMapping(prefix = "hono.kafka.commandInternal") final KafkaAdminClientOptions commandInternalOptions,
+            @ConfigMapping(prefix = "hono.kafka.notification") final KafkaConsumerOptions notificationOptions) {
 
         kafkaTelemetryConfig = new MessagingKafkaProducerConfigProperties(commonOptions, telemetryOptions);
         kafkaEventConfig = new MessagingKafkaProducerConfigProperties(commonOptions, eventOptions);
         kafkaCommandResponseConfig = new MessagingKafkaProducerConfigProperties(commonOptions, commandResponseOptions);
         kafkaCommandConfig = new MessagingKafkaConsumerConfigProperties(commonOptions, commandOptions);
         kafkaCommandInternalConfig = new KafkaAdminClientConfigProperties(commonOptions, commandInternalOptions);
+        kafkaNotificationConfig = new NotificationKafkaConsumerConfigProperties(commonOptions, notificationOptions);
     }
 
     @Override
@@ -271,13 +278,14 @@ public abstract class AbstractProtocolAdapterApplication<C extends ProtocolAdapt
 
         Objects.requireNonNull(adapter);
 
-        final DeviceRegistrationClient registrationClient = registrationClient();
+        final NotificationReceiver notificationReceiver = notificationReceiver();
+        final DeviceRegistrationClient registrationClient = registrationClient(notificationReceiver);
 
         final MessagingClientProvider<TelemetrySender> telemetrySenderProvider = new MessagingClientProvider<>();
         final MessagingClientProvider<EventSender> eventSenderProvider = new MessagingClientProvider<>();
         final MessagingClientProvider<CommandResponseSender> commandResponseSenderProvider = new MessagingClientProvider<>();
         final KafkaClientMetricsSupport kafkaClientMetricsSupport = kafkaClientMetricsSupport(kafkaMetricsOptions);
-        final var tenantClient = tenantClient();
+        final var tenantClient = tenantClient(notificationReceiver);
 
         if (kafkaEventConfig.isConfigured()) {
             LOG.info("Kafka client configuration present, adding Kafka messaging clients");
@@ -343,7 +351,7 @@ public abstract class AbstractProtocolAdapterApplication<C extends ProtocolAdapt
         adapter.setMessagingClientProviders(messagingClientProviders);
         Optional.ofNullable(connectionEventProducer())
             .ifPresent(adapter::setConnectionEventProducer);
-        adapter.setCredentialsClient(credentialsClient());
+        adapter.setCredentialsClient(credentialsClient(notificationReceiver));
         adapter.setHealthCheckServer(healthCheckServer);
         adapter.setRegistrationClient(registrationClient);
         adapter.setResourceLimitChecks(prometheusResourceLimitChecks(resourceLimitChecksConfig, tenantClient));
@@ -378,12 +386,14 @@ public abstract class AbstractProtocolAdapterApplication<C extends ProtocolAdapt
     /**
      * Creates a new client for Hono's Tenant service.
      *
+     * @param notificationReceiver The notification receiver to use.
      * @return The client.
      */
-    protected TenantClient tenantClient() {
+    protected TenantClient tenantClient(final NotificationReceiver notificationReceiver) {
         return new ProtonBasedTenantClient(
                 HonoConnection.newConnection(vertx, tenantClientConfig, tracer),
                 messageSamplerFactory,
+                notificationReceiver,
                 tenantResponseCache());
     }
 
@@ -397,12 +407,14 @@ public abstract class AbstractProtocolAdapterApplication<C extends ProtocolAdapt
     /**
      * Creates a new client for Hono's Device Registration service.
      *
+     * @param notificationReceiver The notification receiver to use.
      * @return The client.
      */
-    protected DeviceRegistrationClient registrationClient() {
+    protected DeviceRegistrationClient registrationClient(final NotificationReceiver notificationReceiver) {
         return new ProtonBasedDeviceRegistrationClient(
                 HonoConnection.newConnection(vertx, deviceRegistrationClientConfig, tracer),
                 messageSamplerFactory,
+                notificationReceiver,
                 registrationResponseCache());
     }
 
@@ -416,12 +428,14 @@ public abstract class AbstractProtocolAdapterApplication<C extends ProtocolAdapt
     /**
      * Creates a new client for Hono's Credentials service.
      *
+     * @param notificationReceiver The notification receiver to use.
      * @return The client.
      */
-    protected CredentialsClient credentialsClient() {
+    protected CredentialsClient credentialsClient(final NotificationReceiver notificationReceiver) {
         return new ProtonBasedCredentialsClient(
                 HonoConnection.newConnection(vertx, credentialsClientConfig, tracer),
                 messageSamplerFactory,
+                notificationReceiver,
                 credentialsResponseCache());
     }
 
@@ -570,4 +584,19 @@ public abstract class AbstractProtocolAdapterApplication<C extends ProtocolAdapt
             return NoopKafkaClientMetricsSupport.INSTANCE;
         }
     }
+
+    /**
+     * Creates the notification receiver.
+     *
+     * @return The bean instance.
+     */
+    public NotificationReceiver notificationReceiver() {
+        if (kafkaNotificationConfig.isConfigured()) {
+            return new KafkaBasedNotificationReceiver(vertx, kafkaNotificationConfig);
+        } else {
+            // TODO provide AMQP based notification receiver
+            return new NoOpNotificationReceiver();
+        }
+    }
+
 }
--- a/clients/amqp-common/src/main/java/org/eclipse/hono/client/amqp/AbstractRequestResponseServiceClient.java
+++ b/clients/amqp-common/src/main/java/org/eclipse/hono/client/amqp/AbstractRequestResponseServiceClient.java
@@ -19,6 +19,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 import org.apache.qpid.proton.amqp.messaging.ApplicationProperties;
 import org.apache.qpid.proton.message.Message;
@@ -38,6 +40,7 @@ import com.github.benmanes.caffeine.cache.Cache;
 import io.opentracing.Span;
 import io.opentracing.tag.Tags;
 import io.vertx.core.Future;
+import io.vertx.core.Handler;
 import io.vertx.core.buffer.Buffer;
 
 /**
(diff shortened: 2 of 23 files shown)


clients/amqp-common/src/main/java/org/eclipse/hono/client/amqp/AbstractRequestResponseServiceClient.java
```diff
@@ -299,6 +301,44 @@ protected final void addToCache(final Object key, final R response) {
...
+     */
+    protected final void removeFromCacheByPattern(final Predicate<Object> keyPredicate) {
+        if (isCachingEnabled()) {
+            Objects.requireNonNull(keyPredicate);
+
+            final var matchingKeys = responseCache.asMap()
+                    .keySet()
+                    .stream()
+                    .filter(keyPredicate)
+                    .collect(Collectors.toSet());
```
sophokles73, line 335: I am a little concerned about the runtime behavior of this when the cache has a lot of entries. Apart from blocking the event loop runtime seems to scale linearly with the cache size. This is probably not an issue with the tenant cache but for devices and/or credentials with dozens of thousands of entries I am not so sure ...

b-abel: @calohmn @kaniyan Do you have any ideas on this?

calohmn: Here is an idea that also involves additional data structures (but I guess is a bit different than envisioned above).
The following description is for the device registration client:

We could keep an extra cache with received device-change notifications, with an eviction timespan that is equivalent to the eviction timespan of the device registration cache.
This device-change cache could have tenant+device id as key, and the notification value (including timestamp) as value.

When a device registration client `assertRegistration` method gets called, there is first the check for a corresponding entry in the device registration cache. 
If a value is found there, 2 extra lookups on the device-change cache are done (with keys "tenant+device id" and "tenant+gateway id"). If there is a cache hit on that, cache timestamps are to be compared: If the device-change cache entry timestamp is newer than the device registration cache entry timestamp, a new device registry request has to be done to get the current device registration data and then update the device registration cache entry.

It would be good to be able to delete the device-change cache entry once a corresponding device registration cache entry has been updated.
However, this gets more complicated when there was a device-change notification regarding a gateway entry, because this may affect multiple device registration cache entries. Maybe if we add some info in the notification message about whether the device is a gateway, we can apply some optimizations here.
To react on *tenant*-change notifications with this approach, we could possible store entries in the device-change cache here as well - with the device id part of the cache key being null.

For the credentials data cache, we could think of a similar approach.

b-abel: Thank you @calohmn for your proposal! I'll have a closer look on it later on.

Two ideas of mine that are independent of this:

1. It might be possible to achieve more efficient scaling with the existing solution by means of sorting.
2. Alternatively, I wonder if it wouldn't be sufficient to take the operation off the event loop (perhaps combined with a parallel stream). That might be sufficient, assuming that these notifications are relatively rare events that are also not particularly time-critical. WDYT?

sophokles73: I wonder if we can simply assume that we do not need to invalidate all entries belonging to a tenant from the registration and credentials caches. In [this comment](https://github.com/eclipse/hono/issues/2984#issuecomment-987703829) @calohmn argued that we can assume that clients would not use values from the registration nor credentials caches if the tenant does not exist or is disabled. If that is true then I guess we can simply wait for the entries to be evicted from the cache instead of actively removing them when a tenant gets deleted, or am I missing something?

b-abel: > I wonder if we can simply assume that we do not need to invalidate all entries belonging to a tenant from the registration and credentials caches.

My understanding of the code is that negative results are also cached. This means that a tenant would not be available after creation until the cache expires.
In addition, there is the case that all devices of a tenant are deleted. Then the tenant still exists and is enabled. For this case, there is a separate notification, where we still have the problem to find the affected devices and credentials only by means of the tenant ID.

b-abel:
@calohmn I think we should first answer my question above about whether it's not enough to take the deletion from the caches down from the event loop. 
So is optimizing a rare, non-time-critical execution path worth the extra code complexity?

If so, I don't see any fundamental problems with your proposal.

> Maybe if we add some info in the notification message about whether the device is a gateway, we can apply some optimizations here.

This small optimization would mean that for each change to a device, additional database accesses would be necessary to obtain the information whether it is a gateway. It's probably not worth it.
I would just skip removing from the "device change cache" for gateway devices (i.e. if gateway id is not `null`). Those entries are removed by expiry.
For the credentials caches, this problem does not exist anyway.

clients/amqp-common/src/main/java/org/eclipse/hono/client/amqp/AbstractRequestResponseServiceClient.java
calohmn: > > I wonder if we can simply assume that we do not need to invalidate all entries belonging to a tenant from the registration and credentials caches.

> My understanding of the code is that negative results are also cached. This means that a tenant would not be available after creation until the cache expires.

But the tenant creation notification would invalidate the cache entry here, wouldn't it? Or which scenario were you thinking of here?

> In addition, there is the case that all devices of a tenant are deleted. 

I guess in practise, this operation would most probably be followed by a deletion of the tenant.  So if we assume that the tenant client is invoked first, the device/credentials client invocations wouldn't happen here anyway.

calohmn:
> @calohmn I think we should first answer my question above about whether it's not enough to take the deletion from the caches down from the event loop.
So is optimizing a rare, non-time-critical execution path worth the extra code complexity?

Good question. I would go one step further and ask if we really need to do the deletion from the caches on a worker thread, or for which cache size we should start doing it. Since this involves just a simple iteration with a simple check on each key, I think this shouldn't be an issue doing it on the event loop thread in _most_ cases, maybe using a worker thread only for cache sizes > 50k  or 100k entries.
So, yes, I think we could probably avoid usage of the extra caches here for now, maybe adding it later as an optimization.

clients/amqp-common/src/main/java/org/eclipse/hono/client/amqp/AbstractRequestResponseServiceClient.java
sophokles73: > My understanding of the code is that negative results are also cached.

That shouldn't be the case. What makes you believe so?

b-abel: Nvermind. I just found the place that prevents negative results from being cached (for the record: `ProtonBasedDeviceRegistrationClient.getResult()` sets the `cacheDirective` property of `RegistrationResult` to `null` and then `AbstractRequestResponseServiceClient.addToCache()` filters for status codes).

b-abel: I will change the PR so that notifications about the creation of tenants, devices, and credentials are not processed because negative results are not cached.
Also, I will no longer clean up the Registration Assertions and Credentials caches when deleting the tenant. This way the cleanup will take place less often.

b-abel:
I suggest performing the cache cleanup off of the event loop. Also, since the verticles in a protocol adapter share a cache instance, one could consider pulling the cleanup out of the individual clients and doing it for each cache centrally per adapter instance. Possibly using a worker verticle. @sophokles73 @calohmn @kaniyan WDYT?

sophokles73:
@b-abel let's start with simply using a worker thread for doing the housekeeping and then add more PRs for optimizing the code.
I also see a potential issue with multiple (say dozens) deletion notifications arriving in rapid succession which might lead to a situation where dozens of housekeeping jobs run in parallel on a cache instance. But again, let's optimize in subsequent PRs ...

clients/amqp-common/src/main/java/org/eclipse/hono/client/amqp/AbstractRequestResponseServiceClient.java
b-abel: > > In addition, there is the case that all devices of a tenant are deleted.
> 
> I guess in practise, this operation would most probably be followed by a deletion of the tenant. 

@sophokles73 Do you agree that we can assume that users will delete the tenant shortly after they invoked the operation that removes all devices and therefore skipping the cleanup of registration and credentials caches?

sophokles73: I tend to agree. The only use case that comes to mind where it wouldn't is that you might want to re-use a (test-)tenant for a new series of tests but want to start with a fresh set of devices, thus removing all registered devices before running the tests. However, I guess that this notification will be sent very seldomly in comparison to the device/credentials updated/removed notifications ...

b-abel: @sophokles73  Makes sense. For the sake of consistency and because we already agreed to use a worker thread I tent to clean up registration and credentials caches on that notification.

b-abel:
I added the discussed changes.

adapter-base-quarkus/src/main/java/org/eclipse/hono/adapter/quarkus/AbstractProtocolAdapterApplication.java
```diff
@@ -570,4 +584,19 @@ public KafkaClientMetricsSupport kafkaClientMetricsSupport(final KafkaMetricsOpt
...
+     * Creates the notification receiver.
+     *
+     * @return The bean instance.
+     */
+    public NotificationReceiver notificationReceiver() {
+        if (kafkaNotificationConfig.isConfigured()) {
+            return new KafkaBasedNotificationReceiver(vertx, kafkaNotificationConfig);
+        } else {
+            // TODO provide AMQP based notification receiver
+            return new NoOpNotificationReceiver();
```
calohmn, line 598: @b-abel Do you want to add usage of `ProtonBasedNotificationReceiver` here now (corresponding PR adding that class has been merged)? 
Otherwise we have to make sure that this gets added in a follow-up PR before the Hono 1.11.0 release.

b-abel: I'll check how much change this would bring. If it is trivial, I could add it here.

b-abel: @calohmn I would prefer to add this in a follow-up PR.

clients/registry-amqp/src/main/java/org/eclipse/hono/client/registry/amqp/AnnotatedCacheKey.java
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * A cache key that supports adding metadata attributes to be used to filter cache entries.
+ * <p>
+ * The metadata attributes are not taken into account by the {@link #equals(Object)} and {@link #hashCode()} methods and
+ * are therefore not part of the identity checks by the cache implementation.
+ *
+ * @param <T> The type of the cache key.
+ */
+public final class AnnotatedCacheKey<T> {
```
sophokles73, line 29: this class doesn't seem to be specific to AMQP, is it? If not, then IMHO we should put it into the generic registry module or maybe even the client-common module instead.

clients/notification/src/main/java/org/eclipse/hono/notification/NoOpNotificationReceiver.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.eclipse.hono.notification;
+
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+
+/**
+ * A no-op implementation for the notification receiver.
+ */
+public class NoOpNotificationReceiver implements NotificationReceiver {
```
sophokles73, line 22: final ?

clients/registry-amqp/src/main/java/org/eclipse/hono/client/registry/amqp/ProtonBasedCredentialsClient.java
```diff
@@ -51,24 +56,32 @@
 /**
  * A vertx-proton based client of Hono's Credentials service.
  *
+ * If a response cache has been provided, a notification receiver can be used to receive notifications about changes in
```
sophokles73, line 59: please insert a `<p>` above this line

clients/registry-amqp/src/main/java/org/eclipse/hono/client/registry/amqp/ProtonBasedCredentialsClient.java
```diff
@@ -51,24 +56,32 @@
...
+    private static final String ATTRIBUTE_KEY_DEVICE_ID = "device-id";
+
+    private final NotificationReceiver notificationReceiver;
 
     /**
      * Creates a new client for a connection.
      *
      * @param connection The connection to the Credentials service.
      * @param samplerFactory The factory for creating samplers for tracing AMQP messages being sent.
+     * @param notificationReceiver The client to receive notifications from the device registry.
```
sophokles73, line 77: please add some text describing that the given receiver will be started/stopped as part of this component's lifecycle

clients/registry-amqp/src/main/java/org/eclipse/hono/client/registry/amqp/ProtonBasedDeviceRegistrationClient.java
```diff
@@ -54,23 +57,29 @@
 /**
  * A vertx-proton based client of Hono's Device Registration service.
  *
+ * If a response cache has been provided, a notification receiver can be used to receive notifications about changes in
```
sophokles73, line 60: `<p>`

clients/registry-amqp/src/main/java/org/eclipse/hono/client/registry/amqp/ProtonBasedDeviceRegistrationClient.java
```diff
@@ -54,23 +57,29 @@
...
 
     private static final Logger LOG = LoggerFactory.getLogger(ProtonBasedDeviceRegistrationClient.class);
+    private final NotificationReceiver notificationReceiver;
 
     /**
      * Creates a new client for a connection.
      *
      * @param connection The connection to the Device Registration service.
      * @param samplerFactory The factory for creating samplers for tracing AMQP messages being sent.
+     * @param notificationReceiver The client to receive notifications from the device registry.
```
sophokles73, line 75: see above

clients/registry-amqp/src/main/java/org/eclipse/hono/client/registry/amqp/ProtonBasedTenantClient.java
```diff
@@ -59,29 +61,37 @@
 /**
  * A vertx-proton based client of Hono's Tenant service.
  *
+ * If a response cache has been provided, a notification receiver can be used to receive notifications about changes in
```
sophokles73, line 64: `<p>`

clients/registry-amqp/src/main/java/org/eclipse/hono/client/registry/amqp/ProtonBasedTenantClient.java
```diff
@@ -59,29 +61,37 @@
...
+    private static final String ATTRIBUTE_KEY_TENANT_ID = "tenant-id";
     private final Map<Object, Future<TenantResult<TenantObject>>> pendingRequests = new HashMap<>();
+    private final NotificationReceiver notificationReceiver;
 
     /**
      * Creates a new client for a connection.
      *
      * @param connection The connection to the service.
      * @param samplerFactory The factory for creating samplers for tracing AMQP messages being sent.
+     * @param notificationReceiver The client to receive notifications from the device registry.
```
sophokles73, line 81: see above

clients/registry-amqp/src/main/java/org/eclipse/hono/client/registry/amqp/ProtonBasedTenantClient.java
```diff
@@ -210,10 +220,59 @@ protected String getKey(final String tenantId) {
...
+        if (isCachingEnabled()) {
+            Optional.ofNullable(tenantResult.getPayload())
+                    .ifPresent(p -> responseCacheKey.putAttribute(ATTRIBUTE_KEY_TENANT_ID, p.getTenantId()));
+
+            addToCache(responseCacheKey, tenantResult);
+        }
+    }
+
+    private void removeResultFromCache(final String tenantId) {
+        removeFromCacheByPattern(key -> ((AnnotatedCacheKey<?>) key)
```
sophokles73, line 242: what about the cache entries where the key is the tenant ID?

clients/registry-amqp/src/main/java/org/eclipse/hono/client/registry/amqp/ProtonBasedTenantClient.java
```diff
@@ -59,29 +61,37 @@
...
      * @param connection The connection to the service.
      * @param samplerFactory The factory for creating samplers for tracing AMQP messages being sent.
+     * @param notificationReceiver The client to receive notifications from the device registry.
      * @param responseCache The cache to use for service responses or {@code null} if responses should not be cached.
      * @throws NullPointerException if any of the parameters other than the response cache are {@code null}.
      */
     public ProtonBasedTenantClient(
             final HonoConnection connection,
             final SendMessageSampler.Factory samplerFactory,
+            final NotificationReceiver notificationReceiver,
```
sophokles73, line 88: it feels a little inconvenient that you always need to specify a receiver. Can't we overload the constructor and simply assume a Noop receiver by default?

sophokles73: I can imagine running the adapters with a custom registry that does not send notifications at all ...

clients/registry-amqp/src/main/java/org/eclipse/hono/client/registry/amqp/ProtonBasedTenantClient.java
b-abel: The cache entries where the key is the tenant ID are also removed by this because every cache key contains this metadata. I added comments to make it clearer. We need to search by the pattern anyways to find the cache entries with the principal as the key.

clients/registry-amqp/src/main/java/org/eclipse/hono/client/registry/amqp/ProtonBasedTenantClient.java
sophokles73: I do not quite see why the response cache parameter should be `null` in that case. Caching the results returned from the registry still makes sense, doesn't it? In fact, that is what we are already doing ...

b-abel: > In fact, that is what we are already doing ...

My understanding was that we add the feature to stop doing that :smile: But you are right, we should not couple the caching to the notifications on that level. I added the parameter.

sophokles73 (approved):
LGTM

sophokles73:
@b-abel any squashing before we merge?

b-abel:
> @b-abel any squashing before we merge?

I squashed it. Thank you.

PR #3001, opened 2021-12-14
Add blockwise status lifetime to configuration.

boaks:
In order to setup the timeout according the network assumptions (e.g.
NAT timeout) this value is added to the configuration.


sophokles73 (approved):
LGTM

thanks for contributing :+1:

